PR #30846, opened 2025-08-26
[CodeHleath] replace intermediate std::vectors with base::span

cmazakas:
Resolves https://github.com/brave/brave-browser/issues/48725


cdesouza-chromium (approved):
LGTM

components/brave_wallet/common/encoding_utils.cc
```diff
@@ -84,15 +86,15 @@ std::optional<std::string> Ss58Address::Encode() {
...
 
   auto hash_prefix = base::byte_span_from_cstring(kSs58HashPrefix);
   auto hash = Blake2bHash<64>(
       {hash_prefix, base::span(buff).first(offset + kSs58PublicKeySize)});
 
   output_span_writer.Write(base::span(hash).first(kSs58HashChecksumSize));
 
-  return Base58Encode(buff);
+  return Base58Encode(buff.first(output_span_writer.num_written()));
 }
```
supermassive: Let's define `buff` as 
```
auto buff = base::span(storage).first(offset + kSs58PublicKeySize +
                            kSs58HashChecksumSize);
```
and don't change these lines

cdesouza-chromium: Using the stack allocated array here is better than than the vector though.

cmazakas: Anton was alluding to just resizing the span up-front because the routines will never write beyond its initial resizing. This is because the length of the prefix can be 1 or 2 bytes.

This is done, as well!

supermassive (commented):
lgtm

supermassive (approved):

PR #30854, opened 2025-08-26
fix(wallet): Wrap Domain Text in Connections Panel

Douglashdaniel:
## Description 

Fixes a bug where the `Domain` text did not wrap on the `Connections` tab

Resolves <https://github.com/brave/brave-browser/issues/48748>

## Test Plan:
1. Visit a website with a really long domain name
2. Open the `Wallet` panel and navigate to the `Connections` tab
3. The `Domain` text should wrap and should not expand off the screen

Before:

Screenshot 58

After:

Screenshot 57


kdenhartog (approved):
LGTM, thanks for the quick fix

PR #30856, opened 2025-08-26
[cr140][Android] Fixed color of search your downloads textbox

AlexeyBarabash:
The color `@color/toolbar_text_box_bg_color` was used which does not depend on Brave's color accent.

Resolves https://github.com/brave/brave-browser/issues/48702

Screenshots:
Flag|Light|Dark
--|--|--
image|image|image
image|image|image


browser/download/internal/android/java/res/drawable/download_search_bar_background.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright (c) 2025 The Brave Authors. All rights reserved.
+     This Source Code Form is subject to the terms of the Mozilla Public
+     License, v. 2.0. If a copy of the MPL was not distributed with this file,
+     You can obtain one at https://mozilla.org/MPL/2.0/.
+-->
+<shape xmlns:android="http://schemas.android.com/apk/res/android"
+  android:shape="rectangle">
+  <solid android:color="@color/toolbar_text_box_bg_color" />
```
samartnik: Will dynamic color work with this one?

AlexeyBarabash: No, this is broken now, screenshot with flag `Dynamic Colors` enabled
image

AlexeyBarabash: @samartnik 
I have updated the PR by using modified background image only if `Dynamic colors` flag is disabled.

I see that actual dynamic colors are applied only when the app theme is Dark. On the Light app color scheme the color of the search box has an orange tint. This is related to color override at this line https://github.com/brave/brave-core/blob/master/android/java/res/values/brave_colors.xml#L38

samartnik (approved):
lgtm, but personally I believe we should follow upstream and use baseline colors for non-dynamic state instead of patching java code for this
cc @aguscruiz

aguscruiz:
I agree, let's follow upstream. Our designs attempt to follow upstream most of the times

AlexeyBarabash:
> lgtm, but personally I believe we should follow upstream and use baseline colors for non-dynamic state instead of patching java code for this

Then, the way would be to remove all overrides at
https://github.com/brave/brave-core/blob/master/android/java/res/values/brave_colors.xml
at section `<!-- Overriding chromium theme colors to brave theme colors -->` and fix all other broken places, but this seems beyond of this PR.

kjozwiak:
Verification **PASSED** on `Samsung Galaxy Tab S8 Ultra` running `Android 15` using the following build(s):

```
Brave | 1.84.18 Chromium: 140.0.7339.41 (Official Build) canary (64-bit)
--- | ---
Revision | 513e5ed4a0fec4b672a6f07ea0bda23d2f2c55f5
```

Using the STR/Cases outlined via https://github.com/brave/brave-browser/issues/48702#issue-3353973958, ensured that the `red` colouring has been removed from `brave://downloads` as per the following:

**`Example`** | **`Example`**
---------------|------------------
Screenshot_20250902_135600_Brave - Nightly | Screenshot_20250902_140909_Brave - Nightly

Created https://github.com/brave/brave-browser/issues/48904 as a follow up. However, this also affects `Chrome` with `C140` so assuming we'll get an upstream fix from a `C140` bump.

PR #30857, opened 2025-08-26
Add more EmailAliasesService unit tests

arthuredelstein:
Added tests:

- RequestAuthentication_ServerErrorObject_NoVerificationToken
- RequestAuthentication_VerificationTokenEmpty
- RequestSession_ParseableWrongShape_RetriesAuthenticating
- RequestSession_VerifiedTrueMissingAuthToken_Retries
- RequestSession_NoResponseBody_RetriesAuthenticating
- Cancel_IgnoresLateCallback
- AddObserver_InitialUnauthenticatedStateDelivered
- Reauth_CancelsInFlightInit
- ReauthDuringPolling_CancelsOldFlow
- Reauth_DoubleClick_CancelsEarlierFlow
- Reauth_NewTokenUsedForSubsequentPolls
- VerifyResult_IsRateLimitedByTwoSeconds
- VerifyResult_StopsAfterMaxDuration

Resolves https://github.com/brave/brave-browser/issues/48696

### DIFF
--- a/components/email_aliases/email_aliases_service_unittest.cc
+++ b/components/email_aliases/email_aliases_service_unittest.cc
@@ -210,10 +210,15 @@ TEST_F(EmailAliasesServiceTest, RequestSession_Success) {
                          "\"service\":\"email-aliases\"}"},
                         AuthenticationStatus::kAuthenticated);
   EXPECT_EQ(service_->GetAuthTokenForTesting(), "auth456");
+  EXPECT_FALSE(service_->IsPollingSessionForTesting());
 }
 
 TEST_F(EmailAliasesServiceTest, RequestSession_InvalidJson) {
-  RunRequestSessionTest({"not a json"}, AuthenticationStatus::kAuthenticating);
+  RunRequestSessionTest({"not a json",
+                         "{\"authToken\":\"auth456\", \"verified\":true, "
+                         "\"service\":\"email-aliases\"}"},
+                        AuthenticationStatus::kAuthenticated);
+  EXPECT_FALSE(service_->IsPollingSessionForTesting());
 }
 
 TEST_F(EmailAliasesServiceTest, RequestSession_RetryOnMissingAuthToken) {
@@ -229,6 +234,7 @@ TEST_F(EmailAliasesServiceTest, RequestSession_RetryOnMissingAuthToken) {
   // unauthenticated, authenticating, authenticated
   EXPECT_EQ(observer_->last_state,
             email_aliases::mojom::AuthenticationStatus::kAuthenticated);
+  EXPECT_FALSE(service_->IsPollingSessionForTesting());
 }
 
 TEST_F(EmailAliasesServiceTest, RequestSession_StopsOnVerificationFailed) {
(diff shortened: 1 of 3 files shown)


components/email_aliases/email_aliases_service_unittest.cc
```diff
@@ -225,6 +314,116 @@ TEST_F(EmailAliasesServiceTest,
...
+                     &called1));
+
+  StartAuthWithInitToken("x@example.com", "tokenX");
+
+  CompleteResultWithSuccess("authX");
+  EXPECT_EQ(service_->GetAuthTokenForTesting(), "authX");
+  EXPECT_FALSE(called1);
+}
+
+TEST_F(EmailAliasesServiceTest, Reauth_NewTokenUsedForSubsequentPolls) {
```
cdesouza-chromium, line 432: Please add another test to verify that we don't keep polling after max polling time.

components/email_aliases/email_aliases_service_unittest.cc
```diff
@@ -225,6 +314,116 @@ TEST_F(EmailAliasesServiceTest,
...
+  test_url_loader_factory_.AddResponse(
+      EmailAliasesService::GetAccountsServiceVerifyResultURL().spec(),
+      "{\"authToken\":null, \"verified\":false, "
+      "\"service\":\"email-aliases\"}");
+
+  // Reauth with a new token and complete successfully.
+  StartAuthWithInitToken("z2@example.com", "t2");
+  CompleteResultWithSuccess("auth2");
+}
+
```
cdesouza-chromium, line 444: Actually looking at this closely, I realise we don't have a proper polling test. We need a test that checks that this class keeps polling, and then stops polling once the max time is exceeded.

components/email_aliases/email_aliases_service_unittest.cc
arthuredelstein: We already have a test in master that does that: https://github.com/brave/brave-core/blob/6d6fcb2d54d08cc1e82a50af1b19753ef1f37edf/components/email_aliases/email_aliases_service_unittest.cc#L318-L333

components/email_aliases/email_aliases_service_unittest.cc
arthuredelstein: please see https://github.com/brave/brave-core/pull/30857#discussion_r2310629092

components/email_aliases/email_aliases_service_unittest.cc
cdesouza-chromium: How is this test managing to forward time?

components/email_aliases/email_aliases_service_unittest.cc
cdesouza-chromium: That test doesn't seem to be doing polling one two three... then it became successful, unless I'm misreading it.

components/email_aliases/email_aliases_service_unittest.cc
arthuredelstein: In my hands, calling `StartAuthAndCaptureRequests` moves through time for these delayed tasks. Thus we end up counting 902 total requests.

components/email_aliases/email_aliases_service_unittest.cc
arthuredelstein: See https://github.com/brave/brave-core/pull/30857#discussion_r2310725508

kdenhartog:
no concerns from me

boocmp:
These tests no more relevant, because email aliases service uses the brave account service for authenthication.

PR #30858, opened 2025-08-26
Move all usages of `crypto::RSAPrivateKey` to `crypto::keypair::PrivateKey`

DJAndries:
Resolves https://github.com/brave/brave-browser/issues/48392

### DIFF
--- a/browser/extensions/brave_crx_generation_browsertest.cc
+++ b/browser/extensions/brave_crx_generation_browsertest.cc
@@ -43,7 +43,7 @@ std::vector<uint8_t> GetPublicKeyHash(const base::FilePath& pem_path) {
     return {};
 
   auto private_key = crypto::keypair::PrivateKey::FromPrivateKeyInfo(
-      std::vector<uint8_t>(private_key_bytes.begin(), private_key_bytes.end()));
+      base::as_byte_span(private_key_bytes));
   if (!private_key)
     return {};
 
--- a/components/web_discovery/browser/background_credential_helper.cc
+++ b/components/web_discovery/browser/background_credential_helper.cc
@@ -10,6 +10,7 @@
 #include "base/base64.h"
 #include "base/check.h"
 #include "base/containers/span_rust.h"
+#include "base/containers/to_vector.h"
 #include "base/logging.h"
 #include "base/threading/thread_restrictions.h"
 #include "brave/components/web_discovery/browser/anonymous_credentials/lib.rs.h"
@@ -50,9 +51,8 @@ class BackgroundCredentialHelperImpl : public BackgroundCredentialHelper {
   }
 
   crypto::keypair::PrivateKey GenerateAndSetRSAKey() override {
-    auto key = web_discovery::GenerateRSAKey();
-    rsa_private_key_ = key;
-    return key;
+    rsa_private_key_ = web_discovery::GenerateRSAKey();
+    return *rsa_private_key_;
   }
 
   void SetRSAKey(crypto::keypair::PrivateKey rsa_private_key) override {
@@ -70,10 +70,9 @@ class BackgroundCredentialHelperImpl : public BackgroundCredentialHelper {
 
     auto signature = RSASign(*rsa_private_key_, join_result.join_request);
 
-    return StartJoinInitialization(
-        base::Base64Encode(join_result.join_request),
-        std::vector<uint8_t>(join_result.gsk.begin(), join_result.gsk.end()),
-        signature);
+    return StartJoinInitialization(base::Base64Encode(join_result.join_request),
+                                   base::ToVector(join_result.gsk),
+                                   std::move(signature));
   }
 
   std::optional<std::string> FinishJoin(
@@ -135,7 +134,7 @@ class BackgroundCredentialHelperImpl : public BackgroundCredentialHelper {
       return std::nullopt;
     }
     auto sig_data = sig_res->unwrap();
-    return std::vector<uint8_t>(sig_data.begin(), sig_data.end());
+    return base::ToVector(sig_data);
   }
 
  private:


browser/extensions/brave_crx_generation_browsertest.cc
```diff
@@ -42,13 +42,12 @@ std::vector<uint8_t> GetPublicKeyHash(const base::FilePath& pem_path) {
   if (!Extension::ParsePEMKeyBytes(private_key_contents, &private_key_bytes))
     return {};
 
-  auto private_key = crypto::RSAPrivateKey::CreateFromPrivateKeyInfo(
+  auto private_key = crypto::keypair::PrivateKey::FromPrivateKeyInfo(
```
cdesouza-chromium, line 45: We can avoid the extra copy into a vector by reading directly from the string:

```cxx
auto private_key = crypto::keypair::PrivateKey::FromPrivateKeyInfo(base::as_byte_span(private_key_bytes));
```

components/web_discovery/browser/background_credential_helper.cc
```diff
@@ -49,20 +49,17 @@ class BackgroundCredentialHelperImpl : public BackgroundCredentialHelper {
...
-  std::unique_ptr<crypto::RSAPrivateKey> GenerateAndSetRSAKey() override {
-    rsa_private_key_ = web_discovery::GenerateRSAKey();
-    if (!rsa_private_key_) {
-      return nullptr;
-    }
-    return rsa_private_key_->Copy();
+  crypto::keypair::PrivateKey GenerateAndSetRSAKey() override {
+    auto key = web_discovery::GenerateRSAKey();
+    rsa_private_key_ = key;
+    return key;
```
cdesouza-chromium: nit: a bit more concise:

```cxx
rsa_private_key_ = web_discovery::GenerateRSAKey();
return rsa_private_key_;
```

components/web_discovery/browser/background_credential_helper.cc
```diff
@@ -71,17 +68,12 @@ class BackgroundCredentialHelperImpl : public BackgroundCredentialHelper {
...
-      VLOG(1) << "RSA signature failed";
-      return std::nullopt;
-    }
+    auto signature = RSASign(*rsa_private_key_, join_result.join_request);
 
     return StartJoinInitialization(
         base::Base64Encode(join_result.join_request),
         std::vector<uint8_t>(join_result.gsk.begin(), join_result.gsk.end()),
-        *signature);
+        signature);
```
cdesouza-chromium: `std::move(signature)`

components/web_discovery/browser/background_credential_helper.cc
```diff
@@ -71,17 +68,12 @@ class BackgroundCredentialHelperImpl : public BackgroundCredentialHelper {
...
-
-    if (!signature) {
-      VLOG(1) << "RSA signature failed";
-      return std::nullopt;
-    }
+    auto signature = RSASign(*rsa_private_key_, join_result.join_request);
 
     return StartJoinInitialization(
         base::Base64Encode(join_result.join_request),
         std::vector<uint8_t>(join_result.gsk.begin(), join_result.gsk.end()),
```
cdesouza-chromium: `base::ToVector(base::as_byte_span(join_result.gsk))`

components/web_discovery/browser/rsa.cc
```diff
@@ -9,43 +9,34 @@
...
-  std::vector<uint8_t> encoded_private_key;
-
-  if (!private_key.ExportPrivateKey(&encoded_private_key) ||
-      !private_key.ExportPublicKey(&encoded_public_key)) {
-    return nullptr;
-  }
+    const crypto::keypair::PrivateKey& private_key) {
+  std::vector<uint8_t> encoded_private_key = private_key.ToPrivateKeyInfo();
+  std::vector<uint8_t> encoded_public_key =
+      private_key.ToSubjectPublicKeyInfo();
```
cdesouza-chromium, line 39: It looks like we don't need `EncodedRSAKeyPair` to begin with, and we should be using `crypto::keypair::PrivateKey` directly. Am I missing something else here?

components/web_discovery/browser/rsa.cc
```diff
@@ -59,33 +50,23 @@ std::unique_ptr<ImportedRSAKey> ImportRSAKey(
...
-
-  auto info = std::make_unique<ImportedRSAKey>();
-
-  info->private_key = std::move(key_pair);
-  info->public_key_b64 = base::Base64Encode(encoded_public_key);
+  std::vector<uint8_t> encoded_public_key = key_pair->ToSubjectPublicKeyInfo();
 
-  return info;
+  return std::make_unique<ImportedRSAKey>(
+      std::move(*key_pair), base::Base64Encode(encoded_public_key));
```
cdesouza-chromium, line 61: `ImportedRSAKey` seems to be duplicating in memory the information that we already have in `crypto::keypair::PrivateKey`. Ideally it is better to avoid these unnecessary conversion to string, or having our own custom types, but rather to use the type provided upstream, because we are paying in memory, performance, and potentially making the code harder to understand too.

components/web_discovery/browser/background_credential_helper.cc
DJAndries: nice, learned something new

components/web_discovery/browser/rsa.cc
DJAndries: This file provides an abstraction for RSA operations in general. This just slightly reduces the scope of the caller. If you disagree, perhaps we can discuss this in a follow-up to reduce the scope of this PR?

components/web_discovery/browser/rsa.cc
DJAndries: The encoded public key is cached in the `CredentialManager` for quick access. `CredentialManager` doesn't own the `PrivateKey`, only the `BackgroundCredentialManager` does. Changing the pattern would require some refactoring which might be out of scope for this PR. As mentioned in https://github.com/brave/brave-core/pull/30858#discussion_r2302735296, perhaps we can discuss this in a follow-up if you disagree with the status quo?

components/web_discovery/browser/rsa.cc
cdesouza-chromium: Yes, let's do in a follow up PR.

components/web_discovery/browser/rsa.cc
cdesouza-chromium: Unless Im misunderstanding something, for all effects and purposes, it looks like `ImportedRSAKey` holds the very same information that `crypto::keypair::PrivateKey` is meant to have, but in a slightly different format, so ultimately they do seem the same, but I could be missing something... let's take this conversation to a follow up PR as you suggested.

PR #30864, opened 2025-08-27
remove`-C path` when passing args to start Brave

bridiver:
Resolves https://github.com/brave/brave-browser/issues/48750


cdesouza-chromium (approved):
Looks good, but it would be nice if the code had a comment over the block explaining "why"

PR #30883, opened 2025-08-27
[Android] Fix Brave VPN server selection radio button interaction.

Olipro:
Resolves brave/brave-browser#48327

## Test Plan

1. Test purchasing the Brave VPN service functions without issue.
2. Test activating the VPN service via an account that already has the VPN purchased also works without issue.
4. Navigate to the VPN menu and ensure automatic selection is turned off.
5. Test the radio buttons to ensure that the region is switched correctly.
6. Test the parent entry to ensure it still brings you into city selection.

### DIFF
--- a/android/java/org/chromium/chrome/browser/vpn/activities/BraveVpnParentActivity.java
+++ b/android/java/org/chromium/chrome/browser/vpn/activities/BraveVpnParentActivity.java
@@ -20,6 +20,7 @@ import com.wireguard.android.backend.GoBackend;
 import com.wireguard.crypto.KeyPair;
 
 import org.chromium.base.Log;
+import org.chromium.base.ThreadUtils;
 import org.chromium.base.supplier.OneshotSupplier;
 import org.chromium.chrome.R;
 import org.chromium.chrome.browser.billing.InAppPurchaseWrapper;
@@ -39,8 +40,10 @@ import org.chromium.chrome.browser.vpn.utils.BraveVpnProfileUtils;
 import org.chromium.chrome.browser.vpn.utils.BraveVpnUtils;
 import org.chromium.chrome.browser.vpn.wireguard.WireguardConfigUtils;
 
-public abstract class BraveVpnParentActivity
-        extends AsyncInitializationActivity implements BraveVpnObserver {
+import java.util.concurrent.CompletableFuture;
+
+public abstract class BraveVpnParentActivity extends AsyncInitializationActivity
+        implements BraveVpnObserver {
     private static final String TAG = "BraveVPN";
     public boolean mIsVerification;
     protected BraveVpnPrefModel mBraveVpnPrefModel;
@@ -265,8 +268,16 @@ public abstract class BraveVpnParentActivity
                             braveVpnWireguardProfileCredentials.getClientId());
                     mBraveVpnPrefModel.setApiAuthToken(
                             braveVpnWireguardProfileCredentials.getApiAuthToken());
-                    BraveVpnPrefUtils.setPrefModel(mBraveVpnPrefModel);
-
+                    CompletableFuture<Void> awaitPrefDone = new CompletableFuture<Void>();
+                    ThreadUtils.postOnUiThread(
+                            () -> {
+                                try {
+                                    BraveVpnPrefUtils.setPrefModel(mBraveVpnPrefModel);
+                                } finally {
+                                    awaitPrefDone.complete(null);
+                                }
+                            });
+                    awaitPrefDone.get();
                     BraveVpnUtils.dismissProgressDialog();
                     Intent intent = GoBackend.VpnService.prepare(BraveVpnParentActivity.this);
                     if (intent != null) {
--- a/android/java/org/chromium/chrome/browser/vpn/activities/BraveVpnParentActivity.java
+++ b/android/java/org/chromium/chrome/browser/vpn/activities/BraveVpnParentActivity.java
@@ -29,7 +29,6 @@ import org.chromium.chrome.browser.init.ActivityProfileProvider;
 import org.chromium.chrome.browser.init.AsyncInitializationActivity;
 import org.chromium.chrome.browser.profiles.ProfileProvider;
 import org.chromium.chrome.browser.util.LiveDataUtil;
-import org.chromium.chrome.browser.util.TabUtils;
 import org.chromium.chrome.browser.vpn.BraveVpnNativeWorker;
 import org.chromium.chrome.browser.vpn.BraveVpnObserver;
 import org.chromium.chrome.browser.vpn.models.BraveVpnPrefModel;
@@ -285,7 +284,6 @@ public abstract class BraveVpnParentActivity extends AsyncInitializationActivity
                         return;
                     }
                     BraveVpnProfileUtils.getInstance().startVpn(BraveVpnParentActivity.this);
-                    TabUtils.bringChromeTabbedActivityToTheTop(BraveVpnParentActivity.this);
                 } catch (Exception e) {
                     BraveVpnUtils.dismissProgressDialog();
                     Log.e(TAG, e.getMessage());
--- a/android/java/org/chromium/chrome/browser/vpn/activities/VpnServerSelectionActivity.java
+++ b/android/java/org/chromium/chrome/browser/vpn/activities/VpnServerSelectionActivity.java
@@ -70,6 +70,10 @@ public class VpnServerSelectionActivity extends BraveVpnParentActivity
         void onServerRegionClick(Region region);
     }
 
+    public interface OnQuickServerSelection {
+        void onQuickServerSelect(Region region);
+    }
+
     @Override
     public boolean onOptionsItemSelected(MenuItem item) {
         if (item.getItemId() == android.R.id.home) {
@@ -168,6 +172,24 @@ public class VpnServerSelectionActivity extends BraveVpnParentActivity
                                             BraveVpnUtils.selectedRegion = region;
                                             BraveVpnUtils.openVpnServerActivity(
                                                     VpnServerSelectionActivity.this, region);
+                                        }
+                                    });
+                            mBraveVpnServerSelectionAdapter.setOnQuickServerSelection(
+                                    new OnQuickServerSelection() {
+                                        @Override
+                                        public void onQuickServerSelect(Region region) {
+                                            BraveVpnUtils.selectedRegion = region;
+                                            BraveVpnUtils.selectedServerRegion =
+                                                    new BraveVpnServerRegion(
+                                                            false,
+                                                            region.country,
+                                                            region.continent,
+                                                            region.countryIsoCode,
+                                                            region.name,
+                                                            region.namePretty,
+                                                            BraveVpnConstants
+                                                                    .REGION_PRECISION_COUNTRY);
+                                            changeServerRegion();
                                         }
                                     });
                             mServerRegionList.setAdapter(mBraveVpnServerSelectionAdapter);
--- a/android/java/org/chromium/chrome/browser/vpn/activities/VpnServerSelectionActivity.java
+++ b/android/java/org/chromium/chrome/browser/vpn/activities/VpnServerSelectionActivity.java
@@ -57,6 +57,7 @@ public class VpnServerSelectionActivity extends BraveVpnParentActivity
 
     private void initVpnService() {
         if (mServiceHandler != null) {
+            mServiceHandler.close();
             mServiceHandler = null;
         }
         mServiceHandler =
@@ -218,4 +219,13 @@ public class VpnServerSelectionActivity extends BraveVpnParentActivity
 
     @Override
     public void updateProfileView() {}
+
+    @Override
+    protected void onDestroy() {
+        super.onDestroy();
+        if (mServiceHandler != null) {
+            mServiceHandler.close();
+            mServiceHandler = null;
+        }
+    }
 }
--- a/android/java/org/chromium/chrome/browser/vpn/adapters/BraveVpnServerSelectionAdapter.java
+++ b/android/java/org/chromium/chrome/browser/vpn/adapters/BraveVpnServerSelectionAdapter.java
@@ -19,6 +19,7 @@ import com.google.android.material.radiobutton.MaterialRadioButton;
 
 import org.chromium.brave_vpn.mojom.Region;
 import org.chromium.chrome.R;
+import org.chromium.chrome.browser.vpn.activities.VpnServerSelectionActivity.OnQuickServerSelection;
 import org.chromium.chrome.browser.vpn.activities.VpnServerSelectionActivity.OnServerRegionSelection;
 import org.chromium.chrome.browser.vpn.utils.BraveVpnPrefUtils;
 import org.chromium.chrome.browser.vpn.utils.BraveVpnUtils;
@@ -31,6 +32,8 @@ public class BraveVpnServerSelectionAdapter
     private final Context mContext;
     private List<Region> mRegions = new ArrayList<>();
     private OnServerRegionSelection mOnServerRegionSelection;
+    private OnQuickServerSelection mOnQuickServerSelection;
+    private Region mActiveRegion;
 
     public BraveVpnServerSelectionAdapter(Context context) {
         this.mContext = context;
@@ -43,6 +46,28 @@ public class BraveVpnServerSelectionAdapter
         View listItem = layoutInflater.inflate(
                 R.layout.brave_vpn_server_selection_item_layout, parent, false);
         return new ViewHolder(listItem);
+    }
+
+    private static enum CheckedState {
+        SET_CHECKED,
+        SET_UNCHECKED,
+    }
+
+    @Override
+    public void onBindViewHolder(
+            @NonNull ViewHolder holder, int position, @NonNull List<Object> payloads) {
+        if (payloads.isEmpty()) {
+            onBindViewHolder(holder, position);
+            return;
+        }
+        Object isCheckedObj = payloads.get(0);
+        if (isCheckedObj instanceof CheckedState) {
+            boolean isChecked = ((CheckedState) isCheckedObj) == CheckedState.SET_CHECKED;
+            holder.serverRadioButton.setChecked(isChecked);
+            if (isChecked) {
+                mActiveRegion = mRegions.get(position);
+            }
+        }
     }
 
     @Override
@@ -62,8 +87,28 @@ public class BraveVpnServerSelectionAdapter
                             .getQuantityString(
                                     R.plurals.server_text, region.serverCount, region.serverCount);
             holder.cityServerText.setText(cityText.concat(serverText));
-            holder.serverRadioButton.setChecked(
-                    BraveVpnPrefUtils.getRegionIsoCode().equals(region.countryIsoCode));
+            boolean isChecked = BraveVpnPrefUtils.getRegionIsoCode().equals(region.countryIsoCode);
+            holder.serverRadioButton.setChecked(isChecked);
+            if (isChecked) {
+                mActiveRegion = region;
+            }
+            holder.serverRadioButton.setOnClickListener(
+                    new View.OnClickListener() {
+                        @Override
+                        public void onClick(View v) {
+                            if (region == mActiveRegion) {
+                                return;
+                            }
+                            mOnQuickServerSelection.onQuickServerSelect(region);
+                            if (mActiveRegion != null) {
+                                notifyItemChanged(
+                                        mRegions.indexOf(mActiveRegion),
+                                        CheckedState.SET_UNCHECKED);
+                            }
+                            notifyItemChanged(
+                                    holder.getBindingAdapterPosition(), CheckedState.SET_CHECKED);
+                        }
+                    });
             holder.serverSelectionItemLayout.setOnClickListener(
                     new View.OnClickListener() {
                         @Override
@@ -85,6 +130,10 @@ public class BraveVpnServerSelectionAdapter
 
     public void setOnServerRegionSelection(OnServerRegionSelection onServerRegionSelection) {
         this.mOnServerRegionSelection = onServerRegionSelection;
+    }
+
+    public void setOnQuickServerSelection(OnQuickServerSelection onQuickServerSelection) {
+        this.mOnQuickServerSelection = onQuickServerSelection;
     }
 
     public static class ViewHolder extends RecyclerView.ViewHolder {
--- a/android/java/org/chromium/chrome/browser/vpn/utils/BraveVpnPrefUtils.java
+++ b/android/java/org/chromium/chrome/browser/vpn/utils/BraveVpnPrefUtils.java
@@ -11,6 +11,7 @@ import org.chromium.chrome.browser.preferences.BravePref;
 import org.chromium.chrome.browser.preferences.ChromeSharedPreferences;
 import org.chromium.chrome.browser.profiles.ProfileManager;
 import org.chromium.chrome.browser.vpn.models.BraveVpnPrefModel;
+import org.chromium.components.prefs.PrefService;
 import org.chromium.components.user_prefs.UserPrefs;
 
 import java.util.Collections;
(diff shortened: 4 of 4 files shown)


android/java/org/chromium/chrome/browser/vpn/activities/BraveVpnParentActivity.java
```diff
@@ -265,8 +267,15 @@ public void run() {
                             braveVpnWireguardProfileCredentials.getClientId());
                     mBraveVpnPrefModel.setApiAuthToken(
                             braveVpnWireguardProfileCredentials.getApiAuthToken());
-                    BraveVpnPrefUtils.setPrefModel(mBraveVpnPrefModel);
-
+                    CompletableFuture<Void> awaitPrefDone = new CompletableFuture<Void>();
+                    BraveVpnParentActivity.this.runOnUiThread(() -> {
```
Olipro: There is perhaps a bigger conversation or task to be had here around ensuring code is being called/dispatched on the appropriate thread. However, for the time being, this does un-break the VPN starting on Debug builds and will ensure we don't have any other issues in production.

android/java/org/chromium/chrome/browser/vpn/adapters/BraveVpnServerSelectionAdapter.java
```diff
@@ -31,6 +34,8 @@ public class BraveVpnServerSelectionAdapter
     private final Context mContext;
     private List<Region> mRegions = new ArrayList<>();
     private OnServerRegionSelection mOnServerRegionSelection;
+    private OnQuickServerSelection mOnQuickServerSelection;
+    private static WeakReference<ViewHolder> sActiveRegion = new WeakReference<>(null);
```
Olipro: This static is used to toggle off the currently-active radio button when the user chooses a different country by tapping on another radio button.

This doesn't strike me as a great way of organising the furniture, but then again, perhaps this is fine. Any input on a better solution (if one exists) would be most welcome.

android/java/org/chromium/chrome/browser/vpn/utils/BraveVpnPrefUtils.java
```diff
@@ -125,20 +126,24 @@ public static String getHostnameDisplay() {
     }
 
     public static void setPurchaseToken(String value) {
+        if (value == null) {
```
Olipro, line 129: see the commit message for context regarding this null check (and the one below)

android/java/org/chromium/chrome/browser/vpn/utils/BraveVpnPrefUtils.java
```diff
@@ -125,20 +126,24 @@ public static String getHostnameDisplay() {
...
+            return;
+        }
         ChromeSharedPreferences.getInstance().writeString(PREF_BRAVE_VPN_PURCHASE_TOKEN, value);
-        UserPrefs.get(ProfileManager.getLastUsedRegularProfile())
-                .setString(BravePref.BRAVE_VPN_PURCHASE_TOKEN_ANDROID, value);
-        UserPrefs.get(ProfileManager.getLastUsedRegularProfile())
-                .setString(
-                        BravePref.BRAVE_VPN_PACKAGE_ANDROID,
-                        ContextUtils.getApplicationContext().getPackageName());
+        PrefService prefSvc = UserPrefs.get(ProfileManager.getLastUsedRegularProfile());
```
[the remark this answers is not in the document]
Olipro: For clarity: This code was refactored to avoid making multiple redundant calls to `UserPrefs.get()` - I did not introduce any semantic change.

android/java/org/chromium/chrome/browser/vpn/activities/VpnServerSelectionActivity.java
```diff
@@ -169,6 +173,23 @@ public void onServerRegionClick(Region region) {
                                                     VpnServerSelectionActivity.this, region);
                                         }
                                     });
+                            mBraveVpnServerSelectionAdapter.setOnQuickServerSelection(
+                                new OnQuickServerSelection() {
+                                    @Override
+                                    public void onQuickServerSelect(Region region) {
```
Olipro: Not sure how descriptive this method name (and its relatives) are. Happy to do some bikeshedding over better names if anyone has one.

SergeyZhukovsky:
Resolves https://github.com/brave/brave-browser/issues/47959 shows a different issue for it, put the correct one. I assume it should be that one https://github.com/brave/brave-browser/issues/48327

Olipro:
> Resolves [brave/brave-browser#47959](https://github.com/brave/brave-browser/issues/47959) shows a different issue for it, put the correct one. I assume it should be that one [brave/brave-browser#48327](https://github.com/brave/brave-browser/issues/48327)

Thanks, not sure how I managed that, it's correct in the commit message in any case.

android/java/org/chromium/chrome/browser/vpn/activities/BraveVpnParentActivity.java
```diff
@@ -265,8 +267,16 @@ public void run() {
                             braveVpnWireguardProfileCredentials.getClientId());
                     mBraveVpnPrefModel.setApiAuthToken(
                             braveVpnWireguardProfileCredentials.getApiAuthToken());
-                    BraveVpnPrefUtils.setPrefModel(mBraveVpnPrefModel);
-
+                    CompletableFuture<Void> awaitPrefDone = new CompletableFuture<Void>();
+                    BraveVpnParentActivity.this.runOnUiThread(
```
SergeyZhukovsky: use chromium wrappers and task runners `ThreadUtils.postOnUiThread` or `PostTask.postTask(TaskTraits.UI_DEFAULT, ...`

android/java/org/chromium/chrome/browser/vpn/activities/BraveVpnParentActivity.java
```diff
@@ -265,8 +267,16 @@ public void run() {
...
+                            () -> {
+                                try {
+                                    BraveVpnPrefUtils.setPrefModel(mBraveVpnPrefModel);
+                                } finally {
+                                    awaitPrefDone.complete(null);
+                                }
+                            });
+                    awaitPrefDone.get();
```
SergeyZhukovsky, line 279: that function is a bit mess as it executes on background thread and in the middle in needs a UI thread and after that again bg thread. In the best case we would refactor it somehow, but it's out of scope of that task.  I would probably create another function where put everything below and call `PostTask.postTask(TaskTraits.BEST_EFFORT_MAY_BLOCK` to run the rest on the bg thread, not sure does it need to run in the bg thread. That way we remove `CompletableFuture<Void> awaitPrefDone = new CompletableFuture<Void>();` and avoid extra blockages of threads.

android/java/org/chromium/chrome/browser/vpn/adapters/BraveVpnServerSelectionAdapter.java
```diff
@@ -62,8 +66,26 @@ public void onBindViewHolder(@NonNull ViewHolder holder, int position) {
                             .getQuantityString(
                                     R.plurals.server_text, region.serverCount, region.serverCount);
             holder.cityServerText.setText(cityText.concat(serverText));
-            holder.serverRadioButton.setChecked(
-                    BraveVpnPrefUtils.getRegionIsoCode().equals(region.countryIsoCode));
+            boolean isChecked = BraveVpnPrefUtils.getRegionIsoCode().equals(region.countryIsoCode);
+            holder.serverRadioButton.setChecked(isChecked);
+            if (isChecked) {
+                sActiveRegion = new WeakReference<>(holder);
```
SergeyZhukovsky: Saving a position of the current active region would work better as to me. We would get rid of storing the `ViewHolder` which in fact could be destroyed especially if the item is not visible currently to the user. So instead of `private static WeakReference<ViewHolder> sActiveRegion = new WeakReference<>(null);` we would have `private int mActivePostion;` When a radio button is clicked we call `notifyItemChanged(mActivePostion)` that triggers a redraw of the item on that particular position. We won't need to call `holder.serverRadioButton.setChecked(true);` explicitly as well inside the listener we just call to redraw 2 elements `notifyItemChanged(mActivePostion)` and `notifyItemChanged(position)` after the call to `mOnQuickServerSelection.onQuickServerSelect(region);`

android/java/org/chromium/chrome/browser/vpn/adapters/BraveVpnServerSelectionAdapter.java
```diff
@@ -62,8 +66,26 @@ public void onBindViewHolder(@NonNull ViewHolder holder, int position) {
...
+                            if (holder != sActiveRegion.get()) {
+                                mOnQuickServerSelection.onQuickServerSelect(region);
+                                holder.serverRadioButton.setChecked(true);
+                                ViewHolder maybeActiveSelection = sActiveRegion.get();
+                                if (maybeActiveSelection != null) {
+                                    maybeActiveSelection.serverRadioButton.setChecked(false);
+                                }
+                                sActiveRegion = new WeakReference<>(holder);
+                            }
+                        }
```
SergeyZhukovsky, line 110: as I mentioned above the whole `onClick` function could look like this
```
                        public void onClick(View v) {
                            if (position == mActivePostion) {
                                return;
                            }
                            mOnQuickServerSelection.onQuickServerSelect(region);
                            notifyItemChanged(mActivePostion);
                            notifyItemChanged(position);
                        }
```

SergeyZhukovsky: `mActivePostion` should also be updated automatically when the item on `position` redraws as it means it would be checked.

android/java/org/chromium/chrome/browser/vpn/activities/BraveVpnParentActivity.java
Olipro: A _bit_ of a mess is an understatement. I agree that it should be refactored, probably as a separate issue - I suppose the most sensible would be to segregate the behaviours into two separate calls: one that does everything that requires a UI thread and another with everything else.

With that said, I'm not sure what the ask is here. I would presume that it's a potential race condition to be calling `startVpn` without first waiting until `setPrefModel` has returned. Are you suggesting that we also move all of the subsequent code below this comment into executing on the UI thread too? This looks tricky given the exception handling.

android/java/org/chromium/chrome/browser/vpn/adapters/BraveVpnServerSelectionAdapter.java
Olipro: Discussed below. Resolving.

android/java/org/chromium/chrome/browser/vpn/adapters/BraveVpnServerSelectionAdapter.java
Olipro: This appears to have broken things in some insidious way. The screen now closes after the VPN switches and _occasionally_ The following stack trace is emitted:

```
java.lang.IllegalStateException: Warning: Router objects should be explicitly closed when no longer required otherwise you may leak handles.
                                                                                                    	at org.chromium.mojo.bindings.AutoCloseableRouter.finalize(chromium-MonochromePublic.apk-default-428300028:27)
                                                                                                    	at java.lang.Daemons$FinalizerDaemon.doFinalize(Daemons.java:387)
                                                                                                    	at java.lang.Daemons$FinalizerDaemon.processReference(Daemons.java:367)
                                                                                                    	at java.lang.Daemons$FinalizerDaemon.runInternal(Daemons.java:339)
                                                                                                    	at java.lang.Daemons$Daemon.run(Daemons.java:132)
                                                                                                    	at java.lang.Thread.run(Thread.java:1119)
                                                                                                    Caused by: java.lang.Exception: AutocloseableRouter allocated at:
                                                                                                    	at org.chromium.mojo.bindings.Interface$Manager.attachProxy(chromium-MonochromePublic.apk-default-428300028:27)
                                                                                                    	at org.chromium.chrome.browser.vpn.activities.VpnServerSelectionActivity.initVpnService(chromium-MonochromePublic.apk-default-428300028:73)
                                                                                                    	at org.chromium.chrome.browser.vpn.activities.VpnServerSelectionActivity.finishNativeInitialization(chromium-MonochromePublic.apk-default-428300028:4)
                                                                                                    	at org.chromium.chrome.browser.init.BrowserParts.startNativeInitialization(chromium-MonochromePublic.apk-default-428300028:1)
                                                                                                    	at org.chromium.chrome.browser.init.ChromeBrowserInitializer$$ExternalSyntheticLambda1.run(chromium-MonochromePublic.apk-default-428300028:61)
                                                                                                    	at org.chromium.base.task.PostTask$TaskOriginRunnable.run(chromium-MonochromePublic.apk-default-428300028:10)
                                                                                                    	at org.chromium.base.task.ChainedTasks$1.run(chromium-MonochromePublic.apk-default-428300028:47)
                                                                                                    	at org.chromium.base.task.PostTask$TaskOriginRunnable.run(chromium-MonochromePublic.apk-default-428300028:10)
                                                                                                    	at org.chromium.base.task.TaskRunnerImpl.runTask(chromium-MonochromePublic.apk-default-428300028:35)
                                                                                                    	at android.os.MessageQueue.nativePollOnce(Native Method)
                                                                                                    	at android.os.MessageQueue.nextLegacy(MessageQueue.java:913)
                                                                                                    	at android.os.MessageQueue.next(MessageQueue.java:1025)
                                                                                                    	at android.os.Looper.loopOnce(Looper.java:196)
                                                                                                    	at android.os.Looper.loop(Looper.java:338)
                                                                                                    	at android.app.ActivityThread.main(ActivityThread.java:9067)
                                                                                                    	at java.lang.reflect.Method.invoke(Native Method)
                                                                                                    	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:593)
                                                                                                    	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:932)
```

Unfortunately, I have no idea what to do with this, nothing stands out to me that should be wrapped in a _try-with-resources_ block.

Olipro: It appears to me that this approach might be invalid. The linter complains about this:

`Error: Do not treat position as fixed; only use immediately and call holder.getAdapterPosition() to look it up later [RecyclerView]`

Since we seemingly can't rely on the position, I'm not sure what alternatives we have here. I originally went with the weak reference option since, as I understand it, if an entry doesn't exist on-screen, we do nothing because its `onBindViewHolder` will execute if the user scrolls it into view - I may be completely wrong here of course. Let me know how you'd like to proceed.

SergeyZhukovsky: the thing is a ViewHolder you store might eventually be unbound, recycled, or even garbage collected. You can't rely on it either. The error honestly is just a warning as we control the dataset `private List<Region> mRegions = new ArrayList<>();` and it doesn't change. But we can in that case store `Region mActiveRegion` and get a position from `mRegions` using `mActiveRegion` and call `notifyItemChanged`

SergeyZhukovsky: With the above we let the linter to relax.

android/java/org/chromium/chrome/browser/vpn/activities/BraveVpnParentActivity.java
SergeyZhukovsky: What I proposed is to move
```
                    BraveVpnUtils.dismissProgressDialog();
                    Intent intent = GoBackend.VpnService.prepare(BraveVpnParentActivity.this);
                    if (intent != null) {
                        mIntentActivityResultLauncher.launch(intent);
                        return;
                    }
                    BraveVpnProfileUtils.getInstance().startVpn(BraveVpnParentActivity.this);
                    TabUtils.bringChromeTabbedActivityToTheTop(BraveVpnParentActivity.this);
```
in a different function and call it from the UI thread using that `PostTask.postTask(TaskTraits.BEST_EFFORT_MAY_BLOCK` that it's executes in a bg thread again.
But that could be done as a follow on refactor. Just change `runOnUiThread` to chromium's wrapper and it's fine.

android/java/org/chromium/chrome/browser/vpn/adapters/BraveVpnServerSelectionAdapter.java
SergeyZhukovsky: for `notifyItemChanged(position);` use `notifyItemChanged(holder.getBindingAdapterPosition()`

SergeyZhukovsky: We can probably store `region.countryIsoCode` as an active item instead of the whole `Region` object.

Olipro: OK, I applied the changes but unfortunately, it didn't work since it appears that the VPN changeover happens completely asynchronously resulting in the radio button checks occurring before it's switched over.

So, instead, I've added the override for `onBindViewHolder` that takes the optional `payload` argument. and I explicitly pass true/false in the click handler.

This works fine and I believe constitutes the better option since having it _just_ update the radio buttons is cheaper than running the existing handler that initialises everything all over again.

android/java/org/chromium/chrome/browser/vpn/activities/BraveVpnParentActivity.java
```diff
@@ -265,16 +267,23 @@ public void run() {
...
                     BraveVpnUtils.dismissProgressDialog();
                     Intent intent = GoBackend.VpnService.prepare(BraveVpnParentActivity.this);
                     if (intent != null) {
                         mIntentActivityResultLauncher.launch(intent);
                         return;
                     }
                     BraveVpnProfileUtils.getInstance().startVpn(BraveVpnParentActivity.this);
-                    TabUtils.bringChromeTabbedActivityToTheTop(BraveVpnParentActivity.this);
```
Olipro, line 277: I've removed this line since it boots you straight out of the selection menu. I can't think of any menus that a user would expect to be unceremoniously kicked out of, especially if they're just using Brave as a way to quickly switch between VPN regions at their leisure.

SergeyZhukovsky: Yeah, I don't think it's right behaviour in Settings, but that Activity is used not only there, it could be removing that line broke something else, probably something related to initial purchase and init of VPN.

android/java/org/chromium/chrome/browser/vpn/adapters/BraveVpnServerSelectionAdapter.java
SergeyZhukovsky: sounds good!

android/java/org/chromium/chrome/browser/vpn/adapters/BraveVpnServerSelectionAdapter.java
```diff
@@ -45,6 +48,22 @@ public ViewHolder onCreateViewHolder(ViewGroup parent, int viewType) {
         return new ViewHolder(listItem);
     }
 
+    @Override
+    public void onBindViewHolder(
+            @NonNull ViewHolder holder, int position, @NonNull List<Object> payloads) {
+        if (payloads.size() == 0) {
```
SergeyZhukovsky: nit: `payloads.isEmpty()`

android/java/org/chromium/chrome/browser/vpn/adapters/BraveVpnServerSelectionAdapter.java
```diff
@@ -62,8 +81,24 @@ public void onBindViewHolder(@NonNull ViewHolder holder, int position) {
...
+            boolean isChecked = BraveVpnPrefUtils.getRegionIsoCode().equals(region.countryIsoCode);
+            holder.serverRadioButton.setChecked(isChecked);
+            if (isChecked) {
+                mActiveRegion = region;
+            }
+            holder.serverRadioButton.setOnClickListener(
+                    new View.OnClickListener() {
+                        @Override
+                        public void onClick(View v) {
+                            if (region != mActiveRegion) {
```
SergeyZhukovsky: nit: could return earlier:
```
if (region == mActiveRegion) {
   return;
}
```

android/java/org/chromium/chrome/browser/vpn/activities/VpnServerSelectionActivity.java
```diff
@@ -218,4 +241,13 @@ public void showRestoreMenu(boolean shouldShowRestore) {}
 
     @Override
     public void updateProfileView() {}
+
+    @Override
+    protected void onDestroy() {
+        super.onDestroy();
+        if (mServiceHandler != null) {
+            mServiceHandler.close();
```
SergeyZhukovsky, line 249: hm, it looks like the mojo channel close was totally missed in that class.

android/java/org/chromium/chrome/browser/vpn/activities/BraveVpnParentActivity.java
SergeyZhukovsky: I would suggest to verify the flow to make sure nothing breaks.

android/java/org/chromium/chrome/browser/vpn/adapters/BraveVpnServerSelectionAdapter.java
Olipro: It returns immediately since the if-statement as it currently stands encloses everything inside its scope.

It won't have any impact on the code's performance to change it, this is purely a matter of style. If we have a hard rule about preferring early return rather than enclosing the executing behaviour, I can change it.

SergeyZhukovsky: yes we don't have any rules regarding to it and it doesn't mean you have to change it right now, but pretty often such constructions make lines longer and it would necessary to break them otherwise lint could complain. Keep them shorter and more readable is always a preferred way to go. No performance impact.

android/java/org/chromium/chrome/browser/vpn/activities/VpnServerSelectionActivity.java
Olipro: Yeah, Do we have some sort of tooling that's possibly been catching this and surfacing it? I find it a bit eyebrow-raising that this wasn't detected.

...or maybe it was and nobody got around to fixing it?

SergeyZhukovsky: we don't have any tooling that detects that. Not sure what happened here, probably no one noticed it and missed on review. It should give a warning in logcat though, so it should be visible on implementation phase at least.

android/java/org/chromium/chrome/browser/vpn/activities/BraveVpnParentActivity.java
Olipro: Tested with a completely fresh install of Brave where I had to authenticate my account to be granted VPN access.

No negative impact to the flow, everything works. I'll update the test plan though so QA can give it a second go.

SergeyZhukovsky: Let it be in that case!

SergeyZhukovsky (approved):
++

android/java/org/chromium/chrome/browser/vpn/activities/VpnServerSelectionActivity.java
Olipro: Yeah, that's how I caught it. I've made a personal note to look into improving our automation to catch things like this.

android/java/org/chromium/chrome/browser/vpn/adapters/BraveVpnServerSelectionAdapter.java
Olipro: Fixed.

android/java/org/chromium/chrome/browser/vpn/adapters/BraveVpnServerSelectionAdapter.java
Olipro: Done.

PR #30889, opened 2025-08-27
[cr141][Android] Update buttons on hub toolbar pane switching

AlexeyBarabash:
This PR is basically uplift of 
https://github.com/brave/brave-core/commit/743e4b8b769ee604222c70553dab080c640c6a91

The fix was done by @samartnik for `cr141` but the issue appeared on `cr140`.

Resolves https://github.com/brave/brave-browser/issues/48776

### DIFF
--- a/browser/hub/internal/android/java/src/org/chromium/chrome/browser/hub/BraveHubToolbarView.java
+++ b/browser/hub/internal/android/java/src/org/chromium/chrome/browser/hub/BraveHubToolbarView.java
@@ -37,6 +37,14 @@ public class BraveHubToolbarView extends HubToolbarView {
     }
 
     @Override
+    void setPaneSwitcherIndex(int index) {
+        super.setPaneSwitcherIndex(index);
+
+        // Update visibility of action and menu buttons based on the switching panes.
+        updateButtonsVisibility();
+    }
+
+    @Override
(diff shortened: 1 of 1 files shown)


kjozwiak:
Verification **PASSED** on `Pixel 6` running `Android 16` using the following build(s):

```
Brave | 1.84.4 Chromium: 140.0.7339.26 (Official Build) canary (64-bit)
--- | ---
Revision | 81e3acfb86177c07e160dcc450bd8089a899d2b5
OS | Android 16; Build/BP2A.250605.031.A5; 36; REL
```

Reproduced the original issue with `1.84.2 Chromium: 140.0.7339.26` using the STR/Cases outlined via https://github.com/brave/brave-browser/issues/48776#issue-3360954697 where the `tab manager` has two `+` buttons after leaving PB as per the following:

Screenshot_20250828-144404

Using the same STR/Cases outlined above, ensured that there's only a single `+` button when exiting PB mode while using `top toolbar` as per the following:

https://github.com/user-attachments/assets/536bd395-f5ef-4b0b-b280-63c989e1e81b

Also ensured that there's no extra `+` buttons being added within the `tab manager` when using `bottom toolbar` after exiting PB mode as per the following:

https://github.com/user-attachments/assets/d007295f-50f2-47d2-a8f2-cd99f9d8536d

PR #30520, opened 2025-08-07
[Brave Midnight] Add Brave Midnight (Darker) theme feature and pref

sangwoo108:
This commit introduces the Brave Midnight theme, a darker variant of the existing dark theme, providing users with enhanced customization options for their browsing experience.

Key changes:
* Add feature flag `kBraveMidnightTheme` to control theme availability
* Create new preference `kBraveDarkerMode` to store user's theme choice
* Implement UI toggle in Customize Chrome side panel for theme switching
* Add Mojo interface methods for theme state management

Technical implementation:
- Created `browser/ui/color/features.{cc,h}` for feature flag definition
- Added feature flag to `about_flags.cc` for experimental access
- Registered preference in `brave_profile_prefs.cc` when feature is enabled
- Extended Customize Chrome UI with `DarkerThemeToggle` component
- Added Mojo interface methods `GetUseDarkerTheme`/`SetUseDarkerTheme`

The feature is disabled by default and can be enabled via the `brave-midnight-theme` flag in brave://flags. When enabled, users can toggle the darker theme through the Customize Chrome side panel.

image
image

Resolves https://github.com/brave/brave-browser/issues/48192
Resolves https://github.com/brave/brave-browser/issues/48193

PR #30892, opened 2025-08-28
[Midnight] Add kDarker variant to SchemeVariant enum for color mixer integration

sangwoo108:
This change introduces support for a "kDarker" theme variant by extending the color system infrastructure:

- Add kDarker to SchemeVariant enum in chromium_src/ui/color/color_provider_key.h
- Update brave_browser_frame to handle darker theme variant detection

The kDarker variant enables the color mixer to properly reference and apply darker theme colors throughout the browser UI

Resolves


browser/ui/color/BUILD.gn
```diff
@@ -25,3 +25,23 @@ source_set("unit_tests") {
     "//ui/color",
   ]
 }
+
+if (!is_android) {
```
simonhong: I think we don't need this guard as this browser_tests target is already added in non-android deps list.

browser/ui/views/frame/brave_browser_frame.cc
```diff
@@ -74,3 +77,18 @@ void BraveBrowserFrame::SetTabDragKind(TabDragKind kind) {
...
+
+ui::ColorProviderKey BraveBrowserFrame::GetColorProviderKey() const {
+  auto key = BrowserFrame::GetColorProviderKey();
+
+  if (base::FeatureList::IsEnabled(color::features::kBraveDarkerTheme)) {
+    if (!key.scheme_variant &&
+        key.color_mode == ui::ColorProviderKey::ColorMode::kDark &&
+        browser_view_->browser()->profile()->GetPrefs()->GetBoolean(
+            color::prefs::kBraveDarkerMode)) {
+      key.scheme_variant = ui::ColorProviderKey::SchemeVariant::kDarker;
```
simonhong: curious what it means if key_scheme_variant is non-null and our darker mode(`kBraveDarkerMode`) is enabled.

it means we don't overwrite if if upstream usea another variant?

scheme variant is very new to me so need to understand :)

browser/ui/color/darker_theme_browsertest.cc
```diff
@@ -0,0 +1,61 @@
...
+  ~DarkerThemeBrowserTest() override = default;
+
+ private:
+  base::test::ScopedFeatureList scoped_feature_list_;
+};
+
+IN_PROC_BROWSER_TEST_F(DarkerThemeBrowserTest, DarkerModeDefaultOff) {
+  // By default, the darker theme should be off.
+  EXPECT_FALSE(browser()->profile()->GetPrefs()->GetBoolean(
+      color::prefs::kBraveDarkerMode));
```
simonhong: I think this simple check could be put at the start of below test case as it just check initial state.
browser test is resource heavy job.

simonhong (commented):
++

browser/ui/color/BUILD.gn
goodov: sometimes you have to do this, because even if a target is not explicitly added to deps, it still becomes part of a global buildgraph when it's declared in `BUILD.gn` that's referenced. This can fail gn check and there's no way to fix this other than remove these files from the buildgraph in this way.

I'm not sure if here is this exact scenario though, but this is something that may require such `if ()` blocks here and there.

sangwoo108: Yes, @goodov is right. That's exactly what happened in Android CI.

```
2025-08-28T13:01:25.350Z] /home/ubuntu/workspace/e-core-build-pr-android_PR-30892/src

[2025-08-28T13:01:25.350Z] > gn gen /home/ubuntu/workspace/e-core-build-pr-android_PR-30892/src/out/android_Component --check

[2025-08-28T13:01:28.612Z] ERROR Unresolved dependencies.

[2025-08-28T13:01:28.612Z] //brave/browser/ui/color:browser_tests(//build/toolchain/android:android_clang_x64)

[2025-08-28T13:01:28.612Z]   needs //chrome/test:test_support_ui(//build/toolchain/android:android_clang_x64)

[2025-08-28T13:01:28.612Z] 

[2025-08-28T13:01:28.867Z] null

```

browser/ui/views/frame/brave_browser_frame.cc
sangwoo108: > it means we don't overwrite if if upstream usea another variant?

Yeah, at the moment that's what I intended. IIUC, scheme variant is used make variation on color scheme like, "dark, light, system", by populating palette dynamically. So I thought "Darker" theme is also be part of scheme variant.

And it's used in Ash, practically we wouldn't bump into it.

browser/ui/color/darker_theme_browsertest.cc
sangwoo108: Good call. Merged two test cases

browser/ui/BUILD.gn
```diff
@@ -607,6 +607,7 @@ source_set("ui") {
     ]
 
     deps += [
+      "//brave/browser/ui/color",
```
sangwoo108: Note: this was added here for "brave_browser_frame", which hasn't split into a smaller target and included in this file directly.

simonhong (approved):
++

browser/ui/views/frame/brave_browser_frame.cc
sangwoo108: Actually I was wrong. When using custom color, this value will be filled per color . I think we should overwrite the value.

ref: https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/ui/webui/cr_components/theme_color_picker/customize_chrome_colors.cc;l=41;drc=ba93f1c7d7a13c567f0af7400c88928fd9d9d18e

sangwoo108: Changed to overwrite.

sangwoo108:
@goodov , @simonhong , Sorry but I added a commit https://github.com/brave/brave-core/pull/30892/commits/f6997af664d01076846c1aa071b3144c812f8d95 . Could you take a look again?

simonhong (approved):

browser/ui/color/BUILD.gn
bridiver: This is not the right fix. If this code is not relevant to android then the target should have `assert(!is_android)`

sangwoo108: > This is not the right fix. If this code is not relevant to android then the target should have assert(!is_android)

I've already tried taht, but the CI failed. I guess that's because of what @goodov mentioned.

bridiver: `//brave/browser/ui/color` dep also needs to be behind the guard, but the correct guard to use here for all the code is actually `toolkit_views` because that's the guard for BrowserFrame

bridiver: there are several other places where `//brave/browser/ui/color` is not properly guarded with `toolkit_views` or behind the wrong guard (`!is_android`), most notably in `brave_chrome_browser_deps` where it was added here https://github.com/brave/brave-core/pull/30520/files and doesn't apply to android https://github.com/brave/brave-core/pull/30520/files#diff-cacbc437964cc20aaaf741b80ecbea5cfce9ca9e67725476f73f14cbab2269fbR489 so you also need `#if defined(TOOLKIT_VIEWS)` around the relevant code in some places. I suggest a separate PR to fix the issues created by https://github.com/brave/brave-core/pull/30520 and then follow-up with this one

bridiver: if this is meant to eventually apply to android as well then this test should be behind `toolkit_views` guard, but either way this seems like the wrong place for this test and it should be in `browser/ui/views/frame` because it's testing BraveBrowserFrame behavior, not `brave/browser/ui/color` behavior

bridiver: so I guess ultimately the complete fix here depends on whether all the relevant code for darker mode is desktop-only or not. If it is, it should all be behind toolkit_views. Either way the test should move and use `toolkit_views` instead of `!is_android` to match browser_frame/brave_browser_frame guard

test/BUILD.gn
```diff
@@ -1159,6 +1159,7 @@ test("brave_browser_tests") {
       "//brave/browser/password_manager:browser_tests",
       "//brave/browser/sharing_hub:browser_tests",
       "//brave/browser/ui/ai_rewriter:browsertest",
+      "//brave/browser/ui/color:browser_tests",
```
sangwoo108: from @bridiver , this should be guarded by toolkit_views, not by !is_android.

browser/ui/color/BUILD.gn
sangwoo108: Now it's migrated into `darker_theme` target and guarded with `toolkit_views`

browser/ui/darker_theme/darker_theme_browsertest.cc
```diff
@@ -0,0 +1,59 @@
...
+#include "ui/color/color_provider_key.h"
+
+class DarkerThemeBrowserTest : public InProcessBrowserTest {
+ public:
+  DarkerThemeBrowserTest()
+      : scoped_feature_list_(darker_theme::features::kBraveDarkerTheme) {}
+  ~DarkerThemeBrowserTest() override = default;
+
+ private:
+  base::test::ScopedFeatureList scoped_feature_list_;
```
bridiver, line 40: test need to check behavior with the feature flag enabled and disabled

browser/ui/darker_theme/darker_theme_browsertest.cc
```diff
@@ -0,0 +1,59 @@
...
+  // By default, the darker theme should be off.
+  ASSERT_FALSE(browser()->profile()->GetPrefs()->GetBoolean(
+      darker_theme::prefs::kBraveDarkerMode));
+
+  auto* browser_view = static_cast<BrowserView*>(browser()->window());
+  auto* browser_frame = static_cast<BraveBrowserFrame*>(browser_view->frame());
+  auto* theme_service =
+      ThemeServiceFactory::GetForProfile(browser()->profile());
+  ASSERT_TRUE(theme_service);
+  theme_service->SetBrowserColorScheme(ThemeService::BrowserColorScheme::kDark);
```
bridiver, line 53: I'm confused about why this is called in the test, isn't this enabled by setting the pref? Why do we need to call this here?

browser/ui/views/frame/brave_browser_frame.cc
```diff
@@ -74,3 +77,17 @@ void BraveBrowserFrame::SetTabDragKind(TabDragKind kind) {
...
   }
 }
+
+ui::ColorProviderKey BraveBrowserFrame::GetColorProviderKey() const {
+  auto key = BrowserFrame::GetColorProviderKey();
+
+  if (base::FeatureList::IsEnabled(darker_theme::features::kBraveDarkerTheme)) {
+    if (key.color_mode == ui::ColorProviderKey::ColorMode::kDark &&
+        browser_view_->browser()->profile()->GetPrefs()->GetBoolean(
+            darker_theme::prefs::kBraveDarkerMode)) {
```
bridiver, line 87: does this not update when if the user toggles the pref?

browser/ui/darker_theme/darker_theme_browsertest.cc
sangwoo108: Yes we need this. This isn't about "Darker" preference. We're setting "Dark" theme, and "Darker" preference will make the "Dark theme" darker.

browser/ui/darker_theme/darker_theme_browsertest.cc
sangwoo108: Added test

browser/ui/views/frame/brave_browser_frame.cc
sangwoo108: I'm not sure if I understand what you mean here, but whenever preference is turned on/off, it'll be notified. And then this method will be invoked from the UI side to refresh ColorProviderKey.

bridiver: I see, through `BraveThemeService::OnDarkerThemePrefChanged`?

browser/ui/darker_theme/darker_theme_browsertest.cc
bridiver: I think I see now, so the only purpose of this pref is to make dark mode darker?

sangwoo108: Yes, exactly

browser/ui/views/frame/brave_browser_frame.cc
sangwoo108: Yes that's correct

bridiver (approved):

PR #30893, opened 2025-08-28
Disabled WebShare in Tor windows on all platforms.

boocmp:
Resolves https://github.com/brave/brave-browser/issues/48770

### DIFF
--- a/chromium_src/third_party/blink/renderer/core/exported/web_view_impl.cc
+++ b/chromium_src/third_party/blink/renderer/core/exported/web_view_impl.cc
@@ -15,10 +15,8 @@
   {                                                        \
     RuntimeEnabledFeatures::SetBraveIsInTorContextEnabled( \
         prefs.is_tor_window);                              \
-    if constexpr (BUILDFLAG(IS_MAC)) {                     \
-      if (prefs.is_tor_window) {                           \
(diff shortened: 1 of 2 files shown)


chromium_src/third_party/blink/renderer/core/exported/web_view_impl.cc
```diff
@@ -15,10 +15,8 @@
   {                                                        \
     RuntimeEnabledFeatures::SetBraveIsInTorContextEnabled( \
         prefs.is_tor_window);                              \
-    if constexpr (BUILDFLAG(IS_MAC)) {                     \
-      if (prefs.is_tor_window) {                           \
-        RuntimeEnabledFeatures::SetWebShareEnabled(false); \
-      }                                                    \
+    if (prefs.is_tor_window) {                             \
```
boocmp, line 18: @fmarier We can check Window version here and disable the web share only on 11 or newer, is it important?

fmarier: No, let's just disable it entirely for all Tor windows. It's not really worth keeping this feature in that mode.

fmarier (approved):

darkdh (approved):

PR #30902, opened 2025-08-28
[CodeHealth] Various `Wexit-time-destructors` fixes

cdesouza-chromium:
This PR concentrates primarily on fixing exit-time destructors on tests,
and for cases that are low-hanging fruits. With tests there is less of a
concern with reinstantiations and copies in general.

Resolves https://github.com/brave/brave-browser/issues/48790

### DIFF
--- a/browser/brave_wallet/solana_provider_renderer_browsertest.cc
+++ b/browser/brave_wallet/solana_provider_renderer_browsertest.cc
@@ -3,14 +3,17 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this file,
  * You can obtain one at https://mozilla.org/MPL/2.0/. */
 
+#include <array>
 #include <optional>
 
 #include "base/containers/flat_map.h"
+#include "base/containers/to_vector.h"
 #include "base/feature_list.h"
 #include "base/memory/weak_ptr.h"
 #include "base/path_service.h"
 #include "base/strings/strcat.h"
 #include "base/strings/string_number_conversions.h"
+#include "base/strings/string_util.h"
 #include "base/test/values_test_util.h"
 #include "brave/browser/brave_content_browser_client.h"
 #include "brave/browser/brave_wallet/brave_wallet_service_factory.h"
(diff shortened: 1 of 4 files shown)


browser/metrics/variations_browsertest.cc
```diff
@@ -38,13 +38,12 @@ const SignedSeedData& GetBraveSignedSeedData() {
...
-                                                 kBase64UncompressedData,
-                                                 kBase64CompressedData,
-                                                 kBase64Signature,
-                                                 /*in_compressed_data=*/{},
-                                                 /*in_compressed_data_size=*/0};
-  return kBraveTestSeedData;
+  return {study_names,
+          kBase64UncompressedData,
+          kBase64CompressedData,
+          kBase64Signature,
```
goodov, line 44: for this to work all `constexpr char` vars in this function should be marked `static`, otherwise it'll trip sanitizer.

browser/brave_wallet/solana_provider_renderer_browsertest.cc
```diff
@@ -128,11 +130,11 @@ constexpr char OnAccountChangedScript[] =
...
+std::string VectorToArrayString(base::span<const uint8_t> data) {
   std::string result;
-  for (size_t i = 0; i < vec.size(); ++i) {
-    base::StrAppend(&result, {base::NumberToString(vec[i])});
-    if (i != vec.size() - 1) {
+  for (size_t i = 0; i < data.size(); ++i) {
+    base::StrAppend(&result, {base::NumberToString(data[i])});
+    if (i != data.size() - 1) {
       base::StrAppend(&result, {", "});
     }
```
goodov: nit: this can be rewritten with a bool that prepends `", "` if it's not the first item. This way you can replace for-loop with a simplified ranged version.

or maybe just this:
```cpp
return base::JoinString(
    base::ToVector(data, [](int value) { return base::NumberToString(value); }), 
    ", ");
```

goodov (approved):

browser/metrics/variations_browsertest.cc
cdesouza-chromium: Fixed.

browser/brave_wallet/solana_provider_renderer_browsertest.cc
cdesouza-chromium: Fixed. I'm going with the `JoinString` version because it is easier to the reader.

cdesouza-chromium:
Thanks @goodov for the helpful review.

SergeyZhukovsky (approved):
++

PR #30903, opened 2025-08-28
Fix legacy_google_update_appid on macOS

mherrmann:
We should mirror what upstream does and set the above GN arg to `keystone_bundle_identifier` on macOS.

No QA / tests required because the affected functionality isn't live yet

Resolves https://github.com/brave/brave-browser/issues/48870.


cdesouza-chromium (approved):
LGTM.

PR #30913, opened 2025-08-28
[AI Chat]: Add Model Capability Labels

Douglashdaniel:
## Description 

Add `Capability` labels to the models in the `Model Selector`

Resolves <https://github.com/brave/brave-browser/issues/48730>

## Test Plan:
1. Open `Leo AI` and start a new conversation.
2. Open the `Model Selector` in the `Input Box`
3. It should now display a list of `Capabilities` for each model instead of the `Description`
4. The `Automatic` model will still have a `Description`
5. All `Custom` models will still have their `modelRequestName`

Screenshot 69 | Screenshot 70
--|--

### DIFF
--- a/components/resources/ai_chat_ui_strings.grdp
+++ b/components/resources/ai_chat_ui_strings.grdp
@@ -439,25 +439,25 @@
     Enable
   </message>
   <message name="IDS_CHAT_UI_CHAT_BASIC_SUBTITLE" desc="A description for default chat model" formatter_data="webui=AiChat">
-    General purpose chat
+    Fast • Search
   </message>
   <message name="IDS_CHAT_UI_CHAT_CLAUDE_INSTANT_SUBTITLE" desc="A description for claude-instant-v1 model" formatter_data="webui=AiChat">
     Conversation and text generation
   </message>
   <message name="IDS_CHAT_UI_CHAT_CLAUDE_HAIKU_SUBTITLE" desc="A description for claude-3-haiku model" formatter_data="webui=AiChat">
-    Lightning fast chat
+    Fast • Search • Tools
   </message>
   <message name="IDS_CHAT_UI_CHAT_CLAUDE_SONNET_SUBTITLE" desc="A description for claude-3-sonnet model" formatter_data="webui=AiChat">
-    Balanced speed and intelligence
+    Search • Vision • Tools
   </message>
   <message name="IDS_CHAT_UI_CHAT_QWEN_SUBTITLE" desc="A description for qwen model" formatter_data="webui=AiChat">
-      General purpose chat, coding, and more
+    Fast • Search
   </message>
   <message name="IDS_CHAT_UI_CHAT_GEMMA_SUBTITLE" desc="A description for gemma model" formatter_data="webui=AiChat">
-    General purpose text and image understanding
+    Search • Vision
   </message>
   <message name="IDS_CHAT_UI_CHAT_DEEPSEEK_R1_SUBTITLE" desc="A description for deepseek-r1 model" formatter_data="webui=AiChat">
-    Reasoning model
+    Search • Thinking
   </message>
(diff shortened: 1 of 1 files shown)


components/ai_chat/core/common/mojom/ai_chat.mojom
```diff
@@ -56,6 +56,14 @@ enum ModelAccess {
   PREMIUM,
 };
 
+enum ModelCapabilities {
+  FAST,
+  SEARCH,
+  THINKING,
+  VISION,
+  TOOLS,
```
Douglashdaniel: On the fence about having `VISION` and `TOOLS` here since we already have `supports_tools` and `supports_vision`.

Happy to remove if we think it would be better to just use those variables.

petemill: I think it's fine to keep them here as whilst some models may support tools, others might be better and we might want to have them stand out as such?

petemill: And if we change to using this to be the source of truth for actual tool and vision support, then we still want to leave them in and eventually migrate over.

components/ai_chat/resources/page/components/model_selector/index.tsx
```diff
@@ -103,11 +126,9 @@ export function ModelSelector() {
...
               <div>{model.displayName}</div>
               <p className={styles.modelSubtitle}>
-                {getLocale(
-                  `CHAT_UI_${model.key
-                    .toUpperCase()
-                    .replaceAll('-', '_')}_SUBTITLE`,
-                )}
+                {model.key === 'chat-automatic'
+                  ? getLocale(S.CHAT_UI_CHAT_AUTOMATIC_SUBTITLE)
+                  : getModelCapabilities(model.capabilities)}
```
petemill: What if we just have a `string model_description` field in `struct Model` and, if this has a value, then we use that instead of the capabilities list? Just trying to reduce and remove any hardcoding of model keys in the frontend. Same with icon by the way that I notice we have `getModelIcon` for: we could just have `model.icon_name`. This preserves the ability for all this data to come from the server and also to be rendered elsewhere.

components/ai_chat/core/common/mojom/ai_chat.mojom
```diff
@@ -56,6 +56,14 @@ enum ModelAccess {
   PREMIUM,
 };
 
+enum ModelCapabilities {
```
petemill: The name should be singular

components/ai_chat/core/common/mojom/ai_chat.mojom
petemill: nit: the values should be cased like `kMyThing`. I've only just learned that they are meant to match C++ style in mojom. But since this is an existing issue we can add to a code health follow-up for this file and references (although this file already has 2 different formats!)

Douglashdaniel: Sounds good to me!

components/ai_chat/resources/page/components/model_selector/index.tsx
Douglashdaniel: Actually better yet, we can check if a model has `capabilities`, if not we fallback to the `description`. Then there is no need for an additional field in the `struct Model`. What do you think?

components/ai_chat/core/common/mojom/ai_chat.mojom
Douglashdaniel: Changed to `ModelCapability`

components/ai_chat/core/common/mojom/ai_chat.mojom
Douglashdaniel: Changed to `kMyThing` formatting!

mkarolin (approved):
`strings`++

PR #30916, opened 2025-08-28
make Email Aliases settings page UI visible again

arthuredelstein:
(remains behind flag)

Resolves https://github.com/brave/brave-browser/issues/48782


browser/resources/settings/email_aliases_page/email_aliases_page_index.html
fallaciousreasoning, line 1: hey so I don't think we need this index page because we only have one route here.

browser/resources/settings/email_aliases_page/email_aliases_page_index.html
```diff
@@ -0,0 +1,9 @@
+<style include="settings-shared"></style>
+
+<div class="cr-centered-card-container">
+  <settings-subpage page-title="Email Aliases">
```
fallaciousreasoning: if email aliases is a `settings-subpage` that should be inside `settings-email-aliases-page` now

browser/resources/settings/email_aliases_page/email_aliases_page_index.ts
fallaciousreasoning, line 1: nit remove

browser/resources/settings/brave_routes.ts
```diff
@@ -156,7 +156,9 @@ export default function addBraveRoutes(r: Partial<SettingsRoutes>) {
       '[Settings] Could not move autofill route to advanced route', r)
   }
   if (loadTimeData.getBoolean('isEmailAliasesEnabled') && r.AUTOFILL) {
-    r.EMAIL_ALIASES = r.AUTOFILL.createChild('/email-aliases')
+    r.EMAIL_ALIASES = r.BASIC.createSection('/email-aliases', 'email-aliases')
```
fallaciousreasoning: this should probably still be a child of `AUTOFILL` - we should add this to the `cr-view-manager` in chrome/browser/resources/settings/autofill_page/autofill_page_index.html.

See brave/browser/resources/settings/brave_overrides/appearance_page_index.ts for an example of how to do this :smile:

browser/resources/settings/brave_overrides/settings_main.ts
```diff
@@ -161,5 +162,21 @@ RegisterPolymerTemplateModifications({
...
+            <template is="dom-if" if="[[renderPlugin_(
+          routes_.EMAIL_ALIASES, lastRoute_, inSearchMode_)]]">
+              <settings-email-aliases-page-index
+                prefs="{{prefs}}"
+                in-search-mode="[[inSearchMode_]]">
+              </settings-email-aliases-page-index>
+            </template>
+          </div>
+        </template>
+      `)
```
fallaciousreasoning: nit: this isn't a top level page so we probably don't want this here

browser/resources/settings/email_aliases_page/email_aliases_page_index.html
arthuredelstein: Removed; now I have `email_aliases_section.html`

browser/resources/settings/email_aliases_page/email_aliases_page_index.html
arthuredelstein: Yes, that's now defined in `email_aliases_section.html`

browser/resources/settings/email_aliases_page/email_aliases_page_index.ts
arthuredelstein: Removed. Now we have `email_aliases_section.ts`

browser/resources/settings/brave_routes.ts
arthuredelstein: That worked! I am no longer changing this file.

browser/resources/settings/brave_overrides/settings_main.ts
arthuredelstein: Yes, now it is instead injected in `autofill_page.ts`.

arthuredelstein:
I force-pushed a new version of this patch because it almost entirely different. (Also rebased to master.)

browser/resources/settings/brave_overrides/autofill_page.ts
```diff
@@ -6,22 +6,40 @@
...
+    prototype.currentRouteChanged =
+      function (newRoute: Route, oldRoute?: Route) {
+      if (typeof original === 'function') {
+        original.call(this, newRoute, oldRoute)
+      }
+
+      if (newRoute === routes.EMAIL_ALIASES) {
+        this.$.viewManager.switchView('email-aliases', 'no-animation',
+                                      'no-animation')
+      }
```
fallaciousreasoning: Hey, if we match this route lets return and not call the super function. That way if upstream adds a default case this won't break.
```suggestion
      if (newRoute === routes.EMAIL_ALIASES) {
        this.$.viewManager.switchView('email-aliases', 'no-animation',
                                      'no-animation')
        return
      }
      original.call(this, newRoute, oldRoute)
```

browser/resources/settings/brave_overrides/autofill_page.ts
```diff
@@ -6,22 +6,40 @@
...
 )
 
+RegisterPolymerPrototypeModification({
+  'settings-autofill-page-index': (prototype) => {
+    const original = prototype.currentRouteChanged
+    prototype.currentRouteChanged =
+      function (newRoute: Route, oldRoute?: Route) {
+      if (typeof original === 'function') {
+        original.call(this, newRoute, oldRoute)
+      }
```
fallaciousreasoning: do we actually need this check?

browser/resources/settings/email_aliases_page/email_aliases_section.html
fallaciousreasoning, line 1: hey, let's not add this file just for wrapping `settings-email-aliases-page` in a `settings-subpage` - let's just make the `settings-subpage` wrap the content inside `settings-subpage` (either in then React or in the web component)

browser/resources/settings/email_aliases_page/email_aliases_section.ts
fallaciousreasoning, line 1: ditto, let's remove this

browser/resources/settings/email_aliases_page/email_aliases_section.html
arthuredelstein: done

browser/resources/settings/email_aliases_page/email_aliases_section.ts
arthuredelstein: done

browser/resources/settings/brave_overrides/autofill_page.ts
arthuredelstein: removed

browser/resources/settings/brave_overrides/autofill_page.ts
```diff
@@ -6,22 +6,45 @@
...
+    prototype.currentRouteChanged =
+      function (newRoute: Route, oldRoute?: Route) {
+      if (newRoute === routes.EMAIL_ALIASES) {
+        this.$.viewManager.switchView('email-aliases', 'no-animation',
+                                      'no-animation')
+        return
+      }
+      originalCurrentRouteChanged.call(this, newRoute, oldRoute)
+    }
+    const originalGetAssociated = prototype.getAssociatedControlFor
```
fallaciousreasoning: `getAssociatedControlFor` needs to be implemented on the parent view for `settings-email-aliases-page`. In this case its `settings-autofill-page`.

Basically its for saying what control will open the email aliases page so we can highlight that during search

browser/resources/settings/brave_overrides/autofill_page.ts
```diff
@@ -37,24 +60,21 @@ RegisterPolymerTemplateModifications({
...
+      return
+    }
+    const viewManager = templateContent.querySelector('cr-view-manager')
+    if (!viewManager) {
+      throw new Error(
+        '[Settings] Unable to find cr-view-manager on autofill-page')
+    }
+    viewManager.append(html`
+      <settings-subpage page-title="Email Aliases" slot="view"
+         id="email-aliases" data-parent-view-id="parent" prefs="{{prefs}}">
```
fallaciousreasoning: Hey, unfortunately its not safe to render nested elements in the cr-view-manager. It expects whatever is slotted to be an implementation of a page (even though it mostly works, it breaks things like internal nesting and `getAssociatedControlFor`).

Either add the wrapping subpage in `brave/browser/resources/settings/email_aliases_page/email_aliases_page.ts` or in `brave/browser/resources/settings/email_aliases_page/email_aliases.tsx`

arthuredelstein: done

browser/resources/settings/brave_overrides/autofill_page.ts
arthuredelstein: Ah, I get it now. We don't currently have a control for that in master. Rather than implementing that here, I opened an issue to do that: https://github.com/brave/brave-browser/issues/48868

browser/resources/settings/email_aliases_page/email_aliases_page.ts
```diff
@@ -13,10 +14,16 @@ import { loadTimeData } from "chrome://resources/js/load_time_data.js"
 
 class EmailAliasesPage extends HTMLElement {
   connectedCallback() {
-    this.attachShadow({ mode: 'open' })
+    const subpage = document.createElement('settings-subpage')
+    this.appendChild(subpage)
```
fallaciousreasoning: Hmmm I think this isn't quite what you want - you're creating a Light Dom `settings-subpage` and then a shadowRoot on its child.

I think you probably want:

```ts
const shadow = this.attachShadow({ mode: 'open' });
const subpage = document.createElement('settings-subpage');
// set subpage attributes
// ...

shadow.appendChild(subpage);

import('/email_aliases.bundle.js' as any)
    .then(() => {
      (window as any).mountEmailAliases(subpage)
    })
```

arthuredelstein: I attached a shadow root under EmailAliasesPage and that seems to be OK: https://github.com/brave/brave-core/pull/30916/commits/0405587f6d81e5cbe1d6400ef84a8bd3c4a3cc7e

It seems I need to keep the the extra div (reactHost) and attach another ShadowRoot there for the layout to be correct.

arthuredelstein: @fallaciousreasoning pointed out I could simplify by mounting on the settings-subpage and use ` <StyleSheetManager target={at.getRootNode() as ShadowRoot}>`: https://github.com/brave/brave-core/pull/30916/commits/243ef0c5f9eaa9b2a755a6140df9dfd6d0f76152

fallaciousreasoning (approved):
lgtm thanks @arthuredelstein !

PR #30918, opened 2025-08-28
[CodeHealth]  Use `base::ElapsedTimer` where applicable

DJAndries:
Resolves https://github.com/brave/brave-browser/issues/48809

### DIFF
--- a/browser/misc_metrics/usage_clock.cc
+++ b/browser/misc_metrics/usage_clock.cc
@@ -30,7 +30,7 @@ UsageClock::~UsageClock() {
 }
 
 base::TimeDelta UsageClock::GetTotalUsageTime() const {
-  auto elapsed_time_in_session = usage_time_in_completed_sessions_;
+  base::TimeDelta elapsed_time_in_session = usage_time_in_completed_sessions_;
   if (current_session_elapsed_timer_) {
     elapsed_time_in_session += current_session_elapsed_timer_->Elapsed();
   }


browser/net/brave_ad_block_tp_network_delegate_helper.cc
```diff
@@ -104,7 +105,7 @@ class AdblockCnameResolveHostClient : public network::mojom::ResolveHostClient {
     if (secure_dns_config.mode() == net::SecureDnsMode::kSecure)
       optional_parameters->source = net::HostResolverSource::DNS;
 
-    start_time_ = base::TimeTicks::Now();
+    elapsed_timer_ = {};
```
cdesouza-chromium, line 108: nit: I think this could be a bit more explicit about being an optional being engaged with:

```
elapsed_timer_.emplace()
```

browser/misc_metrics/usage_clock.cc
```diff
@@ -39,22 +30,22 @@ UsageClock::~UsageClock() {
 }
 
 base::TimeDelta UsageClock::GetTotalUsageTime() const {
-  base::TimeDelta elapsed_time_in_session = usage_time_in_completed_sessions_;
-  if (IsInUse()) {
-    elapsed_time_in_session += NowTicks() - current_usage_session_start_time_;
+  auto elapsed_time_in_session = usage_time_in_completed_sessions_;
```
cdesouza-chromium: nit: The style guide favours the use of `base::TimeDelta`, rather than `auto` here.

> Do not use it merely to avoid the inconvenience of writing an explicit type.
https://google.github.io/styleguide/cppguide.html#Type_deduction

cdesouza-chromium:
Thanks @DJAndries for taking the time to do this PR. This is really good work, and this code is now clearly better with this change.

browser/misc_metrics/usage_clock.cc
DJAndries: 655df8a

PR #30922, opened 2025-08-29
Hide VerticalTabStripRegionView when hiding completely

sangwoo108:
This will prevent the vertical tab strip from being shown partially when it is supposed to be completely hidden.

Resolves https://github.com/brave/brave-browser/issues/47520

### DIFF
--- a/browser/ui/views/frame/vertical_tabs/vertical_tab_strip_region_view.cc
+++ b/browser/ui/views/frame/vertical_tabs/vertical_tab_strip_region_view.cc
@@ -703,11 +703,16 @@ void VerticalTabStripRegionView::SetState(State state) {
   if (gfx::Animation::ShouldRenderRichAnimation()) {
     state_ == State::kCollapsed ? width_animation_.Hide()
                                 : width_animation_.Show();
+  } else if (state_ == State::kCollapsed) {
+    // Call the callback immediately if no animation.
+    OnCollapseAnimationEnded();
   }
 
   if (!GetVisible() && state_ != State::kCollapsed) {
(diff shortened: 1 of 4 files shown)


browser/ui/views/frame/vertical_tabs/vertical_tab_strip_region_view.cc
```diff
@@ -1368,6 +1372,19 @@ views::LabelButton& VerticalTabStripRegionView::GetToggleButtonForTesting() {
...
+void VerticalTabStripRegionView::OnCollapseAnimationEnded() {
+  CHECK_EQ(state_, State::kCollapsed);
+
+  if (IsFloatingEnabledForBrowserFullscreen() ||
+      base::FeatureList::IsEnabled(
+          tabs::features::kBraveVerticalTabHideCompletely)) {
+    // When the animation ends, we should hide the vertical tab strip as we
+    // don't want the tabstrip to be visible partially. This view only takes a
+    // little width and watches mouse movement to expand itself.
+    SetVisible(false);
```
simonhong, line 1384: Q) when it's invisible, it still get mouse entered event?

sangwoo108: Yeah, when I tested, it was expanded well. I guess MouseWatcher works even when the view is invisible. Let me double check.

sangwoo108: Yes, it works

simonhong (approved):
++

PR #30923, opened 2025-08-29
[CodeHealth] Replace `base::WrapUnique` with `std::make_unique` where applicable

DJAndries:
Resolves https://github.com/brave/brave-browser/issues/48816

Other instances of `base::WrapUnique(new X())` call private constructors, so they cannot be converted to `std::make_unique`.

### DIFF
--- a/browser/ui/webui/new_tab_page/brave_new_tab_ui.cc
+++ b/browser/ui/webui/new_tab_page/brave_new_tab_ui.cc
@@ -143,8 +143,7 @@ BraveNewTabUI::BraveNewTabUI(
 
   web_ui->AddMessageHandler(base::WrapUnique(
       BraveNewTabMessageHandler::Create(source, profile, was_restored)));
-  web_ui->AddMessageHandler(
-      base::WrapUnique(new TopSitesMessageHandler(profile)));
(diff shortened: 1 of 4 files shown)


cdesouza-chromium:
> Other instances of base::WrapUnique(new X()) call private constructors, so they cannot be converted to std::make_unique.

@DJAndries To clarify, when a certain class uses a private constructor, which prevent the use of `std::make_unique`, we should always consider if we can make that constructor public, but limit who can call it with the `PassKey` idiom https://source.chromium.org/chromium/chromium/src/+/main:base/types/pass_key.h.

cdesouza-chromium (approved):
LGTM

DJAndries:
> > Other instances of base::WrapUnique(new X()) call private constructors, so they cannot be converted to std::make_unique.
> 
> @DJAndries To clarify, when a certain class uses a private constructor, which prevent the use of `std::make_unique`, we should always consider if we can make that constructor public, but limit who can call it with the `PassKey` idiom https://source.chromium.org/chromium/chromium/src/+/main:base/types/pass_key.h.

Right, a "part 2" PR will warrant more stringent review from other folks. Just knocking out the easy cases in this PR.

PR #30924, opened 2025-08-29
Fixed DCHECK failure from InfoBarView::Layout()

simonhong:
Resolves fix https://github.com/brave/brave-browser/issues/48814

Unlike upstream, Brave adds child views directly to the infobar instead of using the content_container_.
This causes the close button to not always be the last child in the view hierarchy, which triggers layout assertion failures.

Fixed by overriding `ViewHierarchyChanged()` in `BraveConfirmInfoBar` to ensure the close button is always reordered as the last child when views are added.

This upstream changes affected us - https://chromium-review.googlesource.com/c/chromium/src/+/6632793
Filed f/u issue for adding our children to `content_container_`. - https://github.com/brave/brave-browser/issues/48822

TEST=`BraveConfirmInfoBarTest.CloseButtonOrderTest`

### DIFF
--- a/browser/ui/views/infobars/brave_confirm_infobar.cc
+++ b/browser/ui/views/infobars/brave_confirm_infobar.cc
@@ -187,6 +187,22 @@ void BraveConfirmInfoBar::Layout(PassKey) {
   link_->SetPosition(gfx::Point(GetEndX() - link_->width(), OffsetY(link_)));
 }
 
+void BraveConfirmInfoBar::ViewHierarchyChanged(
+    const views::ViewHierarchyChangedDetails& details) {
+  InfoBarView::ViewHierarchyChanged(details);
+
+  // Upstream doesnt need this reorder because all other children
+  // except icon and close button are in |content_container_| that lives
+  // between both. It guarantees that close button is always the last
+  // children. However, we just add children directly instead of
+  // that container.
+  // TODO(https://github.com/brave/brave-browser/issues/48822): Add our children
+  // into content container.
+  if (details.is_add && (details.child == this) && close_button_) {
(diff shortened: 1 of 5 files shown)


browser/ui/views/infobars/brave_confirm_infobar.cc
```diff
@@ -187,6 +187,22 @@ void BraveConfirmInfoBar::Layout(PassKey) {
...
+
+  // Upstream doesnt need this reorder because all other children
+  // except icon and close button are in |content_container_| that lives
+  // between both. It guarantees that close button is always the last
+  // children. However, we just add children directly instead of
+  // that container.
+  // TODO(https://github.com/brave/brave-browser/issues/48822): Add our children
+  // into content container.
+  if (details.is_add && (details.child == this) && close_button_) {
+    ReorderChildView(close_button_, children().size());
```
sangwoo108: How about reordering close button at the end of constructor, where all child views are attached?

simonhong: Can't do it because we have another subclass of `BraveConfirmInforBar`.

simonhong: Oops, sorry. I was confused. we don't have its subclass :) 
We could do it that way.

simonhong: Fixed.

sangwoo108 (approved):
LGTM!

browser/ui/views/infobars/brave_confirm_infobar.cc
sangwoo108: Thanks :)

PR #30934, opened 2025-08-29
Improve Picture-in-picture transition algorithm

simoarpe:
## Resolves https://github.com/brave/brave-browser/issues/48684 - Wrong PiP viewport

This PR improves the picture in picture transition and fixes some issues that could be observed on release builds only.

- To make sure the viewport is correctly calculated when switching into the PIP window the workaround adopted by upstream code is to post a delayed handler (500 ms) where the picture in picture parameters are reset, this trigger also, under the hood, a recalculation of the rect bounds that sometimes may be calculated too early and causing the pip window to look like it moves to the wrong part of the screen and partially clipped before snapping to its normal place.

- To resume the media stream after the PIP transition (`mediaSession.resume()`) the Brave activity now uses a boolean called `mResumeMediaSession` that will trigger the resume method in the appropriate callback.
 Two caveats here:
    1. The boolean flag `mResumeMediaSession` needs to be saved in the bundle and restored as the activity _might_ be destroyed and recreate after calling `enterPictureInPictureMode`. Normally `enterPictureInPictureMode` just resizes the activity but a recreation can happen.
    2. Because of the Brave Activity reference for calling `resumeMediaSession(boolean)` method, I had to move the `BraveYouTubeScriptInjectorNativeHelper` class out of the `youtube_script_injector` and include it in the main app layer.
 - Resuming the media session at the end of the transition always guarantee that the navigation controls are rendered correctly.

## Demo

[Screencast from 29-08-2025 17:44:17.webm](https://github.com/user-attachments/assets/6e6f7373-604f-419d-b53f-ed41ca199162)

### DIFF
--- a/android/java/org/chromium/chrome/browser/app/BraveActivity.java
+++ b/android/java/org/chromium/chrome/browser/app/BraveActivity.java
@@ -8,6 +8,7 @@ package org.chromium.chrome.browser.app;
 import android.app.Activity;
 import android.app.NotificationChannel;
 import android.app.NotificationManager;
+import android.app.PictureInPictureParams;
 import android.content.Context;
 import android.content.DialogInterface;
 import android.content.Intent;
@@ -307,6 +308,7 @@ public abstract class BraveActivity extends ChromeActivity
     private static final List<String> sYandexRegions =
             Arrays.asList("AM", "AZ", "BY", "KG", "KZ", "MD", "RU", "TJ", "TM", "UZ");
 
+    private static final int PIP_UPDATE_DELAY_MS = 500;
     private boolean mIsVerification;
     public boolean mIsDeepLink;
     private BraveWalletService mBraveWalletService;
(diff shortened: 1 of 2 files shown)


android/java/org/chromium/chrome/browser/app/BraveActivity.java
```diff
@@ -305,6 +308,7 @@ public abstract class BraveActivity extends ChromeActivity
     private static final List<String> sYandexRegions =
             Arrays.asList("AM", "AZ", "BY", "KG", "KZ", "MD", "RU", "TJ", "TM", "UZ");
 
+    private static final int PIP_UPDATE_DELAY_MS = 500;
```
samartnik, line 311: Where this number comes from? Do they have the same delay in the upstream?

simoarpe: Yes, correct. I've added a comment below that also references the same Java class so the code can be checked further if required.

simoarpe: More details in the PR description as well.

kjozwiak:
Verification **PASSED** on `Samsung S25 Ultra` running `Android 15` using the following build(s):

```
Brave | 1.84.15 Chromium: 140.0.7339.41 (Official Build) canary (64-bit)
--- | ---
Revision | c5a3f49fc32bd78f0b9d3180b6c54429738b9143
```

Using the STR/Cases outlined via https://github.com/brave/brave-browser/issues/48674#issue-3352610240, https://github.com/brave/brave-browser/issues/48677#issue-3352768415 & https://github.com/brave/brave-browser/issues/48684#issue-3353059271, ensured that:

* ensured the `PiP` navigation controls were being displayed when going into `PiP` mode
* ensured `playback` continued when going into `PiP` mode
* ensured that the video always appeared correct within the `PiP` viewport

https://github.com/user-attachments/assets/3bae2926-075a-4a55-9280-43a724ac679c

PR #30949, opened 2025-08-31
`endpoint_client`: consume requests, use `HttpRequestHeaders`, refine concepts

szilardszaloki:
Resolves https://github.com/brave/brave-browser/issues/48863.

- `Send()` now consumes requests by value (JSON schema compiler types are move-only)
- switch to `net::HttpRequestHeaders` instead of `base::flat_map<std::string, std::string>`
- `ToValue<>` checks `base::WriteJson(t.ToValue())`
- add precise concept docs and file-level comment in `concepts.h`
- use `net::HttpRequestHeaders::kPostMethod` instead of hardcoded "POST"


fmarier (approved):
Looks good to me, but let's wait until @bsclifton 's had a chance to review the `components/brave_account/endpoint_client/client.h` changes closely.

bsclifton (approved):
LGTM!

PR #30951, opened 2025-09-01
Close only active tab from split tab

simonhong:
Resolves https://github.com/brave/brave-browser/issues/48406

In SideBySide, two tabs in one split tab are in selected state.
As `chrome::CloseTab()` closes selected tabs, both tabs are closed together for tab closing shortcut(ex, `cmd + w`).
We want to close only active tab not both tab. So, closing active tab only if current selected tabs are only two and both
are from same split tab.

Handled tab ctx menu's closed tab in a same way.

TEST=SplitViewCommonBrowserTest.SplitViewCloseTabTest
          BraveTabContextMenuContentsTest.SplitViewMenuCustomizationTest

### DIFF
--- a/browser/ui/tabs/brave_tab_strip_model.cc
+++ b/browser/ui/tabs/brave_tab_strip_model.cc
@@ -132,3 +132,24 @@ void BraveTabStripModel::SetCustomTitleForTab(
 
   NotifyTabChanged(tab_interface, TabChangeType::kAll);
 }
+
+void BraveTabStripModel::CloseSelectedTabsFromBrowserCommands() {
+  auto selected_indices = selection_model().selected_indices();
+  if (selected_indices.size() != 2) {
+    return CloseSelectedTabs();
+  }
+
+  // If selected tabs only include two tabs from same split tab,
+  // close active tab only.
+  auto first_tab = selected_indices.begin();
+  auto first_tab_split = GetSplitForTab(*first_tab);
+  if (first_tab_split.has_value() &&
+      first_tab_split == GetSplitForTab(*(first_tab + 1))) {
+    CloseWebContentsAt(active_index(),
+                       TabCloseTypes::CLOSE_USER_GESTURE |
+                           TabCloseTypes::CLOSE_CREATE_HISTORICAL_TAB);
+    return;
+  }
+
+  return CloseSelectedTabs();
+}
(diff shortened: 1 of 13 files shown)


browser/ui/views/tabs/brave_tab_context_menu_contents_browsertest.cc
```diff
@@ -186,19 +185,7 @@ IN_PROC_BROWSER_TEST_F(BraveTabContextMenuContentsTest,
   EXPECT_EQ(incognito_browser->tab_strip_model()->count(), incognito_tab_count);
 }
 
-class BraveTabContextMenuWithSideBySideTest
-    : public BraveTabContextMenuContentsTest {
- public:
-  BraveTabContextMenuWithSideBySideTest() {
-    scoped_features_.InitWithFeatures(
-        /*enabled_features*/ {features::kSideBySide}, {});
```
simonhong, line 194: Deleted as `SideBySide` is enabled by default.

browser/ui/tabs/brave_tab_strip_model.cc
```diff
@@ -132,3 +132,24 @@ void BraveTabStripModel::SetCustomTitleForTab(
...
 }
+
+void BraveTabStripModel::CloseSelectedTabsWithSplitView() {
+  auto selected_indices = selection_model().selected_indices();
+  if (selected_indices.size() != 2) {
+    return CloseSelectedTabs();
+  }
+
+  // If selected tabs only include two tabs from same split tab,
+  // close active tab only.
```
sangwoo108, line 143: Q. What happens if a user selects both of split tabs via Ctrl+click or shift+click and then close? Is it still closing only the active tab?

simonhong: Yes, it'll close only active tab in that case.
Actually, it's same condition with normal state as both tabs in active split tab are always in selected state.
I assume that the majority of close are single active tab closing when split view is activated.
Maybe most(all) of users expects active tab closing instead of closing all.

sangwoo108: Well, I'm not so sure about it. If a user do intentionally 1. select split tabs 2. open context menu and close, they might expect that all tabs would be closed as they selected it explicitly. Maybe it could be better not selecting both split tabs in the first place - but this could make it difficult to drag split tabs simultaneously. So I'll leave it up to your judgement.

simonhong: I think changing selection state could make many side effect such as dragging.
Maybe there is some user behave like that(select explicitely) but it would be very very small number of users than just closing active tab. I think this could be left as known issue.

mkarolin (approved):
`chromium_src`++

PR #30960, opened 2025-09-01
[iOS] - Fix bookmarks search showing duplicate items

Brandon-T:
- Fix bookmarks search duplication by just casting the results instead of getting matching nodes. 
- BookmarkNode inherits from TitleUrlNode so it's okay to downcast.

Resolves https://github.com/brave/brave-browser/issues/48873

### DIFF
--- a/ios/browser/api/bookmarks/brave_bookmarks_api.mm
+++ b/ios/browser/api/bookmarks/brave_bookmarks_api.mm
@@ -657,15 +657,14 @@
                 : query_parser::MatchingAlgorithm::DEFAULT);
 
     NSMutableArray<IOSBookmarkNode*>* nodes = [[NSMutableArray alloc] init];
+
     for (const auto& result : results) {
-      for (const bookmarks::BookmarkNode* bookmark :
-           bookmarks_api->bookmark_model_->GetNodesByURL(
-               result.node->GetTitledUrlNodeUrl())) {
-        IOSBookmarkNode* node = [[IOSBookmarkNode alloc]
-            initWithNode:bookmark
-                   model:bookmarks_api->bookmark_model_];
-        [nodes addObject:node];
-      }
+      const bookmarks::BookmarkNode* bookmark_node =
+          static_cast<const bookmarks::BookmarkNode*>(result.node.get());
+      IOSBookmarkNode* node =
+          [[IOSBookmarkNode alloc] initWithNode:bookmark_node
+                                          model:bookmarks_api->bookmark_model_];
+      [nodes addObject:node];
     }
     completion(nodes);
   };


ios/browser/api/bookmarks/brave_bookmarks_api.mm
```diff
@@ -657,15 +657,14 @@ - (void)searchWithQuery:(NSString*)queryArg
                 : query_parser::MatchingAlgorithm::DEFAULT);
 
     NSMutableArray<IOSBookmarkNode*>* nodes = [[NSMutableArray alloc] init];
+
     for (const auto& result : results) {
-      for (const bookmarks::BookmarkNode* bookmark :
```
nuo-xu, line 661: basically `bookmarks_api->bookmark_model_->GetBookmarksMatching` is all we need? how come users only start getting duplicates recently?
could this also be the reason some users have some bookmark searching problems. like this https://bravesoftware.slack.com/archives/C06UXF3KJ/p1746175312794829 and https://community.brave.app/t/bookmark-search-in-url-bar-behavior-has-changed-partial-word-search-is-broken-ios/623423

Brandon-T: Yes. It is using `ALWAYS_PREFIX_SEARCH` when the `length >= 3` which is from Chromium. 

This is true for `Open Tabs`, `History`, and `Bookmarks` https://source.chromium.org/chromium/chromium/src/+/main:ios/chrome/browser/history/ui_bundled/base_history_view_controller.mm;l=301;drc=391e98eeaee80326c3a9d9e9dd731570dba0981a;bpv=1;bpt=1
and:
https://source.chromium.org/chromium/chromium/src/+/main:components/omnibox/browser/bookmark_provider.cc;l=161;drc=391e98eeaee80326c3a9d9e9dd731570dba0981a;bpv=1;bpt=1
and:
https://source.chromium.org/chromium/chromium/src/+/main:components/omnibox/browser/open_tab_provider.cc;l=148;drc=391e98eeaee80326c3a9d9e9dd731570dba0981a;bpv=0;bpt=1

It's done for the bookmarks screen which is fine. For partial matches (which URL bar should use), we have to eventually stop using Bookmarks, History, and Tab search and just use `OmniboxAutocompleteController` directly: [ios/chrome/browser/omnibox/model/omnibox_autocomplete_controller.h](https://source.chromium.org/chromium/chromium/src/+/main:ios/chrome/browser/omnibox/model/omnibox_autocomplete_controller.h;bpv=0;bpt=1)

The auto-complete client will work just fine with partial matches. Otherwise we can revert the performance fix and use: `GetBookmarksMatchingProperties`.

Atm, we use (much faster):
```c++
std::vector<bookmarks::TitledUrlMatch> results =
        bookmarks_api->bookmark_model_->GetBookmarksMatching(
            base::SysNSStringToUTF16(query), maxCount,
            [query length] >= 3
                ? query_parser::MatchingAlgorithm::ALWAYS_PREFIX_SEARCH
                : query_parser::MatchingAlgorithm::DEFAULT);
```

But we could always revert to (much slower):

```c++
bookmarks::QueryFields queryFields;
queryFields.word_phrase_query.reset(
    new std::u16string(base::SysNSStringToUTF16(query)));
std::vector<const bookmarks::BookmarkNode*> results =
    GetBookmarksMatchingProperties(bookmarks_api->bookmark_model_,
                                   queryFields, maxCount);
```

until we get `OmniboxAutocompleteController` working which implements `AutocompleteController`.
I'd rather not sacrifice performance and go back to stuttering and slow searching, for partial matches. So maybe it's better to wait until `OmniboxAutocompleteController/AutocompleteController` is exported.

nuo-xu (approved):

PR #30963, opened 2025-09-02
[AI Chat] Add backend infrastructure for pushing bookmarks to the frontend and loading them in a backround WebContents

fallaciousreasoning:
Resolves https://github.com/brave/brave-browser/issues/48881

Security/Privacy Review: https://github.com/brave/reviews/issues/2033

This is part of https://github.com/brave/brave-browser/issues/48880. The PR has two parts:
1. Adds a C++ class for pushing the latest bookmarks (and any changes) to the frontend.
2. Adds an implementation of the `AssociatedContentDelegate` which can fetch and extract content from a link in a background webcontents.

A future PR will add the frontend implementation of this feature, to keep this on small and focused.

## Frontend

A work in progress (but working) PR for the frontend: https://github.com/brave/brave-core/pull/30964


components/ai_chat/core/common/mojom/bookmarks.mojom
```diff
@@ -0,0 +1,33 @@
...
+// but this may change in the future.
+struct BookmarksChange {
+  map<string, Bookmark> addedOrUpdated;
+  array<string> removed;
+};
+
+interface BookmarksListener {
+  // Called when the bookmarks change.
+  Changed(BookmarksChange change);
+};
```
darkdh: The interface is confusing, the change structure doesn't always represent the delta between states, it is sometimes used as the whole bookmarks state. And `Changed` got triggered even when there is no actual bookmarks changes ex. bookmark model loaded and a new listener added.

components/ai_chat/content/browser/bookmarks_service.cc
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  return change;
+}
+
+void BookmarksService::BookmarkAllUserNodesRemoved(
+    const std::set<GURL>& removed_urls,
+    const base::Location& location) {
+  // Refresh all bookmarks since all user nodes were removed
+  auto change = GetAllBookmarks();
```
darkdh: why isn't it `auto change = mojom::BookmarksChange::New();` since all bookmarks are gone?

components/ai_chat/content/browser/associated_link_content.cc
```diff
@@ -0,0 +1,160 @@
...
+  params.preview_mode = true;
+  web_contents_ = content::WebContents::Create(params);
+
+  // Start observing the WebContents
+  content::WebContentsObserver::Observe(web_contents_.get());
+}
+
+AssociatedLinkContent::~AssociatedLinkContent() = default;
+
+void AssociatedLinkContent::GetContent(GetPageContentCallback callback) {
```
darkdh, line 53: This needs unit test coverage.

components/ai_chat/content/browser/bookmarks_service.cc
fallaciousreasoning: good point. done.

components/ai_chat/core/common/mojom/bookmarks.mojom
fallaciousreasoning: I've updated the comments on this file - we use the same pattern in a few other places when we need to maintain a list on the frontend, keeping it in sync with a C++ backend. I was using `Changed` to mean there's a delta between what the frontend and the backend know, rather than the bookmarks actually changing.

I could also add a `getBookmarks()` method to the `BookmarksService` so the initial call to `AddListener` wouldn't need to notify of the current bookmarks, but we'd still need to when the `BookmarkModel` loads. Open to suggestions to make it more obvious whats going on.

Other places using this pattern:
- brave/components/brave_news/common/brave_news.mojom
- brave/components/ai_chat/core/common/mojom/tab_tracker.mojom
- brave/components/commands/common/commands.mojom

components/ai_chat/content/browser/bookmarks_service.cc
fallaciousreasoning: addressed in https://github.com/brave/brave-core/pull/30963/commits/6a20bdc50e39ab2ca5cafdb9eaedee3b328c0895

components/ai_chat/content/browser/associated_link_content.cc
fallaciousreasoning: addressed in https://github.com/brave/brave-core/pull/30963/commits/6a20bdc50e39ab2ca5cafdb9eaedee3b328c0895

components/ai_chat/core/common/mojom/bookmarks.mojom
darkdh: Can we add `OnBookmarksReady()` which will emit once when BookmarkModel is loaded? It can either send the initial bookmark state or hinting listener to manually call `GetBookmarks()`.   
If the listener is added before bookmark model is loaded, we should not emit anything so the current implementation is wrong.
If the listener is added after bookmark model is loaded, we emit `OnBookmarksReady()` immediately in `BookmarksService::AddListener()` via checking `bookmarks::BookmarkModel::loaded()`
So `Changed()` can remain only be used when bookmarks actually changed.

browser/ai_chat/associated_link_content_unittest.cc
```diff
@@ -0,0 +1,270 @@
...
+  void set_page_content_result(std::string content, bool is_video) {
+    EXPECT_CALL(*mock_fetcher_, FetchPageContent(_, _))
+        .WillOnce(RunOnceCallback<1>(content, is_video, ""));
+  }
+
+ private:
+  raw_ptr<MockPageContentFetcher> mock_fetcher_;
+};
+
+class AssociatedLinkContentTest : public ChromeRenderViewHostTestHarness {
```
darkdh: Adding new unit test in `//brave/browser` to test something that is in `//brave/components is forbidden`. cc @bridiver 
We should use `content::RenderViewHostTestHarness` and move this test into `brave/components/ai_chat/content/browser`.

components/ai_chat/core/browser/associated_content_manager.cc
```diff
@@ -131,7 +130,17 @@ void AssociatedContentManager::AddContent(AssociatedContentDelegate* delegate,
...
-    delegate->GetContent(base::DoNothing());
+    delegate->GetContent(base::BindOnce(
+        [](base::WeakPtr<AssociatedContentManager> self, PageContent content) {
+          if (!self) {
+            return;
+          }
+
+          // Note: |is_video| may have changed so we need to notify the
+          // conversation.
+          self->conversation_->OnAssociatedContentUpdated();
```
darkdh, line 141: This change should be reflected in `//brave/components/ai_chat/core/browser/associated_content_manager_unittest.cc` that when `AddContent` is called and verify that observer OnAssociatedContentUpdated is notified.

browser/ai_chat/associated_link_content_browsertest.cc
```diff
@@ -0,0 +1,164 @@
...
+#include "content/public/test/browser_test.h"
+#include "content/public/test/browser_test_utils.h"
+#include "content/public/test/content_mock_cert_verifier.h"
+#include "net/test/embedded_test_server/embedded_test_server.h"
+#include "services/network/public/cpp/network_switches.h"
+#include "testing/gtest/include/gtest/gtest.h"
+
+namespace ai_chat {
+
+class AssociatedLinkContentBrowserTest : public InProcessBrowserTest {
```
bridiver: if this is not strictly desktop-only functionality it should use PlatformBrowserTest. I thought we had a semgrep check for this, but I don't recall seeing any semgrep notices lately so I think maybe they aren't working

browser/ai_chat/associated_link_content_unittest.cc
fallaciousreasoning: Fixed this just before I saw the comment :laughing:

components/ai_chat/content/browser/bookmarks_service.h
```diff
@@ -0,0 +1,76 @@
...
+#include "mojo/public/cpp/bindings/receiver.h"
+#include "mojo/public/cpp/bindings/remote_set.h"
+
+namespace bookmarks {
+class BookmarkModel;
+}
+
+namespace ai_chat {
+
+class BookmarksService : public mojom::BookmarksService,
```
bridiver: why do we have a BookmarksService inside aichat? This class definitely needs some explanation in comments, but I don't understand why this would be necessary when there is already a BookmarksService and various observers.

browser/ai_chat/associated_link_content_browsertest.cc
bridiver: I guess there are some in this PR so maybe it is working. I need to check the rules

components/ai_chat/content/browser/bookmarks_service.h
bridiver: `Adds a C++ class for pushing the latest bookmarks (and any changes) to the frontend.` - there is already `chrome.bookmarks` which is available to webui, but also this interface doesn't seem like it's at the right level for a UI because it's sending add/change/remove instead of just getting the current list. We should discuss in slack

components/ai_chat/core/common/mojom/bookmarks.mojom
fallaciousreasoning: Like, we can, I just don't really understand what it buys us. It'll mean you need to understand the lifecycle of the `BookmarkModel` to use the listener.

I don't know if its useful, but this is how I'm using it currently in the frontend:
https://github.com/brave/brave-core/pull/30964/files#diff-fe31700018c66274f2561fcf59637d3aca4782fba4656f4e7c85ad69ce48c523

With your suggestion, we could update it to be something like this:
```
  constructor() {
    super()

    this.service = Mojom.BookmarksService.getRemote()
    this.service.getIsLoaded().then(({ loaded }) => {
      if (loaded) {
        const bookmarks = this.service.getBookmarks()
        this.changed(bookmarks)
      }
    })

    this.service.addListener(this.receiver.$.bindNewPipeAndPassRemote())
  }

  // assuming the onReady event passes a list of bookmarks
  onReady({ bookmarks} ) {
    this.changed(bookmarks);
  }
```

> If the listener is added before bookmark model is loaded, we should not emit anything so the current implementation is wrong.

tbh, I don't  think this is much of a problem - we'll just emit an empty event. We could fairly easily not emit an event if it isn't loaded (as the loaded event will provide a notification for the added **listener).**

> So Changed() can remain only be used when bookmarks actually changed.

Maybe `Changed` is the wrong name for this function? What I want to do is to be able to patch the data that the frontend has without worrying too much about what's going on with the model.

If we really want to separate getting initial state and subsequent updates what would you think of something like this?

```mojom
interface BookmarksService {
  // Resolves after the bookmarks model has been loaded in the backend
  GetBookmarks() => (map<string, Bookmark> bookmarks);
};

interface BookmarkListener {
  // Called when something changes.
  Changed(BookmarksChange event);
};
```

At the end of the day, in the frontend I want a `Bookmark[]` array which gets updated when something on the backend changes.

components/ai_chat/content/browser/associated_link_content_unittest.cc
```diff
@@ -0,0 +1,264 @@
...
+  associated_content->GetContent(future.GetCallback());
+
+  content::MockNavigationHandle mock_handle(test_url(), main_rfh());
+  mock_handle.set_has_committed(false);
+  mock_handle.set_is_error_page(true);
+  mock_handle.set_is_in_primary_main_frame(true);
+
+  associated_content->DidFinishNavigation(&mock_handle);
+
+  task_environment()->RunUntilIdle();
```
darkdh, line 233: why is RunUntilIdle needed when we explicitly call DidFinishNavigation?

components/ai_chat/core/common/mojom/bookmarks.mojom
fallaciousreasoning: Actually, all this might be moot if I just use the existing mojom interface, like @bridiver suggests

components/ai_chat/content/browser/associated_link_content.cc
```diff
@@ -0,0 +1,159 @@
...
+    return;
+  }
+
+  if (navigation_handle->IsErrorPage()) {
+    int net_error = navigation_handle->GetNetErrorCode();
+    std::string error_description = net::ErrorToString(net_error);
+    CompleteWithError("Navigation failed: " + error_description + " (" +
+                      base::NumberToString(net_error) + ")");
+  }
+}
```
darkdh, line 113: we only signal the one shot event when error or timeout? Even when the content is loaded before timeout and we still wait for timeout?

components/ai_chat/content/browser/associated_link_content_unittest.cc
fallaciousreasoning: Because the `OneShotEvent` is posted to the SequencedTaskRunner, so this doesn't complete synchronously. For the other tests I can wait for the future to complete, but in this case, I'm expecting it to not complete.

components/ai_chat/content/browser/associated_link_content.cc
fallaciousreasoning: We signal the event from `DocumentOnLoadCompletedInPrimaryMainFrame` ==> `PageContentExtractor::FetchPageContent` ==> `OnContentExtractionComplete`.'

The `DidFinishNavigation` check is just to see if we hit an error.

components/ai_chat/content/browser/associated_link_content_unittest.cc
darkdh: Currently these two test cases leaves callback uncalled, shouldn't we simulate timeout by fast forwarding time for both to reflect our handling?

components/ai_chat/content/browser/associated_link_content.cc
```diff
@@ -0,0 +1,160 @@
...
+namespace ai_chat {
+
+AssociatedLinkContent::AssociatedLinkContent(
+    GURL url,
+    std::u16string title,
+    content::BrowserContext* browser_context) {
+  DVLOG(2) << __func__ << "Creating link content for: " << url.spec()
+           << " title: " << title;
+
+  set_uuid(base::Uuid::GenerateRandomV4().AsLowercaseString());
```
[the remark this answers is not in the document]
bridiver: what is this uuid for? Maybe it should be `base::Token::CreateRandom()`?  Not in this PR, but curious what this is even used for

components/ai_chat/content/browser/associated_link_content_unittest.cc
fallaciousreasoning: We have a separate test for that:
https://github.com/brave/brave-core/pull/30963/files#diff-b8a1a1dd03633da97cceb64bc1b8857abf1e657bca417c2ff7794fc48ea19375R248

Here I just wanted to test that the callback is not called when we call `DidFinishNavigation` with the wrong params.

I'd still want the `RunUntilIdle(); EXPECT_FALSE(future.IsReady())`, I think to make sure the `DidFinishNavigation` firing the callback, but we could also test the Timeout here, if you think its a good idea?

components/ai_chat/content/browser/bookmarks_service.h
fallaciousreasoning: discussed in Slack - those services can't be used from the `ai_chat` component. We've redesigned this to just have a method for getting bookmarks and moved it to the `core` folder so the implementation can be shared with iOS

browser/ai_chat/associated_link_content_browsertest.cc
fallaciousreasoning: updated

components/ai_chat/core/common/mojom/bookmarks.mojom
fallaciousreasoning: Talked to @bridiver - we're going to test not pushing changes at all, and just pulling them when the dialog is opened.

components/ai_chat/content/browser/associated_link_content.cc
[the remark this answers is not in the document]
fallaciousreasoning: I think its the database id in `AIChatDatabase`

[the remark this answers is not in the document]
bridiver: Seems a bit unusual, I mean obviously nothing is going to change in this PR (or probably at all at this point), but seems like an `AUTOINCREMENT` key would have been simpler?

components/ai_chat/content/browser/associated_link_content_unittest.cc
```diff
@@ -0,0 +1,264 @@
...
+       GetContentResolvesWhenDocumentOnLoadCompletedInPrimaryMainFrame) {
+  auto associated_content =
+      CreateAssociatedLinkContent(test_url(), test_title());
+  associated_content->set_page_content_result(test_content(), false);
+
+  base::test::TestFuture<PageContent> future;
+  associated_content->GetContent(future.GetCallback());
+  EXPECT_TRUE(associated_content->content_loaded_event());
+
+  associated_content->DocumentOnLoadCompletedInPrimaryMainFrame();
```
bridiver, line 136: `NavigationSimulator` should normally be used in these cases instead of calling these method directly. See `PsstTabWebContentsObserverUnitTest`

components/ai_chat/content/browser/associated_link_content_unittest.cc
bridiver: see https://github.com/brave/brave-core/pull/30963/files#r2322581826

browser/ui/webui/ai_chat/ai_chat_ui_page_handler.cc
```diff
@@ -454,6 +457,19 @@ void AIChatUIPageHandler::DisassociateContent(
   service->DisassociateContent(content, conversation_uuid);
 }
 
+void AIChatUIPageHandler::AttachBookmark(ai_chat::mojom::BookmarkPtr bookmark,
+                                         const std::string& conversation_uuid) {
+  // Get TabInterface from the owner WebContents
+  tabs::TabInterface* tab_interface = tabs::TabInterface::GetFromContents(owner_web_contents_);
```
fallaciousreasoning: this is wrong

browser/ui/webui/ai_chat/ai_chat_ui_page_handler.cc
```diff
@@ -76,6 +81,17 @@ namespace ai_chat {
 
 namespace {
 
+class AssociatedLinkContentDelegate
+    : public ai_chat::AssociatedLinkContent::Delegate {
+ public:
+  AssociatedLinkContentDelegate() = default;
+  ~AssociatedLinkContentDelegate() override = default;
+  void AttachTabHelpers(content::WebContents* web_contents) override {
+    ephemeral_storage::EphemeralStorageTabHelper::FromWebContents(web_contents);
```
bridiver, line 90: I don't know if we should explicitly list them here, at least for now these will apply to any similar non-tab contents so maybe we should have a separate global method like `AttachTabHelpersForNonTabContents` or something and call that from the existing global `AttachTabHelpers`. Then we can also call it here and anyplace else we need them? I'd just like to keep the list someplace more visible

fallaciousreasoning: https://github.com/brave/brave-core/pull/31444

fallaciousreasoning:
This PR has been broken up into smaller chunks. See https://github.com/brave/brave-browser/issues/48880 for progress

PR #30965, opened 2025-09-02
BRAVE_SPEEDREADER content settings default value.

boocmp:
Resolves https://github.com/brave/brave-browser/issues/48498


browser/speedreader/speedreader_service_factory.cc
```diff
@@ -31,35 +30,44 @@ SpeedreaderService* SpeedreaderServiceFactory::GetForBrowserContext(
...
 
+// static
+bool SpeedreaderServiceFactory::IsAvailableFor(
+    content::BrowserContext* browser_context) {
+  if (!features::IsSpeedreaderEnabled()) {
+    return false;
+  }
+  if (!GetInstance()->GetBrowserContextToUse(browser_context)) {
+    return false;
+  }
```
simonhong, line 41: If all profiles can have speedreader service, we can replace this with
`CHECK(GetInstance()->GetBrowserContextToUse(browser_context))`.

browser/speedreader/speedreader_service_factory.h
```diff
@@ -8,7 +8,7 @@
 
 #include <memory>
 
-#include "components/keyed_service/content/browser_context_keyed_service_factory.h"
+#include "chrome/browser/profiles/profile_keyed_service_factory.h"
 #include "components/keyed_service/core/keyed_service.h"
```
simonhong: nit: maybe we could replace this with forward decl?

browser/ui/speedreader/speedreader_tab_helper.cc
```diff
@@ -544,7 +556,9 @@ void SpeedreaderTabHelper::OnSiteEnableSettingChanged(
 
 void SpeedreaderTabHelper::OnAllSitesEnableSettingChanged(
     bool enabled_on_all_sites) {
-  if (!is_visible_ || !GetSpeedreaderService()) {
+  if (!is_visible_ || !GetSpeedreaderService() ||
+      !rewriter_service_->URLLooksReadable(
+          web_contents()->GetLastCommittedURL())) {
```
simonhong: Q) This change is related with current issue?

browser/ui/speedreader/speedreader_tab_helper.cc
```diff
@@ -165,6 +165,14 @@ void SpeedreaderTabHelper::ProcessIconClick() {
   }
 }
 
+DistillState SpeedreaderTabHelper::PageDistillState() const {
+  auto* speedreader_service = GetSpeedreaderService();
+  if (!speedreader_service || !speedreader_service->IsFeatureEnabled()) {
+    return {};
+  }
+  return distill_state_;
```
simonhong: When service is null or feature is disabled, `distill_state_` could have non empty value?

components/speedreader/speedreader_service.cc
```diff
@@ -154,8 +146,12 @@ bool SpeedreaderService::IsEnabledForSite(content::WebContents* contents) {
 }
 
 bool SpeedreaderService::IsExplicitlyEnabledForSite(const GURL& url) {
+  if (!url.is_valid()) {
+    return CONTENT_SETTING_BLOCK;
```
simonhong: return type is bool

components/speedreader/speedreader_service.cc
```diff
@@ -167,8 +163,11 @@ bool SpeedreaderService::IsExplicitlyEnabledForSite(
 }
 
 bool SpeedreaderService::IsExplicitlyDisabledForSite(const GURL& url) {
+  if (!url.is_valid()) {
+    return CONTENT_SETTING_BLOCK;
```
simonhong: ditto

browser/speedreader/speedreader_service_factory.cc
boocmp: System profile doesn't have this, but may be it is not relevant, I'll check how it works with the system profile.

browser/ui/speedreader/speedreader_tab_helper.cc
boocmp: It replaces the logic in the icon view, where I removed similar checks

browser/ui/speedreader/speedreader_tab_helper.cc
boocmp: No it can't, I can return `ViewOriginal` instead of monostate with the same results.

components/speedreader/speedreader_service.cc
boocmp: nice catch.

browser/ui/speedreader/speedreader_tab_helper.cc
simonhong: then why not just return `distilled_state_` here?

simonhong (approved):

browser/ui/speedreader/speedreader_tab_helper.cc
boocmp: Sorry, I misunderstood your question, I meant that distill_state_ can never be empty because it's created with ViewOriginal state and never becomes empty if user don't turn off the speedreader. Previously when speedreader feature is disabled we didn't event create tab helper,  but now we have the toggle (that controls IsFeatureEnabled) and if we turn off this toggle when we already have a tab helper it 's state should be empty for foreign functions to display UI in consistent state and we should preserve the internal state for case if user enable it again.

simonhong: I see. I didn't know it can be on/off in runtime. so, return `{ViewOriginal}` would exact return value for disabled state?

boocmp: Yes, it this was introduced recently. And ViewOriginal (with is_distillable = false parameter) will remove the speedreader button from the navigation bar, but empty state represents it much clearer from my point of view.

browser/ui/speedreader/speedreader_tab_helper.cc
```diff
@@ -165,6 +165,14 @@ void SpeedreaderTabHelper::ProcessIconClick() {
   }
 }
 
+DistillState SpeedreaderTabHelper::PageDistillState() const {
+  auto* speedreader_service = GetSpeedreaderService();
+  if (!speedreader_service || !speedreader_service->IsFeatureEnabled()) {
```
simonhong: one more question - this service could be null in this tab helper?

boocmp: No. I'll remove the nullptr check

components/speedreader/speedreader_service.h
```diff
@@ -107,10 +107,20 @@ class SpeedreaderService : public KeyedService {
   SpeedreaderService& operator=(const SpeedreaderService&) = delete;
 
  private:
+  struct Setting {
+    bool is_default = false;
```
bridiver, line 111: I don't understand why this is needed, speedreader is just allow or block, right? So this is just a normal content setting that defaults to BLOCK?

boocmp: Because there is an logic which checks the speedreader turned on or off explicitly for page.

browser/speedreader/speedreader_service_unittest.cc
```diff
@@ -0,0 +1,98 @@
...
+#include "brave/components/speedreader/common/features.h"
+#include "brave/components/speedreader/speedreader_pref_names.h"
+#include "chrome/test/base/testing_profile.h"
+#include "components/prefs/pref_service.h"
+#include "content/public/test/browser_task_environment.h"
+#include "testing/gtest/include/gtest/gtest.h"
+
+namespace speedreader {
+
+class SpeedreaderServiceTest : public testing::Test {
```
bridiver, line 19: there is already a unit test in components

browser/speedreader/speedreader_service_factory.cc
```diff
@@ -31,35 +30,44 @@ SpeedreaderService* SpeedreaderServiceFactory::GetForBrowserContext(
           browser_context, true /*create*/));
 }
 
+// static
+bool SpeedreaderServiceFactory::IsAvailableFor(
```
bridiver, line 34: These changes appear to be unrelated to the content setting, please create a separate PR for refactoring

bridiver: the other changes in this PR were handled here by using ASK https://github.com/brave/brave-core/pull/31409/files

PR #30966, opened 2025-09-02
Set ZOrderLevel::kSecuritySurface to parent widget of PermissionPromptBubble

sangwoo108:
Previously, we only set kSecuritySurface to the prompt widget. But this caused the prompt to be floating without its parent window, which may feel disjointed from the overall UI.

thus, this change sets kSecuritySurface to the parent widget of the prompt. And when the prompt bubble is closed, the parent widget's ZOrderLevel is reset to its original value.

Resolves https://github.com/brave/brave-browser/issues/48093


simonhong:
According to the comment (https://github.com/brave/internal/issues/1318#issuecomment-3105234238), it's fixed by upstream `cr139`. Should we have our own fix if upstream already fixed?

With this PR, Brave works like `always on top` and can't activate to another application window via `cmd + tab`.
Is this behavior is what we want? IMO, it's too aggressive.

sangwoo108:
>According to the comment (https://github.com/brave/internal/issues/1318#issuecomment-3105234238), it's fixed by upstream cr139. Should we have our own fix if upstream already fixed?

I thought @kdenhartog said that upstream patch didn't fix ours.

>Is this behavior is what we want?
Yes, I think so. We're showing the premission bubble like "Always on top" and don't want it to be floating without the parent window.

@kdenhartog , what do you think?

simonhong:
If permission bubble is launched while browser is inactive, it makes browser as top-most application suddently.
and user can't go back to previous active window. Is this situation ok?
Tested https://frozzipies.github.io/googlecam2.html.

@sangwoo108 found that browser dont' go back to original z-order when I did allow/deny the permission bubble of https://frozzipies.github.io/googlecam2.html

sangwoo108:
> If permission bubble is launched while browser is inactive, it makes browser as top-most application suddently. and user can't go back to previous active window. Is this situation ok? Tested https://frozzipies.github.io/googlecam2.html.
> 
> @sangwoo108 found that browser dont' go back to original z-order when I did allow/deny the permission bubble of https://frozzipies.github.io/googlecam2.html

Thanks, maybe it's not closed via "DialogDelegate::Close()" in that scenario. Let me check again

kdenhartog:
> I thought @kdenhartog said that upstream patch didn't fix ours.

Yeah, I was expecting 139 to fix it because I saw it made changes on chrome canary, but then once it actually got migrated into Nightly I was still seeing the error. That's why we ended up needing to address it.

sangwoo108:
> If permission bubble is launched while browser is inactive, it makes browser as top-most application suddently. and user can't go back to previous active window. Is this situation ok? Tested https://frozzipies.github.io/googlecam2.html.
> 
> @sangwoo108 found that browser dont' go back to original z-order when I did allow/deny the permission bubble of https://frozzipies.github.io/googlecam2.html

Thanks @simonhong , I believe both cases are addressed now. Could you take a look again?

sangwoo108:
I tried to resolve CI failure on Mac, but the bubble widget doesn't seem to be activated.

sangwoo108:
@simonhong Do you think it's okay to disable the test on Mac? I think somehow we can't activate the widget in the CI. When tested locally, it didn't happen, the test succeed.

simonhong:
> @simonhong Do you think it's okay to disable the test on Mac? I think somehow we can't activate the widget in the CI. When tested locally, it didn't happen, the test succeed.

I think it's ok as it runs well on other platforms.

PR #29305, opened 2025-05-29
Hide split tabs toolbar button

simonhong:
Resolves https://github.com/brave/brave-browser/issues/46441

PR #30967, opened 2025-09-02
Reenable Split button in toolbar view

sangwoo108:
This reverts commit 476cf2e1e76dd3dfebf45d3ad616a00e3970d157.
- "Hide split tabs toolbar button (#29305)"

As we've defaulted "SideBySide" of upstream, reenabling it makes sense.

Resolves https://github.com/brave/brave-browser/issues/47390

### DIFF
--- a/browser/ui/views/toolbar/brave_toolbar_view.cc
+++ b/browser/ui/views/toolbar/brave_toolbar_view.cc
@@ -166,12 +166,6 @@ void BraveToolbarView::Init() {
     return;
   }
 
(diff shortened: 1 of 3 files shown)


sangwoo108:
@simonhong When do you think is good time to land this? Should we land this once we remove our own split view flag?

simonhong:
> @simonhong When do you think is good time to land this? Should we land this once we remove our own split view flag?

If we want to show this button with side by side, I think it's ok now.

PR #30968, opened 2025-09-02
Revert pinned tab layout

simonhong:
Resolves https://github.com/brave/brave-browser/issues/48869


simonhong:
> So the second tab is not closed, but only hidden because of wrong layout?

Right, need to revisit later.

simonhong:
Reverted code is only valid for SideBySide. so it'll not affect 1.82.x

GeetaSarvadnya:
Verification **PASSED** on

```
Brave | 1.84.22 Chromium: 140.0.7339.80 (Official Build) nightly (64-bit)
-- | --
Revision | 1e968c322d8b4039e00919bcd7f3542dc8ff7893
OS | Windows 10 Version 22H2 (Build 19045.6216)
```

Reproduced the issue on `1.82.160 Chromium: 140.0.7339.80`
issue on RC

Confirmed that issue is fixed and working fine on `1.84.22 Chromium: 140.0.7339.80`
fixed

PR #30970, opened 2025-09-02
Don't load Sparkle when Omaha 4 is active

mherrmann:
Even though we don't explicitly invoke Sparkle when Omaha 4 is active, loading it already has side effects. For example, it deletes the file `/private/tmp/com.brave.Browser[.beta|.nightly].Sparkle.pid`.

Needs #30969.

Resolves https://github.com/brave/brave-browser/issues/48891.

# Test plan

1. Please check that automatic updates still work. (From my perspective, it's not necessary to do that just for this PR. If this check is performed before a release goes live, that is sufficient. But it does need to be checked before a release goes live.)
2. Please check that the "steps to reproduce" in the [issue description](https://github.com/brave/brave-browser/issues/48891) now produce the expected result.

### DIFF
--- a/browser/brave_browser_main_parts_mac.mm
+++ b/browser/brave_browser_main_parts_mac.mm
@@ -26,7 +26,7 @@ void BraveBrowserMainPartsMac::PreCreateMainMessageLoop() {
     return;
   }
 #endif
-  // It would be a no-op if udpates are disabled.
+  // It would be a no-op if updates are disabled.
   [[SparkleGlue sharedSparkleGlue] registerWithSparkle];
 #endif  // BUILDFLAG(ENABLE_SPARKLE)
 }


browser/brave_browser_main_parts_mac.mm
```diff
@@ -7,18 +7,28 @@
...
+#include "brave/browser/updater/features.h"
+#endif
+
 void BraveBrowserMainPartsMac::PreCreateMainMessageLoop() {
   ChromeBrowserMainPartsMac::PreCreateMainMessageLoop();
 
 #if BUILDFLAG(ENABLE_SPARKLE)
-  // It would be no-op if udpate is disabled.
-  [[SparkleGlue sharedSparkleGlue] registerWithSparkle];
+#if BUILDFLAG(ENABLE_OMAHA4)
```
simonhong, line 24: It's confusing that omaha is guarded with sparkle.
sparkle is always enabled when using omaha?

simonhong: Ah, I see. when sparkle is disabled we don't need to care its loading :)

simonhong (approved):
++

PR #30971, opened 2025-09-02
[PSST] Add delegate to interact with UI part

vadimstruts:
Resolves https://github.com/brave/brave-browser/issues/48757

### DIFF
--- a/browser/psst/psst_ui_delegate_impl.cc
+++ b/browser/psst/psst_ui_delegate_impl.cc
@@ -1,0 +1,23 @@
+// Copyright (c) 2025 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#include "brave/browser/psst/psst_ui_delegate_impl.h"
+
+namespace psst {
+
+PsstUiDelegateImpl::PsstUiDelegateImpl() = default;
+PsstUiDelegateImpl::~PsstUiDelegateImpl() = default;
+
+void PsstUiDelegateImpl::SetAppliedItems(
+    long progress,
+    const std::vector<psst_script_responses::PsstTask>& applied_tasks) {
+  // Implementation for setting the current progress.
+}
+
+void PsstUiDelegateImpl::SetComplete() {
+  // Implementation for marking the UI as complete.
+}
+
+}  // namespace psst
--- a/browser/psst/psst_ui_delegate_impl.h
+++ b/browser/psst/psst_ui_delegate_impl.h
@@ -1,0 +1,28 @@
+// Copyright (c) 2025 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#ifndef BRAVE_BROWSER_PSST_PSST_UI_DELEGATE_IMPL_H_
+#define BRAVE_BROWSER_PSST_PSST_UI_DELEGATE_IMPL_H_
+
+#include "brave/components/psst/browser/content/psst_tab_web_contents_observer.h"
+#include "brave/components/psst/common/psst_script_responses.h"
+
+namespace psst {
+
+class PsstUiDelegateImpl : public PsstTabWebContentsObserver::PsstUiDelegate {
+ public:
+  PsstUiDelegateImpl();
+  ~PsstUiDelegateImpl() override;
+
+  // PsstUiDelegate overrides
+  void SetAppliedItems(long progress,
+                       const std::vector<psst_script_responses::PsstTask>&
+                           applied_tasks) override;
+  void SetComplete() override;
+};
+
+}  // namespace psst
+
+#endif  // BRAVE_BROWSER_PSST_PSST_UI_DELEGATE_IMPL_H_
--- a/components/psst/README.md
+++ b/components/psst/README.md
@@ -98,6 +98,24 @@ const params = {
 }
 ;
 <policy script content>
+```
+
+##### Policy script result:
+The policy script returns the next object as result:
 
 ```
+{
+    "status": "<STARTED|COMPLETED>",
+    "progress": "<percent of completion>",
+    "applied": [{
+      "description": "Ads Preferences",
+      "url": "https://x.com/settings/ads_preferences",
+      "error_description": "<optional error description>"
+    }]
+}
+```
+- `status` - represents the flow's status, `started` or `completed`;
+- `progress` - script calculates the percent of operation completion;
+- `applied` - array of the completed tasks, each task is an object `url/description/error_description`
+
 
--- a/components/psst/browser/content/psst_tab_web_contents_observer.cc
+++ b/components/psst/browser/content/psst_tab_web_contents_observer.cc
@@ -70,11 +70,13 @@ std::unique_ptr<PsstTabWebContentsObserver>
 PsstTabWebContentsObserver::MaybeCreateForWebContents(
     content::WebContents* contents,
     content::BrowserContext* browser_context,
+    std::unique_ptr<PsstUiDelegate> ui_delegate,
     PrefService* prefs,
     const int32_t world_id) {
   CHECK(contents);
   CHECK(browser_context);
   CHECK(prefs);
+  CHECK(ui_delegate);
 
   if (browser_context->IsOffTheRecord() ||
       !base::FeatureList::IsEnabled(psst::features::kEnablePsst)) {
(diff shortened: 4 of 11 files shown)


components/psst/common/psst_script_responses.idl
```diff
@@ -0,0 +1,24 @@
+/* Copyright (c) 2025 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+// Schema for Brave PSST script responses.
+namespace psst_script_responses {
```
bridiver: namespace should match component `psst`

components/psst/browser/content/psst_tab_web_contents_observer.h
```diff
@@ -28,9 +29,24 @@ class PsstTabWebContentsObserver : public content::WebContentsObserver {
       const std::string&,
       PsstTabWebContentsObserver::InsertScriptInPageCallback)>;
 
+  class PsstUiDelegate {
+   public:
+    PsstUiDelegate();
```
bridiver: You can drop the constructor/destructor stuff here and make this a pure interface. If the class has no state and no implementation and all methods are pure virtual you don't need these here

components/psst/common/psst_script_responses.idl
vadimstruts: done

components/psst/browser/content/psst_tab_web_contents_observer.cc
```diff
@@ -181,7 +189,26 @@ void PsstTabWebContentsObserver::OnUserScriptResult(
...
+void PsstTabWebContentsObserver::OnPolicyScriptResult(
+    int nav_entry_id,
+    base::Value script_result) {
+  const auto script_result_parsed =
+      psst_script_responses::PolicyScriptResult::FromValue(script_result);
+  if (!script_result_parsed || !ShouldInsertScriptForPage(nav_entry_id)) {
+    return;
+  }
+
+  ui_delegate_->SetAppliedItems(script_result_parsed->progress,
```
bridiver: I don't understand `progress` here. This is a once callback, the script can't keep calling it multiple times.

components/psst/common/psst_script_responses.idl
```diff
@@ -0,0 +1,24 @@
...
+    enum PolicyScriptStatus {
+        STARTED,
+        COMPLETED
```
bridiver: what if it fails?

components/psst/browser/content/psst_tab_web_contents_observer.cc
vadimstruts: the policy script executes for each psst task we want to apply, so after each execution we should actualize the UI.
`SetAppliedItems` sends progress and list of applied tasks UI (i.e. display actual progress, and status of each task)

components/psst/browser/content/psst_tab_web_contents_observer.h
```diff
@@ -28,9 +29,24 @@ class PsstTabWebContentsObserver : public content::WebContentsObserver {
       const std::string&,
       PsstTabWebContentsObserver::InsertScriptInPageCallback)>;
 
+  class PsstUiDelegate {
```
bridiver, line 35: comments for class and methods please

components/psst/common/psst_script_responses.idl
vadimstruts: we just add this task with error here: (https://github.com/brave/brave-core/pull/30971/files#diff-c114099511c81ed1a1fd05ee32bb4b51c2d4769a1a10a0ed2fbd34ff81aff528R22), show it on UI, and go to the next task

components/psst/browser/content/psst_tab_web_contents_observer.cc
bridiver: but that's what I'm saying, you can't do that. It's a promise that gets resolved and it can only get resolved once. This is also a once callback which means you can only call it once. I don't see a test for a policy script that updates the progress by calling it more than once and it doesn't look like that test could work if it did exist.

components/psst/browser/content/psst_tab_web_contents_observer.h
```diff
@@ -28,9 +29,24 @@ class PsstTabWebContentsObserver : public content::WebContentsObserver {
       const std::string&,
       PsstTabWebContentsObserver::InsertScriptInPageCallback)>;
 
+  // Delegate interface for UI-related actions. This class is responsible for
+  // facilitating communication with the consent dialog, ensuring that the UI
+  // reflects the current state accurately.
+  class PsstUiDelegate {
+   public:
+    virtual ~PsstUiDelegate() = default;
+    // Actualize the UI state based on the applied tasks and progress.
```
bridiver: Actualize -> Update

components/psst/browser/content/psst_tab_web_contents_observer.cc
vadimstruts: I have added unittest for that case: `UiDelegateSetAppliedItemsButNotSetCompleteMultipleTasks`

bridiver: That test has two navigations in it so I'm confused about how this is supposed to work. This is a multi-step process that involves several user navigations?

bridiver: I think this process needs to be described in the README because this is not at all clear to me

components/psst/common/psst_script_responses.idl
bridiver: but how do we know it failed if it only returns started/completed?

components/psst/browser/content/psst_tab_web_contents_observer.cc
vadimstruts: I updated README.md: https://github.com/brave/brave-core/tree/48757-psst-script-ui-delegate/components/psst.
Added common workflow description.
cc: @ShivanKaul

components/psst/common/psst_script_responses.idl
vadimstruts: It may be better to rename that enum, as it is more about the whole workflow status
As for current task status: each task contains `error_description` property, so if it is not empty, means that current task failed, but it doesn't mean that whole wrokflow is failed.
The consent dialog should be opened all the time, It will display errors for each task separately

components/psst/browser/content/psst_tab_web_contents_observer_unittest.cc
```diff
@@ -556,22 +592,287 @@ TEST_F(PsstTabWebContentsObserverUnitTest,
...
+                                                             url);
+  observer.Wait();
+
+  check_loop.Run();
+  EXPECT_EQ(script_params, user_script_insert_future.Take());
+  EXPECT_EQ(policy_script_result, policy_script_insert_future.Take());
+  EXPECT_EQ(progress, progress_future.Take());
+}
+
+TEST_F(PsstTabWebContentsObserverUnitTest,
```
bridiver: if this is actually multiple separate navigations that are initiated by the script then we don't need a unit tests for this. I thought the same script was calling it multiple times which clearly wouldn't work.

components/psst/common/psst_script_responses.idl
bridiver: maybe `PolicyStatus` or `WorkflowStatus`? The `Psst` prefix is just redundant here. Also comments explaining would be helpful

components/psst/common/psst_script_responses.idl
```diff
@@ -0,0 +1,24 @@
...
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+// Schema for Brave PSST script responses.
+namespace psst {
+    enum PsstStatus {
+        STARTED,
+        COMPLETED
+    };
+
+    dictionary PsstTask {
```
bridiver: per above maybe this should be `PolicyTask`?

components/psst/README.md
```diff
@@ -5,6 +5,21 @@ The feature helps tweak privacy settings on popular websites.
...
 
+# PSST Common Workflow
+
+1. The user opens any website.
+2. PSST feature detects that it is supported website (it means that browser loads PSST CRX component and find the appropriate user and policy scripts).
+3. The browser injects the user script into the page. The primary goal of the user script is to ensure that the user is signed in to the current website and return a list of URLs where we should apply the privacy settings.
+4. If the user is not signed in, we just break the flow.
+5. If the user is signed in and we have a list of URLs, we will display a consent modal dialog. In this dialog, the user can choose their preferred privacy settings and track the progress of the operation.
+6. When the user clicks the OK button, we inject the policy script and execute it.
+7. The policy script saves the list of URLs (tasks) to local storage to be available after the next navigation, takes the first URL (URL_1), marks it as current, and navigates to it.
```
bridiver: some details from the code like the class and method it calls (`SetAppliedItems`) would be helpful here.

components/psst/browser/content/psst_tab_web_contents_observer.h
```diff
@@ -28,9 +29,24 @@ class PsstTabWebContentsObserver : public content::WebContentsObserver {
...
       PsstTabWebContentsObserver::InsertScriptInPageCallback)>;
 
+  // Delegate interface for UI-related actions. This class is responsible for
+  // facilitating communication with the consent dialog, ensuring that the UI
+  // reflects the current state accurately.
+  class PsstUiDelegate {
+   public:
+    virtual ~PsstUiDelegate() = default;
+    // Update the UI state based on the applied tasks and progress.
+    virtual void SetAppliedItems(
```
bridiver: `SetAppliedItems` -> `UpdateTasks`? Changing terminology is confusing so try to be consistent.

browser/psst/psst_ui_delegate_impl.cc
```diff
@@ -0,0 +1,23 @@
...
+PsstUiDelegateImpl::PsstUiDelegateImpl() = default;
+PsstUiDelegateImpl::~PsstUiDelegateImpl() = default;
+
+void PsstUiDelegateImpl::SetAppliedItems(
+    long progress,
+    const std::vector<PsstTask>& applied_tasks) {
+  // Implementation for setting the current progress.
+}
+
+void PsstUiDelegateImpl::SetComplete() {
```
bridiver: is this not implied by `progress == 100`?

components/psst/browser/content/psst_tab_web_contents_observer_unittest.cc
vadimstruts: Removed

browser/psst/psst_ui_delegate_impl.cc
vadimstruts: removed `SetComplete`

components/psst/README.md
vadimstruts: updated it

browser/psst/psst_ui_delegate_impl.h
```diff
@@ -0,0 +1,26 @@
...
+#include "brave/components/psst/browser/content/psst_tab_web_contents_observer.h"
+#include "brave/components/psst/common/psst_script_responses.h"
+
+namespace psst {
+
+class PsstUiDelegateImpl : public PsstTabWebContentsObserver::PsstUiDelegate {
+ public:
+  PsstUiDelegateImpl();
+  ~PsstUiDelegateImpl() override;
+
```
bridiver, line 18: PsstUiDelegateImpl(const PsstUiDelegateImpl&) = delete;
  PsstUiDelegateImpl& operator=(const PsstUiDelegateImpl&) = delete;

PR #30974, opened 2025-09-02
Fix potential Sparkle/Omaha 4 race condition in system installs

mherrmann:
Resolves https://github.com/brave/brave-browser/issues/47795.

Needs #30969 and #30970.


simonhong:
I think this PR has assumption that sparkle's update is already there when omaha tries to update.
If `Sparkle.pid` exists, it's ok and dmg install from omaha will be cancelled.
If `Sparkle.pid` is not created yet when sparkle is active, dmg from omaha will be applied.
Curios this situation can happen? and it's ok if it's happened?

mherrmann:
Good point. I think it can happen but is very unlikely.

Sparkle's update process conceptually works as follows:

1. `mv Brave.app Brave.app.old`
2. `mv Brave.app.new Brave.app`

Omaha 4 on the other hand modifies the `.app` directory in-place.

Per my analysis, there is only one way in which Sparkle and Omaha 4 (O4) can break Brave during an update. Namely when O4 starts modifying Brave in-place, then Sparkle's steps 1. and 2. run before O4 is done. In this case, O4 will copy some - but not all - files of a potentially different new version into `Brave.app`. The result is a version of Brave.app that has files from different versions, and therefore probably will not work any more.

Sparkle creates the `.pid` file before it starts downloading the new version. With this PR, there is only one way in which the scenario described in the previous paragraph can happen. That is when O4's installation of the new version (which has already been downloaded) starts but does not finish before Sparkle 1. downloads the version 2. extracts it 3. moves it into place. Especially the download (which is not cached, per my reading of the code) should take orders of magnitudes more time, giving O4 plenty of time to complete beforehand.

I would be very grateful if you could also think through this reasoning. Any holes or mistakes you find would be extremely valuable, because they could brick Brave for affected users. The relevant sources are:

1. O4's logic for installing a new version is in [.install.sh](https://source.chromium.org/chromium/chromium/src/+/main:chrome/updater/mac/.install.sh).
2. Sparkle's installation logic is in [SUPlainInstaller.m](https://github.com/brave/Sparkle/blob/main/Sparkle/SUPlainInstaller.m), according to my analysis.
3. Sparkle's logic that creates the `.pid` file is in [SUGlobalUpdateLock.m](https://github.com/brave/Sparkle/blob/main/Sparkle/SUGlobalUpdateLock.m).

simonhong:
BTW, Can't we tell O4 to stop/pause the update? I'm very new to O4 :)

mherrmann:
Sorry, I'm not sure which of these two you mean:

1. stopping/pausing an update that is currently happening (for example, from `.preinstall` in this PR)
2. telling O4 not to check for updates in general.

simonhong:
> Sorry, I'm not sure which of these two you mean:
> 
> 1. stopping/pausing an update that is currently happening (for example, from `.preinstall` in this PR)
> 2. telling O4 not to check for updates in general.

2

mherrmann:
We do 2. by disabling O4's out-of-process background updates and only checking for updates with O4 explicitly from within Brave.

simonhong:
If Brave can ask O4 to stop background update, how update can happen from both side simultaneously?

mherrmann:
Whether Brave uses O4 or Sparkle is controlled by a Griffin flag. I don't believe O4 and Sparkle can conflict in single-user installs. But when there are multiple users that run the same copy of Brave, then some might be on O4 and others on Sparkle.

simonhong:
> Whether Brave uses O4 or Sparkle is controlled by a Griffin flag. I don't believe O4 and Sparkle can conflict in single-user installs. But when there are multiple users that run the same copy of Brave, then some might be on O4 and others on Sparkle.

I see. thanks. It's very interesting situation.

simonhong:
AFAIK, applications are terminated when user logouts on macOS.
That means, Brave will be restarted when uer logins.
During the startup, Brave can ask O4 to start or stop update checking based on current flag(O4 or Sparkle).
If Sparkle is selected, Brave asks O4 to stop and vice-versa.
This could work?

Or some users can't ask O4?

mherrmann:
All users can ask O4, depending on the value of their Griffin flag. But more to what I believe your point is: Multiple users could be logged in at the same time. They may receive different values for the Griffin flag. In this case, logged-in user A might check with Sparkle while at the same time logged-in user B might check with O4.

simonhong:
I think there are two scenarios for preventing update from O4.
I thought we want 1 but maybe we want 2 after reading this PR's test scenario.

```
1. Always preventing update from O4 if any of Braves(from multiple users) are using Sparkle.
2. only prevenint update from O4 if any of Braves are updating with Sparkle
```

I thought this comment as I assumed we want option 1 above. but I think it's ok with option 2 if O4's update is done w/o any issue.

```
I think this PR has assumption that sparkle's update is already there when omaha tries to update.
If Sparkle.pid exists, it's ok and dmg install from omaha will be cancelled.
If Sparkle.pid is not created yet when sparkle is active, dmg from omaha will be applied.
Curios this situation can happen? and it's ok if it's happened?
```

If I thought option 2 is our seleciton, maybe I didn't write above comment :) 
But, you found very rare corner case from my comment that I didn't intend 👍🏼 
As you wrote, I think that corner case could be happen but very very rare (maybe would not be happened because it's inplace update vs, downloading & move)

On macOS, I think sparkle could not touch some file if that file is updating(by O4). So, maybe we don't need to worry about this case?

simonhong (approved):
++

mherrmann:
Thank you @simonhong. Your observation about option 1. is good. I believe option 2. should suffice as well. So I'm going to go ahead and merge. Thank you for your inputs!

PR #30621, opened 2025-08-13
Origin policy provider WIP

DJAndries:
Resolves

PR #30984, opened 2025-09-02
Add Brave Browser Policy Provider for Brave Origin

bbondy:
Adds a Brave browser-level policy provider.
In future work, it will handle global scope policies in Brave Origin. The provider integrates with Chromium's BrowserPolicyConnector to supply policy management for local state prefs.

Resolves https://github.com/brave/brave-browser/issues/48906

### DIFF
--- a/chromium_src/chrome/browser/policy/chrome_browser_policy_connector.cc
+++ b/chromium_src/chrome/browser/policy/chrome_browser_policy_connector.cc
@@ -1,0 +1,39 @@
+/* Copyright (c) 2025 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "chrome/browser/policy/chrome_browser_policy_connector.h"
+
+// Forward declare functions that will be implemented in
+// brave_browser_policy_provider.cc This is done so that we don't need to depend
+// on anything in the brave layer here. Otherwise we'd have a circular
+// dependency.
+namespace brave_policy {
+std::unique_ptr<policy::ConfigurationPolicyProvider>
+CreateBraveBrowserPolicyProvider();
+}  // namespace brave_policy
+
+// Rename CreatePolicyProviders to CreatePolicyProviders_ChromiumImpl
+#define CreatePolicyProviders CreatePolicyProviders_ChromiumImpl
+
+#include <chrome/browser/policy/chrome_browser_policy_connector.cc>  // IWYU pragma: export
+
+#undef CreatePolicyProviders
+
+// And define the new one
+namespace policy {
+
+std::vector<std::unique_ptr<policy::ConfigurationPolicyProvider>>
+ChromeBrowserPolicyConnector::CreatePolicyProviders() {
+  auto providers =
+      ChromeBrowserPolicyConnector::CreatePolicyProviders_ChromiumImpl();
+  // Add browser policy provider for browser-level (local state) policies
+  auto brave_browser_provider =
+      ::brave_policy::CreateBraveBrowserPolicyProvider();
+  // providers takes ownership of brave_browser_provider
+  providers.push_back(std::move(brave_browser_provider));
+  return providers;
+}
+
+}  // namespace policy
--- a/chromium_src/chrome/browser/policy/chrome_browser_policy_connector.h
+++ b/chromium_src/chrome/browser/policy/chrome_browser_policy_connector.h
@@ -1,0 +1,24 @@
+/* Copyright (c) 2025 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#ifndef BRAVE_CHROMIUM_SRC_CHROME_BROWSER_POLICY_CHROME_BROWSER_POLICY_CONNECTOR_H_
+#define BRAVE_CHROMIUM_SRC_CHROME_BROWSER_POLICY_CHROME_BROWSER_POLICY_CONNECTOR_H_
+
+#include "base/gtest_prod_util.h"
+#include "components/policy/core/browser/browser_policy_connector.h"
+#include "components/policy/core/common/policy_service.h"
+
+// Rename CreatePolicyProviders to CreatePolicyProviders_ChromiumImpl
+// and declare a new one
+#define CreatePolicyProviders                                       \
+  CreatePolicyProviders_ChromiumImpl();                             \
+  std::vector<std::unique_ptr<policy::ConfigurationPolicyProvider>> \
+      CreatePolicyProviders
+
+#include <chrome/browser/policy/chrome_browser_policy_connector.h>  // IWYU pragma: export
+
+#undef CreatePolicyProviders
+
+#endif  // BRAVE_CHROMIUM_SRC_CHROME_BROWSER_POLICY_CHROME_BROWSER_POLICY_CONNECTOR_H_
--- a/chromium_src/components/policy/core/browser/browser_policy_connector_base.h
+++ b/chromium_src/components/policy/core/browser/browser_policy_connector_base.h
@@ -1,0 +1,35 @@
+/* Copyright (c) 2025 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#ifndef BRAVE_CHROMIUM_SRC_COMPONENTS_POLICY_CORE_BROWSER_BROWSER_POLICY_CONNECTOR_BASE_H_
+#define BRAVE_CHROMIUM_SRC_COMPONENTS_POLICY_CORE_BROWSER_BROWSER_POLICY_CONNECTOR_BASE_H_
+
+// The reason for this override is to add CreatePolicyProviders_ChromiumImpl in
+// BrowserPolicyConnector.
+// We need that function because the base class's CreatePolicyProviders gets
+// called from another override we did in
+// chromium_src/chrome/browser/policy/chrome_browser_policy_connector.cc
+// which renames CreatePolicyProviders to CreatePolicyProviders_ChromiumImpl,
+// but that also renames the call to its base class's
+// BrowserPolicyConnector::CreatePolicyProviders inside that same
+// chrome/browser/policy/chrome_browser_policy_connector.cc file.
+// eg.:
+//
+// std::vector<std::unique_ptr<policy::ConfigurationPolicyProvider>>
+// ChromeBrowserPolicyConnector::CreatePolicyProviders() {
+//   auto providers = BrowserPolicyConnector::CreatePolicyProviders(); <-----
+//
+// Since you can't just rename only the first instance, the base class also
+// needs an override.
+#define CreatePolicyProviders                               \
+  CreatePolicyProviders();                                  \
+  std::vector<std::unique_ptr<ConfigurationPolicyProvider>> \
+      CreatePolicyProviders_ChromiumImpl
+
+#include <components/policy/core/browser/browser_policy_connector_base.h>  // IWYU pragma: export
+
+#undef CreatePolicyProviders
+
+#endif  // BRAVE_CHROMIUM_SRC_COMPONENTS_POLICY_CORE_BROWSER_BROWSER_POLICY_CONNECTOR_BASE_H_
--- a/components/brave_policy/DEPS
+++ b/components/brave_policy/DEPS
@@ -1,0 +1,4 @@
+include_rules = [
+  "+components/policy",
+  "+components/prefs",
+]
--- a/browser/sources.gni
+++ b/browser/sources.gni
@@ -163,6 +163,7 @@ brave_chrome_browser_deps = [
   "//brave/components/brave_news/common:mojom",
   "//brave/components/brave_origin",
   "//brave/components/brave_perf_predictor/browser",
+  "//brave/components/brave_policy",
   "//brave/components/brave_private_new_tab_ui/common",
   "//brave/components/brave_private_new_tab_ui/common:mojom",
   "//brave/components/brave_referrals/browser",
--- a/chromium_src/components/policy/core/browser/browser_policy_connector_base.cc
+++ b/chromium_src/components/policy/core/browser/browser_policy_connector_base.cc
@@ -1,0 +1,15 @@
+/* Copyright (c) 2025 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include <components/policy/core/browser/browser_policy_connector_base.cc>
+
+namespace policy {
+
+std::vector<std::unique_ptr<ConfigurationPolicyProvider>>
+BrowserPolicyConnectorBase::CreatePolicyProviders_ChromiumImpl() {
+  return {};
+}
+
+}  // namespace policy
--- a/components/BUILD.gn
+++ b/components/BUILD.gn
@@ -25,6 +25,7 @@ test("brave_components_unittests") {
     "//brave/components/api_request_helper:unit_tests",
     "//brave/components/brave_account/endpoint_client:unit_tests",
     "//brave/components/brave_account/endpoints:unit_tests",
+    "//brave/components/brave_policy:unit_tests",
     "//brave/components/brave_shields/core/browser:unit_tests",
     "//brave/components/brave_wallet/common:unit_tests",
     "//brave/components/email_aliases:unit_tests",
--- a/components/brave_policy/BUILD.gn
+++ b/components/brave_policy/BUILD.gn
@@ -1,0 +1,33 @@
+# Copyright (c) 2024 The Brave Authors. All rights reserved.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+static_library("brave_policy") {
+  sources = [
+    "brave_browser_policy_provider.cc",
+    "brave_browser_policy_provider.h",
+  ]
+
+  deps = [
+    "//base",
+    "//components/policy/core/browser",
+    "//components/policy/core/common",
+    "//components/prefs",
+  ]
+}
+
+source_set("unit_tests") {
+  testonly = true
+  sources = [ "brave_browser_policy_provider_unittest.cc" ]
+
+  deps = [
+    ":brave_policy",
+    "//base",
+    "//base/test:test_support",
+    "//components/policy/core/common",
+    "//components/policy/core/common:test_support",
+    "//components/prefs:test_support",
+    "//testing/gtest",
+  ]
+}
--- a/components/brave_policy/brave_browser_policy_provider.cc
+++ b/components/brave_policy/brave_browser_policy_provider.cc
@@ -1,0 +1,61 @@
+/* Copyright (c) 2025 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "brave/components/brave_policy/brave_browser_policy_provider.h"
+
+#include <utility>
+
+#include "base/logging.h"
+#include "base/values.h"
+#include "components/policy/core/common/policy_bundle.h"
+#include "components/policy/core/common/policy_map.h"
+#include "components/policy/core/common/policy_namespace.h"
+#include "components/policy/core/common/policy_types.h"
+#include "components/prefs/pref_service.h"
+
+namespace brave_policy {
+
+BraveBrowserPolicyProvider::BraveBrowserPolicyProvider() = default;
+
+BraveBrowserPolicyProvider::~BraveBrowserPolicyProvider() {}
+
+void BraveBrowserPolicyProvider::Init(policy::SchemaRegistry* registry) {
+  // Call base class Init first
+  policy::ConfigurationPolicyProvider::Init(registry);
+
+  // Trigger immediate policy loading to ensure policies are available in
+  // chrome://policy
+  RefreshPolicies(policy::PolicyFetchReason::kBrowserStart);
+}
+
+void BraveBrowserPolicyProvider::RefreshPolicies(
+    policy::PolicyFetchReason reason) {
+  policy::PolicyBundle bundle = LoadPolicies();
+
+  // Mark as loaded after successfully loading policies
+  first_policies_loaded_ = true;
+
+  UpdatePolicy(std::move(bundle));
+}
+
+bool BraveBrowserPolicyProvider::IsFirstPolicyLoadComplete(
+    policy::PolicyDomain domain) const {
+  return first_policies_loaded_;
+}
+
+policy::PolicyBundle BraveBrowserPolicyProvider::LoadPolicies() {
+  policy::PolicyBundle bundle;
+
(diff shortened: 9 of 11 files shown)


components/policy/brave_browser_policy_provider.cc
```diff
@@ -0,0 +1,62 @@
...
+
+namespace brave_policy {
+
+BraveBrowserPolicyProvider::BraveBrowserPolicyProvider()
+    : policy::ConfigurationPolicyProvider(), first_policies_loaded_(false) {}
+
+BraveBrowserPolicyProvider::~BraveBrowserPolicyProvider() {}
+
+void BraveBrowserPolicyProvider::Initialize(PrefService* local_state,
+                                            policy::SchemaRegistry* registry) {
```
DJAndries: Could we pass the local state via the constructor, and call `RefreshPolicies` there as well?

DJAndries: Also, do we need `registry`?

If we decide to keep the method, maybe a different name is warranted. `ConfigurationPolicyProvider::Init` already exists.

DJAndries: Ah, I see. Local state isn't available at the time of construction. Perhaps we could still consider using a different name (i.e. `InitLocalState`) and dropping the `registry` param.

components/policy/brave_browser_policy_provider.cc
```diff
@@ -0,0 +1,62 @@
...
+  CHECK(local_state);
+  local_state_ = local_state;
+
+  // Trigger immediate policy loading for browser startup
+  RefreshPolicies(policy::PolicyFetchReason::kBrowserStart);
+}
+
+void BraveBrowserPolicyProvider::Shutdown() {
```
DJAndries: Do we need this override?

components/policy/brave_browser_policy_provider.h
```diff
@@ -0,0 +1,48 @@
...
+
+class PrefService;
+
+namespace brave_policy {
+
+// Browser-level policy provider for global scope preferences.
+// This provider handles policies that affect local state (global) preferences
+// and is registered with the BrowserPolicyConnector for machine-wide policy
+// management.
+class BraveBrowserPolicyProvider : public policy::ConfigurationPolicyProvider {
```
DJAndries: Should we consider an Origin-specific name i.e. `OriginBrowserPolicyProvider`? Future policy types unrelated to Origin could have their own policy provider.

components/policy/brave_browser_policy_provider_unittest.cc
```diff
@@ -0,0 +1,127 @@
...
+    pref_service_ = std::make_unique<TestingPrefServiceSimple>();
+    schema_registry_ = std::make_unique<policy::SchemaRegistry>();
+    provider_ = std::make_unique<BraveBrowserPolicyProvider>();
+  }
+
+ protected:
+  base::test::TaskEnvironment task_environment_;
+  std::unique_ptr<TestingPrefServiceSimple> pref_service_;
+  std::unique_ptr<policy::SchemaRegistry> schema_registry_;
+  std::unique_ptr<BraveBrowserPolicyProvider> provider_;
```
DJAndries: nit: do we need `unique_ptr`? these classes have default constructors

components/policy/brave_browser_policy_provider_unittest.cc
```diff
@@ -0,0 +1,127 @@
...
+};
+
+TEST_F(BraveBrowserPolicyProviderTest, Construction) {
+  // Provider should be constructed successfully
+  EXPECT_NE(nullptr, provider_);
```
DJAndries: nit: won't construction always succeed?

components/policy/brave_browser_policy_provider_unittest.cc
```diff
@@ -0,0 +1,127 @@
...
+  provider_->Shutdown();
+
+  // Provider should still report policies as loaded even after shutdown
+  EXPECT_TRUE(
+      provider_->IsFirstPolicyLoadComplete(policy::POLICY_DOMAIN_CHROME));
+}
+
+TEST_F(BraveBrowserPolicyProviderTest, InitializeWithNullPrefService) {
+  // Attempting to initialize with null pref service should fail with CHECK
+  EXPECT_DEATH(provider_->Initialize(nullptr, schema_registry_.get()), "");
```
DJAndries: cool, i didn't know about `EXPECT_DEATH` :+1:

components/policy/brave_browser_policy_provider.cc
bbondy: yes but you won't see why unless you look at the later PRs.

components/policy/brave_browser_policy_provider.h
bbondy: I intentionally made this generic and not using the Origin name. Because there's other things that want to use this within Brave too. It will have Origin inside of it if you look at the full solution in https://github.com/brave/brave-core/pull/30621

components/policy/brave_browser_policy_provider_unittest.cc
bbondy: Dropped it completely in the latest commit and moved the second check into the test below it..

components/policy/brave_browser_policy_provider.cc
bbondy: Removed the `registry` param since I didn't end up using it in the end result. I also renamed `Initialize` to `InitLocalStateAndRefresh`

components/policy/brave_browser_policy_provider.cc
bbondy: I removed it for now, I'll re-add it in the code that needs it.

components/policy/brave_browser_policy_provider_unittest.cc
bbondy: Updated to not be pointers.

components/policy/brave_browser_policy_provider_unittest.cc
```diff
@@ -0,0 +1,114 @@
...
+  BraveBrowserPolicyProviderTest() = default;
+  ~BraveBrowserPolicyProviderTest() override = default;
+
+ protected:
+  base::test::TaskEnvironment task_environment_;
+  TestingPrefServiceSimple pref_service_;
+  BraveBrowserPolicyProvider provider_;
+};
+
+TEST_F(BraveBrowserPolicyProviderTest, InitiAndPolicyLoadComplete) {
```
DJAndries: nit: typo in the test name

chromium_src/chrome/browser/policy/chrome_browser_policy_connector_unittest.cc
```diff
@@ -0,0 +1,106 @@
...
+  ChromeBrowserPolicyConnector connector;
+
+  // Call CreatePolicyProviders to get the list of providers
+  auto providers = connector.CreatePolicyProviders();
+
+  // After CreatePolicyProviders, brave_browser_provider_ should be set
+  EXPECT_NE(nullptr, connector.brave_browser_provider_);
+
+  // Verify that the brave_browser_provider_ is in the providers list
+  bool found_brave_provider = false;
```
DJAndries: nit: `std::ranges::any_of` might work nicely here

DJAndries (approved):

components/policy/brave_browser_policy_provider.cc
```diff
@@ -0,0 +1,57 @@
...
+#include "components/policy/core/common/policy_bundle.h"
+#include "components/policy/core/common/policy_map.h"
+#include "components/policy/core/common/policy_namespace.h"
+#include "components/policy/core/common/policy_types.h"
+#include "components/prefs/pref_service.h"
+
+namespace brave_policy {
+
+BraveBrowserPolicyProvider::BraveBrowserPolicyProvider()
+    : policy::ConfigurationPolicyProvider(), first_policies_loaded_(false) {}
```
DJAndries: nit: could use `bool first_policies_loaded_ = false;` in the header instead

components/policy/brave_browser_policy_provider_unittest.cc
bbondy: Fixed and pushed

chromium_src/chrome/browser/policy/chrome_browser_policy_connector_unittest.cc
bbondy: Fixed and pushed

components/policy/brave_browser_policy_provider.cc
bbondy: Fixed and pushed

chromium_src/chrome/browser/policy/chrome_browser_policy_connector.cc
```diff
@@ -0,0 +1,38 @@
...
+namespace policy {
+
+std::vector<std::unique_ptr<policy::ConfigurationPolicyProvider>>
+ChromeBrowserPolicyConnector::CreatePolicyProviders() {
+  auto providers =
+      ChromeBrowserPolicyConnector::CreatePolicyProviders_ChromiumImpl();
+
+  // Add browser policy provider for browser-level (local state) policies
+  auto brave_browser_provider =
+      std::make_unique<brave_policy::BraveBrowserPolicyProvider>();
```
bridiver: you're missing an include for this, but this override is going to result in more patches when upstream modularizes `chrome_browser_policy_connector`. The way to work around this is to forward declare a method here like `CreateBraveBrowserPolicyProvider` and then implement it in `brave_browser_policy_provider.cc`. That way as long as `//brave/components/policy` is in sources.gni it will still work without patches.

chromium_src/chrome/browser/policy/chrome_browser_policy_connector.h
```diff
@@ -0,0 +1,39 @@
+/* Copyright (c) 2025 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#ifndef BRAVE_CHROMIUM_SRC_CHROME_BROWSER_POLICY_CHROME_BROWSER_POLICY_CONNECTOR_H_
+#define BRAVE_CHROMIUM_SRC_CHROME_BROWSER_POLICY_CHROME_BROWSER_POLICY_CONNECTOR_H_
+
+#include "base/gtest_prod_util.h"
+#include "brave/components/policy/brave_browser_policy_provider.h"
```
bridiver: forward declare this so we don't create a dependency per https://github.com/brave/brave-core/pull/30984/files#r2319852368

chromium_src/chrome/browser/policy/chrome_browser_policy_connector_unittest.cc
```diff
@@ -0,0 +1,94 @@
...
+  // Initialize the connector (this should call CreatePolicyProviders and
+  // InitInternal) We pass nullptr for URL loader factory since we're only
+  // testing policy provider initialization
+  connector.Init(pref_service_.get(), nullptr);
+
+  // After initialization, brave_browser_provider_ should be cleared to prevent
+  // dangling pointer (it was only needed for initialization)
+  EXPECT_EQ(nullptr, connector.brave_browser_provider_);
+}
+
```
bridiver: you need to include the original file so you don't lose upstream tests

chromium_src/chrome/browser/policy/DEPS
```diff
@@ -0,0 +1,6 @@
+from brave_chromium_utils import inline_file
```
bridiver: this should go away with the changes below. We want to avoid creating dependencies here when possible

chromium_src/components/policy/core/browser/browser_policy_connector_base.h
```diff
@@ -0,0 +1,22 @@
...
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#ifndef BRAVE_CHROMIUM_SRC_COMPONENTS_POLICY_CORE_BROWSER_BROWSER_POLICY_CONNECTOR_BASE_H_
+#define BRAVE_CHROMIUM_SRC_COMPONENTS_POLICY_CORE_BROWSER_BROWSER_POLICY_CONNECTOR_BASE_H_
+
+// Insert a declaration for CreatePolicyProviders_ChromiumImpl because
+// an override in
+// chromium_src/chrome/browser/policy/chrome_browser_policy_connector.cc can't
+// avoid calling a base class's CreatePolicyProviders_ChromiumImpl
+#define HasPolicyService                                    \
```
bridiver: you are making this normally protected method public by using `HasPolicyService`. Why not use `CreatePolicyProviders`? Also `#define CreatePolicyProviders()` makes this more readable because you don't have to leave off the `()` for `CreatePolicyProviders_ChromiumImpl`

chromium_src/components/policy/core/browser/browser_policy_connector_base.h
```diff
@@ -0,0 +1,22 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#ifndef BRAVE_CHROMIUM_SRC_COMPONENTS_POLICY_CORE_BROWSER_BROWSER_POLICY_CONNECTOR_BASE_H_
+#define BRAVE_CHROMIUM_SRC_COMPONENTS_POLICY_CORE_BROWSER_BROWSER_POLICY_CONNECTOR_BASE_H_
+
+// Insert a declaration for CreatePolicyProviders_ChromiumImpl because
+// an override in
+// chromium_src/chrome/browser/policy/chrome_browser_policy_connector.cc can't
```
bridiver: I'm not sure I understand this, it sounds like you are trying to avoid calling `BrowserPolicyConnectorBase::CreatePolicyProviders`, but you still call `ChromeBrowserPolicyConnector::CreatePolicyProviders_ChromiumImpl()`? I think this is because of `BrowserPolicyConnectorBase::CreatePolicyProviders_ChromiumImpl`?, but that's not really clear here

components/policy/BUILD.gn
```diff
@@ -3,6 +3,35 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at https://mozilla.org/MPL/2.0/.
 
+source_set("policy") {
```
bridiver: `static_library`

chromium_src/chrome/browser/policy/chrome_browser_policy_connector.cc
```diff
@@ -0,0 +1,38 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "chrome/browser/policy/chrome_browser_policy_connector.h"
+
+// Rename CreatePolicyProviders to CreatePolicyProviders_ChromiumImpl
+#define CreatePolicyProviders CreatePolicyProviders_ChromiumImpl
+// Just after InitInternal, call InitLocalStateAndRefresh for
+// brave_browser_provider_
+#define InitInternal(a, b)                                        \
```
bridiver: `...` and `__VA_ARGS__`

chromium_src/chrome/browser/policy/chrome_browser_policy_connector.cc
```diff
@@ -0,0 +1,38 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "chrome/browser/policy/chrome_browser_policy_connector.h"
+
+// Rename CreatePolicyProviders to CreatePolicyProviders_ChromiumImpl
+#define CreatePolicyProviders CreatePolicyProviders_ChromiumImpl
+// Just after InitInternal, call InitLocalStateAndRefresh for
+// brave_browser_provider_
```
bridiver: some explanation that this is done here because you don't have access to local_state in `CreatePolicyProviders` would be helpful

chromium_src/chrome/browser/policy/chrome_browser_policy_connector.cc
```diff
@@ -0,0 +1,38 @@
...
+#define InitInternal(a, b)                                        \
+  InitInternal(a, b);                                             \
+  brave_browser_provider_->InitLocalStateAndRefresh(local_state); \
+  brave_browser_provider_ = nullptr;
```
bridiver: why set this back to nullptr? It's a bit confusing and `platform_provider_` and `command_line_provider_` don't do that?

bridiver: is this causing a dangling pointer error? I wonder why the others don't

chromium_src/chrome/browser/policy/chrome_browser_policy_connector.cc
bbondy: > you're missing an include for this, but this override 

The include is in the .h file.

chromium_src/chrome/browser/policy/chrome_browser_policy_connector.cc
bbondy: Because otherwise tests fail with a dangling pointer error

chromium_src/chrome/browser/policy/chrome_browser_policy_connector_unittest.cc
bbondy: Move to `brave/browser/policy`

chromium_src/chrome/browser/policy/chrome_browser_policy_connector.h
bbondy: Done in the latest

chromium_src/chrome/browser/policy/DEPS
bbondy: It's gone 🎉

chromium_src/components/policy/core/browser/browser_policy_connector_base.h
bbondy: I'm not sure what you mean. It's just putting a declaration for `CreatePolicyProviders_ChromiumImpl` inside `chromium_src/components/policy/core/browser/browser_policy_connector_base.h`. 
The reason is explained in the comment.  The override in `chromium_src/chrome/browser/policy/chrome_browser_policy_connector.cc` renames `CreatePolicyProviders` to `CreatePolicyProviders_ChromiumImpl` but that also renames `BrowserPolicyConnectorBase::CreatePolicyProviders` inside that same `chrome/browser/policy/chrome_browser_policy_connector.cc` file.


eg.:

```
std::vector<std::unique_ptr<policy::ConfigurationPolicyProvider>>
ChromeBrowserPolicyConnector::CreatePolicyProviders() {
  auto providers = BrowserPolicyConnector::CreatePolicyProviders(); <-----
```

Because you can't just rename only the first instance, the base class also needed an override here.

chromium_src/components/policy/core/browser/browser_policy_connector_base.h
bbondy: explained above.

components/policy/BUILD.gn
bbondy: updated in the latest

chromium_src/chrome/browser/policy/chrome_browser_policy_connector.cc
bbondy: updated in the latest

chromium_src/chrome/browser/policy/chrome_browser_policy_connector.cc
bbondy: Updated in the latest

chromium_src/components/policy/core/browser/browser_policy_connector_base.h
bridiver: either way my point is that imo this comment is a bit confusing and the explanation above references this comment `The reason is explained in the comment` :) It's still not clear to me why this is different from a typical method rename/replacement? Maybe it isn't and this comment is just making it more confusing by looking like it's doing something different than normal?

chromium_src/chrome/browser/policy/chrome_browser_policy_connector.cc
bridiver: I wonder why the other ones don't cause the same problem, but that's fine, maybe just a comment about dangling pointers since it's not immediately obvious why you'd set it to nullptr there, especially since the code is out of context in chromium_src

chromium_src/components/policy/core/browser/browser_policy_connector_base.h
bbondy: the comment just explains why there's an override at all. It's explaining the purpose of the override in that file. I'll write the more verbose explanation though in the file.

bbondy: Per Slack I hung this off of CreatePolicyProviders instead

components/policy/brave_browser_policy_provider.cc
```diff
@@ -0,0 +1,70 @@
...
+
+#include "base/logging.h"
+#include "base/values.h"
+#include "components/policy/core/common/policy_bundle.h"
+#include "components/policy/core/common/policy_map.h"
+#include "components/policy/core/common/policy_namespace.h"
+#include "components/policy/core/common/policy_types.h"
+#include "components/prefs/pref_service.h"
+
+namespace brave_policy {
```
bridiver: this should match the component so either `policy` here or `brave/components/brave_policy`

bridiver: this should maybe be a separate PR if it touches other files to change it everywhere

bbondy: I updated to `policy`

bbondy: I updated back to `brave_policy` and moved the files to an associated dir.

test/BUILD.gn
```diff
@@ -203,6 +204,7 @@ test("brave_unit_tests") {
     "//brave/components/brave_component_updater/browser",
     "//brave/components/brave_news/browser/test:brave_news_unit_tests",
     "//brave/components/brave_perf_predictor/browser",
+    "//brave/components/brave_policy:unit_tests",
```
bridiver: this shouldn't be here, it's already in the correct place here https://github.com/brave/brave-core/pull/30984/files#diff-0213b93d3260f4ad014ca0b858368dc8901b3878d5cdc75f3e0b7ae32c18e570R28

bridiver (approved):

chromium_src/chrome/browser/policy/chrome_browser_policy_connector.h
```diff
@@ -0,0 +1,28 @@
...
+#include "components/policy/core/common/policy_service.h"
+
+// Rename CreatePolicyProviders to CreatePolicyProviders_ChromiumImpl
+// and declare a new one
+#define CreatePolicyProviders                                       \
+  CreatePolicyProviders_ChromiumImpl();                             \
+  std::vector<std::unique_ptr<policy::ConfigurationPolicyProvider>> \
+      CreatePolicyProviders
+
+// Override to add Brave policy provider
```
bridiver: no longer needed I think

bridiver: and https://github.com/brave/brave-core/pull/30984/files#diff-d566a21d43ff476e393b6e69af318c732e12b624ccee8af2ce02375a5868b9f0R25

components/brave_policy/DEPS
```diff
@@ -0,0 +1,6 @@
+include_rules = [
+  "+components/policy/core/browser",
```
bridiver: it's ok to just have `+components/policy` here if you want

components/brave_policy/DEPS
```diff
@@ -0,0 +1,6 @@
...
+  "+components/policy/core/common",
+  "+components/policy/policy_constants.h",
```
bridiver: this looks wrong?

bridiver (approved):
Please handle unresolved comments and then lgtm

chromium_src/chrome/browser/policy/chrome_browser_policy_connector.h
bbondy: Fixed in the latest

components/brave_policy/DEPS
bbondy: generalized in the latest

components/brave_policy/DEPS
bbondy: Fixed, no longer needed because of generalized components/policy

PR #30993, opened 2025-09-02
Only render the thumbnail for full screenshots

darkdh:
And the thumbnail is the first image of screenshots. When a screenshots thumbnail is removed, all staging screenshots should be removed.

Resolves https://github.com/brave/brave-browser/issues/46688

### Capture and submit
1. Use Leo panel to take screenshots on pages that have scroll bar. ex. https://brianbondy.com/blog/184/the-future-of-leo-a-roadmap-for-brave-s-built-in-ai-browser-assistant
2. In the input box, we should see thumbnail of screenshot which should be part of first page of the site.
3. Ask Leo question like "Do you see any screenshots?"
4. Leo should be able to see screenshots
5. And screenshots in the conversation show as a thumbnail too

### Capture and remove
1. Use Leo panel to take screenshots on pages that have scroll bar. ex. https://brianbondy.com/blog/184/the-future-of-leo-a-roadmap-for-brave-s-built-in-ai-browser-assistant
2. In the input box, we should see thumbnail of screenshot which should be part of first page of the site.
3. Remove the thumbnail
4. Ask Leo question like "Do you see any screenshots?"
5. Leo should not see any screenshots.

### Autoscreenshots
1. Navigate to any google doc
2. Click Leo "summarize this page" button
3. Screenshots in the conversation show as a thumbnail

### DIFF
--- a/components/ai_chat/resources/page/components/attachment_item/index.tsx
+++ b/components/ai_chat/resources/page/components/attachment_item/index.tsx
@@ -15,6 +15,7 @@ import * as Mojom from '../../../common/mojom'
 // Styles
 import styles from './style.module.scss'
 import { getLocale } from '$web-common/locale'
+import { isFullPageScreenshot } from '../../../common/conversation_history_utils'
 
 /**
  * Formats file size in bytes to human readable format
@@ -33,6 +34,7 @@ export const formatFileSize = (bytes: number): string => {
 
   return `${size.toFixed(2)} ${units[index]}`
 }
+
 
 type Props = {
   icon: React.ReactNode
@@ -124,14 +126,27 @@ export function AttachmentUploadItems(props: {
   uploadedFiles: Mojom.UploadedFile[]
   remove?: (index: number) => void
 }) {
+  // Calculate first full page screenshot index.
+  const firstFullPageScreenshotIndex = props.uploadedFiles.findIndex(isFullPageScreenshot)
+
   return (
     <>
-      {props.uploadedFiles.map((file, index) => {
-        const isImage = file.type === Mojom.UploadedFileType.kImage ||
-                       file.type === Mojom.UploadedFileType.kScreenshot
-        const isPdf = file.type === Mojom.UploadedFileType.kPdf
+      {props.uploadedFiles
+        .filter((file, index) => {
+          // Only show the first full page screenshot, hide the rest
+          return !isFullPageScreenshot(file) || index === firstFullPageScreenshotIndex
+        })
+        .map((file, filteredIndex) => {
+          // Find the original index in the unfiltered array
+          const originalIndex = props.uploadedFiles.indexOf(file)
 
-        if (isImage) {
+          const isImage = file.type === Mojom.UploadedFileType.kImage ||
+                         file.type === Mojom.UploadedFileType.kScreenshot
+          const isPdf = file.type === Mojom.UploadedFileType.kPdf
+          const isFileFullPageScreenshot = isFullPageScreenshot(file)
+
+          if (isImage) {
+
           const dataUrl = React.useMemo(() => {
             const blob = new Blob([new Uint8Array(file.data)], {
               type: 'image/*'
@@ -143,38 +158,38 @@ export function AttachmentUploadItems(props: {
             return formatFileSize(Number(file.filesize))
           }, [file.filesize])
 
-          return (
-            <AttachmentItem
-              key={`${file.filename}-${index}`}
-              icon={
-                <img
-                  className={styles.image}
-                  src={dataUrl}
-                />
-              }
-              title={file.filename}
-              subtitle={filesize}
-              remove={props.remove ? () => props.remove!(index) : undefined}
-            />
-          )
-        } else if (isPdf) {
-          const filesize = React.useMemo(() => {
-            return formatFileSize(Number(file.filesize))
-          }, [file.filesize])
+            return (
+              <AttachmentItem
+                key={`${file.filename}-${originalIndex}`}
+                icon={
+                  <img
+                    className={styles.image}
+                    src={dataUrl}
+                  />
+                }
+                title={isFileFullPageScreenshot ? getLocale(S.CHAT_UI_FULL_PAGE_SCREENSHOT_TITLE) : file.filename}
+                subtitle={filesize}
+                remove={props.remove ? () => props.remove?.(originalIndex) : undefined}
+              />
+            )
+          } else if (isPdf) {
+            const filesize = React.useMemo(() => {
+              return formatFileSize(Number(file.filesize))
+            }, [file.filesize])
 
-          return (
-            <AttachmentItem
-              key={`${file.filename}-${index}`}
-              icon={<Icon name='file' />}
-              title={file.filename}
-              subtitle={filesize}
-              remove={props.remove ? () => props.remove!(index) : undefined}
-            />
-          )
-        }
+            return (
+              <AttachmentItem
+                key={`${file.filename}-${originalIndex}`}
+                icon={<Icon name='file' />}
+                title={file.filename}
+                subtitle={filesize}
+                remove={props.remove ? () => props.remove!(originalIndex) : undefined}
+              />
+            )
+          }
 
-        return null
-      })}
+          return null
+        })}
     </>
   )
 }
--- a/components/ai_chat/resources/page/state/conversation_context.tsx
+++ b/components/ai_chat/resources/page/state/conversation_context.tsx
@@ -16,7 +16,7 @@ import {
   IGNORE_EXTERNAL_LINK_WARNING_KEY
 } from '../../common/constants'
 import {
-  updateConversationHistory, processUploadedFilesWithLimits
+  updateConversationHistory, processUploadedFilesWithLimits, isFullPageScreenshot
 } from '../../common/conversation_history_utils'
 import useHasConversationStarted from '../hooks/useHasConversationStarted'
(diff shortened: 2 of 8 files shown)


components/ai_chat/resources/page/components/input_box/index.tsx
```diff
@@ -219,6 +220,7 @@ function InputBox(props: InputBoxProps) {
           <AttachmentUploadItems
             uploadedFiles={props.context.pendingMessageFiles}
             remove={(index) => props.context.removeFile(index)}
+            removeScreenshots={props.context.removeScreenshots ? () => props.context.removeScreenshots!() : undefined}
```
fallaciousreasoning: I don't think making this a no-op rather than undefined is fine, and it'll make your other logic simpler too
nit:
```suggestion
            removeScreenshots={() => props.context.removeScreenshots?.()}
```

components/ai_chat/resources/page/components/attachment_item/index.tsx
```diff
@@ -132,6 +133,18 @@ export function AttachmentUploadItems(props: {
         const isPdf = file.type === Mojom.UploadedFileType.kPdf
 
         if (isImage) {
+          const isScreenshot = file.type === Mojom.UploadedFileType.kScreenshot
+
+          // For screenshots, only show the first one and use "Full page screenshot" as title
+          if (isScreenshot) {
+            const firstScreenshotIndex = props.uploadedFiles.findIndex(f =>
+              f.type === Mojom.UploadedFileType.kScreenshot
+            )
```
fallaciousreasoning: nit:

we could store this outside the loop so we only ever need to calculate it once.

optional:
we could add a filter clause checking whether to hide rather than returning `null` which might read more nicely? Up to you though

components/ai_chat/resources/page/components/attachment_item/index.tsx
```diff
@@ -152,9 +165,11 @@ export function AttachmentUploadItems(props: {
                   src={dataUrl}
                 />
               }
-              title={file.filename}
+              title={isScreenshot ? getLocale(S.CHAT_UI_FULL_PAGE_SCREENSHOT_TITLE) : file.filename}
               subtitle={filesize}
-              remove={props.remove ? () => props.remove!(index) : undefined}
+              remove={props.remove ? (isScreenshot ? (
+                props.removeScreenshots ? props.removeScreenshots : undefined
+              ) : () => props.remove!(index)) : undefined}
```
fallaciousreasoning: I think this would read more nicely with fewer ternaries
```suggestion
              remove={props.remove ? () => {
                if (isScreenshot) {
                    props.removeScreenshots?.()
                } else {
                    props.remove?.(index)
                }
              } : undefined}
```

components/ai_chat/resources/page/state/conversation_context.tsx
```diff
@@ -568,6 +570,15 @@ export function ConversationContextProvider(props: React.PropsWithChildren) {
     })
   }
 
+  const removeScreenshots = () => {
+    const updatedImages = context.pendingMessageFiles.filter(file =>
+      file.type !== Mojom.UploadedFileType.kScreenshot
+    )
```
fallaciousreasoning: hey, so in future I think we're going to have functionality for taking screenshots of specific bits of a page (i.e. currently visible) so we might not want to remove every screenshot?

Additionally, does it ever make sense to remove a single one of the fullpage screenshots? If not, maybe we should just update removeFile to remove all fullpage screenshots, if the item being removed is a full page screenshot.

That would simplify the logic in the input_box and attachment item, at the risk of being a bit less intuitive

darkdh: merging into `removeFile` and check full page screenshots pattern specifically in 372012dde533fbfb9177567cf302b0dd8fb134ff

components/ai_chat/resources/page/components/input_box/index.tsx
darkdh: there is no separate `removeScreenshots` prop per https://github.com/brave/brave-core/pull/30993#discussion_r2317368524

components/ai_chat/resources/page/components/attachment_item/index.tsx
darkdh: fixed in https://github.com/brave/brave-core/commit/372012dde533fbfb9177567cf302b0dd8fb134ff

components/ai_chat/resources/page/components/attachment_item/index.tsx
darkdh: fixed in https://github.com/brave/brave-core/commit/372012dde533fbfb9177567cf302b0dd8fb134ff

components/ai_chat/resources/page/components/attachment_item/index.tsx
```diff
@@ -34,6 +34,16 @@ export const formatFileSize = (bytes: number): string => {
   return `${size.toFixed(2)} ${units[index]}`
 }
 
+/**
+ * Checks if a file is a full page screenshot
+ * @param file - The uploaded file to check
+ * @returns True if the file is a full page screenshot
+ */
+const isFullPageScreenshot = (file: Mojom.UploadedFile): boolean => {
```
fallaciousreasoning: Maybe we should move this function into `conversation_context` and export it rather than reimplementing it there?

components/ai_chat/resources/page/state/conversation_context.tsx
```diff
@@ -561,11 +561,23 @@ export function ConversationContextProvider(props: React.PropsWithChildren) {
   }
 
   const removeFile = (index: number) => {
-    const updatedImages = [...context.pendingMessageFiles]
-    updatedImages.splice(index, 1)
-    setPartialContext({
-      pendingMessageFiles: updatedImages
-    })
+    if (index === -1) {
```
fallaciousreasoning: do we need to special case `-1` here? Could we just remove all fullpage screenshots if `index` points to a full page screenshot?

It'd simplify the login in `AttachmentItem` further.

fallaciousreasoning (approved):
lgtm with some optional improvements

components/ai_chat/resources/page/components/attachment_item/index.tsx
darkdh: addressed that in the latest commit

components/ai_chat/resources/page/components/attachment_item/index.tsx
```diff
@@ -15,6 +15,7 @@ import * as Mojom from '../../../common/mojom'
 // Styles
 import styles from './style.module.scss'
 import { getLocale } from '$web-common/locale'
+import { isFullPageScreenshot } from '../../state/conversation_context'
```
fallaciousreasoning: sorry for putting you wrong - I think this import probably isn't safe here because it gets rendered inside the root frame and the untrusted frame and we don't want to pull the trusted context into the untrusted frame (because it'll crash).

Let's put it in a shared utility in `//brave/components/ai_chat/resources/common/` instead.

fallaciousreasoning (approved):
lgtm % moving import

components/ai_chat/resources/page/components/attachment_item/index.tsx
darkdh: moved to  components/ai_chat/resources/common/conversation_history_utils.ts

kjozwiak:

```
Brave | 1.84.67 Chromium: 140.0.7339.133 (Official Build) nightly (64-bit)
-- | --
Revision | d259a1123fc752c0bc730579a40dd755759e9f62
```

Using the STR/Cases outlined via https://github.com/brave/brave-core/pull/30993#issue-3377413682, went through the following:

### `Test Case #1` - `Capture and submit`

* visited https://brianbondy.com/blog/184/the-future-of-leo-a-roadmap-for-brave-s-built-in-ai-browser-assistant
* added screenshot using `Leo` via clicking on the `Attachment` -> `Screenshot`
* ensured that the screenshot was added into `input box` without any issues
* ensured that a thumbnail of the screenshot was visible via the `input box`
* once the screenshot was attached, asked `Leo`  the following --> `Do you see any screenshots?`
* ensured `Leo` mentioned there's an attached screenshot and added a summarization of the screenshot
* also ensured that the screenshot thumbnail is appearing within the `input box`

**`Example`** | **`Example`**
----------------|-----------------
Screenshot 2025-09-16 100235 | Screenshot 2025-09-16 100348

### `Test Case #2` - `Capture and remove`

* visited https://brianbondy.com/blog/184/the-future-of-leo-a-roadmap-for-brave-s-built-in-ai-browser-assistant
* added screenshot using `Leo` via clicking on the `Attachment` -> `Screenshot`
* ensured that the screenshot was added into `input box` without any issues
* ensured that a thumbnail of the screenshot was visible via the `input box`
* removed the screenshot that was added into the `input box` by clicking on the `x` and ensured it was removed
* once removed, asked `Leo` --> `Do you see any screenshots?`
* ensured that `Leo` mentions there's no attached screenshot

**`Example`** | **`Example`**
----------------|-----------------
Screenshot 2025-09-16 100626 | Screenshot 2025-09-16 100720

### `Test Case #3` - `Autoscreenshots`

* visited a public GDoc via https://docs.google.com/document/u/1/d/1ZjXC9OClFuiJRxrMko2sRNkDd-H9KU_ryyixoDVnYPQ/mobilebasic
* opened the `Leo` panel and clicked on `Summarize this page`
* ensured that a screenshot thumbnail appears within the conversation and the page is summarized without any issues

image

PR #30996, opened 2025-09-02
Brave Origin Profile policy provider

bbondy:
Resolves https://github.com/brave/brave-browser/issues/48910

### DIFF
--- a/chromium_src/chrome/browser/policy/profile_policy_connector.cc
+++ b/chromium_src/chrome/browser/policy/profile_policy_connector.cc
@@ -1,0 +1,25 @@
+/* Copyright (c) 2025 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "components/policy/core/common/configuration_policy_provider.h"
+
+namespace brave_policy {
+std::unique_ptr<policy::ConfigurationPolicyProvider>
+CreateBraveProfilePolicyProvider();
+}  // namespace brave_policy
+
+// Create and Init the Brave Profile Policy Provider which is used
+// for Brave Origin and other Brave specific policies.
+// We do not need to define a patch for Shutdown since anything in
+// wrapped_policy_providers_ will automatically call Shutdown.
+#define BRAVE_PROFILE_POLICY_CONNECTOR_INIT                         \
+  auto provider = brave_policy::CreateBraveProfilePolicyProvider(); \
+  policy_providers_.push_back(provider.get());                      \
+  provider->Init(schema_registry);                                  \
+  wrapped_policy_providers_.push_back(std::move(provider));
+
+#include <chrome/browser/policy/profile_policy_connector.cc>  // IWYU pragma: export
+
+#undef BRAVE_PROFILE_POLICY_CONNECTOR_INIT
--- a/components/brave_policy/brave_profile_policy_provider.cc
+++ b/components/brave_policy/brave_profile_policy_provider.cc
@@ -3,7 +3,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this file,
  * You can obtain one at https://mozilla.org/MPL/2.0/. */
 
-#include "brave/components/brave_policy/brave_browser_policy_provider.h"
+#include "brave/components/brave_policy/brave_profile_policy_provider.h"
 
 #include <utility>
 
@@ -12,16 +12,18 @@
 #include "components/policy/core/common/policy_bundle.h"
 #include "components/policy/core/common/policy_map.h"
 #include "components/policy/core/common/policy_namespace.h"
+#include "components/policy/core/common/policy_service.h"
 #include "components/policy/core/common/policy_types.h"
+#include "components/policy/policy_constants.h"
 #include "components/prefs/pref_service.h"
+#include "components/prefs/scoped_user_pref_update.h"
 
 namespace brave_policy {
 
-BraveBrowserPolicyProvider::BraveBrowserPolicyProvider() = default;
+BraveProfilePolicyProvider::BraveProfilePolicyProvider() = default;
+BraveProfilePolicyProvider::~BraveProfilePolicyProvider() = default;
 
-BraveBrowserPolicyProvider::~BraveBrowserPolicyProvider() {}
-
-void BraveBrowserPolicyProvider::Init(policy::SchemaRegistry* registry) {
+void BraveProfilePolicyProvider::Init(policy::SchemaRegistry* registry) {
   // Call base class Init first
   policy::ConfigurationPolicyProvider::Init(registry);
 
@@ -31,7 +33,7 @@ void BraveBrowserPolicyProvider::Init(policy::SchemaRegistry* registry) {
   RefreshPolicies(policy::PolicyFetchReason::kBrowserStart);
 }
 
-void BraveBrowserPolicyProvider::RefreshPolicies(
+void BraveProfilePolicyProvider::RefreshPolicies(
     policy::PolicyFetchReason reason) {
   policy::PolicyBundle bundle = LoadPolicies();
 
@@ -41,22 +43,22 @@ void BraveBrowserPolicyProvider::RefreshPolicies(
   UpdatePolicy(std::move(bundle));
 }
 
-bool BraveBrowserPolicyProvider::IsFirstPolicyLoadComplete(
+bool BraveProfilePolicyProvider::IsFirstPolicyLoadComplete(
     policy::PolicyDomain domain) const {
   return first_policies_loaded_;
 }
 
-policy::PolicyBundle BraveBrowserPolicyProvider::LoadPolicies() {
+policy::PolicyBundle BraveProfilePolicyProvider::LoadPolicies() {
   policy::PolicyBundle bundle;
 
-  // Future work will add policies here
+  // Future work will add Brave Origin profile policies here
 
   return bundle;
 }
 
 std::unique_ptr<policy::ConfigurationPolicyProvider>
-CreateBraveBrowserPolicyProvider() {
-  return std::make_unique<BraveBrowserPolicyProvider>();
+CreateBraveProfilePolicyProvider() {
+  return std::make_unique<BraveProfilePolicyProvider>();
 }
 
 }  // namespace brave_policy
--- a/components/brave_policy/BUILD.gn
+++ b/components/brave_policy/BUILD.gn
@@ -7,6 +7,8 @@ static_library("brave_policy") {
   sources = [
     "brave_browser_policy_provider.cc",
     "brave_browser_policy_provider.h",
+    "brave_profile_policy_provider.cc",
+    "brave_profile_policy_provider.h",
   ]
 
   deps = [
@@ -19,7 +21,10 @@ static_library("brave_policy") {
 
 source_set("unit_tests") {
   testonly = true
-  sources = [ "brave_browser_policy_provider_unittest.cc" ]
+  sources = [
+    "brave_browser_policy_provider_unittest.cc",
+    "brave_profile_policy_provider_unittest.cc",
+  ]
 
   deps = [
     ":brave_policy",
--- a/components/brave_policy/README.md
+++ b/components/brave_policy/README.md
@@ -1,0 +1,7 @@
+There are 2 types of policy providers.
+
+- BraveBrowserPolicyProvider: There is only one of these providers and it manages policies that are stored globally in local state.
+- BraveProfilePolicyProvider: There is one of these providers per profile and it manages policies that are stored in profile prefs.
+
+Policies are defined in `components/policy/resources/templates/policy_definitions/BraveSoftware/` and the YAML file has a `per_profile` attribute which associates it accordingly to the browser or profile provider.
+Both of these providers will be used for Brave Origin, but also for other work within Brave.
--- a/components/brave_policy/brave_profile_policy_provider.h
+++ b/components/brave_policy/brave_profile_policy_provider.h
@@ -1,0 +1,44 @@
+/* Copyright (c) 2025 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#ifndef BRAVE_COMPONENTS_BRAVE_POLICY_BRAVE_PROFILE_POLICY_PROVIDER_H_
+#define BRAVE_COMPONENTS_BRAVE_POLICY_BRAVE_PROFILE_POLICY_PROVIDER_H_
+
+#include "base/containers/flat_set.h"
+#include "base/memory/raw_ptr.h"
+#include "base/memory/weak_ptr.h"
+#include "base/scoped_observation.h"
+#include "base/values.h"
+#include "components/policy/core/common/configuration_policy_provider.h"
+
+namespace brave_policy {
+
+// Policy provider for profile level polciies.
+// Note: When this is created, the profile is not yet initialized.
+class BraveProfilePolicyProvider : public policy::ConfigurationPolicyProvider {
+ public:
+  BraveProfilePolicyProvider();
+  ~BraveProfilePolicyProvider() override;
+
+  BraveProfilePolicyProvider(const BraveProfilePolicyProvider&) = delete;
+  BraveProfilePolicyProvider& operator=(const BraveProfilePolicyProvider&) =
+      delete;
+
+  // ConfigurationPolicyProvider implementation.
+  void Init(policy::SchemaRegistry* registry) override;
+  void RefreshPolicies(policy::PolicyFetchReason reason) override;
+  bool IsFirstPolicyLoadComplete(policy::PolicyDomain domain) const override;
+
+ private:
+  policy::PolicyBundle LoadPolicies();
+
+  bool first_policies_loaded_ = false;
+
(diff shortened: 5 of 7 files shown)


components/policy/brave_profile_policy_provider.cc
```diff
@@ -0,0 +1,65 @@
...
+#include "components/policy/core/common/policy_types.h"
+#include "components/policy/policy_constants.h"
+#include "components/prefs/pref_service.h"
+#include "components/prefs/scoped_user_pref_update.h"
+
+namespace brave_policy {
+
+BraveProfilePolicyProvider::BraveProfilePolicyProvider(PrefService* local_state)
+    : policy::ConfigurationPolicyProvider(),
+      first_policies_loaded_(false),
```
DJAndries: nit: could use `bool first_policies_loaded_ = false;` in the header instead

components/policy/brave_profile_policy_provider_unittest.cc
```diff
@@ -0,0 +1,132 @@
...
+#include "testing/gtest/include/gtest/gtest.h"
+
+namespace brave_policy {
+
+class BraveProfilePolicyProviderTest : public ::testing::Test {
+ public:
+  BraveProfilePolicyProviderTest() : provider_(&local_state_) {}
+  ~BraveProfilePolicyProviderTest() override = default;
+
+  void TearDown() override {
```
DJAndries: Is there a reason why we have a teardown procedure here, but not in `BraveBrowserPolicyProviderTest`?

components/policy/brave_profile_policy_provider_unittest.cc
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+ protected:
+  base::test::TaskEnvironment task_environment_;
+  TestingPrefServiceSimple local_state_;
+  policy::SchemaRegistry schema_registry_;
+  BraveProfilePolicyProvider provider_;
+};
+
+TEST_F(BraveProfilePolicyProviderTest, Construction) {
```
DJAndries: Should we mirror the change you made here? https://github.com/brave/brave-core/pull/30984#discussion_r2317517804

chromium_src/chrome/browser/policy/profile_policy_connector_builder.cc
```diff
@@ -0,0 +1,51 @@
...
+}
+
+std::unique_ptr<ProfilePolicyConnector> CreateAndInitProfilePolicyConnector(
+    SchemaRegistry* schema_registry,
+    policy::ChromeBrowserPolicyConnector* browser_policy_connector,
+    ConfigurationPolicyProvider* policy_provider,
+    const CloudPolicyStore* policy_store,
+    bool force_immediate_load,
+    const user_manager::User* user) {
+  return CreateAndInitBraveProfilePolicyConnector(
```
bridiver: I don't understand what is happening here, why are you renaming the original method and then just calling it here with all the same params?

chromium_src/chrome/browser/policy/profile_policy_connector_builder.cc
```diff
@@ -0,0 +1,51 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "base/files/file_path.h"
+#include "chrome/browser/policy/profile_policy_connector.h"
+
+#define ProfilePolicyConnector BraveProfilePolicyConnector
+#define CreateProfilePolicyConnectorForBrowserContext \
+  CreateBraveProfilePolicyConnectorForBrowserContext
```
bridiver: shouldn't this normally be ChromiumImpl so we can easily recognize it as a renamed upstream method?

chromium_src/chrome/browser/policy/profile_policy_connector.h
```diff
@@ -0,0 +1,54 @@
+/* Copyright (c) 2025 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#ifndef BRAVE_CHROMIUM_SRC_CHROME_BROWSER_POLICY_PROFILE_POLICY_CONNECTOR_H_
+#define BRAVE_CHROMIUM_SRC_CHROME_BROWSER_POLICY_PROFILE_POLICY_CONNECTOR_H_
+
+#include "base/timer/timer.h"
+#include "brave/components/policy/brave_profile_policy_provider.h"
```
bridiver: same as the previous PR, we want to use interface/impl to avoid adding these deps to upstream code where they will likely need patches later

chromium_src/chrome/browser/policy/profile_policy_connector_builder.cc
```diff
@@ -0,0 +1,51 @@
...
+    CloudPolicyManager* cloud_policy_manager,
+    ConfigurationPolicyProvider* policy_provider,
+    policy::ChromeBrowserPolicyConnector* browser_policy_connector,
+    bool force_immediate_load,
+    content::BrowserContext* context) {
+  std::unique_ptr<BraveProfilePolicyConnector> connector =
+      CreateBraveProfilePolicyConnectorForBrowserContext(
+          schema_registry, cloud_policy_manager, policy_provider,
+          browser_policy_connector, force_immediate_load, context);
+  connector->GetBraveProfilePolicyProvider()->Init(schema_registry);
```
bridiver: why doesn't this happen in `CreateAndInitProfilePolicyConnector` when it calls `connector->Init`?

chromium_src/chrome/browser/policy/profile_policy_connector.cc
```diff
@@ -0,0 +1,63 @@
...
+void BraveProfilePolicyConnector::Shutdown() {
+  brave_profile_policy_provider_->Shutdown();
+  ProfilePolicyConnector::Shutdown();
+  // Clear raw pointer after shutdown to prevent dangling pointer
+  brave_profile_policy_provider_ = nullptr;
+}
+
+}  // namespace policy
+
+#undef BRAVE_PROFILE_POLICY_CONNECTOR_INIT
```
DJAndries, line 25: where are these defined?

DJAndries (changes requested):

chromium_src/chrome/browser/policy/profile_policy_connector_builder.cc
bbondy: Take a look at `chrome/browser/policy/profile_policy_connector.cc`
There are a lot of differences from how the browser provider works.
All providers are part of the providers vector, but unlike the browser level one, it does not own them. 
There are some providers that are Init by themselves outside of any vector looping, and some providers that are also part of a wrapped_providers (which we don't use) which it loops over to Init those.

chromium_src/chrome/browser/policy/profile_policy_connector_builder.cc
bbondy: The approach needed to be much different from the browser level provider. This one uses a subclass named `BraveProfilePolicyConnector`.  This particular define is in the .cc file and makes it create the `BraveProfilePolicyConnector` instead of the `ProfilePolicyConnector`.

For the other 2 return types, I'll change them to have the `_ChromiumImpl` suffix instead.
I also documented the defines better which will be in the next push:
```
// This renames ProfilePolicyConnector to BraveProfilePolicyConnector
// so that the Brave subclass gets created instead of the Chromium one.
#define ProfilePolicyConnector BraveProfilePolicyConnector

// This renames away the original CreateProfilePolicyConnectorForBrowserContext
// because the ProfilePolicyConnector rename above changes the return type of
// CreateProfilePolicyConnectorForBrowserContext.
// We also need it because we need to call Init on the Brave profile policy
// provider.
// We provide a replacement with the original return type below.
#define CreateProfilePolicyConnectorForBrowserContext \
  CreateProfilePolicyConnectorForBrowserContext_ChromiumImpl

// This needed to be renamed away only because the #define above for
// ProfilePolicyConnector changes the return type of the original function.
// We define a new one with the original return type below.
#define CreateAndInitProfilePolicyConnector \
  CreateAndInitProfilePolicyConnector_ChromiumImpl
```

chromium_src/chrome/browser/policy/profile_policy_connector_builder.cc
bbondy: I explained that here and documented it in code better:
https://github.com/brave/brave-core/pull/30996#discussion_r2323225459

chromium_src/chrome/browser/policy/profile_policy_connector_builder.cc
bbondy: This is resolved in the latest

chromium_src/chrome/browser/policy/profile_policy_connector_builder.cc
bbondy: This is resolved in the latest

chromium_src/chrome/browser/policy/profile_policy_connector.h
bbondy: This is resolved in the latest

components/policy/brave_profile_policy_provider.cc
bbondy: This is resolved in the latest

components/policy/brave_profile_policy_provider_unittest.cc
bbondy: This is resolved in the latest

chromium_src/chrome/browser/policy/profile_policy_connector.cc
bbondy: This was unused and left over. It's removed in the latest.

components/policy/brave_profile_policy_provider_unittest.cc
bbondy: The reason is because in the browser provider there's an `InitAndRefreshLocalState` function and also an `Init` function.  The browser provider's tests only need to test the former because the later doesn't have any patching.  So those don't actually get `Init`.
In the profile provider here, there is no distinction, and we're actually doing the `Init` too.
It's a bit confusing because you'd think the Connector's `Init` would also `Init` the providers for the browser provider, but that's not the case.  The provider's `Init`s for the browser happen here:

```
PolicyService* BrowserPolicyConnectorBase::GetPolicyService() {
  if (g_testing_policy_service)
    return g_testing_policy_service;

  if (policy_service_)
    return policy_service_.get();

  DCHECK(!is_initialized_);
  is_initialized_ = true;

  policy_providers_ = CreatePolicyProviders();

  if (g_testing_provider)
    g_testing_provider->Init(GetSchemaRegistry());

  for (const auto& provider : policy_providers_)
    provider->Init(GetSchemaRegistry());

  g_created_policy_service = true;
  policy_service_ = std::make_unique<PolicyServiceImpl>(
      GetProvidersForPolicyService(),
      PolicyServiceImpl::ScopeForMetrics::kMachine,
      std::vector<std::unique_ptr<PolicyMigrator>>());
  return policy_service_.get();
}

bool BrowserPolicyConnectorBase::HasPolicyService() {
  return g_testing_policy_service || policy_service_;
}

const ConfigurationPolicyHandlerList*
BrowserPolicyConnectorBase::GetHandlerList() const {
  return handler_list_.get();
}
```

If I was coding those Chromium providers I'd probably have things work the same way? But I didn't write that code, I'm just tying into it :)

chromium_src/chrome/browser/policy/profile_policy_connector.h
```diff
@@ -0,0 +1,52 @@
...
+
+  void Init(const user_manager::User* user,
+            SchemaRegistry* schema_registry,
+            ConfigurationPolicyProvider* configuration_policy_provider,
+            const CloudPolicyStore* policy_store,
+            policy::ChromeBrowserPolicyConnector* browser_policy_connector,
+            bool force_immediate_load) override;
+  void Shutdown() override;
+  raw_ptr<policy::ConfigurationPolicyProvider> GetBraveProfilePolicyProvider();
+  const std::vector<raw_ptr<ConfigurationPolicyProvider, VectorExperimental>>&
```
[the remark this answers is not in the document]
bbondy: This just copies the return type of the upstream file, but it's no longer needed because we're going to test things with a browser test. So you can ignore this.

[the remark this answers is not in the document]
bbondy: This just copies the type of the upstream file, but it was only used for a test (which should have had ForTest on it). But it's being removed anyway because we're going to test this code with a browser test. So you can ignore this.

chromium_src/chrome/browser/policy/profile_policy_connector.cc
```diff
@@ -0,0 +1,59 @@
...
+  // CreateProfilePolicyConnectorForBrowserContext because we have the profile
+  // there which we'll get and pass in the profile_id in later work.
+}
+
+raw_ptr<policy::ConfigurationPolicyProvider>
+BraveProfilePolicyConnector::GetBraveProfilePolicyProvider() {
+  return brave_profile_policy_provider_;
+}
+
+const std::vector<raw_ptr<ConfigurationPolicyProvider, VectorExperimental>>&
```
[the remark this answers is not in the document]
bbondy: This is removed, you can ignore.

chromium_src/chrome/browser/policy/profile_policy_connector.h
```diff
@@ -0,0 +1,50 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#ifndef BRAVE_CHROMIUM_SRC_CHROME_BROWSER_POLICY_PROFILE_POLICY_CONNECTOR_H_
+#define BRAVE_CHROMIUM_SRC_CHROME_BROWSER_POLICY_PROFILE_POLICY_CONNECTOR_H_
+
+#include "base/timer/timer.h"
+
+// Add virtual to Init and Shutdown
+#define Init(user, schema_registry, configuration_policy_provider,         \
```
bridiver: did you include all the params intentionally so it didn't match other `Init` methods? If so, please specify that in the comments so no one changes it to `...`/`__VA_ARGS__`

chromium_src/chrome/browser/policy/profile_policy_connector.h
```diff
@@ -0,0 +1,50 @@
...
+             policy_store, browser_policy_connector, force_immediate_load) \
+  virtual Init(user, schema_registry, configuration_policy_provider,       \
+               policy_store, browser_policy_connector, force_immediate_load)
+#define Shutdown()    \
+  virtual Shutdown(); \
+  friend class BraveProfilePolicyConnector
+#define final
```
bridiver: You can't do this, it's undefined behavior to redefine keywords. You will have to patch or find another way to override

bbondy: reverted to the patch way, I don't think there's any other way around this.

chromium_src/chrome/browser/policy/profile_policy_connector.h
bbondy: no longer applicable since it is reverted to the patch way

chromium_src/chrome/browser/policy/profile_policy_connector.cc
```diff
@@ -0,0 +1,25 @@
...
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "components/policy/core/common/configuration_policy_provider.h"
+
+namespace brave_policy {
+std::unique_ptr<policy::ConfigurationPolicyProvider>
+CreateBraveProfilePolicyProvider();
+}  // namespace brave_policy
+
+// Create and Init the Brave Profilei Policy Provider which is used
```
bsclifton:
```suggestion
// Create and Init the Brave Profile Policy Provider which is used
```

components/brave_policy/brave_profile_policy_provider.cc
```diff
@@ -0,0 +1,63 @@
+/* Copyright (c) 2025 The Brave Authors. All rights reserved.
```
bsclifton, line 1: Could be good to add a README.md under `components/brave_policy` describing the role this (profile policy provider) and browser policy provider play.  The major difference here is `BraveBrowserPolicyProvider` enforces global ones (ex: Tor) and this (`BraveProfilePolicyProvider`) is intended to enforce user specific ones (almost all the others), right? I think a helpful note would be `per_profile` field in the YAML helps differentiate between [browser](https://github.com/brave/brave-core/blob/5d214363319385e3acdbe19998b7421dcb227f32/components/policy/resources/templates/policy_definitions/BraveSoftware/TorDisabled.yaml#L20C3-L20C14) and [profile](https://github.com/brave/brave-core/blob/5d214363319385e3acdbe19998b7421dcb227f32/components/policy/resources/templates/policy_definitions/BraveSoftware/BraveNewsDisabled.yaml#L20).

Also curious - we have some files already under `components/policy` which include the actual policies themself. Separate of this PR, should we move those here? or should these files be under `components/policy`?

bridiver: we intentionally moved this out of `component/policy`, what is there now is more of an override and I'm not sure that is the appropriate place for it anyway. I think probably `pack_policy_templates` should just move into `brave/BUILD.gn`

bbondy: Good idea @bsclifton.  I'll wait to add it until the PR that adds the policies though because that will have the meat of the documentation.

bbondy: Actually I added some documentation for what I can in the `README.md` file in `brave/brave_policy` for now. It's in the latest.

bsclifton: Looks great! 😄

PR #31010, opened 2025-09-03
Split `darker theme` related source into a new target

sangwoo108:
This issues is follow up of https://github.com/brave/brave-core/pull/30520

Currently, we're using `!is_android` or doesn't have proper guard at all. This change puts `toolkit_views` to guard `darker theme` code correctly.

Resolves https://github.com/brave/brave-browser/issues/48917


browser/brave_profile_prefs.cc
```diff
@@ -518,9 +518,11 @@ void RegisterProfilePrefs(user_prefs::PrefRegistrySyncable* registry) {
   containers::RegisterProfilePrefs(registry);
 #endif
 
-  if (base::FeatureList::IsEnabled(color::features::kBraveDarkerTheme)) {
-    registry->RegisterBooleanPref(color::prefs::kBraveDarkerMode, false);
+#if defined(TOOLKIT_VIEWS)
```
simonhong: How about using existing `TOOLKIT_VIEWS` guard above?

browser/about_flags.cc
```diff
@@ -498,7 +505,7 @@ constexpr flags_ui::FeatureEntry::Choice kVerticalTabCollapseDelayChoices[] = {
   })
 #else
 #define BRAVE_MIDDLE_CLICK_AUTOSCROLL_FEATURE_ENTRY
-#endif
+#endif  // defined(TOOLKIT_VIEWS)
```
simonhong: this comment should be at line 493?

sangwoo108: Oops, right. THanks:)

browser/brave_profile_prefs.cc
sangwoo108: Good call

browser/themes/brave_theme_service_unittest.cc
```diff
@@ -26,10 +29,11 @@ TEST(BraveThemeServiceTest, GetBraveThemeListTest) {
   EXPECT_EQ(2UL, list.size());
 }
 
+#if defined(TOOLKIT_VIEWS)
```
bridiver: I would move this into a separate unit test

test/BUILD.gn
```diff
@@ -455,9 +455,9 @@ test("brave_unit_tests") {
       "//brave/browser/download/bubble:unit_tests",
       "//brave/browser/renderer_context_menu/test:unit_tests",
       "//brave/browser/ui/bookmark:unittest",
-      "//brave/browser/ui/color",
       "//brave/browser/ui/color:unit_tests",
       "//brave/browser/ui/commands:unit_tests",
+      "//brave/browser/ui/darker_theme",
```
bridiver: this should go away when you move https://github.com/brave/brave-core/pull/31010/files#r2322559749 into darker_theme

browser/themes/brave_theme_service_unittest.cc
sangwoo108: Made "darker_theme_unittest.cc"

test/BUILD.gn
sangwoo108: Done.

PR #31011, opened 2025-09-03
Added web panel type to sidebar

simonhong:
Resolves https://github.com/brave/brave-browser/issues/48919

This PR introduced web panel type and clicking it toggles web panel view.
For now, only empty view is shown. `SidebarWebPanelController` will controll panel view state.

TEST=SidebarBrowserWithWebPanelTest.WebPanelTest,SidebarServiceTest.WebPanelItemTest
Manual test:
1. Launch Brave with `--enable-features=SidebarWebPanel`
2. Add any web item to sidebar by using plus icon
3. Click that item and check panel view is toggled

https://github.com/user-attachments/assets/50819632-9a77-450d-8113-9f63aa9269a0

### DIFF
--- a/browser/ui/views/frame/split_view/brave_multi_contents_view.h
+++ b/browser/ui/views/frame/split_view/brave_multi_contents_view.h
@@ -40,7 +40,7 @@ class BraveMultiContentsView : public MultiContentsView,
   void UpdateSecondaryLocationBar();
   void UpdateCornerRadius();
   void UseContentsContainerViewForWebPanel();
-  void ShowWebPanel(bool show);
+  void SetWebPanelVisible(bool visible);
   bool IsWebPanelVisible() const;
 
   void SetWebPanelWidth(int width);
--- a/browser/ui/views/sidebar/sidebar_items_contents_view.cc
+++ b/browser/ui/views/sidebar/sidebar_items_contents_view.cc
@@ -512,7 +512,7 @@ void SidebarItemsContentsView::OnItemPressed(const views::View* item,
 
   const auto& item_model = controller->model()->GetAllSidebarItems()[*index];
 
-  // web panel is not a side panel that handled by SidePanelCoordinator.
+  // web panel is not a side panel that's handled by SidePanelCoordinator.
   // It'll be loaded into another contents view in MultiContentsView.
   if (item_model.is_web_panel_type()) {
     controller->ActivateItemAt(index);
--- a/components/sidebar/browser/sidebar_service.cc
+++ b/components/sidebar/browser/sidebar_service.cc
@@ -520,7 +520,6 @@ void SidebarService::LoadSidebarItems() {
       } else {
         continue;
       }
-      // Open in panel for custom items is not yet supported
       bool open_in_panel = false;
       if (const auto value = item.FindBool(kSidebarItemOpenInPanelKey)) {
         open_in_panel = *value;
--- a/browser/ui/sidebar/sidebar_web_panel_controller.cc
+++ b/browser/ui/sidebar/sidebar_web_panel_controller.cc
@@ -19,11 +19,11 @@ SidebarWebPanelController::SidebarWebPanelController(BrowserView& browser_view)
 SidebarWebPanelController::~SidebarWebPanelController() = default;
 
 void SidebarWebPanelController::OpenWebPanel(const SidebarItem& item) {
-  GetMultiContentsView()->ShowWebPanel(true);
+  GetMultiContentsView()->SetWebPanelVisible(true);
 }
 
 void SidebarWebPanelController::CloseWebPanel() {
-  GetMultiContentsView()->ShowWebPanel(false);
+  GetMultiContentsView()->SetWebPanelVisible(false);
 }
 
 bool SidebarWebPanelController::IsShowingWebPanel() const {
--- a/browser/ui/views/frame/split_view/brave_multi_contents_view.cc
+++ b/browser/ui/views/frame/split_view/brave_multi_contents_view.cc
@@ -67,9 +67,9 @@ void BraveMultiContentsView::UseContentsContainerViewForWebPanel() {
   }
 }
 
-void BraveMultiContentsView::ShowWebPanel(bool show) {
+void BraveMultiContentsView::SetWebPanelVisible(bool visible) {
   CHECK(contents_container_view_for_web_panel_);
-  contents_container_view_for_web_panel_->SetVisible(show);
+  contents_container_view_for_web_panel_->SetVisible(visible);
   contents_container_view_for_web_panel_->UpdateBorderAndOverlay(true, true,
                                                                  false);
 }


browser/ui/sidebar/sidebar_controller.cc
```diff
@@ -103,9 +104,21 @@ void SidebarController::ActivateItemAt(std::optional<size_t> index,
   DCHECK_LT(index.value(), sidebar_model_->GetAllSidebarItems().size());
 
   const auto& item = sidebar_model_->GetAllSidebarItems()[*index];
+
+  if (IsWebPanelFeatureEnabled() && item.is_web_panel_type()) {
+    GetWebPanelController()->IsShowingWebPanel()
+        ? GetWebPanelController()->CloseWebPanel()
```
sangwoo108, line 112: Should we rename `ActivateItemAt` to something like `ItemWasClicked`? I think closing web panel in "ActivateItemAt" feels something odd.

browser/ui/sidebar/sidebar_web_panel_controller.h
```diff
@@ -0,0 +1,36 @@
...
+  explicit SidebarWebPanelController(BrowserView& browser);
+  ~SidebarWebPanelController();
+
+  void OpenWebPanel(const SidebarItem& item);
+  void CloseWebPanel();
+  bool IsShowingWebPanel() const;
+
+ private:
+  BraveMultiContentsView* GetMultiContentsView();
+  BraveMultiContentsView* GetMultiContentsView() const;
```
sangwoo108: Should the return type be `const` too?

```suggestion
  const BraveMultiContentsView* GetMultiContentsView() const;
```

browser/ui/sidebar/sidebar_controller.cc
```diff
@@ -103,9 +104,21 @@ void SidebarController::ActivateItemAt(std::optional<size_t> index,
...
+  }
+
   // Only an item for panel can get activated.
   if (item.open_in_panel) {
-    sidebar_model_->SetActiveIndex(index);
+    // TODO(https://github.com/brave/brave-browser/issues/33533): WebPanel item
+    // type also should be activated.
+    if (!item.is_web_panel_type()) {
+      sidebar_model_->SetActiveIndex(index);
+    }
```
sangwoo108: Q. I'm confused why this was added.

simonhong: This web panel type item also should be in active state like other builtin panel(ex, playlist). but it's not ready yet. In this PR, clicking item just toggles web panel. Later we should call SetActiveIndex for web panel type.

browser/ui/sidebar/sidebar_web_panel_controller.h
simonhong: Updated.

browser/ui/sidebar/sidebar_controller.cc
simonhong: So far, we used `ActivateItemAt()` for toggling panel. As you can see passing null for `index`, it'll just close panel.
As web panel item can't have active state, UI calls `ActivateItemAt()` with non-null index now.
In the future, web panel item also can be set as an active item, then calling this method with null will close it instead of close explicitely here.

zeusshomaly: > In the future, web panel item also can be set as an active item

So when you say 'In the future, web panel item also can be set as an active item' Does that also apply to music resuming if played via side panel and you minimise the side panel? as audio should continue if it's paused obviously ;)

sangwoo108: Sounds okay if it's temporary.

browser/ui/sidebar/sidebar_controller.cc
sangwoo108: I see. Thanks :) One more question, I'm curious if all items in |open_in_panel| condition are not web panel types because it seems
* When the feature is enabled, returning from L112?
* When the feature is disabled, open_in_panel can't be true with web panel type?

sangwoo108: Oh, enabling feature -> add web panel -> disabling the feature can make this happen.

browser/ui/sidebar/sidebar_web_panel_controller.h
```diff
@@ -0,0 +1,36 @@
...
+class BraveMultiContentsView;
+
+namespace sidebar {
+
+struct SidebarItem;
+
+class SidebarWebPanelController {
+ public:
+  explicit SidebarWebPanelController(BrowserView& browser);
+  ~SidebarWebPanelController();
```
sangwoo108, line 21: Should we have some code to prevent copying and moving?

browser/ui/views/sidebar/sidebar_items_contents_view.cc
```diff
@@ -511,6 +511,13 @@ void SidebarItemsContentsView::OnItemPressed(const views::View* item,
   }
 
   const auto& item_model = controller->model()->GetAllSidebarItems()[*index];
+
+  // web panel is not a side panel.
```
sangwoo108: I'm not sure what you meant by "not a side panel". Could you elaborate on this?

browser/ui/sidebar/sidebar_controller.cc
```diff
@@ -260,4 +273,14 @@ void SidebarController::SetSidebar(Sidebar* sidebar) {
       profile_, ServiceAccessType::EXPLICIT_ACCESS));
 }
 
+SidebarWebPanelController* SidebarController::GetWebPanelController() {
+  CHECK(IsWebPanelFeatureEnabled());
```
sangwoo108, line 275: How about having this check inside the SidebarWebPanelController's constructor instead of this place?

browser/ui/sidebar/sidebar_web_panel_controller.h
simonhong: Added.

browser/ui/sidebar/sidebar_controller.cc
simonhong: Added there.

browser/ui/sidebar/sidebar_controller.cc
simonhong: Thanks for catching this.
When feature is disabled, all web type should be loaded in a tab. Fixed in that way.

browser/ui/views/sidebar/sidebar_items_contents_view.cc
simonhong: Ok, needs more detail here. Updated.
```
  // web panel is not a side panel that handled by SidePanelCoordinator.
  // It'll be loaded into another contents view in MultiContentsView.
```

browser/ui/views/sidebar/sidebar_items_contents_view.cc
```diff
@@ -511,7 +511,15 @@ void SidebarItemsContentsView::OnItemPressed(const views::View* item,
   }
 
   const auto& item_model = controller->model()->GetAllSidebarItems()[*index];
-  if (item_model.open_in_panel) {
+
+  // web panel is not a side panel that handled by SidePanelCoordinator.
```
sangwoo108: nit: typo

```suggestion
  // web panel is not a side panel that's handled by SidePanelCoordinator.
```

browser/ui/views/frame/split_view/brave_multi_contents_view.h
```diff
@@ -41,6 +40,8 @@ class BraveMultiContentsView : public MultiContentsView,
   void UpdateSecondaryLocationBar();
   void UpdateCornerRadius();
   void UseContentsContainerViewForWebPanel();
+  void ShowWebPanel(bool show);
```
sangwoo108: optional: SetVisible(bool visible) could be better

components/sidebar/browser/sidebar_service.cc
```diff
@@ -522,6 +522,9 @@ void SidebarService::LoadSidebarItems() {
       }
       // Open in panel for custom items is not yet supported
```
sangwoo108: nit: Could we remove this line?

sangwoo108 (approved):
LGTM

browser/ui/views/frame/split_view/brave_multi_contents_view.h
simonhong: Do you mean `SetWebPanelVisible()`? `SetVisible()` is ambiguous.

sangwoo108: Oh yes. That's better

simonhong: Thanks. done.

browser/ui/views/sidebar/sidebar_items_contents_view.cc
simonhong: fixed.

components/sidebar/browser/sidebar_service.cc
simonhong: Removed.

PR #31012, opened 2025-09-03
add minimal PolkadotWalletService bridge

cmazakas:
Resolves https://github.com/brave/brave-browser/issues/49056

Implement initial incomplete PolkadotWalletService bridge, enabling the front-end to trigger calls to a Polkadot testnet via core code.


components/brave_wallet/browser/polkadot/polkadot_rpc.cc
```diff
@@ -0,0 +1,71 @@
...
+                          url_loader_factory) {}
+
+PolkadotRpc::~PolkadotRpc() = default;
+
+void PolkadotRpc::GetChainName(
+    base::OnceCallback<void(const std::string&)> callback) {
+  std::string method = "POST";
+  GURL url("https://westend-rpc.polkadot.io");
+  std::string payload =
+      R"({"id":1, "jsonrpc":"2.0", "method": "system_chain", "params":[]})";
```
cdesouza-chromium: In a follow-up introduce an json schema compiler IDL types for the request and response payload.

components/brave_wallet/browser/polkadot/polkadot_rpc.cc
```diff
@@ -0,0 +1,71 @@
...
+
+  std::string payload_content_type = "application/json";
+
+  api_request_helper_.Request(
+      method, url, payload, payload_content_type,
+      base::BindOnce(
+          [](base::WeakPtr<PolkadotRpc> weak,
```
cdesouza-chromium: this weak_ptr is unused.

components/brave_wallet/browser/polkadot/polkadot_rpc.h
```diff
@@ -0,0 +1,32 @@
...
+#ifndef BRAVE_COMPONENTS_BRAVE_WALLET_BROWSER_POLKADOT_POLKADOT_RPC_H_
+#define BRAVE_COMPONENTS_BRAVE_WALLET_BROWSER_POLKADOT_POLKADOT_RPC_H_
+
+#include "base/memory/weak_ptr.h"
+#include "brave/components/api_request_helper/api_request_helper.h"
+#include "services/network/public/cpp/shared_url_loader_factory.h"
+
+namespace brave_wallet {
+
+class PolkadotRpc {
```
cdesouza-chromium: Add class/member functions/field documentation.

components/brave_wallet/browser/polkadot/polkadot_wallet_service.cc
```diff
@@ -0,0 +1,42 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "brave/components/brave_wallet/browser/polkadot/polkadot_wallet_service.h"
+
+namespace brave_wallet {
+
+PolkadotWalletService::PolkadotWalletService(
+    scoped_refptr<network::SharedURLLoaderFactory> url_loader_factory)
+    : polkadot_rpc_(url_loader_factory) {}
```
cdesouza-chromium: nit: `std::move`?

components/brave_wallet/browser/polkadot/polkadot_wallet_service.cc
```diff
@@ -0,0 +1,42 @@
...
+
+void PolkadotWalletService::GetNetworkName(GetNetworkNameCallback callback) {
+  auto weak = weak_ptr_factory_.GetWeakPtr();
+
+  base::SequencedTaskRunner::GetCurrentDefault()->PostTask(
+      FROM_HERE,
+      base::BindOnce(
+          [](base::WeakPtr<brave_wallet::PolkadotWalletService> weak_ptr,
+             GetNetworkNameCallback callback) {
+            weak_ptr->polkadot_rpc_.GetChainName(base::BindOnce(
```
cdesouza-chromium: This here is not correct, because you are passing the weak ptr into the lambda, so the weak_ptr could be invalidated already by the time the lambda gets called. Rather, have a `PolkadotWalletService::OnGetNetworkName`, and bind it with the weakptr, because that guarantees the completion only triggers if the weakptr is valid, as in:

```cxx
base::BindOnce(
    &PolkadotWalletService::GetNetworkName,
    weak_ptr_factory_.GetWeakPtr(),
   std::move(callback))
```

The function should be something like:

```cxx
void PolkadotWalletService::OnGetNetworkName(GetNetworkNameCallback callback, const std::string& str)
```

components/brave_wallet/browser/polkadot/polkadot_wallet_service.h
```diff
@@ -0,0 +1,38 @@
...
+#define BRAVE_COMPONENTS_BRAVE_WALLET_BROWSER_POLKADOT_POLKADOT_WALLET_SERVICE_H_
+
+#include "brave/components/brave_wallet/browser/polkadot/polkadot_rpc.h"
+#include "brave/components/brave_wallet/common/brave_wallet.mojom.h"
+#include "mojo/public/cpp/bindings/receiver_set.h"
+#include "services/network/public/cpp/shared_url_loader_factory.h"
+
+namespace brave_wallet {
+
+class PolkadotWalletService : public mojom::PolkadotWalletService {
```
cdesouza-chromium, line 17: Same about documentation.

components/brave_wallet/browser/polkadot/polkadot_wallet_service.cc
```diff
@@ -0,0 +1,42 @@
...
+PolkadotWalletService::~PolkadotWalletService() = default;
+
+void PolkadotWalletService::Bind(
+    mojo::PendingReceiver<mojom::PolkadotWalletService> receiver) {
+  receivers_.Add(this, std::move(receiver));
+}
+
+void PolkadotWalletService::Reset() {
+  weak_ptr_factory_.InvalidateWeakPtrs();
+}
```
cdesouza-chromium, line 23: There doesn't seem to be any use for this function...

components/brave_wallet/browser/polkadot/polkadot_rpc.h
cypt4: I'd call this PolkadotSubstrateRpc if it follows Substrate format calls

components/brave_wallet/browser/polkadot/polkadot_rpc.cc
```diff
@@ -0,0 +1,71 @@
...
+
+PolkadotRpc::PolkadotRpc(
+    scoped_refptr<network::SharedURLLoaderFactory> url_loader_factory)
+    : api_request_helper_(GetNetworkTrafficAnnotationTag(),
+                          url_loader_factory) {}
+
+PolkadotRpc::~PolkadotRpc() = default;
+
+void PolkadotRpc::GetChainName(
+    base::OnceCallback<void(const std::string&)> callback) {
```
cypt4: nit: introduce GetChainNameCallback

components/brave_wallet/browser/polkadot/polkadot_rpc.cc
```diff
@@ -0,0 +1,71 @@
...
+using APIRequestResult = api_request_helper::APIRequestResult;
+
+namespace {
+
+net::NetworkTrafficAnnotationTag GetNetworkTrafficAnnotationTag() {
+  return net::DefineNetworkTrafficAnnotation("polkadot_rpc", R"(
+      semantics {
+        sender: "Polkadot RPC"
+        description:
+          "This service is used to communicate with Polkadot nodes "
```
cypt4: nit: Polkadot substrate

components/brave_wallet/browser/polkadot/polkadot_wallet_service.cc
cmazakas: These should be logically equivalent. Binding a member function pointer with a weak_ptr should be completely equivalent to a lambda capture with a weak ptr, and because we don't have stateful lambdas, should be equivalent to a bound function with a weak_ptr.

How do we test this?

cmazakas: Alright, I was definitely wrong, it seems. I'm not sure how BindOnce works now but I added a mutation test:
```cpp
void PolkadotWalletService::GetNetworkName(GetNetworkNameCallback callback) {
  auto weak = weak_ptr_factory_.GetWeakPtr();

  base::SequencedTaskRunner::GetCurrentDefault()->PostTask(
      FROM_HERE, base::BindOnce(&PolkadotWalletService::StartGetNetworkName,
                                weak, std::move(callback)));

  std::this_thread::sleep_for(std::chrono::milliseconds(10));
  weak_ptr_factory_.InvalidateWeakPtrs();
}
```
and then instead of the browser crashing with:
```
[33232:36028:0903/103611.482:FATAL:base\memory\weak_ptr.h:250] Check failed: ref_.IsValid().
```
I now see:
```
[7932:28704:0903/103939.969:FATAL:ave\components\brave_wallet\common\brave_wallet.mojom.cc:63392] DCHECK failed: !connected. PolkadotWalletService::GetNetworkNameCallback was destroyed without first either being run or its corresponding binding being closed. It is an error to drop response callbacks which still correspond to an open interface pipe.
```
I don't inherently understand it but it seems like you're 112% correct.

components/brave_wallet/browser/polkadot/polkadot_rpc.cc
cmazakas: Making a note of this now. Will do!

components/brave_wallet/browser/polkadot/polkadot_rpc.cc
cmazakas: Good catch. This has been removed.

components/brave_wallet/browser/polkadot/polkadot_rpc.h
cmazakas: Done!

components/brave_wallet/browser/polkadot/polkadot_wallet_service.cc
cmazakas: Done!

components/brave_wallet/browser/polkadot/polkadot_wallet_service.cc
cmazakas: I read the docs in `//docs/callback.md` which explained everything. Resolving this convo now.

components/brave_wallet/browser/polkadot/polkadot_rpc.cc
cmazakas: Good catch. The code should aim to look like the other RPC classes we already use.

components/brave_wallet/browser/polkadot/polkadot_rpc.cc
cmazakas: Done!

components/brave_wallet/browser/polkadot/polkadot_rpc.h
cmazakas: Good idea. I've updated it now.

components/brave_wallet/browser/polkadot/polkadot_wallet_service.cc
cmazakas: I've fixed this by making sure `Reset()` is called properly by the parent BraveWalletService.

components/brave_wallet/browser/polkadot/polkadot_wallet_service.h
cmazakas: Done!

components/brave_wallet/browser/polkadot/polkadot_wallet_service.cc
```diff
@@ -0,0 +1,50 @@
...
+  auto weak = weak_ptr_factory_.GetWeakPtr();
+  polkadot_substrate_rpc_.GetChainName(base::BindOnce(
+      &PolkadotWalletService::OnGetnetworkName, weak, std::move(callback)));
+}
+
+void PolkadotWalletService::OnGetnetworkName(GetNetworkNameCallback callback,
+                                             std::string const& str) {
+  std::move(callback).Run(str);
+}
+
```
supermassive: This should be just
```
void PolkadotWalletService::GetNetworkName(GetNetworkNameCallback callback) {
    polkadot_substrate_rpc_.GetChainName(std::move(callback)))
}
```

components/brave_wallet/browser/polkadot/polkadot_wallet_service.h
```diff
@@ -0,0 +1,47 @@
...
+  void Reset();
+
+  // Get the name of the chain currently pointed to by the current network
+  // configuration.
+  void GetNetworkName(GetNetworkNameCallback callback) override;
+
+ private:
+  void StartGetNetworkName(GetNetworkNameCallback callback);
+  void OnGetnetworkName(GetNetworkNameCallback callback,
+                        std::string const& str);
```
supermassive: `const std::string& str` is much more widely used form and should be preferred

components/brave_wallet/browser/polkadot/polkadot_wallet_service.h
```diff
@@ -0,0 +1,47 @@
...
+  // Adds a PolkadotWalletService to the internal ReceiverSet.
+  void Bind(mojo::PendingReceiver<mojom::PolkadotWalletService> receiver);
+
+  // Invalidates all the weak ptrs in use by this service.
+  void Reset();
+
+  // Get the name of the chain currently pointed to by the current network
+  // configuration.
+  void GetNetworkName(GetNetworkNameCallback callback) override;
+
```
supermassive, line 33: Not sure I understand the concept of what is "Current polkadot network" for wallet. Most wallet operations are done with network passed as argument(and/or account if it infers network). Then corresponding rpc object should be picked to serve JSON RPC request.

supermassive: But let's keep it for now, and cleanup interfaces later

components/brave_wallet/browser/polkadot/polkadot_rpc.cc
```diff
@@ -0,0 +1,70 @@
...
+      method, url, payload, payload_content_type,
+      base::BindOnce(
+          [](base::OnceCallback<void(const std::string&)> callback,
+             APIRequestResult res) {
+            const auto& body = res.value_body();
+            const auto* chain_name = body.GetDict().Find("result");
+
+            std::move(callback).Run(chain_name->GetString());
+          },
+          std::move(callback)));
```
supermassive: Let' have `PolkadotSubstrateRpc::OnGetChainName` so it becomes 
```
  api_request_helper_.Request(
      method, url, payload, payload_content_type, 
      base::BindOnce( &PolkadotSubstrateRpc::OnGetChainName, weak_ptr_factory_.GetWeakPtr(), std::move(callback));
```

components/brave_wallet/browser/polkadot/polkadot_rpc.cc
```diff
@@ -0,0 +1,70 @@
...
+  api_request_helper_.Request(
+      method, url, payload, payload_content_type,
+      base::BindOnce(
+          [](base::OnceCallback<void(const std::string&)> callback,
+             APIRequestResult res) {
+            const auto& body = res.value_body();
+            const auto* chain_name = body.GetDict().Find("result");
+
+            std::move(callback).Run(chain_name->GetString());
+          },
```
supermassive: We must check that body is a dict and result is a string.
This method also should support responding with errors: 
`using GetChainNameCallback = base::OnceCallback<void(base::expected<std::string, std::string>)>;`

components/brave_wallet/common/brave_wallet.mojom
```diff
@@ -2290,6 +2292,10 @@ interface ZCashWalletService {
       => (ZCashChainTipStatus? status, string? error_message);
 };
 
+interface PolkadotWalletService {
+  GetNetworkName() => (string network);
```
supermassive: Not sure if we need error_message now, but should support errors is the process of getting network name
`string? network`

components/brave_wallet/browser/polkadot/polkadot_wallet_service.cc
cmazakas: Yeah, this is a good simplification. Done!

components/brave_wallet/browser/polkadot/polkadot_wallet_service.h
cmazakas: Yeah, when we parameterize these methods based on stuff like the `mojom::AccountIdPtr`, we'll have to remember to update the comment as well.

components/brave_wallet/browser/polkadot/polkadot_wallet_service.h
cmazakas: Because the code just uses the RPC service directly, this comment no longer applies because the callbacks are now deleted.

components/brave_wallet/browser/polkadot/polkadot_rpc.cc
cmazakas: I added the error handling, as you suggested, but I wasn't sure how to make it all work with `base::expected`, because I couldn't get mojo to generate that type. Instead, it still expected a callback with `void(const optional<string>&, const optional<string>&)`.

components/brave_wallet/common/brave_wallet.mojom
cmazakas: Added this!

components/brave_wallet/browser/polkadot/polkadot_rpc.cc
cmazakas: Done!

components/brave_wallet/browser/polkadot/polkadot_rpc.cc
supermassive: two optional strings is ok for me

components/brave_wallet/browser/polkadot/polkadot_rpc.h
```diff
@@ -0,0 +1,42 @@
...
+
+#include "base/memory/weak_ptr.h"
+#include "brave/components/api_request_helper/api_request_helper.h"
+#include "services/network/public/cpp/shared_url_loader_factory.h"
+
+namespace brave_wallet {
+
+// The main driver for the Polkadot-based RPC calls against the relay chain and
+// the Substrate-based parachains.
+class PolkadotSubstrateRpc {
```
cypt4: Please rename file too

cmazakas: Done!

components/brave_wallet/browser/polkadot/polkadot_wallet_service_unittest.cc
```diff
@@ -0,0 +1,8 @@
+/* Copyright (c) 2025 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "testing/gtest/include/gtest/gtest.h"
+
+TEST(PolkadotWalletService, HelloWorld) {}
```
supermassive, line 8: There is `PolkadotWalletService::GetNetworkName` public method which needs a unit test

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.cc
```diff
@@ -0,0 +1,88 @@
...
+}  // namespace
+
+PolkadotSubstrateRpc::PolkadotSubstrateRpc(
+    scoped_refptr<network::SharedURLLoaderFactory> url_loader_factory)
+    : api_request_helper_(GetNetworkTrafficAnnotationTag(),
+                          url_loader_factory) {}
+
+PolkadotSubstrateRpc::~PolkadotSubstrateRpc() = default;
+
+void PolkadotSubstrateRpc::GetChainName(GetChainNameCallback callback) {
```
supermassive, line 48: This needs a unit test

components/brave_wallet/browser/polkadot/polkadot_wallet_service_unittest.cc
cmazakas: adding this in a follow-up PR.

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.cc
cmazakas: adding this in a follow-up PR!

PR #31015, opened 2025-09-03
Allow PDF files to be drag and dropped

bbondy:
Resolves https://github.com/brave/brave-browser/issues/48936

### DIFF
--- a/components/ai_chat/resources/page/utils/file_utils.test.ts
+++ b/components/ai_chat/resources/page/utils/file_utils.test.ts
@@ -3,7 +3,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this file,
  * You can obtain one at https://mozilla.org/MPL/2.0/. */
 
-import { convertFileToUploadedFile, FileReadError, ImageProcessingError } from './file_utils'
+import { convertFileToUploadedFile, FileReadError, ImageProcessingError, UnsupportedFileTypeError } from './file_utils'
 import * as Mojom from '../../common/mojom'
 import getAPI from '../api'
 
@@ -19,7 +19,6 @@ jest.mock('$web-common/loadTimeData', () => ({
     getBoolean: jest.fn().mockReturnValue(false)
   }
 }))
-
 
 describe('convertFileToUploadedFile', () => {
   let mockHandler: any
@@ -97,7 +96,6 @@ describe('convertFileToUploadedFile', () => {
     return new File(['mock content'], name, { type, size: size })
   }
 
-
   describe('successful file processing', () => {
     it('converts image file successfully', async () => {
       const file = createMockFile('test.png', 'image/png', 1500)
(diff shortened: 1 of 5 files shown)


components/ai_chat/resources/page/utils/file_utils.ts
```diff
@@ -19,11 +19,17 @@ export const convertFileToUploadedFile = (
       }
 
       const uint8Array = new Uint8Array(arrayBuffer)
+
+      // Determine file type based on MIME type
+      const fileType = file.type.toLowerCase() === 'application/pdf'
+        ? Mojom.UploadedFileType.kPdf
+        : Mojom.UploadedFileType.kImage
```
darkdh: can we error out instead of fallback to image if it is type we can't handle?

components/ai_chat/resources/page/utils/file_utils.test.ts
```diff
@@ -0,0 +1,111 @@
...
+      const result = await convertFileToUploadedFile(imageFile)
+
+      expect(result.type).toBe(Mojom.UploadedFileType.kImage)
+      expect(result.filename).toBe(format.name)
+    }
+  })
+
+  it('converts file content to byte array', async () => {
+    const testContent = 'test file content'
+    const file = createMockFile('test.txt', 'text/plain', testContent)
```
darkdh: can we use known type here?

components/ai_chat/resources/page/utils/file_utils.test.ts
```diff
@@ -0,0 +1,111 @@
...
+    expect(result.data.length).toBeGreaterThan(0)
+    expect(
+      result.data.every(
+        byte => typeof byte === 'number' && byte >= 0 && byte <= 255
+      )
+    ).toBe(true)
+  })
+
+  it('rejects when FileReader fails', async () => {
+    const file = createMockFile('test.txt', 'text/plain')
```
darkdh: ditto and this would throw different error per comment below

components/ai_chat/resources/page/utils/file_utils.ts
bbondy: Addressed in the latest

components/ai_chat/resources/page/utils/file_utils.test.ts
bbondy: Addressed in the latest

kjozwiak:

```
Brave | 1.84.67 Chromium: 140.0.7339.133 (Official Build) nightly (64-bit)
-- | --
Revision | d259a1123fc752c0bc730579a40dd755759e9f62
```

Using the STR/Cases outlined via https://github.com/brave/brave-browser/issues/48936#issue-3380835125, ensured that you can drag & drop PDF's into `Leo's` input box/field without any issues as per the following:

### Dragging & Dropping `Emerging_Tech_Bitcoin_Crypto.pdf`

**`Example`** | **`Example`** | **`Example`**
---------------|-----------------|----------------
image | image | image

### Dragging & Dropping `BasicAttentionTokenWhitePaper-4.pdf`

**`Example`** | **`Example`** | **`Example`**
---------------|-----------------|----------------
image | image | image

PR #31018, opened 2025-09-03
Hit all 4 edges in quick action search widget

SergeyZhukovsky:
Search only form should not hit top anb bottom but a view with top tiles should hit all 4 edges.
Note: Android 10 and 11 still use an old approach as `views.setViewLayoutHeight` is supported Android 12+.
Resolves: https://github.com/brave/brave-browser/issues/48943

[Screen_recording_20250903_154125.webm](https://github.com/user-attachments/assets/eeaaac96-0b0e-4aec-ac98-b3ee10e1f32d)

### DIFF
--- a/android/java/org/chromium/chrome/browser/widget/quickactionsearchandbookmark/QuickActionSearchAndBookmarkWidgetProvider.java
+++ b/android/java/org/chromium/chrome/browser/widget/quickactionsearchandbookmark/QuickActionSearchAndBookmarkWidgetProvider.java
@@ -93,12 +93,11 @@ public class QuickActionSearchAndBookmarkWidgetProvider extends AppWidgetProvide
 
     public static String FROM_SETTINGS = "FROM_SETTINGS";
 
-    private static final int TOTAL_TILES = 16;
+    private static final int TOTAL_TILES = 12;
     private static final int TILES_PER_ROW = 4;
     private static final int MIN_VISIBLE_HEIGHT_ROW_1 = 125;
-    private static final int MIN_VISIBLE_HEIGHT_ROW_2 = 220;
-    private static final int MIN_VISIBLE_HEIGHT_ROW_3 = 252;
-    private static final int MIN_VISIBLE_HEIGHT_ROW_4 = 320;
+    private static final int MIN_VISIBLE_HEIGHT_ROW_2 = 150;
+    private static final int MIN_VISIBLE_HEIGHT_ROW_3 = 210;
     private static final int DESIRED_ICON_SIZE = 44;
     private static final int DESIRED_ICON_RADIUS = 16;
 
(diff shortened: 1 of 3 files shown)


android/java/org/chromium/chrome/browser/widget/quickactionsearchandbookmark/QuickActionSearchAndBookmarkWidgetProvider.java
```diff
@@ -93,12 +93,11 @@ public void accept(SearchActivityPreferences prefs) {
 
     public static String FROM_SETTINGS = "FROM_SETTINGS";
 
-    private static final int TOTAL_TILES = 16;
+    private static final int TOTAL_TILES = 12;
```
SergeyZhukovsky, line 96: We support only 12 top tiles at the moment(4 per row, so 3 rows in total), it could be we had 4, but upstream saves only 12 right now.

android/java/res/layout/quick_action_search_and_bookmark_widget_layout.xml
```diff
@@ -469,133 +468,7 @@
         </LinearLayout>
     </LinearLayout>
 
-    <LinearLayout
-        android:id="@+id/BookmarkLayoutRow4"
```
SergeyZhukovsky, line 473: removed the whole row 4 as there are only 3 rows right now.

android/java/res/xml/quick_action_search_and_bookmark_widget_info.xml
```diff
@@ -8,6 +8,7 @@
     android:initialLayout="@layout/quick_action_search_and_bookmark_widget_layout"
     android:minWidth="304dp"
     android:minHeight="44dp"
+    android:maxResizeHeight="270dp"
```
SergeyZhukovsky, line 11: added the max resize height otherwise it could be expanded on the whole home screen and we have an empty space only after row 3.

AlexeyBarabash (approved):
++

kjozwiak:
Verification **PASSED** on `Samsung S25 Ultra` running `Android 15` using the following build(s):

```
Brave | 1.84.54 Chromium: 140.0.7339.133 (Official Build) canary (64-bit)
--- | ---
Revision | d9a54cf62f3c98a2a6dcc416061ff41d281fa870
```

Using the STR/information via https://github.com/brave/brave-core/pull/31018#issue-3381055606 & https://github.com/brave/brave-browser/issues/48943#issue-3381026772, ensured that the widget is hitting all corners/sides as per the following:

**`Example`** | **`Example`** | **`Example`** | **`Example`** 
----------------|----------------|----------------|------------------
Screenshot_20250912_222511_One UI Home | Screenshot_20250912_222516_One UI Home | Screenshot_20250912_222523_One UI Home | Screenshot_20250912_222530_One UI Home

PR #30910, opened 2025-08-28
[Rewards] Do not filter out zero-valued ads from view counts

zenparsing:
Resolves https://github.com/brave/brave-browser/issues/48799

PR #31019, opened 2025-09-03
do not count Ads in Rewards whose value is 0.0 BAT for connected users

tmancey:
Resolves https://github.com/brave/brave-browser/issues/48929

### DIFF
--- a/components/brave_ads/core/internal/account/statement/ads_summary_util.cc
+++ b/components/brave_ads/core/internal/account/statement/ads_summary_util.cc
@@ -6,6 +6,7 @@
 #include "brave/components/brave_ads/core/internal/account/statement/ads_summary_util.h"
 
 #include "base/time/time.h"
+#include "brave/components/brave_ads/core/internal/settings/settings.h"
 #include "brave/components/brave_ads/core/mojom/brave_ads.mojom.h"
 
 namespace brave_ads {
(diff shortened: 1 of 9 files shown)


components/brave_ads/core/internal/account/statement/ads_summary_util.cc
```diff
@@ -20,7 +21,9 @@ base::flat_map<mojom::AdType, int32_t> GetAdsSummaryForDateRange(
     if (transaction.confirmation_type ==
             mojom::ConfirmationType::kViewedImpression &&
         transaction.created_at >= from_time &&
-        transaction.created_at <= to_time) {
+        transaction.created_at <= to_time &&
+        (UserHasJoinedBraveRewardsAndNotConnectedWallet() ||
```
aseren: `UserHasJoinedBraveRewardsAndNotConnectedWallet()` does several sync mojom calls, so it is better not to call it inside the loop. Maybe we can assign local boolean variable with it outside of the loop?

tmancey: Great catch, fixed. Thanks

zenparsing (approved):
Not sure if you want to in this PR, but we can also now re-enable the "earnings" tooltips that were hidden [here](https://github.com/brave/brave-core/pull/30910/files#diff-5fff7b0e8ace18b6503de9d73893673a76f955cde4c8be4377658d43a28cfbcf) and [here](https://github.com/brave/brave-core/pull/30910/files#diff-bfeccc92126f936e4f99b2e4b7788edc15d72cc701c98dbdc35b1e27e09ab131).

tmancey:
@zenparsing as above I removed `display: none`, if you could please re-review. Thanks

PR #31020, opened 2025-09-03
[Settings]: Reduce spacing between menu items

aguscruiz:
Resolves https://github.com/brave/brave-browser/issues/48940

Results:
image

image

### DIFF
--- a/browser/resources/settings/brave_overrides/settings_menu.ts
+++ b/browser/resources/settings/brave_overrides/settings_menu.ts
@@ -63,14 +63,11 @@ RegisterStyleOverride(
         position: sticky;
         top: var(--brave-settings-menu-margin-v);
         margin: 0 !important;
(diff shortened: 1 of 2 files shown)


aguscruiz:
@simonhong @sangwoo108 if I could bother you for a review here I'd appreciate it. It's a simple fix to tighten up the settings menu and remove poppins from the titles font

browser/resources/settings/brave_overrides/settings_menu.ts
```diff
@@ -98,6 +95,10 @@ RegisterStyleOverride(
         display: none !important;
       }
 
+      .menu-separator {
+        margin: 4px -24px !important;
```
fallaciousreasoning, line 99: is that to have it reach full from left to right?

fallaciousreasoning (approved):

browser/resources/settings/brave_overrides/settings_menu.ts
aguscruiz: Yeah, because there's this weird padding that's being applied that I can't get rid of, and this compensates that

PR #31027, opened 2025-09-03
Implement Email Aliases Manage API

arthuredelstein:
Resolves https://github.com/brave/brave-browser/issues/48945

### DIFF
--- a/components/email_aliases/BUILD.gn
+++ b/components/email_aliases/BUILD.gn
@@ -3,6 +3,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at https://mozilla.org/MPL/2.0/.
 
+import("//brave/build/buildflag_header_with_env_config.gni")
 import("//mojo/public/tools/bindings/mojom.gni")
 import("//tools/json_schema_compiler/json_schema_api.gni")
 
@@ -32,12 +33,18 @@ static_library("features") {
   public_deps = [ "//base" ]
 }
 
+buildflag_header_with_env_config("email_aliases_api_key") {
+  header = "email_aliases_api_key.h"
+  env_config_flags = [ "EMAIL_ALIASES_API_KEY" ]
+}
+
 static_library("service") {
   sources = [
     "email_aliases_service.cc",
     "email_aliases_service.h",
   ]
   deps = [
+    ":email_aliases_api_key",
     ":email_aliases_api_types",
     ":mojom",
     "//base",
--- a/components/email_aliases/email_aliases_service.cc
+++ b/components/email_aliases/email_aliases_service.cc
@@ -11,16 +11,20 @@
 
 #include "absl/strings/str_format.h"
 #include "base/check.h"
+#include "base/feature_list.h"
 #include "base/json/json_reader.h"
 #include "base/json/json_writer.h"
 #include "base/logging.h"
-#include "base/notimplemented.h"
+#include "base/strings/strcat.h"
+#include "base/strings/utf_string_conversions.h"
 #include "base/types/expected.h"
 #include "brave/brave_domains/service_domains.h"
 #include "brave/components/email_aliases/email_aliases.mojom.h"
 #include "brave/components/email_aliases/email_aliases_api.h"
+#include "brave/components/email_aliases/email_aliases_api_key.h"
 #include "brave/components/email_aliases/features.h"
 #include "components/grit/brave_components_strings.h"
+#include "mojo/public/cpp/bindings/clone_traits.h"
 #include "mojo/public/cpp/bindings/pending_remote.h"
 #include "net/traffic_annotation/network_traffic_annotation.h"
 #include "services/network/public/cpp/resource_request.h"
@@ -36,12 +40,17 @@ constexpr char kAccountServiceEndpoint[] = "https://%s/v2/%s";
 constexpr char kAccountsServiceVerifyInitPath[] = "verify/init";
 constexpr char kAccountsServiceVerifyResultPath[] = "verify/result";
 
+constexpr char kEmailAliasesServiceURL[] =
+    "https://aliases.bravesoftware.com/manage";
+
+constexpr char kEmailAliasesServiceAPIKey[] = BUILDFLAG(EMAIL_ALIASES_API_KEY);
+
 // Minimum interval between verify/result polls
 constexpr base::TimeDelta kSessionPollInterval = base::Seconds(2);
 // Maximum total polling duration for a single verification flow.
 constexpr base::TimeDelta kMaxSessionPollDuration = base::Minutes(30);
 
-const net::NetworkTrafficAnnotationTag traffic_annotation =
+const net::NetworkTrafficAnnotationTag kTrafficAnnotation =
     net::DefineNetworkTrafficAnnotation("brave_accounts_service", R"(
       semantics {
         sender: "Email Aliases service"
@@ -56,6 +65,39 @@ const net::NetworkTrafficAnnotationTag traffic_annotation =
     })");
 
 constexpr int kMaxResponseLength = 32768;
+
+// Parses a JSON response body into type T (which must expose a `message` field)
+// from arbitrary JSON. On success, returns base::ok(T)
+// when T parses successfully and T.message == |expected_message|. On failure,
+// returns base::unexpected with a user-facing error string, preferring a
+// backend-reported error (via ErrorMessage) when available, or a generic
+// invalid-response error when the shape is unexpected.
+template <typename T>
+base::expected<T, std::string> ParseResponseDictAs(
+    const std::optional<std::string>& response_body,
+    const std::string_view expected_message) {
+  if (!response_body) {
+    return base::unexpected(l10n_util::GetStringUTF8(
+        IDS_EMAIL_ALIASES_SERVICE_ERROR_NO_RESPONSE_BODY));
+  }
+  auto value_opt = base::JSONReader::Read(*response_body);
+  if (!value_opt) {
+    return base::unexpected(l10n_util::GetStringUTF8(
+        IDS_EMAIL_ALIASES_SERVICE_ERROR_INVALID_RESPONSE_BODY));
+  }
+  auto parsed = T::FromValue(*value_opt);
+  if (parsed.has_value() && parsed.value().message == expected_message) {
+    return base::ok(std::move(*parsed));
+  }
+  auto error_message = ErrorMessage::FromValue(*value_opt);
+  if (error_message.has_value()) {
+    return base::unexpected(
+        l10n_util::GetStringFUTF8(IDS_EMAIL_ALIASES_SERVICE_REPORTED_ERROR,
+                                  base::UTF8ToUTF16(error_message->message)));
+  }
+  return base::unexpected(l10n_util::GetStringUTF8(
+      IDS_EMAIL_ALIASES_SERVICE_ERROR_INVALID_RESPONSE_BODY));
+}
 
 }  // namespace
 
@@ -72,11 +114,23 @@ GURL EmailAliasesService::GetAccountsServiceVerifyResultURL() {
                               kAccountsServiceVerifyResultPath));
 }
 
+// static
+GURL EmailAliasesService::GetEmailAliasesServiceURLForTesting() {
+  return GURL(kEmailAliasesServiceURL);
+}
+
+// static
+std::string EmailAliasesService::GetEmailAliasesServiceAPIKeyForTesting() {
+  return kEmailAliasesServiceAPIKey;
+}
+
 EmailAliasesService::EmailAliasesService(
     scoped_refptr<network::SharedURLLoaderFactory> url_loader_factory)
     : url_loader_factory_(url_loader_factory),
       verify_init_url_(GetAccountsServiceVerifyInitURL()),
-      verify_result_url_(GetAccountsServiceVerifyResultURL()) {
+      verify_result_url_(GetAccountsServiceVerifyResultURL()),
+      email_aliases_service_base_url_(GURL(kEmailAliasesServiceURL)),
+      email_aliases_api_key_(kEmailAliasesServiceAPIKey) {
   CHECK(base::FeatureList::IsEnabled(email_aliases::kEmailAliases));
 }
 
@@ -128,7 +182,7 @@ void EmailAliasesService::RequestAuthentication(
   resource_request->url = verify_init_url_;
   resource_request->method = net::HttpRequestHeaders::kPostMethod;
   verification_simple_url_loader_ = network::SimpleURLLoader::Create(
-      std::move(resource_request), traffic_annotation);
+      std::move(resource_request), kTrafficAnnotation);
   verification_simple_url_loader_->SetRetryOptions(
       /* max_retries=*/3,
       network::SimpleURLLoader::RETRY_ON_5XX |
@@ -195,7 +249,7 @@ void EmailAliasesService::RequestSession() {
   resource_request->headers.SetHeader(
       "Authorization", std::string("Bearer ") + verification_token_);
   verification_simple_url_loader_ = network::SimpleURLLoader::Create(
-      std::move(resource_request), traffic_annotation);
+      std::move(resource_request), kTrafficAnnotation);
   verification_simple_url_loader_->AttachStringForUpload(*body,
                                                          "application/json");
   verification_simple_url_loader_->DownloadToString(
@@ -250,6 +304,8 @@ void EmailAliasesService::OnRequestSessionResponse(
   auth_token_ = *parsed_session->auth_token;
   session_poll_elapsed_timer_.reset();
   NotifyObserversAuthStateChanged(mojom::AuthenticationStatus::kAuthenticated);
+  // Kick off an initial aliases refresh on successful authentication.
+  RefreshAliases();
 }
 
 void EmailAliasesService::MaybeRequestSessionAgain() {
@@ -284,23 +340,43 @@ void EmailAliasesService::CancelAuthenticationOrLogout(
 }
 
 void EmailAliasesService::GenerateAlias(GenerateAliasCallback callback) {
-  NOTIMPLEMENTED();
-  std::move(callback).Run(base::unexpected(std::string("Not implemented")));
+  base::Value::Dict body_value;  // empty JSON object required by the API
+  ApiFetch(email_aliases_service_base_url_,
+           net::HttpRequestHeaders::kPostMethod, body_value,
+           base::BindOnce(&EmailAliasesService::OnGenerateAliasResponse,
+                          weak_factory_.GetWeakPtr(), std::move(callback)));
 }
 
-void EmailAliasesService::UpdateAlias(const std::string& alias_email,
-                                      const std::optional<std::string>& note,
-                                      UpdateAliasCallback callback) {
-  // TODO: Implement alias update logic
-  NOTIMPLEMENTED();
-  std::move(callback).Run(base::unexpected(std::string("Not implemented")));
+void EmailAliasesService::UpdateAlias(
+    const std::string& alias_email,
+    const std::optional<std::string>& /* note */,
+    UpdateAliasCallback callback) {
+  // Build JSON using IDL-defined shape
+  UpdateAliasRequest request;
+  request.alias = alias_email;
+  // TODO(https://github.com/brave/brave-browser/issues/49229):
+  // Add support for storing alias note in the client.
+
+  // For now, we only support active aliases.
+  request.status = "active";
+  auto body_value = request.ToValue();
+  ApiFetch(email_aliases_service_base_url_, net::HttpRequestHeaders::kPutMethod,
+           body_value,
+           base::BindOnce(&EmailAliasesService::OnEditAliasResponse,
+                          weak_factory_.GetWeakPtr(), std::move(callback),
+                          /*update_expected=*/true));
 }
 
 void EmailAliasesService::DeleteAlias(const std::string& alias_email,
                                       DeleteAliasCallback callback) {
-  // TODO: Implement alias deletion logic
-  NOTIMPLEMENTED();
-  std::move(callback).Run(base::unexpected(std::string("Not implemented")));
+  DeleteAliasRequest request;
+  request.alias = alias_email;
+  auto body_value = request.ToValue();
+  ApiFetch(email_aliases_service_base_url_,
+           net::HttpRequestHeaders::kDeleteMethod, body_value,
+           base::BindOnce(&EmailAliasesService::OnEditAliasResponse,
+                          weak_factory_.GetWeakPtr(), std::move(callback),
+                          /*update_expected=*/false));
 }
 
 void EmailAliasesService::AddObserver(
@@ -316,6 +392,134 @@ void EmailAliasesService::AddObserver(
 
 std::string EmailAliasesService::GetAuthTokenForTesting() const {
   return auth_token_;
+}
+
+void EmailAliasesService::ApiFetch(const GURL& url,
+                                   const std::string_view method,
+                                   BodyAsStringCallback callback) {
+  CHECK(method == net::HttpRequestHeaders::kGetMethod ||
+        method == net::HttpRequestHeaders::kHeadMethod);
+  ApiFetchInternal(url, method, /*serialized_body=*/std::nullopt,
+                   std::move(callback));
+}
+
+void EmailAliasesService::ApiFetch(const GURL& url,
+                                   const std::string_view method,
+                                   const base::Value::Dict& body_value,
+                                   BodyAsStringCallback callback) {
+  CHECK(method == net::HttpRequestHeaders::kPostMethod ||
+        method == net::HttpRequestHeaders::kPutMethod ||
+        method == net::HttpRequestHeaders::kDeleteMethod);
+  auto body = base::WriteJson(body_value);
+  CHECK(body);
+  ApiFetchInternal(url, method, /*serialized_body=*/body, std::move(callback));
+}
+
+void EmailAliasesService::ApiFetchInternal(
+    const GURL& url,
+    const std::string_view method,
+    std::optional<std::string> serialized_body,
+    BodyAsStringCallback callback) {
+  auto resource_request = std::make_unique<network::ResourceRequest>();
+  resource_request->url = url;
+  resource_request->method = method;
+  resource_request->headers.SetHeader("Authorization",
+                                      std::string("Bearer ") + auth_token_);
+  resource_request->headers.SetHeader("X-API-key", email_aliases_api_key_);
+  auto simple_url_loader = network::SimpleURLLoader::Create(
+      std::move(resource_request), kTrafficAnnotation);
+  simple_url_loader->SetAllowHttpErrorResults(true);
+  if (serialized_body.has_value()) {
+    simple_url_loader->AttachStringForUpload(*serialized_body, "text/plain");
+  }
+  auto* loader_ptr = simple_url_loader.get();
+  loader_ptr->DownloadToString(
+      url_loader_factory_.get(),
+      base::BindOnce(
+          [](BodyAsStringCallback inner_callback,
+             std::unique_ptr<network::SimpleURLLoader> /*keep_alive_loader*/,
+             std::optional<std::string> response_body) {
+            std::move(inner_callback).Run(std::move(response_body));
+          },
+          std::move(callback), std::move(simple_url_loader)),
+      kMaxResponseLength);
+}
+
+void EmailAliasesService::OnGenerateAliasResponse(
+    GenerateAliasCallback user_callback,
+    std::optional<std::string> response_body) {
+  auto parsed =
+      ParseResponseDictAs<GenerateAliasResponse>(response_body, "created");
+  auto result =
+      parsed.has_value()
+          ? base::expected<std::string, std::string>(parsed.value().alias)
+          : base::unexpected(parsed.error());
+  std::move(user_callback).Run(std::move(result));
+}
+
+void EmailAliasesService::OnEditAliasResponse(
+    base::OnceCallback<void(base::expected<std::monostate, std::string>)>
+        user_callback,
+    bool update_expected,
+    std::optional<std::string> response_body) {
+  RefreshAliases();
+  auto parsed = ParseResponseDictAs<AliasEditedResponse>(
+      response_body, update_expected ? "updated" : "deleted");
+  auto result =
+      parsed.has_value()
+          ? base::expected<std::monostate, std::string>(std::monostate{})
+          : base::unexpected(parsed.error());
+  std::move(user_callback).Run(std::move(result));
+}
+
+void EmailAliasesService::RefreshAliases() {
+  ApiFetch(email_aliases_service_base_url_.Resolve("?status=active"),
+           net::HttpRequestHeaders::kGetMethod,
+           base::BindOnce(&EmailAliasesService::OnRefreshAliasesResponse,
+                          weak_factory_.GetWeakPtr()));
+}
+
+void EmailAliasesService::OnRefreshAliasesResponse(
+    std::optional<std::string> response_body) {
+  // TODO(https://github.com/brave/brave-browser/issues/48959):
+  // In this function, when an error happens, we should show
+  // an error message to the user (requires design work)
+  if (!response_body) {
+    LOG(ERROR) << "Email Aliases service error: No response body";
+    return;
+  }
+  auto parsed = base::JSONReader::Read(*response_body);
+  // TODO(https://github.com/brave/brave-browser/issues/49624):
+  // Remove this check once the backend is updated to return a dictionary.
+  if (parsed && parsed->is_list()) {
+    // Wrap the list in a dictionary to match the AliasListResponse shape.
+    base::Value::Dict dict;
+    dict.Set("result", std::move(*parsed));
+    parsed = base::Value(std::move(dict));
+  }
+  if (!parsed || !parsed->is_dict()) {
+    LOG(ERROR) << "Email Aliases service error: Invalid response format";
+    return;
+  }
+  const base::Value::Dict& parsed_dict = parsed->GetDict();
+  if (const std::string* error_message = parsed_dict.FindString("message")) {
+    LOG(ERROR) << "Email Aliases service error: " << *error_message;
+    return;
+  }
+  auto list_response = AliasListResponse::FromValue(parsed_dict);
+  if (!list_response) {
+    LOG(ERROR) << "Email Aliases service error: Invalid response format";
+    return;
+  }
+  std::vector<email_aliases::mojom::AliasPtr> aliases;
+  for (const auto& entry : list_response->result) {
+    auto alias_obj = email_aliases::mojom::Alias::New();
+    alias_obj->email = entry.alias;
+    aliases.push_back(std::move(alias_obj));
+  }
+  for (auto& observer : observers_) {
+    observer->OnAliasesUpdated(mojo::Clone(aliases));
+  }
 }
 
 }  // namespace email_aliases
--- a/components/email_aliases/email_aliases_service.h
+++ b/components/email_aliases/email_aliases_service.h
@@ -7,10 +7,15 @@
 #define BRAVE_COMPONENTS_EMAIL_ALIASES_EMAIL_ALIASES_SERVICE_H_
 
 #include <memory>
+#include <optional>
 #include <string>
+#include <string_view>
 
+#include "base/functional/callback.h"
+#include "base/memory/weak_ptr.h"
 #include "base/timer/elapsed_timer.h"
 #include "base/timer/timer.h"
+#include "base/values.h"
 #include "brave/components/email_aliases/email_aliases.mojom.h"
 #include "components/keyed_service/core/keyed_service.h"
 #include "mojo/public/cpp/bindings/pending_receiver.h"
@@ -57,15 +62,15 @@ class EmailAliasesService : public KeyedService,
   void CancelAuthenticationOrLogout(
       CancelAuthenticationOrLogoutCallback callback) override;
 
-  // Not implemented yet.
+  // Requests generation of a new alias and returns the result via |callback|.
   void GenerateAlias(GenerateAliasCallback callback) override;
 
-  // Not implemented yet.
+  // Creates or updates an alias identified by |alias_email| with optional note.
   void UpdateAlias(const std::string& alias_email,
                    const std::optional<std::string>& note,
                    UpdateAliasCallback callback) override;
 
-  // Not implemented yet.
+  // Deletes the alias identified by |alias_email|.
   void DeleteAlias(const std::string& alias_email,
                    DeleteAliasCallback callback) override;
 
@@ -86,7 +91,17 @@ class EmailAliasesService : public KeyedService,
   static GURL GetAccountsServiceVerifyInitURL();
   static GURL GetAccountsServiceVerifyResultURL();
 
+  // Returns the base URL for the Email Aliases service.
+  static GURL GetEmailAliasesServiceURLForTesting();
+
+  // Returns the API key for the Email Aliases service.
+  static std::string GetEmailAliasesServiceAPIKeyForTesting();
+
  private:
+  // Callback that receives the response body as an optional string.
+  using BodyAsStringCallback =
+      base::OnceCallback<void(std::optional<std::string> response_body)>;
+
   // Handles the response to the verify/init request. Parses a verification
   // token and, if present, proceeds to poll the session endpoint. Invokes
   // |callback| with an optional error message.
@@ -117,6 +132,41 @@ class EmailAliasesService : public KeyedService,
   // tokens to reset the authentication flow to a clean state.
   void ResetVerificationFlow();
 
+  // Fetch helper for Email Aliases backend. Specifically for GET/HEAD.
+  void ApiFetch(const GURL& url,
+                const std::string_view method,
+                BodyAsStringCallback download_to_string_callback);
+
+  // Fetch helper which uploads |body_value|. Specifically for POST/PUT/DELETE.
+  void ApiFetch(const GURL& url,
+                const std::string_view method,
+                const base::Value::Dict& body_value,
+                BodyAsStringCallback download_to_string_callback);
+
+  // Shared implementation used by the two ApiFetch overloads to make a network
+  // request.
+  void ApiFetchInternal(const GURL& url,
+                        const std::string_view method,
+                        std::optional<std::string> serialized_body,
+                        BodyAsStringCallback download_to_string_callback);
+
+  // Refreshes the aliases list from the server and notifies observers.
+  void RefreshAliases();
+
+  // Parses and applies the aliases list received from the backend.
+  void OnRefreshAliasesResponse(std::optional<std::string> response_body);
+
+  // Processes the server response for a generate-alias request.
+  void OnGenerateAliasResponse(GenerateAliasCallback user_callback,
+                               std::optional<std::string> response_body);
+
+  // Common handler for alias edit responses (update/delete).
+  void OnEditAliasResponse(
+      base::OnceCallback<void(base::expected<std::monostate, std::string>)>
+          user_callback,
+      bool update_expected,
+      std::optional<std::string> response_body);
+
   // Bound Mojo receivers for the EmailAliasesService interface.
   mojo::ReceiverSet<mojom::EmailAliasesService> receivers_;
 
@@ -146,6 +196,12 @@ class EmailAliasesService : public KeyedService,
   // Cached fully-qualified verify/result URL.
   const GURL verify_result_url_;
 
+  // Cached fully-qualified email aliases service base URL.
+  const GURL email_aliases_service_base_url_;
+
+  // Cached email aliases API key.
+  const std::string email_aliases_api_key_;
+
   // One-shot timer used to delay subsequent verify/result polls so that they
   // are not issued more frequently than the minimum interval.
   base::OneShotTimer session_request_timer_;
@@ -153,6 +209,9 @@ class EmailAliasesService : public KeyedService,
   // Elapsed timer for the current verification polling window. Used to
   // enforce a maximum total polling duration.
   std::optional<base::ElapsedTimer> session_poll_elapsed_timer_;
+
+  // WeakPtrFactory to safely bind callbacks across async network operations.
+  base::WeakPtrFactory<EmailAliasesService> weak_factory_{this};
 };
 
 }  // namespace email_aliases
--- a/components/email_aliases/email_aliases_service_unittest.cc
+++ b/components/email_aliases/email_aliases_service_unittest.cc
@@ -7,19 +7,28 @@
 
 #include <string>
 #include <utility>
+#include <variant>
 #include <vector>
 
+#include "base/containers/contains.h"
 #include "base/functional/bind.h"
+#include "base/json/json_reader.h"
 #include "base/test/bind.h"
+#include "base/test/mock_callback.h"
 #include "base/test/run_until.h"
 #include "base/test/scoped_feature_list.h"
 #include "base/test/task_environment.h"
+#include "base/test/test_future.h"
 #include "base/types/expected.h"
 #include "brave/components/email_aliases/features.h"
 #include "components/grit/brave_components_strings.h"
+#include "net/base/net_errors.h"
+#include "net/http/http_status_code.h"
 #include "services/network/public/cpp/weak_wrapper_shared_url_loader_factory.h"
 #include "services/network/test/test_url_loader_factory.h"
+#include "testing/gmock/include/gmock/gmock.h"
 #include "testing/gtest/include/gtest/gtest.h"
+#include "third_party/abseil-cpp/absl/functional/overload.h"
 #include "ui/base/l10n/l10n_util.h"
 
 namespace email_aliases {
@@ -57,10 +66,8 @@ class EmailAliasesServiceTest : public ::testing::Test {
   }
 
   void SetUp() override {
-    url_loader_factory_ =
-        base::MakeRefCounted<network::WeakWrapperSharedURLLoaderFactory>(
-            &test_url_loader_factory_);
-    service_ = std::make_unique<EmailAliasesService>(url_loader_factory_);
+    service_ = std::make_unique<EmailAliasesService>(
+        test_url_loader_factory_.GetSafeWeakWrapper());
     observer_ = std::make_unique<TestObserver>();
     mojo::PendingRemote<email_aliases::mojom::EmailAliasesServiceObserver>
         remote;
@@ -139,7 +146,6 @@ class EmailAliasesServiceTest : public ::testing::Test {
 
   base::test::ScopedFeatureList feature_list_;
   network::TestURLLoaderFactory test_url_loader_factory_;
-  scoped_refptr<network::SharedURLLoaderFactory> url_loader_factory_;
   std::unique_ptr<EmailAliasesService> service_;
   base::test::TaskEnvironment task_environment_;
   std::unique_ptr<TestObserver> observer_;
@@ -234,15 +240,10 @@ TEST_F(EmailAliasesServiceTest,
 // rate limiting and total polling duration.
 class EmailAliasesServiceTimingTest : public ::testing::Test {
  protected:
-  EmailAliasesServiceTimingTest()
-      : url_loader_factory_(/*no args*/),
-        url_loader_wrapper_(
-            base::MakeRefCounted<network::WeakWrapperSharedURLLoaderFactory>(
-                &url_loader_factory_)) {}
-
   void SetUp() override {
     feature_list_.InitAndEnableFeature(email_aliases::kEmailAliases);
-    service_ = std::make_unique<EmailAliasesService>(url_loader_wrapper_);
+    service_ = std::make_unique<EmailAliasesService>(
+        url_loader_factory_.GetSafeWeakWrapper());
     observer_ = std::make_unique<TestObserver>();
     mojo::PendingRemote<email_aliases::mojom::EmailAliasesServiceObserver>
         remote;
@@ -294,7 +295,6 @@ class EmailAliasesServiceTimingTest : public ::testing::Test {
       base::test::TaskEnvironment::TimeSource::MOCK_TIME};
   base::test::ScopedFeatureList feature_list_;
   network::TestURLLoaderFactory url_loader_factory_;
-  scoped_refptr<network::SharedURLLoaderFactory> url_loader_wrapper_;
   std::unique_ptr<EmailAliasesService> service_;
   std::unique_ptr<TestObserver> observer_;
   std::vector<base::TimeTicks> verify_result_request_times_;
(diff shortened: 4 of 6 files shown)


arthuredelstein:
image
image
image
image

components/email_aliases/BUILD.gn
```diff
@@ -3,6 +3,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
```
cdesouza-chromium, line 3: Please add a description to this PR explaining what is being done, descriptive to your fellow engineer, and make sure that that description gets added to the squash merge commit message. The following link has some information on what constitutes a good PR/Commit description:

https://chromium.googlesource.com/chromium/src/+/main/docs/cl_respect.md#provide-context

components/email_aliases/email_aliases_service.cc
```diff
@@ -56,6 +64,32 @@ const net::NetworkTrafficAnnotationTag traffic_annotation =
...
+  if (!dict_opt) {
+    return base::unexpected(l10n_util::GetStringUTF8(
+        IDS_EMAIL_ALIASES_SERVICE_ERROR_INVALID_RESPONSE_BODY));
+  }
+  if (auto err = ErrorResponse::FromValue(*dict_opt)) {
+    LOG(ERROR) << "Email Aliases service error: " << err->error;
+    std::string msg = base::UTF16ToUTF8(
+        l10n_util::GetStringFUTF16(IDS_EMAIL_ALIASES_SERVICE_REPORTED_ERROR,
+                                   base::UTF8ToUTF16(err->error)));
+    return base::unexpected(msg);
```
cdesouza-chromium: `std::move(msg)`

components/email_aliases/email_aliases_service.cc
```diff
@@ -71,11 +105,24 @@ GURL EmailAliasesService::GetAccountsServiceVerifyResultURL() {
                               kAccountsServiceVerifyResultPath));
 }
 
+// static
+GURL EmailAliasesService::GetEmailAliasesServiceBaseURL() {
+  return GURL(absl::StrFormat("%s%s", kEmailAliasesServiceBaseURL,
```
cdesouza-chromium: Nit: use `base::StrCat` as this is just a concatanation.

components/email_aliases/email_aliases_service.cc
```diff
@@ -314,4 +379,147 @@ std::string EmailAliasesService::GetAuthTokenForTesting() const {
...
+  resource_request->headers.SetHeader("X-API-key", email_aliases_api_key_);
+  auto simple_url_loader = network::SimpleURLLoader::Create(
+      std::move(resource_request), kTrafficAnnotation);
+  // For non-GET/HEAD methods attach a JSON body. Backend expects text/plain.
+  if (method != net::HttpRequestHeaders::kGetMethod &&
+      method != net::HttpRequestHeaders::kHeadMethod) {
+    auto body = base::WriteJson(body_value);
+    CHECK(body);
+    simple_url_loader->AttachStringForUpload(body.value(), "text/plain");
+  }
```
cdesouza-chromium: I think it may be better if you have an `ApiFetch` overload that handles takes `body_value` and CHECKs the verb to not be these two, and then we have another overload for no body value, but we shouldn't force the caller to provide `body_value`, for things GET and HEAD, just to silently ignore it.

components/email_aliases/email_aliases_service.cc
```diff
@@ -314,4 +379,147 @@ std::string EmailAliasesService::GetAuthTokenForTesting() const {
...
+                     weak_factory_.GetWeakPtr(),
+                     std::move(download_to_string_callback),
+                     std::move(simple_url_loader)),
+      kMaxResponseLength);
+}
+
+void EmailAliasesService::OnApiFetchDownloadToStringComplete(
+    BodyAsStringCallback callback,
+    std::unique_ptr<network::SimpleURLLoader> /*owned_loader*/,
+    std::unique_ptr<std::string> response_body) {
```
cdesouza-chromium: You are using the old overload. Change this to `std::optional<std::string> response_body`, and things become simplified.

components/email_aliases/email_aliases_service.cc
```diff
@@ -314,4 +379,147 @@ std::string EmailAliasesService::GetAuthTokenForTesting() const {
...
+  loader_ptr->DownloadToString(
+      url_loader_factory_.get(),
+      base::BindOnce(&EmailAliasesService::OnApiFetchDownloadToStringComplete,
+                     weak_factory_.GetWeakPtr(),
+                     std::move(download_to_string_callback),
+                     std::move(simple_url_loader)),
+      kMaxResponseLength);
+}
+
+void EmailAliasesService::OnApiFetchDownloadToStringComplete(
```
cdesouza-chromium: Make the name of this function better reflect that it exists to encapsulate the ownership of the `simple_url_loader` into the closure as self-owned.

components/email_aliases/email_aliases_service.cc
arthuredelstein: fixed

components/email_aliases/email_aliases_service.cc
arthuredelstein: fixed

components/email_aliases/email_aliases_service.cc
arthuredelstein: done

components/email_aliases/email_aliases_service.cc
arthuredelstein: Done. That made it simple enough that I converted it to a lambda.

components/email_aliases/email_aliases_service.cc
arthuredelstein: I replaced this function with a lambda in PerformFetch.

components/email_aliases/email_aliases_service_unittest.cc
```diff
@@ -336,3 +337,355 @@ TEST_F(EmailAliasesServiceTimingTest, VerifyResult_StopsAfterMaxDuration) {
...
+ protected:
+  EmailAliasesAPITest() {
+    feature_list_.InitAndEnableFeature(email_aliases::kEmailAliases);
+  }
+
+  void SetUp() override {
+    url_loader_wrapper_ =
+        base::MakeRefCounted<network::WeakWrapperSharedURLLoaderFactory>(
+            &url_loader_factory_);
+    service_ = std::make_unique<EmailAliasesService>(url_loader_wrapper_);
```
szilardszaloki:
```
service_ = std::make_unique<EmailAliasesService>(url_loader_factory_.GetSafeWeakWrapper());
```
and then you don't need `url_loader_wrapper_`.

components/email_aliases/email_aliases_service_unittest.cc
```diff
@@ -336,3 +337,355 @@ TEST_F(EmailAliasesServiceTimingTest, VerifyResult_StopsAfterMaxDuration) {
...
+                           *out = std::move(res);
+                         },
+                         &called, &result_out));
+    EXPECT_TRUE(base::test::RunUntil([&]() { return called; }));
+    EXPECT_TRUE(observer_.WaitForAliasUpdateCount(1));
+    return result_out;
+  }
+};
+
+TEST_F(EmailAliasesAPITest, GenerateAlias_ReturnsAlias_OnSuccess) {
```
szilardszaloki: Looks like a good fit for parameterized tests (`testing::TestWithParam<>`) so the body doesn't have to be repeated.

arthuredelstein:
Apologies to anyone who received a review-request spam ([here](https://github.com/brave/brave-core/pull/31027#event-19664933529)). This happened when I changed the target branch to master.

components/email_aliases/email_aliases_service_unittest.cc
arthuredelstein: Done. Thanks

components/email_aliases/email_aliases_service_unittest.cc
arthuredelstein: done

components/email_aliases/BUILD.gn
arthuredelstein: done

components/email_aliases/email_aliases_service_unittest.cc
```diff
@@ -345,3 +339,438 @@ TEST_F(EmailAliasesServiceTimingTest, VerifyResult_StopsAfterMaxDuration) {
...
+    UpdateAliasCaseName);
+// ================= DeleteAlias (parameterized) =================
+
+struct DeleteAliasCase {
+  const char* name;
+  std::optional<std::string> delete_body;
+  std::optional<int> net_error;
+  bool expect_success;
+  std::optional<int> expected_error_id;
+  std::optional<std::string> expected_error_substr;
```
szilardszaloki: @arthuredelstein I know you added me by accident and I'm not needed on this PR, but just a heads-up: you can compare `base::expected<>`s directly, so these aren't necessary. That would eliminate the branching in your tests altogether.

Also, since the test cases are almost identical, it'd be simpler to use a single `TestCase` type, for example:
```cpp
template <typename T, typename E>
struct TestCase {
  std::string name;
  int net_error;
  std::string body;
  base::expected<T, E> expected;
};
```

Also, just a note: you can easily verify that a callback is invoked using `base::MockCallback<>`, e.g.:
```cpp
base::MockCallback<base::OnceCallback<void(...)>> callback;
EXPECT_CALL(callback, Run)
    .Times(1)
    .WillOnce([](...) {
      // ...
    });
```
And then pass `callback.Get()` wherever it's needed.

arthuredelstein: Thanks @szilardszaloki. I simplified the tests cases as suggested. Now looking into MockCallback

arthuredelstein: I pushed a commit using MockCallback.

arthuredelstein: (issues in this thread are fixed)

components/email_aliases/email_aliases_service_unittest.cc
```diff
@@ -345,3 +342,394 @@ TEST_F(EmailAliasesServiceTimingTest, VerifyResult_StopsAfterMaxDuration) {
...
+  base::expected<std::string, std::string> CallGenerateAliasWith(
+      const std::optional<std::string>& body,
+      const std::optional<int>& net_error = std::nullopt) {
+    const GURL manage_url =
+        EmailAliasesService::GetEmailAliasesServiceBaseURL();
+    if (net_error.has_value()) {
+      network::URLLoaderCompletionStatus completion(*net_error);
+      url_loader_factory_.AddResponse(manage_url, /*head=*/nullptr,
+                                      /*content=*/"", completion);
+    } else {
```
szilardszaloki: My understanding is that this branch aims to produce a `std::nullopt` `response_body` in `ParseResponseDictOrError()` (in `email_aliases_service.cc`). You can handle this with the same `AddResponse()` overload, e.g.:
```cpp
url_loader_factory_.AddResponse(manage_url.spec(), "", net::HTTP_INTERNAL_SERVER_ERROR);
```
Also, in the test case, `net_error` is only set to `net::ERR_FAILED` when `body` is `std::nullopt`, so `body` already captures the intent and both aren't needed:
```cpp
url_loader_factory_.AddResponse(
    EmailAliasesService::GetEmailAliasesServiceBaseURL().spec(),
    body.value_or(""),
    body ? net::HTTP_OK : net::HTTP_INTERNAL_SERVER_ERROR);
```

components/email_aliases/email_aliases_service_unittest.cc
```diff
@@ -345,3 +342,394 @@ TEST_F(EmailAliasesServiceTimingTest, VerifyResult_StopsAfterMaxDuration) {
...
+    base::expected<std::string, std::string> result_out;
+    EXPECT_CALL(cb, Run(testing::_))
+        .Times(1)
+        .WillOnce([&](base::expected<std::string, std::string> res) {
+          called = true;
+          result_out = std::move(res);
+        });
+    service_->GenerateAlias(cb.Get());
+    EXPECT_TRUE(base::test::RunUntil([&]() { return called; }));
+    return result_out;
```
szilardszaloki: It seems this could be simplified down to the following three lines:
```cpp
base::test::TestFuture<base::expected<std::string, std::string>> future;
service_->GenerateAlias(future.GetCallback());
return future.Take();
```
which would make your `CallGenerateAliasWith()` look like this:
```cpp
base::expected<std::string, std::string> CallGenerateAliasWith(
    const std::optional<std::string>& body) {
  url_loader_factory_.AddResponse(
      EmailAliasesService::GetEmailAliasesServiceBaseURL().spec(),
      body.value_or(""),
      body ? net::HTTP_OK : net::HTTP_INTERNAL_SERVER_ERROR);
  base::test::TestFuture<base::expected<std::string, std::string>> future;
  service_->GenerateAlias(future.GetCallback());
  return future.Take();
}
```

components/email_aliases/email_aliases_service_unittest.cc
```diff
@@ -345,3 +342,394 @@ TEST_F(EmailAliasesServiceTimingTest, VerifyResult_StopsAfterMaxDuration) {
...
+          called = true;
+          result_out = std::move(res);
+        });
+    service_->UpdateAlias(alias_email, /*note=*/std::string("note"), cb.Get());
+    EXPECT_TRUE(base::test::RunUntil([&]() { return called; }));
+    if (wait_for_update) {
+      EXPECT_TRUE(observer_.WaitForAliasUpdateCount(1));
+    }
+    return result_out;
+  }
```
szilardszaloki, line 433: `CallUpdateAliasWith()` would become:
```cpp
base::expected<std::monostate, std::string> CallUpdateAliasWith(
    const std::string& alias_email,
    const std::optional<std::string>& put_body,
    const std::optional<std::string>& refresh_body = std::nullopt,
    bool wait_for_update = true) {
  const GURL manage_url =
      EmailAliasesService::GetEmailAliasesServiceBaseURL();
  url_loader_factory_.AddResponse(
      manage_url.spec(), put_body.value_or(""),
      put_body ? net::HTTP_OK : net::HTTP_INTERNAL_SERVER_ERROR);
  url_loader_factory_.AddResponse(manage_url.Resolve("?status=active").spec(),
                                  refresh_body.value_or("[]"));
  base::test::TestFuture<base::expected<std::monostate, std::string>> future;
  service_->UpdateAlias(alias_email, /*note=*/std::string("note"),
                        future.GetCallback());
  auto result_out = future.Take();
  if (wait_for_update) {
    EXPECT_TRUE(observer_.WaitForAliasUpdateCount(1));
  }
  return result_out;
}
```

components/email_aliases/email_aliases_service_unittest.cc
```diff
@@ -345,3 +342,394 @@ TEST_F(EmailAliasesServiceTimingTest, VerifyResult_StopsAfterMaxDuration) {
...
+        .WillOnce([&](base::expected<std::monostate, std::string> res) {
+          called = true;
+          result_out = std::move(res);
+        });
+    service_->DeleteAlias(alias_email, cb.Get());
+    EXPECT_TRUE(base::test::RunUntil([&]() { return called; }));
+    EXPECT_TRUE(observer_.WaitForAliasUpdateCount(1));
+    return result_out;
+  }
+};
```
szilardszaloki: `CallDeleteAliasWith()` would become:
```cpp
base::expected<std::monostate, std::string> CallDeleteAliasWith(
    const std::string& alias_email,
    const std::optional<std::string>& delete_body) {
  const GURL manage_url =
      EmailAliasesService::GetEmailAliasesServiceBaseURL();
  url_loader_factory_.AddResponse(
      manage_url.spec(), delete_body.value_or(""),
      delete_body ? net::HTTP_OK : net::HTTP_INTERNAL_SERVER_ERROR);
  url_loader_factory_.AddResponse(manage_url.Resolve("?status=active").spec(),
                                  "[]");
  base::test::TestFuture<base::expected<std::monostate, std::string>> future;
  service_->DeleteAlias(alias_email, future.GetCallback());
  auto result_out = future.Take();
  EXPECT_TRUE(observer_.WaitForAliasUpdateCount(1));
  return result_out;
}
```

components/email_aliases/email_aliases_service_unittest.cc
```diff
@@ -345,3 +342,394 @@ TEST_F(EmailAliasesServiceTimingTest, VerifyResult_StopsAfterMaxDuration) {
...
+// Build the success response body for Generate tests using the expected alias.
+std::string MakeGenerateAliasSuccessBody() {
+  return std::string("{\"message\":\"created\",\"alias\":\"") +
+         kExpectedGeneratedAlias + "\"}";
+}
+
+struct EmailAliasesApiCase {
+  const char* name;
+  std::optional<std::string> body;
+  std::optional<int> net_error;
```
szilardszaloki: As per my other comment above, you wouldn't need `net_error`.

components/email_aliases/email_aliases_service_unittest.cc
```diff
@@ -345,3 +342,394 @@ TEST_F(EmailAliasesServiceTimingTest, VerifyResult_StopsAfterMaxDuration) {
...
+
+// Shared expectation helper that only validates success/failure state and
+// error substring when applicable. Value checks (on success) are performed
+// by callers when needed.
+template <typename T>
+void ExpectResultStateMatches(const base::expected<T, std::string>& result,
+                              const EmailAliasesApiCase& p) {
+  if (p.expected_error_substr) {
+    ASSERT_FALSE(result.has_value());
+    EXPECT_NE(result.error().find(*p.expected_error_substr), std::string::npos);
```
szilardszaloki: That's basically:
```cpp
EXPECT_TRUE(base::Contains(result.error(), *p.expected_error_substr));
```

components/email_aliases/email_aliases_service_unittest.cc
```diff
@@ -345,3 +342,394 @@ TEST_F(EmailAliasesServiceTimingTest, VerifyResult_StopsAfterMaxDuration) {
...
+                            /*net_error=*/net::ERR_FAILED,
+                            /*expected_error_substr=*/"No response"},
+        EmailAliasesApiCase{/*name=*/"InvalidJSON",
+                            /*body=*/"not a json",
+                            /*net_error=*/std::nullopt,
+                            /*expected_error_substr=*/"Invalid response"},
+        EmailAliasesApiCase{/*name=*/"UnexpectedPayload",
+                            /*body=*/"{\"message\":\"ok_but_no_alias\"}",
+                            /*net_error=*/std::nullopt,
+                            /*expected_error_substr=*/"Alias not available"}),
```
szilardszaloki: You could use designated initializers here — they make the intent clearer and avoid issues if the `struct`'s field order changes.
```cpp
EmailAliasesApiCase{.name = "Success",
                    .body = MakeGenerateAliasSuccessBody(),
                    .expected_error_substr = std::nullopt},
EmailAliasesApiCase{.name = "BackendError",
                    .body = "{\"message\":\"alias_unavailable\"}",
                    .expected_error_substr = "Alias not available"},
EmailAliasesApiCase{.name = "NoBody",
                    .body = std::nullopt,
                    .expected_error_substr = "No response"},
EmailAliasesApiCase{.name = "InvalidJSON",
                    .body = "not a json",
                    .expected_error_substr = "Invalid response"},
EmailAliasesApiCase{.name = "UnexpectedPayload",
                    .body = "{\"message\":\"ok_but_no_alias\"}",
                    .expected_error_substr = "Alias not available"}),
```

components/email_aliases/email_aliases_service_unittest.cc
```diff
@@ -345,3 +342,394 @@ TEST_F(EmailAliasesServiceTimingTest, VerifyResult_StopsAfterMaxDuration) {
...
+                            /*net_error=*/std::nullopt,
+                            /*expected_error_substr=*/"not_updated"},
+        EmailAliasesApiCase{/*name=*/"NoBody",
+                            /*body=*/std::nullopt,
+                            /*net_error=*/net::ERR_FAILED,
+                            /*expected_error_substr=*/"No response"},
+        EmailAliasesApiCase{/*name=*/"InvalidJSON",
+                            /*body=*/"not a json",
+                            /*net_error=*/std::nullopt,
+                            /*expected_error_substr=*/"Invalid response"}),
```
szilardszaloki:
```cpp
EmailAliasesApiCase{.name = "Success",
                    .body = "{\"message\":\"updated\"}",
                    .expected_error_substr = std::nullopt},
EmailAliasesApiCase{.name = "BackendError",
                    .body = "{\"message\":\"backend_error\"}",
                    .expected_error_substr = "backend_error"},
EmailAliasesApiCase{.name = "NonUpdatedMessage",
                    .body = "{\"message\":\"not_updated\"}",
                    .expected_error_substr = "not_updated"},
EmailAliasesApiCase{.name = "NoBody",
                    .body = std::nullopt,
                    .expected_error_substr = "No response"},
EmailAliasesApiCase{.name = "InvalidJSON",
                    .body = "not a json",
                    .expected_error_substr = "Invalid response"}),
```

components/email_aliases/email_aliases_service_unittest.cc
```diff
@@ -345,3 +342,394 @@ TEST_F(EmailAliasesServiceTimingTest, VerifyResult_StopsAfterMaxDuration) {
...
+                            /*net_error=*/std::nullopt,
+                            /*expected_error_substr=*/"backend_error"},
+        EmailAliasesApiCase{/*name=*/"NoBody",
+                            /*body=*/std::nullopt,
+                            /*net_error=*/net::ERR_FAILED,
+                            /*expected_error_substr=*/"No response"},
+        EmailAliasesApiCase{/*name=*/"InvalidJSON",
+                            /*body=*/"not a json",
+                            /*net_error=*/std::nullopt,
+                            /*expected_error_substr=*/"Invalid response"}),
```
szilardszaloki:
```cpp
EmailAliasesApiCase{.name = "Success",
                    .body = "{\"message\":\"deleted\"}",
                    .expected_error_substr = std::nullopt},
EmailAliasesApiCase{.name = "BackendError",
                    .body = "{\"message\":\"backend_error\"}",
                    .expected_error_substr = "backend_error"},
EmailAliasesApiCase{.name = "NoBody",
                    .body = std::nullopt,
                    .expected_error_substr = "No response"},
EmailAliasesApiCase{.name = "InvalidJSON",
                    .body = "not a json",
                    .expected_error_substr = "Invalid response"}),
```

components/email_aliases/email_aliases_service_unittest.cc
szilardszaloki: Hey @arthuredelstein 🙂,
Thanks — my intent was to introduce a generic test case type and pass the appropriate instantiation to each fixture, so we can compare `base::expected<>` values directly rather than unpacking them. For example:
```cpp
template <typename T>
struct EmailAliasesApiCase {
  const char* name;
  std::optional<std::string> body;
  base::expected<T, std::string> expected_result;
};

class GenerateAliasParamTest
    : public EmailAliasesAPITest,
      public testing::WithParamInterface<EmailAliasesApiCase<std::string>> {};

class UpdateAliasParamTest
    : public EmailAliasesAPITest,
      public testing::WithParamInterface<EmailAliasesApiCase<std::monostate>> {};

class DeleteAliasParamTest
    : public EmailAliasesAPITest,
      public testing::WithParamInterface<EmailAliasesApiCase<std::monostate>> {};
```
Then each test would provide an `expected_result` and we'd assert `EXPECT_EQ(result_out, p.expected_result)`. I realize substring checks wouldn't be covered by equality, but we could add a small matcher for the error case (or switch those cases to exact equality instead of substring containment).

components/email_aliases/email_aliases_service.cc
```diff
@@ -57,6 +65,32 @@ const net::NetworkTrafficAnnotationTag traffic_annotation =
...
+    return base::unexpected(l10n_util::GetStringUTF8(
+        IDS_EMAIL_ALIASES_SERVICE_ERROR_INVALID_RESPONSE_BODY));
+  }
+  if (auto err = ErrorResponse::FromValue(*dict_opt)) {
+    LOG(ERROR) << "Email Aliases service error: " << err->error;
+    std::string msg = base::UTF16ToUTF8(
+        l10n_util::GetStringFUTF16(IDS_EMAIL_ALIASES_SERVICE_REPORTED_ERROR,
+                                   base::UTF8ToUTF16(err->error)));
+    return base::unexpected(std::move(msg));
+  }
```
szilardszaloki: Does this branch ever get hit in practice? It looks like a call to `SetAllowHttpErrorResults(true)` on the `SimpleURLLoader` might be missing — without it, the loader only surfaces bodies for `2xx` responses, so this code path may never execute.

components/email_aliases/email_aliases_service_unittest.cc
arthuredelstein: See simplification commit from @szilardszaloki here: https://github.com/brave/brave-core/pull/31027/commits/ff23527c9376c84387d8ec7d5166d53715934a9f

components/email_aliases/email_aliases_service_unittest.cc
arthuredelstein: See simplification commit from @szilardszaloki here: https://github.com/brave/brave-core/commit/ff23527c9376c84387d8ec7d5166d53715934a9f

components/email_aliases/email_aliases_service_unittest.cc
arthuredelstein: See simplification commit from @szilardszaloki here: https://github.com/brave/brave-core/commit/ff23527c9376c84387d8ec7d5166d53715934a9f

components/email_aliases/email_aliases_service_unittest.cc
arthuredelstein: See simplification commit from @szilardszaloki here: https://github.com/brave/brave-core/commit/ff23527c9376c84387d8ec7d5166d53715934a9f

components/email_aliases/email_aliases_service_unittest.cc
arthuredelstein: See simplification commit from @szilardszaloki here: https://github.com/brave/brave-core/commit/ff23527c9376c84387d8ec7d5166d53715934a9f

components/email_aliases/email_aliases_service_unittest.cc
arthuredelstein: See simplification commit from @szilardszaloki here: https://github.com/brave/brave-core/commit/ff23527c9376c84387d8ec7d5166d53715934a9f

components/email_aliases/email_aliases_service_unittest.cc
arthuredelstein: See simplification commit from @szilardszaloki here: https://github.com/brave/brave-core/commit/ff23527c9376c84387d8ec7d5166d53715934a9f

components/email_aliases/email_aliases_service_unittest.cc
arthuredelstein: See simplification commit from @szilardszaloki here: https://github.com/brave/brave-core/commit/ff23527c9376c84387d8ec7d5166d53715934a9f

components/email_aliases/email_aliases_service.cc
arthuredelstein: Good catch, I have added a call to `SetAllowHttpErrorResults(true)` and updated the unit tests to ensure it is taking effect: 5dcb6b82c53

components/email_aliases/BUILD.gn
cdesouza-chromium: The particular section "Provide context" says:

> Change descriptions are the first impression your change makes, both on reviewers and on code archeologists from the future. A [good description](https://chromium.googlesource.com/chromium/src/+/main/docs/contributing.md#Uploading-a-change-for-review) aims to do two things. First, it conveys at a glance the high level view. Second, it provides references to all the relevant information for a deep dive: design docs, bugs, testing instructions. The bug# is a useful reference, but isn't sufficient on its own. Summarize **what** and **why** in the description. You can additionally provide guidance on how to do the review in the e-mailed message.

The link for good description is this:

https://chromium.googlesource.com/chromium/src/+/main/docs/contributing.md#Uploading-a-change-for-review

This links gives a good example of a good commit message.

To demonstrate this in action, take a look this random CL merged today at Chromium, and how the commit message is being set:

https://chromium-review.googlesource.com/c/chromium/src/+/6939710

components/email_aliases/email_aliases_service.cc
```diff
@@ -57,6 +65,32 @@ const net::NetworkTrafficAnnotationTag traffic_annotation =
 
 constexpr int kMaxResponseLength = 32768;
 
+// Parses a JSON response body into a Dict and surfaces backend error payloads.
+// Returns base::unexpected with a user-facing error string if:
+// - There is no response body
+// - The response is not a dictionary
+// - The response contains an ErrorResponse
+base::expected<base::Value::Dict, std::string> ParseResponseDictOrError(
```
cdesouza-chromium: Maybe  can have a second function that calls this function. Something like:

```cxx
template <typename T>
base::expected<T, std::string> ParseResponseDictAs(
    const std::optional<std::string>& response_body) {
  auto dict_result = ParseResponseDictOrError(response_body);
  if (!dict_result.has_value()) {
    return base::unexpected(dict_result.error());
  }

  auto value_result = T::FromValue(dict_result.value());
  if (!value_result.has_value()) {
    return base::unexpected("Failed to parse value from dictionary.");
  }

  return base::ok(std::move(value_result.value()));
}
```

Not sure, but what do you think?

components/email_aliases/email_aliases_service.cc
```diff
@@ -318,4 +385,155 @@ std::string EmailAliasesService::GetAuthTokenForTesting() const {
...
+                                   BodyAsStringCallback callback) {
+  CHECK(method == net::HttpRequestHeaders::kPostMethod ||
+        method == net::HttpRequestHeaders::kPutMethod ||
+        method == net::HttpRequestHeaders::kDeleteMethod);
+  auto body = base::WriteJson(body_value);
+  CHECK(body);
+  PerformFetch(url, method, /*serialized_body=*/body, std::move(callback));
+}
+
+void EmailAliasesService::PerformFetch(
```
cdesouza-chromium: Maybe `ApiFetchInternal`?

components/email_aliases/email_aliases_service.cc
```diff
@@ -318,4 +385,155 @@ std::string EmailAliasesService::GetAuthTokenForTesting() const {
...
+void EmailAliasesService::OnRefreshAliasesResponse(
+    std::optional<std::string> response_body) {
+  // TODO(https://github.com/brave/brave-browser/issues/48959):
+  // In this function, when an error happens, we should show
+  // an error message to the user (requires design work)
+  if (!response_body) {
+    LOG(ERROR) << "Email Aliases service error: No response body";
+    return;
+  }
+  auto parsed = base::JSONReader::Read(*response_body);
```
cdesouza-chromium, line 491: Is there a reason why we are parsing this manually, rather than having an IDL type for it. It looks like it is just

```
dictionary Foo {
  AliasListEntry messages[];
};
```

components/email_aliases/email_aliases_service.cc
```diff
@@ -318,4 +385,155 @@ std::string EmailAliasesService::GetAuthTokenForTesting() const {
...
+    return;
+  }
+  std::vector<email_aliases::mojom::AliasPtr> aliases;
+  for (const auto& alias_val : parsed->GetList()) {
+    if (!alias_val.is_dict()) {
+      LOG(ERROR) << "Email Aliases service error: Invalid response format";
+      return;
+    }
+    if (auto entry = AliasListEntry::FromValue(alias_val.GetDict())) {
+      auto alias_obj = email_aliases::mojom::Alias::New();
```
cdesouza-chromium: A bit out of scope, but this mojom type looks like:

```
struct Alias {
  // The alias email address for this alias
  string email;
  // Note attached by user
  string? note;
  // Domains where email is used
  array<string>? domains;
};
```

We don't seem to be using any of these fields, or setting them.

components/email_aliases/email_aliases_service.cc
```diff
@@ -318,4 +385,155 @@ std::string EmailAliasesService::GetAuthTokenForTesting() const {
...
+      alias_obj->email = entry->alias;
+      aliases.push_back(std::move(alias_obj));
+    }
+  }
+  for (auto& observer : observers_) {
```
cdesouza-chromium, line 520: Will the application ever have more than one observer for this?

components/email_aliases/email_aliases_service_unittest.cc
```diff
@@ -345,3 +345,337 @@ TEST_F(EmailAliasesServiceTimingTest, VerifyResult_StopsAfterMaxDuration) {
...
+        std::vector<email_aliases::mojom::AliasPtr> aliases) override {
+      ++alias_updates;
+      last_aliases = std::move(aliases);
+    }
+    bool WaitForAliasUpdateCount(size_t count) {
+      return base::test::RunUntil([&]() { return alias_updates >= count; });
+    }
+
+    size_t alias_updates = 0;
+    std::vector<email_aliases::mojom::AliasPtr> last_aliases;
```
cdesouza-chromium: Class fields are expected to be private or protected:

https://google.github.io/styleguide/cppguide.html#Access_Control

components/email_aliases/email_aliases_service_unittest.cc
```diff
@@ -345,3 +345,337 @@ TEST_F(EmailAliasesServiceTimingTest, VerifyResult_StopsAfterMaxDuration) {
...
+            pending) {
+      receiver.Bind(std::move(pending));
+    }
+  };
+
+  base::test::ScopedFeatureList feature_list_;
+  base::test::TaskEnvironment task_environment_;
+  network::TestURLLoaderFactory url_loader_factory_;
+  std::unique_ptr<EmailAliasesService> service_;
+  AliasObserver observer_;
```
cdesouza-chromium: We tend to declare class fields at the bottom of the class, after all member function declarations.

components/email_aliases/email_aliases_service.cc
arthuredelstein: Yes, if the brave://settings/email-aliases page is open in multiple tabs.

components/email_aliases/email_aliases_service.cc
arthuredelstein: Yes, for now we are only using the email address. In the future we will send a note (encrypted) and possibly a domains list, but that is out of scope for this PR.

components/email_aliases/email_aliases_service.cc
arthuredelstein: This api call returns a JSON list, like `[{}, {}, ...]`. So we can't use a dictionary type in the IDL file.

components/email_aliases/email_aliases_service.cc
arthuredelstein: Yes, done.

components/email_aliases/email_aliases_service_unittest.cc
arthuredelstein: fixed

components/email_aliases/email_aliases_service_unittest.cc
arthuredelstein: fixed

components/email_aliases/email_aliases_service.cc
arthuredelstein: I was able to combine `OnUpdateAliasResponse` and `OnDeleteAliasResponse` into a single handler function `OnEditAliasResponse`.

`OnGenerateAliasResponse` is different so I left it as-is.

components/email_aliases/email_aliases_service_unittest.cc
arthuredelstein: I did try to do this, but ran into a problem that results in a crash. It turns out you can't provide a localized string when defining parameters because this happens before `ui::ResourceBundle` is initialized. I was able to switch to using resourceIDs instead of hard-coded English, but I think we will still need to unpack the return values of the tested functions.

arthuredelstein: I pushed a commit that more cleanly uses resource ids. I'm afraid we still need `ExpectResultStateMatches` for the reason I stated above, but it doesn't seem like a major problem to me.

components/email_aliases/email_aliases_service_unittest.cc
```diff
@@ -234,15 +239,12 @@ TEST_F(EmailAliasesServiceTest,
 // rate limiting and total polling duration.
 class EmailAliasesServiceTimingTest : public ::testing::Test {
  protected:
-  EmailAliasesServiceTimingTest()
-      : url_loader_factory_(/*no args*/),
-        url_loader_wrapper_(
-            base::MakeRefCounted<network::WeakWrapperSharedURLLoaderFactory>(
-                &url_loader_factory_)) {}
+  EmailAliasesServiceTimingTest() : url_loader_factory_(/*no args*/) {}
```
szilardszaloki: This can be removed as is.

components/email_aliases/email_aliases_service_unittest.cc
```diff
@@ -345,3 +346,364 @@ TEST_F(EmailAliasesServiceTimingTest, VerifyResult_StopsAfterMaxDuration) {
 // checking cancellation of polling, etc.
 
 }  // namespace email_aliases
+namespace email_aliases {
+
+class AliasObserver : public email_aliases::mojom::EmailAliasesServiceObserver {
```
szilardszaloki: Probably best to handle it in a separate PR, but you can remove the `email_aliases::` prefix everywhere — since you're already inside the `email_aliases` `namespace`, the names don't need to be fully qualified. That way the file will be less cluttered and easier to read.

components/email_aliases/email_aliases_service_unittest.cc
```diff
@@ -345,3 +346,364 @@ TEST_F(EmailAliasesServiceTimingTest, VerifyResult_StopsAfterMaxDuration) {
...
+  std::vector<email_aliases::mojom::AliasPtr> last_aliases;
+  mojo::Receiver<email_aliases::mojom::EmailAliasesServiceObserver> receiver{
+      this};
+};
+
+class EmailAliasesAPITest : public ::testing::Test {
+ protected:
+  EmailAliasesAPITest() {
+    feature_list_.InitAndEnableFeature(email_aliases::kEmailAliases);
+  }
```
szilardszaloki: Nit — you can drop the constructor and just initialize inline:
```cpp
base::test::ScopedFeatureList feature_list_{kEmailAliases};
```

components/email_aliases/email_aliases_service_unittest.cc
```diff
@@ -345,3 +346,364 @@ TEST_F(EmailAliasesServiceTimingTest, VerifyResult_StopsAfterMaxDuration) {
...
+                              const EmailAliasesApiCase& p) {
+  if (p.expected_error) {
+    ASSERT_FALSE(result.has_value());
+    std::string error_string;
+    if (std::holds_alternative<int>(*p.expected_error)) {
+      const int id = std::get<int>(*p.expected_error);
+      error_string = l10n_util::GetStringUTF8(id);
+    } else {
+      error_string = std::get<std::string>(*p.expected_error);
+    }
```
szilardszaloki: The idiomatic approach is to use `std::visit<>()` on the variant rather than unwrapping it by hand, for example:
```cpp
const std::string error_string =
    std::visit(absl::Overload{
                    [](int id) { return l10n_util::GetStringUTF8(id); },
                    [](std::string string) { return string; },
                },
                *p.expected_error);
```

components/email_aliases/email_aliases_service_unittest.cc
arthuredelstein: Thanks for catching that. I removed the unnecessary prefixes for the tests added in this PR.

components/email_aliases/email_aliases_service_unittest.cc
```diff
@@ -344,4 +345,356 @@ TEST_F(EmailAliasesServiceTimingTest, VerifyResult_StopsAfterMaxDuration) {
...
+      public testing::WithParamInterface<EmailAliasesApiCase> {};
+
+TEST_P(GenerateAliasParamTest, HandlesResponses) {
+  const auto& p = GetParam();
+  auto result_out = CallGenerateAliasWith(p.body);
+  ExpectResultStateMatches(result_out, p);
+  if (!p.expected_error) {
+    EXPECT_EQ(result_out.value(), kExpectedGeneratedAlias);
+  }
+}
```
szilardszaloki, line 515: These parameterized tests:
```cpp
TEST_P(GenerateAliasParamTest, HandlesResponses) {
  const auto& p = GetParam();
  auto result_out = CallGenerateAliasWith(p.body);
  ExpectResultStateMatches(result_out, p);
  if (!p.expected_error) {
    EXPECT_EQ(result_out.value(), kExpectedGeneratedAlias);
  }
}
```

```cpp
TEST_P(UpdateAliasParamTest, HandlesResponses) {
  const auto& p = GetParam();
  auto result_out = CallUpdateAliasWith("alias@example.com", p.body);
  ExpectResultStateMatches(result_out, p);
}
```

```cpp
TEST_P(DeleteAliasParamTest, HandlesResponses) {
  const auto& p = GetParam();
  auto result_out = CallDeleteAliasWith("alias@example.com", p.body);
  ExpectResultStateMatches(result_out, p);
}
```
are essentially doing the same: validating that the API under test returns either the expected value or the expected error string for each input case, differing only in which service method (`GenerateAlias()`, `UpdateAlias()`, `DeleteAlias()`) they call.

Ideally, the test bodies remain identical, with the only variation coming from how the underlying parameterized test fixture is instantiated, e.g.:
```cpp
TEST_P(GenerateAliasParamTest, HandlesResponses) {
  RunTestCase();
}
```

```cpp
TEST_P(UpdateAliasParamTest, HandlesResponses) {
  RunTestCase();
}
```

```cpp
TEST_P(DeleteAliasParamTest, HandlesResponses) {
  RunTestCase();
}
```
where
```cpp
using GenerateAliasParamTest = SomeParameterizedTestFixture<GenerateAliasTestCase>;
```

```cpp
using UpdateAliasParamTest = SomeParameterizedTestFixture<UpdateAliasTestCase>;
```

```cpp
using DeleteAliasParamTest = SomeParameterizedTestFixture<DeleteAliasTestCase>;
```

components/email_aliases/email_aliases_service_unittest.cc
```diff
@@ -344,4 +345,356 @@ TEST_F(EmailAliasesServiceTimingTest, VerifyResult_StopsAfterMaxDuration) {
...
+INSTANTIATE_TEST_SUITE_P(
+    EmailAliasesGenerateAlias,
+    GenerateAliasParamTest,
+    testing::Values(
+        EmailAliasesApiCase{.name = "Success",
+                            .body = MakeGenerateAliasSuccessBody(),
+                            .expected_error = std::nullopt},
+        EmailAliasesApiCase{
+            .name = "BackendError",
+            .body = "{\"message\":\"alias_unavailable\"}",
```
szilardszaloki: Nit: consider using raw string literals (e.g. `.body = R"({ "message": "alias_unavailable" })",`) — they avoid the need to escape quotes and make JSON bodies more readable.

szilardszaloki:
@arthuredelstein My suggestion — start by defining a matcher:
```cpp
MATCHER_P(MatchesExpected, expected_result, "") {
  if (expected_result.has_value()) {
    return arg.has_value() && arg.value() == expected_result.value();
  }

  const std::string expected_error =
      std::visit(absl::Overload{
                     [](int id) { return l10n_util::GetStringUTF8(id); },
                     [](std::string string) { return string; },
                 },
                 expected_result.error());

  return !arg.has_value() && base::Contains(arg.error(), expected_error);
}
```
Define a generic test case `struct`, templated on the success type of the `base::expected<>` result:
```cpp
template <typename T>
struct EmailAliasesAPIParamTestCase {
  static constexpr auto kNameGenerator = [](const auto& info) {
    return info.param.name;
  };

  const char* name;
  std::optional<std::string> body;
  base::expected<T, std::variant<int, std::string>> expected_result;
};
```
Define a parameterized test fixture `class` template that takes the test case type as its template parameter:
```cpp
template <typename TestCase>
class EmailAliasesAPIParamTest : public EmailAliasesAPITest,
                                 public testing::WithParamInterface<TestCase> {
 protected:
  void RunTestCase() {
    const auto& test_case = this->GetParam();
    EXPECT_THAT(TestCase::Run(*this, test_case),
                MatchesExpected(test_case.expected_result));
  }
};
```
And finally instantiate the fixture with the appropriate test case types and wire up all your parameterized test suites with the appropriate parameter values, e.g.:
```cpp
struct GenerateAliasTestCase : EmailAliasesAPIParamTestCase<std::string> {
  static auto Run(EmailAliasesAPITest& test,
                  const GenerateAliasTestCase& test_case) {
    return test.CallGenerateAliasWith(test_case.body);
  }
};

using GenerateAliasParamTest = EmailAliasesAPIParamTest<GenerateAliasTestCase>;

TEST_P(GenerateAliasParamTest, HandlesResponses) {
  RunTestCase();
}

INSTANTIATE_TEST_SUITE_P(
    EmailAliasesGenerateAlias,
    GenerateAliasParamTest,
    testing::Values(
        GenerateAliasTestCase{
            {.name = "Success",
             .body =
                 R"({ "message": "created",
                      "alias": "mock-1234@bravealias.com" })",
             .expected_result = base::ok("mock-1234@bravealias.com")}},
        GenerateAliasTestCase{
            {.name = "BackendError",
             .body = R"({ "message": "alias_unavailable" })",
             .expected_result = base::unexpected(
                 IDS_EMAIL_ALIASES_ERROR_ALIAS_NOT_AVAILABLE)}},
        GenerateAliasTestCase{
            {.name = "NoBody",
             .body = std::nullopt,
             .expected_result = base::unexpected(
                 IDS_EMAIL_ALIASES_SERVICE_ERROR_NO_RESPONSE_BODY)}},
        GenerateAliasTestCase{
            {.name = "InvalidJSON",
             .body = "not a json",
             .expected_result = base::unexpected(
                 IDS_EMAIL_ALIASES_SERVICE_ERROR_INVALID_RESPONSE_BODY)}},
        GenerateAliasTestCase{
            {.name = "UnexpectedPayload",
             .body = R"({ "message": "ok_but_no_alias" })",
             .expected_result = base::unexpected(
                 IDS_EMAIL_ALIASES_ERROR_ALIAS_NOT_AVAILABLE)}}),
    GenerateAliasTestCase::kNameGenerator);
```
This way you cut down on boilerplate, keep things consistent, make the cases more declarative, and adding new APIs is just about defining inputs and expected results.

szilardszaloki:
@arthuredelstein I've got these changes on my side — will share the file with you.

components/email_aliases/email_aliases_service_unittest.cc
szilardszaloki: See https://github.com/brave/brave-core/pull/31027#issuecomment-3314372055 below.

components/email_aliases/email_aliases_service_unittest.cc
arthuredelstein: done

components/email_aliases/email_aliases_service_unittest.cc
arthuredelstein: done

components/email_aliases/email_aliases_service_unittest.cc
arthuredelstein: done

components/email_aliases/email_aliases_service_unittest.cc
arthuredelstein: done

components/email_aliases/email_aliases_service.cc
arthuredelstein: Unfortunately it doesn't appear there's a way to create a standalone list type?

components/email_aliases/email_aliases_service_unittest.cc
arthuredelstein: done

components/email_aliases/email_aliases_service_unittest.cc
arthuredelstein: I pushed @szilardszaloki's proposed changes, 309fc60

components/email_aliases/email_aliases_service_unittest.cc
```diff
@@ -344,4 +344,363 @@ TEST_F(EmailAliasesServiceTimingTest, VerifyResult_StopsAfterMaxDuration) {
...
+ public:
+  void OnAuthStateChanged(mojom::AuthStatePtr) override {}
+
+  void OnAliasesUpdated(std::vector<mojom::AliasPtr> aliases) override {
+    ++alias_updates;
+    last_aliases = std::move(aliases);
+  }
+
+  bool WaitForAliasUpdateCount(size_t count) {
+    return base::test::RunUntil([&]() { return alias_updates >= count; });
```
Olipro: I believe the style guide suggests preferring explicit captures. Also, is `RunUntil` returning a callable or executing it straight away? If the former, I believe `count` will be out of its lifetime and causing UB.

My $0.02 would be to make the capture `[this, count]`

components/email_aliases/email_aliases_service_unittest.cc
```diff
@@ -344,4 +344,363 @@ TEST_F(EmailAliasesServiceTimingTest, VerifyResult_StopsAfterMaxDuration) {
...
+    base::test::TestFuture<base::expected<T, std::string>> future;
+    std::forward<InvokeFn>(invoker)(future.GetCallback());
+    return future.Take();
+  }
+
+  base::expected<std::string, std::string> CallGenerateAliasWith(
+      const std::optional<std::string>& body) {
+    AddManageResponseFor(body);
+    return InvokeAndWait<std::string>(
+        [&](auto cb) { service_->GenerateAlias(std::move(cb)); });
```
Olipro: Similar case to my previous comment, `[this]` should do the job and be explicit at the cost of just 3 additional characters :)

components/email_aliases/email_aliases_service_unittest.cc
```diff
@@ -344,4 +344,362 @@ TEST_F(EmailAliasesServiceTimingTest, VerifyResult_StopsAfterMaxDuration) {
...
+  }
+
+  base::expected<std::monostate, std::string> CallUpdateAliasWith(
+      const std::string& alias_email,
+      const std::optional<std::string>& put_body,
+      const std::optional<std::string>& refresh_body = std::nullopt,
+      bool wait_for_update = true) {
+    AddManageResponseFor(put_body);
+    AddRefreshResponseFor(refresh_body);
+    auto result_out = InvokeAndWait<std::monostate>([&](auto cb) {
```
Olipro: Ditto as for my previous comments - `[this, &alias_email]`

I'm not fluent in the mechanics of `InvokeAndWait` - I take it that this never returns the contained lambda itself for later execution since, if it does, whoever calls this member function must ensure that `alias_email` is within its lifetime until the lambda is disposed of.

components/email_aliases/email_aliases_service_unittest.cc
```diff
@@ -344,4 +344,362 @@ TEST_F(EmailAliasesServiceTimingTest, VerifyResult_StopsAfterMaxDuration) {
...
+    return result_out;
+  }
+
+  base::expected<std::monostate, std::string> CallDeleteAliasWith(
+      const std::string& alias_email,
+      const std::optional<std::string>& delete_body) {
+    AddManageResponseFor(delete_body);
+    AddRefreshResponseFor(/*refresh_body=*/std::nullopt);
+    auto result_out = InvokeAndWait<std::monostate>(
+        [&](auto cb) { service_->DeleteAlias(alias_email, std::move(cb)); });
```
Olipro: Ditto as for https://github.com/brave/brave-core/pull/31027/files#r2368959433

components/email_aliases/email_aliases_service_unittest.cc
arthuredelstein: done

components/email_aliases/email_aliases_service_unittest.cc
arthuredelstein: done

components/email_aliases/email_aliases_service_unittest.cc
arthuredelstein: done

components/email_aliases/email_aliases_service_unittest.cc
arthuredelstein: done.

I just want to note that I found many thousands of implicitly capturing lambdas in Chromium code and brave-core. :)

components/email_aliases/email_aliases_service.cc
cdesouza-chromium: Thanks. I did notice though that there are two sites calling `ParseResponseDictOrError`, and both of them are hadling calling `T::ToValue()` which the templated trampoline I suggested above could help with.

components/email_aliases/email_aliases_service.cc
```diff
@@ -72,11 +111,24 @@ GURL EmailAliasesService::GetAccountsServiceVerifyResultURL() {
                               kAccountsServiceVerifyResultPath));
 }
 
+// static
+GURL EmailAliasesService::GetEmailAliasesServiceBaseURL() {
+  return GURL(base::StrCat(
+      {kEmailAliasesServiceBaseURL, kEmailAliasesServiceManagePath}));
```
cdesouza-chromium: We are doing concat for two strings we now at compile time, and then storing the resulting value in a `std::string`. This seems a bit wasteful, as we could just have a single `constexpr` constant with the resulting URL, and to just instantiate the GURL during use.

components/email_aliases/email_aliases_service.cc
```diff
@@ -318,4 +392,144 @@ std::string EmailAliasesService::GetAuthTokenForTesting() const {
   return auth_token_;
 }
 
+void EmailAliasesService::ApiFetch(const GURL& url,
+                                   const char* method,
```
cdesouza-chromium: This should be a `string_view`.

components/email_aliases/email_aliases_service.cc
```diff
@@ -318,4 +392,144 @@ std::string EmailAliasesService::GetAuthTokenForTesting() const {
...
+                                   BodyAsStringCallback callback) {
+  CHECK(method == net::HttpRequestHeaders::kGetMethod ||
+        method == net::HttpRequestHeaders::kHeadMethod);
+  ApiFetchInternal(url, method, /*serialized_body=*/std::nullopt,
+                   std::move(callback));
+}
+
+void EmailAliasesService::ApiFetch(const GURL& url,
+                                   const char* method,
```
cdesouza-chromium: Same here.

components/email_aliases/email_aliases_service.cc
```diff
@@ -318,4 +392,144 @@ std::string EmailAliasesService::GetAuthTokenForTesting() const {
...
+        method == net::HttpRequestHeaders::kPutMethod ||
+        method == net::HttpRequestHeaders::kDeleteMethod);
+  auto body = base::WriteJson(body_value);
+  CHECK(body);
+  ApiFetchInternal(url, method, /*serialized_body=*/body, std::move(callback));
+}
+
+void EmailAliasesService::ApiFetchInternal(
+    const GURL& url,
+    const char* method,
```
cdesouza-chromium: Same.

components/email_aliases/email_aliases_service.cc
cdesouza-chromium: That's okay for now, but going forward, when dealing with mojom types and C++ types, only add fields to a type when you have an actual use for them

components/email_aliases/email_aliases_service.h
```diff
@@ -117,6 +132,41 @@ class EmailAliasesService : public KeyedService,
...
+
+  // Processes the server response for a generate-alias request.
+  void OnGenerateAliasResponse(GenerateAliasCallback user_callback,
+                               std::optional<std::string> response_body);
+
+  // Common handler for alias edit responses (update/delete).
+  void OnEditAliasResponse(
+      base::OnceCallback<void(base::expected<std::monostate, std::string>)>
+          user_callback,
+      std::string_view expected_message,
```
cdesouza-chromium: When calling this function, a string_view is being dispatched through a closure with the implied trust that the underlying pointer is valid. This is error prone if people assume otherwise.

Maybe this should be `OnEditAliasResponseInternal`, and then you should have two function completing the call and calling this function with the expectd messaged, that way this string_view is guaranteed to be safe.

components/email_aliases/email_aliases_service.h
```diff
@@ -146,13 +196,22 @@ class EmailAliasesService : public KeyedService,
   // Cached fully-qualified verify/result URL.
   const GURL verify_result_url_;
 
+  // Cached fully-qualified email aliases service base URL.
+  const GURL email_aliases_service_base_url_;
+
+  // Cached email aliases API key.
+  const std::string email_aliases_api_key_;
```
cdesouza-chromium, line 203: I'm not convinced we need this value taking memory on this type. This is the source of this info:

```cxx
// static
std::string EmailAliasesService::GetEmailAliasesServiceAPIKey() {
  return BUILDFLAG(EMAIL_ALIASES_API_KEY);
}
```

This unnecessarily obfuscate a simple static data known at compile time. This could be just:

```cxx
namespace {

constexpr char EmailAliasesServiceAPIKey[] = BUILDFLAG(EMAIL_ALIASES_API_KEY);

}
```

If  you need to read this value during tests, just have a `ForTesting` static, but that should suffice. Please bear in mind that whenever a value can be just a constexpr static, we should favour that. If this instance can exist throughout the browsers existence, this means we are paying memory use for stationary values, which could be compounded by padding bytes on this type.

I would suggest reviewing the other URL fields to see if they can just be `constexpr` static values.

components/email_aliases/email_aliases_service.cc
arthuredelstein: done

components/email_aliases/email_aliases_service.cc
arthuredelstein: done

components/email_aliases/email_aliases_service.cc
arthuredelstein: done

components/email_aliases/email_aliases_service.h
arthuredelstein: I changed that parameter to a boolean, because there are only two possible expected values. That avoids the need to trust an underlying pointer.

components/email_aliases/email_aliases_service.cc
arthuredelstein: We could do that in principle, but given there are various other differences between `OnGenerateAliasResponse` and `OnEditAliasResponse` it seemed more readable to leave it as-is.

components/email_aliases/email_aliases_service.cc
arthuredelstein: done

components/email_aliases/email_aliases_service.h
arthuredelstein: I changed `kEmailAliasesServiceAPIKey` to a constexpr.

The other URL fields can't be static if we want to use `GetServicesDomain`. If we want to change these I would suggest using a follow-up.

components/email_aliases/email_aliases_service.cc
cdesouza-chromium: Do we control the server API for this?

components/email_aliases/email_aliases_service.cc
cdesouza-chromium: The function above is not to replace `OnGenerateAliasResponse` or `OnEditAliasResponse`. It is rather to move the calls to `T::FromValue()` to a single place, but `OnGenerateAliasResponse` and `OnEditAliasResponse` would still call it.

components/email_aliases/email_aliases_service.cc
arthuredelstein: Fixed.

I added an AliasListResponse type. While the server API still responds with the old JSON list, the code will wrap the list in a dictionary before parsing. In the future, when we update the server API (in https://github.com/brave-intl/email-alias/issues/115) then we can remove the wrapping code.

components/email_aliases/email_aliases_service.cc
arthuredelstein: Fixed. I created a function, `ParseResponseDictAs` that does this: 586b5b5

components/email_aliases/email_aliases_service_unittest.cc
arthuredelstein: We pushed Szilard's suggested code here.

components/email_aliases/BUILD.gn
arthuredelstein: I added more details change description and got approval from @cdesouza-chromium

PR #31028, opened 2025-09-03
Updated WireGuard tunnel library to tag `1.0.20250531`

bsclifton:
This includes the important commit:
https://git.zx2c4.com/wireguard-android/commit/?id=a57ca57e6c93f7db3253194d2880dcc781ea7c88

Fixes https://github.com/brave/brave-browser/issues/48949

### DIFF
--- a/third_party/android_deps/BUILD.gn
+++ b/third_party/android_deps/BUILD.gn
@@ -55,7 +55,7 @@ android_aar_prebuilt("com_github_bumptech_glide_gifdecoder_java") {
 }
 
 android_aar_prebuilt("com_wireguard_android_java") {
-  aar_path = "libs/com_wireguard_android/tunnel-1.0.20230321.aar"
+  aar_path = "libs/com_wireguard_android/tunnel-1.0.20250531.aar"
   info_path = "libs/com_wireguard_android/com_wireguard_android.info"
   deps = [
     "//third_party/androidx:androidx_annotation_annotation_java",


Olipro (approved):
LGTM.

Would there be any interest in having this AAR build from source as part of building `brave-core`? If so, I'll give it a shot.

bsclifton:
@Olipro you are definitely welcome to check that out - I captured that task with https://github.com/brave/brave-browser/issues/39328

You can see the steps that happen for Desktop and then the steps needed for Android I captured there too (most recent post). We can split the two apart (Desktop / Android) - I'm more than happy to tackle the desktop one 😄

bsclifton:
BTW - I build the APK locally for this locally and loaded onto my Pixel 8. I was able to go to account.brave.com where I already have a VPN subscription and click the button there to load the credentials. At that point, I could connect using the `...` menu
image&nbsp;image

bsclifton:
Note to reviewers: we can absolutely run the attached script on a CI node to build the AAR that we want to distribute.
I did notice that the build seems to be deterministic - when I delete the binary and rebuild from source I get the same SHA. It's possible we'd get the same thing on a CI node too.

kdenhartog (approved):
I've done some testing and auditing of the DEP update and this all looks good to me for security

bsclifton:
Added @bridiver as a reviewer as he's [assigned to the dependency review](https://github.com/brave/reviews/issues/2036) also  😄

tools/android/wireguard-build.sh
```diff
@@ -0,0 +1,46 @@
...
+echo "JAVA_HOME=$JAVA_HOME"
+echo "ANDROID_HOME=$ANDROID_HOME"
+
+BRAVE_DEPS_DIR=`realpath ./third_party/android_deps/libs/com_wireguard_android`
+echo "Changing directory to $BRAVE_DEPS_DIR"
+cd "$BRAVE_DEPS_DIR"
+pushd .
+
+echo "If you haven't accepted licenses yet this is interactive."
+sdkmanager --licenses
```
bridiver, line 26: where does this come from?

tools/android/wireguard-build.sh
```diff
@@ -0,0 +1,46 @@
...
+
+WIREGUARD_TAG="1.0.20250531"
+REPO_URL="https://git.zx2c4.com/wireguard-android"
```
bridiver, line 30: this should be a conditional dep in DEPS, maybe `(checkout_wireguard and (checkout_android or checkout_linux))` guard where `checkout_wireguard` is added to `vars = { ...  }`
Also we should use the commit ref instead of tags because tags can be changed and potentially result in building different code than we expect

tools/android/wireguard-build.sh
bsclifton: When you try to build it locally without having accepted the licenses, the gradle step will fail. I just did this so it forces the person to input `Y` over and over again until all are accepted. Only needs to happen once per install, AFAIK.

bsclifton: Maybe @simoarpe or @SergeyZhukovsky know offhand if it's required. Maybe since we're using the Brave version of JDK, we're already guaranteed to have accepted licenses?

tools/android/wireguard-build.sh
bsclifton: This is a rev1 script - I could make this a proper hook, but decided to just capture what I used to build the library. Adding it as a hook was going to be covered by https://github.com/brave/brave-browser/issues/39328 which I was going to take on at a later time.

tools/android/wireguard-build.sh
SergeyZhukovsky: yeah, it's mostly when a new package has to be downloaded and the license needs to be accepted, otherwise it fails. So most likely on the environment where Brave for Android was already built it's not needed, but it won't hurt to have it imo.

bridiver (approved):

tools/android/wireguard-build.sh
bsclifton: Added `WIREGUARD_COMMIT` above

Build is deterministic, running again builds exact same binary.

bsclifton:
Going to merge - the CI failures above are not related at all to this change

hffvld:
Verified on `Pixel 7` using version(s):
```
Device/OS: Pixel 7 / panther_beta-user 16 BP41.250822.010 release-keys
Brave build: 1.85.28 
Chromium: 141.0.7390.55 (Official Build) canary (64-bit)
```
---------------------------

**STEPS:**
1. Follow the STR/TP from https://github.com/brave/brave-core/pull/31028#issue-3381632273
2. *Verify*

**ACTUAL RESULTS:**
- Verified that VPN connection can be established by loading active credentials via `account.brave.com`.
- Verified that the IP address is changing successfully by updating the VPN region and checking it via https://ipleak.net or https://whatismyipaddress.com/.

---------------------------

https://github.com/user-attachments/assets/d17c8474-4f9b-417c-8b31-69241ec3f84d

PR #31042, opened 2025-09-04
Block Lottie themed images for credit card / password / address save dialogs

emerick:
We want to block the various header images show on the credit card, password, and address save dialogs. Since those functions are spread across various classes, we block the images via an override for the upstream helper function that retrieves them.

The set of blocked images are initialized at startup, so we can avoid the need to explicitly reference the blocked resource IDs by name in the helper which would cause a layering violation (they're defined in "chrome/grit/browser_resources.h")/

Resolves https://github.com/brave/brave-browser/issues/48901

### DIFF
--- a/test/filters/browser_tests.filter
+++ b/test/filters/browser_tests.filter
@@ -2454,6 +2454,11 @@
 # see https://github.com/brave/brave-browser/issues/49422
 -LocalNetworkAccessBrowserTest.HtmlScriptSrcAllowPermission
 
+# This test fails because we stub ResourceBundle::GetLottieThemedImageNamed to
+# return an empty ImageModel, but this test expects to run an ImageGenerator off
+# its returned Lottie ImageModel and first DCHECKs that it's not empty
+-FeatureFirstRunDialogHelperBrowserTest.DialogConstructedFromParams
+
 # Tests below this point have not been diagnosed or had issues created yet.
 -_/WebrtcLoggingPrivateApiStartEventLoggingTestPolicyEnabled.*
 -AdsPageLoadMetricsObserverBrowserTest.*


chromium_src/ui/base/resource/resource_bundle.cc
```diff
@@ -0,0 +1,36 @@
...
+
+COMPONENT_EXPORT(UI_BASE)
+void SetBlockedThemedLottieImages(const base::flat_set<int>& blocked_ids) {
+  *g_blocked_lottie_ids = blocked_ids;
+}
+
+const ui::ImageModel& ResourceBundle::GetThemedLottieImageNamed(
+    int resource_id) {
+  // Don't display themed assets that are on our block list
+  if (g_blocked_lottie_ids->count(resource_id)) {
```
bridiver: talking to @rebron we want to block all of these, but returning an empty image will break in some places so we could override `GetLottieData` to return a hardcoded json value that corresponds to an empty image in lottie so it returns a valid value, but displays nothing?

bridiver: I think this should work 
```
{
  "v": "5.5.7",
  "fr": 1,
  "ip": 0,
  "op": 0,
  "w": 0,
  "h": 0,
  "nm": "Blank",
  "ddd": 0,
  "assets": [],
  "layers": []
}
```

emerick: Yeah, I was trying something similar, but unfortunately zero-sized canvases are prohibited:

```
[63272:64112:0909/173656.887:ERROR:cc\paint\skottie_wrapper_impl.cc:46] [Skottie] "Invalid animation params (size: [0.000000 0.000000], frame rate: 30.000000, in-point: 0.000000, out-point: 1.000000)"
```

If I use a 1x1 canvas instead, it does the right thing for two of the images I'm trying to hide. However, the image we want to hide on the password bubble seems to get inflated up to some "preferred size", so while it's blank it also takes up a ton of layout space.

bridiver: Wait, this doesn't crash for `IDR_AUTOFILL_SAVE_CARD_LOCAL_LOTTIE` in `SaveIbanBubbleView::AddedToWidget`? If not why don't we just always return `empty_image` and get rid of the list?

bridiver: what I'm trying to get rid of the the block list so we just block all of them

emerick: I'm fine with blocking all of them, I just thought there could potentially be non-branded images (now or in the future) that we might not want to block. If we don't think that's a valid concern, getting rid of the list would definitely simplify things.

emerick: OK, updated to block all Lottie assets.

emerick:
Per @rebron we're going to replace the images as needed instead of hide them, so abandoning this PR.

PR #31046, opened 2025-09-05
[CodeHealth] Replace blink::StringBuilder with blink::StrCat part 2

DJAndries:
Resolves https://github.com/brave/brave-browser/issues/48992


bridiver (approved):
renderer changes ok

PR #31050, opened 2025-09-05
Removed split view separator menu

simonhong:
Resolves https://github.com/brave/brave-browser/issues/48983


sangwoo108 (approved):
LGTM. How about removing split_view_separator.h, cc ?

simonhong:
> LGTM. How about removing split_view_separator.h, cc ?

Let's delete them together with whole brave split view code later.

sangwoo108:
> > LGTM. How about removing split_view_separator.h, cc ?
> 
> Let's delete them together with whole brave split view code later.

Good call

chromium_src/chrome/browser/ui/views/frame/multi_contents_resize_area.cc
```diff
@@ -0,0 +1,13 @@
+/* Copyright (c) 2025 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "chrome/browser/ui/views/frame/multi_contents_view.h"
+
+// Reset resize area position when double clicked instead of swap.
```
bridiver, line 8: why?

simonhong: Our split view's behavior for double click is reset position. so ported here.
I think it's expected behavior generally - https://bravesoftware.slack.com/archives/C20HMJ62H/p1755864407677579

If we need to rethink behavior, let's f/u in a separated issue.

kjozwiak:

```
Brave | 1.84.93 Chromium: 141.0.7390.30 (Official Build) nightly (64-bit)
-- | --
Revision | 5c2b6ff0f689c574ae2db8fd18e05139c3154c21
```

Using the STR/information outlined via https://github.com/brave/brave-browser/issues/48983#issue-3385425252, went through the following:

* ensured that the middle options/context menu was removed in favour of the icon/settings via the URL toolbar
* ensured that the `Open in split view` icon only appears when viewing a page within a split view
  * changing/switching to another tab that's not in `split view` shouldn't display the `Open in split view` icon
* ensure that the icon is correctly highlighting which side of the `split view` is currently active/in focus
* ensured that clicking on the `Open in split view` icon displays a list of context items without any issues
  * ensured that `unsplit tabs` works as expected
  * ensured that both `Close left views` & `Close right views` works as expected
  * ensured that `Swap tab positions` works as expected (tried multiple times)
  * ensured that `Send feedback` correctly opens the `Send feedback to brave` with a screenshot included
* ensured that right clicking on the `Open in split view` icon displays `Pin` & `Customise toolbar`
  * ensured that enabling/disabling `Pin` enables the icon via the `toolbar` management under `brave://settings/appearance`
  * ensured that clicking on `Customise` opens the `Toolbar` settings without issues
* ensured that you can drag the middle separator without any issues/failures now that the separator menu was removed

Example video of the above STR/Cases being checked/verified:

https://github.com/user-attachments/assets/0e9babaf-4543-46da-9760-381d4cf86171

PR #31051, opened 2025-09-05
Replace "Darker" with "Midnight" for branding consistency.

sangwoo108:
This PR replaces instances of "Darker" with "Midnight" in cases where the term is revealed to users, such as in feature flags and UI strings. For branding consistency, we'll stick with "Midnight" as the theme name.

Resolves https://github.com/brave/brave-browser/issues/48997

### DIFF
--- a/components/resources/customize_chrome_ui_strings.grdp
+++ b/components/resources/customize_chrome_ui_strings.grdp
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <grit-part>
   <message name="IDS_CUSTOMIZE_CHROME_DARKER_THEME_TOGGLE_LABEL" desc="Title for a toggle button to show or hide Midnight Theme in Customize Chrome sidebar." formatter_data="webui=CustomizeChrome">
-    Use Midnight theme
+    Midnight theme
   </message>
 </grit-part>


browser/ui/darker_theme/features.cc
```diff
@@ -8,7 +8,7 @@
 namespace darker_theme::features {
 
 BASE_FEATURE(kBraveDarkerTheme,
```
simonhong, line 12: why not replace `kBraveDarkerTheme` with `kBraveMidnightTheme` also?

sangwoo108: I'm not sure about it. If we go with that, the target name should be renamed too, but I feel like midnight_theme is somewhat ambiguous.

simonhong: For one feature, we're saying it with two names - darker theme/midnight theme. do we have to this?
IMO, we can unify one - midnight theme.

sangwoo108: Yeah, I still think it'd be worth it. Midnight is also used in the new dev channel's name, and this theme will be applied to all channels. So for now, I'd like to keep it as is. But I'm kind of on the fence about it, so I'm looking forward to hearing your thoughts.

simonhong: Well.. for me, both term `BraveDarkerTheme` and `BraveMidnightTheme` seems different theme.
Maybe it's difficult to think `BraveMidnightTheme` equals `BraveDarkerTheme` in the code at first.

As this feature is inside the `darker_theme` namespace, I think we could have another darker theme named  midnight. and it would not be conflict with midnight channel. WDYT?

sangwoo108: So, are you suggesting we have `darker_theme::kBraveMidnightTheme` and leave namespace and build target as is(darker_theme)?

simonhong: Or, why not just changing `about_flags.cc` if what we want is showing it as `midnight` to users?
I think it's sufficient. It's strange that its variable is darker and its name is midnight.

simonhong: At first, it's weird that different two words were used here.
so, asked to use one name (midnight).
and thought again what is the goal of this PR and it's making dark theme visible as midnight to users.
then I thought this file's change is not needed for that purpose as it's sufficient to change other changes except here. and I'm thinking why you changed here also and maybe you think `BASE_FEATURE`'s `name` args is also exported part. Am I right?

simonhong: Ah, I realized that this name could be used by users via command line switch.
Ok, it makes sense to have midnight here. It would be great to have some comments that it's darker in the code but expoed as midnight.

sangwoo108: > and thought again what is the goal of this PR and it's making dark theme visible as midnight to users.

Yes exactly that's what we want. I added the change in this file because users will use this string when they want to enable this feature via --enable-features= command line, which I considered as user visible string.

I think having "darker_theme::kBraveMidnight" and "BraveMidnightTheme" as its name would be good.

sangwoo108: Added comment. Thanks :)

components/resources/customize_chrome_ui_strings.grdp
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <grit-part>
-  <message name="IDS_CUSTOMIZE_CHROME_DARKER_THEME_TOGGLE_LABEL" desc="Title for a toggle button to show or hide Darker Theme in Customize Chrome sidebar." formatter_data="webui=CustomizeChrome">
-    Use darker theme
+  <message name="IDS_CUSTOMIZE_CHROME_DARKER_THEME_TOGGLE_LABEL" desc="Title for a toggle button to show or hide Midnight Theme in Customize Chrome sidebar." formatter_data="webui=CustomizeChrome">
+    Use Midnight theme
```
mkarolin: I am wondering if saying `Use` here would be confusing as this makes me think I am turning Midnight theme on/off here instead of just showing/hiding the toggle. How about just `Midnight theme`? cc: @rebron

sangwoo108: @rebron confirmed that "Midnight theme" is okay. Changed.

PR #31054, opened 2025-09-05
feat(gate3): support price lookups via gate3

onyb:
Resolves


components/brave_wallet/common/brave_wallet.mojom
```diff
@@ -474,11 +474,31 @@ struct AssetTimePrice {
   string price;
 };
 
+struct AssetPriceRequest {
+  CoinType coin_type;
+  string? chain_id;
+  string? address;
```
supermassive, line 483: Let's have a comment what is `address` here and in `AssetPrice`

supermassive: Also I'd rename it to token_address, or contract_address or else. So it is not just `address`

components/brave_wallet/common/brave_wallet.mojom
```diff
@@ -474,11 +474,31 @@ struct AssetTimePrice {
...
 struct AssetPrice {
-  string from_asset;
-  string to_asset;
+  CoinType coin_type;
+  string chain_id;
+  string address;
   string price;
-  string asset_timeframe_change;
+  string vs_currency;
+  Gate3CacheStatus cache_status;
```
supermassive, line 503: Why frontend needs cache hit/miss status?

components/brave_wallet/browser/asset_ratio_response_parser.cc
```diff
@@ -41,75 +70,88 @@ std::optional<std::string> ParseSardineAuthToken(
...
+      continue;
+    }
+
+    auto asset_price = mojom::AssetPrice::New();
+    asset_price->coin_type = *coin_type;
+    if (chain_id) {
+      asset_price->chain_id = *chain_id;
+    }
+    if (address) {
+      asset_price->address = *address;
```
supermassive: Can we start using .idl defintion for this parsing? wdyt?

components/brave_wallet/browser/asset_ratio_response_parser.cc
```diff
@@ -41,75 +70,88 @@ std::optional<std::string> ParseSardineAuthToken(
...
-  //    }
-  //  },
-  //  "lastUpdated": "2021-08-16T15:45:11.901Z"
-  // }
-
-  DCHECK(values);
+std::vector<mojom::AssetPricePtr> ParseAssetPrices(
+    const base::Value& json_value) {
+  // Parses results from the new /api/pricing/v1/getPrices endpoint:
+  // [
```
supermassive: Let's better have a link to format specs

components/brave_wallet/browser/simple_hash_client_unittest.cc
```diff
@@ -1008,7 +1008,7 @@ TEST_F(SimpleHashClientUnitTest, FetchAllNFTsFromSimpleHash) {
   })";
   responses[url] = json;
   GURL next_url = GURL(
-      "https://simplehash.wallet.brave.com/api/v0/nfts/"
+      "https://gate3.wallet.brave.com/simplehash/api/v0/nfts/"
       "owners?chains=ethereum%2Coptimism&wallet_addresses="
```
supermassive, line 1012: Should this file still have `simple_hash` in its name?

components/brave_wallet/common/brave_wallet.mojom
onyb: This is mainly for debugging on the console - never displayed in the UI

components/brave_wallet/browser/simple_hash_client_unittest.cc
onyb: `gate3.wallet.brave.com/simplehash` is a drop-in replacement for the now-deprecated simplehash API. Since we're still using Simplehash response format, I kept the name in the URL. In a future PR, I'll change it to the new response format, which won't have the "simplehash" word in the URL.

components/brave_wallet/browser/asset_ratio_service.cc
```diff
@@ -366,24 +379,32 @@ void AssetRatioService::GetSellUrl(mojom::OffRampProvider provider,
...
   }
-  std::vector<std::string> from_assets_lower = VectorToLowerCase(from_assets);
-  std::vector<std::string> to_assets_lower = VectorToLowerCase(to_assets);
+
+  // Create JSON payload for the new API
+  std::string json_payload = CreatePricingRequestPayload(requests);
+
+  // Create URL with vs_currency query parameter
+  GURL url = GetPriceURL(vs_currency);
+
```
supermassive: I'd say code is self-documenting and we don't need these comments

components/brave_wallet/browser/asset_ratio_service.cc
```diff
@@ -123,19 +107,52 @@ namespace brave_wallet {
...
+    const std::vector<mojom::AssetPriceRequestPtr>& requests) {
+  base::Value::List response;
+
+  for (const auto& request : requests) {
+    base::Value::Dict response_item;
+    auto coin_type_str = GetStringFromCoinType(request->coin_type);
+    if (coin_type_str) {
+      response_item.Set("coin_type", *coin_type_str);
+    }
+
```
supermassive: Should fail request when no valid coin_type

android/java/org/chromium/chrome/browser/crypto_wallet/util/AssetsPricesHelper.java
```diff
@@ -23,56 +18,8 @@ public static void fetchPrices(
             AssetRatioService assetRatioService,
             BlockchainToken[] assets,
             Callbacks.Callback1<HashMap<String, Double>> callback) {
+        // Method is not used - return empty HashMap to maintain compatibility
         HashMap<String, Double> assetsPrices = new HashMap<String, Double>();
```
supermassive: Let's remove completely if it is a no-op now

components/brave_wallet/browser/asset_ratio_service.cc
```diff
@@ -200,17 +217,13 @@ GURL AssetRatioService::GetSardineBuyURL(const std::string& chain_id,
...
-                      from, to, TimeFrameKeyToString(timeframe));
-  return GURL(spec);
+GURL AssetRatioService::GetPriceURL(const std::string& vs_currency) {
+  std::string base_url =
+      base_url_for_test_.is_empty() ? kGate3URL : base_url_for_test_.spec();
+  GURL url = GURL(absl::StrFormat("%s/api/pricing/v1/getPrices", base_url));
+  url = net::AppendQueryParameter(url, "vs_currency",
+                                  base::ToUpperASCII(vs_currency));
+  return url;
 }
```
supermassive, line 226: Should we do something with `GetPriceHistoryURL` below?

components/brave_wallet/browser/asset_ratio_service.cc
```diff
@@ -465,21 +486,17 @@ void AssetRatioService::OnGetStripeBuyURL(GetBuyUrlV1Callback callback,
   std::move(callback).Run(*url, std::nullopt);
 }
 
-void AssetRatioService::OnGetPrice(std::vector<std::string> from_assets,
-                                   std::vector<std::string> to_assets,
-                                   GetPriceCallback callback,
-                                   APIRequestResult api_request_result) {
+void AssetRatioService::OnGetPrice(
+    std::vector<mojom::AssetPriceRequestPtr> requests,
```
supermassive: It seems we don't need to bind requests into this as it is not referenced

components/brave_wallet/common/brave_wallet.mojom
supermassive: Can we always include chain_id?
For example it is up to backend(brave-core) to return 0 for BTC testnet. 
In other words when we request price for a native token we refenece its chain as (coin,chain_id), not just coin. There is acutally already `struct ChainId` to reference chains.

components/brave_wallet_ui/utils/pricing-utils.ts
```diff
@@ -127,3 +128,21 @@ export const getPriceIdForToken = (
...
+export const getPriceRequestForToken = (
+  token: Pick<
+    BraveWallet.BlockchainToken,
+    'coin' | 'chainId' | 'contractAddress'
+  >,
+) => {
+  // Skip price of testnet tokens
+  if (SupportedTestNetworks.includes(token.chainId)) {
+    return
+  }
```
supermassive: I'd rather move that to brave-core

components/brave_wallet_ui/utils/pricing-utils.ts
```diff
@@ -127,3 +128,21 @@ export const getPriceIdForToken = (
...
+
+  return {
+    coinType: token.coin,
+    chainId: token.chainId,
+    address: token.contractAddress || undefined,
+  } as BraveWallet.AssetPriceRequest
```
supermassive: `as` migh result in bugs if `AssetPriceRequest` gets some refactoring.
either `saitsifies BraveWallet.AssetPriceRequest` or just ~`as BraveWallet.AssetPriceRequest`~

components/brave_wallet_ui/utils/pricing-utils.ts
```diff
@@ -16,53 +15,61 @@ import {
...
+    [BraveWallet.CoinType.SOL, BraveWallet.CoinType.ETH].includes(token.coin)
+  ) {
+    return registry.find(
+      (p) =>
+        p.coinType === token.coin
+        && p.chainId === token.chainId
+        && p.address.toLowerCase() === token.contractAddress?.toLowerCase(),
+    )
+  }
+  return registry.find((p) => p.coinType === token.coin)
```
supermassive: this should also check for chain_id equality

supermassive: There is also is_shielded for ZEC which is a sort of a separate native token with seprate balance.
cc: @cypt4

ios/brave-ios/Sources/BraveWallet/Extensions/AssetRatioServiceExtensions.swift
```diff
@@ -7,140 +7,59 @@ import BraveCore
...
-    completion: @escaping (PricesResult) -> Void
-  ) {
-    let uniquePriceIds = uniqueNonEmptyPriceIds(fromAssets)
+extension Array where Element == BraveWallet.AssetPrice {
+  /// Get token price from a list of asset prices
+  /// Returns the matching AssetPrice if found, nil otherwise
+  func getTokenPrice(
+    for token: BraveWallet.BlockchainToken
+  ) -> BraveWallet.AssetPrice? {
+    guard !isEmpty else { return nil }
```
nuo-xu: guard !isEmpty is unnecessary here since `first` will return an optional element if the array is empty

ios/brave-ios/Sources/BraveWallet/Extensions/AssetRatioServiceExtensions.swift
```diff
@@ -7,140 +7,59 @@ import BraveCore
...
-    fromAssets.forEach { asset in
-      dispatchGroup.enter()
-      self.price(fromAssets: [asset], toAssets: toAssets, timeframe: timeframe) { success, prices in
-        defer { dispatchGroup.leave() }
-        pricesResults.append(PricesResult(prices, success ? 0 : 1))
+  /// Updates the existing array with new AssetPrice objects, ensuring no duplicates
+  /// Two AssetPrices are considered duplicates if they have the same coinType, chainId, and address
+  mutating func update(with newPrices: [BraveWallet.AssetPrice]) {
+    for newPrice in newPrices {
+      // Find index of existing price with same coinType, chainId, and address
```
nuo-xu: do we not need to compare differently based on coin type like the `getTokenPrice`? i think we can update `getTokenPrice` to take `coinType`, `chainId` and `address` so that this function can use as well?

ios/brave-ios/Sources/BraveWallet/Extensions/AssetRatioServiceExtensions.swift
```diff
@@ -7,140 +7,59 @@ import BraveCore
...
-  @MainActor func priceWithIndividualRetry(
-    _ fromAssets: [String],
-    toAssets: [String],
-    timeframe: BraveWallet.AssetPriceTimeframe
-  ) async -> PricesResult {
-    let uniquePriceIds = uniqueNonEmptyPriceIds(fromAssets)
+  /// Fetches the prices for a given list of `AssetPriceRequest`, giving a list of `AssetPrice`.
+  /// Returns an empty array in case no prices are found.
+  @MainActor func fetchPrices(
+    for requests: [BraveWallet.AssetPriceRequest],
```
nuo-xu: i think it would be easier to make function take a list of tokens and put the map that transform token to BraveWallet.AssetPriceRequest here so that you don't need to map everywhere when you call this function. because most of the time we have a list of token in hand.

components/brave_wallet/common/brave_wallet.mojom
```diff
@@ -474,11 +474,31 @@ struct AssetTimePrice {
...
+  AssetPriceSource source;
+  string percentage_change_24h;
```
nuo-xu, line 505: do we only have the 24h change? users usually have the ability to change the time in one the choice in `BraveWallet.AssetPriceTimeFrame`

ios/brave-ios/Sources/BraveWallet/Crypto/Stores/PortfolioStore.swift
```diff
@@ -594,21 +594,24 @@ public class PortfolioStore: ObservableObject, WalletObserverStore {
...
-      }
+
+      self.pricesCache.update(with: prices)
 
       // fetch price history for every non-zero balance token
       let nonZeroBalanceAssetRatioIds: [String] =
         allTokens
         .filter { (tokenBalancesCache[$0.id] ?? [:]).values.reduce(0, +) > 0 }
-        .map { $0.assetRatioId }
+        .map { $0.assetRatioId.lowercased() }
```
nuo-xu, line 608: `assetRatioId` is a computed value we created initially to pick from `coingeckId`, `symbol` and `contractAddress`. does this still stay valid with this PR's change?

components/brave_wallet/common/brave_wallet.mojom
onyb: The Ratios service always returned the 24H percentage change regardless of what you provided as input params. This behaviour has been retained in the Gate3 service as well. I therefore decided to use `percentage_change_24h` for correctness.

This is not to be confused with the price history endpoint, which respects the `BraveWallet.AssetPriceTimeFrame` passed an input.

ios/brave-ios/Sources/BraveWallet/Extensions/AssetRatioServiceExtensions.swift
onyb: Good idea! Addressed here - [0a1a472](https://github.com/brave/brave-core/pull/31054/commits/0a1a472988e41ddbe533799542227df80b19b3cc)

ios/brave-ios/Sources/BraveWallet/Extensions/AssetRatioServiceExtensions.swift
onyb: Fixed here: [78aeb9b](https://github.com/brave/brave-core/pull/31054/commits/78aeb9bf00d140f441c11e2c176736bb3a9e5a3e)

components/brave_wallet_ui/utils/pricing-utils.ts
onyb: Fixed here: [dd1071a](https://github.com/brave/brave-core/pull/31054/commits/dd1071abe2f161abe14e6c9474f9bb815ed246f0)

components/brave_wallet/browser/asset_ratio_service.cc
onyb: [2169230](https://github.com/brave/brave-core/pull/31054/commits/21692308b96fa29d071152648b22504120855a8c)

components/brave_wallet/browser/asset_ratio_response_parser.cc
onyb: [2169230](https://github.com/brave/brave-core/pull/31054/commits/21692308b96fa29d071152648b22504120855a8c)

components/brave_wallet/browser/asset_ratio_response_parser.cc
onyb: Switched to IDL parsing: [15e92b2](https://github.com/brave/brave-core/pull/31054/commits/15e92b2b5a9206ccb22d1fc92b18f84e81129b32)

android/java/org/chromium/chrome/browser/crypto_wallet/util/AssetsPricesHelper.java
onyb: Trying not to change too many files on Android in this PR. Here's an issue to clean up all the unused code, including `AssetPricesHelper.java`: https://github.com/brave/brave-browser/issues/37398

components/brave_wallet/browser/asset_ratio_service.cc
onyb: Price history support is not available on gate3 yet. For the time being, we'll continue using Ratios for this.

components/brave_wallet/common/brave_wallet.mojom
onyb: Agree that we need this, but highlighting a small concern here - this will introduce some level of coupling between gate3 (pricing service) and brave-core. For example, `kFilecoinMainnet` and `kFilecoinTestnet` are defined as `f` and `t` respectively, for historical reasons, but coins like bitcoin follow a different convention.

We should standardise the chain ids since we'll now be referencing them outside brave-core.

supermassive: `Coupling between gate3 (pricing service) and brave-core` is actually the only right place where that should be implemented. Wallet's frontend should not know that there is gate3 or coingecko or else being a prices source(in terms of calling interfaces, that's ok to have price source for logging). For wallet's frontend(/brave/components/brave_wallet_ui) there is only wallet's backend(/brave/components/brave_wallet) interfaced by what brave_wallet.mojom provides.

components/brave_wallet_ui/common/slices/endpoints/pricing.endpoints.ts
```diff
@@ -96,80 +91,37 @@ export const pricingEndpoints = ({
...
+            async function (requests: BraveWallet.AssetPriceRequest[]) {
               const { success, values } = await assetRatioService.getPrice(
-                params,
-                [toCurrency],
-                timeframe ?? BraveWallet.AssetPriceTimeframe.Live,
+                requests,
+                vsCurrency,
               )
 
               if (success && values) {
```
supermassive: Do we still need batching? 
`AssetRatioService::GetPrice` does one http request. What is the limit of its json size?

components/brave_wallet/browser/asset_ratio_service.cc
onyb: 6ea8736058892b5c692a1b63b2a9704a7c58d784

components/brave_wallet/common/brave_wallet.mojom
onyb: Commit that includes `chain_id`: 07a8c23f77c3391b037cc4a270bb5c7e592f94ec

components/brave_wallet/browser/asset_ratio_service.cc
onyb: Skipping such items with a `continue`: 07a8c23f77c3391b037cc4a270bb5c7e592f94ec

components/brave_wallet_ui/utils/pricing-utils.ts
onyb: Now that we're passing chain id to gate3, this block no longer required (skipped within gate3).

07a8c23f77c3391b037cc4a270bb5c7e592f94ec

components/brave_wallet_ui/utils/pricing-utils.ts
onyb: 07a8c23f77c3391b037cc4a270bb5c7e592f94ec

onyb: Does ZEC pricing depend on `is_shielded`?

ios/brave-ios/Sources/BraveWallet/Extensions/AssetRatioServiceExtensions.swift
onyb: We no longer use this distinction for ETH and SOL coins, since chain_id is also required by gate3.

Additionally, I've refactored the comparison logic here for code-reuse: [e85d209](https://github.com/brave/brave-core/pull/31054/commits/e85d209e841d12ae370583de0e1765f7d2bc5c40)

ios/brave-ios/Sources/BraveWallet/Crypto/Stores/PortfolioStore.swift
onyb: I think it's still valid. This piece of code is for price history, which still relies on the Ratios service and coingecko IDs for lookups.

components/brave_wallet_ui/common/slices/endpoints/pricing.endpoints.ts
onyb: Let me share some more information.

Pricing lookups are resolved in a single http request from brave-core. Coingecko supports querying prices for 100 tokens in one shot (magic number based on reasonable amount of testing). Gate3 supports a concurrency of up to 4 for the pricing endpoint.

In other words, each gate3 request can concurrently resolve prices of 400 unique assets. This number is quite high even for wallets with large portfolios. So we can skip batching.

There's no limit on payload size in fastapi (gate3's api framework). If the reverse proxy has any limits, I'm sure it'll be more than enough to accommodate even the largest possible POST body that our wallet might ever compose. Practically speaking, we're unlikely to hit any such limit.

onyb: Removing batching logic as discussed on Slack: [3a750df](https://github.com/brave/brave-core/pull/31054/commits/3a750dfb9a8d403967358730b4d9db8c8f2da612)

android/java/org/chromium/chrome/browser/crypto_wallet/util/AssetsPricesHelper.java
```diff
@@ -23,56 +18,8 @@ public static void fetchPrices(
             AssetRatioService assetRatioService,
             BlockchainToken[] assets,
             Callbacks.Callback1<HashMap<String, Double>> callback) {
+        // Method is not used - return empty HashMap to maintain compatibility
```
SergeyZhukovsky: do you know what do we show in wallet panel in that code path https://github.com/brave/brave-core/blob/master/android/java/org/chromium/chrome/browser/crypto_wallet/modal/BraveWalletPanel.java#L137? And also in that place https://github.com/brave/brave-core/blob/master/android/java/org/chromium/chrome/browser/crypto_wallet/util/Utils.java#L1118. It's for tx approval dialog.

PR #30701, opened 2025-08-19
[SideBySide] Updated split view menu entries at tab context menu

simonhong:
Resolves https://github.com/brave/brave-browser/issues/47804

* Prevent showing "new" badge from split menu entries by making `IsNewFeatureAt()` return `false`.
* Updated split view menu entries' icon/string by subclassing `SplitTabMenuModel`.
* Added `//brave/browser/ui/tabs:tab_model` and `//brave/browser/ui/tabs::tab_model_impl` targets.

Screenshot 2025-08-27 191820
Screenshot 2025-08-27 191833
Screenshot 2025-08-27 191842

TEST=`BraveTabContextMenuWithSideBySideTest.SplitViewMenuCustomizationTest`

PR #31055, opened 2025-09-05
Use BraveSplitTabMenuModel as SplitTabsToolbarButton's menu model

simonhong:
Resolves https://github.com/brave/brave-browser/issues/49021

Replaced `SplitTabsToolbarButton`'s menu model with ours(`BraveSplitTabMenuModel`) to have our menu string and icons.

TEST=BraveToolbarViewTest.SplitTabsToolbarButtonTest

image


mkarolin (approved):
LGTM

browser/ui/views/toolbar/brave_toolbar_view_browsertest.cc
```diff
@@ -338,6 +342,24 @@ IN_PROC_BROWSER_TEST_F(BraveToolbarViewTest,
   EXPECT_EQ(false, is_avatar_button_shown());
 }
 
+IN_PROC_BROWSER_TEST_F(BraveToolbarViewTest, SplitTabsToolbarButtonTest) {
+  SplitTabsToolbarButton* split_tabs_toolbar_button =
+      toolbar_view_->split_tabs_toolbar_button_for_testing();
+  ASSERT_TRUE(split_tabs_toolbar_button);
+
+  // Check our menu model is used for split button.
+  auto* menu_model = static_cast<BraveSplitTabMenuModel*>(
```
bridiver, line 352: this really isn't a great way to test this and we should avoid `for_testing` methods in general. I'm not sure how I feel about this entire approach of `SetMenuModel` so looking at this a little more closely

browser/ui/views/toolbar/brave_toolbar_view.cc
```diff
@@ -166,6 +167,12 @@ void BraveToolbarView::Init() {
     return;
   }
 
+  if (split_tabs_) {
+    split_tabs_->SetMenuModel(std::make_unique<BraveSplitTabMenuModel>(
```
bridiver: this code is problematic as I mentioned, but I don't understand what we're doing here since we're already replacing SplitTabMenuModel with BraveSplitTabMenuModel here https://github.com/brave/brave-core/pull/30701/files#r2334484736

bridiver: it looks like there is quite a bit of copied code like `BraveBrowserTabStripController`, `BraveTabContextMenuContents`, etc... unless we are completely changing the upstream implementation and don't care about any of their changes at all, we should not be copying code. I think we need to take a step back here and look at the existing changes because it seems like we're just adding more problems here instead of just replacing `SplitTabMenuModel` directly

browser/ui/views/toolbar/brave_toolbar_view_browsertest.cc
simonhong: removed `_for_testing()` method.

browser/ui/views/toolbar/brave_toolbar_view.cc
simonhong: Directly replaced via overriding file.

chromium_src/chrome/browser/ui/views/toolbar/split_tabs_button.cc
```diff
@@ -0,0 +1,14 @@
+/* Copyright (c) 2025 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "chrome/browser/ui/views/toolbar/split_tabs_button.h"
+
+#include "brave/browser/ui/tabs/brave_split_tab_menu_model.h"
+
+#define SplitTabMenuModel BraveSplitTabMenuModel
```
bridiver: use `CreateBraveSplitTabMenuModel()` with forward decl so we don't have to add a dependency in `chrome/browser/ui/views/toolbar`

browser/ui/views/toolbar/sources.gni
```diff
@@ -0,0 +1,6 @@
+# Copyright (c) 2025 The Brave Authors. All rights reserved.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+brave_browser_ui_views_toolbar_deps = [ "//brave/browser/ui/tabs:tab_model" ]
```
bridiver: per https://github.com/brave/brave-core/pull/31055/files#r2353615059 this should no longer be needed

chromium_src/chrome/browser/ui/views/toolbar/split_tabs_button.cc
simonhong: Done.

chromium_src/chrome/browser/ui/views/toolbar/split_tabs_button.cc
```diff
@@ -0,0 +1,30 @@
...
+std::unique_ptr<ui::SimpleMenuModel> CreateBraveSplitTabMenuModel(
+    TabStripModel* tab_strip_model,
+    SplitTabMenuModel::MenuSource source);
+
+SplitTabsToolbarButton::SplitTabsToolbarButton(Browser* browser)
+    : SplitTabsToolbarButton_ChromiumImpl(browser) {
+  split_tab_menu_ = CreateBraveSplitTabMenuModel(
+      browser_->tab_strip_model(),
+      SplitTabMenuModel::MenuSource::kToolbarButton);
+}
```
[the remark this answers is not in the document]
simonhong: Base (upstream) class uses `Browser *` as a param.

kjozwiak:

```
Brave | 1.84.93 Chromium: 141.0.7390.30 (Official Build) nightly (64-bit)
-- | --
Revision | 5c2b6ff0f689c574ae2db8fd18e05139c3154c21
```

Using the STR/Cases and information via https://github.com/brave/brave-browser/issues/49021#issue-3389089429, ensured that the correct strings/order and icons are being used as per the following:

**`Light`** | **`Dark`**
------------|--------------
image | image

The actual icon/menu verification were completed via https://github.com/brave/brave-core/pull/31050#issuecomment-3330292603. The above is just checking the strings/icons and order of the context menu.

PR #31058, opened 2025-09-05
[NTP Next] Allow sanitized images to be lazy loaded

zenparsing:
Resolves https://github.com/brave/brave-browser/issues/49005

### DIFF
--- a/browser/resources/brave_new_tab_page_refresh/components/common/safe_image.tsx
+++ b/browser/resources/brave_new_tab_page_refresh/components/common/safe_image.tsx
@@ -5,7 +5,7 @@
 
 import * as React from 'react'
 
-import { loadImage, placeholderImageSrc } from '../../lib/image_loader'
+import { placeholderImageSrc } from '../../lib/image_loader'
 
 interface Props {
   src: string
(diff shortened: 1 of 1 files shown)


browser/resources/brave_new_tab_page_refresh/components/common/safe_image.tsx
```diff
@@ -5,22 +5,20 @@
...
-    const url = 'chrome://image?url=' + encodeURIComponent(props.src || '')
-    loadImage(url).then((loaded) => {
-      setImageURL(loaded ? url : placeholderImageSrc)
-    })
-  }, [props.src])
-
-  return <img src={imageURL} loading='lazy' className={props.className} />
+  return (
+    <img
+      src={'chrome://image?url=' + encodeURIComponent(props.src || '')}
```
fallaciousreasoning: can we move the `|| ''` outside the `chrome://` url so we don't make unnecessary requests?


```suggestion
      src={props.src ? 'chrome://image?url=' + encodeURIComponent(props.src) : ''}
```

fallaciousreasoning (approved):

browser/resources/brave_new_tab_page_refresh/components/common/safe_image.tsx
zenparsing: Done.

PR #31065, opened 2025-09-05
[Android] Changes condition for PIP script injection

samartnik:
Resolves https://github.com/brave/brave-browser/issues/48682
Resolves https://github.com/brave/brave-browser/issues/49075

### DIFF
--- a/browser/android/youtube_script_injector/youtube_script_injector_tab_helper.cc
+++ b/browser/android/youtube_script_injector/youtube_script_injector_tab_helper.cc
@@ -221,7 +221,7 @@ void YouTubeScriptInjectorTabHelper::PrimaryMainDocumentElementAvailable() {
   SetFullscreenRequested(false);
   content::WebContents* contents = web_contents();
   // Filter only YouTube videos.
-  if (!IsYouTubeVideo()) {
+  if (!IsYouTubeDomain()) {
     return;
   }
   content::RenderFrameHost::AllowInjectingJavaScript();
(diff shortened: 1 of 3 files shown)


kjozwiak:
Verification **PASSED** on `Pixel 6` running `Android 16` using the following build(s):

```
Brave | 1.84.39 Chromium: 140.0.7339.80 (Official Build) canary (64-bit)
--- | ---
Revision | 4e2c70fc0c2eb18349970a2741dc7d375e877079
OS | Android 16; Build/BP2A.250605.031.A5; 36; REL
```

### `Test Case #1` - https://github.com/brave/brave-browser/issues/48682

Using `1.84.38 Chromium: 140.0.7339.80`, reproduced the original issue mentioned via https://github.com/brave/brave-browser/issues/48682#issue-3353003518 where the `PiP` window would automatically close which appears as it crashed.

Using the same process/STR cases mentioned above, ensured that the issue wasn't happening anymore while using `1.84.39 Chromium: 140.0.7339.80` as per the following:

Video of verification re: `PiP` working with the fix --> https://youtube.com/shorts/JEgYZElRfoA

### `Test Case #2` - https://github.com/brave/brave-browser/issues/49075

Using `1.84.38 Chromium: 140.0.7339.80`, reproduced the original issue mentioned via https://github.com/brave/brave-browser/issues/49075#issue-3395036963 where `Playback` stopped working when users either sleep/lock the device or switch to another tab.

Using the same process/STR cases mentioned above, ensured that `Playback` worked when the device is locked or when a user switches to another application as per the following:

* On `Nightly`, ensured that `Brave Experimental Adblock Rules` was disabled via `Content Filtering`
* visited YT and played a random video
* ensured that `playback` was still working once switching to another tab
* ensured that `playback` was still working when locking/sleeping the device
* ensured that you can `pause` the video via the `notification tray` without any issues

https://github.com/user-attachments/assets/2f7e1e46-acd9-441b-935a-44bedcd58333

PR #29750, opened 2025-06-24
Enable a proxy when WireGuard is enabled on Windows

bsclifton:
For more information on how the proxy strategy used in this PR works please see: https://learn.microsoft.com/en-us/archive/blogs/askie/how-can-i-configure-proxy-autoconfigurl-setting-using-group-policy-preference-gpp

Fixes https://github.com/brave/brave-browser/issues/43749

Brave employees can see more information in the `#code-reviewers` Slack Channel here:
https://bravesoftware.slack.com/archives/C02HW17JAR5/p1750201069888559

PR #31066, opened 2025-09-05
Cleanup to fix crash that happened on dereference

bsclifton:
Fixes https://github.com/brave/brave-browser/issues/49003

May also fix https://github.com/brave/brave-browser/issues/48361 (due to undefined behavior)

Need to get binary and install/run locally to verify. Unintentionally introduced crash in
https://github.com/brave/brave-core/pull/29750 when implementing review feedback ☹️

### DIFF
--- a/browser/brave_vpn/win/wireguard_utils_win.cc
+++ b/browser/brave_vpn/win/wireguard_utils_win.cc
@@ -48,13 +48,16 @@ ServiceRegistrationStatus g_wireguard_service_registered_for_testing;
 // survive the owning process being restarted.
 // TODO(https://github.com/brave/brave-browser/issues/47115): remove this value
 // and pass in the real value each time.
-base::NoDestructor<std::string> g_smart_proxy_url("");
+std::string& GetSmartProxyUrlInstance() {
+  static base::NoDestructor<std::string> instance("");
+  return *instance;
+}
 constexpr wchar_t kSystemProxyRegistryKey[] =
     L"Software\\Microsoft\\Windows\\CurrentVersion\\Internet Settings";
 constexpr wchar_t kAutoConfigURL[] = L"AutoConfigURL";
 
 void MaybeEnableSystemProxy() {
-  if (g_smart_proxy_url->empty()) {
+  if (GetSmartProxyUrlInstance().empty()) {
     return;
   }
 
@@ -70,7 +73,7 @@ void MaybeEnableSystemProxy() {
               << "\".\nBrave VPN smart proxy routing can't be used.";
     } else {
       // Set the system proxy URL.
-      current_proxy_url = base::UTF8ToWide(*g_smart_proxy_url);
+      current_proxy_url = base::UTF8ToWide(GetSmartProxyUrlInstance());
       rv = key.WriteValue(kAutoConfigURL, current_proxy_url.c_str());
       VLOG(1) << L"Set Brave VPN smart proxy routing to = \""
               << current_proxy_url << "\"";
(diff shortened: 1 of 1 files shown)


Olipro (approved):
Since the scope of what used to be `g_smart_proxy_url` is local to the TU, the only change I'd suggest is one of:

1. Make `GetSmartProxyUrlInstance()` a `static` function since it has no reason to have external linkage - though I doubt anyone in the future would cause an ODR violation by managing to create a function with the same name elsewhere.
2. Move the variable into an anonymous namespace or similar and access it that way.

cdesouza-chromium:
> Since the scope of what used to be `g_smart_proxy_url` is local to the TU, the only change I'd suggest is one of:
> 
> 1. Make `GetSmartProxyUrlInstance()` a `static` function since it has no reason to have external linkage - though I doubt anyone in the future would cause an ODR violation by managing to create a function with the same name elsewhere.
> 2. Move the variable into an anonymous namespace or similar and access it that way.

`GetSmartProxyUrlInstance` is already in an anonymous namespace, so we don't mark that as static as being in the unnamed namespace gives it internal linkage.

browser/brave_vpn/win/wireguard_utils_win.cc
```diff
@@ -48,13 +48,16 @@ ServiceRegistrationStatus g_wireguard_service_registered_for_testing;
...
+  static base::NoDestructor<std::string> instance("");
+  return *instance;
+}
 constexpr wchar_t kSystemProxyRegistryKey[] =
     L"Software\\Microsoft\\Windows\\CurrentVersion\\Internet Settings";
 constexpr wchar_t kAutoConfigURL[] = L"AutoConfigURL";
 
 void MaybeEnableSystemProxy() {
-  if (g_smart_proxy_url->empty()) {
+  if (GetSmartProxyUrlInstance().empty()) {
```
cdesouza-chromium, line 60: This looks a bit ambiguous if you think about it, but I'm not sure if we should care. Basically `smart_proxy_url` is a `std::optional<std::string>`, but here we treat `""` as a sentinel to mean not set. Maybe this is not a meaningful difference...

Olipro:
> `GetSmartProxyUrlInstance` is already in an anonymous namespace, so we don't mark that as static as being in the unnamed namespace gives it internal linkage.

Yep, all good, that's what I get for looking at the diff and not the whole file.

bsclifton:
Pulled down the binary- installed via `brave_installer_140_1_83_0.exe --system-level` and logged into account.brave.com to load credentials.

Verified this works great! 😄 Tested both code paths with WireGuard specifically.
image

Switching servers is good, etc.

MadhaviSeelam:
Verification `PASSED` using 
```
Brave | 1.84.36 Chromium: 140.0.7339.80 (Official Build) nightly (64-bit)
-- | --
Revision | 9d7ca1afc5d61bc0c1026eb0a35e964c68e7f42b
OS | Windows 11 Version 24H2 (Build 26100.4946)
```

1. Installed 1.84.36 (with admin privileges)
2. launched Brave
3. verified `Use WireGuard protocol in Brave VPN` toggle is not enabled
4. logged into account.brave.com with existing credentials
5. refreshed credentials
6. confirmed `Use WireGuard protocol in Brave VPN` toggle is enabled
7. enabled VPN via VPN panel 
8. confirmed VPN connected successfully to default region (USA)
9. switched to different region while Smart proxy enabled (eg. Poland)
10. successfully connected to different region without any Brave crash
11. switched back to USA and disabled Smart proxy and successfully connected
12. switched to different region and connected to Switzerland

Confirmed no Brave crashes and VPN button didn't freeze


https://github.com/user-attachments/assets/5a8a4647-1d7f-4a46-b3f3-c619b818c31b

PR #31073, opened 2025-09-07
Remove top sites suggestions provider

bbondy:
The list was not updated nor maintained, and even had invalid entries.

On device which uses a local ML model was tied to this preference for some reason too, so that keeps the preference around and has a checkbox for just that now.

Resolves https://github.com/brave/brave-browser/issues/48802


bbondy:
Note for the Android reviewer: I don't currently have an Android env setup, so I haven't tested the removal on Android. Please help if you can if you think that's needed.

fallaciousreasoning:
I think there are a bunch of people asking to be able to toggle top sites and the zero suggestions separately, so this should give them what they want too.

@rebron do we want to rename the toggle now?

@bbondy I think we did this so we could avoid having even more toggles in settings, and people wanted to be able to toggle them off.

fallaciousreasoning (approved):
lgtm - always nice to remove a bunch of code!

bsclifton (approved):
string reviews ++, overall change for Desktop side LGTM also 👍

SergeyZhukovsky:
> Note for the Android reviewer: I don't currently have an Android env setup, so I haven't tested the removal on Android. Please help if you can if you think that's needed.

Screenshot_20250908_101138

PR #31077, opened 2025-09-07
Add new POLICY_SOURCE_BRAVE_ORIGIN policy source

bbondy:
Resolves https://github.com/brave/brave-browser/issues/49042


chromium_src/components/policy/core/browser/DEPS
```diff
@@ -0,0 +1,6 @@
+from brave_chromium_utils import inline_file
+inline_file("//components/policy/core/browser/DEPS", globals(), locals())
+
+include_rules = [
+  "+components/grit/brave_components_strings.h",
```
bbondy, line 5: wondering if you have a way to not need this. Maybe moving the string somewhere else?

chromium_src/components/enterprise/browser/reporting/policy_info.cc
```diff
@@ -0,0 +1,15 @@
+/* Copyright (c) 2025 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "components/policy/core/common/policy_types.h"
+
+#define POLICY_SOURCE_MERGED                                \
+  POLICY_SOURCE_BRAVE_ORIGIN:                               \
+  return em::Policy_PolicySource_SOURCE_ENTERPRISE_DEFAULT; \
```
bridiver: shouldn't this be POLICY_SOURCE_PLATFORM?  https://source.chromium.org/chromium/chromium/src/+/main:components/policy/core/common/policy_types.h;l=60;drc=4128411589187a396829a827f59a655bed876aa7;bpv=1;bpt=1

chromium_src/components/policy/core/browser/DEPS
```diff
@@ -0,0 +1,6 @@
...
+]
```
bridiver: please configure git/editor to always add eol

chromium_src/components/policy/core/browser/DEPS
bridiver: no, this is fine, we already add brave_components_strings as a dep of components/strings

chromium_src/components/policy/core/common/policy_types.h
```diff
@@ -0,0 +1,22 @@
...
+#ifndef BRAVE_CHROMIUM_SRC_COMPONENTS_POLICY_CORE_COMMON_POLICY_TYPES_H_
+#define BRAVE_CHROMIUM_SRC_COMPONENTS_POLICY_CORE_COMMON_POLICY_TYPES_H_
+
+// Inject a new POLICY_SOURCE_BRAVE_ORIGIN at the end so it doesn't affect other
+// integer values which could be stored in logs or sent via network
+// serialization
+#define POLICY_SOURCE_COUNT POLICY_SOURCE_BRAVE_ORIGIN, POLICY_SOURCE_COUNT
+
+// Define a new kBraveOriginPriority to be used as the new lowest priority
+#define kEnterpriseDefault kBraveOriginPriority, kEnterpriseDefault
```
bridiver: do we need a new priority for this? Shouldn't it be either kPlatformUser or kEnterpriseDefault depending on this? https://github.com/brave/brave-core/pull/31077/files#r2331303165

bbondy: Yes because we want it lower than enterprise default.

chromium_src/components/enterprise/browser/reporting/policy_info.cc
bbondy: We could put origin here but it would require patching a proto file and I don't think it matters for this code. I think the closest value to ORIGIN which is lower than ENTERPRISE default by 1 is ENTERPRISE_DEFAULT so I think it should stay that value.

chromium_src/components/policy/core/browser/DEPS
bbondy: fixed and pushed, it's when you use AI to edit things in Claude Code or Cursor it doesn't put them. I tried to add a precommit hook to detect this. We should probably add this to presubmit though.

chromium_src/components/policy/core/common/policy_types.h
```diff
@@ -0,0 +1,22 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#ifndef BRAVE_CHROMIUM_SRC_COMPONENTS_POLICY_CORE_COMMON_POLICY_TYPES_H_
+#define BRAVE_CHROMIUM_SRC_COMPONENTS_POLICY_CORE_COMMON_POLICY_TYPES_H_
+
+// Inject a new POLICY_SOURCE_BRAVE_ORIGIN at the end so it doesn't affect other
+// integer values which could be stored in logs or sent via network
+// serialization
+#define POLICY_SOURCE_COUNT POLICY_SOURCE_BRAVE_ORIGIN, POLICY_SOURCE_COUNT
```
bridiver: couldn't this potentially overlap with some future policy that is added here? Is it persisted anywhere? Also any issue that there's no corresponding value in device_management_backend.proto? Maybe we should actually map it to 0 so it maps to `SOURCE_UNKNOWN` in the proto file and never overlaps with anything?

chromium_src/components/policy/core/browser/policy_conversions.cc
```diff
@@ -0,0 +1,26 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "components/grit/brave_components_strings.h"
+#include "components/policy/core/common/policy_types.h"
+#include "components/strings/grit/components_strings.h"
+
+// Capture the original number
+enum {
+  kBravePolicySourceRestrictedManagedGuest =
```
bridiver: enum values should be all caps with underscores, surprised format doesn't complain about this?

bridiver: actually why an enum for this at all? `constexpr int`?

chromium_src/components/enterprise/browser/reporting/policy_info.cc
bridiver: per below maybe this should actually be `em::Policy_PolicySource_SOURCE_UNKNOWN` for the proto file

chromium_src/components/policy/core/common/policy_types.h
bbondy: I added a comment about why this value (10) is used and added a static assert so we know when it changes. It seems safer than using -1 and an offset like +10 which can causing indexing problems into arrays initialized with the policy count as the size. Or potentially into other arrays like string arrays which have COUNT items.

chromium_src/components/policy/core/browser/policy_conversions.cc
bbondy: no reason, updated to `constexpr int`

chromium_src/components/enterprise/browser/reporting/policy_info.cc
bbondy: OK updated to SOURCE_UNKNOWN.

PR #30947, opened 2025-08-31
Use result<> in email_aliases.mojom

arthuredelstein:
Resolves https://github.com/brave/brave-browser/issues/48515

PR #30668, opened 2025-08-15
Brave Account registration flow.

szilardszaloki:
Resolves https://github.com/brave/brave-browser/issues/47400.

**Hey all — just a heads up: this PR currently includes only the production changes. I'll add the tests in a separate commit early next week, but didn't want to block reviews on them.**

PR #31088, opened 2025-09-08
Brave Account registration flow (frontend changes)

szilardszaloki:
Resolves https://github.com/brave/brave-browser/issues/49058.


components/brave_account/brave_account_ui_base.h
```diff
@@ -32,24 +34,42 @@
...
+    auto* source = WebUIDataSource::CreateAndAdd(profile_, kBraveAccountHost);
     std::move(setup_webui_data_source)
         .Run(source, kBraveAccountResources,
              IDR_BRAVE_ACCOUNT_BRAVE_ACCOUNT_PAGE_HTML);
     SetupWebUIDataSource(source);
   }
 
+  void BindInterface(mojo::PendingReceiver<brave_account::mojom::Authentication>
+                         pending_receiver) {
+    auto* brave_account_service = BraveAccountServiceFactory::GetFor(profile_);
```
fallaciousreasoning: Hey so is there a reason we can't just pass the service into the constructor, rather than making this generic on the factory? We're trying to move away from looking everything up from a profile in these leaf classes

components/brave_account/brave_account_ui_base.h
```diff
@@ -32,24 +34,42 @@
...
-          setup_webui_data_source = base::DoNothing()) {
+      Profile* profile,
+      base::OnceCallback<void(WebUIDataSource*,
+                              base::span<const webui::ResourcePath>,
+                              int)> setup_webui_data_source = base::DoNothing())
+      : profile_(profile) {
     CHECK(brave_account::features::IsBraveAccountEnabled());
 
-    auto* source = WebUIDataSource::CreateAndAdd(profile, kBraveAccountHost);
+    auto* source = WebUIDataSource::CreateAndAdd(profile_, kBraveAccountHost);
```
fallaciousreasoning: similarly, `CreateAndAdd` can take a `BrowserContext` - does that not exist on iOS? If it does, we could get rid of the generic `Profile` param too

components/brave_account/mojom/brave_account.mojom
```diff
@@ -0,0 +1,36 @@
...
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+module brave_account.mojom;
+
+struct RegisterInitializeResult {
+  string encryptedVerificationToken;
+  string serializedResponse;
+};
+
+struct RegisterFinalizeResult {};
```
fallaciousreasoning: is this going to have more on it later? If not, I think @arthuredelstein used https://github.com/brave/brave-core/pull/30947/files#diff-9317c7beb83ead693e7d1e4b8d94d29d82c3ecdbee264da00adcd4c32f017ed3R55 `mojom.Empty` instead for a void returning promise

components/brave_account/mojom/brave_account.mojom
```diff
@@ -0,0 +1,36 @@
...
+  kInitializeUnknown,
+  kInitializeUnexpected,
+  kFinalizeBadRequest,     // /v2/accounts/password/finalize: 400
+  kFinalizeUnauthorized,   // /v2/accounts/password/finalize: 401
+  kFinalizeForbidden,      // /v2/accounts/password/finalize: 403
+  kFinalizeNotFound,       // /v2/accounts/password/finalize: 404
+  kFinalizeServerError,    // /v2/accounts/password/finalize: 5XX
+  kFinalizeUnknown,
+  kFinalizeUnexpected,
+};
```
fallaciousreasoning, line 32: do we need to know all these different failure modes on the frontend? Maybe it'd be better to just send an error message across if we're just going to use them to look up a string error message anyway

ios/browser/ui/webui/brave_account/BUILD.gn
```diff
@@ -12,11 +12,14 @@ source_set("brave_account") {
   deps = [
     "//base",
     "//brave/components/brave_account",
+    "//brave/components/brave_account/mojom",
```
fallaciousreasoning, line 15: should the `mojom` files be a `public_dep` of `brave_account`?

components/brave_account/brave_account_ui_base.h
fallaciousreasoning: ah, I see its unhelpfully something else. Makes sense.

components/brave_account/resources/brave_account_browser_proxy.ts
fallaciousreasoning, line 1: just curious, but why do we need the `chrome.send` here?

components/brave_account/mojom/brave_account.mojom
szilardszaloki: I had used `mojom.Empty` for that before, but I find spelling it out a bit more consistent.

components/brave_account/mojom/brave_account.mojom
szilardszaloki: The error handling work is already on the way. We won't need to keep all of these long-term, but I'd like to leave them in place until we land the concrete error conditions.

components/brave_account/resources/brave_account_browser_proxy.ts
szilardszaloki: Because of [`ConstrainedWebDialogUI`](https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/ui/webui/constrained_web_dialog_ui.cc;l=73-77):
```cpp
// Add a "dialogClose" callback which matches WebDialogUI behavior.
web_ui()->RegisterMessageCallback(
    "dialogClose",
    base::BindRepeating(&ConstrainedWebDialogUI::OnDialogCloseMessage,
                        base::Unretained(this)));
```

components/brave_account/brave_account_ui_base.h
szilardszaloki: I added `BraveWebUIIOSDataSource::CreateAndAdd()` on iOS a while back. It takes a `web::BrowserState*` (whereas `WebUIDataSource::CreateAndAdd()` takes a `content::BrowserContext*`). Passing the profile works in both cases (because of `ProfileIOS` subclassing `web::BrowserState` and `Profile` subclassing `content::BrowserContext`) and makes the datasource setup uniform.

components/brave_account/resources/brave_account_browser_proxy.ts
fallaciousreasoning: :+1:

components/brave_account/brave_account_ui_base.h
szilardszaloki: We'd need to duplicate this logic:
```cpp
auto* brave_account_service = BraveAccountServiceFactory::GetFor(profile_);
DCHECK(brave_account_service)
    << "BraveAccountService is unexpectedly nullptr - the Brave Account UI "
        "should be launched only from Settings with a non-Guest, "
        "non-System, original (non-OTR) regular profile!";
if (!brave_account_service) {
  return;
}
```
in both
- `brave/ios/browser/ui/webui/brave_account/brave_account_ui.mm` and
- `brave/browser/ui/webui/brave_account/brave_account_ui.cc`,

and then store a reference to the `BraveAccountService` in `BraveAccountUIBase`.
That's why I leaned toward keeping it generic on the factory.

ios/browser/ui/webui/brave_account/BUILD.gn
szilardszaloki: It could be, but other targets that use `//brave/components/brave_account` don't necessarily use `//brave/components/brave_account/mojom` (e.g. `//brave/browser/brave_account`, `//brave/browser/ui/webui/brave_account`, `//brave/ios/browser/brave_account`), so I kept it private.

darkdh (approved):
unblock trivial browser/sources.gni changes

components/brave_account/brave_account_service.h
```diff
@@ -29,9 +34,23 @@ class BraveAccountService : public KeyedService {
 
   ~BraveAccountService() override;
 
+  void BindInterface(
+      mojo::PendingReceiver<mojom::Authentication> pending_receiver);
+
+  void RegisterInitialize(
```
bridiver, line 41: these methods should be private so they can only be called on the mojo interface and not the raw pointer

components/brave_account/brave_account_ui_base.h
```diff
@@ -32,24 +34,42 @@
...
     std::move(setup_webui_data_source)
         .Run(source, kBraveAccountResources,
              IDR_BRAVE_ACCOUNT_BRAVE_ACCOUNT_PAGE_HTML);
     SetupWebUIDataSource(source);
   }
 
+  void BindInterface(mojo::PendingReceiver<brave_account::mojom::Authentication>
+                         pending_receiver) {
+    auto* brave_account_service = BraveAccountServiceFactory::GetFor(profile_);
+    DCHECK(brave_account_service)
```
bridiver: This should be a `CHECK` and it looks like `BraveAccountServiceFactory` is missing ProfileSelections to enforce the correct profile type.

components/brave_account/brave_account_ui_base.h
bridiver: it's actually not possible afaik due to the way webui are constructed

components/brave_account/mojom/brave_account.mojom
```diff
@@ -0,0 +1,36 @@
...
+
+interface Authentication {
+  RegisterInitialize(string email, string blindedMessage)
```
bridiver, line 44: please add comments for the interface/methods

components/brave_account/brave_account_ui_base.h
bridiver: oic, it's the superclass so it could get passed in to the constructor and that code would not have to be duplicated because `if (!brave_account_service) {` isn't relevant per my other comment

components/brave_account/brave_account_service.cc
```diff
@@ -17,4 +19,23 @@ BraveAccountService::BraveAccountService(
...
+void BraveAccountService::BindInterface(
+    mojo::PendingReceiver<mojom::Authentication> pending_receiver) {
+  authentication_receivers_.Add(this, std::move(pending_receiver));
+}
+
+void BraveAccountService::RegisterInitialize(
+    const std::string& email,
+    const std::string& blinded_message,
+    mojom::Authentication::RegisterInitializeCallback callback) {
+  std::move(callback).Run(mojom::RegisterInitializeResult::New());
```
bridiver, line 35: TODO/NOTREACHED

components/brave_account/brave_account_service.cc
```diff
@@ -17,4 +19,23 @@ BraveAccountService::BraveAccountService(
...
+}
+
+void BraveAccountService::RegisterFinalize(
+    const std::string& encrypted_verification_token,
+    const std::string& serialized_record,
+    mojom::Authentication::RegisterFinalizeCallback callback) {
+  std::move(callback).Run(mojom::RegisterFinalizeResult::New());
```
bridiver, line 45: TODO/NOTREACHED

components/brave_account/brave_account_service.h
szilardszaloki: Good catch — they needed to be public because `BraveAccountPageHandler` calls into them in https://github.com/brave/brave-core/pull/30668.

components/brave_account/brave_account_ui_base.h
szilardszaloki: It's intentional — `BraveAccountServiceFactory` uses the default `ProfileSelections`:
- regular: original only, no OTR
- guest: no original, no OTR
- system: no original, no OTR

bridiver: Ok, I was thinking the default allowed more than that

szilardszaloki:
Splitting it further as requested by @bridiver.

PR #31098, opened 2025-09-08
Tor guest profile should have Tor text unless disabled by mandatory policy

bbondy:
Resolves https://github.com/brave/brave-browser/issues/49066

### DIFF
--- a/browser/ui/webui/private_new_tab_page/brave_private_new_tab_page_handler.cc
+++ b/browser/ui/webui/private_new_tab_page/brave_private_new_tab_page_handler.cc
@@ -25,7 +25,9 @@
 
 #if BUILDFLAG(ENABLE_TOR)
 #include "brave/browser/tor/tor_profile_service_factory.h"
+#include "brave/components/tor/pref_names.h"
 #include "brave/components/tor/tor_launcher_factory.h"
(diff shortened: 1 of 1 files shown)


browser/ui/webui/private_new_tab_page/brave_private_new_tab_page_handler.cc
```diff
@@ -98,7 +100,14 @@ void BravePrivateNewTabPageHandler::GetIsTorDisabled(
     GetIsTorDisabledCallback callback) {
   CHECK(profile_);
 #if BUILDFLAG(ENABLE_TOR)
-  bool is_disabled = TorProfileServiceFactory::IsTorDisabled(profile_);
+  // For private new tab page, only check if Tor is explicitly
+  // disabled by enterprise policy (managed and set to true)
+  bool is_disabled = false;
+  if (g_browser_process && g_browser_process->local_state()) {
```
fallaciousreasoning: nit: in other places we don't seem to check whether `g_browser_process` is set, so this could probably be

```suggestion
  if (auto* prefs = g_browser_process->local_state()) {
    auto* pref = prefs->FindPreference(tor::prefs::kTorDisabled);
```

PR #31107, opened 2025-09-09
[perf] Explicitly set --fake-variations-channel

atuchin-m:
Resolves https://github.com/brave/brave-browser/issues/49104

The windows binary doesn't use a static (compiled) channel for variations (Griffin). Instead, it relies on install information (windows registry?).
To fix usage an incorrect channel, the PR explicitly pass `--fake-variations-channel=<channel>` to the browser.
https://source.chromium.org/chromium/chromium/src/+/main:components/variations/variations_switches.cc;l=44?q=fake-variations-channel


bsclifton (approved):
LGTM! This new flag is used for CI when running perf test. 😄👍

PR #31125, opened 2025-09-10
[vpn]: Migrate Brave VPN to the new strings system.

fallaciousreasoning:
Resolves https://github.com/brave/brave-browser/issues/49130

First commit sets up the string bindings and converts a single string. Second commit converts all the other VPN strings over.

Unfortunately I also needed to rename the style imports from `S` to `Styles` so it didn't clash with the strings.

@simonhong do you mind testing this on Windows?


simonhong:
it works well on macOS thanks! 👍🏼

PR #31131, opened 2025-09-10
[PSST] Add user or policy script time out handling

vadimstruts:
Resolves https://github.com/brave/brave-browser/issues/49113

- added handling of the `script times out` failure case for user and policy script execution
- added unittest for the case when user script is failed to to finish running (in that case should be called the Timeout handler instead of user script result handler) 
- updated `README.md` with information about handling the case when the script times out

### DIFF
--- a/components/psst/README.md
+++ b/components/psst/README.md
@@ -29,6 +29,34 @@ https://docs.google.com/document/d/1ccnBWBV_KkknZpZYxcOXTwtfIiaSzeLS5dMd08N2pbs/
 12. Then we enumerate each available URL in the list and do the same as in points #8-11.
 13. When all tasks are completed, the user sees all status and progress information in the consent dialog.
 
+### Workflow failure cases
+
+1. <a id="script-times-out"></a>The script times out or page crashes.</br> This happens when an injected script (either from a user or a policy) never finishes running. Common causes include an infinite loop, a promise that never resolves, or renderer crash can lead to the same situation, when we need to stop the process and notify the front end.
+
+#### Handling Script Timeouts and page crashes
+
+The `base::OneShotTimer timeout_timer_;` is used to handle cases when a script fails to finish running, (i.e. ["The script times out or page crashes"](#script-times-out)).</br>
+Use the next method for script execution as it starts the timer first (with interval: 15 seconds):
+```
+void PsstTabWebContentsObserver::RunWithTimeout(
+    const int last_committed_entry_id,
+    const std::string& script,
+    InsertScriptInPageCallback callback) {
+  timeout_timer_.Start(
+      FROM_HERE, kScriptTimeout,
+      base::BindOnce(&PsstTabWebContentsObserver::OnScriptTimeout,
+                     weak_factory_.GetWeakPtr(), last_committed_entry_id));
+  inject_script_callback_.Run(script, std::move(callback));
+}
+
+```
+Parameters:</br>
+`const int last_committed_entry_id` -  the unique last committed entry ID;</br>
+`const std::string& script` - the script to be injected;</br>
+`InsertScriptInPageCallback callback` - callback function that handles the script result once it executes;
+
+The `PsstTabWebContentsObserver::OnScriptTimeout` is special timeout handler, which stops the other script execution result handlers by calling the `weak_factory_.InvalidateWeakPtrs();`
+
 # PSST CRX Component
 
 Contains set of rules and scripts for small number of very popular sites (Google, Facebook, Twitter, Twitch, etc.). 
--- a/components/psst/browser/content/psst_tab_web_contents_observer.cc
+++ b/components/psst/browser/content/psst_tab_web_contents_observer.cc
@@ -10,8 +10,12 @@
 
 #include "base/debug/crash_logging.h"
 #include "base/debug/dump_without_crashing.h"
+#include "base/functional/bind.h"
 #include "base/json/json_writer.h"
 #include "base/strings/utf_string_conversions.h"
+#include "base/time/time.h"
+#include "base/timer/timer.h"
+#include "base/values.h"
 #include "brave/components/psst/browser/core/psst_rule.h"
 #include "brave/components/psst/browser/core/psst_rule_registry.h"
 #include "brave/components/psst/common/features.h"
@@ -26,6 +30,7 @@ namespace psst {
 
 namespace {
 
+constexpr base::TimeDelta kScriptTimeout = base::Seconds(15);
 const char kShouldProcessKey[] = "should_process_key";
 const char kSignedUserId[] = "user";
 
@@ -157,10 +162,9 @@ void PsstTabWebContentsObserver::InsertUserScript(
   if (!rule || !ShouldInsertScriptForPage(id)) {
     return;
   }
-
   const std::string user_script = rule->user_script();
-  inject_script_callback_.Run(
-      user_script,
+  RunWithTimeout(
+      id, user_script,
       base::BindOnce(&PsstTabWebContentsObserver::OnUserScriptResult,
                      weak_factory_.GetWeakPtr(), id, std::move(rule)));
 }
@@ -169,10 +173,15 @@ void PsstTabWebContentsObserver::OnUserScriptResult(
     int id,
     std::unique_ptr<MatchedRule> rule,
     base::Value user_script_result) {
+  if (!ShouldInsertScriptForPage(id)) {
+    return;
+  }
+
+  timeout_timer_.Stop();
+
   // We should break the flow in case of policy script is not available or user
   // script result is not a dictionary
-  if (!rule || !ShouldInsertScriptForPage(id) ||
-      rule->policy_script().empty() || !user_script_result.is_dict()) {
+  if (!rule || rule->policy_script().empty() || !user_script_result.is_dict()) {
     ui_delegate_->UpdateTasks(100, {}, mojom::PsstStatus::kFailed);
     return;
   }
@@ -185,7 +194,8 @@ void PsstTabWebContentsObserver::OnUserScriptResult(
     return;
   }
 
-  inject_script_callback_.Run(
+  RunWithTimeout(
+      id,
       MaybeAddParamsToScript(std::move(rule),
                              std::move(user_script_result).TakeDict()),
       base::BindOnce(&PsstTabWebContentsObserver::OnPolicyScriptResult,
@@ -199,6 +209,8 @@ void PsstTabWebContentsObserver::OnPolicyScriptResult(
     return;
   }
 
+  timeout_timer_.Stop();
+
   const auto script_result_parsed =
       PolicyScriptResult::FromValue(script_result);
   if (!script_result_parsed) {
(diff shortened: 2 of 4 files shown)


components/psst/browser/content/psst_tab_web_contents_observer.cc
```diff
@@ -199,8 +263,18 @@ void PsstTabWebContentsObserver::OnPolicyScriptResult(
...
 
-  ui_delegate_->UpdateTasks(script_result_parsed->progress,
-                            script_result_parsed->applied_tasks);
+  ui_delegate_->UpdateTasks(
+      script_result_parsed->progress, script_result_parsed->applied_tasks,
+      script_result_parsed->progress == 100 ? PsstStatus::kCompleted
+                                            : PsstStatus::kInProgress);
+}
+
+void PsstTabWebContentsObserver::OnScriptTimeout(int id) {
```
bridiver, line 238: I don't understand why we need a separate callback for the timeout. Why wouldn't it just call the original callback with these values?

components/psst/browser/content/psst_tab_web_contents_observer_unittest.cc
```diff
@@ -43,6 +46,10 @@ using ::testing::InvokeArgument;
 
 namespace psst {
 
+namespace {
+constexpr base::TimeDelta kScriptTimeout = base::Seconds(3);
```
bridiver: never actually wait in tests, it would take way too long to run them if we actually waited for every timeout. Use TaskEnvironment::FastForwardBy instead

components/psst/browser/content/psst_tab_web_contents_observer.cc
vadimstruts: In case of a timeout, we just need to break the flow and inform the UI about the problem (i.e. call `PsstUiDelegate::UpdateTasks` and do not call the next step of the flow).
However, I see some obstacles with simply calling the original callbacks:
Currently, timeouts may occur in two cases: when executing the user script or when executing the policy script.
There are two callbacks for those cases, and their parameter parsing is complicated. Only one of them can call `PsstUiDelegate::UpdateTasks` :
- `PsstTabWebContentsObserver::OnUserScriptResult`
- `PsstTabWebContentsObserver::OnPolicyScriptResult`

Nevertheless, we must make sure that it follows the same flow, with the help of `ShouldInsertScriptForPage(id)`  , since the timeout is also part of the flow.
It is why I defined separate method for Timeout case

bridiver: oh, because `UpdateTasks` would normally be called by the policy script, not the user script? This still seems like a confusing way to short-circuit the process, I think it would be better to pass some something to `OnUserScriptResult` to indicate the script has timed out and `UpdateTasks` can be called from there with the relevant params for the timeout

bridiver: I think I should have been a little more clear here in my response above. TimedScriptRepeatingCallback just seems to be unnecessarily complicating things, but I don't think it's clear in my comment that I don't think it's needed at all. It makes the entire thing harder to understand because there are now at least three layers of callbacks you have to trace through to figure out what's going on. If the timer fires, invalidate the weak ptr and call whatever you need for `UpdateTasks`. I got a bit confused by the initial implementation because of the chain of callbacks so it's ok to have `OnScriptTimeout` which handles `UpdateTasks`, but it should all  happen within `PsstTabWebContentsObserver` with no extra layers of callbacks like this.

components/psst/common/psst_common.h
```diff
@@ -0,0 +1,20 @@
...
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#ifndef BRAVE_COMPONENTS_PSST_COMMON_PSST_COMMON_H_
+#define BRAVE_COMPONENTS_PSST_COMMON_PSST_COMMON_H_
+
+namespace psst {
+
+enum class PsstStatus {
```
bridiver: I think you're eventually passing this back to the UI so shouldn't it be defined in mojo?

components/psst/browser/content/psst_tab_web_contents_observer.h
```diff
@@ -37,7 +41,39 @@ class PsstTabWebContentsObserver : public content::WebContentsObserver {
...
     // Update the UI state based on the applied tasks and progress.
     virtual void UpdateTasks(long progress,
-                             const std::vector<PolicyTask>& applied_tasks) = 0;
+                             const std::vector<PolicyTask>& applied_tasks,
+                             const PsstStatus status) = 0;
+  };
+
+  // Helper class that watches for hanging injected JS scripts and ensures
+  // their callbacks eventually fire.
+  class TimedScriptRepeatingCallback {
```
bridiver: This isn't actually a callback and definitely not a repeating callback so this naming seems misleading

bridiver: also this is an internal implementation detail so it doesn't need to be in the header at al and definitely not public

bridiver: btw this is just for future reference, see https://github.com/brave/brave-core/pull/31131/files#r2345292337

components/psst/browser/content/psst_tab_web_contents_observer.cc
```diff
@@ -199,8 +206,31 @@ void PsstTabWebContentsObserver::OnPolicyScriptResult(
...
+      base::BindOnce(&PsstTabWebContentsObserver::OnScriptTimeout,
+                     weak_factory_.GetWeakPtr(), last_committed_entry_id));
+  inject_script_callback_.Run(script, std::move(callback));
+}
+
+void PsstTabWebContentsObserver::OnScriptTimeout(int id) {
+  if (!ShouldInsertScriptForPage(id)) {
+    return;
+  }
+  weak_factory_.InvalidateWeakPtrs();
```
bridiver, line 244: please leave a comment here explaining that we want to make sure any in-progress user script that returns after the timeout is a no-op

components/psst/README.md
```diff
@@ -29,6 +29,35 @@ https://docs.google.com/document/d/1ccnBWBV_KkknZpZYxcOXTwtfIiaSzeLS5dMd08N2pbs/
 12. Then we enumerate each available URL in the list and do the same as in points #8-11.
 13. When all tasks are completed, the user sees all status and progress information in the consent dialog.
 
+### Workflow failure cases
+
+1. <a id="script-times-out"></a>The script times out or page crashes.</br> This happens when an injected script (either from a user or a policy) never finishes running. Common causes include an infinite loop, a promise that never resolves, or renderer crash can lead to the same situation, when we need to stop the process and notify the front end.
+
+#### Handling Script Timeouts and page crashes
+
+The `TimedScriptRepeatingCallback` class is used to handle cases when a script fails to finish running, (i.e. ["The script times out or page crashes"](#script-times-out)).</br>
```
bridiver: this needs to be updated

components/psst/browser/content/psst_tab_web_contents_observer.cc
```diff
@@ -176,14 +180,17 @@ void PsstTabWebContentsObserver::OnUserScriptResult(
     return;
   }
 
+  timeout_timer_.Stop();
```
bridiver, line 180: Shouldn't this be at the top of this method? If `OnUserScriptResult` then it didn't time out, right? If the policy script is not available we don't want to fire the timeout? Separately (for follow-up) in these early return cases shouldn't we be updating the status?

components/psst/browser/content/psst_tab_web_contents_observer_unittest.cc
```diff
@@ -685,6 +724,61 @@ TEST_F(PsstTabWebContentsObserverUnitTest, UiDelegateUpdateTasksCalled) {
...
+
+  // No policy script executed
+  EXPECT_CALL(inject_script_callback(), Run(policy_script, _)).Times(0);
+
+  DocumentOnLoadObserver observer(web_contents());
+  content::NavigationSimulator::NavigateAndCommitFromBrowser(web_contents(),
+                                                             url);
+  observer.Wait();
+  check_loop.Run();
+  // It means that
```
bridiver: ?

components/psst/browser/content/psst_tab_web_contents_observer.cc
```diff
@@ -199,8 +206,31 @@ void PsstTabWebContentsObserver::OnPolicyScriptResult(
...
-                            script_result_parsed->applied_tasks);
+  timeout_timer_.Stop();
+  ui_delegate_->UpdateTasks(
+      script_result_parsed->progress, script_result_parsed->applied_tasks,
+      script_result_parsed->progress == 100 ? mojom::PsstStatus::kCompleted
+                                            : mojom::PsstStatus::kInProgress);
+}
+
+void PsstTabWebContentsObserver::RunWithTimeout(
+    const int last_committed_entry_id,
```
bridiver, line 228: in a follow-up I think we can drop `last_committed_entry_id` and use `weak_factory_.InvalidateWeakPtrs();` for that case as well to simplify things.

browser/psst/psst_ui_delegate_impl.cc
```diff
@@ -12,7 +12,8 @@ PsstUiDelegateImpl::~PsstUiDelegateImpl() = default;
 
 void PsstUiDelegateImpl::UpdateTasks(
     long progress,
-    const std::vector<PolicyTask>& applied_tasks) {
+    const std::vector<PolicyTask>& applied_tasks,
+    const mojom::PsstStatus status) {
```
bridiver: can we split out adding `PsstStatus` into a separate PR? It can go before or after this one, doesn't matter to me. This seems like a distinct change that isn't directly related to the time out

components/psst/browser/content/psst_tab_web_contents_observer.cc
vadimstruts: you are right, it should be at the top

components/psst/browser/content/psst_tab_web_contents_observer_unittest.cc
```diff
@@ -97,6 +103,30 @@ ACTION_P(InsertScriptInPageCallback, future, value) {
...
+           fut = future, val = value.Clone()]() {
+            std::move(
+                const_cast<
+                    PsstTabWebContentsObserver::InsertScriptInPageCallback&>(
+                    callback))
+                .Run(val.Clone());
+            fut->SetValue(val.Clone());
+          }),
+      base::Seconds(delay_in_secs));
+  task_environment->FastForwardBy(base::Seconds(delay_in_secs));
```
bridiver, line 127: this should be more than the PostDelayedTask delay to ensure it has time to run

components/psst/browser/content/psst_tab_web_contents_observer_unittest.cc
```diff
@@ -97,6 +103,30 @@ ACTION_P(InsertScriptInPageCallback, future, value) {
...
+ACTION_P(InsertScriptInPageDelayedCallback,
+         future,
+         task_environment,
+         delay_in_secs,
+         value) {
+  task_environment->GetMainThreadTaskRunner()->PostDelayedTask(
+      FROM_HERE,
+      base::BindLambdaForTesting(
+          [callback = std::move(
+               const_cast<
```
bridiver, line 115: why are you using const_cast here?

bridiver: nvm, but the way this is written is making the lambda a bit hard to read and understand. I think it would be better to move these assignments out of the lambda capture and then just pass in the the assigned vars just make it easier to see what is going on here.

components/psst/browser/content/psst_tab_web_contents_observer_unittest.cc
bridiver: in fact that still isn't reliable, it should use RunLoop to verify that `InsertScriptInPageCallback` does run, but doesn't update the status or run the the policy script

bridiver: need to think a bit about how to do this since the callback gets invalidated by the weakptr

bridiver: moving this to follow-up

PR #31132, opened 2025-09-10
[plaster] Add an example using plaster to generate a patch file

bridiver:
Add an example using plaster to generate a patch file using re_pattern

### DIFF
--- a/patches/chrome-utility-BUILD.gn.patch
+++ b/patches/chrome-utility-BUILD.gn.patch
@@ -1,5 +1,5 @@
 diff --git a/chrome/utility/BUILD.gn b/chrome/utility/BUILD.gn
-index edd099547589f6f87f0ebe10f7d56e9fadb4c679..8db347dd358650eafd86c9a0d00b30ee921d9a3e 100644
+index 88b6b588f581338a8873767e64d08dcf5333240c..3241c78ddc8d2a5bf32190d22cc6593fa73792db 100644
 --- a/chrome/utility/BUILD.gn
 +++ b/chrome/utility/BUILD.gn
 @@ -268,6 +268,7 @@ static_library("utility") {


rewrite/chrome/utility/BUILD.gn.toml
```diff
@@ -0,0 +1,10 @@
+# Copyright (c) 2025 The Brave Authors. All rights reserved.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+[[substitution]]
+description = 'Add brave/utility sources'
+re_pattern = '''(static_library\("utility"\) {[\s\S]*?)(\s)(^})'''
```
mkarolin: Is there a reason to have the white space before the closing brace in a separate group? Wouldn't
```
'''(static_library\("utility"\) {[\s\S]*?)(^})'''
```
work the same way with 
```
replace = '''\1  import("//brave/utility/sources.gni") sources += brave_utility_sources deps += brave_utility_deps public_deps += brave_utility_public_deps\n\2'''
```

mkarolin (approved):
LGTM (with a question)

rewrite/chrome/utility/BUILD.gn.toml
bridiver: There was some issue with that and I don't recall now, but I don't want to micro-analyze the regexes unless there is a clear functionality problem/benefit.

bridiver: it does seem like it would make it easier to understand though so I'll try again and see

bridiver:
unrelated test failures

PR #31135, opened 2025-09-10
use chromium icon for folders in bookmarks bar

aguscruiz:
Resolves https://github.com/brave/brave-browser/issues/49147

### DIFF
--- a/chromium_src/chrome/browser/ui/bookmarks/bookmark_utils.cc
+++ b/chromium_src/chrome/browser/ui/bookmarks/bookmark_utils.cc
@@ -40,9 +40,7 @@ void BraveToggleBookmarkBarState(content::BrowserContext* browser_context) {
 #define ShouldShowAppsShortcutInBookmarkBar \
   ShouldShowAppsShortcutInBookmarkBar_Unused
 
-#if defined(TOOLKIT_VIEWS)
-#define GetBookmarkFolderIcon GetBookmarkFolderIcon_UnUsed
-#endif
+// No Brave overrides for folder icon; use Chromium implementation.
 
 #define ToggleBookmarkBarWhenVisible                                       \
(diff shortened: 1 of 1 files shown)


aguscruiz:
@simonhong @sangwoo108 I'm getting a windows error it seems. Would you happen to know what the problem is and what I should do? 🥺

sangwoo108:
Let me restart the CI

aguscruiz:
> Let me restart the CI

No luck, still faililng :(

chromium_src/chrome/browser/ui/bookmarks/bookmark_utils.cc
```diff
@@ -40,9 +40,7 @@ void BraveToggleBookmarkBarState(content::BrowserContext* browser_context) {
 #define ShouldShowAppsShortcutInBookmarkBar \
   ShouldShowAppsShortcutInBookmarkBar_Unused
 
-#if defined(TOOLKIT_VIEWS)
-#define GetBookmarkFolderIcon GetBookmarkFolderIcon_UnUsed
-#endif
+// No Brave overrides for folder icon; use Chromium implementation.
```
sangwoo108: I think we can do without this comment. Let me add a commit to remove this line

simonhong:
> LGTM. @simonhong Do you have any other thoughts?

I think we could delete resource files also. Maybe all `IDR_BRAVE_BOOKMARK_FOLDER_XXX` images?

aguscruiz:
> LGTM, but want to note that in the past we got very strong push back from users when we tried to move away from the yellow folders on Windows (cc: @rebron).

Yeah, but Rafa and I talked and he agreed to make it part of macOS26 / Windows 11 to let go of the old icons. They're veeeery old looking by now.

Chrome for reference
image

PR #31141, opened 2025-09-11
Don't reset custom title when it's initial navigation

sangwoo108:
When a tab is newly created, e.g. restoring tabs, it has not yet committed a real page. In this case the origin could be empty but we should not reset the custom title. As the real page will be committed soon and the custom title will be reset or persisted based on the origin of the real page once it's committed.

Resolves https://github.com/brave/brave-browser/issues/49125

### DIFF
--- a/chromium_src/chrome/browser/ui/tab_ui_helper.cc
+++ b/chromium_src/chrome/browser/ui/tab_ui_helper.cc
@@ -38,7 +38,7 @@ void TabUIHelper::UpdateLastOrigin() {
   // custom title. This is to ensure that the custom title is not stale.
   const auto origin =
       web_contents()->GetPrimaryMainFrame()->GetLastCommittedOrigin();
-  if (last_origin_.IsSameOriginWith(origin)) {
+  if (last_origin_ && last_origin_->IsSameOriginWith(origin)) {
     return;
   }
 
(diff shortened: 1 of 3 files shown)


chromium_src/chrome/browser/ui/tab_ui_helper.cc
```diff
@@ -42,6 +42,21 @@ void TabUIHelper::UpdateLastOrigin() {
...
+  // restoring, the real page will be commited soon and the custom title will be
+  // reset or persisted based on the origin of the real page.
+  if (web_contents()->GetController().IsInitialNavigation()) {
+    return;
+  }
+
+  // We reset the custom title only when the last origin is initialized. When
+  // restoring tabs, last origin could be uninitialized yet, and we do not want
+  // to reset the custom title in that case.
+  if (last_origin_initialized_) {
```
simonhong: Instead of having another bool flag,
Could we simply check `last_origin_` is valid or not like `if (!last_origin().opaque()) `?

sangwoo108: I don't think so. IIUC, `opaque` isn't saying that it's invalid or empty. It's just invisible to users. So I believe it can become `opaque` even after the initial loading. As for the `is_valid`, I couldn't find such a thing. Do you happen to compatible API?

sangwoo108: If we'd like to avoid bool flag, we could wrap the last_origin_ with unique_ptr and check it's validity. How about this?

simonhong: I see. 
How about having `last_origin_` as optional if you think it makes more simpler? it's up to you.

sangwoo108: That'd be definitely better!

PR #31143, opened 2025-09-11
[AI Chat]: Send page title to backend as part of page content

fallaciousreasoning:
Resolves https://github.com/brave/brave-browser/issues/49169

After a bit of discussion with @yrliou she mentioned the LLM might understand just sending `Title: <page>` title better than some structured data. In my testing she was 100% right, and it works pretty well. It also doesn't require any backend changes and its easy to share the implementation with BYOM.

### DIFF
--- a/components/ai_chat/core/browser/engine/engine_consumer.cc
+++ b/components/ai_chat/core/browser/engine/engine_consumer.cc
@@ -10,8 +10,15 @@
 
 #include "base/base64.h"
 #include "base/strings/strcat.h"
+#include "base/strings/utf_string_conversions.h"
+#include "third_party/abseil-cpp/absl/strings/str_format.h"
 
 namespace ai_chat {
+
+std::string FormatPageContentWithTitle(const PageContent& content) {
+  return absl::StrFormat("Title: %v\n%v", base::UTF16ToUTF8(content.title),
+                         content.content);
+}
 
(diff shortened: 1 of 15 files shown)


components/ai_chat/core/browser/engine/engine_consumer.cc
```diff
@@ -10,9 +10,16 @@
 
 #include "base/base64.h"
 #include "base/strings/strcat.h"
+#include "base/strings/utf_string_conversions.h"
+#include "third_party/abseil-cpp/absl/strings/str_format.h"
 
 namespace ai_chat {
 
+std::string FormatPageContentWithTitle(const PageContent& content) {
+  return absl::StrFormat("Title: %v\n%v", base::UTF16ToUTF8(content.title),
```
fallaciousreasoning, line 19: I wonder if instead of this `Title: <pageTitle>` we should use a `#` so its kind of markdowny?

fallaciousreasoning: Alternatively, I could prepend a `<title>Title</title>` tag

fallaciousreasoning:
reviewers please don't take a look yet - things are in a bit of flux while we work out the best way to ensure TabHelpers and TabFeatures don't fall out of sync

PR #30839, opened 2025-08-26
BRAVE_COSMETIC_FILTERING no longer uses the firstParty pattern marker.

boocmp:
Resolves https://github.com/brave/brave-browser/issues/48500

1. The default value for `BRAVE_COSMETIC_FILTERING` has been changed from `CONTENT_SETTING_DEFAULT` to `CONTENT_SETTING_ASK`. This new default, which maps to `ControlType::BLOCK_THIRD_PARTY`, aligns with the value previously returned by `GetCosmeticFilteringControlType` by default.

2. Changed the `BRAVE_COSMETIC_FILTERING` content setting to no longer use the `https://firstparty` pattern marker. The associated preference has been renamed to `profile.content_settings.exceptions.cosmeticFilteringV2`, and a migration was implemented in `BravePrefProvider`.

PR #31146, opened 2025-09-11
Utilities for shields content settings values.

boocmp:
The split of https://github.com/brave/brave-core/pull/30839, requested [here ](https://github.com/brave/brave-core/pull/30839#discussion_r2337322307).

1. `ControlType` enum has been moved to `common`
2. `ShieldsContentSettingsTypes` has been changed to be constexpr
3. Added utilities for serialize web settings into base:::Value

### DIFF
--- a/components/brave_shields/core/common/brave_shield_constants.h
+++ b/components/brave_shields/core/common/brave_shield_constants.h
@@ -6,10 +6,15 @@
 #ifndef BRAVE_COMPONENTS_BRAVE_SHIELDS_CORE_COMMON_BRAVE_SHIELD_CONSTANTS_H_
 #define BRAVE_COMPONENTS_BRAVE_SHIELDS_CORE_COMMON_BRAVE_SHIELD_CONSTANTS_H_
 
+#include "base/containers/fixed_flat_map.h"
+#include "base/containers/fixed_flat_set.h"
+#include "base/containers/map_util.h"
 #include "base/files/file_path.h"
+#include "components/content_settings/core/common/content_settings_types.h"
 
 namespace brave_shields {
 
+// Content/Web settings:
 inline constexpr char kAds[] = "shieldsAds";
 inline constexpr char kCosmeticFiltering[] = "cosmeticFiltering";
 inline constexpr char kTrackers[] = "trackers";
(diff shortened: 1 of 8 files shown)


bridiver (approved):

components/brave_shields/core/common/brave_shields_settings_values.h
```diff
@@ -0,0 +1,107 @@
...
+
+  static SettingType FromValue(const base::Value& value) {
+    if (const auto* dict = value.GetIfDict()) {
+      if (const auto v = dict->FindInt(BraveShieldsSetting::kName);
+          v.has_value()) {
+        return traits::SettingTraits<SettingType>::From(*v).value_or(
+            kDefaultValue);
+      }
+    }
+    LOG(ERROR) << "ShieldSetting " << BraveShieldsSetting::kName
```
bridiver, line 96: I somehow missed this log message, I think this should probably be a CHECK here? I believe that's what happens in normal content settings when parsing the enum? Whatever happens there, we should do the same thing here

boocmp: Content setting fallbacks to a default value (the code that you was pointed to while fixing the crash with an invalid prefs).

bridiver: ok, then DCHECK is fine here but logging should be removed

components/content_settings/core/browser/brave_content_settings_utils.cc
```diff
@@ -28,64 +27,42 @@ bool CanPatternBeConvertedToWildcardSchemeAndPort(
...
-      break;
-  }
-  NOTREACHED() << "All handled shields content type above.";
+  const auto* name =
+      IsShieldsContentSettingsType(content_type)
+          ? base::FindOrNull(brave_shields::kShieldsContentTypeNames,
+                             content_type)
+          : nullptr;
+  CHECK(name) << "Content type isn't a shield content type.";
+  return *name;
```
cdesouza-chromium, line 65: This function should be returning `std::string_view` as the value will always be a compile time string.

components/brave_shields/core/common/brave_shield_constants.h
```diff
@@ -21,12 +26,67 @@ inline constexpr char kBraveShields[] = "braveShields";
...
+        ContentSettingsType::BRAVE_FINGERPRINTING_V2,
+        ContentSettingsType::BRAVE_SHIELDS,
+        ContentSettingsType::BRAVE_REFERRERS,
+        ContentSettingsType::BRAVE_COOKIES,
+    });
+
+using ShieldsContentSettingsTypes = decltype(kShieldsContentSettingsTypes);
+
+inline constexpr auto kShieldsContentTypeNames =
+    base::MakeFixedFlatMap<ContentSettingsType, const char*>({
```
cdesouza-chromium, line 50: it would have been better to use `std::string_view` here, so as to better pass it around without relying on `strlen`.

components/brave_shields/core/common/brave_shields_settings_values.h
```diff
@@ -0,0 +1,107 @@
...
+}
+
+consteval const char* GetShieldsContentTypeName(
+    ContentSettingsType content_type) {
+  // Linear, because `find` isn't constexpr/consteval.
+  for (const auto& v : kShieldsContentTypeNames) {
+    if (v.first == content_type) {
+      return v.second;
+    }
+  }
```
cdesouza-chromium, line 58: what is the point of doing this manual search. Shouldn't we be using the regular `fixed_flat_map` search here?

boocmp: As said in the comment at line 53, the regular isn't consteval

PR #31149, opened 2025-09-11
Add starter pack for "Ask Brave Search"

zenparsing:
Resolves https://github.com/brave/brave-browser/issues/49179

Chromium maintains a list of search engine template URLs that are associated with omnibox keywords. For example:

* `@bookmarks` allows searching bookmarks
* `@history` allows searching history

These built-in "engines" are called "starter packs". We current have overrides to disable the recently added `@gemini` and `@aimode` starter packs.

This change introduces our own list of starter packs, along with a new Brave-defined starter pack `@ask`, which will direct a query to the Ask interface of Brave Search. Since we now maintain our own list of starter packs, some of the overrides which hide Chromium's starter packs are no longer required.

Screenshot 2025-09-26 at 11 26 00 AM

Screenshot 2025-09-22 at 2 11 40 PM


patches/components-search_engines-template_url_starter_pack_data.h.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/components/search_engines/template_url_starter_pack_data.h b/components/search_engines/template_url_starter_pack_data.h
+index 86fbd587b65b194fdc528b6dbd51f020168eb50e..07aa67626cd9369bc637eee9a8b3bc38245dcf35 100644
+--- a/components/search_engines/template_url_starter_pack_data.h
++++ b/components/search_engines/template_url_starter_pack_data.h
+@@ -28,6 +28,7 @@ typedef enum {
+   kGemini = 4,
+   kPage = 5,
+   kAiMode = 6,
++  kAskBraveSearch = 7,
```
bridiver: `#define kMaxStarterPackId kAskBraveSearch, kMaxStarterPackId`
notice I left off  ` = 7` to avoid conflicts, it will default to the next int value

patches/components-search_engines-search_engine_type.h.patch
```diff
@@ -0,0 +1,20 @@
+diff --git a/components/search_engines/search_engine_type.h b/components/search_engines/search_engine_type.h
+index e1228cbf36a79367d10319d6c8f8661cc2978916..0102983617da898c563240bfef3df8fc39bf3f1f 100644
+--- a/components/search_engines/search_engine_type.h
++++ b/components/search_engines/search_engine_type.h
+@@ -98,6 +98,7 @@ enum SearchEngineType {
+   SEARCH_ENGINE_FREESPOKE = 78,
+   SEARCH_ENGINE_KAGI = 79,
+   SEARCH_ENGINE_STARTER_PACK_AI_MODE = 80,
++  SEARCH_ENGINE_STARTER_PACK_ASK_BRAVE_SEARCH = 81,
```
bridiver: `#define SEARCH_ENGINE_MAX ...` similar to https://github.com/brave/brave-core/pull/31149/files#r2350205843

patches/components-search_engines-template_url.cc.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/components/search_engines/template_url.cc b/components/search_engines/template_url.cc
+index 33638c629aa01a01fd71cd7ad234f20138290861..b87e871b2ae76487dbdecfad50cfa210a904d529 100644
+--- a/components/search_engines/template_url.cc
++++ b/components/search_engines/template_url.cc
+@@ -1859,6 +1859,7 @@ BuiltinEngineType TemplateURL::GetBuiltinEngineType() const {
+         return KEYWORD_MODE_STARTER_PACK_PAGE;
+       case template_url_starter_pack_data::kAiMode:
+         return KEYWORD_MODE_STARTER_PACK_AI_MODE;
```
bridiver: `#define KEYWORD_MODE_STARTER_PACK_AI_MODE KEYWORD_MODE_STARTER_PACK_AI_MODE; case ....`

patches/components-omnibox-browser-omnibox_controller.cc.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/components/omnibox/browser/omnibox_controller.cc b/components/omnibox/browser/omnibox_controller.cc
+index eeb1d6ea66171fd514127f0b06e7c7378d973dfe..459e2315dee726ddcdf79d8340ddd893d0f0cfcf 100644
+--- a/components/omnibox/browser/omnibox_controller.cc
++++ b/components/omnibox/browser/omnibox_controller.cc
+@@ -173,6 +173,7 @@ bool OmniboxController::IsSuggestionHidden(
+         turl->starter_pack_id() == template_url_starter_pack_data::kGemini) {
+       return true;
+     }
++    BRAVE_OMNIBOX_CONTROLLER_IS_SUGGESTION_HIDDEN
```
bridiver: `#define kGemini kGemini) { return true; } if turl && turl->starter_pack_id() == template_url_starter_pack_data::kAskBraveSearch`

chromium_src/chrome/browser/ui/views/location_bar/selected_keyword_view.cc
```diff
@@ -0,0 +1,18 @@
+/* Copyright (c) 2025 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "brave/components/vector_icons/vector_icons.h"
```
bridiver: avoid adding brave dependencies in chromium_src overrides. Use interface/impl pattern and do something like

```
if (IsBraveSearchStarterPack(template_url) { 
  vector_icon = GetBraveSearchStarterPackIcon();
}
```
where `IsBraveSearchStarterPack` and `GetBraveSearchStarterPackIcon` are forward declared

chromium_src/chrome/browser/ui/views/location_bar/selected_keyword_view.cc
```diff
@@ -0,0 +1,18 @@
...
+
+// Sets the icon that is displayed when a Brave-defined starter pack engines is
+// active.
+#define BRAVE_SELECTED_KEYWORD_VIEW_SET_CUSTOM_IMAGE                         \
+  if (template_url && template_url->starter_pack_id() ==                     \
+                          template_url_starter_pack_data::kAskBraveSearch) { \
+    vector_icon = &kLeoMessageBubbleAskIcon;                                 \
```
bridiver: I thought we were dropping leo in favor of aichat in the code? Or is this different because it's using search?

patches/components-omnibox-browser-featured_search_provider.cc.patch
```diff
@@ -1,8 +1,16 @@
...
-index 3f93bf5689e8b51f6e842373e530b81fa7a905ed..e0ab6d9dfd5c44a7ec5e71168b8ad0a2126f75ef 100644
+index 3f93bf5689e8b51f6e842373e530b81fa7a905ed..7c1e35c18616cfefc93a8b48d43a034d7ab8ccfb 100644
 --- a/components/omnibox/browser/featured_search_provider.cc
 +++ b/components/omnibox/browser/featured_search_provider.cc
-@@ -290,6 +290,7 @@ void FeaturedSearchProvider::AddFeaturedKeywordMatches(
+@@ -74,6 +74,7 @@ constexpr int kIphRelevance = 5000;
+ int StarterPackRelevance(
+     template_url_starter_pack_data::StarterPackId starter_pack_id) {
+   switch (starter_pack_id) {
++    BRAVE_FEATURED_SEARCH_PROVIDER_STARTER_PACK_RELEVANCE
```
bridiver: override `kMaxStarterPackId` similar to other examples

patches/chrome-browser-ui-views-location_bar-selected_keyword_view.cc.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/chrome/browser/ui/views/location_bar/selected_keyword_view.cc b/chrome/browser/ui/views/location_bar/selected_keyword_view.cc
+index f824957011655e33004ff528d5a791878cbb3868..77ce646a829661087760894119495f4d198130c0 100644
+--- a/chrome/browser/ui/views/location_bar/selected_keyword_view.cc
++++ b/chrome/browser/ui/views/location_bar/selected_keyword_view.cc
+@@ -130,6 +130,7 @@ void SelectedKeywordView::SetCustomImage(const gfx::Image& image) {
+     vector_icon = &vector_icons::kGoogleAgentspaceMonochromeLogoIcon;
+ #endif
+   }
++  BRAVE_SELECTED_KEYWORD_VIEW_SET_CUSTOM_IMAGE
```
bridiver: `#define kSearchAggregator kSearchAggregator) { /* this is guarded by GOOGLE_CHROME_BRANDING anyway */ } if (...) { ... } else if (false`

patches/components-search_engines-template_url.h.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/components/search_engines/template_url.h b/components/search_engines/template_url.h
+index 176918e83a4f784d819985d27e0d31da0ab94e80..011709fa6df7345d2854f7d170c88e29cb5f46c7 100644
+--- a/components/search_engines/template_url.h
++++ b/components/search_engines/template_url.h
+@@ -828,6 +828,7 @@ class TemplateURL {
+   // Some starter packs are considered 'ask a question' kind of starter packs.
+   // This can be used to condition UI text or a11y strings.
+   bool is_ask_starter_pack() const {
```
bridiver:
```
#define is_ask_starter_pack() is_ask_starter_pack_chromium()

bool is_ask_starter_pack() {
  if (starter_pack_id() == template_url_starter_pack_data::kAskBraveSearch) { \
    return true;                                                              \
  }
  return is_ask_starter_pack_chromium();
}
```

patches/components-omnibox-browser-autocomplete_match.cc.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/components/omnibox/browser/autocomplete_match.cc b/components/omnibox/browser/autocomplete_match.cc
+index 5750f28aa23674724ef8bf3eca021cf60f26c970..0543d6ffae50ed3a8db6f923b08b6bb045f4dfb6 100644
+--- a/components/omnibox/browser/autocomplete_match.cc
++++ b/components/omnibox/browser/autocomplete_match.cc
+@@ -1346,6 +1346,7 @@ std::u16string AutocompleteMatch::GetKeywordPlaceholder(
+     case template_url_starter_pack_data::kAiMode:
+       message_id = IDS_OMNIBOX_AI_MODE_SCOPE_PLACEHOLDER_TEXT;
```
bridiver: `#define IDS_OMNIBOX_AI_MODE_SCOPE_PLACEHOLDER_TEXT IDS_OMNIBOX_AI_MODE_SCOPE_PLACEHOLDER_TEXT; break; case ...`

chromium_src/components/omnibox/browser/keyword_provider.cc
```diff
@@ -0,0 +1,16 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "components/search_engines/template_url_starter_pack_data.h"
+
+// In `KeywordProvider::FillInUrlAndContents`, when the "@ask" keyword is in
+// use, we want the empty query placeholder text to be "question oriented
+// rather than "search oriented". Use the same logic for kAskBraveSearch as
+// upstream uses for kAiMode.
+#define kAiMode kAskBraveSearch
```
bridiver: you're removing kAiMode with this, please use `kMaxStarterPackId` to anchor this

chromium_src/components/omnibox/browser/featured_search_provider.cc
```diff
@@ -4,12 +4,22 @@
  * You can obtain one at https://mozilla.org/MPL/2.0/. */
 
 #include "components/omnibox/browser/omnibox_field_trial.h"
+#include "components/search_engines/template_url_starter_pack_data.h"
+
+// Return a relevance value for any Brave-defined starter pack engines.
+#define kMaxStarterPackId \
+  kAskBraveSearch:        \
+  return 1461;            \
```
bridiver: what do these ids correspond to?

chromium_src/chrome/browser/ui/views/location_bar/selected_keyword_view.cc
zenparsing: This "Leo" refers to the vector icons coming from "nala".

chromium_src/components/omnibox/browser/keyword_provider.cc
zenparsing: `kMaxStarterPackId` isn't in this particular file. We can still use `kAiMode` as the anchor, but add a branch instead of replace.

chromium_src/components/omnibox/browser/featured_search_provider.cc
zenparsing: I believe this corresponds to the omnibox result relevance score for the various starter packs. It should match (or be higher than) the value for `kAiMode`. Instead of hardcoding a value, another possibility would be to call the function recursively.

```
return StarterPackRelevance(template_url_starter_pack_data::StarterPackId::kAiMode);
```

chromium_src/chrome/browser/ui/views/location_bar/selected_keyword_view.cc
zenparsing: @bridiver In what target or folder should those helpers be defined?

chromium_src/components/search_engines/DEPS
```diff
@@ -1,4 +1,5 @@
 include_rules = [
   "+brave/components/search_engines",
   "+components/country_codes",
+  "+components/grit/brave_components_strings.h",
```
zenparsing, line 4: Required due to the introduction of new strings for the "ask" starter pack.

@bridiver Will this introduce a missing dep to `//components/search_engines`? If so, how should it be resolved?

patches/chrome-browser-ui-views-location_bar-selected_keyword_view.cc.patch
zenparsing: Would it be OK to hang this off of `kAiMode` (as in current version), since it's a bit more closely related?

chromium_src/chrome/browser/ui/views/location_bar/selected_keyword_view.cc
bridiver: corresponding brave/browser/ui/... and add the dep directly to brave/browser or brave/browser/ui. It just needs to exist somewhere at link time

chromium_src/components/omnibox/browser/keyword_provider.cc
bridiver: maybe I mixed something up, I'll take a look

chromium_src/components/omnibox/browser/featured_search_provider.cc
bridiver: I would definitely not use a hard-coded value here unless there is some static or similar check

chromium_src/chrome/browser/ui/views/location_bar/selected_keyword_view.cc
zenparsing: Added to `//brave/browser/ui/views/location_bar/brave_selected_keyword_view_util`. I went with just `GetBraveSearchStarterPackIcon` because `IsBraveSearchStarterPack` seemed redundant, but I can add it if you prefer.

chromium_src/components/omnibox/browser/omnibox_edit_model_unittest.cc
```diff
@@ -0,0 +1,14 @@
+/* Copyright (c) 2025 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "components/search_engines/template_url_starter_pack_data.h"
+
+// Let the upstream unit test use upstream's list of starter pack engines,
+// rather than Brave's overridden list.
+#define GetStarterPackEngines GetStarterPackEngines_ChromiumImpl
```
bridiver: this doesn't necessarily seem useful, we run the upstream unit tests to guard against breaking things, but we have intentionally made changes here so checking the unchanged code seems like it kind of defeats the purpose. I think I would disable the relevant test instead in the filters file

mkarolin (approved):
`strings`++

chromium_src/components/omnibox/browser/omnibox_edit_model_unittest.cc
zenparsing: Added to test filters.

chromium_src/components/search_engines/DEPS
goodov: yes, please add a GN dependency into `brave/components/search_engines/sources.gni`

to test if you did everything right, add `#include "components/grit/brave_components_strings.h"` into original chromium file and run gn check.

zenparsing: > to test if you did everything right, add #include "components/grit/brave_components_strings.h" into original chromium file and run gn check.

I still get an include error.

It looks like the deps defined in `brave/components/search_engines/sources.gni`  are used in the patch for `//third_party/search_engines_data/BUILD.gn` and not `//components/search_engines/BUILD.gn` (which we aren't currently patching).

Any suggestions on the cleanest way to proceed?

goodov: yes, you're right. The target was moved to third_party at some point, but this file was not adjusted.

please rename these vars and update BUILD.gn patch accordingly:
```
brave_components_search_engines_sources -> brave_third_party_search_engines_data_prepopulated_engines_sources

brave_components_search_engines_deps -> brave_third_party_search_engines_data_prepopulated_engines_deps
```

use `brave_components_search_engines_deps` to add required dep into a `//components/search_engines` (you need to add a patch).

zenparsing: Done - thanks!

goodov (approved):

PR #31151, opened 2025-09-11
Intermittent browser test failures

SergeyZhukovsky:
From time to time we have mass browser tests failures on Android. We usually just rebuild and it works.
It's hard to replicate it locally, basically it happens when you run it in a loop. It happens 3-4 times out of 100 tries.
I created a follow up to figure it out and re-enable as it's time consuming https://github.com/brave/brave-browser/issues/49176
Disables AndroidPageAppearingBrowserTest.TestSwapPageAppearing temporary.
Resolves: https://github.com/brave/brave-browser/issues/49175

### DIFF
--- a/components/brave_wallet/browser/android_page_appearing_browsertest.cc
+++ b/components/brave_wallet/browser/android_page_appearing_browsertest.cc
@@ -399,7 +399,8 @@ IN_PROC_BROWSER_TEST_F(AndroidPageAppearingBrowserTest,
   }
 }
 
-IN_PROC_BROWSER_TEST_F(AndroidPageAppearingBrowserTest, TestSwapPageAppearing) {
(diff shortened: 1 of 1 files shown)


components/brave_wallet/browser/android_page_appearing_browsertest.cc
```diff
@@ -399,7 +399,8 @@ IN_PROC_BROWSER_TEST_F(AndroidPageAppearingBrowserTest,
   }
 }
 
-IN_PROC_BROWSER_TEST_F(AndroidPageAppearingBrowserTest, TestSwapPageAppearing) {
+IN_PROC_BROWSER_TEST_F(AndroidPageAppearingBrowserTest,
```
bridiver, line 402: in the future please add `TODO(https://github.com/brave/brave-browser/issues/49176) ...` in the comments

PR #31165, opened 2025-09-11
[AI Chat]: Suggested Models in Model Selector

Douglashdaniel:
## Description 

Updates the sorting in the `Model Selector` for recommend models to be rendered in the following order.

- Auto
- Default model for the user (if not auto)
- Current active model (if not auto or any of the others in this list)
- Default free model (if different from any of above)
- Default premium model (if different from any of above)

Resolves <https://github.com/brave/brave-browser/issues/48852>

## Test Plan:
1. Start a new conversation with `Leo AI`
2. Open the `Model Selector`
3. You should `Automatic` at the top
4. Your `Default Model` from settings if not `Automatic`
5. and then `Llama 3.1 8B` (Basic Suggested) and `Claude Sonnet` (Premium Suggested)
6. Click `Show all models` and select some another model in the list
7. Reopen the `Model Selector` and that model should be visible in the `Recommend` list
8. Go to settings and change you `Default Model`
9. The `Model selector` should observe this change and update

https://github.com/user-attachments/assets/05585c76-4f81-4d7b-9324-dec9b48b7e22

### DIFF
--- a/components/ai_chat/core/common/mojom/ai_chat.mojom
+++ b/components/ai_chat/core/common/mojom/ai_chat.mojom
@@ -389,6 +389,8 @@ struct Model {
   bool vision_support;
   // Model supports tool use
   bool supports_tools = false;
+  // Is a suggested model
+  bool is_suggested_model;
 };
 
 struct ModelWithSubtitle {
@@ -545,6 +547,7 @@ struct ConversationState {
   bool is_request_in_progress;
   array<Model> all_models;
   string current_model_key;
+  string default_model_key;
   array<string> suggested_questions;
   SuggestionGenerationStatus suggestion_status;
   array<AssociatedContent> associated_content;
@@ -684,7 +687,9 @@ interface ConversationUI {
   // it could be changed from the browser.
   // When the browser changes the model, it will also provide a new list
   // of models.
-  OnModelDataChanged(string conversation_model_key, array<Model> all_models);
+  OnModelDataChanged(string conversation_model_key,
+                     string default_model_key,
+                     array<Model> all_models);
   // Provides the full list of questions for the active tab in the page's
   // Browser window.
   OnSuggestedQuestionsChanged(
--- a/components/ai_chat/resources/page/components/model_selector/index.tsx
+++ b/components/ai_chat/resources/page/components/model_selector/index.tsx
@@ -16,6 +16,8 @@ import { useConversation } from '../../state/conversation_context'
 import { getModelIcon } from '../../../common/constants'
 import styles from './style.module.scss'
 
+const AUTO_MODEL_KEY = 'chat-automatic'
+
 export function ModelSelector() {
   const aiChatContext = useAIChat()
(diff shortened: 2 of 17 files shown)


components/ai_chat/core/common/mojom/ai_chat.mojom
```diff
@@ -684,7 +687,9 @@ interface ConversationUI {
   // it could be changed from the browser.
   // When the browser changes the model, it will also provide a new list
   // of models.
-  OnModelDataChanged(string conversation_model_key, array<Model> all_models);
+  OnModelDataChanged(string conversation_model_key,
+                     string default_model_key,
```
fallaciousreasoning, line 691: One thing I was wondering looking at this is whether the `defaultModelKey` should be coming from the AIChatService, seeing as its global state.

On the other hand, it seems like all the other model stuff flows through here, so it makes sense and looks like it was simpler to implement.

components/ai_chat/resources/page/components/model_selector/index.tsx
```diff
@@ -24,16 +26,64 @@ export function ModelSelector() {
...
+  const suggestedModels = React.useMemo(() => {
+    return conversationContext.allModels.filter(
+      (model) => model.isSuggestedModel,
+    )
+  }, [conversationContext.allModels])
+
+  const recommendedModels = React.useMemo(() => {
+    // Show all BASIC_AND_PREMIUM models if showAllModels is true
+    if (showAllModels) {
+      return conversationContext.allModels.filter(
```
fallaciousreasoning, line 39: Sorry, slightly confused by this. If we set `showAllModels` then we only show models which have a different basic and premium tier?

Should this be doing something like:

```
BASIC_AND_PREMIUM
    + (isPremium ? PREMIUM : BASIC)
```

components/ai_chat/resources/page/components/model_selector/index.tsx
```diff
@@ -24,16 +26,64 @@ export function ModelSelector() {
...
-    () =>
-      conversationContext.allModels.filter(
+  // Memos
+  const suggestedModels = React.useMemo(() => {
+    return conversationContext.allModels.filter(
+      (model) => model.isSuggestedModel,
+    )
+  }, [conversationContext.allModels])
+
+  const recommendedModels = React.useMemo(() => {
```
fallaciousreasoning: should this just be `models`? Its not necessarily just the recommended ones

Douglashdaniel: Changed to `models`

components/ai_chat/resources/page/components/model_selector/index.tsx
Douglashdaniel: Filtering by `BASIC` actually does not return any models, but I found that `BASIC_AND_PREMIUM` will include all `Basic` models since they are technically also available to `PREMIUM` users

kylehickinson (approved):

PR #31172, opened 2025-09-12
Implement new preference screen "Customize menu"

simoarpe:
## Resolves https://github.com/brave/brave-browser/issues/41801

This PR introduces a new preference screen that can be opened in two different ways:
- Tapping on Brave overflow menu and selecting Customize Menu
- Tapping on Brave overflow menu and navigating to Setting > Appearance > Customize menu

Android_Emulator_-_Pixel_9_Pro_API_33_5554

Android_Emulator_-_Pixel_9_Pro_API_33_5554

The custom menu item preference screen follows the Figma design and is composed by two sections, `Main menu` and `Page actions`.

The screen shows only those items that are supported by the device, items that will never appear won't be shown.
Each item's visibility will default to `true` and each preference is saved on shared preferences.

The only two items whose visibility cannot be modified are the Main Settings item and the Custom Menu item itself.

### ☀️ Light Theme 

| Layout Bounds Off | Layout Bounds On |
|---------|--------|
| Screenshot 2025-09-12 at 1 39 22 PM  | Screenshot 2025-09-12 at 1 39 57 PM |
 

### 🌒 Dark Theme

| Layout Bounds Off | Layout Bounds On |
|---------|--------|
| Screenshot 2025-09-12 at 1 41 38 PM  | Screenshot 2025-09-12 at 1 41 59 PM |
 
## Menu icon standardization and fetching

_Some_ menu icons don't have the appropriate size to be displayed correctly in the switch preference icon (`ChromeSwitchPreference`) so when its width or height is less than 24 DP it gets scaled appropriately.

Drawables (and vector drawables in particular) cannot be parceled, so the easiest (and fastest) solution is to retrieve them by their menu ID, hence the creation of `MENU_ICON_MAP`.

## Demo

https://github.com/user-attachments/assets/b5695a3f-dc5b-48f4-9466-602924eb90f8


browser/customize_menu/android/java/src/org/chromium/brave/browser/customize_menu/CustomizeBraveMenu.java
```diff
@@ -0,0 +1,396 @@
...
+    }
+
+    /**
+     * Checks whether a menu item should be visible based on user preferences. Returns true by
+     * default if no preference has been set.
+     *
+     * @param itemId the resource ID of the menu item to check
+     * @return {@code true} if the item should be visible, {@code false} if it should be hidden
+     */
+    public static boolean isVisible(final int itemId) {
```
SergeyZhukovsky: The integer IDs generated in the R class, including R.id values, are not permanent between builds and can change when resources are modified. I would better use the resource name (not the R.id value).
`String resourceName = getResources().getResourceEntryName(R.id.....);`

browser/customize_menu/android/java/src/org/chromium/brave/browser/customize_menu/CustomizeBraveMenu.java
```diff
@@ -0,0 +1,396 @@
...
+        }
+
+        // If required, standardize icon size to match menu icons.
+        if (width < iconSizePx || height < iconSizePx) {
+            Bitmap bitmap = Bitmap.createBitmap(iconSizePx, iconSizePx, Bitmap.Config.ARGB_8888);
+            bitmap.setDensity(context.getResources().getDisplayMetrics().densityDpi);
+            Canvas canvas = new Canvas(bitmap);
+            drawable.setBounds(0, 0, canvas.getWidth(), canvas.getHeight());
+            drawable.draw(canvas);
+            drawable = new BitmapDrawable(context.getResources(), bitmap);
```
SergeyZhukovsky: It's better to move the drawable creation in a bg thread as it can cause performance issues, including ANR errors, especially on low-end devices.

browser/customize_menu/android/java/src/org/chromium/brave/browser/customize_menu/CustomizeBraveMenu.java
```diff
@@ -0,0 +1,396 @@
...
+        MENU_ICON_MAP.put(R.id.brave_wallet_id, R.drawable.ic_crypto_wallets);
+        MENU_ICON_MAP.put(R.id.brave_playlist_id, R.drawable.ic_open_playlist);
+        MENU_ICON_MAP.put(R.id.add_to_playlist_id, R.drawable.ic_baseline_add_24);
+        MENU_ICON_MAP.put(R.id.brave_news_id, R.drawable.ic_news);
+        MENU_ICON_MAP.put(R.id.brave_speedreader_id, R.drawable.ic_readermode);
+        MENU_ICON_MAP.put(R.id.brave_leo_id, R.drawable.ic_brave_ai);
+        MENU_ICON_MAP.put(R.id.request_brave_vpn_id, R.drawable.ic_vpn);
+        MENU_ICON_MAP.put(R.id.brave_rewards_id, R.drawable.brave_menu_rewards);
+        MENU_ICON_MAP.put(R.id.set_default_browser, R.drawable.brave_menu_set_as_default);
+        MENU_ICON_MAP.put(R.id.exit_id, R.drawable.brave_menu_exit);
```
SergeyZhukovsky, line 125: what if upstream adds a new menu item? Will we just show it without ability to customize or do we just miss it and it won't be shown at all?

android/brave_java_resources.gni
```diff
@@ -537,6 +537,7 @@ brave_java_resources = [
   "java/res/drawable/ic_web.xml",
   "java/res/drawable/ic_widget_generic.xml",
   "java/res/drawable/ic_widget_generic_with_container.xml",
+  "java/res/drawable/ic_window_screwdriver.xml",
```
samartnik: not a blocker for this PR, but we really should add our resources directly instead of copying them to the upstream folder and add them from there, copy should be reserved only for overriding upstream's resources that we are taking ownership of

SergeyZhukovsky:
Not sure why don't we want to control visibility of Custom Menu item itself. `The only two items whose visibility cannot be modified are the Main Settings item and the Custom Menu item itself.` I'm pretty sure there will be a decent cohort of users who won't like it. It's still accessible via `Settings` even if it's hidden in the main menu.

browser/customize_menu/android/java/src/org/chromium/brave/browser/customize_menu/CustomizeBraveMenu.java
simoarpe: Introduced `getResourceEntryName(int)` by commit 👉 [`42c7ba8` (#31172)](https://github.com/brave/brave-core/pull/31172/commits/42c7ba8f6fbf539b9bebc4255fcf98477e28825c)

browser/customize_menu/android/java/src/org/chromium/brave/browser/customize_menu/CustomizeBraveMenu.java
simoarpe: Offloaded `Bitmap`s recreation to a background thread using `PostTaks`s 👉 [`8c202ec` (#31172)](https://github.com/brave/brave-core/pull/31172/commits/8c202ec019f468a9cc52a4bfe33d3a47fb0076e4)

android/java/org/chromium/chrome/browser/app/BraveActivity.java
```diff
@@ -445,6 +448,24 @@ public boolean onMenuOrKeyboardAction(
...
+            assert delegate instanceof BraveTabbedAppMenuPropertiesDelegate;
+            final BraveTabbedAppMenuPropertiesDelegate braveTabbedAppMenuPropertiesDelegate =
+                    (BraveTabbedAppMenuPropertiesDelegate) delegate;
+
+            final Bundle bundle =
+                    CustomizeBraveMenu.populateBundle(
+                            getResources(),
+                            new Bundle(),
+                            braveTabbedAppMenuPropertiesDelegate.buildMainMenuModelList(),
+                            braveTabbedAppMenuPropertiesDelegate.buildPageActionsModelList());
```
SergeyZhukovsky, line 462: I wonder how fast is that execution? Maybe you can profile it to make sure it doesn't make a performance much worse? As we do that on every 3 dot menu tap. We could potentially cache it to avoid rebuilding each time. The measurements could probably be done with `Android Studio Profiler`.

browser/customize_menu/android/java/src/org/chromium/brave/browser/customize_menu/CustomizeBraveMenu.java
simoarpe: If upstream adds a new item it will be hard to miss it as our tests will fail. If for some reason we don't notice this,  we will show it without the ability to customize its visibility as the whole customization implementation gracefully returns without any modification if an item is not recognized.
On debug builds we assert emitting the menu item ID that is not recognized.

android/java/org/chromium/chrome/browser/app/BraveActivity.java
simoarpe: Sure! I did it already and it was pretty fast, hence I've preferred to skip caching to avoid further complexity.
Even if it seems heavy at first glance the drawables are pretty small, the bitmap recreation happens only for two items and the whole UI didn't suffer at all.
Let me provide some numbers.

simoarpe:
> Not sure why don't we want to control visibility of Custom Menu item itself. The only two items whose visibility cannot be modified are the Main Settings item and the Custom Menu item itself. I'm pretty sure there will be a decent cohort of users who won't like it. It's still accessible via Settings even if it's hidden in the main menu.

Yes, that may happen. On the other hand they might hide the "Customize menu" item by mistake and then panic as they are unable to navigate back to that screen. So someone else will complain, may less savvy users.
If we receive many complaints we can add the option to hide the "Customize menu" itself at a second stage.
I don't have a strong position, but it would probably give me some headache if I wouldn't know that the menu is accessible also from Settings > Appearance > Customize menu and it disappears from main settings.

@aguscruiz , suggestions?

aguscruiz:
> > Not sure why don't we want to control visibility of Custom Menu item itself. The only two items whose visibility cannot be modified are the Main Settings item and the Custom Menu item itself. I'm pretty sure there will be a decent cohort of users who won't like it. It's still accessible via Settings even if it's hidden in the main menu.
> 
> Yes, that may happen. On the other hand they might hide the "Customize menu" item by mistake and then panic as they are unable to navigate back to that screen. So someone else will complain, may less savvy users. If we receive many complaints we can add the option to hide the "Customize menu" itself at a second stage. I don't have a strong position, but it would probably give me some headache if I wouldn't know that the menu is accessible also from Settings > Appearance > Customize menu and it disappears from main settings.
> 
> @aguscruiz , suggestions?

I agree. I would keep it as non-hidable for now and then if we get feedback asking to get rid of it.

Or we could do something like this on the settings page itself, to always indicate to users the alternative way to find this option in case they want to toggle it off. 

image

simoarpe:
Let's start simple by keeping it as is (not removable) and gather some feedback. If we receive some complaints about this I'll create an issue to adopt your suggested solution @aguscruiz.

android/brave_java_resources.gni
simoarpe: Absolutely! Thanks for noticing this. Moved drawable out of main app resources 👉 [`83de44c` (#31172)](https://github.com/brave/brave-core/pull/31172/commits/83de44c4548ff1a932d8ef12ee1177fb495e83a6)

android/java/org/chromium/chrome/browser/app/BraveActivity.java
simoarpe: I've introduced this:

```java
    public static Bundle populateBundle(
            final Resources resources,
            final Bundle bundle,
            final MVCListAdapter.ModelList menuItems,
            final MVCListAdapter.ModelList pageActions) {
        final long startTime = System.nanoTime(); // <--- START time.

        // Convert menu items to parcelable data.
        final ArrayList<MenuItemData> menuItemDataList = new ArrayList<>();
        populateMenuItemData(resources, menuItemDataList, menuItems);

        // Convert page action items to parcelable data.
        final ArrayList<MenuItemData> pageActionDataList = new ArrayList<>();
        populateMenuItemData(resources, pageActionDataList, pageActions);

        // Add data list to fragment bundle.
        bundle.putParcelableArrayList(KEY_MAIN_MENU_ITEM_LIST, menuItemDataList);
        bundle.putParcelableArrayList(KEY_PAGE_ACTION_ITEM_LIST, pageActionDataList);

        final long endTime = System.nanoTime();  // <--- END time.
        final double durationMs = (endTime - startTime) / 1_000_000.0;
        Log.i(TAG, String.format(Locale.US, "populateBundle: %.2fms (menu: %d, page: %d items)",
                durationMs, menuItemDataList.size(), pageActionDataList.size()));

        return bundle;
    }
```

On average I've obtained around `0.66ms`:

```
populateBundle: 0.66ms (menu: 16, page: 13 items)
```

On release builds this should be even quicker because all the optimizations.

SergeyZhukovsky (approved):
++

browser/ui/android/strings/android_brave_strings.grd
```diff
@@ -1064,6 +1064,21 @@ This file contains all "about" strings.  It is set to NOT be translated, in tran
...
+      </message>
+      <message name="IDS_CUSTOMIZE_PAGE_ACTIONS_SECTION_TITLE" desc="Title for page actions customization settings">
+        Page actions
+      </message>
+      <message name="IDS_CUSTOMIZE_PAGE_ACTIONS_SECTION_SUMMARY" desc="Description for page actions customization settings">
+        The following items will only be visible on the menu when supported by the current open webpage.
+      </message>
+        <message name="IDS_WEBAPP" desc="WebApp placeholder used in menu item customization preferences">
+        WebApp
+      </message>
```
mkarolin, line 1081: Is this one being used?

simoarpe: Yes it is 

Screenshot_20250915-210420

PR #31175, opened 2025-09-12
Add Brave Origin KeyedService / Factory

bbondy:
Resolves https://github.com/brave/brave-browser/issues/49214

This adds a keyed service and factory in support of upcoming Brave Origin work. 
Tracking issue: https://github.com/brave/brave-browser/issues/49212
The prototype draft to see how this all fits together is here: https://github.com/brave/brave-core/pull/30621

### DIFF
--- a/browser/brave_origin/brave_origin_service_factory.cc
+++ b/browser/brave_origin/brave_origin_service_factory.cc
@@ -1,0 +1,45 @@
+/* Copyright (c) 2025 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "brave/browser/brave_origin/brave_origin_service_factory.h"
+
+#include "base/no_destructor.h"
+#include "brave/components/brave_origin/brave_origin_service.h"
+#include "chrome/browser/profiles/profile.h"
+#include "chrome/browser/profiles/profile_selections.h"
+
+namespace brave_origin {
+
+// static
+BraveOriginService* BraveOriginServiceFactory::GetForProfile(Profile* profile) {
+  return static_cast<BraveOriginService*>(
+      GetInstance()->GetServiceForBrowserContext(profile, true));
+}
+
+// static
+BraveOriginServiceFactory* BraveOriginServiceFactory::GetInstance() {
+  static base::NoDestructor<BraveOriginServiceFactory> instance;
+  return instance.get();
+}
+
+BraveOriginServiceFactory::BraveOriginServiceFactory()
+    : ProfileKeyedServiceFactory(
+          "BraveOriginService",
+          ProfileSelections::BuildRedirectedInIncognito()) {}
+
+BraveOriginServiceFactory::~BraveOriginServiceFactory() = default;
+
+std::unique_ptr<KeyedService>
+BraveOriginServiceFactory::BuildServiceInstanceForBrowserContext(
+    content::BrowserContext* context) const {
+  // Pass the profile ID here
+  return std::make_unique<BraveOriginService>("");
+}
+
+bool BraveOriginServiceFactory::ServiceIsCreatedWithBrowserContext() const {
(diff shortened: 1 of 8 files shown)


browser/brave_origin/brave_origin_service_factory.cc
```diff
@@ -0,0 +1,45 @@
...
+// static
+BraveOriginServiceFactory* BraveOriginServiceFactory::GetInstance() {
+  static base::NoDestructor<BraveOriginServiceFactory> instance;
+  return instance.get();
+}
+
+BraveOriginServiceFactory::BraveOriginServiceFactory()
+    : ProfileKeyedServiceFactory(
+          "BraveOriginService",
+          ProfileSelections::BuildForRegularAndIncognito()) {}
```
DJAndries: should we use `ProfileSelections::BuildRedirectedInIncognito` here?

browser/brave_origin/brave_origin_service_factory.cc
```diff
@@ -0,0 +1,45 @@
...
+
+BraveOriginServiceFactory::~BraveOriginServiceFactory() = default;
+
+std::unique_ptr<KeyedService>
+BraveOriginServiceFactory::BuildServiceInstanceForBrowserContext(
+    content::BrowserContext* context) const {
+  // Pass the profile ID here
+  return std::make_unique<BraveOriginService>("");
```
DJAndries, line 38: will you pass the profile id in a future PR?

browser/brave_origin/brave_origin_service_factory.h
```diff
@@ -0,0 +1,48 @@
...
+template <typename T>
+class NoDestructor;
+}
+
+class Profile;
+
+namespace brave_origin {
+
+class BraveOriginService;
+enum class BraveOriginPolicyScope;
```
DJAndries: what's this forward declaration for?

components/brave_origin/brave_origin_service.h
```diff
@@ -0,0 +1,42 @@
...
+namespace brave_origin {
+
+// This keyed service will maintain the definitions/mappings of policies to
+// preferences.
+//
+// This is separate from BraveProfilePolicyProvider which handles the actual
+// integration with Chromium's policy framework.
+class BraveOriginService : public KeyedService {
+ public:
+  explicit BraveOriginService(const std::string& profile_id);
```
DJAndries, line 21: nit: could be either `const std::string` or `std::string_view`. this applies to other string references

bbondy: What do you mean? Passing in strings by const reference is very common.

browser/brave_origin/brave_origin_service_factory.cc
bbondy: Yes that's what the comment is indicating

browser/brave_origin/brave_origin_service_factory.cc
bbondy: I don't think so but not sure. @bridiver ?

components/brave_origin/brave_origin_service.h
DJAndries: Ah, I thought `std::string_view` was preferred based on feedback I've received in the past; I might be mistaken here

browser/brave_origin/brave_origin_service_factory.cc
DJAndries: nit: Perhaps it would make sense to add the TODO prefix for such comments for clarity. At first glance I was not sure whether the empty string was a mistake or a placeholder.

bbondy: I did have that at first but then presubmit wanted me to add an issue so that's why I removed it.

DJAndries (approved):

browser/brave_origin/brave_origin_service_factory.cc
bbondy: Thought I commented on this.  You're right we don't need a separate one for Incognito, it is updated in the latest.

bbondy:
if any sources.gni concerns, I'll adjust in a follow up.

PR #31185, opened 2025-09-12
Unify the Android Dynamic NTT for reuse on iOS

aseren:
The current PR is pre-requisite of Dynamic NTT on iOS as it allows to reuse android `New Tab Takeover WebUI` on iOS. The PR removes dependency of android `New Tab Takeover WebUI` from `ViewCounterService` which is not available on iOS. It also removes `current_wallpaper` base::Value cache in ViewCounterService which allows to save memory and make code cleaner.

## Test plan

* Start Brave browser on Android
* Check that dynamic NTT can shown
* Check that dynamic NTT navigates to target URL on user tap
* Check that metrics are sent

Resolves https://github.com/brave/brave-browser/issues/49216

### DIFF
--- a/android/java/org/chromium/chrome/browser/ntp/BraveNewTabPageLayout.java
+++ b/android/java/org/chromium/chrome/browser/ntp/BraveNewTabPageLayout.java
@@ -1255,19 +1255,22 @@ public class BraveNewTabPageLayout extends NewTabPageLayout
 
         boolean wasWallpaperShown = true;
         if (ntpImage instanceof Wallpaper && ((Wallpaper) ntpImage).isRichMedia()) {
-            setupSponsoredBackgroundContent();
-        } else if (ntpImage instanceof Wallpaper
-                && NTPImageUtil.isReferralEnabled()
-                && Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
-            setBackgroundImage(ntpImage);
-
-        } else if (UserPrefs.get(ProfileManager.getLastUsedRegularProfile())
-                        .getBoolean(BravePref.NEW_TAB_PAGE_SHOW_BACKGROUND_IMAGE)
-                && mSponsoredTab != null
-                && NTPImageUtil.shouldEnableNTPFeature()) {
-            setBackgroundImage(ntpImage);
+            setupSponsoredBackgroundContent((Wallpaper) ntpImage);
         } else {
-            wasWallpaperShown = false;
+            maybeResetSponsoredRichMediaBackground();
+
+            if (ntpImage instanceof Wallpaper
+                    && NTPImageUtil.isReferralEnabled()
+                    && Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+                setBackgroundImage(ntpImage);
+            } else if (UserPrefs.get(ProfileManager.getLastUsedRegularProfile())
+                            .getBoolean(BravePref.NEW_TAB_PAGE_SHOW_BACKGROUND_IMAGE)
+                    && mSponsoredTab != null
+                    && NTPImageUtil.shouldEnableNTPFeature()) {
+                setBackgroundImage(ntpImage);
+            } else {
+                wasWallpaperShown = false;
+            }
         }
 
         if (wasWallpaperShown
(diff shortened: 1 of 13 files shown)


components/new_tab_takeover/App.tsx
```diff
@@ -13,35 +13,63 @@ import {
   SponsoredRichMediaBackgroundInfo, SponsoredRichMediaBackground
 } from '../brave_new_tab_ui/containers/newTab/sponsored_rich_media_background'
 
+function sanitizeId(value: string | null): string | null {
+  if (!value || value.length === 0) {
+    return null;
+  }
+
+  if (!/^[0-9a-fA-F-]+$/.test(value)) {
```
tmancey, line 23: This regex seems overly permissive, can we not check the string is a valid v4 uuid?

components/ntp_background_images/browser/ntp_sponsored_images_data.cc
```diff
@@ -463,6 +463,18 @@ bool NTPSponsoredImagesData::IsSuperReferral() const {
...
 
+const Creative* NTPSponsoredImagesData::GetCreativeByInstanceId(
+    const std::string& creative_instance_id) const {
+  for (const Campaign& campaign : campaigns) {
+    for (const Creative& creative : campaign.creatives) {
+      if (creative.creative_instance_id == creative_instance_id) {
+        return &creative;
+      }
+    }
+  }
```
tmancey, line 476: The complexity is O(N * M). We dont have many campaigns/creatives for NTT, so seems fine for now. However, we could build a hash map after downloading campaigns.json so that complexity is O(1).

components/new_tab_takeover/App.tsx
aseren: No need to do full v4 uuid check, as this is only needed for security reasons to exclude non [ascii, `-`] characters.

components/ntp_background_images/browser/view_counter_service_unittest.cc
```diff
@@ -375,7 +374,7 @@ class ViewCounterServiceTest : public testing::Test {
     return view_counter_service_->GetCurrentWallpaperForDisplay();
   }
 
-  void VerifyGetCurrentBrandedWallpaperExpectation() {
+  void VerifyGetBrandedWallpaperExpectation() {
```
tmancey: Tests have `GetNewTabTakeoverWallpaper`, your thoughts on `VerifyGetNewTabTakeoverWallpaperExpectation`?

components/new_tab_takeover/App.tsx
tmancey: Could we use `base::IsStringASCII` if we are not too concerned it is a valid v4 uuid?

components/ntp_background_images/browser/view_counter_service_unittest.cc
aseren: Done

components/new_tab_takeover/App.tsx
aseren: `base::IsStringASCII` can be used from C++ code, but not from typescript/javascript. I've added a following comment show why this check was added:
```
  // Restrict input to alphanumeric characters and hyphens to prevent
  // potential injections.
```

components/ntp_background_images/browser/ntp_sponsored_images_data.cc
aseren: Raised an issue for this: https://github.com/brave/brave-browser/issues/49222. Also added TODO in code:
```
  // TODO(https://github.com/brave/brave-browser/issues/49222):
  // Use a map-based lookup for creatives to improve performance.
```

components/new_tab_takeover/App.tsx
tmancey: Ah yes 😆

android/java/org/chromium/chrome/browser/ntp/BraveNewTabPageLayout.java
```diff
@@ -1255,19 +1255,22 @@ private void showNTPImage(NTPImage ntpImage) {
...
-                && NTPImageUtil.shouldEnableNTPFeature()) {
-            setBackgroundImage(ntpImage);
+            setupSponsoredBackgroundContent((Wallpaper) ntpImage);
         } else {
-            wasWallpaperShown = false;
+            maybeResetSponsoredRichMediaBackground();
+
+            if (ntpImage instanceof Wallpaper
+                    && NTPImageUtil.isReferralEnabled()
+                    && Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
```
SergeyZhukovsky, line 1264: nit: we don't support `Build.VERSION_CODES.M` and lower already, so could be removed.

aseren: These checks were in code before my changes. I filed an issue to fix it in our codebase: https://github.com/brave/brave-browser/issues/49506. There are a lots of other places.

PR #31186, opened 2025-09-12
replace chromium folder icon with nala

aguscruiz:
Resolves https://github.com/brave/brave-browser/issues/49147

Follow up to previous PR to use Nala icon instead of default chromium one.

image

### DIFF
--- a/chromium_src/chrome/browser/ui/bookmarks/bookmark_utils.cc
+++ b/chromium_src/chrome/browser/ui/bookmarks/bookmark_utils.cc
@@ -50,11 +50,17 @@ void BraveToggleBookmarkBarState(content::BrowserContext* browser_context) {
   }                                                                        \
   void ToggleBookmarkBarWhenVisible_ChromiumImpl
 
+#if defined(TOOLKIT_VIEWS)
 // Override folder icon to use a fixed 20px size similar to saved groups.
 #define GetBookmarkFolderIcon GetBookmarkFolderIcon_ChromiumImpl
+#endif  // defined(TOOLKIT_VIEWS)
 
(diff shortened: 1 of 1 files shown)


aguscruiz:
@simonhong @sangwoo108 I created this PR in order to use the Nala icons for the bookmarks folder, but I'm not sure if it's the right way to do it, if you have any feedback it's always appreciated

sangwoo108:
Let me append a few commits to fix build and nits

chromium_src/chrome/browser/ui/bookmarks/bookmark_utils.cc
```diff
@@ -50,11 +50,17 @@ void BraveToggleBookmarkBarState(content::BrowserContext* browser_context) {
   }                                                                        \
   void ToggleBookmarkBarWhenVisible_ChromiumImpl
 
+#if defined(TOOLKIT_VIEWS)
```
simonhong, line 53: Do we need this guard? W/o this, there is a build failure?
I think it's not necessary because this file `bookmark_utils.cc` is in `//chrome/browser/ui/bookmarks:impl` target.
and that target is only added in deps of non-android `ui` target.

sangwoo108: Didn't know that but I think we'd better to have this because the upstream's GetBookmarkFolderIcon() is guarded with TOOLKIT_VIEWS in the source code.

simonhong: ok, their guard was added in 10y ago. I think it should be cleaned up.

simonhong (approved):

PR #31191, opened 2025-09-12
add support for NetworkManager

cmazakas:
Resolves https://github.com/brave/brave-browser/issues/49223

Unit tests for the PolkadotSubstrateRpc layer are coming up in a follow-up PR to this.

### DIFF
--- a/components/brave_wallet/browser/network_manager.cc
+++ b/components/brave_wallet/browser/network_manager.cc
@@ -751,7 +751,7 @@ const mojom::NetworkInfo* GetPolkadotTestnet() {
        {},
        0,
        {PolkadotTestnetRpcUrl()},
-       "DOT",
+       "WND",
        "Polkadot",
        12,
        coin,


components/brave_wallet/common/test_utils.h
```diff
@@ -71,6 +72,7 @@ inline constexpr mojom::KeyringId kAllKeyrings[] = {
     mojom::KeyringId::kZCashTestnet,
     mojom::KeyringId::kCardanoMainnet,
     mojom::KeyringId::kCardanoTestnet,
+    // TODO(cmazakas): Polkadot Keyring Testing
```
supermassive: Let's have an issue for polkadot keyrings and mention it TODO comments.

components/brave_wallet/common/common_utils_unittest.cc
```diff
@@ -653,7 +686,8 @@ TEST(CommonUtils, MakeAccountId) {
   EXPECT_DCHECK_DEATH(MakeAccountId(mojom::CoinType::ADA,
                                     mojom::KeyringId::kCardanoMainnet,
                                     mojom::AccountKind::kDerived, "0xabc"));
-  static_assert(AllCoinsTested<6>());
+  // TODO(cmazakas): Polkadot Keyring Testing
```
supermassive: issue link

components/brave_wallet/common/common_utils.h
```diff
@@ -72,6 +72,9 @@ std::string GetNetworkForCardanoKeyring(const mojom::KeyringId& keyring_id);
 std::string GetNetworkForCardanoAccount(const mojom::AccountIdPtr& account_id);
 
 bool IsPolkadotKeyring(mojom::KeyringId keyring_id);
+bool IsPolkadotNetwork(std::string_view network_id);
+std::string GetNetworkForPolkadotKeyring(const mojom::KeyringId& keyring_id);
+std::string GetNetworkForPolkadotAccount(const mojom::AccountIdPtr& account_id);
```
supermassive, line 77: `GetNetworkForPolkadotKeyring`
`GetNetworkForPolkadotAccount`
need unit tests

components/brave_wallet/browser/network_manager.cc
```diff
@@ -699,6 +721,44 @@ const mojom::NetworkInfo* GetCardanoTestnet() {
...
+       0,
+       {PolkadotTestnetRpcUrl()},
+       "DOT",
+       "Polkadot",
+       10,
+       coin,
+       GetSupportedKeyringsForNetwork(coin, chain_id)});
+  return network_info.get();
+}
+
```
supermassive, line 761: Can we stick to some test network from list in https://docs.polkadot.com/develop/networks/ ?
I think `Kusama` should be our choice at seems to be the closest one to mainnet. If yes please add scan url, and token info(decimals, symbol, etc)

components/brave_wallet/browser/network_manager.cc
```diff
@@ -67,25 +67,27 @@ constexpr auto kEip1559ForKnownChains =
...
+         // Bitcoin chains.
+         {mojom::kBitcoinMainnet, "bitcoin-mainnet"},
+
+         // Cardano chains.
+         {mojom::kCardanoMainnet, "cardano-mainnet"},
+         {mojom::kCardanoTestnet, "cardano-preprod"},
+
+         // Polkadot chains.
+         {mojom::kPolkadotMainnet, "polkadot-mainnet"},
+         {mojom::kPolkadotTestnet, "polkadot-testnet"}},
```
supermassive: If we pick Kusama as testnet please change it to `polkadot-kusama`. This would be part of alias in our proxy which should have a name similar to the entity being proxied.

components/brave_wallet/common/test_utils.h
cmazakas: You know, I tried making one but I don't know how to get these things to link up and recognize each other. See: https://github.com/brave/brave-browser/issues/49224 and https://github.com/brave/brave-browser/issues/49225

components/brave_wallet/common/common_utils_unittest.cc
cmazakas: same response here: https://github.com/brave/brave-core/pull/31191#discussion_r2352904170

components/brave_wallet/browser/network_manager.cc
cmazakas: I was thinking we'd be using Westend because the coins have no economic value so they're a true testnet we can just spam all day. I think Kusama is more for people doing parachain dev but I'm not sure.

@cypt4 what do you think here?

Technically, I've been pointing my local development binary to the testnet using the switches like this: `--polkadot-testnet-rpc-url=https://westend-rpc.polkadot.io`

components/brave_wallet/browser/network_manager.cc
cmazakas: Ah, interesting.

I chose the names I did because it looked like every other coin type we supported had names like that, and didn't include the name of the testnet itself in it. It just said "Bitcoin Testnet" and "ZCash Testnet" and stuff like that, so I assumed we would be doing the same thing here for Polkadot.

components/brave_wallet/common/common_utils.h
cmazakas: This is done!

components/brave_wallet/common/common_utils.cc
```diff
@@ -447,6 +485,11 @@ std::vector<mojom::KeyringId> GetSupportedKeyringsForNetwork(
       } else {
         return {mojom::KeyringId::kCardanoTestnet};
       }
+    case mojom::CoinType::DOT:
+      if (chain_id == mojom::kPolkadotMainnet) {
+        return {mojom::KeyringId::kPolkadotMainnet};
+      }
+      return {mojom::KeyringId::kPolkadotTestnet};
```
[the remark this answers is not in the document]
cmazakas: This should be address now. I didn't realize the flow analysis was so sensitive, and I've added the missing `else` clause.

components/brave_wallet/common/common_utils_unittest.cc
supermassive: should be `TODO(https://github.com/brave/brave-browser/issues/49225): Polkadot Keyring Testing`

supermassive: So presubmit checker would be happy

components/brave_wallet/browser/network_manager.cc
supermassive: There is `cardano-preprod` proxy alias for kCardanoTestnet. And all other proxy aliases above relate to some real network.
Something like this should be eventually picked for Polkadot.

components/brave_wallet/common/common_utils_unittest.cc
cmazakas: Oh, interesting.
I got confused because there's other issues like with the Cardano P3A one that seems to pass fine.
Alright, well, I've confirmed my fix passes locally. Thanks for the help!

components/brave_wallet/browser/network_manager.cc
cmazakas: I see. Alright, we need to settle on a real testnet for Polkadot as a team then, I'd say.
I vote Westend because it's a true fake testnet whereas Kusama just has cheaper coins, but they still have economic value.
Should I go ahead and make an issue for this and we can consider the issue in this PR resolved? Or should we try and just address it now as a team?

supermassive: I'm OK with either Westend or Kusama. Let's just pick one and use its name here and in its props in `GetPolkadotTestnet()`

nuo-xu (approved):
iOS++

components/brave_wallet/browser/network_manager.cc
cmazakas: Alright, I'm gonna stick with Westend for now because I've already hard-coded some stuff related to it in other places, like the PolkadotKeyring and how we derive keypairs.

components/brave_wallet/common/test_utils.h
cmazakas: This has been fixed!

components/brave_wallet/common/common_utils_unittest.cc
cmazakas: This is fixed!

components/brave_wallet/browser/network_manager.cc
cmazakas: I've chosen westend as previously discussed in other comments; this should be addressed!

components/brave_wallet/browser/network_manager.cc
cmazakas: This should be fixed!

components/brave_wallet/browser/network_manager.cc
```diff
@@ -699,6 +721,44 @@ const mojom::NetworkInfo* GetCardanoTestnet() {
...
+  const auto* chain_id = mojom::kPolkadotTestnet;
+
+  static base::NoDestructor<mojom::NetworkInfo> network_info(
+      {chain_id,
+       "Polkadot Westend",
+       {"https://westend.subscan.io/"},
+       {},
+       0,
+       {PolkadotTestnetRpcUrl()},
+       "DOT",
```
supermassive: https://docs.polkadot.com/develop/networks/#test-networks
~"DOT"~
"WND"

supermassive (approved):

components/brave_wallet/common/common_utils.cc
```diff
@@ -447,6 +485,10 @@ std::vector<mojom::KeyringId> GetSupportedKeyringsForNetwork(
       } else {
         return {mojom::KeyringId::kCardanoTestnet};
       }
+    case mojom::CoinType::DOT:
+      return {(chain_id == mojom::kPolkadotMainnet
+                   ? mojom::KeyringId::kPolkadotMainnet
+                   : mojom::KeyringId::kPolkadotTestnet)};
```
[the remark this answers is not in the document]
kdenhartog: this looks like it should be addressed

components/brave_wallet/common/common_utils.cc
```diff
@@ -289,6 +289,32 @@ bool IsPolkadotKeyring(mojom::KeyringId keyring_id) {
...
+}
+
+std::string GetNetworkForPolkadotKeyring(const mojom::KeyringId& keyring_id) {
+  if (keyring_id == mojom::KeyringId::kPolkadotMainnet) {
+    return mojom::kPolkadotMainnet;
+  }
+  if (keyring_id == mojom::KeyringId::kPolkadotTestnet) {
+    return mojom::kPolkadotTestnet;
+  }
+  NOTREACHED();
```
kdenhartog, line 309: nit: Is NOTREACHED fatal and best practice for cases like this? I was thinking the option return type with return null and bubbling up would make more sense for a function like this

cmazakas: Good question, actually.

I wrote this function the way I did to keep consistency with how this is implemented in other places, e.g. `GetNetworkForCardanoKeyring` and `GetNetworkForBitcoinKeyring`.

My impression was that it was written this way to check developer bugs, which are unexpected and show that we 112% need to refactor our code to be correct.

Maybe @cypt4 or @supermassive can chime in here.

In my own personal opinions, optionals are better in the cases where we'd either want to handle the error further up or we actually expect a failure/error. In this case, we certainly don't expect an error/failure as we should just be calling this function ourselves with the correct input. As for where to best handle the error, that's a different matter. Personally, I don't mind just killing the process and blowing up our debug logs.

components/brave_wallet/browser/network_manager.cc
cmazakas: This is done! Thanks for the keen eye.

components/brave_wallet/common/common_utils.cc
supermassive: GetNetworkForPolkadotKeyring
GetNetworkForBitcoinKeyring
GetNetworkForCardanoKeyring
etc
are to be called from a corresponding coin area. That's OK to fail fast(in tests or nightly) when this gets violated. Returning nullopt would just hide potential bug

cmazakas: Alright, I'm gonna mark this convo as resolved so I can merge the PR.

If we want to change this code, I think we should do it via a CodeHealth issue.

PR #31196, opened 2025-09-12
Fix intermittent failure WalletPanelUIBrowserTest.CustomNetworkInSettings

bbondy:
This ensures that the value will be loaded and there will be no timing issues.

Resolves https://github.com/brave/brave-browser/issues/34765

### DIFF
--- a/browser/ui/webui/brave_wallet/wallet_panel_ui_browsertest.cc
+++ b/browser/ui/webui/brave_wallet/wallet_panel_ui_browsertest.cc
@@ -324,6 +324,8 @@ IN_PROC_BROWSER_TEST_F(WalletPanelUIBrowserTest, CustomNetworkInSettings) {
 
   // Chain name for Neon EVM changes to 'Custom Network' in wallet.
   ActivateWalletTab();
+  wallet()->GetController().Reload(content::ReloadType::NORMAL, true);
+  EXPECT_TRUE(WaitForLoadStop(wallet()));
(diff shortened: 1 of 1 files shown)


browser/ui/webui/brave_wallet/wallet_panel_ui_browsertest.cc
```diff
@@ -324,6 +324,8 @@ IN_PROC_BROWSER_TEST_F(WalletPanelUIBrowserTest, CustomNetworkInSettings) {
 
   // Chain name for Neon EVM changes to 'Custom Network' in wallet.
   ActivateWalletTab();
+  wallet()->GetController().Reload(content::ReloadType::NORMAL, true);
```
SergeyZhukovsky, line 327: should `Reload` be triggered here? It seems like not necessary.

bbondy: Not necessary but it will fix the timing issue. I think the purpose of that test is not for the observer but maybe I'm wrong?

bbondy: The test file already does the reload btw for another test

SergeyZhukovsky: I would think the wait would be enough, but  fine if it passes.

SergeyZhukovsky (approved):
++

PR #31201, opened 2025-09-15
[AI Chat] Add DetachedTabInterface class

fallaciousreasoning:
Introduces a new DetachedTabInterface class that implements TabInterface for tabs that are not attached to a browser window. This provides a standalone tab implementation for use cases where tabs need to exist independently of the main browser UI.

Resolves


fallaciousreasoning:
Closing this - I don't think its the right approach

PR #31202, opened 2025-09-15
Add pref to control "Hide vertical tabs completely when minimized" feature

sangwoo108:
This PR adds a new preference to control the "Hide vertical tabs completely when collapsed" feature.

image

part of https://github.com/brave/brave-browser/issues/49115

### DIFF
--- a/app/brave_settings_strings.grdp
+++ b/app/brave_settings_strings.grdp
@@ -157,7 +157,7 @@
     Show title bar
   </message>
   <message name="IDS_SETTINGS_APPEARANCE_SETTINGS_TABS_HIDE_COMPLETELY_WHEN_COLLPASED" desc="The label for hiding vertical tabs completely when collapsed" formatter_data="webui=BraveSettings">
-    Hide completely when collapsed
+    Hide completely when minimized
   </message>
   <message name="IDS_SETTINGS_APPEARANCE_SETTINGS_TABS_USE_FLOATING_VERTICAL_TABS" desc="The label for enabling floating mode for vertical tab strip">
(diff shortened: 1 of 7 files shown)


app/brave_settings_strings.grdp
```diff
@@ -156,6 +156,9 @@
   <message name="IDS_SETTINGS_APPEARANCE_SETTINGS_TABS_SHOW_TITLE_BAR" desc="The label for showing window title when vertical tab strip is enabled.">
     Show title bar
   </message>
+  <message name="IDS_SETTINGS_APPEARANCE_SETTINGS_TABS_HIDE_COMPLETELY_WHEN_COLLPASED" desc="The label for hiding vertical tabs completely when collapsed" formatter_data="webui=BraveSettings">
+    Hide completely when collapsed
```
mkarolin: Just wanted to note that the button that "collapses" the verticals tabs bar has the tooltip text that says `Minimize tabs`. We should probably use consistent terminology everywhere (either collapse or minimize). cc: @rebron

mkarolin (approved):
`strings`++

app/brave_settings_strings.grdp
sangwoo108: Nice, let me change it to "Hide completely when minimized"

mkarolin: @sangwoo108 I think we'd have to change the other option `Expand vertical tabs panel on mouseover when collapsed` to `Expand vertical tabs panel on mouseover when minimized` as well.

sangwoo108: That'd be even better.

PR #31209, opened 2025-09-15
Add Brave Origin Local State pref

bbondy:
Brave Origin will use a local state key to set the Brave Origin policy values the user wants to use so that policies can be setup that way on startup.

Resolves https://github.com/brave/brave-browser/issues/49285


components/brave_origin/brave_origin_constants.h
```diff
@@ -0,0 +1,16 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#ifndef BRAVE_COMPONENTS_BRAVE_ORIGIN_BRAVE_ORIGIN_CONSTANTS_H_
+#define BRAVE_COMPONENTS_BRAVE_ORIGIN_BRAVE_ORIGIN_CONSTANTS_H_
+
+namespace brave_origin {
+
+// Base key for all profile-scoped BraveOrigin policy preferences in local state
+inline constexpr char kBraveOriginPolicies[] = "brave_origin_policies";
```
DJAndries: consider using a name with a dotted prefix such as `brave.origin.policies` or `origin.policies`. If we decide to add more Origin-related prefs later, then it would be nice if they were grouped accordingly.

components/brave_origin/brave_origin_constants.h
```diff
@@ -0,0 +1,16 @@
+/* Copyright (c) 2025 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#ifndef BRAVE_COMPONENTS_BRAVE_ORIGIN_BRAVE_ORIGIN_CONSTANTS_H_
+#define BRAVE_COMPONENTS_BRAVE_ORIGIN_BRAVE_ORIGIN_CONSTANTS_H_
```
DJAndries: should we rename this to pref_names.h as per convention?

components/brave_origin/brave_origin_constants.h
bbondy: updated in the latest

components/brave_origin/brave_origin_constants.h
bbondy: updated in the latest

bbondy:
If any DEPS concerns I'll do it in a follow up. Merging.

PR #31211, opened 2025-09-15
Parity scale codec

cmazakas:
Resolves https://github.com/brave/brave-browser/issues/49718

Note: I had to manually create the BUILD.gn for the rustversion crate, as that exists upstream in Chromium and isn't available outside of their `//third_party` so I had to adjust the visiblity rules as well.


cmazakas:
@brave/rust-reviewers @bridiver can we get some mobility on this?

third_party/rust/rustversion/v1/BUILD.gn
```diff
@@ -0,0 +1,90 @@
...
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+group("lib") {
+  public_deps = [ "//third_party/rust/rustversion/v1:lib" ]
+}
+
+# copy-paste from upstream and modify to add the proper buildrs_support
+# All deps should be changed from lib -> buildrs_support
+# TODO(cmazakas) find a better way to handle this, as Brian alludes to in proc_macro2/v1/BUILD.gn
+import("//build/rust/cargo_crate.gni")
```
bridiver: on second thought lets patch out the build script similar to slab and remove the build_deps in gnrt_config.toml. We don't need anything the build script is doing and this is just a maintenance headache

cmazakas: This has been addressed!

PR #31213, opened 2025-09-15
fix(privacy): Crash removing external filter list on iOS 26

StephenHeaps:
- On iOS 26 (regardless on which SDK built the app), the Binding `ForEach` with a `Toggle` in the rows will crash when removing the last item after the `onDelete(perform:)` action is executed due to an index out of bounds access. 
    - This can be reproduced in a small sample project: https://gist.github.com/StephenHeaps/9449c1b77263db516018f3fc48ca6c2a
- The workaround to resolve the crash is using the non-Binding `ForEach` initializer for the external filter lists.

Resolves https://github.com/brave/brave-browser/issues/49277

### Testplan

1. In Settings > Shields & Privacy > Content Filtering, add an external filter list by URL, ex: https://raw.githubusercontent.com/StephenHeaps/adblock-lists/refs/heads/main/hide-brave-site.txt
    - Note: This list has 1 rule which will render `brave.com` as blank / hide the root element: 
    `brave.com##html:matches-attr(data-theme)`
2. Swipe to delete the external list
3. Verify the app does not crash and the list is removed.
4. Re-add the list in Settings > Shields & Privacy > Content Filtering.
5. Visit `brave.com`, verify site is blank with shields up.
6. Verify the toggle still works as expected:
    1. Disable it, leave the Content Filtering screen, re-visit Content Filtering screen. Verify toggle is still disabled.
    2. Visit `brave.com`, verify site is no longer blank
        - Note: There is a delay between disabling and/or removing the list and changes being applied. If using the list from step 1, brave.com may still be blank immediately after but the list will be removed and engines are re-compiled shortly.


kylehickinson (approved):

kjozwiak:
Verification **PASSED** on `iPhone 11` running `iOS 26` using the following build(s):

```
Brave | 1.84.71 Chromium: 140.0.7339.133 (Official Build) nightly (64-bit) 
--- | ---
Revision | f96e70e765dc...
OS | iOS
```

Using the STR/Cases outlined via https://github.com/brave/brave-core/pull/31213#issue-3419574098, reproduced the original crash with `1.84.67 Chromium: 140.0.7339.133` when swiping/removing a URL list as per the following:

https://github.com/user-attachments/assets/525bfad7-d15f-40d0-a009-5dd72784a3f0

Using the same STR/Cases as above, ensured that Brave wasn't crashing when removing URLs via `content filtering` as per the following:

* ensured that removing `hide-brave-site.txt` via swiping didn't crash `Brave`
* ensured that removing `hide-brave-site.txt` via swiping and then tapping on `Delete` didn't crash `Brave`
* ensured that adding `hide-brave-site.txt` sets the list as `enabled` as default
  * ensured that the `enabled` state persists after several browser restarts
* ensured that you can disable `hide-brave-site.txt` via the toggle without any issues/crashes
  * ensured that the `disabled` state persists after several browser restarts
* ensured that `brave.com` appears is blank when `hide-brave-site.txt` is disabled (can take ~1 minute or more to apply rule)
* ensured that re-enabling `hide-brave-site.txt` re-displays `brave.com` (can take ~1 minute or more to apply rule)

PR #31214, opened 2025-09-15
[PSST] Create status definition enum

vadimstruts:
Resolves https://github.com/brave/brave-browser/issues/49286


components/psst/browser/content/psst_tab_web_contents_observer.cc
```diff
@@ -199,8 +201,10 @@ void PsstTabWebContentsObserver::OnPolicyScriptResult(
     return;
```
bridiver, line 206: I think we also need to call `UpdateTasks` here for `!script_result_parsed`, but not for `!ShouldInsertScriptForPage`

PR #31219, opened 2025-09-16
`BindInterface()` wiring for `mojom::Authentication`

szilardszaloki:
Add `BindInterface()` wiring for `brave_account::mojom::Authentication` in `BraveAccountService`, allowing the UI to bind and call the `Authentication` remote in the registration flow.

Resolves https://github.com/brave/brave-browser/issues/49296, resolves https://github.com/brave/brave-browser/issues/49058.

Follow-ups:
- https://github.com/brave/brave-browser/issues/49349
- https://github.com/brave/brave-browser/issues/49350


kdenhartog (approved):
no concerns from me

fallaciousreasoning (approved):

kylehickinson (approved):

darkdh (approved):

components/brave_account/brave_account_service.cc
```diff
@@ -20,6 +20,11 @@ BraveAccountService::BraveAccountService(
 
 BraveAccountService::~BraveAccountService() = default;
 
+void BraveAccountService::BindInterface(
```
bridiver, line 23: nit `Bind`. There are a lot of variations, but just `Bind` is the most common outside webui

bridiver: actually I should check recent usages in upstream, the `Bind` ones might be older

bridiver: nvm, this is fine. We might as well standardize on `BindInterface` since it is used in webui anyway

components/brave_account/resources/brave_account_create_dialog.ts
```diff
@@ -162,6 +162,29 @@ export class BraveAccountCreateDialogElement extends CrLitElement {
     this.isCheckboxChecked = detail.checked
   }
 
+  protected async onCreateAccountButtonClicked() {
```
bridiver, line 172: We really need to get tests working for this before we keep adding ts code. It's fine to leave brave_account_browser_proxy.ts changes in this PR, but I think we need to move the other ts changes out. I don't think they really belong in this PR anyway

ios/browser/ui/webui/brave_account/brave_account_ui.mm
```diff
@@ -52,14 +54,28 @@
...
+  web_ui->GetWebState()->GetInterfaceBinderForMainFrame()->AddInterface(
+      base::BindRepeating(static_cast<PasswordStrengthMeter>(
+                              &BraveAccountUIBase::BindInterface),
                           base::Unretained(this)));
 
   web_ui->AddMessageHandler(std::make_unique<BraveAccountUIMessageHandler>());
 }
 
 BraveAccountUI::~BraveAccountUI() {
+  web_ui()->GetWebState()->GetInterfaceBinderForMainFrame()->RemoveInterface(
```
bridiver, line 76: This doesn't seem like it should be necessary and seems fragile and error prone anyway. @kylehickinson do you know why this is here? Not an issue for this PR, but I want to look at this as a follow-up.

szilardszaloki: https://github.com/brave/brave-core/pull/29549#discussion_r2143396031

szilardszaloki: https://github.com/brave/brave-browser/issues/46677

bridiver: I guess this is from upstream so 🤷

components/brave_account/resources/brave_account_create_dialog.ts
szilardszaloki: PTAL: https://github.com/brave/brave-core/pull/31241

components/brave_account/resources/brave_account_create_dialog.ts
```diff
@@ -162,6 +162,29 @@ export class BraveAccountCreateDialogElement extends CrLitElement {
...
+    try {
+      const blindedMessage = this.registration.start(this.password)
```
bridiver, line 174: this seems to violate the one user action == one method principle. Why does this sequence of events happen in js instead inside the service?

components/brave_account/brave_account_service.cc
szilardszaloki: The name `BindInterface` comes from the `InterfaceRegistrationHelper` by the way — https://source.chromium.org/chromium/chromium/src/+/main:content/public/browser/web_ui_browser_interface_broker_registry.h;l=41:
```cpp
auto* concrete_controller = controller->GetAs<ControllerType>();

DCHECK(concrete_controller)
    << "The requesting WebUIController is of a different type.";

concrete_controller->BindInterface(std::move(receiver));
```
where `ControllerType` is `BraveAccountUI` in this case.

components/brave_account/resources/brave_account_create_dialog.ts
szilardszaloki: The reason this happens here is that both `registration.start()` and `registration.finish()` invoke the OPAQUE protocol in our WASM (compiled from Rust), and so the flow must run in the renderer to manage the transient cryptographic state — the service only transports the two server round trips (`registerInitialize`/`registerFinalize`).

bridiver: I see, maybe comment here as placeholders?

bridiver (approved):

components/brave_account/resources/brave_account_create_dialog.ts
szilardszaloki: Added comment here: https://github.com/brave/brave-core/pull/31219/commits/373b6b48d3168e893715bde20699494298f410b9
```
// The reason this happens here (rather than in BraveAccountService) is that
// both `registration.start()` and `registration.finish()` invoke the OPAQUE
// protocol in our WASM (compiled from Rust), and so the flow must run in the
// renderer to manage the transient cryptographic state — the service only
// transports the two server round trips
// (`registerInitialize`/`registerFinalize`). We'll revisit handling this
// through Mojo in C++ if that proves practical.
```

components/brave_account/resources/brave_account_create_dialog.ts
szilardszaloki: As for tests, we don't have Mocha in Brave yet. I flagged this for @fmarier as well. I suggested adding the registration/login flow to QA's manual test plan for now. Once the login flow lands (this PR is registration only), I'll start enabling Mocha for Brave — at which point we can drop these cases from QA's manual plan.

szilardszaloki: Filed it here: https://github.com/brave/brave-browser/issues/49349 — I'll add more details later.

components/brave_account/resources/brave_account_create_dialog.ts
szilardszaloki: https://github.com/brave/brave-browser/issues/49350

PR #31221, opened 2025-09-16
[Nala]: Fix scheme relative loading of variables

fallaciousreasoning:
Resolves https://github.com/brave/brave-browser/issues/49298

### DIFF
--- a/ui/webui/resources/leo/web_components.ts
+++ b/ui/webui/resources/leo/web_components.ts
@@ -31,12 +31,14 @@ setIconBasePath(`${scheme}resources/brave-icons`)
 
 // In Chromium UI Nala variables haven't necessarily been included. We
 // make sure the variables are imported so the controls look correct.
-const variablesUrl = `//resources/brave/css/nala.css`
-const variablesLink = document.querySelector(`link[rel=stylesheet][href$="${variablesUrl}"]`)
+const variablesUrl = `resources/brave/css/nala.css`
+const variablesLink = document.querySelector(`link[rel=stylesheet][href$="//${variablesUrl}"]`)
 if (!variablesLink) {
   const link = document.createElement('link')
   link.setAttribute('rel', 'stylesheet')
-  link.setAttribute('href', variablesUrl)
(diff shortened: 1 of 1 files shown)


ui/webui/resources/leo/web_components.ts
```diff
@@ -31,12 +31,14 @@ setIconBasePath(`${scheme}resources/brave-icons`)
...
-const variablesLink = document.querySelector(`link[rel=stylesheet][href$="${variablesUrl}"]`)
+const variablesUrl = `resources/brave/css/nala.css`
+const variablesLink = document.querySelector(`link[rel=stylesheet][href$="//${variablesUrl}"]`)
 if (!variablesLink) {
   const link = document.createElement('link')
   link.setAttribute('rel', 'stylesheet')
-  link.setAttribute('href', variablesUrl)
+  // When we create a link to the variables file, use `scheme` rather than a
+  // scheme relative url, to ensure the link loads correctly for extensions.
+  link.setAttribute('href', `${scheme}${variablesUrl}`)
```
fallaciousreasoning, line 41: So the problem here was that on `chrome-extension://` urls we'd try and load the CSS variables from 

`chrome-extension://resources/brave/css/nala.css` which doesn't exist.

ui/webui/resources/leo/web_components.ts
```diff
@@ -31,12 +31,14 @@ setIconBasePath(`${scheme}resources/brave-icons`)
 
 // In Chromium UI Nala variables haven't necessarily been included. We
 // make sure the variables are imported so the controls look correct.
-const variablesUrl = `//resources/brave/css/nala.css`
-const variablesLink = document.querySelector(`link[rel=stylesheet][href$="${variablesUrl}"]`)
+const variablesUrl = `resources/brave/css/nala.css`
+const variablesLink = document.querySelector(`link[rel=stylesheet][href$="//${variablesUrl}"]`)
```
fallaciousreasoning, line 35: Note: This is an `endsWith` attribute selector, so it doesn't care about the scheme

PR #31222, opened 2025-09-16
Setup Brave Origin Profile ID

bbondy:
The profile ID is used in local state to scope the preferences to the specific profile

Resolves https://github.com/brave/brave-browser/issues/49299


components/brave_policy/brave_profile_policy_provider.h
```diff
@@ -31,10 +31,15 @@ class BraveProfilePolicyProvider : public policy::ConfigurationPolicyProvider {
...
   bool IsFirstPolicyLoadComplete(policy::PolicyDomain domain) const override;
 
+  void SetProfileID(const std::string& profile_id);
+
  private:
   policy::PolicyBundle LoadPolicies();
 
   bool first_policies_loaded_ = false;
+  // This tracks whether or not the policy definitions have been loaded
+  bool policies_ready_ = false;
```
bbondy: I'll use this in the next PR. I just wanted to guard the RefreshPolicies call and not forget about that for now.

browser/brave_policy/brave_policy_profile_id.cc
```diff
@@ -0,0 +1,27 @@
...
+#include "brave/components/brave_policy/brave_profile_policy_provider.h"
+#include "chrome/browser/profiles/profile.h"
+#include "components/policy/core/common/configuration_policy_provider.h"
+
+namespace brave_policy {
+
+// This is the definition of a declaration a chromium_src for setting
+// the profile ID on a profile's BraveProfilePolicyProvider.
+// The declaration is in
+// chromium_src/chrome/browser/policyprofile_policy_connector_builder.cc
```
DJAndries: Why is the declaration in chromium_src instead of brave_policy_profile_id.h?

components/brave_policy/brave_profile_policy_provider.cc
```diff
@@ -56,6 +56,16 @@ policy::PolicyBundle BraveProfilePolicyProvider::LoadPolicies() {
   return bundle;
 }
 
+void BraveProfilePolicyProvider::SetProfileID(const std::string& profile_id) {
```
DJAndries, line 59: Slightly unrelated, but curious to hear your thoughts on moving the contents of `//brave/components/brave_policy` to `//brave/components/policy`

browser/brave_origin/brave_origin_service_factory.cc
```diff
@@ -154,8 +155,10 @@ BraveOriginServiceFactory::~BraveOriginServiceFactory() = default;
 std::unique_ptr<KeyedService>
 BraveOriginServiceFactory::BuildServiceInstanceForBrowserContext(
     content::BrowserContext* context) const {
-  // Pass the profile ID here
-  return std::make_unique<BraveOriginService>("");
+  Profile* profile = Profile::FromBrowserContext(context);
+  std::string profile_id =
+      GetProfileId(profile->GetPath().BaseName().AsUTF8Unsafe());
+  return std::make_unique<BraveOriginService>(profile_id);
```
DJAndries, line 159: feel free to ignore this comment since it only pertains to style, but you could consider passing the profile path, and adding a call to `GetProfileID` inside the constructor to slightly simplify the factory.

components/brave_origin/profile_id.h
```diff
@@ -0,0 +1,18 @@
...
+
+#ifndef BRAVE_COMPONENTS_BRAVE_ORIGIN_PROFILE_ID_H_
+#define BRAVE_COMPONENTS_BRAVE_ORIGIN_PROFILE_ID_H_
+
+#include <string>
+
+namespace brave_origin {
+
+// Creates a base64url encoded profile identifier for stable, safe identifier
+std::string GetProfileId(const std::string& profile_path_base_name);
```
DJAndries: consider accepting a `base::FilePath` instead of a string to simplify call sites

browser/brave_policy/brave_policy_profile_id.cc
```diff
@@ -0,0 +1,27 @@
...
+void SetBraveProfilePolicyProviderProfileID(
+    policy::ConfigurationPolicyProvider* provider,
+    content::BrowserContext* context) {
+  Profile* profile = Profile::FromBrowserContext(context);
+  auto profile_id = ::brave_origin::GetProfileId(
+      profile->GetPath().BaseName().AsUTF8Unsafe());
```
DJAndries: `BrowserContext` also has `GetPath`, so we might not need `Profile::FromBrowserContext`

browser/brave_policy/brave_policy_profile_id.cc
bbondy: This is a trick to work around needing a circular dependency where chromium_src would need to depend on the brave origin component

components/brave_policy/brave_profile_policy_provider.cc
bbondy: That's actually how I had it originally but BJ suggested we split it out from the stuff that's there and move it to brave_policy instead.

components/brave_origin/profile_id.h
bbondy: Good call. Update in the latest

browser/brave_policy/brave_policy_profile_id.cc
bbondy: Nice. Doing that allowed me to move it into components too and avoid the browser/ BUILD.gn that I had originally.

components/brave_policy/brave_policy_profile_id.cc
```diff
@@ -0,0 +1,25 @@
...
+#include "brave/components/brave_policy/brave_profile_policy_provider.h"
+#include "components/policy/core/common/configuration_policy_provider.h"
+#include "content/public/browser/browser_context.h"
+
+namespace brave_policy {
+
+// This is the definition of a declaration a chromium_src for setting
+// the profile ID on a profile's BraveProfilePolicyProvider.
+// The declaration is in
+// chromium_src/chrome/browser/policyprofile_policy_connector_builder.cc
```
DJAndries: nit: typo

browser/brave_policy/brave_policy_profile_id.cc
DJAndries: What if we placed profile_id.h/cc in a separate target (i.e. `//brave/components/brave_origin:profile_id` or `//brave/components/brave_origin:util`) and added a dependency to that target?

bbondy: It's beyond that, BJ asked for it to be like this specifically. I did do that same thing in the past though. He doesn't want any components as a dep in chromium_src layer.

components/brave_policy/brave_policy_profile_id.cc
bbondy: updated in the latest

PR #30743, opened 2025-08-21
[AI Chat] [Content Agent] Introduce other content agent tools

petemill:
This introduces the rest of the tools for manipulating content via the Actor service (click, drag, scroll, select, type, navigate history).

In https://github.com/brave/brave-core/pull/31224 we introduced a single content agent tool (the navigation tool) and in this PR we repeat the same pattern for the rest of the tools.

These tools commonly define "target" parameters - elements on which to act. See `target_util.h`, `target_util_unittest.cc` and `target_test_util.h` for all the common functionality around "target".

This feature continues, most recently, from these branches:
- https://github.com/brave/brave-core/pull/31223
- https://github.com/brave/brave-core/pull/31224
- Tool and ToolProvider interfaces in components/ai_chat
- Tool loop in ConversationHandler
- Tool sending in EngineConsumer implementations
- An experimental profile which will be the only place where content agent tools are provided

It does not include other parallel branches:
- Task UI - groups tool uses and previews the tab being worked on, splitting details and a current summary behind two tabs within the conversation response.
- Annotated Page Content result - expands the AnnotatedPageContent representation of the WebContent to AI Chat ContentBlocks as result of the content agent tool action.
- Related tools that make multi-step flows (such as content agent conversations) better:
  - A TODO tool that is, for now, only used for CONTENT_AGENT conversations
  - AssistantDetailStorageTool so that the AI can store internal data for a conversation that may be wiped when large tool results are omitted from the context

**I didn't split this PR up any further because the tools all repeat the same pattern and there's little to gain merging them until they have all been reviewed and changed together. 1 tool and its pattern was already reviewed separately in #31224.**

**The bulk of the size is due to there being multiple simple tools but each one getting it's own set of unit tests to ensure we're parsing the inputs, handling bad input and creating the actor framework requests correctly. They are quite repetitive and, since we've already reviewed one tool in #31224, it would be great to review the rest of them, which are using the same pattern, at the same time.**

PR #31223, opened 2025-09-16
Introduce ConversationCapability with types CHAT and CONTENT_AGENT

petemill:
Introduces the concept of conversation "capability" with 2 options: CHAT and CONTENT_AGENT, passing to the Conversation API. This allows:
- the system prompt to be adjusted and the built-in (and competing) RAG tools to be disabled
- tools to be filtered based on conversation type

This is experimental, not user-toggleable and only active (for all conversations) in the separate agent profile See the following integration branch for wider context:
-  https://github.com/brave/brave-core/pull/30743


browser/ai_chat/ai_chat_service_factory.cc
```diff
@@ -85,7 +86,12 @@ AIChatServiceFactory::BuildServiceInstanceForBrowserContext(
   tool_provider_factories.push_back(
       std::make_unique<BrowserToolProviderFactory>());
 
-  return std::make_unique<AIChatService>(
+#if BUILDFLAG(ENABLE_BRAVE_AI_CHAT_AGENT_PROFILE)
+  bool is_actor_allowed = features::IsAIChatAgentProfileEnabled() &&
+                          Profile::FromBrowserContext(context)->IsAIChatAgent();
+#endif
```
yrliou, line 92: Let's combine this section with `service->SetIsContentAgentAllowed(is_actor_allowed);` below in one build flag guard.

components/ai_chat/core/browser/ai_chat_service.h
```diff
@@ -191,6 +191,9 @@ class AIChatService : public KeyedService,
   void BindObserver(mojo::PendingRemote<mojom::ServiceObserver> ui,
                     BindObserverCallback callback) override;
 
+  bool GetIsContentAgentAllowed();
```
yrliou: bool GetIsContentAgentAllowed() const;

components/ai_chat/core/browser/engine/conversation_api_client.cc
```diff
@@ -278,11 +279,22 @@ std::string ConversationAPIClient::CreateJSONRequestBody(
...
+    mojom::ConversationCapability conversation_capability,
     const std::optional<std::string>& model_name,
     const bool is_sse_enabled) {
   base::Value::Dict dict;
 
+  static const base::NoDestructor<
+      std::map<mojom::ConversationCapability, std::string>>
+      kCapabilityMap(
+          {{mojom::ConversationCapability::CHAT, "chat"},
+           {mojom::ConversationCapability::CONTENT_AGENT, "content_agent"}});
```
yrliou:
```
static constexpr auto kCapabilityMap = base::MakeFixedFlatMap<
      mojom::ConversationCapability, std::string_view>({
      {mojom::ConversationCapability::CHAT, "chat"},
      {mojom::ConversationCapability::CONTENT_AGENT, "content_agent"}
});
```

components/ai_chat/core/browser/engine/conversation_api_client_unittest.cc
```diff
@@ -883,6 +886,9 @@ TEST_F(ConversationAPIUnitTest,
         const std::string* model = dict->FindString("model");
         EXPECT_TRUE(model);
         EXPECT_EQ(*model, override_model_name);
+        const std::string* capability = dict->FindString("capability");
+        EXPECT_TRUE(capability);
+        EXPECT_EQ(*capability, "chat");
```
yrliou: Seems like a bit random place to add this, I would suggest adding in PerformRequest_PremiumHeaders and PerformRequest_NonPremium like selected_language and system_language case.

browser/ai_chat/ai_chat_service_factory.cc
petemill: It's separate because we will need to use `is_actor_allowed` to create some things the `AIChatService` ctor will need, in the following PR which immediately follows this one:
- https://github.com/brave/brave-core/pull/31224

components/ai_chat/core/browser/engine/conversation_api_client_unittest.cc
```diff
@@ -364,32 +364,23 @@ class ConversationAPIUnitTest : public testing::Test {
 
   void TearDown() override {}
 
-  base::Value::List GetEvents(std::string_view body_json) {
-    auto dict = base::JSONReader::ReadDict(body_json);
-    EXPECT_TRUE(dict.has_value());
-    base::Value::List* events = dict->FindList("events");
+  base::Value::List& GetEvents(base::Value::Dict& body) {
```
yrliou, line 367: const base::Value::Dict& body

petemill: Then return type needs to be `const base::Value::List&` and I didn't think we need to add the restriction that the caller can't modify the list, especially in a test.

yrliou: I can't think of a use case that we would want to manipulate the events we get from the body, but I'm okay with it if you prefer to stay as is.

PR #31224, opened 2025-09-16
Integrate AIChat with the Chromium ActorKeyedService

petemill:
This introduces the connection for providing Tools which use the ActorKeyedService to AIChat conversations. It fulfills a per-conversation owner of an Actor Task, which will be available to all Tools which will also be owned by the same ToolProvider. We provide this functionality to the Tools by means of a TaskProvider so that we won't need to pass the ContentAgentToolProvider definition to each Tool and can limit what we expose in that circular relationship.
- Introduce ContentAgentToolProvider which implements AIChat's ToolProvider. It will be used to group a set of Tools for content agent purposes and is only created for the agent profile
- ContentAgentToolProvider will be responsible for using the Actor Service to create and own Tasks and the Tabs within the Tasks.
- Add 1 tool in order to provide context. The rest will follow in subsequent PR.
- Overrides Chromium's actor site policy so that the tool can act on and with any URL. We change the fallback to allow instead of deny. We restrict it's use to Tabs created within the conversation's task.

## Tests:
- The navigation tool has unit tests verifying the parsing of incoming parameters
- There is a browser test for verifying the content tools are only created in the agent profile
- This PR does not have any tests verifying that the actor framework actually works with the tool requests and the tasks created by our tool provider. That will be provided in the following PR, or one split from it, as that branch completes the lifecycle of the tool request by providing a response instead of a NOTIMPLEMENTED in this PR:
  - https://github.com/brave/brave-core/pull/30743

See the integration branch for more context and a working version:
- https://github.com/brave/brave-core/pull/30743


components/ai_chat/core/browser/tools/tool_provider.h
```diff
@@ -52,6 +52,10 @@ class ToolProvider {
   // Note: any filtering conditions required by ToolProviders can be added as
   // params here.
   virtual std::vector<base::WeakPtr<Tool>> GetTools() = 0;
+
+  // Attempts to stops all current tasks started by Tools from this
+  // ToolProvider.
+  virtual void StopAllTasks() {}
```
petemill, line 58: Calling StopAllTasks for each ToolProvider, as well as stopping the agentic tool loop when cancelling an in-progress generation request, will come in the ConversationHandler in https://github.com/brave/brave-browser/issues/49293.

components/ai_chat/core/browser/DEPS
```diff
@@ -10,6 +10,7 @@ include_rules = [
   "+components/os_crypt/sync",
   "+components/prefs",
   "+components/sync_preferences/testing_pref_service_syncable.h",
+  "+components/tabs",
```
fallaciousreasoning: do we need this? doesn't seem to like we have any imports of it yet

browser/ai_chat/content_agent_tool_provider.h
```diff
@@ -0,0 +1,79 @@
...
+
+// Provides content agent tools to conversations, manages the lifecycle of
+// those tools, and provides the actor tasks for the tools to act on as well as
+// interfacing with the actor service to execute the actions, deciding
+// which tabs to act on.
+class ContentAgentToolProvider : public ToolProvider,
+                                 public ContentAgentTaskProvider {
+ public:
+  ContentAgentToolProvider(Profile* profile,
+                           raw_ptr<actor::ActorKeyedService> actor_service);
```
fallaciousreasoning: Should we be passing a `actor::ActorKeyedServce*` in here, instead of a raw_ptr

browser/ai_chat/content_agent_task_provider.h
```diff
@@ -0,0 +1,50 @@
...
+
+  // Get the current tab for the task.
+  // TODO(https://github.com/brave/brave-browser/issues/49258): re-architect
+  // so that multiple tab can be added to the task, observed and acted on. The
+  // AI can choose which tab to act on via a tab ID parameter, as the actor
+  // framework expects.
+  virtual void GetOrCreateTabHandleForTask(
+      base::OnceCallback<void(tabs::TabHandle)> callback) = 0;
+
+  // TODO(https://github.com/brave/brave-browser/issues/49289): Now that we can
```
fallaciousreasoning, line 39: Is there a reason we're not just doing this now? CR140 has landed

goodov (approved):
`DEPS` lgtm

browser/ai_chat/ai_chat_service_factory.cc
```diff
@@ -82,13 +89,23 @@ AIChatServiceFactory::BuildServiceInstanceForBrowserContext(
       misc_metrics::ProfileMiscMetricsServiceFactory::GetServiceForContext(
           context);
 
+  // Tool sets provided from this layer to conversations
   std::vector<std::unique_ptr<ToolProviderFactory>> tool_provider_factories;
-  tool_provider_factories.push_back(
-      std::make_unique<BrowserToolProviderFactory>());
```
yrliou, line 87: Are we removing this first on purpose? It currently doesn't have any tools so I guess it's technically fine if you plan to add it back later when adding a browser tool.

browser/ai_chat/ai_chat_service_factory.cc
```diff
@@ -82,13 +89,23 @@ AIChatServiceFactory::BuildServiceInstanceForBrowserContext(
...
 
 #if BUILDFLAG(ENABLE_BRAVE_AI_CHAT_AGENT_PROFILE)
   bool is_actor_allowed = features::IsAIChatAgentProfileEnabled() &&
                           Profile::FromBrowserContext(context)->IsAIChatAgent();
```
yrliou: Missed it in previous review, it should just be `context->IsAIChatAgent()`.

browser/ai_chat/content_agent_tool_provider.cc
```diff
@@ -0,0 +1,183 @@
...
+#include "components/optimization_guide/content/browser/page_content_proto_provider.h"
+#include "components/tabs/public/tab_interface.h"
+#include "content/public/browser/web_contents.h"
+#include "ui/base/page_transition_types.h"
+
+namespace ai_chat {
+
+ContentAgentToolProvider::ContentAgentToolProvider(
+    Profile* profile,
+    raw_ptr<actor::ActorKeyedService> actor_service)
```
yrliou: just ActorKeyedService* here

browser/ai_chat/content_agent_tool_provider.h
yrliou: actor::ActorKeyedService*

browser/ai_chat/content_agent_tool_provider.cc
```diff
@@ -0,0 +1,183 @@
...
+    : actor_service_(actor_service), profile_(profile) {
+  // Only create page actor tools if this profile is allowed to.
+  if (actor_service_) {
```
yrliou: Should it be `CHECK(actor_service_);` instead as we won't create this provider if actor_service is null in the first place?

browser/ai_chat/content_agent_task_provider.h
```diff
@@ -0,0 +1,50 @@
...
+#include "chrome/browser/actor/task_id.h"
+#include "components/optimization_guide/proto/features/actions_data.pb.h"
+#include "components/tabs/public/tab_interface.h"
+
+namespace ai_chat {
+
+// Provides glue between tools and the actor framework, specifically around
+// actor tasks and tool execution. A task provider may choose how to execute
+// the actions and provide the results of those actions. This could be
+// implemented on platforms that don't support the chromium actor framework.
```
fallaciousreasoning, line 19: what platforms that support `//brave/browser/ai_chat` don't support the chromium actor framework? Just android?

browser/ai_chat/content_agent_tool_provider.cc
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  actor_service_->PerformActions(
+      actor::TaskId(actions.task_id()), std::move(requests.value()),
+      base::BindOnce(&ContentAgentToolProvider::OnActionsFinished,
+                     weak_ptr_factory_.GetWeakPtr(), std::move(callback)));
+}
+
+void ContentAgentToolProvider::CreateTools() {
+  tools_.clear();
```
fallaciousreasoning, line 126: why do we need to clear the tools here? This method is only called from the constructor (and its private)

browser/ai_chat/content_agent_tool_provider.cc
```diff
@@ -0,0 +1,183 @@
...
+    NavigateParams params(profile_, GURL(url::kAboutBlankURL),
+                          ui::PAGE_TRANSITION_FROM_API);
+    params.disposition = WindowOpenDisposition::NEW_BACKGROUND_TAB;
+    Navigate(&params);
+    content::WebContents* new_contents = params.navigated_or_inserted_contents;
+
+    task_tab_handle_ =
+        tabs::TabInterface::GetFromContents(new_contents)->GetHandle();
+  }
+  actor_service_->GetTask(task_id_)->AddTab(
```
fallaciousreasoning, line 95: So in the other methods, we check whether the `actor_service_` exists. Is there a reason we don't need to do it here too (this is public)?

browser/ai_chat/content_agent_tool_provider.cc
```diff
@@ -0,0 +1,183 @@
...
+      task_tab_handle_, base::BindOnce(&ContentAgentToolProvider::TabAdded,
```
fallaciousreasoning: Is there a reason we don't just invoke the callback directly here? Is it so the callback gets dropped if this class is destroyed?

browser/ai_chat/content_agent_tool_provider.cc
```diff
@@ -0,0 +1,183 @@
...
+std::vector<base::WeakPtr<Tool>> ContentAgentToolProvider::GetTools() {
+  // Note: We don't have the ability to filter tools based on conversation
+  // capability here. But for now we don't need to as we only create the content
+  // agent tools if we don't have the actor_service_ (which is only provided for
+  // agent profiles).
+  std::vector<base::WeakPtr<Tool>> tool_ptrs;
+  tool_ptrs.reserve(tools_.size());
+  for (const auto& tool : tools_) {
+    tool_ptrs.push_back(tool->GetWeakPtr());
+  }
```
fallaciousreasoning: optional: can we do a std::ranges::transform here with `&Tool::GetWeakPtr` as the transform function?

browser/ai_chat/content_agent_tool_provider.h
```diff
@@ -0,0 +1,79 @@
...
+      std::optional<optimization_guide::AIPageContentResult> content);
+
+  // Browser-specific tools owned by this provider.
+  // Note: if it becomes an advantage to refer directly to a specific tool,
+  // then there's no need for a vector - we can simply store each tool in its
+  // own member.
+  std::vector<std::unique_ptr<Tool>> tools_;
+
+  actor::TaskId task_id_;
+  tabs::TabHandle task_tab_handle_;
```
fallaciousreasoning, line 69: Hey so one of these `ContentAgentToolProvider`s is bound to a single tab? This doesn't feel like it fits with the description on this class:
`deciding which tabs to act on.`

I'm not sure that this is the right layer to be storing `TaskTabHandle`. In the case of the NavigationTool, it seems like the NavigationTool should decide what tab it wants to navigate and hold onto that handle.

browser/ai_chat/content_agent_task_provider.h
petemill: Just haven't got round to it and don't want to delay. There's a lot of converting I'll need to do both in classes and unit tests that are contained in the next PR which is already ready. I'd rather do that refactoring after the feature is complete.

browser/ai_chat/content_agent_tool_provider.cc
```diff
@@ -0,0 +1,183 @@
...
+
+    // TODO(https://github.com/brave/brave-browser/issues/49259):
+    // Use multi_source_page_context_fetcher.h (or use it
+    // via ActorKeyedService), now that this API is public outside of glic.
+
+    auto options = blink::mojom::AIPageContentOptions::New();
+    options->mode = blink::mojom::AIPageContentMode::kActionableElements;
+
+    optimization_guide::GetAIPageContent(
+        task_tab_handle_.Get()->GetContents(), std::move(options),
```
fallaciousreasoning, line 155: should we `CHECK` TaskTabHandle?

browser/ai_chat/content_agent_tool_provider.h
petemill: I, er, yeah

browser/ai_chat/content_agent_tool_provider.cc
fallaciousreasoning: Also, is it true that we want to get the page content after every action? I guess it makes sense

browser/ai_chat/content_agent_task_provider.h
petemill: We could conceivably move content_agent_task_provider to components/ai_chat and have it working on iOS. The mechanism of LLM response -> Tools -> Task started -> Observe Tabs could be shared on all platforms. But, for now, desktop only.

browser/ai_chat/content_agent_tool_provider.h
petemill: Right now the ContentAgentToolProvider only supports acting on a single and new tab. We will most likely expand it to being able to create multiple tabs for a task, as well as be able to act on existing tabs. We just haven't really tested that or decided what the UX for attaching a tab to be acted on looks like rather than attaching for content retrieval. 
I can imagine that yes the NavigationTool or the ClickTool, etc. will decide which Tab to act on, but it should only be within a list of tabs owned by the Task. It wouldn't just get to choose any tabs the user has open, I don't think.
Also, the NavigationTool can't be the one owning that tab handle because then the other tools won't know about that handle. The ContentAgentToolProvider is the glue between the set of tools to act on the content.

browser/ai_chat/content_agent_tool_provider.cc
petemill: I wanted to do it "by the book" as far as the ActorService / Task is concerned and ensure we're not acting on the content before `ActorTask::AddTab` does what it needs to do. Otherwise there might be some error about acting on the Tab before it's added to the task.

browser/ai_chat/content_agent_tool_provider.cc
petemill: A fair observation. I did it because we're not checking the existance of any of the tools or being intelligent about when to add each one. It's possible we may want to re-create the tools during `ToolProvider::OnNewGenerationLoop()` if the tools hang on to state. I might have had that happen previously and removed it. Is there a disadvantage of leaving the clear here?

browser/ai_chat/content_agent_tool_provider.cc
petemill: Yeah I can't think of any actions where the new state isn't important.

browser/ai_chat/content_agent_task_provider.h
fallaciousreasoning: I mean, we don't have `TabHandle`s on iOS, so seems pretty unlikely. Can we remove this interface if we aren't going to need it (everything exists in the same layer for now), so not sure how useful it is

browser/ai_chat/content_agent_tool_provider.cc
fallaciousreasoning: I guess not, it just read as slightly jarring given its only called from the ctor

browser/ai_chat/content_agent_tool_provider.cc
fallaciousreasoning: Sorry, I meant doing this:

```
actor_service_->GetTask(task_id_)->AddTab(
      task_tab_handle_, base::BindOnce(std::move(callback), other_stuff));
```

instead of adding an instance method

browser/ai_chat/content_agent_tool_provider.h
fallaciousreasoning: Ah okay - so concievably this will migrate to a `std::vector` once we need it? And `ContentAgentToolProvider` is responsible for creating/adding all the tabs the other tools are allowed to act on?

fallaciousreasoning:
Hey, reading this I'm wondering if we should maybe add a `README.md` file to this directory to explain how ContentAgentTools should work

browser/ai_chat/content_agent_task_provider.h
petemill: ah well another purpose of `ContentAgentTaskProvider` is being able to pass something to the tools (`NavigationTool` in this PRs case) so that they could use a Task without passing their owner, `ContentAgentToolProvider` in a circular reference fashion. Whilst we can probably get around the circular reference because of header files, passing this reduced interface helps with the unit testing so that we don't need to mock the entire `ContentAgentToolProvider` for each Tool's unit test (see `browser/ai_chat/tools/navigation_tool_unittest.cc` and `mock_content_agent_task_provider.{h,cc}`. Do you think there's value enough there?

browser/ai_chat/content_agent_tool_provider.cc
petemill: In a previous version of this PR, I had the actor_service_ sometimes be passed in as `nullptr`. But now the `ContentAgentToolProvider` class itself is not created by `AIChatServiceFactory` if there is no `ActorKeyedService`. So I think I'll just remove all the individual checks and `CHECK` in the ctor.

browser/ai_chat/content_agent_tool_provider.cc
petemill: I think mainly because I didn't want `actor::mojom::ActionResultPtr result` to be part of the callback direct to the tools. It doesn't seem necessary and if we do get in to some error cases then we should probably handle it in `ContentAgentToolProvider` and not in each of the many tools.
I suppose we could handle that with a base class for all the Tools, but either way we'd still need a class method which intercepts this result, handles the success or error condition and either continues with the Tool execution or sends an error response back to the LLM.

browser/ai_chat/content_agent_tool_provider.h
petemill: `ContentAgentToolProvider` will own the Task(s) for a conversation (or potentially a single message generation in a single conversation). The Task could have multiple Tabs, so yes one way is for this to be a vector created from conversation message attachments at the point of submission - maybe these would be provided via metadata in `ToolProvider::OnNewGenerationLoop`. Perhaps there would be a "list or observe tabs" tool which would tell the LLM about which tabs are available to be acted on in the current session. And then the LLM could choose to act on one of those tabs, or on a new one.

browser/ai_chat/content_agent_tool_provider.cc
```diff
@@ -0,0 +1,189 @@
...
+    Navigate(&params);
+    content::WebContents* new_contents = params.navigated_or_inserted_contents;
+
+    task_tab_handle_ =
+        tabs::TabInterface::GetFromContents(new_contents)->GetHandle();
+  }
+  actor_service_->GetTask(task_id_)->AddTab(
+      task_tab_handle_, base::BindOnce(&ContentAgentToolProvider::TabAdded,
+                                       weak_ptr_factory_.GetWeakPtr(),
+                                       task_tab_handle_, std::move(callback)));
```
yrliou: Seems no need to pass task_tab_handle_ here, just std::move(callback).Run(task_tab_handler_) in TabAdded?

browser/ai_chat/content_agent_tool_provider.cc
```diff
@@ -0,0 +1,189 @@
...
+#include "chrome/browser/ui/browser_window/public/browser_window_interface_iterator.h"
+#include "chrome/common/actor.mojom.h"
+#include "components/optimization_guide/content/browser/page_content_proto_provider.h"
+#include "components/tabs/public/tab_interface.h"
+#include "content/public/browser/web_contents.h"
+#include "ui/base/page_transition_types.h"
+
+namespace ai_chat {
+
+ContentAgentToolProvider::ContentAgentToolProvider(
```
yrliou, line 33: Not super sure about the full picture, I do see there're some browser tests added in the subsequent PR (BrowserTools...), ideally we would also need to figure out a way to test edge (error handling) cases in a unit test here.

browser/ai_chat/content_agent_task_provider.h
fallaciousreasoning: sure, happy to be convinced if it makes unit tests easier :laughing:

browser/ai_chat/tools/navigation_tool.cc
```diff
@@ -0,0 +1,99 @@
...
+    const {
+  return std::optional<std::vector<std::string>>({"website_url"});
+}
+
+void NavigationTool::UseTool(const std::string& input_json,
+                             UseToolCallback callback) {
+  data_decoder::DataDecoder::ParseJsonIsolated(
+      input_json,
+      base::BindOnce(&NavigationTool::OnParseJson,
+                     weak_ptr_factory_.GetWeakPtr(), std::move(callback)));
```
yrliou: Could we move to base::JSONReader::ReadDict instead, it's simpler to read without introducing OnParseJson method.

fallaciousreasoning (commented):
lgtm % README. Happy for you to resolve any lingering comment threads

browser/ai_chat/content_agent_tool_provider.cc
petemill: Already done from previous review

browser/ai_chat/content_agent_tool_provider.h
petemill: Already done from previous review

browser/ai_chat/content_agent_tool_provider.cc
petemill: Already done from previous review

fallaciousreasoning (approved):

browser/ai_chat/content_agent_tool_provider.cc
petemill: As far as I can tell, there is not much unit testable in this class for now. For sure in the future where it's doing more things. But creating, manipulating and receiving the annotated contents of WebContents is more for browser tests, which come with the next PR, which will also serialize the annotated page content and will contain unit tests for that. If you have an idea of something that is worth unit testing before then them I'm open to it.

browser/ai_chat/content_agent_tool_provider.cc
petemill: I suppose that's true in the current state. I did it like this because the sentiment around this function is that it will get or create the appropriate tab given the task but, since we don't have that functionality yet, it's always the same tab. I'll adjust and we can re-introduce it later.

browser/ai_chat/ai_chat_service_factory.cc
petemill: I put it back. It will be used by at least TabManagementTool

yrliou (approved):

browser/ai_chat/tools/navigation_tool.cc
```diff
@@ -0,0 +1,93 @@
...
+    return;
+  }
+
+  task_provider_->GetOrCreateTabHandleForTask(
+      base::BindOnce(&NavigationTool::OnTabHandleCreated,
+                     weak_ptr_factory_.GetWeakPtr(), std::move(callback), url));
+}
+
+void NavigationTool::OnTabHandleCreated(UseToolCallback callback,
+                                        GURL url,
```
yrliou, line 83: const GURL&

browser/ai_chat/content_agent_tool_provider.cc
```diff
@@ -0,0 +1,188 @@
...
+
+  if (!apc.has_root_node()) {
+    DLOG(ERROR) << "No root node";
+    std::move(callback).Run(CreateContentBlocksForText("No root node"));
+    return;
+  }
+
+  // TODO(https://github.com/brave/brave-browser/issues/49301): implement
+  // structured representation of page content
+  NOTREACHED();
```
netzenbot, line 185: bbondy asked me to review this already closed PR.

`NOTREACHED()` will crash the browser if this code path is hit. Since this is a TODO stub for unimplemented functionality (not a security invariant), consider returning an error via `callback` instead so the browser does not crash during development. [best practice](https://github.com/brave-experiments/brave-core-tools/tree/master/docs/best-practices/coding-standards.md#CS-026)

browser/ai_chat/tools/navigation_tool_unittest.cc
```diff
@@ -0,0 +1,211 @@
...
+  std::unique_ptr<MockContentAgentTaskProvider> mock_task_provider_;
+  std::unique_ptr<NavigationTool> navigation_tool_;
+  tabs::TabHandle test_tab_handle_;
+  actor::TaskId test_task_id_;
+};
+
+TEST_F(NavigationToolTest, ValidInputHttpsUrl) {
+  std::string test_url = "https://www.example.com";
+  std::string input_json = CreateValidNavigationJson(test_url);
+  base::RunLoop run_loop;
```
netzenbot, line 108: bbondy asked me to review this already closed PR.

This test uses the `RunLoop` + lambda + `Quit()` pattern that TA-009 recommends replacing with `TestFuture`. Consider using `base::test::TestFuture` for the `UseTool` callback (as `RunWithExpectedError` already does) and `testing::SaveArg` to capture the `actions` proto from the mock. [best practice](https://github.com/brave-experiments/brave-core-tools/tree/master/docs/best-practices/testing-async.md#TA-009)

browser/ai_chat/tools/navigation_tool_unittest.cc
```diff
@@ -0,0 +1,211 @@
...
+  run_loop.Run();
+
+  // Verify navigation action properties
+  VerifyNavigationAction(captured_actions, test_url);
+}
+
+TEST_F(NavigationToolTest, ValidInputComplexUrl) {
+  std::string test_url = "https://search.brave.com/search?q=test&source=web";
+  std::string input_json = CreateValidNavigationJson(test_url);
+  base::RunLoop run_loop;
```
netzenbot, line 129: bbondy asked me to review this already closed PR.

Same `RunLoop` pattern as above. Since this mock already invokes the callback, this is a straightforward conversion: pass `TestFuture::GetCallback()` to `UseTool` instead of `base::DoNothing()`, use `testing::SaveArg` for the captured actions, and remove the `RunLoop`. [best practice](https://github.com/brave-experiments/brave-core-tools/tree/master/docs/best-practices/testing-async.md#TA-009)

browser/ai_chat/tools/navigation_tool_unittest.cc
```diff
@@ -0,0 +1,211 @@
...
+  optimization_guide::proto::Actions captured_actions;
+  EXPECT_CALL(*mock_task_provider_, ExecuteActions(testing::_, testing::_))
+      .WillOnce(testing::Invoke([&captured_actions, &run_loop](
+                                    optimization_guide::proto::Actions actions,
+                                    Tool::UseToolCallback callback) {
+        captured_actions = std::move(actions);
+        run_loop.Quit();
+      }));
+
+  navigation_tool_->UseTool(input_json, base::DoNothing());
```
netzenbot, line 119: bbondy asked me to review this already closed PR.

Both `ValidInputHttpsUrl` and `ValidInputComplexUrl` pass `base::DoNothing()` as the `UseTool` callback and never verify the final result. Consider invoking the captured callback and asserting the tool's final output to verify the full success path end-to-end. [best practice](https://github.com/brave-experiments/brave-core-tools/tree/master/docs/best-practices/testing-isolation.md#TI-021)

browser/ai_chat/BUILD.gn
```diff
@@ -150,6 +174,27 @@ if (enable_brave_ai_chat_agent_profile) {
   }
 }
 
+if (enable_brave_ai_chat_agent_profile) {
+  source_set("content_agent_test_support") {
```
netzenbot, line 178: bbondy asked me to review this already closed PR.

nit: `source_set("content_agent_test_support")` should have a `visibility` restriction (e.g., `visibility = [ ":*" ]` or list specific test targets) to prevent unintended dependencies. [best practice](https://github.com/brave-experiments/brave-core-tools/tree/master/docs/best-practices/build-system.md#BS-028)

browser/ai_chat/content_agent_tool_provider.cc
```diff
@@ -0,0 +1,188 @@
...
+    // whilst this feature is focused on AI-initiated tasks instead
+    // of acting on existing tabs.
+    NavigateParams params(profile_, GURL(url::kAboutBlankURL),
+                          ui::PAGE_TRANSITION_FROM_API);
+    params.disposition = WindowOpenDisposition::NEW_BACKGROUND_TAB;
+    Navigate(&params);
+    content::WebContents* new_contents = params.navigated_or_inserted_contents;
+
+    task_tab_handle_ =
+        tabs::TabInterface::GetFromContents(new_contents)->GetHandle();
```
netzenbot, line 93: bbondy asked me to review this already closed PR.

`GetFromContents(new_contents)` can return nullptr — `params.navigated_or_inserted_contents` may be null if navigation fails, and even with a valid WebContents, `GetFromContents` is not guaranteed to return non-null. Store the result and null-check before calling `->GetHandle()`. [best practice](https://github.com/brave-experiments/brave-core-tools/tree/master/docs/best-practices/coding-standards-memory.md#CSM-029)

browser/ai_chat/content_agent_tool_provider.cc
```diff
@@ -0,0 +1,188 @@
...
+  }
+  actor_service_->GetTask(task_id_)->AddTab(
```
netzenbot, line 95: bbondy asked me to review this already closed PR.

`GetTask(task_id_)` can return nullptr if the task was stopped or removed (the upstream API documents: "Returns nullptr if the task does not exist"). Store the result and null-check before calling `->AddTab()`. [best practice](https://github.com/brave-experiments/brave-core-tools/tree/master/docs/best-practices/coding-standards-memory.md#CSM-029)

PR #31226, opened 2025-09-16
[NTP Next] Account for scroll position when posting NTT safe area

zenparsing:
Resolves https://github.com/brave/brave-browser/issues/49184


tmancey:
Tested locally and all good.

btlechowski:
Verified with

Brave | 1.84.83 Chromium: 140.0.7339.186 (Official Build) nightly (64-bit)
-- | --
Revision | e0c2fd58aa7ccaf7595022f392784fe4e7483f8b
OS | Linux

Verified test plan from https://github.com/brave/brave-browser/issues/49184

<details>

Verified safe area does not change based on scrolling.

image|image
--|--

Verified safe area does not change on reload
image


</details>

PR #31230, opened 2025-09-16
Brave origin utils

bbondy:
Resolves https://github.com/brave/brave-browser/issues/49316

### DIFF
--- a/components/brave_origin/brave_origin_utils.cc
+++ b/components/brave_origin/brave_origin_utils.cc
@@ -6,6 +6,9 @@
 #include "brave/components/brave_origin/brave_origin_utils.h"
 
 #include "base/feature_list.h"
+#include "base/logging.h"
+#include "base/strings/strcat.h"
+#include "brave/components/brave_origin/brave_origin_policy_info.h"
 #include "brave/components/brave_origin/features.h"
 
 namespace brave_origin {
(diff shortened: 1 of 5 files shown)


components/brave_origin/brave_origin_utils.h
```diff
@@ -0,0 +1,29 @@
...
+bool IsBraveOriginEnabled();
+
+// Forward declaration
+struct BraveOriginPolicyInfo;
+
+// Gets the correct brave_origin_pref_key for a preference based on its scope
+// and the provided profile_id. For global prefs, returns pref_name directly.
+// For profile prefs, returns "profile_id.pref_name".
+std::string GetBraveOriginPrefKey(const BraveOriginPolicyInfo& pref_info,
+                                  const std::string& profile_id);
```
DJAndries: nit: consider changing this to `std::optional<std::string_view>`

bbondy: Returning a `string_view` would give the caller a pointer into that built string's temporary’s buffer. So it would be unsafe. `std::string` is right here.

components/brave_origin/brave_origin_utils.cc
```diff
@@ -0,0 +1,38 @@
...
+std::string GetBraveOriginPrefKey(const BraveOriginPolicyInfo& pref_info,
+                                  const std::string& profile_id) {
+  // For global prefs, use policy_key directly
+  // For profile prefs, use profile_id.policy_key format
+  if (pref_info.scope == BraveOriginPolicyScope::kGlobal) {
+    return pref_info.policy_key;
+  } else {
+    CHECK(!profile_id.empty())
+        << "Profile ID cannot be empty for profile-scoped preferences";
+    return profile_id + "." + pref_info.policy_key;
```
DJAndries: nit: `base::StrCat`

components/brave_origin/brave_origin_utils.h
DJAndries: I meant to recommend replacing the `profile_id` type, but it's not really applicable as discussed over DM.

DJAndries (approved):

PR #31231, opened 2025-09-16
Add starter pack for "Ask Brave Search" (v2)

zenparsing:
Resolves https://github.com/brave/brave-browser/issues/49179

### DIFF
--- a/chromium_src/chrome/browser/ui/search_engines/template_url_table_model.cc
+++ b/chromium_src/chrome/browser/ui/search_engines/template_url_table_model.cc
@@ -7,11 +7,14 @@
 #include "components/search_engines/template_url.h"
 #include "components/search_engines/template_url_starter_pack_data.h"
 
-// We disable starter pack expansion to hide @gemini search keyword. Piggy back
-// on it to also disable @aimode.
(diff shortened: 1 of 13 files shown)


chromium_src/chrome/browser/ui/search_engines/template_url_table_model.cc
```diff
@@ -7,11 +7,14 @@
...
 #include "components/search_engines/template_url_starter_pack_data.h"
 
-// We disable starter pack expansion to hide @gemini search keyword. Piggy back
-// on it to also disable @aimode.
-#define kGemini                                 \
-  kGemini || template_url->starter_pack_id() == \
-                 template_url_starter_pack_data::kAiMode
+// Unconditionally skip @aimode keyword. Always allow `kGemini`, which is
+// reused to implement `@ask`.
+#define kGemini kAiMode) {                        \
```
bridiver, line 12: why is this changing it to `kAiMode`?

bridiver: I don't understand this change at all, why do we need to do anything with `kAiMode` here?

PR #31235, opened 2025-09-16
Add policy for Brave fingerprinting V2

aseren:
Resolves https://github.com/brave/brave-browser/issues/49335


browser/policy/brave_configuration_policy_pref_store_unittest.cc
```diff
@@ -0,0 +1,54 @@
...
+      handler_list_.AddHandler(std::make_unique<SimplePolicyHandler>(
+          entry.policy_name, entry.preference_path, entry.value_type));
+    }
+  }
+};
+
+TEST_F(BraveConfigurationPolicyPrefStoreTest, GetDefaultFingerprintingV2) {
+  EXPECT_FALSE(store_->GetValue(kManagedDefaultBraveFingerprintingV2, nullptr));
+
+  const base::Value* value = nullptr;
```
bridiver: What exactly are we testing here? These tests ideally be split out into multiple test cases, but this seems like we're testing the functionality of managed policies in general? There is no class called `BraveConfigurationPolicy` that I can see

aseren: Thanks for comment, you are right and purpose of this test is quite vague. I renamed it to `BraveSimplePolicyMapTest` and simplified the test code.
So now it checks following:
* policy was added to `kBraveSimplePolicyMap` 
* policy can be set with specific value type

aseren: I'm removing the test per discussion in DM as it tests functionality of managed policies in general

aseren: unneeded unit-test removed

components/policy/resources/templates/policy_definitions/BraveSoftware/DefaultBraveFingerprintingV2Setting.yaml
```diff
@@ -0,0 +1,44 @@
...
+example_value: 3
+features:
+  can_be_mandatory: true
+  can_be_recommended: false
+  dynamic_refresh: true
+  per_profile: true
+items:
+- caption: Disable fingerprinting protection
+  name: DisableFingerprintingProtection
+  value: 1
```
bridiver, line 21: this is not correct, disable == CONTENT_SETTING_BLOCK which is 2

aseren: yeah, `brave_shields::GetFingerprintingControlType` functionality is really confusing: if `brave_shields::GetFingerprintingControlType` returns `ControlType::ALLOW` it means that fingerprinting is off.

Here is code which shows it:
- https://github.com/brave/brave-core/blob/master/components/brave_shields/core/browser/brave_shields_utils.cc#L381
- https://github.com/brave/brave-core/blob/master/browser/resources/settings/default_brave_shields_page/default_brave_shields_page.ts#L255

bridiver: I see, yeah that does make sense in retrospect

components/constants/pref_names.h
```diff
@@ -16,6 +16,8 @@ inline constexpr char kManagedBraveShieldsDisabledForUrls[] =
     "brave.managed_shields_disabled";
 inline constexpr char kManagedBraveShieldsEnabledForUrls[] =
     "brave.managed_shields_enabled";
+inline constexpr char kManagedDefaultBraveFingerprintingV2[] =
```
bridiver, line 19: Doesn't this also have to be registered like `kManagedDefaultPrefs` `PolicyProvider::RegisterProfilePrefs`? Since we're adding a new pref here and not just making an existing pref managed I think in this case unit tests also make sense here

components/policy/resources/templates/policy_definitions/BraveSoftware/DefaultBraveFingerprintingV2Setting.yaml
```diff
@@ -0,0 +1,44 @@
...
+- caption: Enable fingerprinting protection strict mode
```
ShivanKaul: We have deprecated strict fingerprinting mode: https://brave.com/privacy-updates/28-sunsetting-strict-fingerprinting-mode/. We shouldn't allow tweaking this.

bridiver: oh, we should do a settings migration then so we're not using ask anymore, that would really simplify this

bridiver: I mean not this specifically, but fingerprinting checks in general

aseren: `Enable fingerprinting protection strict mode` option is removed

components/constants/pref_names.h
aseren: Added registering of `kManagedDefaultBraveFingerprintingV2` as a separate PR: https://github.com/brave/brave-core/pull/31564. Should I add to the current PR?

bridiver: I would combine the two, they are both small and make sense together I think. This would be a bit like adding a pref in one PR, but not actually using it

aseren: Combined, thanks!

patches/components-content_settings-core-browser-content_settings_policy_provider.cc.patch
```diff
@@ -18,3 +18,19 @@ index 174e7f0c9e714bc7a792d9dc69fbc8c01d064146..44f63ba6a842eed74bc395e9b39274bc
      prefs::kManagedAutomaticFullscreenAllowedForUrls,
      prefs::kManagedAutomaticFullscreenBlockedForUrls,
      prefs::kManagedAutoSelectCertificateForUrls,
+@@ -234,6 +236,7 @@ constexpr const char* kManagedPrefs[] = {
+ // the preference default content settings. If a default content settings type
+ // is managed any user defined exceptions (patterns) for this type are ignored.
+ constexpr const char* kManagedDefaultPrefs[] = {
++    BRAVE_MANAGED_DEFAULT_PREFS
```
bridiver, line 25: why are these different from the existing `kManagedPrefs` and `kPrefsForManagedContentSettingsMap` overrides?

aseren: Thanks for comment! 
`kManagedPrefs` and `kManagedDefaultPrefs` are used for different things. 
`kManagedPrefs` - is used to apply specific content value across urls, defined by patterns from policy
`kManagedDefaultPrefs` - is used to define Default value for content setting, which is defined by policy. 

I used the same patter as following chromium managed prefs:
- policy::key::kDefaultJavaScriptSetting
- policy::key::kDefaultCookiesSetting

bridiver: ah right, I see. Thanks!

components/policy/resources/templates/policy_definitions/BraveSoftware/DefaultBraveFingerprintingV2Setting.yaml
```diff
@@ -0,0 +1,37 @@
...
+  per_profile: true
+items:
+- caption: Disable fingerprinting protection
+  name: DisableFingerprintingProtection
+  value: 1
+- caption: Enable fingerprinting protection standard mode
+  name: EnableFingerprintingProtectionStandardMode
+  value: 3
+owners:
+- bbondy@brave.com
```
bsclifton, line 26: Can you add yourself here @aseren

components/policy/resources/templates/policy_definitions/BraveSoftware/DefaultBraveFingerprintingV2Setting.yaml
```diff
@@ -0,0 +1,37 @@
...
+- clifton@brave.com
+schema:
+  enum:
+  - 1
+  - 3
+  type: integer
+supported_on:
+- chrome.*:140-
```
bsclifton: we'll want to make sure this matches whatever Nightly is on when we merge; so like 141

components/policy/resources/templates/policy_definitions/BraveSoftware/DefaultBraveFingerprintingV2Setting.yaml
aseren: Added, thanks!

components/policy/resources/templates/policy_definitions/BraveSoftware/DefaultBraveFingerprintingV2Setting.yaml
aseren: Great catch! Changed and also rebase to latest master, to match version.

PR #31236, opened 2025-09-16
[plaster] Allow exactly one of pattern (exact match) or re_pattern (python regex)

bridiver:
We should allow either pattern or re_pattern and error if both are present, none are present or replace is missing

### DIFF
--- a/docs/plaster.md
+++ b/docs/plaster.md
@@ -63,26 +63,26 @@ occurrences of `ChromeAutocompleteSchemeClassifier` with
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at https://mozilla.org/MPL/2.0/.
 
-[[substitutions]]
+[[substitution]]
 description = 'Adding header for BraveAutocompleteSchemeClassifier'
-pattern = '#include "extensions/buildflags/buildflags.h"'
-re_pattern = '''#include "extensions/buildflags/buildflags.h"
-#include "brave/browser/autocomplete/brave_autocomplete_scheme_classifier.h"'''
+re_pattern = '(#include "[\s\S]*)'
+replace = '#include "brave/browser/autocomplete/brave_autocomplete_scheme_classifier.h"\n\1'
 
-[[substitutions]]
+[[substitution]]
 description = 'Patching in BraveAutocompleteSchemeClassifier'
-re_pattern = 'ChromeAutocompleteSchemeClassifier'
+pattern = 'ChromeAutocompleteSchemeClassifier'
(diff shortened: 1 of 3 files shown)


tools/cr/plaster_test.py
```diff
@@ -255,6 +255,178 @@ def __init__(self):
...
+                plaster_path.parent.mkdir(parents=True, exist_ok=True)
+                plaster_path.write_text(toml_content)
+
+                plaster_file = plaster.PlasterFile(plaster_path)
+                with self.assertRaises(ValueError) as context:
+                    plaster_file.apply()
+
+                self.assertIn(expected_error, str(context.exception))
+
+    def test_pattern_exact_match_works(self):
```
mkarolin, line 412: This test and the ones below seem to have the same code pattern. Maybe make it into a function that takes the source file content, the test plaster file content, and expected result as parameters?

docs/plaster.md
```diff
@@ -63,27 +63,27 @@ occurrences of `ChromeAutocompleteSchemeClassifier` with
...
 description = ''
-re_pattern = ''
+# One of either pattern or re_pattern must be specified
+pattern = ''  # non-regex pattern (string will be escaped)
+re_pattern = ''  # regex pattern
 replace = ''
-re_flags = ''  # These are traditional python regex flags.
-count = 0
+re_flags = ''  # python regex flags ASCII, IGNORECASE, etc...
+count = 1  # use -1 to ignore the match count and replace all
```
mkarolin: I think this change probably belongs to https://github.com/brave/brave-core/pull/31238?

bridiver: yep, good catch

tools/cr/plaster_test.py
bridiver: yeah I've been tweaking these tests, they're all AI generated

docs/plaster.md
```diff
@@ -63,26 +63,26 @@ occurrences of `ChromeAutocompleteSchemeClassifier` with
...
 
 The basic format for a `[[substitution]]` entry is as follow:
 
 ```toml
-[[plaster]]
+[[substition]]
 description = ''
-re_pattern = ''
+# One of either pattern or re_pattern must be specified
+pattern = ''  # non-regex pattern (string will be escaped)
```
cdesouza-chromium, line 82: This is the part where python object files were better, because it allowed calling `re.escape` in the JSON-like file itself.

tools/cr/plaster_test.py
bridiver: There are a bunch of new tests now from different branches so I'm going to follow-up to clean this stuff up

PR #31238, opened 2025-09-16
Check plaster config count against num_changes

bridiver:
Resolves


tools/cr/plaster.py
```diff
@@ -258,7 +258,7 @@ def apply(self, dry_run=False):
         for substitution in plaster_file.get('substitution', []):
             pattern = substitution.get('re_pattern', '')
             replace = substitution.get('replace', '')
-            count = substitution.get('count', 0)
+            count = substitution.get('count', 1)
```
mkarolin: Does this mean if count is not specified we default to only the first match?

tools/cr/plaster.py
```diff
@@ -281,6 +281,10 @@ def apply(self, dry_run=False):
                 raise ValueError(
                     f'No matches found for pattern {pattern} in {info.source}')
 
+            if count not in (num_changes, -1, 0):
```
mkarolin: So if we are using `-1` to mean all matches, should 0 be a valid config?

tools/cr/plaster.py
bridiver: yes, I think that's safer than allowing any number of matches by default?

tools/cr/plaster.py
bridiver: something got messed up here in a rebase I think, looking at it again

bridiver: this should be `if num_changes not in (count, -1, 0):`

bridiver: and https://github.com/brave/brave-core/pull/31238/files#diff-772ca294e499082c33b318d08eacd858594cfd11a2a7c7bc6748f9c20f237cc3R280 should go away

tools/cr/plaster.py
bridiver: what I worry about is someone not specifying this and having a single match, but then there's some change to the upstream file and it gets applied somewhere that it shouldn't. I think in those cases it's better to verify that it's correct. In cases where we know we want to replace all instances we can specify `-1`

tools/cr/plaster.py
mkarolin: It seems we should have 2 checks:
1. To verify the parameter is correct: `count > 0 || count == -1`
Then, for `-1` case, we need to re-assign count to be 0 since `re.subn` won't understand `-1`
2. To verify the number of substitutions is correct: 
    a. if (count > 0) then `num_changes == count`
    b. if (count == 0) then `num_changes != 0`

bridiver: actually I think we should always pass 0 to `subn` and control what it actually matches with the regex. If we only want to match one of them we should add enough context that it doesn't match the others

tools/cr/plaster.py
```diff
@@ -274,12 +274,13 @@ def apply(self, dry_run=False):
                                             replace,
                                             contents,
                                             flags=re_flags,
-                                            count=count)
+                                            count=0)
```
mkarolin: Wouldn't this always pass 0?

bridiver: yes, that's what I want. The goal is not to limit the number of matches here, we should only limit the number of matches by providing appropriate context in the regex and then check to make sure it only matched what we expected

bridiver: I'll add comments

PR #31240, opened 2025-09-16
[plaster] Allow any valid regex flag

bridiver:
Replace hard-coded list of regex flags with check against python regex lib


mkarolin (approved):
LGTM

PR #31242, opened 2025-09-16
Add Brave Origin Policy Manager

bbondy:
Resolves https://github.com/brave/brave-browser/issues/49343

### DIFF
--- a/components/brave_origin/brave_origin_policy_manager.cc
+++ b/components/brave_origin/brave_origin_policy_manager.cc
@@ -1,0 +1,168 @@
+/* Copyright (c) 2025 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "brave/components/brave_origin/brave_origin_policy_manager.h"
+
+#include "base/containers/map_util.h"
+#include "base/logging.h"
+#include "base/no_destructor.h"
+#include "brave/components/brave_origin/brave_origin_utils.h"
+#include "brave/components/brave_origin/pref_names.h"
+#include "components/prefs/pref_service.h"
+
+namespace brave_origin {
+
+// static
+BraveOriginPolicyManager* BraveOriginPolicyManager::GetInstance() {
+  static base::NoDestructor<BraveOriginPolicyManager> instance;
+  return instance.get();
+}
+
+void BraveOriginPolicyManager::Init(
+    BraveOriginPolicyMap&& browser_policy_definitions,
+    BraveOriginPolicyMap&& profile_policy_definitions,
+    PrefService* local_state) {
+  CHECK(!initialized_) << "BraveOriginPolicyManager already initialized";
+  CHECK(local_state) << "BraveOriginPolicyManager local state should exist";
+
+  browser_policy_definitions_ = std::move(browser_policy_definitions);
+  profile_policy_definitions_ = std::move(profile_policy_definitions);
+  local_state_ = local_state;
+  initialized_ = true;
+
+  // Notify observers that policies are now ready
+  for (auto& observer : observers_) {
+    observer.OnBraveOriginPoliciesReady();
+  }
+}
+
+void BraveOriginPolicyManager::AddObserver(Observer* observer) {
+  observers_.AddObserver(observer);
+
+  // If local state is already available, notify immediately
+  if (local_state_) {
+    observer->OnBraveOriginPoliciesReady();
+  }
+}
+
+void BraveOriginPolicyManager::RemoveObserver(Observer* observer) {
+  observers_.RemoveObserver(observer);
+}
+
+std::optional<bool> BraveOriginPolicyManager::GetPolicyValue(
+    std::string_view pref_name,
+    std::optional<std::string_view> profile_id) const {
+  CHECK(initialized_) << "BraveOriginPolicyManager not initialized";
+  CHECK(local_state_) << "BraveOriginPolicyManager local state should exist";
+
+  // First check browser-level policies
+  const BraveOriginPolicyInfo* policy_info =
+      base::FindOrNull(browser_policy_definitions_, pref_name);
+
+  // If not found in browser policies, check profile policies
+  if (!policy_info) {
+    policy_info = base::FindOrNull(profile_policy_definitions_, pref_name);
+  }
+
+  if (!policy_info) {
+    LOG(ERROR) << "Unknown pref name: " << pref_name;
+    return std::nullopt;
+  }
+
+  // Get policies dict once and pass to internal helper
+  const base::Value::Dict& policies_dict =
+      local_state_->GetDict(kBraveOriginPolicies);
+  return GetPolicyValueInternal(*policy_info, policies_dict, profile_id);
+}
+
+PoliciesEnabledMap BraveOriginPolicyManager::GetAllBrowserPolicies() const {
+  CHECK(initialized_) << "BraveOriginPolicyManager not initialized";
+  CHECK(local_state_) << "BraveOriginPolicyManager local state should exist";
+
+  PoliciesEnabledMap policies;
+
+  // Get policies dict once for all lookups
+  const base::Value::Dict& policies_dict =
+      local_state_->GetDict(kBraveOriginPolicies);
+
+  for (const auto& [pref_name, policy_info] : browser_policy_definitions_) {
+    auto value = GetPolicyValueInternal(policy_info, policies_dict);
+    policies[policy_info.policy_key] = value;
+  }
+  return policies;
+}
+
+PoliciesEnabledMap BraveOriginPolicyManager::GetAllProfilePolicies(
+    std::string_view profile_id) const {
+  CHECK(initialized_) << "BraveOriginPolicyManager not initialized";
+  CHECK(local_state_) << "BraveOriginPolicyManager local state should exist";
+
+  PoliciesEnabledMap policies;
+
+  // Get policies dict once for all lookups
+  const base::Value::Dict& policies_dict =
+      local_state_->GetDict(kBraveOriginPolicies);
+
+  for (const auto& [pref_name, policy_info] : profile_policy_definitions_) {
+    auto value = GetPolicyValueInternal(policy_info, policies_dict, profile_id);
+    policies[policy_info.policy_key] = value;
+  }
+  return policies;
+}
+
+// Helper function to get pref info from pref definitions
+const BraveOriginPolicyInfo* BraveOriginPolicyManager::GetPrefInfo(
+    std::string_view pref_name) {
+  // First check browser-level policies
+  const BraveOriginPolicyInfo* policy_info =
+      base::FindOrNull(browser_policy_definitions_, pref_name);
+
+  // If not found in browser policies, check profile policies
+  if (!policy_info) {
+    policy_info = base::FindOrNull(profile_policy_definitions_, pref_name);
+  }
+
+  return policy_info;
+}
+
+bool BraveOriginPolicyManager::IsInitialized() const {
+  return initialized_;
+}
+
+void BraveOriginPolicyManager::ResetForTesting() {
+  initialized_ = false;
+  browser_policy_definitions_.clear();
+  profile_policy_definitions_.clear();
+  local_state_ = nullptr;
+  observers_.Clear();
+}
+
+bool BraveOriginPolicyManager::GetPolicyValueInternal(
+    const BraveOriginPolicyInfo& policy_info,
+    const base::Value::Dict& policies_dict,
+    std::optional<std::string_view> profile_id) const {
+  std::string policy_key;
+  if (!profile_id.has_value()) {
+    // Browser-scoped preference
+    policy_key = GetBraveOriginBrowserPrefKey(policy_info);
+  } else {
+    // Profile-scoped preference
+    policy_key = GetBraveOriginProfilePrefKey(policy_info, profile_id.value());
+  }
+
+  const base::Value* policy_value = policies_dict.Find(policy_key);
+  if (policy_value && policy_value->is_bool()) {
+    return policy_value->GetBool();
+  }
+
+  // Return default value if no policy value found
+  return policy_info.default_value;
+}
+
+BraveOriginPolicyManager::BraveOriginPolicyManager() = default;
+
+BraveOriginPolicyManager::~BraveOriginPolicyManager() = default;
+
+}  // namespace brave_origin
--- a/components/brave_origin/brave_origin_policy_manager.h
+++ b/components/brave_origin/brave_origin_policy_manager.h
@@ -1,0 +1,102 @@
+/* Copyright (c) 2025 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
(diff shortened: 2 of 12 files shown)


components/brave_origin/brave_origin_policy_manager.h
```diff
@@ -0,0 +1,83 @@
...
+  void RemoveObserver(Observer* observer);
+
+  // Get policy value for a specific pref, using local state value or default
+  // For profile-scoped prefs, profile_id is required to construct the correct
+  // key. Returns std::nullopt if the pref is unknown.
+  std::optional<bool> GetPolicyValue(const std::string& pref_name,
+                                     const std::string& profile_id = "") const;
+
+  // Get all pref definitions (metadata only)
+  const BraveOriginPolicyMap& GetDefinitions() const;
```
DJAndries: What do you think about changing the signature to be something like `GetDefinitions(std::optional<std::string_view> profile_id)`, or splitting this function into two: `GetBrowserDefinitions()` and `GetProfileDefinitions(const std::string& profile_id)`?

The method(s) could return a map of policy keys to booleans. This could shift complexity away from the providers, and potentially allow the providers to share some of the filtering logic.

components/brave_origin/brave_origin_policy_manager.h
```diff
@@ -0,0 +1,83 @@
...
+  void Init(BraveOriginPolicyMap&& pref_definitions, PrefService* local_state);
+
+  // Add/remove observers for policy readiness notifications
+  void AddObserver(Observer* observer);
+  void RemoveObserver(Observer* observer);
+
+  // Get policy value for a specific pref, using local state value or default
+  // For profile-scoped prefs, profile_id is required to construct the correct
+  // key. Returns std::nullopt if the pref is unknown.
+  std::optional<bool> GetPolicyValue(const std::string& pref_name,
```
DJAndries: if the first suggestion is accepted, i think we can remove this method

components/brave_origin/brave_origin_policy_manager.h
```diff
@@ -0,0 +1,83 @@
...
+  const BraveOriginPolicyMap& GetDefinitions() const;
+
+  // Check if the singleton has been initialized
+  bool IsInitialized() const;
+
+  // Helper function to get pref info from pref definitions
+  const BraveOriginPolicyInfo* GetPrefInfo(const std::string& pref_name);
+
+  // Reset singleton state for testing
+  void ResetForTesting();
```
DJAndries, line 74: Potential alternative: Making the unit test a friend of this class and re-constructing the class as needed within the test

components/brave_origin/brave_origin_policy_manager.cc
```diff
@@ -0,0 +1,104 @@
...
+}
+
+void BraveOriginPolicyManager::ResetForTesting() {
+  initialized_ = false;
+  policy_definitions_.clear();
+  local_state_ = nullptr;
+  observers_.Clear();
+}
+
+BraveOriginPolicyManager::BraveOriginPolicyManager() : initialized_(false) {}
```
DJAndries: nit: could move the boolean initialization to the header, since we already do this for `local_state_`

components/brave_origin/brave_origin_policy_manager.h
```diff
@@ -0,0 +1,83 @@
...
+   public:
+    // Called when BraveOrigin policies become available or are updated.
+    virtual void OnBraveOriginPoliciesReady() = 0;
+  };
+
+  static BraveOriginPolicyManager* GetInstance();
+
+  // Initialize with pref definitions mappings from browser layer and local
+  // state
+  void Init(BraveOriginPolicyMap&& pref_definitions, PrefService* local_state);
```
DJAndries: Slightly unrelated, but now I'm wondering if it's appropriate for the service factory to create the policy map. What do you think about mvoing the static `BuildBraveOriginPolicyDefinitions` function to somewhere outside of the service factory?

components/brave_origin/brave_origin_policy_manager.h
bbondy: I think this will still be needed for up to date local state values vs the static policy definitions that have the default value only.

components/brave_origin/brave_origin_policy_manager.h
bbondy: Ya I was considering doing that too instead of storing them together.

components/brave_origin/brave_origin_policy_manager.h
bbondy: BJ asked me to put it there to avoid the various deps within `components/brave_origin` in a previous review.

components/brave_origin/brave_origin_policy_manager.h
bbondy: Forgot to comment but this is updated in the latest

components/brave_origin/brave_origin_policy_manager.cc
bbondy: Updated in the latest

components/brave_origin/brave_origin_policy_manager.h
bbondy: Split up these here and elsewhere removed the enum for global vs profile. It's in the latest

DJAndries: The split looks good.

Is it correct to say that the policy providers only need the policy key, and whether the policy is enabled? If so, it may be prudent to change the return type to `base:flat_map<std::string_view, bool>` or something similar.

components/brave_origin/brave_origin_policy_manager.h
DJAndries: If https://github.com/brave/brave-core/pull/31242#discussion_r2357619891 makes sense then would that eliminate the need to call `GetPolicyValue`?

components/brave_origin/brave_origin_policy_manager.h
```diff
@@ -0,0 +1,88 @@
...
+// access to policy values from local state. This completely abstracts away
+// the local state management from policy providers.
+class BraveOriginPolicyManager {
+ public:
+  // Observer interface for objects that need to be notified when
+  // BraveOrigin policies are loaded/changed.
+  class Observer : public base::CheckedObserver {
+   public:
+    // Called when BraveOrigin policies become available or are updated.
+    virtual void OnBraveOriginPoliciesReady() = 0;
```
DJAndries, line 37: This is semi-related to the draft you created earlier, but do you plan to call this observer method when `BraveOriginPolicyManager::SetPolicyValue` is called?

components/brave_origin/brave_origin_policy_manager.h
```diff
@@ -0,0 +1,88 @@
...
+  friend base::NoDestructor<BraveOriginPolicyManager>;
+  friend class BraveOriginPolicyManagerTest;
+
+  BraveOriginPolicyManager();
+  ~BraveOriginPolicyManager();
+
+  bool initialized_ = false;
+  BraveOriginPolicyMap browser_policy_definitions_;
+  BraveOriginPolicyMap profile_policy_definitions_;
+  raw_ptr<PrefService> local_state_ = nullptr;
```
DJAndries, line 96: During browser shutdown, are there any issues with holding a `PrefService` pointer in a singleton? I don't think there are, but thought I'd check.

components/brave_origin/brave_origin_policy_manager_unittest.cc
```diff
@@ -0,0 +1,288 @@
...
+  void SetPolicyInLocalState(const std::string& key, bool value) {
+    base::Value::Dict policies_dict =
+        pref_service_->GetDict(kBraveOriginPolicies).Clone();
+    policies_dict.Set(key, value);
+    pref_service_->SetDict(kBraveOriginPolicies, std::move(policies_dict));
+  }
+
+  base::test::TaskEnvironment task_environment_;
+  std::unique_ptr<TestingPrefServiceSimple> pref_service_;
+  raw_ptr<BraveOriginPolicyManager> manager_ = nullptr;
```
DJAndries: re: https://github.com/brave/brave-core/pull/31242#discussion_r2353882534

I was thinking this could just be a `std::unique_ptr<BraveOriginPolicyManager>` that you construct in the `SetUp` method. This would allow you to remove the manual resetting code in `TearDown`. It could make the unit test more robust, especially in the event that we wish to add additional members to the policy manager, since the author will not have to worry about updating the reset code in this test.

components/brave_origin/brave_origin_policy_manager_unittest.cc
```diff
@@ -0,0 +1,288 @@
...
+
+  void SetPolicyInLocalState(const std::string& key, bool value) {
+    base::Value::Dict policies_dict =
+        pref_service_->GetDict(kBraveOriginPolicies).Clone();
+    policies_dict.Set(key, value);
+    pref_service_->SetDict(kBraveOriginPolicies, std::move(policies_dict));
+  }
+
+  base::test::TaskEnvironment task_environment_;
+  std::unique_ptr<TestingPrefServiceSimple> pref_service_;
```
DJAndries: nit: `std::unique_ptr` might not be needed here

components/brave_origin/brave_origin_policy_manager_unittest.cc
DJAndries: Actually, you might be able to get away with not using `std::unique_ptr` since the policy manager has a default constructor. The declaration could simply be: 

```
BraveOriginPolicyManager manager_;
```

components/brave_origin/brave_origin_policy_manager_unittest.cc
```diff
@@ -0,0 +1,288 @@
...
+};
+
+class TestObserver : public BraveOriginPolicyManager::Observer {
+ public:
+  void OnBraveOriginPoliciesReady() override { ready_called_count_++; }
+
+  int ready_called_count() const { return ready_called_count_; }
+
+ private:
+  int ready_called_count_ = 0;
```
DJAndries: Perhaps this is an answer to my question in https://github.com/brave/brave-core/pull/31242/files#r2357632544. Does a count (as opposed to a boolean like `ready_called` or something similar) imply that the ready method can be called multiple times per session?

components/brave_origin/brave_origin_policy_manager.h
bbondy: I think it only really needs the policy key and the value. But that return is a static definition, so it needs to get the latest values from local state.  The get method is still needed for the keyed service because that will get things 1 value at a time. 
I'll remove these though and just make a function to get all the keys and values without other data. It'll have to check local state for each too.

components/brave_origin/brave_origin_policy_manager.h
bbondy: yep correct.

components/brave_origin/brave_origin_policy_manager.h
bbondy: This will be used by the keyed service still, so will be kept, but it's refactored in the next push so it doesn't use it from the policy providers.

components/brave_origin/brave_origin_policy_manager.h
bbondy: I don't think there's an issue.

components/brave_origin/brave_origin_policy_manager.h
bbondy: No this is just used by the policy observers to know they can setup the policies. Because the policy providers are created super early in startup, even before feature flags exist.  So when it's ready to with definitions, it signals to the providers that it's ready to setup policies. Then the bundle for policies is created and a feature check can at that point be made to see if Brave Origin exists.

bbondy: What you're thinking is a new observer even that I mentioned that I will also add to this interface in a future PR that will notify the policy providers to do a refresh of values.  
This one is only used for the original setup. It's possible that future PR will combine both of those observer events though. But for now this one is only for setup.

components/brave_origin/brave_origin_policy_manager_unittest.cc
bbondy: A boolean makes more sense for now since it's only for setup. Will update.

components/brave_origin/brave_origin_policy_manager.h
bbondy: I think ResetForTesting is probably cleaner, so went back to that after a detour.

components/brave_origin/brave_origin_policy_manager_unittest.cc
bbondy: Since it's a singleton I think it's better for the tests to use the singleton

components/brave_origin/brave_origin_policy_manager_unittest.cc
bbondy: updated

components/brave_origin/brave_origin_policy_manager_unittest.cc
bbondy: updated

components/brave_origin/brave_origin_policy_manager.h
```diff
@@ -0,0 +1,100 @@
...
+
+  // Get policy value for a specific pref, using local state value or default
+  // For profile-scoped prefs, profile_id is required to construct the correct
+  // key. Returns std::nullopt if the pref is unknown.
+  std::optional<bool> GetPolicyValue(const std::string& pref_name,
+                                     const std::string& profile_id = "") const;
+
+  // Get browser-level policy key-value pairs (policy values from local state or
+  // defaults)
+  base::flat_map<std::string, bool> GetAllBrowserPolicies() const;
```
DJAndries: nit: perhaps we should add a type alias?

components/brave_origin/brave_origin_policy_manager.cc
```diff
@@ -0,0 +1,176 @@
...
+  local_state_ = nullptr;
+  observers_.Clear();
+}
+
+std::optional<bool> BraveOriginPolicyManager::GetPolicyValueInternal(
+    const BraveOriginPolicyInfo& policy_info,
+    const base::Value::Dict& policies_dict,
+    const std::string& profile_id) const {
+  std::string policy_key;
+  if (profile_id.empty()) {
```
DJAndries: nit: using an optional for profile_id could be a good fit here, but it's probably fine as is

components/brave_origin/brave_origin_policy_manager.cc
```diff
@@ -0,0 +1,176 @@
...
+    policy_key = GetBraveOriginProfilePrefKey(policy_info, profile_id);
+  }
+
+  const base::Value* policy_value = policies_dict.Find(policy_key);
+  if (policy_value && policy_value->is_bool()) {
+    return policy_value->GetBool();
+  }
+
+  // Return default value if no policy value found
+  return policy_info.default_value;
```
DJAndries, line 161: Would it be correct to say that this method always returns a bool, and never returns `std::nullopt`? If so, consider simplifying the return type

components/brave_origin/brave_origin_policy_manager.h
bbondy: added

components/brave_origin/brave_origin_policy_manager.cc
bbondy: updated to optional and string_view. Also strinvg_view elsewhere

components/brave_origin/brave_origin_policy_manager.cc
bbondy: ya that was a relic of an older version of the function. Now always return boolean

DJAndries (approved):

PR #31138, opened 2025-09-10
Search engine selection dialog should use previous style in Japan.

mkarolin:
Resolves https://github.com/brave/brave-browser/issues/49215

PR #31243, opened 2025-09-16
Search engine selection dialog should use previous style in Japan. (#31138)

mkarolin:
Resolves https://github.com/brave/brave-browser/issues/49215
Uplift from https://github.com/brave/brave-core/pull/31138


LaurenWags (approved):
Uplift to 1.83.x is approved 👍🏻 QA has verified the PR on Nightly using 1.84.x as per https://github.com/brave/brave-core/pull/31138#issuecomment-3304280031.

PR #31248, opened 2025-09-16
fix(privacy): Farbling not working as expected

StephenHeaps:
- Incorrectly checking if Block Scripts is enabled, instead of Block Fingerprinting. If Block Scripts is not enabled (the default), the farbling protection script is not injected.
    - Regressed in refactoring for https://github.com/brave/brave-core/pull/30104

Resolves https://github.com/brave/brave-browser/issues/49346

### Testplan
1. Visit https://dev-pages.brave.software/fingerprinting/farbling.html
2. Tap `Generate fingerprints`. Make note of values for Plugins, Hardware concurrency and Speech Voices.
3. Quit the app and re-launch
4. On same page as step 1, tap `Generate fingerprints`.
5. Verify values differ from step 2.

### DIFF
--- a/ios/brave-ios/Sources/Brave/Frontend/Browser/BrowserViewController/BVC+TabPolicyDecider.swift
+++ b/ios/brave-ios/Sources/Brave/Frontend/Browser/BrowserViewController/BVC+TabPolicyDecider.swift
@@ -276,7 +276,7 @@ extension BrowserViewController: TabPolicyDecider {
       let isBlockFingerprintingEnabled =
         tab.braveShieldsHelper?.isShieldExpected(
           for: mainDocumentURL,
-          shield: .noScript,
+          shield: .fpProtection,
           considerAllShieldsOption: true
         ) ?? true
 


kylehickinson (approved):

nuo-xu (approved):

hffvld:
Verified on `iPhone 14` using version(s):
```
Device/OS: iPhone 14 / iOS 17.7.2
Brave build: 1.84 (79)
BraveCore: 1.84.79 (140.0.7339.133)
```
---------------------------

**STEPS:**
1. Follow the STR/TP from https://github.com/brave/brave-core/pull/31248#issue-3423977984
2. *Verify*

**ACTUAL RESULTS:**
- Verified that the values on both pages are different.
- Verified that the values are different after Brave relaunch.

---------------------------

| `Before relaunch / Page 1` | `Before relaunch / Page 2` |
|:-:|:-:|
|  |  |
| `After relaunch / Page 1` | `After relaunch / Page 2` |
|  |  |

PR #31249, opened 2025-09-16
Brave origin working

bbondy:
Resolves


browser/brave_browser_process_impl.cc
```diff
@@ -67,6 +67,8 @@
 #include "brave/browser/ai_chat/ai_chat_agent_profile_manager.h"
 #endif
 
+#include "brave/components/brave_origin/brave_origin_policy_manager.h"
```
DJAndries: nit: this include could be placed around line 33

components/brave_origin/brave_origin_service.cc
```diff
@@ -5,43 +5,127 @@
...
+BraveOriginService::BraveOriginService(PrefService* local_state,
+                                       PrefService* profile_prefs,
+                                       std::string_view profile_id,
+                                       policy::PolicyService* policy_service)
+    : local_state_(local_state),
+      profile_prefs_(profile_prefs),
+      profile_id_(profile_id),
+      policy_service_(policy_service) {
+  CHECK(local_state_);
+  CHECK(profile_prefs_);
```
DJAndries, line 31: Might be good to check that `profile_id` is not empty

browser/brave_origin/brave_origin_service_factory.cc
```diff
@@ -150,8 +160,20 @@ BraveOriginServiceFactory::~BraveOriginServiceFactory() = default;
...
-  std::string profile_id = GetProfileId(context->GetPath());
-  return std::make_unique<BraveOriginService>(profile_id);
+  Profile* profile = Profile::FromBrowserContext(context);
+
+  // Lazy initialization of BraveOriginPolicyManager
+  auto* policy_manager = BraveOriginPolicyManager::GetInstance();
+  std::string profile_id = GetProfileId(profile->GetPath());
+  if (!policy_manager->IsInitialized()) {
+    policy_manager->Init(GetBrowserPolicyDefinitions(),
+                         GetProfilePolicyDefinitions(profile_id),
```
DJAndries: I think we can remove the profile id parameter since it's not being used in the function.

browser/brave_origin/brave_origin_service_factory.cc
```diff
@@ -150,8 +160,20 @@ BraveOriginServiceFactory::~BraveOriginServiceFactory() = default;
...
+                         g_browser_process->local_state());
+  }
+
+  return std::make_unique<BraveOriginService>(
+      g_browser_process->local_state(), profile->GetPrefs(), profile_id,
+      GetPolicyServiceFromProfile(profile));
```
DJAndries, line 176: Do we also need to get the machine-level policy service?

components/brave_origin/brave_origin_service.cc
```diff
@@ -5,43 +5,127 @@
...
+      BraveOriginPolicyManager::GetInstance()->GetPrefInfo(pref_name);
+  if (!pref_info) {
+    return false;
+  }
+
+  // Check if the active policy source is POLICY_SOURCE_BRAVE
+  const policy::PolicyMap& policies = policy_service_->GetPolicies(
+      policy::PolicyNamespace(policy::POLICY_DOMAIN_CHROME, std::string()));
+
+  const policy::PolicyMap::Entry* entry = policies.Get(pref_info->policy_key);
```
DJAndries, line 55: Will this check work for machine-level policies?

components/brave_origin/brave_origin_policy_manager.cc
```diff
@@ -112,6 +113,49 @@ PoliciesEnabledMap BraveOriginPolicyManager::GetAllProfilePolicies(
...
+      base::FindOrNull(browser_policy_definitions_, pref_name);
+  if (!policy_info) {
+    LOG(ERROR) << "Unknown browser pref name: " << pref_name;
+    return;
+  }
+
+  // Update the value in the dictionary
+  ScopedDictPrefUpdate update(local_state_, kBraveOriginPolicies);
+  std::string browser_key = GetBraveOriginBrowserPrefKey(*policy_info);
+  update->Set(browser_key, base::Value(value));
```
DJAndries: nit: the `base::Value` construction could be removed here

components/brave_origin/brave_origin_policy_manager.cc
```diff
@@ -112,6 +113,49 @@ PoliciesEnabledMap BraveOriginPolicyManager::GetAllProfilePolicies(
...
+}
+
+void BraveOriginPolicyManager::SetProfilePolicyValue(
```
DJAndries, line 122: Since the function above has the same logic, consider consolidating the two functions with a `std::optional<std::string_view> profile_id` parameter.

components/brave_origin/brave_origin_policy_manager.cc
```diff
@@ -112,6 +113,49 @@ PoliciesEnabledMap BraveOriginPolicyManager::GetAllProfilePolicies(
...
 }
 
+void BraveOriginPolicyManager::SetBrowserPolicyValue(std::string_view pref_name,
+                                                     bool value) {
+  CHECK(initialized_) << "BraveOriginPolicyManager not initialized";
+  CHECK(local_state_) << "BraveOriginPolicyManager local state should exist";
+
+  // Check if this is a valid browser-level pref
+  const BraveOriginPolicyInfo* policy_info =
+      base::FindOrNull(browser_policy_definitions_, pref_name);
```
DJAndries: I still wonder if it would be optimal if we keyed policy definitions by policy key, rather than pref name. Thoughts?

components/brave_origin/brave_origin_service.cc
```diff
@@ -5,43 +5,127 @@
...
+  // Set the policy value in BraveOriginPolicyManager
+  BraveOriginPolicyManager::GetInstance()->SetBrowserPolicyValue(pref_name,
+                                                                 value);
+
+  // Also set the corresponding pref value in local_state
+  // If not user-settable and value equals default, clear the pref instead
+  // We don't want a user to have a value set in local state if the value is not
+  // user settable as they would be stuck with that value if they stopped being
+  // a Brave Origin user
+  if (!pref_info->user_settable && value == pref_info->default_value) {
```
DJAndries, line 80: What if we checked the value against `prefs->GetDefaultPrefValue(pref_name)` instead of storing the default value in the pref info?

components/brave_origin/brave_origin_service.cc
```diff
@@ -5,43 +5,127 @@
...
+  } else {
+    local_state_->SetBoolean(pref_info->pref_name, value);
+  }
+
+  return true;
 }
 
-base::Value BraveOriginService::GetBraveOriginPrefValue(
-    const std::string& pref_name) const {
+bool BraveOriginService::SetProfilePolicyValue(std::string_view pref_name,
```
DJAndries, line 89: Similar to my comment on the policy manager, consider consolidating this method and the one above.

components/brave_origin/brave_origin_service.cc
```diff
@@ -5,43 +5,127 @@
...
+                                               bool value) {
   if (!IsBraveOriginEnabled()) {
-    return base::Value();
+    return false;
+  }
+
+  // Get policy info to access pref_key and user_settable
+  const BraveOriginPolicyInfo* pref_info =
+      BraveOriginPolicyManager::GetInstance()->GetPrefInfo(pref_name);
```
DJAndries: Should we just hold a reference to the policy manager as a class member rather than calling `GetInstance` here and elsewhere?

components/brave_policy/brave_browser_policy_provider.cc
```diff
@@ -20,16 +20,18 @@ namespace brave_policy {
...
   policy::ConfigurationPolicyProvider::Init(registry);
 
-  // Trigger immediate policy loading to ensure policies are available in
-  // chrome://policy
-  // The actual policies will be conditionally added in LoadPolicies.
-  RefreshPolicies(policy::PolicyFetchReason::kBrowserStart);
+  // Register as BraveOriginPolicyManager observer.
+  // This ensures feature flags and local state are available before policy
+  // loading.
+  brave_origin::BraveOriginPolicyManager::GetInstance()->AddObserver(this);
```
DJAndries: Consider holding a `base::ScopedObservation` rather than calling `AddObserver`/`RemoveObserver` explicitly.

components/brave_origin/brave_origin_policy_manager.cc
```diff
@@ -112,6 +113,49 @@ PoliciesEnabledMap BraveOriginPolicyManager::GetAllProfilePolicies(
...
+  if (!policy_info) {
+    LOG(ERROR) << "Unknown profile pref name: " << pref_name;
+    return;
+  }
+
+  // Update the value in the dictionary with profile-scoped key
+  ScopedDictPrefUpdate update(local_state_, kBraveOriginPolicies);
+  std::string profile_key =
+      GetBraveOriginProfilePrefKey(*policy_info, profile_id);
+  update->Set(profile_key, base::Value(value));
```
DJAndries: Do we notify the policy providers of changes in this PR, or will that be in a future PR?

DJAndries: Also, will we notify the policy providers of changes to the Origin enabled status?

components/brave_origin/brave_origin_service.cc
bbondy: The default for Brave Origin is not the same as the default of the pref without origin.

components/brave_origin/brave_origin_service.cc
bbondy: I don't think there's a benefit to that

components/brave_origin/brave_origin_policy_manager.cc
bbondy: > Do we notify the policy providers of changes in this PR, or will that be in a future PR?

That will be in a future PR. 

> Also, will we notify the policy providers of changes to the Origin enabled status?

No plans yet but we might eventually.

components/brave_origin/brave_origin_policy_manager.cc
bbondy: The lookups by the mojom functions will be by pref name.

components/brave_origin/brave_origin_policy_manager.cc
bbondy: I'll refactor out the common logic but keep them separate since one of them CHECKs for profile id and they operate on different policy definitions.

browser/brave_origin/brave_origin_service_factory.cc
bbondy: I believe that it will include browser level policies too here but if not I'll address it in a different PR.

components/brave_origin/brave_origin_service.cc
bbondy: See above.

browser/brave_browser_process_impl.cc
bbondy: Fixed in b2bdea0c21691ba6d58fa6f59f2f7bd931d3fac4

components/brave_origin/brave_origin_service.cc
bbondy: Fixed in b2bdea0c21691ba6d58fa6f59f2f7bd931d3fac4

browser/brave_origin/brave_origin_service_factory.cc
bbondy: Good catch. Fixed in b2bdea0c21691ba6d58fa6f59f2f7bd931d3fac4

components/brave_origin/brave_origin_policy_manager.cc
bbondy: Fixed in b2bdea0c21691ba6d58fa6f59f2f7bd931d3fac4

components/brave_origin/brave_origin_policy_manager.cc
bbondy: Fixed in b2bdea0c21691ba6d58fa6f59f2f7bd931d3fac4

components/brave_origin/brave_origin_policy_manager.cc
bbondy: Let's wait until after the mojom stuff lands then see if it makes sense.

components/brave_origin/brave_origin_service.cc
bbondy: I think these are different enough to warrant different implementations.  Probably a helper internal function to refactor the get pref info and is origin check makes the code harder to read. Would prefer to leave this one.

browser/brave_origin/brave_origin_service_factory.cc
DJAndries: Sounds good. It looks like the browser policy connector owns its own policy service: https://source.chromium.org/chromium/chromium/src/+/main:components/policy/core/browser/browser_policy_connector_base.h;drc=e63cbaf6fdcfdac0f04883d956b00e83330bd0a2;l=122

components/brave_origin/brave_origin_service.cc
DJAndries: No performance benefit, just a bit more concise. Feel free to resolve if you wish to keep it as is.

components/brave_origin/brave_origin_policy_manager.cc
DJAndries: > > Also, will we notify the policy providers of changes to the Origin enabled status?
> 
> No plans yet but we might eventually.

How do we plan to handle Origin subscription cancellations during runtime in the MVP? Will we prompt the user to restart their browser in order to update policies?

components/brave_origin/brave_origin_policy_manager.cc
DJAndries: I would be curious to see if the mojo interface could lookup via policy key as well. The reduction in complexity could make it worth it. Happy to discuss later.

components/brave_origin/brave_origin_service.cc
DJAndries: Ah, makes sense

components/brave_origin/brave_origin_service.cc
bbondy: Prefer it as is, so resolving

components/brave_policy/brave_browser_policy_provider.cc
bbondy: Forgot to reply but this was fixed in the last commit too.

components/brave_origin/brave_origin_policy_manager.cc
bbondy: we might make it dynamic but that's definitely a possibility too

components/brave_origin/brave_origin_service.cc
bbondy: I did change the 2 functions that had the BraveOriginPolicyManager ::GetInstance to store it in an `auto* manager` and just pushed though. But I think it's more readable and standard to show it's a singleton to not have this in a class variable.

chromium_src/chrome/browser/profiles/profile.cc
```diff
@@ -68,3 +69,15 @@ bool Profile::OTRProfileID::IsSearchBackupResults() const {
...
                           base::CompareCase::SENSITIVE);
 }
+
+// This is to avoid a circular dep on chrome/browser in the Factory for
+// the brave_origin keyed service.
+namespace brave_origin {
+policy::PolicyService* GetPolicyServiceFromProfile(Profile* profile) {
+  policy::PolicyService* policy_service = nullptr;
+  if (auto* connector = profile->GetProfilePolicyConnector()) {
+    policy_service = connector->policy_service();
```
DJAndries: nit: you could `return connector->policy_service()` here and `return nullptr` at the end of the function

components/brave_origin/brave_origin_policy_manager.cc
```diff
@@ -112,6 +113,49 @@ PoliciesEnabledMap BraveOriginPolicyManager::GetAllProfilePolicies(
...
+    const BraveOriginPolicyMap& policy_definitions,
+    std::optional<std::string_view> profile_id) {
+  CHECK(initialized_) << "BraveOriginPolicyManager not initialized";
+  CHECK(local_state_) << "BraveOriginPolicyManager local state should exist";
+
+  // Check if this is a valid pref
+  const BraveOriginPolicyInfo* policy_info =
+      base::FindOrNull(policy_definitions, pref_name);
+  if (!policy_info) {
+    LOG(ERROR) << "Unknown " << (profile_id.has_value() ? "profile" : "browser")
```
DJAndries, line 145: Is this error log permanent? If so, should we change it to `VLOG` to avoid noise?

components/brave_policy/brave_profile_policy_provider.cc
```diff
@@ -22,16 +22,18 @@
...
   policy::ConfigurationPolicyProvider::Init(registry);
 
-  // Trigger immediate policy loading to ensure policies are available in
-  // chrome://policy
-  // The actual policies will be conditionally added in LoadPolicies.
-  RefreshPolicies(policy::PolicyFetchReason::kBrowserStart);
+  // Register as BraveOriginPolicyManager observer.
+  // This ensures feature flags and local state are available before policy
+  // loading.
+  brave_origin::BraveOriginPolicyManager::GetInstance()->AddObserver(this);
```
DJAndries: This class could also use scoped observation

components/brave_origin/brave_origin_policy_manager.cc
bbondy: more like a warning of someone trying to add code that looks up the wrong value. I don't want it to be a CHECK though in case we get rid of a preference at some point and something uses this. I think they probably wouldn't see a `VLOG` and I think the `LOG(ERROR)` won't be hit normally

components/brave_policy/brave_profile_policy_provider.cc
bbondy: Sorry should have did that earlier. updated

chromium_src/chrome/browser/profiles/profile.cc
bbondy: updated in the latest

components/brave_origin/brave_origin_policy_manager_unittest.cc
```diff
@@ -282,4 +282,137 @@ TEST_F(BraveOriginPolicyManagerTest, RemoveObserver_StopsNotifications) {
...
 }
 
+TEST_F(BraveOriginPolicyManagerTest, SetBrowserPolicyValue_UpdatesValue) {
+  InitializeManager();
+
+  auto* manager = BraveOriginPolicyManager::GetInstance();
+
+  // Initially should return default value (false)
+  auto result = manager->GetPolicyValue("test.global.pref");
+  EXPECT_TRUE(result.has_value());
```
DJAndries: nit: `ASSERT_TRUE` may be more appropriate to prevent crashing on the subsequent `result.value()` call

components/brave_origin/brave_origin_policy_manager_unittest.cc
```diff
@@ -282,4 +282,137 @@ TEST_F(BraveOriginPolicyManagerTest, RemoveObserver_StopsNotifications) {
...
+  auto* manager = BraveOriginPolicyManager::GetInstance();
+  const std::string profile_id = "test-profile";
+
+  // Initially should return default value (true)
+  auto result = manager->GetPolicyValue("test.profile.pref", profile_id);
+  EXPECT_TRUE(result.has_value());
+  EXPECT_TRUE(result.value());
+
+  // Set the value to false
+  manager->SetProfilePolicyValue("test.profile.pref", false, profile_id);
```
DJAndries: nit: consider creating consts for these test strings like you did for `profile_id` in the test below

components/brave_origin/brave_origin_service_unittest.cc
```diff
@@ -0,0 +1,513 @@
...
+    // Enable the BraveOrigin feature for tests
+    feature_list_.InitAndEnableFeature(features::kBraveOrigin);
+
+    // Register the BraveOrigin policies dictionary pref in local_state
+    local_state_.registry()->RegisterDictionaryPref(kBraveOriginPolicies);
+
+    // Register test browser preferences in local_state
+    // These are needed because BraveOriginService::SetBrowserPolicyValue()
+    // calls local_state_->SetBoolean() on these prefs after updating the policy
+    local_state_.registry()->RegisterBooleanPref("test.browser.pref", false);
```
DJAndries: These pref strings are used in most of the tests. consts would be a good fit.

emerick (approved):
++chromium_src

components/brave_origin/brave_origin_service_unittest.cc
bbondy: Updated in the latest commit

components/brave_origin/brave_origin_policy_manager_unittest.cc
bbondy: Updated in the latest commit

components/brave_origin/brave_origin_policy_manager_unittest.cc
bbondy: Good call. Updated there and elsewhere in the latest commit

DJAndries: I see changes in the test above, but not in this test

components/brave_origin/brave_origin_policy_manager_unittest.cc
DJAndries: `test.profile.pref` and `test.global.pref` string literals are still present in this test

bbondy: Should have them all now

components/brave_origin/brave_origin_policy_manager_unittest.cc
bbondy: Should have them all now

bbondy:
FYI I also added a commit to have the dcheck only apply to the policy providers so that it doesn't mess with Release build unittests on the service. ec9a5031a06684b757264ff145849e41c4401779

components/brave_policy/brave_profile_policy_provider.cc
```diff
@@ -49,21 +50,75 @@ bool BraveProfilePolicyProvider::IsFirstPolicyLoadComplete(
...
+  }
+}
+
 policy::PolicyBundle BraveProfilePolicyProvider::LoadPolicies() {
   policy::PolicyBundle bundle;
 
-  // Future work will add policies here
+  // TODO(https://github.com/brave/brave-browser/issues/47463)
+  // Get the actual purchase state from SKU service.
+#if DCHECK_IS_ON()  // Debug builds only
```
bridiver, line 67: DCHECK is not equivalent to debug builds, but also why isn't all brave origin code behind `IsBraveOriginEnabled`? see https://github.com/brave/brave-core/pull/31485/files#r2434119991

bridiver: seems like this should just be behind the feature flag here? Or maybe you need a second feature flag for this specific function?

bbondy: Being fixed here https://github.com/brave/brave-core/pull/31869

PR #31250, opened 2025-09-17
fix(privacy): iOS not masking forced aggressive mode in shields panel

StephenHeaps:
- On some websites, we force aggressive mode when user has standard mode selected. We mask this in the Shields panel UI across all platforms.
- This change fixes a regression on iOS where this behaviour was not respected by moving the always aggressive check to the tab helper, with a function property to respect the always aggressive check as needed. In the shields panel (using view model `ShieldsSettingsViewModel.swift`), we do not respect the always aggressive check for display purposes to align with other platforms and prior iOS behaviour.

Resolves https://github.com/brave/brave-browser/issues/49348

### Testplan

1. Fresh install the app
2. Open Settings > Shields & Privacy and confirm Trackers & Ad Blocking is set to `Standard`
3. Visit `youtube.com`
4. Open shields panel, open advanced controls
5. Verify Trackers & Ad Blocking displays as `Standard`.
    - Do not change this setting or open the dropdown
6. Open Settings > Shields & Privacy and updated Trackers & Ad Blocking to `Aggressive`
7. On `youtube.com` tab, check shields panel displays Trackers & Ad Blocking as `Aggressive`
8. Open Settings > Shields & Privacy and updated Trackers & Ad Blocking to `Disabled`
9. On `youtube.com` tab, check shields panel displays Trackers & Ad Blocking as `Disabled`

### DIFF
--- a/ios/brave-ios/Sources/Brave/Frontend/Shields/ShieldsSettingsViewModel.swift
+++ b/ios/brave-ios/Sources/Brave/Frontend/Shields/ShieldsSettingsViewModel.swift
@@ -54,9 +54,14 @@ class ShieldsSettingsViewModel: ObservableObject {
     self.shieldsEnabled =
       tab.braveShieldsHelper?.isBraveShieldsEnabled(for: tab.visibleURL)
       ?? true
+    // `considerAlwaysAggressiveETLDs` is `false` to mask that we may force
+    // aggressive mode on some domains.
     self.blockAdsAndTrackingLevel =
-      tab.braveShieldsHelper?.shieldLevel(for: tab.visibleURL, considerAllShieldsOption: true)
-      ?? .standard
+      tab.braveShieldsHelper?.shieldLevel(
+        for: tab.visibleURL,
+        considerAllShieldsOption: true,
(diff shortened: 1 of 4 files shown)


hffvld:
Verified on iPhone 14` using version(s):
```
Device/OS: iPhone 14 / iOS 17.7.2
Brave build: 1.84 (79)
BraveCore: 1.84.79 (140.0.7339.133)
```
---------------------------

**STEPS:**
1. Follow the STR/TP from https://github.com/brave/brave-core/pull/31250#issue-3424129120
2. *Verify*

**ACTUAL RESULTS:**
- Verified that `Trackers & Ads Blocking` for the YT site is in sync with global settings by default:
  - If `Global` settings are set to `Aggressive`, then `Shields` settings are also set to `Aggressive`.
  - If `Global` settings are set to `Standard,` then `Shields` settings are also set to `Standard`.
  - If `Global` settings are set to `Disabled,` then `Shields` settings are also set to `Disabled`.

---------------------------

| `Standard` | `Aggressive` | `Disabled` |
|:-:|:-:|:-:|
|  |  |  |
|  |  |  |

PR #31251, opened 2025-09-17
Hide vertical tabs completely based on pref

sangwoo108:
Previously, we hid vertical tabs when the flag is on. Now we have a pref to control this feature, so we should consider the pref together.

Resolves https://github.com/brave/brave-browser/issues/49115

### DIFF
--- a/browser/ui/views/frame/vertical_tabs/vertical_tab_strip_browsertest.cc
+++ b/browser/ui/views/frame/vertical_tabs/vertical_tab_strip_browsertest.cc
@@ -1148,6 +1148,17 @@ class VerticalTabStripHideCompletelyTest : public VerticalTabStripBrowserTest {
 
   ~VerticalTabStripHideCompletelyTest() override = default;
 
+  void SetHideCompletelyWhenCollapsed(bool hide) {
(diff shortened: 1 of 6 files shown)


browser/ui/views/frame/vertical_tabs/vertical_tab_strip_browsertest.cc
```diff
@@ -1169,9 +1180,13 @@ IN_PROC_BROWSER_TEST_F(VerticalTabStripHideCompletelyTest, GetMinimumWidth) {
   // The minimum width of the region view should be 4px, which is narrower than
   // it used to be.
   EXPECT_EQ(4, region_view->GetMinimumSize().width());
+
+  // When the preference is disabled, the minimum width should be back to 8px.
```
simonhong:
```suggestion
  // When the preference is disabled, the minimum width should be back to 41px.
```

simonhong (approved):
++

browser/ui/views/frame/vertical_tabs/vertical_tab_strip_browsertest.cc
sangwoo108: Oops, thanks :)

PR #31252, opened 2025-09-17
Leo local code execution tool POC

DJAndries:
Resolves

### DIFF
--- a/browser/ai_chat/code_execution_tool.cc
+++ b/browser/ai_chat/code_execution_tool.cc
@@ -1,0 +1,174 @@
+// Copyright (c) 2025 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#include "brave/browser/ai_chat/code_execution_tool.h"
+
+#include <utility>
+
+#include "base/json/json_reader.h"
+#include "base/strings/strcat.h"
+#include "base/strings/string_util.h"
+#include "base/strings/utf_string_conversions.h"
+#include "base/task/single_thread_task_runner.h"
+#include "base/time/time.h"
+#include "base/values.h"
+#include "brave/components/ai_chat/core/browser/tools/tool_input_properties.h"
+#include "brave/components/ai_chat/core/browser/tools/tool_utils.h"
+#include "brave/components/ai_chat/core/common/mojom/ai_chat.mojom.h"
+#include "chrome/browser/profiles/profile.h"
+#include "content/public/browser/browser_context.h"
+#include "content/public/browser/navigation_controller.h"
+#include "content/public/browser/render_frame_host.h"
+#include "content/public/browser/web_contents.h"
+#include "content/public/browser/web_contents_observer.h"
+#include "ui/base/page_transition_types.h"
+#include "url/gurl.h"
+
+namespace ai_chat {
+
+// Helper class to capture console messages from WebContents
+class ConsoleMessageObserver : public content::WebContentsObserver {
+ public:
+  explicit ConsoleMessageObserver(content::WebContents* web_contents)
+      : content::WebContentsObserver(web_contents) {}
+
+  ~ConsoleMessageObserver() override = default;
+
+  void OnDidAddMessageToConsole(
+      content::RenderFrameHost* source_frame,
+      blink::mojom::ConsoleMessageLevel log_level,
+      const std::u16string& message,
+      int32_t line_no,
+      const std::u16string& source_id,
+      const std::optional<std::u16string>& untrusted_stack_trace) override {
+    console_messages_.push_back(base::UTF16ToUTF8(message));
+  }
+
+  std::string GetCombinedConsoleOutput() const {
+    if (console_messages_.empty()) {
+      return "(No console output)";
+    }
+    return base::JoinString(console_messages_, "\n");
+  }
+
+ private:
+  std::vector<std::string> console_messages_;
+};
+
+CodeExecutionTool::CodeExecutionTool(content::BrowserContext* browser_context)
+    : browser_context_(browser_context) {}
+
+CodeExecutionTool::~CodeExecutionTool() = default;
+
+std::string_view CodeExecutionTool::Name() const {
+  return mojom::kCodeExecutionToolName;
+}
+
+std::string_view CodeExecutionTool::Description() const {
+  return "Execute JavaScript code and return console.log output. "
+         "Use this tool to run JavaScript code snippets and see their output. "
+         "The code will be executed in a sandboxed environment and any "
+         "console.log statements will be captured and returned as the result.";
+}
+
+std::optional<base::Value::Dict> CodeExecutionTool::InputProperties() const {
+  return CreateInputProperties(
+      {{"program", StringProperty("The JavaScript code to execute")}});
+}
+
+std::optional<std::vector<std::string>> CodeExecutionTool::RequiredProperties()
+    const {
+  return std::vector<std::string>{"program"};
+}
+
+bool CodeExecutionTool::RequiresUserInteractionBeforeHandling() const {
+  return false;
+}
+
+bool CodeExecutionTool::SupportsConversation(bool is_temporary,
+                                             bool has_untrusted_content) const {
+  // Support all conversation types for now
+  return true;
+}
+
+void CodeExecutionTool::UseTool(const std::string& input_json,
+                                UseToolCallback callback) {
+  // Parse the input JSON
+  auto input_dict = base::JSONReader::ReadDict(input_json);
+  if (!input_dict.has_value()) {
+    std::move(callback).Run(CreateContentBlocksForText(
+        "Error: Invalid JSON input, input must be a JSON object"));
+    return;
+  }
+
+  const std::string* program = input_dict->FindString("program");
+
+  if (!program || program->empty()) {
+    std::move(callback).Run(
+        CreateContentBlocksForText("Error: Missing or empty 'program' field"));
+    return;
+  }
+
+  // Get the primary OTR profile for isolated execution
+  Profile* profile = Profile::FromBrowserContext(browser_context_);
+  Profile* otr_profile =
+      profile->GetPrimaryOTRProfile(true /* create_if_needed */);
+
+  // Create WebContents for JavaScript execution using the OTR profile
+  content::WebContents::CreateParams create_params(otr_profile);
+  web_contents_ = content::WebContents::Create(create_params);
+
+  // Set up console message observer
+  console_observer_ =
+      std::make_unique<ConsoleMessageObserver>(web_contents_.get());
+
+  // Load about:blank to ensure clean state
+  web_contents_->GetController().LoadURL(
+      GURL("about:blank"), content::Referrer(), ui::PAGE_TRANSITION_TYPED,
+      std::string());
+
+  // Wrap the JavaScript code in try-catch
+  std::string wrapped_js =
+      base::StrCat({"try {", "  ", *program, "\n", "} catch (error) {",
+                    "  console.error('Execution error:', error);", "}"});
+
+  // Execute JavaScript after a delay to ensure about:blank has loaded
+  base::SingleThreadTaskRunner::GetCurrentDefault()->PostDelayedTask(
+      FROM_HERE,
+      base::BindOnce(
+          [](base::WeakPtr<CodeExecutionTool> tool, UseToolCallback callback,
+             const std::string& js_code) {
+            if (!tool || !tool->web_contents()) {
+              std::move(callback).Run(CreateContentBlocksForText(
+                  "Error: Tool was destroyed during execution"));
+              return;
+            }
+
+            tool->web_contents()->GetPrimaryMainFrame()->ExecuteJavaScript(
+                base::UTF8ToUTF16(js_code),
+                base::BindOnce(
+                    &CodeExecutionTool::OnJavaScriptExecutionComplete, tool,
+                    std::move(callback)));
+          },
+          weak_ptr_factory_.GetWeakPtr(), std::move(callback), wrapped_js),
+      base::Milliseconds(100));
+}
+
+void CodeExecutionTool::OnJavaScriptExecutionComplete(UseToolCallback callback,
+                                                      base::Value result) {
+  // Get console output from the observer
+  std::string console_output;
+  if (console_observer_) {
+    console_output = console_observer_->GetCombinedConsoleOutput();
+  }
+
+  // Clean up the WebContents and observer after execution
+  web_contents_.reset();
+  console_observer_.reset();
+
+  std::move(callback).Run(CreateContentBlocksForText(console_output));
+}
+
+}  // namespace ai_chat
--- a/browser/ai_chat/BUILD.gn
+++ b/browser/ai_chat/BUILD.gn
@@ -21,6 +21,8 @@ static_library("ai_chat") {
     "browser_tool_provider.h",
     "browser_tool_provider_factory.cc",
     "browser_tool_provider_factory.h",
+    "code_execution_tool.cc",
+    "code_execution_tool.h",
     "tab_data_web_contents_observer.cc",
     "tab_data_web_contents_observer.h",
     "tab_tracker_service_factory.cc",
--- a/browser/ai_chat/ai_chat_service_factory.cc
+++ b/browser/ai_chat/ai_chat_service_factory.cc
@@ -83,7 +83,7 @@ AIChatServiceFactory::BuildServiceInstanceForBrowserContext(
 
   std::vector<std::unique_ptr<ToolProviderFactory>> tool_provider_factories;
   tool_provider_factories.push_back(
-      std::make_unique<BrowserToolProviderFactory>());
+      std::make_unique<BrowserToolProviderFactory>(context));
 
   return std::make_unique<AIChatService>(
       ModelServiceFactory::GetForBrowserContext(context),
--- a/browser/ai_chat/browser_tool_provider.cc
+++ b/browser/ai_chat/browser_tool_provider.cc
@@ -5,27 +5,34 @@
 
 #include "brave/browser/ai_chat/browser_tool_provider.h"
 
+#include <memory>
 #include <vector>
 
 #include "base/memory/weak_ptr.h"
+#include "brave/browser/ai_chat/code_execution_tool.h"
 #include "brave/components/ai_chat/core/browser/tools/tool.h"
+#include "content/public/browser/browser_context.h"
 
 namespace ai_chat {
 
-BrowserToolProvider::BrowserToolProvider() {
-  CreateTools();
+BrowserToolProvider::BrowserToolProvider(
+    content::BrowserContext* browser_context) {
+  CreateTools(browser_context);
 }
 
 BrowserToolProvider::~BrowserToolProvider() = default;
 
 std::vector<base::WeakPtr<Tool>> BrowserToolProvider::GetTools() {
   std::vector<base::WeakPtr<Tool>> tool_ptrs;
-  // TODO(petemill): Return some tools
+  if (code_execution_tool_) {
+    tool_ptrs.push_back(code_execution_tool_->GetWeakPtr());
+  }
   return tool_ptrs;
 }
 
-void BrowserToolProvider::CreateTools() {
-  // TODO(petemill): Construct some tools and own them
+void BrowserToolProvider::CreateTools(
+    content::BrowserContext* browser_context) {
+  code_execution_tool_ = std::make_unique<CodeExecutionTool>(browser_context);
 }
 
 }  // namespace ai_chat
--- a/browser/ai_chat/browser_tool_provider.h
+++ b/browser/ai_chat/browser_tool_provider.h
@@ -6,13 +6,20 @@
 #ifndef BRAVE_BROWSER_AI_CHAT_BROWSER_TOOL_PROVIDER_H_
 #define BRAVE_BROWSER_AI_CHAT_BROWSER_TOOL_PROVIDER_H_
 
+#include <memory>
 #include <vector>
 
 #include "base/memory/weak_ptr.h"
 #include "brave/components/ai_chat/core/browser/tools/tool.h"
 #include "brave/components/ai_chat/core/browser/tools/tool_provider.h"
 
+namespace content {
+class BrowserContext;
+}
+
 namespace ai_chat {
+
+class CodeExecutionTool;
 
 // Implementation of ToolProvider that provides browser-specific
 // tools for conversations.
@@ -20,7 +27,7 @@ namespace ai_chat {
 // that the tools for a conversation perform actions on.
 class BrowserToolProvider : public ToolProvider {
  public:
-  BrowserToolProvider();
+  explicit BrowserToolProvider(content::BrowserContext* browser_context);
 
   ~BrowserToolProvider() override;
 
@@ -31,7 +38,9 @@ class BrowserToolProvider : public ToolProvider {
   std::vector<base::WeakPtr<Tool>> GetTools() override;
 
  private:
-  void CreateTools();
+  void CreateTools(content::BrowserContext* browser_context);
+
+  std::unique_ptr<CodeExecutionTool> code_execution_tool_;
 };
 
 }  // namespace ai_chat
--- a/browser/ai_chat/browser_tool_provider_factory.cc
+++ b/browser/ai_chat/browser_tool_provider_factory.cc
@@ -9,15 +9,18 @@
 
 #include "brave/browser/ai_chat/browser_tool_provider.h"
 #include "brave/components/ai_chat/core/browser/tools/tool_provider.h"
+#include "content/public/browser/browser_context.h"
 
 namespace ai_chat {
 
-BrowserToolProviderFactory::BrowserToolProviderFactory() = default;
+BrowserToolProviderFactory::BrowserToolProviderFactory(
+    content::BrowserContext* browser_context)
+    : browser_context_(browser_context) {}
 
 BrowserToolProviderFactory::~BrowserToolProviderFactory() = default;
 
 std::unique_ptr<ToolProvider> BrowserToolProviderFactory::CreateToolProvider() {
-  return std::make_unique<BrowserToolProvider>();
+  return std::make_unique<BrowserToolProvider>(browser_context_);
 }
 
 }  // namespace ai_chat
--- a/browser/ai_chat/browser_tool_provider_factory.h
+++ b/browser/ai_chat/browser_tool_provider_factory.h
@@ -8,15 +8,20 @@
 
 #include <memory>
 
+#include "base/memory/raw_ptr.h"
 #include "brave/components/ai_chat/core/browser/tools/tool_provider.h"
 #include "brave/components/ai_chat/core/browser/tools/tool_provider_factory.h"
+
+namespace content {
+class BrowserContext;
+}
 
 namespace ai_chat {
 
 // Factory for creating ToolProvider instances in the browser layer
 class BrowserToolProviderFactory : public ToolProviderFactory {
  public:
-  BrowserToolProviderFactory();
+  explicit BrowserToolProviderFactory(content::BrowserContext* browser_context);
   ~BrowserToolProviderFactory() override;
 
   BrowserToolProviderFactory(const BrowserToolProviderFactory&) = delete;
@@ -25,6 +30,9 @@ class BrowserToolProviderFactory : public ToolProviderFactory {
 
   // ToolProviderFactory implementation
   std::unique_ptr<ToolProvider> CreateToolProvider() override;
+
+ private:
+  raw_ptr<content::BrowserContext> browser_context_;
 };
 
 }  // namespace ai_chat
--- a/browser/ai_chat/code_execution_tool.h
+++ b/browser/ai_chat/code_execution_tool.h
@@ -1,0 +1,67 @@
+// Copyright (c) 2025 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#ifndef BRAVE_BROWSER_AI_CHAT_CODE_EXECUTION_TOOL_H_
+#define BRAVE_BROWSER_AI_CHAT_CODE_EXECUTION_TOOL_H_
+
+#include <memory>
+#include <optional>
+#include <string>
+#include <string_view>
+#include <vector>
+
+#include "base/memory/raw_ptr.h"
+#include "base/memory/weak_ptr.h"
+#include "base/values.h"
+#include "brave/components/ai_chat/core/browser/tools/tool.h"
+#include "content/public/browser/web_contents_observer.h"
+
+namespace content {
+class WebContents;
+class BrowserContext;
+}  // namespace content
+
+namespace ai_chat {
+
+class ConsoleMessageObserver;
+
+// Tool for executing JavaScript code and returning console.log output.
+// This tool is provided by the browser and allows AI assistants to run
+// JavaScript code in a sandboxed environment.
+class CodeExecutionTool : public Tool {
+ public:
+  explicit CodeExecutionTool(content::BrowserContext* browser_context);
+  ~CodeExecutionTool() override;
+
+  CodeExecutionTool(const CodeExecutionTool&) = delete;
+  CodeExecutionTool& operator=(const CodeExecutionTool&) = delete;
+
+  // Tool overrides
+  std::string_view Name() const override;
+  std::string_view Description() const override;
+  std::optional<base::Value::Dict> InputProperties() const override;
+  std::optional<std::vector<std::string>> RequiredProperties() const override;
+  bool RequiresUserInteractionBeforeHandling() const override;
+  bool SupportsConversation(bool is_temporary,
+                            bool has_untrusted_content) const override;
+  void UseTool(const std::string& input_json,
+               UseToolCallback callback) override;
+
+  // Public accessor for lambda access
+  content::WebContents* web_contents() const { return web_contents_.get(); }
+
+ private:
+  void OnJavaScriptExecutionComplete(UseToolCallback callback,
+                                     base::Value result);
+
+  raw_ptr<content::BrowserContext> browser_context_;
+  std::unique_ptr<content::WebContents> web_contents_;
+  std::unique_ptr<ConsoleMessageObserver> console_observer_;
+  base::WeakPtrFactory<CodeExecutionTool> weak_ptr_factory_{this};
+};
+
+}  // namespace ai_chat
+
+#endif  // BRAVE_BROWSER_AI_CHAT_CODE_EXECUTION_TOOL_H_
--- a/components/ai_chat/core/common/mojom/ai_chat.mojom
+++ b/components/ai_chat/core/common/mojom/ai_chat.mojom
@@ -15,6 +15,7 @@ import "url/mojom/url.mojom";
 // custom user actions and UI representations can be created.
 const string kUserChoiceToolName = "user_choice_tool";
 const string kMemoryStorageToolName = "memory_storage_tool";
+const string kCodeExecutionToolName = "code_execution_tool";
 
 // Maximum length for memory and customization records.
 const int32 kMaxMemoryRecordLength = 512;
--- a/components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/code_execution_tool_event.module.scss
+++ b/components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/code_execution_tool_event.module.scss
@@ -1,0 +1,36 @@
+/* Copyright (c) 2025 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+.codeExecutionTool {
+  display: flex;
+  flex-direction: column;
+  gap: 8px;
+  width: 100%;
+}
+
+.codeExecutionContent {
+  display: flex;
+  flex-direction: column;
+  gap: 16px;
+  margin-top: 8px;
+}
+
+.codeSection,
+.outputSection {
+  display: flex;
+  flex-direction: column;
+  gap: 8px;
+}
+
+.sectionLabel {
+  font-size: 14px;
+  font-weight: 600;
+  color: var(--leo-color-text-primary);
+}
+
+.outputContent {
+  border-left: 3px solid var(--leo-color-systemfeedback-success-icon);
+  padding-left: 12px;
+}
--- a/components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/code_execution_tool_event.tsx
+++ b/components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/code_execution_tool_event.tsx
@@ -1,0 +1,78 @@
+// Copyright (c) 2025 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import * as React from 'react'
+import Icon from '@brave/leo/react/icon'
+import ProgressRing from '@brave/leo/react/progressRing'
+import CodeBlock from '../code_block'
+import { ToolComponent } from './tool_event'
+import styles from './code_execution_tool_event.module.scss'
+
+const CodeExecutionToolEvent: ToolComponent = ({
+  toolUseEvent,
+  toolInput,
+  content,
+  children,
+}) => {
+  // Extract the JavaScript code from the tool input
+  const jsCode = React.useMemo(() => {
+    if (!toolInput?.program) return ''
+    return typeof toolInput.program === 'string' ? toolInput.program : ''
+  }, [toolInput])
+
+  // Extract the output from the tool use event
+  const output = React.useMemo(() => {
+    if (!toolUseEvent.output?.[0]?.textContentBlock?.text) return ''
+    return toolUseEvent.output[0].textContentBlock.text
+  }, [toolUseEvent.output])
+
+  const isComplete = !!toolUseEvent.output
+
+  // Custom content for the code execution tool
+  const customContent = {
+    ...content,
+    toolText: (
+      <div className={styles.codeExecutionTool}>
+        <span>Code Execution</span>
+        {jsCode && (
+          <div className={styles.codeExecutionContent}>
+            <div className={styles.codeSection}>
+              <div className={styles.sectionLabel}>JavaScript Code:</div>
+              <CodeBlock.Block
+                code={jsCode}
+                lang='javascript'
+              />
+            </div>
+            {isComplete && output && (
+              <div className={styles.outputSection}>
+                <div className={styles.sectionLabel}>Output:</div>
+                <div className={styles.outputContent}>
+                  <CodeBlock.Block
+                    code={output}
+                    lang='text'
+                  />
+                </div>
+              </div>
+            )}
+          </div>
+        )}
+      </div>
+    ),
+    statusIcon: (
+      <span data-testid='code-execution-completed-icon'>
+        <Icon name='code' />
+      </span>
+    ),
+    progressIcon: (
+      <span data-testid='code-execution-progress-icon'>
+        <ProgressRing />
+      </span>
+    ),
+  }
+
+  return children(customContent)
+}
+
+export default CodeExecutionToolEvent
--- a/components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/tool_event.tsx
+++ b/components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/tool_event.tsx
@@ -10,6 +10,7 @@ import Tooltip from '@brave/leo/react/tooltip'
 import classnames from '$web-common/classnames'
 import * as Mojom from '../../../common/mojom'
 import ToolEventContentUserChoice from './tool_event_content_user_choice'
+import CodeExecutionToolEvent from './code_execution_tool_event'
 import styles from './tool_event.module.scss'
 
 interface Props {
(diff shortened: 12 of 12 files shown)


DJAndries:
Will add tests tomorrow.

stoletheminerals:
@DJAndries can we disable all networking in the renderer where we execute the code?

DJAndries:
> @DJAndries can we disable all networking in the renderer where we execute the code?

The CSP will handle this. The browser test will ensure that this works

thypon:
@DJAndries isn't better to use a plain WebWorker, instead of a full WebContents, here?

It would get OTB sandboxing, and it should be easier to extra sandbox it.

components/ai_chat/core/browser/code_sandbox_script_storage.h
```diff
@@ -0,0 +1,47 @@
...
+#include "base/memory/scoped_refptr.h"
+
+namespace base {
+template <typename T>
+class NoDestructor;
+}
+
+namespace ai_chat {
+
+// Singleton class to manage script storage for code execution requests
```
bridiver: Why is this a singleton?

components/ai_chat/core/browser/code_sandbox_script_storage.cc
```diff
@@ -0,0 +1,40 @@
...
+CodeSandboxScriptStorage::~CodeSandboxScriptStorage() = default;
+
+std::string CodeSandboxScriptStorage::StoreScript(std::string script) {
+  auto request_id = base::UnguessableToken::Create().ToString();
+  scripts_[request_id] =
+      base::MakeRefCounted<base::RefCountedString>(std::move(script));
+  return request_id;
+}
+
+scoped_refptr<base::RefCountedString> CodeSandboxScriptStorage::ConsumeScript(
```
bridiver: I don't understand why this is called `ConsumeScript` or why you're using scoped_refptr. This does not have shared ownership so scoped_refptr is not appropriate here https://www.chromium.org/developers/smart-pointer-guidelines/

browser/ui/webui/ai_chat/code_sandbox_ui.cc
```diff
@@ -0,0 +1,101 @@
...
+CodeSandboxUIConfig::~CodeSandboxUIConfig() = default;
+
+CodeSandboxUI::CodeSandboxUI(content::WebUI* web_ui)
+    : ui::UntrustedWebUIController(web_ui) {
+  auto* source = content::WebUIDataSource::CreateAndAdd(
+      web_ui->GetWebContents()->GetBrowserContext(), kAIChatCodeSandboxUIURL);
+
+  source->SetDefaultResource(IDR_AI_CHAT_CODE_SANDBOX_UI_HTML);
+
+  source->SetRequestFilter(
```
bridiver: this needs some explanation, why are we handling it this way instead of exposing a mojo api for the scripts?

bridiver: it also seems strange to handle responses through the console api where the mojo interface could return a response without this limitation `The code must be synchronous and not use any asynchronous functions`

bridiver: I got some info from @thypon related to this and there is already a well tested model for this which is to have a webui with an untrusted webui iframe and communicate via postMessage

chromium_src/chrome/browser/ui/webui/chrome_untrusted_web_ui_configs.cc
```diff
@@ -69,5 +70,7 @@ void RegisterChromeUntrustedWebUIConfigs() {
   if (ai_chat::features::IsAIChatEnabled()) {
     content::WebUIConfigMap::GetInstance().AddUntrustedWebUIConfig(
         std::make_unique<AIChatUntrustedConversationUIConfig>());
+    content::WebUIConfigMap::GetInstance().AddUntrustedWebUIConfig(
+        std::make_unique<ai_chat::CodeSandboxUIConfig>());
   }
```
fallaciousreasoning, line 87: Hey I'm not sure why this is here instead of with the other interfaces in `brave/browser/brave_content_browser_client.cc`.

Either this one or the other one is in the wrong place (I'd assume that its better to be in `brave_content_browser_client` as its not in a ChromiumSrc override).

cc @bridiver

components/ai_chat/core/browser/code_sandbox_script_storage.h
```diff
@@ -0,0 +1,47 @@
...
+  scoped_refptr<base::RefCountedString> ConsumeScript(
+      std::string_view request_id);
+
+ private:
+  friend class base::NoDestructor<CodeSandboxScriptStorage>;
+
+  CodeSandboxScriptStorage();
+  ~CodeSandboxScriptStorage();
+
+  base::flat_map<std::string, scoped_refptr<base::RefCountedString>> scripts_;
```
fallaciousreasoning: this seems like it could just be storing the strings rather than using a smart pointer at all - we pass in a copy of the string, and we remove it from this map when we consume it:
```suggestion
  base::flat_map<std::string, std::string> scripts_;
```

DJAndries:
> @DJAndries isn't better to use a plain WebWorker, instead of a full WebContents, here?
> 
> It would get OTB sandboxing, and it should be easier to extra sandbox it.

WebContents seems to provide the highest degree of isolation compared to workers. With a worker, you need to maintain a message channel with the host which introduces a small risk of hijacking. With WebContents, we just observe the console logs of the renderer and kill it once the page has finished loading, without any need for a message channel. Using an OTR profile when spawning the renderer also ensures that any stored data gets destroyed after execution (i.e. IndexedDB), which provides data isolation between executions.

browser/ui/webui/ai_chat/code_sandbox_ui.cc
DJAndries: > this needs some explanation, why are we handling it this way instead of exposing a mojo api for the scripts?

The upstream doc states that webui bindings should not be used with untrusted webui: https://source.chromium.org/chromium/chromium/src/+/main:ui/webui/untrusted_web_ui_controller.h;l=17
This webui implementation takes inspiration from the print preview webui, which uses the same simple mechanism to transfer the pdf to the page: https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/ui/webui/print_preview/print_preview_ui_untrusted.cc;bpv=0;bpt=1

> it also seems strange to handle responses through the console api where the mojo interface could return a response without this limitation "The code must be synchronous and not use any asynchronous functions"

We could probably find an easy workaround for this limitation, but I can't think of any reason why we would want an LLM to execute async code anyway. The sandbox isn't allowed to make network requests, or anything else that would require async I/O.

> I got some info from @thypon related to this and there is already a well tested model for this which is to have a webui with an untrusted webui iframe and communicate via postMessage

Please see my comment below: https://github.com/brave/brave-core/pull/31252#issuecomment-3387103070

components/ai_chat/core/browser/code_sandbox_script_storage.h
DJAndries: Moved ownership to `AIChatService`

components/ai_chat/core/browser/code_sandbox_script_storage.cc
DJAndries: The webui request filter callback accepts `base::RefCountedString` so I thought I'd just construct it here, but I realize that it makes things unclear. Changed to store & return `std::string`.

chromium_src/chrome/browser/ui/webui/chrome_untrusted_web_ui_configs.cc
DJAndries: This is an untrusted webui which has no bindings/mojo. Untrusted webuis are registered via `AddUntrustedWebUIConfig`, like the `AIChatUntrustedConversationUIConfig` above.

fallaciousreasoning: :facepalm: didn't notice this was for the config not a mojo binding

browser/ui/webui/ai_chat/code_sandbox_ui.cc
fallaciousreasoning: I don't think there's any rule against having an untrusted WebUI having mojom interfaces?

Upstream has multiple examples of providing an interface to an untrusted UI:
- ReadAnythingUntrustedUI
- LensSidePanelUntrustedUI
- DemoModeAppUntrustedUI
- HelpAppUntrustedUI

bridiver: I think you're misunderstanding @DJAndries , webui bindings are a specific binding for webui https://source.chromium.org/chromium/chromium/src/+/main:content/common/web_ui.mojom?q=web_ui.mojom&ss=chromium

DJAndries: Gotcha, I wasn't aware of the distinction, thanks for noting that. I wonder why the `UntrustedWebUIController` comments advises against webui bindings, but not mojo...

petemill: >  well tested model for this which is to have a webui with an untrusted webui iframe and communicate via postMessage

@bridiver @thypon what is this well tested model? We avoid postMessage between trusted and untrusted because the chromium docs state this. We instead should use a mojo interface between child and parent frames across trusted/unstrusted boundary. But I haven't started reviewing this PR yet. I assume the whole thing can be untrusted.

petemill: This actually (loading the script via a `script.js` request) does seem like quite a good way to avoid having to have _any_ mojo bindings or postMessage handlers in the frame at all, especially because the purpose of this WebContents is to execute untrusted script directly. Any time we're using a mojom pattern is because we want to protect against unknown weaknesses that might lead to untrusted code execution, but this scenario definitely does perform untrusted code execution, so it seems better to not expose any extra APIs that can be directly called. If we had the script transferred over mojom, we'd just be `eval`-ing it. Not supporting async JS might seem like a limitation but are there actually any async JS APIs we want to provide for this tool which exists mainly to perform calculations?

DJAndries: > Not supporting async JS might seem like a limitation but are there actually any async JS APIs we want to provide for this tool which exists mainly to perform calculations?

Not that I can think of.

bridiver: I wonder if there are any potential race conditions here where console.log messages may actually come in after the page onload event. Generally the messages should be ordered so I don't think so, but one advantage of a mojo api is that we are waiting for a well defined response. I'm going to discuss this with @thypon today

bridiver: looking at this again, why aren't we just using components/script_injector to run the JS and get the response?

components/ai_chat/core/browser/code_sandbox_script_storage.h
```diff
@@ -0,0 +1,36 @@
...
+#include <optional>
+#include <string>
+#include <string_view>
+
+#include "base/containers/flat_map.h"
+
+namespace ai_chat {
+
+// Class to manage script storage for code execution requests
+class CodeSandboxScriptStorage {
```
bridiver: why is this class necessary? It seems like a memory leak that could grow without bounds.

browser/ui/webui/ai_chat/code_sandbox_ui.cc
DJAndries: Did you mean `components/on_load_script_injector`? I'm unsure what advantage this component would provide compared to loading the script via webui data. Interested in learning more.

components/ai_chat/core/browser/code_sandbox_script_storage.h
DJAndries: I created this to avoid increasing the code size of the `AIChatService`. We could enforce a maximum size to the map if that would address any memory-related concerns.

browser/ai_chat/code_execution_tool_browsertest.cc
```diff
@@ -0,0 +1,132 @@
...
+    console.log('Starting computation...');
+    const result = fibonacci(45);
+    console.log('Result: ' + result);
+  )";
+
+  std::string output;
+  ExecuteCode(script, &output);
+  EXPECT_EQ(output, "Error: Time limit exceeded");
+}
+
```
thypon, line 154: I advise to add a couple of tests here to check if the security boundaries are correctly set:

1. check if the CSP is setup properly maybe with `console.log(document.contentSecurityPolicy);` or something similar
2. check the window.location for understanding the execution context. It should say something like `brave://` or `chrome://` or `chrome-untrusted://`

browser/ai_chat/code_execution_tool.cc
```diff
@@ -0,0 +1,219 @@
...
+#include "content/public/browser/web_contents.h"
+#include "content/public/browser/web_contents_observer.h"
+#include "ui/base/page_transition_types.h"
+#include "url/gurl.h"
+
+namespace ai_chat {
+
+namespace {
+
+constexpr base::TimeDelta kExecutionTimeLimit = base::Seconds(10);
```
thypon, line 36: can we test that the execution time limit works properly in the code execution tool, with a test?

If we don't we risk to leak webcontents forever.

browser/ui/webui/ai_chat/code_sandbox_ui.cc
thypon: I'm not super in love with the architecture, but I suspect the LLM would be very very confused if we don't keep as simple as it is right now. I suspect calling a specific mojom interface would be quite a stretch.

thypon (changes requested):

components/ai_chat/core/browser/code_sandbox_script_storage.h
bridiver: But I'm asking why we need to keep these here at all. In its current form this class is also just a wrapper around a map and that doesn't provide any value over just having a map.

browser/ai_chat/code_execution_tool.cc
DJAndries: We have one here https://github.com/brave/brave-core/blob/625c12ff3b9b2388846da3aa5760b77a3886c935/browser/ai_chat/code_execution_tool_browsertest.cc#L114

components/ai_chat/core/browser/code_sandbox_script_storage.h
DJAndries: Removed the class and moved the single member/methods to `AIChatService`

browser/ai_chat/code_execution_tool_browsertest.cc
DJAndries: > check if the CSP is setup properly maybe with console.log(document.contentSecurityPolicy); or something similar

I'm unsure if we're able to inspect the CSP via JS (at least, `document.contentSecurityPolicy` doesn't work for me), but we do have a test to ensure XHR network requests are blocked due to CSP.

> check the window.location for understanding the execution context

I omitted the `window` and `location` objects from the execution context and added a test to ensure they are inaccessible.

browser/ui/webui/ai_chat/code_sandbox_ui.cc
```diff
@@ -0,0 +1,113 @@
...
+      AIChatServiceFactory::GetForBrowserContext(original_profile);
+  if (ai_chat_service) {
+    auto script = ai_chat_service->ConsumeCodeExecutionToolScript(request_id);
+    if (script) {
+      script_content = std::move(*script);
+    }
+  }
+
+  auto script =
+      base::MakeRefCounted<base::RefCountedString>(std::move(script_content));
```
bridiver: why are you making this refcounted?

browser/ui/webui/ai_chat/code_sandbox_ui.cc
bridiver: `calling a specific mojom interface would be quite a stretch` - there is no mojom interface for the llm to call here, we're just executing javascript from the browser process and returning the result. The webui/script/llm doesn't have to know anything about it. Script injector is similar to calling `RenderFrameHost::ExecuteJavaScript` (which we could also use)

goodov: Creating an OTR profile for **each** exec request feels heavy performance and battery-wise. 

What are your thoughts on:
1. use current profile or a new OTR, but same for **all** chat sessions
2. load `chrome-untrusted://...` with **sandbox CSP**. This will create an opaque origin which will prevent storage sharing and make it effectively ephemeral (no need for OTR profile). 
3. bind the same loaded page to the chat session to act as REPL. Enable/disable it via "repl" tool parameter.
4. use `//brave/components/script_injector` to execute JS (I wasn't aware of that, thanks @bridiver for mentioning).

DJAndries: >load chrome-untrusted://... with sandbox CSP. This will create an opaque origin which will prevent storage sharing and make it effectively ephemeral (no need for OTR profile).

Will give this a try.

>bind the same loaded page to the chat session to act as REPL. Enable/disable it via "repl" tool parameter.

This would be a neat mode to have, but I'm hoping that we can add that as a follow-up to keep this PR simple.

> use //brave/components/script_injector to execute JS (I wasn't aware of that, thanks @bridiver for mentioning).

We explored this option. Here's an exploratory draft: https://github.com/brave/brave-core/pull/32347

The latest commit uses `ExecuteJavascriptInIsolatedWorld`. I had a tough time getting the script injector to work on the webui (it worked fine on HTTPS and about:blank pages however). It did not return a result whatsoever.  `ExecuteJavascriptInIsolatedWorld` worked fine, but the execution did not respect CSP; I was able to perform fetch/XHR calls inside the sandbox.

thypon:
>Memory Leak Risk: The code_execution_tool_scripts_ map in AIChatService stores scripts indefinitely if they're never consumed. Consider adding TTL or size limits.

Is this correct @DJAndries ? Scripts might contain user data embedded. We should clean them up after sometime

DJAndries:
> Is this correct @DJAndries ? Scripts might contain user data embedded. We should clean them up after sometime

In theory, yes, but practically no. The renderer immediately consumes the script at startup, so the risk is very low. Nevertheless, I can add a size limit to the container.

browser/ui/webui/ai_chat/code_sandbox_ui.cc
DJAndries: It's what the callback expects

browser/ui/webui/ai_chat/code_sandbox_ui.cc
DJAndries: Using a CSP sandbox directive did not work, so I changed it to use a sandboxed iframe

goodov: Thanks for looking into this alternative, it looks a bit better. Now we don't spawn profiles for each request.

Wonder if we can replace console parser with something more sensible.

browser/ai_chat/code_execution_tool.cc
```diff
@@ -0,0 +1,196 @@
...
+        CreateContentBlocksForText("Error: Missing or empty 'script' field"));
+    return;
+  }
+
+  auto wrapped_js = base::StrCat({"try {", *script,
+                                  "} catch (error) {console.error('Error:', "
+                                  "error.toString()); }"});
+
+  auto request_id =
+      ai_chat_service_->StoreCodeExecutionToolScript(std::move(wrapped_js));
```
petemill: Perhaps we can avoid making this very purpose-specific temporary store on the `AIChatService` because the script is already stored in the conversation history, how about we have the CodeSandBoxUI lookup via Conversation uuid, ConversationTurn uuid and ToolUseEvent id. With those 3, it should be able to do a quick lookup via `AIChatService::GetConversation(uuid)` (it can use the sync version of that function since the conversation is needs to already be loaded anyway). We can also construct the wrapped JS with error handling in the CodeSandboxUI.
Having said that, there _would_ be a more justifiable reason to have a separate store for the script if we are moving away from console log reading and to a mojom interface where we might have to store a callback somewhere if we can't provide the callback directly from here to the WebContents. Although we might be able to call `GetRemoteInterfaces()->GetInterface` on the render frame if we go down that path anyway...

browser/ui/webui/ai_chat/code_sandbox_ui.cc
DJAndries: >Wonder if we can replace console parser with something more sensible.

When considering the script injection approach, we explored returning a human readable string as a result. Without script injection, I think this might be the easiest method.

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/tool_event_code_execution.tsx
```diff
@@ -0,0 +1,63 @@
...
+    return props.toolUseEvent.output[0].textContentBlock.text
+  }, [props.toolUseEvent.output])
+
+  const content: ToolUseContent = {
+    toolLabel: props.content.toolLabel,
+    expandedContent: (
+      <div className={styles.codeExecutionTool}>
+        {jsCode && (
+          <div className={styles.codeExecutionContent}>
+            <div className={styles.codeSection}>
```
petemill: `expandedContent` should be `null` if `!jsCode`

components/ai_chat/core/browser/code_sandbox_script_storage.h
petemill: I suggested an alternative lookup for the scripts where they are already stored to avoid this complication

browser/ai_chat/code_execution_tool.cc
DJAndries: > Perhaps we can avoid making this very purpose-specific temporary store on the `AIChatService` because the script is already stored in the conversation history, how about we have the CodeSandBoxUI lookup via Conversation uuid, ConversationTurn uuid and ToolUseEvent id.

Good point. If others are satisfied with the current approach of having the webui config provide the script, then I'll be happy to update accordingly.

browser/ui/webui/ai_chat/code_sandbox_ui.cc
goodov: > The latest commit uses `ExecuteJavascriptInIsolatedWorld`. I had a tough time getting the script injector to work on the webui (it worked fine on HTTPS and about:blank pages however). It did not return a result whatsoever. `ExecuteJavascriptInIsolatedWorld` worked fine, but the execution did not respect CSP; I was able to perform fetch/XHR calls inside the sandbox.

1. you need to wait for the page/frame to load before using script injector 
2. you need to launch JS in the main world to adhere to CSP.
3. sandbox in the main frame in webui requires few patches to work correctly (that was unexpected to me).

please take a look at the working impl here: https://github.com/brave/brave-core/commit/560696d1dc508cadff68fd0ab9f48ace94554708

browser/ai_chat/code_execution_tool.cc
DJAndries: Removed the temporary script storage

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/tool_event_code_execution.tsx
DJAndries: e0b4bfedd9604e06fed5f16fa8526f3dceec3e3a

components/ai_chat/core/browser/code_sandbox_script_storage.h
DJAndries: Removed script storage entirely. 85b7b83b11d689c54c4426f527b403776dcafd46

browser/ui/webui/ai_chat/code_sandbox_ui.cc
DJAndries: Good catch. Pushed 85b7b83b11d689c54c4426f527b403776dcafd46 to use script injector.

browser/ai_chat/code_execution_tool.cc
```diff
@@ -0,0 +1,207 @@
...
+
+  if (!script || script->empty()) {
+    std::move(callback).Run(
+        CreateContentBlocksForText("Error: Missing or empty 'script' field"));
+    return;
+  }
+
+  auto wrapped_js =
+      base::StrCat({"(async function() { try {", *script,
+                    "} catch (error) { return error.toString(); } })()"});
```
goodov: do we need to wrap this with a try/catch block? what will happen if there's a syntax error? will the error be passed into a callback in this case?

This is something that should work in an expected way. Please add required tests for all such scenarios and adjust/fix logic if you need to (maybe you even have to change ScriptInjector).

browser/ai_chat/code_execution_tool.cc
```diff
@@ -0,0 +1,207 @@
...
+    bool is_temporary,
+    bool has_untrusted_content,
+    mojom::ConversationCapability conversation_capability) const {
+  // Support all conversation types for now
+  return true;
+}
+
+void CodeExecutionTool::SetExecutionTimeLimitForTesting(
+    base::TimeDelta time_limit) {
+  execution_time_limit_for_testing_ = time_limit;
```
goodov: `execution_time_limit_for_testing_` should just be `execution_time_limit_` that you initialize in the constructor.

in the test you either call `SetExecutionTimeLimitForTesting` or initialize the timeout via constructor.

this part shouldn't exist in this case:
```
  auto time_limit =
      execution_time_limit_for_testing_.value_or(kExecutionTimeLimit);
```

browser/ai_chat/code_execution_tool.cc
```diff
@@ -0,0 +1,207 @@
...
+  }
+
+  auto wrapped_js =
+      base::StrCat({"(async function() { try {", *script,
+                    "} catch (error) { return error.toString(); } })()"});
+
+  auto* otr_profile =
+      profile_->GetPrimaryOTRProfile(true /* create_if_needed */);
+  content::WebContents::CreateParams create_params(otr_profile);
+  auto web_contents = content::WebContents::Create(create_params);
```
goodov: the fact that this creates WebContents for each request is killing me. This needs to be reworked in some way to make it bound to a chat session.

It's fine for a follow up as now it seems Tools just do not have this context knowledge to make it work like that.

browser/ai_chat/code_execution_tool.cc
```diff
@@ -0,0 +1,207 @@
...
+      execution_time_limit_for_testing_.value_or(kExecutionTimeLimit);
+  request_it->timeout_timer.Start(
+      FROM_HERE, time_limit,
+      base::BindOnce(&CodeExecutionTool::ResolveRequest,
+                     weak_ptr_factory_.GetWeakPtr(), request_it,
+                     "Error: Time limit exceeded"));
+}
+
+void CodeExecutionTool::OnPageLoadComplete(
+    std::list<CodeExecutionRequest>::iterator request_it,
```
goodov: You already have `CodeExecutionRequest`. Just add methods there to handle all callbacks and timers, no need to have data in `CodeExecutionRequest` and handlers in `CodeExecutionTool`.

You also shouldn't have `CodeSandboxWebContentsObserver` as a separate class.

Just try to unify everything related to a single request in one class please. You can make it have a static method like `Create` that will prepare everything and pass into a constructor.

In `CodeExecutionTool` you can have a single member with currently running requests, and a single callback that will remove those `CodeExecutionRequest` on completion.

In this setup on `CodeExecutionTool` destructor all currently active requests (with bound webcontents/timers) should be destroyed automatically.

browser/ai_chat/code_execution_tool.cc
```diff
@@ -0,0 +1,207 @@
...
+  auto wrapped_js_utf16 = base::UTF8ToUTF16(request_it->wrapped_js);
+  request_it->injector->RequestAsyncExecuteScript(
+      content::ISOLATED_WORLD_ID_GLOBAL, wrapped_js_utf16,
+      blink::mojom::UserActivationOption::kActivate,
+      blink::mojom::PromiseResultOption::kAwait,
+      base::BindOnce(&CodeExecutionTool::HandleResult,
+                     weak_ptr_factory_.GetWeakPtr(), request_it));
+}
+
+void CodeExecutionTool::HandleResult(
```
goodov: You already have `CodeExecutionRequest`. Just add methods there to handle all callbacks and timers, no need to have data in `CodeExecutionRequest` and handlers in `CodeExecutionTool`.

You also shouldn't have `CodeSandboxWebContentsObserver` as a separate class.

Just try to unify everything related to a single request in one class please. You can make it have a static method like `Create` that will prepare everything and pass into a constructor.

In `CodeExecutionTool` you can have a single member with currently running requests, and a single callback that will remove those `CodeExecutionRequest` on completion.

In this setup on `CodeExecutionTool` destructor all currently active requests (with bound webcontents/timers) should be destroyed automatically.

browser/ai_chat/code_execution_tool.cc
DJAndries: Added additional tests: 5debd6c26465f7fa8ac33641cf419996cae958ef

browser/ai_chat/code_execution_tool.cc
DJAndries: 5debd6c26465f7fa8ac33641cf419996cae958ef

browser/ai_chat/code_execution_tool.cc
DJAndries: Invocations are expected to be infrequent to start, happy to follow up if that expectation changes.

browser/ai_chat/code_execution_tool.cc
DJAndries: 5debd6c26465f7fa8ac33641cf419996cae958ef

goodov (approved):

goodov:
thanks for fixing most critical things. This is lgtm for now, but I believe we can do better. Waiting for nice follow ups!

browser/ai_chat/code_execution_tool.cc
```diff
@@ -0,0 +1,183 @@
...
+    content::RenderFrameHost* render_frame_host,
+    const GURL& validated_url) {
+  if (!render_frame_host->GetParent()) {
+    return;
+  }
+
+  render_frame_host->GetRemoteAssociatedInterfaces()->GetInterface(&injector_);
+
+  auto wrapped_js_utf16 = base::UTF8ToUTF16(wrapped_js_);
+  injector_->RequestAsyncExecuteScript(
```
goodov, line 86: this should be a single operation for a request. You need to move/clear `wrapped_js_` after you send it for execution and don't attempt to exec script here again.

you cannot just blindly assume that `DidFinishLoad` with a sandbox frame will be called once. There are many ways when it can be called, so you need to be a bit defensive here and only run your script once.

browser/ai_chat/code_execution_tool.cc
```diff
@@ -0,0 +1,183 @@
...
+      content::ISOLATED_WORLD_ID_GLOBAL, wrapped_js_utf16,
+      blink::mojom::UserActivationOption::kActivate,
+      blink::mojom::PromiseResultOption::kAwait,
+      base::BindOnce(&CodeExecutionRequest::HandleResult,
+                     weak_ptr_factory_.GetWeakPtr()));
+}
+
+void CodeExecutionTool::CodeExecutionRequest::HandleResult(base::Value result) {
+  std::string output;
+  if (result.is_string()) {
```
goodov, line 96: this seems like an unnecessary restriction, but can be adjusted later.

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/tool_event_code_execution.tsx
```diff
@@ -0,0 +1,61 @@
...
+import styles from './tool_event_code_execution.module.scss'
+import '../../../common/strings'
+
+const ToolEventCodeExecution: ToolComponent = (props) => {
+  const jsCode = React.useMemo(() => {
+    if (!props.toolInput?.script) return ''
+    return typeof props.toolInput.script === 'string'
+      ? props.toolInput.script
+      : ''
+  }, [props.toolInput])
```
petemill: Seems a little overkill to `useMemo` this when it's not doing any parsing

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/tool_event_code_execution.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  const output = React.useMemo(() => {
+    if (!props.toolUseEvent.output?.[0]?.textContentBlock?.text) return ''
+    return props.toolUseEvent.output[0].textContentBlock.text
+  }, [props.toolUseEvent.output])
```
petemill: ditto, re: useMemo

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/tool_event_code_execution.tsx
```diff
@@ -0,0 +1,61 @@
...
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import * as React from 'react'
+import { getLocale } from '$web-common/locale'
+import CodeBlock from '../code_block'
+import type { ToolComponent, ToolUseContent } from './tool_event'
+import styles from './tool_event_code_execution.module.scss'
+import '../../../common/strings'
+
+const ToolEventCodeExecution: ToolComponent = (props) => {
```
petemill, line 13: We need a tool use event that uses this tool added to Storybook, in the `ToolUse` and/or `Assistant Task` and/or `Panel` sections. Ideally all kinds of states, which should simply need example `ToolUseEvent` items adding to the sample data.

components/ai_chat/resources/code_sandbox/code_sandbox.html
```diff
@@ -0,0 +1,11 @@
+<!DOCTYPE html>
```
petemill, line 1: I don't think this belongs in `/page` as that's the directory for the AI Chat conversation WebUI. We have `/page` and `/untrusted_conversation_frame` and `/custom_distiller_scripts` as "top level" containers. This could be `/code_sandbox`, for example.

components/ai_chat/core/browser/ai_chat_service.h
```diff
@@ -11,11 +11,13 @@
 #include <functional>
 #include <map>
 #include <memory>
+#include <optional>
 #include <string>
 #include <string_view>
 #include <utility>
 #include <vector>
 
+#include "base/containers/flat_map.h"
```
petemill: Not needed anymore

components/ai_chat/core/browser/ai_chat_service.cc
```diff
@@ -31,6 +31,7 @@
 #include "base/task/thread_pool.h"
 #include "base/time/time.h"
 #include "base/types/strong_alias.h"
+#include "base/unguessable_token.h"
```
petemill: not needed anymore

browser/ai_chat/code_execution_tool.cc
```diff
@@ -0,0 +1,183 @@
...
+std::string_view CodeExecutionTool::Name() const {
+  return mojom::kCodeExecutionToolName;
+}
+
+std::string_view CodeExecutionTool::Description() const {
+  return "Execute JavaScript code and return a human-readable formatted "
+         "string as a result. "
+         "Do not use console.log statements or similar statements. Always "
+         "return a string as a result."
+         "The code will be executed in a sandboxed environment.";
```
petemill, line 134: I've seen even Claude Haiku still fail to end the code snippet with `return x;` and just print `x;` as the last statement. Perhaps providing a short example param value would help here.

browser/ai_chat/code_execution_tool.cc
```diff
@@ -0,0 +1,183 @@
...
+}
+
+std::optional<base::Value::Dict> CodeExecutionTool::InputProperties() const {
+  return CreateInputProperties(
+      {{"script", StringProperty("The JavaScript code to execute")}});
+}
+
+std::optional<std::vector<std::string>> CodeExecutionTool::RequiredProperties()
+    const {
+  return std::vector<std::string>{"script"};
```
petemill: nit: `script` could be a locall-named constant

browser/ai_chat/code_execution_tool.cc
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  const std::string* script = input_dict->FindString("script");
+
+  if (!script || script->empty()) {
+    std::move(callback).Run(
+        CreateContentBlocksForText("Error: Missing or empty 'script' field"));
+    return;
+  }
+
```
petemill, line 183: Needs unit tests for bad input. See other tools for quick examples if you need.

petemill (changes requested):
As long as someone else is ok with the lifecycle of the `CodeExeuctionRequest` and the security of the sandbox, the Tool definition, execution and UI LGTM aside from the minor comments and additional test needed below:

browser/ui/webui/ai_chat/code_sandbox_ui.cc
```diff
@@ -0,0 +1,80 @@
...
+
+  webui::SetupWebUIDataSource(source, kAiChatUiGenerated,
+                              IDR_AI_CHAT_CODE_SANDBOX_HTML);
+
+  source->OverrideContentSecurityPolicy(
+      network::mojom::CSPDirectiveName::DefaultSrc, "default-src 'none';");
+
+  source->OverrideContentSecurityPolicy(
+      network::mojom::CSPDirectiveName::ScriptSrc,
+      "script-src 'unsafe-inline';");
```
thypon: Is it worth using an hash-based CSP, based on the content of the script we know we want to run, itself?

browser/ai_chat/code_execution_tool.cc
```diff
@@ -0,0 +1,183 @@
...
+    const GURL& validated_url) {
+  if (!render_frame_host->GetParent()) {
+    return;
+  }
+
+  render_frame_host->GetRemoteAssociatedInterfaces()->GetInterface(&injector_);
+
+  auto wrapped_js_utf16 = base::UTF8ToUTF16(wrapped_js_);
+  injector_->RequestAsyncExecuteScript(
+      content::ISOLATED_WORLD_ID_GLOBAL, wrapped_js_utf16,
```
thypon, line 87: should we use a dedicated id here, instead of the GLOBAL one?

browser/ai_chat/code_execution_tool.cc
```diff
@@ -0,0 +1,183 @@
...
+
+}  // namespace
+
+CodeExecutionTool::CodeExecutionRequest::CodeExecutionRequest(
+    Profile* profile,
+    const std::string& script,
+    base::TimeDelta execution_time_limit)
+    : content::WebContentsObserver(nullptr), wrapped_js_(WrapScript(script)) {
+  auto* otr_profile =
+      profile->GetPrimaryOTRProfile(true /* create_if_needed */);
```
thypon: Should we create a dedicated OTR profile here, so that we don't interact with the primary one?

browser/ui/webui/ai_chat/code_sandbox_ui.cc
```diff
@@ -0,0 +1,80 @@
...
+
+  source->OverrideContentSecurityPolicy(
+      network::mojom::CSPDirectiveName::ConnectSrc, "connect-src 'none';");
```
thypon, line 61: Can/should we have a network throttle, instead of this single `connect-src none`?

browser/ui/webui/ai_chat/code_sandbox_ui.cc
```diff
@@ -0,0 +1,80 @@
...
+      network::mojom::CSPDirectiveName::FrameSrc, "frame-src 'none';");
+  source->OverrideContentSecurityPolicy(
+      network::mojom::CSPDirectiveName::FrameAncestors,
+      "frame-ancestors 'self';");
+  source->OverrideContentSecurityPolicy(
+      network::mojom::CSPDirectiveName::WorkerSrc, "worker-src 'none';");
+  source->OverrideContentSecurityPolicy(
+      network::mojom::CSPDirectiveName::FormAction, "form-action 'none';");
+  source->OverrideContentSecurityPolicy(
+      network::mojom::CSPDirectiveName::BaseURI, "base-uri 'none';");
```
thypon, line 74: NIT: unneeded 

```  source->OverrideContentSecurityPolicy(
      network::mojom::CSPDirectiveName::PermissionsPolicy,
      "geolocation=(), camera=(), microphone=(), usb=()");
```

thypon (changes requested):

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/tool_event_code_execution.tsx
DJAndries: cea5d7d31533e5e07e39d13e878fd2a670a654bb

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/tool_event_code_execution.tsx
DJAndries: cea5d7d31533e5e07e39d13e878fd2a670a654bb

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/tool_event_code_execution.tsx
DJAndries: cea5d7d31533e5e07e39d13e878fd2a670a654bb

components/ai_chat/resources/code_sandbox/code_sandbox.html
DJAndries: cea5d7d31533e5e07e39d13e878fd2a670a654bb

components/ai_chat/core/browser/ai_chat_service.h
DJAndries: cea5d7d31533e5e07e39d13e878fd2a670a654bb

components/ai_chat/core/browser/ai_chat_service.cc
DJAndries: cea5d7d31533e5e07e39d13e878fd2a670a654bb

browser/ai_chat/code_execution_tool.cc
DJAndries: cea5d7d31533e5e07e39d13e878fd2a670a654bb

browser/ai_chat/code_execution_tool.cc
DJAndries: cea5d7d31533e5e07e39d13e878fd2a670a654bb

browser/ai_chat/code_execution_tool.cc
DJAndries: cea5d7d31533e5e07e39d13e878fd2a670a654bb

browser/ui/webui/ai_chat/code_sandbox_ui.cc
DJAndries: I forgot to update this to `none`. We no longer need this since we're now using script injection rather than supplying the script via the webui.

cea5d7d31533e5e07e39d13e878fd2a670a654bb

browser/ai_chat/code_execution_tool.cc
DJAndries: We need to use the global world id in order to inherit the CSP defined in the webui config

browser/ai_chat/code_execution_tool.cc
DJAndries: It shouldn't be required. Now that we are using a sandboxed iframe with an opaque origin, there is no risk of data being persisted with the profile, so it doesn't matter which profile we use.

browser/ui/webui/ai_chat/code_sandbox_ui.cc
DJAndries: the connect-src CSP will ensure no fetch requests are made, and the frame-src CSP will ensure that the script cannot navigate away from the page. A throttle would be redundant. We also have tests to ensure that fetches and external navigations are prohibited.

browser/ui/webui/ai_chat/code_sandbox_ui.cc
DJAndries: The `PermissionsPolicy` name is not available

browser/ai_chat/code_execution_tool.cc
DJAndries: cea5d7d31533e5e07e39d13e878fd2a670a654bb

DJAndries:
@thypon Added a variations feature `AIChatCodeExecutionTool`. Will enable this in nightly for the interim: https://github.com/brave/brave-variations/pull/1576

browser/ui/webui/ai_chat/code_sandbox_ui.cc
thypon: yup

browser/ui/webui/ai_chat/code_sandbox_ui.cc
thypon: I wonder what happens if the renderer gets compromised, here. CSP should not bubble to any network throttle, but it should be safe even if compromised if we pass a nonetwork throttle

browser/ai_chat/code_execution_tool.cc
thypon: Does creating that introduce high performance penalty? I would do in any case as a defense in depth here

DJAndries: A single secondary OTR profile shouldn't hurt perf

d627b8c440945b9a5c19ff99ae2399dd098bf11a

thypon: https://github.com/brave/brave-browser/issues/51220

browser/ui/webui/ai_chat/code_sandbox_ui.cc
thypon: https://github.com/brave/brave-browser/issues/51220

thypon (approved):
conditionally approved https://github.com/brave/brave-browser/issues/51220

Let's investigate hardening options for GA

browser/ai_chat/code_execution_tool.cc
```diff
@@ -0,0 +1,192 @@
...
+
+  auto wrapped_js_utf16 = base::UTF8ToUTF16(wrapped_js_);
+
+  // Clear the wrapped script to avoid re-using it.
+  wrapped_js_ = {};
+
+  injector_->RequestAsyncExecuteScript(
+      content::ISOLATED_WORLD_ID_GLOBAL, wrapped_js_utf16,
+      blink::mojom::UserActivationOption::kActivate,
+      blink::mojom::PromiseResultOption::kAwait,
```
petemill, line 89: Are we sure we want to call `kAwait` - this means the input script can technically return a promise even though we're asking it to return a string only?

DJAndries: The input script is wrapped in a promise, so the final script will always return a promise

PR #31253, opened 2025-09-17
Do not get clipboard text while checking an input is pasted to the omnibox.

boocmp:
Resolves https://github.com/brave/brave-browser/issues/49272


simonhong (approved):
++ 👍🏼

bsclifton:
@boocmp should we uplift this to Beta / Release?

boocmp:
Yes we should

PR #31254, opened 2025-09-17
adds ts folders as data

nikhedonia:
Resolves

Currently most webpacked assets are not considered by `gn analyze` for change detection and does not yield a list of tests to run.

This is because files are not explicitly declared as inputs. Whilst a list of assets could be manually maintained we choose to simplify by just listing the folders that contain assets.


browser/resources/brave_new_tab_page_refresh/BUILD.gn
```diff
@@ -14,6 +14,10 @@ transpile_web_ui("resources") {
       ] ]
   resource_name = "brave_new_tab_page_refresh"
   output_module = true
+  data = [
```
goodov: I'd say such declaration should be enforced and asserted inside `transpile_web_ui` during the build. Can you compare the generated depsfile with the deps listed here? This will ensure any misconfiguration is caught and is always reflected properly in GN.

I'd also name this variable more specific to somehow reflect that it's required for webpack actions where we don't list all sources explicitly.

goodov: I see you wrote it in "future improvements", but this feels more like a requirement tbh.

nikhedonia: > I'd also name this variable more specific to somehow reflect that it's required for webpack actions where we don't list all sources explicitly.

any name suggestions? 
- files? 
- sourceset?

> I'd say such declaration should be enforced and asserted inside transpile_web_ui during the build

I could cross validate it with the generated manifest file / dep file

goodov: > any name suggestions?
> 
> * files?
> * sourceset?

`source_dirs`?

> I could cross validate it with the generated manifest file / dep file

yes, that would be the right thing to do.

nikhedonia: Source dirs works but technically you could also list individual files...

goodov: `webpack_deps` maybe to be really explicit.

petemill: Let's not use the word webpack

nikhedonia: deps by convention usually takes gn targets?
How about webSourceRoots?

in case we don't find a better name I'm tempted to just call it data since that is the gn nomenclature

components/common/typescript.gni
```diff
@@ -137,6 +137,7 @@ template("transpile_web_ui") {
     # https://gn.googlesource.com/gn/+/master/docs/reference.md#var_depfile
     # https://chromium.googlesource.com/chromium/src/build/+/HEAD/docs/writing_gn_templates.md#inputs-and-depfiles
     depfile = "${target_gen_dir}/${target_name}.d"
+    data_deps_path = "${target_gen_dir}/${target_name}.data_deps.json"
```
petemill: We might want to comment what the purpose of this file is as it's not clear just reading this what the different purposes of `depfile` and `data_deps_path` is. 
Also the name `"data_deps_path"` seems a little confusing since `data_deps` is a GN thing that we're not using here, but we're filling this file with `data` paths. (i.e. "is this the list of paths of the data deps or is this the file that contains the data deps?" where it's actually the path to a file that contains the data paths, but not the data_deps paths). How about `data_paths_file`?

nikhedonia: yeah struggle a bit with naming...
- so in gn we pass the folders as through via the data attribute. 
- we then ask gn to write it to json
- this path tells us where the json file is

we were arguing whether we should rename `data` to `webpack_deps` or something else...
Whatever we choose will probably affect the name of this var too

script/transpile-web-ui.py
```diff
@@ -8,8 +8,61 @@
...
+        outDir  # generated assets are deps and handled by gn already
+    ] + extra_modules
+
+    notContained = getNonContained(allRoots, files)
+
+    if len(notContained) > 0:
+        print(
+            "error occured cross-referencing data folders. Expected following srcRoots:"
+        )
+        print("  " + ", ".join(srcRoots))
```
goodov: 1. this list may be empty so the error text might be confusing
2. pls list items via new lines. something like `"\n  ".join()`

script/transpile-web-ui.py
```diff
@@ -8,8 +8,61 @@
...
+        print("to contain:")
+        print("  " + ", ".join([str(file) for file in notContained]))
```
goodov: pls list items via new lines

script/transpile-web-ui.py
```diff
@@ -8,8 +8,61 @@
...
+
+def verify_webpack_srcs(root_gen_dir, data_deps_path, depfile_path,
+                        extra_modules):
+    srcFolder = os.path.abspath(os.path.join(root_gen_dir, '..', '..', '..'))
+    outDir = os.path.abspath(os.path.join(root_gen_dir, '..'))
+    srcRoots = []
+
+    with open(data_deps_path) as f:
+        srcRoots = json.loads(f.read())
+        srcRoots = [root.replace('//', srcFolder + '/') for root in srcRoots]
```
goodov: Seems like you're going with absolute paths to make it easier to match everything, but for error message it make sense to display GN paths, it's much easier to read and understand at glance what is missing (it will be same on CI and locally this way).

script/transpile-web-ui.py
```diff
@@ -8,8 +8,61 @@
...
+
+    if len(notContained) > 0:
+        print(
+            "error occured cross-referencing data folders. Expected following srcRoots:"
+        )
+        print("  " + ", ".join(srcRoots))
+        print("to contain:")
+        print("  " + ", ".join([str(file) for file in notContained]))
+        print(
+            "fix this issue by adding the containing source folders into the transpile_web_ui target"
```
goodov: this should mention exact variable to fill and ideally the exact GN target.

script/transpile-web-ui.py
```diff
@@ -8,8 +8,61 @@
...
+    Returns list of paths that were not contained
+    """
+    roots = [Path(r).resolve() for r in srcRoots]
+    test_paths = [Path(p).resolve() for p in paths]
+
+    not_contained = []
+
+    for p in test_paths:
+        if not any(p.is_relative_to(root) for root in roots):
+            not_contained.append(p)
```
goodov: `str(p)` and remove `[str(file) for file in notContained]` below.

script/transpile-web-ui.py
```diff
@@ -8,8 +8,61 @@
 import os
 import sys
 import shutil
+import json
 from lib.util import execute_stdout, scoped_cwd
 
+from pathlib import Path
+
+
+def getNonContained(srcRoots, paths):
```
goodov: pls follow naming style from pep-8. this is not JS :)

browser/resources/brave_new_tab_page_refresh/BUILD.gn
fallaciousreasoning: Personally I'm in favor of `data` :smile: `source_dirs` is also good, and I think other gn targets use `inputs`

components/ai_chat/resources/custom_site_distiller_scripts/BUILD.gn
```diff
@@ -19,6 +19,10 @@ transpile_web_ui("custom_site_distiller_scripts") {
       rebase_path("scripts/github.com/distiller.ts"),
     ],
   ]
+  data = [
+    "//brave/components/ai_chat/resources/custom_site_distiller_scripts/",
+    "//brave/components/common/",
```
fallaciousreasoning: I wonder if we should automatically include this in the target - all webuis are going to depend on it 'cause that's where our localization stuff lives

components/ai_chat/resources/BUILD.gn
```diff
@@ -18,6 +18,13 @@ transpile_web_ui("ai_chat_ui") {
           "untrusted_conversation_frame/untrusted_conversation_frame.tsx"),
     ],
   ]
+
+  data = [
+    "//brave/components/ai_chat/resources/",
+    "//brave/components/web-components",
```
fallaciousreasoning, line 26: not for this PR but @petemill we're importing `app.global.scss` - I don't think we should be needing to do import this (and it looks like it sets all the fonts wrong)

components/brave_private_new_tab_ui/resources/page/BUILD.gn
```diff
@@ -1,3 +1,8 @@
+# Copyright (c) 2025 The Brave Authors. All rights reserved.
```
fallaciousreasoning: copyright year should probably be from when the file was created (`2022`)

fallaciousreasoning:
```suggestion
# Copyright (c) 2022 The Brave Authors. All rights reserved.
```

browser/resources/brave_new_tab_page_refresh/BUILD.gn
fallaciousreasoning: `input_folders`?

components/brave_wallet/resources/BUILD.gn
```diff
@@ -1,3 +1,8 @@
+# Copyright (c) 2025 The Brave Authors. All rights reserved.
```
fallaciousreasoning:
```suggestion
# Copyright (c) 2021 The Brave Authors. All rights reserved.
```

components/tor/resources/internals/BUILD.gn
```diff
@@ -1,3 +1,8 @@
+# Copyright (c) 2025 The Brave Authors. All rights reserved.
```
fallaciousreasoning:
```suggestion
# Copyright (c) 2021 The Brave Authors. All rights reserved.
```

ui/webui/resources/BUILD.gn
```diff
@@ -94,6 +94,8 @@ transpile_web_ui("leo_web_components") {
   generate_grdp = true
   resource_path_prefix = "brave"
 
+  data = [ "//brave/ui/webui/resources/" ]
```
fallaciousreasoning: this should probably point to `//brave/ui/webui/resources/leo/`

ui/webui/resources/BUILD.gn
```diff
@@ -117,6 +119,8 @@ transpile_web_ui("opaque_ke") {
     rebase_path("wasm/opaque_ke/src/lib.rs"),
   ]
 
+  data = [ "//brave/ui/webui/resources/" ]
```
fallaciousreasoning: maybe with a `/wasm`?

fallaciousreasoning (commented):
Hey, at the risk of reiterating the massive conversation from Slack:

Why do we need to list out the folders when we're able to verify that we've listed everything during the build anyway?  We could be generating these `data_deps` as the basename of all files in the deps file, right (even with the example of dynamic imports you gave in Slack).

nikhedonia:
> Hey, at the risk of reiterating the massive conversation from Slack:
> Why do we need to list out the folders when we're able to verify that we've listed everything > during the build anyway? We could be generating these data_deps as the basename of all files > in the deps file, right (even with the example of dynamic imports you gave in Slack).

In short because we need this information before we start the build and because gn doesn't process the generated depfile and incoporate the information for eg. `gn analyze`
This affects test determination on CI and RBE. The latter we currently don't care about.

An added benefit of this is that we get to review our dependencies.


There is an alternative solution which is to either:
- manually list every file
- use `exec_script()` in gn and walk the import graph to generate the list. The latter is feasible but needs to be made fast as it will run every time you invoke build. C++ targets don't suffer from this problem because they have special integration with a `scandeps_server` that runs in the background. We could replicate this by integrating with eg. https://facebook.github.io/watchman/ (that's what facebooks buildsystem does) and a library like https://www.npmjs.com/package/depcheck. There might be a way of having two gn configs, one that runs this expensive operation on CI but is disabled local / incremental dev but it is unexplored and may pose a maintainance burden as we'd have two different solutions

components/common/typescript.gni
nikhedonia: renamed it to data_deps_path and added a comment

browser/resources/brave_new_tab_page_refresh/BUILD.gn
nikhedonia: I left it as data since:
1) data is a gn term and we are not changing it's meaning (we just validate completeness)
2) it also supports files not just folders

components/ai_chat/resources/custom_site_distiller_scripts/BUILD.gn
goodov: I also think `"//brave/components/common/"` should be part of the template.

components/common/typescript.gni
```diff
@@ -139,6 +139,11 @@ template("transpile_web_ui") {
     # https://chromium.googlesource.com/chromium/src/build/+/HEAD/docs/writing_gn_templates.md#inputs-and-depfiles
     depfile = "${target_gen_dir}/${target_name}.d"
 
+    # This file contains the data entries and describe the folders and files (indirectly) imported by the entry point
+    # This is important to ensure gn knows about these implicit dependencies.
+    # We write those to json and cross reference them in a later stage to ensure the list is complete
+    data_paths_file = "${target_gen_dir}/${target_name}.data_deps.json"
```
goodov: `.data_deps.json` -> `.data_paths.json`?

script/transpile-web-ui.py
```diff
@@ -8,8 +8,70 @@
 import os
 import sys
 import shutil
+import json
```
goodov: sort pls. let's also put all system imports first followed by local imports. `import` and `from` should also be separated with an empty line. This is what upstream does in `depot_tools`.

```python
import argparse
import json
import os
import sys
import shutil

from pathlib import Path

from lib.util import execute_stdout, scoped_cwd
```

script/transpile-web-ui.py
```diff
@@ -8,8 +8,70 @@
...
+
+    not_contained = [
+        str(path) for path in test_paths
+        if not any(path.startswith(root) for root in roots)
+    ]
+
+    return not_contained
+
+
+def verify_webpack_srcs(root_gen_dir, data_paths_file, depfile_path,
```
goodov: let's declare these new functions in order of use. It's fine to do it this way in python. Other functions are put _after_ `main` already.

script/transpile-web-ui.py
```diff
@@ -8,8 +8,70 @@
...
+                        extra_modules):
+
+    src_folder = Path(os.path.abspath(os.path.join(
+        __file__, '..', '..', '..'))).resolve().as_posix()
+
+    make_source_absolute = lambda path: Path(path).resolve().as_posix(
```
goodov: thank you!

script/transpile-web-ui.py
```diff
@@ -168,5 +177,63 @@ def generate_grd(target_include_dir, grd_name, resource_name,
         execute_stdout(args, env)
 
 
+def get_not_contained(roots, test_paths):
```
goodov: nit: `get_not_contained` should go after `verify_webpack_srcs`.

script/transpile-web-ui.py
```diff
@@ -168,5 +177,63 @@ def generate_grd(target_include_dir, grd_name, resource_name,
...
+        __file__, '..', '..', '..'))).resolve().as_posix()
+
+    make_source_absolute = lambda path: Path(path).resolve().as_posix(
+    ).replace(src_folder, '/')
+
+    out_dir = make_source_absolute(Path(root_gen_dir).resolve().parents[1])
+    src_roots = []
+
+    with open(data_paths_file) as f:
+        src_roots = json.loads(f.read())
```
goodov, line 199: `assert isinstance(src_roots, list)`

script/transpile-web-ui.py
```diff
@@ -168,5 +177,63 @@ def generate_grd(target_include_dir, grd_name, resource_name,
...
+                        extra_modules):
+
+    src_folder = Path(os.path.abspath(os.path.join(
+        __file__, '..', '..', '..'))).resolve().as_posix()
+
+    make_source_absolute = lambda path: Path(path).resolve().as_posix(
+    ).replace(src_folder, '/')
+
+    out_dir = make_source_absolute(Path(root_gen_dir).resolve().parents[1])
+    src_roots = []
```
goodov:
```suggestion
```

goodov (approved):

zenparsing (approved):
Rewards 👍 
NTP Refresh 👍

Douglashdaniel (approved):
++ Wallet Front-end

boocmp (approved):
Tor ++

browser/resources/brave_new_tab_page_refresh/BUILD.gn
petemill: `data` still feels a _little_ weird for this use case. An alternative name could be `imports_from` which still could be files or folders but more explicitly informs what the user is meant to use it for. Otherwise, `data` in the context of webui seems like it could mean JSON files, etc.

components/new_tab_takeover/BUILD.gn
```diff
@@ -20,6 +20,10 @@ transpile_web_ui("new_tab_takeover_ui") {
     "//brave/components/ntp_background_images/browser/mojom:mojom_js",
     "//mojo/public/mojom/base",
   ]
+  data = [
+    "//brave/components/new_tab_takeover/",
+    "//brave/components/brave_new_tab_ui/containers/newTab/",
+  ]
```
petemill: `data` still feels a _little_ weird for this use case. An alternative name could be `imports_from = [` which still reads as if it could contain files or folders but more explicitly informs what the user is meant to use it for. Otherwise, "data" in the context of webui seems like it could imply only JSON files, etc.

script/transpile-web-ui.py
```diff
@@ -168,5 +177,63 @@ def generate_grd(target_include_dir, grd_name, resource_name,
         execute_stdout(args, env)
 
 
+def verify_webpack_srcs(root_gen_dir, data_paths_file, depfile_path,
```
petemill: Can we add a comment as to what this function is doing as I don't think the name is conveying enough about how they will be verified or what they will be verified against, (and what the webpack srcs are). i.e. is it verifying the data paths or the depfile paths?

PR #31280, opened 2025-09-18
require commit hash for sync rebase pr workflow

nikhedonia:
Resolves  https://bravesoftware.slack.com/archives/C6R461GF4/p1757771585820629

There is a race condition between pr review and workflow execution that could be exploited to run untrusted code on our CI.

This PR introduces an additional check against the reviewed commit hash that will need to be provided by the user of the workflow

### DIFF
--- a/.github/workflows/sync-and-rebase-pr-from-fork.yml
+++ b/.github/workflows/sync-and-rebase-pr-from-fork.yml
@@ -6,6 +6,9 @@ on:
       PR_NUMBER:
         required: true
         description: "Number of the PR from the fork"
+      COMMIT_HASH:
+        required: true
+        description: "Commit hash of the PR"
 
 jobs:
   sync-and-rebase-pr-from-fork:
@@ -26,6 +29,7 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
           PR_NUMBER: ${{ inputs.PR_NUMBER }}
+          COMMIT_HASH: ${{ inputs.COMMIT_HASH }}
         run: |
           set -x
           JQ="jq -e -r"
@@ -48,6 +52,13 @@ jobs:
           # Rebase the branch with the latest targeting branch HEAD
           git checkout "$contribHeadRefName"
           git fetch origin "$baseRefName"
+
+          ACTUAL_PR_HASH=$(git rev-parse HEAD)
+          if [[ "$ACTUAL_PR_HASH" != "$COMMI_THASH" ]]; then
+            echo "FORK HASH does not match provided hash! Make sure the PR has no changes that have not been reviewed. Exiting!"
+            exit 1
+          fi
+
           if ! git rebase FETCH_HEAD; then
             echo "Rebase conflict detected. Please resolve the conflicts and push the changes to your fork. Exiting!"
             exit 1
--- a/.github/workflows/sync-and-rebase-pr-from-fork.yml
+++ b/.github/workflows/sync-and-rebase-pr-from-fork.yml
@@ -48,11 +48,12 @@ jobs:
           git remote add contributor "https://github.com/$headRepositoryOwnerLogin/$headRepositoryName.git" || :
           contribHeadRefName="contributor-$headRepositoryOwnerLogin-$headRefName"
           git fetch --no-tags contributor +"$headRefName":"$contribHeadRefName"
-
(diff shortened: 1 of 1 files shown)


.github/workflows/sync-and-rebase-pr-from-fork.yml
```diff
@@ -44,9 +48,11 @@ jobs:
           git remote add contributor "https://github.com/$headRepositoryOwnerLogin/$headRepositoryName.git" || :
           contribHeadRefName="contributor-$headRepositoryOwnerLogin-$headRefName"
           git fetch --no-tags contributor +"$headRefName":"$contribHeadRefName"
-
+          # Note we are checking out the provided commit hash instead of the pr-branch to avoid race condition
+          # The author could have commited malicious code between the last review and the execution of this workflow
+          git checkout "$COMMIT_HASH"
```
wknapik, line 58: Should we make sure we got a sha?
```bash
grep -qix '[0-9a-f]\{40,40\}' <<<"$COMMIT_HASH"
```

.github/workflows/sync-and-rebase-pr-from-fork.yml
```diff
@@ -6,6 +6,9 @@ on:
       PR_NUMBER:
         required: true
         description: "Number of the PR from the fork"
+      COMMIT_HASH:
```
wknapik, line 9: So the person overseeing the external contribution would have to, at the end, verify that the commit that was tested is the last commit on the external PR branch before merging, right? How do we make sure that happens?

And, theoretically, they'd still race the original author with merging, before they push additional commits. So we'd also have to check after merging, that we won that race, no?

nikhedonia: This is how i see things:

1) brave dev reviews a pr of an external contributor and confirms that it's save
2) brave dev takes note of the commit hash
3) brave dev triggers workflow to copy the branch over with the given commit hash
  - if the commit hash doesn't exist because of rebase then we fail
  - if a commit exist then we proceed with that commit
4) CI runs...
5) If external contributor commits new changes you then need to run the workflow again

> So the person overseeing the external contribution would have to, at the end, verify that the commit that was tested is the last commit on the external PR branch before merging, right? How do we make sure that happens?

yes I believe so. I hope there is clear communication between reviewer and external contributor. And hopefully there is an understanding what gets merged

wknapik: As long as merging is done by force-merging the PR from fork via the GitHub webui, there's presumably a race condition, right? The author could push another (potentially malicious) commit as the PR is being force-merged and we might not realize we merged the additional commit, right?

We could change how we merge those PRs though. If we automate it, we can control exactly what commit gets merged.

nikhedonia: That is true.
I don't know the policies but I'd hope since we went through the effort of copying the PR to a brave-core branch and ran CI against it, that we only merge that PR?

wknapik: I'm not sure whether that would end up properly attributing the change to the original author

goodov (commented):
While this change is in the right direction, I believe the process can be hardened further.

Can we add a special label like `contributor-safe-to-merge` that will be set from this workflow in contributor's PR, and will be unset by another workflow that will be run in contributor's PR using `pull_request_target` trigger?

There's an important bit that should allow us do this:
> Workflows triggered by `pull_request_target` events are run in the context of the base branch. Since the base branch is considered trusted, workflows triggered by these events will always run, regardless of approval settings. For more information about the pull_request_target event, see [Events that trigger workflows](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target).


https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#controlling-changes-from-forks-to-workflows-in-public-repositories

nikhedonia:
@goodov I think it's worth reviewing other approaches, this solution is a bit manual.

We need to verify if this event is safe from race-conditions in particular:
> review -> approve -> action queued -> add naughty commit -> action starts running

The event trigger provides the commit sha of the base branch but not of the last approved commit. We can probably query the last approved commit using the github api. 
We need to make sure we are not being taken advantage via a time-of-check, time-of-use race-condition (TOCTOU).

PR #31281, opened 2025-09-18
[NTP Next] Fix handling of google search engine host

zenparsing:
Resolves https://github.com/brave/brave-browser/issues/49380

### DIFF
--- a/browser/resources/brave_new_tab_page_refresh/components/search/engine_icon.tsx
+++ b/browser/resources/brave_new_tab_page_refresh/components/search/engine_icon.tsx
@@ -24,7 +24,7 @@ export function EngineIcon(props: Props) {
 
 function getNamedIcon(engineHost: string) {
   switch (engineHost) {
-    case 'google.com':
+    case 'www.google.com':
       return 'google-color'
     case 'duckduckgo.com':
       return 'duckduckgo-color'
--- a/browser/ui/webui/brave_new_tab_page_refresh/new_tab_page_handler.cc
+++ b/browser/ui/webui/brave_new_tab_page_refresh/new_tab_page_handler.cc
@@ -221,12 +221,14 @@ void NewTabPageHandler::GetAvailableSearchEngines(
         BuiltinEngineType::KEYWORD_MODE_PREPOPULATED_ENGINE) {
       continue;
     }
+    const GURL url(template_url->GenerateSearchURL(
+        template_url_service_->search_terms_data()));
+    if (!url.is_valid() || !url.has_host()) {
+      continue;
+    }
     auto search_engine = mojom::SearchEngineInfo::New();
     search_engine->prepopulate_id = template_url->prepopulate_id();
-    search_engine->host = GURL(template_url->url()).host();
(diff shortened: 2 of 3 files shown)


browser/resources/brave_new_tab_page_refresh/components/search/engine_icon.tsx
```diff
@@ -24,7 +24,7 @@ export function EngineIcon(props: Props) {
 
 function getNamedIcon(engineHost: string) {
   switch (engineHost) {
-    case 'google.com':
+    case 'www.google.com':
```
fallaciousreasoning, line 27: oh weird, why does this one need the `www.` but nothing else?

zenparsing: The google host is stored in the map used by `template_url_service_->GetTemplateURLForHost` as `www.google.com`, and that's host you get when you call `template_url->GenerateSearchURL`. I had it as `google.com` incorrectly here.

browser/ui/webui/brave_new_tab_page_refresh/new_tab_page_handler.cc
```diff
@@ -221,12 +221,14 @@ void NewTabPageHandler::GetAvailableSearchEngines(
         BuiltinEngineType::KEYWORD_MODE_PREPOPULATED_ENGINE) {
       continue;
     }
+    const GURL url(template_url->GenerateSearchURL(
```
zenparsing, line 224: Instead of using `template_url->url()` which may be an invalid URL (for google it is because the hostname is part of the template), I'm using `GenerateSearchURL` which will give us the host that will be required by `template_url_service_->GetTemplateURLForHost` (used in the `OpenSearch` method).

kjozwiak:

```
Brave | 1.84.91 Chromium: 140.0.7339.186 (Official Build) nightly (64-bit)
-- | --
Revision | 7e8a4fbbe4518a30502e47cf8d751057789bad18
```

Using the STR/Cases outlined via https://github.com/brave/brave-browser/issues/49380#issue-3428060765, went through the following:

* installed `1.84.87 Chromium: 140.0.7339.186`
* changed the default search engine from `Brave` -> `Google` via `brave://newtab` using the `Customise dashboard` panel
* closed `1.84.87 Chromium: 140.0.7339.186` and updated to `1.84.89 Chromium: 140.0.7339.186`
* notice that the default search engine is set as `Brave` instead of `Google`

Using the same STR/Cases mentioned above, updated to `1.84.91 Chromium: 140.0.7339.186` which includes the above fix and ensured:

* installed `1.84.87 Chromium: 140.0.7339.186`
* changed the default search engine from `Brave` -> `Google` via `brave://newtab` using the `Customise dashboard` panel
* closed `1.84.87 Chromium: 140.0.7339.186` and updated to `1.84.91 Chromium: 140.0.7339.186`
* ensured that `Google` is still set as the default search engine via `brave://newtab`
* ensured that `Google` is still selected as the default via `Customize New Tab Page`

Screenshot 2025-09-22 182459

PR #31282, opened 2025-09-18
[Unit testing] Ignore death message.

boocmp:
Resolves https://github.com/brave/brave-browser/issues/49400


wknapik:
> I'd like to get to the bottom of why we have a difference in CI environments between PR builds and release builds, but also want to unblock this since it's a non-material unit testing change. Based on DMs with @boocmp, this seems fine to merge.

https://github.com/brave/devops/wiki/Browser-CI-CD#build-types describes all the details and differences between public, test and PR builds. If you'd like to change any of them, please start a thread in #browser-dev.

PR #31285, opened 2025-09-18
Detects URLs better from QR code scanner

SergeyZhukovsky:
Sometimes scanning QR codes searches for the URL within search.brave.com rather than either loading the targeted URL or displaying the URL within the omnibox so the user can load the website that was scanned.
Resolves: https://github.com/brave/brave-browser/issues/48806

Security/Privacy review:
https://github.com/brave/reviews/issues/2054

Test plan is described in the issue.


simoarpe (approved):
👍 LGTM

PR #31293, opened 2025-09-18
add schnorrkel crate to polkadot wasm bundle

cmazakas:
Resolves TBD

This PR adds the schnorrkel Rust crate as a usable wasm bundle, wrapped as a polkadot-named component.

Note: I could only get it to work in practice by changing the global webpack config like so:
```
target: ["web", "es2020"],
```

this is undoubtedly wrong but I'm not sure how to address it. See this branch for example usage: https://github.com/brave/brave-core/tree/polkadot-wasm-bridge

i.e. the polkadot-wasm-bridge branch I have is a fully working demo we can use as a canary in the coal mine for refactoring this PR and cleaning it up. In the ideal case, we'd simply update the polkadot bridge itself to correctly load the wasm modules but I'm unsure how to express that in the build system.

Note: this PR is a continuation of the goals of moving Wallet deps into wasm, specifically implementing a Polkadot keyring via a webui component.

This PR should be blocked on https://github.com/brave/brave-core/pull/31312, in which case rebasing against master will result in a solitary commit introducing the change.


szilardszaloki:
Hi @cmazakas,

Would you mind adding `polkadot_bridge_wasm` to `ui/webui/resources/wasm` in a separate PR?  
I meant specifically these steps should be done in a different PR:
- create a folder for it under https://github.com/brave/brave-core/tree/master/ui/webui/resources/wasm
- add an entry for it in https://github.com/brave/brave-core/blob/master/ui/webui/resources/wasm/Cargo.toml
- run `npm run update_wasm_resources`

cmazakas:
> Hi @cmazakas,
> 
> Would you mind adding `polkadot_bridge_wasm` to `ui/webui/resources/wasm` in a separate PR? I meant specifically these steps should be done in a different PR:
> 
> * create a folder for it under https://github.com/brave/brave-core/tree/master/ui/webui/resources/wasm
> * add an entry for it in https://github.com/brave/brave-core/blob/master/ui/webui/resources/wasm/Cargo.toml
> * run `npm run update_wasm_resources`

This PR lives at https://github.com/brave/brave-core/pull/31312

PR #31295, opened 2025-09-18
fix(l10n/iOS): Portuguese(Portugal & Brazil) l10n not working correctly

nuo-xu:
Remove support for pt-PT localizable identifier. re-import pt-BR translations.

Resolves https://github.com/brave/brave-browser/issues/49424


kjozwiak (approved):
Uplift into `1.82.x` approved 👍

PR #31297, opened 2025-09-19
[Tree Tabs] Implement tab creation and node wrapping functionality

sangwoo108:
This commit introduces core functionality for the Tree Tabs feature, enabling tab creation and node wrapping capabilities when adding new tabs.

Key changes:
- Added tab creation logic with proper node wrapping in tree structure
- When adding a new tab with a opener tab and the opener tab is the previous tab, the new tab's node will be set as a child of the opener tab's node.
- In this way, we'll have a tree structure like below when we open multiple tabs with the previous tab as opener sequentially:

```
    - Tab 1
      - Tab 2
        - Tab 3
          - Tab 4
```

Resolves https://github.com/brave/brave-browser/issues/49754


simonhong:
Sorry, review could be delayed. I have some urgent split view issues from `cr141`.

sangwoo108:
> Sorry, review could be delayed. I have some urgent split view issues from `cr141`.

No worries

components/tabs/impl/brave_tab_strip_collection.cc
```diff
@@ -0,0 +1,57 @@
...
+    bool new_pinned_state) {
+  if (!in_tree_tab_mode_) {
+    TabStripCollection::AddTabRecursive(std::move(tab), index, new_group_id,
+                                        new_pinned_state);
+    return;
+  }
+
+  // If the new |tab|'s opener is the previous tab, we need to insert the new
+  // tab into the opener's TreeTabNode as a child node
+  if (const auto* opener = tab->GetOpener();
```
sangwoo108: Actually, I was wrong about this.
We should check if the previous tab is in the same tree and if not, we should create a new tree.

sangwoo108: Changed.

browser/ui/tabs/test/tree_tabs_browsertest.cc
```diff
@@ -440,3 +446,131 @@ IN_PROC_BROWSER_TEST_F(TreeTabsBrowserTest,
...
+  // nested in other tree tab node ---------------------------------------------
+  // Enable tree tabs to enter tree mode.
+  SetTreeTabsEnabled(true);
+
+  // Create a tab with no opener.
+  tab_interface =
+      std::make_unique<tabs::TabModel>(CreateWebContents(), &tab_strip_model());
+  ASSERT_EQ(nullptr, tab_interface->opener());
+  ASSERT_EQ(nullptr, tab_interface->GetOpener());
+  ASSERT_EQ(tab_interface->GetOpener(), tab_interface->opener());
```
simonhong: Checking opener before adding to model is necessary?

browser/ui/tabs/test/tree_tabs_browsertest.cc
```diff
@@ -440,3 +446,131 @@ IN_PROC_BROWSER_TEST_F(TreeTabsBrowserTest,
...
+
+  // Add tab in tree mode without opener.
+  // note that PAGE_TRANSITION_AUTO_BOOKMARK is used as LINK or TYPED transition
+  // type would be treated as having an opener (the current active tab).
```
simonhong, line 485: According to this comment, maybe we can check `added_tab`'s opener is active tab or not?

browser/ui/tabs/test/tree_tabs_browsertest.cc
```diff
@@ -440,3 +446,131 @@ IN_PROC_BROWSER_TEST_F(TreeTabsBrowserTest,
...
+  // Verify tab was added as child of opener's TreeTabNode.
+  added_tab = tab_strip_model().GetTabAtIndex(index);
+  EXPECT_EQ(added_tab->GetParentCollection()->type(),
+            tabs::TabCollection::Type::TREE_NODE);
+
+  EXPECT_EQ(added_tab->GetParentCollection()->GetParentCollection()->type(),
+            tabs::TabCollection::Type::TREE_NODE);
+
+  EXPECT_EQ(added_tab->GetParentCollection()->GetParentCollection(),
+            tab_strip_model().GetTabAtIndex(index - 1)->GetParentCollection());
```
simonhong: nit: how about comparing tree tab node id?

browser/ui/tabs/test/tree_tabs_browsertest.cc
```diff
@@ -440,3 +446,131 @@ IN_PROC_BROWSER_TEST_F(TreeTabsBrowserTest,
...
+  // The opener's TreeTabNode should now have 2 children (1 original + 1 added).
+  EXPECT_EQ(2u, tab_strip_model()
+                    .GetTabAtIndex(index - 1)
+                    ->GetParentCollection()
+                    ->ChildCount());
+
+  // 4. In tree mode with opener. The previous tab is not the opener  but the
+  // previous tab is a child of opener. In this case, the new tab should be
+  // added as a child of the opener's TreeTabNode.
+  // ----------------------------------- add tab in tree modee with opener as
```
simonhong:
```suggestion
  // ----------------------------------- add tab in tree mode with opener as
```

browser/ui/tabs/test/tree_tabs_browsertest.cc
simonhong: Also it's difficult to read the code with one `index` var.
Can't infer its value easily whenever it refers in this whole test code.
At least having a comment in each place something like `it's 3 now` would be helpful.

browser/ui/tabs/test/tree_tabs_browsertest.cc
```diff
@@ -440,3 +446,131 @@ IN_PROC_BROWSER_TEST_F(TreeTabsBrowserTest,
...
+  tab_strip_model().AddTab(std::move(tab_interface), index,
+                           ui::PAGE_TRANSITION_AUTO_BOOKMARK, ADD_NONE);
+
+  // Verify tab was added and wrapped in its own TreeTabNode, but not as child
+  // of the opener's TreeTabNode.
+  added_tab = tab_strip_model().GetTabAtIndex(index);
+  EXPECT_EQ(opener_tab, added_tab->GetOpener());
+  EXPECT_EQ(added_tab->GetParentCollection()->type(),
+            tabs::TabCollection::Type::TREE_NODE);
+  EXPECT_EQ(added_tab->GetParentCollection()->GetParentCollection(),
```
simonhong, line 575: This means `added_tab` and `opener_tab` have same parent collection?

browser/ui/tabs/test/tree_tabs_browsertest.cc
```diff
@@ -440,3 +446,131 @@ IN_PROC_BROWSER_TEST_F(TreeTabsBrowserTest,
...
+  EXPECT_EQ(opener_tab, added_tab->GetOpener());
+  EXPECT_EQ(added_tab->GetParentCollection()->type(),
+            tabs::TabCollection::Type::TREE_NODE);
+  EXPECT_EQ(static_cast<const TreeTabNode*>(opener_tab->GetParentCollection())
+                ->GetTopLevelAncestor(),
+            static_cast<const TreeTabNode*>(added_tab->GetParentCollection())
+                ->GetTopLevelAncestor());
+
+  // 5. In tree mode with opener not as previous tab: should wrap in new
+  // TreeTabNode but should not be a child of the opener's TreeTabNode. -------
```
simonhong, line 556: Curious why we have the difference when opener is previous tab or not?

chromium_src/chrome/browser/ui/tabs/tab_strip_model.cc
```diff
@@ -3,10 +3,19 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this file,
  * You can obtain one at https://mozilla.org/MPL/2.0/. */
 
+#include "brave/components/tabs/public/brave_tab_strip_collection.h"
```
simonhong: This adding requires `//chrome/browser/ui/tabs:tab_strip_impl` should depends on `//brave/components/tabs/public`.

browser/ui/tabs/test/tree_tabs_browsertest.cc
sangwoo108: Actually, it means when using PAGE_TRANSITION_AUTO_LINK, opener will be set to the active tab during new tab addition automatically inside `TabStripModel`'s method. i.e. this is for prevent opener is set/reset by TabStripModel unintentionally.

browser/ui/tabs/test/tree_tabs_browsertest.cc
sangwoo108: Yes, it was added to make it sure opener is nullptr before calling AddTab() below.
https://github.com/brave/brave-core/pull/31297#discussion_r2377364045

browser/ui/tabs/test/tree_tabs_browsertest.cc
sangwoo108: That means we have another tab tree between opener and newly added tab. In this case, if we want to add the new tab to opener's tree hierarchy, we need to change index to the last child of opener's leaves. I don't think we'd want that.

browser/ui/tabs/test/tree_tabs_browsertest.cc
sangwoo108: Yes, both should be the direct child of unpinned tab collection.

browser/ui/tabs/test/tree_tabs_browsertest.cc
sangwoo108: Removed index stuff 👍

> nit: how about comparing tree tab node id?

I'm not sure what's benefit from using it. I think comparing pointers is enough. Additionally we should down cast the parent collection to tree tab node for it.

browser/ui/tabs/test/tree_tabs_browsertest.cc
sangwoo108: thanks. cleaned up

chromium_src/chrome/browser/ui/tabs/tab_strip_model.cc
sangwoo108: 👍 added deps via patch file.

simonhong: Hmm, maybe we can avoid patching.
Instead how about `contents_data_ = std::make_unique<tabs::BraveTabStripCollection>();` in `BraveTabStripModel`'s ctor?

components/tabs/impl/brave_tab_strip_collection.cc
```diff
@@ -0,0 +1,67 @@
...
+    return;
+  }
+
+  // If the previous tab is in the same hierarchy of tree of the opener, we can
+  // add the new tab to the same tree node.
+  if (const auto* opener = tab->GetOpener()) {
+    auto* opener_collection = opener->GetParentCollection(GetPassKey());
+    CHECK_EQ(opener_collection->type(), TabCollection::Type::TREE_NODE);
+
+    CHECK_GE(index, 1u);
```
simonhong: It's impossible to add at 0 with opener?

simonhong:
Give me some more time. I think I need to understand tab colletions fully.

components/tabs/impl/brave_tab_strip_collection.cc
sangwoo108: I'm not 100% sure but that's my expectation and put the CHECK() to see what happens.

chromium_src/chrome/browser/ui/tabs/tab_strip_model.cc
sangwoo108: That'd be much better.

sangwoo108:
> Give me some more time. I think I need to understand tab colletions fully.

Take your time

browser/ui/tabs/test/tree_tabs_browsertest.cc
```diff
@@ -440,3 +446,126 @@ IN_PROC_BROWSER_TEST_F(TreeTabsBrowserTest,
...
+                           ui::PAGE_TRANSITION_AUTO_BOOKMARK, ADD_NONE);
+
+  added_tab = tab_strip_model().GetTabAtIndex(tab_strip_model().count() - 1);
+  EXPECT_EQ(opener_tab, added_tab->GetOpener());
+  EXPECT_EQ(added_tab->GetParentCollection()->type(),
+            tabs::TabCollection::Type::TREE_NODE);
+  EXPECT_EQ(static_cast<const TreeTabNode*>(opener_tab->GetParentCollection())
+                ->GetTopLevelAncestor(),
+            static_cast<const TreeTabNode*>(added_tab->GetParentCollection())
+                ->GetTopLevelAncestor());
```
simonhong, line 547: How about having this expection also here for checking new tab's parent precisely?
```
  EXPECT_EQ(opener_tab->GetParentCollection(),
            added_tab->GetParentCollection()->GetParentCollection());
```
also it would be great to check added tab is positioned before the previous tab to confirm this comment` 
// add tab in tree mode with opener as the tab before the previous tab`?

browser/ui/tabs/test/tree_tabs_browsertest.cc
```diff
@@ -440,3 +446,126 @@ IN_PROC_BROWSER_TEST_F(TreeTabsBrowserTest,
...
+  EXPECT_EQ(added_tab->GetParentCollection()->GetParentCollection(),
+            opener_tab->GetParentCollection());
+
+  // The opener's TreeTabNode should now have 2 children (1 original + 1 added).
+  EXPECT_EQ(2u, opener_tab->GetParentCollection()->ChildCount());
+
+  // 4. In tree mode with opener. The previous tab is not the opener  but the
+  // previous tab is a child of opener. In this case, the new tab should be
+  // added as a child of the opener's TreeTabNode. -----------------------------
+  // add tab in tree modee with opener as the tab before the previous tab.
```
simonhong:
```suggestion
  // add tab in tree mode with opener as the tab before the previous tab.
```

browser/ui/tabs/brave_tab_strip_model.cc
```diff
@@ -36,17 +37,21 @@ BraveTabStripModel::BraveTabStripModel(
...
-        base::BindRepeating(&BraveTabStripModel::OnTreeTabRelatedPrefChanged,
-                            base::Unretained(this)));
-    vertical_tabs_enabled_.Init(
-        brave_tabs::kVerticalTabsEnabled, profile->GetPrefs(),
-        base::BindRepeating(&BraveTabStripModel::OnTreeTabRelatedPrefChanged,
-                            base::Unretained(this)));
-    OnTreeTabRelatedPrefChanged();
+  if (base::FeatureList::IsEnabled(tabs::features::kBraveTreeTab)) {
+    // Replace the default TabStripCollection with Brave's version
+    contents_data_ = std::make_unique<tabs::BraveTabStripCollection>();
```
simonhong, line 42: we need `BraveTabStripCollection` always? maybe only for normal window?

simonhong: Ah, ok. `in_tree_tab_mode()` is used.

components/tabs/impl/brave_tab_strip_collection.cc
simonhong: Hmm, maybe it's possible when moving a tab from pinned to unpinned?
Or tab moving in unpinned container?

components/tabs/impl/brave_tab_strip_collection.cc
```diff
@@ -0,0 +1,67 @@
...
+    auto* previous_tab_collection =
+        previous_tab->GetParentCollection(GetPassKey());
+    CHECK_EQ(previous_tab_collection->type(), TabCollection::Type::TREE_NODE);
+
+    if (static_cast<TreeTabNode*>(opener_collection)->GetTopLevelAncestor() ==
+        static_cast<TreeTabNode*>(previous_tab_collection)
+            ->GetTopLevelAncestor()) {
+      auto tree_tab_node = std::make_unique<TreeTabNode>(
+          tree_tab::TreeTabNodeId::GenerateNew(), std::move(tab));
+      opener_collection->AddCollection(std::move(tree_tab_node), 1);
```
simonhong: why wrapped tree node is inserted at `1` always?

sangwoo108: Because #0 is reserved for the "Tab" for the collection itself. 
i.e. opener collection will have the opener tab at #0

simonhong: IIUC, it seems `previous tab` and `opener tab` are in same hierachy, new tab will be added to right after position of `operner tab` as it's added at `1` always. Am I right?

sangwoo108: Yes, that's correct.

```
- Opener tab collection(treenode)
   - #0 opener tab interface
   - #1 new tab collection(treenode)
       - #0 new tab interface
   - #2 other tab collection
       - #0 other tab interface
```

Maybe it'd better to add new tabs to the last of children? Is it what you're implying?

simonhong: IMO, last position is more natural as it's added lastly?

components/tabs/impl/brave_tab_strip_collection.cc
sangwoo108: Yeah, could be. Let's move this to conditional.

components/tabs/impl/brave_tab_strip_collection.cc
sangwoo108: Yeah that sounds better. I'm sold

sangwoo108:
Checking some failures

chromium_src/chrome/browser/ui/tabs/tab_strip_model.cc
```diff
@@ -4,9 +4,12 @@
  * You can obtain one at https://mozilla.org/MPL/2.0/. */
 
 #include "chrome/browser/ui/views/tabs/dragging/tab_drag_controller.h"
+#include "components/tabs/public/tab_strip_collection.h"
```
simonhong, line 7: we don't need any changes in this file?

sangwoo108: Oops, thanks :)

chromium_src/components/tabs/public/tab_interface.h
```diff
@@ -0,0 +1,25 @@
...
+#ifndef BRAVE_CHROMIUM_SRC_COMPONENTS_TABS_PUBLIC_TAB_INTERFACE_H_
+#define BRAVE_CHROMIUM_SRC_COMPONENTS_TABS_PUBLIC_TAB_INTERFACE_H_
+
+#include "build/build_config.h"
+
+// Add method declarations to TabInterface to get opener of this tab.
+#if !BUILDFLAG(IS_ANDROID)
+#define IsInNormalWindow                 \
+  IsInNormalWindow() const = 0;          \
+  virtual TabInterface* GetOpener() = 0; \
```
bridiver: we generally should not be modifying the public api for things like TabInterface. If you want to add something like this you should go through upstream

chromium_src/chrome/browser/ui/tabs/tab_model.cc
```diff
@@ -16,3 +16,14 @@
 #include <chrome/browser/ui/tabs/tab_model.cc>  // IWYU pragma: export
 
 #undef UNPINNED
+
+namespace tabs {
+
+TabInterface* TabModel::GetOpener() {
```
bridiver: this is already available through `TabModel::opener()`

components/tabs/public/brave_tab_strip_collection.h
```diff
@@ -0,0 +1,40 @@
...
+
+#ifndef BRAVE_COMPONENTS_TABS_PUBLIC_BRAVE_TAB_STRIP_COLLECTION_H_
+#define BRAVE_COMPONENTS_TABS_PUBLIC_BRAVE_TAB_STRIP_COLLECTION_H_
+
+#include "components/tabs/public/tab_strip_collection.h"
+
+namespace tabs {
+
+// BraveTabStripCollection is used to override methods for tree tabs.
+// * In AddTabRecursive(), we need to to create TreeTabNode with a given new tab
```
bridiver, line 14: why do we only need to do this in `AddTabRecursive` and not `MoveTabRecursive` or `MoveTabsRecursive`?

sangwoo108: That's a planned work. I tried to focus on "add" event in order to minimize the PR

bridiver: that makes sense, but this should have a feature flag then and just add `MoveTabRecursive`, etc... with NOTREACHED and TODO(issue link)

chromium_src/components/tabs/public/tab_interface.h
sangwoo108: I think this is a Brave specific requirement, so upstream folks wouldn't think this is needed in upstream. Is there a better way to achieve this without modifying the API?

chromium_src/chrome/browser/ui/tabs/tab_model.cc
sangwoo108: We can't use TabModel from TreeTabNode, as it's layering violation. TabMoel is in //chrome layer implmenting TabInterface in //components/tabs/public .

chromium_src/components/tabs/public/tab_interface.h
bridiver: I don't see why they would object to adding `opener()` here as a virtual method since it's already in TabModel

bridiver: I think you could also accomplish this with WebContents::GetFirstWebContentsInLiveOriginalOpenerChain and then convert back to TabInterface again. Per the docs this is available even if the opener was suppressed

sangwoo108: > I don't see why they would object to adding opener() here as a virtual method since it's already in TabModel

I may be biased, but based on my previous experience, upstream folks have generally been skeptical about accepting changes that don't have usage within the chromium code base. The reason they give was that when the usage isn't included together in the codebase, it becomes difficult to maintain

> WebContents::GetFirstWebContentsInLiveOriginalOpenerChain

This looked promising but has sutle difference betwen TabModel::opener(). When you open a link via wheel button of mouse, TabModel::opener() is valid but the WebContents's method returns null.

---

I think we can take opener as a parameter of AddTabRecursive(). As TabStripModel is allowed to access TabModel directly, we won't need TabInterface modification.

chromium_src/chrome/browser/ui/tabs/tab_model.cc
sangwoo108: Removed.

components/tabs/public/brave_tab_strip_collection.h
sangwoo108: Added stub code and NOTIMPELEMENTED() macros so that browser tests passes.

sangwoo108:
Converting this into draft as we're talking about overall architecture first.

PR #31302, opened 2025-09-19
Merge MV2 extension deprecation stages into DeprecationWarning

boocmp:
We decided/clarified that we want to merge the 3 states mentioned in https://github.com/brave/brave-browser/issues/49257#issue-3416097569 into `ExtensionManifestV2DeprecationWarning`, with an updated warning message for MV2 extensions pointing users to Brave's MV2 settings page.

The idea is that we want to just show a warning to Brave's users about impending MV2 removal and point them to the Brave-hosted extensions for as long as we can.

### DIFF
--- a/chromium_src/chrome/browser/extensions/manifest_v2_experiment_manager.cc
+++ b/chromium_src/chrome/browser/extensions/manifest_v2_experiment_manager.cc
@@ -1,0 +1,31 @@
+/* Copyright (c) 2025 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "chrome/browser/extensions/manifest_v2_experiment_manager.h"
+
+#include "brave/browser/extensions/manifest_v2/features.h"
+#include "chrome/browser/extensions/mv2_experiment_stage.h"
+#include "extensions/browser/extension_registry.h"
+
+namespace {
+
+extensions::MV2ExperimentStage AdjustMV2ExperimentStage(
+    extensions::MV2ExperimentStage experiment_stage) {
+  if (!base::FeatureList::IsEnabled(
+          extensions_mv2::features::kExtensionsManifestV2)) {
+    return experiment_stage;
+  }
+  // We relax any stage into the Warning stage.
+  return extensions::MV2ExperimentStage::kWarning;
+}
+
+}  // namespace
+
+#define experiment_stage_(...) \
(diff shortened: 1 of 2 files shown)


chromium_src/chrome/browser/extensions/manifest_v2_experiment_manager.cc
```diff
@@ -0,0 +1,34 @@
...
+    experiment_stage = extensions::MV2ExperimentStage::kWarning;
+  }
+}
+
+}  // namespace
+
+#define Observe(...)        \
+  Observe(__VA_ARGS__);     \
+  AdjustMV2ExperimentStage( \
+      const_cast<extensions::MV2ExperimentStage&>(experiment_stage_))
```
emerick: I don't think it's generally safe to cast away `const` like this and the Chromium guidelines discourage it too. Since `experiment_stage_` is initialized via `experiment_stage_(CalculateCurrentExperimentStage())` I think you can define `experiment_stage_` as a function-like macro and safely implement the new behavior that way instead.

emerick: Example of this technique is here https://github.com/brave/brave-core/blob/master/chromium_src/components/update_client/request_sender.cc#L35-L36.

boocmp: Thanks for example! I've updated the patch

chromium_src/chrome/browser/extensions/manifest_v2_experiment_manager.cc
```diff
@@ -0,0 +1,33 @@
+/* Copyright (c) 2025 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "chrome/browser/extensions/manifest_v2_experiment_manager.h"
+
+#include "brave/browser/extensions/manifest_v2/features.h"
```
emerick, line 8: Can you use this forward-declaration technique to eliminate the need to include a brave file here? https://github.com/brave/brave-core/pull/30984/files#diff-07d7235ead46632a187c2fb39037a32eaad6676580d9c07c8aa7553ef518d5cbR8

Otherwise, we would need to add a dep and it gets problematic quickly. Sorry I didn't notice this the first time.

boocmp: But chrome/browser/extensions already depends on brave/browser/extensions, or I missed something ?

emerick: OK, I see now. That seems fine then.

emerick (approved):

PR #31303, opened 2025-09-19
[Android] Initial integration of Brave adaptive buttons

samartnik:
The Bookmarks button is the first in the list; the rest of the buttons will be added in separate PRs.

Important notes:
- override via `chromium_src` doesn't work for `patches/chrome-browser-ui-android-toolbar-adaptive_toolbar_enums.h.patch` since this file is parsed by `java_cpp_enum.py` to convert it to java enum
- patch to `java_cpp_enum.py` was added to be able to do 1 line patch, since upsteam parses each value as a separate line
- in `patches/chrome-browser-ui-android-toolbar-adaptive_toolbar_enums.h.patch` were added values for all planned Brave buttons to avoid additional patching for each consecutive PR

Partially resolves https://github.com/brave/brave-browser/issues/49409


patches/chrome-browser-ui-android-toolbar-adaptive_toolbar_enums.h.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/chrome/browser/ui/android/toolbar/adaptive_toolbar_enums.h b/chrome/browser/ui/android/toolbar/adaptive_toolbar_enums.h
+index 346d0fc2b0fb5a9dcb093814511ffb2afa90f463..4faecaa82111ba9d684ebb0a08c64bf8dbc18db8 100644
+--- a/chrome/browser/ui/android/toolbar/adaptive_toolbar_enums.h
++++ b/chrome/browser/ui/android/toolbar/adaptive_toolbar_enums.h
+@@ -50,7 +50,7 @@ enum class AdaptiveToolbarButtonVariant {
+   // Tab grouping action.
+   kTabGrouping = 16,
+ 
+-  kMaxValue = kTabGrouping,
++  /* BRAVE_JAVA_ENUM: */kBookmarks = kTabGrouping + 1, kHistory = kBookmarks + 1, kDownloads = kHistory + 1, kPlaylist = kDownloads + 1, kLeo = kPlaylist + 1, kWallet = kLeo + 1, kNews = kWallet + 1, kMaxValue = kNews,
```
samartnik: override via `chromium_src` doesn't work here since this file is parsed by `java_cpp_enum.py` to convert it to java enum
all planned values for Brave buttons were added here to avoid additional patching for each consecutive PR

patches/build-android-gyp-java_cpp_enum.py.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/build/android/gyp/java_cpp_enum.py b/build/android/gyp/java_cpp_enum.py
+index ffd896d9bc7ecb937e9f367400eea0cb18defbd9..87dc0091b74f80d5df3674566bc23478738a7fb3 100755
+--- a/build/android/gyp/java_cpp_enum.py
++++ b/build/android/gyp/java_cpp_enum.py
+@@ -262,6 +262,7 @@ class HeaderParser:
+     self._generator_directives = DirectiveSet()
+ 
+   def ParseDefinitions(self):
++    from brave_java_cpp_enum_utils import SplitBraveJavaEnumEntries; self._lines = SplitBraveJavaEnumEntries(self._lines)
```
samartnik: added to be able to do 1 line patch in `patches/chrome-browser-ui-android-toolbar-adaptive_toolbar_enums.h.patch`, since upstream parses each value as a separate line

AlexeyBarabash (approved):
LGTM!

patches/build-android-gyp-java_cpp_enum.py.patch
goodov: please do this via `chromium_src` file patch.

we should not add new python patches via `scripts/*.py` files, this prevents correct rebuilds and prevents `siso` actions to be run correctly on remote.

samartnik: thank you for clarifications, pushed the change

goodov: After looking at this closely I see that you did this patch only for the purpose of unwrapping a single line patch done in the header.

It would be reasonable if the patch was done via `#define` that was somehow unwrapped via python script, but I don't see an easy way of doing this correctly. So we end up with a patch that will need to be changed each time we want to add/modify a value.

In this case I'd say we should go with a multiline patch with a comment on why this is done like that, and drop the python override. We had a similar multiline patch in content settings types before they were moved to mojo, and it was fine as a multiline patch.

Although maybe you can make it work for `#define`-inlined patches, then it would be great to have it work automatically for all enums that are exposed to Java via this custom Chromium machinery.

goodov (approved):
lgtm, but the python patch for java enum should either be removed in favor of a multiline patch in the header or improved to make it work for define-style patch.

patches/build-android-gyp-java_cpp_enum.py.patch
samartnik: I'll go with multiline patch, since `#define` patch is not straightforward, besides `plaster` will make such workaround obsolete anyways

PR #31304, opened 2025-09-19
[NTP] Fix settings modal popping over News modal

zenparsing:
Resolves https://github.com/brave/brave-browser/issues/49423

### Explanation

Both the NTP settings modal and News modal are HTML top-layer dialogs. The settings modal is configured to close when the user clicks outside of the modal rectangle. When the user opens the News modal from the settings modal, both modals are displayed. When the user clicks "outside" the settings modal, it closes the modal. When the React component for the settings modal is re-rendered (due to a state change), the modal is re-opened. Since it is a top-layer element, it displays on top of the News modal.

It's not clear why this worked previously (perhaps there was a blink rendering change?). The solution is to disable the "close on click outside" behavior if the News modal is open.

### DIFF
--- a/browser/resources/brave_new_tab_page_refresh/components/settings/settings_modal.tsx
+++ b/browser/resources/brave_new_tab_page_refresh/components/settings/settings_modal.tsx
@@ -125,14 +125,8 @@ export function SettingsModal(props: Props) {
       <Dialog
         isOpen={props.isOpen}
         showClose
-        onClose={() => {
-          // If the News dialog is open, keep the settings dialog open so that
-          // closing the News dialog will bring the user back to the settings
-          // dialog.
-          if (!braveNews.customizePage) {
-            props.onClose()
-          }
-        }}
+        onClose={() => props.onClose()}
+        backdropClickCloses={!braveNews.customizePage}
       >
         <h3>
           {getString('settingsTitle')}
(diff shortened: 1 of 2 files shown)


fallaciousreasoning (approved):

kjozwiak:

```
Brave | 1.84.91 Chromium: 140.0.7339.186 (Official Build) nightly (64-bit)
-- | --
Revision | 7e8a4fbbe4518a30502e47cf8d751057789bad18
```

Reproduced the original issue with `1.84.89 Chromium: 140.0.7339.186` using the STR/Cases outlined via https://github.com/brave/brave-browser/issues/49423#issue-3431994243 as per the following:

https://github.com/user-attachments/assets/28cf7849-a7fb-4ad5-a717-a26af2673a3e

Using the same STR/Cases outline above, ensured that the `Dashboard` wasn't being moved into the forefront when interacting with the `News` settings as per the following:

* ensured that `Customize New Tab Page` isn't being moved to the forefront when interacting with `Open articles in`
* ensured that `Customize New Tab Page` isn't being moved to the forefront when adding `Popular` news sources
* ensured that `Customize New Tab Page` isn't being moved to the forefront when adding `Suggested` news sources
* ensured that clicking on `Back to dashboard` closes `Brave News` and displays `Customize New Tab Page`

https://github.com/user-attachments/assets/a3f8870b-9bcd-4343-8ef9-34d3feeb0860

PR #31308, opened 2025-09-19
[ads] Dynamic NTT Brave Search integration

tmancey:
This PR enables the browser to query Brave Search autocomplete and inform the Dynamic NTT when suggestions are available. The Dynamic NTT, distributed through the component updater, is responsible for displaying those suggestions and allowing the user to navigate to the selected URL.

NOTE: This feature will remain a no-op until a dynamic NTT that supports it is published. Before publishing, discussions with the relevant teams are needed to decide how to handle cases where the search widget is visible, since displaying both does not make sense. We plan to use SmartNTT condition matchers to hide this feature for users with the widget enabled, or consider alternative solutions.

Pre-requisites:
- Discuss with @tmancey to test against a demo Dynamic NTT (https://github.com/brave/ads-rich-ntt-campaigns/pull/108), which he can provide. The dynamic NTT is only a demo and should not be tested beyond the core functionality listed below.

Testing:
- Suggestions should appear in the search box as you type a query
- Selecting a suggestion should navigate to the corresponding page
- Verify that clicked ad events are triggered

Resolves https://github.com/brave/brave-browser/issues/49338

### DIFF
--- a/browser/resources/brave_new_tab_page_refresh/components/background/background.tsx
+++ b/browser/resources/brave_new_tab_page_refresh/components/background/background.tsx
@@ -10,6 +10,8 @@ import { openLink } from '../common/link'
 import { loadImage } from '../../lib/image_loader'
 import { useWidgetLayoutReady } from '../app_layout_ready'
 import { debounce } from '$web-common/debounce'
+import { useSearchState, useSearchActions } from '../../context/search_context'
+import { AutocompleteMatch } from '../../state/search_state'
 
 import {
   useBackgroundState,
@@ -95,21 +97,59 @@ function SponsoredRichMediaBackground(
 
   useSafeAreaReporter(frameHandle)
 
+  // TODO(https://github.com/brave/brave-browser/issues/49471): [NTP Next]
+  // Refactor rich media background components.
+  const searchActions = useSearchActions()
+  const searchMatches = useSearchState((s) => s.searchMatches)
+  useSearchMatchesReporter(frameHandle, searchMatches)
+
   return (
     <IframeBackground
       url={props.background.imageUrl}
       expectedOrigin={new URL(sponsoredRichMediaBaseUrl).origin}
       onReady={setFrameHandle}
-      onMessage={(data) => {
-        const eventType = getRichMediaEventType(data)
-        if (eventType) {
-          actions.notifySponsoredRichMediaEvent(eventType)
+      onMessage={(data: any) => {
+        if (!data ||
+          typeof data !== 'object' ||
+          typeof data.type !== 'string') {
+          return
         }
-        if (eventType === NewTabPageAdEventType.kClicked) {
-          const url = props.background.logo?.destinationUrl
-          if (url) {
-            openLink(url)
+
+        switch (data.type) {
+          case 'richMediaEvent': {
+            const value = String(data.value ?? '')
+            const eventType = getRichMediaEventType(value)
+            if (eventType) {
+              actions.notifySponsoredRichMediaEvent(eventType)
+              if (eventType === NewTabPageAdEventType.kClicked) {
+                const url = props.background.logo?.destinationUrl
+                if (url) {
+                  openLink(url)
+                }
+              }
+            }
+            break
           }
+          case 'richMediaQueryBraveSearchAutocomplete': {
+            const value = String(data.value ?? '')
+            if (value) {
+              searchActions.queryAutocomplete(value, 'search.brave.com')
+            }
+            break
+          }
+          case 'richMediaOpenBraveSearchWithQuery': {
+            const value = String(data.value ?? '')
+            if (value) {
+              // Opening Brave Search from rich media is treated as an ad click
+              // for reporting purposes.
+              actions.notifySponsoredRichMediaEvent(
+                NewTabPageAdEventType.kClicked)
+              openLink(`https://search.brave.com/${value}`)
+            }
+            break
+          }
+          default:
+            break
         }
       }}
     />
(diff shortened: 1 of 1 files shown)


bsclifton (approved):
Code changes LGTM; couldn't hurt to have some tests - but I like it 🙂

browser/resources/brave_new_tab_page_refresh/components/background/background.tsx
```diff
@@ -95,21 +97,57 @@ function SponsoredRichMediaBackground(
...
+                  openLink(url)
+                }
+              }
+            }
+            break
+          }
+          case 'richMediaQueryBraveSearchAutocomplete': {
+            const value = String(data.value)
+            if (value) {
+              searchActions.queryAutocomplete(value, "search.brave.com")
```
zenparsing: nit: single-quotes

browser/resources/brave_new_tab_page_refresh/components/background/background.tsx
```diff
@@ -156,19 +194,37 @@ function useSafeAreaReporter(frameHandle?: IframeBackgroundHandle) {
...
+) {
+  React.useEffect(() => {
+    if (!frameHandle) {
+      return
+    }
+
+    const postSearchMatches = debounce(() => {
+      frameHandle.postMessage({
+        type: 'richMediaSearchMatches',
+        value: JSON.stringify(searchMatches ?? [])
```
zenparsing, line 224: Can we just send the data that we need, rather than the entire array as a serialized string?

```ts
searchMatches.map((match) => ({
   // whatever item data we need here.
})
```

browser/resources/brave_new_tab_page_refresh/components/background/background.tsx
```diff
@@ -156,19 +194,37 @@ function useSafeAreaReporter(frameHandle?: IframeBackgroundHandle) {
...
+      })
+    }, 120)
+
+    postSearchMatches()
```
zenparsing, line 228: Just a heads up that this will post the search matches into all background iframes, even if the iframe is not interested in listening to autocomplete results.

browser/resources/brave_new_tab_page_refresh/components/background/background.tsx
```diff
@@ -95,21 +97,57 @@ function SponsoredRichMediaBackground(
...
+            break
           }
+          case 'richMediaOpenBraveSearchWithQuery': {
+            const value = String(data.value)
+            if (value) {
+              // Opening Brave Search from rich media is treated as an ad click
+              // for reporting purposes.
+              actions.notifySponsoredRichMediaEvent(
+                NewTabPageAdEventType.kClicked)
+              openLink(`https://search.brave.com/search?${value}`)
```
zenparsing: You may want to send the entire URL or the entire path plus query, so things like `https://search.brave.com/ask?q={query}` would be possible.

If you want to send the entire URL, we could validate the hostname on this side, as in your previous version.

browser/resources/brave_new_tab_page_refresh/components/background/background.tsx
```diff
@@ -95,21 +97,57 @@ function SponsoredRichMediaBackground(
...
   return (
     <IframeBackground
       url={props.background.imageUrl}
       expectedOrigin={new URL(sponsoredRichMediaBaseUrl).origin}
       onReady={setFrameHandle}
-      onMessage={(data) => {
-        const eventType = getRichMediaEventType(data)
-        if (eventType) {
-          actions.notifySponsoredRichMediaEvent(eventType)
+      onMessage={(data: SponsoredRichMediaMessage) => {
```
zenparsing: minor: It would be better to use `any` for the type here, rather than use `SponsoredRichMediaMessage` which won't necessarily be reliable.

browser/resources/brave_new_tab_page_refresh/components/background/background.tsx
```diff
@@ -95,21 +97,57 @@ function SponsoredRichMediaBackground(
...
+          return
         }
-        if (eventType === NewTabPageAdEventType.kClicked) {
-          const url = props.background.logo?.destinationUrl
-          if (url) {
-            openLink(url)
+
+        switch (data.type) {
+          case 'richMediaEvent': {
+            const value = String(data.value)
```
zenparsing: `String(data.value)` could yield `"undefined"`.

```
const value = String(data.value ?? '')
```

browser/resources/brave_new_tab_page_refresh/components/background/background.tsx
```diff
@@ -95,21 +97,57 @@ function SponsoredRichMediaBackground(
 
   useSafeAreaReporter(frameHandle)
 
+  const searchActions = useSearchActions()
+  const searchMatches = useSearchState((s) => s.searchMatches)
+  useSearchMatchesReporter(frameHandle, searchMatches)
+
```
zenparsing, line 105: Please add a TODO pointed to https://github.com/brave/brave-browser/issues/49471.

browser/resources/brave_new_tab_page_refresh/components/background/background.tsx
tmancey: The current dynamic NTT uses a few values, but future NTTs may include additional data. This approach avoided the need to make browser changes which ride the trains if we want to display more information later. Thanks

browser/resources/brave_new_tab_page_refresh/components/background/background.tsx
tmancey: Fixed, thanks

browser/resources/brave_new_tab_page_refresh/components/background/background.tsx
tmancey: Fixed, thanks

browser/resources/brave_new_tab_page_refresh/components/background/background.tsx
tmancey: Fixed, thanks

browser/resources/brave_new_tab_page_refresh/components/background/background.tsx
tmancey: Added, thanks

browser/resources/brave_new_tab_page_refresh/components/background/background.tsx
tmancey: 👀

tmancey: Changed so that we now pass the path?query, thanks

browser/resources/brave_new_tab_page_refresh/components/background/background.tsx
zenparsing: The [robustness principle](https://en.wikipedia.org/wiki/Robustness_principle) is generally a good idea in protocol design.

> be conservative in what you send, be liberal in what you accept


But your call.

browser/resources/brave_new_tab_page_refresh/components/background/background.tsx
tmancey: Let me raise an issue to follow up on this. Thanks

browser/resources/brave_new_tab_page_refresh/components/background/background.tsx
tmancey: Agreed and will raise a follow up issue once I understand the bare minimum the NTTs will use. Thanks

zenparsing: Could this lead to version mismatch problems, if the protocol is changed in a backward incompatible way? Or are NTTs targeted to specific versions of the browser? Ideally it would send structured data and not a JSON string.

zenparsing: [window.postMessage](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) should be able to send structured data using [structured clone](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm). Technically JSON isn't required.

tmancey: We control NTT by browser version, but once this merges let me investigate the structured clone. Thanks for this.

PR #31311, opened 2025-09-19
Ad Block Only mode settings UI

aseren:
The PR adds `Ad Block Only mode` alert and `Ad Block Only mode` toggle to the Shields page. 

The UI is under a feature which is turned off by default.

Shields settings with `Ad Block Only mode` enabled:
image

Shields settings with `Ad Block Only mode` disabled:
image

Epic for this task is: https://github.com/brave/brave-browser/issues/48521
Spec: https://docs.google.com/document/d/1-y3NIRvt0scNK7qJ-XUuY0wRcnFuT08xsxivbUtr9SE/edit?tab=t.0

Resolves https://github.com/brave/brave-browser/issues/48559

### DIFF
--- a/app/brave_settings_strings.grdp
+++ b/app/brave_settings_strings.grdp
@@ -605,6 +605,26 @@
     Show Brave Rewards button
   </message>
 
+  <!-- Settings / Ad block only mode -->
+  <message name="IDS_SETTINGS_AD_BLOCK_ONLY_MODE" desc="Title for an item in the 'Ad block only mode' section of Shields Settings">
+    Ad Block Only mode
+  </message>
+  <message name="IDS_SETTINGS_AD_BLOCK_ONLY_MODE_LABEL" desc="The text for settings option to enable shields ad block only mode">
+    Use Ad Block Only mode
+  </message>
+  <message name="IDS_SETTINGS_AD_BLOCK_ONLY_MODE_DESC" desc="The description for settings option to enable shields ad block only mode">
+    Ad Block Only mode retains core ad blocking rules of Brave Shields. Warning: removes all privacy protections. Please note that for individual websites broken by Brave's privacy protections, you can set Shields to DOWN. This is an experimental mode. <ph name="BEGIN_LINK">&lt;a target="_blank" href="$1"&gt;</ph>Learn more<ph name="END_LINK">&lt;/a&gt;</ph>
+  </message>
+  <message name="IDS_SETTINGS_AD_BLOCK_ONLY_MODE_ALERT_TITLE" desc="Title for the alert shown when ad block only mode is enabled">
+    Shields options are currently disabled.
+  </message>
+  <message name="IDS_SETTINGS_AD_BLOCK_ONLY_MODE_ALERT_DESC" desc="Description for the alert shown when ad block only mode is enabled">
+    Ad Block Only mode retains core ad blocking rules of Brave Shields and turns off privacy protections. If you would like to enable standard Shields, please disable this mode.
+  </message>
+  <message name="IDS_SETTINGS_AD_BLOCK_ONLY_MODE_ALERT_TURN_OFF_BUTTON" desc="Alert button text to turn off ad block only mode">
+    Turn off Ad Block Only mode
+  </message>
+
   <!-- Settings / Social media blocking -->
   <message name="IDS_SETTINGS_SOCIAL_BLOCKING_DEFAULTS_TITLE" desc="The title the settings section which has options for blocking third-party embeds, commonly from social media sites">
     Social media blocking
--- a/browser/resources/settings/ad_block_only_mode_page/ad_block_only_mode_alert.html
+++ b/browser/resources/settings/ad_block_only_mode_page/ad_block_only_mode_alert.html
@@ -1,0 +1,26 @@
+<style include="settings-shared">
+  .alert-container {
+    padding-top: var(--leo-spacing-m);
+    padding-bottom: var(--leo-spacing-xl);
+  }
+  .alert-title {
+    font: var(--leo-font-heading-h4);
+  }
+  .alert-content {
+    font: var(--leo-font-default-regular);
+  }
+</style>
+
+<template is="dom-if" if="[[isAdBlockOnlyModeEnabled_]]">
+  <div class="alert-container">
+    <leo-alert type="info" hideIcon="true">
+      <div class="alert-title">$i18n{adBlockOnlyModeAlertTitle}</div>
+      <div class="alert-content">$i18n{adBlockOnlyModeAlertDesc}</div>
+      <div slot="actions">
+        <leo-button on-click="onTurnOffAdBlockOnlyMode_" class="alert-button">
+          $i18n{adBlockOnlyModeAlertTurnOffButton}
+        </leo-button>
+      </div>
+    </leo-alert>
+  </div>
+</template>
--- a/browser/resources/settings/ad_block_only_mode_page/ad_block_only_mode_alert.ts
+++ b/browser/resources/settings/ad_block_only_mode_page/ad_block_only_mode_alert.ts
@@ -1,0 +1,68 @@
+// Copyright (c) 2025 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import {PolymerElement} from 'chrome://resources/polymer/v3_0/polymer/polymer_bundled.min.js'
+import {PrefsMixin} from '/shared/settings/prefs/prefs_mixin.js'
+import {loadTimeData} from '../i18n_setup.js'
+
+import {getTemplate} from './ad_block_only_mode_alert.html.js'
+
+class BraveSettingsAdBlockOnlyModeAlert extends PrefsMixin(PolymerElement) {
+  static readonly AD_BLOCK_ONLY_MODE_ENABLED_PREF =
+      'brave.shields.adblock_only_mode_enabled'
+
+  static get is() {
+    return 'settings-ad-block-only-mode-alert'
+  }
+
+  static get template() {
+    return getTemplate()
+  }
+
+  static get properties () {
+    return {
+      isAdBlockOnlyModeEnabled_: {
+        type: Boolean,
+        value: false,
+      },
+    }
+  }
+
+  private declare isAdBlockOnlyModeEnabled_: boolean
+
+  override ready() {
+    super.ready();
+
+    chrome.settingsPrivate.getPref(
+      BraveSettingsAdBlockOnlyModeAlert.AD_BLOCK_ONLY_MODE_ENABLED_PREF)
+      .then((pref: chrome.settingsPrivate.PrefObject) =>
+        this.onAdBlockOnlyModeEnabledUpdated_(pref));
+
+    chrome.settingsPrivate.onPrefsChanged.addListener(
+      (prefs: chrome.settingsPrivate.PrefObject[]) => {
+        const pref = prefs.find(
+          p => p.key === BraveSettingsAdBlockOnlyModeAlert.AD_BLOCK_ONLY_MODE_ENABLED_PREF
+        )
+        if (pref) {
+          this.onAdBlockOnlyModeEnabledUpdated_(pref)
+        }
+      });
+  }
+
+  private onAdBlockOnlyModeEnabledUpdated_(pref: chrome.settingsPrivate.PrefObject): void {
+    this.isAdBlockOnlyModeEnabled_ =
+        pref.value &&
+        loadTimeData.getBoolean('isAdBlockOnlyModeFeatureEnabled')
+  }
+
+  onTurnOffAdBlockOnlyMode_() {
+    this.setPrefValue(
+        BraveSettingsAdBlockOnlyModeAlert.AD_BLOCK_ONLY_MODE_ENABLED_PREF,
+        false)
+  }
+}
+
+customElements.define(BraveSettingsAdBlockOnlyModeAlert.is,
+                      BraveSettingsAdBlockOnlyModeAlert);
--- a/browser/resources/settings/ad_block_only_mode_page/ad_block_only_mode_page.html
+++ b/browser/resources/settings/ad_block_only_mode_page/ad_block_only_mode_page.html
@@ -1,6 +1,6 @@
-<style include="settings-shared md-select">
+<style include="settings-shared">
   .description {
-    padding: 1em 0;
+    padding: var(--leo-spacing-m) 0;
   }
 
   .flex-center {
--- a/browser/resources/settings/ad_block_only_mode_page/ad_block_only_mode_page.html
+++ b/browser/resources/settings/ad_block_only_mode_page/ad_block_only_mode_page.html
@@ -1,0 +1,25 @@
+<style include="settings-shared md-select">
+  .description {
+    padding: 1em 0;
+  }
+
+  .flex-center {
+    align-items: center;
+    display: flex;
+  }
+</style>
+
+<settings-section page-title="$i18n{adBlockOnlyMode}">
+  <div class="settings-box first">
+    <div class="flex-center">
+      <div class="label secondary shields-lite-description">
+        <p>$i18nRaw{adBlockOnlyModeDesc}</p>
+      </div>
+    </div>
+  </div>
+  <settings-toggle-button
+      class="cr-row"
+      pref="{{prefs.brave.shields.adblock_only_mode_enabled}}"
+      label="$i18n{adBlockOnlyModeLabel}">
+  </settings-toggle-button>
+</settings-section>
--- a/browser/resources/settings/ad_block_only_mode_page/ad_block_only_mode_page.ts
+++ b/browser/resources/settings/ad_block_only_mode_page/ad_block_only_mode_page.ts
@@ -1,0 +1,22 @@
+// Copyright (c) 2025 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import {PolymerElement} from 'chrome://resources/polymer/v3_0/polymer/polymer_bundled.min.js'
+import {PrefsMixin} from '/shared/settings/prefs/prefs_mixin.js'
+
+import {getTemplate} from './ad_block_only_mode_page.html.js'
+
+class BraveSettingsAdBlockOnlyModePage extends PrefsMixin(PolymerElement) {
+  static get is() {
+    return 'settings-ad-block-only-mode-page'
+  }
+
+  static get template() {
+    return getTemplate()
+  }
+}
+
+customElements.define(BraveSettingsAdBlockOnlyModePage.is,
+                      BraveSettingsAdBlockOnlyModePage);
--- a/browser/resources/settings/default_brave_shields_page/default_brave_shields_page.ts
+++ b/browser/resources/settings/default_brave_shields_page/default_brave_shields_page.ts
@@ -24,6 +24,7 @@ import {
 
 import {getTemplate} from './default_brave_shields_page.html.js'
 
+import '../ad_block_only_mode_page/ad_block_only_mode_alert.js'
 import '../social_blocking_page/social_blocking_page.js';
 
 interface BraveShieldsPage {
--- a/browser/resources/settings/default_brave_shields_page/shields_page_index.ts
+++ b/browser/resources/settings/default_brave_shields_page/shields_page_index.ts
@@ -15,12 +15,14 @@ import { RouteObserverMixin } from '../router.js';
 import type { Route } from '../router.js';
 import type { SettingsPlugin } from '../settings_main/settings_plugin.js';
 import { SearchableViewContainerMixin } from '../settings_page/searchable_view_container_mixin.js';
+import { loadTimeData } from '../i18n_setup.js'
 
 import { getTemplate } from './shields_page_index.html.js';
 
 // Subpages
 import './default_brave_shields_page.js';
 import './brave_adblock_subpage.js';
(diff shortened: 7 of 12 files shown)


browser/resources/settings/default_brave_shields_page/shields_page_index.ts
```diff
@@ -51,8 +52,13 @@ export class SettingsShieldsPageIndexElement extends
   declare prefs: { [key: string]: any };
 
   private showDefaultViews_() {
-    this.$.viewManager.switchViews(
-      ['parent', 'socialBlocking'], 'no-animation', 'no-animation');
+    let views = ['parent']
```
fallaciousreasoning: nit: should still be `const` as the reference isn't modified

browser/resources/settings/brave_overrides/basic_page.ts
```diff
@@ -19,6 +19,7 @@ import '../brave_wallet_page/brave_wallet_page.js'
 import '../brave_web3_domains_page/brave_web3_domains_page.js'
 import '../default_brave_shields_page/default_brave_shields_page.js'
 import '../getting_started_page/getting_started.js'
+import '../ad_block_only_mode_page/ad_block_only_mode_page.js'
```
fallaciousreasoning: this import should be in `shields_page_index.ts` not here

fallaciousreasoning: (this file is being removed upstream)

browser/resources/settings/default_brave_shields_page/default_brave_shields_page.ts
```diff
@@ -24,6 +24,8 @@ import {
 
 import {getTemplate} from './default_brave_shields_page.html.js'
 
+import '../ad_block_only_mode_page/ad_block_only_mode_alert.js'
+import '../ad_block_only_mode_page/ad_block_only_mode_page.js'
```
fallaciousreasoning: this import should be in the index - its not used here

browser/resources/settings/brave_routes.ts
```diff
@@ -46,6 +46,10 @@ export default function addBraveRoutes(r: Partial<SettingsRoutes>) {
       console.error('[Settings] Routes: could not find SYSTEM page')
     }
   }
+  if (pageVisibility.adBlockOnlyMode) {
+    r.AD_BLOCK_ONLY_MODE =
+        r.BASIC.createSection('/adBlockOnlyMode', 'adBlockOnlyMode')
```
fallaciousreasoning: I think this should be
```suggestion
        r.SHIELDS.createChild('/adBlockOnlyMode')
```
as this page is controlled by the `shields_page_index` which is rendered at `SHIELDS` (even though its rendered at the root)

browser/resources/settings/brave_routes.ts
```diff
@@ -46,6 +46,10 @@ export default function addBraveRoutes(r: Partial<SettingsRoutes>) {
       console.error('[Settings] Routes: could not find SYSTEM page')
     }
   }
+  if (pageVisibility.adBlockOnlyMode) {
+    r.AD_BLOCK_ONLY_MODE =
```
fallaciousreasoning: Hey actually, do we actually need this route? It doesn't seem like its used?

browser/resources/settings/ad_block_only_mode_page/ad_block_only_mode_alert.html
```diff
@@ -0,0 +1,28 @@
...
+  }
+  .shields-options-disabled-alert-title {
+    font: var(--leo-font-heading-h4);
+  }
+  .shields-options-disabled-alert-content {
+    font: var(--leo-font-default-regular);
+  }
+  .shields-options-disabled-alert-button {
+    padding-top: var(--leo-spacing-xl);
+  }
```
fallaciousreasoning: optional: we don't need to be this specific with the class names - they're scoped to the shadow root which is just this file

browser/resources/settings/ad_block_only_mode_page/ad_block_only_mode_alert.html
```diff
@@ -0,0 +1,28 @@
+<link rel="stylesheet" href="chrome://resources/brave/leo/css/variables.css">
```
fallaciousreasoning: we shouldn't need to import this here - variables pierce the shadow root and they should be loaded at the document level

browser/resources/settings/ad_block_only_mode_page/ad_block_only_mode_alert.html
```diff
@@ -0,0 +1,28 @@
...
+  }
+</style>
+
+<template is="dom-if" if="[[isAdBlockOnlyModeEnabled_]]">
+  <div class="shields-options-disabled-container">
+    <leo-alert type="info" hideIcon="true" has-actions="false">
+      <div class="shields-options-disabled-alert-title">$i18n{adBlockOnlyModeAlertTitle}</div>
+      <div class="shields-options-disabled-alert-content">$i18n{adBlockOnlyModeAlertDesc}</div>
+      <leo-button on-click="onTurnOffAdBlockOnlyMode_" class="shields-options-disabled-alert-button">$i18n{adBlockOnlyModeAlertTurnOffButton}</leo-button>
+    </leo-alert>
```
fallaciousreasoning: hey, I think this shouldn't be setting `has-actions` to false, as it has actions - likely you just need to slot the button:


```suggestion
    <leo-alert type="info" hideIcon="true">
      <div class="shields-options-disabled-alert-title">$i18n{adBlockOnlyModeAlertTitle}</div>
      <div class="shields-options-disabled-alert-content">$i18n{adBlockOnlyModeAlertDesc}</div>
      <div slot="actions">
          <leo-button on-click="onTurnOffAdBlockOnlyMode_" class="shields-options-disabled-alert-button">$i18n{adBlockOnlyModeAlertTurnOffButton}</leo-button>
      </div>
    </leo-alert>
```

it might also work without the wrapping div (i.e. if you just set the slot directly on the `leo-button`)

browser/resources/settings/ad_block_only_mode_page/ad_block_only_mode_browser_proxy.html
fallaciousreasoning, line 1: I don't think this file should exist

browser/ui/webui/settings/default_brave_shields_handler.cc
```diff
@@ -80,6 +82,16 @@ void DefaultBraveShieldsHandler::RegisterMessages() {
...
+      "getAdBlockOnlyModeEnabled",
+      base::BindRepeating(
+          &DefaultBraveShieldsHandler::GetAdBlockOnlyModeEnabled,
+          base::Unretained(this)));
+  web_ui()->RegisterMessageCallback(
+      "setAdBlockOnlyModeEnabled",
+      base::BindRepeating(
+          &DefaultBraveShieldsHandler::SetAdBlockOnlyModeEnabled,
+          base::Unretained(this)));
   web_ui()->RegisterMessageCallback(
```
fallaciousreasoning: I think we shouldn't be adding new page handler methods here - instead we should introduce a mojom interface for shields and use that for the communication.

fallaciousreasoning: components/containers/core/mojom/containers.mojom is a good example of how to integrate into settings :smile:

browser/resources/settings/ad_block_only_mode_page/ad_block_only_mode_browser_proxy.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export class AdBlockOnlyModeBrowserProxyImpl
+implements AdBlockOnlyModeBrowserProxy {
+  getAdBlockOnlyModeEnabled () {
+    return sendWithPromise('getAdBlockOnlyModeEnabled')
+  }
+
+  setAdBlockOnlyModeEnabled (value: boolean) {
+    chrome.send('setAdBlockOnlyModeEnabled', [value])
+  }
```
fallaciousreasoning: I think we should probably use a mojom interface here sorry

browser/resources/settings/ad_block_only_mode_page/ad_block_only_mode_page.html
```diff
@@ -0,0 +1,30 @@
+<style include="settings-shared md-select">
+  .label.shields-lite-description {
```
fallaciousreasoning: nit: these styles are scoped to this file so no need for such specific classes
```suggestion
  .description {
```

browser/resources/settings/ad_block_only_mode_page/ad_block_only_mode_page.html
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  .flex-center {
+    align-items: center;
+    display: flex;
+  }
+</style>
+
+<settings-section
+  id="adBlockOnlyMode"
```
fallaciousreasoning: nit: unused id

browser/resources/settings/ad_block_only_mode_page/ad_block_only_mode_page.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  private browserProxy_: AdBlockOnlyModeBrowserProxy =
+    AdBlockOnlyModeBrowserProxyImpl.getInstance()
+
+  static get properties () {
+    return {
+      isAdBlockOnlyModeEnabled_: {
+        type: Object,
+        value: {
+          key: '',
```
fallaciousreasoning: hey so rather than adding a new Mojom service or special page handler functions could we just use the existing machinery for exposing prefs (like we do for almost all the other prefs)?

You can expose them from here:
brave/browser/extensions/api/settings_private/brave_prefs_util.cc


And you should be able to bind the toggle to it directly. For the Alert you can have a look at what the toggle does to change the pref and just do that :smile:

patches/chrome-browser-resources-settings-router.ts.patch
```diff
@@ -6,7 +6,7 @@ index ed94fc0a562c31c8a6dd11e15eb07b5b46343334..421e6047efdc76dda6177d4c0d19a3c8
...
++  GET_STARTED: Route; SHIELDS: Route; AD_BLOCK_ONLY_MODE: Route; SOCIAL_BLOCKING: Route; EXTENSIONS: Route; EXTENSIONS_V2: Route; BRAVE_SYNC: Route; BRAVE_SYNC_SETUP: Route; BRAVE_WALLET: Route; BRAVE_WEB3: Route; BRAVE_NEW_TAB: Route; THEMES: Route; SITE_SETTINGS_AUTOPLAY: Route; SITE_SETTINGS_GOOGLE_SIGN_IN: Route; SITE_SETTINGS_LOCALHOST_ACCESS: Route; SITE_SETTINGS_BRAVE_OPEN_AI_CHAT: Route; SITE_SETTINGS_ETHEREUM: Route; SITE_SETTINGS_SOLANA: Route; SITE_SETTINGS_CARDANO: Route; SITE_SETTINGS_SHIELDS_STATUS: Route; SHIELDS_ADBLOCK: Route; SHORTCUTS: Route; BRAVE_WALLET_NETWORKS: Route; BRAVE_LEO_ASSISTANT: Route; BRAVE_LEO_CUSTOMIZATION: Route; BRAVE_CONTENT: Route; BRAVE_SURVEY_PANELIST: Route; DEFAULT_SEARCH: Route; PRIVATE_SEARCH: Route; EMAIL_ALIASES: Route;
```
fallaciousreasoning: I suspect we don't need the new route (and can probably revert this)

browser/resources/settings/ad_block_only_mode_page/ad_block_only_mode_page.ts
aseren: hey, thanks for comment and explanation! Yeah, using machinery for prefs makes sense.
 Does this work for LocalState preferences?

fallaciousreasoning: oooooh maybe it doesn't actually. Sorry, missed it was a local state pref

browser/resources/settings/brave_routes.ts
aseren: Thanks for comment! yeah, this root is not really required by a spec, currently navigation to `brave://settings/adBlockOnlyMode` gives an empty page. So this root is not needed and does not really show anything. I will remove it.

browser/resources/settings/brave_routes.ts
aseren: Thanks for comment! Yeah, root is not really required by a spec, currently navigation to brave://settings/adBlockOnlyMode gives an empty page. I will remove it.

browser/resources/settings/ad_block_only_mode_page/ad_block_only_mode_page.ts
aseren: I will check

patches/chrome-browser-resources-settings-router.ts.patch
aseren: Yeah, root is not really required by a spec, currently navigation to brave://settings/adBlockOnlyMode gives an empty page. I will remove it.

browser/resources/settings/default_brave_shields_page/shields_page_index.ts
aseren: Fixed

browser/resources/settings/brave_overrides/basic_page.ts
aseren: Fixed

browser/resources/settings/default_brave_shields_page/default_brave_shields_page.ts
aseren: Fixed

browser/resources/settings/brave_routes.ts
aseren: @tmancey noticed that it can be possible needed. So I'm leaving it here for now, until we discuss it with @tmancey and @iambrianfung today on a call.

browser/resources/settings/brave_routes.ts
aseren: @tmancey noticed that it can be possibly needed. So I'm leaving it here for now, until we discuss it with @tmancey and @iambrianfung today on a call.

browser/resources/settings/ad_block_only_mode_page/ad_block_only_mode_alert.html
aseren: Fixed

browser/resources/settings/ad_block_only_mode_page/ad_block_only_mode_alert.html
aseren: removed

browser/resources/settings/ad_block_only_mode_page/ad_block_only_mode_browser_proxy.html
aseren: removed

browser/ui/webui/settings/default_brave_shields_handler.cc
aseren: Removed changes from default_brave_shields_handler.cc, as preferences are now handled from TypeScript part.

browser/resources/settings/ad_block_only_mode_page/ad_block_only_mode_browser_proxy.ts
aseren: Removed these changes, as preferences are now handled from TypeScript part.

browser/resources/settings/ad_block_only_mode_page/ad_block_only_mode_page.html
aseren: Removed

browser/resources/settings/ad_block_only_mode_page/ad_block_only_mode_page.ts
aseren: Local State is working fine. I changed code to handle preferences from TypeScript part.

patches/chrome-browser-resources-settings-router.ts.patch
aseren: I'm leaving it here for now as it can possibly needed. Will maybe remove it later after discussion

browser/resources/settings/brave_routes.ts
tmancey: Excellent, all working as expected lets remove old one.

aseren: The route is removed now.

browser/resources/settings/brave_routes.ts
aseren: The route is removed now.

patches/chrome-browser-resources-settings-router.ts.patch
aseren: The route is removed now.

app/brave_settings_strings.grdp
```diff
@@ -605,6 +605,26 @@
...
+    Ad Block Only mode retains core ad blocking rules of Brave Shields. Warning: removes all privacy protections. Please note that for individual websites broken by Brave's privacy protections, you can set Shields to DOWN. This is an experimental mode. <ph name="BEGIN_LINK">&lt;a target="_blank" href="$1"&gt;</ph>Learn more<ph name="END_LINK">&lt;/a&gt;</ph>
+  </message>
+  <message name="IDS_SETTINGS_AD_BLOCK_ONLY_MODE_ALERT_TITLE" desc="Title for the alert shown when ad block only mode is enabled">
+    Shields options are currently disabled.
+  </message>
+  <message name="IDS_SETTINGS_AD_BLOCK_ONLY_MODE_ALERT_DESC" desc="Description for the alert shown when ad block only mode is enabled">
+    Ad Block Only mode retains core ad blocking rules of Brave Shields and turns off privacy protections. If you would like to enable standard Shields, please disable this mode.
+  </message>
+  <message name="IDS_SETTINGS_AD_BLOCK_ONLY_MODE_ALERT_TURN_OFF_BUTTON" desc="Alert button text to turn off ad block only mode">
+    Turn off Ad Block Only mode
```
mkarolin, line 625: @rmcfadden3 should `Ad Block` be hyphenated here? Thanks.

browser/resources/settings/ad_block_only_mode_page/ad_block_only_mode_alert.ts
```diff
@@ -0,0 +1,68 @@
...
+          this.onAdBlockOnlyModeEnabledUpdated_(pref)
+        }
+      });
+  }
+
+  private onAdBlockOnlyModeEnabledUpdated_(pref: chrome.settingsPrivate.PrefObject): void {
+    this.isAdBlockOnlyModeEnabled_ =
+        pref.value &&
+        loadTimeData.getBoolean('isAdBlockOnlyModeFeatureEnabled')
+  }
```
fallaciousreasoning, line 58: happy to defer to you on this but should we just handle this in `ready()`? We could just early return if `isAdBlockOnlyModeFeatureEnabled` is false because you can't change the feature state without restarting the browser

browser/resources/settings/ad_block_only_mode_page/ad_block_only_mode_page.html
```diff
@@ -0,0 +1,25 @@
+<style include="settings-shared md-select">
```
fallaciousreasoning: nit: md-select styles are unused (we don't have a `<select>` on the page
```suggestion
<style include="settings-shared">
```

browser/resources/settings/ad_block_only_mode_page/ad_block_only_mode_page.html
```diff
@@ -0,0 +1,25 @@
...
+  .description {
+    padding: 1em 0;
```
fallaciousreasoning: Hey sorry, is this intentional? Normally we avoid `em` units in Brave in favor of pixel units. Possibly `var(--leo-spacing-m)` but I'm not sure.

browser/resources/settings/ad_block_only_mode_page/ad_block_only_mode_page.html
```diff
@@ -0,0 +1,25 @@
...
+  <div class="settings-box first">
+    <div class="flex-center">
+      <div class="label secondary shields-lite-description">
+        <p>$i18nRaw{adBlockOnlyModeDesc}</p>
+      </div>
+    </div>
+  </div>
+  <settings-toggle-button
+      class="cr-row"
+      pref="{{prefs.brave.shields.adblock_only_mode_enabled}}"
```
fallaciousreasoning, line 22: 👍

browser/resources/settings/ad_block_only_mode_page/ad_block_only_mode_alert.ts
fallaciousreasoning: Actually possibly dumb question, can we ever render this alert when the feature is disabled? If not maybe let's just get rid of this check of the feature flag?

fallaciousreasoning (approved):
lgtm % nits (happy for you to resolve comments and merge when you're happy)

browser/resources/settings/ad_block_only_mode_page/ad_block_only_mode_page.html
aseren: changed as `var(--leo-spacing-m)` fits good. Thanks

browser/resources/settings/ad_block_only_mode_page/ad_block_only_mode_alert.ts
aseren: Thanks for comment! yeah, if the feature is disabled then we are not showing alert. The feature check is added here in case if user enables preference, but then the feature is turned off for some reason - we still don't want to show Alert.
I also feel it more robust to set `isAdBlockOnlyModeEnabled_` basing on feature value, rather then early returning from ready().

app/brave_settings_strings.grdp
rmcfadden3: @mkarolin — I suppose there's an argument to be made for hyphenation, since this is a modifying phrase. But in this case I think it's fine as-is. `Ad Block Only mode` is a quasi-branded feature, and we never want to insert unnecessary punctuation in brand terms.

Long story short, nah, no hyphen.

PR #30835, opened 2025-08-26
Brave Account password endpoints.

szilardszaloki:
Resolves https://github.com/brave/brave-browser/issues/48485, resolves https://github.com/brave/brave-browser/issues/48486.

PR #31314, opened 2025-09-19
Brave Account registration flow

szilardszaloki:
Resolves https://github.com/brave/brave-browser/issues/47400, resolves https://github.com/brave/brave-browser/issues/48488.

### DIFF
--- a/components/brave_account/brave_account_service.cc
+++ b/components/brave_account/brave_account_service.cc
@@ -7,16 +7,98 @@
 
 #include <utility>
 
-#include "base/notimplemented.h"
+#include "base/base64.h"
+#include "base/check_deref.h"
+#include "base/functional/bind.h"
+#include "base/functional/callback.h"
+#include "base/strings/strcat.h"
+#include "brave/components/brave_account/endpoint_client/client.h"
+#include "brave/components/brave_account/pref_names.h"
+#include "components/os_crypt/sync/os_crypt.h"
 #include "components/prefs/pref_service.h"
+#include "net/http/http_request_headers.h"
+#include "net/http/http_status_code.h"
+#include "net/traffic_annotation/network_traffic_annotation.h"
 #include "services/network/public/cpp/shared_url_loader_factory.h"
 
+namespace {
+inline constexpr net::NetworkTrafficAnnotationTag kTrafficAnnotation =
+    net::DefineNetworkTrafficAnnotation("brave_account_endpoints",
+                                        R"(
+  semantics {
+    sender: "Brave Account client"
+    description:
+      "Implements the creation or sign-in process for a Brave Account."
+    trigger:
+      "User attempts to create or sign in to a Brave Account from settings."
+    user_data: {
+      type: EMAIL
+    }
+    data:
+      "Blinded cryptographic message for secure password setup "
+      "and account email address."
+      "Verification token for account activation and "
+      "serialized cryptographic record for account finalization."
+    destination: OTHER
+    destination_other: "Brave Account service"
+  }
+  policy {
+    cookies_allowed: NO
+    policy_exception_justification:
+      "These requests are essential for Brave Account creation and sign-in "
+      "and cannot be disabled by policy."
+  }
+)");
+
+std::string Encrypt(const std::string& plain_text) {
+  if (plain_text.empty()) {
+    return std::string();
+  }
+
+  std::string encrypted;
+  if (!OSCrypt::EncryptString(plain_text, &encrypted)) {
+    return std::string();
+  }
+
+  return base::Base64Encode(encrypted);
+}
+
+std::string Decrypt(const std::string& base64) {
+  if (base64.empty()) {
+    return std::string();
+  }
+
+  std::string encrypted;
+  if (!base::Base64Decode(base64, &encrypted)) {
+    return std::string();
+  }
+
+  std::string plain_text;
+  if (!OSCrypt::DecryptString(encrypted, &plain_text)) {
+    return std::string();
+  }
+
+  return plain_text;
+}
+}  // namespace
+
 namespace brave_account {
+
+using endpoint_client::Client;
+using endpoint_client::WithHeaders;
+using endpoints::PasswordFinalize;
+using endpoints::PasswordInit;
 
 BraveAccountService::BraveAccountService(
     PrefService* pref_service,
     scoped_refptr<network::SharedURLLoaderFactory> url_loader_factory)
-    : pref_service_(pref_service), url_loader_factory_(url_loader_factory) {}
+    : BraveAccountService(
+          pref_service,
+          std::make_unique<api_request_helper::APIRequestHelper>(
+              kTrafficAnnotation,
+              std::move(url_loader_factory)),
+          base::BindRepeating(&Encrypt),
+          base::BindRepeating(&Decrypt)) {}
 
 BraveAccountService::~BraveAccountService() = default;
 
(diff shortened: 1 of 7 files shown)


components/brave_account/brave_account_service_test.h
```diff
@@ -0,0 +1,108 @@
...
+ protected:
+  void SetUp() override {
+    prefs::RegisterPrefs(pref_service_.registry());
+    auto mock_api_request_helper =
+        std::make_unique<api_request_helper::MockAPIRequestHelper>(
+            TRAFFIC_ANNOTATION_FOR_TESTS,
+            test_url_loader_factory_.GetSafeWeakWrapper());
+    mock_api_request_helper_ = mock_api_request_helper.get();
+    auto crypt_callback = base::BindRepeating(&BraveAccountServiceTest::Crypt,
+                                              base::Unretained(this));
```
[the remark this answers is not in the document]
szilardszaloki: False positive — the test fixture is guaranteed to outlive `brave_account_service_` (and any callbacks bound to it), so using `base::Unretained<>()` here is safe.

components/brave_account/brave_account_service.cc
```diff
@@ -25,24 +107,167 @@ void BraveAccountService::BindInterface(
...
 void BraveAccountService::RegisterInitialize(
     const std::string& email,
     const std::string& blinded_message,
-    mojom::Authentication::RegisterInitializeCallback callback) {
-  // TODO(https://github.com/brave/brave-browser/issues/47400)
-  // TODO(https://github.com/brave/brave-browser/issues/48488)
-  NOTIMPLEMENTED();
-  std::move(callback).Run(mojom::RegisterInitializeResult::New());
+    RegisterInitializeCallback callback) {
+  DCHECK(!email.empty());
```
bridiver: this doesn't seem like an appropriate use of DCHECK https://chromium.googlesource.com/chromium/src/+/HEAD/styleguide/c++/checks.md

components/brave_account/brave_account_service.cc
```diff
@@ -25,24 +107,167 @@ void BraveAccountService::BindInterface(
...
 void BraveAccountService::RegisterFinalize(
     const std::string& encrypted_verification_token,
     const std::string& serialized_record,
-    mojom::Authentication::RegisterFinalizeCallback callback) {
-  // TODO(https://github.com/brave/brave-browser/issues/47400)
-  // TODO(https://github.com/brave/brave-browser/issues/48488)
-  NOTIMPLEMENTED();
-  std::move(callback).Run(mojom::RegisterFinalizeResult::New());
+    RegisterFinalizeCallback callback) {
+  DCHECK(!encrypted_verification_token.empty());
```
bridiver: same as above re: DCHECK

bsclifton:
Nice work here 😄 

I wanted to share some reviewer notes - these are from me digging in and taking the time to understand the code. 
There are basically two concerns here:

1. `os_crypt` dependency was added to `DEPS` file for `components/brave_accounts`. The reason for adding this was so that we can encrypt the verification token sent back by the Accounts system. This is what triggered the @brave/deps-reviewers group which I am not a part of.

2. `BraveAccountService::RegisterFinalize` method is now added to `components/brave_account/brave_account_service.h`.
I was looking at the code and I had some questions initially. @szilardszaloki helped me resolve these via DM and these findings were critical in me understanding the how things work. I wanted to share them:
    - When `Client<PasswordInit>::Send` and `Client<PasswordFinalize>::Send` methods are called, the `components/brave_account/endpoint_client/client.h` abstraction makes the outbound call. It's important to look at https://github.com/brave/brave-core/blob/a333426313b03cd6bed3ef969f43b5c8ecd3d312/components/brave_account/endpoint_client/client.h#L64-L82 to see where the reply is handled. A 2xx triggers a `base::expected` response and others trigger a `base::unexpected` response.
    - The responses themselves (success or error) are objects Chromium is generating via IDL and the JSON schema compiler. These responses [were added in a prior PR](https://github.com/brave/brave-core/pull/30835) (which is why they're not included here) and consist of [PasswordInitResponse](https://github.com/brave/brave-core/blob/master/components/brave_account/endpoints/password_init_response.idl) and [PasswordFinalizeResponse](https://github.com/brave/brave-core/blob/master/components/brave_account/endpoints/password_finalize_response.idl).
    - The code via these lines:
    https://github.com/brave/brave-core/blob/a333426313b03cd6bed3ef969f43b5c8ecd3d312/components/brave_account/endpoint_client/client.h#L69-L70
    will use the IDL generated code (`FromValue`) to deserialize the response object. If the schema doesn't match, the generated code will set the response as `std::nullopt` (basically a `std::optional` with `has_value` as false).  Chromium's JSON schema compiler is generating this code based on the IDL.  This [nuance (where it can be `std::nullopt`) is captured in the docs](https://github.com/brave/brave-core/tree/master/components/brave_account/endpoint_client#usage).  However, it's important to note that this shouldn't realistically ever happen. Only if the server sends fields client isn't expecting (ex: we messed up an endpoint upgrade). 
    - `base::expected` and `base::unexpected` were completely new to me. I looked at C++ 23's [std::expected](https://en.cppreference.com/w/cpp/utility/expected.html) for an example how how things work and it seems comparable. The reply type (having signature `base::expected<std::optional<PasswordInit::Response>, std::optional<PasswordInit::Error>>`) is basically an object which could have a success or a failure. The caller doesn't know until they use a method to resolve this. A person could call `has_value()` to determine if it's the one we expect OR if it's the error one.
    - The syntax like `transform_error` is a no-op if the reply type is a success. That code only executes on failure.  And if you look at it, you can see it will test to see if the object is `std::nullopt`. That is how the caller would catch a situation where JSON schema compiler was not able to deserialize the code.

components/brave_account/brave_account_service.cc
```diff
@@ -25,24 +107,159 @@ void BraveAccountService::BindInterface(
...
+      CHECK_DEREF(api_request_helper_.get()), std::move(request),
+      base::BindOnce(&BraveAccountService::OnRegisterFinalize,
+                     weak_factory_.GetWeakPtr(), std::move(callback),
+                     encrypted_verification_token));
+}
+
+void BraveAccountService::OnRegisterInitialize(
+    RegisterInitializeCallback callback,
+    int response_code,
+    base::expected<std::optional<PasswordInit::Response>,
```
bridiver, line 173: as discussed I have some concerns around `expected<optional, optional>`, but we'll defer that for now because it would require other refactoring anyway that we don't want in this PR.

components/brave_account/brave_account_service_unittest.cc
```diff
@@ -0,0 +1,409 @@
...
+    .mojo_expected = mojom::RegisterFinalizeResult::New(),
+};
+
+using RegisterFinalizeTest = BraveAccountServiceTest<RegisterFinalizeTestCase>;
+
+TEST_P(RegisterFinalizeTest, MapsEndpointExpectedToMojoExpected) {
+  RunTestCase();
+
+  const auto& test_case = CHECK_DEREF(this->GetParam());
+  EXPECT_EQ(!pref_service_.GetString(prefs::kVerificationToken).empty(),
```
bridiver: why is this only testing for non-empty value vs the actual value?

components/brave_account/brave_account_service.cc
```diff
@@ -7,16 +7,98 @@
...
+
+  std::string encrypted;
+  if (!OSCrypt::EncryptString(plain_text, &encrypted)) {
+    return std::string();
+  }
+
+  return base::Base64Encode(encrypted);
+}
+
+std::string Decrypt(const std::string& base64) {
```
bridiver: seems like something should be testing these cases including base64 encoding? Maybe instead of mocking the entire Decrypt/Encrypt you should just mock `OSCrypt` and pass that to Encrypt/Decrypt?

szilardszaloki: Upstream doesn't seem too concerned about testing their `OSCrypt` + `base64` logic — e.g. for [`HashPasswordManager`](https://source.chromium.org/chromium/chromium/src/+/main:components/password_manager/core/browser/hash_password_manager.cc;l=55-62), the unit test basically just reimplements `EncryptString()` (see [here](https://source.chromium.org/chromium/chromium/src/+/main:components/password_manager/core/browser/hash_password_manager_unittest.cc;l=30-34)).

szilardszaloki: Even the seemingly cleanest workaround isn't ideal — using `OSCryptMocker` and exposing `Encrypt()` as a `private` `static` in `BraveAccountService`. That works for most test cases, but fails when you need to encrypt the verification token during test-case definition, since `OSCryptMocker::SetUp()` hasn't run yet.

szilardszaloki: Given [this](https://github.com/brave/brave-core/pull/31314#discussion_r2373461550), and that mocking `Encrypt()`/`Decrypt()` at least makes it possible to test how we handle crypto failures, I'd leave it as-is — WDYT?

components/brave_account/brave_account_service.cc
szilardszaloki: https://github.com/brave/brave-core/pull/31314/commits/f1885b37e56d7188c0493a6ebeb20d995b6a3dee

components/brave_account/brave_account_service.cc
szilardszaloki: https://github.com/brave/brave-core/pull/31314/commits/f1885b37e56d7188c0493a6ebeb20d995b6a3dee

components/brave_account/brave_account_service_unittest.cc
szilardszaloki: https://github.com/brave/brave-core/pull/31314/commits/f1885b37e56d7188c0493a6ebeb20d995b6a3dee

components/brave_account/brave_account_service.cc
bridiver: That's not what what I'm suggesting. What I'm saying is that instead of passing in a callback to replace Encrypt/Decrypt, you'd pass one in to replace OSCrypt::Encrypt/Decrypt and then pass those callbacks to Encrypt/Decrypt. It's almost exactly the same thing you're doing now

bridiver: so `std::string Decrypt(const std::string& base64, OSDecryptCallback callback) {`

szilardszaloki: Sure thing: https://github.com/brave/brave-core/pull/31314/commits/7aba1c2299cfed16037fbfd9fbbc8906e34c7967

components/brave_account/brave_account_service_test.h
```diff
@@ -0,0 +1,109 @@
...
+ protected:
+  void SetUp() override {
+    prefs::RegisterPrefs(pref_service_.registry());
+    auto mock_api_request_helper =
+        std::make_unique<api_request_helper::MockAPIRequestHelper>(
+            TRAFFIC_ANNOTATION_FOR_TESTS,
+            test_url_loader_factory_.GetSafeWeakWrapper());
+    mock_api_request_helper_ = mock_api_request_helper.get();
+    auto os_crypt_callback = base::BindRepeating(
+        &BraveAccountServiceTest::OSCrypt, base::Unretained(this));
```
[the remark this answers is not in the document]
szilardszaloki: Same as https://github.com/brave/brave-core/pull/31314#discussion_r2364577610.

PR #31317, opened 2025-09-20
Added rust ciborium

supermassive:
Resolves https://github.com/brave/brave-browser/issues/49542
Sec review: https://github.com/brave/reviews/issues/2059

```
--- a/components/brave_wallet/common/Cargo.toml
+++ b/components/brave_wallet/common/Cargo.toml
@@ -12,6 +12,7 @@ curve25519-dalek = "4.1.3"
 bech32 = "0.9.1"
 bls-signatures = { version = "0.15", default-features = false, features = ["pairing"] }
 schnorrkel = "0.11.4"
+ciborium = "0.2.2"
 
 [lib]
 name = "brave_wallet"
```

`vpython3 ./tools/crates/run_gnrt.py vendor`
`vpython3 ./tools/crates/run_gnrt.py gen`


third_party/rust/crunchy/v0_2/BUILD.gn
```diff
@@ -24,7 +24,11 @@ cargo_crate("lib") {
       "Crunchy unroller: deterministically unroll constant loops"
   cargo_pkg_version = "0.2.2"
 
-  features = [ "std" ]
+  features = [
+    "default",
+    "limit_128",
```
bridiver, line 29: it's not clear to me where this is coming from? cc @cdesouza-chromium

bridiver: I think this is maybe a bug in gnrt https://github.com/brave/brave-core/pull/31317/files#diff-473365ea8510a7ea1831004b69bbf7a25c6864ff9730c09f6f3f22ba1b7b040eR132

supermassive: @bridiver @cdesouza-chromium 
is it blocking or not?

cdesouza-chromium: I'm not sure if I can infer either why these features are being added.

bridiver: we should either fix gnrt to add this target or patch it out of Cargo.toml

image

bridiver: or remove in gnrt_config.toml actually is better as I suggested in dm

bridiver (approved):

PR #31329, opened 2025-09-21
[Android] Update light/dark colours from Nala XML definitions

Olipro:
This change has been produced using a Ruby script that implements the following logic:

1. For each <color> tag in the destination, try to find one in the source with the same name
2. If found, update the destination tag with the colour from the source tag
3. If not found, do nothing and move onto the next tag.
4. Tags that exist in the source but not the destination are ignored.

Currently, some tags exist in the destination XML that are missing from the source and also vice versa.

Once this has been resolved, we should be able to delete the files in this diff and move to simply copying the generated XML files straight out of brave/leo.

Closes brave/brave-browser#49143

### DIFF
--- a/android/java/res/values-night/brave_colors.xml
+++ b/android/java/res/values-night/brave_colors.xml
@@ -1,10 +1,9 @@
-<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="UTF-8"?>
 <!-- Copyright (c) 2019 The Brave Authors. All rights reserved.
      This Source Code Form is subject to the terms of the Mozilla Public
      License, v. 2.0. If a copy of the MPL was not distributed with this file,
      You can obtain one at https://mozilla.org/MPL/2.0/.
 -->
-
 <resources>
     <!-- Theme colors -->
     <!-- Overriding chromium theme colors to brave theme colors -->
@@ -50,10 +49,10 @@
 
     <!-- Brave wallet 2.0 colors -->
     <color name="brave_wallet_dn_panel_bg">@color/wallet_bg</color>
-    <color name="text_disabled">#80EBEEF0</color>
+    <color name="text_disabled">#ebeef0</color>
     <color name="text_on_primary">#251491</color>
-    <color name="icon_interactive">#7C91FF</color>
-    <color name="divider_subtle">#66687B85</color>
+    <color name="icon_interactive">#bcc6f3</color>
+    <color name="divider_subtle">#303032</color>
     <color name="system_feedback_error">#FFB1B0</color>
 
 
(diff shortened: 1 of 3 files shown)


fallaciousreasoning (approved):
this lgtm, thanks heaps @Olipro

Olipro:
I've generated the two files below that give you a complete list of `<color>` tag names that are missing in either direction. If we can make changes to Nala so that missing definitions are populated (at least, definitions that are in our Android XML but not in Nala) then we should be able to transition to using Nala's XML verbatim.

@fallaciousreasoning PTAL :)

[dark_color_discrepancies.txt](https://github.com/user-attachments/files/22468136/dark_color_discrepancies.txt)
[light_color_discrepancies.txt](https://github.com/user-attachments/files/22468138/light_color_discrepancies.txt)

fallaciousreasoning:
@Olipro so had a quick glance at those - I think we don't want to verbatim add most of them to Nala because they're already there (just with slightly different names).

I'd prefer to add all the Nala variables to the Android build and then work on gradually migrating these over :smile:

Olipro:
@fallaciousreasoning Alright, sounds like a plan. I take it you're fine with us merging this PR for now?

samartnik:
change itself looks good, as long as UI elements that used those colors look as expected, but where is the script itself?

samartnik:
accidentally closed it, sorry, I didn't mean it, only comment

Olipro:
> LGTM @Olipro could you please attach the screenshots of the updated app?

I can - could you let me know which screens you'd like captured?

AlexeyBarabash:
> I can - could you let me know which screens you'd like captured?

yes, it would be good to see, dark/light schemes:
- first time run dialog;
- Brave Sync screen;
- new tab page for regular profile;
- new tab page for private profile;
- Brave Wallet onboarding screen.

AlexeyBarabash:
... Just to ensure there are no obvious problems

Olipro:
Screenshot_1758729471
Screenshot_1758729485
Screenshot_1758729491
Screenshot_1758729527
Screenshot_1758729540
Screenshot_1758728866
Screenshot_1758728901
Screenshot_1758728911
Screenshot_1758729343
Screenshot_1758729363

Looks reasonable to me - @fallaciousreasoning @AlexeyBarabash let me know if anything's not as it should be.

Olipro:
@aguscruiz - It was mentioned that you might be interested in this work, you have been duly pinged! 😄

AlexeyBarabash:
Button on Sync Dark theme looks broken to me.

Olipro:
> Button on Sync Dark theme looks broken to me.

What does "broken" mean in this context? What should we do to fix it?

aguscruiz:
I think it refers to the color on the button on dark mode being white on a light purple. 

image

The button on dark mode should have a dark purple text on light purple background, like this

image

Olipro:
OK, presumably this is defined by the tag `button_background`

On the Nala side, this is defined as `<color name="button_background">#bcc6f3</color>` which is equal to the colour in the screenshot.

Thus, that should be why this is the colour it's using right now.

aguscruiz:
The background color is the correct color, what's not right is the text+icon colors.

Olipro:
> The background color is the correct color, what's not right is the text+icon colors.

Hmm, then this appears to be orthogonal to the change made in this commit - The text colour appears to be `#ffffff` but there aren't any lines in the diff for the dark scheme XML that set a color tag to white. PTAL to confirm though and let me know how we should proceed.

fallaciousreasoning:
Discussed offline and the text/icon color should be `schemes_on_primary`.

AlexeyBarabash:
indeed, the change
```
--- a/android/java/res/values/brave_styles.xml
+++ b/android/java/res/values/brave_styles.xml
@@ -163,7 +163,7 @@
 
     <!-- Buttons for Brave Sync -->
     <style name="BraveSyncButton">
-        <item name="android:textColor">@android:color/white</item>
+        <item name="android:textColor">@color/schemes_on_primary</item>
         <item name="android:background">@drawable/brave_sync_btn</item>
         <item name="android:layout_marginStart">15dp</item>
         <item name="android:layout_marginEnd">15dp</item>
```
fixes the button:

image


@Olipro could you please apply the change to the PR?

PR #31330, opened 2025-09-22
Remove NoDestructor usage against DataDecoder in CustomBackgroundFileManager

sangwoo108:
The NoDestructor is designed to be alive until process exit, which is not necessary for the DataDecoder instance here. We can just use a unique_ptr and bind it to the callback to keep it alive until the decoding is done. This will address concerns about memory concerns.

Resolves https://github.com/brave/brave-browser/issues/49468


browser/ntp_background/custom_background_file_manager.cc
```diff
@@ -176,9 +167,26 @@ void CustomBackgroundFileManager::DecodeImageInIsolatedProcess(
   if (!image_decoder_)
     image_decoder_ = std::make_unique<ImageDecoderImpl>();
 
+  std::unique_ptr<data_decoder::DataDecoder> data_decoder =
```
atuchin-m: Can we use `DecodeImageIsolated` instead of making `DataDecoder` instance each time?

atuchin-m: https://source.chromium.org/chromium/chromium/src/+/main:services/data_decoder/public/cpp/decode_image.h;l=45

sangwoo108: I found that we're using ImageDecoderImpl defined in chrome layer, but the reference is using the one in service layer. I think we should keep this version so that we can avoid unwanted regression by replacing implementation. Additionally, I'm not sure what I should fill rest of parameters like `shrink_to_fit`, `max_size_in_bytes` and etc.

atuchin-m: ~~Wow, if it's this [DataDecoderImpl](https://source.chromium.org/chromium/chromium/src/+/main:services/data_decoder/image_decoder_impl.h;l=15), we can't use it at all, it's a layer violation. We can't use non-public `/services/<service_name>/`~~

Internally ImageDecoderImpl uses the same `that data_decoder::DecodeImage[Isolated]`:
https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/image_decoder/image_decoder.cc;l=61-65
The rest parameters (`shrink_to_fit`, `max_size_in_bytes`)  are in the default state.
Why just don't use it directly?

sangwoo108: Okay. Switched to using the function directly.

browser/ntp_background/custom_background_file_manager.cc
```diff
@@ -17,21 +17,14 @@
 #include "brave/browser/ntp_background/constants.h"
 #include "chrome/browser/image_fetcher/image_decoder_impl.h"
```
atuchin-m: please clean up unused includes

PR #31331, opened 2025-09-22
[AI Chat]: Fix renderer crash

fallaciousreasoning:
Resolves https://github.com/brave/brave-browser/issues/49493

My theory about why this was happening is that because `kWantResult` is set the promise is resolving to `undefined` when the navigation happens. We aren't checking the optional so we do a bad dereference there.

We don't really have steps to repro this, so this is speculative based on the stacktrace

### DIFF
--- a/components/ai_chat/renderer/page_text_distilling.cc
+++ b/components/ai_chat/renderer/page_text_distilling.cc
@@ -202,7 +202,7 @@ void DistillPageText(
         [](base::OnceCallback<void(const std::optional<std::string>&)> callback,
            std::optional<base::Value> value, base::TimeTicks start_time) {
           if (value && value->is_string()) {
-            std::move(callback).Run(value->GetString());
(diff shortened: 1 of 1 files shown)


components/ai_chat/renderer/page_text_distilling.cc
```diff
@@ -201,7 +201,7 @@ void DistillPageText(
     auto on_script_executed =
         [](base::OnceCallback<void(const std::optional<std::string>&)> callback,
            std::optional<base::Value> value, base::TimeTicks start_time) {
-          if (value->is_string()) {
+          if (value && value->is_string()) {
             std::move(callback).Run(value->GetString());
```
goodov: The string can be quite big, and we can move it here. Let's just do this:
`std::move(*value)->TakeString()`

goodov (approved):

PR #31335, opened 2025-09-22
Enable kBraveVerticalTabHideCompletely flag by default

sangwoo108:
As of this commit, the kBraveVerticalTabHideCompletely feature flag is enabled by default. Even with this flag is enabled, users can still toggle the feature via settings page.

Resolves https://github.com/brave/brave-browser/issues/49494


browser/ui/tabs/features.cc
```diff
@@ -5,6 +5,7 @@
 
 #include "brave/browser/ui/tabs/features.h"
 
+#include "base/feature_list.h"
```
simonhong: nit: it's necessary? features.h already have it.

sangwoo108: Oops, no clue why it was added.

PR #31344, opened 2025-09-22
Ad block only mode support for adblock updater

aseren:
The PR adds unloading of ad-block filter lists other then "default" when ad-block only mode is enabled.

The functionality is under a feature which is turned off by default.

Epic for this task is: https://github.com/brave/brave-browser/issues/48521
Spec: https://docs.google.com/document/d/1-y3NIRvt0scNK7qJ-XUuY0wRcnFuT08xsxivbUtr9SE/edit?tab=t.0

Resolves https://github.com/brave/brave-browser/issues/48561

### DIFF
--- a/components/brave_shields/core/common/brave_shield_constants.h
+++ b/components/brave_shields/core/common/brave_shield_constants.h
@@ -142,6 +142,10 @@ inline constexpr char kCookieListEnabledHistogram[] =
 inline constexpr char kCookieListPromptHistogram[] =
     "Brave.Shields.CookieListPrompt";
 
+// The list of UUIDs of filter lists that will be loaded by AdBlockOnlyMode.
+inline constexpr auto kAdblockOnlyModeFilerListUUIDs =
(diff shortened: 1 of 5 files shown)


components/brave_shields/core/common/brave_shield_constants.h
```diff
@@ -142,6 +142,10 @@ inline constexpr char kCookieListEnabledHistogram[] =
 inline constexpr char kCookieListPromptHistogram[] =
     "Brave.Shields.CookieListPrompt";
 
+// The list of UUIDs of filter lists that will be loaded by AdBlockOnlyMode.
+inline constexpr auto kAdblockOnlyModeFilerListUUIDs =
+    base::MakeFixedFlatSet<std::string_view>({"default"});
+
```
antonok-edm, line 148: Just for extra context:

We are planning to remove the UUIDs and migrate to using component IDs instead. The UUID kind of fits better with the existing code in this context, so no changes necessary, but may be good to keep in mind

PR #31346, opened 2025-09-22
[PSST] Add permission support (content settings part)

vadimstruts:
Resolves https://github.com/brave/brave-browser/issues/49309

- added content settings changes in context of permissions


chromium_src/components/content_settings/core/browser/content_settings_registry.cc
```diff
@@ -393,6 +394,16 @@ void ContentSettingsRegistry::BraveInit() {
           WebsiteSettingsRegistry::PLATFORM_ANDROID |
           WebsiteSettingsRegistry::PLATFORM_IOS,
       WebsiteSettingsInfo::INHERIT_IN_INCOGNITO);
+
+#if BUILDFLAG(ENABLE_PSST)
+  website_settings_registry_->Register(
+      ContentSettingsType::BRAVE_PSST, brave_shields::kBravePsst, base::Value(),
+      WebsiteSettingsInfo::UNSYNCABLE, WebsiteSettingsInfo::NOT_LOSSY,
+      WebsiteSettingsInfo::REQUESTING_ORIGIN_ONLY_SCOPE,
```
bridiver: Why is this still per-frame instead of top origin? Are there cases where we would be doing this in sub frames that are not the same as the top origin?

chromium_src/components/content_settings/core/browser/content_settings_registry.cc
```diff
@@ -393,6 +394,16 @@ void ContentSettingsRegistry::BraveInit() {
           WebsiteSettingsRegistry::PLATFORM_ANDROID |
           WebsiteSettingsRegistry::PLATFORM_IOS,
       WebsiteSettingsInfo::INHERIT_IN_INCOGNITO);
+
+#if BUILDFLAG(ENABLE_PSST)
```
bridiver, line 406: there's also a feature flag, right?

components/brave_shields/core/common/brave_shield_constants.h
```diff
@@ -31,6 +31,7 @@ inline constexpr char kCookies[] = "shieldsCookiesV3";
 inline constexpr char kFacebookEmbeds[] = "fb-embeds";
 inline constexpr char kTwitterEmbeds[] = "twitter-embeds";
 inline constexpr char kLinkedInEmbeds[] = "linked-in-embeds";
+inline constexpr char kBravePsst[] = "bravePsst";
```
bridiver: this isn't part of shields so doesn't seem like this belongs here?

components/psst/common/BUILD.gn
```diff
@@ -13,11 +13,15 @@ mojom("mojom") {
 }
 
 static_library("common") {
+  public = [
+    "brave_psst_constants.h",
```
bridiver: just constants.h
also public isn't really needed here because nothing is non-public

PR #31348, opened 2025-09-22
[PSST] Add website settings support

vadimstruts:
Resolves https://github.com/brave/brave-browser/issues/49309

Added saving of the PSST settings with using of the website settings.  We should save it when user accepts the consent dialog.

### DIFF
--- a/browser/psst/psst_ui_delegate_impl.cc
+++ b/browser/psst/psst_ui_delegate_impl.cc
@@ -50,7 +50,7 @@ void PsstUiDelegateImpl::OnUserAcceptedPsstSettings(
   // Save the PSST settings when user accepts the dialog
   SetPsstWebsiteSettings(host_content_settings_map_, origin,
                          ConsentStatus::kAllow, dialog_data_->script_version,
-                         dialog_data_->user_id, std::move(urls_to_skip));
+                         dialog_data_->user_id, urls_to_skip.Clone());
 
   if (apply_changes_callback_) {
     std::move(apply_changes_callback_).Run(std::move(urls_to_skip));


browser/psst/psst_ui_delegate_impl.cc
```diff
@@ -5,16 +5,75 @@
...
     const std::vector<PolicyTask>& applied_tasks,
     const mojom::PsstStatus status) {
   // Implementation for setting the current progress.
 }
 
+std::optional<PsstPermissionInfo> PsstUiDelegateImpl::GetPsstPermissionInfo(
+    const url::Origin& origin,
+    const std::string& user_id) {
+  auto* psst_permission_context =
+      BravePsstPermissionContextFactory::GetForProfile(profile_);
```
bridiver: use dependency injection

browser/psst/psst_ui_delegate_impl.cc
```diff
@@ -5,16 +5,75 @@
 
 #include "brave/browser/psst/psst_ui_delegate_impl.h"
 
+#include "brave/browser/psst/brave_psst_permission_context_factory.h"
+#include "chrome/browser/profiles/profile.h"
+#include "content/public/browser/web_contents.h"
+
+namespace {
+std::vector<std::string> ListToVector(const base::Value::List list) {
```
bridiver: what is the point of this? Why not just base the Value::List to the callback?

components/psst/browser/core/psst_consent_data.h
```diff
@@ -0,0 +1,43 @@
...
+  PsstConsentData& operator=(PsstConsentData&&) noexcept;
+
+  PsstConsentData(const PsstConsentData&) = delete;
+  PsstConsentData& operator=(const PsstConsentData&) = delete;
+
+  std::string user_id;
+  std::string site_name;
+  base::Value::List request_infos;
+  int script_version;
+  ConsentCallback apply_changes_callback;
```
bridiver: why is the callback part of the consent data?

browser/psst/psst_ui_delegate_impl.cc
```diff
@@ -5,16 +5,75 @@
...
-PsstUiDelegateImpl::PsstUiDelegateImpl() = default;
+PsstUiDelegateImpl::PsstUiDelegateImpl(Profile* profile,
+                                       content::WebContents* contents)
+    : web_contents_(contents), profile_(profile) {
+  CHECK(profile);
+}
+
 PsstUiDelegateImpl::~PsstUiDelegateImpl() = default;
 
+void PsstUiDelegateImpl::Show(PsstConsentData dialog_data) {
```
bridiver: why does this get passed into Show?

browser/psst/psst_ui_delegate_impl.h
```diff
@@ -6,24 +6,46 @@
...
 class PsstUiDelegateImpl : public PsstTabWebContentsObserver::PsstUiDelegate {
  public:
-  PsstUiDelegateImpl();
+  PsstUiDelegateImpl(Profile* profile, content::WebContents* contents);
   ~PsstUiDelegateImpl() override;
 
   PsstUiDelegateImpl(const PsstUiDelegateImpl&) = delete;
   PsstUiDelegateImpl& operator=(const PsstUiDelegateImpl&) = delete;
 
+  void Show(PsstConsentData dialog_data) override;
```
bridiver: per above please add comments https://github.com/brave/brave-core/pull/31348/files#r2409059572

components/psst/browser/core/psst_consent_data.h
vadimstruts: to get the user response from the consent dialog

browser/psst/psst_ui_delegate_impl.cc
vadimstruts: PsstConsentData contains the data required to display the consent dialog to the user

bridiver: actually shouldn't this be implemented as a PermissionRequest similar to WidevinePermissionRequest? cc @goodov

goodov: > actually shouldn't this be implemented as a PermissionRequest similar to WidevinePermissionRequest? cc [@goodov](https://github.com/goodov)

don't think `PermissionRequest` is applicable here as it handles basic values such as ask/reject. It's not applicable for object-based permissions.

browser/psst/psst_ui_delegate_impl.cc
```diff
@@ -5,16 +5,53 @@
...
+    const url::Origin& origin,
+    const std::string& user_id) {
+  return permission_context_->GetPsstPermissionInfo(origin, user_id);
+}
+
+void PsstUiDelegateImpl::OnUserAcceptedPsstSettings(
+    base::Value::List urls_to_skip) {
+  // Create the PSST permission when user accepts the dialog
+  permission_context_->GrantPermission(
+      url::Origin::Create(web_contents_->GetLastCommittedURL()),
```
goodov: this is wrong. web contents can be navigated to another URL while the dialog is opened.

vadimstruts: changed to use the origin instead (we pass the origin at the moment of showing the dialog)

browser/psst/psst_ui_delegate_impl.cc
bridiver: @goodov it's still basically ask/reject, there is just additional metadata associated with it. Doing this outside of that framework would mean re-implementing a bunch of logic around active tab/navigation, unless we're making this modal? We would have a subclass here, but even the base PermissionRequest object has some additional permission metadata like GetRequestedAudioCaptureDeviceIds. Is there a specific place where you think this runs into a problem? SmartCardPermissionRequest has an additional field `reader_name` and `RegisterProtocolHandlerPermissionRequest` also has `protocol` so this seems like it make sense here as well?

browser/psst/psst_ui_delegate_impl.cc
bridiver: this is part of my point here https://github.com/brave/brave-core/pull/31348#discussion_r2411819240 that there are a bunch of edge cases to handle for permission dialogs

bridiver: some of the other related events that PermissionRequestManager handles
```
void OnTabActiveStatusChanged(bool is_active,
                                tabs::TabInterface* tab_interface);
  void OnTabVisibleStatusChanged(bool is_visible,
                                 tabs::TabInterface* tab_interface);
  void OnTabDetached(tabs::TabInterface* tab_interface,
                     tabs::TabInterface::DetachReason reason);
  void OnTabAttached(tabs::TabInterface* tab_interface);
  void OnTabActiveChanged();
```

browser/ui/tabs/brave_tab_features.cc
```diff
@@ -47,11 +49,15 @@ void BraveTabFeatures::Init(TabInterface& tab, Profile* profile) {
...
-          std::make_unique<psst::PsstUiDelegateImpl>(), profile->GetPrefs(),
-          ISOLATED_WORLD_ID_BRAVE_INTERNAL);
+  if (base::FeatureList::IsEnabled(psst::features::kEnablePsst)) {
+    psst_web_contents_observer_ =
+        psst::PsstTabWebContentsObserver::MaybeCreateForWebContents(
+            tab.GetContents(), profile,
+            std::make_unique<psst::PsstUiDelegateImpl>(
+                psst::BravePsstPermissionContextFactory::GetForProfile(
+                    profile)),
+            profile->GetPrefs(), ISOLATED_WORLD_ID_BRAVE_INTERNAL);
```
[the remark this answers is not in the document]
bridiver: this is not adding a new usage

components/psst/browser/core/psst_consent_data.h
bridiver: but why is it part of this struct? It doesn't seem like it belongs here

browser/psst/psst_ui_delegate_impl.cc
goodov: I guess we can try, but it might be too problematic to try to wire those. PSST has its own UI.

USB/HID/Serial (object-based permissions) do not go through `PermissionRequestManager` and `PermissionRequest`.

browser/psst/brave_psst_utils_unittest.cc
```diff
@@ -0,0 +1,179 @@
...
+
+base::Value::List VectorToList(const std::vector<std::string>& values) {
+  base::Value::List list;
+  for (auto& value : values) {
+    list.Append(value);
+  }
+  return list;
+}
+
+base::Value::Dict CreatePsstPermissionDict(
```
goodov: this is not permission, this is PSST settings. Please remove any mention of "Permission" from this work to make this less confusing for everyone.

browser/psst/psst_ui_delegate_impl.h
```diff
@@ -6,24 +6,46 @@
...
+  std::optional<PsstMetadata> GetPsstMetadata(
+      const url::Origin& origin,
+      const std::string& user_id) override;
+
+ private:
+  void OnUserAcceptedPsstSettings(base::Value::List urls_to_skip);
+
+  raw_ptr<content::WebContents> web_contents_;
+  std::optional<PsstConsentData> dialog_data_;
+  raw_ptr<HostContentSettingsMap> host_content_settings_map_;
```
goodov: `= nullptr;` all raw_ptrs please.

browser/browser_context_keyed_service_factories.cc
```diff
@@ -103,7 +99,7 @@
 #endif
 
 #if BUILDFLAG(ENABLE_PSST)
-#include "brave/browser/psst/brave_psst_permission_context_factory.h"
+// #include "brave/browser/psst/brave_psst_permission_context_factory.h"
```
goodov: ?

components/psst/browser/core/brave_psst_utils.cc
```diff
@@ -0,0 +1,113 @@
...
+             ToString(psst_metadata.consent_status));
+  object.Set(kScriptVersionPermissionKey, psst_metadata.script_version);
+  object.Set(kUrlsToSkipPermissionKey,
+             VectorToList(std::move(psst_metadata.urls_to_skip)));
+  return object;
+}
+
+}  // namespace
+
+std::optional<PsstMetadata> GetPsstMetadata(HostContentSettingsMap* map,
```
goodov: let's use `PsstSettings` or `PsstWebsiteSettings` instead of `PsstMetadata/Permission`.

components/psst/browser/core/brave_psst_utils.cc
```diff
@@ -0,0 +1,113 @@
...
+          .Set(kScriptVersionPermissionKey, script_version)
+          .Set(kUrlsToSkipPermissionKey, std::move(urls_to_skip)));
+  if (!psst_metadata) {
+    return;
+  }
+
+  SetPsstMetadata(map, origin, std::move(*psst_metadata));
+}
+
+void SetPsstMetadata(HostContentSettingsMap* map,
```
goodov: please add a service which you will use to store/get `PsstSettings` for a website. Let's not go the legacy brave shields style with global function.

components/psst/browser/core/psst_consent_data.h
```diff
@@ -0,0 +1,52 @@
...
+
+namespace psst {
+
+using ConsentCallback =
+    base::OnceCallback<void(const base::Value::List disabled_checks)>;
+
+// Represents all the data required to display the consent dialog to the user.
+// This includes the information to be shown, as well as configuration for user
+// interaction and capturing their consent response.
+struct PsstConsentData {
```
goodov: why have two separate structs `PsstConsentData` and `PsstSettings/Metadata`? Can't this be the one struct? And you need to remove callback from this struct. Just make it the data storage and pass UI-specific callback separately.

components/psst/browser/core/brave_psst_utils.cc
vadimstruts: Could we move it to the follow up PR?

goodov: yes, fine for a follow up.

components/psst/browser/core/psst_consent_data.h
goodov: seems like this is resolved and now we have only a single struct. nice!

browser/psst/psst_ui_delegate_impl.cc
```diff
@@ -5,16 +5,56 @@
...
+void PsstUiDelegateImpl::OnUserAcceptedPsstSettings(
+    const url::Origin& origin,
+    base::Value::List urls_to_skip) {
+  // Save the PSST settings when user accepts the dialog
+  SetPsstWebsiteSettings(host_content_settings_map_, origin,
+                         ConsentStatus::kAllow, dialog_data_->script_version,
+                         dialog_data_->user_id, std::move(urls_to_skip));
+
+  if (apply_changes_callback_) {
+    std::move(apply_changes_callback_).Run(std::move(urls_to_skip));
```
goodov, line 56: `urls_to_skip` is already moved.

components/psst/browser/core/psst_consent_data.h
vadimstruts: removed the callback from the struct

components/psst/browser/core/brave_psst_utils.cc
vadimstruts: Follow up: https://github.com/brave/brave-browser/issues/51307

PR #31353, opened 2025-09-22
[Android] Attempt to fix cut-off background image after app restart

Olipro:
Occasionally, users have reported that the NTP (New Tab Page) background image is partially cut-off after restarting the app. The best working-theory that I have around this is that there is some sort of layout inconsistency happening when the layout is internally saved and restored.

It is entirely possible that this problem is caused by the XML layout definition causing some sort of inconsistency when Android decides how to size elements - the most reasonable explanation being that when restoring the view instead of drawing it from scratch, elements are evaluated in a different order.

This commit attempts to programatically compute and apply the size to the background image after built-in sizing has occured. There is quite possibly a better, cleaner fix that could possibly be done entirely in XML.

Might sort out brave/brave-browser#28664 - but it should be closed manually once we've established whether or not this does anything useful.

### DIFF
--- a/android/java/org/chromium/chrome/browser/ntp/BraveNewTabPageLayout.java
+++ b/android/java/org/chromium/chrome/browser/ntp/BraveNewTabPageLayout.java
@@ -1596,5 +1596,16 @@ public class BraveNewTabPageLayout extends NewTabPageLayout
         }
 
         super.onMeasure(widthMeasureSpec, heightMeasureSpec);
+
+        // This is an attempt to fix the occasional issue where the browser is restarted
+        // and the NTP cuts off the bottom of the background image. There may well be a
+        // cleaner fix (assuming this fixes it at all) but I have not managed to ascertain
+        // exactly how the problem is able to occur.
+        if (mBgImageView != null && mBgImageView.getMeasuredHeight() < getMeasuredHeight()) {
+            int bgWidthSpec = MeasureSpec.makeMeasureSpec(getMeasuredWidth(), MeasureSpec.EXACTLY);
+            int bgHeightSpec =
+                    MeasureSpec.makeMeasureSpec(getMeasuredHeight(), MeasureSpec.EXACTLY);
+            mBgImageView.measure(bgWidthSpec, bgHeightSpec);
+        }
     }
 }


android/java/org/chromium/chrome/browser/ntp/BraveNewTabPageLayout.java
```diff
@@ -1596,5 +1596,16 @@ protected void onMeasure(int widthMeasureSpec, int heightMeasureSpec) {
         }
 
         super.onMeasure(widthMeasureSpec, heightMeasureSpec);
+
+        // This is an attempt to fix the occasional issue where the browser is restarted
+        // and the NTP cuts off the bottom of the background image. There may well be a
+        // cleaner fix (assuming this fixes it at all) but I have not managed to ascertain
+        // exactly how the problem is able to occur.
+        if (mBgImageView != null && mBgImageView.getMeasuredHeight() < getMeasuredHeight()) {
```
samartnik, line 1604: It seems like problem is with incorrect `mBgImageView.getMeasuredHeight()` when we create `FetchWallpaperWorkerTask`. Apparently sometimes `mBgImageView` is not measured yet at this point.
Not sure if this `if` will ever happen after `super.onMeasure`.
I believe cleaner,  but still only possible, fix would be to make sure that we call `setBackgroundImage` after `onMeasure`.

Olipro: The theory is that during a restore, the order of size computation is different which then causes the background image's dimensions to be wrong. Based on that idea, the thought is that after `super.onMeasure()` has returned, since all elements have been calculated we can check if the height is a mis-match.

Of course, this is very speculative since I can't get the bug to manifest on my emulator. What's your thinking with explicitly setting the image? Does this guarantee a re-computation of the element's sizing and layout?

samartnik: It seems like we should move these 2 calls from `onAttachedToWindow` to `onMeasure`, cause we need proper measurements at time of fetching image. But that's also a speculation, that move may introduce some other unintended issues.
```
checkAndShowNTPImage(false);
mNTPBackgroundImagesBridge.addObserver(mNTPBackgroundImageServiceObserver);
```

samartnik: generally this change lgtm, if you believe it may help

Olipro: My thinking is that at worst, it does nothing useful nor harmful, but at least rules something out. If it turns out this doesn't fix it, I'll give your approach a shot - I'd rather not do both at the same time or we won't know which change fixed it (assuming one of them does)

PR #31354, opened 2025-09-22
chore(wallet): Wallet Web Screen Updates for iOS

nuo-xu:
1. encode asset image url in select asset button
2. using isMobile flag to maintain width and padding in onboarding screen
3. using minWidth for welcome actions image to avoid compressing

Resolves https://github.com/brave/brave-browser/issues/49516

### DIFF
--- a/components/brave_wallet_ui/page/screens/fund-wallet/components/select_asset_button/select_asset_button.tsx
+++ b/components/brave_wallet_ui/page/screens/fund-wallet/components/select_asset_button/select_asset_button.tsx
@@ -58,6 +58,7 @@ export const SelectAssetButton = (props: SelectAssetButtonProps) => {
 
   // Computed
   const assetSymbol = selectedAsset ? getAssetSymbol(selectedAsset) : ''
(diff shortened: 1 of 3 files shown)


Douglashdaniel (approved):
Looks good to me!

Lets make sure @kdenhartog is good with the `encodeURIComponent` change, but I think that should have actually been there to begin with.

Can you also test that those `Icons` still load on the `Desktop` and `Panel`?

nuo-xu:
thanks @Douglashdaniel 
> Looks good to me!
> 
> Lets make sure @kdenhartog is good with the `encodeURIComponent` change, but I think that should have actually been there to begin with.

Yes other places are using `encodeURIComponent` and i noticed this when i was running on my webui branch and those un-encoded icons are not displayed.

> 
> Can you also test that those `Icons` still load on the `Desktop` and `Panel`?

Yes i have tested on mac. this displays correctly with this change.

kdenhartog (approved):
Thanks for making these fixes

PR #31355, opened 2025-09-22
[ads] Search matches should only be posted after autocomplete is called

tmancey:
Currently, search matches are posted into all background iframes, even when those iframes have not requested or are not listening for autocomplete results. This can cause unnecessary message traffic.

Resolves https://github.com/brave/brave-browser/issues/49515

### DIFF
--- a/browser/resources/brave_new_tab_page_refresh/components/background/background.tsx
+++ b/browser/resources/brave_new_tab_page_refresh/components/background/background.tsx
@@ -101,7 +101,8 @@ function SponsoredRichMediaBackground(
   // Refactor rich media background components.
   const searchActions = useSearchActions()
   const searchMatches = useSearchState((s) => s.searchMatches)
-  useSearchMatchesReporter(frameHandle, searchMatches)
+  const shouldMatchSearches =
+    useSearchMatchesReporter(frameHandle, searchMatches)
 
   return (
     <IframeBackground
@@ -133,6 +134,7 @@ function SponsoredRichMediaBackground(
           case 'richMediaQueryBraveSearchAutocomplete': {
             const value = String(data.value ?? '')
             if (value) {
+              shouldMatchSearches()
               searchActions.queryAutocomplete(value, 'search.brave.com')
             }
             break
@@ -213,8 +215,10 @@ function useSearchMatchesReporter(
   frameHandle?: IframeBackgroundHandle,
   searchMatches?: AutocompleteMatch[]
 ) {
+  const [shouldReport, setShouldReport] = React.useState(false)
+
   React.useEffect(() => {
-    if (!frameHandle) {
+    if (!frameHandle || !shouldReport) {
       return
     }
 
@@ -226,7 +230,9 @@ function useSearchMatchesReporter(
     }, 120)
 
     postSearchMatches()
-  }, [frameHandle, searchMatches])
+  }, [frameHandle, searchMatches, shouldReport])
+
+  return () => setShouldReport(true)
 }
 
 interface IframeBackgroundHandle {


browser/resources/brave_new_tab_page_refresh/components/background/background.tsx
```diff
@@ -213,20 +220,18 @@ function useSearchMatchesReporter(
...
+    return
+  }
 
-    const postSearchMatches = debounce(() => {
-      frameHandle.postMessage({
-        type: 'richMediaSearchMatches',
-        value: JSON.stringify(searchMatches ?? [])
-      })
-    }, 120)
+  const postSearchMatches = debounce(() => {
```
zenparsing: Since this function no longer uses any React hooks, we should rename it so that it doesn't start with "use".

We might just want to inline the `postMessage` call into the message handler above and remove this separate function. I don't believe that we need to `debounce` any more, since we only post once for each autocomplete query.

browser/resources/brave_new_tab_page_refresh/components/background/background.tsx
```diff
@@ -94,14 +94,21 @@ function SponsoredRichMediaBackground(
...
-  const searchActions = useSearchActions()
-  const searchMatches = useSearchState((s) => s.searchMatches)
-  useSearchMatchesReporter(frameHandle, searchMatches)
+  // Only report search matches when flag is set
+  React.useEffect(() => {
+    if (shouldReportSearchMatches && frameHandle) {
+      useSearchMatchesReporter(frameHandle, searchMatches)
+      setShouldReportSearchMatches(false)
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
zenparsing: Can we remove this lint-disabling comment?

browser/resources/brave_new_tab_page_refresh/components/background/background.tsx
```diff
@@ -94,14 +94,21 @@ function SponsoredRichMediaBackground(
...
+  const searchActions = useSearchActions()
 
   useSafeAreaReporter(frameHandle)
 
-  // TODO(https://github.com/brave/brave-browser/issues/49471): [NTP Next]
-  // Refactor rich media background components.
-  const searchActions = useSearchActions()
-  const searchMatches = useSearchState((s) => s.searchMatches)
-  useSearchMatchesReporter(frameHandle, searchMatches)
+  // Only report search matches when flag is set
```
zenparsing: I think we lost the refactoring TODO.

tmancey: Resolving as pushed different solution.

browser/resources/brave_new_tab_page_refresh/components/background/background.tsx
tmancey: Resolving as pushed different solution.

browser/resources/brave_new_tab_page_refresh/components/background/background.tsx
tmancey: Resolving as pushed different solution.

browser/resources/brave_new_tab_page_refresh/components/background/background.tsx
```diff
@@ -223,10 +231,11 @@ function useSearchMatchesReporter(
         type: 'richMediaSearchMatches',
         value: JSON.stringify(searchMatches ?? [])
       })
+      onReported?.()
     }, 120)
 
     postSearchMatches()
-  }, [frameHandle, searchMatches])
+  }, [frameHandle, searchMatches, shouldReport, onReported])
```
zenparsing: Since `onReported` is a new function each time, this will cause the effect to run after each render.

There are a couple of solutions I can think of. You could use `React.useCallback` to ensure a stable callback function.

(Note: I haven't tested either of these snippets.)

```ts

const onReported = React.useCallback(() => {
  setShouldReportSearchMatches(false)
}, [])

useSearchMatchesReporter(frameHandle, searchMatches, onReported)

```

Or you could put the "shouldReport" state into this hook, and return a function that will active the "shouldReport" flag.

```ts

function useSearchMatchesReporter(
  frameHandle?: IframeBackgroundHandle,
  searchMatches?: AutocompleteMatch[]
) {
  const [shouldReport, setShouldReport] = React.useState(false)

  // ...

  return () => setShouldReport(true)
}
```

tmancey: Pushed changes. Thanks

browser/resources/brave_new_tab_page_refresh/components/background/background.tsx
```diff
@@ -226,7 +230,10 @@ function useSearchMatchesReporter(
     }, 120)
 
     postSearchMatches()
-  }, [frameHandle, searchMatches])
+    setShouldReport(false)
```
zenparsing: Since we turn off the match reporter straight after sending the matches, does that mean the iframe only gets the existing set of matches, before the autocomplete query has had time to complete?

tmancey: good question, and once my build finishes, I will test that. If so, we need another solution.

PR #31357, opened 2025-09-22
Adblock only mode pref and feature

aseren:
Resolves https://github.com/brave/brave-browser/issues/49526


components/brave_shields/core/common/BUILD.gn
```diff
@@ -61,11 +61,17 @@ mojom_component("mojom") {
 source_set("unit_tests") {
   testonly = true
 
-  sources = [ "brave_shields_settings_values_unittest.cc" ]
+  sources = [
+    "brave_shield_utils_unittest.cc",
```
tmancey, line 65: brave_shields_utils_unittest to match fixture name

components/brave_shields/core/common/brave_shield_utils_unittest.cc
```diff
@@ -0,0 +1,60 @@
...
+  RegisterLocalStatePrefs(local_state.registry());
+
+  local_state.SetBoolean(prefs::kAdBlockOnlyModeEnabled, true);
+  EXPECT_TRUE(IsAdblockOnlyModeFeatureEnabled(&local_state));
+
+  local_state.SetBoolean(prefs::kAdBlockOnlyModeEnabled, false);
+  EXPECT_FALSE(IsAdblockOnlyModeFeatureEnabled(&local_state));
+}
+
+TEST(BraveShieldsUtilsTest, SetBraveShieldsAdBlockOnlyModeEnabled) {
```
tmancey, line 49: The test is named Set...Enabled but it tests both enabled and disabled

components/brave_shields/core/common/brave_shield_utils_unittest.cc
```diff
@@ -0,0 +1,60 @@
...
+  {
+    base::test::ScopedFeatureList feature_list;
+    feature_list.InitAndEnableFeature(features::kAdblockOnlyMode);
+    EXPECT_TRUE(IsAdblockOnlyModeFeatureEnabled());
+  }
+  {
+    base::test::ScopedFeatureList feature_list;
+    feature_list.InitAndDisableFeature(features::kAdblockOnlyMode);
+    EXPECT_FALSE(IsAdblockOnlyModeFeatureEnabled());
+  }
```
tmancey, line 35: Ideally this should test the expected state and for now that is disabled

components/brave_shields/core/common/brave_shield_utils_unittest.cc
```diff
@@ -0,0 +1,60 @@
...
+}
+
+TEST(BraveShieldsUtilsTest, IsBraveShieldsAdBlockOnlyModeEnabled) {
```
tmancey, line 38: nit: test name does not need BraveShields because fixture has that

components/brave_shields/core/common/BUILD.gn
aseren: brave_shield_util.cc is the filename, for util, so it should match it

components/brave_shields/core/common/brave_shield_utils_unittest.cc
aseren: it checks SetBraveShieldsAdBlockOnlyModeEnabled function

components/brave_shields/core/common/brave_shield_utils_unittest.cc
aseren: it checks IsAdblockOnlyModeFeatureEnabled function

components/brave_shields/core/common/brave_shield_utils_unittest.cc
aseren: IsBraveShieldsAdBlockOnlyModeEnabled tests specific function

components/brave_shields/core/common/BUILD.gn
tmancey: Ok so the unit test fixture should be BraveShieldUtilsTest instead of BraveShields...

components/brave_shields/core/common/brave_shield_utils_unittest.cc
tmancey: Discussed and another test will verify that the pref is false.

components/brave_shields/core/common/brave_shield_utils.cc
```diff
@@ -99,4 +103,19 @@ ShieldsSettingCounts GetSettingCountFromCosmeticFilteringRules(
   return result;
 }
 
+bool IsAdblockOnlyModeFeatureEnabled() {
```
bridiver, line 106: helper methods for single pref or feature accessors are not necessary, but if they do exist, they should be in features or pref_names so they don't create unnecessary dependencies

components/brave_shields/core/common/brave_shield_utils_unittest.cc
bridiver: please do not create change detector tests https://source.chromium.org/chromium/chromium/src/+/main:docs/chrome_browser_design_principles.md

components/brave_shields/core/common/brave_shield_utils.h
```diff
@@ -36,6 +37,12 @@ ShieldsSettingCounts GetSettingCountFromRules(
 ShieldsSettingCounts GetSettingCountFromCosmeticFilteringRules(
     const ContentSettingsForOneType& fp_rules);
 
+bool IsAdblockOnlyModeFeatureEnabled();
+
+bool IsBraveShieldsAdBlockOnlyModeEnabled(PrefService* local_state);
```
bridiver, line 42: why are these added in common? PrefService is not available outside of the browser process

components/brave_shields/core/common/DEPS
```diff
@@ -1,5 +1,6 @@
 include_rules = [
   "+components/content_settings",
   "+components/grit/brave_components_strings.h",
+  "+components/prefs",
```
bridiver, line 4: @mkarolin this dep is not appropriate for core/common because PrefService can only be accessed from the browser process

components/brave_shields/core/common/brave_shield_utils.cc
aseren: Filed an issue for removal https://github.com/brave/brave-browser/issues/50414

components/brave_shields/core/common/DEPS
aseren: sorry, missed this. this dep will be removed in https://github.com/brave/brave-browser/issues/50414

components/brave_shields/core/common/brave_shield_utils.h
aseren: sorry, missed this. Pref service dep will be removed in https://github.com/brave/brave-browser/issues/50414

PR #31358, opened 2025-09-22
fix(privacy): Shred on app exit default value

StephenHeaps:
- CoreData predicate being `"shield_shredLevel == "nil""` vs `"shield_shredLevel == nil"` caused Domains without a explicit shred level set to not match

Resolves https://github.com/brave/brave-browser/issues/49464
 

### Testplan

1. Open 2 tabs with different sites on each
2. On tab 1, open shields panel and set Auto Shred level to App Close. Don't touch tab 2 shields panel Auto Shred level.
3. Open Settings > Shields & Privacy and set Auto Shred level to App Close
4. Quit the app and relaunch
5. Verify both tabs are closed
6. Verify data for both sites is removed from Settings > Shields & Privacy > Manage Site Data
7. Verify history item(s) are removed

### DIFF
--- a/ios/brave-ios/Sources/BraveShields/ShieldPreferences.swift
+++ b/ios/brave-ios/Sources/BraveShields/ShieldPreferences.swift
@@ -33,9 +33,9 @@ public class ShieldPreferences {
     default: nil
   )
 
-  /// Get the level of the https upgrade setting as a stored preference
-  /// - Warning: You should not access this directly but  through ``httpsUpgradeLevel``
-  private static var shredLevelRaw = Preferences.Option<String?>(
+  /// Get the auto shred level setting as a stored preference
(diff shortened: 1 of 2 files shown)


kjozwiak:
Verification **PASSED** on `iPhone 11` using `iOS 26` via the following build(s):

```
Brave | 1.84.97 Chromium: 141.0.7390.30 (Official Build) nightly (64-bit) 
--- | ---
Revision | fa6c2a0c5191...
OS | iOS
```

Using the STR/Cases outlined via https://github.com/brave/brave-core/pull/31358#issue-3443109717, ensured that the currently opened tabs were removed/closed and both `Manage Site Data` and `History` were cleared as per the following:
 
https://github.com/user-attachments/assets/440e33c6-939f-4e4b-bd87-92045692b595

PR #31364, opened 2025-09-23
[CR141] Fixed split view border drawing regressions from cr141

simonhong:
Resolves https://github.com/brave/brave-browser/issues/49529

In cr141, chromium introduced `ContentsContainerOutline` for handling split view's outline and it's conflicted with ours.
With this PR, only our border drawing is done.

Screenshot 2025-09-23 161951
Screenshot 2025-09-23 162015


simonhong:
Set `skip-ios` label as this PR definitely only includes desktop only file changes.
I think this `BraveShieldsTabHelperTests.swift:513: XCTAssertEqual failed: ("appExit") is not equal to ("never")` failure on ios CI could be resolved with rebase.

PR #31365, opened 2025-09-23
"Forget me when I close this site" (not working) on Android. Cookies and site data persist.

vadimstruts:
Resolves https://github.com/brave/brave-browser/issues/47595

- Added additional management for TLD ephemeral lifetimes to prevent cases where cookies and site data were not deleted because the WebContentsObserver was not destructed


goodov (approved):
can't say if it'll fix the issue, but worth the shot.

PR #31367, opened 2025-09-23
add polkadot substrate rpc unittests

cmazakas:
Resolves https://github.com/brave/brave-browser/issues/49545

* Updates code to use JSON compiler for response parsing
* Update error handling

The Polkadot RPC nodes seem to return a 200 Ok when there's an error and they instead push the error itself into the JSON response, so the tests have been updated to reflect that.

### DIFF
--- a/components/brave_wallet/browser/polkadot/polkadot_substrate_rpc_unittest.cc
+++ b/components/brave_wallet/browser/polkadot/polkadot_substrate_rpc_unittest.cc
@@ -75,11 +75,9 @@ TEST_F(PolkadotSubstrateRpcUnitTest, GetChainName) {
       "id": 1 })");
   polkadot_substrate_rpc_->GetChainName(chain_id, future.GetCallback());
 
-  {
-    auto const [network_name, error] = future.Take();
-    EXPECT_EQ(network_name, "Westend");
-    EXPECT_EQ(error, std::nullopt);
-  }
+  auto [network_name, error] = future.Take();
+  EXPECT_EQ(network_name, "Westend");
+  EXPECT_EQ(error, std::nullopt);
 
   url_loader_factory_.AddResponse(testnet_url, R"(
     { "jsonrpc": "2.0",
@@ -87,21 +85,17 @@ TEST_F(PolkadotSubstrateRpcUnitTest, GetChainName) {
       "id": 1 })");
   polkadot_substrate_rpc_->GetChainName(chain_id, future.GetCallback());
 
-  {
-    auto const [network_name, error] = future.Take();
-    EXPECT_EQ(network_name, std::nullopt);
-    EXPECT_EQ(error, WalletParsingErrorMessage());
-  }
+  std::tie(network_name, error) = future.Take();
+  EXPECT_EQ(network_name, std::nullopt);
+  EXPECT_EQ(error, WalletParsingErrorMessage());
 
   url_loader_factory_.AddResponse(testnet_url, R"(
     { "id": 1 })");
   polkadot_substrate_rpc_->GetChainName(chain_id, future.GetCallback());
 
-  {
-    auto const [network_name, error] = future.Take();
-    EXPECT_EQ(network_name, std::nullopt);
-    EXPECT_EQ(error, WalletParsingErrorMessage());
-  }
+  std::tie(network_name, error) = future.Take();
+  EXPECT_EQ(network_name, std::nullopt);
+  EXPECT_EQ(error, WalletParsingErrorMessage());
 
   url_loader_factory_.AddResponse(testnet_url, R"(
     {"jsonrpc":"2.0",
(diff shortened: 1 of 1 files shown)


components/brave_wallet/browser/polkadot/polkadot_substrate_rpc_unittest.cc
```diff
@@ -0,0 +1,114 @@
...
+
+  EXPECT_CALL(callback, Run(std::optional<std::string>("westend"),
+                            std::optional<std::string>(std::nullopt)));
+  url_loader_factory_.AddResponse(testnet_url, R"(
+    { "jsonrpc": "2.0",
+      "result": "westend",
+      "id": 1 })");
+  polkadot_substrate_rpc_->GetChainName(chain_id, callback.Get());
+  task_environment_.RunUntilIdle();
+  testing::Mock::VerifyAndClearExpectations(&callback);
```
supermassive: We should avoid `RunUntilIdle()`
Instead of `MockCallback`, `EXPECT_CALL`, `RunUntilIdle`, `VerifyAndClearExpectations`
let's use `base::TestFuture`
https://github.com/brave/brave-core/blob/master/components/brave_wallet/browser/cardano/cardano_get_utxos_task_unittest.cc#L95-L113

components/brave_wallet/browser/json_rpc_responses.idl
```diff
@@ -69,4 +69,11 @@ namespace json_rpc_responses {
     DOMString totalBalanceUsd;
     AnkrAssetBalance[] assets;
   };
+
+  dictionary PolkadotSystemChainResponse {
+    DOMString jsonrpc;
+    long id;
+    DOMString? result;
+    Error? error;
+  };
```
supermassive: Should be in a separate polkadot_rpc similar to
https://github.com/brave/brave-core/blob/master/components/brave_wallet/browser/cardano/cardano_rpc_blockfrost_api.idl#L6

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc_unittest.cc
cmazakas: This should be addressed! Super cool tip, btw.
Really simplified the tests.

components/brave_wallet/browser/json_rpc_responses.idl
cmazakas: This is done too! Good idea to keep this all separated out. I didn't realize we really _have_ to put that BUILD.gn there. Good to see the other coin types have it too.

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc_responses.idl
```diff
@@ -0,0 +1,19 @@
+/* Copyright (c) 2022 The Brave Authors. All rights reserved.
```
supermassive: 2025

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.cc
```diff
@@ -48,45 +51,59 @@ PolkadotSubstrateRpc::PolkadotSubstrateRpc(
...
 
-  const auto* chain_name = result->GetIfString();
-  if (!chain_name) {
+  if (res->error) {
+    if (res->error->message) {
+      return std::move(callback).Run(std::nullopt, res->error->message.value());
+    }
     return std::move(callback).Run(std::nullopt,
-                                   "result field was not a string");
+                                   WalletUserRejectedRequestErrorMessage());
```
supermassive: Typically if don't know what happened we just return WalletInternalError

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc_unittest.cc
```diff
@@ -0,0 +1,125 @@
...
+  url_loader_factory_.AddResponse(testnet_url, R"(
+    { "jsonrpc": "2.0",
+      "not_result": "westend",
+      "id": 1 })");
+  polkadot_substrate_rpc_->GetChainName(chain_id, future.GetCallback());
+
+  EXPECT_TRUE(future.Wait());
+  EXPECT_EQ(future.Get<0>(), std::nullopt);
+  EXPECT_EQ(future.Get<1>(), WalletParsingErrorMessage());
+  future.Clear();
```
supermassive: nit: future.Wait() + future.Clear() could be replaced with future.Take()

cmazakas: Done!

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc_responses.idl
cmazakas: Good catch. Always the copy-paste errors. This is done.

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.cc
cmazakas: I also realized this branch was missing coverage as well. This has been addressed with an updated unit test.

PR #31371, opened 2025-09-23
Bitcoin-core compilation fix

cypt4:
Resolves https://github.com/brave/brave-browser/issues/49546

### DIFF
--- a/chromium_src/brave/third_party/bitcoin-core/src/src/span.h
+++ b/chromium_src/brave/third_party/bitcoin-core/src/src/span.h
@@ -1,0 +1,22 @@
+/* Copyright (c) 2025 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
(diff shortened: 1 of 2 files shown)


chromium_src/brave/third_party/bitcoin-core/src/src/span.h
```diff
@@ -0,0 +1,22 @@
+/* Copyright (c) 2025 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef BRAVE_CHROMIUM_SRC_BRAVE_THIRD_PARTY_BITCOIN_CORE_SRC_SRC_SPAN_H_
+#define BRAVE_CHROMIUM_SRC_BRAVE_THIRD_PARTY_BITCOIN_CORE_SRC_SRC_SPAN_H_
+
+// clang-format off
+#include <utility>
```
cdesouza-chromium, line 10: I think there's a way to comment `  // NO LINT` here with the specific failure that you don't want to lint, and then you won't need the little code below.

cypt4: I've tried several variants and they didn't work

PR #31382, opened 2025-09-23
Update string conversion to use `base::` methods

bsclifton:
Fixes https://github.com/brave/brave-browser/issues/49425

### DIFF
--- a/browser/brave_vpn/win/wireguard_utils_win.cc
+++ b/browser/brave_vpn/win/wireguard_utils_win.cc
@@ -307,11 +307,8 @@ wireguard::WireguardKeyPair WireguardGenerateKeypairImpl() {
     return std::nullopt;
   }
 
-  std::string public_key = reinterpret_cast<std::string::value_type*>(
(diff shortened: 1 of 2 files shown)


fmarier:
If you haven't done so already, it may be worth grepping for `reinterpret_cash` recursively in the vpn component to see if there are other instances worth looking at.

bsclifton:
Going to let PR builder finish this and will install / run locally to verify everything looks good 😄

bsclifton:
OK installed locally using the PR builder executable:
`brave_installer_141_1_84_0.exe --system-level`

Connected/disconnected a bunch of times and made sure WireGuard was being used. Things look good!

bsclifton:
The CI that failed wasn't related to this - but I pulled latest/rebased to pull in the fixes folks have done (https://github.com/brave/brave-core/pull/31420 and https://github.com/brave/brave-core/pull/31396). Will let CI run again - then I think we're good 😄

PR #31380, opened 2025-09-23
Shields Panel Ad Block Only mode UI

aseren:
The PR adds Ad Block Only mode dialogs to the Shields Panel UI.

The UI is under a feature which is turned off by default.
QA testing will be done when MVP functionality is landed via Epic https://github.com/brave/brave-browser/issues/48521

Shields panel screenshots:

image

image

image

image

Backend part will be added in https://github.com/brave/brave-browser/issues/49565

Epic for this task is: https://github.com/brave/brave-browser/issues/48521
Spec: https://docs.google.com/document/d/1-y3NIRvt0scNK7qJ-XUuY0wRcnFuT08xsxivbUtr9SE/edit?tab=t.0

Resolves https://github.com/brave/brave-browser/issues/49564

PR #31384, opened 2025-09-23
Shields Panel Ad Block Only mode support

aseren:
The PR adds Ad Block Only mode business logic for Shields Panel. Ad Block Only mode UI part was added here: https://github.com/brave/brave-core/pull/31380

The PR implements following functionality:

1) Enables `Ad Block Only mode` from Shields Panel on button press. Handles `Dismiss` button press, so prompt is no longer shown.
image

2) Shows following prompt when user disable shield specific number of times.
image

This is controlled by `AdBlockOnlyMode` feature parameters:
```
"prompt_after_shields_disabled_count" (default is 5)
```

3) Shows following prompt when user reloads web-page several times during specific time interval.
image

This is controlled by `AdBlockOnlyMode` feature parameters:
```
"prompt_after_page_reloads_interval" (default is 10 seconds)
"prompt_after_page_reloads_min" (default is 0)
"prompt_after_page_reloads_max" (default is 0)
```
When `prompt_after_page_reloads_min` and `prompt_after_page_reloads_max` both equal to 0 then the prompt is not shown.

Resolves https://github.com/brave/brave-browser/issues/49565

### DIFF
--- a/browser/brave_shields/brave_shields_tab_helper.cc
+++ b/browser/brave_shields/brave_shields_tab_helper.cc
@@ -17,6 +17,9 @@
 #include "brave/components/brave_shields/core/browser/brave_shields_settings.h"
 #include "brave/components/brave_shields/core/browser/brave_shields_utils.h"
 #include "brave/components/brave_shields/core/common/brave_shield_constants.h"
+#include "brave/components/brave_shields/core/common/brave_shield_utils.h"
+#include "brave/components/brave_shields/core/common/features.h"
+#include "brave/components/brave_shields/core/common/pref_names.h"
 #include "chrome/browser/browser_process.h"
 #include "chrome/browser/content_settings/cookie_settings_factory.h"
 #include "chrome/browser/content_settings/host_content_settings_map_factory.h"
@@ -24,6 +27,7 @@
 #include "components/content_settings/core/browser/cookie_settings.h"
 #include "components/content_settings/core/common/content_settings_pattern.h"
 #include "components/favicon/content/content_favicon_driver.h"
+#include "components/prefs/pref_service.h"
 #include "content/public/browser/navigation_controller.h"
 #include "content/public/browser/navigation_handle.h"
 #include "content/public/browser/reload_type.h"
@@ -59,6 +63,13 @@ BraveShieldsTabHelper::BraveShieldsTabHelper(content::WebContents* web_contents)
       std::make_unique<brave_shields::BraveShieldsSettings>(
           *host_content_settings_map_, g_browser_process->local_state(),
           profile_prefs);
+
+  local_state_change_registrar_.Init(g_browser_process->local_state());
+  local_state_change_registrar_.Add(
+      brave_shields::prefs::kAdBlockOnlyModeEnabled,
+      base::BindRepeating(
+          &BraveShieldsTabHelper::OnShieldsAdBlockOnlyModeEnabledChanged,
+          base::Unretained(this)));
 }
 
 void BraveShieldsTabHelper::DidFinishNavigation(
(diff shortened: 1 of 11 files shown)


browser/brave_shields/brave_shields_tab_helper.cc
```diff
@@ -59,6 +63,13 @@ BraveShieldsTabHelper::BraveShieldsTabHelper(content::WebContents* web_contents)
       std::make_unique<brave_shields::BraveShieldsSettings>(
           *host_content_settings_map_, g_browser_process->local_state(),
           profile_prefs);
+
+  local_state_change_registrar_.Init(g_browser_process->local_state());
+  local_state_change_registrar_.Add(
+      brave_shields::prefs::kAdBlockOnlyModeEnabled,
+      base::BindRepeating(
+          &BraveShieldsTabHelper::OnShieldsAdBlockOnlyModeEnabledChanged,
+          base::Unretained(this)));
```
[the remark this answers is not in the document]
aseren: base::Unretained is used for Local State `PrefChangeRegistrar`

fallaciousreasoning (approved):
lgtm - thanks heaps for splitting this up :smile

browser/brave_shields/brave_shields_tab_helper_unittest.cc
```diff
@@ -0,0 +1,305 @@
...
+TEST_F(BraveShieldsTabHelperUnitTest,
+       DontTriggerOnRepeatedReloadsDetectedOnNotReloadNavigation) {
+  feature_list_.InitAndEnableFeatureWithParameters(
+      features::kAdblockOnlyMode,
+      {
+          {features::kAdblockOnlyModePromptAfterPageReloadsMin.name, "1"},
+          {features::kAdblockOnlyModePromptAfterPageReloadsMax.name, "2"},
+          {features::kAdblockOnlyModePromptAfterPageReloadsInterval.name, "1s"},
+      });
+  SetBraveShieldsAdBlockOnlyModeEnabled(local_state(), false);
```
bridiver, line 142: why are you explicitly setting this to the default value?

browser/brave_shields/brave_shields_tab_helper.cc
```diff
@@ -71,6 +82,72 @@ void BraveShieldsTabHelper::DidFinishNavigation(
...
+
+    if (!navigation_triggered_by_shields_changes_) {
+      MaybeNotifyRepeatedReloads(navigation_handle);
+    }
+
+    navigation_triggered_by_shields_changes_ = false;
+  }
+}
+
+void BraveShieldsTabHelper::MaybeNotifyRepeatedReloads(
```
bridiver, line 94: This is not appropriate functionality to add to this tab helper, it's purpose is to handle shields setting changes, not to handle prompts to enable adblock only mode

browser/brave_shields/brave_shields_tab_helper.cc
```diff
@@ -71,6 +82,72 @@ void BraveShieldsTabHelper::DidFinishNavigation(
...
+          features::kAdblockOnlyModePromptAfterPageReloadsMin.Get() &&
+      repeated_reloads_counter_->reloads_count <=
+          features::kAdblockOnlyModePromptAfterPageReloadsMax.Get()) {
+    for (Observer& observer : observer_list_) {
+      observer.OnRepeatedReloadsDetected();
+    }
+  }
+}
+
+void BraveShieldsTabHelper::OnShieldsAdBlockOnlyModeEnabledChanged() {
```
bridiver, line 148: I don't understand why we need a specific method for adblock only mode, it already updates for settings changes so this should not be necessary.

browser/brave_shields/brave_shields_tab_helper.cc
```diff
@@ -185,10 +264,52 @@ bool BraveShieldsTabHelper::GetBraveShieldsEnabled() {
 void BraveShieldsTabHelper::SetBraveShieldsEnabled(bool is_enabled) {
   brave_shields_settings_->SetBraveShieldsEnabled(is_enabled,
                                                   GetCurrentSiteURL());
+
+  if (brave_shields::IsAdblockOnlyModeFeatureEnabled() && !is_enabled) {
+    PrefService* prefs =
```
bridiver, line 269: please keep prefs as an instance var and ideally pass it in to the constructor

browser/brave_shields/brave_shields_tab_helper.cc
aseren: Filed an issue to fix: https://github.com/brave/brave-browser/issues/50413

browser/brave_shields/brave_shields_tab_helper_unittest.cc
aseren: yeah, this was not actually needed. Default value should be enough

browser/brave_shields/brave_shields_tab_helper.cc
aseren: `Adblock Only Mode` can be also enabled/disabled from from Shields settings page, so Shields panel should be updated to reflect `Adblock Only Mode` status.

bridiver: but that would also trigger a content settings update so it's still not clear to me why this is needed

bridiver: there are pref and content setting listeners so we shouldn't need a adblock only mode specific observer method

aseren: ah I got what  you mean, agreed that's no longer needed, because adblock only mode updates content settings
will fix it here: https://github.com/brave/brave-browser/issues/50413

PR #31391, opened 2025-09-24
[CR141] Fixed crash when clicking a split tab

simonhong:
Resolves https://github.com/brave/brave-browser/issues/49533

When split tab has tab modal, it's blocked tab and other tab in that split can't be activated.
Regarding to split tab activation, Brave and chromium have difference.
When clicking any split tab in chromium, always activated last active tab in that split.
However, we activate clicked tab. This works well in most cases.
But when one of split tab has tab modal, only that tab could be activated.
This caused crash from Brave.
To fix this, we follow chromium's activation logic when there is tab modal in split.

TEST=SideBySideEnabledBrowserTest.SelectTabTest

### DIFF
--- a/browser/ui/views/split_view/split_view_browsertest.cc
+++ b/browser/ui/views/split_view/split_view_browsertest.cc
@@ -119,6 +119,22 @@ class SideBySideEnabledBrowserTest : public InProcessBrowserTest {
     browser_non_client_frame_view()->DeprecatedLayoutImmediately();
   }
 
+  javascript_dialogs::TabModalDialogManager* GetTabModalDialogManagerAt(
+      int index) {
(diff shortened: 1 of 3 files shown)


simonhong:
Weird -  `SideBySideEnabledBrowserTest.SelectTabTest` is failed only on macOS CI. it works on local macOS..
Disabled on macOS. No idea why it's flaky on macOS CI about dialog test.

simonhong:
`testShredLevel() – BraveShieldsTabHelperTests` failure on ios CI is expected.
`WebAppTabStripBrowserTest` failure on linux CI is not related with this PR.

mkarolin (approved):
++

PR #31394, opened 2025-09-24
Introduce chrome://brave-image WebUI handler

atuchin-m:
Resolves https://github.com/brave/brave-browser/issues/48874

The PR resolves the most critical Brave News issue with CPU/memory by resizing the images before encoding them and sending to the renderer. 
As the result we use only 1/5 of original memory and are more than 20x faster in the CPU usage.   

The PR:
* adds a new `brave_sanitized_image_source.*` via `chrome://brave-image` that supports `&target_size=WxH` and uses faster `FastEncodeBGRASkBitmap` impl for encoding.
* adopts Brave News code to use the new handler; 
* reverts `sanitized_image_source.*` to the original state.

Because we don't use most of the code in the original `sanitized_image_source.{cc,h}` and we're adding more customizations here, it makes sense to introduce our custom version `brave_sanitized_image_source.{cc,h}` and revert the upstream file to the original state.

TODO:
* Add tests;
* Unify this with iOS version: [sanitized_image_source.mm](https://github.com/brave/brave-core/blob/master/ios/browser/ui/webui/sanitized_image_source.mm#L6)

image
before-after-cpu


components/brave_news/browser/resources/feed/Card.tsx
```diff
@@ -33,23 +33,35 @@ export const Title = styled.h3`
...
+  }, [onLoad])
+
+  let src = 'chrome://brave-image?url=' + encodeURIComponent(imgUrl);
+  if (targetHeight && targetWidth) {
+    const width = Math.round(targetWidth * window.devicePixelRatio)
+    const height = Math.round(targetHeight * window.devicePixelRatio)
+    src += `&target_size=${width}x${height}`
+  }
+
+  return <img {...rest} src={src} ref={ref as any} onError={handleError} onLoad={handleLoad} />
```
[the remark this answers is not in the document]
atuchin-m: Was before my changes in `master`

browser/ui/webui/brave_sanitized_image_source_unittest.cc
fallaciousreasoning, line 1: nit: we should rename this to `brave_sanitized_image_source_unittest.cc`

browser/ui/webui/brave_sanitized_image_source_unittest.cc
```diff
@@ -79,35 +91,48 @@ class SanitizedImageSourceTest : public testing::Test {
...
 }
 
-TEST_F(SanitizedImageSourceTest, PaddedImageIsDecoded) {
+TEST_F(BraveSanitizedImageSourceTest, ImageIsDecodedAndResized) {
+  // original size is 204x212
+  auto url = GURL("https://example.com/image.png");
+  auto result = Decode(url, kValidImage, "10x20");
+  EXPECT_TRUE(result);
+  // Check the result is PNG resized to 19x20 (to be able cut into 10x20)
+  auto bitmap = gfx::PNGCodec::Decode(base::as_byte_span(*result),
```
fallaciousreasoning: Hey this doesn't seem right - the image should be resized to fit into a 10x20 box, so it should probably be something like `10x12` (just guestimating)

components/brave_news/browser/resources/feed/Card.tsx
```diff
@@ -33,23 +33,35 @@ export const Title = styled.h3`
 
   &> a { all: unset; }
 `
-
-const HidableImage = ({ onError, ...rest }: React.DetailedHTMLProps<React.ImgHTMLAttributes<HTMLImageElement>, HTMLImageElement>) => {
+type HidabeImageProps = React.DetailedHTMLProps<React.ImgHTMLAttributes<HTMLImageElement>, HTMLImageElement> & {
+  imgUrl: string,
```
fallaciousreasoning: Hey, we should still pass `src` here as the image soure - we can pull it out in the props destructure. As-is the props accept `imgUrl` and `src` (but src will be ignored)

components/brave_news/browser/resources/feed/Card.tsx
```diff
@@ -33,23 +33,35 @@ export const Title = styled.h3`
...
+  targetHeight?: number,
+  targetWidth?: number,
+};
+const HidableImage = ({ onError, onLoad, targetHeight, targetWidth, imgUrl, ...rest }: HidabeImageProps) => {
```
fallaciousreasoning: pull `src` out here instead of imgUrl

browser/ui/webui/brave_sanitized_image_source.cc
```diff
@@ -0,0 +1,264 @@
...
+  if (!target_size.IsEmpty() && bitmap.width() > 0 && bitmap.height() > 0) {
+    // Resize to the target size, keeping the aspect ratio:
+    auto target_width = target_size.width();
+    auto target_height = target_size.height();
+    auto width_with_ratio = target_height * bitmap.width() / bitmap.height();
+    if (width_with_ratio > target_width) {
+      target_width = width_with_ratio;
+    } else {
+      target_height = target_width * bitmap.height() / bitmap.width();
+    }
```
fallaciousreasoning, line 70: This doesn't seem bad, but it doesn't quite work how I would've expected passing in a box, I'd expected the resulting image to fit in the box - instead we resize the height to fit.

fallaciousreasoning: I guess this is probably better for making sure the image isn't pixellated though

browser/ui/webui/brave_sanitized_image_source.h
```diff
@@ -0,0 +1,82 @@
...
+class Profile;
+class SkBitmap;
+
+namespace network {
+class SharedURLLoaderFactory;
+class SimpleURLLoader;
+}  // namespace network
+
+// Brave versions of SanitizedImageSource available via
+// chrome://brave-image/?url=<image_url>&target_size=<width>x<height>
```
fallaciousreasoning, line 28: I think we should note that only one dimension is guaranteed to fit in `target_size`

fallaciousreasoning (approved):

browser/ui/webui/brave_sanitized_image_source_unittest.cc
atuchin-m: in fact on NTP we also crop the image: So if we ask 10x20 the right answer is 19x20: the minimal size that can be cropped to 10x20.
Resizing it to 10x12 results in a) having a heigh gap (20-12) or b) upscaling to 19x20, then cropping.

browser/ui/webui/brave_sanitized_image_source.h
atuchin-m: It was my initial attempt. In fact we need the both dimensions if we don't want to have some gap (now in `master` we crop in the image)

components/brave_news/browser/resources/feed/Card.tsx
atuchin-m: done.

components/brave_news/browser/resources/feed/Card.tsx
atuchin-m: done

browser/ui/webui/brave_sanitized_image_source_unittest.cc
atuchin-m: Done

browser/ui/webui/brave_sanitized_image_source.cc
atuchin-m: Well, we can add cropping or clarify the explanation.
The idea of this code is to get the same pixels as we have now in `master`: we resized to a the dimensions to be perfectly cropped to the target size, without extra up/downscaling.

atuchin-m:
>One for the sanitized image source and one for the ntp changes

Discussed in DM: making 2 PRs implies supporting 2 providers simultaneously, that result in duplicating the tests in the first PR and removing the original tests in the second. It doesn't looks good.

browser/ui/webui/brave_sanitized_image_source_unittest.cc
fallaciousreasoning: realised this further down :smile:

fallaciousreasoning (approved):

browser/resources/brave_new_tab_page_refresh/components/common/safe_image.tsx
```diff
@@ -17,7 +17,7 @@ export function SafeImage(props: Props) {
     <img
       src={
         props.src
-          ? 'chrome://image?url=' + encodeURIComponent(props.src)
+          ? 'chrome://brave-image?url=' + encodeURIComponent(props.src)
```
petemill, line 20: Should this component accept the target image size?

atuchin-m: Discussed with @zenparsing : we will address this in the follow up PR.

PR #31400, opened 2025-09-24
Fix: add bottom padding to settings-box in Shields Developer mode (Fi…

teafany:
…xes brave/brave-browser#47782)

Resolves


bsclifton:
Sorry for the long turn-around time here. @zenparsing could you check this out?

zenparsing (commented):
This change doesn't appear to fix the issue. The changed file is for the `brave://settings/wallet/networks` subpage.

PR #31404, opened 2025-09-24
Brave Origin mojo for Settings UI

bbondy:
Resolves https://github.com/brave/brave-browser/issues/49605

### DIFF
--- a/browser/brave_origin/brave_origin_service_factory.cc
+++ b/browser/brave_origin/brave_origin_service_factory.cc
@@ -21,6 +21,7 @@
 #include "chrome/browser/browser_process.h"
 #include "chrome/browser/profiles/profile.h"
 #include "chrome/browser/profiles/profile_selections.h"
+#include "components/policy/core/browser/browser_policy_connector_base.h"
 
 #if BUILDFLAG(IS_WIN) || BUILDFLAG(IS_MAC) || BUILDFLAG(IS_LINUX)
 #include "brave/components/brave_rewards/core/pref_names.h"
@@ -43,11 +44,27 @@ namespace policy {
 class PolicyService;
 }
 
+namespace policy {
+class ChromeBrowserPolicyConnector;
+}
+
 namespace brave_origin {
 
 // Defined in chromium_src/chrome/browser/profiles/profile.cc to avoid a
 // circular dep on chrome/browser for the policy connector include.
 policy::PolicyService* GetPolicyServiceFromProfile(Profile* profile);
+
+policy::PolicyService* GetBrowserPolicyService() {
+  if (g_browser_process && g_browser_process->browser_policy_connector()) {
+    auto* connector = reinterpret_cast<policy::BrowserPolicyConnectorBase*>(
+        g_browser_process->browser_policy_connector());
+    // Only get policy service if we have one already or not in testing mode
+    if (connector->HasPolicyService()) {
+      return connector->GetPolicyService();
+    }
+  }
+  return nullptr;
+}
 
 namespace {
 
@@ -173,7 +190,7 @@ BraveOriginServiceFactory::BuildServiceInstanceForBrowserContext(
   std::string profile_id = GetProfileId(profile->GetPath());
   return std::make_unique<BraveOriginService>(
       g_browser_process->local_state(), profile->GetPrefs(), profile_id,
-      GetPolicyServiceFromProfile(profile));
+      GetPolicyServiceFromProfile(profile), GetBrowserPolicyService());
 }
 
 bool BraveOriginServiceFactory::ServiceIsCreatedWithBrowserContext() const {
--- a/browser/brave_origin/brave_origin_service_factory.cc
+++ b/browser/brave_origin/brave_origin_service_factory.cc
@@ -209,10 +209,9 @@ BraveOriginPolicyMap BraveOriginServiceFactory::GetBrowserPolicyDefinitions() {
       std::string brave_origin_pref_key = pref_name;
 
       browser_policy_definitions.emplace(
-          policy_key,
-          BraveOriginPolicyInfo(pref_name, metadata->origin_default_value,
-                                metadata->user_settable, policy_key,
-                                brave_origin_pref_key));
+          policy_key, BraveOriginPolicyInfo(
+                          pref_name, metadata->origin_default_value,
+                          metadata->user_settable, brave_origin_pref_key));
     }
   }
 
@@ -231,10 +230,9 @@ BraveOriginPolicyMap BraveOriginServiceFactory::GetProfilePolicyDefinitions() {
       std::string brave_origin_pref_key = pref_name;
 
       profile_policy_definitions.emplace(
-          policy_key,
-          BraveOriginPolicyInfo(pref_name, metadata->origin_default_value,
-                                metadata->user_settable, policy_key,
-                                brave_origin_pref_key));
+          policy_key, BraveOriginPolicyInfo(
+                          pref_name, metadata->origin_default_value,
+                          metadata->user_settable, brave_origin_pref_key));
     }
   }
 
--- a/browser/brave_origin/brave_origin_service_factory.cc
+++ b/browser/brave_origin/brave_origin_service_factory.cc
@@ -209,7 +209,7 @@ BraveOriginPolicyMap BraveOriginServiceFactory::GetBrowserPolicyDefinitions() {
       std::string brave_origin_pref_key = pref_name;
 
       browser_policy_definitions.emplace(
-          pref_name,
+          policy_key,
           BraveOriginPolicyInfo(pref_name, metadata->origin_default_value,
                                 metadata->user_settable, policy_key,
                                 brave_origin_pref_key));
@@ -231,7 +231,7 @@ BraveOriginPolicyMap BraveOriginServiceFactory::GetProfilePolicyDefinitions() {
       std::string brave_origin_pref_key = pref_name;
 
       profile_policy_definitions.emplace(
-          pref_name,
+          policy_key,
           BraveOriginPolicyInfo(pref_name, metadata->origin_default_value,
                                 metadata->user_settable, policy_key,
                                 brave_origin_pref_key));
--- a/components/brave_origin/brave_origin_handler.cc
+++ b/components/brave_origin/brave_origin_handler.cc
@@ -13,7 +13,7 @@ namespace brave_origin {
 BraveOriginSettingsHandlerImpl::BraveOriginSettingsHandlerImpl(
     BraveOriginService* brave_origin_service)
     : brave_origin_service_(brave_origin_service) {
-  DCHECK(brave_origin_service_);
+  CHECK(brave_origin_service_);
 }
 
 BraveOriginSettingsHandlerImpl::~BraveOriginSettingsHandlerImpl() = default;
@@ -30,28 +30,28 @@ void BraveOriginSettingsHandlerImpl::IsBraveOriginUser(
   std::move(callback).Run(is_brave_origin_user);
 }
 
-void BraveOriginSettingsHandlerImpl::IsPrefControlledByBraveOrigin(
-    const std::string& pref_name,
-    IsPrefControlledByBraveOriginCallback callback) {
+void BraveOriginSettingsHandlerImpl::IsPolicyControlledByBraveOrigin(
+    const std::string& policy_key,
+    IsPolicyControlledByBraveOriginCallback callback) {
   bool is_controlled =
-      brave_origin_service_->IsPrefControlledByBraveOrigin(pref_name);
+      brave_origin_service_->IsPolicyControlledByBraveOrigin(policy_key);
   std::move(callback).Run(is_controlled);
 }
 
 void BraveOriginSettingsHandlerImpl::GetPolicyValue(
-    const std::string& pref_name,
+    const std::string& policy_key,
     GetPolicyValueCallback callback) {
   if (!brave_origin::IsBraveOriginEnabled()) {
     std::move(callback).Run(std::nullopt);
     return;
   }
 
-  std::optional<bool> value = brave_origin_service_->GetPolicyValue(pref_name);
+  std::optional<bool> value = brave_origin_service_->GetPolicyValue(policy_key);
   std::move(callback).Run(value);
 }
 
 void BraveOriginSettingsHandlerImpl::SetPolicyValue(
-    const std::string& pref_name,
+    const std::string& policy_key,
     bool value,
     SetPolicyValueCallback callback) {
   if (!brave_origin::IsBraveOriginEnabled()) {
@@ -59,7 +59,7 @@ void BraveOriginSettingsHandlerImpl::SetPolicyValue(
     return;
   }
 
-  bool success = brave_origin_service_->SetPolicyValue(pref_name, value);
+  bool success = brave_origin_service_->SetPolicyValue(policy_key, value);
   std::move(callback).Run(success);
 }
 
--- a/components/brave_origin/brave_origin_handler.cc
+++ b/components/brave_origin/brave_origin_handler.cc
@@ -38,55 +38,28 @@ void BraveOriginSettingsHandlerImpl::IsPrefControlledByBraveOrigin(
   std::move(callback).Run(is_controlled);
 }
 
-void BraveOriginSettingsHandlerImpl::GetBrowserPolicyValue(
+void BraveOriginSettingsHandlerImpl::GetPolicyValue(
     const std::string& pref_name,
-    GetBrowserPolicyValueCallback callback) {
+    GetPolicyValueCallback callback) {
   if (!brave_origin::IsBraveOriginEnabled()) {
     std::move(callback).Run(std::nullopt);
     return;
   }
 
-  std::optional<bool> value =
-      brave_origin_service_->GetBrowserPolicyValue(pref_name);
+  std::optional<bool> value = brave_origin_service_->GetPolicyValue(pref_name);
   std::move(callback).Run(value);
 }
 
-void BraveOriginSettingsHandlerImpl::GetProfilePolicyValue(
-    const std::string& pref_name,
-    GetProfilePolicyValueCallback callback) {
-  if (!brave_origin::IsBraveOriginEnabled()) {
-    std::move(callback).Run(std::nullopt);
-    return;
-  }
-
-  std::optional<bool> value =
-      brave_origin_service_->GetProfilePolicyValue(pref_name);
-  std::move(callback).Run(value);
-}
-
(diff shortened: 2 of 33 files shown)


chromium_src/chrome/browser/policy/chrome_browser_policy_connector.cc
```diff
@@ -37,3 +37,14 @@ ChromeBrowserPolicyConnector::CreatePolicyProviders() {
 }
 
 }  // namespace policy
+
+// This is to avoid a circular dep on chrome/browser in the Factory for
+// the brave_origin keyed service.
+namespace brave_origin {
+policy::PolicyService* GetBrowserPolicyService() {
+  if (g_browser_process && g_browser_process->browser_policy_connector()) {
+    return g_browser_process->browser_policy_connector()->GetPolicyService();
```
DJAndries: I'm wondering if we need a chromium_src change here. It looks like `GetPolicyService` is defined in `components/policy/core/browser/browser_policy_connector_base.h`. Could we just include that header and call the method directly?

components/brave_origin/common/mojom/brave_origin_settings.mojom
```diff
@@ -0,0 +1,24 @@
...
+  // Check if the current user is a Brave Origin user
+  IsBraveOriginUser() => (bool is_brave_origin_user);
+
+  // Check if a specific preference is controlled by Brave Origin
+  IsPrefControlledByBraveOrigin(string pref_name) =>
+      (bool is_controlled);
+
+  // Get the current value of a Brave Origin controlled preference
+  GetBrowserPolicyValue(string pref_name) => (bool? value);
+  GetProfilePolicyValue(string pref_name) => (bool? value);
```
DJAndries: It's not clear to me how the UI would know whether a policy is for the browser or profile. It also seems odd that `IsPrefControlledByBraveOrigin` works for both browser and profile policies, while these functions are separate. IMO, it would be nice if we combined these functions (so we have `GetPolicyValue` and `SetPolicyValue`) so we could simplify the UI logic. Thoughts?

bbondy: The UI logic can definitely know if it is a browser or profile pref, because the UI dev can look at where the pref lives. It's not something that is dynamic or a guess. 

There were previous requested changes that led to the split of browser vs profile APIs, but we're here now with a split API. I think there is no difference though for that API, you just want to know if a browser is Brave Origin controlled or not, where it is controlled doesn't matter.  For the Get/Set functions though, it does matter.

chromium_src/chrome/browser/policy/chrome_browser_policy_connector.cc
bbondy: `browser_policy_connector()` it returns a `ChromeBrowserPolicyConnector` which is defined in `chrome/browser/policy/chrome_browser_policy_connector.cc` which is in `chrome/browser/BUILD.gn`.  But you're right that `GetPolicyService` is part of the base class. So I think I can just include the base class and forward declare `ChromeBrowserPolicyConnector` and cast it accordingly and I should be able to avoid it.

mkarolin (approved):
`DEPS`++

components/brave_origin/common/mojom/brave_origin_settings.mojom
bbondy: FYI: I'm updating to use 1 API for Get* and Set* from the service and the mojo interface. Will update when done.

chromium_src/chrome/browser/policy/chrome_browser_policy_connector.cc
bbondy: Fixed in https://github.com/brave/brave-core/pull/31404/commits/335693769a2d39730af3335371e59ce19b5f1cf1

components/brave_origin/common/mojom/brave_origin_settings.mojom
bbondy: Fixed in https://github.com/brave/brave-core/pull/31404/commits/60ad806c3fd511222c385a40faf895b2847291b2

components/brave_origin/brave_origin_policy_manager.cc
```diff
@@ -78,6 +70,16 @@ std::optional<bool> BraveOriginPolicyManager::GetPolicyValue(
   return GetPolicyValueInternal(*policy_info, policies_dict, profile_id);
 }
 
+bool BraveOriginPolicyManager::IsBrowserPolicy(
+    std::string_view pref_name) const {
+  return base::FindOrNull(browser_policy_definitions_, pref_name) != nullptr;
```
DJAndries: Could this be `browser_policy_definitions_.contains(pref_name)`? This also applies to the method below

components/brave_origin/brave_origin_service.cc
```diff
@@ -70,61 +88,50 @@ bool BraveOriginService::SetBrowserPolicyValue(std::string_view pref_name,
...
-  // We don't want a user to have a value set in local state if the value is not
-  // user settable as they would be stuck with that value if they stopped being
-  // a Brave Origin user
-  if (!pref_info->user_settable && value == pref_info->default_value) {
-    local_state_->ClearPref(pref_info->pref_name);
-  } else {
-    local_state_->SetBoolean(pref_info->pref_name, value);
+  PrefService* target_prefs = nullptr;
+  if (manager->IsBrowserPolicy(pref_info->pref_name)) {
+    manager->SetPolicyValue(pref_name, value);
```
DJAndries, line 93: What do you think about having `SetPolicyValue` detect whether the policy is browser or profile-based internally, so callers don't have to call `IsBrowserPolicy`/`IsProfilePolicy`?

components/brave_origin/brave_origin_service.cc
```diff
@@ -70,61 +88,50 @@ bool BraveOriginService::SetBrowserPolicyValue(std::string_view pref_name,
...
 
-  return true;
-}
-
-bool BraveOriginService::SetProfilePolicyValue(std::string_view pref_name,
-                                               bool value) {
-  if (!IsBraveOriginEnabled()) {
-    return false;
+  // Also set the corresponding pref value
+  if (target_prefs) {
```
DJAndries: Would it be appropriate to `CHECK(target_prefs)` instead?

components/brave_origin/brave_origin_service.cc
```diff
@@ -70,61 +88,50 @@ bool BraveOriginService::SetBrowserPolicyValue(std::string_view pref_name,
...
 
-  // Get policy info to access pref_key and user_settable
-  auto* manager = BraveOriginPolicyManager::GetInstance();
-  const BraveOriginPolicyInfo* pref_info = manager->GetPrefInfo(pref_name);
-  if (!pref_info) {
-    return false;
+  // Reload browser policies to ensure the new policy value takes effect
+  if (manager->IsBrowserPolicy(pref_info->pref_name) &&
+      browser_policy_service_) {
+    browser_policy_service_->RefreshPolicies(
```
DJAndries: Would it make sense to notify the providers via the observer interface so they can just call `UpdatePolicy` instead of refreshing all policies? IIRC, this might be more efficient as it shouldn't prompt non-Origin policy providers for a refresh.

components/brave_origin/brave_origin_utils.cc
```diff
@@ -19,18 +19,14 @@ bool IsBraveOriginEnabled() {
...
+std::string GetBraveOriginPrefKey(const BraveOriginPolicyInfo& pref_info,
+                                  std::optional<std::string_view> profile_id) {
+  if (!profile_id.has_value()) {
+    return pref_info.policy_key;
+  }
 
-std::string GetBraveOriginProfilePrefKey(const BraveOriginPolicyInfo& pref_info,
-                                         std::string_view profile_id) {
   // For profile prefs, use profile_id.policy_key format
-  CHECK(!profile_id.empty())
```
DJAndries, line 31: Should we keep this check?

components/brave_origin/brave_origin_policy_manager.cc
bbondy: Fixed in https://github.com/brave/brave-core/pull/31404/commits/88e5862230f679d6efcb230d8a66d0f656e3d890

components/brave_origin/brave_origin_service.cc
bbondy: Fixed in https://github.com/brave/brave-core/pull/31404/commits/88e5862230f679d6efcb230d8a66d0f656e3d890

components/brave_origin/brave_origin_service.cc
bbondy: I forgot I said I was going to do it that way. Updated in https://github.com/brave/brave-core/pull/31404/commits/88e5862230f679d6efcb230d8a66d0f656e3d890

components/brave_origin/brave_origin_utils.cc
bbondy: Fixed  in https://github.com/brave/brave-core/pull/31404/commits/88e5862230f679d6efcb230d8a66d0f656e3d890

components/brave_origin/brave_origin_policy_manager.cc
```diff
@@ -149,16 +131,24 @@ void BraveOriginPolicyManager::SetPolicyValueInternal(
...
-  std::string key =
-      profile_id.has_value()
-          ? GetBraveOriginProfilePrefKey(*policy_info, *profile_id)
-          : GetBraveOriginBrowserPrefKey(*policy_info);
+  std::string key = GetBraveOriginPrefKey(*policy_info, profile_id);
   update->Set(key, value);
+
+  // Notify observers of the policy change
+  if (profile_id.has_value()) {
+    for (auto& observer : observers_) {
```
DJAndries: I think we may be able to call `observers_.Notify` instead of iterating through the list. If so, we should update all `ObserverList` notifications to use this pattern.

bbondy: Good call. Updated in 9653fac8a79392faa102992c3bb370bc3b9f1e59

components/brave_origin/brave_origin_handler.cc
```diff
@@ -0,0 +1,66 @@
...
+
+#include "brave/components/brave_origin/brave_origin_service.h"
+#include "brave/components/brave_origin/brave_origin_utils.h"
+
+namespace brave_origin {
+
+BraveOriginSettingsHandlerImpl::BraveOriginSettingsHandlerImpl(
+    BraveOriginService* brave_origin_service)
+    : brave_origin_service_(brave_origin_service) {
+  DCHECK(brave_origin_service_);
```
DJAndries: Should this just be `CHECK`?

components/brave_origin/brave_origin_service.cc
```diff
@@ -70,61 +86,35 @@ bool BraveOriginService::SetBrowserPolicyValue(std::string_view pref_name,
...
-  // If not user-settable and value equals default, clear the pref instead
-  // We don't want a user to have a value set in local state if the value is not
-  // user settable as they would be stuck with that value if they stopped being
-  // a Brave Origin user
-  if (!pref_info->user_settable && value == pref_info->default_value) {
-    local_state_->ClearPref(pref_info->pref_name);
-  } else {
-    local_state_->SetBoolean(pref_info->pref_name, value);
+  PrefService* target_prefs = nullptr;
+  if (manager->IsBrowserPolicy(pref_info->pref_name)) {
```
DJAndries: In the last PR I suggested keying the policy maps by policy key rather than pref name. Is my assessment correct that we only need the pref name for updating the `target_prefs` with an explicit value? If so, I think the implementing the suggestion could be fruitful since the policy key seems to be more frequently referenced compared to the pref name. It could save us from having to perform frequent map lookups to get the key. This suggestion would also include changing the mojom interface to use policy keys as well.

bbondy: I don't see why it would be more efficient. The API as it was keyed things by pref names. And the lookups looked up by pref name. The only time the policy key is used is in setting up the policies in the policy service, but in those cases, it gets all values, so there's no lookups there.

You're suggesting to update to policy values, and keying by policy.  
The lookups never looked up the key names repeatedly to get a policy key, because it didn't need that.  They were keyed by pref name, and they were looked up by pref name. 

I'll update, but I don't think it's gaining us anything. Just a different parameter and key. The user of the API will just need to use a policy name instead of a key.

components/brave_origin/brave_origin_handler.cc
bbondy: Updated here: https://github.com/brave/brave-core/pull/31404/commits/be0ff9629b641943ebabea092589d64a312b6292

components/brave_origin/brave_origin_service.cc
bbondy: Updated to the new API parameter of using policy key instead of pref key and updated the dict accordingly here: https://github.com/brave/brave-core/pull/31404/commits/be0ff9629b641943ebabea092589d64a312b6292 and retested all. Should be functionally equivalent and perf equivalent but maybe has a nicer api signature? not sure.

components/brave_origin/brave_origin_utils.cc
```diff
@@ -19,18 +19,15 @@ bool IsBraveOriginEnabled() {
   return base::FeatureList::IsEnabled(features::kBraveOrigin);
 }
 
-std::string GetBraveOriginBrowserPrefKey(
-    const BraveOriginPolicyInfo& pref_info) {
-  // For browser prefs, use policy_key directly
-  return pref_info.policy_key;
-}
+std::string GetBraveOriginPrefKey(const BraveOriginPolicyInfo& pref_info,
```
DJAndries: nit: it looks like this could be simplified to take a policy key instead of policy info

components/brave_origin/brave_origin_service.cc
```diff
@@ -19,112 +19,104 @@
...
 
-BraveOriginService::BraveOriginService(PrefService* local_state,
-                                       PrefService* profile_prefs,
-                                       std::string_view profile_id,
-                                       policy::PolicyService* policy_service)
+namespace {
+// Helper function to check if a policy is controlled by BraveOrigin in a given
+// policy service
+bool IsPolicyControlledByBraveOrigin(policy::PolicyService* policy_service,
+                                     const BraveOriginPolicyInfo* pref_info) {
```
DJAndries: nit: looks like this could also be simplified to take the policy key instead

components/brave_origin/brave_origin_policy_manager.cc
```diff
@@ -106,66 +106,51 @@ PoliciesEnabledMap BraveOriginPolicyManager::GetAllProfilePolicies(
...
     return;
   }
 
   // Update the value in the dictionary
   ScopedDictPrefUpdate update(local_state_, kBraveOriginPolicies);
-  std::string key =
-      profile_id.has_value()
-          ? GetBraveOriginProfilePrefKey(*policy_info, *profile_id)
-          : GetBraveOriginBrowserPrefKey(*policy_info);
+  std::string key = GetBraveOriginPrefKey(*policy_info, profile_id);
```
DJAndries: if the `GetBraveOriginPrefKey` signature is simplified, then we might not need to `GetPolicyInfo` call above

components/brave_origin/brave_origin_service.cc
DJAndries: I made the suggestion to open up opportunities for code simplification. Left some relevant comments below

components/brave_origin/brave_origin_utils.cc
bbondy: Updated in https://github.com/brave/brave-core/pull/31404/commits/020d2278b7da642bbca618c6d9d06851bbb638f2

components/brave_origin/brave_origin_service.cc
bbondy: Updated in https://github.com/brave/brave-core/pull/31404/commits/020d2278b7da642bbca618c6d9d06851bbb638f2

components/brave_origin/brave_origin_policy_manager.cc
bbondy: Updated in https://github.com/brave/brave-core/pull/31404/commits/020d2278b7da642bbca618c6d9d06851bbb638f2

DJAndries: It looks like we still call `GetPolicyInfo`, but we already call `GetPolicyInfo` in `BraveOriginService:::SetPolicyValue`. Do we still want to keep this call in the policy manager?

browser/brave_origin/brave_origin_service_factory.cc
```diff
@@ -214,7 +231,7 @@ BraveOriginPolicyMap BraveOriginServiceFactory::GetProfilePolicyDefinitions() {
       std::string brave_origin_pref_key = pref_name;
 
       profile_policy_definitions.emplace(
-          pref_name,
+          policy_key,
           BraveOriginPolicyInfo(pref_name, metadata->origin_default_value,
                                 metadata->user_settable, policy_key,
```
DJAndries: nit: i think we can remove policy key from the policy info struct. thoughts?

bbondy: I think we can

bbondy: Removed in the latest

bbondy:
If any objection to the sources.gni change I'll do in another commit. I'd like to actually split that file out into other gni files for common uses like this so it doesn't trigger codeowners.

components/brave_origin/brave_origin_policy_manager.cc
bbondy: Discussed on Slack but the check is needed here because you can't be sure all callers in the future will check that before calling this API.

browser/brave_origin/brave_origin_service_factory.cc
```diff
@@ -43,12 +44,28 @@ namespace policy {
...
+}
+
 namespace brave_origin {
 
 // Defined in chromium_src/chrome/browser/profiles/profile.cc to avoid a
 // circular dep on chrome/browser for the policy connector include.
 policy::PolicyService* GetPolicyServiceFromProfile(Profile* profile);
 
+policy::PolicyService* GetBrowserPolicyService() {
+  if (g_browser_process && g_browser_process->browser_policy_connector()) {
```
bridiver, line 58: this should not be null unless `CHECK_IS_TEST();` so should be
```
if (!g_browser_process) {
  CHECK_IS_TEST();
   return nullptr;
}
```
in all other cases `g_browser_process` is an invariant here

bbondy: This is fixed here https://github.com/brave/brave-core/pull/31870

PR #31414, opened 2025-09-24
[CodeHealth][Speedreader] use dependency injection for speedreader service in tab helper

bridiver:
Resolves https://github.com/brave/brave-browser/issues/49700

### DIFF
--- a/browser/ui/speedreader/speedreader_tab_helper.h
+++ b/browser/ui/speedreader/speedreader_tab_helper.h
@@ -107,6 +107,7 @@ class SpeedreaderTabHelper
  private:
   friend class content::WebContentsUserData<SpeedreaderTabHelper>;
   explicit SpeedreaderTabHelper(content::WebContents* web_contents,
+                                SpeedreaderService& speedreader_service,
                                 SpeedreaderRewriterService* rewriter_service);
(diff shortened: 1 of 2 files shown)


bridiver:
unrelated test failures

browser/ui/speedreader/speedreader_tab_helper.h
```diff
@@ -187,8 +186,11 @@ class SpeedreaderTabHelper
 
   DistillState distill_state_{DistillStates::ViewOriginal()};
 
+  raw_ref<SpeedreaderService> speedreader_service_;
+
   const raw_ptr<SpeedreaderRewriterService> rewriter_service_ =
       nullptr;  // NOT OWNED
```
fallaciousreasoning, line 192: unrelated, but why do we have a `NOT OWNED` commend on a `raw_ptr`?

fallaciousreasoning (approved):
lgtm

browser/ui/speedreader/speedreader_tab_helper.h
bridiver: probably it was converted from a regular pointer and not removed?

PR #31415, opened 2025-09-24
[Speedreader] cleanup pref/feature flag naming

bridiver:
Resolves https://github.com/brave/brave-browser/issues/49745

This PR does not contain any functional changes, it is only renaming/non-functional refactoring (reviewers should verify no accidental changes)


components/speedreader/speedreader_service.cc
```diff
@@ -106,12 +94,12 @@ void SpeedreaderService::RemoveObserver(Observer* observer) {
...
+bool SpeedreaderService::IsEnabled() {
+  return prefs_->GetBoolean(kSpeedreaderEnabled);
 }
 
 bool SpeedreaderService::IsEnabledForAllSites() {
-  if (!IsFeatureEnabled()) {
+  if (!IsEnabled()) {
     return false;
   }
   return prefs_->GetBoolean(kSpeedreaderPrefEnabledForAllSites);
```
bbondy: Just noting this is a confusing name for a setting that automatically loads speedreader when eligible pages load.

components/speedreader/common/features.cc
```diff
@@ -15,6 +15,10 @@ BASE_FEATURE(kSpeedreaderFeature,
              base::FEATURE_ENABLED_BY_DEFAULT);
 #endif
 
+bool IsSpeedreaderEnabled() {
```
bbondy: We should probably use a different name for this function like `IsSpeedreaderFeatureFlagEnabled` but I think it's going away in a different PR

bridiver: this code was a bit weird before, features.h just used speedreader, but then this method was inside features namespace. I think I'm just going to move all of feautres.h into speedreader::features

bridiver: actually I'm going to remove this method which is what I originally intended to do and forgot :)

bridiver:
unrelated test error on android and renderer changes are trivial

PR #31416, opened 2025-09-24
[Leo] Smart Modes

yrliou:
Resolves https://github.com/brave/brave-browser/issues/49622

Backend PR: https://github.com/brave/brave-core/pull/31422

This PR is an initial PR for Smart Modes feature, which allows users to create custom prompt shortcuts and use them during conversations (limited to 1 shortcut per user turn).
The goal is to start with smallest implementation footprint first so slash tools are not migrated in this PR.
We also limit the smart modes to be only at the beginning of user input to have a simplified version of highlighting working first before https://github.com/brave/brave-browser/issues/48610 is resolved.

With this PR, we can:
- `Save as smart mode` on submitted user prompts, which would open the smart mode dialog for adding a new smart mode with the user prompt populated.
- Can use a smart mode inside input box like `/shortcut with some added context`, which will add a message right before the user message to ask agent to interpret /shortcut as the saved user prompts, works for both Leo and custom models.

We haven't match exactly with https://www.figma.com/design/m0Gdbf0wtqyfEFGm32VLLc/%F0%9F%94%84-Leo--IN-PROGRESS-?node-id=14222-277272&p=f&m=dev, we would need to have a UI follow-up issue/PR to fix it after 1) migrating slash tools and 2) inputs at anywhere is supported.

https://github.com/user-attachments/assets/21263c11-7508-4c6c-a4f2-0749b73b3775
(Note that highlight in the input box in this video is no longer added by this PR.)

Disclaimer: Claude Code was used for the most of frontend changes, I tried my best self-reviewing it but could still be not meeting the best frontend practices here and there.

Test plan:
- Type something that you would like to re-use in the future, "create a playlist with 10 songs" for example, and submit this user prompt.
- In the three-dot menu next to the submitted prompt, click on `Save as smart mode`.
- After saving it, type your defined shortcut, `/playlist` for example, in the input box with some additional context, agent should do the corresponding action with the additional context (ex: /playlist from 2000).

### DIFF
--- a/components/ai_chat/resources/page/components/smart_mode_dialog/index.tsx
+++ b/components/ai_chat/resources/page/components/smart_mode_dialog/index.tsx
@@ -31,6 +31,9 @@ export default function SmartModeDialog() {
     if (!value.trim()) {
       return getLocale(S.CHAT_UI_SHORTCUT_REQUIRED_ERROR)
     }
+    if (value.trim().length > 64) {
+      return getLocale(S.CHAT_UI_SHORTCUT_TOO_LONG_ERROR)
+    }
     if (!/^[a-zA-Z0-9_-]+$/.test(value.trim())) {
       return getLocale(S.CHAT_UI_SHORTCUT_FORMAT_ERROR)
     }
--- a/components/resources/ai_chat_ui_strings.grdp
+++ b/components/resources/ai_chat_ui_strings.grdp
@@ -690,6 +690,9 @@
   <message name="IDS_CHAT_UI_SHORTCUT_FORMAT_ERROR" desc="Error message when shortcut format is invalid in smart mode dialog" formatter_data="webui=AiChat">
     Shortcut can only contain letters, numbers, underscore, and hyphen
   </message>
+  <message name="IDS_CHAT_UI_SHORTCUT_TOO_LONG_ERROR" desc="Error message when shortcut exceeds maximum length in smart mode dialog" formatter_data="webui=AiChat">
+    Shortcut must be 64 characters or less
+  </message>
   <message name="IDS_CHAT_UI_SHORTCUT_DUPLICATE_ERROR" desc="Error message when shortcut already exists" formatter_data="webui=AiChat">
     Shortcut already exists
   </message>


fallaciousreasoning:
Hey, just noting that the highlighting is breaking scrolls for long inputs:

<details>
<summary>input</summary>
/translate this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of txt this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text this really long big of text
</details>

[Screencast From 2025-09-25 15-46-32.webm](https://github.com/user-attachments/assets/707fbba8-e7e7-4a5e-b31a-92e5825000ee)

yrliou:
Split backend change to https://github.com/brave/brave-core/pull/31422, please focus on reviewing the frontend part of this PR, thanks.

components/ai_chat/core/common/mojom/untrusted_frame.mojom
```diff
@@ -41,6 +41,9 @@ interface ParentUIFrame {
   // Otherwise the menu can sometimes be overlapped by the Suggested question
   // buttons.
   RegenerateAnswerMenuIsOpen(bool is_open);
+
+  // Show the smart mode creation dialog with the provided prompt text
+  ShowSmartModeDialog(string prompt);
```
petemill, line 46: nit: `CreateSmartMode(string prompt)` (exposing that it's a dialog doesn't seem neccessary)

components/ai_chat/resources/page/stories/components_panel.tsx
```diff
@@ -1374,3 +1484,50 @@ export const _MemoryEvents: Story = {
     useMemoryHistory: true,
   },
 }
+
+export const _SmartModeDialogNew: Story = {
```
fallaciousreasoning, line 1488: hey, rather than adding this here, I think we should add a new story in the `smart_mode_dialog` folder (probably called `smart_mode_dialog.stories.tsx` or something)

components/ai_chat/resources/untrusted_conversation_frame/components/conversation_entries/index.tsx
```diff
@@ -34,6 +34,27 @@ import {
 import useConversationEventClipboardCopyHandler from './use_conversation_event_clipboard_copy_handler'
 import styles from './style.module.scss'
 
+// Utility function to highlight smart mode shortcuts in text
+const highlightSmartModeText = (
```
fallaciousreasoning, line 38: I think this should be 
```suggestion
const maybeHighlightSmartModeText = (
```

as it won't necessarily highlight the text

components/ai_chat/resources/page/components/filter_menu/tools_menu.tsx
```diff
@@ -4,43 +4,113 @@
...
+  if (entry.details) {
+    return matches(query, entry.details.label)
+  } else if (entry.smartMode) {
+    return matches(query, entry.smartMode.shortcut)
+  }
+  return false
 }
 
 export default function ToolsMenu(props: ToolsMenuProps) {
+  const { smartModes } = useAIChat()
```
fallaciousreasoning, line 46: hey, so I think we probably want to `Pick` this off the context like we do for `categories`.

This component, ideally doesn't depend on the `AIChatContext` because it might be embedded in other places (NTP, Rewriter Dialog ect).

components/ai_chat/resources/page/components/filter_menu/tools_menu.tsx
```diff
@@ -4,43 +4,113 @@
...
+  'categories' | 'isOpen' | 'setIsOpen' | 'query'
+>
 
-function matchesQuery(query: string, entry: ActionEntry) {
-  return entry.details && matches(query, entry.details.label)
+// Extended entry type that can include smart modes before we migrate
+// existing slash tools (actions) to smart modes.
+type ExtendedActionEntry = ActionEntry & {
+  smartMode?: SmartMode
+}
```
fallaciousreasoning, line 34: This should be an `|` not an `&` - `&` implies it will have all the properties of both, whereas this will only ever be one or the other

components/ai_chat/resources/page/components/filter_menu/tools_menu.tsx
```diff
@@ -4,43 +4,113 @@
...
 import FilterMenu, { Props } from './filter_menu'
 import { matches } from './query'
 import styles from './style.module.scss'
+import { getLocale } from '$web-common/locale'
+import { useAIChat } from '../../state/ai_chat_context'
 
 type ToolsMenuProps = {
   handleClick: (type: ActionType) => void
-} & Pick<Props<ActionEntry>, 'categories' | 'isOpen' | 'setIsOpen' | 'query'>
+  handleSmartModeClick?: (smartMode: SmartMode) => void
```
fallaciousreasoning, line 23: Hey, I don't think we need these separate `click` handlers - `handleClick` should just take an `item: ActionType | SmartMode`

components/ai_chat/resources/page/components/filter_menu/tools_menu.tsx
```diff
@@ -4,43 +4,113 @@
...
+  handleSmartModeEdit?: (smartMode: SmartMode) => void
```
fallaciousreasoning, line 24: in what scenario would this be `undefined`?

fallaciousreasoning: If this is for the rewriter_ui, let's just pass in an empty function there

components/ai_chat/resources/page/components/filter_menu/tools_menu.tsx
```diff
@@ -4,43 +4,113 @@
...
+      smartModes.length > 0
+      && props.handleSmartModeClick
+      && props.handleSmartModeEdit
+    ) {
+      const smartModeEntries: ExtendedActionEntry[] = smartModes.map(
+        (mode: SmartMode) => ({
+          smartMode: mode,
+          subheading: undefined,
+          details: undefined,
+        }),
```
fallaciousreasoning, line 62: if we use the `|` we could probably just return the `mode` here, I think

components/ai_chat/resources/page/components/filter_menu/tools_menu.tsx
```diff
@@ -4,43 +4,113 @@
...
+      )
+
+      categories.push({
+        category: getLocale(S.CHAT_UI_SMART_MODES_GROUP),
+        entries: smartModeEntries,
+      })
```
fallaciousreasoning, line 68: so smart modes should always show first?

components/ai_chat/resources/page/components/filter_menu/tools_menu.tsx
```diff
@@ -4,43 +4,113 @@
...
+    }
+
+    categories.push(...props.categories)
+
+    return categories
+  }, [smartModes])
```
fallaciousreasoning, line 74: this also needs to depend on `props.categories`

fallaciousreasoning: And probably `props.handleSmartModeClick` and `props.handleSmartModeEdit` if they actually can be falsey, otherwise we're going to have weird issues if they aren't set on initial render.

Unfortunately if it does need to depend on them we'll probably need to wrap them in a useCallback or it'll rerun this every render.

In my ideal world these just wouldn't be falsey

components/ai_chat/resources/page/components/filter_menu/tools_menu.tsx
```diff
@@ -4,43 +4,113 @@
...
-      categories={props.categories}
+      categories={combinedCategories}
       isOpen={props.isOpen}
       setIsOpen={props.setIsOpen}
       query={props.query}
       matchesQuery={matchesQuery}
     >
-      {(item) =>
-        !item.details ? (
+      {(item: ExtendedActionEntry) =>
```
fallaciousreasoning, line 84: I think we shouldn't need this type here - possibly the type of `combinedCategories` isn't being inferred properly.

If that's the case, we should probably add a type param to the `useMemo` above

components/ai_chat/resources/page/components/filter_menu/tools_menu.tsx
```diff
@@ -4,43 +4,113 @@
...
-        ) : (
+        ) : item.details ? (
           <leo-menu-item
             key={item.details.type}
             onClick={() => props.handleClick(item.details!.type)}
           >
             {item.details.label}
           </leo-menu-item>
-        )
+        ) : item.smartMode ? (
```
fallaciousreasoning, line 94: hmmm coould we improve the nested ternary formatting? Its a bit hard to follow here:

```ts
  return a
    ? b
      ? c
      : d
    : e
```

alternatively, we could make this a function

components/ai_chat/resources/page/components/filter_menu/tools_menu.tsx
```diff
@@ -4,43 +4,113 @@
...
+            key={item.smartMode.id}
+            onClick={() => props.handleSmartModeClick?.(item.smartMode!)}
+          >
+            <div className={styles.smartModeWithEdit}>
+              <span className={styles.shortcut}>{item.smartMode.shortcut}</span>
+              <Button
+                fab
+                kind='plain-faint'
+                className={styles.editIcon}
+                onClick={(e) => {
```
fallaciousreasoning, line 105: nit: unused param
```suggestion
                onClick={() => {
```

components/ai_chat/resources/page/components/filter_menu/tools_menu.tsx
```diff
@@ -4,43 +4,113 @@
...
+                  props.handleSmartModeEdit?.(item.smartMode!)
+                }}
+              >
+                <Icon name='edit-pencil' />
+              </Button>
+            </div>
+          </leo-menu-item>
+        ) : null
```
fallaciousreasoning, line 113: Could we actually return `null` here? I think this is a result of our type definition for `ExtendedActionEntry`

components/ai_chat/resources/page/components/filter_menu/tools_menu.test.tsx
```diff
@@ -7,13 +7,42 @@ import '$test-utils/disable_custom_elements'
...
+
+// Mock locale function
+jest.mock('$web-common/locale', () => ({
+  getLocale: (key: string) => {
+    const mockStrings: { [key: string]: string } = {
+      CHAT_UI_SMART_MODES_GROUP: 'Smart Modes',
+    }
+    return mockStrings[key] || key
+  },
+}))
```
fallaciousreasoning, line 24: hey, this should automatically be mocked in tests. Additionally, rather than adding a translation here, we should just use the key `CHAT_UI_SMART_MODES_GROUP` in the tests

components/ai_chat/resources/page/components/filter_menu/tools_menu.test.tsx
```diff
@@ -60,41 +89,164 @@ describe('tools_menu', () => {
...
-    expect(getByText('subheading 1')).toBeInTheDocument()
-    expect(getByText('subheading 2')).toBeInTheDocument()
-    expect(getByText('entry 1')).toBeInTheDocument()
-    expect(getByText('entry 2')).toBeInTheDocument()
-    expect(getByText('entry 3')).toBeInTheDocument()
-
-    expect(container.querySelector('leo-buttonmenu')).toHaveProperty(
-      'isOpen',
-      true,
+  const renderWithContext = (
```
fallaciousreasoning, line 92: hey so this shouldn't be using the context (see https://github.com/brave/brave-core/pull/31416#discussion_r2389442088)

components/ai_chat/resources/page/components/filter_menu/tools_menu.test.tsx
```diff
@@ -60,41 +89,164 @@ describe('tools_menu', () => {
...
+    return render(
+      <AIChatReactContext.Provider value={aiChatContext as any}>
+        <ToolsMenu {...props} />
+      </AIChatReactContext.Provider>,
     )
+  }
+
+  beforeEach(() => {
+    jest.clearAllMocks()
   })
```
fallaciousreasoning, line 107: do we actually need this?

components/ai_chat/resources/page/components/filter_menu/tools_menu.test.tsx
```diff
@@ -60,41 +89,164 @@ describe('tools_menu', () => {
...
-    const { queryByText } = render(
-      <ToolsMenu
-        {...defaultProps}
-        query='e'
-      />,
-    )
-    expect(queryByText('subheading 1')).not.toBeInTheDocument()
-    expect(queryByText('subheading 2')).not.toBeInTheDocument()
+  describe('without smart modes', () => {
+    it('should render categories without smart modes', () => {
```
fallaciousreasoning, line 109: this is just checking that the smart modes category is hidden when there are no smart modes right?

Maybe 
```suggestion
    it('smart modes category is hidden when there are no smart modes', () => {
```

components/ai_chat/resources/page/components/filter_menu/tools_menu.test.tsx
```diff
@@ -60,41 +89,164 @@ describe('tools_menu', () => {
...
-    )
-    expect(queryByText('category 1')).not.toBeInTheDocument()
-    expect(queryByText('category 2')).toBeInTheDocument()
+  describe('with smart modes', () => {
+    it('should render smart modes category first', () => {
+      const { getByText } = renderWithContext()
+      expect(getByText('Smart Modes')).toBeInTheDocument()
+      expect(getByText('translate')).toBeInTheDocument()
+      expect(getByText('explain')).toBeInTheDocument()
+    })
```
fallaciousreasoning, line 150: we're not actually checking ordering here

components/ai_chat/resources/page/components/filter_menu/tools_menu.test.tsx
```diff
@@ -60,41 +89,164 @@ describe('tools_menu', () => {
...
+      const menuItems = Array.from(container.querySelectorAll('leo-menu-item'))
+      const categoryHeaders = Array.from(
+        container.querySelectorAll('.menuSectionTitle, .menuSubtitle'),
+      )
+
+      // Smart Modes should appear first, then other categories
+      expect(categoryHeaders[0]).toHaveTextContent('Smart Modes')
+      expect(menuItems[0]).toHaveTextContent('translate')
+      expect(menuItems[1]).toHaveTextContent('explain')
+    })
```
fallaciousreasoning, line 242: we don't check ordering here

components/ai_chat/resources/page/components/filter_menu/tools_menu.test.tsx
```diff
@@ -60,41 +89,164 @@ describe('tools_menu', () => {
...
+
+    it('should handle empty smart modes array', () => {
+      const { queryByText } = renderWithContext([])
+
+      expect(queryByText('Smart Modes')).not.toBeInTheDocument()
+      expect(queryByText('translate')).not.toBeInTheDocument()
+      expect(queryByText('explain')).not.toBeInTheDocument()
+    })
```
fallaciousreasoning, line 250: I think we have a very similar test https://github.com/brave/brave-core/pull/31416#discussion_r2389477012?

components/ai_chat/resources/page/components/filter_menu/tools_menu.test.tsx
```diff
@@ -60,41 +89,164 @@ describe('tools_menu', () => {
...
+        query: 'translate',
+      })
+
+      expect(getByText('Smart Modes')).toBeInTheDocument()
+      expect(getByText('translate')).toBeInTheDocument()
+      expect(queryByText('category 1')).not.toBeInTheDocument()
+      expect(queryByText('category 2')).not.toBeInTheDocument()
+    })
+
+    it('should not render smart modes section when no handlers', () => {
```
fallaciousreasoning, line 215: I think this is the only time this code path is exercised (testonly)

components/ai_chat/resources/page/components/filter_menu/tools_menu.test.tsx
```diff
@@ -60,41 +89,164 @@ describe('tools_menu', () => {
...
+
+      const translateItem = screen
+        .getByText('translate')
+        .closest('leo-menu-item')
+      fireEvent.click(translateItem!)
+
+      expect(mockHandleSmartModeClick).toHaveBeenCalledWith(mockSmartModes[0])
+    })
+
+    it('should call handleSmartModeEdit when edit button is clicked', () => {
```
fallaciousreasoning, line 179: should we test these with filtering too?

fallaciousreasoning:
**Notes**
- Menu on the message with the `Create smart mode` entry disappears when the mouse moves off the menu
- Save button doesn't enable as soon as I've typed text in the shortcut box (possibly the input is using `onChange` instead of `onInput`
- Same for the `prompt` box
- Edit button doesn't show up when the first item is default selected (I think probably because it doesn't have focus - there's a class that gets put on it, I think)
- If you click outside the `Delete` confirmation dialog it closes the edit dialog (potentially losing changes)

yrliou:
Changes in this PR has been split and merged in other PRs, closing this now.

PR #31422, opened 2025-09-25
[Leo] Smart Modes - backend changes

yrliou:
Resolves https://github.com/brave/brave-browser/issues/49630

This is the backend changes split from https://github.com/brave/brave-core/pull/31416 (full PR with both backend and frontend changes).


components/ai_chat/core/browser/engine/engine_consumer.cc
```diff
@@ -35,6 +36,13 @@ std::string EngineConsumer::GetPromptForEntry(
   return prompt_entry->prompt.value_or(prompt_entry->text);
 }
 
+// static
+std::string EngineConsumer::BuildSmartModeDefinitionMessage(
+    const mojom::SmartModeEntryPtr& smart_mode) {
+  return absl::StrFormat("Interpret '/%s' as '%s' when handling the request.",
```
petemill: Do we need to escape quotes in the prompt?

components/ai_chat/core/browser/engine/engine_consumer_oai.cc
```diff
@@ -508,6 +508,17 @@ base::Value::List EngineConsumerOAIRemote::BuildMessages(
...
+
+    // Add Smart Mode definition message if this turn has one
+    if (turn->character_type == CharacterType::HUMAN && turn->smart_mode) {
+      std::string mode_definition =
+          BuildSmartModeDefinitionMessage(turn->smart_mode);
+      base::Value::Dict smart_mode_message;
+      smart_mode_message.Set("role", "user");
+      smart_mode_message.Set("content", mode_definition);
+      messages.Append(std::move(smart_mode_message));
+    }
```
petemill: Perhaps this should instead be an additional content block on the user message? IIRC some API providers complain if the messages aren't alternating or state that their models are trained on alternating messages and suggest multiple content blocks instead.

components/ai_chat/core/browser/engine/engine_consumer_conversation_api_unittest.cc
```diff
@@ -3325,4 +3338,68 @@ TEST_F(EngineConsumerConversationAPIUnitTest,
...
+        EXPECT_EQ(conversation[0].type, ConversationEventType::kChatMessage);
+        EXPECT_EQ(GetContentStrings(conversation[0].content)[0],
+                  "Interpret '/summarize' as 'Please summarize the content' "
+                  "when handling the request.");
+
+        // Second event should be the actual user message
+        EXPECT_EQ(conversation[1].role, ConversationEventRole::kUser);
+        EXPECT_EQ(conversation[1].type, ConversationEventType::kChatMessage);
+        EXPECT_EQ(GetContentStrings(conversation[1].content)[0],
+                  "What is artificial intelligence?");
```
darkdh: Can we change the user prompt to include /summarize for better readability?

components/ai_chat/core/browser/engine/engine_consumer_oai_unittest.cc
```diff
@@ -2421,4 +2451,72 @@ TEST_F(EngineConsumerOAIUnitTest,
...
+            EXPECT_EQ(
+                *smart_mode_msg.FindString("content"),
+                "Interpret '/summarize' as 'Please summarize the content' "
+                "when handling the request.");
+
+            // Third message should be the actual user message
+            const auto& user_msg = messages[2].GetDict();
+            EXPECT_EQ(*user_msg.FindString("role"), "user");
+            EXPECT_EQ(*user_msg.FindString("content"),
+                      "What is artificial intelligence?");
```
darkdh: ditto

components/ai_chat/core/common/prefs.cc
```diff
@@ -149,4 +241,90 @@ std::optional<base::Value::Dict> GetUserMemoryDictFromPrefs(
...
+  return SmartModeDictToStruct(id, mode_value->GetDict());
+}
+
+void AddSmartModeToPrefs(const std::string& shortcut,
+                         const std::string& prompt,
+                         const std::optional<std::string>& model,
+                         PrefService& prefs) {
+  if (prompt.empty() || !IsValidAndUniqueShortcut(prefs, shortcut, "")) {
+    return;
+  }
```
darkdh, line 284: I saw we have maximum 1000 char length limit in spec, shouldn't we check it here? 
Also should we enforce shortcut length limit too?

components/ai_chat/core/common/prefs.cc
```diff
@@ -7,13 +7,105 @@
...
+                              const std::string& exclude_id) {
+  // Check format validation
+  if (shortcut.empty()) {
+    return false;
+  }
+
+  if (!std::all_of(shortcut.begin(), shortcut.end(), [](char c) -> bool {
+        return base::IsAsciiAlphaNumeric(c) || c == '_' || c == '-';
+      })) {
+    return false;
```
darkdh, line 80: Seems like we don't have unit test case cover this

components/ai_chat/core/common/prefs.cc
yrliou: We do enforce 1000 limit in the frontend side to give a real-time feedback direct in frontend which would prevent user from saving it. Didn't have one for shortcut before, I'll add 64 for shortcut to the frontend.

components/ai_chat/core/browser/engine/engine_consumer.cc
yrliou: Hard to say, I played around with this for quite a while, LLM seems to works fine when there are multiple single quotes inside.
```
  Create a summary that includes the following requirements: Use 'bullet points' for key information, include 'direct quotes' from the source material, and don't forget to
  mention the author's 'main argument'. When writing, please use contractions like 'don't', 'won't', 'can't', and 'shouldn't'. Also include phrases like 'it's important to
  note', 'here's what we found', and 'that's the conclusion'. Make sure to handle nested quotes like "The author said 'this is crucial' in the introduction" and complex cases
  like 'the user's input's validation'. Finally, use possessives like 'company's policy', 'users' feedback', and 'system's performance'.
```
Tried to use other ways to wrap this, even tried XML tag, seem not giving better result or even make it worse than just the simple prompt we had.

components/ai_chat/core/browser/engine/engine_consumer_oai.cc
yrliou: I tried with couple popular models on Ollama (gpt-oss:20b, gemma3:4b, qwen3:8b, llama3.1:8b), they were working fine either way, I'll update to content blocks still just in case.

components/ai_chat/core/common/prefs.cc
yrliou: AddSmartModeToPrefs_InvalidShortcut covers this, I'll add UpdateSmartModeToPrefs_InvalidShortcut too.

yrliou: added UpdateSmartModeToPrefs_InvalidShortcut in https://github.com/brave/brave-core/pull/31422/commits/c08ed6c623230b6affe73e51af28e6c265ec9690

components/ai_chat/core/browser/engine/engine_consumer_oai.cc
yrliou: https://github.com/brave/brave-core/pull/31422/commits/0fda566ecb4379e1feaca3c25898e5d890c46c9e

components/ai_chat/core/browser/engine/engine_consumer.cc
yrliou: https://github.com/brave/brave-core/pull/31422/commits/4bcb2e6e5e32eee7415c723011f840066a664aaa only slightly tweaking on the sentence, but still with single quote.

components/ai_chat/core/browser/engine/engine_consumer_oai_unittest.cc
yrliou: https://github.com/brave/brave-core/pull/31422/commits/0fda566ecb4379e1feaca3c25898e5d890c46c9e

components/ai_chat/core/browser/engine/engine_consumer_conversation_api_unittest.cc
yrliou: https://github.com/brave/brave-core/pull/31422/commits/4bcb2e6e5e32eee7415c723011f840066a664aaa

components/ai_chat/core/browser/engine/engine_consumer.cc
petemill: Perhaps we don't need the quotes at all since it's in a content block? Something like:
```
Interpret '/%s' as the following within the rest of this user message:
%s
```

yrliou: Could be, from a quick test it seems fine, but I would keep it as is for now as I haven't really seen bad result due to single quotes involved. I would keep this in mind with more testing during further feature development.

PR #31425, opened 2025-09-25
[Cardano] Use Ciborium for signTx backend method

cypt4:
Resolves https://github.com/brave/brave-browser/issues/49634


components/brave_wallet/browser/internal/cardano_tx_decoder.cc
```diff
@@ -0,0 +1,211 @@
...
+
+#include <memory>
+#include <optional>
+#include <string>
+#include <vector>
+
+#include "base/containers/span.h"
+#include "base/containers/span_writer.h"
+#include "base/containers/to_vector.h"
+#include "base/logging.h"
```
supermassive: ~#include "base/logging.h"~

components/brave_wallet/browser/internal/cardano_tx_decoder.rs
```diff
@@ -0,0 +1,451 @@
...
+                Error::InvalidTransactionFormat,
+            )));
+        }
+    };
+
+    if transaction_array.len() < 2 {
+        return Box::new(CxxSignedCardanoTransactionResult::from(Err(
+            Error::InvalidTransactionFormat,
+        )));
+    }
```
supermassive: 2 -> WITNESS_SET_INDEX

components/brave_wallet/browser/internal/cardano_tx_decoder.rs
```diff
@@ -0,0 +1,451 @@
...
+                    if let CborValue::Array(witness_array) = value {
+                        return witness_array.clone();
+                    }
+                    break;
+                }
+            }
+        }
+    }
+    Vec::new()
+}
```
supermassive: Why we do this?
We just need witnesses array from tx and append array of two items to it. We don't need to parse what is in witness_array

components/brave_wallet/browser/cardano/cardano_api_impl.cc
```diff
@@ -561,11 +559,16 @@ void CardanoApiImpl::OnSignTransactionRequestProcessed(
               kAPIErrorInternalError, WalletInternalErrorMessage(), nullptr));
       return;
     }
-    sign_results.push_back(sign_result.value());
+    // Convert CardanoSignMessageResult to
+    // CardanoTxDecoder::CardanoSignMessageResult
+    CardanoTxDecoder::CardanoSignMessageResult converted_result;
+    converted_result.public_key = base::ToVector(sign_result->pubkey);
+    converted_result.signature_bytes = base::ToVector(sign_result->signature);
+    sign_results.push_back(std::move(converted_result));
```
supermassive: Let's have a CardanoSignMessageResult ctor for this case so this becomes a oneliner

components/brave_wallet/browser/internal/cardano_tx_decoder.rs
```diff
@@ -0,0 +1,451 @@
...
+impl_result!(
+    CxxSignedCardanoTransaction,
+    CxxSignedCardanoTransactionResult,
+    CxxSignedCardanoTransactionValue
+);
+
+/// Extracts the transaction body from a Cardano transaction CBOR
+///
+/// Cardano transactions follow the structure: [body, witness_set, metadata]
+/// This function extracts and parses the body component.
```
supermassive: That's not strictly correct.
Let's have a link also/instead https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/conway/impl/cddl-files/conway.cddl#L17-L18

components/brave_wallet/browser/internal/cardano_tx_decoder.rs
```diff
@@ -0,0 +1,451 @@
...
+#[allow(unused_imports)]
+use crate::ffi::CxxRestoredCardanoInput;
+#[allow(unused_imports)]
+use crate::ffi::CxxRestoredCardanoOutput;
+#[allow(unused_imports)]
+use crate::ffi::CxxWittness;
+
+// Constants for Cardano transaction structure
+const TRANSACTION_BODY_INDEX: u8 = 0;
+const WITNESS_SET_INDEX: u8 = 1;
```
supermassive: can this be `usize`, so no casts below?

components/brave_wallet/browser/internal/cardano_tx_decoder.cc
```diff
@@ -0,0 +1,211 @@
...
+
+namespace {
+std::optional<CardanoTxDecoder::RestoredTransactionInput> ConvertInput(
+    const CxxRestoredCardanoInput& cxx_input) {
+  if (cxx_input.tx_hash.size() != 32u) {
+    return std::nullopt;
+  }
+
+  CardanoTxDecoder::RestoredTransactionInput restored_input;
+  std::copy(cxx_input.tx_hash.begin(), cxx_input.tx_hash.end(),
```
supermassive: std::ranges::copy

components/brave_wallet/browser/internal/cardano_tx_decoder.cc
```diff
@@ -0,0 +1,211 @@
...
+#include "brave/components/brave_wallet/browser/internal/cardano_tx_decoder.rs.h"
+#include "brave/components/brave_wallet/common/cardano_address.h"
+#include "brave/components/brave_wallet/common/zcash_utils.h"
+
+namespace brave_wallet {
+
+namespace {
+std::optional<CardanoTxDecoder::RestoredTransactionInput> ConvertInput(
+    const CxxRestoredCardanoInput& cxx_input) {
+  if (cxx_input.tx_hash.size() != 32u) {
```
supermassive: let's compare with `restored_input.tx_hash.size()`

components/brave_wallet/browser/internal/cardano_tx_decoder.cc
```diff
@@ -0,0 +1,211 @@
...
+  if (!cxx_output.addr.empty()) {
+    auto address = CardanoAddress::FromCborBytes(cxx_output.addr);
+    if (address) {
+      restored_output.address = std::move(address.value());
+    }
+  }
+
+  restored_output.amount = cxx_output.amount;
+
+  // Return RestoredTransactionOutput
```
supermassive: That`s a bad comment

components/brave_wallet/browser/internal/cardano_tx_decoder.cc
```diff
@@ -0,0 +1,211 @@
...
+  CardanoTxDecoder::RestoredTransactionOutput restored_output;
+
+  // Extract addr and amount
+  // Convert addr from vector<uint8_t> to CardanoAddress using FromCborBytes
+  if (!cxx_output.addr.empty()) {
+    auto address = CardanoAddress::FromCborBytes(cxx_output.addr);
+    if (address) {
+      restored_output.address = std::move(address.value());
+    }
+  }
```
supermassive, line 41: Either fail when `FromCborBytes` fails, or just never return optional from `FromCborBytes`

components/brave_wallet/browser/internal/cardano_tx_decoder.cc
```diff
@@ -0,0 +1,211 @@
...
+
+CardanoTxDecoder::CardanoTxDecoder() = default;
+CardanoTxDecoder::~CardanoTxDecoder() = default;
+
+// static
+std::optional<CardanoTxDecoder::RestoredTransaction>
+CardanoTxDecoder::DecodeTransaction(base::span<const uint8_t> cbor_bytes) {
+  auto result = decode_cardano_transaction(
+      ::rust::Slice<const uint8_t>{cbor_bytes.data(), cbor_bytes.size()});
+
```
supermassive: base::SpanToRustSlice

components/brave_wallet/browser/internal/cardano_tx_decoder.cc
```diff
@@ -0,0 +1,211 @@
...
+CardanoTxDecoder::CardanoSignMessageResult::CardanoSignMessageResult(
+    CardanoSignMessageResult&&) = default;
+
+CardanoTxDecoder::CardanoSignMessageResult::~CardanoSignMessageResult() =
+    default;
+
+// static
+std::optional<std::vector<uint8_t>> CardanoTxDecoder::ApplySignResults(
+    const std::vector<uint8_t>& unsigned_tx_bytes,
+    const std::vector<CardanoSignMessageResult>& witness_results) {
```
supermassive: Let's rename it to AddWitnessesToTransaction

components/brave_wallet/browser/internal/cardano_tx_decoder.cc
```diff
@@ -0,0 +1,211 @@
...
+  }
+
+  // Convert std::vector to Rust Vec for the function call using ranges::copy
+  ::rust::Vec<CxxWittness> rust_witnesses;
+  rust_witnesses.reserve(cxx_witnesses.size());
+  std::ranges::copy(cxx_witnesses, std::back_inserter(rust_witnesses));
+
+  // Call apply_signatures(unsigned_tx_bytes, witnesses) from Rust
+  auto result =
+      apply_signatures(::rust::Slice<const uint8_t>{unsigned_tx_bytes.data(),
```
supermassive: SpanToRustSlice

components/brave_wallet/browser/internal/cardano_tx_decoder.cc
```diff
@@ -0,0 +1,211 @@
...
+
+  // If successful, call result->unwrap() to get CxxSignedCardanoTransaction
+  auto signed_tx = result->unwrap();
+
+  // Convert Rust Vec back to C++ vector
+  auto rust_bytes = signed_tx->bytes();
+  std::vector<uint8_t> cpp_bytes;
+  cpp_bytes.reserve(rust_bytes.size());
+  std::ranges::copy(rust_bytes, std::back_inserter(cpp_bytes));
+
```
supermassive: std::vector<uint8_t> cpp_bytes = base::ToVector(signed_tx->bytes());

components/brave_wallet/browser/internal/cardano_tx_decoder.rs
```diff
@@ -0,0 +1,451 @@
...
+
+// Re-export types for external use
+#[allow(unused_imports)]
+use crate::ffi::CxxRestoredCardanoBody;
+#[allow(unused_imports)]
+use crate::ffi::CxxRestoredCardanoInput;
+#[allow(unused_imports)]
+use crate::ffi::CxxRestoredCardanoOutput;
+#[allow(unused_imports)]
+use crate::ffi::CxxWittness;
```
supermassive: Wittness -> Witness
everywhere

components/brave_wallet/browser/internal/cardano_tx_decoder.cc
```diff
@@ -0,0 +1,195 @@
...
+#include <optional>
+#include <string>
+#include <vector>
+
+#include "base/containers/span.h"
+#include "base/containers/span_rust.h"
+#include "base/containers/to_vector.h"
+#include "brave/components/brave_wallet/browser/internal/cardano_tx_decoder.rs.h"
+#include "brave/components/brave_wallet/common/cardano_address.h"
+#include "brave/components/brave_wallet/common/zcash_utils.h"
```
cmazakas: These includes can be cleaned up to just:
```cpp
#include "brave/components/brave_wallet/browser/internal/cardano_tx_decoder.h"

#include <optional>
#include <vector>

#include "base/containers/span.h"
#include "base/containers/to_vector.h"
#include "brave/components/brave_wallet/browser/internal/cardano_tx_decoder.rs.h"
#include "brave/components/brave_wallet/common/cardano_address.h"
```

components/brave_wallet/browser/internal/cardano_tx_decoder.h
```diff
@@ -0,0 +1,94 @@
...
+
+#include <array>
+#include <memory>
+#include <optional>
+#include <string>
+#include <vector>
+
+#include "base/containers/span.h"
+#include "brave/components/brave_wallet/common/cardano_address.h"
+#include "brave/components/brave_wallet/common/zcash_utils.h"
```
cmazakas: These includes can be cleaned up to just:
```cpp
#include <array>
#include <optional>
#include <vector>

#include "base/containers/span.h"
#include "brave/components/brave_wallet/common/cardano_address.h"
```

components/brave_wallet/browser/internal/cardano_tx_decoder.cc
```diff
@@ -0,0 +1,195 @@
...
+    restored_output.address = std::move(address.value());
+  }
+
+  restored_output.amount = cxx_output.amount;
+
+  return restored_output;
+}
+
+std::optional<CardanoTxDecoder::RestoredTransactionBody> ConvertBody(
+    const ::brave_wallet::CxxRestoredCardanoBody& cxx_body) {
```
cmazakas: This namespace qualification seems unnecessary. Even though the code is inside an unnamed namespace, lookup should still apply as normal and the build seems to succeed.

cmazakas (approved):
Rust code looks great to me!

components/brave_wallet/browser/internal/cardano_tx_decoder.rs
```diff
@@ -0,0 +1,410 @@
...
+            _ => return Err(Error::InvalidOutputFormat),
+        };
+
+        let addr_bytes = match &output_array[0] {
+            CborValue::Bytes(bytes) => bytes.clone(),
+            _ => return Err(Error::InvalidOutputFormat),
+        };
+
+        let amount = match &output_array[1] {
+            CborValue::Integer(i) => (*i).try_into().unwrap_or(0),
```
supermassive: should fail instead of `unwrap_or(0)`

components/brave_wallet/browser/internal/cardano_tx_decoder.rs
```diff
@@ -0,0 +1,410 @@
...
+        };
+
+        // Extract transaction hash and index
+        let tx_hash_bytes = match &input_array[0] {
+            CborValue::Bytes(bytes) => bytes.clone(),
+            _ => return Err(Error::InvalidInputFormat),
+        };
+
+        let index = match &input_array[1] {
+            CborValue::Integer(i) => (*i).try_into().unwrap_or(0),
```
supermassive: should fail instead of `unwrap_or(0)`

supermassive (approved):

components/brave_wallet/browser/internal/cardano_tx_decoder.cc
```diff
@@ -0,0 +1,192 @@
...
+    const RestoredTransaction&) = default;
+CardanoTxDecoder::RestoredTransaction::RestoredTransaction(
+    RestoredTransaction&&) = default;
+
+CardanoTxDecoder::CardanoTxDecoder() = default;
+CardanoTxDecoder::~CardanoTxDecoder() = default;
+
+// static
+std::optional<CardanoTxDecoder::RestoredTransaction>
+CardanoTxDecoder::DecodeTransaction(base::span<const uint8_t> cbor_bytes) {
```
kdenhartog, line 111: Can we add a transaction size limit here so that we don't end up with a massive transaction which slows things down or consumes a bunch of memory?

From what I can find in documents the largest this should be that would be accepted by the network is 16KB. It's probably worth testing a larger transaction with @supermassive test site to see if the network still uses this limits though because the docs I'm finding are from 2021 (`maxTxSize` in [CIP-0009](https://cips.cardano.org/cip/CIP-0009))

supermassive: https://cardanoscan.io/protocolparams
```
Max Transaction Size
16384
```
This a blockchain parameter similar to fee constants which are technically varying parameters, but are unlikely to change.
I can add a quick check with hardcoded 16384 limit and later implement fetching that from RPC endpoint like we do for transaction fees.

kdenhartog: yeah just a quick hardcoded check in the code and an issue to come back and fetch it from there later makes sense to me. Thanks

supermassive: https://github.com/brave/brave-core/pull/32176

PR #31429, opened 2025-09-25
[NTP] Isolate autocomplete results for separate search inputs

zenparsing:
Resolves https://github.com/brave/brave-browser/issues/49591

The approach here is to add an "active search input key" to the application state. Search boxes will set this state on focus and will not display autocomplete results if the current key does not match its own key.

### DIFF
--- a/browser/resources/brave_new_tab_page_refresh/components/search/search_input_state.ts
+++ b/browser/resources/brave_new_tab_page_refresh/components/search/search_input_state.ts
@@ -10,11 +10,15 @@ import {
   AutocompleteMatch,
   ClickEvent } from '../../state/search_state'
 
+import {
+  useSearchState,
+  useSearchActions,
+  useSearchMatches } from '../../context/search_context'
+
 import { ResultOption } from './search_results'
-import { useSearchState, useSearchActions } from '../../context/search_context'
 import { urlFromInput } from '../../lib/url_input'
 
-export function useSearchInputState() {
+export function useSearchInputState(inputKey: string) {
   const actions = useSearchActions()
 
   const showSearchBox = useSearchState((s) => s.showSearchBox)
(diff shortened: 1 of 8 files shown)


browser/resources/brave_new_tab_page_refresh/components/search/search_input_state.ts
```diff
@@ -49,10 +50,15 @@ export function useSearchInputState() {
...
-  const resultOptions = React.useMemo(
-      () => getResultOptions(query, searchMatches),
-      [query, searchMatches])
+  // URL (if the user has typed a URL) or a list of autocomplete options. If
+  // this input is not the currently active search input, then the list will be
+  // empty.
+  const resultOptions = React.useMemo(() => {
+    if (inputKey !== activeSearchInputKey) {
+      return []
+    }
```
fallaciousreasoning: do we need this? It looks like we reset the matches when the `inputKey` changes.

browser/resources/brave_new_tab_page_refresh/state/search_state.ts
```diff
@@ -36,6 +36,7 @@ export interface SearchState {
   lastUsedSearchEngine: string
   searchSuggestionsEnabled: boolean
   searchSuggestionsPromptDismissed: boolean
+  activeSearchInputKey: string
```
fallaciousreasoning, line 39: Hey I think lets add some comments about why we need to do this :smile:

fallaciousreasoning (approved):
this lgtm, but do you mind adding some comments explaining why we need to do this so people don't need to come back to the PR/issue to find out?

browser/resources/brave_new_tab_page_refresh/components/search/search_input_state.ts
zenparsing: This has been refactored out after introducing the `useSearchMatches` hook.

tmancey:
@zenparsing, great work. I will sanity test this today with the dynamic NTT POC. Thanks and much appreciated.

PR #31431, opened 2025-09-25
Android origin skus

SergeyZhukovsky:
We need to pass order token to the SKUs SDK to be able to register Brave Origin purchase from Google Play Store on Android. The PR exposes Brave domains util functions and SkusService via mojom to Java.
There is no any entry point to the logic of Brave Origin purchase on Android. There will be future PRs that connect everything together.
Resolves: https://github.com/brave/brave-browser/issues/49619

### DIFF
--- a/android/java/org/chromium/chrome/browser/billing/InAppPurchaseWrapper.java
+++ b/android/java/org/chromium/chrome/browser/billing/InAppPurchaseWrapper.java
@@ -453,7 +453,7 @@ public class InAppPurchaseWrapper {
      * @param profile The current user profile for context (can be null)
      */
     public void initiatePurchase(
-            Activity activity, ProductDetails productDetails, @NonNull Profile profile) {
+            Activity activity, ProductDetails productDetails, Profile profile) {
         // Store the profile for use in the asynchronous purchase callback flow
         mPurchaseProfile = profile;
 
--- a/android/java/org/chromium/chrome/browser/billing/InAppPurchaseWrapper.java
+++ b/android/java/org/chromium/chrome/browser/billing/InAppPurchaseWrapper.java
@@ -30,10 +30,13 @@ import com.android.billingclient.api.QueryPurchasesParams;
 
 import org.chromium.base.ContextUtils;
 import org.chromium.base.Log;
+import org.chromium.base.ThreadUtils;
 import org.chromium.chrome.R;
 import org.chromium.chrome.browser.app.BraveActivity;
 import org.chromium.chrome.browser.brave_leo.BraveLeoPrefUtils;
 import org.chromium.chrome.browser.brave_leo.BraveLeoUtils;
+import org.chromium.chrome.browser.brave_origin.BraveOriginSubscriptionPrefs;
+import org.chromium.chrome.browser.profiles.Profile;
 import org.chromium.chrome.browser.util.BraveConstants;
 import org.chromium.chrome.browser.util.LiveDataUtil;
 import org.chromium.chrome.browser.vpn.utils.BraveVpnPrefUtils;
@@ -74,6 +77,9 @@ public class InAppPurchaseWrapper {
     public static final String VPN_RELEASE_MONTHLY_SUBSCRIPTION = "brave.vpn.monthly";
     public static final String VPN_RELEASE_YEARLY_SUBSCRIPTION = "brave.vpn.yearly";
     private BillingClient mBillingClient;
+
+    // Profile context for current purchase flow
+    private Profile mPurchaseProfile;
 
     private static final long MICRO_UNITS =
             1000000; // 1,000,000 micro-units equal one unit of the currency
@@ -439,6 +445,22 @@ public class InAppPurchaseWrapper {
                 });
     }
 
+    /**
+     * Initiates a subscription purchase flow with profile context.
+     *
+     * @param activity The activity to launch the billing flow from
+     * @param productDetails The subscription product details to purchase
+     * @param profile The current user profile for context (can be null)
+     */
+    public void initiatePurchase(
+            Activity activity, ProductDetails productDetails, @NonNull Profile profile) {
+        // Store the profile for use in the asynchronous purchase callback flow
+        mPurchaseProfile = profile;
+
+        // Delegate to the original method
+        initiatePurchase(activity, productDetails);
+    }
+
     public void processPurchases(Context context, Purchase activePurchase) {
         acknowledgePurchase(context, activePurchase);
     }
@@ -524,16 +546,22 @@ public class InAppPurchaseWrapper {
                             BraveLeoUtils.bringMainActivityOnTop();
                         }
                     });
-        } else if (isOriginProduct && activity != null) {
-            activity.runOnUiThread(
-                    new Runnable() {
-                        @Override
-                        public void run() {
-                            // TODO(sergz): Add Origin subscription handling
-                        }
+        } else if (isOriginProduct) {
+            ThreadUtils.runOnUiThread(
+                    () -> {
+                        BraveOriginSubscriptionPrefs.setIsSubscriptionActive(
+                                mPurchaseProfile, true);
+                        BraveOriginSubscriptionPrefs.setOriginPackageName(mPurchaseProfile);
+                        BraveOriginSubscriptionPrefs.setOriginProductId(
+                                mPurchaseProfile, purchase.getProducts().get(0));
+                        BraveOriginSubscriptionPrefs.setOriginPurchaseToken(
+                                mPurchaseProfile, purchase.getPurchaseToken());
                     });
         }
         showToast(context.getResources().getString(R.string.subscription_consumed));
+
+        // Clear the stored profile after processing
+        mPurchaseProfile = null;
     }
 
     private boolean isVPNProduct(List<String> productIds) {
--- a/android/java/org/chromium/chrome/browser/brave_origin/BraveOriginPlansActivity.java
+++ b/android/java/org/chromium/chrome/browser/brave_origin/BraveOriginPlansActivity.java
@@ -26,6 +26,7 @@ import org.chromium.chrome.browser.billing.InAppPurchaseWrapper;
 import org.chromium.chrome.browser.billing.LinkSubscriptionUtils;
 import org.chromium.chrome.browser.init.ActivityProfileProvider;
 import org.chromium.chrome.browser.init.AsyncInitializationActivity;
+import org.chromium.chrome.browser.profiles.Profile;
 import org.chromium.chrome.browser.profiles.ProfileProvider;
 import org.chromium.chrome.browser.util.LiveDataUtil;
 import org.chromium.chrome.browser.util.TabUtils;
@@ -127,9 +128,14 @@ public class BraveOriginPlansActivity extends AsyncInitializationActivity {
                     if (productDetails == null) {
                         return;
                     }
-                    // Initiate purchase flow for the selected plan
+                    // Get the current profile and initiate purchase flow for the selected plan. The
+                    // purchase
+                    // initiated on original profile only.
+                    Profile currentProfile =
+                            getProfileProviderSupplier().get().getOriginalProfile();
                     InAppPurchaseWrapper.getInstance()
-                            .initiatePurchase(BraveOriginPlansActivity.this, productDetails);
+                            .initiatePurchase(
+                                    BraveOriginPlansActivity.this, productDetails, currentProfile);
                 });
 
         TextView refreshCredentialsButton = findViewById(R.id.refresh_credentials_button);
--- a/android/java/org/chromium/chrome/browser/brave_origin/BraveOriginPlansActivity.java
+++ b/android/java/org/chromium/chrome/browser/brave_origin/BraveOriginPlansActivity.java
@@ -129,8 +129,7 @@ public class BraveOriginPlansActivity extends AsyncInitializationActivity {
                         return;
                     }
                     // Get the current profile and initiate purchase flow for the selected plan. The
-                    // purchase
-                    // initiated on original profile only.
+                    // purchase is initiated on original profile only.
                     Profile currentProfile =
                             getProfileProviderSupplier().get().getOriginalProfile();
                     InAppPurchaseWrapper.getInstance()
--- a/android/java/org/chromium/chrome/browser/brave_origin/BraveOriginSubscriptionPrefs.java
+++ b/android/java/org/chromium/chrome/browser/brave_origin/BraveOriginSubscriptionPrefs.java
@@ -1,4 +1,4 @@
-/* Copyright (c) 2024 The Brave Authors. All rights reserved.
+/* Copyright (c) 2025 The Brave Authors. All rights reserved.
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this file,
  * You can obtain one at https://mozilla.org/MPL/2.0/. */
--- a/android/java/org/chromium/chrome/browser/brave_origin/BraveOriginSubscriptionPrefs.java
+++ b/android/java/org/chromium/chrome/browser/brave_origin/BraveOriginSubscriptionPrefs.java
@@ -1,0 +1,373 @@
+/* Copyright (c) 2024 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+package org.chromium.chrome.browser.brave_origin;
+
+import android.util.Base64;
+
+import org.json.JSONException;
+import org.json.JSONObject;
+
+import org.chromium.base.Callback;
+import org.chromium.base.ContextUtils;
+import org.chromium.base.Log;
+import org.chromium.base.task.PostTask;
+import org.chromium.base.task.TaskTraits;
+import org.chromium.brave.browser.skus.SkusServiceFactory;
+import org.chromium.brave.browser.util.BraveDomainsUtils;
+import org.chromium.brave.browser.util.ServicesEnvironment;
+import org.chromium.build.annotations.NullMarked;
+import org.chromium.build.annotations.Nullable;
+import org.chromium.chrome.browser.preferences.BravePref;
+import org.chromium.chrome.browser.profiles.Profile;
+import org.chromium.components.prefs.PrefService;
+import org.chromium.components.user_prefs.UserPrefs;
+import org.chromium.skus.mojom.SkusResult;
+import org.chromium.skus.mojom.SkusResultCode;
+import org.chromium.skus.mojom.SkusService;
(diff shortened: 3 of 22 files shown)


android/java/org/chromium/chrome/browser/billing/InAppPurchaseWrapper.java
```diff
@@ -439,6 +445,22 @@ public void initiatePurchase(Activity activity, ProductDetails productDetails) {
...
 
+    /**
+     * Initiates a subscription purchase flow with profile context.
+     *
+     * @param activity The activity to launch the billing flow from
+     * @param productDetails The subscription product details to purchase
+     * @param profile The current user profile for context (can be null)
+     */
+    public void initiatePurchase(
+            Activity activity, ProductDetails productDetails, @NonNull Profile profile) {
```
AlexeyBarabash: probably it makes sense to mark whole  `class InAppPurchaseWrapper` as `@Nullmarked`

SergeyZhukovsky: hm, I wonder why presubmit didn't complain about it.

AlexeyBarabash: I also don't know.
I would expect exclusion at https://github.com/brave/brave-core/blob/master/chromium_presubmit_config.json5#L171 , but `InAppPurchaseWrapper.java` is not there

browser/skus/android/java/src/org/chromium/brave/browser/skus/SkusServiceFactory.java
```diff
@@ -0,0 +1,65 @@
...
+            }
+        }
+        return sInstance;
+    }
+
+    private SkusServiceFactory() {}
+
+    public @Nullable SkusService getSkusService(
+            Profile profile, @Nullable ConnectionErrorHandler connectionErrorHandler) {
+        if (profile == null) {
```
Olipro: We don't allow this to be null, but we check for it anyway. It might be better if we remove this check since `Profile::FromJavaObject()` on the C++ side _already_ has its own null check. Nonetheless, this check right here is also the only thing standing between us and invoking UB on the C++ side.

It does mean that you'll need to make the C++ code check that `auto* profile` isn't `nullptr` and either return a `jlong` that represents an invalid value, or alter the return type to something that communicates expected vs unexpected. Perhaps `std::numeric_limits<jlong>::max()` would suffice though. it also eliminates the risk of UB if someone else calls the JNI function with a null object.

android/java/org/chromium/chrome/browser/billing/InAppPurchaseWrapper.java
SergeyZhukovsky: I'll create a follow up for it as it looks like the whole purchase logic will have to retested on diff products after marking everything

android/java/org/chromium/chrome/browser/brave_origin/BraveOriginSubscriptionPrefs.java
```diff
@@ -0,0 +1,373 @@
...
+
+    /**
+     * Sets the Origin subscription active status for the given profile.
+     *
+     * @param profile The profile to use for preference storage
+     * @param value The subscription active status
+     */
+    public static void setIsSubscriptionActive(@Nullable Profile profile, boolean value) {
+        if (profile == null) {
+            Log.e(TAG, "BraveOriginSubscriptionPrefs.setIsSubscriptionActive profile is null");
```
AlexeyBarabash: nit: `TAG` already contains `BraveOriginSubsPrefs`, so I would propose to skip class name at `BraveOriginSubscriptionPrefs.setIsSubscriptionActive profile is null` and below.

browser/skus/android/java/src/org/chromium/brave/browser/skus/SkusServiceFactory.java
SergeyZhukovsky: yep, my bad, it could be removed

brave_domains/BUILD.gn
```diff
@@ -6,6 +6,10 @@
 import("//build/buildflag_header.gni")
 import("buildflags.gni")
 
+if (is_android) {
+  import("//build/config/android/rules.gni")
```
AlexeyBarabash: why do we need this import?

brave_domains/android/brave_domains_utils_android.cc
```diff
@@ -0,0 +1,24 @@
...
+#include "brave/brave_domains/service_domains.h"
+
+namespace brave_domains {
+namespace android {
+
+static std::string JNI_BraveDomainsUtils_GetServicesDomain(JNIEnv* env,
+                                                           std::string& prefix,
+                                                           jint environment) {
+  // Direct cast since Java enum values match C++ enum values exactly
+  auto env_value = static_cast<brave_domains::ServicesEnvironment>(environment);
```
Olipro: Since `enum ServicesEnvironment` doesn't have an explicit size (nor an implicit one because it isn't an `enum class`) This code is UB if `environment` ever contains a value greater than the number of bits used to store the enum's largest value.

We should either explicitly ensure that `environment` is within the range of the enum's possible values, or alter the `enum` to have an explicit integral type that is at least the same size as and compatible with `jint` - of course, that means it won't be UB any more but can still hold a value that doesn't represent any of its defined values.

brave_domains/BUILD.gn
AlexeyBarabash: as the Android targets are inside of `brave_domains/android/BUILD.gn` it seems to me `import("//build/config/android/rules.gni")` is not required here at `brave_domains/BUILD.gn`

browser/skus/BUILD.gn
```diff
@@ -3,6 +3,10 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at https://mozilla.org/MPL/2.0/.
 
+if (is_android) {
+  import("//build/config/android/rules.gni")
```
AlexeyBarabash: I think, this import is also redundant here, because `browser/skus/android/BUILD.gn` already has it.

AlexeyBarabash (approved):
LGTM with a few minor notices

brave_domains/android/brave_domains_utils_android.cc
```diff
@@ -0,0 +1,24 @@
...
+#include "base/android/jni_android.h"
+#include "base/android/jni_string.h"
+#include "brave/brave_domains/android/jni_headers/BraveDomainsUtils_jni.h"
+#include "brave/brave_domains/service_domains.h"
+
+namespace brave_domains {
+namespace android {
+
+static std::string JNI_BraveDomainsUtils_GetServicesDomain(JNIEnv* env,
+                                                           std::string& prefix,
```
Olipro, line 28: `GetServicesDomain()` takes the prefix by-value so, if it's safe to do, `std::move(prefix)` below.

If it isn't safe to mutate, make this param `const` - I'm guessing the JNI wrapper shouldn't be doing anything crazy where it needs to look at the object itself after returning so I'd hope it's safe to move.

browser/skus/android/skus_service_factory_android.cc
```diff
@@ -0,0 +1,26 @@
...
+#include "chrome/browser/profiles/profile.h"
+#include "mojo/public/cpp/bindings/pending_receiver.h"
+
+namespace brave {
+namespace android {
+
+static jlong JNI_SkusServiceFactory_GetInterfaceToSkusService(
+    JNIEnv* env,
+    const base::android::JavaParamRef<jobject>& profile_android) {
+  auto* profile = Profile::FromJavaObject(profile_android);
```
Olipro, line 19: See: https://github.com/brave/brave-core/pull/31431/files/3b23ee9ba7b970d6062d4610aa4ab4a9fa466e5f#r2379723156

brave_domains/android/brave_domains_utils_android.cc
SergeyZhukovsky: great point! I will do the validation as not sure exactly where that enum could be used elsewhere.

browser/skus/android/java/src/org/chromium/brave/browser/skus/SkusServiceFactory.java
SergeyZhukovsky: Fixed https://github.com/brave/brave-core/pull/31431/commits/8890468e9e9aaaf196a4b800659b953acd7071fe

android/java/org/chromium/chrome/browser/brave_origin/BraveOriginSubscriptionPrefs.java
SergeyZhukovsky: Removed https://github.com/brave/brave-core/pull/31431/commits/8890468e9e9aaaf196a4b800659b953acd7071fe

brave_domains/BUILD.gn
SergeyZhukovsky: removed https://github.com/brave/brave-core/pull/31431/commits/8890468e9e9aaaf196a4b800659b953acd7071fe

brave_domains/android/brave_domains_utils_android.cc
SergeyZhukovsky: added https://github.com/brave/brave-core/pull/31431/commits/8890468e9e9aaaf196a4b800659b953acd7071fe

browser/skus/BUILD.gn
SergeyZhukovsky: removed https://github.com/brave/brave-core/pull/31431/commits/8890468e9e9aaaf196a4b800659b953acd7071fe

brave_domains/android/brave_domains_utils_android.cc
SergeyZhukovsky: made `sts::move` https://github.com/brave/brave-core/pull/31431/commits/8890468e9e9aaaf196a4b800659b953acd7071fe

browser/skus/android/skus_service_factory_android.cc
SergeyZhukovsky: added the check https://github.com/brave/brave-core/pull/31431/commits/8890468e9e9aaaf196a4b800659b953acd7071fe

brave_domains/android/brave_domains_utils_android.cc
```diff
@@ -0,0 +1,29 @@
...
+                                                           jint environment) {
+  // Validate that environment is within the valid range of enum values,
+  // if not, default to DEV.
+  if (environment < static_cast<jint>(brave_domains::DEV) ||
```
Olipro: Let's simplify this to two `!=` operations - that way, we don't depend upon the declaration order of the items in the enum.

...or make the whole thing a ternary, whichever floats your boat:

`environment = environment != static_cast<jint>(brave_domains::DEV) && environment != static_cast<jint>(brave_domains::PROD) ? static_cast<jint>(brave_domains::DEV) : environment;`

browser/skus/android/java/src/org/chromium/brave/browser/skus/SkusServiceFactory.java
```diff
@@ -0,0 +1,62 @@
...
+import org.chromium.skus.mojom.SkusService;
+
+@NullMarked
+@JNINamespace("brave::android")
+public class SkusServiceFactory {
+    private static final Object sLock = new Object();
+    private static @Nullable SkusServiceFactory sInstance;
+
+    public static SkusServiceFactory getInstance() {
+        synchronized (sLock) {
```
Olipro: I take it that this isn't a hot enough code path to warrant an initial check outside of the `synchronized` so that we can early-return without contention?

With that said. If that is the case, is there any reason we can't simply remove this and declare this method as `public static synchronized SkusServiceFactory getInstance()` ?

browser/skus/android/skus_service_factory_android.cc
```diff
@@ -0,0 +1,29 @@
...
+  auto pending = mojo::PendingRemote<skus::mojom::SkusService>();
```
Olipro, line 20: Just for the benefit of my own understanding: Does this guarantee that when we return, the handle value is invalid because we have a default-constructed instance?

feel free to self-resolve this comment.

brave_domains/android/brave_domains_utils_android.cc
SergeyZhukovsky: we are going to miss everything that is going to maybe be added inside in between in the future.

SergeyZhukovsky: but actually with < and > we are going to miss everything that is maybe going to be added before or after. That's a trade off :).

SergeyZhukovsky: not sure honestly that what I have is something worse of what you propose.

SergeyZhukovsky: they are both not ideal

browser/skus/android/skus_service_factory_android.cc
SergeyZhukovsky: it is true, I checked that.

browser/skus/android/java/src/org/chromium/brave/browser/skus/SkusServiceFactory.java
SergeyZhukovsky: yeah, it could be done, would make the code cleaner at least.

SergeyZhukovsky: actually lint gives a warning on that `brave/browser/skus/android/java/src/org/chromium/brave/browser/skus/SkusServiceFactory.java:26: warning: [NoSynchronizedMethodCheck] Used synchronized modifier in public method getInstance`

brave_domains/android/brave_domains_utils_android.cc
Olipro: I agree with the matter of not catching new values, but that's another matter. using `!=` is orthogonal.

If we want to assure this, we could create a dummy `constexpr` function with a `switch` on the enum values (and no `default` case)

Our build will error out if anyone adds a new element to the enum since we treat missing switch cases on an enum as an error. Though this sounds like it should live in a file that builds with the unit tests.

...There may be an even better way that we enforce these sort of checks that I'm not aware of, of course.

browser/skus/android/java/src/org/chromium/brave/browser/skus/SkusServiceFactory.java
Olipro: For what the code is doing, it's functionally the same. I note that Google have used `@SuppressWarning` to suppress this themselves in places where it's not of value.

Unless it's worth it to do a double-checked lock, I'd be of the opinion to suppress it.

Olipro: Actually, scratch that, this should probably be reduced to using this pattern: https://en.wikipedia.org/wiki/Initialization-on-demand_holder_idiom if we _must_ lazy-init the object.

brave_domains/android/brave_domains_utils_android.cc
cdesouza-chromium: I think a switch would feel more natural here.

SergeyZhukovsky: added the switch compile time protection in both codebase and unit test https://github.com/brave/brave-core/pull/31431/commits/4d7c9d5ec7a7f0074ecdaf72b3bea7aa5a38cfe2

browser/skus/android/java/src/org/chromium/brave/browser/skus/SkusServiceFactory.java
SergeyZhukovsky: switched to it https://github.com/brave/brave-core/pull/31431/commits/4d7c9d5ec7a7f0074ecdaf72b3bea7aa5a38cfe2

android/java/org/chromium/chrome/browser/billing/InAppPurchaseWrapper.java
SergeyZhukovsky: the follow up https://github.com/brave/brave-browser/issues/49649

android/java/org/chromium/chrome/browser/brave_origin/BraveOriginPlansActivity.java
```diff
@@ -127,9 +128,13 @@ protected void triggerLayoutInflation() {
                     if (productDetails == null) {
                         return;
                     }
-                    // Initiate purchase flow for the selected plan
+                    // Get the current profile and initiate purchase flow for the selected plan. The
+                    // purchase is initiated on original profile only.
+                    Profile currentProfile =
+                            getProfileProviderSupplier().get().getOriginalProfile();
```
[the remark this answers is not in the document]
bridiver: as discussed ideally this profile is passed from the initiator (settings) where selecting regular profile is always ok, but this is ok for now. Please add a TODO with issue link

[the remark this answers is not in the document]
SergeyZhukovsky: https://github.com/brave/brave-browser/issues/49669. It's going to be done when Settings for them are implemented.

brave_domains/android/brave_domains_utils_android.cc
```diff
@@ -0,0 +1,41 @@
...
+#include "brave/brave_domains/service_domains.h"
+
+namespace brave_domains {
+namespace android {
+
+// Validate enum values using a switch statement without default case.
+// This ensures all enum values are handled and will cause a compile error
+// if new enum values are added without updating this function.
+constexpr bool IsValidServicesEnvironment(jint environment) {
+  switch (static_cast<brave_domains::ServicesEnvironment>(environment)) {
```
Olipro: Well, we're almost there with this.

Since we aren't able to call this function in a `constexpr` context (because `jint` is going to be assigned at runtime) the compiler can't tell you if there will ever be UB and it'll go ahead and compile. This is incidentally one of the reasons why `consteval` became a thing. But we can't use that of course because the `jint` can't be constant-evaluated.

Fortunately, we can remove the `static_cast` entirely since the enum is unscoped (allowing its values to implicitly convert to integrals in the case statements) and there will be no chance of UB. The unit test below then acts as the "canary in the coal mine" for the future if anyone ever alters the `enum`

SergeyZhukovsky: I would think we already will avoid the UB as we won't be able to compile, but ok, removing the `static_cast`

SergeyZhukovsky:
Added skip label to everything but Android as only Android file changed from this point

brave_domains/android/brave_domains_utils_android.cc
SergeyZhukovsky: pushed 13ff6223d3809ef1694fd2ac04bdfa07754c15bb

Olipro: Sadly not - here's an example you can play with to see where UB will be caught, and where it won't:

https://godbolt.org/z/cfjza936Y

PR #31441, opened 2025-09-25
[Android] Remaining Brave options for quick shortcut settings

samartnik:
Resolves https://github.com/brave/brave-browser/issues/49409


android/java/org/chromium/chrome/browser/toolbar/adaptive/BraveNewsButtonController.java
```diff
@@ -0,0 +1,71 @@
...
+    @Override
+    public void onClick(View view) {
+        Profile profile = mProfileSupplier.get();
+        if (profile == null) return;
+
+        assert mContext instanceof BraveActivity : "Context is not an BraveActivity";
+
+        // Open Brave News Settings
+        if (mContext instanceof BraveActivity braveActivity) {
+            braveActivity.openBraveNewsSettings();
```
AlexeyBarabash: should it open news settings or the news itself at the new tab?

samartnik: That's a good catch, currently it does the same action as main menu item does (which is opening settings and doesn't look really helpful). @timchilds what is actual desired behavior for `Brave News` shortcut?

android/java/org/chromium/chrome/browser/toolbar/adaptive/BraveLeoButtonController.java
```diff
@@ -0,0 +1,72 @@
...
+                /* tooltipTextResId= */ R.string.menu_brave_leo);
+
+        mContext = context;
+        mProfileSupplier = profileSupplier;
+    }
+
+    @Override
+    public void onClick(View view) {
+        Profile profile = mProfileSupplier.get();
+        if (profile == null) return;
```
SergeyZhukovsky: why do we need to check on `profile` in that function and most of other `onClick`?

SergeyZhukovsky: `mProfileSupplier` maybe not needed at all for such classes when we don't use it.

samartnik: yeah, that's leftovers from my previous implementation, I'll clean it up

PR #31442, opened 2025-09-25
Autocomplete: commit suggestion on hardware Right Arrow (Fixes #44927)

3lvis:
**Summary**
This change ensures the Right Arrow key from hardware keyboards is treated like the autocomplete accessory arrow: when a suggestion is active, pressing Right Arrow commits the suggestion into the text field. Some hardware keyboards send arrow events as `UIPress` rather than triggering `UIKeyCommand`; handling `pressesEnded` provides a robust fallback.

**Motivation**
Reproduces and fixes the behavior described in brave/brave-browser#44927: on iPad (and simulator with hardware keyboard), pressing Right Arrow while an autocomplete suggestion is highlighted should commit the suggestion instead of moving the caret one character. The existing `UIKeyCommand` path was not triggered in certain contexts where the arrow arrives as a `UIPress`.

**What changed**
* Added a `pressesEnded(_:with:)` override to `AutocompleteTextField` that:

  * Detects Right Arrow via `charactersIgnoringModifiers` and HID usage (keyCode raw value 79) for compatibility.
  * If a suggestion is active (`isSelectionActive`), calls `applyCompletion()` and moves caret to the end.
  * Leaves normal caret movement untouched when no suggestion is active.

**How I tested**
* iPad Simulator with hardware keyboard connected:

  1. Focus the location/search bar.
  2. Type a prefix that triggers a known suggestion (e.g. `old.` for `old.reddit.com`).
  3. Press Right Arrow — suggestion is committed and caret placed at end.
  4. Press Right Arrow when no suggestion is active — caret behaves normally.
* Verified behavior on Simulator; recommend test on an actual iPad + external keyboard for final verification.

**Related**
Fixes brave/brave-browser#44927

### DIFF
--- a/ios/brave-ios/Sources/Brave/Frontend/Widgets/AutocompleteTextField.swift
+++ b/ios/brave-ios/Sources/Brave/Frontend/Widgets/AutocompleteTextField.swift
@@ -427,4 +427,19 @@ public class AutocompleteTextField: UITextField, UITextFieldDelegate {
     applyCompletion()
     super.touchesBegan(touches, with: event)
   }
+
+  public override func pressesEnded(_ presses: Set<UIPress>, with event: UIPressesEvent?) {
+    for press in presses {
+      guard let key = press.key else { continue }
+      let chars = key.charactersIgnoringModifiers
+      let isRightArrowChar = chars == UIKeyCommand.inputRightArrow || chars == UIKeyCommand.inputRightArrow
+      let isRightArrowKeyCode = key.keyCode.rawValue == 79
+      if (isRightArrowChar || isRightArrowKeyCode) && isSelectionActive {
+        applyCompletion()
+        selectedTextRange = textRange(from: endOfDocument, to: endOfDocument)
+        return
+      }
+    }
+    super.pressesEnded(presses, with: event)
+  }
 }


3lvis:
**Video testing on iPhone simulator as well.**

[Watch the video](https://github.com/user-attachments/assets/b6d7373d-dc8a-4ea9-ac14-b1274597548f)

ios/brave-ios/Sources/Brave/Frontend/Widgets/AutocompleteTextField.swift
```diff
@@ -427,4 +427,25 @@ public class AutocompleteTextField: UITextField, UITextFieldDelegate {
...
+    }
+
+    for press in presses {
+      guard let key = press.key else { continue }
+      let chars = key.charactersIgnoringModifiers
+      let isRightArrowChar = chars == UIKeyCommand.inputRightArrow
+      let isRightArrowKeyCode = key.keyCode == .keyboardRightArrow
+      if isRightArrowChar || isRightArrowKeyCode {
+        applyCompletion()
+        selectedTextRange = textRange(from: endOfDocument, to: endOfDocument)
```
StephenHeaps: this shouldn't be needed since it happens in `applyCompletion()` (`removeCompletion()` returns true when `isSelectionActive` which is guarded against above).

ios/brave-ios/Sources/Brave/Frontend/Widgets/AutocompleteTextField.swift
```diff
@@ -427,4 +427,25 @@ public class AutocompleteTextField: UITextField, UITextFieldDelegate {
...
+        return
```
StephenHeaps: I don't think early return is needed here unless I'm mistaken?

ios/brave-ios/Sources/Brave/Frontend/Widgets/AutocompleteTextField.swift
3lvis: You're right! Removed.

ios/brave-ios/Sources/Brave/Frontend/Widgets/AutocompleteTextField.swift
3lvis: We should stop the loop once the right arrow check was handled, I did change the return for a break so super gets called at the end.

3lvis:
Updated to handle the comments 👍

PR #31443, opened 2025-09-26
[AI Chat]: Add mojom bookmarks service

fallaciousreasoning:
Resolves https://github.com/brave/brave-browser/issues/48881
Resolves https://github.com/brave/brave-browser/issues/49652

This PR adds a mojom service for getting a list of bookmarks in the frontend. It uses the `//components/bookmark` component and can be shared with iOS. This service will be used in a follow up PR to allow the frontend to attach bookmarks to conversations.

+204 lines of tests
+149 lines of implementation


browser/resources/settings/sources.gni
```diff
@@ -172,6 +172,7 @@ brave_settings_ts_extra_deps =
 brave_settings_mojo_files = [
   "$root_gen_dir/brave/components/ai_chat/core/common/mojom/settings_helper.mojom-webui.ts",
   "$root_gen_dir/brave/components/ai_chat/core/common/mojom/ai_chat.mojom-webui.ts",
+  "$root_gen_dir/brave/components/ai_chat/core/common/mojom/bookmarks.mojom-webui.ts",
```
petemill, line 175: Do we need to include this in the comment about why it needs to be added despite not being used on settings?

components/ai_chat/core/common/mojom/bookmarks.mojom
```diff
@@ -0,0 +1,20 @@
...
+module ai_chat.mojom;
+
+import "url/mojom/url.mojom";
+
+// Represents a bookmark (guaranteed to be non-folder).
+struct Bookmark {
+  int64 id;
+  string title;
+  url.mojom.Url url;
+};
```
petemill, line 15: How are we going to sort bookmarks in the UI if the user chooses since we don't have something like hierarchy or time of last visit. Do we need any other data here?

components/ai_chat/core/common/mojom/bookmarks.mojom
```diff
@@ -0,0 +1,20 @@
...
+
+interface BookmarksPageHandler {
+  // Gets the current list of bookmarks.
+  GetBookmarks() => (array<Bookmark> bookmarks);
```
petemill, line 19: When will we call this? Do we need to observe the list and get notified of changes?

fallaciousreasoning: I originally had this, but @bridiver asked me to remove it until we're sure we'll need it. It's not a massive deal because we'll refetch the bookmarks every time we open the dialog, and I can't imagine users are going to be regularly adding/removing bookmarks while the dialog is open.

Happy to add it back, but I'd prefer to do so in a followup PR.

components/ai_chat/core/common/mojom/bookmarks.mojom
fallaciousreasoning: Currently the UI has no affordances for sorting. I think lets cross this bridge when we need to, rather than worrying about it now - its easy to add new fields to a mojom struct.

browser/resources/settings/sources.gni
fallaciousreasoning: sure, I can update the comment to be about both.

goodov (approved):

PR #31444, opened 2025-09-26
[TabHelpers]: Add a section for required tab helpers

fallaciousreasoning:
Resolves https://github.com/brave/brave-browser/issues/49653

For AI Chat, we need to be able to load a URL from a bookmark or history entry in the background. While doing this, we want to ensure that we attach the relevant tab helpers for privacy/security.

This PR adds a `AttachRequireTabHelpers` method in which we can place `TabHelpers` which should be applied even to background WebContents. We can call this method to attach all required TabHelpers to our background contents

### DIFF
--- a/browser/brave_tab_helpers.cc
+++ b/browser/brave_tab_helpers.cc
@@ -107,8 +107,10 @@
 namespace brave {
 
 void AttachTabHelpers(content::WebContents* web_contents) {
-  brave_shields::BraveShieldsWebContentsObserver::CreateForWebContents(
-      web_contents);
+  // If your TabHelper is related to privacy consider whether it belongs in this
+  // method, so it can also be attached to background WebContents.
+  AttachPrivacySensitiveTabHelpers(web_contents);
+
 #if BUILDFLAG(IS_ANDROID)
   YouTubeScriptInjectorTabHelper::CreateForWebContents(web_contents);
 #else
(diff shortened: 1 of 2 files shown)


browser/brave_tab_helpers.h
```diff
@@ -13,7 +14,11 @@ namespace brave {
 
 void AttachTabHelpers(content::WebContents* web_contents);
 
+// Note: These TabHelpers are related to security/privacy and should be attached
+// event to background WebContents, which aren't displayed to the user.
+// As such, these TabHelpers must not depend on being displayed in a tab.
+void AttachRequiredTabHelpers(content::WebContents* web_contents);
```
ShivanKaul: The name right now doesn't indicate that these are required for security/privacy reasons. What about `AttachPrivacySensitiveTabHelpers`?

browser/brave_tab_helpers.h
```diff
@@ -13,7 +14,11 @@ namespace brave {
 
 void AttachTabHelpers(content::WebContents* web_contents);
 
+// Note: These TabHelpers are related to security/privacy and should be attached
+// event to background WebContents, which aren't displayed to the user.
```
ShivanKaul:
```suggestion
// even to background WebContents, which aren't displayed to the user.
```

browser/brave_tab_helpers.cc
```diff
@@ -107,8 +107,8 @@
 namespace brave {
 
 void AttachTabHelpers(content::WebContents* web_contents) {
-  brave_shields::BraveShieldsWebContentsObserver::CreateForWebContents(
-      web_contents);
+  AttachRequiredTabHelpers(web_contents);
```
ShivanKaul: I know we are adding the comment to the declaration, but I think we should add a comment or pointer here too, since this is where a developer would think about adding a new tab helper, and would be easy to miss during code review if it's only in the header file.

ShivanKaul (approved):
I don't have very strong feelings on what to call the method, or where exactly the comment lives, so won't block on them.

PR #31448, opened 2025-09-26
[AI Chat]: Introduce AssociatedLinkContent for loading page content from a URL

fallaciousreasoning:
Resolves https://github.com/brave/brave-browser/issues/49660

**Note:** This PR only introduces and tests a class for loading a URL in a background WebContents. Attaching Bookmarks/History will happen in a subsequent PR(s).

See https://github.com/brave/brave-browser/issues/48880 for the tracking issue

This PR has been split out of https://github.com/brave/brave-core/pull/30963

### DIFF
--- a/components/ai_chat/content/browser/associated_link_content.h
+++ b/components/ai_chat/content/browser/associated_link_content.h
@@ -1,0 +1,71 @@
+// Copyright (c) 2025 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#ifndef BRAVE_COMPONENTS_AI_CHAT_CONTENT_BROWSER_ASSOCIATED_LINK_CONTENT_H_
+#define BRAVE_COMPONENTS_AI_CHAT_CONTENT_BROWSER_ASSOCIATED_LINK_CONTENT_H_
+
+#include <memory>
+#include <string>
+
+#include "base/memory/weak_ptr.h"
+#include "base/one_shot_event.h"
+#include "base/timer/timer.h"
+#include "brave/components/ai_chat/core/browser/associated_content_delegate.h"
+#include "content/public/browser/web_contents_observer.h"
+#include "url/gurl.h"
+
+namespace content {
+class BrowserContext;
+class WebContents;
+class NavigationHandle;
+}  // namespace content
+
+namespace ai_chat {
+
+class PageContentFetcher;
+
+// Represents a link that has been attached to a conversation.
+// The link will be loaded asynchronously in a background WebContents when
+// |GetContent| is called.
+class AssociatedLinkContent : public AssociatedContentDelegate,
+                              public content::WebContentsObserver {
+ public:
+  AssociatedLinkContent(
+      GURL url,
+      std::u16string title,
+      content::BrowserContext* browser_context,
+      base::OnceCallback<void(content::WebContents*)> attach_tab_helpers);
+  ~AssociatedLinkContent() override;
+  AssociatedLinkContent(const AssociatedLinkContent&) = delete;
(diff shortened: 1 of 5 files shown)


components/ai_chat/content/browser/associated_link_content.h
```diff
@@ -0,0 +1,71 @@
...
+}  // namespace content
+
+namespace ai_chat {
+
+class PageContentFetcher;
+
+// Represents a link that has been attached to a conversation.
+// The link will be loaded asynchronously in a background WebContents when
+// |GetContent| is called.
+class AssociatedLinkContent : public AssociatedContentDelegate,
```
petemill: nit: maybe `AssociatedURLContent`? Link sounds like a UI element or a relationship when this could be completely standalone.

components/ai_chat/content/browser/associated_url_content.cc
```diff
@@ -0,0 +1,163 @@
...
+
+void AssociatedLinkContent::CompleteWithError(const std::string& error) {
+  // Note: We don't actually do anything with the error, just log it in debug
+  // mode.
+  DVLOG(2) << __func__ << "Background content loading failed: " << error;
+
+  timeout_timer_.Stop();
+
+  // Clear cached content
+  set_cached_page_content(PageContent());
```
petemill, line 161: Should we try to read the page content anyway? Even though the page didn't finish loading, we might have enough content.

components/ai_chat/content/browser/associated_url_content.cc
```diff
@@ -0,0 +1,163 @@
...
+}
+
+void AssociatedLinkContent::DocumentOnLoadCompletedInPrimaryMainFrame() {
+  DVLOG(2) << __func__ << "Page fully loaded for URL: " << url().spec();
+  SetTitle(web_contents_->GetTitle());
+
+  content_fetcher_->FetchPageContent(
+      /*invalidation_token=*/"",
+      base::BindOnce(&AssociatedLinkContent::OnContentExtractionComplete,
+                     weak_ptr_factory_.GetWeakPtr()));
```
petemill, line 129: We aren't including any of the features in AIChatTabHelper::GetPageContent: PDF, print preview extraction. If we care about that then do we either need to attach the AIChatTabHelper and use that, or move that functionality to a shared class? For the longer-term, I do intend to suggest we move implementations to Chromium's `page_content_annotations::FetchPageContext` but that is a little way off and it only gets the PDF bytes and not the text so we probably still need what's in `AIChatTabHelper`...

fallaciousreasoning: Yeah the problem with attaching an `AIChatTabHelper` is that it complicates the lifecycle significantly (what owns the `WebContents`). 

I'd like to separate out the loading logic but I think it'll be okay to do that in a follow up. For now, this will work, and I think it'll be good enough for the initial version.

components/ai_chat/content/browser/associated_url_content.cc
fallaciousreasoning: Sure yeah! Can't hurt to try :laughing:

components/ai_chat/content/browser/associated_url_content.cc
```diff
@@ -0,0 +1,163 @@
...
+
+AssociatedLinkContent::AssociatedLinkContent(
+    GURL url,
+    std::u16string title,
+    content::BrowserContext* browser_context,
+    base::OnceCallback<void(content::WebContents*)> attach_tab_helpers) {
+  DVLOG(2) << __func__ << "Creating link content for: " << url.spec()
+           << " title: " << title;
+
+  set_uuid(base::Uuid::GenerateRandomV4().AsLowercaseString());
```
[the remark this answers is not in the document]
fallaciousreasoning: Marking this as resolved so the PR can merge (with the understanding that security review is meant to be non blocking). FWIW, this is the same as other AssociatedContent classes. 

There's a full security review for this feature which will be completed before the feature is enabled.
 https://github.com/brave/reviews/issues/2033

PR #31449, opened 2025-09-26
[AI Chat]: Make it possible for AssociatedContentManager to own non-archive content

fallaciousreasoning:
Resolves https://github.com/brave/brave-browser/issues/49661

This has no functional changes but paves the way for the AssociatedContentManager to own non-archive content, such as Bookmarks and History


yrliou (approved):
LGTM

PR #31457, opened 2025-09-26
Hide Managed UI when there are only internal Brave policies

bbondy:
Resolves https://github.com/brave/brave-browser/issues/49679

### DIFF
--- a/chromium_src/chrome/browser/ui/managed_ui.cc
+++ b/chromium_src/chrome/browser/ui/managed_ui.cc
@@ -1,0 +1,54 @@
+/* Copyright (c) 2025 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include <algorithm>
+
+#include "chrome/browser/profiles/profile.h"
+#include "components/policy/core/common/policy_map.h"
+#include "components/policy/core/common/policy_namespace.h"
+#include "components/policy/core/common/policy_service.h"
+#include "components/policy/core/common/policy_types.h"
+
+#define ShouldDisplayManagedUi ShouldDisplayManagedUi_ChromiumImpl
+#include <chrome/browser/ui/managed_ui.cc>
+#undef ShouldDisplayManagedUi
+
+namespace brave_policy {
+
+bool ShouldHideManagedUI(const policy::PolicyMap& policies) {
+  // Empty policy map means no management at all
+  if (policies.empty()) {
+    return false;
+  }
+
+  // Check all policies, not just BraveOrigin-related ones
+  return std::ranges::all_of(policies, [](const auto& policy_pair) {
+    const auto& [policy_name, entry] = policy_pair;
+    return entry.source == policy::POLICY_SOURCE_BRAVE;
+  });
+}
+
+}  // namespace brave_policy
+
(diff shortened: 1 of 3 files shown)


chromium_src/chrome/browser/ui/managed_ui.cc
```diff
@@ -0,0 +1,54 @@
...
+  }
+  return true;
+}
+
+}  // namespace brave_policy
+
+// Our override implementation
+bool ShouldDisplayManagedUi(Profile* profile) {
+  // Check if we should hide due to Brave-only management
+  if (enterprise_util::IsBrowserManaged(profile)) {
```
DJAndries: nit: what if we call the original function and perform an early return instead of calling this function?

```
if (!ShouldDisplayManagedUi_ChromiumImpl(profile)) {
  return false;
}
```

chromium_src/chrome/browser/ui/managed_ui.cc
```diff
@@ -0,0 +1,54 @@
...
+    policy::PolicyService* policy_service =
+        profile->GetProfilePolicyConnector()->policy_service();
```
DJAndries: Do we also need to check browser-based policies?

bbondy: I checked that and it doesn't seem to. 
This did trigger something I want to look more into in later work for Origin though.

I noticed that chrome://policy doesn't list browser level policies, only the profile level ones.
That was known and not a problem.

When you do this:
`sudo /usr/libexec/PlistBuddy -c "Add :TorDisabled bool true" /Library/Managed\ Preferences/com.brave.Browser.development.plist`

It does list Tor in chrome://policy though. 
So I think it's being registered in the profile provider when you do that. 
And it does find the policy in the above code.

But I can't get the local state related policies to work in the profile provider in Brave Origin code. I retried and even swapped the browser/profile policies, and nothing gets set or disabled at all.  When I swapped, both the profile policies don't work and the browser policies don't work.

So this is something I need to dig in more. But I can't find a way to set a browser level policy outside of doing it through code. And this managedUi check seems to be only for profile related policies. 

The code seems to work as expected without passing in the browser policy service.

chromium_src/chrome/browser/ui/managed_ui.cc
bbondy: Sounds good. Updated in the latest commit.

chromium_src/chrome/browser/ui/managed_ui.cc
DJAndries: Very interesting. I'll try to dig in a little bit as well.

chromium_src/chrome/browser/ui/managed_ui.cc
```diff
@@ -0,0 +1,54 @@
...
+namespace brave_policy {
+
+bool ShouldHideManagedUI(const policy::PolicyMap& policies) {
+  // Empty policy map means no management at all
+  if (policies.empty()) {
+    return false;
+  }
+
+  // Check all policies, not just BraveOrigin-related ones
+  for (const auto& [policy_name, entry] : policies) {
```
DJAndries: nit: could use `std::ranges::any_of`

DJAndries (approved):
It looks like this branch might need a rebase, but looks good otherwise.

chromium_src/chrome/browser/ui/managed_ui.cc
bbondy: Updated in the latest

PR #31459, opened 2025-09-26
Revert "Hit all 4 edges in quick action search widget (uplift to 1.83.x)"

SergeyZhukovsky:
There are still issues with it per https://github.com/brave/brave-browser/issues/49110#issuecomment-3313641851. So better to revert and we are going to make a scroll there.
Reverts brave/brave-core#31035


kjozwiak (approved):
Uplift into `1.83.x` approved 👍

PR #31461, opened 2025-09-26
Fixes a crash in PIP.

mkarolin:
Resolves https://github.com/brave/brave-browser/issues/49558

### DIFF
--- a/browser/ui/BUILD.gn
+++ b/browser/ui/BUILD.gn
@@ -64,6 +64,7 @@ source_set("ui") {
     "//chrome/browser/ui:browser_navigator_params_headers",
     "//chrome/browser/ui/startup",
     "//chrome/browser/ui/tabs:tabs_public",
+    "//media",
   ]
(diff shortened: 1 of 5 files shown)


kjozwiak:

```
Brave | 1.85.8 Chromium: 141.0.7390.43 (Official Build) nightly (64-bit)
-- | --
Revision | 2a9706a72366eb979f3ac6304a3c07ad1658e06f
OS | Windows 11 Version 25H2 (Build 26200.6713)
```

Using the STR/Cases outlined via https://github.com/brave/brave-browser/issues/49558#issue-3446444961, reproduced the original crash using `1.85.4 Chromium: 141.0.7390.43`. Clicked on the `PiP` icon via the `Media` toolbar while watching a YT video and reproduced the crash.

Using the same STR/Cases outlined above, ensured that `Brave` doesn't crash when clicking on `PiP` as per the following:

image

PR #31465, opened 2025-09-26
Enforce Trusted Types and allow Lit's `lit-html-desktop` policy — now extended to iOS.

szilardszaloki:
Resolves https://github.com/brave/brave-browser/issues/49642.

On iOS/iPadOS 26 (Safari 26/WebKit update), Trusted Types enforcement is now active (see more on that [here](https://webkit.org/blog/17333/webkit-features-in-safari-26-0/)). With the default WebUI CSP on iOS, Lit can't create its trusted type policy (`lit-html-desktop`), which prevents it from loading and leaves `brave://account` blank. This change explicitly enforces Trusted Types and overrides `trusted-types` to allow Lit to initialize properly.

Note: desktop and Android already go through a similar Trusted Types path in `webui_util.cc` [here](https://source.chromium.org/chromium/chromium/src/+/main:ui/webui/webui_util.cc;l=95-102):
```cpp
void EnableTrustedTypesCSP(content::WebUIDataSource* source) {
      network::mojom::CSPDirectiveName::RequireTrustedTypesFor,
      "require-trusted-types-for 'script';");
      network::mojom::CSPDirectiveName::TrustedTypes,
      base::StrCat({kDefaultTrustedTypesPolicies, ";"}));
}
```
where [`kDefaultTrustedTypesPolicies`](https://source.chromium.org/chromium/chromium/src/+/main:ui/webui/webui_util.h;l=18-25) is:
```cpp
inline constexpr char kDefaultTrustedTypesPolicies[] =
    "trusted-types parse-html-subset sanitize-inner-html static-types "
    // Add TrustedTypes policies used during tests.
    "webui-test-script webui-test-html "
    // Add TrustedTypes policies necessary for using Polymer.
    "polymer-html-literal polymer-template-event-attribute-policy "
    // Add TrustedTypes policies necessary for using Desktop's Lit bundle.
    "lit-html-desktop";
```
and will also be overridden with this change in `brave_account_ui_base.h`.


kdenhartog (approved):
looks to be a good improvement to me

fmarier (approved):

PR #31477, opened 2025-09-29
[NTP] Cache rewards balance between page loads

zenparsing:
Resolves https://github.com/brave/brave-browser/issues/49267

### DIFF
--- a/browser/resources/brave_new_tab_page_refresh/components/widgets/rewards_widget.tsx
+++ b/browser/resources/brave_new_tab_page_refresh/components/widgets/rewards_widget.tsx
@@ -11,6 +11,7 @@ import Tooltip from '@brave/leo/react/tooltip'
 import { getString } from '../../lib/strings'
 import { useRewardsState, useRewardsActions } from '../../context/rewards_context'
 import { usePluralString } from '../../lib/plural_string'
+import { usePersistedJSON } from '$web-common/usePersistedState'
 import { WidgetMenu } from './widget_menu'
 import { Link, openLink } from '../common/link'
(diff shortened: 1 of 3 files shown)


browser/resources/brave_new_tab_page_refresh/components/widgets/rewards_widget.tsx
```diff
@@ -11,6 +11,7 @@ import Tooltip from '@brave/leo/react/tooltip'
 import { getString } from '../../lib/strings'
 import { useRewardsState, useRewardsActions } from '../../context/rewards_context'
 import { usePluralString } from '../../lib/plural_string'
+import { usePersistedJSON } from '$web-common/usePersistedState'
```
fallaciousreasoning, line 14: nioce!

browser/resources/brave_new_tab_page_refresh/components/widgets/rewards_widget.tsx
```diff
@@ -244,6 +254,29 @@ export function RewardsWidget() {
     )
   }
 
+  function renderBalance() {
+    const balance = rewardsBalance ?? cachedBalance
```
fallaciousreasoning, line 258: Hey, I think we could just always be rendering the `cachedBalance` as we set it every time `rewardsBalance` changes. Makes the logic (very mildly) simpler

fallaciousreasoning (approved):
lgtm % nits

browser/resources/brave_new_tab_page_refresh/components/widgets/rewards_widget.tsx
zenparsing: ah - good call!

PR #31485, opened 2025-09-29
Apply policies with Ad Block Only mode enabled

aseren:
When the Ad Block Only mode preference is enabled, PR applies the corresponding policies through
the Brave Policy Provider.

The Ad Block Only mode functionality is under a feature which is turned off by default.

Resolves https://github.com/brave/brave-browser/issues/48907

### DIFF
--- a/components/brave_policy/BUILD.gn
+++ b/components/brave_policy/BUILD.gn
@@ -16,6 +16,7 @@ static_library("brave_policy") {
     ":brave_policy_observer",
     "//base",
     "//brave/components/brave_origin",
+    "//brave/components/brave_policy/ad_block_only_mode",
     "//components/policy/core/browser",
     "//components/policy/core/common",
     "//components/prefs",
--- a/components/brave_policy/ad_block_only_mode/buildflags/BUILD.gn
+++ b/components/brave_policy/ad_block_only_mode/buildflags/BUILD.gn
@@ -1,0 +1,15 @@
+# Copyright (c) 2025 The Brave Authors. All rights reserved.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import(
+    "//brave/components/brave_policy/ad_block_only_mode/buildflags/buildflags.gni")
+import("//build/buildflag_header.gni")
+
+buildflag_header("buildflags") {
+  header = "buildflags.h"
+  flags = [
+    "ENABLE_AD_BLOCK_ONLY_MODE_POLICIES=$enable_ad_block_only_mode_policies",
+  ]
+}
--- a/components/brave_policy/ad_block_only_mode/buildflags/buildflags.gni
+++ b/components/brave_policy/ad_block_only_mode/buildflags/buildflags.gni
@@ -1,0 +1,10 @@
+# Copyright (c) 2025 The Brave Authors. All rights reserved.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+declare_args() {
+  # Controls whether Ad Block Only Mode can override some Brave browser policies
+  # for improved website compatibility. Disabled on iOS where not supported.
+  enable_ad_block_only_mode_policies = !is_ios
+}
--- a/components/brave_policy/brave_profile_policy_provider.cc
+++ b/components/brave_policy/brave_profile_policy_provider.cc
@@ -10,6 +10,7 @@
 #include "base/logging.h"
 #include "base/values.h"
 #include "brave/components/brave_origin/brave_origin_utils.h"
+#include "brave/components/brave_policy/ad_block_only_mode/ad_block_only_mode_policy_manager.h"
 #include "components/policy/core/common/policy_bundle.h"
 #include "components/policy/core/common/policy_map.h"
 #include "components/policy/core/common/policy_namespace.h"
@@ -33,6 +34,10 @@ void BraveProfilePolicyProvider::Init(policy::SchemaRegistry* registry) {
   // loading.
   brave_origin_observation_.Observe(
       brave_origin::BraveOriginPolicyManager::GetInstance());
+
+  // Register as AdBlockOnlyModePolicyManager observer.
+  ad_block_only_mode_policy_observation_.Observe(
+      AdBlockOnlyModePolicyManager::GetInstance());
 }
 
 void BraveProfilePolicyProvider::RefreshPolicies(
@@ -55,7 +60,12 @@ void BraveProfilePolicyProvider::OnBravePoliciesReady() {
 
   // Once we have Brave policies and a profile ID trigger Refresh policies
   if (!profile_id_.empty()) {
-    RefreshPolicies(policy::PolicyFetchReason::kBrowserStart);
+    // `OnBravePoliciesReady()` can be called multiple times, so we need to
+    // choose the appropriate reason for the policies refresh basing on the
+    // first policies load complete state.
+    RefreshPolicies(first_policies_loaded_
+                        ? policy::PolicyFetchReason::kUserRequest
+                        : policy::PolicyFetchReason::kBrowserStart);
   }
 }
 
@@ -71,6 +81,8 @@ policy::PolicyBundle BraveProfilePolicyProvider::LoadPolicies() {
 #else
   // Always disabled in official builds
 #endif
+
+  MaybeLoadAdBlockOnlyModePolicies(bundle);
 
   return bundle;
 }
@@ -109,6 +121,27 @@ void BraveProfilePolicyProvider::LoadBraveOriginPolicy(
   bundle_policy_map.Set(std::string(policy_key), policy::POLICY_LEVEL_MANDATORY,
                         policy::POLICY_SCOPE_USER, policy::POLICY_SOURCE_BRAVE,
                         base::Value(enabled), nullptr);
+}
+
+void BraveProfilePolicyProvider::MaybeLoadAdBlockOnlyModePolicies(
+    policy::PolicyBundle& bundle) {
+  auto ad_block_only_mode_policies =
+      AdBlockOnlyModePolicyManager::GetInstance()->GetPolicies();
+  // Ad Block Only mode policies can be empty if the feature is disabled or
+  // the Ad Block Only mode is not enabled.
+  if (ad_block_only_mode_policies.empty()) {
+    return;
+  }
+
+  policy::PolicyMap& bundle_policy_map = bundle.Get(
+      policy::PolicyNamespace(policy::POLICY_DOMAIN_CHROME, std::string()));
+
+  for (auto& [policy_key, value] : ad_block_only_mode_policies) {
+    bundle_policy_map.Set(
+        std::string(policy_key), policy::POLICY_LEVEL_MANDATORY,
+        policy::POLICY_SCOPE_USER, policy::POLICY_SOURCE_BRAVE,
+        std::move(value), nullptr);
+  }
 }
 
 void BraveProfilePolicyProvider::OnProfilePolicyChanged(
--- a/components/brave_policy/brave_profile_policy_provider.h
+++ b/components/brave_policy/brave_profile_policy_provider.h
@@ -13,6 +13,7 @@
 #include "base/values.h"
 #include "brave/components/brave_origin/brave_origin_policy_info.h"
 #include "brave/components/brave_origin/brave_origin_policy_manager.h"
+#include "brave/components/brave_policy/ad_block_only_mode/ad_block_only_mode_policy_manager.h"
 #include "brave/components/brave_policy/brave_policy_observer.h"
 #include "components/policy/core/common/configuration_policy_provider.h"
 
@@ -50,6 +51,8 @@ class BraveProfilePolicyProvider : public policy::ConfigurationPolicyProvider,
                              std::string_view policy_key,
                              bool enabled);
 
+  void MaybeLoadAdBlockOnlyModePolicies(policy::PolicyBundle& bundle);
+
   bool first_policies_loaded_ = false;
   bool policies_ready_ = false;
   std::string profile_id_;
@@ -57,6 +60,9 @@ class BraveProfilePolicyProvider : public policy::ConfigurationPolicyProvider,
   base::ScopedObservation<brave_origin::BraveOriginPolicyManager,
                           BravePolicyObserver>
       brave_origin_observation_{this};
+
+  base::ScopedObservation<AdBlockOnlyModePolicyManager, BravePolicyObserver>
+      ad_block_only_mode_policy_observation_{this};
 
   base::WeakPtrFactory<BraveProfilePolicyProvider> weak_factory_{this};
 };
--- a/browser/DEPS
+++ b/browser/DEPS
@@ -35,6 +35,7 @@ include_rules += [
   "+brave/components/brave_news/browser",
   "+brave/components/brave_news/common",
   "+brave/components/brave_origin",
+  "+brave/components/brave_policy/ad_block_only_mode",
   "+brave/components/brave_rewards/content",
   "+brave/components/brave_rewards/core",
   "+brave/components/brave_shields/content/browser",
--- a/browser/brave_browser_process_impl.cc
+++ b/browser/brave_browser_process_impl.cc
@@ -31,6 +31,7 @@
 #include "brave/components/brave_component_updater/browser/brave_component_updater_delegate.h"
 #include "brave/components/brave_component_updater/browser/local_data_files_service.h"
 #include "brave/components/brave_origin/brave_origin_policy_manager.h"
+#include "brave/components/brave_policy/ad_block_only_mode/ad_block_only_mode_policy_manager.h"
 #include "brave/components/brave_referrals/browser/brave_referrals_service.h"
 #include "brave/components/brave_shields/content/browser/ad_block_service.h"
 #include "brave/components/brave_shields/content/browser/ad_block_subscription_service_manager.h"
@@ -192,6 +193,10 @@ void BraveBrowserProcessImpl::Init() {
     CreateAIChatAgentProfileManager();
   }
 #endif
+
+  // Lazy initialization of AdBlockOnlyModePolicyManager
+  brave_policy::AdBlockOnlyModePolicyManager::GetInstance()->Init(
+      local_state());
 }
 
 void BraveBrowserProcessImpl::PreMainMessageLoopRun() {
@@ -220,6 +225,7 @@ void BraveBrowserProcessImpl::StartTearDown() {
 #endif
   // Reset BraveOriginPolicyManager to prevent dangling pointer to local_state_
   brave_origin::BraveOriginPolicyManager::GetInstance()->Shutdown();
+  brave_policy::AdBlockOnlyModePolicyManager::GetInstance()->Shutdown();
   brave_sync::NetworkTimeHelper::GetInstance()->Shutdown();
   BrowserProcessImpl::StartTearDown();
 }
--- a/browser/sources.gni
+++ b/browser/sources.gni
@@ -167,6 +167,7 @@ brave_chrome_browser_deps = [
   "//brave/components/brave_origin",
   "//brave/components/brave_perf_predictor/browser",
   "//brave/components/brave_policy",
+  "//brave/components/brave_policy/ad_block_only_mode",
   "//brave/components/brave_private_new_tab_ui/common",
   "//brave/components/brave_private_new_tab_ui/common:mojom",
   "//brave/components/brave_referrals/browser",
--- a/components/BUILD.gn
+++ b/components/BUILD.gn
@@ -28,6 +28,7 @@ test("brave_components_unittests") {
     "//brave/components/brave_account/endpoints:unit_tests",
     "//brave/components/brave_origin:unit_tests",
     "//brave/components/brave_policy:unit_tests",
+    "//brave/components/brave_policy/ad_block_only_mode:unit_tests",
     "//brave/components/brave_shields/core/browser:unit_tests",
     "//brave/components/brave_shields/core/common:unit_tests",
     "//brave/components/brave_wallet/common:unit_tests",
--- a/components/brave_policy/ad_block_only_mode/BUILD.gn
+++ b/components/brave_policy/ad_block_only_mode/BUILD.gn
@@ -1,0 +1,46 @@
+# Copyright (c) 2025 The Brave Authors. All rights reserved.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import(
+    "//brave/components/brave_policy/ad_block_only_mode/buildflags/buildflags.gni")
+
+static_library("ad_block_only_mode") {
+  public_deps =
+      [ "//brave/components/brave_policy/ad_block_only_mode/buildflags" ]
+
+  sources = [
+    "ad_block_only_mode_policy_manager.cc",
+    "ad_block_only_mode_policy_manager.h",
+  ]
+
+  deps = [
+    "//base",
+    "//brave/components/brave_policy:brave_policy_observer",
+    "//brave/components/brave_shields/core/common",
+    "//components/content_settings/core/common",
+    "//components/policy/core/browser",
+    "//components/policy/core/common",
+    "//components/prefs",
+  ]
+}
+
+source_set("unit_tests") {
+  testonly = true
+
+  if (enable_ad_block_only_mode_policies) {
+    sources = [ "ad_block_only_mode_policy_manager_unittest.cc" ]
+
+    deps = [
+      ":ad_block_only_mode",
+      "//base/test:test_support",
+      "//brave/components/brave_policy:brave_policy_observer",
+      "//brave/components/brave_shields/core/common",
+      "//components/prefs",
+      "//components/prefs:test_support",
+      "//testing/gmock",
+      "//testing/gtest",
+    ]
+  }
+}
--- a/components/brave_policy/ad_block_only_mode/DEPS
+++ b/components/brave_policy/ad_block_only_mode/DEPS
@@ -1,0 +1,3 @@
+include_rules = [
+  "+components/content_settings",
+]
--- a/components/brave_policy/ad_block_only_mode/ad_block_only_mode_policy_manager.cc
+++ b/components/brave_policy/ad_block_only_mode/ad_block_only_mode_policy_manager.cc
@@ -1,0 +1,113 @@
+/* Copyright (c) 2025 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "brave/components/brave_policy/ad_block_only_mode/ad_block_only_mode_policy_manager.h"
+
+#include "base/feature_list.h"
+#include "base/no_destructor.h"
+#include "brave/components/brave_policy/ad_block_only_mode/buildflags/buildflags.h"
+#include "brave/components/brave_shields/core/common/features.h"
+#include "brave/components/brave_shields/core/common/pref_names.h"
+#include "components/content_settings/core/common/content_settings.h"
+#include "components/policy/policy_constants.h"
+#include "components/prefs/pref_service.h"
+
+namespace brave_policy {
+
+// static
+AdBlockOnlyModePolicyManager* AdBlockOnlyModePolicyManager::GetInstance() {
+  static base::NoDestructor<AdBlockOnlyModePolicyManager> instance;
+  return instance.get();
+}
+
+void AdBlockOnlyModePolicyManager::Init(PrefService* local_state) {
+  CHECK(local_state) << "AdBlockOnlyModePolicyManager local state should exist";
+
+  local_state_ = local_state;
+
+  pref_change_registrar_.Init(local_state_);
+  pref_change_registrar_.Add(
+      brave_shields::prefs::kAdBlockOnlyModeEnabled,
+      base::BindRepeating(
+          &AdBlockOnlyModePolicyManager::OnAdBlockOnlyModeChanged,
+          base::Unretained(this)));
+
+  OnAdBlockOnlyModeChanged();
+}
+
+void AdBlockOnlyModePolicyManager::Shutdown() {
+  pref_change_registrar_.RemoveAll();
+  observers_.Clear();
+  local_state_ = nullptr;
+}
+
+void AdBlockOnlyModePolicyManager::AddObserver(BravePolicyObserver* observer) {
+  observers_.AddObserver(observer);
+
+  if (local_state_) {
+    // Notify the observer to fetch Ad Block Only mode policies.
+    observer->OnBravePoliciesReady();
+  }
+}
+
+void AdBlockOnlyModePolicyManager::RemoveObserver(
+    BravePolicyObserver* observer) {
+  observers_.RemoveObserver(observer);
+}
+
+AdBlockOnlyModePolicies AdBlockOnlyModePolicyManager::GetPolicies() const {
+  if (!base::FeatureList::IsEnabled(
+          brave_shields::features::kAdblockOnlyMode)) {
+    return {};
+  }
+
+  if (!local_state_ || !local_state_->GetBoolean(
+                           brave_shields::prefs::kAdBlockOnlyModeEnabled)) {
+    return {};
+  }
+
+  return GetPoliciesImpl();
+}
+
+AdBlockOnlyModePolicyManager::AdBlockOnlyModePolicyManager() = default;
+
+AdBlockOnlyModePolicyManager::~AdBlockOnlyModePolicyManager() = default;
+
+void AdBlockOnlyModePolicyManager::OnAdBlockOnlyModeChanged() {
+  observers_.Notify(&BravePolicyObserver::OnBravePoliciesReady);
+}
+
+AdBlockOnlyModePolicies AdBlockOnlyModePolicyManager::GetPoliciesImpl() const {
+#if BUILDFLAG(ENABLE_AD_BLOCK_ONLY_MODE_POLICIES)
+  AdBlockOnlyModePolicies policies;
+
+  // Allow JavaScript globally.
+  policies.emplace(policy::key::kDefaultJavaScriptSetting,
+                   base::Value(CONTENT_SETTING_ALLOW));
+
+  // Allow all cookies.
+  policies.emplace(policy::key::kDefaultCookiesSetting,
+                   base::Value(CONTENT_SETTING_ALLOW));
+
+  // Do not block third-party cookies.
+  policies.emplace(policy::key::kBlockThirdPartyCookies, base::Value(false));
+
+  // Disable language fingerprinting reduction.
+  policies.emplace(policy::key::kBraveReduceLanguageEnabled,
+                   base::Value(false));
+
+  // Disable De-AMP.
+  policies.emplace(policy::key::kBraveDeAmpEnabled, base::Value(false));
+
+  // Disable URL debouncing.
+  policies.emplace(policy::key::kBraveDebouncingEnabled, base::Value(false));
+
+  return policies;
+#else
+  return {};
+#endif  // BUILDFLAG(ENABLE_AD_BLOCK_ONLY_MODE_POLICIES)
+}
+
+}  // namespace brave_policy
--- a/components/brave_policy/ad_block_only_mode/ad_block_only_mode_policy_manager.h
+++ b/components/brave_policy/ad_block_only_mode/ad_block_only_mode_policy_manager.h
@@ -1,0 +1,63 @@
+/* Copyright (c) 2025 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#ifndef BRAVE_COMPONENTS_BRAVE_POLICY_AD_BLOCK_ONLY_MODE_AD_BLOCK_ONLY_MODE_POLICY_MANAGER_H_
+#define BRAVE_COMPONENTS_BRAVE_POLICY_AD_BLOCK_ONLY_MODE_AD_BLOCK_ONLY_MODE_POLICY_MANAGER_H_
+
+#include "base/containers/flat_map.h"
+#include "base/memory/raw_ptr.h"
+#include "base/observer_list.h"
+#include "base/values.h"
+#include "brave/components/brave_policy/brave_policy_observer.h"
+#include "components/prefs/pref_change_registrar.h"
+
+namespace base {
+template <typename T>
+class NoDestructor;
+}
+
+class PrefService;
+
+namespace brave_policy {
+
+using AdBlockOnlyModePolicies = base::flat_map<std::string, base::Value>;
+
+// Singleton that holds Ad Block Only mode preference handling and manages
+// setting policy values when the preference is changed. This
+// abstracts away the local state management from policy provider.
+// TODO(https://github.com/brave/brave-browser/issues/50077): Refactor this
+// class when `BravePolicyManager` is introduced.
+class AdBlockOnlyModePolicyManager final {
+ public:
+  static AdBlockOnlyModePolicyManager* GetInstance();
+
+  void Init(PrefService* local_state);
+  void Shutdown();
+
+  void AddObserver(BravePolicyObserver* observer);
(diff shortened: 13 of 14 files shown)


components/brave_shields/core/common/ad_block_only_mode/ad_block_only_mode_policy_provider.cc
```diff
@@ -0,0 +1,100 @@
...
+  SetPolicy(policies, policy::key::kBraveDeAmpEnabled, base::Value(false));
+
+  // Disable URL debouncing.
+  SetPolicy(policies, policy::key::kBraveDebouncingEnabled, base::Value(false));
+#endif  // !defined(IS_IOS)
+}
+
+}  // namespace
+
+AdBlockOnlyModePolicyProvider::AdBlockOnlyModePolicyProvider(
```
bridiver: This doesn't seem right, I don't think we want a separate policy provider? I think we want something like BraveOriginPolicyManager? Need to think about this, wdyt @bbondy ?

components/brave_shields/core/common/ad_block_only_mode/ad_block_only_mode_policy_provider.cc
```diff
@@ -0,0 +1,100 @@
...
+  pref_change_registrar_.Add(
+      prefs::kAdBlockOnlyModeEnabled,
+      base::BindRepeating(
+          &AdBlockOnlyModePolicyProvider::OnAdBlockOnlyModeChanged,
+          weak_factory_.GetWeakPtr()));
+}
+
+void AdBlockOnlyModePolicyProvider::MaybeLoadPolicies(
+    policy::PolicyBundle& bundle) {
+  if (!IsAdblockOnlyModeFeatureEnabled()) {
```
bridiver: please avoid wrapping simple feature checks in methods, in particular when they pull in additional deps that would otherwise not be needed. This kind of method (if it exists at all) should be features.h in a separate `features` target so it doesn't pull in any more dependencies than necessary.

components/brave_shields/core/common/ad_block_only_mode/ad_block_only_mode_policy_provider.h
```diff
@@ -0,0 +1,52 @@
...
+class PrefService;
+
+namespace policy {
+class ConfigurationPolicyProvider;
+class PolicyBundle;
+}  // namespace policy
+
+namespace brave_shields {
+
+class AdBlockOnlyModePolicyProvider {
```
bridiver: per above this doesn't even appear to be a policy provider because it doesn't subclass any type of policy provider.

components/brave_shields/core/common/ad_block_only_mode/ad_block_only_mode_policy_provider.cc
```diff
@@ -0,0 +1,100 @@
...
+void SetPolicy(policy::PolicyMap& policies,
+               std::string_view policy_key,
+               base::Value value) {
+  policies.Set(policy_key.data(), policy::POLICY_LEVEL_MANDATORY,
+               policy::POLICY_SCOPE_USER, policy::POLICY_SOURCE_BRAVE,
+               std::move(value), nullptr);
+}
+
+void SetPolicies(policy::PolicyMap& policies) {
+#if !defined(IS_IOS)
```
bridiver: why is this guard here?

components/brave_shields/core/common/ad_block_only_mode/ad_block_only_mode_policy_provider.cc
aseren: `AdBlockOnlyModePolicyProvider` functionality is closer to `BraveProfilePolicyProvider` and just needs to load specific profile level policies. 
Currently `AdBlockOnlyModePolicyProvider::MaybeLoadPolicies` is being called from `BraveProfilePolicyProvider::LoadPolicies()`.

aseren: yeah, naming is quite confusing, so maybe `AdBlockOnlyModePoliciesLoader` would make more sense.

components/brave_shields/core/common/ad_block_only_mode/ad_block_only_mode_policy_provider.h
aseren: yeah, naming is quite confusing, so maybe `AdBlockOnlyModePolicyLoader` would make more sense.

components/brave_shields/core/common/ad_block_only_mode/ad_block_only_mode_policy_provider.cc
aseren: ~~I did it because CI for iOS gave me a compilation error with these policies, as they are not defined for iOS.
The strange thing is that compilation passes for me locally. Now I'm trying to find why.~~

~~The guard was triggered by tests. Now I removed the guard as it is no longer needed, because I moved the test to a proper target.~~

components/brave_shields/core/common/ad_block_only_mode/ad_block_only_mode_policy_provider.cc
aseren: makes sense

aseren: Fixed

components/brave_shields/core/common/ad_block_only_mode/ad_block_only_mode_policy_provider.h
aseren: Renamed the class to `AdBlockOnlyModePoliciesLoader` so now it makes more sense now.

components/brave_shields/core/common/ad_block_only_mode/ad_block_only_mode_policy_provider.cc
aseren: Renamed the class to AdBlockOnlyModePoliciesLoader so it makes more sense now.

components/brave_shields/core/common/ad_block_only_mode/ad_block_only_mode_policy_provider.cc
aseren: `AdBlockOnlyModePoliciesLoader` is compiled for all platforms, including iOS. 
Guard is needed because these policies are not supported for iOS for now.

aseren: Guard itself was wrong, I changed it to `#if !BUILDFLAG(IS_IOS)`

components/brave_shields/core/common/ad_block_only_mode/ad_block_only_mode_policy_provider.cc
bbondy: why don't we just use `components/brave_policy/brave_profile_policy_provider.h` /.cc instead of a new provider?

aseren: yeah, `AdBlockOnlyModePolicyProvider` is an old class name which is confusing.  Now I renamed it to `AdBlockOnlyModePoliciesLoader` which makes more sense it just loads `Ad Block Only mode` policies. 

We are actually using `BraveProfilePolicyProvider` here because `AdBlockOnlyModePolicyProvider:: MaybeLoadPolicies` is being called from `BraveProfilePolicyProvider::LoadPolicies()`: https://github.com/brave/brave-core/pull/31485/files#diff-038f53c17e6e112a2f01c61d67772688b188e588efc629a07431a9e560b0166bR84

components/brave_policy/ad_block_only_mode/ad_block_only_mode_policies_utils.cc
```diff
@@ -0,0 +1,106 @@
+/* Copyright (c) 2025 The Brave Authors. All rights reserved.
```
bbondy: I think you want this file to be a singleton that follows the same pattern as `components/brave_origin/brave_origin_policy_manager.cc`/`.h`
You could refactor out and generalize the `BraveOriginPolicyManager` to be `BravePolicyManager` or leave that for a future refactor and just follow the same pattern for now. If you leave it until later just post an issue to track that at least for now. 

It wouldn't need a dependency on `components/policy` you could just have it track the policies to set.

chromium_src/chrome/browser/policy/profile_policy_connector.cc
```diff
@@ -3,22 +3,25 @@
...
 // Create and Init the Brave Profile Policy Provider which is used
 // for Brave Origin and other Brave specific policies.
 // We do not need to define a patch for Shutdown since anything in
 // wrapped_policy_providers_ will automatically call Shutdown.
-#define BRAVE_PROFILE_POLICY_CONNECTOR_INIT                         \
-  auto provider = brave_policy::CreateBraveProfilePolicyProvider(); \
-  brave_profile_policy_provider_ = provider.get();                  \
-  policy_providers_.push_back(provider.get());                      \
-  provider->Init(schema_registry);                                  \
+#define BRAVE_PROFILE_POLICY_CONNECTOR_INIT                       \
```
bbondy: I think you can get away with no changes to chromium_src overrides. I.e. you don't need to have `local_state` passed into the profile policy provider.

components/brave_policy/brave_profile_policy_provider.cc
```diff
@@ -33,6 +40,8 @@ void BraveProfilePolicyProvider::Init(policy::SchemaRegistry* registry) {
   // loading.
   brave_origin_observation_.Observe(
       brave_origin::BraveOriginPolicyManager::GetInstance());
+
+  ad_block_only_mode_policies_loader_->Init();
```
bbondy: Follow pattern of `brave_origin::BraveOriginPolicyManager` instead to not need local state.

components/brave_policy/brave_profile_policy_provider.cc
```diff
@@ -72,12 +81,14 @@ policy::PolicyBundle BraveProfilePolicyProvider::LoadPolicies() {
...
+  ad_block_only_mode_policies_loader_->MaybeLoadPolicies(bundle);
+
   return bundle;
 }
 
 std::unique_ptr<policy::ConfigurationPolicyProvider>
-CreateBraveProfilePolicyProvider() {
-  return std::make_unique<BraveProfilePolicyProvider>();
+CreateBraveProfilePolicyProvider(PrefService* local_state) {
+  return std::make_unique<BraveProfilePolicyProvider>(local_state);
```
bbondy: not needed

components/brave_policy/brave_profile_policy_provider.h
```diff
@@ -58,6 +66,9 @@ class BraveProfilePolicyProvider
                           brave_origin::BraveOriginPolicyManager::Observer>
       brave_origin_observation_{this};
 
+  std::unique_ptr<brave_shields::AdBlockOnlyModePoliciesLoader>
+      ad_block_only_mode_policies_loader_;
```
bbondy: not needed if it's a singleton

components/brave_policy/brave_profile_policy_provider.h
```diff
@@ -23,7 +29,7 @@ class BraveProfilePolicyProvider
     : public policy::ConfigurationPolicyProvider,
       public brave_origin::BraveOriginPolicyManager::Observer {
  public:
-  BraveProfilePolicyProvider();
+  explicit BraveProfilePolicyProvider(PrefService* local_state);
```
bbondy: not needed

components/brave_policy/brave_profile_policy_provider.h
```diff
@@ -15,6 +15,12 @@
 #include "brave/components/brave_origin/brave_origin_policy_manager.h"
 #include "components/policy/core/common/configuration_policy_provider.h"
 
+class PrefService;
+
+namespace brave_shields {
+class AdBlockOnlyModePoliciesLoader;
+}
```
bbondy: I think you can remove this and just include the header with the observer.

components/brave_policy/brave_profile_policy_provider_unittest.cc
```diff
@@ -19,7 +19,7 @@ namespace brave_policy {
 
 class BraveProfilePolicyProviderTest : public ::testing::Test {
  public:
-  BraveProfilePolicyProviderTest() = default;
+  BraveProfilePolicyProviderTest() : provider_(&local_state_) {}
```
bbondy: change not needed

components/brave_policy/brave_profile_policy_provider_unittest.cc
```diff
@@ -31,6 +31,7 @@ class BraveProfilePolicyProviderTest : public ::testing::Test {
  protected:
   base::test::TaskEnvironment task_environment_;
   policy::SchemaRegistry schema_registry_;
+  TestingPrefServiceSimple local_state_;
```
bbondy: change not needed

components/brave_policy/ad_block_only_mode/ad_block_only_mode_policies_utils.cc
```diff
@@ -0,0 +1,106 @@
...
+#include "base/feature_list.h"
+#include "base/values.h"
+#include "brave/components/brave_shields/core/common/features.h"
+#include "brave/components/brave_shields/core/common/pref_names.h"
+#include "components/content_settings/core/common/content_settings.h"
+#include "components/policy/core/common/configuration_policy_provider.h"
+#include "components/policy/core/common/policy_bundle.h"
+#include "components/policy/core/common/policy_map.h"
+#include "components/policy/core/common/policy_namespace.h"
+#include "components/policy/core/common/policy_types.h"
```
bbondy: None of these components/ policy will be needed after refactoring.

components/brave_shields/core/browser/ad_block_only_mode/ad_block_only_mode_policies_loader.cc
```diff
@@ -0,0 +1,106 @@
...
+    return;
+  }
+
+  if (!local_state_->GetBoolean(prefs::kAdBlockOnlyModeEnabled)) {
+    return;
+  }
+
+  policy::PolicyMap& policies = bundle.Get(
+      policy::PolicyNamespace(policy::POLICY_DOMAIN_CHROME, std::string()));
+  SetPolicies(policies);
```
bbondy: code to set the policy can go in the profile provider.

bbondy (changes requested):

components/brave_shields/core/common/ad_block_only_mode/ad_block_only_mode_policy_provider.cc
aseren: `AdBlockOnlyModePolicyLoader` is now removed. 
Now we manage AOM policies via `AdBlockOnlyModePolicyManager` and `BraveProfilePolicyProvider`

components/brave_shields/core/common/ad_block_only_mode/ad_block_only_mode_policy_provider.h
aseren: `AdBlockOnlyModePolicyLoader` is now removed. 
Now we manage AOM policies via `AdBlockOnlyModePolicyManager` and `BraveProfilePolicyProvider`

components/brave_shields/core/common/ad_block_only_mode/ad_block_only_mode_policy_provider.cc
aseren: `AdBlockOnlyModePolicyLoader` is now removed. 
Now we manage AOM policies via `AdBlockOnlyModePolicyManager` and `BraveProfilePolicyProvider`

components/brave_policy/ad_block_only_mode/ad_block_only_mode_policies_utils.cc
aseren: Introduced `AdBlockOnlyModePolicyManager` which notifies`BraveProfilePolicyProvider` when AOM policies needs to be updated. This follows `BraveOriginPolicyManager` pattern.

Also introduced `AdBlockOnlyModeService` which calls `AdBlockOnlyModePolicyManager` when Ad Block Only mode is enabled/disabled (via preference). This follows `BraveOriginService` pattern.

Thanks!

chromium_src/chrome/browser/policy/profile_policy_connector.cc
aseren: Removed, thanks!

components/brave_policy/brave_profile_policy_provider.cc
aseren: Removed, thanks!

components/brave_policy/brave_profile_policy_provider.cc
aseren: Removed, thanks!

components/brave_policy/brave_profile_policy_provider.h
aseren: Removed, thanks!

components/brave_policy/brave_profile_policy_provider.h
aseren: Removed

components/brave_policy/brave_profile_policy_provider.h
aseren: Removed

components/brave_policy/brave_profile_policy_provider_unittest.cc
aseren: Removed, thanks!

components/brave_policy/brave_profile_policy_provider_unittest.cc
aseren: Removed, thanks!

components/brave_shields/core/browser/ad_block_only_mode/ad_block_only_mode_policies_loader.cc
aseren: Moved this code to `components/brave_policy/ad_block_only_mode/ad_block_only_mode_policies_utils.cc`

components/brave_policy/ad_block_only_mode/ad_block_only_mode_policies_utils.cc
aseren: Moved this code to `components/brave_policy/ad_block_only_mode/ad_block_only_mode_policies_utils.cc`

components/brave_policy/ad_block_only_mode/ad_block_only_mode_policies_utils.cc
bbondy: I think that the `AdBlockOnlyModeService` is probably not needed. Would it work to just have the pref change registrar in the manager instead?

aseren: Thanks for comment! 

`AdBlockOnlyModeService` allows to keep and operate the `profile_id` which is used currently used in `BraveOriginPolicyManager`.
So the using `AdBlockOnlyModeService` well fits with existing logic in `BraveOriginPolicyManager` where we don't set policies until `profile_id` is not empty.

I'm also planning to use `AdBlockOnlyModeService` in future with other AOM functionality.

aseren: `AdBlockOnlyModeService` as discuss in DM

aseren: Filed the issue for refactoring BraveOriginPolicyManager to BravePolicyManager: https://github.com/brave/brave-browser/issues/50077

components/brave_origin/ad_block_only_mode_policy_manager.cc
```diff
@@ -0,0 +1,86 @@
...
+}
+
+void AdBlockOnlyModePolicyManager::AddObserver(Observer* observer) {
+  observers_.AddObserver(observer);
+
+  if (local_state_ &&
+      base::FeatureList::IsEnabled(brave_shields::features::kAdblockOnlyMode)) {
+    // Notify the observer about Ad Block Only mode if local state is already
+    // set.
+    observer->OnAdBlockOnlyModePoliciesChanged();
```
bbondy: nit: Maybe keep it consitent with the other call in the same file:
`observers_.Notify(&Observer::OnAdBlockOnlyModePoliciesChanged);`

aseren: Ah, I just realized that if we do it, then all other observers will be notified that preference was changed, which is not true. So I think it is better to notify the current observer only which the current code does. This also follows pattern from BraveOriginPolicyManager. Thanks!

bbondy: oh you're right. Thx

aseren:
Android build keeps failing, so I'm rebasing to latest master branch to try to fix it.

components/brave_policy/brave_profile_policy_provider.cc
```diff
@@ -111,6 +118,16 @@ void BraveProfilePolicyProvider::LoadBraveOriginPolicy(
                         base::Value(enabled), nullptr);
 }
 
+void BraveProfilePolicyProvider::MaybeLoadAdBlockOnlyModePolicies(
+    policy::PolicyBundle& bundle) {
+  if (!brave_origin::AdBlockOnlyModePolicyManager::GetInstance()
+           ->IsAdBlockOnlyModeEnabled()) {
+    return;
```
bridiver, line 133: do we need to do anything to unload previous policies?

components/brave_policy/brave_profile_policy_provider.cc
```diff
@@ -120,6 +137,13 @@ void BraveProfilePolicyProvider::OnProfilePolicyChanged(
   }
 }
 
+void BraveProfilePolicyProvider::OnAdBlockOnlyModePoliciesChanged() {
+  // Only update if we have a profile ID.
+  if (!profile_id_.empty()) {
```
bridiver: why would this be empty?

components/brave_policy/brave_profile_policy_provider.cc
aseren: No, we just don't need to load them with the current `policy::PolicyBundle& bundle`.

components/brave_policy/brave_profile_policy_provider.cc
aseren: This follows the pattern from `BraveProfilePolicyProvider::OnProfilePolicyChanged()` and `BraveProfilePolicyProvider::OnBraveOriginPoliciesReady()` in case if they are called before `BraveProfilePolicyProvider::SetProfileID` is called.

bridiver: @bbondy should this be a CHECK instead? If this can potentially happen in normal use then shouldn't we remember that it was called and trigger the updates when profile id is set? If not then we probably need some additional comments

aseren: yep, we do refresh policies when profile id is set:

```
void BraveProfilePolicyProvider::SetProfileID(const std::string& profile_id) {
  profile_id_ = profile_id;

  // If policies are ready already and we now have a profile_Id, refresh the
  // policies
  if (policies_ready_) {
    RefreshPolicies(policy::PolicyFetchReason::kBrowserStart);
  }
}
```

components/brave_policy/ad_block_only_mode/ad_block_only_mode_policies_utils.cc
```diff
@@ -0,0 +1,55 @@
...
+#include "build/build_config.h"
+#include "components/content_settings/core/common/content_settings.h"
+#include "components/policy/core/common/policy_bundle.h"
+#include "components/policy/core/common/policy_map.h"
+#include "components/policy/core/common/policy_namespace.h"
+#include "components/policy/core/common/policy_types.h"
+#include "components/policy/policy_constants.h"
+
+void LoadAdBlockOnlyModePolicies(policy::PolicyBundle& bundle) {
+  // This guard is necessary because the Ad Block Only mode policies below are
```
bridiver: os guards should only be used for os-specific code, not to guard feature-specific code that doesn't apply to all platforms. I'm not clear on why these need a buildflag guard in the first place, does ios not build if you remove it?

components/brave_policy/ad_block_only_mode/ad_block_only_mode_policies_utils.cc
```diff
@@ -0,0 +1,55 @@
...
+#include "base/values.h"
+#include "build/build_config.h"
+#include "components/content_settings/core/common/content_settings.h"
+#include "components/policy/core/common/policy_bundle.h"
+#include "components/policy/core/common/policy_map.h"
+#include "components/policy/core/common/policy_namespace.h"
+#include "components/policy/core/common/policy_types.h"
+#include "components/policy/policy_constants.h"
+
+void LoadAdBlockOnlyModePolicies(policy::PolicyBundle& bundle) {
```
bridiver: why is this in a separate utils file?

components/brave_policy/brave_profile_policy_provider.cc
bridiver: but is it expected that it will call this method before profile id is set?

components/brave_policy/BUILD.gn
```diff
@@ -5,6 +5,8 @@
 
 static_library("brave_policy") {
   sources = [
+    "ad_block_only_mode/ad_block_only_mode_policies_utils.cc",
```
bridiver: per my other comment it's not clear why these files are needed, but each directory should have its own BUILD.gn

components/brave_policy/ad_block_only_mode/ad_block_only_mode_policies_utils.cc
aseren: yes, the iOS build fails if guard is removed. These policies are not available on iOS.

bridiver: why is the provider used at all on ios is none of the policies work on ios

components/brave_origin/ad_block_only_mode_policy_manager.cc
```diff
@@ -0,0 +1,86 @@
...
+#include "base/feature_list.h"
+#include "base/no_destructor.h"
+#include "brave/components/brave_shields/core/common/features.h"
+#include "brave/components/brave_shields/core/common/pref_names.h"
+#include "components/prefs/pref_service.h"
+
+namespace brave_origin {
+
+// static
+AdBlockOnlyModePolicyManager* AdBlockOnlyModePolicyManager::GetInstance() {
```
bridiver: why is this in brave_origin?

bridiver: this seems like it should be in brave_shields?

components/brave_policy/ad_block_only_mode/ad_block_only_mode_policies_utils.cc
aseren: I wanted to separate code which sets `ad block only mode` policies from `BraveProfilePolicyProvider`. So I moved it to a separate util file and added to separate directory. Do you think it is better to move it to parent `components/brave_policy` directory?

components/brave_policy/BUILD.gn
aseren: I wanted to separate code which sets `ad block only mode` policies from `BraveProfilePolicyProvider`. So I moved it to a separate util file and added to separate directory. Do you think it is better to move it to parent `components/brave_policy` directory?

components/brave_origin/ad_block_only_mode_policy_manager.cc
```diff
@@ -0,0 +1,86 @@
...
+void AdBlockOnlyModePolicyManager::RemoveObserver(Observer* observer) {
+  observers_.RemoveObserver(observer);
+}
+
+bool AdBlockOnlyModePolicyManager::IsAdBlockOnlyModeEnabled() const {
+  if (!local_state_) {
+    return false;
+  }
+
+  return base::FeatureList::IsEnabled(
```
bridiver: shouldn't this entire policy be behind the feature flag?

browser/brave_origin/brave_origin_service_factory.cc
```diff
@@ -193,6 +194,13 @@ BraveOriginServiceFactory::BuildServiceInstanceForBrowserContext(
                          g_browser_process->local_state());
   }
 
+  // Lazy initialization of AdBlockOnlyModePolicyManager
+  auto* ad_block_only_mode_policy_manager =
```
bridiver: why is adblock only mode inside brave origin?

components/brave_origin/ad_block_only_mode_policy_manager.cc
aseren: I did it after discussion with @bbondy that BraveOriginPolicyManager will be generalized to BravePolicyManager. And AdBlockOnlyModePolicyManager can stay here until it is done. There is a ticket for generalization: https://github.com/brave/brave-browser/issues/50077

browser/brave_origin/brave_origin_service_factory.cc
aseren: I did it after discussion with @bbondy that BraveOriginPolicyManager will be generalized to BravePolicyManager. And AdBlockOnlyModePolicyManager can stay here until it is done. There is a ticket for generalization: https://github.com/brave/brave-browser/issues/50077

components/brave_policy/ad_block_only_mode/ad_block_only_mode_policies_utils.cc
bridiver: in any case if there any buildflag, it should not be an os buildflag because this is not desktop specific code, it's feature specific code and should its own buildflag. Since it doesn't do anything at all on ios now probably the entire thing should go behind the buildflag

components/brave_origin/ad_block_only_mode_policy_manager.cc
bridiver: putting the feature flag inside the code like this defeats part of the purpose of having buildflags which is to be able to disable the code if there are problems with it. This only disables parts of the code.

components/brave_policy/ad_block_only_mode/ad_block_only_mode_policies_utils.cc
aseren: BravePolicyManager which loads AOM policies is compiled for iOS, so we need to place a guard somewhere.
Another option will be to use separate files for iOS and for other platforms, like:
* `ad_block_only_mode_policies_utils.cc`
* `ad_block_only_mode_policies_utils_ios.cc`

Which option would it be better by your opinion?

components/brave_policy/brave_profile_policy_provider.cc
aseren: I followed the existing pattern from the BraveProfilePolicyProvider so yes, I think that potentially can happen. 
Maybe it would be better to fix these issues separately as a follow-up (I mean put CHECKs and add more comments).

components/brave_origin/ad_block_only_mode_policy_manager.cc
aseren: This makes sense, I will think of a way of how to do it

components/brave_policy/ad_block_only_mode/ad_block_only_mode_policies_utils.cc
bridiver: neither because this file shouldn't exist, it should be in AdBlockOnlyPolicyManager::GetAllProfilePolicies like brave origin and the entire AdBlockOnlyPolicyManager should be guarded be an ad_block_only_mode feature flag, not ios.

components/brave_policy/ad_block_only_mode/ad_block_only_mode_policies_utils.cc
bridiver: https://github.com/brave/brave-core/pull/31485/files#r2434140979

bridiver: and this belongs in components/brave_shields just like the brave origin stuff is in components/brave_origin

components/brave_policy/brave_profile_policy_provider.cc
bridiver: also `OnBraveOriginPoliciesReady` should now just be `OnProfilePoliciesReady` or similar

components/brave_origin/ad_block_only_mode_policy_manager.cc
```diff
@@ -0,0 +1,86 @@
...
+    // set.
+    observer->OnAdBlockOnlyModePoliciesChanged();
+  }
+}
+
+void AdBlockOnlyModePolicyManager::RemoveObserver(Observer* observer) {
+  observers_.RemoveObserver(observer);
+}
+
+bool AdBlockOnlyModePolicyManager::IsAdBlockOnlyModeEnabled() const {
```
bridiver: as mentioned recently in slack we should avoid bringing in unnecessary dependencies for feature/pref checks. `IsAdBlockOnlyModeEnabled` should be in the same file as kAdBlockOnlyModeEnabled and pass in local state

bridiver: for convenience you could wrap that helper inside `AdBlockOnlyModePolicyManager::IsAdBlockOnlyModeEnabled`, but this should not be the source of truth for the value because you shouldn't need a dependency on `AdBlockOnlyModePolicyManager` to use it

components/brave_origin/ad_block_only_mode_policy_manager.cc
```diff
@@ -0,0 +1,86 @@
...
+}
+
+bool AdBlockOnlyModePolicyManager::IsInitialized() const {
+  return initialized_;
+}
+
+void AdBlockOnlyModePolicyManager::AddObserver(Observer* observer) {
+  observers_.AddObserver(observer);
+
+  if (local_state_ &&
```
bridiver: shouldn't this be `IsAdBlockOnlyModeEnabled`? we don't want to notify observers that policies have changed if the feature isn't even enabled?

components/brave_policy/ad_block_only_mode/ad_block_only_mode_policies_utils.cc
aseren: @bbondy commented that AOM policies should be set in `components/brave_policies`, so we don't create any new dependencies from `components/policy/*`.

Are we fine, if we add dependency on `components/policy/*` to some other target?

bridiver: what I'm suggesting is exactly what brave origin does. There shouldn't be any components/policy dependencies because those can be handled with interface/impl/forward declarations

bridiver: I guess it's a component target so maybe a bit more complicated and we should probably brave brave_policy a component as well to deal with it, but the only dependency it needs is on brave_policy, the rest is all internal

components/brave_origin/ad_block_only_mode_policy_manager.cc
aseren: yeah, if feature is disabled, then observers shouldn't be notified

components/brave_origin/ad_block_only_mode_policy_manager.cc
bridiver: I feel like that puts this in a very confusing state and it would be better to have a simple PR to change the name before doing this. It's a very small amount of work to reduce a lot of confusion in this PR

components/brave_policy/ad_block_only_mode/ad_block_only_mode_policies_utils.cc
aseren: Thanks for comments, so I will do following:
- add a new file to `brave_policy` target: `components/brave_policy/ad_block_only_mode_policies_utils.h/.cc` 
- target which contains `AdBlockOnlyPolicyManager` will depend on `brave_policy` 
- `AdBlockOnlyPolicyManager` file implementaion will include `components/brave_policy/ad_block_only_mode_policies_utils.h`

components/brave_policy/brave_profile_policy_provider.cc
aseren: Changed to `OnBravePoliciesReady`

components/brave_policy/ad_block_only_mode/ad_block_only_mode_policies_utils.cc
aseren: Moved policies set part to `AdBlockOnlyPolicyManager`

components/brave_policy/ad_block_only_mode/ad_block_only_mode_policies_utils.cc
aseren: Moved policies set part to AdBlockOnlyPolicyManager

components/brave_policy/BUILD.gn
aseren: Moved policies set part to AdBlockOnlyPolicyManager

components/brave_origin/ad_block_only_mode_policy_manager.cc
aseren: Moved `AdBlockOnlyPolicyManager` to components/brave_policy/ad_block_only_mode

components/brave_origin/ad_block_only_mode_policy_manager.cc
aseren: Changed code to contain only one `base::FeatureList::IsEnabled ` check

browser/brave_origin/brave_origin_service_factory.cc
aseren: Moved `AdBlockOnlyPolicyManager` to `components/brave_policy/ad_block_only_mode`

components/brave_origin/ad_block_only_mode_policy_manager.cc
aseren: `AdBlockOnlyModePolicyManager::IsAdBlockOnlyModeEnabled` is removed

components/brave_origin/ad_block_only_mode_policy_manager.cc
aseren: refactored the code to have only one is feature enabled check

components/brave_policy/ad_block_only_mode/buildflags/BUILD.gn
```diff
@@ -0,0 +1,13 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import(
+    "//brave/components/brave_policy/ad_block_only_mode/buildflags/buildflags.gni")
+import("//build/buildflag_header.gni")
+
+buildflag_header("buildflags") {
+  header = "buildflags.h"
+  flags = [ "ENABLE_AD_BLOCK_ONLY_MODE_POLICIES=$enable_ad_block_only_mode" ]
```
bridiver: please keep consistent naming

components/brave_policy/ad_block_only_mode/buildflags/buildflags.gni
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) 2025 The Brave Authors. All rights reserved.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+declare_args() {
+  enable_ad_block_only_mode = is_mac || is_linux || is_win || is_android
```
bridiver: !is_ios

components/brave_policy/ad_block_only_mode/buildflags/BUILD.gn
aseren: re-named to use `enable_ad_block_only_mode_policies`

components/brave_policy/ad_block_only_mode/buildflags/buildflags.gni
aseren: changed

components/brave_policy/BUILD.gn
```diff
@@ -41,6 +42,7 @@ source_set("unit_tests") {
     ":brave_policy",
     "//base",
     "//base/test:test_support",
+    "//brave/components/brave_policy/ad_block_only_mode:unit_tests",
```
bridiver: this so go in brave/components/BUILD.gn

components/brave_policy/ad_block_only_mode/buildflags/buildflags.gni
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) 2025 The Brave Authors. All rights reserved.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+declare_args() {
+  enable_ad_block_only_mode_policies = !is_ios
```
bridiver, line 9: please add a description

aseren: Added following description:
```
  # Controls whether Ad Block Only Mode can override some Brave browser policies
  # for improved website compatibility. Disabled on iOS where not supported.
```

components/brave_policy/BUILD.gn
aseren: moved, thanks!

aseren:
The PR is rebased to latest master and all comments are address now.

aseren:
Rebased to the latest master before merging.

PR #31497, opened 2025-09-29
Enabled injecting command line options from a config file on Linux

wknapik:
**EARLY DRAFT, PLEASE DON'T REVIEW, UNLESS EXPLICITLY REQUESTED**

Resolves https://github.com/brave/brave-browser/issues/49220

This will fix a number of issues on Arch and port a feature that was Arch-specific to all Linux distributions, with improvements.

The feature allows users to add command-line options to a file (~/.config/brave-flags.conf, ~/.config/brave-beta-flags.conf, ~/.config/brave-nightly-flags.conf), to be injected each time the browser is launched.

This was originally implemented by previous maintainers of Arch packages, before Brave took over maintenance. As such, users are already relying on this behavior.

On Arch, we would drop the custom wrapper and rely on the newly patched upstream one, which would, most importantly, fix PWAs, probably fix https://github.com/brave/brave-browser/issues/14603 (haven't tested yet) and bring in a few other improvements (see: https://github.com/brave/brave-browser/issues/49220).

On all other distributions, it would enable this useful feature that only Arch users had access to.

Beyond that, we'd be fixing some issues with the [previous](https://aur.archlinux.org/cgit/aur.git/tree/brave-bin.sh?h=brave-bin) [wrappers](https://aur.archlinux.org/cgit/aur.git/tree/brave-nightly-bin.sh?h=brave-nightly-bin) (max one option per line, no inline comment support, inadequate quoting, no whitespace stripping, use of tools from packages that aren't dependencies, issues with passing multiple options).

This also removes a workaround for https://github.com/brave/brave-browser/issues/4142, which, at this point, was leaving a shell process in the tree unnecessarily and unconditionally changing the exit code to 0.

To anyone who thinks this should be implemented in the browser - right on. Until a PR for that materializes, this is the solution.

### DIFF
--- a/patches/chrome-installer-linux-common-wrapper.patch
+++ b/patches/chrome-installer-linux-common-wrapper.patch
@@ -1,8 +1,8 @@
 diff --git a/chrome/installer/linux/common/wrapper b/chrome/installer/linux/common/wrapper
-index aaa46bf71f6a3..22e4412a127e1 100755
+index aaa46bf71f6a3..58fd403372220 100755
 --- a/chrome/installer/linux/common/wrapper
 +++ b/chrome/installer/linux/common/wrapper
-@@ -27,6 +27,40 @@ fi
+@@ -27,6 +27,41 @@ fi
  
  export CHROME_VERSION_EXTRA="@@CHANNEL@@"
  
@@ -12,6 +12,7 @@ index aaa46bf71f6a3..22e4412a127e1 100755
 +cfg_options=()
 +
 +shopt -s extglob
++
 +if [[ -f "$cfg_file" ]]; then
 +  # Parse options after comments have been removed.
 +  parse_cfg_options() {
@@ -31,7 +32,7 @@ index aaa46bf71f6a3..22e4412a127e1 100755
 +    parse_cfg_options "$rest"
 +  }
 +
-+  # Remove comments and parse the remaining options.
++  # Read the config file, remove comments and parse the remaining options.
 +  while read -r cfg_line; do
 +    read -ra cfg_line_items <<<"${cfg_line%%[[:space:]]#*}"
 +    if [[ "${cfg_line_items[0]:0:1}" != "#" ]]; then
@@ -43,7 +44,7 @@ index aaa46bf71f6a3..22e4412a127e1 100755
  # We don't want bug-buddy intercepting our crashes. http://crbug.com/24120
  export GNOME_DISABLE_CRASH_DIALOG=SET_BY_GOOGLE_CHROME
  
-@@ -37,4 +71,4 @@ exec > >(exec cat)
+@@ -37,4 +72,4 @@ exec > >(exec cat)
  exec 2> >(exec cat >&2)
  
  # Note: exec -a below is a bashism.


patches/chrome-installer-linux-common-wrapper.patch
```diff
@@ -1,11 +1,50 @@
...
 --- a/chrome/installer/linux/common/wrapper
 +++ b/chrome/installer/linux/common/wrapper
-@@ -36,5 +36,4 @@ exec < /dev/null
- exec > >(exec cat)
+@@ -27,6 +27,39 @@ fi
+ 
+ export CHROME_VERSION_EXTRA="@@CHANNEL@@"
+ 
++# Add command line flags from a config file.
++# One option per line, surrounding whitespace is stripped.
```
wknapik: It would be possible to handle any number of options per line, if we prohibit whitespace in the values passed, or require some indicators of the start and end of a value, or complicate the parsing beyond what should be done here.

Every simple approach has some corner cases w.r.t. copy-pastability from a script, or a shell (which is how I think most users would populate the config file).

We can choose different ones, go for a proper implementation in the browser, or merge this.

wknapik: Basically potential whitespace in values is causing all the trouble here. Otherwise this would all be a few lines.

patches/chrome-installer-linux-common-wrapper.patch
```diff
@@ -1,11 +1,50 @@
...
+ # We don't want bug-buddy intercepting our crashes. http://crbug.com/24120
+ export GNOME_DISABLE_CRASH_DIALOG=SET_BY_GOOGLE_CHROME
+ 
+@@ -37,4 +70,4 @@ exec > >(exec cat)
  exec 2> >(exec cat >&2)
  
--# Note: exec -a below is a bashism.
+ # Note: exec -a below is a bashism.
 -exec -a "$0" "$HERE/@@PROGNAME@@" "$@"
-+"$HERE/@@PROGNAME@@" "$@" || true
```
fmarier: That `|| true` was added a while back because a failure here would end up starting Firefox IIRC.

patches/chrome-installer-linux-common-wrapper.patch
```diff
@@ -1,11 +1,50 @@
...
- exec > >(exec cat)
+@@ -27,6 +27,39 @@ fi
+ 
+ export CHROME_VERSION_EXTRA="@@CHANNEL@@"
+ 
++# Add command line flags from a config file.
++# One option per line, surrounding whitespace is stripped.
++# Empty lines and comments beginning with "#" are ignored.
++config_channel_infix="$([[ "$CHROME_VERSION_EXTRA" == stable ]] && echo || echo "$CHROME_VERSION_EXTRA-")"
++config_file="${XDG_CONFIG_HOME:-"$HOME/.config"}/brave-${config_channel_infix}flags.conf"
```
fmarier: Why not inside the `~/.config/BraveSoftware/Brave-Browser-<channel>` directory?

wknapik: A very good question. However, we have no choice at this point. At least not on Arch, unless we want to migrate existing users to the new path, which I'd definitely rather avoid.

patches/chrome-installer-linux-common-wrapper.patch
wknapik: How would that start firefox?

patches/chrome-installer-linux-common-wrapper.patch
wknapik: The "one option per line" limitation is bugging me.

I'm also wondering about copy-pastability and the potential for things to go wrong.

We allow pasting quoted values, including the quotes, that works.

Backslash-escaped whitespace wouldn't work copy-pasted, the backslashes would be escaped and become literal.

It would be reasonable to say that there's no escaping of any kind needed in the config file.

Then we could just strip out backslashes and, as a result, support copy-pasting backslash-escaped values.

But then there are possible literal backslashes to handle.

I'm also weary of complicating this any further.

wknapik: Ok, I think I see a way to drastically simplify this and support any number of options per line. Will post later.

wknapik: Pushed. Any number of options can be added per line, comments (including inline) work and if spaces are present, quoting the value takes care of it (when there are no spaces, quotes aren't required).

patches/chrome-installer-linux-common-wrapper.patch
```diff
@@ -1,11 +1,33 @@
...
+ export CHROME_VERSION_EXTRA="@@CHANNEL@@"
+ 
++# Add command-line options from a config file. Empty lines and #-comments are ignored.
++cfg_channel="$([[ "$CHROME_VERSION_EXTRA" == stable ]] && echo || echo "$CHROME_VERSION_EXTRA-")"
++cfg_file="${XDG_CONFIG_HOME:-"$HOME/.config"}/brave-${cfg_channel}flags.conf"
++cfg_options=()
++if [[ -f "$cfg_file" ]]; then
++  while read -r line; do
++    read -ra line_items <<<"${line%%[[:space:]]#*}"
++    case "${line_items[*]}" in
```
wknapik: Need to loop over line items too, will push a fix

patches/chrome-installer-linux-common-wrapper.patch
fmarier: IIRC it was something like `xdg-open` (or maybe it was `sensible-browser` on Debian) falls back to the next browser when the first one fails.

patches/chrome-installer-linux-common-wrapper.patch
wknapik: Fixed

patches/chrome-installer-linux-common-wrapper.patch
```diff
@@ -1,11 +1,47 @@
...
++
++if [[ -f "$cfg_file" ]]; then
++  parse_cfg_options() {
++    local line="$*" key val val_and_rest rest
++
++    case "$line" in
++      "") ;;
++      [[:space:]]*)
++        parse_cfg_options "${line:1}";;
++      --*=\"*\"*|--*=\'*\'*)
```
wknapik: Gotta tighten up this match

patches/chrome-installer-linux-common-wrapper.patch
wknapik: I don't get it. We're not executing `xdg-open`, or `sensible-browser`. How would a brave script/binary end up starting `firefox`?

wknapik:
Looks like on Arch we'll have to modify the last line of the launcher from the PKGBUILD. At least initially.

Also, looks like on other distros, the packages are consistently installed in /opt/brave.com, whereas the Arch release channel package installs in /opt/brave-bin. We need to unify the Arch packages anyway, regardless of this, so we'd likely go in that direction.

patches/chrome-installer-linux-common-wrapper.patch
fmarier: See https://github.com/brave/brave-browser/issues/4142 / https://github.com/brave/brave-core/commit/daa14f1c8af8b33d3ee0ee2632fc72544561de8f

PR #31498, opened 2025-09-29
Set default engine in network audit tests

DJAndries:
Resolves https://github.com/brave/brave-browser/issues/49743


browser/net/brave_network_audit_search_ad_browsertest.cc
```diff
@@ -73,6 +76,16 @@ class BraveNetworkAuditSearchAdTest : public InProcessBrowserTest {
     mock_cert_verifier_.mock_cert_verifier()->set_default_result(net::OK);
     host_resolver()->AddRule("*", "127.0.0.1");
 
+    auto* template_url_service =
+        TemplateURLServiceFactory::GetForProfile(browser()->profile());
+    search_test_utils::WaitForTemplateURLServiceToLoad(template_url_service);
+
+    auto* google_template_url =
```
fmarier: Does the search ad test need Brave Search to be the default in order to work? It's clicking on a search ad and checking for extra network connections IIRC.

DJAndries: Woops! Just switched the defaults

fmarier (approved):

PR #31499, opened 2025-09-29
Sandboxed adblock filter parsing

antonok-edm:
Resolves https://github.com/brave/brave-browser/issues/49744

### DIFF
--- a/browser/brave_shields/ad_block_service_browsertest.cc
+++ b/browser/brave_shields/ad_block_service_browsertest.cc
@@ -162,17 +162,17 @@ void AdBlockServiceTest::PreRunTestOnMainThread() {
   // platforms
   ASSERT_TRUE(base::test::RunUntil([&]() {
     auto counts = histogram_tester_.GetTotalCountsForPrefix(
-        "Brave.Adblock.MakeEngineWithRules.Default");
-    return counts.find("Brave.Adblock.MakeEngineWithRules.Default") !=
+        "Brave.Adblock.EngineDeserialize.Default");
+    return counts.find("Brave.Adblock.EngineDeserialize.Default") !=
                counts.end() &&
-           counts.at("Brave.Adblock.MakeEngineWithRules.Default") >= 1;
+           counts.at("Brave.Adblock.EngineDeserialize.Default") >= 1;
   })) << "Timeout waiting for initial engine creation";
 
   histogram_tester_.ExpectTotalCount(
-      "Brave.Adblock.MakeEngineWithRules.Default", 1);
+      "Brave.Adblock.EngineDeserialize.Default", 1);
   InstallDefaultAdBlockComponent();
   histogram_tester_.ExpectTotalCount(
-      "Brave.Adblock.MakeEngineWithRules.Default", 2);
+      "Brave.Adblock.EngineDeserialize.Default", 2);
   EXPECT_EQ(profile()->GetPrefs()->GetUint64(kAdsBlocked), 0ULL);
 }
 
@@ -529,9 +529,9 @@ IN_PROC_BROWSER_TEST_F(AdBlockServiceTest, DefaultStartupWithCookieList) {
   // The empty ruleset building until the components are loaded.
   // TODO(matuchin): remove that excessive work.
   histogram_tester_.ExpectTotalCount(
-      "Brave.Adblock.MakeEngineWithRules.Default", 2);
+      "Brave.Adblock.EngineDeserialize.Default", 2);
   histogram_tester_.ExpectTotalCount(
-      "Brave.Adblock.MakeEngineWithRules.Additional", 1);
+      "Brave.Adblock.EngineDeserialize.Additional", 1);
 
   // Loads the default list first, then the additional cookie list.
   InstallRegionalAdBlockComponent(brave_shields::kCookieListUuid, true);
@@ -539,9 +539,9 @@ IN_PROC_BROWSER_TEST_F(AdBlockServiceTest, DefaultStartupWithCookieList) {
   // Only one new rebuild is expected. Loading the extra list must not
   // trigger another rebuilding of the default engine and vice versa.
   histogram_tester_.ExpectTotalCount(
-      "Brave.Adblock.MakeEngineWithRules.Default", 2);
+      "Brave.Adblock.EngineDeserialize.Default", 2);
   histogram_tester_.ExpectTotalCount(
-      "Brave.Adblock.MakeEngineWithRules.Additional", 2);
+      "Brave.Adblock.EngineDeserialize.Additional", 2);
 }
 
 // Load a page with an ad image, and make sure it is blocked by a filter from a
--- a/components/brave_shields/content/browser/ad_block_custom_filters_provider.cc
+++ b/components/brave_shields/content/browser/ad_block_custom_filters_provider.cc
@@ -14,7 +14,6 @@
 #include "base/task/single_thread_task_runner.h"
 #include "base/trace_event/trace_event.h"
 #include "brave/components/brave_shields/content/browser/ad_block_custom_filter_reset_util.h"
-#include "brave/components/brave_shields/core/browser/ad_block_filters_provider_manager.h"
 #include "brave/components/brave_shields/core/browser/brave_shields_utils.h"
 #include "brave/components/brave_shields/core/common/pref_names.h"
 #include "components/prefs/pref_service.h"
@@ -22,15 +21,6 @@
 namespace brave_shields {
 
 namespace {
-
-void AddDATBufferToFilterSet(uint8_t permission_mask,
-                             DATFileDataBuffer buffer,
-                             uint64_t flow_id,
-                             rust::Box<adblock::FilterSet>* filter_set) {
-  TRACE_EVENT("brave.adblock", "AddDATBufferToFilterSet_CustomFiltersProvider",
-              perfetto::TerminatingFlow::ProcessScoped(flow_id));
-  (*filter_set)->add_filter_list_with_permissions(buffer, permission_mask);
-}
 
 // Custom filters get all permissions granted, i.e. all bits of the mask set,
 // i.e. the maximum possible uint8_t.
@@ -117,11 +107,11 @@ bool AdBlockCustomFiltersProvider::UpdateCustomFiltersFromSettings(
   return UpdateCustomFilters(custom_filters);
 }
 
-void AdBlockCustomFiltersProvider::LoadFilterSet(
+void AdBlockCustomFiltersProvider::LoadFilters(
     base::OnceCallback<
-        void(base::OnceCallback<void(rust::Box<adblock::FilterSet>*)>)> cb) {
+        void(std::vector<unsigned char> filter_buffer, uint8_t permission_mask)> cb) {
   const uint64_t flow_id = base::RandUint64();
-  TRACE_EVENT("brave.adblock", "AdBlockCustomFiltersProvider::LoadFilterSet",
+  TRACE_EVENT("brave.adblock", "AdBlockCustomFiltersProvider::LoadFilters",
               perfetto::TerminatingFlow::ProcessScoped(flow_id));
   DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
   auto custom_filters = GetCustomFilters();
@@ -132,9 +122,7 @@ void AdBlockCustomFiltersProvider::LoadFilterSet(
   // PostTask so this has an async return to match other loaders
   base::SingleThreadTaskRunner::GetCurrentDefault()->PostTask(
       FROM_HERE, base::BindOnce(std::move(cb),
-                                base::BindOnce(&AddDATBufferToFilterSet,
-                                               kCustomFiltersPermissionLevel,
-                                               buffer, flow_id)));
+                                buffer, kCustomFiltersPermissionLevel));
 }
 
 }  // namespace brave_shields
--- a/components/brave_shields/content/browser/ad_block_localhost_filters_provider.cc
+++ b/components/brave_shields/content/browser/ad_block_localhost_filters_provider.cc
@@ -9,17 +9,10 @@
 #include <vector>
 
 #include "base/task/single_thread_task_runner.h"
-#include "brave/components/brave_shields/core/browser/ad_block_filters_provider_manager.h"
-#include "components/prefs/pref_service.h"
 
 namespace brave_shields {
 
 namespace {
-
-void AddDATBufferToFilterSet(DATFileDataBuffer buffer,
-                             rust::Box<adblock::FilterSet>* filter_set) {
-  (*filter_set)->add_filter_list(buffer);
-}
 
 constexpr char kLocalhostBadfilters[] = R"(
 ||0.0.0.0^$third-party,domain=~[::]|~[::ffff:0:0],badfilter
@@ -44,9 +37,9 @@ std::string AdBlockLocalhostFiltersProvider::GetNameForDebugging() {
   return "AdBlockLocalhostFiltersProvider";
 }
 
-void AdBlockLocalhostFiltersProvider::LoadFilterSet(
+void AdBlockLocalhostFiltersProvider::LoadFilters(
     base::OnceCallback<
-        void(base::OnceCallback<void(rust::Box<adblock::FilterSet>*)>)> cb) {
+        void(std::vector<unsigned char> filter_buffer, uint8_t permission_mask)> cb) {
   DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
 
   auto buffer = std::vector<unsigned char>(std::begin(kLocalhostBadfilters),
@@ -55,8 +48,7 @@ void AdBlockLocalhostFiltersProvider::LoadFilterSet(
   // PostTask so this has an async return to match other loaders
   base::SingleThreadTaskRunner::GetCurrentDefault()->PostTask(
       FROM_HERE,
-      base::BindOnce(std::move(cb),
-                     base::BindOnce(&AddDATBufferToFilterSet, buffer)));
+      base::BindOnce(std::move(cb), buffer, UINT8_MAX));
 }
 
 }  // namespace brave_shields
--- a/components/brave_shields/content/browser/ad_block_service.cc
+++ b/components/brave_shields/content/browser/ad_block_service.cc
@@ -65,37 +65,22 @@ void AdBlockService::SourceProviderObserver::OnChanged(bool is_default_engine) {
     return;
   }
   auto on_loaded_cb = base::BindOnce(
-      &AdBlockService::SourceProviderObserver::OnFilterSetCallbackLoaded,
+      &AdBlockService::SourceProviderObserver::OnDATCreated,
       weak_factory_.GetWeakPtr());
   if (is_filter_provider_manager_) {
     static_cast<AdBlockFiltersProviderManager*>(filters_provider_.get())
-        ->LoadFilterSetForEngine(is_default_engine, std::move(on_loaded_cb));
+        ->LoadFiltersForEngine(is_default_engine, std::move(on_loaded_cb));
   } else {
-    filters_provider_->LoadFilterSet(std::move(on_loaded_cb));
+    // Hopefully we can remove this case, although it is currently used by other legacy
+    // tests (i.e. other than AdBlockServiceTest)
+    CHECK(false);
   }
 }
 
-void AdBlockService::SourceProviderObserver::OnFilterSetCallbackLoaded(
-    base::OnceCallback<void(rust::Box<adblock::FilterSet>*)> cb) {
-  task_runner_->PostTaskAndReplyWithResult(
-      FROM_HERE,
-      base::BindOnce(
-          [](base::OnceCallback<void(rust::Box<adblock::FilterSet>*)> cb) {
-            auto filter_set = std::make_unique<rust::Box<adblock::FilterSet>>(
-                adblock::new_filter_set());
-            std::move(cb).Run(filter_set.get());
-            return filter_set;
-          },
-          std::move(cb)),
-      base::BindOnce(
-          &AdBlockService::SourceProviderObserver::OnFilterSetCreated,
-          weak_factory_.GetWeakPtr()));
-}
-
-void AdBlockService::SourceProviderObserver::OnFilterSetCreated(
-    std::unique_ptr<rust::Box<adblock::FilterSet>> filter_set) {
-  TRACE_EVENT("brave.adblock", "OnFilterSetCreated");
-  filter_set_ = std::move(filter_set);
+void AdBlockService::SourceProviderObserver::OnDATCreated(
+    const std::vector<unsigned char>& verified_engine_dat) {
+  TRACE_EVENT("brave.adblock", "OnDATCreated");
+  pending_dat_ = std::move(verified_engine_dat);
   // multiple AddObserver calls are ignored
   resource_provider_->AddObserver(this);
   resource_provider_->LoadResources(base::BindOnce(
@@ -104,7 +89,7 @@ void AdBlockService::SourceProviderObserver::OnFilterSetCreated(
 
 void AdBlockService::SourceProviderObserver::OnResourcesLoaded(
     const std::string& resources_json) {
-  if (!filter_set_) {
+  if (pending_dat_.empty()) {
     task_runner_->PostTask(
         FROM_HERE,
         base::BindOnce(&AdBlockEngine::UseResources,
@@ -112,13 +97,13 @@ void AdBlockService::SourceProviderObserver::OnResourcesLoaded(
   } else {
     auto engine_load_callback = base::BindOnce(
         [](base::WeakPtr<AdBlockEngine> engine,
-           std::unique_ptr<rust::Box<adblock::FilterSet>> filter_set,
+           std::vector<unsigned char> dat_buffer,
            const std::string& resources_json) {
           if (engine) {
-            engine->Load(std::move(*filter_set.get()), resources_json);
+            engine->Load(true, dat_buffer, resources_json);
           }
         },
-        adblock_engine_->AsWeakPtr(), std::move(filter_set_), resources_json);
+        adblock_engine_->AsWeakPtr(), std::move(pending_dat_), resources_json);
     task_runner_->PostTask(FROM_HERE, std::move(engine_load_callback));
   }
 }
--- a/components/brave_shields/content/browser/ad_block_service.h
+++ b/components/brave_shields/content/browser/ad_block_service.h
@@ -73,9 +73,7 @@ class AdBlockService {
     ~SourceProviderObserver() override;
 
    private:
-    void OnFilterSetCallbackLoaded(
-        base::OnceCallback<void(rust::Box<adblock::FilterSet>*)> cb);
-    void OnFilterSetCreated(std::unique_ptr<rust::Box<adblock::FilterSet>>);
+    void OnDATCreated(const std::vector<unsigned char>& verified_engine_dat);
 
     // AdBlockFiltersProvider::Observer
     void OnChanged(bool is_default_engine) override;
@@ -83,7 +81,7 @@ class AdBlockService {
     // AdBlockResourceProvider::Observer
     void OnResourcesLoaded(const std::string& resources_json) override;
 
-    std::unique_ptr<rust::Box<adblock::FilterSet>> filter_set_;
+    std::vector<unsigned char> pending_dat_;
     raw_ptr<AdBlockEngine> adblock_engine_ = nullptr;               // not owned
     raw_ptr<AdBlockFiltersProvider> filters_provider_ = nullptr;    // not owned
     raw_ptr<AdBlockResourceProvider> resource_provider_ = nullptr;  // not owned
--- a/components/brave_shields/core/browser/ad_block_filters_provider_manager.cc
+++ b/components/brave_shields/core/browser/ad_block_filters_provider_manager.cc
@@ -4,8 +4,11 @@
 // You can obtain one at https://mozilla.org/MPL/2.0/.
 
 #include "brave/components/brave_shields/core/browser/ad_block_filters_provider_manager.h"
+#include "brave/components/brave_shields/core/services/filter_set/mojom/filter_set.mojom-forward.h"
+#include "brave/components/brave_shields/core/services/filter_set/mojom/filter_set.mojom.h"
+#include "content/public/browser/service_process_host.h"
+#include "mojo/public/cpp/bindings/remote.h"
 
-#include <memory>
 #include <string>
 #include <utility>
 
@@ -18,15 +21,32 @@
 #include "base/task/sequenced_task_runner.h"
 #include "base/trace_event/trace_event.h"
 
+namespace {
+
+// Translates the 2-argument `LoadFilters` callback into one that takes a single tuple argument.
+// Intended for compatibility with `BarrierCallback`.
+void LoadFiltersTuple(base::RepeatingCallback<void(std::tuple<std::vector<unsigned char>, uint8_t>)> cb, std::vector<unsigned char> filter_buffer, uint8_t permission_mask) {
+  cb.Run(std::tuple(filter_buffer, permission_mask));
+}
+
+}  // namespace
+
 namespace brave_shields {
 
 // static
 AdBlockFiltersProviderManager* AdBlockFiltersProviderManager::GetInstance() {
-  static base::NoDestructor<AdBlockFiltersProviderManager> instance;
+  static mojo::Remote<filter_set::mojom::UtilParseFilterSet> filter_set_service =
+    content::ServiceProcessHost::Launch<filter_set::mojom::UtilParseFilterSet>(
+        content::ServiceProcessHost::Options()
+            .WithDisplayName("Adblock Filter Parsing")
+            .Pass());
+
+  static base::NoDestructor<AdBlockFiltersProviderManager> instance(std::move(filter_set_service));
   return instance.get();
 }
 
-AdBlockFiltersProviderManager::AdBlockFiltersProviderManager() = default;
+AdBlockFiltersProviderManager::AdBlockFiltersProviderManager(mojo::Remote<filter_set::mojom::UtilParseFilterSet> filter_set_service) : filter_set_service_(std::move(filter_set_service)) {
+}
 
 AdBlockFiltersProviderManager::~AdBlockFiltersProviderManager() = default;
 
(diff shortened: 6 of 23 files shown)


components/services/brave_shields/BUILD.gn
```diff
@@ -0,0 +1,17 @@
+# Copyright (c) 2026 The Brave Authors. All rights reserved.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+source_set("filter_set") {
```
bridiver: I think this should normally be components/services/brave_shields?

components/brave_shields/core/browser/adblock/BUILD.gn
```diff
@@ -10,6 +10,7 @@ group("adblock") {
     "//brave/components/brave_shields/content/test:unit_tests",
     "//brave/components/brave_shields/core/browser:*",
     "//brave/components/brave_shields/core/common:*",
+    "//brave/components/brave_shields/core/services/filter_set",
```
bridiver: this doesn't look right, only the utility process should depend on the non-mojom dep

bridiver: sorry, the other way around, the utility process code should not depend on browser process code. Any shared code needs to be in common

browser/brave_shields/ad_block_service_browsertest.cc
```diff
@@ -160,17 +160,17 @@ void AdBlockServiceTest::PreRunTestOnMainThread() {
   // platforms
   ASSERT_TRUE(base::test::RunUntil([&]() {
     auto counts = histogram_tester_.GetTotalCountsForPrefix(
-        "Brave.Adblock.MakeEngineWithRules.Default");
-    return counts.find("Brave.Adblock.MakeEngineWithRules.Default") !=
+        "Brave.Adblock.EngineDeserialize.Default");
```
bridiver, line 163: this change seems like it's not necessary for this PR? I also see some refactoring involving moving/renaming files that should also be separated from this.

ios/browser/api/brave_shields/adblock_service.mm
```diff
@@ -128,7 +129,9 @@ - (instancetype)initWithComponentUpdater:
     _catalogProvider =
         std::make_unique<brave_shields::AdBlockFilterListCatalogProvider>(_cus);
     _filtersProviderManager =
-        std::make_unique<brave_shields::AdBlockFiltersProviderManager>();
+        std::make_unique<brave_shields::AdBlockFiltersProviderManager>(
+            base::BindRepeating(
+                &brave_shields::LaunchInProcessFilterSetParser));
```
antonok-edm: I'm still using the `InProcess` parser on iOS:
- The sandboxing implementation is based off of upstream Chromium's `Unzipper`
- I noticed that upstream also used an in-process implementation on iOS, but I don't fully understand the reasoning at the moment

I figured we can use the simpler version for now; later @StephenHeaps can test out using a real sandboxed implementation like in `browser/brave_browser_process_impl.cc`.

StephenHeaps: Only one process available for iOS (+ WebKit runs in separate process)

chromium_src/chrome/test/sources.gni
```diff
@@ -17,6 +17,7 @@ brave_chromium_src_chrome_test_test_support_deps = [
   "//brave/components/brave_shields/content/browser",
   "//brave/components/brave_shields/content/browser:test_support",
   "//brave/components/brave_vpn/common/buildflags",
+  "//brave/components/services/brave_shields:filter_set",
```
bridiver: why does this go in upstream test deps?

components/brave_shields/content/browser/ad_block_custom_filters_provider.cc
```diff
@@ -117,11 +108,13 @@ bool AdBlockCustomFiltersProvider::UpdateCustomFiltersFromSettings(
   return UpdateCustomFilters(custom_filters);
 }
 
-void AdBlockCustomFiltersProvider::LoadFilterSet(
+void AdBlockCustomFiltersProvider::LoadFilters(
```
bridiver, line 111: maybe `LoadFiltersFile`?

bridiver: or why is this now `LoadFilters` instead of `LoadFilterSet`?

components/brave_shields/content/browser/ad_block_localhost_filters_provider.cc
```diff
@@ -45,15 +39,23 @@ std::string AdBlockLocalhostFiltersProvider::GetNameForDebugging() {
...
+void AdBlockLocalhostFiltersProvider::LoadFilters(
     base::OnceCallback<
-        void(base::OnceCallback<void(rust::Box<adblock::FilterSet>*)>)> cb) {
+        void(std::vector<unsigned char> filter_buffer,
+             uint8_t permission_mask,
+             base::OnceCallback<void(adblock::FilterListMetadata)> on_metadata)>
+        cb) {
   DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
 
+  std::vector<unsigned char> buffer;
```
bridiver: seems like we should probably be using `base::HeapArray` here? There's a lot of overhead and copying going on passing around std::vector all over the place, especially without `std::move`.

components/brave_shields/content/browser/ad_block_localhost_filters_provider.cc
```diff
@@ -45,15 +39,23 @@ std::string AdBlockLocalhostFiltersProvider::GetNameForDebugging() {
...
+  buffer.reserve(sizeof(kLocalhostBadfilters) - 1);
+  std::copy_n(kLocalhostBadfilters, sizeof(kLocalhostBadfilters) - 1,
+              std::back_inserter(buffer));
+
   // PostTask so this has an async return to match other loaders
   base::SingleThreadTaskRunner::GetCurrentDefault()->PostTask(
       FROM_HERE,
-      base::BindOnce(std::move(cb), base::BindOnce(&AddDATBufferToFilterSet)));
+      base::BindOnce(std::move(cb), buffer, UINT8_MAX, base::DoNothing()));
```
bridiver: seems very strange that we're passing `UINT8_MAX` here to a callback? What does this actually call?

components/brave_shields/content/browser/ad_block_service.cc
```diff
@@ -64,34 +64,17 @@ void AdBlockService::SourceProviderObserver::OnChanged(bool is_default_engine) {
...
-}
-
-void AdBlockService::SourceProviderObserver::OnFilterSetCreated(
-    std::unique_ptr<rust::Box<adblock::FilterSet>> filter_set) {
-  TRACE_EVENT("brave.adblock", "OnFilterSetCreated");
-  filter_set_ = std::move(filter_set);
+void AdBlockService::SourceProviderObserver::OnDATCreated(
+    const std::vector<unsigned char>& verified_engine_dat) {
+  TRACE_EVENT("brave.adblock", "OnDATCreated");
+  pending_dat_ = std::move(verified_engine_dat);
```
bbondy, line 77: Review via brave-core-bot: `std::move(verified_engine_dat)` has no effect here because `verified_engine_dat` is a `const std::vector<unsigned char>&`. Moving from a const reference silently falls back to a copy since the move constructor requires a non-const rvalue reference. Either change the parameter to take by value (`std::vector<unsigned char> verified_engine_dat`) so the move works, or just assign directly without `std::move` to make the copy explicit.

components/brave_shields/core/browser/ad_block_filters_provider_manager.cc
```diff
@@ -16,10 +15,33 @@
...
+// tuple argument. Intended for compatibility with `BarrierCallback`.
+void LoadFiltersTuple(
+    base::RepeatingCallback<void(
+        std::tuple<std::vector<unsigned char>,
+                   uint8_t,
+                   base::OnceCallback<void(adblock::FilterListMetadata)>>)> cb,
+    std::vector<unsigned char> filter_buffer,
+    uint8_t permission_mask,
+    base::OnceCallback<void(adblock::FilterListMetadata)> on_metadata) {
+  cb.Run(std::tuple(filter_buffer, permission_mask, std::move(on_metadata)));
```
bbondy: Review via brave-core-bot: `filter_buffer` is taken by value but then copied into the tuple via `std::tuple(filter_buffer, ...)`. Use `std::move(filter_buffer)` to avoid an unnecessary copy of the entire filter buffer: `cb.Run(std::tuple(std::move(filter_buffer), permission_mask, std::move(on_metadata)));`

components/brave_shields/core/browser/ad_block_filters_provider_manager.cc
```diff
@@ -62,57 +84,104 @@ void AdBlockFiltersProviderManager::OnChanged(bool is_for_default_engine) {
...
-  std::move(cb).Run(base::BindOnce(&RunAllResults, std::move(results)));
+
+  std::vector<filter_set::mojom::FilterListInputPtr> inputs;
+  std::transform(
+      results.cbegin(), results.cend(), std::back_inserter(inputs),
+      [](const std::tuple<
+          std::vector<unsigned char>, uint8_t,
+          base::OnceCallback<void(adblock::FilterListMetadata)>>& tuple) {
+        auto r = filter_set::mojom::FilterListInput::New();
+        r->filters = std::get<0>(tuple);
```
bbondy: Review via brave-core-bot: `r->filters = std::get<0>(tuple)` copies the vector since `tuple` is a const reference. Consider restructuring the two `std::transform` calls into a single loop over `results` (using non-const iterators) so you can `std::move` both the filter buffer and the OnceCallback out of each tuple element in one pass, avoiding the copy.

components/services/brave_shields/filter_set_service.cc
```diff
@@ -0,0 +1,56 @@
...
+      }
+      if (this_metadata.value.expires_hours.has_value) {
+        mojom_metadata->expires_hours = this_metadata.value.expires_hours.value;
+      }
+    }
+    metadata.push_back(std::move(mojom_metadata));
+  }
+
+  const auto e = adblock::engine_from_filter_set(std::move(*filter_set));
+  CHECK(e.result_kind == adblock::ResultKind::Success);
```
bbondy: Review via brave-core-bot: `CHECK(e.result_kind == adblock::ResultKind::Success)` will crash the browser process (or utility process) if engine creation fails. Since the input comes from user-supplied filter lists (including custom subscriptions), this should handle the error gracefully instead of crashing. Consider returning an empty DAT buffer or an error via the Mojo callback.

components/services/brave_shields/filter_set_service.cc
```diff
@@ -0,0 +1,56 @@
...
+  const auto dat = e.value->serialize();
+  std::vector<unsigned char> output_dat;
+  std::copy(dat.cbegin(), dat.cend(), std::back_inserter(output_dat));
+
+  std::move(callback).Run(output_dat, std::move(metadata));
```
bbondy: Review via brave-core-bot: `output_dat` is passed to `callback.Run(output_dat, ...)` by lvalue, which copies the entire serialized engine buffer. Use `std::move(output_dat)` to avoid this copy.

components/brave_shields/content/browser/ad_block_localhost_filters_provider.cc
bbondy: Review via brave-core-bot: `buffer` is passed to `base::BindOnce(std::move(cb), buffer, ...)` without `std::move`, which copies the entire filter buffer. Use `std::move(buffer)` instead.

components/brave_shields/content/browser/ad_block_custom_filters_provider.cc
```diff
@@ -131,10 +124,9 @@ void AdBlockCustomFiltersProvider::LoadFilterSet(
 
   // PostTask so this has an async return to match other loaders
   base::SingleThreadTaskRunner::GetCurrentDefault()->PostTask(
-      FROM_HERE, base::BindOnce(std::move(cb),
-                                base::BindOnce(&AddDATBufferToFilterSet,
-                                               kCustomFiltersPermissionLevel,
-                                               buffer, flow_id)));
+      FROM_HERE,
+      base::BindOnce(std::move(cb), buffer, kCustomFiltersPermissionLevel,
```
bbondy: Review via brave-core-bot: `buffer` is passed to `base::BindOnce(std::move(cb), buffer, ...)` without `std::move`, which copies the entire filter buffer. Use `std::move(buffer)` instead.

components/brave_shields/content/browser/ad_block_localhost_filters_provider.cc
bbondy: Review via brave-core-bot: nit: The manual `reserve` + `std::copy_n` + `std::back_inserter` pattern can be simplified with `base::ToVector`. For example: `auto buffer = base::ToVector(base::span(kLocalhostBadfilters).first(sizeof(kLocalhostBadfilters) - 1));`

components/services/brave_shields/filter_set_service_in_process_launcher.cc
```diff
@@ -0,0 +1,36 @@
...
+namespace {
+void BindInProcessFilterSetParser(
+    mojo::PendingReceiver<filter_set::mojom::UtilParseFilterSet> receiver) {
+  mojo::MakeSelfOwnedReceiver(
+      std::make_unique<brave_shields::FilterSetService>(), std::move(receiver));
+}
+}  // namespace
+
+mojo::PendingRemote<filter_set::mojom::UtilParseFilterSet>
+LaunchInProcessFilterSetParser() {
```
bridiver: I would just make these launch methods (in and out of process) static methods on FilterSetService.

antonok-edm:
Split and rebased on top of https://github.com/brave/brave-core/pull/34099 - we will merge that first to simplify review of this PR.

components/brave_shields/content/browser/ad_block_service.cc
antonok-edm: updated to take by value 👍

components/brave_shields/core/browser/ad_block_filters_provider_manager.cc
antonok-edm: Propose a revision for this code

edit: nevermind, I did it myself :facepalm:

components/services/brave_shields/filter_set_service.cc
antonok-edm: The crashing is basically the motivation behind sandboxing it in the first place... if crashing here isn't safe, then the PR is completely useless :sweat_smile:

components/services/brave_shields/mojom/filter_set.mojom
```diff
@@ -0,0 +1,27 @@
...
+struct FilterListMetadata {
+  string? homepage;
+  string? title;
+  uint16? expires_hours;
+};
+
+// Utility process interface exposed to the browser process. Allows for
+// sandboxing when parsing adblock filters.
+[ServiceSandbox=sandbox.mojom.Sandbox.kService]
+interface UtilParseFilterSet {
```
bridiver: we don't have to include `Util` in the name here, but since this probably makes more sense as `FilterSetParser` and that's consistent with the rest of the code like `BindInProcessFilterSetParser`

components/services/brave_shields/filter_set_service.cc
bridiver: I'm not sure I understand, the motivation is to make sure we don't crash the browser process, but why would a failure to parse intentionally crash the utility process? Shouldn't this report some error back to the browser process?

components/brave_shields/content/browser/ad_block_service.h
```diff
@@ -71,17 +71,15 @@ class AdBlockService {
     ~SourceProviderObserver() override;
 
    private:
-    void OnFilterSetCallbackLoaded(
-        base::OnceCallback<void(rust::Box<adblock::FilterSet>*)> cb);
-    void OnFilterSetCreated(std::unique_ptr<rust::Box<adblock::FilterSet>>);
+    void OnDATCreated(const std::vector<unsigned char> verified_engine_dat);
```
netzenbot: `OnDATCreated` takes `const std::vector<unsigned char> verified_engine_dat` by const value. The `const` qualifier prevents `std::move` from actually moving the data -- the `pending_dat_ = std::move(verified_engine_dat)` in the .cc file will silently copy instead of move. Remove the `const` so the vector can be moved into `pending_dat_`. [best practice](https://github.com/brave-experiments/brave-core-bot/tree/master/docs/best-practices/coding-standards.md#-prefer-stdmove-over-clone)

components/brave_shields/core/browser/ad_block_filters_provider_manager.cc
```diff
@@ -62,57 +85,98 @@ void AdBlockFiltersProviderManager::OnChanged(bool is_for_default_engine) {
...
+}
+
+void AdBlockFiltersProviderManager::OnParseFilters(
+    base::OnceCallback<
+        void(const std::vector<unsigned char> verified_engine_dat)> cb,
+    std::vector<base::OnceCallback<void(adblock::FilterListMetadata)>>
+        on_metadata_cbs,
+    const std::vector<unsigned char>& verified_engine_dat,
+    const std::vector<filter_set::mojom::FilterListMetadataPtr> metadata) {
+  CHECK_EQ(on_metadata_cbs.size(), metadata.size());
```
netzenbot: `CHECK_EQ(on_metadata_cbs.size(), metadata.size())` will crash the browser process if the utility process sends back a mismatched number of metadata entries. Since this data crosses an IPC boundary (sandboxed filter parser), it is not fully within this code's control. Consider validating gracefully (e.g., early return with `LOG(ERROR)`) instead of CHECK. [best practice](https://github.com/brave-experiments/brave-core-bot/tree/master/docs/best-practices/coding-standards.md#-use-check-only-for-invariants-within-codes-control)

components/brave_shields/core/browser/ad_block_filters_provider_manager.cc
```diff
@@ -62,57 +85,98 @@ void AdBlockFiltersProviderManager::OnChanged(bool is_for_default_engine) {
...
+  for (unsigned int i = 0; i < metadata.size(); i++) {
```
netzenbot: nit: `metadata.size()` returns `size_t` (64-bit on most platforms), but `unsigned int` is 32-bit. Use `size_t` for the loop variable to avoid potential truncation.

components/services/brave_shields/filter_set_service.cc
netzenbot: `CHECK(e.result_kind == adblock::ResultKind::Success)` crashes the utility process if filter parsing fails. Since input may include user-supplied custom filters or third-party subscription lists, parsing failure is a plausible non-security error. Consider handling the error gracefully (e.g., returning an empty DAT buffer with error metadata) so remaining filter lists can still be processed.

components/services/brave_shields/filter_set_service.cc
```diff
@@ -0,0 +1,56 @@
...
+      }
+    }
+    metadata.push_back(std::move(mojom_metadata));
+  }
+
+  const auto e = adblock::engine_from_filter_set(std::move(*filter_set));
+  CHECK(e.result_kind == adblock::ResultKind::Success);
+  const auto dat = e.value->serialize();
+  std::vector<unsigned char> output_dat;
+  std::copy(dat.cbegin(), dat.cend(), std::back_inserter(output_dat));
```
netzenbot: nit: `std::copy(dat.cbegin(), dat.cend(), std::back_inserter(output_dat))` can be simplified to `auto output_dat = base::ToVector(dat);` which handles reserve and iteration automatically. [best practice](https://github.com/brave-experiments/brave-core-bot/tree/master/docs/best-practices/coding-standards.md#-use-basetovector-for-range-to-vector-conversions)

components/services/brave_shields/filter_set_service_in_process_launcher.cc
```diff
@@ -0,0 +1,36 @@
...
+  mojo::PendingRemote<filter_set::mojom::UtilParseFilterSet> remote;
+  base::ThreadPool::CreateSequencedTaskRunner(
```
netzenbot: The thread pool task runner is created without explicit `base::TaskPriority` or `base::TaskShutdownBehavior`. Since this blocks filter engine creation, consider specifying `base::TaskPriority::USER_BLOCKING` and `base::TaskShutdownBehavior::SKIP_ON_SHUTDOWN`. [best practice](https://github.com/brave-experiments/brave-core-bot/tree/master/docs/best-practices/coding-standards.md#-explicitly-specify-basetaskpriority-in-thread-pool-tasks)

components/services/brave_shields/filter_set_service.h
```diff
@@ -0,0 +1,38 @@
...
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#ifndef BRAVE_COMPONENTS_SERVICES_BRAVE_SHIELDS_FILTER_SET_SERVICE_H_
+#define BRAVE_COMPONENTS_SERVICES_BRAVE_SHIELDS_FILTER_SET_SERVICE_H_
+
+#include "brave/components/services/brave_shields/mojom/filter_set.mojom.h"
+#include "mojo/public/cpp/bindings/receiver.h"
+#include "mojo/public/cpp/bindings/remote.h"
```
netzenbot: nit: This header includes `mojo/public/cpp/bindings/remote.h` but `FilterSetService` only has a `mojo::Receiver` member, not a `mojo::Remote`. The include appears unused. [best practice](https://github.com/brave-experiments/brave-core-bot/tree/master/docs/best-practices/coding-standards.md#-always-include-what-you-use-iwyu)

chromium_src/chrome/test/sources.gni
antonok-edm: `gn_check` asked for it:

```
ERROR at //brave/test/base/testing_brave_browser_process.cc:19:11: Can't include this header from here.
#include "brave/components/services/brave_shields/filter_set_service_in_process_launcher.h"
          ^-------------------------------------------------------------------------------
The target:
  //chrome/test:test_support
is including a file from the target:
  //brave/components/services/brave_shields:filter_set

It's usually best to depend directly on the destination target.
In some cases, the destination target is considered a subcomponent
of an intermediate target. In this case, the intermediate target
should depend publicly on the destination to forward the ability
to include headers.

Dependency chain (there may also be others):
  //chrome/test:test_support -->
  //chrome/utility:utility --[private]-->
  //brave/components/services/brave_shields:filter_set
```

components/brave_shields/content/browser/ad_block_localhost_filters_provider.cc
antonok-edm: that is the permission mask for scriptlet injections, we use all `1` bits since it is a trusted (hardcoded) filter list source

components/services/brave_shields/filter_set_service.cc
antonok-edm: What kind of error would you want it to report?

adblock-rust doesn't have a `Result` codepath here; the only reason this result might be unsuccessful is if some string decoding goes wrong at the cxx bindings level. I don't see a good reason to specialize for that error type at the moment. If something really goes wrong there it may as well crash the same way an unexpected adblock-rust panic would.

browser/brave_shields/ad_block_service_browsertest.cc
antonok-edm: The adblock engine is being initialized through a different constructor now, so the histogram metric is updated.

(That constructor and histogram metric already existed previously, which is why you don't see a corresponding definition changed in this PR)

components/brave_shields/content/browser/ad_block_custom_filters_provider.cc
antonok-edm: `LoadFilterSet` was previously referring to the adblock-rust [`FilterSet`](https://docs.rs/adblock/latest/adblock/struct.FilterSet.html) struct representing an aggregation of multiple filter lists.

Now we're just passing a string buffer to be interpreted as adblock filters.

components/services/brave_shields/filter_set_service.cc
antonok-edm: duplicate of https://github.com/brave/brave-core/pull/31499#discussion_r2833858617

components/services/brave_shields/filter_set_service_in_process_launcher.cc
antonok-edm: `USER_BLOCKING` is default, but added `SKIP_ON_SHUTDOWN`

chromium_src/chrome/test/sources.gni
bridiver: but what is causing `test_support` to depend on `filter_set`? I'm looking for the root of the problem here

components/services/brave_shields/filter_set_service.cc
bridiver: but this code will run in-process in ios? It would defeat the entire purpose to intentionally crash?

chromium_src/chrome/test/sources.gni
antonok-edm: it's from `brave/test/base/testing_brave_browser_process.cc`, needed for initializing a `FilterSetService` for `TestingBraveBrowserProcess::ad_block_service()`. IIRC that file is included in upstream tests via `chromium_src`.

components/services/brave_shields/filter_set_service.cc
antonok-edm: iOS will need followup changes either way, but I can change this to just be a `VLOG` like it was prior to being sandboxed.

components/brave_shields/content/browser/ad_block_localhost_filters_provider.cc
antonok-edm: the `std::move` parts should be fixed now; let's potentially investigate if `base::HeapArray` is better in a followup since `vector` is already fairly entrenched in other callback APIs here

components/services/brave_shields/mojom/filter_set.mojom
antonok-edm: did some renaming of all the `brave/components/services/brave_shields/*` parts for consistency

components/brave_shields/content/browser/ad_block_custom_filters_provider.cc
bridiver: but now it's returning `FilterListMetadata` so shouldn't it be `LoadFilterListMetadata`? The terminology gets a bit confusing otherwise. We don't really care where it's loading from, we care what we're getting.

bridiver: or even `LoadFilterListMetadataFromBuffer`

components/brave_shields/core/browser/ad_block_filters_provider_manager.cc
```diff
@@ -62,57 +85,99 @@ void AdBlockFiltersProviderManager::OnChanged(bool is_for_default_engine) {
...
+                     weak_factory_.GetWeakPtr(), std::move(cb),
+                     std::move(on_metadata_cbs)));
+}
+
+void AdBlockFiltersProviderManager::OnParseFilters(
+    base::OnceCallback<
+        void(const std::vector<unsigned char> verified_engine_dat)> cb,
+    std::vector<base::OnceCallback<void(adblock::FilterListMetadata)>>
+        on_metadata_cbs,
+    const std::vector<unsigned char>& verified_engine_dat,
```
bridiver: `BigBuffer` would eliminate std::vector copy over mojo

bridiver: even without `BigBuffer` using `unsigned char` here is problematic as it can lead to UB. For type safety it should be `uint8_t` and for cases where we aren't passing ownership it should be `base::span<const uint8_t>` (does not apply here)

components/services/brave_shields/filter_parsing_service.cc
```diff
@@ -0,0 +1,99 @@
...
+      }
+    }
+    metadata.push_back(std::move(mojom_metadata));
+  }
+
+  auto e = adblock::engine_from_filter_set(std::move(*filter_set));
+  if (e.result_kind != adblock::ResultKind::Success) {
+    VLOG(0) << "AdBlockEngine::OnFilterSetLoaded failed: "
+            << e.error_message.c_str();
+    return;
```
netzenbot, line 92: When `engine_from_filter_set` fails, `callback` is dropped without being invoked. Mojo response callbacks must always be called — dropping it disconnects the IPC pipe, so `OnParseFilters` in `AdBlockFiltersProviderManager` never fires and the engine is left permanently stale with no recovery. Call `std::move(callback).Run(mojo_base::BigBuffer(), std::move(metadata));` on the error path.

components/services/brave_shields/mojom/adblock_filter_list_parser.mojom
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) 2026 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+module adblock_filter_list_parser.mojom;
```
bridiver: the module here is actually `brave_shields`, but it's fine to use just `adblock` here

bridiver: I'm confused though because I see both adblock::FilterListMetadata and adblock_filter_list_parser FilterListMetadata. Please avoid using the same class name with two different namespaces.

bridiver: we've typically been using the convention CxxFilterListMetadata for the rust side

components/brave_shields/content/browser/ad_block_engine.cc
```diff
@@ -338,12 +341,12 @@ void AdBlockEngine::OnListSourceLoaded(
...
 
 void AdBlockEngine::OnDATLoaded(
-    const DATFileDataBuffer& dat_buf,
+    DATFileDataBuffer dat_buf,
     const adblock::BraveCoreResourceStorage& storage) {
   DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
 
   // An empty buffer will not load successfully.
-  if (dat_buf.empty()) {
+  if (dat_buf.size() == 0) {
```
bridiver, line 349: in a follow-up we should make this optional

components/brave_shields/content/browser/ad_block_localhost_filters_provider.cc
```diff
@@ -5,17 +5,20 @@
...
 #include "base/task/single_thread_task_runner.h"
-#include "brave/components/brave_component_updater/browser/dat_file_util.h"
+#include "mojo/public/cpp/base/big_buffer.h"
 
 namespace brave_shields {
 
 namespace {
 
-constexpr uint8_t kLocalhostBadfilters[] = R"(
+constexpr char kLocalhostBadfilters[] = R"(
```
boocmp, line 21: nit: constexpr std::string_view ? if this change only needed to construct the string view.

browser/brave_browser_process_impl.cc
```diff
@@ -320,9 +320,20 @@ brave_shields::AdBlockService* BraveBrowserProcessImpl::ad_block_service() {
         base::ThreadPool::CreateSequencedTaskRunner(
             {base::MayBlock(), base::TaskPriority::USER_BLOCKING,
              base::TaskShutdownBehavior::SKIP_ON_SHUTDOWN}));
+    brave_shields::FilterParsingServiceFactory filter_list_parser_factory =
+        base::BindRepeating([]() {
+          mojo::PendingRemote<adblock::mojom::AdblockFilterListParser> remote;
+          content::ServiceProcessHost::Launch<
```
atuchin-m, line 326: What is the lifetime of this process? If we have done building and start enough one soon after, do we make another process?

components/brave_shields/core/browser/ad_block_filters_provider_manager.h
```diff
@@ -58,18 +62,28 @@ class AdBlockFiltersProviderManager : public AdBlockFiltersProvider,
...
+                     uint8_t,
+                     base::OnceCallback<void(adblock::CxxFilterListMetadata)>>>
           results);
 
   base::flat_set<AdBlockFiltersProvider*> default_engine_filters_providers_;
   base::flat_set<AdBlockFiltersProvider*> additional_engine_filters_providers_;
 
   base::CancelableTaskTracker task_tracker_;
 
+  mojo::Remote<adblock::mojom::AdblockFilterListParser> list_parser_service_;
```
atuchin-m, line 85: Looks like the associated process will live forever. Is it expected behavior?

browser/brave_browser_process_impl.cc
```diff
@@ -320,9 +320,20 @@ brave_shields::AdBlockService* BraveBrowserProcessImpl::ad_block_service() {
...
+              adblock::mojom::AdblockFilterListParser>(
+              remote.InitWithNewPipeAndPassReceiver(),
+              content::ServiceProcessHost::Options()
+                  .WithDisplayName("Adblock Filter Parsing")
+                  .Pass());
+          return remote;
```
boocmp, line 332: I dont see any `.reset` calls on this remote, is it by design? I'd expect to drop the utility process when the parsing is done.

browser/brave_shields/ad_block_service_browsertest.cc
```diff
@@ -537,20 +537,20 @@ IN_PROC_BROWSER_TEST_F(AdBlockServiceTest,
 IN_PROC_BROWSER_TEST_F(AdBlockServiceTest, DefaultStartupWithCookieList) {
   // The empty ruleset building until the components are loaded.
```
atuchin-m, line 538: Does this "empty ruleset building" about immediately result in starting the new process?

components/brave_shields/core/browser/ad_block_filters_provider_manager.cc
```diff
@@ -62,57 +86,94 @@ void AdBlockFiltersProviderManager::OnChanged(bool is_for_default_engine) {
...
+                     std::move(on_metadata_cbs)));
+}
+
+void AdBlockFiltersProviderManager::OnParseFilters(
+    base::OnceCallback<void(mojo_base::BigBuffer verified_engine_dat)> cb,
+    std::vector<base::OnceCallback<void(adblock::CxxFilterListMetadata)>>
+        on_metadata_cbs,
+    mojo_base::BigBuffer verified_engine_dat,
+    const std::vector<adblock::mojom::FilterListMetadataPtr> metadata) {
+  DCHECK_EQ(on_metadata_cbs.size(), metadata.size());
```
atuchin-m, line 159: This must be CHECK_EQ(), because it's security-related. We shouldn't trust a less privileged process.
Otherwise it's give ability to trigger a segfault in the main process via `on_metadata_cbs[i]`

browser/brave_browser_process_impl.cc
bridiver: I don't think I would worry about that. If we see any problems we can tune it, but for now I think we should just let the service code handle lifetime when it's needed

browser/brave_browser_process_impl.cc
bridiver: now this part of it is important, but we don't need an explicit reset, the remote just needs to go out of scope when we're done which I think it does, but should verify

browser/brave_browser_process_impl.cc
atuchin-m: We definitely don't want:
* another permanent process
* spawn multiple extra processes, especially on startup

components/brave_shields/core/browser/ad_block_filters_provider_manager.cc
```diff
@@ -62,57 +86,94 @@ void AdBlockFiltersProviderManager::OnChanged(bool is_for_default_engine) {
...
+      on_metadata_cbs;
+  on_metadata_cbs.reserve(results.size());
+  for (auto& [filters, permission_mask, on_metadata_cb] : results) {
+    inputs.push_back(adblock::mojom::FilterListInput::New(std::move(filters),
+                                                          permission_mask));
+    on_metadata_cbs.push_back(std::move(on_metadata_cb));
+  }
+  list_parser_service_->ParseFilters(
+      std::move(inputs),
+      base::BindOnce(&AdBlockFiltersProviderManager::OnParseFilters,
```
atuchin-m, line 148: What happens if the process actually crashes? Do we unblock the navigation throttle? How many attempt to parse we make after?

Have we tested this scenario?
Sounds like it make sense to add a browser tests for that case.

atuchin-m: I don't see any `set_disconnect_handler*` in the PR, so looks like we just continue to wait forever.

components/services/brave_shields/filter_parsing_service.cc
```diff
@@ -0,0 +1,101 @@
...
+
+void FilterParsingService::ParseFilters(
+    std::vector<adblock::mojom::FilterListInputPtr> filters,
+    ParseFiltersCallback callback) {
+  auto filter_set = std::make_unique<rust::Box<adblock::FilterSet>>(
+      adblock::new_filter_set());
+
+  std::vector<adblock::mojom::FilterListMetadataPtr> metadata;
+  for (auto& filter_list : filters) {
+    auto filter_vec = base::ToVector(filter_list->filters.byte_span());
```
atuchin-m, line 78: `.byte_span()` [doesn't work](https://source.chromium.org/chromium/chromium/src/+/main:mojo/public/cpp/base/big_buffer.h;l=132-139?q=big_buffer) for shared memory buffer.
As the result, Static build crashes on:
```
[61615:24972667:0309/153339.729154:FATAL:mojo/public/cpp/base/big_buffer.h:136] Check failed: storage_type_ == StorageType::kBytes (1 vs. 0). 
0   Brave Browser Development Framework 0x000000013213a3bc base::debug::CollectStackTrace(base::span<void const*, 18446744073709551615ul, void const**>) + 28
1   Brave Browser Development Framework 0x0000000132129548 base::debug::StackTrace::StackTrace(unsigned long) + 224
2   Brave Browser Development Framework 0x0000000132036354 logging::LogMessage::Flush() + 152
3   Brave Browser Development Framework 0x0000000132036234 logging::LogMessage::~LogMessage() + 36
4   Brave Browser Development Framework 0x000000013201dd1c logging::(anonymous namespace)::CheckLogMessage::~CheckLogMessage() + 76
5   Brave Browser Development Framework 0x000000013201d9e4 logging::CheckNoreturnError::~CheckNoreturnError() + 16
6   Brave Browser Development Framework 0x000000013201d9f4 logging::CheckNoreturnError::PCheck(char const*, base::Location const&) + 0
7   Brave Browser Development Framework 0x00000001307f748c auto base::as_byte_span<rust::cxxbridge1::Vec<unsigned char>>(T0 const&) + 0
8   Brave Browser Development Framework 0x00000001307f6e6c brave_shields::FilterParsingService::ParseFilters(std::__Cr::vector<mojo::StructPtr<adblock::mojom::FilterListInput>, std::__Cr::allocator<mojo::StructPtr<adblock::mojom::FilterListInput>>>, base::OnceCallback<void (mojo_base::BigBuffer, std::__Cr::vector<mojo::InlinedStructPtr<adblock::mojom::FilterListMetadata>, std::__Cr::allocator<mojo::InlinedStructPtr<adblock::mojom::FilterListMetadata>>>)>) + 216
```

components/brave_shields/core/browser/ad_block_filters_provider_manager.h
atuchin-m: Currently, the process never dies:
Image

browser/brave_shields/ad_block_service_browsertest.cc
atuchin-m: It does. In fact we always start it (even there are no lists).

browser/brave_browser_process_impl.cc
bridiver: https://github.com/brave/brave-core/pull/31499/files#r2895859713

PR #31502, opened 2025-09-30
Check browser service for Brave policies when hiding Managed UI

DJAndries:
Resolves https://github.com/brave/brave-browser/issues/49749

### DIFF
--- a/chromium_src/chrome/browser/ui/managed_ui.cc
+++ b/chromium_src/chrome/browser/ui/managed_ui.cc
@@ -5,6 +5,7 @@
 
 #include <algorithm>
 
+#include "chrome/browser/browser_process.h"
 #include "chrome/browser/profiles/profile.h"
 #include "components/policy/core/common/policy_map.h"
 #include "components/policy/core/common/policy_namespace.h"
@@ -17,12 +18,7 @@
 
 namespace brave_policy {
 
-bool ShouldHideManagedUI(const policy::PolicyMap& policies) {
-  // Empty policy map means no management at all
-  if (policies.empty()) {
-    return false;
-  }
-
+bool HasOnlyBravePolicies(const policy::PolicyMap& policies) {
   // Check all policies, not just BraveOrigin-related ones
   return std::ranges::all_of(policies, [](const auto& policy_pair) {
     const auto& [policy_name, entry] = policy_pair;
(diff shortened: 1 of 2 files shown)


bbondy:
Did you confirm that the managed UI is shown when only a browser level policy is set? 
I guess you could test this by changing the recently added chromium_src override to just call the Chromium one. And then change the policy types that are set by the brave policy provider to enterprise default.

chromium_src/chrome/browser/ui/managed_ui.cc
```diff
@@ -39,16 +35,28 @@ bool ShouldDisplayManagedUi(Profile* profile) {
...
+    const auto& profile_policies = profile_policy_service->GetPolicies(
+        policy::PolicyNamespace(policy::POLICY_DOMAIN_CHROME, std::string()));
+
+    if (!brave_policy::HasOnlyBravePolicies(profile_policies)) {
+      return true;
+    }
+  }
+
+  // Also check browser-level policies
+  auto* browser_policy_service = g_browser_process->policy_service();
```
bridiver, line 50: I'm confused, I thought the original PR was only about hiding `your installation is managed in chrome://settings and in the hamburger menu`? Why would we need to check this if it's not checked in the original method?

bbondy: The check that exists in the repo now is only checking the profile provider (The policies associated with a profile pref). I think this code also checks the browser provider. But I think we didn't check yet if we do need to check the browser provider or not yet.

bridiver: seems like we shouldn't have to since the upstream code doesn't? Not sure why it doesn't though

bridiver: in the issue I asked for a specific example of why we'd need this

chromium_src/chrome/browser/ui/managed_ui.cc
```diff
@@ -39,16 +35,28 @@ bool ShouldDisplayManagedUi(Profile* profile) {
   }
 
   // Check if we should hide due to Brave-only management
-  policy::PolicyService* policy_service =
+  auto* profile_policy_service =
```
netzenbot, line 38: The previous code used explicit `policy::PolicyService*` and `const policy::PolicyMap&` types. The new code replaces these with `auto*` and `const auto&` in four places. Per coding standards, spell out types rather than using `auto` merely to avoid writing them — especially in chromium_src overrides where readers lack surrounding file context. [best practice](https://github.com/brave-experiments/brave-core-tools/tree/master/docs/best-practices/coding-standards.md#-dont-use-auto-where-style-guide-wants-explicit-types)

bbondy: I adjusted the best practice here to trigger less. Feel free to ignore this one.

PR #31503, opened 2025-09-30
Fixes RPM build on Linux.

mkarolin:
Resolves https://github.com/brave/brave-browser/issues/49748


bridiver (approved):
Please follow-up to copy/rename these in the script or just use the same name as upstream so we don't have to patch here

PR #31506, opened 2025-09-30
Disable and hide kSplitViewDragAndDropEnabled and settings option

simonhong:
Resolves https://github.com/brave/brave-browser/issues/49711

Disable `kSplitViewDragAndDropEnabled` by default and hide it from settings.

TEST=BraveProfilePrefsBrowserTest.DisableGoogleServicesByDefault

Manual test:
1. Launch brave with clean profile
2. Load brave://settings/braveContent
3. Check there is no settings for split view DnD

### DIFF
--- a/browser/brave_prefs_browsertest.cc
+++ b/browser/brave_prefs_browsertest.cc
@@ -157,6 +157,9 @@ IN_PROC_BROWSER_TEST_F(BraveProfilePrefsBrowserTest,
 #endif
   EXPECT_TRUE(chrome_test_utils::GetProfile(this)->GetPrefs()->GetBoolean(
       policy::policy_prefs::kHideWebStoreIcon));
+
+  EXPECT_FALSE(chrome_test_utils::GetProfile(this)->GetPrefs()->GetBoolean(
+      prefs::kSplitViewDragAndDropEnabled));
 }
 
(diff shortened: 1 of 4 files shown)


mkarolin (approved):
LGTM

kjozwiak:

```
Brave | 1.85.15 Chromium: 141.0.7390.43 (Official Build) nightly (64-bit)
-- | --
Revision | 09df39440ba1e2a8386c78b4b2886e033fbed14b
OS | Windows 11 Version 25H2 (Build 26200.6713)
```

Using the STR/Cases outlined via https://github.com/brave/brave-browser/issues/49711#issue-3465025541, reproduced the crash using `1.85.14 Chromium: 141.0.7390.43` where `DnD`  a `tab group` would crash the browser. Checked `brave://settings/braveContent` as per https://github.com/brave/brave-core/pull/31506#issue-3467275151 and verified that `Allow split view drag-and-drop on left or right edge of window` existed.

Screenshot 2025-09-30 191315

Using the same method/STR/Cases mentioned above, ensured that:

* ensured that `Allow split view drag-and-drop on left or right edge of window` isn't available via `brave://settings/braveContent`
* ensured that Brave doesn't crash when `DnD`/detaching `tab groups` into a separate window (tried ~10 times)

image

PR #31510, opened 2025-09-30
[Tree Tabs] Rename TreeTabNode::GetChildren to GetDirectChildren

sangwoo108:
As we already have TabCollection::GetChildren method, we shouldn't hide it. So rename TreeTabNode::GetChildren to GetDirectChildren.

Resolves https://github.com/brave/brave-browser/issues/49753


sangwoo108:
Will be addressed from cr142.

PR #31511, opened 2025-09-30
[AI Chat][Content Agent] convert annotated page content to tool use output content blocks

petemill:
Continuing to merge parts of the content agent feature, this provides LLM tool use response content blocks that represent page content.

The response is pseudo-html focusing on maintaining:
- Text content
- Interactivity specifications
- Targeting via DOM ID or coordinates
- Scroll state
- Hierarchy (via xml-like syntax and indentation)
- Minimal characters (via flattening containers and supressing attributes where not useful)

See the main added function as well as associated tests:
```cpp
// Converts web content (in the form of AnnotatedPageContent) in to LLM-readable
// content (in the form of ContentBlocks), suitable for AI Tool responses.
// The conversion creates structured text that helps
// AI understand page content, target actions against specific elements and
// understand the state of the viewport.
//
// The output includes:
// - Summary of interactive elements with DOM IDs and coordinates
// - Hierarchical content structure
// - Form data and controls
// - Accessibility information
// - Scroll data for viewport and elements
std::vector<mojom::ContentBlockPtr> ConvertAnnotatedPageContentToBlocks(
    const optimization_guide::proto::AnnotatedPageContent& page_content);
```

~560 LOC, single class
~1000 LOT (lines of test, in case that's mentally helpful info)

Resolves:
- Fix https://github.com/brave/brave-browser/issues/49301

### DIFF
--- a/browser/ai_chat/page_content_blocks.cc
+++ b/browser/ai_chat/page_content_blocks.cc
@@ -11,8 +11,8 @@
 #include "base/strings/strcat.h"
 #include "base/strings/string_util.h"
 #include "brave/components/ai_chat/core/browser/tools/tool_utils.h"
-#include "brave/components/ai_chat/core/common/features.h"
 #include "brave/components/ai_chat/core/common/constants.h"
+#include "brave/components/ai_chat/core/common/features.h"
 #include "brave/components/ai_chat/core/common/mojom/ai_chat.mojom.h"
 #include "third_party/abseil-cpp/absl/strings/str_format.h"
 
@@ -126,18 +126,18 @@ std::string BuildAttributes(const ContentAttributes& attrs,
   // Add DOM node ID if available
   if ((is_interactive || !id_only_for_interactive) &&
       attrs.has_common_ancestor_dom_node_id()) {
-    attr_result +=
-        absl::StrFormat(" dom_id=\"%d\"", attrs.common_ancestor_dom_node_id());
+    absl::StrAppendFormat(&attr_result, " dom_id=\"%d\"",
+                          attrs.common_ancestor_dom_node_id());
   }
 
   // Add interaction capabilities
   if (attrs.has_interaction_info()) {
     const auto& interaction = attrs.interaction_info();
     if (interaction.is_clickable()) {
-      attr_result += " clickable";
+      attr_result.append(" clickable");
     }
     if (interaction.is_editable()) {
-      attr_result += " editable";
+      attr_result.append(" editable");
     }
     if (interaction.has_scroller_info()) {
       const auto& scroller_info = interaction.scroller_info();
@@ -145,14 +145,14 @@ std::string BuildAttributes(const ContentAttributes& attrs,
           scroller_info.user_scrollable_vertical()) {
         auto size = scroller_info.scrolling_bounds();
         auto visible_area = scroller_info.visible_area();
-        attr_result += " scrollable";
+        attr_result.append(" scrollable");
         // Size in XxY
-        attr_result +=
-            absl::StrFormat(" size=\"%dx%d\"", size.width(), size.height());
+        absl::StrAppendFormat(&attr_result, " size=\"%dx%d\"", size.width(),
+                              size.height());
         // Visible area size and position
-        attr_result += absl::StrFormat(
-            " visible_area=\"%dx%d,%d,%d\"", visible_area.width(),
-            visible_area.height(), visible_area.x(), visible_area.y());
+        absl::StrAppendFormat(&attr_result, " visible_area=\"%dx%d,%d,%d\"",
+                              visible_area.width(), visible_area.height(),
+                              visible_area.x(), visible_area.y());
       }
     }
     // if (interaction.is_focusable()) {
@@ -169,9 +169,9 @@ std::string BuildAttributes(const ContentAttributes& attrs,
     if (kShouldIncludeGeometry && is_interactive && attrs.has_geometry() &&
         attrs.geometry().has_outer_bounding_box()) {
       const auto& bbox = attrs.geometry().outer_bounding_box();
-      attr_result +=
-          absl::StrFormat(" x=\"%d\" y=\"%d\" width=\"%d\" height=\"%d\"",
-                          bbox.x(), bbox.y(), bbox.width(), bbox.height());
+      absl::StrAppendFormat(&attr_result,
+                            " x=\"%d\" y=\"%d\" width=\"%d\" height=\"%d\"",
+                            bbox.x(), bbox.y(), bbox.width(), bbox.height());
     }
   }
 
@@ -216,7 +216,8 @@ std::string BuildAttributes(const ContentAttributes& attrs,
       }
     }
     if (!important_roles.empty()) {
-      attr_result += base::StrCat(
+      base::StrAppend(
+          &attr_result,
           {" role=\"",
            XmlEscapeAndSanitizeText(base::JoinString(important_roles, " ")),
            "\""});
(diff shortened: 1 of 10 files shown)


browser/ai_chat/page_content_blocks.cc
```diff
@@ -0,0 +1,562 @@
...
+std::vector<mojom::ContentBlockPtr> ConvertAnnotatedPageContentToBlocks(
+    const AnnotatedPageContent& page_content) {
+  if (!page_content.has_root_node()) {
+    return {};
+  }
+
+  // Indicate that the content is external and is untrusted
+  std::string result = base::StrCat({kBraveUntrustedContentOpenTag, "\n"});
+
+  result += "=== PAGE METADATA ===\n\n";
```
darkdh: `base::StrCat({kBraveUntrustedContentOpenTag, "\n", "=== PAGE METADATA ===\n\n"});`

browser/ai_chat/page_content_blocks.cc
```diff
@@ -0,0 +1,562 @@
...
+
+  // Add page metadata
+  if (page_content.has_main_frame_data()) {
+    const auto& frame_data = page_content.main_frame_data();
+    if (frame_data.has_title()) {
+      result += base::StrCat(
+          {"PAGE TITLE: ", SanitizeContentText(frame_data.title()), "\n"});
+    }
```
darkdh: `base::StrAppend(&result, {"PAGE TITLE: ", SanitizeContentText(frame_data.title()), "\n"});`

Please update all the `+= base::StrCat`  usage to `base::StrAppend`  per
```
// StrAppend -------------------------------------------------------------------
//
// Appends a sequence of strings to a destination. Prefer:
//   StrAppend(&foo, ...);
// over:
//   foo += StrCat(...);
// because it avoids a temporary string allocation and copy.
```

browser/ai_chat/page_content_blocks.cc
```diff
@@ -0,0 +1,562 @@
...
+  // Only consider flattening if there's exactly one child
+  if (node.children_nodes_size() != 1) {
+    return false;
+  }
+
+  const auto& attrs = node.content_attributes();
+  const auto& interaction = attrs.interaction_info();
+  // Flatten non-interactive containers with only 1 child
+  // but don't consider focusable, selectable and draggable for now
+  if (interaction.is_clickable() || interaction.is_editable()) {
```
darkdh, line 41: Seems like we didn't check draggable per comment says here

browser/ai_chat/page_content_blocks.cc
```diff
@@ -0,0 +1,562 @@
...
+  }
+
+  return attr_result;
+}
+
+// Generates XML-like structured content representation with interaction
+// attributes
+std::string GenerateContentStructure(const ContentNode& node, int depth = 0) {
+  std::string content;
+  std::string indent(kShouldIndent ? depth * 2 : 0, ' ');
```
darkdh: The const bool is always false. What is the reason for this ternary?

browser/ai_chat/page_content_blocks.cc
```diff
@@ -0,0 +1,562 @@
...
+    }
+    // if (interaction.is_focusable()) {
+    //   attr_result += " focusable=\"true\"";
+    // }
+    // if (interaction.is_selectable()) {
+    //   attr_result += " selectable=\"true\"";
+    // }
+    // if (interaction.is_draggable()) {
+    //   attr_result += " draggable=\"true\"";
+    // }
```
darkdh: leftover?

browser/ai_chat/page_content_blocks.cc
```diff
@@ -0,0 +1,562 @@
...
+
+    // Add geometry only if interactive
+    if (kShouldIncludeGeometry && is_interactive && attrs.has_geometry() &&
```
darkdh: what is this const bool for?

browser/ai_chat/page_content_blocks.cc
```diff
@@ -0,0 +1,562 @@
...
+    tree_string = tree_string.substr(0, kMaxTreeStringLength) + "...</root>\n";
+    tree_string +=
+        "PAGE STRUCTURE (XML) was too long to display. Truncated.\n\n";
+  }
+
+  result += "\n=== PAGE STRUCTURE (XML representation) ===";
+
+  // Replace all occurances of the untrusted tag with an empty string
+  base::ReplaceSubstringsAfterOffset(&tree_string, 0,
+                                     kBraveUntrustedContentTagName, "");
```
darkdh: shouldn't this call `SanitizeContentText`?

browser/ai_chat/page_content_blocks.cc
```diff
@@ -0,0 +1,562 @@
...
+  result +=
+      "- x,y,width,height: Use the position/size only when neccessary or to "
+      "infer hierarchy.\n";
+  result += "- clickable: Element can be clicked\n";
+  result += "- editable: Element can receive text input\n";
+  // result += "- focusable=\"true\": Element can receive focus\n";
+  // result +=
+  //     "- role: Semantic role (header, nav, main, search, article, section, "
+  //     "aside, footer, hidden, paid)\n";
+  // result += "- label: Accessibility label for the element\n";
```
darkdh: should we keep this?

browser/ai_chat/page_content_blocks.cc
```diff
@@ -0,0 +1,562 @@
...
+  std::string output = input;
+
+  SanitizeContentText(output);
+
+  // Escape XML to avoid breaking out of pseudo-XML serialization
+  base::ReplaceChars(output, "&", "&amp;", &output);
+  base::ReplaceChars(output, "<", "&lt;", &output);
+  base::ReplaceChars(output, ">", "&gt;", &output);
+  base::ReplaceChars(output, "\"", "&quot;", &output);
+  base::ReplaceChars(output, "'", "&apos;", &output);
```
darkdh: can we use `base::EscapeForHTML`?
which is basically does the same replacement
```
 static constexpr struct {
    char key;
    std::string_view replacement;
  } kCharsToEscape[] = {
      {'<', "&lt;"},   {'>', "&gt;"},   {'&', "&amp;"},
      {'"', "&quot;"}, {'\'', "&#39;"},
  };
  ```
  `&#39;` should be compatible with both XML and HTML

browser/ai_chat/page_content_blocks.cc
petemill: Yeah I removed it as part of an effort to reduce unnecessary characters because it was taking up too much space as nearly everything was draggable. I think. Either way, I'll re-introduce in a follow-up when I can analyze it more.

browser/ai_chat/page_content_blocks.cc
petemill: Thanks, done

browser/ai_chat/page_content_blocks.cc
petemill: done

browser/ai_chat/page_content_blocks.cc
petemill: experimentation on cost-benefit on taking up space in the context vs what makes for a better LLM understanding of the structure. I've converted this to a `FeatureParam<bool>`

browser/ai_chat/page_content_blocks.cc
petemill: Experimentation. I'll remove for now. We can re-introduce as a function parameter if we use this output for regular non-agent content.

browser/ai_chat/page_content_blocks.cc
petemill: removed for now

browser/ai_chat/page_content_blocks.cc
petemill: removed for now

browser/ai_chat/page_content_blocks.cc
petemill: I can't remember why I didn't use this. I think I saw it! But I'll use it and hopefully there was no reason.

browser/ai_chat/page_content_blocks.cc
petemill: Thanks, missed one after I introduced function

browser/ai_chat/page_content_blocks.cc
petemill: I'm not concerned with XML and HTML strictness, it's about LLM interpretation.  Hopefully LLMs will understand `&#39;` just as well as `&apos;`.

darkdh (approved):
++

PR #31514, opened 2025-09-30
Ignore os.link failure for race condition on android

bridiver:
Resolves https://github.com/brave/brave-browser/issues/49751


mkarolin (approved):
++

PR #31516, opened 2025-09-30
[perf] Load brave news images on demand

atuchin-m:
Resolves https://github.com/brave/brave-browser/issues/48998
The PR tracks the scroll event and postpone loading the most news-related images (marked as `loading=lazy`) until the scroll is started.
It also add 400ms fade effect when a loaded image appears to add smoothness (without on-disk caching it's still impossible to load images instantly).
A video how it looks like: https://github.com/user-attachments/assets/f57dbfa1-ae04-4548-82fd-ec4766e33369

### DIFF
--- a/browser/resources/brave_new_tab_page_refresh/components/common/safe_image.tsx
+++ b/browser/resources/brave_new_tab_page_refresh/components/common/safe_image.tsx
@@ -29,17 +29,16 @@ export function SafeImage(props: Props) {
     src = placeholderImageSrc
   }
 
-  const { scrolledToNews } = useBraveNews()
-  const delayLoad = !scrolledToNews && loading === 'lazy'
+  const { shouldRenderImages } = useBraveNews()
+  const delayLoad = !shouldRenderImages && loading === 'lazy'
 
   return (
     <img
-      src={!delayLoad ? src: undefined}
+      src={delayLoad ? undefined : src}
       loading={loading ?? 'lazy'}
       className={props.className}
       onError={(event) => { event.currentTarget.src = placeholderImageSrc }}
-      onLoad={(event) => { event.currentTarget.style.opacity = '1' }}
-      style={{ opacity: 0, transition: 'opacity 400ms ease-in-out' }}
+      onLoad={(event) => { event.currentTarget?.classList.add('loaded') }}
     />
   )
 }
--- a/browser/resources/brave_new_tab_page_refresh/components/common/safe_image.tsx
+++ b/browser/resources/brave_new_tab_page_refresh/components/common/safe_image.tsx
@@ -10,7 +10,6 @@ import { placeholderImageSrc } from '../../lib/image_loader'
 interface Props {
   src: string
   className?: string
-  loading?: 'lazy' | 'eager'
   targetSize?: { width: number, height: number }
 }
 
--- a/browser/resources/brave_new_tab_page_refresh/components/common/safe_image.tsx
+++ b/browser/resources/brave_new_tab_page_refresh/components/common/safe_image.tsx
@@ -6,15 +6,17 @@
 import * as React from 'react'
 
 import { placeholderImageSrc } from '../../lib/image_loader'
+import { useBraveNews } from '../../../../../components/brave_news/browser/resources/shared/Context'
 
 interface Props {
   src: string
   className?: string
+  loading?: 'lazy' | 'eager'
   targetSize?: { width: number, height: number }
 }
 
 export function SafeImage(props: Props) {
-  let { src } = props
+  let { src, loading } = props
   if (src) {
     src = 'chrome://brave-image?url=' + encodeURIComponent(src);
     if (props.targetSize) {
@@ -23,13 +25,21 @@ export function SafeImage(props: Props) {
       height = Math.round(height * window.devicePixelRatio)
       src += `&target_size=${width}x${height}`
     }
+  } else {
+    src = placeholderImageSrc
   }
+
+  const { scrolledToNews } = useBraveNews()
+  const delayLoad = !scrolledToNews && loading === 'lazy'
+
   return (
     <img
-      src={src || placeholderImageSrc}
-      loading='lazy'
+      src={!delayLoad ? src: undefined}
+      loading={loading ?? 'lazy'}
       className={props.className}
       onError={(event) => { event.currentTarget.src = placeholderImageSrc }}
+      onLoad={(event) => { event.currentTarget.style.opacity = '1' }}
+      style={{ opacity: 0, transition: 'opacity 400ms ease-in-out' }}
     />
   )
 }
--- a/browser/resources/brave_new_tab_page_refresh/components/common/safe_image.tsx
+++ b/browser/resources/brave_new_tab_page_refresh/components/common/safe_image.tsx
@@ -15,7 +15,7 @@ interface Props {
 }
 
 export function SafeImage(props: Props) {
-  let { src, loading } = props
+  let { src } = props
   if (src) {
     src = 'chrome://brave-image?url=' + encodeURIComponent(src);
     if (props.targetSize) {
@@ -24,14 +24,12 @@ export function SafeImage(props: Props) {
       height = Math.round(height * window.devicePixelRatio)
       src += `&target_size=${width}x${height}`
     }
-  } else {
-    src = placeholderImageSrc
   }
 
   return (
     <img
-      src={src}
-      loading={loading ?? 'lazy'}
+      src={src || placeholderImageSrc}
+      loading='lazy'
       className={props.className}
       onError={(event) => { event.currentTarget.src = placeholderImageSrc }}
       onLoad={(event) => { event.currentTarget?.classList.add('loaded') }}
--- a/browser/resources/brave_new_tab_page_refresh/components/common/safe_image.tsx
+++ b/browser/resources/brave_new_tab_page_refresh/components/common/safe_image.tsx
@@ -6,7 +6,6 @@
 import * as React from 'react'
 
 import { placeholderImageSrc } from '../../lib/image_loader'
-import { useBraveNews } from '../../../../../components/brave_news/browser/resources/shared/Context'
 
 interface Props {
   src: string
@@ -29,12 +28,9 @@ export function SafeImage(props: Props) {
     src = placeholderImageSrc
   }
 
-  const { shouldRenderImages } = useBraveNews()
-  const delayLoad = !shouldRenderImages && loading === 'lazy'
-
   return (
     <img
-      src={delayLoad ? undefined : src}
+      src={src}
       loading={loading ?? 'lazy'}
       className={props.className}
       onError={(event) => { event.currentTarget.src = placeholderImageSrc }}
--- a/browser/resources/brave_new_tab_page_refresh/components/news/news_feed.tsx
+++ b/browser/resources/brave_new_tab_page_refresh/components/news/news_feed.tsx
@@ -22,12 +22,6 @@ export function NewsFeed() {
     setTimeout(() => setShouldRenderNews(true), 1000)
   }, [])
 
-  React.useEffect(() => {
-    const handleScroll = () => braveNews.setShouldRenderImages(true)
-    document.addEventListener('scroll', handleScroll, { once: true })
-    return () => document.removeEventListener('scroll', handleScroll)
-  }, [braveNews.shouldRenderImages])
-
   if (!newsFeatureEnabled) {
     return null
   }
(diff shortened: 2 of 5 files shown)


browser/resources/brave_new_tab_page_refresh/components/common/safe_image.tsx
```diff
@@ -6,15 +6,17 @@
...
+import { useBraveNews } from '../../../../../components/brave_news/browser/resources/shared/Context'
 
 interface Props {
   src: string
   className?: string
+  loading?: 'lazy' | 'eager'
   targetSize?: { width: number, height: number }
 }
 
 export function SafeImage(props: Props) {
```
atuchin-m, line 16: The current usage of `SafeImage` don't look like they need to be postponed.
But I also added this positioning capability to SafeImage anyway.

components/brave_news/browser/resources/feed/Card.tsx
```diff
@@ -69,7 +73,14 @@ const HidableImage = ({
     src += `&target_size=${width}x${height}`
   }
 
-  return <img {...rest} src={src} ref={ref as any} onError={handleError} onLoad={handleLoad} />
+  return <img
+    {...rest}
```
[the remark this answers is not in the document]
atuchin-m: Was here before my changes.
I can fix it if @brave/brave-core-ntp is okay with that.

components/brave_news/browser/resources/shared/Context.tsx
```diff
@@ -47,6 +47,9 @@ interface BraveNewsContext {
   reportVisit: (depth: number) => void
   reportSidebarFilterUsage: () => void
   reportSessionStart: () => void
+
+  scrolledToNews: boolean
```
zenparsing: Since this UI is designed to work standalone, I would generalize this name to something like `isNewsVisible`. Or it could also be more specific to the intended behavior, as in `shouldRenderImages`.

components/brave_news/browser/resources/shared/Context.tsx
```diff
@@ -115,6 +121,16 @@ export function BraveNewsContextProvider(props: { children: React.ReactNode }) {
     return () => configurationCache.removeListener(setConfiguration)
   }, [])
 
+  React.useEffect(() => {
+    if (scrolledToNews) {
+      return
+    }
+
+    const handleScroll = () => setScrolledToNews(true)
+    document.addEventListener('scroll', handleScroll, { once: true })
```
zenparsing: If we want to keep the News UI as potentially standalone, then we may want to set this flag from outside of news itself.

Moving this effect to `news_feed.tsx`, for example, might work.

atuchin-m:
@rebron the PR has a small visual improvement: 400ms fade when we show loaded images. Can you check the video and tell me is it okay?

browser/resources/brave_new_tab_page_refresh/components/common/safe_image.tsx
```diff
@@ -23,13 +25,21 @@ export function SafeImage(props: Props) {
...
+
   return (
     <img
-      src={src || placeholderImageSrc}
-      loading='lazy'
+      src={!delayLoad ? src: undefined}
+      loading={loading ?? 'lazy'}
       className={props.className}
       onError={(event) => { event.currentTarget.src = placeholderImageSrc }}
+      onLoad={(event) => { event.currentTarget.style.opacity = '1' }}
```
zenparsing: The fade in looks nice. I would specifically target the news widget element for this fade in though.

Here's one way we might do that: in this `onLoad` event, set a `className` on the element (like 'loaded').

Then, in `news_widget.style.ts`, add the rules to the `img` selector:

```css
  img {
    /* ... */
    opacity: 0;
    transition: opacity 400ms ease-in-out;
    &.loaded {
      opacity: 1;
    }
  }
```

wdyt?

browser/resources/brave_new_tab_page_refresh/components/widgets/news_widget.style.ts
```diff
@@ -88,6 +88,16 @@ export const style = scoped.css`
     }
   }
 
+  .peek {
+    img {
+      opacity: 0;
```
zenparsing: Totally optional, but these rules could also be put into the selector starting at line 60.

browser/resources/brave_new_tab_page_refresh/components/common/safe_image.tsx
atuchin-m: Done, thanks for the suggestion.

components/brave_news/browser/resources/shared/Context.tsx
atuchin-m: Done.

components/brave_news/browser/resources/shared/Context.tsx
atuchin-m: Done.

rebron:
> @rebron the PR has a small visual improvement: 400ms fade when we show loaded images. Can you check the video and tell me is it okay?

Looks nice.

components/brave_news/browser/resources/feed/Card.tsx
```diff
@@ -69,7 +73,14 @@ const HidableImage = ({
...
+    src={delayLoad ? undefined : src}
+    ref={ref as any}
+    onError={handleError}
+    onLoad={handleLoad}
+    style={{ opacity: 0, transition: 'opacity 400ms ease-in-out', ...rest.style }}
```
fallaciousreasoning: I think rather than setting this style like this we should have a styled component:

```
const BaseImg = styled.img`
  opacity: 0;
  transition: 'opacity, 400ms, ease-in-out'
`;
```

and use that instead of the `img` here.

components/brave_news/browser/resources/feed/Card.tsx
```diff
@@ -69,7 +73,14 @@ const HidableImage = ({
     src += `&target_size=${width}x${height}`
   }
 
-  return <img {...rest} src={src} ref={ref as any} onError={handleError} onLoad={handleLoad} />
+  return <img
+    {...rest}
+    src={delayLoad ? undefined : src}
+    ref={ref as any}
+    onError={handleError}
```
fallaciousreasoning: do we still need this `onError` handler, now the image isn't set to visible until it loads?

components/brave_news/browser/resources/feed/Card.tsx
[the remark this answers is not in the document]
fallaciousreasoning: I think the spread is correct here, as this is just a wrapper for the img component.

browser/resources/brave_new_tab_page_refresh/components/news/news_feed.tsx
```diff
@@ -22,6 +22,12 @@ export function NewsFeed() {
     setTimeout(() => setShouldRenderNews(true), 1000)
   }, [])
 
+  React.useEffect(() => {
+    const handleScroll = () => braveNews.setShouldRenderImages(true)
+    document.addEventListener('scroll', handleScroll, { once: true })
+    return () => document.removeEventListener('scroll', handleScroll)
+  }, [braveNews.shouldRenderImages])
```
fallaciousreasoning: Hey, I wonder if this should just live in `Context.tsx`, as it doesn't really depend on anything in this file.

Then we could remove the `setShouldRenderImages` from the context (as nothing outside there would ever set it).

Also, if we don't this will break all the images in the old New Tab Page

browser/resources/brave_new_tab_page_refresh/components/news/news_feed.tsx
```diff
@@ -22,6 +22,12 @@ export function NewsFeed() {
     setTimeout(() => setShouldRenderNews(true), 1000)
   }, [])
 
+  React.useEffect(() => {
+    const handleScroll = () => braveNews.setShouldRenderImages(true)
```
fallaciousreasoning: no point attaching these handlers if `shouldRenderImages` is true
```suggestion
    if (braveNews.shouldRenderImages) {
        return
    }
    const handleScroll = () => braveNews.setShouldRenderImages(true)
```

browser/resources/brave_new_tab_page_refresh/components/common/safe_image.tsx
```diff
@@ -23,13 +24,17 @@ export function SafeImage(props: Props) {
       height = Math.round(height * window.devicePixelRatio)
       src += `&target_size=${width}x${height}`
     }
+  } else {
+    src = placeholderImageSrc
```
fallaciousreasoning: why this change?

browser/resources/brave_new_tab_page_refresh/components/common/safe_image.tsx
fallaciousreasoning: Hey so I wonder if we should revert the `loading` change - doesn't ever get used and its already defaulting to `loading=lazy`. If we don't need `loading=eagre` anywhere yet, lets not add it

browser/resources/brave_new_tab_page_refresh/components/news/news_feed.tsx
zenparsing: I suggested moving the scroll detection here, since News is presumably designed to work standalone as well?

components/brave_news/browser/resources/feed/Card.tsx
atuchin-m: Looks like we can drop it there.

browser/resources/brave_new_tab_page_refresh/components/news/news_feed.tsx
fallaciousreasoning: The fade in already isn't going to work in standalone (and we don't have a standalone :laughing:), so I think better to put it in the Context for now. 

If we want it to work in Standalone, we'll probably need to revisit how the fade works anyway

browser/resources/brave_new_tab_page_refresh/components/common/safe_image.tsx
atuchin-m: If was used in a version with `src=undefined`.
Reverted.

browser/resources/brave_new_tab_page_refresh/components/news/news_feed.tsx
atuchin-m: Ok, I've decided to move the code back to `Context.tsx`.
In case of doubts that approach just has less code and exposes less outside.

browser/resources/brave_new_tab_page_refresh/components/widgets/news_widget.style.ts
atuchin-m: Done

components/brave_news/browser/resources/feed/Card.tsx
atuchin-m: Done

components/brave_news/browser/resources/feed/Card.tsx
atuchin-m: Nice, done.

browser/resources/brave_new_tab_page_refresh/components/news/news_feed.tsx
atuchin-m: done.

browser/resources/brave_new_tab_page_refresh/components/common/safe_image.tsx
atuchin-m: Ok, reverted `loading` changes.

browser/resources/brave_new_tab_page_refresh/components/common/safe_image.tsx
```diff
@@ -10,6 +10,7 @@ import { placeholderImageSrc } from '../../lib/image_loader'
 interface Props {
   src: string
   className?: string
+  loading?: 'lazy' | 'eager'
```
fallaciousreasoning: nit: think you forgot to move the pop

components/brave_news/browser/resources/shared/Context.tsx
```diff
@@ -127,6 +131,15 @@ export function BraveNewsContextProvider(props: { children: React.ReactNode }) {
     return () => { publishersCache.removeListener(handler) }
   }, [])
 
+  React.useEffect(() => {
+    if (shouldRenderImages) {
+      return
+    }
```
fallaciousreasoning: sorry, probably don't need this anymore now its been removed from the deps array

fallaciousreasoning (approved):
lgtm % nits. feel free to resolve and merge

components/brave_news/browser/resources/shared/Context.tsx
atuchin-m: done.

browser/resources/brave_new_tab_page_refresh/components/common/safe_image.tsx
atuchin-m: Ah, my bad, thanks.
Fixed.

PR #31528, opened 2025-09-30
[Android] Update header image for quick shortcuts and optimization

samartnik:
Optimizations done:
- Initialize leo and wallet controllers regardless if feature enabled or not. This is because at this point native is not guaranteed to be initialized to check the feature. Check for option visibility is done in `BraveRadioButtonGroupAdaptiveToolbarPreference` instead.
- Override `buildUiStateForStats` to avoid asserts on Brave specific options.

Resolves https://github.com/brave/brave-browser/issues/49779


Olipro:
Looks like this broke the `BraveAdaptiveToolbarSettingsFragmentTest.testSelectShortcuts` test. Perhaps the `isNativeInitialized()` check should be left in? The comment makes me think it's only false in a testing scenario, but I'm not 100%

samartnik:
Here is the look of the new icon
Screenshot_20250930_132237

AlexeyBarabash (approved):
++

PR #31532, opened 2025-09-30
Use IsPrimaryMainFrame for DomainBlockNavigationThrottle

bridiver:
…n frame

This may possibly fix https://github.com/brave/brave-browser/issues/44657 it's not clear what is going on there, but either way this should only run for the primary main frame.


bridiver:
> I'm not sure `IsInMainFrame-->IsInPrimaryMainFrame` fixes the crash, but `IsInPrimaryMainFrame` is used in upstream throttles and it looks like it's more properly distinguish the primary main frame.

Yes, I'm not sure either, I think it may just be an upstream issue, but it's definitely what we should have here so it won't hurt either way

bridiver:
unrelated test failures

PR #31542, opened 2025-10-01
[AI Chat] [Content Agent] Task UI

petemill:
Simplifies the display of regular tool use UI components, but adds a new Task UI that groups response events and allows the user to view either only the most recent LLM text completion and tool use requests, or a list of all the messages during the current event loop

- Resolves https://github.com/brave/brave-browser/issues/49797

### DIFF
--- a/components/resources/ai_chat_ui_strings.grdp
+++ b/components/resources/ai_chat_ui_strings.grdp
@@ -645,4 +645,10 @@
   <message name="IDS_CHAT_UI_MEMORY_ERROR_LABEL" desc="Message shown when memory storage fails" formatter_data="webui=AiChat">
     Failed to save memory. Please try again.
   </message>
+  <message name="IDS_AI_CHAT_TASK_PROGRESS_LABEL" desc="Label for the task progress tab in the assistant response" formatter_data="webui=AiChat">
+    Progress
+  </message>
+  <message name="IDS_AI_CHAT_TASK_STEPS_LABEL" desc="Label for the task steps tab in the assistant response" formatter_data="webui=AiChat">
+    Steps
+  </message>
 </grit-part>
--- a/components/ai_chat/resources/page/stories/components_panel.tsx
+++ b/components/ai_chat/resources/page/stories/components_panel.tsx
@@ -133,7 +133,19 @@ const CONVERSATIONS: Mojom.Conversation[] = [
 const toolEvents: Mojom.ToolUseEvent[] = [
   {
     id: 'abc123d',
+    toolName: 'tool_one',
+    argumentsJson: JSON.stringify({ choices: ['7:00pm', '8:00pm'] }),
+    output: [createTextContentBlock('7:00pm')],
+  },
+  {
+    id: 'abc123d',
     toolName: 'user_choice_tool',
+    argumentsJson: JSON.stringify({ choices: ['7:00pm', '8:00pm'] }),
+    output: [createTextContentBlock('7:00pm')],
+  },
+  {
+    id: 'abc123d',
+    toolName: 'tool_two',
     argumentsJson: JSON.stringify({ choices: ['7:00pm', '8:00pm'] }),
     output: [createTextContentBlock('7:00pm')],
   },
@@ -755,6 +767,9 @@ const HISTORY: Mojom.ConversationTurn[] = [
           + '1962              |\n'
           + "\n\n Let me know if you'd like more details!",
       ),
+      ...toolEvents
+        .slice(0, 3)
+        .map((toolUseEvent) => ({ ...eventTemplate, toolUseEvent })),
     ],
     uploadedFiles: [],
     fromBraveSearchSERP: false,
@@ -775,8 +790,8 @@ const HISTORY: Mojom.ConversationTurn[] = [
         'Pointer compression is a memory optimization technique where pointers are stored in a compressed format to save memory.',
       ),
       ...toolEvents
-        .slice(0, 3)
-        .map((toolUseEvent) => ({ ...eventTemplate, toolUseEvent })),
+        .slice(3, 5)
+        .map((toolEvent) => ({ ...eventTemplate, toolUseEvent: toolEvent })),
     ],
     uploadedFiles: [],
     fromBraveSearchSERP: false,
@@ -793,9 +808,12 @@ const HISTORY: Mojom.ConversationTurn[] = [
     edits: [],
     createdTime: { internalValue: BigInt('13278618001000000') },
     events: [
-      ...toolEvents
-        .slice(3)
-        .map((toolEvent) => ({ ...eventTemplate, toolUseEvent: toolEvent })),
+      getCompletionEvent('Ah, now I have the answer!'),
+      {
+        ...eventTemplate,
+        toolUseEvent: toolEvents[4],
+      },
+      },
     ],
     uploadedFiles: [],
     fromBraveSearchSERP: false,
--- a/components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/assistant_task.module.scss
+++ b/components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/assistant_task.module.scss
@@ -1,0 +1,58 @@
+.task {
+  border: 1px solid var(--leo-color-divider-subtle);
+  border-radius: var(--leo-radius-xl);
+  min-height: 350px;
+  box-shadow: var(--leo-effect-elevation-01);
+}
+
+.taskContent {
+  padding: var(--leo-spacing-l);
+  display: flex;
+  flex-direction: row;
+  gap: var(--leo-spacing-xs);
+  flex-wrap: wrap;
+}
+
+.taskData {
+  flex: 1 1 0;
+  min-width: min(300px, 100%);
+}
+
+.taskImage {
+  // display: none;
+  flex: 1 1 0;
+  min-width: min(300px, 100%);
+}
+
+.taskStep {
+  min-height: 30px;
+  margin-left: 11px;
+  border-left: var(--leo-color-neutral-20) 2px solid;
+  padding: 8px 12px 8px 22px;
+  font: var(--leo-font-default-regular);
+  color: var(--leo-color-text-primary);
+  position: relative;
+
+  .taskIcon {
+    --leo-icon-size: 16px;
+    --leo-progressring-size: 16px;
+    position: absolute;
+    top: calc(8px + 0.5lh - 11px);
+    left: -12px;
+    padding: 1px;
+    width: 20px;
+    height: 20px;
+    // circle
+    border-radius: 50%;
+    border: var(--leo-color-divider-subtle) 1px solid;
+    background: var(--leo-color-container-background);
+  }
+
+  &.isComplete {
+    .taskIcon {
+      --leo-icon-color: var(--leo-color-systemfeedback-success-icon);
+      background: var(--leo-color-systemfeedback-success-background);
+      border-color: var(--leo-color-systemfeedback-success-background);
+    }
+  }
+}
--- a/components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/assistant_task.tsx
+++ b/components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/assistant_task.tsx
@@ -1,0 +1,211 @@
+// Copyright (c) 2025 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import * as React from 'react'
+import Tabs from '@brave/leo/react/tabs'
+import TabItem from '@brave/leo/react/tabItem'
+import { getLocale } from '$web-common/locale'
+import API from '../../untrusted_conversation_frame_api'
+import * as Mojom from '../../../common/mojom'
+import { useUntrustedConversationContext } from '../../untrusted_conversation_context'
+import AssistantResponse from '.'
+import ToolEvent from './tool_event'
+import styles from './assistant_task.module.scss'
+import classnames from '$web-common/classnames'
+import ProgressRing from '@brave/leo/react/progressRing'
+import Icon from '@brave/leo/react/icon'
+
+interface Props {
+  // Entries that make up the task loop
+  assistantEntries: Mojom.ConversationTurn[]
+  isActiveTask: boolean
+  isGenerating: boolean
+}
+
+// Should be Navigate and TODO
+const importantToolNames = [Mojom.NAVIGATE_TOOL_NAME]
+
+function getImportantToolUseEvents(
+  assistantEntries: Mojom.ConversationTurn[],
+): Mojom.ToolUseEvent[] {
+  const importantToolUseEvents: Partial<
+    Record<(typeof importantToolNames)[number], Mojom.ToolUseEvent>
+  > = {}
+
+  for (const event of assistantEntries
+    .toReversed()
+    .flatMap((entry) => entry.events?.toReversed() ?? [])) {
+    if (
+      !!event
+      && !!event.toolUseEvent
+      && importantToolNames.includes(event.toolUseEvent.toolName ?? '')
+    ) {
+      if (importantToolUseEvents[event.toolUseEvent.toolName]) {
+        continue
+      }
+      importantToolUseEvents[event.toolUseEvent.toolName] = event.toolUseEvent!
+      // Don't keep iterating if we've found all the important tool use events
+      if (
+        Object.keys(importantToolUseEvents).length === importantToolNames.length
+      ) {
+        break
+      }
+    }
+  }
+  return Object.values(importantToolUseEvents).filter((event) => !!event)
+}
+
+export default function AssistantTask(props: Props) {
+  const [showSteps, setShowSteps] = React.useState(false)
+  const [taskThumbnail, setTaskThumbnail] = React.useState<string>()
+  const conversationContext = useUntrustedConversationContext()
+
+  React.useEffect(() => {
+    if (!conversationContext.contentTaskTabId) {
+      return
+    }
+    const id = API.getInstance().uiObserver.thumbnailUpdated.addListener(
+      (tabId: number, dataURI: string) => {
+        console.error(dataURI)
+        if (
+          tabId === conversationContext.contentTaskTabId
+          && props.isActiveTask
+        ) {
+          setTaskThumbnail(dataURI)
+        }
+      },
+    )
+
+    return () => {
+      API.getInstance().uiObserver.removeListener(id)
+    }
+  }, [props.isActiveTask, conversationContext.contentTaskTabId])
+
+  return (
+    <div className={styles.task}>
+      <Tabs
+        onChange={() => setShowSteps(!showSteps)}
+        value={showSteps ? 'steps' : 'progress'}
+        size='medium'
+      >
+        <TabItem value='progress'>
+          {getLocale(S.AI_CHAT_TASK_PROGRESS_LABEL)}
+        </TabItem>
+        <TabItem value='steps'>{getLocale(S.AI_CHAT_TASK_STEPS_LABEL)}</TabItem>
+      </Tabs>
+      <div className={styles.taskContent}>
+        <div className={styles.taskData}>
+          {showSteps && <Steps {...props} />}
+
+          {!showSteps && <Progress {...props} />}
+        </div>
+        {taskThumbnail && (
+          <div className={styles.taskImage}>
+            <img src={taskThumbnail} />
+          </div>
+        )}
+      </div>
+    </div>
+  )
+}
+
+function Progress(props: Props) {
+  // The Progress tab should show:
+  // - Any last active complete "important" tool (TODO, navigate)
+  // - the most recent completion event
+  // - any tool use events from the most recent entry in the group (the active
+  // events)
+  const conversationContext = useUntrustedConversationContext()
+
+  const lastCompletionEvent = props.assistantEntries
+    .flatMap(
+      (entry) =>
+        entry.events?.filter((event) => !!event && event.completionEvent) ?? [],
+    )
+    .at(-1)
+
+  const lastEntryToolUseEvents =
+    props.assistantEntries
+      .at(-1)
+      ?.events?.filter((event) => !!event && !!event.toolUseEvent) ?? []
+
+  const importantToolUseEvents = getImportantToolUseEvents(
+    props.assistantEntries,
+  )
+
+  return (
+    <div>
+      {importantToolUseEvents.map((event, index) => (
+        <ToolEvent
+          key={index}
+          toolUseEvent={event}
+          isEntryActive={props.isActiveTask}
+        />
+      ))}
+      {lastCompletionEvent && (
+        <AssistantResponse
+          events={[lastCompletionEvent]}
+          isEntryInteractivityAllowed={false}
+          isEntryInProgress={props.isGenerating}
+          allowedLinks={[]}
+          isLeoModel={conversationContext.isLeoModel}
+        />
+      )}
+      {lastEntryToolUseEvents.map((event, index) => (
+        <ToolEvent
+          key={index}
+          toolUseEvent={event.toolUseEvent!}
+          isEntryActive={props.isActiveTask}
+        />
+      ))}
+    </div>
+  )
+}
+
+function Steps(props: Props) {
+  // Render every event in the task, split by completion event
+  // so that the LLM tells a story of the task by it's own progress
+  // description.
+  const conversationContext = useUntrustedConversationContext()
+  let taskItems: Mojom.ConversationEntryEvent[][] = [[]]
+  const allowedLinks: string[] = []
+  for (const event of props.assistantEntries.flatMap(
+    (entry) => entry.events ?? [],
+  )) {
+    if (!!event && !!event.completionEvent) {
+      taskItems.push([event])
+    } else if (event) {
+      taskItems[taskItems.length - 1].push(event)
+      if (event.sourcesEvent) {
+        allowedLinks.push(
+          ...event.sourcesEvent.sources.map((source) => source.url.url),
+        )
+      }
+    }
+  }
+
+  taskItems = taskItems.filter((taskItem) => taskItem.length > 0)
+
+  return taskItems.map((taskItem, index) => {
+    const isActive = index === taskItems.length - 1
+    return (
+      <div
+        key={index}
+        className={classnames(styles.taskStep, !isActive && styles.isComplete)}
+      >
+        <div className={styles.taskIcon}>
+          {isActive ? <ProgressRing /> : <Icon name='check-circle-outline' />}
+        </div>
+        <AssistantResponse
+          events={taskItem}
+          isEntryInteractivityAllowed={isActive}
+          isEntryInProgress={props.isGenerating && isActive}
+          allowedLinks={allowedLinks}
+          isLeoModel={conversationContext.isLeoModel}
+        />
+      </div>
+    )
+  })
+}
--- a/components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/tool_event.module.scss
+++ b/components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/tool_event.module.scss
@@ -3,51 +3,42 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this file,
 // You can obtain one at https://mozilla.org/MPL/2.0/.
 
-.toolUse {
-  margin-left: 11px;
-  border-left: var(--leo-color-neutral-20) 2px solid;
-  padding: 8px 12px 8px 22px;
-  font: var(--leo-font-default-regular);
-  color: var(--leo-color-text-secondary);
-  position: relative;
-
-  .toolUseIcon {
-    --leo-icon-size: 16px;
-    --leo-progressring-size: 16px;
-    position: absolute;
-    top: calc(8px + 0.5lh - 11px);
-    left: -12px;
-    padding: 2px;
-    width: 22px;
-    height: 22px;
-    // circle
-    border-radius: 50%;
-    border: var(--leo-color-divider-subtle) 1px solid;
-    background: var(--leo-color-container-background);
+@keyframes shimmer {
+  0% {
+    background-position: 200% center;
   }
-
-  &:not(&.toolUseComplete):global(.tool-user_choice_tool) {
-    .toolUseIcon {
-      --leo-icon-color: var(--leo-color-primary-5);
-      background-color: var(--leo-color-primary-40);
-      border-color: var(--leo-color-primary-50);
-    }
-  }
-
-  &.toolUseComplete {
-    color: var(--leo-color-text-tertiary);
-
-    .toolUseIcon {
-      --leo-icon-color: var(--leo-color-systemfeedback-success-icon);
-      background: var(--leo-color-systemfeedback-success-background);
-      border-color: var(--leo-color-green-20);
-    }
+  100% {
+    background-position: -200% center;
   }
 }
 
-.toolUseContentPreview {
-  margin: 4px;
-  max-width: 80vw;
+.toolUse {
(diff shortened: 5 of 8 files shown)


components/resources/ai_chat_ui_strings.grdp
```diff
@@ -744,4 +744,10 @@
   <message name="IDS_CHAT_UI_TOOL_LABEL_NAVIGATE_WEB_PAGE_WITH_INPUT" desc="Label for the navigate web page tool with a known URL destination" formatter_data="webui=AiChat">
     Navigated the web page to <ph name="URL">$1<ex>https://www.example.com</ex></ph>
   </message>
+  <message name="IDS_AI_CHAT_TASK_PROGRESS_LABEL" desc="Label for the task progress tab in the assistant response" formatter_data="webui=AiChat">
+    Progress
+  </message>
+  <message name="IDS_AI_CHAT_TASK_STEPS_LABEL" desc="Label for the task steps tab in the assistant response" formatter_data="webui=AiChat">
+    Steps
+  </message>
```
petemill, line 755: Tab titles for this:

Image

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/assistant_task.tsx
```diff
@@ -0,0 +1,179 @@
...
+        <TabItem value='progress'>
+          {getLocale(S.AI_CHAT_TASK_PROGRESS_LABEL)}
+        </TabItem>
+        <TabItem value='steps'>{getLocale(S.AI_CHAT_TASK_STEPS_LABEL)}</TabItem>
+      </Tabs>
+      <div className={styles.taskContent}>
+        <div className={styles.taskData}>
+          {showSteps && (
+            <Steps
+              {...props}
```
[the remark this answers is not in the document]
petemill: internal shared props

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/assistant_task.tsx
```diff
@@ -0,0 +1,179 @@
...
+              taskData={taskData}
+            />
+          )}
+
+          {!showSteps && (
+            <Progress
+              {...props}
```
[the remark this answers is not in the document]
petemill: internal shared props

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/assistant_task.tsx
```diff
@@ -4,13 +4,178 @@
...
+ */
 export default function AssistantTask(props: Props) {
-  return <h1>Task stub</h1>
+  const [showSteps, setShowSteps] = React.useState(false)
+
+  const taskData = useExtractTaskData(props.assistantEntries)
+
+  return (
+    <div className={styles.task}>
+      <Tabs
```
petemill, line 50: Spacing is not quite right for inside the Tabs component and should probably be solved in `@brave/leo/Tabs`:
- https://github.com/brave/leo/issues/1238

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/assistant_task.tsx
```diff
@@ -4,13 +4,177 @@
...
 
+/**
+ * The Task is split *not* by ConversationTurn but by CompletionEvent, which is
+ * better for descriptive UI.
+ * The Progress tab displays the most recent item in that split (the
+ * completion text and the subsequent tool uses) as well as any previous
+ * "important" tool use (e.g. TODO list or tab navigation).
+ * The Steps tab displays everything in timeline order.
+ */
 export default function AssistantTask(props: Props) {
```
petemill, line 43: @fallaciousreasoning Whilst I'm testing all the data for this component and its 2 children via `use_extract_task_data.test.ts`, I'm not testing anything else in the rendering. If you see something worth testing, let me know. I couldn't really see anything aside from maybe the classnames or icon in `<Steps>`, but that's relatively simple logic.

components/test/testPolyfills.ts
```diff
@@ -21,3 +21,11 @@ const windowAsAny = window as any
 
 windowAsAny.chrome = getMockChrome()
 windowAsAny.loadTimeData = getMockLoadTimeData()
+
+windowAsAny.ResizeObserver = class ResizeObserverPolyfill {
```
petemill, line 25: This polyfill is needed for the Tabs component. I guess we're not previously using it anywhere in this repo, or at least where it gets rendered by a test.

fallaciousreasoning: Alternatively, you can import components/test/disable_custom_elements.ts which will stop the code inside the customelements running - it makes a lot of testing simpler because we generally don't care what the customElement is doing (just whats happening in the light-dom)

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/use_extract_task_data.ts
```diff
@@ -0,0 +1,71 @@
...
+        }
+      }
+    }
+
+    // Filter out empty task items since we start with an empty one and we might
+    // have encountered a completion event first, which starts a new task item.
+    taskItems = taskItems.filter((taskItem) => taskItem.length > 0)
+
+    return {
+      taskItems,
```
fallaciousreasoning: optional nit: merge these and make `taskItems` `const`:


```suggestion
    return {
      // Filter out empty task items since we start with an empty one and we might
      // have encountered a completion event first, which starts a new task item.
      taskItems: taskItems.filter((taskItem) => taskItem.length > 0),
```

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/use_extract_task_data.ts
```diff
@@ -0,0 +1,71 @@
...
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import * as React from 'react'
+import * as Mojom from '../../../common/mojom'
+
+// TODO(https://github.com/brave/brave-browser/issues/48535):
+// Should be Navigate and TODO tools.
+const importantToolNames = [Mojom.NAVIGATE_TOOL_NAME]
```
fallaciousreasoning: Do you mind adding a comment explaining what makes a tool important & why we need to keep track of them?

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/use_extract_task_data.ts
```diff
@@ -0,0 +1,71 @@
...
+    const importantToolUseEvents: Partial<
+      Record<(typeof importantToolNames)[number], Mojom.ToolUseEvent>
+    > = {}
+
+    for (const event of assistantEntryGroup.flatMap(
+      (entry) => entry.events ?? [],
+    )) {
+      if (!event) {
+        continue
+      }
```
fallaciousreasoning: can we actually have falsey events in the events array?

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/use_extract_task_data.ts
```diff
@@ -0,0 +1,71 @@
...
+      if (!!event.completionEvent) {
```
fallaciousreasoning: mega nit - no need to coeerce as this is an if statement so we only care about the truthiness, we don't need it to be a boolean
```suggestion
      if (event.completionEvent) {
```

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/use_extract_task_data.ts
```diff
@@ -0,0 +1,71 @@
...
+        // Start a new task item for every completion event
+        taskItems.push([event])
+      } else {
+        // Add any other event types to the last task item
+        taskItems[taskItems.length - 1].push(event)
```
fallaciousreasoning:
```suggestion
        taskItems.at(-1).push(event)
```

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/use_extract_task_data.ts
```diff
@@ -0,0 +1,71 @@
...
+        ) {
+          importantToolUseEvents[event.toolUseEvent.toolName] =
+            event.toolUseEvent
+        }
+      }
+    }
+
+    // Filter out empty task items since we start with an empty one and we might
+    // have encountered a completion event first, which starts a new task item.
+    taskItems = taskItems.filter((taskItem) => taskItem.length > 0)
```
fallaciousreasoning: optional, but alternatively we could have an 
```ts
if (!taskItems.at(-1)) {
    taskItems.push([])
}
```

to the else branch above. Might keep the logic a bit closer together.

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/use_extract_task_data.ts
```diff
@@ -0,0 +1,71 @@
...
+  return React.useMemo(() => {
+    // Individual tasks, split by CompletionEvent
+    let taskItems: Mojom.ConversationEntryEvent[][] = [[]]
+    // All completion events are allowed the links provided by the whole response
+    // group.
+    const allowedLinks: string[] = []
+    // Most recent of each type of important tool use event
+    const importantToolUseEvents: Partial<
+      Record<(typeof importantToolNames)[number], Mojom.ToolUseEvent>
+    > = {}
```
fallaciousreasoning: Hey does this work? `importantToolNames` isn't declared as `as const` so I supect this is a record of `<string, Mojom.ToolUseEvent>`

Probably we want an `as const` on the declaration

bsclifton (approved):
string reviewers ++

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/use_extract_task_data.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    // Filter out empty task items since we start with an empty one and we might
+    // have encountered a completion event first, which starts a new task item.
+    taskItems = taskItems.filter((taskItem) => taskItem.length > 0)
+
+    return {
+      taskItems,
+      allowedLinks,
+      importantToolUseEvents: Object.values(importantToolUseEvents).filter(
+        (event) => !!event,
```
fallaciousreasoning, line 68: Looking at the code above, we only insert non-falsey events, so do we need this?

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/use_extract_task_data.ts
fallaciousreasoning: alternatively, I guess this could be a `Record<string, Mojom.ToolUseEvent>` or `{ [key: string]: Mojom.ToolUseEvent }` if the as const doesn't work

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/use_extract_task_data.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    for (const event of assistantEntryGroup.flatMap(
+      (entry) => entry.events ?? [],
+    )) {
+      if (!event) {
+        continue
+      }
+      if (event.completionEvent) {
+        // Start a new task item for every completion event
+        taskItems.push([event])
```
fallaciousreasoning, line 39: Hey sorry, does this mean completion events will be grouped with the next task?

`[1, 2, 'complete', 3, 4, 'complete']`
will result in
```
[1, 2]
['complete', 3, 4]
['complete']
```

Is that intentional? I (possibly naively) thought the completion should be for the tasks before:
```
[1, 2, 'complete']
[3, 4, 'complete']
```

I guess the above would be working if we were iterating in reverse order?

fallaciousreasoning: That said, I'm pretty sure I'm just misunderstanding `completetionEvent`

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/use_extract_task_data.test.ts
```diff
@@ -0,0 +1,345 @@
...
+        createConversationTurnWithDefaults({
+          characterType: Mojom.CharacterType.ASSISTANT,
+          events: [
+            getToolUseEvent({
+              toolName: Mojom.NAVIGATE_TOOL_NAME,
+              id: '2',
+              argumentsJson: '{"url": "https://second.com"}',
+              output: undefined,
+            }),
+            getCompletionEvent('Second navigation'),
```
fallaciousreasoning, line 179: sorry, slightly confused here - why does the completion event happen before the first navigation, but after the second/third? Is it not deterministic?

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/use_extract_task_data.test.ts
```diff
@@ -0,0 +1,345 @@
...
+
+      const { result } = renderHook(() => useExtractTaskData(assistantEntries))
+
+      expect(result.current.importantToolUseEvents.length).toEqual(1)
+      expect(
+        result.current.importantToolUseEvents.filter(
+          (t) => t.toolName === Mojom.NAVIGATE_TOOL_NAME,
+        ),
+      ).toHaveLength(1)
+    })
```
fallaciousreasoning: I think we should just remove this until we get another tool

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/use_extract_task_data.test.ts
```diff
@@ -0,0 +1,345 @@
...
+      ]
+
+      const { result } = renderHook(() => useExtractTaskData(assistantEntries))
+
+      // Should only have 1 navigate tool event (the most recent one)
+      expect(result.current.importantToolUseEvents).toHaveLength(1)
+      expect(result.current.importantToolUseEvents[0].id).toBe('3')
+    })
+
+    it('should not include non-important tools', () => {
```
fallaciousreasoning, line 203: can we have a test mixing important/non-important

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/use_extract_task_data.test.ts
```diff
@@ -0,0 +1,345 @@
...
+              {
+                url: { url: 'https://example2.com' },
+                title: 'Example 2',
+                faviconUrl: { url: 'https://example2.com/favicon.ico' },
+              },
+              {
+                url: { url: 'https://example3.com' },
+                title: 'Example 3',
+                faviconUrl: { url: 'https://example3.com/favicon.ico' },
+              },
```
fallaciousreasoning, line 294: could we test some duplicate links across sources?

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/use_extract_task_data.test.ts
```diff
@@ -0,0 +1,345 @@
...
+
+      const { result } = renderHook(() => useExtractTaskData(assistantEntries))
+
+      expect(result.current.allowedLinks).toHaveLength(3)
+      expect(result.current.allowedLinks).toContain('https://example1.com')
+      expect(result.current.allowedLinks).toContain('https://example2.com')
+      expect(result.current.allowedLinks).toContain('https://example3.com')
+    })
+
+    it('should handle entries with no sources events', () => {
```
fallaciousreasoning, line 309: also no events at all, maybe?

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/use_extract_task_data.test.ts
fallaciousreasoning, line 1: cool! Didn't think of testing hooks on their own like this but it makes a lot of sense!

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/assistant_task.module.scss
```diff
@@ -0,0 +1,82 @@
...
+  color: var(--leo-color-text-primary);
+  position: relative;
+
+  .taskIcon {
+    --border-size: 1px;
+    --leo-icon-size: 16px;
+    --leo-progressring-size: 16px;
+
+    position: absolute;
+    top: calc(0.5lh - var(--half-task-icon-size));
```
fallaciousreasoning, line 50: wow, cool!

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/assistant_task.module.scss
```diff
@@ -0,0 +1,82 @@
...
+    left: calc(var(--half-task-icon-size) * -1);
+    width: calc(var(--task-icon-size) - calc(2 * var(--border-size)));
+    height: calc(var(--task-icon-size) - calc(2 * var(--border-size)));
```
fallaciousreasoning: just curious but any reason we can't use an outline here instead, so we don't need to worry about these size calculations?

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/assistant_task.stories.tsx
```diff
@@ -5,16 +5,36 @@
 
 import * as React from 'react'
 import { Meta } from '@storybook/react'
+import { InferControlsFromArgs } from '../../../../../../.storybook/utils'
```
fallaciousreasoning, line 8: weird casing on this method

fallaciousreasoning: I guess not for this PR though :shrug:

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/assistant_task.tsx
```diff
@@ -4,13 +4,177 @@
...
+  // events)
+  const lastTaskItem = props.taskData.taskItems.at(-1)
+
+  if (!lastTaskItem || lastTaskItem.length === 0) {
+    return null
+  }
+
+  const currentCompletionEvent = lastTaskItem.find(
+    (event) => !!event.completionEvent,
+  )
```
fallaciousreasoning: mega nit:

```suggestion
  const currentCompletionEvent = lastTaskItem.find(
    (event) => event.completionEvent,
  )
```

fallaciousreasoning: At the moment, this can only ever be the first item in the taskItem - should we check that instead?

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/assistant_task.tsx
```diff
@@ -4,13 +4,177 @@
...
+
+  // Include any current tool use events that are not already included in the
+  // "important" section.
+  const currentToolUseEvents = lastTaskItem.filter(
+    (event) =>
+      !!event.toolUseEvent
```
fallaciousreasoning:
```suggestion
      event.toolUseEvent
```

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/assistant_task.tsx
```diff
@@ -4,13 +4,177 @@
...
+    const isRunnable =
+      props.isActiveTask && index === props.taskData.taskItems.length - 1
+
+    // Are we generating this task item?
+    const isActive = isRunnable && props.isGenerating && props.isActiveTask
+
+    // Are we waiting for this task item to be completed (shouldn't show as complete)
+    const isPending =
+      isRunnable
+      && !isActive
```
fallaciousreasoning, line 153: is a task thats generating not pending?

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/assistant_task.tsx
```diff
@@ -4,13 +4,177 @@
...
+      && taskItem.some(
+        (event) => !!event.toolUseEvent && !event.toolUseEvent.output,
```
fallaciousreasoning: mega nit
```suggestion
        (event) => event.toolUseEvent && !event.toolUseEvent.output,
```

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/assistant_task.tsx
fallaciousreasoning: this seems fine to me

fallaciousreasoning (commented):
this looks great!

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/use_extract_task_data.ts
petemill: The name `CompletionEvent` is probably confusing and is a holdover from when the APIs were not conversational - it's a _text_ completion, i.e. it did what LLMs do: completed the submitted text with a response.

I guess we should just rename it something like `TextEvent` because that's what it's used for - just a regular text part of a response.

So yes it is correct and the tests confirm any CompletionEvent should start a new task item because it's the LLM text that we want to start new items with. They are the "thoughts" that we want to highlight, and the events associated with that thought would come after. Event if it's a reaction to the events from the previous task item, the design is that we start a new task item with each piece of text.

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/use_extract_task_data.test.ts
petemill: These are just examples, e.g.

- "I am going to navigate to site 1"
- Navigate: site 1
- Navigate: site 2
- "It didn't work navigating to site 1 so I navigated to site 2, which worked"

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/use_extract_task_data.ts
petemill: - https://github.com/brave/brave-browser/issues/50184

components/test/testPolyfills.ts
petemill: Something already in `conversation_entries/index.test.ts` seems to rely on the nala code running. 

```
 ● ConversationEntries visualContentUsedPercentage handling › should show text content warning when visualContentUsedPercentage is fine but trimmed tokens exist

    expect(received).toBeInTheDocument()

    received value must be an HTMLElement or an SVGElement.
    Received has value: null

      473 |     // Should show text content warning due to trimmed tokens (highest priority)
      474 |     const infoElement = container.querySelector('leo-icon[name="info-outline"]')
    > 475 |     expect(infoElement).toBeInTheDocument()
```

petemill: Seems fine to have the "polyfill" because the environment should at least know about that API anyway

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/use_extract_task_data.ts
petemill: I have elaborated and made it clear it could change after collecting feedback:

```
// Important tools are ones worth always showing the latest version of
// to the user because they convey a longer-term status of the task compared
// to other tools. For example a navigation in a web page vs a click, or an
// up to date TODO list of items in the current task.
// TODO(https://github.com/brave/brave-browser/issues/48535):
// Should be Navigate and TODO tools.
// Note: Navigate could be removed if we display the current Url of a related
// content tab in the Task UI instead (since history navigation affects the
// "current" URL and not just the most recent navigation tool. This is all
// experimental.
```

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/use_extract_task_data.ts
petemill: I guess not. I thought something was complaining but will try

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/use_extract_task_data.ts
petemill: It has to be `taskItems.at(-1)?.push(event)`

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/use_extract_task_data.ts
petemill: The `as const` causes issues when using it for `importantToolNames.includes(event.toolUseEvent.toolName)`. You're right in that it's just treating it as a string, so I could just be explicit about that. It's not necessary to type the key, just to know it is a valid tool name, which is a string.

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/use_extract_task_data.ts
petemill: It's to satisfy typescript otherwise it defines this property as ` importantToolUseEvents: (Mojom.ToolUseEvent | undefined)[]`

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/use_extract_task_data.test.ts
petemill: We do have that already with `'should extract an important tool'` or were you thinking to have some different kind of mix?

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/assistant_task.tsx
petemill: `isPending` is only `true` when `props.isGenerating` is false because `isPending` controls some UI state and not the event interactivity itself. 
We have `active` for "is working on it (receiving input or running the tool automatically)" , `pending` for "is waiting for something (user input probably)" and `complete` for "is a past event".

So when the task is generating it will look like:
image

And when it is not generating but is the active task:
image

And when it is no longer the active task:
image

You can play around with this on the Story for this component. Happy to take naming suggestions.

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/assistant_task.tsx
petemill: done 246a466

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/assistant_task.tsx
petemill: done

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/assistant_task.tsx
petemill: done 246a466

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/use_extract_task_data.test.ts
petemill: added 246a466

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/use_extract_task_data.test.ts
petemill: added 246a466

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/use_extract_task_data.test.ts
petemill: removed for the other PR 246a466

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/use_extract_task_data.ts
petemill: changed in 246a466

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/use_extract_task_data.ts
petemill: done in 246a466

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/use_extract_task_data.ts
petemill: which I've added in 246a466

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/use_extract_task_data.ts
petemill: done in 246a466

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/use_extract_task_data.ts
petemill: removed in 246a466

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/use_extract_task_data.ts
petemill: done in 246a466

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/assistant_task.module.scss
petemill: That's a good point!
At first I thought you were talking about the steps vertical line and I was thinking an outline can't be on a single side, but I could use a box-shadow. For that, though, there's not much benefit and I think the margin and padding spacing are meant to be on either side of the line, as happens by default with border.
But for the icon - yes that makes sense and I made that change in bc2f6dd7e54446dd7e94ab1b6173ca4665c86910. Not sure it's fewer calculations - it might be more actually but it feels more definitely centered on the line?

petemill: ...and the outline-offset a07e5e864cf25886eaa9f39f28167054ba979bb0

petemill:
I rebased on master (new chromium update) and so invalidated the refs I added to review comments. But the feedback fixes are still split in to separate commits.

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/use_extract_task_data.ts
```diff
@@ -0,0 +1,74 @@
...
+    )) {
+      if (event.completionEvent) {
+        // Start a new task item for every completion event
+        taskItems.push([event])
+      } else {
+        if (!taskItems.at(-1)) {
+          taskItems.push([])
+        }
+        // Add any other event types to the last task item
+        taskItems.at(-1)?.push(event)
```
fallaciousreasoning, line 45: nit: should be `!` I think
```suggestion
        taskItems.at(-1)!.push(event)
```

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/assistant_task.stories.tsx
```diff
@@ -5,16 +5,36 @@
...
 export const _AssistantTask = {
-  render: () => {
-    return <AssistantTask assistantEntries={taskConversationEntries} />
+  render: (args: CustomArgs) => {
+    return (
+      <AssistantTask
+        assistantEntries={taskConversationEntries}
+        isActiveTask={args.isActiveTask}
+        isGenerating={args.isGenerating}
+        isLeoModel={true}
```
fallaciousreasoning, line 29: mega nit:
```suggestion
        isLeoModel
```

PR #31558, opened 2025-10-01
[CodeHealth] Replace ternary ops with logical expressions

cdesouza-chromium:
This PR is part of a couple of changes correcting these redundant
ternary operations:

```
x ? y : false -> x && y
x ? false : y -> !x && y
x ? y : true - > !x || y
x ? true : y -> x || y
```

This PR mirrors upstream work being done in the same direction:
https://issues.chromium.org/issues/416294715

Bug: https://github.com/brave/brave-browser/issues/49819

### DIFF
--- a/ios/brave-ios/Sources/Brave/Frontend/UserContent/UserScripts/AllFrames/AtDocumentStart/FullscreenHelper.js
+++ b/ios/brave-ios/Sources/Brave/Frontend/UserContent/UserScripts/AllFrames/AtDocumentStart/FullscreenHelper.js
@@ -3,11 +3,10 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this file,
 // You can obtain one at https://mozilla.org/MPL/2.0/.
 
-window.__firefox__.includeOnce("FullscreenHelper", function($) {
+window.__firefox__.includeOnce('FullscreenHelper', function($) {
   let isFullscreenSupportedNatively = document.fullscreenEnabled ||
-                                      document.webkitFullscreenEnabled ||
-                                      document.mozFullScreenEnabled ||
-                                      document.msFullscreenEnabled ? true : false;
+      document.webkitFullscreenEnabled || document.mozFullScreenEnabled ||
+      document.msFullscreenEnabled;
 
   let videosSupportFullscreen = HTMLVideoElement.prototype.webkitEnterFullscreen !== undefined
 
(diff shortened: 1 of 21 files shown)


AlexeyBarabash:
for me the version `before` is more clear than `after`

cdesouza-chromium:
> for me the version `before` is more clear than `after`

This is also being done in upstream https://issues.chromium.org/issues/416294715. The thing is that using ternary for what are simple && operations can be error prone, and a bit harder for readability.

darkdh (approved):

kylehickinson (approved):
ios++

AlexeyBarabash (approved):

nuo-xu (approved):

test/data/speedreader/rewriter/pages/news_pages/nypost.com/original.html
```diff
@@ -120,7 +120,7 @@
 				// Check if we have a cookie.
 				if ( GACookie.length ) {
 					// If the cookie value is not empty we have a GA cookie.
-					hasGACookie = GACookie[0].trim().length > 4 ? true : false;
+					hasGACookie = GACookie[0].trim().length > 4;
```
boocmp, line 123: Not sure why it's required for test data. In this case it doesn't matter because speedreader ignores any js code.

ios/brave-ios/Sources/Brave/Frontend/UserContent/UserScripts/AllFrames/AtDocumentStart/FullscreenHelper.js
```diff
@@ -3,11 +3,10 @@
...
 // You can obtain one at https://mozilla.org/MPL/2.0/.
 
-window.__firefox__.includeOnce("FullscreenHelper", function($) {
+window.__firefox__.includeOnce('FullscreenHelper', function($) {
   let isFullscreenSupportedNatively = document.fullscreenEnabled ||
-                                      document.webkitFullscreenEnabled ||
-                                      document.mozFullScreenEnabled ||
-                                      document.msFullscreenEnabled ? true : false;
+      document.webkitFullscreenEnabled || document.mozFullScreenEnabled ||
+      document.msFullscreenEnabled;
```
supermassive, line 9: This would strictly match previous behavior
```
let isFullscreenSupportedNatively = !!(
  document.fullscreenEnabled ||
  document.webkitFullscreenEnabled || 
  document.mozFullScreenEnabled ||
  document.msFullscreenEnabled)
```

PR #31565, opened 2025-10-01
codeowners devops

mihaiplesa:
Resolves wrong owners on GitHub Actions.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,3 +1,11 @@
+# GitHub configuration
+/.github/CODEOWNERS @bridiver @bsclifton
+/.github/PULL_REQUEST_TEMPLATE.md @bsclifton
+/.github/ @brave/devops-team
+
+# Semgrep configuration
+/.semgrepignore @thypon
+
 build/commands/ @brave/build-commands-reviewers
 patches/ @brave/patch-reviewers
 chromium_src/ @brave/chromium-src-reviewers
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -1,3 +1,11 @@
+# GitHub configuration
```
bridiver: why move these to the top? If anything they should move down below line 211?

mihaiplesa: We don't want them in alphabetic/appearance in file tree order?

bridiver: no, the last matching rule applies so order is significant

bridiver: see https://github.com/brave/brave-core/pull/31565/files#diff-3d36a1bf06148bc6ba1ce2ed3d19de32ea708d955fed212c0d27c536f0bd4da7R209

mihaiplesa: Sorry, I meant as paths in the whole CODEOWNERS file not order of lines for .github ownership, that is already in the right order with the specific files first then the remaining owned by DevOps team.

.github/CODEOWNERS
```diff
@@ -1,3 +1,11 @@
...
+/.github/CODEOWNERS @bridiver @bsclifton
+/.github/PULL_REQUEST_TEMPLATE.md @bsclifton
```
wknapik: I'd suggest adding devops-team here as well

.github/CODEOWNERS
mihaiplesa: Moved back.

.github/CODEOWNERS
mihaiplesa: That's covered already by the last rule.

.github/CODEOWNERS
```diff
@@ -194,10 +194,9 @@ android/java/res @brave/Android
 android/java/org/chromium/chrome/browser/preferences/BravePrefServiceBridge.java @AlexeyBarabash @samartnik @SergeyZhukovsky
 
 # GitHub configuration
-/.github/ @bridiver @cdesouza-chromium
 /.github/CODEOWNERS @bridiver @bsclifton
 /.github/PULL_REQUEST_TEMPLATE.md @bsclifton
-
+/.github/ @brave/devops-team
```
bridiver: this still has to go above CODEOWNERS or it will take precedence over the other two. As the first entry it will match anything except `CODEOWNERS` or `PULL_REQUEST_TEMPLATE.md`

PR #31568, opened 2025-10-01
[ollama] backend

thypon:
- Settings Integration: Adds kBraveAIChatOllamaSyncEnabled preference to allow/disallow Ollama sync
  - Ollama Client: HTTP client connecting to localhost:11434 to check Ollama availability and fetch model
  list
  - Model Sync Service: Automatically syncs local Ollama models to Brave's ModelService:
    - Adds newly detected models
    - Removes obsolete models no longer in Ollama
    - Formats model names for UI display (e.g., "llama2:7b" → "Llama 2 7B")
    - Estimates context sizes and content limits per model
  - Build System: New BUILD.gn target for Ollama components with unit tests
  - Tests: Comprehensive unit tests for client, sync logic, and name formatting


thypon:
errors in CI seems unrelated to me

components/ai_chat/core/browser/ollama/ollama_model_sync.cc
```diff
@@ -0,0 +1,385 @@
...
+                          weak_ptr_factory_.GetWeakPtr()));
+}
+
+OllamaModelSync::~OllamaModelSync() = default;
+
+void OllamaModelSync::OnOllamaSyncEnabledChanged() {
+  bool ollama_sync_enabled =
+      prefs_->GetBoolean(prefs::kBraveAIChatOllamaSyncEnabled);
+
+  DVLOG(1) << "Ollama sync preference changed to: " << ollama_sync_enabled;
```
bridiver: This is an excessive amount of logging to have at 1. It's also an unusually large amount of logging in general and we typically don't have DVLOG in nearly every method like this. Is all of this really necessary?

bridiver: I would say most of this is at least 2

components/ai_chat/core/browser/ollama/ollama_model_sync.h
```diff
@@ -0,0 +1,89 @@
...
+#include "services/network/public/cpp/shared_url_loader_factory.h"
+
+namespace ai_chat {
+
+class ModelService;
+class OllamaClient;
+
+// Manages synchronization of models between Ollama and the AI Chat
+// ModelService.
+class OllamaModelSync {
```
bridiver: OllamaModelFetcher would be more typical naming

test/BUILD.gn
```diff
@@ -195,6 +195,7 @@ test("brave_unit_tests") {
     "//brave/chromium_src/components/privacy_sandbox:unit_tests",
     "//brave/chromium_src/components/webui/flags:unit_tests",
     "//brave/chromium_src/net/base:unit_tests",
+    "//brave/components/ai_chat/core/browser/ollama:unit_tests",
```
bridiver: this belongs in brave/components/BUILD.gn

components/ai_chat/core/browser/ollama/ollama_model_sync.cc
thypon: removed them, needed for debugging only

components/ai_chat/core/browser/ollama/ollama_model_sync.h
thypon: Done

test/BUILD.gn
thypon: Done

components/ai_chat/core/browser/ollama/ollama_model_fetcher.cc
```diff
@@ -0,0 +1,335 @@
...
+
+  ProcessModelsResponse(response_body);
+}
+
+void OllamaModelFetcher::ProcessModelsResponse(
+    const std::string& response_body) {
+  auto json_value = base::JSONReader::Read(response_body);
+  if (!json_value || !json_value->is_dict()) {
+    return;
+  }
```
yrliou: Just use base::JSONReader::ReadDict.

components/ai_chat/core/browser/ollama/ollama_client.cc
```diff
@@ -0,0 +1,131 @@
...
+      base::BindOnce(&OllamaClient::OnConnectionCheckComplete,
+                     weak_ptr_factory_.GetWeakPtr(), std::move(callback),
+                     std::move(loader)),
+      1024);  // Small response expected
+}
+
+void OllamaClient::OnConnectionCheckComplete(
+    ConnectionCallback callback,
+    std::unique_ptr<network::SimpleURLLoader> loader,
+    std::unique_ptr<std::string> response) {
```
yrliou: `std::optional<std::string>` instead https://source.chromium.org/chromium/chromium/src/+/main:services/network/public/cpp/simple_url_loader.h;l=113-116;drc=4d01ab4367797ae1dceb72d38f70b5d7c577e897
https://source.chromium.org/chromium/chromium/src/+/main:services/network/public/cpp/simple_url_loader.h;l=202-208;drc=4d01ab4367797ae1dceb72d38f70b5d7c577e897

components/ai_chat/core/browser/ollama/ollama_client.cc
```diff
@@ -0,0 +1,131 @@
...
+  ConnectionResult result;
+
+  if (response && response->find("Ollama is running") != std::string::npos &&
+      loader->ResponseInfo() && loader->ResponseInfo()->headers &&
+      loader->ResponseInfo()->headers->response_code() == 200) {
+    result.connected = true;
+  } else {
+    result.connected = false;
+    result.error = "Ollama is not running at localhost:11434";
```
yrliou: Is this error message really needed? It seems we only have two possibilities, either it's connected, or it's not.
It also doesn't seem like really an error to me, just not connected.

components/ai_chat/core/browser/ollama/ollama_client.cc
```diff
@@ -0,0 +1,131 @@
...
+      base::BindOnce(&OllamaClient::OnModelsListComplete,
+                     weak_ptr_factory_.GetWeakPtr(), std::move(callback),
+                     std::move(loader)),
+      1024 * 1024);  // 1MB should be enough for model list
+}
+
+void OllamaClient::OnModelsListComplete(
+    ModelsCallback callback,
+    std::unique_ptr<network::SimpleURLLoader> loader,
+    std::unique_ptr<std::string> response) {
```
yrliou: DITTO

components/ai_chat/core/browser/ollama/ollama_client.h
```diff
@@ -0,0 +1,62 @@
...
+// Handles network communication with a local Ollama instance.
+class OllamaClient {
+ public:
+  struct ConnectionResult {
+    bool connected = false;
+    std::string error;
+  };
+
+  using ConnectionCallback = base::OnceCallback<void(ConnectionResult)>;
+  using ModelsCallback = base::OnceCallback<void(std::string response_body)>;
```
yrliou: `std::optional<std::string>` to match the signature change of callback being passed to DownloadToString.

components/ai_chat/core/browser/ollama/ollama_model_fetcher.cc
```diff
@@ -0,0 +1,335 @@
...
+    custom_model->key = "";  // Empty for new models
+    custom_model->display_name = FormatOllamaModelName(*model_name);
+    custom_model->vision_support = vision_support;
+    custom_model->supports_tools = false;
+    custom_model->is_suggested_model = false;
+
+    auto custom_options = mojom::CustomModelOptions::New();
+    custom_options->model_request_name = *model_name;
+    custom_options->endpoint = GURL(mojom::kOllamaEndpoint);
+    custom_options->api_key = "ollama";
```
yrliou: Why do we need to set this?

components/ai_chat/core/browser/ollama/ollama_model_fetcher.cc
```diff
@@ -0,0 +1,335 @@
...
+    custom_options->context_size = context_size;
+
+    // Scale associated content and warning limits based on context size
+    uint32_t max_content =
+        std::min(static_cast<uint32_t>(context_size * 4), kMaxContentLength);
+    custom_options->max_associated_content_length = max_content;
+    custom_options->long_conversation_warning_character_limit = max_content;
```
yrliou: We shouldn't need to set this manually here, should follow the same flow for `AddCustomModel` triggered by current UI, which would set these when it calls `InitModels()` inside `AddCustomModel` and leads to `GetCustomModelsFromPrefs`, which will validate model options and set metrics for AI Chat content length warnings.

components/ai_chat/core/browser/ollama/ollama_model_fetcher.cc
```diff
@@ -0,0 +1,335 @@
...
+    RemoveModels();
+  }
+}
+
+void OllamaModelFetcher::FetchModels() {
+  ollama_client_->FetchModels(base::BindOnce(
+      &OllamaModelFetcher::OnModelsResponse, weak_ptr_factory_.GetWeakPtr()));
+}
+
+void OllamaModelFetcher::OnModelsResponse(std::string response_body) {
```
yrliou: Maybe better to move the response parsing to ollama client.
And in this class, rename this to OnModelsFetched which take an optional models_list and continue processing received mode_list here.

components/ai_chat/core/browser/ollama/ollama_model_fetcher.cc
```diff
@@ -0,0 +1,335 @@
...
+  if (response_body.empty()) {
+    return;
+  }
+
+  ProcessModelsResponse(response_body);
```
yrliou: I don't quite see why we want to separate logics into two functions here where this one only has a simple empty string check.

components/ai_chat/core/browser/ollama/ollama_model_fetcher.cc
```diff
@@ -0,0 +1,335 @@
...
+                mojom::kOllamaEndpoint &&
+            model->options->get_custom_model_options()->model_request_name ==
+                existing_model_name) {
+          // Convert full model list index to custom model list index
+          size_t custom_model_index = 0;
+          for (size_t j = 0; j < i; ++j) {
+            if (existing_models[j]->options &&
+                existing_models[j]->options->is_custom_model_options()) {
+              custom_model_index++;
+            }
```
yrliou: I feel weird that we have to find and calculate these indexes to actually remove a model, we could have method added as needed in model_service, maybe something that could directly find the target model in pref and remove it given the conditions.

components/ai_chat/core/browser/ollama/ollama_model_fetcher.cc
```diff
@@ -0,0 +1,335 @@
...
+    return;
+  }
+
+  const base::Value::Dict& json_dict = json_value->GetDict();
+  const base::Value::List* models_list = json_dict.FindList("models");
+  if (!models_list) {
+    return;
+  }
+
+  const auto& existing_models = model_service_->GetModels();
```
yrliou: We should just directly act on custom models instead of getting all models and check if it's custom model below.

components/ai_chat/core/browser/ollama/ollama_model_fetcher.cc
```diff
@@ -0,0 +1,335 @@
...
+      }
+      custom_model_index++;
+    }
+  }
+
+  // Remove Ollama models (in reverse order to maintain indices)
+  for (auto it = ollama_custom_indices.rbegin();
+       it != ollama_custom_indices.rend(); ++it) {
+    model_service_->DeleteCustomModel(*it);
+  }
```
yrliou: Again, these index calculating to use this index only API is confusing and error-prone, can't we just have a method that would remove all models with a specific endpoint from a pref exposed?

components/ai_chat/core/browser/ollama/ollama_model_fetcher.cc
```diff
@@ -0,0 +1,335 @@
...
+}
+
+// static
+std::string OllamaModelFetcher::FormatOllamaModelName(
```
yrliou: Is there really much benefit doing this? Personally I just put the same as the request name whenever I added an ollama model manually.

components/ai_chat/core/browser/ollama/ollama_model_fetcher.cc
```diff
@@ -0,0 +1,335 @@
...
+OllamaModelFetcher::~OllamaModelFetcher() = default;
+
+void OllamaModelFetcher::OnOllamaFetchEnabledChanged() {
+  bool ollama_fetch_enabled =
+      prefs_->GetBoolean(prefs::kBraveAIChatOllamaFetchEnabled);
+
+  if (ollama_fetch_enabled) {
+    FetchModels();
+  } else {
+    RemoveModels();
```
yrliou: Is it expected that when user disable this pref, any models with ollama endpoints would be deleted, even for those they added manually?

components/ai_chat/core/browser/ollama/ollama_client.cc
thypon: removed, not really. I liked the extra error

components/ai_chat/core/browser/ollama/ollama_client.cc
thypon: Done

components/ai_chat/core/browser/ollama/ollama_client.cc
thypon: Done

components/ai_chat/core/browser/ollama/ollama_client.h
thypon: Done

components/ai_chat/core/browser/ollama/ollama_model_fetcher.cc
thypon: yup, because they also become managed once they get added. There is no indication they are `ollama` models, if not that they have the same localhost endpoint. If we need to change this, we need to add a new attribute to models too.

components/ai_chat/core/browser/ollama/ollama_model_fetcher.cc
thypon: Set it as empty now, so it won't be added. It should still be required by mojom, I guess.

components/ai_chat/core/browser/ollama/ollama_model_fetcher.cc
thypon: I personally like to have them formatted as the other models pre-installed in brave, but it's a nice-to-have, not very required.

thypon: We can remove this if it's too much

thypon:
addressed part of @yrliou comments. Will do a second pass tomorrow!

components/ai_chat/core/browser/ollama/ollama_model_fetcher.cc
thypon: Removed. I noticed when I split the PR I lost `/api/show` capability too, so the creation was a bit bogus. I added that in the "yrliou fixes #2" commit

components/ai_chat/core/browser/ollama/ollama_model_fetcher.cc
thypon: Screenshot 2025-10-08 at 08 33 14

components/ai_chat/core/browser/ollama/ollama_model_fetcher.cc
thypon: fixed

components/ai_chat/core/browser/ollama/ollama_model_fetcher.cc
thypon: fixed https://github.com/brave/brave-core/pull/31568/commits/f530c3eb58d76887f21bc3510388e62700ef782e

components/ai_chat/core/browser/ollama/ollama_model_fetcher.cc
thypon: https://github.com/brave/brave-core/pull/31568/commits/d151a4bc8616f5db1aa176464e3526e3906fc2c6

components/ai_chat/core/browser/ollama/ollama_model_fetcher.cc
thypon: https://github.com/brave/brave-core/pull/31568/commits/d151a4bc8616f5db1aa176464e3526e3906fc2c6

components/ai_chat/core/browser/ollama/ollama_model_fetcher.cc
thypon: https://github.com/brave/brave-core/pull/31568/commits/d151a4bc8616f5db1aa176464e3526e3906fc2c6

components/ai_chat/core/browser/ollama/ollama_client.h
```diff
@@ -0,0 +1,102 @@
...
+namespace network {
+class SimpleURLLoader;
+class SharedURLLoaderFactory;
+}  // namespace network
+
+namespace ai_chat {
+
+// Handles network communication with a local Ollama instance.
+// Implements the mojom::OllamaService interface for UI communication.
+class OllamaClient : public KeyedService, public mojom::OllamaService {
```
yrliou: Should this class be named as OllamaService instead?

components/ai_chat/core/browser/model_service.cc
```diff
@@ -650,6 +650,49 @@ void ModelService::DeleteCustomModel(uint32_t index) {
...
+  // Remove models matching predicate (iterate in reverse to avoid index issues)
+  for (size_t i = custom_models_pref.size(); i > 0; --i) {
+    const base::Value::Dict& model_dict = custom_models_pref[i - 1].GetDict();
+
+    if (predicate.Run(model_dict)) {
+      std::string removed_key = *model_dict.FindString(kCustomModelItemKey);
+      removed_keys.push_back(removed_key);
+
+      // Check if this is the default model
+      if (current_default_key == removed_key) {
```
yrliou: `if (GetDefaultModelKey() == removed_key)` here, there's no need to maintain it in a variable.

components/ai_chat/core/browser/model_service.cc
```diff
@@ -650,6 +650,49 @@ void ModelService::DeleteCustomModel(uint32_t index) {
   }
 }
 
+void ModelService::DeleteCustomModelsIf(CustomModelPredicate predicate) {
+  base::Value::List custom_models_pref =
+      pref_service_->GetList(kCustomModelsList).Clone();
```
yrliou: Let's use ScopedListPrefUpdate instead.

components/ai_chat/core/browser/model_service_unittest.cc
```diff
@@ -385,4 +389,166 @@ TEST_F(ModelServiceTest, GetLeoModelKeyByName_And_GetLeoModelNameByKey) {
...
+    ASSERT_EQ(custom_models.size(), 1u);
+    custom_model_key = custom_models[0]->key;
+  }
+
+  // Set the custom model as default
+  GetService()->SetDefaultModelKey(custom_model_key);
+  EXPECT_EQ(GetService()->GetDefaultModelKey(), custom_model_key);
+
+  // Expect observer to be called when default model is removed
+  EXPECT_CALL(*observer_, OnDefaultModelChanged(custom_model_key, _)).Times(1);
```
yrliou: Instead of _, let's use chat-basic (features::kAIModelsDefaultKey.name set in this test).

components/ai_chat/core/browser/model_service.cc
```diff
@@ -650,6 +650,49 @@ void ModelService::DeleteCustomModel(uint32_t index) {
...
+
+  if (!removed_keys.empty()) {
+    pref_service_->SetList(kCustomModelsList, std::move(custom_models_pref));
+    InitModels();
+
+    for (const auto& removed_key : removed_keys) {
+      for (auto& obs : observers_) {
+        obs.OnModelRemoved(removed_key);
+      }
+    }
```
yrliou: These should be verified too in the unit test cases.

yrliou: And probably a good idea to verify OnModelListUpdated is called too (from InitModels).

components/ai_chat/core/browser/ollama/ollama_model_fetcher.cc
yrliou: It might be a surprise for users trying this out then have their previous added models removed when they decide to disable this, cc @aguscruiz @mattmcalister for UX inputs regarding how we want the flow to be.
If we do decide to remove them all when disabled, we would at least need to display some warning to users regarding this destructive action, though it won't help the fact that as long as I choose to enable this feature then disable, I'll get previous models I manually added being deleted, so maybe we need to tell them upon enabling? Not sure what's the best flow, I'll defer to @aguscruiz and @mattmcalister to decide.

components/ai_chat/core/common/mojom/ollama.mojom
```diff
@@ -0,0 +1,20 @@
...
+const string kOllamaApiTagsEndpoint = "http://localhost:11434/api/tags";
+const string kOllamaApiShowEndpoint = "http://localhost:11434/api/show";
+const string kOllamaEndpoint = "http://localhost:11434/v1/chat/completions";
+
+// Interface for communicating with a local Ollama instance.
+// This interface is implemented directly by OllamaClient and allows
+// the UI to check connection status.
+interface OllamaService {
+  // Check if Ollama is running at localhost:11434
+  Connected() => (bool connected);
```
yrliou: IsConnected is more clear.

components/ai_chat/core/browser/ollama/ollama_client.cc
```diff
@@ -0,0 +1,270 @@
...
+  auto loader = network::SimpleURLLoader::Create(std::move(request),
+                                                 kOllamaConnectionAnnotation);
+
+  auto* loader_ptr = loader.get();
+  loader_ptr->DownloadToString(
+      url_loader_factory_.get(),
+      base::BindOnce(&OllamaClient::OnConnectionCheckComplete,
+                     weak_ptr_factory_.GetWeakPtr(), std::move(callback),
+                     std::move(loader)),
+      1024);  // Small response expected
```
yrliou: Suggest to make this a constant with readable names.

components/ai_chat/core/browser/ollama/ollama_client.cc
```diff
@@ -0,0 +1,270 @@
...
+  auto loader = network::SimpleURLLoader::Create(std::move(request),
+                                                 kOllamaModelsAnnotation);
+
+  auto* loader_ptr = loader.get();
+  loader_ptr->DownloadToString(
+      url_loader_factory_.get(),
+      base::BindOnce(&OllamaClient::OnModelsListComplete,
+                     weak_ptr_factory_.GetWeakPtr(), std::move(callback),
+                     std::move(loader)),
+      1024 * 1024);  // 1MB should be enough for model list
```
yrliou: DITTO

components/ai_chat/core/browser/ollama/ollama_client.cc
```diff
@@ -0,0 +1,270 @@
...
+  std::string body = "{\"model\":\"" + model_name + "\"}";
+  loader->AttachStringForUpload(body, "application/json");
+
+  auto* loader_ptr = loader.get();
+  loader_ptr->DownloadToString(
+      url_loader_factory_.get(),
+      base::BindOnce(&OllamaClient::OnModelDetailsComplete,
+                     weak_ptr_factory_.GetWeakPtr(), std::move(callback),
+                     std::move(loader)),
+      1024 * 1024);  // 1MB should be enough for model details
```
yrliou: DITTO

components/ai_chat/core/common/mojom/ollama.mojom
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) 2025 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+module ai_chat.mojom;
+
+// Ollama local endpoints
+const string kOllamaBaseUrl = "http://localhost:11434";
+const string kOllamaApiTagsEndpoint = "http://localhost:11434/api/tags";
```
yrliou: Suggest to name it to something like kOllamaListModelsAPIEndpoint instead, easier for other devs to understand what this is without checking the API doc.

components/ai_chat/core/common/mojom/ollama.mojom
```diff
@@ -0,0 +1,20 @@
...
+const string kOllamaApiShowEndpoint = "http://localhost:11434/api/show";
```
yrliou: Also suggest to name this as kOllamaShowModelInfoAPIEndpoint or similar.

components/ai_chat/core/browser/ollama/ollama_client.cc
```diff
@@ -0,0 +1,270 @@
...
+                             ModelDetailsCallback callback) {
+  auto request = std::make_unique<network::ResourceRequest>();
+  request->url = GURL(mojom::kOllamaApiShowEndpoint);
+  request->method = "POST";
+  request->headers.SetHeader("Content-Type", "application/json");
+
+  auto loader = network::SimpleURLLoader::Create(std::move(request),
+                                                 kOllamaModelDetailsAnnotation);
+
+  std::string body = "{\"model\":\"" + model_name + "\"}";
```
yrliou: StrCat instead of +.

components/ai_chat/core/browser/ollama/ollama_client.cc
```diff
@@ -0,0 +1,270 @@
...
+  }
+  std::move(callback).Run(ParseModelsResponse(*response));
+}
+
+void OllamaClient::ShowModel(const std::string& model_name,
+                             ModelDetailsCallback callback) {
+  auto request = std::make_unique<network::ResourceRequest>();
+  request->url = GURL(mojom::kOllamaApiShowEndpoint);
+  request->method = "POST";
+  request->headers.SetHeader("Content-Type", "application/json");
```
yrliou: Seems redundant cuz we have AttachStringForUpload called below.

components/ai_chat/core/browser/ollama/ollama_service.h
```diff
@@ -0,0 +1,102 @@
...
+                            std::optional<std::string> response);
+
+  void OnModelDetailsComplete(ModelDetailsCallback callback,
+                              std::unique_ptr<network::SimpleURLLoader> loader,
+                              std::optional<std::string> response);
+
+  std::optional<std::vector<ModelInfo>> ParseModelsResponse(
+      const std::string& response_body);
+  std::optional<ModelDetails> ParseModelDetailsResponse(
+      const std::string& response_body);
```
yrliou: We should have unit tests for these two methods, for any conditions we handle in the parsing method.

components/ai_chat/core/browser/ollama/ollama_model_fetcher.cc
yrliou: I don't feel strong that this is that beneficial to introduce all these code.
User should be able to customize these names as they preferred (and other options too) as usual.
Could we at least separate this out to another PR if we really want to have this nice-to-have feature?

components/ai_chat/core/browser/ollama/ollama_model_fetcher_unittest.cc
```diff
@@ -0,0 +1,409 @@
...
+  base::test::TaskEnvironment task_environment_;
+  sync_preferences::TestingPrefServiceSyncable pref_service_;
+  std::unique_ptr<ModelService> model_service_;
+  network::TestURLLoaderFactory test_url_loader_factory_;
+  scoped_refptr<network::SharedURLLoaderFactory> shared_url_loader_factory_;
+  std::unique_ptr<OllamaModelFetcher> ollama_model_fetcher_;
+};
+
+TEST_F(OllamaModelFetcherTest, FetchModelsAddsNewModels) {
+  test_url_loader_factory()->AddResponse(ai_chat::mojom::kOllamaApiTagsEndpoint,
```
yrliou: Since it's a unit test for OllamaModelFetcher, I think it's better we mock ollama client and verify expect calls with mock responses rather than using the real ollama client and mock HTTP responses directly.

components/ai_chat/core/browser/ollama/ollama_model_fetcher.cc
thypon: https://github.com/brave/brave-core/pull/31568/commits/8ed196160bc4ca7c6b0d534b5405c59da8d30f99 reverts the change. https://github.com/brave/brave-core/pull/31568/commits/67bf516f9be2445d4fc9febcd64fd74dd655b785 readds the change.

Once we decided how we want to proceed, we can remove one or either of the two commits, or I can move to another branch.

@aguscruiz and @mattmcalister to page in on this one, as well

components/ai_chat/core/browser/ollama/ollama_client.h
thypon: Done in 64ca5057289b2a0b95b15868844857ce18d62a13

components/ai_chat/core/browser/model_service.cc
thypon: Done in d3668de4277fb6f80f672627e99da15dd1e065b8

components/ai_chat/core/browser/model_service.cc
thypon: Done in d3668de4277fb6f80f672627e99da15dd1e065b8

components/ai_chat/core/browser/model_service_unittest.cc
thypon: Done in 266d64d006bc42cba3d1044a3dd73bcb9f3bb0bb

components/ai_chat/core/common/mojom/ollama.mojom
thypon: Done in 64ca5057289b2a0b95b15868844857ce18d62a13

components/ai_chat/core/browser/ollama/ollama_client.cc
thypon: Done in 64ca5057289b2a0b95b15868844857ce18d62a13

components/ai_chat/core/browser/ollama/ollama_client.cc
thypon: Done in 64ca5057289b2a0b95b15868844857ce18d62a13

components/ai_chat/core/browser/ollama/ollama_client.cc
thypon: Done in 64ca5057289b2a0b95b15868844857ce18d62a13

components/ai_chat/core/common/mojom/ollama.mojom
thypon: Done in 64ca5057289b2a0b95b15868844857ce18d62a13

components/ai_chat/core/common/mojom/ollama.mojom
thypon: Done in 64ca5057289b2a0b95b15868844857ce18d62a13

components/ai_chat/core/browser/ollama/ollama_client.cc
thypon: Done in d3668de4277fb6f80f672627e99da15dd1e065b8

components/ai_chat/core/browser/ollama/ollama_client.cc
thypon: Done in d3668de4277fb6f80f672627e99da15dd1e065b8

components/ai_chat/core/browser/ollama/ollama_model_fetcher_unittest.cc
thypon: Done in 266d64d006bc42cba3d1044a3dd73bcb9f3bb0bb

components/ai_chat/core/browser/model_service.cc
thypon: Done in 266d64d006bc42cba3d1044a3dd73bcb9f3bb0bb

components/ai_chat/core/browser/ollama/ollama_service.h
thypon: Fixed in 0fa5a9e4bf6c3bee51b72f6e88be2457b5b7193e

components/ai_chat/core/browser/ollama/ollama_model_fetcher.cc
mattmcalister: we shouldn't auto-remove models in BYOM. if a model isn't serviceable then it should be unselectable or show a "can't connect" error message. or maybe it doesn't appear in the model-selector list but still exists in settings.

components/ai_chat/core/browser/ollama/ollama_model_fetcher.cc
aguscruiz: +1 to Jocelyn, I think having the ability to rename them and customize the options and system prompt, etc. would be needed for each model

components/ai_chat/content/browser/ollama_service_factory.h
```diff
@@ -0,0 +1,48 @@
...
+
+template <typename T>
+class NoDestructor;
+}  // namespace base
+
+namespace ai_chat {
+
+class OllamaService;
+
+class OllamaServiceFactory : public BrowserContextKeyedServiceFactory {
```
bridiver: this should be a ProfileKeyedServiceFactory

components/ai_chat/content/browser/ollama_service_factory.cc
```diff
@@ -0,0 +1,51 @@
...
+// static
+OllamaServiceFactory* OllamaServiceFactory::GetInstance() {
+  static base::NoDestructor<OllamaServiceFactory> instance;
+  return instance.get();
+}
+
+// static
+OllamaService* OllamaServiceFactory::GetForBrowserContext(
+    content::BrowserContext* context) {
+  DCHECK(context);
```
bridiver: this is not necessary and also see https://chromium.googlesource.com/chromium/src/+/HEAD/styleguide/c++/checks.md

components/ai_chat/content/browser/ollama_service_factory.cc
```diff
@@ -0,0 +1,51 @@
...
+  if (features::IsAIChatEnabled()) {
```
bridiver: please see https://github.com/brave/brave-core/blob/master/docs/keyed_services.md#keyed-services

only GetBrowserContextToUse should ever return nullptr

components/ai_chat/core/browser/model_service.cc
```diff
@@ -651,6 +652,43 @@ void ModelService::DeleteCustomModel(uint32_t index) {
   }
 }
 
+void ModelService::DeleteCustomModelsIf(CustomModelPredicate predicate) {
```
bridiver: typical convention here would be `MaybeDeleteCustomModels`

components/ai_chat/core/browser/ollama/ollama_service.cc
```diff
@@ -0,0 +1,275 @@
...
+                     std::move(loader)),
+      kConnectionCheckMaxSize);
+}
+
+void OllamaService::OnConnectionCheckComplete(
+    IsConnectedCallback callback,
+    std::unique_ptr<network::SimpleURLLoader> loader,
+    std::optional<std::string> response) {
+  bool connected = response &&
+                   response->find("Ollama is running") != std::string::npos &&
```
bridiver: This seems a bit hacky and maybe the text changes depending on locale? Is a 200 response code not sufficient?

components/ai_chat/core/browser/ollama/ollama_service.cc
```diff
@@ -0,0 +1,275 @@
...
+                   loader->ResponseInfo() && loader->ResponseInfo()->headers &&
+                   loader->ResponseInfo()->headers->response_code() == 200;
+
+  std::move(callback).Run(connected);
+}
+
+void OllamaService::FetchModels(ModelsCallback callback) {
+  auto request = std::make_unique<network::ResourceRequest>();
+  request->url = GURL(mojom::kOllamaListModelsAPIEndpoint);
+  request->method = "GET";
```
bridiver: `net::HttpRequestHeaders::kGetMethod`

components/ai_chat/core/browser/ollama/ollama_service.cc
```diff
@@ -0,0 +1,275 @@
...
+    std::optional<std::string> response) {
+  bool connected = response &&
+                   response->find("Ollama is running") != std::string::npos &&
+                   loader->ResponseInfo() && loader->ResponseInfo()->headers &&
+                   loader->ResponseInfo()->headers->response_code() == 200;
+
+  std::move(callback).Run(connected);
+}
+
+void OllamaService::FetchModels(ModelsCallback callback) {
```
bridiver: I think model fetching should moved to a separate PR and this should only include the base service with `IsConnected`. This PR is over 1000 changes which is quite large

components/ai_chat/content/browser/ollama_service_factory.h
thypon: This is not possible per include rules:

```
> git cl presubmit origin/master --force --upload
Running presubmit upload checks on branch features/ollama-auto-discover-backend ...
  checking for commit objects in tree took a long time: 4.1s
** Presubmit ERRORS: 1 **
You added one or more #includes that violate checkdeps rules.
  brave/components/ai_chat/content/browser/ollama_service_factory.cc
      Illegal include: "chrome/browser/profiles/profile_selections.h"
    Because of "-chrome" from brave's include_rules.
  brave/components/ai_chat/content/browser/ollama_service_factory.h
      Illegal include: "chrome/browser/profiles/profile_keyed_service_factory.h"
    Because of "-chrome" from brave's include_rules.

```

components/ai_chat/content/browser/ollama_service_factory.cc
thypon: https://github.com/brave/brave-core/pull/31568/commits/9b21a9d1df1141048c1d8e957b484889d77eec1e

components/ai_chat/core/browser/model_service.cc
thypon: https://github.com/brave/brave-core/pull/31568/commits/9b21a9d1df1141048c1d8e957b484889d77eec1e

components/ai_chat/core/browser/ollama/ollama_service.cc
thypon: https://github.com/brave/brave-core/pull/31568/commits/9b21a9d1df1141048c1d8e957b484889d77eec1e

components/ai_chat/core/browser/ollama/ollama_service.cc
thypon: https://github.com/brave/brave-core/pull/31568/commits/9b21a9d1df1141048c1d8e957b484889d77eec1e

components/ai_chat/core/browser/ollama/ollama_service.cc
thypon: I don't think we should escorporate the two http connections changes. IsConnected is quite small as it is.

components/ai_chat/content/browser/ollama_service_factory.h
thypon: https://github.com/brave/brave-core/pull/31568/commits/a05cf3ea45ff198018e1cc03c813cf24576d76f3

components/ai_chat/core/browser/ollama/ollama_service.cc
thypon: https://github.com/brave/brave-core/pull/31568/commits/fc387e2ee0c4404aa3c53c68887dc9d135a5d204 addressed, this PR is now 600 lines instead, the model format is still the majority of this code, and it's still 1200 lines long. This is due to the tests mainly, which are comprehensive

bridiver: the point of leaving IsConnected is exactly that it is small, but I'm fine moving that as well

browser/ai_chat/BUILD.gn
```diff
@@ -19,6 +19,8 @@ static_library("ai_chat") {
     "browser_tool_provider.h",
     "browser_tool_provider_factory.cc",
     "browser_tool_provider_factory.h",
+    "ollama_service_factory.cc",
```
bridiver: this should probably be under ollama directory (with separate BUILD.gn) to match the service. That creates fewer dependencies.

browser/ai_chat/ollama/ollama_service_factory.cc
```diff
@@ -0,0 +1,46 @@
...
+    content::BrowserContext* context) {
+  return static_cast<OllamaService*>(
+      GetInstance()->GetServiceForBrowserContext(context, true));
+}
+
+OllamaServiceFactory::OllamaServiceFactory()
+    : ProfileKeyedServiceFactory(
+          "OllamaServiceFactory",
+          ProfileSelections::Builder()
+              .WithRegular(ProfileSelection::kOriginalOnly)
```
bridiver: wasn't there also a feature check? You can still override `GetBrowserContextToUse`, call the superclass method to get the profile selection and then apply any additional checks

browser/ai_chat/ollama_service_factory.h
```diff
@@ -0,0 +1,48 @@
...
+class NoDestructor;
+}  // namespace base
+
+namespace ai_chat {
+
+class OllamaService;
+
+class OllamaServiceFactory : public ProfileKeyedServiceFactory {
+ public:
+  OllamaServiceFactory(const OllamaServiceFactory&) = delete;
```
bridiver: these are already deleted in the superclass

browser/ai_chat/ollama_service_factory.h
```diff
@@ -0,0 +1,48 @@
...
+  OllamaServiceFactory& operator=(const OllamaServiceFactory&) = delete;
+
+  static OllamaServiceFactory* GetInstance();
+  static OllamaService* GetForBrowserContext(content::BrowserContext* context);
```
bridiver: GetForProfile and Profile* since this is a ProfileKeyedSserviceFactory

components/ai_chat/core/browser/model_service.h
```diff
@@ -74,6 +74,14 @@ class ModelService : public KeyedService {
   void AddCustomModel(mojom::ModelPtr model);
   void SaveCustomModel(uint32_t index, mojom::ModelPtr model);
   void DeleteCustomModel(uint32_t index);
+
+  // Delete custom models matching a predicate
+  using CustomModelPredicate =
```
bridiver: These changes also belong in a separate PR. I would change the name of this PR to `Create Ollama Keyed Service` or similar

thypon: Removed in https://github.com/brave/brave-core/pull/31568/commits/495029362c4a6fd10193db03572e115f52afe7cd

browser/ai_chat/ollama/ollama_service_factory.cc
thypon: There was, but I guess we should have removed that given this comment https://github.com/brave/brave-core/blob/master/docs/keyed_services.md#discussion_r2452819793 . I'm still not sure if I did anything better with https://github.com/brave/brave-core/pull/31568/commits/4fce27ae7a09e979e851dcd122ea11ca47e9616f

browser/ai_chat/ollama_service_factory.h
thypon: fixed in https://github.com/brave/brave-core/pull/31568/commits/4fce27ae7a09e979e851dcd122ea11ca47e9616f

browser/ai_chat/ollama_service_factory.h
thypon: fixed in https://github.com/brave/brave-core/pull/31568/commits/4fce27ae7a09e979e851dcd122ea11ca47e9616f

browser/ai_chat/BUILD.gn
thypon: ~~not sure if worth it, as well it complicates this file more possibly https://github.com/brave/brave-core/pull/31568/commits/b04a7c0fa00872c1d8f0a3569d6285113169cf28~~

https://github.com/brave/brave-core/pull/31568/commits/2c64a5c9b8f427ebd5fc6c0b196c90e55ef8c8db

components/ai_chat/core/browser/ollama/ollama_service_unittest.cc
```diff
@@ -0,0 +1,111 @@
...
+ private:
+  base::test::TaskEnvironment task_environment_;
+  network::TestURLLoaderFactory test_url_loader_factory_;
+  scoped_refptr<network::SharedURLLoaderFactory> shared_url_loader_factory_;
+  std::unique_ptr<OllamaService> ollama_client_;
+};
+
+TEST_F(OllamaServiceTest, ConnectedSuccess) {
+  test_url_loader_factory()->AddResponse(ai_chat::mojom::kOllamaBaseUrl,
+                                         kOllamaSuccessResponse);
```
yrliou, line 57: Since we no longer check the exact string, maybe just put "TEST" here and remove `ConnectedWrongResponse` case below.

components/ai_chat/core/browser/ollama/ollama_model_fetcher.cc
bridiver: the ability to rename is fine, but that can be added later, no?

components/ai_chat/core/browser/ollama/ollama_model_fetcher.cc
bridiver: these could be quite large, no? That might be a problem on mobile if we don't provide some way to delete?

bridiver: please keep in mind that we can and should start with something simple and we can open follow-up issues

components/ai_chat/core/browser/ollama/ollama_service_unittest.cc
thypon: https://github.com/brave/brave-core/pull/31568/commits/baba38be59fbb96b5e796432943cf2173b90c9d9

browser/ai_chat/ollama/ollama_service_factory.cc
```diff
@@ -0,0 +1,50 @@
...
+              .Build()) {}
+
+OllamaServiceFactory::~OllamaServiceFactory() = default;
+
+std::unique_ptr<KeyedService>
+OllamaServiceFactory::BuildServiceInstanceForBrowserContext(
+    content::BrowserContext* context) const {
+  if (!features::IsAIChatEnabled()) {
```
bridiver: only `GetBrowserContextToUse` should return nullptr. To keep using ProfileSelections just override, check the feature and then fallback to superclass

bridiver: https://github.com/brave/brave-core/blob/master/docs/keyed_services.md

browser/ai_chat/ollama/ollama_service_factory.cc
```diff
@@ -0,0 +1,50 @@
...
+  return instance.get();
+}
+
+// static
+OllamaService* OllamaServiceFactory::GetForProfile(Profile* profile) {
+  return static_cast<OllamaService*>(
+      GetInstance()->GetServiceForBrowserContext(profile, true));
+}
+
+OllamaServiceFactory::OllamaServiceFactory()
```
bridiver, line 40: `OllamaServiceFactory::GetInstance()` should be added to `EnsureBrowserContextKeyedServiceFactoriesBuilt`. I'll update keyed services doc with this info

browser/ai_chat/ollama/ollama_service_factory.cc
thypon: https://github.com/brave/brave-core/pull/31568/commits/87c0d323b2e682c4195360b49822d5fcc7656e1a

browser/ai_chat/ollama/ollama_service_factory.cc
thypon: https://github.com/brave/brave-core/pull/31568/commits/53b3dc53a96aa6510e6c150c4983cade6ebecef1

browser/ai_chat/BUILD.gn
```diff
@@ -27,6 +27,8 @@ static_library("ai_chat") {
     "upload_file_helper.h",
   ]
 
+  public_deps = [ "//brave/browser/ai_chat/ollama" ]
```
bridiver: why is this a public dep?

bridiver: If the intention here is to only add a dep for `ai_chat`, that's the opposite of what we want. We want smaller more modularized targets with fewer dependencies. I think there should not be a dep here at all on ollama because nothing in this target uses it right now

thypon: fixed in https://github.com/brave/brave-core/pull/31568/commits/6129548100176288b9e3f7d03adbacddd1ffc362

PR #31569, opened 2025-10-01
[ollama] auto discover integration

thypon:
Bugfix conversatoin_handler.cc when the selected model is not existing anymore.
This caused a DCHECK crash without.

Readded automatic model fetcher sync

### DIFF
--- a/components/ai_chat/core/browser/ollama/ollama_model_fetcher.h
+++ b/components/ai_chat/core/browser/ollama/ollama_model_fetcher.h
@@ -92,7 +92,7 @@ class OllamaModelFetcher {
   // (https://github.com/brave/brave-core/commit/dadaf97432efd7d64289bd92b7d31cc6e16722f0)
   const raw_ref<ModelService> model_service_;
   raw_ptr<PrefService> prefs_;
-  raw_ptr<Delegate> delegate_;
+  raw_ptr<Delegate> delegate_ = nullptr;
   PrefChangeRegistrar pref_change_registrar_;
   std::map<std::string, PendingModelInfo> pending_models_;
   base::WeakPtrFactory<OllamaModelFetcher> weak_ptr_factory_{this};


browser/ai_chat/ai_chat_settings_helper.cc
```diff
@@ -227,6 +250,41 @@ void AIChatSettingsHelper::DeleteConversations(int time_period) {
                                browsing_data::CalculateEndDeleteTime(period));
 }
 
+void AIChatSettingsHelper::CheckOllamaConnection(
```
bridiver: I'm confused, in the other PR you have `OllamaClient::CheckConnection` which does the same thing. When splitting up PRs you should either change the base branch to build on each one or leave unimplemented methods at NOTREACHED. Also it looks like this class should probably be deprecated in general because we don't need a layer between the UI and whatever is implementing the mojo interface. In this case rather than add a new method to this existing interface, you should create a new interface implemented directly by OllamaClient

thypon: Is the suggestion to move this call in `ai_chat.mojom` or `settings_helper.mojom`, or should we create an `ollama.mojom` too?

bridiver: I would create a new mojom file for it, we don't want to dump everything in ai_chat.mojom

thypon:
ollama_deps

Added OllamaFetcher instantiation in OllamaServiceFactory, where we were already instantiating OllamaService mojom. OllamaFetcher needs OllamaService as a constructor parameter with DI

components/ai_chat/core/browser/ollama/ollama_service.h
```diff
@@ -64,6 +66,11 @@ class OllamaService : public KeyedService, public mojom::OllamaService {
   virtual void ShowModel(const std::string& model_name,
                          ModelDetailsCallback callback);
 
+  // Set the model fetcher that will be owned by this service.
+  // This should be called after construction when the service is created
+  // through the factory.
+  void SetModelFetcher(std::unique_ptr<OllamaModelFetcher> model_fetcher);
```
goodov: this should be set via constructor.

components/ai_chat/core/browser/ollama/ollama_service.h
```diff
@@ -102,6 +109,7 @@ class OllamaService : public KeyedService, public mojom::OllamaService {
 
   scoped_refptr<network::SharedURLLoaderFactory> url_loader_factory_;
   mojo::ReceiverSet<mojom::OllamaService> receivers_;
+  std::unique_ptr<OllamaModelFetcher> model_fetcher_;
```
goodov, line 101: this should be reset in `::Shutdown()`.

components/ai_chat/core/browser/ollama/ollama_model_fetcher.h
```diff
@@ -62,7 +61,7 @@ class OllamaModelFetcher {
   // (https://github.com/brave/brave-core/commit/dadaf97432efd7d64289bd92b7d31cc6e16722f0)
   const raw_ref<ModelService> model_service_;
   raw_ptr<PrefService> prefs_;
-  std::unique_ptr<OllamaService> ollama_service_;
+  raw_ptr<OllamaService> ollama_service_;
```
goodov: make this a delegate, don't make these classes depend on each other.

components/ai_chat/core/browser/ollama/ollama_model_fetcher.h
```diff
@@ -62,7 +61,7 @@ class OllamaModelFetcher {
   // (https://github.com/brave/brave-core/commit/dadaf97432efd7d64289bd92b7d31cc6e16722f0)
   const raw_ref<ModelService> model_service_;
   raw_ptr<PrefService> prefs_;
-  std::unique_ptr<OllamaService> ollama_service_;
```
goodov, line 65: o_O

the way we had a unique_ptr to a `KeyedService` class here before is wild. We should never do that.

components/ai_chat/core/browser/ollama/ollama_service.h
thypon: https://github.com/brave/brave-core/pull/31569/commits/ed8640869d979ee2bdeaf3a65f496818b9b325fe

components/ai_chat/core/browser/ollama/ollama_service.h
thypon: https://github.com/brave/brave-core/pull/31569/commits/ed8640869d979ee2bdeaf3a65f496818b9b325fe

components/ai_chat/core/browser/ollama/ollama_model_fetcher.h
thypon: https://github.com/brave/brave-core/pull/31569/commits/ed8640869d979ee2bdeaf3a65f496818b9b325fe

components/ai_chat/core/browser/ollama/ollama_model_fetcher.h
```diff
@@ -51,17 +83,16 @@ class OllamaModelFetcher {
...
+                             std::optional<ModelDetails> details);
 
   // TODO(https://github.com/brave/brave-browser/issues/49828): display_name is
   // currently identical to model_name but will be used for proper name
   // formatting when we reland the formatting patch
   // (https://github.com/brave/brave-core/commit/dadaf97432efd7d64289bd92b7d31cc6e16722f0)
   const raw_ref<ModelService> model_service_;
   raw_ptr<PrefService> prefs_;
-  raw_ptr<OllamaService> ollama_service_;
+  raw_ptr<Delegate> delegate_;
```
goodov: `= nullptr;`

thypon: fixed in https://github.com/brave/brave-core/pull/31569/commits/8a7c527c1d53a75c4313a02bb01cd86ccabd11bb

PR #31573, opened 2025-10-01
[iOS] Disable cancelling save custom filters while saving

StephenHeaps:
- Disable the cancel button when save has started

Resolves https://github.com/brave/brave-browser/issues/49651

### Testplan

1. In Shields & Privacy > Content filtering > Custom filters, add a filter
2. Tap save then try and tap cancel right after
3. Verify cancel is disabled while saving


StephenHeaps:
> I assume there's no way to actually cancel the compilation so disabling the cancel is the only way?

Not without further refactoring (if the adblock engine recompile has started we'd need to wait and re-compile the previous rules) so we chose to just disable cancel once save has started for this lesser used screen.

PR #31562, opened 2025-10-01
Separate AOM pref handling to mixin

aseren:
This would allow handling AOM pref within different pages without code duplication

Resolves https://github.com/brave/brave-browser/issues/49820

PR #31575, opened 2025-10-02
Disable related settings when AOM is enabled

aseren:
Disable `Ad Block Only mode` related settings when `Ad Block Only mode` is enabled

This PR uses Mixin, which was introduced https://github.com/brave/brave-core/pull/31562.

Resolves https://github.com/brave/brave-browser/issues/49820


browser/resources/settings/default_brave_shields_page/default_brave_shields_page.ts
```diff
@@ -208,6 +209,11 @@ class BraveShieldsPage extends BraveShieldsPageBase {
   }
 
   onAdblockPageClick_() {
+    // Prevent navigation to the content filters page when Ad Block Only mode
+    // is enabled.
+    if (this.isAdBlockOnlyModeEnabled_) {
```
fallaciousreasoning: Should we just hide the UI element when this is enabled, rather than preventing the click?

fallaciousreasoning: At the very least, seems like we should give it some disabled styles

aseren: Thanks for the comment. Disabled styles makes sense! I've removed this code and added following style:
```
  #contentFiltersRow[disabled] {
    background-color: var(--leo-color-container-disabled);
    pointer-events: none;
  }
```

fallaciousreasoning (approved):

PR #31585, opened 2025-10-02
[NTP] Remove max width on NTT click-through region

zenparsing:
Resolves https://github.com/brave/brave-browser/issues/49818


bsclifton (approved):
LGTM++

PR #31589, opened 2025-10-02
Retry on failure in sign_binaries.py

wknapik:
Resolves https://bravesoftware.slack.com/archives/C5DT3GPCH/p1759355863614809?thread_ts=1759346895.732879&cid=C5DT3GPCH

### DIFF
--- a/script/sign_binaries.py
+++ b/script/sign_binaries.py
@@ -34,7 +34,7 @@ assert cert or cert_hash or signtool_args, \
     'Store. It is ambiguous and will likely be deprecated in the future.'
 
 
-# pylint: disable=W0102
+# pylint: disable=dangerous-default-value
 def execute_with_retry(argv, env=os.environ, count=5):
     return exponential_backoff_retry(lambda: execute(argv, env),
                                      excs=(RuntimeError, CalledProcessError),


script/sign_binaries.py
```diff
@@ -32,6 +33,20 @@
     'CERT is a part of the name in the //CurrentUser/My Windows Certificate ' \
     'Store. It is ambiguous and will likely be deprecated in the future.'
 
+def execute_with_retry(cmd, max_attempts, sleep_sec = 10):
+    for attempt in range(max_attempts + 1):
+        while True:
+            try:
+                execute(cmd)
+            except Exception:
```
mherrmann: As mentioned on Slack, I find `Exception` too broad. For example, it would also retry on a `FileNotFoundError`, caused by a missing `signtool` binary. It makes no sense to do this.
```suggestion
            except (RuntimeError, CalledProcessError):
```

script/sign_binaries.py
```diff
@@ -32,6 +33,20 @@
     'CERT is a part of the name in the //CurrentUser/My Windows Certificate ' \
     'Store. It is ambiguous and will likely be deprecated in the future.'
 
+def execute_with_retry(cmd, max_attempts, sleep_sec = 10):
+    for attempt in range(max_attempts + 1):
+        while True:
```
mherrmann: I believe this `while True` will loop forever, as long as exceptions occur, which renders `max_attempts` meaningless. I believe the `while True` needs to be removed.

script/sign_binaries.py
```diff
@@ -32,6 +33,20 @@
...
+            try:
+                execute(cmd)
+            except Exception:
+                if attempt == max_attempts:
+                    print(f'Command `{cmd}\' failed. Maximum number of retries reached.', file=sys.stderr)
```
mherrmann:
```suggestion
                    print(f"Command `{cmd}' failed. Maximum number of retries reached.", file=sys.stderr)
```

script/sign_binaries.py
```diff
@@ -32,6 +33,20 @@
...
+                    raise
+                else:
+                    print(f'Command `{cmd}\' failed. Retrying in {sleep_sec}s.', file=sys.stderr)
```
mherrmann:
```suggestion
                    print(f"Command `{cmd}' failed. Retrying in {sleep_sec}s.", file=sys.stderr)
```

script/sign_binaries.py
mherrmann: Please also reformat to obey 80 char line lengths, like the rest of the files in `script/`.

script/sign_binaries.py
wknapik: What's the downside? We might retry when we don't need to, at an expense of <1m.

script/sign_binaries.py
wknapik: Removed

script/sign_binaries.py
wknapik: Done

script/sign_binaries.py
wknapik: Done

script/sign_binaries.py
wknapik: Are we guaranteed not to get any other exception where we might succeed on retry?

mherrmann: I want code to blow up when something unexpected occurs. For example, an intermittent `FileNotFoundError` may highlight an underlying issue that we could (and should) fix. With the broad `except Exception`, we will not notice here, when the problem occurs, but possibly at some random point later that is far removed and harder to debug. I believe it is also your philosophy to not gloss over errors, even if rare, but to fix them at their root. We can still catch further exceptions later. But by catching `Exception` now, we may never learn about other potential troubles in the system.

wknapik: Alright, changed

mherrmann: :pray:

script/sign_binaries.py
```diff
@@ -33,6 +35,27 @@
...
+    for attempt in range(max_attempts + 1):
+        try:
+            execute(cmd)
+        except (RuntimeError, CalledProcessError):
+            err = f"Command `{cmd}' failed."
+            if attempt == max_attempts:
+                print(f"{err} Max number of retries reached.", file=sys.stderr)
+                raise
+            else:
+                sleep_sec = base_sleep_sec * pow(backoff_mult, attempt)
```
mherrmann:
```suggestion
                sleep_sec = base_sleep_sec * backoff_mult ** attempt
```

script/sign_binaries.py
```diff
@@ -33,6 +35,27 @@
     'Store. It is ambiguous and will likely be deprecated in the future.'
 
 
+# pylint: disable=no-else-raise
```
mherrmann: What I'm about to write is my stylistic preference and you can leave the code as-is. Having said that:

Why fight the linter and the language? Get rid of the pylint: disable, remove the else: and de-indent its body. The code will be shorter, readers will not need to cognitively deal with "why is the pylint: disable necessary?" and the structure of the code will still make the `raise` stand out visually.

What's more, if you feel this way about `no-else-raise` here, then I'd imagine you'd feel the same way about it everywhere. In this case, the "proper" solution would be to disable `no-else-raise` globally, instead of a single comment here.

But as I said, it's a nit that's not very important and shouldn't hold us up.

mherrmann (changes requested):

script/sign_binaries.py
wknapik: Done

script/sign_binaries.py
wknapik: > Why fight the linter

IMO this is a harmful rule.

> and the language?

The specific language has nothing to do with it.

The change from
```python
if condition:
    foo
    exit/raise/etc.
else:
    bar
```
to
```python
if condition:
    foo
    exit/raise/etc.

bar
```
makes the code "more cleverer", i.e. harder to read for humans, while providing no benefit.

The original structure can be followed without actually reading the text. You know the the `else` part will only happen if the `condition` is false.

The updated structure hides that and requires you to read the code to realize that the jump in the `if` means subsequent code won't be executed.

I'd call that a shot in the foot.

> Get rid of the pylint: disable, remove the else: and de-indent its body. The code will be shorter

The number of lines wouldn't change. But the structure would become less clear.

> readers will not need to cognitively deal with "why is the pylint: disable necessary?"

I don't think anyone would ask that question when skimming the code. Maybe when digging in deeper, but that's a different situation altogether.

> and the structure of the code will still make the raise stand out visually.

I don't see anything that makes the raise stand out visually. It's another bit of text in a block of text.

My argument is that I want to code structure to be clear without having to actually read the text.

> What's more, if you feel this way about no-else-raise here, then I'd imagine you'd feel the same way about it everywhere.

Absolutely.

> In this case, the "proper" solution would be to disable no-else-raise globally, instead of a single comment here.

I wouldn't consider fighting that for a second. It would be a massive time sink and would get nowhere. 

> But as I said, it's a nit that's not very important and shouldn't hold us up.

It's good that you brought it up.

wknapik: I can change it, if you think it should be changed. Up to you.

mherrmann: It's OK, it's one preference against another and in that case I think it's better to go with the PR author's preference.

script/sign_binaries.py
```diff
@@ -33,6 +35,27 @@
...
+def execute_with_retry(cmd, max_attempts=5, base_sleep_sec=1, backoff_mult=2):
```
goodov: depot_tools has a nice generic python-way helper [1]. Pls consider calling `execute` via this helper. `lib/util.py` seems like a good place for both.

```
def exponential_backoff_retry(...):
    ...

def execute_with_retry(argv, env=os.environ, max_attempts=5):
    return exponential_backoff_retry(lambda: execute(argv, env),
                                     excs=(RuntimeError, CalledProcessError),
                                     max_attempts=max_attempts)
```

`printerr` param can be removed in favor of unconditional call to `print(file=sys.stderr)`. .

1. https://source.chromium.org/chromium/chromium/src/+/main:third_party/depot_tools/gclient_utils.py;l=1423-1459?q=gclient_utils&ss=chromium%2Fchromium%2Fsrc

goodov:
@wknapik did you run test-brave-browser-build (mac) to ensure this works fine?

script/sign_binaries.py
```diff
@@ -33,6 +34,12 @@
     'Store. It is ambiguous and will likely be deprecated in the future.'
 
 
+# pylint: disable=W0102
```
mherrmann: Is it possible to change this to the more readable `pylint: disable=dangerous-default-value`?

mherrmann:
Follow-up refactoring: https://github.com/brave/brave-browser/issues/50661

script/sign_binaries.py
wknapik: Done

mherrmann (approved):

wknapik:
@goodov yes

PR #31590, opened 2025-10-02
[Leo] Add new models

Steeeephen:
Resolves https://github.com/brave/brave-browser/issues/49218

image

### DIFF
--- a/components/ai_chat/core/browser/model_service.cc
+++ b/components/ai_chat/core/browser/model_service.cc
@@ -281,6 +281,7 @@ const std::vector<mojom::ModelPtr>& GetLeoModels() {
       model->vision_support = true;
       model->supports_tools = false;
       model->is_suggested_model = false;
+      model->is_near_model = false;
       model->options =
           mojom::ModelOptions::NewLeoModelOptions(std::move(options));
 
@@ -305,6 +306,7 @@ const std::vector<mojom::ModelPtr>& GetLeoModels() {
       model->vision_support = true;
       model->supports_tools = false;
       model->is_suggested_model = false;
+      model->is_near_model = false;
       model->options =
           mojom::ModelOptions::NewLeoModelOptions(std::move(options));
 
@@ -329,6 +331,7 @@ const std::vector<mojom::ModelPtr>& GetLeoModels() {
       model->vision_support = false;
       model->supports_tools = false;
       model->is_suggested_model = false;
+      model->is_near_model = false;
       model->options =
           mojom::ModelOptions::NewLeoModelOptions(std::move(options));
 
@@ -351,6 +354,7 @@ const std::vector<mojom::ModelPtr>& GetLeoModels() {
       model->vision_support = false;
       model->supports_tools = false;
       model->is_suggested_model = false;
+      model->is_near_model = false;
       model->options =
           mojom::ModelOptions::NewLeoModelOptions(std::move(options));
 
@@ -373,6 +377,7 @@ const std::vector<mojom::ModelPtr>& GetLeoModels() {
       model->vision_support = true;
       model->supports_tools = false;
       model->is_suggested_model = false;
+      model->is_near_model = false;
       model->options =
           mojom::ModelOptions::NewLeoModelOptions(std::move(options));
 
@@ -395,6 +400,7 @@ const std::vector<mojom::ModelPtr>& GetLeoModels() {
       model->vision_support = true;
       model->supports_tools = false;
       model->is_suggested_model = false;
+      model->is_near_model = false;
       model->options =
           mojom::ModelOptions::NewLeoModelOptions(std::move(options));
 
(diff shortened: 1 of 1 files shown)


components/ai_chat/core/browser/model_service.cc
```diff
@@ -253,6 +253,232 @@ const std::vector<mojom::ModelPtr>& GetLeoModels() {
       models.push_back(std::move(model));
     }
 
+    // Llama 4 Scout
+    {
```
bridiver, line 267: This is very verbose, you could either pass the properties to the `New` call, create a method to encapsulate creating them, etc...

browser/ai_chat/ai_chat_settings_helper.cc
```diff
@@ -134,6 +134,36 @@ void AIChatSettingsHelper::GetModelsWithSubtitles(
       } else if (model->key == "chat-gemma") {
         modelWithSubtitle.subtitle =
             l10n_util::GetStringUTF8(IDS_CHAT_UI_CHAT_GEMMA_SUBTITLE);
+      } else if (model->key == "chat-llama-4-scout") {
+        modelWithSubtitle.subtitle =
+            l10n_util::GetStringUTF8(IDS_CHAT_UI_CHAT_LLAMA_4_SCOUT_SUBTITLE);
+      } else if (model->key == "chat-llama-4-maverick") {
```
bridiver: `chat-llama-4-maverick`, etc.. should be constants

components/ai_chat/core/browser/model_service.cc
Steeeephen: I'm not much of a c++ guy so I was following the existing format, didn't wanna rock the boat. @Douglashdaniel wdyt?

bridiver: oh, there are already more of these? Maybe it needs a refactor before this PR because even the list of newly added models is very long. Probably should move to another file

bridiver: could be a follow-up as well

browser/ai_chat/ai_chat_settings_helper.cc
bridiver: I guess per https://github.com/brave/brave-core/pull/31590/files#r2410675182 we could leave this to be consistent and do a follow-up to fix all of them. Using string literals is very error-prone

Steeeephen: that works for me! 👍

bridiver: can you please create a follow-up issue for both?

yrliou (approved):
Code LGTM

components/ai_chat/core/browser/model_service.cc
```diff
@@ -254,6 +254,232 @@ const std::vector<mojom::ModelPtr>& GetLeoModels() {
...
+                            ? mojom::ModelAccess::BASIC_AND_PREMIUM
+                            : mojom::ModelAccess::BASIC;
+      options->max_associated_content_length = 64000;
+      options->long_conversation_warning_character_limit = 9700;
+
+      auto model = mojom::Model::New();
+      model->key = "chat-llama-4-scout";
+      model->display_name = "Llama 4 Scout";
+      model->vision_support = true;
+      model->supports_tools = true;
```
yrliou: @stoletheminerals we're adding quite a few models here with tool support, do you have any concern with this, specifically for the prompt injection protection aspect? Not sure if we would want to audit if they have good enough quality of following security prompts and if alignment check is as effective as existing models? Should we open a security and privacy review when adding new models? It would be nice if we could have a clear guideline on when we need to and when we don't.

Douglashdaniel (approved):
Front-end LGTM!

components/ai_chat/core/browser/model_service.cc
Steeeephen: after https://github.com/brave/brave-core/pull/32011 - thinking we should just set to `false` for now and we can eval later

components/resources/ai_chat_ui_strings.grdp
```diff
@@ -552,6 +582,36 @@
...
+    Search • Tools
+  </message>
+  <message name="IDS_CHAT_UI_CHAT_DEEPSEEK_V3_1_SUBTITLE" desc="A description for deepseek-v3-1 model" formatter_data="webui=AiChat">
+    Search
+  </message>
+  <message name="IDS_CHAT_UI_CHAT_QWEN_3_CODER_480B_SUBTITLE" desc="A description for qwen-3-coder-480b model" formatter_data="webui=AiChat">
+    Search • Tools
+  </message>
+  <message name="IDS_CHAT_UI_CHAT_CLAUDE_OPUS_SUBTITLE" desc="A description for claude-opus model" formatter_data="webui=AiChat">
+    Search • Vision • Tools
```
Douglashdaniel: Hey, since we are disabling tools for all of these. We probably should update these strings to not include `Tools` too

Steeeephen: good spot

mkarolin (approved):
`strings`++

PR #31597, opened 2025-10-02
Add npm run gn <gn_command> [build_config] [gn_args...] command.

goodov:
Resolves


build/commands/scripts/gn.js
```diff
@@ -0,0 +1,47 @@
...
+    'gn',
+    [gnCommand, config.outputDir, ...gnArgs, ...getUnknownOptions(program)],
+    config.defaultOptions,
+  )
+}
+
+function* getUnknownOptions(command) {
+  const opts = command.opts()
+  for (const option of command.parseOptions(process.argv).unknown) {
+    // Filter out actually known options passed as --option=value.
```
bridiver, line 40: it's not clear to me what we are considering "unknown" options here with this?

goodov: > it's not clear to me what we are considering "unknown" options here with this?

everything that should be passed to GN except those listed in the npm command itself: `--target_os, --target_arch, --target_environment`.

bridiver: oh nvm, I thought you were filtering out everything that had `--option=value` because I missed `.unknown` above

petemill (approved):

PR #31601, opened 2025-10-02
Run gn clean automatically on aborted build on CI.

goodov:
Resolves


bridiver:
This seems like it will cause a lot of unnecessary rebuilds, I squash commits all the time or maybe rebase to pull in some missing commits and a clean is definitely not needed. I think the only we'd want to automatically clean is if the chromium major version has changed

bridiver (changes requested):
This seems like it will cause a lot of unnecessary rebuilds, I squash commits all the time or maybe rebase to pull in some missing commits and a clean is definitely not needed. I think the only we'd want to automatically clean is if the chromium major version has changed

goodov:
> This seems like it will cause a lot of unnecessary rebuilds, I squash commits all the time or maybe rebase to pull in some missing commits and a clean is definitely not needed. I think the only we'd want to automatically clean is if the chromium major version has changed

You're not wrong, but we either fix all things at once or fallback to manually clicking GN_CLEAN/WIPE_WORKSPACE from time to time. I'm pretty sure we should go with auto-clean and let build run 5-10 minutes longer in some scenarios instead of figuring random failures manually.

Clean is only executed if you force-push into a build that was currently running `autoninja` command. With RBE the actual build in PR is pretty fast, so I don't see this as a real issue tbh.

Upstream does exactly this: https://source.chromium.org/chromium/infra/infra_superproject/+/main:build/recipes/recipe_modules/chromium/api.py;l=693-715;drc=af0f535913291930256660205a2cfd6a8c9bf0ab

build/commands/lib/util.js
```diff
@@ -289,6 +289,27 @@ const util = {
...
         resolve(stdout)
+      }
+      prog.on('close', (statusCode, signal) => {
+        if (config.isCI && (statusCode || signal)) {
+          // When running in CI, we delay handling process termination by 1
+          // second to distinguish between two scenarios:
+          // 1. A build failure (where autoninja exits with code 1)
+          // 2. CI killing the process tree with SIGTERM
+          //
+          // Without this delay, both scenarios would appear the same since
```
bridiver, line 300: in this case wouldn't the npm process have already received SIGTERM? I assume it starts with the root of the process tree to prevent re-spawning of child processes? Could we track that here instead of using the timeout?

goodov: > in this case wouldn't the npm process have already received SIGTERM? I assume it starts with the root of the process tree to prevent re-spawning of child processes? Could we track that here instead of using the timeout?

this is exactly what I tried to do, but for some reason `SIGTERM` handling by node happens later than autoninja exit. After putting logs here and there I saw autoninja exit was handled _before_ the handler set by `process.addListener(signal, signalHandler)` was executed.

I also expected `SIGTERM` signal handler in node to react immediately, but I have a feeling the process shutdown happens in reversed order.

PR #31602, opened 2025-10-02
[Rounded Corners] Replace flag with a pref and settings toggle

zenparsing:
Resolves


simonhong:
Yay! :)

chromium_src/chrome/browser/ui/views/side_panel/side_panel_web_ui_view.cc
```diff
@@ -3,11 +3,13 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this file,
  * You can obtain one at https://mozilla.org/MPL/2.0/. */
 
-#include "brave/browser/brave_browser_features.h"
 #include "brave/browser/ui/views/frame/brave_contents_view_util.h"
 #include "brave/browser/ui/views/side_panel/side_panel_web_ui_view_utils.h"
+#include "brave/components/constants/pref_names.h"
```
simonhong: I think we need to handle this new dependency for `//chrome/browser/ui/views/side_panel` target.
Instead of adding new deps in `chromium_src` here, how about declare method here and define it in our code base? something like `void UpdateSidePanelCornerRadius(...);`

simonhong (approved):
++

app/brave_settings_strings.grdp
```diff
@@ -198,6 +198,9 @@
   <message name="IDS_SETTINGS_APPEARANCE_SETTINGS_LOCATION_BAR_IS_WIDE" desc="The label for whether the location bar should display wide or not">
     Use wide address bar
   </message>
+  <message name="IDS_SETTINGS_APPEARANCE_SETTINGS_WEB_VIEW_ROUNDED_CORNERS" desc="The label for whether the main content areas should have rounded corners">
+    Show rounded corners on main content areas
```
mkarolin, line 202: I am not sure, but maybe `in` instead of `on`?

browser/speedreader/speedreader_browsertest.cc
```diff
@@ -1097,8 +1098,12 @@ IN_PROC_BROWSER_TEST_F(SpeedReaderBrowserTest, ToolbarWithRoundedCorners) {
   EXPECT_TRUE(speedreader::DistillStates::IsDistilled(
       tab_helper()->PageDistillState()));
 
+  const bool rounded_contents =
```
bridiver, line 1101: this doesn't seem right, why does speedreader need to know about this?

bridiver: oic, nvm

bridiver: errr... actually I take that back. What is the reason for adding this here?

browser/ui/brave_browser.cc
```diff
@@ -63,12 +63,11 @@ void BraveBrowser::SuppressBrowserWindowClosingDialogForTesting(bool suppress) {
 }
 
 // static
-bool BraveBrowser::IsBraveWebViewRoundedCornersFeatureEnabled(
-    Browser* browser) {
-  // If we have prefs for toggling rounded corners in runtime, check that prefs
-  // here.
-  return base::FeatureList::IsEnabled(features::kBraveWebViewRoundedCorners) &&
-         browser->is_type_normal();
+bool BraveBrowser::IsBraveWebViewRoundedCornersEnabled(Browser* browser) {
```
bridiver, line 67: I don't understand why we have this method on BraveBrowser in the first place, it's only ever called from BraveBrowserView. This should just be internal to `BraveBrowserView`.

browser/speedreader/speedreader_browsertest.cc
bridiver: this functionality is not specific to speedreader so this is not the appropriate place to test it. @simonhong can you please fix this in a follow-up? This kind of functionality should not just be tacked on to some existing test.

zenparsing: I think that this test is making sure that the speedreader toolbar has rounded corners if split view has been enabled (as split view will enable "rounded corners mode" regardless of flags or settings).

The test previously assumed that rounded corners mode only applies when split view is enabled, which is no longer the case.

components/constants/pref_names.h
```diff
@@ -134,6 +134,8 @@ inline constexpr char kEnableWindowClosingConfirm[] =
 inline constexpr char kEnableClosingLastTab[] = "brave.enable_closing_last_tab";
 inline constexpr char kShowFullscreenReminder[] =
     "brave.show_fullscreen_reminder";
+inline constexpr char kWebViewRoundedCorners[] =
```
bridiver: As the codeowners file specifies this is deprecated and new prefs should not be added here. This is browser code so it doesn't need to be in components anyway. It should probably go somewhere in browser/ui, but it's not clear to me why it's called `kWebViewRoundedCorners` when it applies to any browser that is not "normal". That list includes popups, apps, and devtools

browser/ui/brave_browser.cc
bridiver: this should be a follow-up

browser/ui/sidebar/sidebar_browsertest.cc
```diff
@@ -617,13 +617,18 @@ class SidebarBrowserWithSplitViewTest
...
-         {tabs::features::kBraveSplitView, std::get<1>(GetParam())},
+        {{tabs::features::kBraveSplitView, std::get<1>(GetParam())},
          {::features::kSideBySide, std::get<2>(GetParam())}});
   }
 
   ~SidebarBrowserWithSplitViewTest() override = default;
 
+  void SetUpOnMainThread() override {
+    SidebarBrowserTest::SetUpOnMainThread();
+    browser()->profile()->GetPrefs()->SetBoolean(kWebViewRoundedCorners,
```
bridiver, line 628: Same as speedreader, this is not the right place to test this. The sidebar test should be focused on sidebar functionality, not generic BraveBrowserView functionality

browser/ui/views/side_panel/side_panel_web_ui_view_utils.cc
```diff
@@ -14,4 +17,9 @@ bool ShouldEnableContextMenu(const GURL& url) {
   return url.is_valid() && url.host() == kAIChatUIHost;
 }
 
+int GetSidePanelCornerRadius(Profile* profile) {
```
bridiver: Please do not create wrappers for simple pref getters. In particular a pref getter for BraveBrowserView functionality should not be in side_panel_web_ui_view_utils

bridiver: also in general pass in what you need (PrefService not Profile)

components/constants/pref_names.h
zenparsing: This pref name has been moved to `common/pref_names.h`.

browser/ui/views/side_panel/side_panel_web_ui_view_utils.cc
zenparsing: The changes to this file have been removed. The pref is now checked directly in `side_panel_web_ui_view.cc`.

browser/speedreader/speedreader_browsertest.cc
simonhong: As @zenparsing wrote, speedreader toolbar's corner radius should be updated whenever rounded corners is changed.

browser/ui/brave_browser.cc
simonhong: @zenparsing I have a f/u idea.
If we have another prefs(ex, `kWebViewRoundedCornersRunTime`) for below `BraveBrowser::ShouldUseBraveWebViewRoundedCornersForContents()`,
I think we could remove both methods from here.
With this new prefs, other place just need observe it.

browser/ui/sidebar/sidebar_browsertest.cc
bridiver: for follow-up also since it appears this was already being tested here

browser/speedreader/speedreader_browsertest.cc
bridiver: @simonhong but this is not speedreader functionality, this is BrowserView functionality. We shouldn't be testing this here in speedreader unless we were making some specific exception for speedreader when it normally would not apply.

common/pref_names.h
```diff
@@ -0,0 +1,16 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#ifndef BRAVE_COMMON_PREF_NAMES_H_
+#define BRAVE_COMMON_PREF_NAMES_H_
+
+#include "build/build_config.h"
+
+#if !BUILDFLAG(IS_ANDROID)
```
bridiver: this should actually be toolkit views instead of android, but I don't think the guard is strictly necessary here. Please add a description though

chromium_src/chrome/browser/ui/views/side_panel/side_panel_web_ui_view.cc
```diff
@@ -32,8 +34,12 @@ bool SidePanelWebUIView::HandleContextMenu(
 void SidePanelWebUIView::AddedToWidget() {
   WebView::AddedToWidget();
 
-  if (base::FeatureList::IsEnabled(features::kBraveWebViewRoundedCorners)) {
-    constexpr auto kRadius = BraveContentsViewUtil::kBorderRadius;
-    holder()->SetCornerRadii(gfx::RoundedCornersF(kRadius));
+  if (contents_wrapper_ && contents_wrapper_->web_contents()) {
```
bridiver: when would either of these be null?

browser/ui/brave_browser.cc
bridiver: @simonhong can you create follow-up issues for these?

browser/ui/views/frame/brave_browser_view.cc
```diff
@@ -328,6 +334,11 @@ void BraveBrowserView::OnPreferenceChanged(const std::string& pref_name) {
     return;
   }
 
+  if (pref_name == kWebViewRoundedCorners) {
+    UpdateRoundedCornersUI();
+    return;
```
bridiver, line 339: also for follow-up this should probably be if/else instead of a bunch of if/return. Looked potentially concerning until I expanded the context that was visible. Technically the return isn't even necessary anyway, but it's fine to follow the existing ones for this PR

chromium_src/chrome/browser/ui/views/side_panel/side_panel_web_ui_view.cc
zenparsing: Fixed. `contents_wrapper_` can't be null, and less obviously, `contents_wrapper_->web_contents()` should never be null. Added a `DCHECK` for `contents_wrapper_->web_contents()`.

common/pref_names.h
zenparsing: Removed guard and added comment.

browser/ui/brave_browser.cc
simonhong: Filed two f/u issues.
* https://github.com/brave/brave-browser/issues/50302 
* https://github.com/brave/brave-browser/issues/50303

app/brave_settings_strings.grdp
zenparsing: I think "on" is probably best, if "main content areas" means the webviews themselves.

PR #31606, opened 2025-10-02
[perf] Load NTP wallpaper once

atuchin-m:
Resolves https://github.com/brave/brave-browser/issues/49859
The PR removes the second redundant loading of `chrome://background-wallpaper/<something>` by using blob URL.

### DIFF
--- a/browser/resources/brave_new_tab_page_refresh/components/background/background.tsx
+++ b/browser/resources/brave_new_tab_page_refresh/components/background/background.tsx
@@ -7,7 +7,7 @@ import * as React from 'react'
 
 import { NewTabPageAdEventType, SponsoredImageBackground } from '../../state/background_state'
 import { openLink } from '../common/link'
-import { loadImage } from '../../lib/image_loader'
+import { loadImageAsBlob } from '../../lib/image_loader'
 import { useWidgetLayoutReady } from '../app_layout_ready'
 import { debounce } from '$web-common/debounce'
 import { useSearchState, useSearchActions } from '../../context/search_context'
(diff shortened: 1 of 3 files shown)


browser/resources/brave_new_tab_page_refresh/lib/image_loader.ts
```diff
@@ -6,33 +6,34 @@
 export const placeholderImageSrc =
   'data:image/svg+xml,%3Csvg xmlns="http://www.w3.org/2000/svg"%3E%3C/svg%3E'
 
-// Loads an image in the background and resolves when the image has either
-// loaded or was unable to load.
-export function loadImage(url: string): Promise<boolean> {
-  return new Promise((resolve) => {
+// Loads an image in the background and resolves when it has loaded.
+// Returns a blob URL if successful, rejects if failed.
+export function loadImageAsBlob(url: string): Promise<string> {
```
zenparsing, line 11: Can this just be an async function now?

browser/resources/brave_new_tab_page_refresh/components/background/background.tsx
```diff
@@ -71,17 +71,27 @@ function setBackgroundVariable(value: string) {
 }
 
 function ImageBackground(props: { url: string, onLoadError?: () => void }) {
+  const [blobUrl, setBlobUrl] = React.useState<string | null>(null)
```
zenparsing, line 74: Curious: do we need to keep the `blobUrl` in component state, or could it just be  a local variable in the `useEffect` callback?

browser/resources/brave_new_tab_page_refresh/lib/image_loader.ts
```diff
@@ -6,33 +6,34 @@
...
-    const onError = () => {
-      unlisten()
-      resolve(false)
-    }
-
-    const image = new Image()
-    image.addEventListener('load', onLoad)
-    image.addEventListener('error', onError)
-    image.src = url
+    fetch(url)
```
fallaciousreasoning, line 18: oh interesting - I didn't think you could load `chrome://` urls in `fetch`

fallaciousreasoning (commented):
Hey so more broadly, was this not just getting served from the cache when it was loaded the second time?

I wouldn't be that surprised if this wasn't much different to what was already here.

fallaciousreasoning (commented):
Like, would we be better off fixing the cache headers if this is causing an impact?

atuchin-m:
>Hey so more broadly, was this not just getting served from the cache when it was loaded the second time?
I wouldn't be that surprised if this wasn't much different to what was already here.

I double checked and looks like it's my local issues: `no-cache` checkbox in the devtools. The `chrome://<something>` cache policy is unclear, It looked like we still have `Cache control: no-cache` anyway. But in fact we don't.
I've checked a fresh profile plus Beta and everything is okay.

So there is no real issue to fix, caching works as expected ( see the screenshots).
For now I believe we can just decline this PR.

Screenshot 2025-10-03 at 14 05 10
Screenshot 2025-10-03 at 14 05 02

browser/resources/brave_new_tab_page_refresh/lib/image_loader.ts
atuchin-m: In fact it works for some chrome:// schemes, but you need to add a CSP exclusion.

PR #31607, opened 2025-10-02
add GetAccountBalance to PolkadotWalletService

cmazakas:
Add our first real RPC call to the Polkadot nodes: getting account balance and other information.

RPC calls seem to work for this by creating a StorageKey which is a hex-encoded string containing a concatenation of the xxhash of the pallet name ("System") and the command name ("Account") along with the account's public key and a small checksum.

The response is a binary encoded hex string containing all the relevant information, defined here:
```
    "id": 3,
    "type": {
      "path": [
        "frame_system",
        "AccountInfo"
      ],
     ...
```

https://raw.githubusercontent.com/polkadot-js/api/refs/heads/master/packages/types-support/src/metadata/v16/substrate-types.json

### DIFF
--- a/components/brave_wallet/browser/keyring_service.cc
+++ b/components/brave_wallet/browser/keyring_service.cc
@@ -2599,6 +2599,14 @@ KeyringService::SignCip30MessageByCardanoKeyring(
   return result;
 }
 
+std::array<uint8_t, kSr25519PublicKeySize> KeyringService::GetPolkadotPubKey(
+    const mojom::AccountIdPtr& account_id) {
+  CHECK(account_id);
+  auto* keyring = this->GetKeyring<PolkadotKeyring>(account_id->keyring_id);
+  auto key = keyring->GetPublicKey(account_id->account_index);
+  return {key};
+}
+
 void KeyringService::UpdateNextUnusedAddressForBitcoinAccount(
     const mojom::AccountIdPtr& account_id,
     std::optional<uint32_t> next_receive_index,
--- a/components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.cc
+++ b/components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.cc
@@ -5,11 +5,17 @@
 
 #include "brave/components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.h"
 
+#include "base/containers/span.h"
 #include "base/functional/bind.h"
 #include "base/json/json_writer.h"
+#include "base/numerics/byte_conversions.h"
+#include "base/strings/strcat.h"
+#include "base/strings/string_number_conversions.h"
 #include "brave/components/brave_wallet/browser/brave_wallet_utils.h"
 #include "brave/components/brave_wallet/browser/network_manager.h"
 #include "brave/components/brave_wallet/browser/polkadot/polkadot_substrate_rpc_responses.h"
+#include "brave/components/brave_wallet/common/brave_wallet_types.h"
+#include "brave/components/brave_wallet/common/hash_utils.h"
 
 namespace brave_wallet {
 
@@ -38,6 +44,160 @@ net::NetworkTrafficAnnotationTag GetNetworkTrafficAnnotationTag() {
           "Not implemented."
       }
     )");
+}
+
+mojom::PolkadotAccountInfoPtr ParseAccountInfoAsHex(std::string_view sv) {
+  auto account = mojom::PolkadotAccountInfo::New();
+
+  DCHECK(sv.size() == 162);
+  sv.remove_prefix(2);  // remove leading 0x
+
+  if (!base::HexStringToUInt(sv.substr(0, 8), &account->nonce)) {
+    return nullptr;
+  }
+  sv.remove_prefix(8);
+
+  if (!base::HexStringToUInt(sv.substr(0, 8), &account->consumers)) {
+    return nullptr;
+  }
+  sv.remove_prefix(8);
+
+  if (!base::HexStringToUInt(sv.substr(0, 8), &account->providers)) {
+    return nullptr;
+  }
+  sv.remove_prefix(8);
+
+  if (!base::HexStringToUInt(sv.substr(0, 8), &account->sufficients)) {
+    return nullptr;
+  }
+  sv.remove_prefix(8);
+
+  account->nonce = base::ByteSwap(account->nonce);
+  account->consumers = base::ByteSwap(account->consumers);
+  account->providers = base::ByteSwap(account->providers);
+  account->sufficients = base::ByteSwap(account->sufficients);
+
+  account->data = mojom::PolkadotAccountBalance::New();
+
+  uint64_t low = 0, high = 0;
+  if (!base::HexStringToUInt64(sv.substr(0, 16), &low)) {
+    return nullptr;
+  }
+  sv.remove_prefix(16);
+  low = base::ByteSwap(low);
+
+  if (!base::HexStringToUInt64(sv.substr(0, 16), &high)) {
+    return nullptr;
+  }
+  sv.remove_prefix(16);
+  high = base::ByteSwap(high);
+
+  account->data->free = mojom::uint128Ptr(std::in_place, high, low);
+
+  if (!base::HexStringToUInt64(sv.substr(0, 16), &low)) {
+    return nullptr;
+  }
+  sv.remove_prefix(16);
+  low = base::ByteSwap(low);
+
+  if (!base::HexStringToUInt64(sv.substr(0, 16), &high)) {
+    return nullptr;
+  }
+  sv.remove_prefix(16);
+  high = base::ByteSwap(high);
+
+  account->data->reserved = mojom::uint128Ptr(std::in_place, high, low);
+
+  if (!base::HexStringToUInt64(sv.substr(0, 16), &low)) {
+    return nullptr;
+  }
+  sv.remove_prefix(16);
+  low = base::ByteSwap(low);
+
+  if (!base::HexStringToUInt64(sv.substr(0, 16), &high)) {
+    return nullptr;
+  }
+  sv.remove_prefix(16);
+  high = base::ByteSwap(high);
+
+  account->data->frozen = mojom::uint128Ptr(std::in_place, high, low);
+
+  if (!base::HexStringToUInt64(sv.substr(0, 16), &low)) {
+    return nullptr;
+  }
+  sv.remove_prefix(16);
+  low = base::ByteSwap(low);
+
+  if (!base::HexStringToUInt64(sv.substr(0, 16), &high)) {
+    return nullptr;
+  }
+  sv.remove_prefix(16);
+  high = base::ByteSwap(high);
+
+  account->data->flags = mojom::uint128Ptr(std::in_place, high, low);
+
+  return account;
+}
+
+mojom::PolkadotAccountInfoPtr ParseAccountInfoFromJson(
+    const std::optional<
+        std::vector<polkadot_substrate_rpc_responses::AccountInfo>>& result) {
+  /*
+
+  The shape of the resturned JSON is:
+
+  {
+    ...,
+    "result": [
+      {
+        "block": "<block hash>",
+        "changes": [["<storage key>", "<account info>" | null], ...]
+      },
+      ...
+    ]
+  }
+
+  */
+
+  if (!result) {
+    return nullptr;
+  }
+
+  auto& accounts = *result;
+  if (accounts.size() != 1) {
+    return nullptr;
+  }
+
+  auto& changes = accounts[0].changes;
+  if (changes.empty()) {
+    return nullptr;
+  }
+
+  auto* list = changes[0].GetIfList();
+  if (!list || list->size() != 2) {
+    return nullptr;
+  }
+
+  constexpr const char kFallback[] =
+      "0x0000000000000000000000000000000000000000000000000000000000000000"
+      "000000000000000000000000000000000000000000000000000000000000000000"
+      "000000000000000000000000000080";
+
+  std::string_view sv;
+  if (auto* str = (*list)[1].GetIfString()) {
+    sv = *str;
+  } else {
+    sv = kFallback;
+  }
+
+  if (sv.size() == 162) {
+    auto account = ParseAccountInfoAsHex(sv);
+    if (account) {
+      return account;
+    }
+  }
+
+  return nullptr;
 }
 
 }  // namespace
@@ -103,6 +263,73 @@ void PolkadotSubstrateRpc::OnGetChainName(GetChainNameCallback callback,
   }
 
   return std::move(callback).Run(std::nullopt, WalletParsingErrorMessage());
+}
+
+void PolkadotSubstrateRpc::GetAccountBalance(
+    const std::string& chain_id,
+    base::span<const uint8_t, kSr25519PublicKeySize> pubkey,
+    GetAccountBalanceCallback callback) {
+  // xxhash("System") | xxhash("Account")
+  //
+  // https://github.com/polkadot-js/common/blob/047840319ef3f758880cc112b987888b8b2749d0/packages/util-crypto/src/xxhash/asU8a.ts#L24
+  // https://github.com/paritytech/polkadot-sdk/blob/cf439301b2a9571e5fcb04e4550167a878187182/substrate/primitives/crypto/hashing/src/lib.rs#L77-L82
+  //
+  // pub fn twox_128_into(data: &[u8], dest: &mut [u8; 16]) {
+  //     let r0 = twox_hash::XxHash::with_seed(0).chain_update(data).finish();
+  //     let r1 = twox_hash::XxHash::with_seed(1).chain_update(data).finish();
+  //     LittleEndian::write_u64(&mut dest[0..8], r0);
+  //     LittleEndian::write_u64(&mut dest[8..16], r1);
+  // }
+  constexpr char const kStoragePallet[] = "26AA394EEA5630E07C48AE0C9558CEF7";
+  constexpr char const kAccount[] = "B99D880EC681799C0CF30E8886371DA9";
+
+  auto checksum = Blake2bHash<16>({pubkey});
+
+  auto rpc_cmd =
+      base::StrCat({"0x", kStoragePallet, kAccount, base::HexEncode(checksum),
+                    base::HexEncode(pubkey)});
+
+  auto payload = base::WriteJson(MakeRpcRequestJson(
+      "state_queryStorageAt",
+      base::Value::List().Append(
+          base::Value(base::Value::List().Append(std::move(rpc_cmd))))));
+
+  CHECK(payload);
+
+  auto url = GetNetworkURL(chain_id);
+
+  std::string method = net::HttpRequestHeaders::kPostMethod;
+  std::string payload_content_type = "application/json";
+
+  api_request_helper_.Request(
+      method, url, *payload, payload_content_type,
+      base::BindOnce(&PolkadotSubstrateRpc::OnGetAccountBalance,
+                     weak_ptr_factory_.GetWeakPtr(), std::move(callback)));
+}
+
+void PolkadotSubstrateRpc::OnGetAccountBalance(
+    GetAccountBalanceCallback callback,
+    APIRequestResult api_result) {
+  auto res = polkadot_substrate_rpc_responses::PolkadotAccountBalanceResponse::
+      FromValue(api_result.value_body());
+
+  if (!res) {
+    return std::move(callback).Run(nullptr, WalletParsingErrorMessage());
+  }
+
+  if (res->error) {
+    if (res->error->message) {
+      return std::move(callback).Run(nullptr, res->error->message.value());
+    }
+    return std::move(callback).Run(nullptr, WalletInternalErrorMessage());
+  }
+
+  auto account = ParseAccountInfoFromJson(res->result);
+  if (account) {
+    return std::move(callback).Run(std::move(account), std::nullopt);
+  }
+
+  return std::move(callback).Run(nullptr, WalletParsingErrorMessage());
 }
 
 GURL PolkadotSubstrateRpc::GetNetworkURL(const std::string& chain_id) {
--- a/components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.h
+++ b/components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.h
@@ -8,6 +8,8 @@
 
 #include "base/memory/weak_ptr.h"
 #include "brave/components/api_request_helper/api_request_helper.h"
+#include "brave/components/brave_wallet/browser/internal/hd_key_sr25519.h"
+#include "brave/components/brave_wallet/common/brave_wallet.mojom.h"
 #include "services/network/public/cpp/shared_url_loader_factory.h"
 
 namespace brave_wallet {
@@ -27,9 +29,18 @@ class PolkadotSubstrateRpc {
       base::OnceCallback<void(const std::optional<std::string>&,
                               const std::optional<std::string>&)>;
 
+  using GetAccountBalanceCallback =
+      base::OnceCallback<void(mojom::PolkadotAccountInfoPtr,
+                              const std::optional<std::string>&)>;
+
   // Get the name of the chain pointed to by the current network configuration.
   // "Westend" or "Paseo" for the testnets, "Polkadot" for the mainnet.
   void GetChainName(const std::string& chain_id, GetChainNameCallback callback);
+
+  void GetAccountBalance(
+      const std::string& chain_id,
+      base::span<const uint8_t, kSr25519PublicKeySize> pubkey,
+      GetAccountBalanceCallback callback);
 
  private:
   using APIRequestResult = api_request_helper::APIRequestResult;
@@ -39,6 +50,7 @@ class PolkadotSubstrateRpc {
 
   GURL GetNetworkURL(const std::string& chain_id);
   void OnGetChainName(GetChainNameCallback callback, APIRequestResult res);
+  void OnGetAccountBalance(GetAccountBalanceCallback, APIRequestResult res);
 
   const raw_ref<NetworkManager> network_manager_;
   api_request_helper::APIRequestHelper api_request_helper_;
--- a/components/brave_wallet/browser/polkadot/polkadot_substrate_rpc_responses.idl
+++ b/components/brave_wallet/browser/polkadot/polkadot_substrate_rpc_responses.idl
@@ -16,4 +16,17 @@ namespace polkadot_substrate_rpc_responses {
     DOMString? result;
     Error? error;
   };
+
+  // any[] => sequence<sequence<DOMString>>
+  dictionary AccountInfo {
+    DOMString block;
+    any[] changes;
+  };
+
+  dictionary PolkadotAccountBalanceResponse {
+    DOMString jsonrpc;
+    long id;
+    AccountInfo[]? result;
+    Error? error;
+  };
 };
--- a/components/brave_wallet/browser/polkadot/polkadot_substrate_rpc_unittest.cc
+++ b/components/brave_wallet/browser/polkadot/polkadot_substrate_rpc_unittest.cc
@@ -5,13 +5,16 @@
 
 #include "brave/components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.h"
 
+#include "base/json/json_reader.h"
 #include "base/test/scoped_feature_list.h"
 #include "base/test/task_environment.h"
 #include "base/test/test_future.h"
 #include "brave/components/brave_wallet/browser/brave_wallet_prefs.h"
 #include "brave/components/brave_wallet/browser/brave_wallet_utils.h"
 #include "brave/components/brave_wallet/browser/network_manager.h"
+#include "brave/components/brave_wallet/browser/polkadot/polkadot_substrate_rpc_responses.h"
 #include "brave/components/brave_wallet/common/features.h"
+#include "brave/components/brave_wallet/common/hash_utils.h"
 #include "components/sync_preferences/testing_pref_service_syncable.h"
 #include "services/network/public/cpp/weak_wrapper_shared_url_loader_factory.h"
 #include "services/network/test/test_url_loader_factory.h"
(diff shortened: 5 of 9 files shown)


components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.cc
```diff
@@ -40,6 +46,160 @@ net::NetworkTrafficAnnotationTag GetNetworkTrafficAnnotationTag() {
...
+
+  if (sv.size() == 162) {
+    auto account = ParseAccountInfoAsHex(sv);
+    if (account) {
+      return account;
+    }
+  }
+
+  return nullptr;
+}
```
supermassive: `return ParseAccountInfoAsHex(sv);`
And let's move size checking into ParseAccountInfoAsHex closer to where correct size is relevant

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.cc
```diff
@@ -40,6 +46,160 @@ net::NetworkTrafficAnnotationTag GetNetworkTrafficAnnotationTag() {
     )");
 }
 
+mojom::PolkadotAccountInfoPtr ParseAccountInfoAsHex(std::string_view sv) {
+  auto account = mojom::PolkadotAccountInfo::New();
+
+  DCHECK(sv.size() == 162);
+  sv.remove_prefix(2);  // remove leading 0x
```
supermassive: We should not assume it is 0x.

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.cc
```diff
@@ -40,6 +46,160 @@ net::NetworkTrafficAnnotationTag GetNetworkTrafficAnnotationTag() {
     )");
 }
 
+mojom::PolkadotAccountInfoPtr ParseAccountInfoAsHex(std::string_view sv) {
```
supermassive: Let's convert everything to a byte vector an pass `base::span<const uint8_t>` here.
Use base::SpanReader to walk through span.
Depnding on endianess read ints by ReadU64LittleEndian, ReadU64BigEndian etc(no ByteSwap calls)

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.cc
```diff
@@ -105,6 +265,73 @@ void PolkadotSubstrateRpc::OnGetChainName(GetChainNameCallback callback,
...
+  api_request_helper_.Request(
+      method, url, *payload, payload_content_type,
+      base::BindOnce(&PolkadotSubstrateRpc::OnGetAccountBalance,
+                     weak_ptr_factory_.GetWeakPtr(), std::move(callback)));
+}
+
+void PolkadotSubstrateRpc::OnGetAccountBalance(
+    GetAccountBalanceCallback callback,
+    APIRequestResult api_result) {
+  auto res = polkadot_substrate_rpc_responses::PolkadotAccountBalanceResponse::
```
cypt4, line 308: We should check api_result error_code here first

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.cc
```diff
@@ -40,6 +46,160 @@ net::NetworkTrafficAnnotationTag GetNetworkTrafficAnnotationTag() {
...
+  low = base::ByteSwap(low);
+
+  if (!base::HexStringToUInt64(sv.substr(0, 16), &high)) {
+    return nullptr;
+  }
+  sv.remove_prefix(16);
+  high = base::ByteSwap(high);
+
+  account->data->free = mojom::uint128Ptr(std::in_place, high, low);
+
```
supermassive: Typically created as 
`mojom::uint128::New(high, low)`

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.cc
```diff
@@ -40,6 +46,160 @@ net::NetworkTrafficAnnotationTag GetNetworkTrafficAnnotationTag() {
...
+      "000000000000000000000000000000000000000000000000000000000000000000"
+      "000000000000000000000000000080";
+
+  std::string_view sv;
+  if (auto* str = (*list)[1].GetIfString()) {
+    sv = *str;
+  } else {
+    sv = kFallback;
+  }
+
```
supermassive: What is the prupose of kFallback info? Is this a legitimate response when not a string returned?

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc_responses.idl
```diff
@@ -16,4 +16,17 @@ namespace polkadot_substrate_rpc_responses {
     DOMString? result;
     Error? error;
   };
+
+  // any[] => sequence<sequence<DOMString>>
+  dictionary AccountInfo {
```
cypt4, line 23: Please add links to the format description

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.cc
```diff
@@ -105,6 +265,73 @@ void PolkadotSubstrateRpc::OnGetChainName(GetChainNameCallback callback,
...
+  CHECK(payload);
+
+  auto url = GetNetworkURL(chain_id);
+
+  std::string method = net::HttpRequestHeaders::kPostMethod;
+  std::string payload_content_type = "application/json";
+
+  api_request_helper_.Request(
+      method, url, *payload, payload_content_type,
```
supermassive: Don't need for `method` and `payload` to be variables, just use values inplace

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.cc
supermassive: If we have to return some account info, let's return an instance of PolkadotAccountInfoPtr filled with fallback values

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.h
```diff
@@ -27,10 +29,19 @@ class PolkadotSubstrateRpc {
...
+
   // Get the name of the chain pointed to by the current network configuration.
   // "Westend" or "Paseo" for the testnets, "Polkadot" for the mainnet.
   void GetChainName(const std::string& chain_id, GetChainNameCallback callback);
 
+  void GetAccountBalance(
+      const std::string& chain_id,
+      base::span<const uint8_t, kPolkadotSubstrateAccountIdSize> pubkey,
+      GetAccountBalanceCallback callback);
+
```
supermassive, line 44: I'd suggest to follow approach in cardano where CardanoWalletService has per chain set of CardanoRpc objects, so each  CardanoRpc object has an assigned chain_id and we don't need to pass into every function as an arg.

components/brave_wallet/common/brave_wallet.mojom
```diff
@@ -2316,9 +2316,32 @@ interface ZCashWalletService {
       => (ZCashChainTipStatus? status, string? error_message);
 };
 
+struct uint128 {
+  uint64 high;
+  uint64 low;
+};
```
supermassive, line 2322: That maybe ok for now, but I think eventually we would find it more convenient to pass such very large integers as hex strings

components/brave_wallet/browser/keyring_service.cc
```diff
@@ -2599,6 +2599,14 @@ KeyringService::SignCip30MessageByCardanoKeyring(
   return result;
 }
 
+std::array<uint8_t, kPolkadotSubstrateAccountIdSize>
+KeyringService::GetPolkadotPubKey(const mojom::AccountIdPtr& account_id) {
+  CHECK(account_id);
+  auto* keyring = this->GetKeyring<PolkadotKeyring>(account_id->keyring_id);
```
cypt4: This should be checked, i believe

supermassive: ~this->~
Also please check keyring for nullptr

components/brave_wallet/browser/polkadot/polkadot_wallet_service.cc
```diff
@@ -38,4 +39,14 @@ void PolkadotWalletService::GetNetworkName(mojom::AccountIdPtr account_id,
                                        std::move(callback));
 }
 
+void PolkadotWalletService::GetAccountBalance(
```
cypt4, line 42: I think chain_id should be given as a method argument, similar way to the Eth, to be able to fetch different balances for Substrate networks

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.h
cypt4: Actually Polkadot is closer to Ethereum, where there are a lot of chains and they can be added manually

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.cc
cmazakas: kFallback is the default value used by the polkadot-js implementations, which I'm treating as a canonical implementation. I should've included a link where I derived it from. I've updated the code to include that.

I like using the string here because it means in the future if we ever need to update our code for the different versions of Substrate runtimes, we can just pass in the string and not having to touch all the different data members.

Though I will say in this case, it's pretty obvious we should just return a struct filled with nothing but zeroes, save for the `flags` field.

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.h
cmazakas: I'm not overwhelmingly experienced in this domain so I'll implement whatever you two decide upon. No matter what, we'll eventually learn what the correct choice is/was, ha ha.

components/brave_wallet/common/brave_wallet.mojom
cmazakas: This was something I debated a lot.

I wound up asking Gemini and this is where I got the idea from. The AI-generated reasoning was that string parsing was always a potential security factor because it's a string coming through an opaque source. I'm not sure I found that super compelling.

But JS does have a BigInt type, so all the frontend has to do is do:
```js
(high << BigInt(64) | low
```
because `uint64` in mojom gets translated to a BigInt automatically for us.

It also makes consuming the data easier on the C++ side as well, because if we receive the data from a mojom pipe, we don't need any string parsing and can just use normal bitshifts and bitwise or. Imo, it was nicer to replace the requirement for hex encoding/decoding.

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.cc
cmazakas: This is done!

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.cc
cmazakas: Good catch. Fixed!

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.cc
cmazakas: Yeah, this was a dramatically better way of doing this. Good tip, thank you.

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.cc
cmazakas: Good catch. I updated the unit tests.

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.cc
cmazakas: I could've sworn I had previously tried that but it didn't compile. This works fine. Done!

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc_responses.idl
cmazakas: Done! Let me know if the links don't provide enough information.

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.cc
cmazakas: Done!

components/brave_wallet/browser/keyring_service.cc
cmazakas: Done!

cmazakas: Done!

components/brave_wallet/browser/polkadot/polkadot_wallet_service.cc
cmazakas: Done!

components/brave_wallet/browser/keyring_service.cc
```diff
@@ -2599,6 +2599,17 @@ KeyringService::SignCip30MessageByCardanoKeyring(
   return result;
 }
 
+std::array<uint8_t, kPolkadotSubstrateAccountIdSize>
+KeyringService::GetPolkadotPubKey(const mojom::AccountIdPtr& account_id) {
+  CHECK(account_id);
+
+  auto* keyring = GetKeyring<PolkadotKeyring>(account_id->keyring_id);
+  CHECK(keyring);
```
cypt4: Sorry, i was meaning that result should be optional here like in other similar places (KeyringService::GetBitcoinAddress for ex)

components/brave_wallet/browser/polkadot/polkadot_utils.h
```diff
@@ -0,0 +1,17 @@
...
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#ifndef BRAVE_COMPONENTS_BRAVE_WALLET_COMMON_POLKADOT_UTILS_H_
+#define BRAVE_COMPONENTS_BRAVE_WALLET_COMMON_POLKADOT_UTILS_H_
+
+namespace brave_wallet {
+
+// See definition for "path": ["sp_core", "crypto", "AccountId32"]
+// https://raw.githubusercontent.com/polkadot-js/api/refs/heads/master/packages/types-support/src/metadata/v16/substrate-types.json
+inline constexpr const size_t kPolkadotSubstrateAccountIdSize = 32;
```
cypt4, line 13: Will it be used outside /browser/? "common" directory is used to share between renderer and browser. 
I know there are a lot of unrelated files for now in the brave_wallet/comon/ directory but anyway please
consider moving to /browser/ instead

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.cc
```diff
@@ -105,7 +254,75 @@ void PolkadotSubstrateRpc::OnGetChainName(GetChainNameCallback callback,
...
+
+  if (!res) {
+    return std::move(callback).Run(nullptr, WalletParsingErrorMessage());
+  }
+
+  if (res->error) {
+    if (res->error->message) {
+      return std::move(callback).Run(nullptr, res->error->message.value());
+    }
+    return std::move(callback).Run(nullptr, WalletInternalErrorMessage());
```
cypt4, line 319: Shouldn't it be also a parsing error?

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc_unittest.cc
```diff
@@ -128,4 +128,253 @@ TEST_F(PolkadotSubstrateRpcUnitTest, GetChainName) {
...
+    polkadot_substrate_rpc_->GetAccountBalance(chain_id, pubkey,
+                                               future.GetCallback());
+
+    auto [account_info, error] = future.Take();
+    EXPECT_EQ(error, WalletParsingErrorMessage());
+    EXPECT_FALSE(account_info);
+  }
+
+  {
+    // contains invalid hex in account info
```
cypt4: Please follow common comments format - starts with capital and ends with .

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.cc
cmazakas: I realize, I don't think it should be. Because in this case, we're getting an error back from the RPC service itself. It means we (wallet) made an invalid RPC call, either because we passed an ill-formed storage key or something like that. This is a valid JSON response as far as the spec of the RPC is concerned.

components/brave_wallet/browser/keyring_service.cc
cmazakas: Done! We now return `std::nullopt` when the `keyring` pointer is null,  because it'll be automatically null when the Polkadot feature is disabled.

components/brave_wallet/browser/polkadot/polkadot_utils.h
cmazakas: It probably won't be used outside browser, so I moved it to `browser/polkadot/` and it seems like `npm run gn_check` is happy with that, which was my original motivation anyway.

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc_unittest.cc
cmazakas: Done!

components/brave_wallet/browser/polkadot/polkadot_wallet_service.cc
```diff
@@ -43,8 +43,9 @@ void PolkadotWalletService::GetAccountBalance(
     const std::string& chain_id,
     GetAccountBalanceCallback callback) {
   auto pubkey = keyring_service_->GetPolkadotPubKey(account_id);
+  CHECK(pubkey);
```
cypt4: Run callback with an error here

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.cc
```diff
@@ -40,6 +45,151 @@ net::NetworkTrafficAnnotationTag GetNetworkTrafficAnnotationTag() {
     )");
 }
 
+mojom::PolkadotAccountInfoPtr ParseAccountInfoAsHex(std::string_view sv) {
+  // leading 0x, 160 hex chars worth of data
```
cypt4: Comment

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.cc
```diff
@@ -40,6 +45,151 @@ net::NetworkTrafficAnnotationTag GetNetworkTrafficAnnotationTag() {
...
+  if (sv.size() != 162) {
+    return nullptr;
+  }
+
+  if (!sv.starts_with("0x")) {
+    return nullptr;
+  }
+
+  std::vector<uint8_t> bytes;
+  sv.remove_prefix(2);  // remove leading 0x
```
cypt4: Comment

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.cc
```diff
@@ -40,6 +45,151 @@ net::NetworkTrafficAnnotationTag GetNetworkTrafficAnnotationTag() {
...
+
+  account->data->flags = mojom::uint128::New(high, low);
+
+  return account;
+}
+
+mojom::PolkadotAccountInfoPtr ParseAccountInfoFromJson(
+    const std::optional<
+        std::vector<polkadot_substrate_rpc_responses::AccountInfo>>& result) {
+  // see: `"id": 3` for more information
```
cypt4: Please check other comments too

cypt4 (changes requested):

components/brave_wallet/browser/polkadot/polkadot_wallet_service.cc
cmazakas: I wasn't exactly sure what error string to use here so I invented one. Let me know if it's up to par.

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.cc
cmazakas: Done!

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.cc
cmazakas: Done!

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.cc
cmazakas: I did a comb through and hopefully we're all good now!

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc_unittest.cc
```diff
@@ -128,4 +128,254 @@ TEST_F(PolkadotSubstrateRpcUnitTest, GetChainName) {
...
+    polkadot_substrate_rpc_->GetAccountBalance(chain_id, pubkey,
+                                               future.GetCallback());
+
+    auto [account_info, error] = future.Take();
+    EXPECT_EQ(error, WalletParsingErrorMessage());
+    EXPECT_FALSE(account_info);
+  }
+
+  {
+    url_loader_factory_.AddResponse(
```
cypt4: Please also add comment why this and next test are expecting false account_info

cmazakas: Done!

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.h
supermassive: > Actually Polkadot is closer to Ethereum, where there are a lot of chains and they can be added manually

That doesn't mean Ethereum RPC is done in the right way :)
But OK, let's leave chain_id in args by now. Might be addresses in some follow up

components/brave_wallet/common/brave_wallet.mojom
supermassive: Let's refactor this later when/if we need

components/brave_wallet/browser/keyring_service.cc
```diff
@@ -2599,6 +2599,19 @@ KeyringService::SignCip30MessageByCardanoKeyring(
   return result;
 }
 
+std::optional<std::array<uint8_t, kPolkadotSubstrateAccountIdSize>>
+KeyringService::GetPolkadotPubKey(const mojom::AccountIdPtr& account_id) {
+  CHECK(account_id);
+
+  auto* keyring = GetKeyring<PolkadotKeyring>(account_id->keyring_id);
+  if (!keyring) {
+    return std::nullopt;
```
supermassive, line 2608: nit: there also could be GetKeyring specialization so `GetKeyring<PolkadotKeyring>(account_id)` works and checks if account is of polkdat type

supermassive: Ok to leave it as is

components/brave_wallet/browser/polkadot/polkadot_wallet_service.cc
```diff
@@ -38,4 +38,18 @@ void PolkadotWalletService::GetNetworkName(mojom::AccountIdPtr account_id,
...
 
+void PolkadotWalletService::GetAccountBalance(
+    mojom::AccountIdPtr account_id,
+    const std::string& chain_id,
+    GetAccountBalanceCallback callback) {
+  auto pubkey = keyring_service_->GetPolkadotPubKey(account_id);
+  if (!pubkey) {
+    return std::move(callback).Run(
+        nullptr, "Cannot get Polkadot public key for this account.");
+  }
```
supermassive, line 49: Let's use generic `WalletInternalErrorMessage()` error message

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc_unittest.cc
```diff
@@ -128,4 +128,262 @@ TEST_F(PolkadotSubstrateRpcUnitTest, GetChainName) {
...
+        "{\"id\":1,\"jsonrpc\":\"2.0\",\"result\":[{\"block\":"
+        "\"0x1bcd3e074b91ef25740714dc63671f4a36d2781ff93877ef9ef31b849d1ad69c\""
+        ",\"changes\":[["
+        "\"0x26aa394eea5630e07c48ae0c9558cef7b99d880ec681799c0cf30e8886371da9de"
+        "1e86a9a8c739864cf3cc5ec2bea59fd43593c715fdd31c61141abd04a99fd6822c8558"
+        "854ccde39a5684e7a56da27d\",null]]}]}");
+
+    polkadot_substrate_rpc_->GetAccountBalance(chain_id, pubkey,
+                                               future.GetCallback());
+
```
supermassive: nit: prefer raw string literals for test jsons like
```
  const std::string utxo_json = R"([
    {
      "amount": [{
        "quantity": 406560,
        "unit": "lovelace"
      }]
    }
  ])";
```

components/brave_wallet/common/brave_wallet.mojom
cmazakas: Alright, sounds good to me. I wanted to also keep the C++ side relatively simple as well, and just one shift seemed fine for recreating the u128 vs converting it from a string repr. On JS, it's pretty trivial to do `BigInt(0x....)`.

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.cc
cmazakas: Alright, I addressed this by relocating the size and `0x` checks outside the function and then we decode the hex string into a stack-local array, passing in a `base::span<const uint8_t>` now. 

This should be fully resolved.

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.cc
cmazakas: I'm going to mark this as resolved unless there's any further concerns!

components/brave_wallet/browser/polkadot/polkadot_wallet_service.cc
cmazakas: I worked this out with Anton in another review comment.

components/brave_wallet/browser/keyring_service.cc
cmazakas: I'll mark this as resolved then. I'll probably touch this up at some point in the future.

components/brave_wallet/browser/polkadot/polkadot_wallet_service.cc
cmazakas: Done!

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc_unittest.cc
cmazakas: I didn't realize we could ignore presubmit warnings. Using these raw string literals is dramatically nicer.

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.cc
```diff
@@ -40,6 +45,153 @@ net::NetworkTrafficAnnotationTag GetNetworkTrafficAnnotationTag() {
...
+  if (!reader.ReadU64LittleEndian(low)) {
+    return nullptr;
+  }
+
+  if (!reader.ReadU64LittleEndian(high)) {
+    return nullptr;
+  }
+
+  account->data->free = mojom::uint128::New(high, low);
+
```
supermassive: To avoid tempororaies:

```
  account->data->free = mojom::uint128::New();

  if (!reader.ReadU64LittleEndian(account->data->free->low)) {
    return nullptr;
  }

  if (!reader.ReadU64LittleEndian(account->data->free->high)) {
    return nullptr;
  }
```

supermassive: Or even some anonymous function to avoid copypaste:
```
bool ReadU128(base::SpanReader<const uint8_t>& reader, mojom::uint128Ptr& out) {
  auto val = mojom::uint128::New();

  if (!reader.ReadU64LittleEndian(val->low)) {
    return false;  
  }  

  if (!reader.ReadU64LittleEndian(val->high)) {  
    return false;  
  }  

  out = std::move(val);
  return true;
}
```

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.cc
```diff
@@ -40,6 +45,153 @@ net::NetworkTrafficAnnotationTag GetNetworkTrafficAnnotationTag() {
...
+    return nullptr;
+  }
+
+  if (!reader.ReadU64LittleEndian(high)) {
+    return nullptr;
+  }
+
+  account->data->flags = mojom::uint128::New(high, low);
+
+  return account;
```
supermassive, line 122: We usually add such here
`DCHECK_EQ(reader.remaining(), 0u);`
to make sure every byte was meaningful.

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc_unittest.cc
```diff
@@ -128,4 +128,317 @@ TEST_F(PolkadotSubstrateRpcUnitTest, GetChainName) {
...
+                 account_info->data->reserved->low) == reserved);
+
+    uint128_t frozen = std::numeric_limits<uint128_t>::max();
+    EXPECT_TRUE((uint128_t{account_info->data->frozen->high} << 64 |
+                 account_info->data->frozen->low) == frozen);
+
+    uint128_t flags = std::numeric_limits<uint128_t>::max();
+    EXPECT_TRUE((uint128_t{account_info->data->flags->high} << 64 |
+                 account_info->data->flags->low) == flags);
+  }
```
supermassive, line 441: For mojom::uint128 it would be actually quite natural to have implicit conversion to native uint128_t.
https://chromium.googlesource.com/chromium/src/+/master/mojo/public/cpp/bindings/README.md#type-mapping

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.cc
cmazakas: This was a good idea. Done!

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc_unittest.cc
cmazakas: I've added this as a CodeHealth item here: https://github.com/brave/brave-browser/issues/50066

components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.cc
cmazakas: Good idea. Done!

PR #31608, opened 2025-10-03
[AI Chat]: Add mojom interface for attaching URLs to conversations

fallaciousreasoning:
Resolves https://github.com/brave/brave-browser/issues/49864


components/ai_chat/core/common/mojom/ai_chat.mojom
```diff
@@ -551,6 +551,9 @@ interface AIChatUIHandler {
   // Associates an open tab with a conversation.
   AssociateTab(TabData tab, string conversation_uuid);
 
+  // Associates content from a URL with a conversation.
+  AssociateUrlContent(url.mojom.Url url, string title, string conversation_uuid);
```
petemill, line 555: So is this instead of something like `AssociateHistoryItem` or `AssociateBookmark`? If so, we don't need to preserve (for UI) that the URL is one of those things?

components/ai_chat/core/browser/ai_chat_service.cc
```diff
@@ -1191,6 +1199,25 @@ void AIChatService::MaybeAssociateContent(
...
+    const std::string& conversation_uuid) {
+  CHECK(delegate);
+
+  auto* conversation = GetConversation(conversation_uuid);
+  if (!conversation) {
+    return;
+  }
+
+  // Don't associate the content if it isn't allowed.
+  if (!CanAssociateContent(delegate.get())) {
```
petemill, line 1212: Can we unit test trying to associate bad/non-allowed content?

browser/ui/webui/ai_chat/ai_chat_ui_page_handler_unittest.cc
```diff
@@ -0,0 +1,134 @@
...
+  // Should have attached shields observer and ephemeral storage tab helper
+  EXPECT_TRUE(brave_shields::BraveShieldsWebContentsObserver::FromWebContents(
+      associated_link_content->GetWebContentsForTesting()));
+  EXPECT_TRUE(ephemeral_storage::EphemeralStorageTabHelper::FromWebContents(
+      associated_link_content->GetWebContentsForTesting()));
+
+  // EXPECT_EQ(associated_link_content->GetWebContentsForTesting()->GetURL(),
+  //           test_url);
+  // EXPECT_EQ(associated_link_content->GetWebContentsForTesting()->GetTitle(),
+  //           title);
```
petemill: still needed?

components/ai_chat/core/browser/ai_chat_service.cc
petemill: nevermind, I see you have that covered 1 layer up in `AIChatUIPageHandlerTest, AssociateUrlContent_InvalidScheme`.

petemill (approved):
Seems good as long as the interface function name / UX note isn't a concern

petemill:
GN issue:
```
> gn gen /home/ubuntu/workspace/core-build-pr-linux-x64_PR-31608/src/out/Component --check

[2025-10-03T00:32:07.196Z] ERROR at //brave/components/ai_chat/core/browser/ai_chat_service.h:26:11: Include not allowed.

[2025-10-03T00:32:07.196Z] #include "brave/components/ai_chat/content/browser/associated_url_content.h"

[2025-10-03T00:32:07.197Z]           ^----------------------------------------------------------------

[2025-10-03T00:32:07.197Z] It is not in any dependency of

[2025-10-03T00:32:07.197Z]   //brave/components/ai_chat/core/browser:browser

[2025-10-03T00:32:07.197Z] The include file is in the target(s):

[2025-10-03T00:32:07.197Z]   //brave/components/ai_chat/content/browser:browser

[2025-10-03T00:32:07.197Z] which should somehow be reachable.
```

You can't depend on `content/browser` from `core/browser`, so you'll need to use an interface.

components/ai_chat/core/common/mojom/ai_chat.mojom
fallaciousreasoning: For now, we don't have any reason to distinguish - they both show in the UI the same way, and I think there's no point treating a history entry as separate to a bookmark entry with the same URL (as they'll have the same content)

fallaciousreasoning:
Ah yeah, that shouldn't have even been there!

browser/ui/webui/ai_chat/ai_chat_ui_page_handler.h
```diff
@@ -77,6 +78,9 @@ class AIChatUIPageHandler : public mojom::AIChatUIHandler,
       override;
   void AssociateTab(mojom::TabDataPtr tab,
                     const std::string& conversation_uuid) override;
+  void AssociateUrlContent(const GURL& url,
+                           const std::string& title,
+                           const std::string& conversation_uuid) override;
```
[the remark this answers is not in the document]
fallaciousreasoning: Marking as resolved to unblock merge, will reopen when merged

PR #31635, opened 2025-10-03
Android widget scroll

SergeyZhukovsky:
Resolves https://github.com/brave/brave-browser/issues/49890

[Screen_recording_20251003_133135.webm](https://github.com/user-attachments/assets/37284e58-dd6b-48e2-8f46-5ea23ffe0894)

Test Plan:
Make sure quick action search widget is working and scroll is working in case of 4+ fav tiles.


samartnik (approved):
lgtm

android/java/org/chromium/chrome/browser/widget/quickactionsearchandbookmark/BookmarkRemoteViewsFactory.java
```diff
@@ -0,0 +1,234 @@
...
+        ThreadUtils.runOnUiThreadBlocking(
+                () -> {
+                    try {
+                        // Widgets are top-level entry points similar to Activities, so accessing
+                        // the last used profile here is acceptable since there's no existing
+                        // Profile context to pass through the RemoteViewsFactory chain. The same
+                        // approach is used in upstream Chromium's bookmark widget.
+                        LargeIconBridge largeIconBridge =
+                                new LargeIconBridge(
+                                        ProfileManager.getLastUsedRegularProfile()); // nocheck
```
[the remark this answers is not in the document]
SergeyZhukovsky: upstream's implementation is inside `BookmarkWidgetServiceImpl.java`

browser/ui/android/strings/android_brave_strings.grd
```diff
@@ -3047,10 +3047,7 @@ If you don't accept this request, VPN will not reconnect and your internet conne
       Top Visited Site
     </message>
     <message name="IDS_SEARCH_WITH_BRAVE" desc="String used in QuickActionSearchAndBookmarkWidget's search bar">
-      Search with Brave
-    </message>
-    <message name="IDS_SEARCH_WITH_SEARCH_ENGINE" desc="Dynamic string used in QuickActionSearchAndBookmarkWidget's search bar. Placeholder is replaced by default search engine.">
-      Search with <ph name="DEFAULT_SEARCH_ENGINE">%1$s</ph>
+      Search
```
SergeyZhukovsky, line 3053: we decided to just do `Search` as `Search with Brave` doesn't fit and doesn't look good on some small screens.

vadimstruts (approved):
LGTM

android/java/org/chromium/chrome/browser/settings/BraveSearchEngineUtils.java
```diff
@@ -69,19 +67,18 @@ private static void initializeDSEPrefs(Profile profile) {
         // At first run, we should set initial default prefs to each standard/private DSE prefs.
         // Those pref values will be used until user change DES options explicitly.
         final String notInitialized = "notInitialized";
-        if (notInitialized.equals(ContextUtils.getAppSharedPreferences().getString(
-                    BraveSearchEngineAdapter.STANDARD_DSE_SHORTNAME, notInitialized))) {
+        SharedPreferencesManager sharedPreferences = ChromeSharedPreferences.getInstance();
+        if (notInitialized.equals(
```
Olipro, line 71: A minor nit if I'm understanding this correctly:

`notInitialized` represents the string that will be returned as a default if no real defined value exists.

Since it's a literal, it will be interned and you can safely use `==` to compare the references - which is of course also faster than invoking `.equals`

That said, maybe our linter isn't smart enough to know when we're checking for an interned string, but I imagine there's a way around this.

In any case, it can be left as-is or changed, feel free to self-resolve this comment no matter what you choose.

android/java/org/chromium/chrome/browser/widget/quickactionsearchandbookmark/BookmarkRemoteViewsFactory.java
```diff
@@ -0,0 +1,234 @@
...
+    }
+
+    @Override
+    public int getCount() {
+        return mWidgetTiles.size();
+    }
+
+    @Override
+    public @Nullable RemoteViews getViewAt(int position) {
+        if (position >= mWidgetTiles.size()) {
```
Olipro: Since we care about overflow, would it hurt to also check if they've passed in a negative number?

android/java/org/chromium/chrome/browser/widget/quickactionsearchandbookmark/BookmarkRemoteViewsFactory.java
```diff
@@ -0,0 +1,234 @@
...
+    private void loadFaviconForTile(RemoteViews rv, WidgetTile tile) {
+        GURL gurl = tile.getGURL();
+        if (gurl == null || !gurl.isValid()) {
+            return;
+        }
+
+        String url = gurl.getSpec();
+
+        // Check cache first
+        if (mIconCache.containsKey(url)) {
```
Olipro: We're doing a double lookup here by checking for the key and then calling `get`

Since `get` will return null if the key isn't found, we can assign it to an lvalue and this conditional can be reduced to a `null` check.

android/java/org/chromium/chrome/browser/widget/quickactionsearchandbookmark/BookmarkRemoteViewsFactory.java
```diff
@@ -0,0 +1,234 @@
...
+            rv.setImageViewBitmap(R.id.bookmark_icon, mIconCache.get(url));
+            return;
+        }
+
+        // Load icon on UI thread with timeout
+        // Use arrays to allow modification inside lambda (Java closure requirement)
+        final @Nullable Bitmap[] iconResult = new Bitmap[1];
+        final int[] fallbackColorResult = new int[1];
```
Olipro: Since this runs on a different thread and there's otherwise nothing special about these variables, I suspect Java may technically be allowed to assume later on in the code that these were never modified. Though maybe the `CountDownLatch` prevents that in Java's memory model, I'm not 100% sure.

In any case, it's more idiomatic (and guaranteed to be a barrier) if we turn these into `AtomicReference<T>`

android/java/org/chromium/chrome/browser/widget/quickactionsearchandbookmark/BookmarkRemoteViewsFactory.java
```diff
@@ -0,0 +1,234 @@
...
+    @Override
+    public int getViewTypeCount() {
+        // All items use the same layout (widget_bookmark_grid_item.xml)
+        return 1;
+    }
+
+    @Override
+    public long getItemId(int position) {
+        // Use URL hash as stable identifier for better update handling
+        if (position < mWidgetTiles.size()) {
```
Olipro: Ditto https://github.com/brave/brave-core/pull/31635/files#r2406130423

android/java/org/chromium/chrome/browser/settings/BraveSearchEngineUtils.java
SergeyZhukovsky: So `==` would work when the preference isn't set (same reference), but if someone actually saved the string `notInitialized` as a value, it would be a different instance and `==` would fail! I would play safe and don't use `==` here. The code isn't used in a loop or where performance is really critical. As I mentioned I'm not 100% sure we didn't save an actual `notInitialized` in the past.

android/java/org/chromium/chrome/browser/widget/quickactionsearchandbookmark/BookmarkRemoteViewsFactory.java
simoarpe: @SergeyZhukovsky, I would change this into something like:

```java
final Bitmap cachedIcon =  mIconCache.get(url);
if (cachedIcon != null) {
    rv.setImageViewBitmap(R.id.bookmark_icon, cachedIcon);
    return;
}
// ...
```

At least you don't hit the fetching twice.

android/java/org/chromium/chrome/browser/settings/BraveSearchEngineUtils.java
Olipro: Ah, I think I've misunderstood the intention of `notInitialized` in this case then.

As I see it, we have 3 possibilities and perhaps we only care about some of them:

1. The pref doesn't have a value and it simply returns our interned `notInitialized` literal.
2. The pref does have a value and for whatever reason, it's set to "notInitialized"
3. The pref does have a value and it's something valid.

If we use `==` then we are explicitly checking for our unique literal (assuming that the prefs code doesn't call `.intern()` anywhere)

So, with that said, is it a legitimate case that we want to allow the pref to be set to `"notInitialized"` and have it behave the same way as when we're getting our own interned literal back? This sounds problematic in that the pref essentially represents a hidden tri-state where we have a magic value that mirrors it not being set.

android/java/org/chromium/chrome/browser/widget/quickactionsearchandbookmark/BookmarkRemoteViewsFactory.java
```diff
@@ -0,0 +1,234 @@
...
+        final CountDownLatch latch = new CountDownLatch(1);
+
+        // Background Thread          UI Thread
+        //     |                         |
+        //     |--runOnUiThreadBlocking--|
+        //     |    (wait at latch)      |
+        //     |                    Load Icon
+        //     |                    countDown()
+        //     |<--latch released--------|
+        // Continue...
```
simoarpe, line 158: Not a huge fan of countdown latches, but widgets tend to be problematic on Android, so I would say, as long as it does not create UI stuttering, to me is fine. 
In case this will create issues in the future or will require a restyling I would lean more towards a solution that loads a placeholder first, and then silently swap it with the actual icon.
But, again, take this just as a suggestion for the future.
If the UI results in a scrolling _smooth_  enough I'm totally okay with this 👍

simoarpe (approved):
👍 Changes look good.

Added just one small nit, and some food for thought for the future.

android/java/org/chromium/chrome/browser/settings/BraveSearchEngineUtils.java
SergeyZhukovsky: not sure about that. That code is pretty old and I would avoid modifying it in the current issue, I just fixed presubmit warnings.

android/java/org/chromium/chrome/browser/widget/quickactionsearchandbookmark/BookmarkRemoteViewsFactory.java
SergeyZhukovsky: yeah, I implemented an approach with placeholders first but after I tried it on bunch of low end emulators I never saw them actually appear, so it just overcomplicated logic via adding more code and really not needed. I decided to remove them and use as is. The icons are already there btw when the scroll is active, they just need to be loaded on a widget init, that's is basically done in the background in most cases when the device boots up

android/java/org/chromium/chrome/browser/widget/quickactionsearchandbookmark/BookmarkRemoteViewsFactory.java
SergeyZhukovsky: added https://github.com/brave/brave-core/pull/31635/commits/e3c9df970f87f1dc8b9de8c1c41cb2f39ecabc71

android/java/org/chromium/chrome/browser/widget/quickactionsearchandbookmark/BookmarkRemoteViewsFactory.java
SergeyZhukovsky: fixed https://github.com/brave/brave-core/pull/31635/commits/e3c9df970f87f1dc8b9de8c1c41cb2f39ecabc71

android/java/org/chromium/chrome/browser/widget/quickactionsearchandbookmark/BookmarkRemoteViewsFactory.java
SergeyZhukovsky: changed https://github.com/brave/brave-core/pull/31635/commits/e3c9df970f87f1dc8b9de8c1c41cb2f39ecabc71

android/java/org/chromium/chrome/browser/widget/quickactionsearchandbookmark/BookmarkRemoteViewsFactory.java
SergeyZhukovsky: added https://github.com/brave/brave-core/pull/31635/commits/e3c9df970f87f1dc8b9de8c1c41cb2f39ecabc71

mkarolin (approved):
`strings`++

android/java/org/chromium/chrome/browser/widget/quickactionsearchandbookmark/BookmarkRemoteViewsFactory.java
```diff
@@ -0,0 +1,239 @@
...
+        ThreadUtils.runOnUiThreadBlocking(
```
Olipro: Just noticed this - `runOnUiThreadBlocking` will result in us halting execution here until the lambda has returned.

So... the only asynchronous part of this operation is the `LargeIconCallback callback` object.

That raises a few questions:

1. What is it that requires us to run this code on the UI thread?
2. If we _do_ need to post it to the UI thread, what is the purpose of doing it in a blocking manner when we have a synchronisation object anyway?
3. Why do we need a countdown latch when what we seem to be doing is checking for success/failure which could be represented by an (atomic) boolean?

SergeyZhukovsky:
```
new LargeIconBridge(
                                            ProfileManager.getLastUsedRegularProfile())
```
requires to be executed on UI thread. Yeah, you are right, I changed it to `ThreadUtils.runOnUiThread`. I don't see any pros of using atomic boolean vs countdown latch that could be returned earlier and we just use a generated icon in that case.
`runOnUiThreadBlocking-->runOnUiThread` is done here https://github.com/brave/brave-core/pull/31635/commits/f0367c13f99eb32634887c89fb942d0685655c9b

android/java/org/chromium/chrome/browser/widget/quickactionsearchandbookmark/BookmarkRemoteViewsFactory.java
```diff
@@ -0,0 +1,255 @@
...
+            return;
+        }
+
+        // Load icon on UI thread with timeout
+        // Use AtomicReference for thread-safe access across UI and background threads
+        final AtomicReference<@Nullable Bitmap> iconResult = new AtomicReference<>();
+        final AtomicReference<Integer> fallbackColorResult = new AtomicReference<>();
+        // CountDownLatch provides both signaling (countDown) and blocking (await) mechanisms,
+        // which is cleaner than pairing an AtomicBoolean with Object.wait/notify
+        final CountDownLatch latch = new CountDownLatch(1);
```
Olipro, line 147: Resurrecting the previous discussion since the comment became outdated:

To my mind, the latch doesn't convey intent quite as clearly compared to an AtomicBoolean that we can check and sleep on - something like:

`if (!theAtomicBool.get()) Thread.Sleep(ICON_LOAD_TIMEOUT_MS);` strikes me as more obvious since the CountDownLatch isn't actually being used for any sort of countdown.

With that said, I could equally be bikeshedding over whether you want six of one or half a dozen of the other. That is to say, feel free to self-resolve this, whatever you decide.

android/java/org/chromium/chrome/browser/widget/quickactionsearchandbookmark/BookmarkRemoteViewsFactory.java
bridiver: This code does need to run on the UI thread, but yes this seems concerning and it's not clear to me why we need this vs calling `PostTask.runOrPostTask` with a callback? Not sure what thread this is running on, but we really should avoid blocking any thread like this.
```
Note that non-test usage of this function is heavily discouraged. For non-tests, use callbacks rather than blocking threads.
```

android/java/org/chromium/chrome/browser/widget/quickactionsearchandbookmark/BookmarkRemoteViewsFactory.java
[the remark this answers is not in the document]
bridiver: `getLastUsedRegularProfile` seems ok here since upstream is doing the same, but I echo @Olipro's concern above with `runOnUiThreadBlocking`

android/java/org/chromium/chrome/browser/widget/quickactionsearchandbookmark/BookmarkRemoteViewsFactory.java
Olipro: Agreed. From a design standpoint, if we were able to do the follow-up work _inside_ the executed callback, that would eliminate the need for any sort of blocking - however, it seems that as it currently stands, it wants to get the icon ASAP, synchronously. If we need to merge this urgently, then maybe there could be a follow-up to refactor it?

SergeyZhukovsky: `getViewAt()` is called on a binder thread (not UI thread). We need to block that binder thread until the icon loads on the UI thread. We need to block because `getViewAt()` is a synchronous method that must return a fully-populated `RemoteViews` object immediately.

SergeyZhukovsky: exactly the same approach is used inside upstream's widget `BookmarkWidgetServiceImpl.java`. The block binder thread uses a countdown latch https://github.com/chromium/chromium/blob/141.0.7390.64/chrome/android/java/src/org/chromium/chrome/browser/bookmarkswidget/BookmarkWidgetServiceImpl.java#L375 and https://github.com/chromium/chromium/blob/141.0.7390.64/chrome/android/java/src/org/chromium/chrome/browser/bookmarkswidget/BookmarkWidgetServiceImpl.java#L412

kjozwiak:
Verification **PASSED** on `Pixel 8 Pro` running `Android 16` using the following build(s):

```
Brave | 1.85.79 Chromium: 142.0.7444.34 (Official Build) canary (64-bit)
--- | ---
Revision | 20d691ca9892a31848e55f1cab5917222f4a31ee
OS | Android 16; Build/BP3A.251005.004.B1; 36; REL
```

Using the STR/Cases and information outlined via https://github.com/brave/brave-core/pull/31635#issue-3482139687, ensured that you can scroll through the current favourites via the `Quick Action` widget without any issues/failures as per the following:

* Video Example of the above working as expected ---> https://youtube.com/shorts/04Xiqkp76qE

--------------------------

Verification **PASSED** on `Samsung S25 Ultra` running `Android 16` using the following build(s):

```
Brave | 1.85.53 Chromium: 141.0.7390.70 (Official Build) canary (64-bit)
--- | ---
Revision | f891b4e9a5183471b5ad804a8430d4b3b8c9985e
OS | Android 16; Build/BP2A.250605.031.A3; 36; REL
```

Using the STR/Cases and information outlined via https://github.com/brave/brave-core/pull/31635#issue-3482139687, ensured that you can scroll through the current favourites via the `Quick Action` widget without any issues/failures as per the following:

https://github.com/user-attachments/assets/1916949c-656a-4977-b8a9-d8c4cce56e43

PR #31637, opened 2025-10-03
refactor code to use higher-order function and switch

cmazakas:
Resolves https://github.com/brave/brave-browser/issues/49018

Update NetworkManager methods to handle the supplied CoinType with a switch, so that a new coin type being added and unhandled results in compilation failure with `-Werror`. Also update the code to use a function pointer to increase code clarity and reduce duplication.


supermassive (approved):

cdesouza-chromium (approved):
LGTM

PR #31640, opened 2025-10-03
Modify `Managed` icon tooltip text for settings

aseren:
PR changes `Managed` icon tooltip to cover a case when setting is managed by Brave policy.

Resolves

### DIFF
--- a/app/generated_resources.grd
+++ b/app/generated_resources.grd
@@ -12831,7 +12831,7 @@ Check your passwords anytime in <ph name="GOOGLE_PASSWORD_MANAGER">$1<ex>Brave P
         </message>
 
         <message name="IDS_CONTROLLED_SETTING_POLICY" desc="Text displayed in the controlled settings bubble when a setting's value is managed by policy.">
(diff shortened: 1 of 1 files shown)


app/generated_resources.grd
```diff
@@ -12907,7 +12907,7 @@ Check your passwords anytime in <ph name="GOOGLE_PASSWORD_MANAGER">$1<ex>Brave P
         </message>
 
         <message name="IDS_CONTROLLED_SETTING_POLICY" desc="Text displayed in the controlled settings bubble when a setting's value is managed by policy.">
-          This setting is managed by your administrator.
+          This setting is managed by your administrator or by another Brave policy.
```
mkarolin: This file is get auto-replaced with every bump. We can't change strings in here. Typically you'd add our own string in brave_generated_resources.grd and then have an override for where this string ID is used to replace upstream string ID with our own.

aseren: Fixed, thanks for comment!

PR #31643, opened 2025-10-04
Correctly handle Leo enabled state transitions

DJAndries:
Resolves https://github.com/brave/brave-browser/issues/49899

### DIFF
--- a/browser/ai_chat/BUILD.gn
+++ b/browser/ai_chat/BUILD.gn
@@ -195,6 +195,38 @@ if (enable_brave_ai_chat_agent_profile) {
   }
 }
 
+source_set("enabled_state_transition") {
+  sources = [
+    "enabled_state_transition_service.cc",
+    "enabled_state_transition_service.h",
+    "enabled_state_transition_service_factory.cc",
+    "enabled_state_transition_service_factory.h",
+  ]
+
+  deps = [
+    "//brave/browser/ai_chat",
+    "//brave/components/ai_chat/core/browser",
+    "//brave/components/ai_chat/core/common",
+    "//brave/components/constants",
+    "//chrome/browser/profiles:profile",
+    "//components/keyed_service/content",
+    "//components/prefs",
+    "//components/user_prefs",
+    "//content/public/browser",
+  ]
+
+  if (!is_android) {
+    deps += [
+      "//brave/browser/ui/sidebar",
+      "//brave/components/sidebar/browser",
+      "//chrome/browser/ui",
+      "//chrome/browser/ui:browser_list",
+      "//chrome/browser/ui/tabs:tabs_public",
+      "//chrome/browser/ui/views/side_panel",
+    ]
+  }
+}
+
 source_set("unit_tests") {
   testonly = true
   sources = [
@@ -262,12 +294,14 @@ source_set("browser_tests") {
       "ai_chat_render_view_context_menu_browsertest.cc",
       "ai_chat_sidepanel_behavior_browsertest.cc",
       "ai_chat_ui_browsertest.cc",
+      "enabled_state_transition_service_browsertest.cc",
       "page_content_fetcher_browsertest.cc",
     ]
     deps = [
       "//brave/app:generate_command_metadata",
       "//brave/browser/ai_chat",
       "//brave/browser/ai_chat:ai_chat_agent_profile",
+      "//brave/browser/ai_chat:enabled_state_transition",
       "//brave/browser/misc_metrics",
       "//brave/browser/ui/sidebar",
       "//brave/components/ai_chat/content/browser",
--- a/browser/ai_chat/enabled_state_transition_service.cc
+++ b/browser/ai_chat/enabled_state_transition_service.cc
@@ -1,0 +1,118 @@
+// Copyright (c) 2025 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#include "brave/browser/ai_chat/enabled_state_transition_service.h"
+
+#include "base/check.h"
+#include "brave/components/ai_chat/core/browser/utils.h"
+#include "brave/components/ai_chat/core/common/pref_names.h"
+#include "brave/components/constants/webui_url_constants.h"
+#include "chrome/browser/profiles/profile.h"
+#include "components/user_prefs/user_prefs.h"
+#include "content/public/browser/browser_context.h"
+#include "content/public/browser/web_ui_data_source.h"
+
+#if !BUILDFLAG(IS_ANDROID)
+#include <algorithm>
+
+#include "brave/browser/ui/sidebar/sidebar_service_factory.h"
+#include "brave/components/sidebar/browser/sidebar_service.h"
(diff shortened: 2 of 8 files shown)


browser/ai_chat/enabled_state_transition_service.cc
```diff
@@ -0,0 +1,118 @@
...
+#include "chrome/browser/ui/tabs/public/tab_features.h"
+#include "chrome/browser/ui/tabs/tab_strip_model.h"
+#include "chrome/browser/ui/views/side_panel/side_panel_registry.h"
+#include "content/public/browser/web_contents.h"
+#include "content/public/common/url_constants.h"
+#endif
+
+namespace ai_chat {
+
+EnabledStateTransitionService::EnabledStateTransitionService(
```
bridiver, line 33: This approach seems odd to me, shouldn't these things already be listening for pref changes and updating themselves?

bridiver: in particular
`Leo sidebar panel should be closed for all tabs`
`Leo sidebar icon should be hidden`
should definitely be handled internally. Still looking at the tab stuff

browser/ai_chat/enabled_state_transition_service.cc
```diff
@@ -0,0 +1,118 @@
...
+          &EnabledStateTransitionService::OnEnabledByPolicyChanged,
+          base::Unretained(this)));
+}
+
+EnabledStateTransitionService::~EnabledStateTransitionService() = default;
+
+void EnabledStateTransitionService::OnEnabledByPolicyChanged() {
+  const bool enabled = IsAIChatEnabled(prefs_);
+  if (!enabled) {
+    content::WebUIDataSource::CreateAndAdd(profile_, kAIChatUIHost);
```
bridiver, line 55: why are you adding a WebUIDataSource here?

browser/ai_chat/enabled_state_transition_service.cc
```diff
@@ -0,0 +1,118 @@
...
+#endif
+  }
+#if !BUILDFLAG(IS_ANDROID)
+  UpdateSidebarState(enabled);
+#endif
+}
+
+#if !BUILDFLAG(IS_ANDROID)
+void EnabledStateTransitionService::CloseAIChatTabs() {
+  for (Browser* browser : *BrowserList::GetInstance()) {
```
bridiver, line 67: `GetAllBrowserWindowInterfaces` should be used intead of BrowserList and also works on android

browser/ai_chat/enabled_state_transition_service.cc
```diff
@@ -0,0 +1,118 @@
...
+
+void EnabledStateTransitionService::OnEnabledByPolicyChanged() {
+  const bool enabled = IsAIChatEnabled(prefs_);
+  if (!enabled) {
+    content::WebUIDataSource::CreateAndAdd(profile_, kAIChatUIHost);
+#if !BUILDFLAG(IS_ANDROID)
+    CloseAIChatTabs();
+#endif
+  }
+#if !BUILDFLAG(IS_ANDROID)
```
bridiver, line 60: sidebar is guarded by toolkit_views, not is_android

browser/ai_chat/enabled_state_transition_service.cc
```diff
@@ -0,0 +1,118 @@
...
+#include "brave/components/ai_chat/core/browser/utils.h"
+#include "brave/components/ai_chat/core/common/pref_names.h"
+#include "brave/components/constants/webui_url_constants.h"
+#include "chrome/browser/profiles/profile.h"
+#include "components/user_prefs/user_prefs.h"
+#include "content/public/browser/browser_context.h"
+#include "content/public/browser/web_ui_data_source.h"
+
+#if !BUILDFLAG(IS_ANDROID)
+#include <algorithm>
```
bridiver, line 18: most of these includes do not belong behind `!IS_ANDROID` guard. Guards in gn and cpp should always match.

browser/ai_chat/BUILD.gn
```diff
@@ -195,6 +195,38 @@ if (enable_brave_ai_chat_agent_profile) {
...
+    "//brave/components/ai_chat/core/common",
+    "//brave/components/constants",
+    "//chrome/browser/profiles:profile",
+    "//components/keyed_service/content",
+    "//components/prefs",
+    "//components/user_prefs",
+    "//content/public/browser",
+  ]
+
+  if (!is_android) {
```
bridiver, line 218: some of these are android and some are toolkit_views. Please make sure buildflag guards are consistent

browser/ai_chat/enabled_state_transition_service.h
```diff
@@ -0,0 +1,55 @@
...
+
+#if !BUILDFLAG(IS_ANDROID)
+namespace sidebar {
+class SidebarService;
+}
+#endif
+
+namespace ai_chat {
+
+class EnabledStateTransitionService : public KeyedService {
```
bridiver, line 29: I would not create a separate KeyedService for this because you need to keep it in sync with AIChatService in terms of which profiles it is enabled for, but also you should be using ProfileKeyedService for things in brave/browser

browser/ai_chat/enabled_state_transition_service.cc
```diff
@@ -0,0 +1,118 @@
...
+#include "content/public/common/url_constants.h"
+#endif
+
+namespace ai_chat {
+
+EnabledStateTransitionService::EnabledStateTransitionService(
+    content::BrowserContext* context)
+    : profile_(Profile::FromBrowserContext(context)),
+#if !BUILDFLAG(IS_ANDROID)
+      sidebar_service_(sidebar::SidebarServiceFactory::GetForProfile(profile_)),
```
bridiver, line 37: use dependency injection

browser/ai_chat/enabled_state_transition_service.cc
```diff
@@ -0,0 +1,118 @@
...
+#endif
+      prefs_(user_prefs::UserPrefs::Get(context)) {
```
bridiver, line 39: use dependency injection

browser/ai_chat/enabled_state_transition_service.cc
```diff
@@ -0,0 +1,118 @@
...
+#include "chrome/browser/ui/views/side_panel/side_panel_registry.h"
+#include "content/public/browser/web_contents.h"
+#include "content/public/common/url_constants.h"
+#endif
+
+namespace ai_chat {
+
+EnabledStateTransitionService::EnabledStateTransitionService(
+    content::BrowserContext* context)
+    : profile_(Profile::FromBrowserContext(context)),
```
bridiver, line 35: why are you passing in BrowserContext and converting to Profile for something that is in brave/browser?

browser/ai_chat/enabled_state_transition_service.h
bridiver: if you use dependency injection you can move this class to components/ai_chat and pass it into AIChatService in the factory

bridiver: this class also needs comments

browser/ai_chat/enabled_state_transition_service.cc
bridiver: I also feel like in general having a separate service to handle toggling the pref is not a good idea. If we are going to be able to toggle this on and off, the aichat code needs to be aware of it directly.

PR #31647, opened 2025-10-06
Support signing tagged Cardano transactions

supermassive:
Resolves https://github.com/brave/brave-browser/issues/49933

### DIFF
--- a/components/brave_wallet/browser/internal/cardano_tx_decoder.rs
+++ b/components/brave_wallet/browser/internal/cardano_tx_decoder.rs
@@ -223,9 +223,10 @@ fn extract_inputs(
     body_map: &[(CborValue, CborValue)],
 ) -> Result<Vec<CxxRestoredCardanoInput>, Error> {
     let inputs_value = find_map_value(body_map, INPUTS_KEY)?;
-    let inputs_array = match inputs_value {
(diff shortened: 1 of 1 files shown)


components/brave_wallet/browser/internal/cardano_tx_decoder.rs
```diff
@@ -226,6 +225,10 @@ fn extract_inputs(
     let inputs_value = find_map_value(body_map, INPUTS_KEY)?;
     let inputs_array = match inputs_value {
         Some(CborValue::Array(arr)) => arr,
+        Some(CborValue::Tag(SET_TAG, boxed)) => match &**boxed {
```
cmazakas: Nit: the `&**boxed` kinda threw me for a bit.

Maybe something like:
```rs
    let inputs_array: &Vec<_> = match inputs_value {
        Some(CborValue::Array(arr)) => arr,
        Some(CborValue::Tag(SET_TAG, ref boxed)) => match &**boxed {
```
is a bit more readable at a glance what this code is doing? It wasn't inherently obvious to me without some type annotations here that we were getting a borrow in `inputs_array`.

supermassive: Fixed!

cypt4 (approved):

PR #31660, opened 2025-10-06
[iOS][Leo] Add Gemma Model Copies

nuo-xu:
Resolves https://github.com/brave/brave-browser/issues/49767

IMG_8884
IMG_8885


Uni-verse:
Verified on `iPhone 15` running `iOS 18.6.2` using version `1.85 (33)`

- ensured that copy is updated in the chat tool tip for `Gemma 3 12 B`
- ensured that the copy text matches the criteria mentioned in https://github.com/brave/brave-browser/issues/48329#issue-3312107825 like other models

IMG_0359

PR #31661, opened 2025-10-06
[AI Chat]: Wire up Smart Mode to Front End

Douglashdaniel:
## Description 

Wires up the `Smart Mode` feature to the Front-end UI

- Add the ability to save a `Smart Mode` from the human message menu
- Add the ability to select a `Smart Mode` from the `Tools menu`
- Adds the ability to `Edit/Delete` a `Smart Mode`

Resolves <https://github.com/brave/brave-browser/issues/49772>

## Test Plan:
1. Start a new conversation with `Leo AI` and prompt a question you want to save as a `Smart Mode`
2. After you ask the `Question` open the human message `Menu` and click `Save as Smart Mode`
3. Give your `Smart Mode` a shortcut name and click save
4. Now in the `Input Box` start the conversation with a `/` and the tools menu should popup
5. Select your `Smart Mode` and test
6. Enter a `/` in the input box again to open the `Tools Menu`
7. Hover over your `Smart Mode` and click the `Edit` button
8. Test editing and `Saving` your `Smart Mode`
9. Do this again and test `Deleting` your `Smart Mode`

https://github.com/user-attachments/assets/07dd9399-2a50-48af-b248-23b37a48bab9


components/ai_chat/resources/page/components/main/index.tsx
```diff
@@ -217,6 +217,54 @@ function Main() {
...
+  // Transform smartModes into ActionGroup format and append to actionList
+  const categoriesWithSmartModes = React.useMemo(() => {
+    if (!aiChatContext.smartModes || aiChatContext.smartModes.length === 0) {
+      return aiChatContext.actionList
+    }
+
+    const smartModeEntries = aiChatContext.smartModes.map((smartMode) => ({
+      subheading: undefined,
+      details: {
+        label: `${smartMode.shortcut} id:${smartMode.id}`,
```
fallaciousreasoning: Hey, I think rather doing this we should either:
1. Add an id property to this struct
2. Make the type a `SmartMode | Action`

I think needing to parse the id from a string isn't ideal

components/ai_chat/resources/page/components/main/index.tsx
```diff
@@ -217,6 +217,54 @@ function Main() {
...
+        type: Mojom.ActionType.SMART_MODE,
```
fallaciousreasoning: If we make this a `SmartMode | Action` then we won't need to introduce this new type on the mojom enum (which is only used on the frontend)

components/ai_chat/resources/page/components/main/index.tsx
```diff
@@ -217,6 +217,54 @@ function Main() {
...
+      category: getLocale(S.CHAT_UI_SMART_MODES_GROUP),
+      entries: smartModeEntries,
+    }
+
+    return [smartModeGroup, ...aiChatContext.actionList]
+  }, [aiChatContext.actionList, aiChatContext.smartModes])
+
+  const handleToolsMenuClick = React.useCallback(
+    (type: Mojom.ActionType, id?: string) => {
+      const smartMode = aiChatContext.smartModes.find(
```
fallaciousreasoning: here for example, you could test for a property which only exists on `Action` or `SmartMode` instead of needing the new enum type (though you'll need to change the signature to take a `Action | SmartMode`)

components/ai_chat/resources/page/components/input_box/index.tsx
```diff
@@ -89,10 +94,58 @@ function usePlaceholderText(
...
+        const currentModeShortcut = `/${props.context.selectedSmartMode.shortcut}`
+        if (!newValue.startsWith(currentModeShortcut)) {
+          props.context.resetSelectedSmartMode()
+        }
+        return
+      }
+      if (newValue.startsWith('/') && props.context.smartModes.length > 0) {
+        // No smart mode selected, but input starts with /, try to detect
+        detectAndSetSmartMode(newValue)
+      }
```
fallaciousreasoning, line 141: wait, so the selected smart mode is constantly updated while you're typing? You don't need to select it from the menu?

components/ai_chat/resources/page/components/main/index.tsx
Douglashdaniel: Switched to type `ActionEntry | SmartMode`!

components/ai_chat/resources/page/components/main/index.tsx
Douglashdaniel: This is removed, actually I looks like the mojom enum was recently removed so this would not have worked any more 😅

components/ai_chat/resources/page/components/main/index.tsx
Douglashdaniel: Done!

fallaciousreasoning: probably moved to `common.mojom`, I think!

components/ai_chat/resources/page/components/filter_menu/tools_menu.test.tsx
```diff
@@ -97,4 +115,15 @@ describe('tools_menu', () => {
     expect(queryByText('category 1')).not.toBeInTheDocument()
     expect(queryByText('category 2')).toBeInTheDocument()
   })
+
+  it('should show smart modes when filtering', () => {
+    const { getByText } = render(
+      <ToolsMenu
+        {...defaultProps}
```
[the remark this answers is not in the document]
fallaciousreasoning: not an issue here, this is a test. Maybe worth an exclusion for .test.tsx files?

components/ai_chat/resources/page/components/input_box/index.tsx
```diff
@@ -89,10 +94,58 @@ function usePlaceholderText(
...
+    },
+    [
+      props.context.selectedSmartMode,
+      props.context.smartModes,
+      detectAndSetSmartMode,
+    ],
+  )
```
fallaciousreasoning, line 148: do we need this? Won't this be handled by selecting something in the menu?

components/ai_chat/resources/page/components/filter_menu/tools_menu.tsx
```diff
@@ -4,21 +4,44 @@
...
+const getIsActionEntry = (item: ExtendedActionEntry): item is ActionEntry => {
+  return 'details' in item && item.details !== undefined
+}
 
-function matchesQuery(query: string, entry: ActionEntry) {
-  return entry.details && matches(query, entry.details.label)
+export const getIsSmartMode = (
+  item: ExtendedActionEntry,
+): item is SmartMode => {
+  return 'id' in item && 'shortcut' in item
```
fallaciousreasoning: optional: probably okay to just test one of these

components/ai_chat/resources/page/components/filter_menu/tools_menu.test.tsx
```diff
@@ -97,4 +115,15 @@ describe('tools_menu', () => {
     expect(queryByText('category 1')).not.toBeInTheDocument()
     expect(queryByText('category 2')).toBeInTheDocument()
   })
+
+  it('should show smart modes when filtering', () => {
+    const { getByText } = render(
```
fallaciousreasoning, line 120: could we add a test for firing the click handler on these? Or does that not play nice with the test harness

fallaciousreasoning: and maybe an edit button click test too?

fallaciousreasoning: all good if its not easy to set up with jest - I know some of the webcomponents stuff is a bit painful

components/ai_chat/resources/page/components/filter_menu/tools_menu.tsx
```diff
@@ -4,21 +4,44 @@
...
+}
+
+function matchesQuery(query: string, entry: ExtendedActionEntry) {
+  if (getIsActionEntry(entry)) {
+    return matches(query, entry.details!.label)
+  } else if (getIsSmartMode(entry)) {
```
fallaciousreasoning: I think its probably okay to just have this as an `else` - Typescript should infer its a SmartMode entry!

components/ai_chat/resources/page/components/input_box/index.tsx
yrliou: IMO users should be free to type the shortcut directly without needing to choose from the menu, I vaguely recalled it was also the same in Dia though it's been a while since I tried.

Douglashdaniel: When the `SmartMode` is selected, it will be highlighted in the `InputBox`, so we also need a way to unselect it when the user backspaces the input, otherwise there is no way to unselect it.

components/ai_chat/resources/page/components/filter_menu/tools_menu.tsx
Douglashdaniel: Done!

components/ai_chat/resources/page/components/filter_menu/tools_menu.test.tsx
Douglashdaniel: Tests added for `handleClick` and `handleEditClick`

components/ai_chat/resources/page/components/filter_menu/tools_menu.test.tsx
```diff
@@ -97,4 +115,111 @@ describe('tools_menu', () => {
...
+      })
+
+      expect(handleClick).toHaveBeenCalledTimes(2)
+    })
+
+    it('should call handleEditClick when clicking edit', () => {
+      const handleEditClick = jest.fn()
+      const { container } = render(
+        <ToolsMenu
+          {...defaultProps}
```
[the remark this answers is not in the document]
Douglashdaniel: not an issue here, this is a test.

components/ai_chat/resources/page/components/filter_menu/tools_menu.test.tsx
```diff
@@ -97,4 +115,111 @@ describe('tools_menu', () => {
...
+      })
+
+      expect(handleClick).toHaveBeenCalledTimes(2)
+    })
+
+    it('should call handleClick when clicking on smart modes', () => {
+      const handleClick = jest.fn()
+      const { getByText } = render(
+        <ToolsMenu
+          {...defaultProps}
```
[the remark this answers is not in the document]
Douglashdaniel: not an issue here, this is a test.

components/ai_chat/resources/page/components/filter_menu/tools_menu.test.tsx
```diff
@@ -97,4 +115,111 @@ describe('tools_menu', () => {
...
+    expect(getByText('smart mode 1')).toBeInTheDocument()
+    expect(getByText('smart mode 2')).toBeInTheDocument()
+  })
+
+  describe('tools menu click handling', () => {
+    it('should call handleClick when clicking on action entries', () => {
+      const handleClick = jest.fn()
+      const { getByText } = render(
+        <ToolsMenu
+          {...defaultProps}
```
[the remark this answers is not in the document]
Douglashdaniel: not an issue here, this is a test.

components/ai_chat/resources/page/components/input_box/index.tsx
fallaciousreasoning: Okay sweet, seems reasonable enough!

PR #31667, opened 2025-10-06
RichNTT iOS

aseren:
Resolves https://github.com/brave/brave-browser/issues/43514


components/constants/webui_url_constants.h
```diff
@@ -102,6 +103,9 @@ inline constexpr char kBraveUINewTabURL[] = "chrome://newtab/";
 inline constexpr char kBraveUINewTabTakeoverURL[] =
     "chrome://new-tab-takeover/";
 
+inline constexpr char kBraveUINewTabTakeoverPageURL[] =
+    "chrome://new-tab-takeover-page/";
```
aseren, line 108: For Android we use `chrome://new-tab-takeover` url for WebUI and 
`chrome-untrusted://new-tab-takeover` url for data source which serves NTT html files.

As I understand, this is not possible on iOS because different data sources on iOS cannot have the same host, but different schemes. That's why new url is needed.

iOS team to confirm.

ios/brave-ios/Sources/Brave/Frontend/Browser/New Tab Page/Backgrounds/NewTabPageBackgroundView.swift
```diff
@@ -18,10 +19,28 @@ class NewTabPageBackgroundView: UIView {
 
   let playerLayer = AVPlayerLayer()
 
+  var webUIController: ChromeWebUIController?
```
aseren: we are using Chrome Web UI to show NTTs in the background (HTML pages)

ios/browser/ui/webui/brave_web_ui_controller_factory.mm
```diff
@@ -94,6 +101,15 @@ WebUIIOSFactoryFunction GetWebUIIOSFactoryFunction(const GURL& url) {
 BraveWebUIControllerFactory::CreateWebUIIOSControllerForURL(
     WebUIIOS* web_ui,
     const GURL& url) const {
+  const std::string url_host = url.host();
+  if (url_host == kNewTabTakeoverHost) {
+    return std::make_unique<NTPSponsoredRichMediaController>(
```
aseren, line 134: `NTPSponsoredRichMediaController` serves HTML files from NTT component updater.

ios/browser/ui/webui/brave_web_ui_controller_factory.mm
```diff
@@ -94,6 +101,15 @@ WebUIIOSFactoryFunction GetWebUIIOSFactoryFunction(const GURL& url) {
...
+        web_ui, url, ntp_background_images_service_);
+  } else if (url_host == kNewTabTakeoverPageHost) {
+    return std::make_unique<NewTabTakeoverUIIOS>(
```
aseren, line 137: `NewTabTakeoverUIIOS` shows webUI, which:
* Display NTT HTML page in an untrusted iFrame with strict CSP
* handle communication between browser and the iFrame

ios/browser/ui/webui/new_tab_takeover_ui/new_tab_takeover_ui_handler.cc
```diff
@@ -0,0 +1,73 @@
...
+    return failed();
+  }
+
+  std::move(callback).Run(creative->url,
+                          creative->should_metrics_fallback_to_p3a,
+                          GURL(creative->logo.destination_url));
+}
+
+void NewTabTakeoverUIHandler::NavigateToUrl(const GURL& url) {
+  // TODO(aseren): Implement Navigation.
```
aseren, line 72: what would be the best way to do trigger a navigation for iOS tab from here?

cc iOS team

ios/app/brave_profile_controller.mm
```diff
@@ -153,12 +154,17 @@ - (instancetype)initWithProfileKeepAlive:
...
             _profile->GetOffTheRecordProfile());
 
+    auto service =
+        std::make_unique<ntp_background_images::NTPBackgroundImagesService>(
+            GetApplicationContext()->GetVariationsService(),
+            GetApplicationContext()->GetComponentUpdateService(),
+            GetApplicationContext()->GetLocalState());
+
+    BraveWebUIControllerFactory::GetInstance()->SetNTPBackgroundImagesService(
+        service.get());
```
aseren, line 162: Didn't find a better way of passing NTPBackgroundImagesService to WebUI.

components/new_tab_takeover/new_tab_takeover.html
```diff
@@ -6,9 +6,14 @@
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <meta name="referrer" content="no-referrer">
   <title>New Tab Takeover</title>
+  <if expr="is_ios">
```
kylehickinson, line 9: These are never processed currently, you'll need to run the html through the expression preprocessor then update the grdp to copy it from the generated folder instead. Right now these scripts will attempt to load on all platform

ios/brave-ios/Sources/Brave/Frontend/Browser/New Tab Page/Backgrounds/NewTabPageBackgroundView.swift
```diff
@@ -18,10 +19,28 @@ class NewTabPageBackgroundView: UIView {
...
+  func setupRichNewTabTakeoverLayer(
+    braveProileController: BraveProfileController,
+    richNewTabTakeoverURL: URL
+  ) {
+    webUIController = ChromeWebUIController(
+      braveCore: braveProileController,
+      isPrivateBrowsing: false
+    ).then {
+      $0.webView.load(URLRequest(url: richNewTabTakeoverURL))
+      addSubview($0.view)
```
kylehickinson: In this current version of the code there will literally be no way to interact with it because its on the background under a scroll view

ios/browser/ui/webui/brave_web_ui_controller_factory.mm
```diff
@@ -83,6 +85,11 @@ WebUIIOSFactoryFunction GetWebUIIOSFactoryFunction(const GURL& url) {
     return NSURLErrorNotConnectedToInternet;
   }
 
+  const std::string url_host = url.host();
+  if (url_host == kNewTabTakeoverHost || url_host == kNewTabTakeoverPageHost) {
+    return 0;
+  }
+
```
kylehickinson: Why?

ios/browser/ui/webui/brave_web_ui_controller_factory.mm
kylehickinson: These should all be added to `GetWebUIIOSFactoryFunction` not `CreateWebUIIOSControllerForURL`

ios/browser/ui/webui/brave_web_ui_controller_factory.mm
```diff
@@ -104,6 +120,11 @@ WebUIIOSFactoryFunction GetWebUIIOSFactoryFunction(const GURL& url) {
   return (*function)(web_ui, url);
 }
 
+void BraveWebUIControllerFactory::SetNTPBackgroundImagesService(
+    ntp_background_images::NTPBackgroundImagesService* service) {
+  ntp_background_images_service_ = service;
+}
```
kylehickinson, line 154: Definitely shouldn't be here

ios/browser/ui/webui/new_tab_takeover_ui/new_tab_takeover_ui_handler.cc
kylehickinson: Currently there is no pattern for doing so, but we're working on it to support AI Chat/Wallet WebUIs

ios/browser/ui/webui/brave_web_ui_controller_factory.mm
aseren: Thanks for a comment! the problem is that `NewTabTakeoverUIIOS` and `NTPSponsoredRichMediaController` controllers take NTPBackgroundImagesService as an argument, but `GetWebUIIOSFactoryFunction` does not allow this for now.
I will try to find a way of how to make `GetWebUIIOSFactoryFunction` to pass additional args.

ios/browser/ui/webui/new_tab_takeover_ui/new_tab_takeover_ui_handler.cc
aseren: gotcha, thanks!

components/new_tab_takeover/new_tab_takeover.html
aseren: will do thanks!

ios/browser/ui/webui/brave_web_ui_controller_factory.mm
aseren: Thanks for a comment! That was done because `GetWebUIIOSFactoryFunction` doesn't allow to pass additional args. I will try to find how to do it

ios/browser/ui/webui/brave_web_ui_controller_factory.mm
aseren: will try to find another way of passing this, thanks!

ios/brave-ios/Sources/Brave/Frontend/Browser/New Tab Page/Backgrounds/NewTabPageBackgroundView.swift
aseren: yeah, clicks and interaction is supposed to be added in follow-up PRs.

aseren:
@kylehickinson , thanks for comments and for initial PR review. We will try to address all comments in the current or follow-up PRs.
Sorry for instructions to setup RichNTT component didn't work. It sometimes hard to set it up, due to component id can be different depending on current device region or ip address if rewards is on.

> * actual usage of this should be feature flagged and off by default
> * this implementation has no logic around where to adjust elements in the WebUI based on where existing elements are in the native UI
> * this implementation doesn't seem to have any plan for supporting interaction, since the current native NTP already contains a scrollable container for Brave News
> * this implementation doesn't offload NTP WebUI's when NTPs are removed from the view hierarchy so memory increases are permanent per session until the tab is closed
> * the current PR isn't testable other than to see WebUI does load the empty page so hard to know what other problems will arise

kylehickinson:
Zoom should be disabled for the WebUI somehow, otherwise it can break scrolling. Also I don't see the web view's `isScrollEnabled = false` on the BraveWebView's scrollView

https://github.com/user-attachments/assets/4c55e5c0-64db-4245-84e2-adf50b6abc7b

kylehickinson:
We should ensure touch exclusivity for gestures so that you can't scroll while actively dragging on the NTP background

https://github.com/user-attachments/assets/7fc285bf-19e0-4867-b72b-f75f3d8a1f74

kylehickinson:
There's a noticable gap in resource load time that we need to fix, the time it takes for the image to load is too noticable 

https://github.com/user-attachments/assets/30b6feff-1d3e-4857-b426-a3eafa389403

kylehickinson:
Other minor issues:

You can long-press the entire web view and it'll select the whole thing, showing the standard edit menu
selection

kylehickinson:
We'll need to blur the background and disable interaction when you scroll into brave news, otherwise you can end up accidentally tapping or interacting with NTP elements while in News content

news

https://github.com/user-attachments/assets/2889cd74-20cb-4ff1-b45b-8c1fc585bae3

aseren:
Thanks for checking and finding RichNTT iOS issues. They will be fixed via following issues:
https://github.com/brave/brave-browser/issues/50455
https://github.com/brave/brave-browser/issues/50456

aseren:
Closed as we are no longer proceeding with this functionality.

PR #31679, opened 2025-10-07
[AI Chat] Automatic model supports tools

petemill:
I've left this as a content flag for now because of the description in the issue
> So that in content agent conversations, the automatic model can be used to route to a capable model that can also use the content agent tools.
> For regular conversations, we might want to update the server routing so that it ignores tools unless the content agent capability is provided, otherwise automatic model chat conversations will never use non-tools capable models.

- Resolves https://github.com/brave/brave-browser/issues/49965

### DIFF
--- a/components/ai_chat/core/common/features.h
+++ b/components/ai_chat/core/common/features.h
@@ -38,14 +38,21 @@ extern const base::FeatureParam<double> kAITemperature;
 COMPONENT_EXPORT(AI_CHAT_COMMON)
 extern const base::FeatureParam<size_t> kMaxCountLargeToolUseEvents;
 
-// Whether should add indentation to page content structure for tool results.
-COMPONENT_EXPORT(AI_CHAT_COMMON)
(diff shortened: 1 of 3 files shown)


components/ai_chat/core/common/features.h
```diff
@@ -38,15 +38,22 @@ extern const base::FeatureParam<double> kAITemperature;
 COMPONENT_EXPORT(AI_CHAT_COMMON)
 extern const base::FeatureParam<size_t> kMaxCountLargeToolUseEvents;
 
-// Whether should add indentation to page content structure for tool results.
-COMPONENT_EXPORT(AI_CHAT_COMMON)
-extern const base::FeatureParam<bool> kShouldIndentPageContentBlocks;
-
```
petemill, line 44: I moved this lower so that the "large tool use events" related flags are grouped.

yrliou (approved):

PR #31680, opened 2025-10-07
[Android] Top sites text needs to have bottom padding

SergeyZhukovsky:
Adds bottom padding for top sites text on NTP. It is regressed after adding `mv_tiles_layout_stub` in https://github.com/brave/brave-core/commit/3a31b31511d73b9370201cb841cf354f58f0f3a1. We can't use margin anymore, the more appropriate way is to use padding.
Resolves: https://github.com/brave/brave-browser/issues/49500


samartnik (approved):
++

emerick (approved):

AlexeyBarabash (approved):
++ thanks

PR #31681, opened 2025-10-07
[Wallet]: New Confirm Cancel Transaction Panel

Douglashdaniel:
## Description 

Builds out the new `Confirm Cancel Transaction` panel

Resolves <https://github.com/brave/brave-browser/issues/49719>

Screenshot 2 | Screenshot 3
--|--

## Test Plan:
1. Create and confirm a `Slow` ETH transaction
2. When the `Transaction Submitted` panel is displayed
3. Click `Cancel Transaction` and create a `Cancel` transaction
4. The new `Confirm Cancel Transaction` panel should popup
5. Confirm the transaction and make sure it completes.

https://github.com/user-attachments/assets/a081163e-8b47-4116-b2c3-da9c31a1f696


components/brave_wallet_ui/components/extension/cancel_speedup_transaction/cancel_speedup_transaction.tsx
```diff
@@ -0,0 +1,178 @@
...
+  Card,
+  Title,
+  PriorityLabel,
+  FeeButton,
+  InfoText,
+  InfoIcon,
+} from './cancel_speedup_transaction.style'
+import { Column, Row } from '../../shared/style'
+
+export function CancelSpeedupTransaction() {
```
onyb, line 46: Why is this called `CancelSpeedupTransaction` and not just `CancelTransaction`?

onyb, line 46: We should display the pending transaction being cancelled, otherwise it's difficult to say which cancel transaction is cancelling what, when there are multiple ones.

components/brave_wallet_ui/utils/tx-utils.ts
```diff
@@ -1809,6 +1809,63 @@ export const parseTransactionWithoutPrices = ({
...
+  } else if (ethTxData) {
+    // Legacy transaction
+    value = ethTxData.value || '0'
+    data = ethTxData.data || []
+    toAddress = ethTxData.to || ''
+  }
+
+  const fromAddress = tx.fromAccountId?.address || ''
+
+  // Cancel transactions have:
```
onyb, line 1872: See my previous comment about showing the transaction being cancelled. There's another criteria that we should include here - the cancel transaction must have the same nonce as that of the pending transaction(s) being cancelled.

components/brave_wallet_ui/components/extension/cancel_speedup_transaction/cancel_speedup_transaction.tsx
Douglashdaniel: Hey, so the UI for `Speedup` is going to be the same so we are making this a shared component.
Speedup will be handled in https://github.com/brave/brave-browser/issues/49720

Douglashdaniel: Talked with @kleantzogu and agreed to show `Transaction to be canceled` with the intent string. I'm going to be handling parsing `Cancel` transaction in intent via https://github.com/brave/brave-browser/issues/49972 so lets add this then.

Screenshot 58

components/brave_wallet_ui/utils/tx-utils.ts
Douglashdaniel: Glad you caught this, I've added an additional check for `nonce` in https://github.com/brave/brave-core/compare/6fd9cc8cce6bdf7a7af443eb4d33e4b99fd6ea9e..47075982f128967e80e5f63253cd3558cb9d8e6c

onyb (approved):
LGTM!

PR #31692, opened 2025-10-07
[NTP Next] Display the default image if backgrounds are not ready

zenparsing:
Resolves https://github.com/brave/brave-browser/issues/49980

Screenshot 2025-10-07 at 6 10 38 PM

### DIFF
--- a/browser/resources/brave_new_tab_page_refresh/state/background_handler.ts
+++ b/browser/resources/brave_new_tab_page_refresh/state/background_handler.ts
@@ -8,7 +8,16 @@ import { SponsoredRichMediaAdEventHandler } from 'gen/brave/components/ntp_backg
 import { NewTabPageProxy } from './new_tab_page_proxy'
 import { Store } from '../lib/store'
 import { debounce } from '$web-common/debounce'
-import { BackgroundState, BackgroundActions } from './background_state'
+import { BackgroundState, BackgroundActions, BraveBackground } from './background_state'
+
+// A pre-loaded background image resource that can be used if a new profile
+// opens the NTP before the NTPBackgroundImagesService has finished loading
+// the current collection of Brave backgrounds.
+const defaultBraveBackground: BraveBackground = {
+  imageUrl: 'dylan-malval_sea-min.webp',
+  author: 'Dylan Malval',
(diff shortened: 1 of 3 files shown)


browser/resources/brave_new_tab_page_refresh/state/background_handler.ts
```diff
@@ -39,7 +49,9 @@ export function createBackgroundHandler(
 
   async function updateBraveBackgrounds() {
     const { backgrounds } = await handler.getBraveBackgrounds()
-    store.update({ braveBackgrounds: backgrounds })
+    if (backgrounds.length > 0) {
+      store.update({ braveBackgrounds: backgrounds })
+    }
```
fallaciousreasoning: instead of the check and the update above, would it be worth
```suggestion
    store.update({ braveBackgrounds: backgrounds.length ? backgrounds : [defaultBraveBackground] })
```

fallaciousreasoning (approved):
lgtm, but its missing ~~a comment~~ an issue

browser/resources/brave_new_tab_page_refresh/state/background_handler.ts
zenparsing: Yes - that's better. Thanks!

kjozwiak:

```
Brave | 1.85.38 Chromium: 141.0.7390.70 (Official Build) nightly (64-bit)
-- | --
Revision | 96d981769f3d4b89f839932413b07ebb0e47977f
OS | Windows 11 Version 25H2 (Build 26200.6718)
```

Reproduced the original issue using `1.85.33 Chromium: 141.0.7390.55` with the STR/Cases outlined via https://github.com/brave/brave-browser/issues/49980#issue-3493248736 as per the following:

**`Broken`** | **`Broken`**
---------------|----------------
Screenshot 2025-10-08 134235 | Screenshot 2025-10-08 134242

Using the same STR/Cases mentioned above, ensured that the issue was resolved while running through `1.85.38 Chromium: 141.0.7390.70`. Went through the STR/Cases ~10 times and didn't hit the regression/problem after the fix.

**`Fixed`** | **`Fixed`**
------------|----------------
Screenshot 2025-10-08 134453 | Screenshot 2025-10-08 134459

PR #31695, opened 2025-10-08
[webpack]: `output_module=true` implies `no_iife=true`

fallaciousreasoning:
Resolves https://github.com/brave/brave-browser/issues/49990

### DIFF
--- a/components/webpack/webpack.config.js
+++ b/components/webpack/webpack.config.js
@@ -78,6 +78,7 @@ module.exports = async function (env, argv) {
 
   if (env.output_module) {
     output.library = { type: 'module' }
+    output.iife = false
   }
 
   if (env.output_public_path) {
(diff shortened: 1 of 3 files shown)


fallaciousreasoning:
@jonathansampson just curious if you think it'd be possible to use `output_module` for the distiller stuff too? It'd be nice to get rid of one of the output formats we support

components/webpack/webpack.config.js
```diff
@@ -78,6 +78,7 @@ module.exports = async function (env, argv) {
 
   if (env.output_module) {
     output.library = { type: 'module' }
+    output.iife = false
```
szilardszaloki, line 81: Maybe throw an error when `env.output_module && !env.no_iife`?

fallaciousreasoning: I was considering that but `output.iife` considers the definedyness of `iife` which isn't necessarily set :/

PR #31700, opened 2025-10-08
[webpack]: Default `output_module=true` for new WebUIs

fallaciousreasoning:
Resolves https://github.com/brave/brave-browser/issues/49993

This will make the migration easier when upstream eventually removes `load_time_data_deprecated.js` which we have to use when not outputting a module.

Most of these WebUIs work fine with `output_module=true` but I wanted to land this without worrying about breaking existing pages.


zenparsing (approved):
Rewards, NTP refresh 👍

PR #31701, opened 2025-10-08
Email Aliases: Add application menu item

arthuredelstein:
Implement the Email Aliases menu item in the application menu at Hamburger menu > Passwords and Autofill > Email Aliases. This menu item is only visible when the EmailAliases flag is enabled.

When the user clicks on this item, the Email Aliases settings page opens at brave://settings/email-aliases

Similar to other menu items in the application menu, this menu item includes an icon for Email Aliases service.

Two browser tests:
- EmailAliasesMenuInsertion
- EmailAliasesOpensSettings

Resolves https://github.com/brave/brave-browser/issues/45539


arthuredelstein:
(To be clear: this PR has nothing to do with passwords, except it is adding an item in the passwords and autofill menu, as designed in https://www.figma.com/design/N0Xsctr5jt7I3Ots0G355V/%F0%9F%93%A7-Email-aliases?node-id=22-2900&t=7cIRBEDRIae7wBoh-0.)

components/constants/webui_url_constants.h
```diff
@@ -47,6 +47,8 @@ inline constexpr char kBraveUIWalletPageURL[] = "chrome://wallet/";
 inline constexpr char kWalletPageHost[] = "wallet";
 inline constexpr char kExtensionSettingsURL[] = "brave://settings/extensions";
 inline constexpr char kWalletSettingsURL[] = "brave://settings/wallet";
+inline constexpr char kEmailAliasesSettingsURL[] =
```
bridiver: this belongs in the email aliases component, this file is deprecated per https://github.com/brave/brave-core/blob/master/.github/CODEOWNERS#L273

browser/ui/brave_pages.cc
```diff
@@ -90,6 +90,10 @@ void ShowWalletSettings(Browser* browser) {
   ShowSingletonTabOverwritingNTP(browser, GURL(kWalletSettingsURL));
 }
 
+void ShowEmailAliases(Browser* browser) {
+  ShowSingletonTabOverwritingNTP(browser, GURL(kEmailAliasesSettingsURL));
+}
```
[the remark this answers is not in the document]
bridiver: these should all be updated together

browser/ui/toolbar/brave_app_menu_model.cc
```diff
@@ -127,6 +129,28 @@ void BraveAppMenuModel::Build() {
   }
 }
 
+void BraveAppMenuModel::InsertEmailAliasesItem() {
+  if (!base::FeatureList::IsEnabled(email_aliases::kEmailAliases)) {
```
bridiver: this should never get hit if email aliases is disabled, right? So it should CHECK, not silently return

browser/ui/toolbar/brave_app_menu_model.cc
```diff
@@ -127,6 +129,28 @@ void BraveAppMenuModel::Build() {
   }
 }
 
+void BraveAppMenuModel::InsertEmailAliasesItem() {
```
bridiver: other similar methods are named `Build*Section`

browser/ui/toolbar/brave_app_menu_model.cc
```diff
@@ -127,6 +129,28 @@ void BraveAppMenuModel::Build() {
...
+  if (!base::FeatureList::IsEnabled(email_aliases::kEmailAliases)) {
+    return;
+  }
+
+  auto* autofill_menu_model =
+      static_cast<ui::SimpleMenuModel*>(GetSubmenuModelAt(
+          GetIndexOfCommandId(IDC_PASSWORDS_AND_AUTOFILL_MENU).value()));
+  if (!autofill_menu_model) {
```
bridiver: same as above, this should CHECK

app/brave_command_ids.h
```diff
@@ -140,6 +140,9 @@
 
 #define IDC_SHOW_APPS_PAGE 56352
 
+// Email Aliases
+#define IDC_SHOW_EMAIL_ALIASES 56455
```
bridiver: why is this above lower values below? That's almost certainly going to lead to someone accidentally duplicating it

browser/ui/toolbar/brave_app_menu_model.cc
```diff
@@ -103,6 +103,9 @@ void BraveAppMenuModel::Build() {
   BuildBraveProductsSection();
   BuildBrowserSection();
   BuildMoreToolsSubMenu();
+  if (GetIndexOfCommandId(IDC_PASSWORDS_AND_AUTOFILL_MENU)) {
```
goodov: consider moving this condition inside `BuildPasswordsAndAutofillSubmenu()` call. It looks out of the line here.

PR #31702, opened 2025-10-08
Add dep crx to target create_dist_mac

mherrmann:
This lets us use `npm run create_dist` to build the crx for macOS. It's a preparation for [uploading Omaha 4 builds to the update server during releases](https://github.com/brave/devops/issues/13476).

Pure dev concern, no QA required.

### DIFF
--- a/build/mac/BUILD.gn
+++ b/build/mac/BUILD.gn
@@ -24,6 +24,8 @@ packaged_basename =
 
 dmg_path = "$packaged_basename.dmg"
 
+crx_private_key = "//brave/updater/crx-private-key.der"
+
 if (is_universal_binary) {
   assert(target_cpu == "arm64")
   x64_arch_app_path = "$root_out_dir/x64/$brave_exe"
@@ -401,16 +403,17 @@ group("create_dist_mac") {
       }
     }
   }
+  if (path_exists(crx_private_key)) {
+    deps += [ ":crx" ]
+  }
 }
 
-crx3("crx") {
-  inputs = [ dmg_path ]
-  output = "$packaged_basename.crx3"
-  base_dir = get_path_info(dmg_path, "dir")
-
-  # The private key file is not in Git for obvious reasons. It needs to be
-  # copied into this directory manually. Changing the private key requires
-  # also changing kBravePublisherKeyHash in crx_verifier.cc.
-  key = "//brave/updater/crx-private-key.der"
-  deps = [ ":package" ]
+if (path_exists(crx_private_key)) {
+  crx3("crx") {
+    inputs = [ dmg_path ]
+    output = "$packaged_basename.crx3"
+    base_dir = get_path_info(dmg_path, "dir")
+    key = crx_private_key
+    deps = [ ":package" ]
+  }
 }


build/mac/BUILD.gn
```diff
@@ -401,16 +403,17 @@ group("create_dist_mac") {
...
-  inputs = [ dmg_path ]
-  output = "$packaged_basename.crx3"
-  base_dir = get_path_info(dmg_path, "dir")
-
-  # The private key file is not in Git for obvious reasons. It needs to be
-  # copied into this directory manually. Changing the private key requires
-  # also changing kBravePublisherKeyHash in crx_verifier.cc.
-  key = "//brave/updater/crx-private-key.der"
-  deps = [ ":package" ]
+if (path_exists(crx_private_key)) {
```
goodov: this is a bit unusual approach in my opinion. For this to work you are required to re-run `gn gen` after putting a file at this location. There's no clear way to validate if `crx` target ever runs or not.

Usually you control GN targets with `args.gn`.

mherrmann: interesting; what would be a better approach?

goodov: > interesting; what would be a better approach?

The sensible approach to me is to make `crx_private_key` a GN argument and set it explicitly when needed. When it's set, the `crx` target should run and (supposedly) fail if the file is not found or invalid.

mherrmann: That makes much more sense, thank you for this input @goodov. I implemented your suggestion in bdbcdebafd2a9caadb7e4e1342a04817ffbf8a11. What do you think?

updater/.gitignore
```diff
@@ -1 +0,0 @@
-crx-private-key.der
```
goodov, line 1: should this file be removed entirely?

build/mac/BUILD.gn
goodov: Looks good.

build/mac/config.gni
```diff
@@ -23,6 +23,9 @@ declare_args() {
   # we have fully migrated to Omaha 4, we will remove this GN arg with a default
   # value of true.
   install_omaha4_with_pkg = false
+
+  # Path to the private key that is used to sign CRX files.
+  crx_private_key = ""
```
goodov: nit: maybe name this more explicit to mention that this key is used to sign mac dmg.

right now it's easy to assume that this key is used for all `crx3()` invocations.

updater/.gitignore
mherrmann: I think it has been removed entirely.

build/mac/config.gni
mherrmann: I'm finding it a little hard to come up with a better name. As a comment to what you wrote, this key is used for all _our_ `crx3()` invocations - because we only have one. Upstream also has a few, but they are only used in tests.

Alternative names I can think of:

 * dmg_crx_key
 * dmg_crx_private_key
 * browser_updates_crx_key
 * browser_updates_private_key

Do you prefer any of those (or another one)?

goodov: `dmg_crx_private_key` looks fine for a single use related to dmg signing.

If there's a chance it may be used in other parts of the build, we can keep `crx_private_key`, however in this case it's better to share it via `//brave/build/config.gni` or `//brave/build/crx_private_key.gni` (I prefer `crx_private_key.gni`).

mherrmann: I renamed to `dmg_crx_private_key` in 60a51a46cc0b267e862d4c655bcfbeebf1a083cd.

PR #31703, opened 2025-10-08
[NTP Next] Hide sponsored images toggle if Rewards is not supported

zenparsing:
Resolves https://github.com/brave/brave-browser/issues/49998

With rewards disabled by policy or by region check:

Screenshot 2025-10-08 at 8 40 40 AM

### DIFF
--- a/browser/resources/brave_new_tab_page_refresh/components/settings/background_panel.tsx
+++ b/browser/resources/brave_new_tab_page_refresh/components/settings/background_panel.tsx
@@ -36,6 +36,7 @@ export function BackgroundPanel() {
   const selectedBackground = useBackgroundState((s) => s.selectedBackground)
   const braveBackgrounds = useBackgroundState((s) => s.braveBackgrounds)
   const customBackgrounds = useBackgroundState((s) => s.customBackgrounds)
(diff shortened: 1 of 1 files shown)


browser/resources/brave_new_tab_page_refresh/components/settings/background_panel.tsx
```diff
@@ -36,6 +36,7 @@ export function BackgroundPanel() {
   const selectedBackground = useBackgroundState((s) => s.selectedBackground)
   const braveBackgrounds = useBackgroundState((s) => s.braveBackgrounds)
   const customBackgrounds = useBackgroundState((s) => s.customBackgrounds)
+  const rewardsFeatureEnabled = useRewardsState((s) => s.rewardsFeatureEnabled)
   const rewardsEnabled = useRewardsState((s) => s.rewardsEnabled)
```
fallaciousreasoning, line 40: would we want to hide it when rewards is disabled, rather than the whole feature?

zenparsing: There's a label that gets displayed with the toggle if Rewards is not enabled, encouraging the user to enable Rewards.

PR #31704, opened 2025-10-08
[Android] Attempt to fix cut-off background image after app restart

Olipro:
Occasionally, users have reported that the NTP (New Tab Page) background image is partially cut-off after restarting the app.

It now appears likely that this is being caused by the call to `removeOnGlobalLayoutListener(this)` which results in nothing happening when the background image needs to be resized.

This commit deletes the line that unregisters the listener and instead adds code that checks if the measured height/width of the background image differs from the values in a previous run. If they are different, we re-fetch the image.

Resolves brave/brave-browser#28664

### DIFF
--- a/android/java/org/chromium/chrome/browser/ntp/BraveNewTabPageLayout.java
+++ b/android/java/org/chromium/chrome/browser/ntp/BraveNewTabPageLayout.java
@@ -194,6 +194,7 @@ public class BraveNewTabPageLayout extends NewTabPageLayout
     private boolean mIsDisplayNewsFeed;
     private boolean mIsDisplayNewsOptin;
     private long mNewsFeedLastViewTime;
+    private ViewTreeObserver.OnGlobalLayoutListener mBgImageViewOnGlobalLayoutListener;
 
     private static final int SHOW_BRAVE_RATE_ENTRY_AT = 10; // 10th row
 
@@ -897,6 +898,15 @@ public class BraveNewTabPageLayout extends NewTabPageLayout
             mWorkerTask = null;
         }
 
+        var observer =
+                mBgImageView != null && mBgImageViewOnGlobalLayoutListener != null
+                        ? mBgImageView.getViewTreeObserver()
+                        : null;
+        if (observer != null && observer.isAlive()) {
+            observer.removeOnGlobalLayoutListener(mBgImageViewOnGlobalLayoutListener);
+            mBgImageViewOnGlobalLayoutListener = null;
+        }
+
         if (!mIsFromBottomSheet) {
             setBackgroundResource(0);
             if (mImageDrawable != null && mImageDrawable.getBitmap() != null
@@ -1308,22 +1318,42 @@ public class BraveNewTabPageLayout extends NewTabPageLayout
         mBgImageView = (ImageView) findViewById(R.id.bg_image_view);
         mBgImageView.setScaleType(ImageView.ScaleType.MATRIX);
 
-        ViewTreeObserver observer = mBgImageView.getViewTreeObserver();
-        observer.addOnGlobalLayoutListener(
-                new ViewTreeObserver.OnGlobalLayoutListener() {
-                    @Override
-                    public void onGlobalLayout() {
-                        mWorkerTask =
-                                new FetchWallpaperWorkerTask(
-                                        ntpImage,
-                                        mBgImageView.getMeasuredWidth(),
-                                        mBgImageView.getMeasuredHeight(),
-                                        mWallpaperRetrievedCallback);
-                        mWorkerTask.executeOnExecutor(AsyncTask.THREAD_POOL_EXECUTOR);
+        if (mBgImageViewOnGlobalLayoutListener == null) {
+            mBgImageViewOnGlobalLayoutListener =
+                    new ViewTreeObserver.OnGlobalLayoutListener() {
+                        private int mLastWidth;
+                        private int mLastHeight;
 
-                        mBgImageView.getViewTreeObserver().removeOnGlobalLayoutListener(this);
-                    }
-                });
+                        @Override
+                        public void onGlobalLayout() {
+                            int currentWidth = mBgImageView.getMeasuredWidth();
+                            int currentHeight = mBgImageView.getMeasuredHeight();
+
+                            // Only re-fetch if dimensions actually changed
+                            if (currentWidth > 0
+                                    && currentHeight > 0
+                                    && (currentWidth != mLastWidth
+                                            || currentHeight != mLastHeight)) {
+                                mLastWidth = currentWidth;
+                                mLastHeight = currentHeight;
+
+                                if (mWorkerTask != null) {
+                                    mWorkerTask.cancel(true);
+                                }
+
+                                mWorkerTask =
+                                        new FetchWallpaperWorkerTask(
+                                                ntpImage,
+                                                currentWidth,
+                                                currentHeight,
+                                                mWallpaperRetrievedCallback);
+                                mWorkerTask.executeOnExecutor(AsyncTask.THREAD_POOL_EXECUTOR);
+                            }
+                        }
+                    };
+        }
+        ViewTreeObserver observer = mBgImageView.getViewTreeObserver();
+        observer.addOnGlobalLayoutListener(mBgImageViewOnGlobalLayoutListener);
     }
 
     private void checkAndShowNTPImage(boolean isReset) {


android/java/org/chromium/chrome/browser/ntp/BraveNewTabPageLayout.java
```diff
@@ -1314,17 +1314,33 @@ private void setBackgroundImage(NTPImage ntpImage) {
...
                     public void onGlobalLayout() {
-                        mWorkerTask =
-                                new FetchWallpaperWorkerTask(
-                                        ntpImage,
-                                        mBgImageView.getMeasuredWidth(),
-                                        mBgImageView.getMeasuredHeight(),
-                                        mWallpaperRetrievedCallback);
-                        mWorkerTask.executeOnExecutor(AsyncTask.THREAD_POOL_EXECUTOR);
-
-                        mBgImageView.getViewTreeObserver().removeOnGlobalLayoutListener(this);
```
SergeyZhukovsky, line 1324: We still need to remove the listener, it could potentially causing memory leaks if not removed. I would do it inside `onDetachedFromWindow`. Create a member `private ViewTreeObserver.OnGlobalLayoutListener mBackgroundImageLayoutListener;` and clean it up
```
        if (mBackgroundImageLayoutListener != null && mBgImageView != null) {
            mBgImageView.getViewTreeObserver().removeOnGlobalLayoutListener(
                    mBackgroundImageLayoutListener);
            mBackgroundImageLayoutListener = null;
        }
```

android/java/org/chromium/chrome/browser/ntp/BraveNewTabPageLayout.java
```diff
@@ -1314,17 +1314,33 @@ private void setBackgroundImage(NTPImage ntpImage) {
...
+
+                        // Only re-fetch if dimensions actually changed
+                        if (currentWidth > 0
+                                && currentHeight > 0
+                                && (currentWidth != mLastWidth || currentHeight != mLastHeight)) {
+                            mLastWidth = currentWidth;
+                            mLastHeight = currentHeight;
+
+                            if (mWorkerTask != null) {
+                                mWorkerTask.cancel(true);
```
SergeyZhukovsky: from a description to `cancel`.
```
After invoking this method, you should check the
     * value returned by {@link #isCancelled()} periodically from
     * {@link #doInBackground()} to finish the task as early as
     * possible.
```
We need to add `isCancelled` checks and return earlier inside `FetchWallpaperWorkerTask.doInBackground` in a few places at least.

android/java/org/chromium/chrome/browser/ntp/BraveNewTabPageLayout.java
SergeyZhukovsky: does `setBackgroundImage` got called multiple times? If yes, we need to clean the listener as well before creating a new one.

SergeyZhukovsky: or just create it only once in a function that is called once, or don't create it multiple times.

Olipro: > We still need to remove the listener, it could potentially causing memory leaks if not removed. I would do it inside `onDetachedFromWindow`. Create a member `private ViewTreeObserver.OnGlobalLayoutListener mBackgroundImageLayoutListener;` and clean it up

This does not sound right.

As I understand it, the only reason we could possibly cause a memory leak is if this can outlive the View - but if that is the case, holding a `WeakReference` sounds like it would be the right approach. I'm out for the night now but if you could explain how this could leak I'll pick it up in the morning.

android/java/org/chromium/chrome/browser/ntp/BraveNewTabPageLayout.java
Olipro: We could, but when is the user ever going to be changing their DPI (or whatever else can invoke this) so fast and so often that it will outpace the rate at which these redraws are cancelling and therefore need to early-return within the executing task?

android/java/org/chromium/chrome/browser/ntp/BraveNewTabPageLayout.java
SergeyZhukovsky: While the docs don't have a single bold "you must remove this" statement, the pattern is clear:
1. They provide a removal method - `removeOnGlobalLayoutListener()` exists specifically for cleanup.
2. They cared enough to update the API in level 16 to improve removal when deprecated an old `removeGlobalOnLayoutListener()`.
3. A real world example of leaking https://stackoverflow.com/questions/28264139/view-getviewtreeobserver-addongloballayoutlistener-leaks-fragment (Discusses Fragment leaks caused by not removing `OnGlobalLayoutListener`).
4. `ViewTreeObserver` can outlive views in certain edge case scenarios.
5. Best practice hygiene - Clean up what you register.
6. Upstream usage:
  - `ViewRectProvider.java`: Adds listener in `startObserving()`, removes in `stopObserving()` with `isAlive()` check. (We should also add the `isAlive` in our cleaning place);
  - `KeyboardHideHelper.java`: removes in `cleanUp()` method;
  - `PermissionDialogCoordinator.java`: Removes immediately after getting needed info.

android/java/org/chromium/chrome/browser/ntp/BraveNewTabPageLayout.java
SergeyZhukovsky: I don't know, that's why I asked above https://github.com/brave/brave-core/pull/31704#discussion_r2414409436 and also how many times do you see the listener got fired? Better to follow the recommendation and put it there as to me.

android/java/org/chromium/chrome/browser/ntp/BraveNewTabPageLayout.java
bridiver: I think regardless of whether this check turns out to be correct, not correct (prevents it from being called more than once if that is needed) or not strictly needed but following upstream conventions, when removing something like this I would either check in slack or comment on the PR yourself to draw reviewer attention to it. Sometimes there are non-obvious reasons. In this case @SergeyZhukovsky did notice it, but another reviewer might have missed it. 

We also shouldn't blindly follow upstream conventions, we should make sure we understand why they are doing something so we know whether it applies in this situation or not, but for instance if this check turns out to not be strictly necessary but upstream typically does it anyway, we should follow their lead.

Olipro: Yes, agreed - I'm fine with copying anything upstream (or anyone else) does as long as the reason why is known and can be explained, otherwise we're getting into the territory of cargo-culting software development.

With that said, in this instance, I'd like to understand _how_ this code introduces a memory leak. I have also found this SO post: https://stackoverflow.com/questions/36788740/can-a-not-removed-viewtreeobserver-listener-cause-memory-leaks

It makes sense to me that whether or not you could cause a memory leak is dependent on how the code is structured (or, the "architecture" as the author in that SO post worded it) - so, if this is going to leak, it would be beneficial to understand how that's the case.

I'm by no means an expert of Android UI vagaries so perhaps we really do need to explicitly remove this listener. I'd still like to be sure we do that in the right place and what would happen if we don't.

bridiver: @Olipro well to be clear I'm not sure if doing this here is correct or not, in particular I wonder what happens if there is more than one global layout change? We definitely should understand what is happening here, was any of the information @SergeyZhukovsky provided helpful? In particular there was a link (I didn't look at it yet) talking about this issue.

bridiver: Although one thing I would say here is that ref counting makes lifetimes very hard to reason about and for that reason alone I think it's best to be explicit and not force people to think through whether something might leak or not. It's much easier to accidentally leak something when you rely on ref counting. If we don't need it anymore to handle subsequent layout changes, we should explicitly remove it for clarity imo. Not sure if there is anything in the chromium/android style guide about this or not https://chromium.googlesource.com/chromium/src/+/lkgr/styleguide/java/java.md

bridiver: The cpp style guide discourages using ref counted types for exactly this reason. Obviously it's impossible to avoid ref counting in java, but we can do things to make it easier to understand and reason about.

Olipro: OK, so, after testing, Kamil has confirmed that removing the listener was indeed our smoking gun (this change fixes the problem)

Here's my analysis/understanding:

1. We're registering a listener specifically off of the `ImageView` - not the entire view/fragment.
2. The only objects that we capture that I can see potentially preventing garbage collection are `mBgImageView`, `mWorkerTask`, `ntpImage` and `mWallpaperRetrievedCallback`

With that said however, I don't see _how_ that would occur given that Java is able to detect and garbage collect orphaned cyclical references - that is to say, when the `BraveNewTabPageLayout` instance gets collected, the fact that the listener holds a reference to its members shouldn't be an issue (if my understanding holds).

Do we integrate anything like LeakCanary into our builds for detecting leaks? If not, irrefutably proving it is going to be a bit difficult.

And with all of that out of the way; I believe it shouldn't reintroduce the bug if we do explicitly unregister it in `onDetachedFromWindow` as serg suggested. - I hope you don't mind me throwing another APK your way @kjozwiak 😅

android/java/org/chromium/chrome/browser/ntp/BraveNewTabPageLayout.java
Olipro: From looking at the `FetchWallpaperWorkerTask` class, Everything it does looks to be cheap - with the possible exception of `BitmapFactory.decodeStream()` and _maybe_ the call to `NTPImageUtil.getWallpaperBitmap()`

Bearing in mind that these only fire when the screen dimensions have changed, I think we're better off leaving this alone.

SergeyZhukovsky: I just want to be sure it's not executed 10 times in a row. Could you confirm that it's not the case by putting a log there? I'm ok to leave it as is if it's not the case.

Olipro: Ah I see - I dropped in a Log line (executes directly before the `mWorkerTask != null` check) and it only runs once.

Image

(Nothing else pops up after letting it sit there)

android/java/org/chromium/chrome/browser/ntp/BraveNewTabPageLayout.java
Olipro: We now remove the listener in the detach callback - dropped an APK to @kjozwiak to give it a once-over.

android/java/org/chromium/chrome/browser/ntp/BraveNewTabPageLayout.java
SergeyZhukovsky: ok, fine to not cancel in that case.

android/java/org/chromium/chrome/browser/ntp/BraveNewTabPageLayout.java
SergeyZhukovsky: awesome, one more thing. We should check on `isAlive` on the `ViewTreeObserver` before calling `removeOnGlobalLayoutListener`.
```
ViewTreeObserver observer = mBgImageView.getViewTreeObserver();
if (observer.isAlive()) {
    observer.removeOnGlobalLayoutListener(mBackgroundImageLayoutListener);
    mBackgroundImageLayoutListener = null;
}
```
`IllegalStateException` could occur if the `ViewTreeObserver` is invalid. The view `mBgImageView` could already be detached at this point.

Olipro: Alright, I've pared down the logic so that we do the check if everything else is in a suitable state.

SergeyZhukovsky (approved):
++

PR #31746, opened 2025-10-10
Add policy pref interceptor

### LINKED ISSUE #52122: Some policy values should be cached for lifetime of process
Certain features that are configurable via policy do not support policy changes during runtime. Cache policy values for UI heavy features to prevent undefined behavior.

DJAndries:
Resolves https://github.com/brave/brave-browser/issues/52122


patches/components-policy-core-browser-BUILD.gn.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/components/policy/core/browser/BUILD.gn b/components/policy/core/browser/BUILD.gn
+index 08f79c4eb76e1b83cd7a8d39f8344d7c5b97d35b..1b64d5d75310e166f799b7af79c6b9e014b19aff 100644
+--- a/components/policy/core/browser/BUILD.gn
++++ b/components/policy/core/browser/BUILD.gn
+@@ -111,6 +111,8 @@ source_set("internal") {
+     "//net",
+     "//third_party/icu",
+   ]
++
```
bridiver: add the file inline in configuration_policy_pref_store.cc chromium_src override to avoid adding deps

patches/components-policy-core-browser-configuration_policy_pref_store.cc.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/components/policy/core/browser/configuration_policy_pref_store.cc b/components/policy/core/browser/configuration_policy_pref_store.cc
+index 7440287dc5a08d93af1353ec3bb65141dbfb293f..9098815d3376c831f95f8a2c9b688954fd1780da 100644
+--- a/components/policy/core/browser/configuration_policy_pref_store.cc
++++ b/components/policy/core/browser/configuration_policy_pref_store.cc
+@@ -155,6 +155,8 @@ ConfigurationPolicyPrefStore::CreatePreferencesFromPolicies() {
+     }
+   }
+ 
++  BRAVE_CREATE_PREFERENCES_FROM_POLICIES
```
bridiver: `#define ApplyPolicySettings(...) ApplyPolicySettings(__VA_ARGS__); ...`

patches/components-policy-core-browser-BUILD.gn.patch
DJAndries: Will this work? The interceptor depends on `//brave/components/ai_chat/core/common`, so it seems we need to add a dep for that at the very least

DJAndries:
Will add tests when I return.

patches/components-policy-core-browser-BUILD.gn.patch
bridiver: it gets a little tricky with component targets, but since `policy_pref_interceptor` is also a component target I think you can put the import declaration in configuration_policy_pref_store.cc and only (conditionally) put the export declaration in policy_pref_interceptor.h

components/brave_policy/policy_pref_interceptor.cc
```diff
@@ -0,0 +1,100 @@
...
+#endif
+
+#if BUILDFLAG(ENABLE_SPEEDREADER)
+#include "brave/components/speedreader/speedreader_pref_names.h"
+#endif
+
+#if BUILDFLAG(ENABLE_BRAVE_TALK)
+#include "brave/components/brave_talk/pref_names.h"
+#endif
+
```
netzenbot: `DisableCachingForTesting()` sets a global bool to `true` but provides no way to reset it. This will leak state across tests running in the same process. Consider using a `base::AutoReset`-based scoped helper instead. [best practice](https://github.com/brave-experiments/brave-core-tools/tree/master/docs/best-practices/testing-isolation.md)

chromium_src/components/policy/core/browser/configuration_policy_pref_store.cc
```diff
@@ -0,0 +1,14 @@
+/* Copyright (c) 2025 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "components/policy/core/browser/configuration_policy_handler_list.h"
```
netzenbot, line 6: nit: This chromium_src override file lacks a comment explaining why the override is needed. Add a brief comment describing that this intercepts `ApplyPolicySettings` to cache non-dynamic-refresh policy pref values. [best practice](https://github.com/brave-experiments/brave-core-tools/tree/master/docs/best-practices/chromium-src-overrides.md#-add-explanation-comments-in-chromium_src-override-files)

chromium_src/components/policy/core/browser/configuration_policy_pref_store.h
```diff
@@ -0,0 +1,19 @@
+/* Copyright (c) 2025 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#ifndef BRAVE_CHROMIUM_SRC_COMPONENTS_POLICY_CORE_BROWSER_CONFIGURATION_POLICY_PREF_STORE_H_
+#define BRAVE_CHROMIUM_SRC_COMPONENTS_POLICY_CORE_BROWSER_CONFIGURATION_POLICY_PREF_STORE_H_
+
+#include "brave/components/brave_policy/policy_pref_interceptor.h"
```
netzenbot, line 9: nit: This chromium_src header override lacks a comment explaining why the override is needed. Add a brief comment describing that this injects a `PolicyPrefInterceptor` member to preserve non-dynamic-refresh pref values. [best practice](https://github.com/brave-experiments/brave-core-tools/tree/master/docs/best-practices/chromium-src-overrides.md#-add-explanation-comments-in-chromium_src-override-files)

chromium_src/components/policy/core/browser/configuration_policy_pref_store.cc
DJAndries: documentation is available on the `PolicyPrefInterceptor` itself

chromium_src/components/policy/core/browser/configuration_policy_pref_store.h
DJAndries: https://github.com/brave/brave-core/pull/31746#discussion_r3239334313

components/brave_policy/policy_pref_interceptor.cc
DJAndries: It's fine if it leaks; all policy browser tests don't test caching

patches/components-policy-core-browser-BUILD.gn.patch
DJAndries: changed to use a singleton which will be populated in `//browser/policy`, as discussed over DM

browser/policy/policy_pref_interceptor_utils.cc
```diff
@@ -0,0 +1,65 @@
...
+    ai_chat::prefs::kEnabledByPolicy,
+#endif
+    brave_news::prefs::kBraveNewsDisabledByPolicy,
+    brave_rewards::prefs::kDisabledByPolicy,
+#if BUILDFLAG(ENABLE_BRAVE_VPN)
+    brave_vpn::prefs::kManagedBraveVPNDisabled,
+#endif
+    brave_wallet::kBraveWalletDisabledByPolicy,
+    brave_talk::prefs::kDisabledByPolicy,
+    playlist::kPlaylistEnabledPref,
```
netzenbot: `brave_talk::prefs::kDisabledByPolicy` is used unconditionally here, but `brave/components/brave_talk/pref_names.h` is only included when `BUILDFLAG(ENABLE_BRAVE_TALK)` is true. This will fail to compile when `enable_brave_talk=false`. Wrap the entry in `#if BUILDFLAG(ENABLE_BRAVE_TALK)` the same way the include is guarded. ([✅ Place Includes Inside BUILDFLAG Guards When Only Used There](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-034))

browser/policy/policy_pref_interceptor_utils.cc
```diff
@@ -0,0 +1,65 @@
...
+#if BUILDFLAG(ENABLE_SPEEDREADER)
+    speedreader::kSpeedreaderEnabled,
+#endif
+};
+
+}  // namespace
+
+void InitializePolicyPrefInterceptorList() {
```
bridiver: why do we need this method instead of just calling `PolicyPrefInterceptorList::GetInstance()->Init()` or something? It's not clear to me why we need to pass in the non-dynamic pref list.

browser/ui/brave_browser_command_controller_browsertest.cc
```diff
@@ -96,6 +97,7 @@ class BraveBrowserCommandControllerTest : public InProcessBrowserTest {
 
 #if BUILDFLAG(ENABLE_AI_CHAT)
   void BlockAIChatByPolicy(bool value) {
+    brave_policy::PolicyPrefInterceptor::DisableCachingForTesting();
```
bridiver, line 99: this whole pattern of setting the policy in tests seems fragile, we should follow-up to create some kind of wrapper method where we don't have to duplicate things like policy level, source, scope, etc...

patches/components-policy-core-browser-BUILD.gn.patch
bridiver: still figuring out whether we need this or not, but patches should not add new lines

browser/policy/policy_pref_interceptor_utils.cc
bridiver: Oh I remember now, it was a dependency issue

browser/policy/policy_pref_interceptor_utils.cc
```diff
@@ -0,0 +1,65 @@
...
+#include "brave/components/ai_chat/core/common/buildflags/buildflags.h"
+#include "brave/components/brave_news/common/pref_names.h"
+#include "brave/components/brave_policy/policy_pref_interceptor_list.h"
+#include "brave/components/brave_rewards/core/pref_names.h"
+#include "brave/components/brave_talk/buildflags/buildflags.h"
+#include "brave/components/brave_vpn/common/buildflags/buildflags.h"
+#include "brave/components/brave_wallet/browser/pref_names.h"
+#include "brave/components/constants/pref_names.h"
+#include "brave/components/playlist/core/common/pref_names.h"
+#include "brave/components/speedreader/common/buildflags/buildflags.h"
```
netzenbot: `brave/components/playlist/core/common/pref_names.h` has `static_assert(BUILDFLAG(ENABLE_PLAYLIST))`, so this unconditional include will fail to compile when `enable_playlist=false`. Move both the include and the `playlist::kPlaylistEnabledPref` entry inside `#if BUILDFLAG(ENABLE_PLAYLIST)`, and make the `//brave/components/playlist/core/common` dep in BUILD.gn conditional on `enable_playlist`. ([BS-034](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-034))

browser/policy/BUILD.gn
```diff
@@ -17,17 +17,24 @@ import("//brave/components/tor/buildflags/buildflags.gni")
...
+  sources = [
+    "brave_simple_policy_map.h",
+    "policy_pref_interceptor_utils.cc",
+    "policy_pref_interceptor_utils.h",
+  ]
 
   deps = [
     "//base",
     "//brave/components/ai_chat/core/common/buildflags",
+    "//brave/components/brave_news/common",
```
netzenbot: `//brave/components/brave_news/common`, `//brave/components/brave_wallet/browser:pref_names`, and `//brave/components/playlist/core/common` all have `assert(enable_*)` in their own BUILD.gn files, so adding them as unconditional deps here will break builds when those features are disabled. Each should be wrapped under its corresponding `if (enable_brave_news)`, `if (enable_brave_wallet)`, and `if (enable_playlist)` guard — matching the existing pattern for `playlist/content/browser` already in this file. ([BS-040](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-040))

test/BUILD.gn
```diff
@@ -988,6 +988,7 @@ test("brave_browser_tests") {
     "//brave/components/brave_component_updater/browser",
     "//brave/components/brave_perf_predictor/browser",
     "//brave/components/brave_perf_predictor/common",
+    "//brave/components/brave_policy:policy_pref_interceptor",
```
bridiver, line 1019: this belongs in `brave_components_unittests` target

bridiver: wait actually I'm confused, why is this dependency added here?

DJAndries: brave_browser_command_controller_browsertest.cc and brave_toolbar_view_browsertest.cc are included in this target and call `DisableCachingForTesting`

components/brave_policy/policy_pref_interceptor.cc
```diff
@@ -0,0 +1,57 @@
...
+}  // namespace
+
+PolicyPrefInterceptor::PolicyPrefInterceptor() = default;
+
+PolicyPrefInterceptor::~PolicyPrefInterceptor() = default;
+
+// static
+void PolicyPrefInterceptor::DisableCachingForTesting() {
+  g_disable_caching_for_testing = true;
+}
```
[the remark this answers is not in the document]
DJAndries: discussed previously, leaving as follow-up

netzenbot (approved):

browser/policy/policy_pref_interceptor_utils.cc
```diff
@@ -0,0 +1,85 @@
...
+#endif
+#if BUILDFLAG(ENABLE_SPEEDREADER)
+    speedreader::kSpeedreaderEnabled,
+#endif
+};
+
+}  // namespace
+
+void InitializePolicyPrefInterceptorList() {
+  PolicyPrefInterceptorList::GetInstance()->SetPrefs(kNonDynamicPrefs);
```
bridiver: I think this name is a bit misleading/confusing. How about `PolicyPrefInterceptor` -> `NonDynamicPolicyPrefCache`?

browser/policy/brave_simple_policy_map.h
```diff
@@ -80,6 +80,11 @@
 
 namespace policy {
 
+// Note: If a policy without dynamic refresh support is being added
```
bridiver, line 86: unfortunately most people never read comments so we should probably also add this to best practices

patches/components-policy-core-browser-BUILD.gn.patch
bridiver: I guess this is a component target in component builds, they just did it really strangely, but patches shouldn't add new lines

browser/ui/views/toolbar/brave_toolbar_view_browsertest.cc
```diff
@@ -109,6 +110,7 @@ class BraveToolbarViewTest : public InProcessBrowserTest {
 
 #if BUILDFLAG(ENABLE_BRAVE_VPN)
   void BlockVPNByPolicy(bool value) {
+    brave_policy::PolicyPrefInterceptor::DisableCachingForTesting();
```
bridiver, line 113: I'm a bit worried this is going to cause confusion for people when creating new tests because they won't understand why they are not behaving as expected. Maybe rather than putting `kNonDynamicPrefs` in policy_pref_interceptor_utils.cc it should go in brave_simple_policy_map.h and then we just concatenate the two lists. At least that way it's more clearly marked as cached somewhere instead of being hidden inside policy_pref_interceptor_utils.cc

browser/policy/policy_pref_interceptor_utils.cc
DJAndries: IMHO, I think the name fits. Could you elaborate on your concern?

browser/policy/brave_simple_policy_map.h
DJAndries: 60c7f92

browser/ui/views/toolbar/brave_toolbar_view_browsertest.cc
DJAndries: 60c7f92

browser/policy/policy_pref_interceptor_utils.cc
DJAndries: discussed over dm, leaving as is

bridiver (approved):

PR #31779, opened 2025-10-13
[tab helpers]: Move required TabHelpers to follow upstream patterns

### LINKED ISSUE #49741: [TabHelpers]: We should look at moving critical TabHelper functionality to NavigationThrottles
TabHelpers are not attached to all WebContents (for example, background WebContents). Some privacy critical functionality is implemented as `TabHelpers` (`EphemeralStorageTabHelper`, `BraveShieldsTabHelper`).

We should look to migrate as much of the functionality from these TabHelpers to NavigationThrottles, so they run for all web contents. 

In https://github.com/brave/brave-core/pull/31444, I've separated these TabHelpers into a separate section so all critical TabHelpers can easily be attached to a background WebContents, but this still isn't ideal because its easy to forget (and similar features upstream won't apply them).

fallaciousreasoning:
Resolves https://github.com/brave/brave-browser/issues/49741

This is a follow up to https://github.com/brave/brave-core/pull/31444 which does things in a more idiomatic way. There are no functional changes.

Summary:
- Remove `AttachPrivacySensitiveTabHelpers`
- Add `AttachBraveUniversalWebContentsObservers`
- Migrate tab helpers to this new function
- Remove usages of `AttachPrivacySensitiveTabHelpers` (they're automatic now)

### DIFF
--- a/browser/brave_content_browser_client.h
+++ b/browser/brave_content_browser_client.h
@@ -17,6 +17,7 @@
 #include "chrome/browser/chrome_content_browser_client.h"
 #include "content/public/browser/browser_thread.h"
 #include "content/public/browser/content_browser_client.h"
+#include "content/public/browser/web_contents.h"
 #include "mojo/public/cpp/bindings/pending_remote.h"
 #include "services/metrics/public/cpp/ukm_source_id.h"
 #include "third_party/blink/public/mojom/loader/referrer.mojom.h"
@@ -204,6 +205,8 @@ class BraveContentBrowserClient : public ChromeContentBrowserClient {
 #endif
 
   content::BluetoothDelegate* GetBluetoothDelegate() override;
+
+  void OnWebContentsCreated(content::WebContents* web_contents) override;
 
  private:
   void OnAllowGoogleAuthChanged();
--- a/browser/brave_universal_web_contents_observers.h
+++ b/browser/brave_universal_web_contents_observers.h
@@ -1,0 +1,16 @@
+// Copyright (c) 2025 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#ifndef BRAVE_BROWSER_BRAVE_UNIVERSAL_WEB_CONTENTS_OBSERVERS_H_
+#define BRAVE_BROWSER_BRAVE_UNIVERSAL_WEB_CONTENTS_OBSERVERS_H_
+
+namespace content {
+class WebContents;
+}
+
+void AttachBraveUniversalWebContentsObservers(
+    content::WebContents* web_contents);
+
+#endif  // BRAVE_BROWSER_BRAVE_UNIVERSAL_WEB_CONTENTS_OBSERVERS_H_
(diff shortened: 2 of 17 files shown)


browser/brave_universal_web_contents_observers.h
```diff
@@ -0,0 +1,16 @@
...
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#ifndef BRAVE_BROWSER_BRAVE_UNIVERSAL_WEB_CONTENTS_OBSERVERS_H_
+#define BRAVE_BROWSER_BRAVE_UNIVERSAL_WEB_CONTENTS_OBSERVERS_H_
+
+namespace content {
+class WebContents;
+}
+
+void AttachBraveUniversalWebContentsObservers(
```
ShivanKaul, line 15:
```suggestion
// These TabHelpers always attach to WebContents (including background ones).
// Ones related to privacy and security should probably go here.
// See |universal_web_contents_observers.h| upstream.
void AttachBraveUniversalWebContentsObservers(
```

ShivanKaul (approved):
This does look more idiomatic, thanks for doing this. Defer to @bridiver for brave-core implementation review but approach lgtm.

browser/sources.gni
```diff
@@ -90,6 +90,8 @@ brave_chrome_browser_sources = [
   "//brave/browser/brave_shell_integration.h",
   "//brave/browser/brave_tab_helpers.cc",
   "//brave/browser/brave_tab_helpers.h",
+  "//brave/browser/brave_universal_web_contents_observers.cc",
```
bridiver, line 109: don't add these to sources, create a target for them using public/impl to break circular deps

browser/brave_content_browser_client.h
```diff
@@ -14,6 +14,7 @@
 #include "chrome/browser/chrome_content_browser_client.h"
 #include "content/public/browser/browser_thread.h"
 #include "content/public/browser/content_browser_client.h"
+#include "content/public/browser/web_contents.h"
```
netzenbot, line 20: The full `#include "content/public/browser/web_contents.h"` is unnecessary here — `OnWebContentsCreated` only uses `content::WebContents*` (a pointer), so a forward declaration suffices. The base class header already forward-declares `content::WebContents`. [best practice](https://github.com/brave-experiments/brave-core-tools/tree/master/docs/best-practices/coding-standards.md#-use-forward-declarations-in-headers-include-in-cc)

browser/sources.gni
```diff
@@ -106,6 +106,8 @@ brave_chrome_browser_sources = [
...
+  "//brave/browser/brave_universal_web_contents_observers.h",
```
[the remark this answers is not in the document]
darkdh: https://github.com/brave/brave-core/pull/31779#discussion_r2433906596

browser/sources.gni
```diff
@@ -106,6 +106,8 @@ brave_chrome_browser_sources = [
   "//brave/browser/brave_shell_integration.h",
   "//brave/browser/brave_tab_helpers.cc",
   "//brave/browser/brave_tab_helpers.h",
+  "//brave/browser/brave_universal_web_contents_observers.cc",
```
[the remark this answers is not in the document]
darkdh: https://github.com/brave/brave-core/pull/31779#discussion_r2433906596

fallaciousreasoning:
I've just been looking into whether I can ressurrect this PR, but I think the timing of OnWebContentsCreated isn't going to work for the EphemeralStorageHelper tests.

It breaks session restore because it runs before navigations commit, so storage would be cleared for a tab when you reopen a window (i.e. after a restart) which I think wouldn't match what we do currently.

PR #31864, opened 2025-10-16
[Android] Build Wireguard-android directly from source.

Olipro:
We now build the Wireguard code directly from source via a submodule within `third_party/wireguard-android`. This is achieved through the addition of a BUILD.gn to our own fork of the wireguard-android repository.

For further details, see the HEAD commit of
https://github.com/brave-experiments/wireguard-android

Resolves <TODO>


patches/chrome-android-chrome_public_apk_tmpl.gni.patch
```diff
@@ -18,11 +18,21 @@ index eb45f6da579aaf099d848ab0f5e98b8d0c9b0b71..6a3a079741b37a5c7cb94bd085e76364
...
          proguard_configs +=
              [ "//android_webview/nonembedded/java/proguard.flags" ]
-@@ -763,6 +765,7 @@ template("chrome_common_apk_or_module_tmpl") {
+@@ -763,6 +765,17 @@ template("chrome_common_apk_or_module_tmpl") {
                                   "version_code",
                                   "version_name",
                                 ])
-+    _wireguard_target = "//brave/third_party/android_deps:com_wireguard_android_J__unpack_aar" deps += [ _wireguard_target] _libwireguard_dir = get_label_info("//brave/third_party/android_deps:com_wireguard_android($default_toolchain)","target_out_dir") + "/com_wireguard_android_java/jni" loadable_modules +=[ "$_libwireguard_dir/$android_app_abi/libwg-go.so" ]
++    # WireGuard native libraries built from source
++    deps += [
```
Olipro, line 26: I'd like to concoct a better way of doing this such that the definitions are added within the Brave repo rather than directly in this patch... The key question being: _"Where would we put it"_

script/brave_license_helper.py
```diff
@@ -56,6 +56,12 @@ def AddBraveCredits(root, prune_paths, special_cases, prune_dirs,
         os.path.join('brave', 'vendor', 'brave-ios'),
         os.path.join('brave', 'vendor', 'brave_base'),
 
+        # TODO: This can probably be removed entirely before merging.
```
Olipro, line 59: TODO as written.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "third_party/wireguard-android"]
+	path = third_party/wireguard-android
+	url = https://github.com/brave-experiments/wireguard-android.git
```
bsclifton: We should use `DEPS` file instead - see https://github.com/brave/brave-core/blob/master/DEPS (and then we can remove submodule too).

This executes during the `run_hooks` portion of `npm run sync` / `npm run init`.

third_party/android_deps/BUILD.gn
```diff
@@ -54,17 +53,6 @@ android_aar_prebuilt("com_github_bumptech_glide_gifdecoder_java") {
   deps = [ "//third_party/androidx:androidx_annotation_annotation_java" ]
 }
 
-android_aar_prebuilt("com_wireguard_android_java") {
-  aar_path = "libs/com_wireguard_android/tunnel-1.0.20250531.aar"
```
bsclifton, line 72: We should remove the file itself at `third_party/android_deps/libs/com_wireguard_android/tunnel-1.0.20250531.aar`

We can also delete the prior tools script `tools/android/wireguard-build.sh`

.gitmodules
bsclifton: Fixed in https://github.com/brave/brave-core/pull/31864/commits/26d2ea2c05edf76d1b25435163c80c7473c2dc60

third_party/android_deps/BUILD.gn
bsclifton: Fixed in https://github.com/brave/brave-core/pull/31864/commits/26d2ea2c05edf76d1b25435163c80c7473c2dc60

bsclifton:
Closing this for now as stale - we can revisit! 
There are some exciting things happening with VPN for Desktop, maybe we can properly revisit Android too.

PR #32242, opened 2025-11-11
add security tests for cardano dapp support

kdenhartog:
These were mostly tests generated during static code analaysis of the cardano dapp functionality. I wanted to make sure that I understood what the code was doing by having an agent add these tests and then I reviewed them for functional cover of edge cases that I felt could impact the security of the wallet.

Related to https://github.com/brave/reviews/issues/2000

### DIFF
--- a/components/brave_wallet/browser/cardano/cardano_cip30_serializer_unittest.cc
+++ b/components/brave_wallet/browser/cardano/cardano_cip30_serializer_unittest.cc
@@ -124,4 +124,112 @@ TEST(CardanoCip30SerializerTest, SerializeUtxos) {
             CardanoCip30Serializer::SerializeUtxos({}));
 }
 
+TEST(CardanoCip30SerializerTest, SerializeUtxos_Single) {
+  cardano_rpc::UnspentOutputs utxos;
+  cardano_rpc::UnspentOutput output;
+  output.output_index = 0;
+  output.lovelace_amount = 1000000u;
+  std::vector<uint8_t> tx_bytes;
+  base::HexStringToBytes(
+      "d9ef8dcd983c6fe996d5029e010e224bec191d0f63ff695cdab046abfd79dfbd",
+      &tx_bytes);
+  base::SpanWriter(base::span(output.tx_hash)).Write(tx_bytes);
+  output.address_to = *CardanoAddress::FromString(
+      "addr1qyx2zscdearcexdktcgq6g27jkyff65dw82w6catczfwxz2qjy"
+      "nwf42y3c7ejrrekj5r2fh6kx5m9gcrmywpqxw3np5qjeh38p");
+  utxos.push_back(std::move(output));
+
+  auto result = CardanoCip30Serializer::SerializeUtxos(utxos);
+  EXPECT_TRUE(result.has_value());
+  EXPECT_EQ(result->size(), 1u);
+}
+
+TEST(CardanoCip30SerializerTest, DeserializeAmount_InvalidCBOR) {
+  EXPECT_FALSE(CardanoCip30Serializer::DeserializeAmount("ZZZZ"));
+  EXPECT_FALSE(CardanoCip30Serializer::DeserializeAmount("1a00"));
+  EXPECT_FALSE(CardanoCip30Serializer::DeserializeAmount("6568656c6c6f"));
+}
+
+TEST(CardanoCip30SerializerTest, DeserializeAmount_Zero) {
+  auto serialized = CardanoCip30Serializer::SerializeAmount(0u);
+  EXPECT_TRUE(serialized.has_value());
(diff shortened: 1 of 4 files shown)


components/brave_wallet/browser/cardano/cardano_wallet_service_unittest.cc
```diff
@@ -175,6 +175,130 @@ TEST_F(CardanoWalletServiceUnitTest, GetChangeAddress) {
...
+TEST_F(CardanoWalletServiceUnitTest, NoAddressMethod_ReturnsMultipleAddresses) {
+  SetupCardanoAccount();
+
+  // GetUsedAddresses should NOT return multiple addresses
+  auto used_addresses = cardano_wallet_service_->GetUsedAddresses(account_id());
+  EXPECT_EQ(used_addresses.size(), 1u)
+      << "GetUsedAddresses returned " << used_addresses.size()
+      << " addresses. If this is intentional (Issue #45278), update this test "
+         "and review privacy implications in "
+         "task-1.8-address-correlation-risk.md";
```
kdenhartog: todo remove these

components/brave_wallet/browser/cardano/cardano_wallet_service_unittest.cc
```diff
@@ -228,4 +352,180 @@ TEST_F(CardanoWalletServiceUnitTest, CreateAndSignCardanoTransaction) {
       cardano_test_rpc_server_->captured_raw_tx());
 }
 
+TEST_F(CardanoWalletServiceUnitTest,
+       SignAndPostTransaction_AccountSwitch_TOCTOU) {
```
kdenhartog: this test kind of feels like junk - I ended up manually verifying this which is where I discovered that the sync calls hold the transactions if multiple are awaiting user interaction. I'll probably drop this unless you see a good reason to keep @supermassive

components/brave_wallet/browser/cardano/cardano_provider_impl_unittest.cc
```diff
@@ -447,4 +447,314 @@ TEST_F(CardanoProviderImplUnitTest, EnableFails_OnWalletUnlock_TabNotActive) {
...
+
+  TestFuture<mojom::CardanoProviderErrorBundlePtr> future;
+  mojo::Remote<mojom::CardanoApi> api;
+  provider()->Enable(api.BindNewPipeAndPassReceiver(), future.GetCallback());
+
+  auto error = future.Take();
+  EXPECT_FALSE(error);
+}
+
+TEST_F(CardanoProviderImplUnitTest, Enable_UnlockRetainsCapturedOrigin) {
```
kdenhartog, line 514: this isn't really needed since the response calls back to the origin via the promise, but is kind of useful for the defense in depth check

kdenhartog: decided to keep this since it's useful for defense in depth

components/brave_wallet/browser/cardano/cardano_wallet_service_unittest.cc
kdenhartog: deleted

components/brave_wallet/browser/cardano/cardano_wallet_service_unittest.cc
kdenhartog: removed

kdenhartog:
Converting this back to draft until I've had a chance to clean this up better and make sure the code is maintainable by the crypto-wallet-core team. The primary purpose of generating these tests were for me to gain confidence in edge cases of the ADA implementation, but I didn't focus on making sure they were runnable in CI or production ready.

components/brave_wallet/browser/cardano/cardano_cip30_serializer_unittest.cc
```diff
@@ -124,4 +124,112 @@ TEST(CardanoCip30SerializerTest, SerializeUtxos) {
...
+      "d9ef8dcd983c6fe996d5029e010e224bec191d0f63ff695cdab046abfd79dfbd",
+      &tx_bytes);
+  base::SpanWriter(base::span(output.tx_hash)).Write(tx_bytes);
+  output.address_to = *CardanoAddress::FromString(
+      "addr1qyx2zscdearcexdktcgq6g27jkyff65dw82w6catczfwxz2qjy"
+      "nwf42y3c7ejrrekj5r2fh6kx5m9gcrmywpqxw3np5qjeh38p");
+  utxos.push_back(std::move(output));
+
+  auto result = CardanoCip30Serializer::SerializeUtxos(utxos);
+  EXPECT_TRUE(result.has_value());
```
cypt4, line 143: We should use valid result test vectors, not just testing result has value

PR #32637, opened 2025-12-02
Get codebase ready for compute_inputs_for_analyze=true for gn analyze

nikhedonia:
This PR attempts to improve the correctness of `gn analyze` by getting the codebase ready to support `compute_inputs_for_analyze=true`.

This flag front loads some of the dependency analysis from build time to analysis time by invoking exec_script during `gn gen`.

This enforces a more correct dependency graph and allows us to determine better what tests need to run given a file change without running the build.
We intent to enable this flag only on CI as it has a negative build performance during local development. By doing so there is a small chance that a build might succeed locally but fail on CI. However the error messages are usually quite actionable


components/ai_chat/resources/BUILD.gn
goodov, line 58: shouldn't this be listed here via `public_deps` instead of pulling `preprocess_html` into `//brave/components/resources:static_resources_grit`?

the current solution feels like a workaround.. I dunno.

browser/resources/ai_chat_agent_new_tab_page/BUILD.gn
```diff
@@ -45,6 +45,10 @@ generate_grd("static_resources_grd") {
 
 brave_grit("pack_static_resources") {
   source = static_grd_path
+
+  # The .grd file is generated by generate_grd at build time by transpile_web_ui
```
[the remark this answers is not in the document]
fallaciousreasoning: This comment from copilot seems correct - nothing to do with transpile_web_ui here (only JS/CSS is going through it)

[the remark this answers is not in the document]
fallaciousreasoning: The generated pak is sourced on line 65

components/resources/BUILD.gn
```diff
@@ -32,13 +32,21 @@ brave_grit("static_resources") {
   ]
 
   if (enable_ai_chat) {
-    deps += [ "//brave/components/ai_chat/resources" ]
+    deps += [
+      "//brave/components/ai_chat/resources",
+
+      # Needed for preprocessed HTML files referenced in ai_chat_ui_resources.grdp
+      "//brave/components/ai_chat/resources:preprocess_html",
```
fallaciousreasoning, line 39: Why do we need this dep? Its a dependency of `//brave/components/ai_chat/resources:resource`? 

If we need it to be able to iterate over the files in the pak, maybe it should be a public dep?

components/resources/BUILD.gn
```diff
@@ -32,13 +32,21 @@ brave_grit("static_resources") {
...
+    ]
   }
 
   if (enable_brave_wallet) {
     deps += [
       "//brave/components/brave_wallet_ui/page:brave_wallet_page_generated",
       "//brave/components/decentralized_dns/content/resources:bundle_js",
+
+      # Needed for preprocessed HTML files referenced in brave_components_resources.grd
+      "//brave/components/brave_wallet_ui/page:preprocess_html",
```
fallaciousreasoning, line 49: ditto

components/ai_chat/resources/BUILD.gn
```diff
@@ -36,7 +36,12 @@ transpile_web_ui("ai_chat_ui") {
 }
 
 preprocess_if_expr("preprocess_html") {
-  visibility = [ ":resources_grit" ]
+  visibility = [
+    ":resources_grit",
+
+    # Also used by brave_components_resources.grd via ai_chat_ui_resources.grdp
+    "//brave/components/resources:static_resources_grit",
```
fallaciousreasoning, line 43: If this was a public dep of `:resources` would we need this?

components/brave_wallet_ui/page/BUILD.gn
```diff
@@ -29,7 +29,12 @@ transpile_web_ui("brave_wallet_page_ui") {
 }
 
 preprocess_if_expr("preprocess_html") {
-  visibility = [ ":brave_wallet_page_generated_grit" ]
+  visibility = [
+    ":brave_wallet_page_generated_grit",
+
+    # Also used by brave_components_resources.grd
+    "//brave/components/resources:static_resources_grit",
```
fallaciousreasoning, line 36: ditto

netzenbot (approved):

wknapik:
Could anyone please pick this up?

Linux PR builds have an average runtime of 45min. 38min of that is a fixed time spent executing all tests. This PR is the basis for cutting out any part of that 38min per build.

PR #32717, opened 2025-12-08
[AI Chat]: Add inline mentions of attachments

fallaciousreasoning:
Series https://github.com/brave/brave-browser/issues/51151

https://github.com/user-attachments/assets/14a8752e-0370-4cbf-875a-5ede4ab480dc

### DIFF
--- a/components/ai_chat/resources/page/components/filter_menu/attachments_menu.tsx
+++ b/components/ai_chat/resources/page/components/filter_menu/attachments_menu.tsx
@@ -53,24 +53,23 @@ export default function TabsMenu() {
     }
   }, [query !== null])
 
-  const handleClose = React.useCallback(
-    (isOpen: boolean) => {
-      if (!isOpen) {
-        const editor = document.querySelector<HTMLElement>('[data-editor]')
-        if (!editor) return
+  const handleClose = React.useCallback((isOpen: boolean) => {
+    if (isOpen) return
 
-        makeEdit(editor).selectRangeToTriggerChar('@').replaceSelectedRange('')
-        setIsOpen(false)
+    const editor = document.querySelector<HTMLElement>('[data-editor]')
+    if (!editor) return
 
-        editor?.focus()
-      }
-    },
-    [conversation.setInputText],
-  )
+    // If we have a trigger character, but the menu is closed make sure we remove it.
+    makeEdit(editor)
+      .selectRangeToTriggerChar('@')
+      .ifHasSelection()
+      ?.replaceSelectedRange('')
+
+    editor?.focus()
+  }, [])
 
   const selectAttachment = React.useCallback(
     (attachment: Attachment) => {
-      setIsOpen(false)
       if ('contentId' in attachment) {
         aiChat.api.uiHandler.associateTab(
           attachment,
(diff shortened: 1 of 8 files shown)


components/ai_chat/resources/page/components/filter_menu/filter_menu.tsx
```diff
@@ -52,6 +54,7 @@ export function MatchedText(props: { text: string; match?: Match }) {
 }
 
 export default function FilterMenu<T>(props: Props<T>) {
+  const domToItem = React.useRef(new Map<HTMLElement, T>())
```
petemill: This function could do with a comment especially explaining what role `T` plays, that might help explain what `domToItem` is storing - otherwise that needs a comment because I'm not sure the variable name is self explanatory.
Also that `leo-menu-item`s are the expected children and they are built by the parent but 1 by 1 via function-as-children calls. Sorry if this was missed on a previous review.

components/ai_chat/resources/page/components/filter_menu/filter_menu.tsx
```diff
@@ -136,6 +139,36 @@ export default function FilterMenu<T>(props: Props<T>) {
...
+      }
+
+      props.onSelectedItemChanged?.(item)
+    }
+    handler()
+    ref.current?.addEventListener('focusin', handler)
+    return () => {
+      ref.current?.removeEventListener('focusout', handler)
+    }
+  }, [props.isOpen, props.onSelectedItemChanged, filtered])
```
petemill: Would be nice for this block to have at least one comment explaining what it's for.

components/ai_chat/resources/untrusted_conversation_frame/components/conversation_entries/index.test.tsx
```diff
@@ -596,7 +596,7 @@ describe('ConversationEntries visualContentUsedPercentage handling', () => {
   })
 
   test('should highlight skill shortcuts in text', () => {
```
petemill, line 797: shouldn't we add some tests for content attachment mentions?

components/ai_chat/resources/page/components/filter_menu/attachments_menu.tsx
```diff
@@ -41,6 +41,7 @@ const icon: Record<ReturnType<typeof getType>, string> = {
 export default function TabsMenu() {
   const aiChat = useAIChat()
   const conversation = useConversation()
+  const [selected, setSelected] = React.useState<Attachment | null>(null)
```
petemill: formatting (extra indentation), wonder why this isn't getting caught

petemill (commented):
Minor comments, but couple issues I've found when testing are as follows. Probably follow-ups:
- If you call up the attachment menu with `@` but then don't select anything and dismiss the popup, the last highlighted item becomes attached and selected.
- Unattaching a Tab does not cause the mention to lose it's structure. Maybe it should get at least a strikethrough?
- I think that the inline mention should have a tooltip or title which shows the URL. This will be sueful when there are multiple attachments with the same name.

components/ai_chat/resources/page/components/filter_menu/attachments_menu.tsx
fallaciousreasoning: removed

components/ai_chat/resources/page/components/filter_menu/filter_menu.tsx
fallaciousreasoning: I've simplified this a bit - I'll add the comment as a separate PR

components/ai_chat/resources/page/components/filter_menu/filter_menu.tsx
fallaciousreasoning: Removed

components/ai_chat/resources/untrusted_conversation_frame/components/conversation_entries/index.test.tsx
fallaciousreasoning: Huh thought I had but doesn't seem like it - I'll dop that now

fallaciousreasoning (commented):
Simplified the logic for attaching stuff and added tests. I think I've addressed your other comments too :smile:

netzenbot (approved):

Douglashdaniel (approved):

fallaciousreasoning:
I think the problems with editing will be unblocked by https://github.com/brave/brave-core/pull/36130

Douglashdaniel (approved):
Additional changes LGTM!

PR #32721, opened 2025-12-08
[AI Chat] Use brave:// URLs on android

### LINKED ISSUE #49938: [Android] Leo fullscreen URL displays chrome:// instead of brave://
### Description

When opening Leo the URL starts with chrome:// instead of brave://

Slack thread with context: https://bravesoftware.slack.com/archives/C05NDQP03AQ/p1759551651548459

Image

### Steps to reproduce

1. Open Leo
2. Check URL

### Actual result

The URL starts with chrome://

### Expected result

The URL starts with brave://

### Reproduces how often

Easily reproduced

### Brave version

1.82

### Device

N/A

### Channel information

- [x] release (stable)
- [x] beta
- [x] nightly

### Reproducibility

petemill:
- Resolves https://github.com/brave/brave-browser/issues/49938

This is a workaround whilst android does not rewrite chrome:// URLs for display
- https://github.com/brave/brave-browser/issues/51302

### DIFF
--- a/components/ai_chat/core/common/ai_chat_urls.cc
+++ b/components/ai_chat/core/common/ai_chat_urls.cc
@@ -14,12 +14,33 @@
 
 namespace ai_chat {
 
+namespace {
+
+#if BUILDFLAG(IS_ANDROID)
+constexpr char kBraveScheme[] = "brave";
+#endif
+
+GURL CreateAIChatUIURL(std::string_view path) {
+  GURL url(base::StrCat({kAIChatUIURL, path}));
+#if BUILDFLAG(IS_ANDROID)
+  // TODO(https://github.com/brave/brave-browser/issues/51302): Remove this
+  // override once chrome:// URLs are rewritten as brave:// URLs in the android
+  // UI.
+  GURL::Replacements replacements;
+  replacements.SetSchemeStr(kBraveScheme);
+  url = url.ReplaceComponents(replacements);
+#endif
+  return url;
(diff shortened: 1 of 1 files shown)


netzenbot (approved):

fallaciousreasoning (approved):
lgtm, sorry for force pushing to your branch! I reverted to your version as its a bit nicer!

components/ai_chat/core/common/ai_chat_urls.cc
```diff
@@ -14,12 +14,33 @@
...
+  url = url.ReplaceComponents(replacements);
+#endif
+  return url;
+}
+
+}  // namespace
+
 GURL TabAssociatedConversationUrl() {
-  return GURL(base::StrCat({kAIChatUIURL, "tab"}));
+  return CreateAIChatUIURL("tab");
```
bridiver, line 39: Just want to be sure that this is display only? We should not be storing any URLs as brave or using brave anywhere where there might be security checks

petemill: We reverted this didn't we?

bridiver: Actually I think this is before my PR? If so it can probably be removed

petemill: This was reverted https://github.com/brave/brave-core/pull/34627

PR #32780, opened 2025-12-10
Update rust tools build process to share intermediate targets.

### LINKED ISSUE #51430: Rework rust build tools process (wasm-related) to avoid build bottlenecks
There are few issues currently:

1. `wasm-pack`, `wasm-opt`, `wasm-bindgen-cli` are build in separate trees which prevents already built dependencies sharing across them.
2. `wasm-opt` is a build dependency for other two tools, which doesn't have to be. This is the main bottleneck which makes the build of those 3 packages non-parallel.

This build bottleneck becomes very real on a fully cached build, so we should fix it.

goodov:
This PR removes rust tools build bottleneck by doing few things:
1. `crate build` reuses already built dependencies.
2. wasm-related binaries are built in parallel without a dependency on `wasm-opt`.
3. release optimizations are lowered to speedup the build a bit
4. custom action scripts are replaced with a generalized one to support environment vars passing

before:
```
    444.6s weighted time (47066.6s elapsed time sum, 105.9x parallelism)
    60095 build steps completed, average of 135.17/s
```

after:
```
    360.2s weighted time (58289.9s elapsed time sum, 161.8x parallelism)
    60096 build steps completed, average of 166.82/s
```

Saved ~80 seconds on fully cached `--target=brave` build.

Resolves https://github.com/brave/brave-browser/issues/51430


tools/crates/.cargo/config.toml
```diff
@@ -3,3 +3,7 @@ replace-with = "vendored-sources"
 
 [source.vendored-sources]
 directory = "vendor"
+
+[profile.release]
+opt-level = 1
+lto = "off"
```
cdesouza-chromium, line 9: Are we sure about this? Will this affect production wasm code, or is this exclusively contained to the tooling only?

goodov: Initially I had a separate build profile here for the rust build tools, but I figured this vendor directory is only used for the build tools, so went with adjusting `release` profile and using it as is. Happy to be corrected though if I miss something cc @szilardszaloki.

Wasm packages use their own build call which requires an in-package cargo config:
https://github.com/brave/brave-core/blob/70c5352cb11203312e72d2954bcfa34316a6559e/components/common/rust_to_wasm.gni#L63-L64

cdesouza-chromium (approved):
LGTM

goodov:
did a rebase to ensure nothing is broken

tools/crates/build_crate.gni
```diff
@@ -3,60 +3,85 @@
...
+  } else {
+    crates = [ target_name ]
+  }
+
+  if (defined(invoker["executable_outputs"])) {
+    executable_outputs = invoker["executable_outputs"]
+  } else {
+    executable_outputs = [ target_name ]
+  }
+
```
szilardszaloki, line 20:
```gn
assert(len(crates) == len(executable_outputs))
```

tools/crates/.cargo/config.toml
szilardszaloki: Correct — this will not affect WebUI WASM resources directly. That said, it will be blown away the next time `npm run update_brave_tools_crates` is run. It probably makes sense to move this into `create_cargo_config()` in `tools/crates/update.py` instead:
https://github.com/brave/brave-core/blob/master/tools/crates/update.py.

tools/crates/build_crate.gni
```diff
@@ -3,60 +3,85 @@
...
+  cargo_target_dir = "$root_build_dir/brave_tools_crates"
```
szilardszaloki: This could turn into a race. All `build_crate` targets would share the same `cargo_target_dir = "$root_build_dir/brave_tools_crates"`, but declare different executable outputs. GN would see no conflict and might run them in parallel, even though they are all poking at the same Cargo target directory.

tools/crates/.cargo/config.toml
goodov: ah, I wasn't aware this file is regenerated. Adjusted `update.py`.

tools/crates/build_crate.gni
goodov: A package can potentially generate more than one executable or less than one, so I decided to not add such assert here.

tools/crates/build_crate.gni
goodov: This was intended, because `cargo build` should lock the build directory and prevent multiple build commands to work in the same directory. However I'm a bit hesitant about this, so yeah, let's use a separate dir for each `build_crate()` target as it was before.

tools/crates/build_crate.gni
szilardszaloki: That makes sense — actually, the `assert()` is redundant. I just wanted to make sure there isn't a case where executables not listed in `executable_outputs` get built but aren't copied to `$root_build_dir`, but this will be caught by dependent targets.

BUILD.gn
```diff
@@ -723,25 +723,31 @@ if (is_mac) {
...
-  if (audit_dev_deps) {
-    args += [ "--audit_dev_deps" ]
-  }
+    # Action must have outputs, and these outputs must be created, otherwise
+    # SISO will fail.
+    outputs = [ output ]
+
+    args = [
+      "--source_root=" + rebase_path("//brave"),
+      "--cargo_audit_exe=" + rebase_path(cargo_audit_outputs[0]),
```
szilardszaloki: Maybe simplify this to `$root_build_dir/cargo-audit$host_exe_suffix`?

szilardszaloki (approved):
LGTM — good stuff! 👍🏻

tools/crates/build_crate.gni
```diff
@@ -3,60 +3,85 @@
...
-    args = [
-      "--temp_dir_path",
-      rebase_path(crate_target_dir),
+    response_file_contents = [
+      "TMPDIR=" + rebase_path(cargo_target_dir),
+      "TEMP=" + rebase_path(cargo_target_dir),
+      "TMP=" + rebase_path(cargo_target_dir),
+      "PATH=" +
+          rebase_path("//third_party/rust-toolchain/bin", root_build_dir) +
+          host_path_sep + getenv("PATH"),
```
bridiver, line 60: Why are we added PATH here? We specifically don't want arbitrary user paths here as that could cause unexpected behavior and we don't want to use anything outside of the chromium toolchain.

goodov: the idea is simple: PATHs are usually appended to have sys-wide paths be also available to whatever is happening inside. This call does just that.

PR #32812, opened 2025-12-12
Potential Vulnerability in Cloned Code

tabudz:
### Summary
Our tool detected a potential vulnerability in tools/crates/vendor/wasm-opt-sys/binaryen/third_party/llvm-project/SmallVector.cpp which was cloned from facebook/hermes but did not receive the security patch applied. The original issue was reported and fixed under https://nvd.nist.gov/vuln/detail/cve-2022-32234.

### Proposed Fix
Apply the same patch as the one in facebook/hermes to eliminate the vulnerability.

### Reference
https://nvd.nist.gov/vuln/detail/cve-2022-32234
https://github.com/facebook/hermes/commit/06eaec767e376bfdb883d912cb15e987ddf2bda1

Resolves


bsclifton:
This code is vendored; not part of ours. We update as needed. Closing

PR #9602, opened 2021-08-01
Fix#14496 - Added a toggle for OpenSearch that is off by default

rahulsnkr:
Resolves brave/brave-browser#14496. 

https://user-images.githubusercontent.com/28645536/127785125-5a481b4b-9460-4aef-a3f5-c95108dcdaf4.mp4

## Submitter Checklist:

- [x] I confirm that no security/privacy review [is needed](https://github.com/brave/brave-browser/wiki/Security-reviews), or that I have [requested](https://github.com/brave/security/issues/new/choose) one
- [x] There is a [ticket](https://github.com/brave/brave-browser/issues) for my issue
- [x] Used Github [auto-closing keywords](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue) in the PR description above
- [x] Wrote a good [PR/commit description](https://google.github.io/eng-practices/review/developer/cl-descriptions.html)
- [ ] Added appropriate labels (`QA/Yes` or `QA/No`; `release-notes/include` or `release-notes/exclude`; `OS/...`) to the associated issue
- [x] Checked the PR locally: `npm run test -- brave_browser_tests`, `npm run test -- brave_unit_tests`, `npm run lint`, `npm run gn_check`, `npm run tslint`
- [ ] Ran `git rebase master` (if needed)

## Reviewer Checklist:

- [ ] A security review [is not needed](https://github.com/brave/brave-browser/wiki/Security-reviews), or a link to one is included in the PR description
- [ ] New files have MPL-2.0 license header
- [ ] Adequate test coverage exists to prevent regressions
- [ ] Major classes, functions and non-trivial code blocks are well-commented
- [ ] Changes in component dependencies are properly reflected in `gn`
- [ ] Code follows the [style guide](https://chromium.googlesource.com/chromium/src/+/HEAD/styleguide/c++/c++.md)
- [ ] Test plan is specified in PR before merging

## After-merge Checklist:

- [ ] The associated issue milestone is set to the smallest version that the
  changes has landed on
- [ ] All relevant documentation has been updated, for instance:
  - [ ] https://github.com/brave/brave-browser/wiki/Deviations-from-Chromium-(features-we-disable-or-remove)
  - [ ] https://github.com/brave/brave-browser/wiki/Proxy-redirected-URLs
  - [ ] https://github.com/brave/brave-browser/wiki/Fingerprinting-Protections
[...]

PR #32950, opened 2025-12-24
Remove obsolete setting from search settings

### LINKED ISSUE #51656: Remove obsolete `Index other search engines` setting from brave://settings/search
## Test plan
1. Open Brave and visit brave://settings/search
2. Verify there is no `Index other search engines` option. It should have been removed.

## Description
Originally, Chromium would be auto-discovering search engines for sites that people visited. For example, if you went to homedepot.com and did a search, it would save this. When someone went to brave://settings/search, you'd be able to see all the sites they visited and these sites would interfere with the omnibox behavior too.

We had created https://github.com/brave/brave-browser/issues/14496 to track this and put the logic behind a flag (default disabled).  This rework on our side was done with https://github.com/brave/brave-core/pull/9602.

Image

Upstream, this unwanted behavior was tracked with https://issues.chromium.org/issues/40088547.  [Eventually, the search settings page was reworked and this behavior was removed](https://issues.chromium.org/issues/40088547#comment92).

We should also delete our patch as it's not doing anything anymore.

bsclifton:
Fixes https://github.com/brave/brave-browser/issues/51656

For more information on why this is obsolete, see https://github.com/brave/brave-browser/issues/14496#issuecomment-3690503501

This is basically reverting:
- https://github.com/brave/brave-core/pull/9602
- https://github.com/brave/brave-core/pull/23659


bridiver (approved):

bsclifton:
Going to ++ myself as a fellow string reviewer; since it's just the removal of strings

antoinelibert:
Hello,

Sorry if this is not the right place to continue discussing this feature.

I would like Brave to reconsider reverting and put back the setting to disable [Open Search auto discovery feature](https://github.com/brave/brave-browser/issues/14496).

According to the issue, [upstream chromium "removed](https://github.com/brave/brave-browser/issues/14496#issuecomment-3690503501) the feature but that is not what happened. The search engine still get added but they are now in disabled state by default instead. They still pollute the user session and obviously the privacy problem of the original issue still exists (search engine get added with no user confirmation and be used as a browsing history).

Thank you.

Brave Origin 1.92.134 Chromium: 150.0.7871.63 (Official Build) (64-bit)

DerrickRice:
@antoinelibert , we're investigating in https://github.com/brave/brave-browser/issues/58688.

PR #33000, opened 2026-01-02
[iOS] NTP fav page icons are being incorrectly cached

### LINKED ISSUE #50741: [iOS] NTP fav page icons are being incorrectly cached
### Description

All `google.com/<site>` fav icons displayed on NTP always use the icon of the most recently used site. This issue is iOS-specific (not reproducible on desktop).

### Steps to reproduce

1. Add [google.com](https://google.com), [google.com/finance](https://google.com/finance), and [google.com/flights](https://google.com/flights) to NTP fav sites (in this particular order)
2. Open each of them and check what icons get displayed on NTP afterwards

### Actual result

The screenshots were taken in mid-August, but the problem persists to this day.
 
1. All the fav site icons changed to the Google Flights icon (the Google service I visited last; 1st screenshot)
2. After opening Google Search, all the icons were updated accordingly (2nd screenshot)
 
Image

---

Image

### Expected result

Each site should use its own unique icon.

### Reproduces how often

Easily reproduced

### Brave version

Brave iOS 1.83.118
Brave iOS Nightly 1.86.58

### Device/iOS version

iPhone 17 Pro
iOS 26.0.1

### Affected browser versions

- [x] latest AppStore
- [x] latest TestFlight
- [x] previous TestFlight

### Reproducibility

### Miscellaneous information

https://bravesoftware.slack.com/archives/C06UXF3KJ/p1755519802968999

nuo-xu:
Fix favicon is not cached correctly for sites have same domain but different favicons with different fragmented parts.
We used to cache favicon only using domain value not the full url. This will cause some same domain sites that have different favicons based on their query parameters or fragmented parts. 
We did this this way to avoid issues when some websites still only have a favicon for the FULL url including the fragmented parts, but no favicon for their domain. Back then, we were only experiencing this scenario for `https://app.uniswap.org/`. But now they have fixed. 
This change will start caching favicons using the FULL urls. As long as user visit any site, the favicon of this site will be cached with its FULL url and will be fetched later using its FULL url.

Resolves https://github.com/brave/brave-browser/issues/50741

Test:
1. open a build without this change and visiting some sites have favicons that don't share domain. Save them as a Favorite. (You can only see 4 Favorites in NTP. Full list is revealed in the edit mode by tapping the omnibox. Verify all favicons are displayed correctly.
2. add some sites as Favorites that share the same domain but have different favicons. For example:  google.com, google.com/finance, google.com/flights
3. Verify all the sites's favicon in step 2 are displayed as the last visited site's favicon
4. Upgrade this build to a build that has this PR's changes
5. visit those Favorites in step 2 
6. Verify both Favorite favicons in step 1 and step 2 are displayed correctly in NTP and edit mode 
7. Relaunch the build and verify again.

### DIFF
--- a/ios/brave-ios/Sources/Favicon/FaviconFetcher.swift
+++ b/ios/brave-ios/Sources/Favicon/FaviconFetcher.swift
@@ -108,6 +108,14 @@ public class FaviconFetcher {
       return favicon
     }
 
+    // We stored favicons using `url.domain` before.
+    // Now we are using the full url.
+    // Try to fetch with url.domain if favicon has not yet
+    // updated with the full url.
+    if let favicon = await getFromCache(for: url.domainURL) {
+      return favicon
+    }
+
     // When we search for a domain in the URL bar,
     // it automatically makes the scheme `http`
     // Even if the website loads/redirects as `https`
@@ -127,9 +135,8 @@ public class FaviconFetcher {
   /// Updates the Favicon in the cache with the specified icon if any, otherwise removes the favicon from the cache.
   public static func updateCache(_ favicon: Favicon?, for url: URL, persistent: Bool) async {
     guard let favicon, !favicon.isMonogramImage else {
-      let cachedURL = cacheURL(for: url)
-      SDImageCache.shared.memoryCache.removeObject(forKey: cachedURL.absoluteString)
-      SDImageCache.shared.diskCache.removeData(forKey: cachedURL.absoluteString)
+      SDImageCache.shared.memoryCache.removeObject(forKey: url.absoluteString)
+      SDImageCache.shared.diskCache.removeData(forKey: url.absoluteString)
       return
     }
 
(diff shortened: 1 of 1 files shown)


ios/brave-ios/Sources/Favicon/FaviconFetcher.swift
```diff
@@ -121,6 +121,14 @@ public class FaviconFetcher {
       return favicon
     }
 
+    // We stored favicons using `url.domain` before.
+    // Now we are using the full url.
+    // Try to fetch with url.domain if favicon has not yet
+    // updated with the full url.
+    if let favicon = await getFromCache(for: url.domainURL) {
```
kylehickinson, line 115: Do we even need the `SDImageCache` aspect anymore? Is using the Chromium fetcher for all favicons not viable?

nuo-xu: this is in case the new fetched icon has not yet cached. and we check if there is a cache favicon with the old key.

kylehickinson: Sure but I mean wont it already be cached in Chromium?

netzenbot (approved):

ios/brave-ios/Sources/Favicon/FaviconFetcher.swift
nuo-xu: this PR got forgotten lol
I think SDImageCache is still used for monogramIcon and those bundled favicon.

PR #33053, opened 2026-01-07
Fix new tab redirect before actual website

### LINKED ISSUE #51052: Opening a new tab redirects to an internal url first at "internal://local/about/home?uuidkey=..." before going to the actual web page
### Description

I've noticed this recently where opening a new tab redirects to an internal url first at "internal://local/about/home?uuidkey=..." before going to the actual web page. It is fairly quick so hard to spot but I've noticed this on both iOS and iPadOS versions of brave. Not sure at this point if this is a bug or a malware on my device.

### Steps to reproduce

1. Open brave browser on iOS or IpadOS (Ideally with network slowed down to get more time to see the redirect page)
2. Go to any web page either directly or over search.
3. It'd briefly redirect to "internal://local/about/home?uuidkey=..." before going to the actual web page

### Actual result

The browser goes to the internal url first before going to the actual website

### Expected result

It shouldn't have this internal redirect.

### Reproduces how often

Easily reproduced

### Brave version

Everytime I open the browser

### Device/iOS version

I've tested this on an iPhone 13 Pro running iOS 18.7.2 and brave browser version: 1.84

### Affected browser versions

- [x] latest AppStore
- [ ] latest TestFlight
- [ ] previous TestFlight

### Reproducibility

### Miscellaneous information

_No response_

EliHini:
Resolves [51052](https://github.com/brave/brave-browser/issues/51052)

## Problem
When opening a new tab with a URL (either directly or via search), the browser was briefly redirecting to `internal://local/about/home?uuidkey=...` before navigating to the actual web page. This can create a visible flash of the internal URL in the address bar and caused a poor user experience.

## Investigation: Root Cause
The `selectTab` method in `TabManager.swift` was incorrectly treating new tabs with a `nil` `visibleURL` as zombie tabs that needed restoration. When a new tab is created with a URL request:

1. The tab's `visibleURL` is `nil` temporarily before navigation commits
2. The code assumed this meant the tab was a zombie tab (a saved tab that needs restoration)
3. It would restore the tab with the New Tab Page (NTP) URL (`internal://local/about/home`)
4. On a slow enough network connection, this will cause the unwanted redirect before the actual URL loaded

## Solution
To ensure that new tabs with URL requests navigate directly to their target URL without the intermediate redirect to the internal about/home page I added additional checks in the `selectTab` method to only restore tabs that are actually zombie tabs by:

1. **Checking if tab is loading** (`!selectedTab.isLoading`) - Tabs with pending navigations should not be restored
2. **Check for SessionTab entry** (`SessionTab.from(tabId: selectedTab.id) != nil`) - Only restore tabs that have been saved to storage


ios/brave-ios/Sources/Brave/Frontend/Browser/TabManager.swift
```diff
@@ -383,8 +383,14 @@ class TabManager: NSObject {
...
     }
 
+    // Only restore zombie tabs that actually need restoration and do not restore those
+    // that are currently loading or have a pending navigation. They may have a temporary nil visibleURL
+    // before navigation commits.
+    // Finally check if the tab's webview has a URL, which ia an indicator of a pending navigation.
     if let selectedTab = selectedTab,
-      selectedTab.visibleURL == nil
+      selectedTab.visibleURL == nil,
+      !selectedTab.isLoading,
```
kylehickinson, line 392: Do we need an `isLoading` check here?

ios/brave-ios/Sources/Brave/Frontend/Browser/TabManager.swift
```diff
@@ -383,8 +383,14 @@ class TabManager: NSObject {
...
+      SessionTab.from(tabId: selectedTab.id) != nil
```
kylehickinson: Maybe worth checking that `SessionTab.from(tabId: selectedTab.id)?.interactionState != nil`

EliHini: good idea actually. Let me do a quick test on that

ios/brave-ios/Sources/Brave/Frontend/Browser/TabManager.swift
EliHini: Yeah we need it here because `visibleURL` can be temporarily `nil` while  `isLoading == true` during cross-origin navigation. So without this check we could restore tab a tab that's actively navigating, which would interrupt the navigation.

ios/brave-ios/Sources/Brave/Frontend/Browser/TabManager.swift
EliHini: Done

netzenbot (approved):

EliHini:
Closing because it has gone stale and there's a potential alternate solution

PR #33125, opened 2026-01-09
[Wallet] update code to cache origin-unique callbacks

### LINKED ISSUE #51897: EthereumProviderImpl should cache request callbacks from different origins
Right now the EthereumProviderImpl will reject _any_ incoming requests while there's a callback pending before Wallet unlock.

cmazakas:
Resolves https://github.com/brave/brave-browser/issues/51897

We persist a callback-per-unique-origin and then invoke them all once the Wallet is unlocked.

Tests are updated to cover this new codepath.

### DIFF
--- a/components/brave_wallet/browser/ethereum_provider_impl.cc
+++ b/components/brave_wallet/browser/ethereum_provider_impl.cc
@@ -161,6 +161,26 @@ std::vector<std::string> FilterAccountsByPermissionIdentifier(
 
 }  // namespace
 
+EthereumProviderImpl::PendingRequestEthereumPersmissionsCallback::
+    PendingRequestEthereumPersmissionsCallback() = default;
+
+EthereumProviderImpl::PendingRequestEthereumPersmissionsCallback::
+    ~PendingRequestEthereumPersmissionsCallback() = default;
+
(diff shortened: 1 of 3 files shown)


components/brave_wallet/browser/ethereum_provider_impl.cc
```diff
@@ -1244,18 +1264,24 @@ void EthereumProviderImpl::RequestEthereumPermissions(
   }
 
   if (keyring_service_->IsLockedSync()) {
-    if (pending_request_ethereum_permissions_callback_) {
+    if (pending_request_ethereum_permissions_callbacks_.contains(origin)) {
+      // Only reject requests coming from the same origin. Requests from other
+      // origins should be cached and invoked later.
```
supermassive, line 1269: EthereumProviderImpl is created for each frame for each page. If frame navigates to another origin new EthereumProviderImpl is created. So each EthereumProviderImpl is supposed to receive requests only from same origin. I don't think we need per origin mapping then

cmazakas: This is exactly what I thought as well, and because of this, we don't actually need this PR then because the code's behavior would be unchanged...

I'll get clarification on slack.

netzenbot (approved):

PR #33666, opened 2026-02-04
Fix BraveBrowserViewWithRoundedCornersTest.RoundedCornersForContentTest/1 for macos Tahoe

bridiver:
```
Expected equality of these values:
  rounded_corners_border_radius_at_window_corner
    Which is: 17
  contents_view_radii.lower_left()
    Which is: 6
```

f/u PR to https://github.com/brave/brave-core/pull/33131

Resolves

### DIFF
--- a/browser/ui/views/brave_layout_provider_mac.mm
+++ b/browser/ui/views/brave_layout_provider_mac.mm
@@ -11,9 +11,6 @@ int BraveLayoutProviderMac::GetCornerRadiusMetric(
     views::ShapeContextTokensOverride token) const {
   switch (token) {
     case views::ShapeContextTokensOverride::kRoundedCornersBorderRadius:
-      if (@available(macOS 26, *)) {
-        return 17;
-      }
       return 6;
     case views::ShapeContextTokensOverride::
         kRoundedCornersBorderRadiusAtWindowCorner:
--- a/browser/ui/views/brave_layout_provider_mac.mm
+++ b/browser/ui/views/brave_layout_provider_mac.mm
@@ -11,6 +11,9 @@ int BraveLayoutProviderMac::GetCornerRadiusMetric(
     views::ShapeContextTokensOverride token) const {
   switch (token) {
     case views::ShapeContextTokensOverride::kRoundedCornersBorderRadius:
+      if (@available(macOS 26, *)) {
+        return 17;
+      }
       return 6;
     case views::ShapeContextTokensOverride::
         kRoundedCornersBorderRadiusAtWindowCorner:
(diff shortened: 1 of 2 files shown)


browser/ui/views/brave_layout_provider_mac.mm
```diff
@@ -11,6 +11,9 @@
     views::ShapeContextTokensOverride token) const {
   switch (token) {
     case views::ShapeContextTokensOverride::kRoundedCornersBorderRadius:
+      if (@available(macOS 26, *)) {
```
bridiver: I don't know if this is the correct fix or if the test should be checking `kRoundedCornersBorderRadius` instead of `kRoundedCornersBorderRadiusAtWindowCorner`

simonhong: We don't need this change. `kRoundedCornersBorderRadius` should be always `6`.
Only we need different `kRoundedCornersBorderRadiusAtWindowCorner` per macOS version.

Sometimes ago, I tried to fix similar issue like this but different point - https://github.com/brave/brave-browser/issues/51913

Contents's roundness depends on UI status (vertical tab/sidebar visibility and split view enabled state).
This `RoundedCornersForContentTest` checks corrent roundness by changing these status.
I think having more `RunUntil` after changing UI could help this status.

simonhong: I'll update this PR with `RunUntil`.

bridiver: That is not correct. This test fails 100% of the time on Tahoe and it's not a race condition. If `kRoundedCornersBorderRadius` should always be 6 then the assertion is wrong because it's checking the value of `kRoundedCornersBorderRadiusAtWindowCorner`, but the code uses the value from `kRoundedCornersBorderRadius`

bridiver: I guess the wait I tried didn't work because there were issues further up in the test. As discussed `RunScheduledLayouts()` seems like the correct way to handle this

simonhong (approved):

fallaciousreasoning:
This is failing for me again
https://ci.brave.com/job/brave-core-build-pr-origin-macos-arm64/job/PR-36950/4/testReport/

simonhong:
> This is failing for me again https://ci.brave.com/job/brave-core-build-pr-origin-macos-arm64/job/PR-36950/4/testReport/

`BraveBrowserViewWithRoundedCornersTest.RoundedCornersWithSidePanelTest` was disabled recently.
https://github.com/brave/brave-core/pull/36910

PR #33986, opened 2026-02-17
[AIChat] Tab Management Tool gets "list" operation

petemill:
- Series https://github.com/brave/brave-browser/issues/48536

The tool was defined in:
- https://github.com/brave/brave-core/pull/33985

The follow-up implements more actions:
- https://github.com/brave/brave-core/pull/33987

This first action ("list") is implemented separately so that we can have the Profile class member and associated GN changes setup in a more minimal PR.

The list response is also good to implement as the first action because:
- It is the response provided to any action, to provide the current state of the tabs post-manipulation.
- There is enforcement that it is the first action asked for by the LLM
- This PR adds browser tests to verify the structure is as expected after any other action.

Data provided to the LLM as a response to any operation (including "list") includes each window's list of tabs and groups, each tab's url, title and state, and each group's color and title, and looks like:

```json
{
  "windows": [ {
     "active_tab_index": 5,
     "groups": {
     },
     "is_active": false,
     "tabs": [ {
        "index": 0,
        "is_active": false,
        "is_pinned": false,
        "tab_id": 1,
        "title": "about:blank",
        "url": "about:blank"
     }, {
        "index": 1,
        "is_active": false,
        "is_pinned": false,
        "tab_id": 2,
        "title": "a.test",
        "url": "https://a.test/"
     }, {
        "index": 2,
        "is_active": false,
        "is_pinned": false,
        "tab_id": 3,
        "title": "b.test",
        "url": "https://b.test/"
     }, {
        "index": 3,
        "is_active": false,
        "is_pinned": false,
        "tab_id": 16,
        "title": "remove.test",
        "url": "https://remove.test/"
     }, {
        "index": 4,
        "is_active": false,
        "is_pinned": false,
        "tab_id": 17,
        "title": "cross1.test",
        "url": "https://cross1.test/"
     }, {
        "index": 5,
        "is_active": true,
        "is_pinned": false,
        "tab_id": 19,
        "title": "another1.test",
        "url": "https://another1.test/"
     }, {
        "index": 6,
        "is_active": false,
        "is_pinned": false,
        "tab_id": 20,
        "title": "another2.test",
        "url": "https://another2.test/"
     }, {
        "index": 7,
        "is_active": false,
        "is_pinned": false,
        "tab_id": 18,
        "title": "cross2.test",
        "url": "https://cross2.test/"
     } ],
     "window_id": 1115739677
  }, {
     "active_tab_index": 2,
     "groups": {
     },
     "is_active": false,
     "tabs": [ {
        "index": 0,
        "is_active": false,
        "is_pinned": false,
        "tab_id": 4,
        "title": "about:blank",
        "url": "about:blank"
     }, {
        "index": 1,
        "is_active": false,
        "is_pinned": false,
        "tab_id": 5,
        "title": "c.test",
        "url": "https://c.test/"
     }, {
        "index": 2,
        "is_active": true,
        "is_pinned": false,
        "tab_id": 6,
        "title": "d.test",
        "url": "https://d.test/"
     } ],
     "window_id": 1115739681
  }, {
     "active_tab_index": 1,
     "groups": {
        "BECB3D4494AEF0C741BCAE1BE461E56A": {
           "color": "red",
           "is_collapsed": false,
           "title": "G1B"
        }
     },
     "is_active": true,
     "tabs": [ {
        "index": 0,
        "is_active": false,
        "is_pinned": false,
        "tab_id": 14,
        "title": "extra1.test",
        "url": "https://extra1.test/"
     }, {
        "index": 1,
        "is_active": true,
        "is_pinned": false,
        "tab_id": 15,
        "title": "extra2.test",
        "url": "https://extra2.test/"
     }, {
        "group_id": "BECB3D4494AEF0C741BCAE1BE461E56A",
        "index": 2,
        "is_active": false,
        "is_pinned": false,
        "tab_id": 10,
        "title": "example.test",
        "url": "https://example.test/"
     }, {
        "group_id": "BECB3D4494AEF0C741BCAE1BE461E56A",
        "index": 3,
        "is_active": false,
        "is_pinned": false,
        "tab_id": 11,
        "title": "example2.test",
        "url": "https://example2.test/"
     }, {
        "group_id": "BECB3D4494AEF0C741BCAE1BE461E56A",
        "index": 4,
        "is_active": false,
        "is_pinned": false,
        "tab_id": 7,
        "title": "one.test",
        "url": "https://one.test/"
     }, {
        "group_id": "BECB3D4494AEF0C741BCAE1BE461E56A",
        "index": 5,
        "is_active": false,
        "is_pinned": false,
        "tab_id": 12,
        "title": "move.test",
        "url": "https://move.test/"
     }, {
        "group_id": "BECB3D4494AEF0C741BCAE1BE461E56A",
        "index": 6,
        "is_active": false,
        "is_pinned": false,
        "tab_id": 13,
        "title": "move-window.test",
        "url": "https://move-window.test/"
     } ],
     "window_id": 1115739692
  }, {
     "active_tab_index": 0,
     "groups": {
     },
     "is_active": false,
     "tabs": [ {
        "index": 0,
        "is_active": true,
        "is_pinned": false,
        "tab_id": 21,
        "title": "about:blank",
        "url": "about:blank"
     } ],
     "window_id": 1115739700
  }, {
     "active_tab_index": 0,
     "groups": {
     },
     "is_active": true,
     "tabs": [ {
        "index": 0,
        "is_active": true,
        "is_pinned": false,
        "tab_id": 24,
        "title": "",
        "url": "https://move-all.test/"
     }, {
        "index": 1,
        "is_active": false,
        "is_pinned": false,
        "tab_id": 23,
        "title": "",
        "url": "https://move-all.test/"
     }, {
        "index": 2,
        "is_active": false,
        "is_pinned": false,
        "tab_id": 22,
        "title": "",
        "url": "https://move-all.test/"
     } ],
     "window_id": 1115739705
  } ]
}
```

### DIFF
--- a/browser/ai_chat/BUILD.gn
+++ b/browser/ai_chat/BUILD.gn
@@ -194,7 +194,6 @@ if (enable_ai_chat_tab_management_tool) {
       "//chrome/browser/profiles:profile",
       "//chrome/browser/profiles:profile_util",
       "//chrome/browser/ui",
-      "//chrome/browser/ui:browser_list",
       "//chrome/browser/ui/browser_window",
       "//chrome/browser/ui/tabs:tab_strip",
       "//components/tab_groups",
--- a/browser/ai_chat/BUILD.gn
+++ b/browser/ai_chat/BUILD.gn
@@ -38,10 +38,7 @@ static_library("ai_chat") {
   ]
 
   if (enable_ai_chat_tab_management_tool) {
-    sources += [
-      "tools/tab_management_tool.cc",
-      "tools/tab_management_tool.h",
-    ]
+    sources += [ "tools/tab_management_tool.h" ]
   }
 
   public_deps = [ "//brave/components/restricted_web_contents_delegate" ]
@@ -178,6 +175,36 @@ if (enable_print_preview) {
   }
 }
 
+if (enable_tab_management_tool) {
+  source_set("ai_chat_tab_management_tool_impl") {
+    visibility = [
+      ":browser_tests",
+      ":unit_tests",
+      "//chrome/browser",
+    ]
+
+    sources = [ "tools/tab_management_tool.cc" ]
+
+    public_deps = [ ":ai_chat" ]
+
+    deps = [
+      "//base",
+      "//brave/components/ai_chat/core/browser",
+      "//brave/components/ai_chat/core/common/mojom",
+      "//chrome/browser/profiles:profile",
+      "//chrome/browser/profiles:profile_util",
+      "//chrome/browser/ui",
+      "//chrome/browser/ui:browser_list",
+      "//chrome/browser/ui/browser_window",
+      "//chrome/browser/ui/tabs:tab_strip",
+      "//components/tab_groups",
+      "//components/tabs:public",
+      "//content/public/browser",
+      "//services/data_decoder/public/cpp",
+    ]
+  }
+}
+
 if (enable_brave_ai_chat_agent_profile) {
   source_set("ai_chat_agent_profile") {
     sources = [ "ai_chat_agent_profile_helper.h" ]
@@ -287,6 +314,7 @@ source_set("unit_tests") {
 
   if (enable_ai_chat_tab_management_tool) {
     sources += [ "tools/tab_management_tool_unittest.cc" ]
+    deps += [ "//brave/browser/ai_chat:ai_chat_tab_management_tool_impl" ]
   }
 
   if (!is_android) {
@@ -401,6 +429,15 @@ source_set("browser_tests") {
       "//ui/accessibility:ax_base",
       "//ui/accessibility:ax_features_mojo",
     ]
+
+    if (enable_tab_management_tool) {
+      sources += [ "tools/tab_management_tool_browsertest.cc" ]
+      deps += [
+        ":ai_chat_tab_management_tool_impl",
+        "//chrome/browser/ui/tabs:tabs_public",
+        "//components/tab_groups",
+      ]
+    }
 
     if (enable_print_preview) {
       sources += [
--- a/browser/ai_chat/BUILD.gn
+++ b/browser/ai_chat/BUILD.gn
@@ -175,7 +175,7 @@ if (enable_print_preview) {
   }
 }
 
-if (enable_tab_management_tool) {
+if (enable_ai_chat_tab_management_tool) {
   source_set("ai_chat_tab_management_tool_impl") {
     visibility = [
       ":browser_tests",
@@ -430,7 +430,7 @@ source_set("browser_tests") {
       "//ui/accessibility:ax_features_mojo",
     ]
 
-    if (enable_tab_management_tool) {
+    if (enable_ai_chat_tab_management_tool) {
       sources += [ "tools/tab_management_tool_browsertest.cc" ]
       deps += [
         ":ai_chat_tab_management_tool_impl",
--- a/browser/ai_chat/browser_tool_provider.h
+++ b/browser/ai_chat/browser_tool_provider.h
@@ -14,6 +14,8 @@
 #include "brave/components/ai_chat/core/browser/tools/tool_provider.h"
 #include "brave/components/ai_chat/core/common/buildflags/buildflags.h"
 
+class Profile;
+
 namespace content {
 class BrowserContext;
 }
--- a/browser/ai_chat/browser_tool_provider.h
+++ b/browser/ai_chat/browser_tool_provider.h
@@ -29,7 +29,7 @@ class TabManagementTool;
 // that the tools for a conversation perform actions on.
 class BrowserToolProvider : public ToolProvider {
  public:
-  explicit BrowserToolProvider(content::BrowserContext* browser_context);
+  explicit BrowserToolProvider(Profile* profile);
 
   ~BrowserToolProvider() override;
 
@@ -47,6 +47,7 @@ class BrowserToolProvider : public ToolProvider {
 #if BUILDFLAG(ENABLE_AI_CHAT_TAB_MANAGEMENT_TOOL)
   std::unique_ptr<TabManagementTool> tab_management_tool_;
 #endif
+  raw_ptr<Profile> profile_ = nullptr;
 };
 
 }  // namespace ai_chat
--- a/browser/ai_chat/tools/tab_management_tool.cc
+++ b/browser/ai_chat/tools/tab_management_tool.cc
@@ -12,7 +12,6 @@
 
 #include "base/json/json_reader.h"
 #include "base/json/json_writer.h"
-#include "base/strings/string_number_conversions.h"
 #include "base/strings/utf_string_conversions.h"
 #include "base/values.h"
 #include "brave/components/ai_chat/core/browser/tools/tool_input_properties.h"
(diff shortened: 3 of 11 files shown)


browser/ai_chat/BUILD.gn
```diff
@@ -155,6 +152,36 @@ if (enable_print_preview) {
   }
 }
 
+if (enable_tab_management_tool) {
```
netzenbot: Bug: This uses `enable_tab_management_tool` but the buildflag is defined as `enable_ai_chat_tab_management_tool` (in buildflags.gni from #33985). The same mismatch appears at line 384 in this file and line 757 in `browser/sources.gni`. This source_set will never be compiled. [best practice](https://github.com/brave-experiments/brave-core-tools/tree/master/docs/best-practices/build-system.md#-conditional-compilation)

petemill: This was not fixed after rebase, yes, but I think you're mistaken that this would not cause a compile error - it does for me. Perhaps therefore it shouldn't be in the best practices document as it's as obvious as saying that variable uses should not be misspelled from the variable definitions.

bbondy: GN does error on undefined variables, so the claim in the best practice that this would be "silently never compiled" was incorrect. I've removed that rule: https://github.com/brave-experiments/brave-core-bot/pull/25

The bug observation on this PR was valid (mismatched variable name), but it shouldn't have been elevated to a best practice since the build system already catches it.

browser/ai_chat/browser_tool_provider.h
```diff
@@ -29,7 +31,7 @@ class TabManagementTool;
 // that the tools for a conversation perform actions on.
 class BrowserToolProvider : public ToolProvider {
  public:
-  explicit BrowserToolProvider(content::BrowserContext* browser_context);
+  explicit BrowserToolProvider(Profile* profile);
```
netzenbot, line 34: Consider whether `Profile*` coupling is necessary here — a lighter dependency (like `PrefService*` or a specific interface) would improve testability and reduce the dependency surface. [best practice](https://github.com/brave-experiments/brave-core-tools/tree/master/docs/best-practices/architecture.md#dependency-injection)

browser/ai_chat/tools/tab_management_tool.cc
```diff
@@ -11,19 +11,63 @@
...
+    case tab_groups::TabGroupColorId::kPink:
+      return "pink";
+    case tab_groups::TabGroupColorId::kPurple:
+      return "purple";
+    case tab_groups::TabGroupColorId::kCyan:
+      return "cyan";
+    case tab_groups::TabGroupColorId::kOrange:
+      return "orange";
+    default:
+      return "unknown";
```
netzenbot: nit: consider using `NOTREACHED()` in the default case to catch unexpected enum values. [best practice](https://github.com/brave-experiments/brave-core-tools/tree/master/docs/best-practices/coding-standards.md#use-check-for-impossible-conditions)

browser/ai_chat/browser_tool_provider.h
```diff
@@ -47,6 +49,7 @@ class BrowserToolProvider : public ToolProvider {
 #if BUILDFLAG(ENABLE_AI_CHAT_TAB_MANAGEMENT_TOOL)
   std::unique_ptr<TabManagementTool> tab_management_tool_;
 #endif
+  raw_ptr<Profile> profile_ = nullptr;
```
netzenbot, line 52: nit: raw_ptr members benefit from a `// not owned` or similar ownership comment for clarity.

browser/ai_chat/tools/tab_management_tool.h
```diff
@@ -36,6 +37,14 @@ class TabManagementTool : public Tool {
...
 
  private:
+  // Action handlers
+  void HandleListTabs(UseToolCallback callback);
+
+  // Helper to generate tab list that can be reused by all handlers
+  base::Value::Dict GenerateTabList() const;
+  // Profile with which to restrict all window and tab operations
+  raw_ptr<Profile> profile_ = nullptr;
+
```
netzenbot, line 46: nit: raw pointer members should be grouped together and typically placed after owned members for consistency.

browser/ai_chat/tools/tab_management_tool_browsertest.cc
```diff
@@ -0,0 +1,193 @@
...
+  SCOPED_TRACE(testing::Message() << location.ToString());
+  base::Value out_val = base::test::ParseJson(out_json);
+  EXPECT_THAT(out_val, base::test::IsSupersetOfValue(
+                           BuildExpectedOutputSubset(profile)));
+}
+
+}  // namespace
+
+class TabManagementToolBrowserTest : public InProcessBrowserTest {
+ public:
```
netzenbot, line 136: nit: prefer `PlatformBrowserTest` over `InProcessBrowserTest` for cross-platform compatibility. [best practice](https://github.com/brave-experiments/brave-core-tools/tree/master/docs/best-practices/testing-async.md#use-platformbrowsertest)

browser/ai_chat/browser_tool_provider.h
petemill: Profile is needed because the tool needs to filter browser windows to only the current profile. Use the context from other files in the PR before commenting.

browser/ai_chat/tools/tab_management_tool.cc
sangwoo108: Maybe not having `default:` case at all could be better, so that compiler can catch if there are unhandled cases.

browser/ai_chat/tools/tab_management_tool.cc
```diff
@@ -11,19 +11,63 @@
...
 
 namespace ai_chat {
 
-TabManagementTool::TabManagementTool() = default;
+namespace {
+
+std::string GetTabGroupColorString(tab_groups::TabGroupColorId color) {
+  switch (color) {
+    case tab_groups::TabGroupColorId::kGrey:
+      return "grey";
```
sangwoo108, line 46: Q. I'm curious we expect "string" so that the LLM can understand what color is.

---

Can we use [GetTabGroupColorLabelMap()](https://source.chromium.org/chromium/chromium/src/+/main:components/tab_groups/tab_group_color.h;l=60;drc=50dbcddad2f8e36ddfcec21d4551f389df425c37) instead?

sangwoo108 (commented):
LGTM!

---

I'm wondering if we have room for using browser_api component in the future(, which is under development at the moment) 

https://source.chromium.org/chromium/chromium/src/+/main:components/browser_apis/;bpv=1;bpt=0

browser/ai_chat/tools/tab_management_tool.cc
petemill: It's a good suggestion, but my issue with this is that those strings are translated for the user, whereas we want the agent to provide constant strings back to us for operations where it chooses a tab group color:
```cpp
       {"group_color",
        StringProperty(
            "Color for new or updated group",
            std::vector<std::string>{"grey", "blue", "red", "yellow", "green",
                                     "pink", "purple", "cyan", "orange"})},
```

And we'll use that value to map to `TabGroupColorId` (implemented in https://github.com/brave/brave-core/pull/33987). I don't think we'll want to do something like make a new reverse map (or lookup the existing map) because all the LLM instructions are in english and this isn't user-facing.

What do you think?

sangwoo108: sounds sensible! let's leave as is 👍

goodov (approved):

netzenbot (approved):

browser/ai_chat/tools/tab_management_tool_browsertest.cc
petemill: This test only runs on desktop

browser/ai_chat/browser_tool_provider.h
petemill: added ownership info

browser/ai_chat/tools/tab_management_tool.h
petemill: moved

browser/ai_chat/tools/tab_management_tool.cc
petemill: added

PR #33987, opened 2026-02-17
[AIChat] Tab Management Tool actions

petemill:
Implements tab movement, grouping, ~closing~ and window creation actions

- Series https://github.com/brave/brave-browser/issues/48536

Builds on:
- https://github.com/brave/brave-core/pull/33985
- https://github.com/brave/brave-core/pull/33986

Even though this is only 1 class being modified (4 files with tests), I know it's a lot of lines in the tab_management_tool (~800 with the rest being in tests). I did remove the setup (previous PRs) and tab closing action (follow-up, below) as they were easy to isolate. The rest seems more intertwined with internal functions. I'm happy to split it up more if difficult to review.

Follow-ups:
- https://github.com/brave/brave-core/pull/36894

### DIFF
--- a/browser/ai_chat/tools/tab_management_tool.cc
+++ b/browser/ai_chat/tools/tab_management_tool.cc
@@ -178,6 +178,12 @@ std::vector<ResolvedTab> ResolveTabs(const std::vector<int>& tab_ids,
   for (int handle_id : tab_ids) {
     auto* tab = tabs::TabHandle(handle_id).Get();
     if (!tab) {
+      // A handle that resolves to no live tab usually means the model is
+      // operating on a stale tab list. Surface it for diagnostics; we still
+      // proceed with whatever tabs did resolve rather than failing the whole
+      // call, since tabs can legitimately close between list and action.
+      DVLOG(2) << "TabManagementTool: tab id " << handle_id
+               << " did not resolve to a live tab (stale or invalid handle)";
       continue;
     }
     if (require_grouped && !tab->GetGroup().has_value()) {
@@ -589,6 +595,8 @@ BrowserWindowInterface* TabManagementTool::FindOrCreateTargetWindow(
     std::optional<int> window_id,
     std::string* error_out,
     bool* did_create_window_out) {
+  *did_create_window_out = false;
+
   if (!window_id.has_value()) {
     return nullptr;  // Use source window
   }
@@ -643,6 +651,7 @@ BrowserWindowInterface* TabManagementTool::FindOrCreateTargetWindow(
 BrowserWindowInterface* TabManagementTool::FindWindowWithGroup(
     const std::string& group_id,
     TabGroup** group_out) {
+  *group_out = nullptr;
   for (BrowserWindowInterface* browser : GetAllBrowserWindowInterfaces()) {
     if (browser->GetProfile() != profile_ ||
         browser->GetType() != BrowserWindowInterface::Type::TYPE_NORMAL) {
@@ -667,6 +676,10 @@ bool TabManagementTool::ValidateMoveTarget(
     std::optional<int> index,
     std::string* error) const {
   CHECK(target_window) << "Don't call this function with a nullptr";
+  // All paths that produce a target window (FindOrCreateTargetWindow,
+  // FindWindowWithGroup, and a moved tab's own window) restrict to this
+  // profile, so a cross-profile target is a programming error.
+  CHECK_EQ(target_window->GetProfile(), profile_.get());
 
   TabStripModel* tab_strip = target_window->GetTabStripModel();
   if (!tab_strip) {
@@ -696,7 +709,7 @@ void TabManagementTool::HandleMoveGroup(UseToolCallback callback,
                                         const std::string& group_id,
                                         std::optional<int> window_id,
                                         std::optional<int> index) {
-  TabGroup* group;
+  TabGroup* group = nullptr;
   BrowserWindowInterface* source_window = FindWindowWithGroup(group_id, &group);
 
   if (!source_window || !group) {
@@ -1063,7 +1076,7 @@ void TabManagementTool::HandleUpdateGroup(UseToolCallback callback,
   bool found = false;
 
   // Find and update the group
-  TabGroup* group;
+  TabGroup* group = nullptr;
   if (BrowserWindowInterface* window =
           FindWindowWithGroup(*group_id_str, &group)) {
     tab_groups::TabGroupVisualData visual_data = ApplyGroupVisualOverrides(
--- a/browser/ai_chat/tools/tab_management_tool.cc
+++ b/browser/ai_chat/tools/tab_management_tool.cc
@@ -242,26 +242,26 @@ std::string_view TabManagementTool::Name() const {
 }
 
 std::string_view TabManagementTool::Description() const {
-  return "Manage browser tabs - list, move, close tabs and manage tab "
-         "groups. Only use this tool when the user explicitly requests tab "
+  return "Manage browser tabs - list and move tabs, move groups, close tabs "
+         "and manage tab groups. Only use this tool when the user explicitly "
+         "requests tab "
          "organization, grouping, moving, or closing tabs. Do not use this "
          "tool proactively or when the user's request can be answered without "
          "modifying their tabs. "
          "This tool can list all open tabs with their window, group, URL and "
          "title information, "
-         "move tabs or entire groups between windows or positions, "
+         "move individual tabs (move_tabs) or an entire group (move_group) "
+         "between windows or positions, "
          "close tabs, "
          "and create or modify tab groups. Groups are per-window, so make any "
          "moves before grouping. Use window_id=-1 to move tabs/groups to a new "
          "window. "
-         "Use move_group_id to move an entire group at once (cannot be "
-         "combined with tab_ids in the same request). "
+         "Use move_tabs to move individual tabs and move_group to move an "
+         "entire group at once. "
          "When moving tabs, the active tab state is preserved - if you move "
          "the active tab, "
          "it remains active in its new location. "
          "After each operation, the updated tab list is returned. "
-         "To move both individual tabs and groups, make separate move "
-         "requests. "
          "If possible and you know the operations and IDs ahead of time, "
          "try to make multiple parallel requests to use this tool without "
          "waiting for the answer. Every time this tool needs to be used, if "
@@ -277,8 +277,9 @@ std::optional<base::DictValue> TabManagementTool::InputProperties() const {
   return CreateInputProperties(
       {{"action", StringProperty("The action to perform",
                                  std::vector<std::string>{
-                                     "list", "move", "close", "create_group",
-                                     "update_group", "remove_from_group"})},
+                                     "list", "move_tabs", "move_group", "close",
+                                     "create_group", "update_group",
+                                     "remove_from_group"})},
        {"plan",
         StringProperty("Human readable plan of what the assistant intends to "
                        "do with the list of tabs and with the tab management "
@@ -286,34 +287,35 @@ std::optional<base::DictValue> TabManagementTool::InputProperties() const {
                        "list operation in a conversation and allows the user "
                        "to approve or deny the tab management operations.")},
        {"tab_ids",
-        ArrayProperty("List of tab IDs to operate on (for move, "
-                      "close, create_group, remove_from_group). Cannot be used "
-                      "with move_group_id.",
+        ArrayProperty("List of tab IDs to operate on (for move_tabs, "
+                      "close, create_group, remove_from_group).",
                       IntegerProperty("Tab ID"))},
-       {"move_group_id",
-        StringProperty("Group ID to move entirely (for move operation). "
-                       "Mutually exclusive with tab_ids - use separate "
-                       "requests to move both individual tabs and groups.")},
        {"window_id",
         IntegerProperty(
-            "Target window ID (for move operation). Use -1 to create a new "
-            "window. If group_id is provided, this will be inferred from that "
-            "group's window and window_id should not be provided.")},
+            "Destination window ID (for move_tabs and move_group). Use -1 to "
+            "create a new window. For move_tabs this is mutually exclusive "
+            "with "
+            "destination_group_id.")},
        {"group_id",
-        StringProperty("Target group ID for update_group or move (for move "
-                       "operation window ID will be inferred by the group "
-                       "window and is mutually exclusive) or group to "
-                       "update")},
+        StringProperty("The existing group to act on: the group to move "
+                       "(move_group) or the group to update (update_group).")},
+       {"destination_group_id",
+        StringProperty(
+            "Destination group to move tabs into (for move_tabs). Mutually "
+            "exclusive with window_id; the target window is inferred from this "
+            "group's window.")},
        {"index",
-        IntegerProperty("Target index position (for move operations)")},
+        IntegerProperty(
+            "Destination index position (for move_tabs and move_group).")},
        {"group_title", StringProperty("Title for new or updated group")},
        {"group_color",
         StringProperty(
             "Color for new or updated group",
             std::vector<std::string>{"grey", "blue", "red", "yellow", "green",
                                      "pink", "purple", "cyan", "orange"})},
-       {"add_to_end", BooleanProperty("Add tabs to end of group instead of "
-                                      "beginning (for move to group)")}});
+       {"add_to_end",
+        BooleanProperty("Add tabs to the end of the destination group instead "
+                        "of the beginning (for move_tabs into a group).")}});
 }
 
 std::optional<std::vector<std::string>> TabManagementTool::RequiredProperties()
@@ -400,17 +402,19 @@ void TabManagementTool::UseTool(const std::string& input_json,
   if (!action) {
     std::move(callback).Run(
         CreateContentBlocksForText(
-            "Missing required 'action' field. Must be one of: list, move, "
-            "close, "
-            "create_group, update_group, remove_from_group"),
+            "Missing required 'action' field. Must be one of: list, move_tabs, "
+            "move_group, close, create_group, update_group, "
+            "remove_from_group"),
         {});
     return;
   }
 
   if (*action == "list") {
     HandleListTabs(std::move(callback));
-  } else if (*action == "move") {
+  } else if (*action == "move_tabs") {
     HandleMoveTabs(std::move(callback), dict);
+  } else if (*action == "move_group") {
+    HandleMoveGroup(std::move(callback), dict);
   } else if (*action == "create_group") {
     HandleCreateGroup(std::move(callback), dict);
   } else if (*action == "update_group") {
@@ -420,8 +424,8 @@ void TabManagementTool::UseTool(const std::string& input_json,
   } else {
     std::move(callback).Run(
         CreateContentBlocksForText(
-            "Invalid action. Must be one of: list, move, close, "
-            "create_group, update_group, remove_from_group"),
+            "Invalid action. Must be one of: list, move_tabs, move_group, "
+            "close, create_group, update_group, remove_from_group"),
         {});
   }
 }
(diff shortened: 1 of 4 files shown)


browser/ai_chat/tools/tab_management_tool.h
fallaciousreasoning, line 1: this is a bit more of an architectural question, but for tab management we have a bunch of different actions that are controlled by one tool. 

However, in agentic mode, every action gets its own tool (i.e. there's no `MouseManagement` tool, there's a `MouseMove` and a `MouseClick` tool).

Do you think we should split this up to match those ones?

browser/ai_chat/tools/tab_management_tool.cc
```diff
@@ -331,4 +538,600 @@ void TabManagementTool::HandleListTabs(UseToolCallback callback) {
...
+    }
+  }
+  return nullptr;
+}
+
+bool TabManagementTool::ValidateMoveTarget(
+    BrowserWindowInterface* target_window,
+    std::optional<int> index,
+    std::string* error) const {
+  CHECK(target_window) << "Don't call this function with a nullptr";
```
fallaciousreasoning, line 819: Not sure how we'd manage it, but should we also CHECK its using the same profile?

browser/ai_chat/tools/tab_management_tool.cc
```diff
@@ -40,6 +52,56 @@ namespace ai_chat {
...
+// provided TabStripModel. Any indices outside of the current tab bounds are
+// dropped.
+std::vector<int> MakeSortedUniqueValidIndices(const std::vector<int>& indices,
+                                              const TabStripModel* tab_strip) {
+  CHECK(tab_strip);
+  std::vector<int> filtered_indices;
+  const int tab_count = tab_strip->count();
+  filtered_indices.reserve(indices.size());
+  for (int index : indices) {
+    if (index >= 0 && index < tab_count) {
```
fallaciousreasoning, line 70: optional nit: Is there a reason we don't filter duplicates here rather than having a separate uniquify step down below? probably extremely marginal memory/perf improvement?

browser/ai_chat/tools/tab_management_tool.cc
```diff
@@ -65,6 +127,102 @@ std::string GetTabGroupColorString(tab_groups::TabGroupColorId color) {
...
+// it receives the TabStripModel of the first valid tab in input order, which
+// callers use to choose a destination window.
+base::flat_map<TabStripModel*, std::vector<int>> ResolveTabIdsToStripIndices(
+    const base::ListValue& tab_ids,
+    Profile* profile,
+    bool require_grouped,
+    TabStripModel** first_strip_out) {
+  base::flat_map<TabStripModel*, std::vector<int>> by_strip;
+  for (const ResolvedTab& resolved :
+       ResolveTabs(HandleIdsFromList(tab_ids), profile, require_grouped)) {
```
fallaciousreasoning, line 223: Should we log if we aren't getting the same number of ids out of tab_ids as we put in?

fallaciousreasoning: or fail the tool call? Probably indicates the LLM is doing something funky

browser/ai_chat/tools/tab_management_tool.cc
```diff
@@ -187,10 +346,14 @@ void TabManagementTool::UserPermissionGranted(const std::string& tool_use_id) {
 
 void TabManagementTool::UseTool(const std::string& input_json,
                                 UseToolCallback callback) {
+  // TODO: keep sidebar open to a AI Chat conversation if it's open in the
```
fallaciousreasoning: nit: file an issue for this and link it here

browser/ai_chat/tools/tab_management_tool.cc
```diff
@@ -322,6 +493,42 @@ base::DictValue TabManagementTool::GenerateTabList() const {
   return result;
 }
 
+void TabManagementTool::SendResultWithTabList(UseToolCallback callback,
+                                              base::DictValue result) {
+  base::DictValue tab_list = GenerateTabList();
+  result.Set("windows", std::move(*tab_list.FindList("windows")));
```
fallaciousreasoning, line 600: windows is the only bit we're interested in from |result|?

browser/ai_chat/tools/tab_management_tool.cc
```diff
@@ -331,4 +538,600 @@ void TabManagementTool::HandleListTabs(UseToolCallback callback) {
 
   std::move(callback).Run(CreateContentBlocksForText(json_output), {});
 }
+
+void TabManagementTool::HandleMoveTabs(UseToolCallback callback,
+                                       const base::DictValue& params) {
+  const auto* tab_ids = params.FindList("tab_ids");
+  const auto* group_to_move = params.FindString("move_group_id");
```
fallaciousreasoning: similar to the comment on the header but at a lower level - should we separate this at the `action` level so we can avoid this?

`move_tabs` and `move_group`?

It'd be one less level of unions :D

browser/ai_chat/tools/tab_management_tool.cc
```diff
@@ -331,4 +538,600 @@ void TabManagementTool::HandleListTabs(UseToolCallback callback) {
...
+    // their active conversation tab.
+    target_browser->GetWindow()->ShowInactive();
+    *did_create_window_out = true;
+    // Get BrowserWindowInterface from session ID of the new browser
+    return BrowserWindowInterface::FromSessionID(
+        target_browser->GetSessionID());
+  }
+
+  // Find existing window using BrowserWindowInterface
+  SessionID target_session_id = SessionID::FromSerializedValue(*window_id);
```
fallaciousreasoning, line 768: should we explicitly set `did_create_window_out = false` in this branch, to guarantee we set it? Not sure about C++ but C# requires you to set out params and it does seem like good hygeine

browser/ai_chat/tools/tab_management_tool.cc
```diff
@@ -331,4 +538,600 @@ void TabManagementTool::HandleListTabs(UseToolCallback callback) {
...
+  // end of the tab strip.
+  return true;
+}
+
+void TabManagementTool::HandleMoveGroup(UseToolCallback callback,
+                                        const std::string& group_id,
+                                        std::optional<int> window_id,
+                                        std::optional<int> index) {
+  TabGroup* group;
+  BrowserWindowInterface* source_window = FindWindowWithGroup(group_id, &group);
```
fallaciousreasoning: `FindWindowWithGroup` doesn't guarantee we set `group` - I think it its okay because we check `!source_window` first but I think we should either:
1. Set `group = nullptr` when we initialize
2. Set `group = nullptr` when we return from FindWindowWithGroup = nullptr

browser/ai_chat/tools/tab_management_tool.cc
```diff
@@ -331,4 +538,600 @@ void TabManagementTool::HandleListTabs(UseToolCallback callback) {
...
+      continue;
+    }
+    for (const auto& g : tab_strip->group_model()->ListTabGroups()) {
+      if (g.ToString() == group_id) {
+        *group_out = tab_strip->group_model()->GetTabGroup(g);
+        return BrowserWindowInterface::FromSessionID(browser->GetSessionID());
+      }
+    }
+  }
+  return nullptr;
```
fallaciousreasoning, line 812: probably want to set `group_out = nullptr` here to guarantee the variable isn't uninitialized

browser/ai_chat/tools/tab_management_tool.cc
```diff
@@ -331,4 +538,600 @@ void TabManagementTool::HandleListTabs(UseToolCallback callback) {
...
+}
+
+void TabManagementTool::HandleMoveIndividualTabs(
+    UseToolCallback callback,
+    const std::vector<int>& tab_handles,
+    std::optional<int> window_id,
+    std::optional<int> index,
+    const std::string* group_id,
+    bool add_to_end) {
+  if (group_id && !group_id->empty() && window_id.has_value()) {
```
fallaciousreasoning: similarly, should this be split into `move_to_window` and `move_to_group` actions?

browser/ai_chat/tools/tab_management_tool.h
petemill: In general, the fewer the tools the better so that the context isn't polluted. Claude has computer use as a single tool, but I implemented ours as multiple tools to match chromium's tool definitions and simplify the argument validation. I think there is a roadmap item to see about reducing those tools to a single "browser tab use" tool. So I tried it this way for the tab management tool from the start, and it has been good in my testing.

browser/ai_chat/tools/tab_management_tool.cc
petemill: yeah, added `CHECK` considering that it should already be filtered

browser/ai_chat/tools/tab_management_tool.cc
```diff
@@ -331,4 +544,607 @@ void TabManagementTool::HandleListTabs(UseToolCallback callback) {
...
+    PostTaskSendResultWithTabList(std::move(callback), std::move(result),
+                                  tab_to_reactivate);
+  }
+}
+
+void TabManagementTool::HandleMoveIndividualTabs(
+    UseToolCallback callback,
+    const std::vector<int>& tab_handles,
+    std::optional<int> window_id,
+    std::optional<int> index,
```
netzenbot: The 'already at target position' early-return runs the callback synchronously, while the actual-move path defers via `PostTaskSendResultWithTabList`. [ARCH-060](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-060) requires that for a given callback execution must be consistently sync or async. Consider routing through `PostTaskSendResultWithTabList` here too (passing an empty `std::nullopt` for `active_moved_tab`), or add a comment explaining why the no-op case deliberately skips the task post.

browser/ai_chat/tools/tab_management_tool_browsertest.cc
```diff
@@ -191,6 +299,531 @@ IN_PROC_BROWSER_TEST_F(TabManagementToolBrowserTest, TabManagementToolTest) {
...
+        "tab_ids": [%d],
+        "group_title": "TestGroup"
+      })JSON",
+                                                        a_handle));
+    // Verify the tab is grouped and the returned group id is valid
+    base::DictValue create_dict = base::test::ParseJsonDict(create_response);
+    const std::string* created_group_id =
+        create_dict.FindString("created_group_id");
+    ASSERT_TRUE(created_group_id);
+    ASSERT_FALSE(created_group_id->empty());
```
netzenbot, line 370: `EXPECT_TRUE(tab->GetGroup().has_value())` is non-fatal, so if it fails execution continues to `tab->GetGroup().value().ToString()`, which throws `std::bad_optional_access` and crashes the test process. Use `ASSERT_TRUE(tab->GetGroup().has_value())` here so the test stops on failure before reaching the dereference. See https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-025

browser/ai_chat/tools/tab_management_tool.cc
```diff
@@ -40,6 +52,56 @@ namespace ai_chat {
 
 namespace {
```
netzenbot, line 54: nit: `const std::vector<int>&` for a read-only parameter should be `base::span<const int>` — spans accept any contiguous container and avoid forcing callers to use `std::vector`. See [CSA-017](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-017).

browser/ai_chat/tools/tab_management_tool.h
```diff
@@ -40,16 +44,61 @@ class TabManagementTool : public Tool {
...
+
+  // Helper to take window param and validate then find or create a window
+  BrowserWindowInterface* FindOrCreateTargetWindow(std::optional<int> window_id,
+                                                   std::string* error_out,
+                                                   bool* did_create_window_out);
+  // Provides the window and the group after validation from an incoming group
+  // id param
+  BrowserWindowInterface* FindWindowWithGroup(const std::string& group_id,
+                                              TabGroup** group_out);
+
```
netzenbot, line 64: nit: `bool ValidateMoveTarget(..., std::string* error)` fits the pattern replaced by `std::optional<std::string>` — return `std::nullopt` on success and the error string on failure. See [CSA-024](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-024).

browser/ai_chat/tools/tab_management_tool.cc
```diff
@@ -40,6 +52,56 @@ namespace ai_chat {
...
+    }
+  }
+
+  if (filtered_indices.empty()) {
+    return filtered_indices;
+  }
+
+  std::sort(filtered_indices.begin(), filtered_indices.end());
+  filtered_indices.erase(
+      std::unique(filtered_indices.begin(), filtered_indices.end()),
```
netzenbot: nit: This static string-to-enum mapping is a good fit for `base::MakeFixedFlatMap` per [CSA-045](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-045). A `constexpr auto kColorMap = base::MakeFixedFlatMap<std::string_view, tab_groups::TabGroupColorId>(...)` would replace the if/else chain with a compile-time verified lookup.

browser/ai_chat/tools/tab_management_tool.cc
petemill: hmm, at first I thought accumulating in to a set here would be worse performance. But then found a way to do the filtering, sorting and de-duping in 1 step. It's not the easiest read, but maybe it's better?

```cpp
  // The incoming values are positions within |tab_strip|, so they are bounded
  // by [0, tab_count). Marking the valid ones in a bitmap keyed on the index
  // and then reading them back in order filters, de-duplicates, and sorts in a
  // single linear pass, with no comparison sort.
  std::vector<bool> is_valid_index(tab_count, false);
  for (int index : indices) {
    if (index >= 0 && index < tab_count) {
      is_valid_index[index] = true;
    }
  }

  std::vector<int> result;
  for (int i = 0; i < tab_count; ++i) {
    if (is_valid_index[i]) {
      result.push_back(i);
    }
  }
  return result;
```

fc5f52a6e76

browser/ai_chat/tools/tab_management_tool.cc
petemill: It's expected that sometimes it will be different because tabs can close whilst the LLM is generating. I added a `DVLOG(2)` for it.

browser/ai_chat/tools/tab_management_tool.cc
petemill: Thanks for the reminder - added issue https://github.com/brave/brave-browser/issues/56024

browser/ai_chat/tools/tab_management_tool.cc
petemill: This is adding the `windows` property to `result` - adding to whatever the operation also sets, usually `message` and operation-specific properties, e.g. `new_window_id`.

`windows` is the only property on `tab_list`. Not sure why that doesn't return a List, but I guess that comes from the _list_ operation (`HandleListTabs`) sending that result directly and not going via `SendResultWithTabList`, as well as reserving ability to set other fields that aren't just windows in the future (sidebars??).

browser/ai_chat/tools/tab_management_tool.cc
petemill: done

browser/ai_chat/tools/tab_management_tool.cc
petemill: Initialized `TabGroup* group = nullptr` (also did the same in `HandleUpdateGroup`).

browser/ai_chat/tools/tab_management_tool.cc
petemill: I set `*group_out = nullptr` at function entry

browser/ai_chat/tools/tab_management_tool.cc
petemill: This one I don't feel as good about - I think it's simple for the LLM to decide whether the target is a window or a group, especially with `window_id` and the new `destination_group_id` params, compared to the previous complexity of the subject item: whether to move tabs or a group. And then this could balloon: should we have `move_tabs_to_position`, and `move_group_to_position`? (there are 3 valid destinations for tabs - window, group, index, and 2 for groups - window, index)

browser/ai_chat/tools/tab_management_tool.cc
petemill: Good suggestion - added `move_tabs` action, and `destination_group_id` param. I considered re-using the `group_id` param but that might add some confusion as it's the subject of some actions and would now be the target in this action.
e226bce8da8

browser/ai_chat/tools/tab_management_tool_browsertest.cc
petemill: done

browser/ai_chat/tools/tab_management_tool.cc
petemill: done

browser/ai_chat/tools/tab_management_tool.cc
petemill: changed to map

browser/ai_chat/tools/tab_management_tool.cc
petemill: done but there are some early returns which I didn't change. I think the expectation with most Tool is that they early return and call the callback immediately for errors.

browser/ai_chat/tools/tab_management_tool.cc
fallaciousreasoning: I actually worked this out later in the review but forgot to reply to myself sorry :laughing:

browser/ai_chat/tools/tab_management_tool.cc
fallaciousreasoning: happy to defer to you on this :smile:

browser/ai_chat/tools/tab_management_tool.cc
fallaciousreasoning: seems fine to me

fallaciousreasoning (approved):

browser/ai_chat/tools/tab_management_tool.cc
```diff
@@ -40,6 +53,55 @@ namespace ai_chat {
 
 namespace {
 
+// Returns a sorted, de-duplicated list of indices that are valid for the
+// provided TabStripModel. Any indices outside of the current tab bounds are
+// dropped.
+std::vector<int> MakeSortedUniqueValidIndices(base::span<const int> indices,
```
sangwoo108, line 59: Should `indices` be `const &` of span?

browser/ai_chat/tools/tab_management_tool.cc
```diff
@@ -40,6 +53,55 @@ namespace ai_chat {
...
+    }
+  }
+
+  std::vector<int> result;
+  for (int i = 0; i < tab_count; ++i) {
+    if (is_valid_index[i]) {
+      result.push_back(i);
+    }
+  }
+  return result;
```
sangwoo108, line 81: Optional:

It wouldn't be the case for this, but `std::vector<bool>` is what I avoid in general as it's a pseudo vector, which may cause subtle bugs.

And using std::sort and unique can be more idiomatic.

```
std::vector<int> result;
result.reserve(indices.size());

for (int index : indices) {
  if (index >= 0 && index < tab_count) {
    result.push_back(index);
  }
}  // this also can be replaced with std::copy_if but not that worth.

std::sort(result.begin(), result.end());
result.erase(std::unique(result.begin(), result.end()), result.end());

return result;
```

browser/ai_chat/tools/tab_management_tool.cc
```diff
@@ -40,6 +53,55 @@ namespace ai_chat {
...
+           {"pink", tab_groups::TabGroupColorId::kPink},
+           {"purple", tab_groups::TabGroupColorId::kPurple},
+           {"cyan", tab_groups::TabGroupColorId::kCyan},
+           {"orange", tab_groups::TabGroupColorId::kOrange}});
+
+  const auto it = kColorNameToId.find(group_color);
+  if (it != kColorNameToId.end()) {
+    return it->second;
+  }
+  return std::nullopt;
```
sangwoo108, line 102: Can we do `return base::FindOrNull(kColorNameToId, group_color)` ?

browser/ai_chat/tools/tab_management_tool.cc
petemill: 😄 I did it this way because of feedback from @fallaciousreasoning here https://github.com/brave/brave-core/pull/33987#discussion_r3337433043 😄 

But I've changed it to something like you suggest here: https://github.com/brave/brave-core/pull/36976

browser/ai_chat/tools/tab_management_tool.cc
petemill: > `base::span` is a non-owning view (just a pointer + size), so the idiom is to pass it by value — a `const&` only adds an indirection with no benefit

What do you think? I implemented it as `base::span<const int>` because of @netzenbot feedback here: https://github.com/brave/brave-core/pull/33987#discussion_r3343149035

browser/ai_chat/tools/tab_management_tool.cc
petemill: Submitted in https://github.com/brave/brave-core/pull/36976

browser/ai_chat/tools/tab_management_tool.cc
sangwoo108: aha, thanks for the explanation. I think the feedback makes sense

browser/ai_chat/tools/tab_management_tool.cc
fallaciousreasoning: :laughing: happy to defer to @sangwoo108 here

PR #34030, opened 2026-02-18
Remove gs_lcrp URL parameter from omnibox searches to Google

### LINKED ISSUE #49954: Google searches from omnibox leak searchbox stats via tracking parameter in Chromium
### Description

I've freshly installed brave on mac, did some google search, and now am seeing tracking in URL bar.

### Steps to reproduce

1. Install brave on latest macos
2. Do a google search using built-in address bar

### Actual result

The URL would be something like this, with base64-encoded tracking data:

    https://www.google.com/search?q=hello&oq=hello&gs_lcrp=EgZjaHJvbWUq{...111-characters}&sourceid=chrome&ie=UTF-8

Notice `gs_lcrp`. I've removed full `gs_lcrp` 111-character content because i'm not sure which information is getting included.

### Expected result

URL should not include any tracking params. For example, on Safari it's clean & fresh:

    https://www.google.com/search?client=safari&rls=en&q=hello&ie=UTF-8&oe=UTF-8

### Reproduces how often

Easily reproduced

- 1.83.109 Chromium: 141.0.7390.55 (Official Build) (arm64) 
- 6c9b7bdded46e59d445cb0c067bff9f3bcd8fdd0
- macOS Version 26.0.1 (Build 25A362)

### Channel information

- [x] release (stable)
- [ ] beta
- [ ] nightly

### Reproducibility

- [ ] in the latest version of Chrome *(did not test there, Chrome is garbage)*

### Miscellaneous information

_No response_

r0hit0303:
Chromium browsers sends a `gs_lcrp` parameter on searches made to Google via its omnibox. This parameter captures statistical information about omnibox autocomplete impressions which is not relevant for Brave, and instead adds a passive fingerprinting surface.   Please see the [document](https://docs.google.com/document/d/1Ru1qGTSyp3R1bmQYO_aWFoIHEuQW2W1ygBTvKNlFFLs/edit?tab=t.0) for more details on the internal mechanism. 

Fix verified locally on Brave debug build: [Recording](https://drive.google.com/file/d/1tNpCWu5XdWfmcX1I9iZXH3SC0kbm2sod/view?usp=sharing)

Resolves brave/brave-browser#49954


r0hit0303:
Hello @fmarier, 
I'm tagging you since I saw that you have worked on such issues extensively in the past. This change is not yet completely ready for review. I wanted to just run the unit_tests on the cloud with this patch, while I'm building it locally. I've verified the fix on Brave debug build on my machine, and added a video for the same. 

I wanted to point out that we would unconditionally remove the `gs_lcrp` query parameter from *all* URLs I think. Given that this parameter seems unique to {Google[dot]com,Chromium} pair, would you rather prefer that we remove it conditionally on google[dot]com? Also, there were other query parameters that Brave was sending to Google[dot]com but not Safari (which I've captured in the doc), should we also include them as well? 

It's my first change, so please don't mind if I missed something very obvious. Thanks!

bbondy (commented):
Review via brave-core-bot: The test URL `"https://google.com/?gs_lcrp=12345?q=abc"` uses a second `?` instead of `&` to separate query parameters. It should be `"https://google.com/?gs_lcrp=12345&q=abc"`. With the current malformed URL, GURL treats the entire query as a single parameter `gs_lcrp=12345?q=abc`, so after stripping, the result would be `https://google.com/` rather than the expected `https://google.com/?q=abc`.

fmarier (changes requested):
This specific approach is not actually going to work because we don't filter same-site requests: https://github.com/brave/brave-browser/wiki/Query-String-Filter

fmarier:
I wonder whether there's something in Chromium  (e.g. a flag, a feature) that's causing this to be enabled in Brave/Chrome and not in Safari. Or whether it's user-agent sniffing on the server.

In general, we're not really keen to remove arbitrary parameters from a website's own internal navigations, unless we have a good reason for it.

fmarier:
Also, we would want to scope this removal to just `google.com` since that's not a parameter that's used in cross-site navigations: https://github.com/brave/brave-core/blob/24749dff7db503a0b2b1f5a7367e025af74e9688/components/query_filter/utils.cc#L145-L165

fmarier:
> I wonder whether there's something in Chromium (e.g. a flag, a feature) that's causing this to be enabled in Brave/Chrome

Looking at https://docs.google.com/document/d/1Ru1qGTSyp3R1bmQYO_aWFoIHEuQW2W1ygBTvKNlFFLs/edit?tab=t.0, it seems like you already found the code responsible for this in Chromium. Sorry, I missed that. Preventing the params from getting added in the first place seems like the best approach to me, despite the code change being larger.

r0hit0303:
Thanks fmarier@ for the additional context and pointers around the query string filter. This helps me to get the ball rolling on this change. I'll update this patch inline with the other approach that overrides on the chromium side.  (https://docs.google.com/document/d/1Ru1qGTSyp3R1bmQYO_aWFoIHEuQW2W1ygBTvKNlFFLs/edit?tab=t.0#bookmark=id.127ios7dm15p)

fmarier (approved):
Approach looks good to me.

r0hit0303:
Closing this PR in favour of the new [PR](https://github.com/brave/brave-core/pull/35120) directly in the brave-core repo to tackle the permission issues around github actions.

PR #34124, opened 2026-02-23
Upgrade from Chromium 145.0.7632.109 to Chromium 145.0.7632.117 (1.87.x)

emerick:
Resolves https://github.com/brave/brave-browser/issues/53078


emerick:
Abandoning for 145.0.7632.120 instead.

PR #34152, opened 2026-02-24
[News] Swap-and-pop instead of erasing articles when building feed

zenparsing:
Subtask of https://github.com/brave/brave-browser/issues/48969


fallaciousreasoning (approved):

fallaciousreasoning (commented):
just double checking whether this relies on sorting

I think this might break the publisher feed, which is meant to be ordered chronologically - do you mind checking?

zenparsing:
@fallaciousreasoning Good catch - this does result in an out-of-order publisher feed.

zenparsing:
Closing this one - it looks like the order of feed items is important. Also, I'm not seeing much of a performance benefit.

PR #34209, opened 2026-02-26
[Polkadot] Present Polkadot import keyring

cypt4:
Polkadot import keyring required to store polkadot imported accounts. This PR also moves encoding\decoding methods to polkadot_utils to reuse them.

Resolves


components/brave_wallet/browser/polkadot/polkadot_utils_unittest.cc
```diff
@@ -97,4 +111,358 @@ TEST(PolkadotUtils, Uint128MojomConversions) {
...
+                                                   &salt_bytes, &nonce_bytes);
+    std::optional<base::Value> json_value =
+        base::JSONReader::Read(*private_key_0, 5);
+    constexpr const char kExpectedJson[] =
+        R"({
+            "address":"5HGiBcFgEBMgT6GEuo9SA98sBnGgwHtPKDXiUukT6aqCrKEx",
+            "encoded":"AQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEAgAAAAQAAAAgAAAACAgICAgICAgICAgICAgICAgICAgICAgK8fiV6lGVSFewy6uxsU1C4kjcrjm4Ge+Vh5Lwh9XlJ3vAbQs+/0vSBKGe71ik2n/owjXKi9/fBCOIbIDIttGbopozlHDAyPEOrftWc4aiMZfTHRTh/Hb0kJ4C79LBJ84Y6sm59VMs51zUalbzBwa9c75OqlTCtqRouH8891IU51jczQkHY",
+            "encoding":{"content":["pkcs8","sr25519"],"type":["scrypt","xsalsa20-poly1305"],"version":"3"}})";
+    std::optional<base::Value> expected_json_value =
+        base::JSONReader::Read(kExpectedJson, 5);
```
supermassive: ~base::JSONReader::Read~
base::test::ParseJsonDict

components/brave_wallet/browser/polkadot/polkadot_utils.cc
```diff
@@ -5,11 +5,248 @@
...
 namespace brave_wallet {
 
+namespace {
+
+// Allowed scrypt parameters matching Polkadot.js wallet standards.
+struct AllowedScryptParams {
+  uint32_t n;
+  uint32_t p;
+  uint32_t r;
+};
```
supermassive, line 30: = 0

components/brave_wallet/browser/polkadot/polkadot_utils.h
```diff
@@ -46,6 +57,25 @@ struct PolkadotAddress {
...
+// Uses scrypt + xsalsa20-poly1305. Optional salt_for_testing and
+// nonce_for_testing allow deterministic output in tests; pass nullptr to use
+// random bytes.
+std::optional<std::string> EncodePrivateKeyForExport(
+    base::span<const uint8_t, kSr25519Pkcs8Size> pkcs8_key,
+    std::string_view address,
+    std::string_view password,
+    const std::array<uint8_t, kScryptSaltSize>* salt_for_testing = nullptr,
+    const std::array<uint8_t, kSecretboxNonceSize>* nonce_for_testing =
+        nullptr);
```
supermassive: These easily could be done without `for_testing` args. With the idead that caller always provides salt and nonce spans.

components/brave_wallet/browser/polkadot/polkadot_import_keyring.h
```diff
@@ -0,0 +1,60 @@
+/* Copyright (c) 2025 The Brave Authors. All rights reserved.
```
supermassive: 2026

components/brave_wallet/browser/polkadot/polkadot_import_keyring.h
```diff
@@ -0,0 +1,60 @@
...
+#include <string>
+
+#include "base/containers/span.h"
+#include "brave/components/brave_wallet/browser/internal/hd_key_sr25519.h"
+#include "brave/components/brave_wallet/common/brave_wallet.mojom.h"
+#include "brave/components/brave_wallet/common/brave_wallet_constants.h"
+
+namespace brave_wallet {
+
+class PolkadotImportKeyring {
```
netzenbot, line 25: nit: The new `PolkadotImportKeyring` class is missing a class-level documentation comment explaining its purpose and how it differs from `PolkadotKeyring` (manages individually imported sr25519 keys vs. HD-derived keys). Public methods like `AddAccount`, `RemoveAccount`, `SignMessage`, and `EncodePrivateKeyForExport` also lack doc comments — only `GetAccountAddress` is documented. [best practice](https://github.com/brave-experiments/brave-core-tools/tree/master/docs/best-practices/coding-standards.md#CS-020)

components/brave_wallet/browser/polkadot/polkadot_import_keyring.h
```diff
@@ -0,0 +1,60 @@
...
+  bool IsTestnet() const;
+  mojom::KeyringId keyring_id() const { return keyring_id_; }
+
+ private:
+  HDKeySr25519* GetAccountByIndex(uint32_t account);
+
+  mojom::KeyringId keyring_id_;
+  std::map<uint32_t, HDKeySr25519> accounts_;
+};
+
```
netzenbot, line 70: nit: Chromium guidelines recommend `base::flat_map` as the default associative container. For a per-keyring account map (expected to be small), `base::flat_map<uint32_t, HDKeySr25519>` would provide better cache locality and lower overhead. [best practice](https://github.com/brave-experiments/brave-core-tools/tree/master/docs/best-practices/coding-standards-apis.md#CSA-005)

components/brave_wallet/browser/polkadot/polkadot_import_keyring_unittest.cc
```diff
@@ -0,0 +1,176 @@
...
+    PolkadotImportKeyring import_keyring(mojom::KeyringId::kPolkadotImport);
+
+    EXPECT_TRUE(
+        import_keyring.AddAccount(0, hd_keyring.GetPkcs8KeyForTesting(0)));
+    EXPECT_EQ(import_keyring.GetAccountAddress(0).value(), kMainnetAddress0);
+    EXPECT_EQ(import_keyring.GetAddress(0, 0u).value(), kMainnetAddress0);
+
+    EXPECT_TRUE(
+        import_keyring.AddAccount(1, hd_keyring.GetPkcs8KeyForTesting(1)));
+    EXPECT_EQ(import_keyring.GetAccountAddress(1).value(), kMainnetAddress1);
```
netzenbot, line 56: `EXPECT_TRUE` should be `ASSERT_TRUE` here — if `AddAccount` fails, the subsequent `GetAccountAddress(0).value()` will crash rather than produce a clear test failure. This pattern repeats throughout the file (lines 56, 61, 69, 74, etc.). [best practice](https://github.com/brave-experiments/brave-core-tools/tree/master/docs/best-practices/testing-isolation.md#TI-034)

cmazakas:
Hmm, I'm wondering: do we need to introduce a new class here or can we just extend/modify the existing PolkadotKeyring?

If all we're doing is just storing something like this:
```cpp
base::flat_map<uint32_t, HDKeySr25519> secondary_keys_;
```
it feels like we should be able to reify all this functionality into a single umbrella, no?

The only reason I push back against a new keyring type is that it seems to duplicate a lot of functionality and it increases our testing burden a lot more.

components/brave_wallet/browser/polkadot/polkadot_keyring.cc
```diff
@@ -168,231 +131,27 @@ void PolkadotKeyring::SetSignatureRngForTesting() {
...
-  // Substrate/Polkadot standard parameters: n=32768, r=8, p=1.
-  // https://github.com/polkadot-js/common/blob/fe0886be239526e6c559e98d1099815d4b4f4a7f/packages/util-crypto/src/scrypt/defaults.ts#L10
-  crypto::kdf::ScryptParams scrypt_params = {
-      .cost = 32768,                         // n
-      .block_size = 8,                       // r
-      .parallelization = 1,                  // p
-      .max_memory_bytes = 64 * 1024 * 1024,  // 64 MB
-  };
+  std::array<uint8_t, kScryptSaltSize> salt;
+  std::array<uint8_t, kSecretboxNonceSize> nonce;
```
supermassive: nit: = {}

supermassive (approved):

components/brave_wallet/browser/polkadot/polkadot_keyring.cc
```diff
@@ -168,231 +131,27 @@ void PolkadotKeyring::SetSignatureRngForTesting() {
...
-      encrypted_data, nonce, base::span(*scrypt_key).first<kScryptKeyBytes>());
-  if (!decrypt_result || decrypt_result->size() != kSr25519Pkcs8Size) {
-    return std::nullopt;
-  }
-
-  std::array<uint8_t, kSr25519Pkcs8Size> secret_key = {};
-  base::span(secret_key).copy_from_nonoverlapping(base::span(*decrypt_result));
-
-  return secret_key;
+  return ::brave_wallet::EncodePrivateKeyForExport(pkcs8_key, address, password,
```
netzenbot: The old code explicitly zeroed `pkcs8_key` via `crypto::internal::SecureZeroBuffer(pkcs8_key)` after copying to `SecureVector`. This refactoring removes that zeroing — the PKCS8 private key now sits unzeroed on the caller's stack. Consider storing the result, zeroing `pkcs8_key`, then returning. Same issue in `PolkadotImportKeyring::EncodePrivateKeyForExport`.

cypt4:
> Hmm, I'm wondering: do we need to introduce a new class here or can we just extend/modify the existing PolkadotKeyring?
> 
> If all we're doing is just storing something like this:
> 
> ```c++
> base::flat_map<uint32_t, HDKeySr25519> secondary_keys_;
> ```
> 
> it feels like we should be able to reify all this functionality into a single umbrella, no?
> 
> The only reason I push back against a new keyring type is that it seems to duplicate a lot of functionality and it increases our testing burden a lot more.

I am following bitcoin approach here which is latest coin that supports import.
Actually both polkadot and polkadot_import reuse major part of code and they only bring code related to managing accounts.

components/brave_wallet/browser/polkadot/polkadot_utils_unittest.cc
```diff
@@ -97,4 +111,349 @@ TEST(PolkadotUtils, Uint128MojomConversions) {
...
+  {
+    auto encoded_json = keyring.EncodePrivateKeyForExport(1, kPassword);
+    ASSERT_TRUE(encoded_json.has_value());
+
+    auto original_pkcs8_key = keyring.GetPkcs8KeyForTesting(1);
+    auto decoded_pkcs8_key =
+        DecodePrivateKeyFromExport(*encoded_json, kPassword);
+    ASSERT_TRUE(decoded_pkcs8_key.has_value());
+    EXPECT_EQ(*decoded_pkcs8_key, original_pkcs8_key);
+  }
```
cmazakas, line 262: I notice this test doesn't generate a `decoded_keypair` like the one immediately above it does. Shouldn't this one also ensure that we can create the decoded `HDKeySr25519`?

cmazakas (approved):

components/brave_wallet/browser/polkadot/polkadot_utils.cc
```diff
@@ -5,11 +5,233 @@
...
+  auto decrypt_result = XSalsaPolyDecrypt(
+      encrypted_data, nonce, base::span(*scrypt_key).first<kScryptKeyBytes>());
+  if (!decrypt_result || decrypt_result->size() != kSr25519Pkcs8Size) {
+    return std::nullopt;
+  }
+
+  std::array<uint8_t, kSr25519Pkcs8Size> secret_key = {};
+  base::span(secret_key).copy_from_nonoverlapping(base::span(*decrypt_result));
+
+  return secret_key;
```
netzenbot, line 160: `decrypt_result` holds the plaintext PKCS#8 private key but is never zeroed — the key material lingers in heap memory. The encode path zeroes both `pkcs8_key` and `derived_key` via `SecureZeroBuffer`; this decode path should do the same for `decrypt_result` and `scrypt_key`.

cypt4: decrypt_result and scrypt_key are optional of SecureVector so they are erased on destruction.

components/brave_wallet/common/common_utils.cc
```diff
@@ -489,9 +499,12 @@ std::vector<mojom::KeyringId> GetSupportedKeyringsForNetwork(
...
     case mojom::CoinType::DOT:
-      return {(chain_id == mojom::kPolkadotMainnet
-                   ? mojom::KeyringId::kPolkadotMainnet
-                   : mojom::KeyringId::kPolkadotTestnet)};
+      if (chain_id == mojom::kPolkadotMainnet) {
+        return {mojom::KeyringId::kPolkadotMainnet,
+                mojom::KeyringId::kPolkadotImport};
+      }
+      return {mojom::KeyringId::kPolkadotTestnet,
+              mojom::KeyringId::kPolkadotImportTestnet};
```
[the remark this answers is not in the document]
cypt4: No break is required since return is used

components/brave_wallet/common/brave_wallet_types.h
```diff
@@ -23,6 +24,7 @@ using int256_t = _BitInt(256);
 
 using uint128_t = unsigned _BitInt(128);
 using int128_t = _BitInt(128);
+using SecureVector = std::vector<uint8_t, crypto::SecureAllocator<uint8_t>>;
```
kdenhartog, line 27: Thanks for using secureAllocator here

components/brave_wallet/browser/polkadot/polkadot_utils_unittest.cc
```diff
@@ -97,4 +111,354 @@ TEST(PolkadotUtils, Uint128MojomConversions) {
...
+  auto encoded_json = keyring.EncodePrivateKeyForExport(0, kPassword);
+  ASSERT_TRUE(encoded_json.has_value());
+
+  auto original_pkcs8_key = keyring.GetPkcs8KeyForTesting(0);
+  auto decoded_pkcs8_key = DecodePrivateKeyFromExport(*encoded_json, kPassword);
+  ASSERT_TRUE(decoded_pkcs8_key.has_value());
+  EXPECT_EQ(*decoded_pkcs8_key, original_pkcs8_key);
+}
+
+TEST(PolkadotUtils, DecodePrivateKeyFromExport_MissingParts) {
```
kdenhartog, line 321: nice job adding all these unit tests for potential failure cases.

components/brave_wallet/browser/polkadot/polkadot_utils.cc
```diff
@@ -5,11 +5,233 @@
...
+      return true;
+    }
+  }
+  return false;
+}
+
+}  // namespace
+
+std::optional<std::array<uint8_t, kSr25519Pkcs8Size>>
+DecodePrivateKeyFromExport(std::string_view json_export,
```
kdenhartog, line 55: nit: the code complexity of this function is rather large. It would be useful to split it up so it's easier to read, but AFAICT it's implemented properly currently. Seems better to just open a separate code health issue for that though.

kdenhartog (commented):
No concerns from me for security perspective

components/brave_wallet/browser/BUILD.gn
```diff
@@ -454,18 +452,24 @@ source_set("utils") {
     "brave_wallet_utils.h",
     "network_manager.cc",
     "network_manager.h",
+    "polkadot/polkadot_utils.cc",
```
bridiver, line 455: `polkadot` directory needs its own BUILD.gn file.

cmazakas: Filed here https://github.com/brave/brave-browser/issues/56562

PR #34296, opened 2026-02-27
Add default search engine to daily usage ping

aseren:
Report default search in usage pings via the "defaultSearch" query parameter and limited to Brave, Google, and Other only.

Resolves

### DIFF
--- a/browser/brave_stats/brave_stats_updater.cc
+++ b/browser/brave_stats/brave_stats_updater.cc
@@ -40,10 +40,13 @@
 #include "chrome/browser/net/system_network_context_manager.h"
 #include "chrome/browser/profiles/profile.h"
 #include "chrome/browser/profiles/profile_manager.h"
+#include "chrome/browser/search_engines/template_url_service_factory.h"
 #include "chrome/common/channel_info.h"
 #include "components/prefs/pref_change_registrar.h"
 #include "components/prefs/pref_registry_simple.h"
 #include "components/prefs/pref_service.h"
+#include "components/search_engines/search_engine_type.h"
+#include "components/search_engines/template_url_service.h"
 #include "content/public/browser/browser_thread.h"
 #include "net/base/load_flags.h"
 #include "net/base/url_util.h"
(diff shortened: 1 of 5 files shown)


netzenbot (commented):
`TemplateURLServiceFactory::GetForProfile` (in `brave_stats_updater.cc`) can return null for non-regular profiles (system, guest). Since the profile comes from `GetInitialOrLastUsedProfile` which may return unexpected profile types, consider handling this gracefully (return `{}`) like the null profile check above, rather than crashing with CHECK. [best practice](https://github.com/brave-experiments/brave-core-tools/tree/master/docs/best-practices/coding-standards.md#CS-027)

netzenbot (dismissed):

aseren:
```
TemplateURLServiceFactory::GetForProfile (in brave_stats_updater.cc) can return null for non-regular profiles (system, guest). Since the profile comes from GetInitialOrLastUsedProfile which may return unexpected profile types, consider handling this gracefully (return {}) like the null profile check above, rather than crashing with CHECK. [best practice](https://github.com/brave-experiments/brave-core-tools/tree/master/docs/best-practices/coding-standards.md#CS-027)
```

The `GetInitialOrLastUsedProfile ` function can only return regular profiles. To make it clear I renamed it to `GetInitialOrLastUsedRegularProfile` and added checks.

browser/brave_stats/brave_stats_updater.cc
```diff
@@ -123,8 +117,47 @@ serp_metrics::SerpMetrics* GetSerpMetrics(ProfileManager* profile_manager) {
     // TODO(https://github.com/brave/brave-browser/issues/52492): Migrate SERP
     // metrics to profile attributes and remove temporary invariant checks.
     DUMP_WILL_BE_CHECK(profile);
+  }
+
+  return profile;
+}
+
+std::string GetDefaultSearchEngine(ProfileManager* profile_manager) {
```
netzenbot, line 125: The new unit tests only check that a string passed into `GetUpdateURL()` comes back out as `defaultSearch` — the actual new logic in `GetDefaultSearchEngine()` (the `SEARCH_ENGINE_*` -> `"Brave"`/`"Google"`/`"Other"` mapping and the empty-string paths) is untested. `BraveStatsUpdaterBrowserTest` already captures the ping URL via `GetUpdateURL()` and asserts on query params, so asserting `defaultSearch` there would cover the mapping end to end. ([✅ PRs Should Include Reasonable Test Coverage](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-039))

PR #34451, opened 2026-03-04
Sites and shields redux

### LINKED ISSUE #55724: [Shields] Update Shields panel design
## Description
- This captures work originally started with https://github.com/brave/brave-browser/issues/49408
- Figma captured for employees with https://www.figma.com/design/yv8ZAUPl43TI3PlzSElm2m/%F0%9F%9B%A1%EF%B8%8F-Shields-redesign?t=VMgYmdD1NDQ35R5j-0

Goal here is to redesign the shields experience on Android.  [Desktop has already implemented this redesign](https://github.com/brave/brave-browser/issues/53111).

### Regular shields panel
Image

### Advanced shields panel
Image

Olipro:
This commit series represents a scaled-back version of the original Sites & Shields UI design. Most of the retained functionality centers around the Shields UI which now appears when tapping on the Brave Lion icon in the URL bar.

Commits are arranged into logical groupings so it is likely easier to review by examining each commit rather than the entire changeset all at once.

Closes https://github.com/brave/brave-browser/issues/55724


android/java/org/chromium/chrome/browser/shields/BraveUnifiedPanelHandler.java
```diff
@@ -0,0 +1,1666 @@
...
+                    learnMoreEnd,
+                    Spanned.SPAN_EXCLUSIVE_EXCLUSIVE);
+            spannable.setSpan(
+                    new ClickableSpan() {
+                        @Override
+                        public void onClick(View widget) {
+                            if (mCurrentTab != null) {
+                                mCurrentTab.loadUrl(
+                                        new LoadUrlParams(
+                                                "https://community.brave.com/t/broken-site-report/"));
```
Olipro: Questions for reviewers:

1. Is this the correct URL? I think so. It fires when the user clicks "Learn More" when shields is disabled from our UI.

2. Where should this string ideally live? Presumably there's a better place for it than just hardcoding it inline

android/java/org/chromium/chrome/browser/shields/BraveUnifiedPanelHandler.java
```diff
@@ -0,0 +1,1666 @@
...
+                                        if (bitmap != null) {
+                                            onFaviconResult(origin, bitmap);
+                                        } else {
+                                            tryRegistrableDomainFavicon(origin);
+                                        }
+                                    });
+                });
+    }
+
+    private void tryRegistrableDomainFavicon(GURL origin) {
```
Olipro, line 1797: This method (and the related ones following it, are a hack put together while I was testing because stuff like favicons for Facebook trackers wouldn't display even though I'd visited facebook.com because they (and others) use an alternative TLD.

Ideally I can toss all of this away but it would mean that someone has to take care of shipping Brave with whatever favicon entries are needed in the SQLite DB for whatever TLDs the major players use.

In essence, either get rid of this, or don't, but come up with a more systematic approach.

android/java/org/chromium/chrome/browser/shields/BraveUnifiedPanelHandler.java
```diff
@@ -0,0 +1,1721 @@
...
+    private void addBlockOverlay(FrameLayout container, int iconSize) {
+        if (mContext == null) return;
+
+        ImageView overlay = new ImageView(mContext);
+        overlay.setLayoutParams(new FrameLayout.LayoutParams(iconSize, iconSize));
+        overlay.setImageResource(R.drawable.ic_blocked_item_overlay);
+        overlay.setScaleType(ImageView.ScaleType.FIT_CENTER);
+        container.addView(overlay);
+    }
+}
```
netzenbot, line 2077: The favicon callback at this line casts `mContext` to `Activity` and posts UI work via `runOnUiThread`, but there is no check for whether the Activity is finishing or destroyed. Between when the favicon request is made and when the callback fires, the Activity could be destroyed, causing a crash or operation on stale state. Add a check like `if (((Activity) mContext).isFinishing() || ((Activity) mContext).isDestroyed()) return;` before the `runOnUiThread` call. The same pattern repeats in `tryRegistrableDomainFavicon` and `tryNextTldCandidate`. ([✅ Check `isActivityFinishingOrDestroyed()` Before UI Operations in Async Callbacks](https://github.com/brave/brave-core/tree/master/docs/best-practices/android.md#AND-001))

android/java/brave-res/values/unified_panel_strings.xml
```diff
@@ -0,0 +1,46 @@
...
+    <!-- Report Broken Site Panel -->
+    <string name="report_broken_site_panel_title">Report a broken site</string>
+    <string name="report_broken_site_panel_desc1">Let Brave\'s developers know that this site doesn\'t work properly with this site.</string>
+    <string name="report_broken_site_panel_desc2">Thank you for helping make Brave better for all. This report will only contain information necessary for us to fix this site. </string>
+    <string name="report_broken_site_panel_details_hint">What is not working on this site?</string>
+    <string name="report_broken_site_panel_contact_label">Contact me at (optional)</string>
+    <string name="report_broken_site_panel_contact_hint">Email, X username, etc.</string>
+    <string name="report_broken_site_panel_contact_info">If you provide contact info it will be stored for future reports</string>
+    <string name="report_broken_site_panel_submit">Submit</string>
+    <string name="report_broken_site_panel_submitted">Report submitted. Thank you!</string>
```
netzenbot: The string `report_broken_site_panel_desc1` ends with "doesn't work properly with this site" — "with this site" looks like a copy-paste error from the site domain reference earlier in the string. Based on the old removed strings (and the feature context), this should probably read "with Shields" or "with Shields up". ([✅ String Text Must Be Complete and Grammatically Correct](https://github.com/brave/brave-core/tree/master/docs/best-practices/localization.md#L10N-016))

android/java/brave-res/drawable/ic_unified_shields_disabled.xml
```diff
@@ -0,0 +1,15 @@
+<!-- Copyright (c) 2026 The Brave Authors. All rights reserved.
+     This Source Code Form is subject to the terms of the Mozilla Public
+     License, v. 2.0. If a copy of the MPL was not distributed with this file,
+     You can obtain one at https://mozilla.org/MPL/2.0/. -->
+
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
netzenbot: nit: This icon should be added to `android/nala/icons.gni` instead of creating a custom drawable file directly. See the [NA-009 rule](https://github.com/brave/brave-core/tree/master/docs/best-practices/nala.md#NA-009). Tag @nala-token-reviewers.

android/java/brave-res/drawable/ic_unified_shields_enabled.xml
```diff
@@ -0,0 +1,15 @@
+<!-- Copyright (c) 2026 The Brave Authors. All rights reserved.
+     This Source Code Form is subject to the terms of the Mozilla Public
+     License, v. 2.0. If a copy of the MPL was not distributed with this file,
+     You can obtain one at https://mozilla.org/MPL/2.0/. -->
+
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
netzenbot: nit: This icon should be added to `android/nala/icons.gni` instead of creating a custom drawable file directly. See the [NA-009 rule](https://github.com/brave/brave-core/tree/master/docs/best-practices/nala.md#NA-009). Tag @nala-token-reviewers.

android/java/brave-res/drawable/ic_blocked_item_overlay.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright (c) 2026 The Brave Authors. All rights reserved.
+     This Source Code Form is subject to the terms of the Mozilla Public
+     License, v. 2.0. If a copy of the MPL was not distributed with this file,
+     You can obtain one at https://mozilla.org/MPL/2.0/. -->
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
netzenbot: nit: This icon should be added to `android/nala/icons.gni` instead of creating a custom drawable file directly. See the [NA-009 rule](https://github.com/brave/brave-core/tree/master/docs/best-practices/nala.md#NA-009). Tag @nala-token-reviewers.

android/java/brave-res/drawable/filled_button_background.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright (c) 2026 The Brave Authors. All rights reserved.
+     This Source Code Form is subject to the terms of the Mozilla Public
+     License, v. 2.0. If a copy of the MPL was not distributed with this file,
+     You can obtain one at https://mozilla.org/MPL/2.0/. -->
+
+<shape xmlns:android="http://schemas.android.com/apk/res/android"
+    android:shape="rectangle">
+    <corners android:radius="8dp" />
+    <solid android:color="#434FCF" />
```
AlexeyBarabash: Does  this button look the same at Light and Dark mode? Maybe it would better to use a named color?

android/java/brave-res/values/unified_panel_strings.xml
```diff
@@ -0,0 +1,45 @@
...
+
+    <!-- Shred Site Data Panel -->
+    <string name="shred_site_data_title">Shred Site Data</string>
+    <string name="shred_auto_shred">Auto Shred</string>
+    <string name="shred_option_site_tabs_closed">When site tabs are closed</string>
```
netzenbot: nit: Menu items should use title case per the localization style guide. `shred_option_site_tabs_closed` and `shred_option_app_exit` should be "When Site Tabs Are Closed" and "When App Exits" to be consistent with the other title-case strings in this panel ("Auto Shred", "Shred Site Data Now"). ([L10N-009](https://github.com/brave/brave-core/tree/master/docs/best-practices/localization.md#L10N-009))

android/java/brave-res/values/unified_panel_strings.xml
```diff
@@ -0,0 +1,45 @@
...
+    <string name="brave_shields_global_settings_title">Shields global settings</string>
+
+    <!-- Report Broken Site -->
+    <string name="report_broken_site_prompt">Tell us if the site wasn\'t working properly with Shields up.</string>
+    <string name="report_broken_site_button">Report</string>
+
+    <!-- Report Broken Site Panel -->
+    <string name="report_broken_site_panel_title">Report a broken site</string>
+    <string name="report_broken_site_panel_desc1">Let Brave\'s developers know that this site doesn\'t work properly with this site.</string>
+    <string name="report_broken_site_panel_desc2">Thank you for helping make Brave better for all. This report will only contain information necessary for us to fix this site. </string>
```
netzenbot: nit: `report_broken_site_panel_desc2` has a trailing space after the final period — `"...necessary for us to fix this site. "` should be `"...necessary for us to fix this site."`. Android string resources include literal whitespace. ([L10N-016](https://github.com/brave/brave-core/tree/master/docs/best-practices/localization.md#L10N-016))

android/java/brave-res/drawable/ic_brave_unified_arrow_down.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
```
netzenbot: nit: This icon should be added to `android/nala/icons.gni` instead of creating a custom drawable file directly. See the [NA-009 rule](https://github.com/brave/brave-core/tree/master/docs/best-practices/nala.md#NA-009). Tag @nala-token-reviewers.

android/java/brave-res/drawable/ic_shred_dropdown_arrows.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright (c) 2026 The Brave Authors. All rights reserved.
+     This Source Code Form is subject to the terms of the Mozilla Public
+     License, v. 2.0. If a copy of the MPL was not distributed with this file,
+     You can obtain one at https://mozilla.org/MPL/2.0/. -->
+
+<!-- TODO: Destroy this when we have the actual form control -->
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
netzenbot: nit: Even as a temporary implementation, this icon should be added to `android/nala/icons.gni` rather than as a custom drawable file. The TODO comment already flags this for removal — if it's truly temporary, track it with a bug. See the [NA-009 rule](https://github.com/brave/brave-core/tree/master/docs/best-practices/nala.md#NA-009). Tag @nala-token-reviewers.

android/java/brave-res/drawable/filled_button_background.xml
Olipro: Yes, that's exactly what it should be, fixed.

android/java/brave-res/values/unified_panel_strings.xml
Olipro: @aguscruiz Could you chime in on this? This is actually currently slightly different from the more abbreviated copy that you have in Figma (reduced to just `App close`)

Let me know what you'd like to go with and if it should be capitalised according to the bot or how you have it in Figma

android/java/brave-res/values/unified_panel_strings.xml
Olipro: Will be fixed by next push.

android/java/brave-res/values/unified_panel_strings.xml
```diff
@@ -0,0 +1,45 @@
...
+    <string name="report_broken_site_panel_submit">Submit</string>
+    <string name="report_broken_site_panel_submitted">Report submitted. Thank you!</string>
+
+    <!-- Shred Site Data Panel -->
+    <string name="shred_site_data_title">Shred Site Data</string>
+    <string name="shred_auto_shred">Auto Shred</string>
+    <string name="shred_option_site_tabs_closed">When site tabs are closed</string>
+    <string name="shred_option_app_exit">When app exits</string>
+    <string name="shred_site_data_now">Shred Site Data Now</string>
+</resources>
```
netzenbot: nit: `report_broken_site_panel_contact_hint` uses "etc." — per the style guide, use "and more" instead. Change to "Email, X username, and more". ([L10N-010](https://github.com/brave/brave-core/tree/master/docs/best-practices/localization.md#L10N-010))

netzenbot: nit: `report_broken_site_panel_contact_info` is missing a comma — "If you provide contact info it will be stored..." should be "If you provide contact info, it will be stored for future reports". ([L10N-016](https://github.com/brave/brave-core/tree/master/docs/best-practices/localization.md#L10N-016))

android/java/res/values/brave_colors.xml
```diff
@@ -390,4 +381,9 @@
     <color name="material_tertiary_80">@color/primitive_tertiary_80</color>
```
netzenbot: nit: `shields_surface`, `shields_container_background`, `shields_text_warning`, and `divider_line_bg_color` add new hardcoded color resources rather than referencing existing Nala/Leo tokens. `shields_container_background` in particular duplicates `schemes_surface_container_lowest` (light) and `schemes_surface_container` (dark) exactly — consider using the existing tokens directly. ([NA-002](https://github.com/brave/brave-core/tree/master/docs/best-practices/nala.md#NA-002)) Tag @nala-token-reviewers.

android/java/brave-res/values/unified_panel_strings.xml
```diff
@@ -0,0 +1,45 @@
...
+    <string name="report_broken_site_panel_desc2">Thank you for helping make Brave better for all. This report will only contain information necessary for us to fix this site.</string>
+    <string name="report_broken_site_panel_details_hint">What is not working on this site?</string>
+    <string name="report_broken_site_panel_contact_label">Contact me at (optional)</string>
+    <string name="report_broken_site_panel_contact_hint">Email, X username, etc.</string>
+    <string name="report_broken_site_panel_contact_info">If you provide contact info it will be stored for future reports</string>
+    <string name="report_broken_site_panel_submit">Submit</string>
+    <string name="report_broken_site_panel_submitted">Report submitted. Thank you!</string>
+
+    <!-- Shred Site Data Panel -->
+    <string name="shred_site_data_title">Shred Site Data</string>
```
aguscruiz: This @Olipro though should say "Shred site data" in sentence case

android/java/brave-res/values/unified_panel_strings.xml
aguscruiz: @Olipro I was gonna say let's not mess with this but no, Camel case is not the right solution for Android. Android uses sentence case, we should follow that. I'll change the string to say what they say here, but it should be sentence case for sure.

android/java/brave-res/values/unified_panel_strings.xml
```diff
@@ -0,0 +1,45 @@
...
+    <string name="shred_auto_shred">Auto Shred</string>
+    <string name="shred_option_site_tabs_closed">When site tabs are closed</string>
+    <string name="shred_option_app_exit">When app exits</string>
+    <string name="shred_site_data_now">Shred Site Data Now</string>
```
aguscruiz: @Olipro this should say "Shred site data now"

Olipro:
@aguscruiz What do we want to do regarding:

> report_broken_site_screenshot_view.xml deleted — The screenshot preview dialog UI was removed entirely. The new code captures screenshots but never shows a preview to the user before submission.

Maybe you already did a UI for it that I missed?

Olipro:
> Report submission strips query but not fragment — Uri.parse(mUrl.getSpec()).buildUpon().clearQuery().fragment(null).build().toString() — if fragment(null) doesn't properly strip fragments on all Android versions, potentially sensitive URL fragments could be sent in the webcompat report

@brave/android This seems a bit of a stupid AI suggestion: _"If the user's system has a buggy implementation, we end up with incorrect output"_? You don't say.

I'm guessing there's an Chromium internal equivalent? GURL? Figure I'd get someone's $0.02 before I go and find yet another new way of doing it that the bot still doesn't like.

aguscruiz:
> @aguscruiz What do we want to do regarding:
> 
> > report_broken_site_screenshot_view.xml deleted — The screenshot preview dialog UI was removed entirely. The new code captures screenshots but never shows a preview to the user before submission.
> 
> Maybe you already did a UI for it that I missed?

I think you're refering to this? 
https://www.figma.com/design/yv8ZAUPl43TI3PlzSElm2m/%F0%9F%9B%A1%EF%B8%8F-Shields-redesign?node-id=6399-155212&t=wEMtTCK6xCHULrGG-1

As far as I know the webcompat reporter was a separate thing from Shields (it is on desktop at least). But I do have the design there. It's pretty much the same behavior and everything our current one has, just using the new styles.

**Oh no, sorry, I see now. I didn't realize there was a screenshot preview (we don't have that on Desktop). I'll build that right away @Olipro, but it's gonna be the same as what we already have in place. Won't change anything**

aguscruiz:
@Olipro Here's that missing part 
https://www.figma.com/design/yv8ZAUPl43TI3PlzSElm2m/%F0%9F%9B%A1%EF%B8%8F-Shields-redesign?node-id=6668-37902&t=wEMtTCK6xCHULrGG-1

I also added for good measure how the dropdown for the "What's the main issue you're seeing?" dropdown should be styled, which is pretty much standard material design, so I hope it's not a big deal and it's maybe already implemented.

android/java/org/chromium/chrome/browser/shields/BraveUnifiedPanelHandler.java
```diff
@@ -0,0 +1,1760 @@
...
+            blockersInfo.mFingerprintsBlocked++;
+        }
+
+        if (isShowing()
+                && mCurrentTab != null
+                && mCurrentTab.getId() == tabId
+                && mBlockCountNumber != null
+                && mContext != null) {
+            int totalBlocked = getTotalBlockedCount(tabId);
+            ((Activity) mContext)
```
netzenbot: The `addStat` method posts a UI update via `runOnUiThread` without first checking whether the Activity is finishing or destroyed. Between when `addStat` is called and when the lambda executes, the Activity could be torn down. Add `if (((Activity) mContext).isFinishing() || ((Activity) mContext).isDestroyed()) return;` before the `runOnUiThread` call, consistent with the guard already used in `loadBlockedItemFavicon`, `tryRegistrableDomainFavicon`, and `tryNextTldCandidate`. ([AND-001](https://github.com/brave/brave-core/tree/master/docs/best-practices/android.md#AND-001))

android/java/org/chromium/chrome/browser/shields/BraveUnifiedPanelHandler.java
```diff
@@ -0,0 +1,1760 @@
...
+            if (tabFavicon != null) {
+                siteFavicon.setImageBitmap(tabFavicon);
+            }
+        }
+
+        TextView descLearnMore = panel.findViewById(R.id.report_description_learn_more);
+        if (descLearnMore != null && mContext != null) {
+            String descText = mContext.getString(R.string.report_broken_site_panel_desc2);
+            String learnMoreText = mContext.getString(R.string.learn_more);
+            String fullText = descText + learnMoreText + ".";
```
netzenbot: The previous review flagged a trailing space at the end of `report_broken_site_panel_desc2`, and that space appears to have been removed. However, the Java code here concatenates `descText + learnMoreText + "."` without a space separator, which now produces "...to fix this site.Learn more." — missing the space between the sentence-ending period and "Learn more". Fix by inserting a space: `String fullText = descText + " " + learnMoreText + ".";`. ([✅ String Text Must Be Complete and Grammatically Correct](https://github.com/brave/brave-core/tree/master/docs/best-practices/localization.md#L10N-016))

android/java/org/chromium/chrome/browser/shields/BraveUnifiedPanelHandler.java
```diff
@@ -0,0 +1,1760 @@
...
+    private void addBlockOverlay(FrameLayout container, int iconSize) {
+        if (mContext == null) return;
+
+        ImageView overlay = new ImageView(mContext);
+        overlay.setLayoutParams(new FrameLayout.LayoutParams(iconSize, iconSize));
+        overlay.setImageResource(R.drawable.ic_disable_outline);
+        overlay.setScaleType(ImageView.ScaleType.FIT_CENTER);
+        container.addView(overlay);
+    }
+}
```
netzenbot, line 2077: nit: The constructor is missing a Javadoc. The old `BraveShieldsHandler(Context context)` had `@param context Context that is using the BraveShieldsMenu.` — worth preserving that contract description in the replacement class. Same applies to the public `show()` method. ([✅ Method Documentation Should Describe the Contract](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-004))

android/java/org/chromium/chrome/browser/shields/BraveUnifiedPanelHandler.java
Olipro: I thought I'd fixed this. Is this a new issue or one that was previously reported?

SergeyZhukovsky:
@Olipro could you please include an issue?
Resolves: <issue link>
and also a few pictures on how does the new shield panel looks like.
And a figma link please.

android/java/brave-res/values/unified_panel_strings.xml
```diff
@@ -0,0 +1,45 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright (c) 2026 The Brave Authors. All rights reserved.
+     This Source Code Form is subject to the terms of the Mozilla Public
+     License, v. 2.0. If a copy of the MPL was not distributed with this file,
+     You can obtain one at https://mozilla.org/MPL/2.0/. -->
+
+<resources xmlns:tools="http://schemas.android.com/tools">
+    <!-- Shields Panel Strings -->
+
+    <string name="shields_warning_line2">Note: this may reduce Brave\'s privacy protections.</string>
```
SergeyZhukovsky: not sure why the strings are in that new file? I don't think translations are going to auto apply to them. Android strings that are in base module should land in `android_brave_strings.grd`

Olipro:
> @Olipro could you please include an issue? Resolves: and also a few pictures on how does the new shield panel looks like. And a figma link please.

Ah, yes, thought of everything but stating what it resolves. Will amend the commit message(s) ~~and attach some screenshots when I have the build passing~~ - it seems something's up with the corp VPN right now so probably tomorrow if it isn't sorted soon.

Edit: Build is in fact passing although I can't access it. Here's the Figma: https://www.figma.com/design/yv8ZAUPl43TI3PlzSElm2m/%F0%9F%9B%A1%EF%B8%8F-Shields-redesign?node-id=6668-37902&m=dev - I actually need to make a couple more tweaks from a few changes today so I'll get the screenshots in when it fully reflects the design.

SergeyZhukovsky:
> > Report submission strips query but not fragment — Uri.parse(mUrl.getSpec()).buildUpon().clearQuery().fragment(null).build().toString() — if fragment(null) doesn't properly strip fragments on all Android versions, potentially sensitive URL fragments could be sent in the webcompat report
> 
> @brave/android This seems a bit of a stupid AI suggestion: _"If the user's system has a buggy implementation, we end up with incorrect output"_? You don't say.
> 
> I'm guessing there's an Chromium internal equivalent? GURL? Figure I'd get someone's $0.02 before I go and find yet another new way of doing it that the bot still doesn't like.

I also think it's fine as is the way you implemented. Could be done via GURL as well something like
```
  reportInfo.reportUrl = mUrl.getScheme() + "://" + mUrl.getHost()
          + (!mUrl.getPort().isEmpty() ? ":" + mUrl.getPort() : "")                                                                                                                                                                                                                                                                                    
          + mUrl.getPath();

```

android/java/org/chromium/chrome/browser/shields/BraveUnifiedPanelHandler.java
```diff
@@ -0,0 +1,1760 @@
...
+        }
+
+        if (isShowing()
+                && mCurrentTab != null
+                && mCurrentTab.getId() == tabId
+                && mBlockCountNumber != null
+                && mContext != null) {
+            int totalBlocked = getTotalBlockedCount(tabId);
+            ((Activity) mContext)
+                    .runOnUiThread(
```
SergeyZhukovsky: please use `PostTask.postTask(TaskTraits.UI_DEFAULT` to run something on UI thread. `PostTask` goes through Chromium's `TaskRunner` infrastructure. In short — `runOnUiThread` is an Android API that bypasses Chromium's entire task infrastructure, which means you lose lifecycle safety, priority scheduling, shutdown handling, and testability.

android/java/res/values-night/brave_colors.xml
```diff
@@ -229,4 +224,10 @@
     <color name="empty_state_tab_border_color">#828CFF</color>
     <color name="empty_state_tab_group_border_color">#828CFF</color>
 
+    <!-- Shields Panel Colors -->
+    <color name="shields_surface">#2E3038</color>
+    <color name="shields_container_background">#1F1F23</color>
+    <color name="shields_text_warning">#FFB800</color>
+    <color name="divider_line_bg_color">#3B3E4F</color>
```
SergeyZhukovsky: do we have newly added colors in figma? We should use everything from there only.

android/java/org/chromium/chrome/browser/shields/BraveUnifiedPanelHandler.java
```diff
@@ -0,0 +1,1760 @@
...
+        return mPopupWindow != null && mPopupWindow.isShowing();
+    }
+
+    public void hide() {
+        if (mPopupWindow != null && mPopupWindow.isShowing()) {
+            mPopupWindow.dismiss();
+        }
+    }
+
+    public void destroy() {
```
SergeyZhukovsky, line 1103: I don't see anyone calls `mUnifiedPanelHandler.destroy()`. Should it be called on hide()? Otherwise `mFaviconHelper` leaks and the mojo isn't closed. Or just nuke destroy and move everything inside `hide()`
```
      public void hide() {                                                                                                                                                                                                                                                                                                                             
          if (mPopupWindow != null && mPopupWindow.isShowing()) {
              mPopupWindow.dismiss();                                                                                                                                                                                                                                                                                                                  
          }       
          if (mFaviconHelper != null) {
              mFaviconHelper.destroy();
              mFaviconHelper = null;
          }
          if (mWebcompatReporterHandler != null) {
              mWebcompatReporterHandler.close();
              mWebcompatReporterHandler = null;
          }
      }

```

android/java/org/chromium/chrome/browser/shields/BraveUnifiedPanelHandler.java
```diff
@@ -0,0 +1,1760 @@
...
+
+    public void addObserver(BraveShieldsMenuObserver menuObserver) {
+        mMenuObserver = menuObserver;
+    }
+
+    public void show(@Nullable View anchorView, @Nullable Tab tab) {
+        if (mHardwareButtonMenuAnchor == null || mContext == null) {
+            return;
+        }
+
```
SergeyZhukovsky, line 276: Have you checked does the panel work in Custom Tab(it's a webview mode only when you open links from external apps). Old implementation called `ensureInitializedForCustomTabs` inside `show()`
```
    private void ensureInitializedForCustomTabs() {
        if (mHardwareButtonMenuAnchor == null && mContext == null) {
            mContext = BraveActivity.getCustomTabActivity();
            mHardwareButtonMenuAnchor = ((Activity) mContext).findViewById(R.id.menu_anchor_stub);
        }
    }
```

android/java/org/chromium/chrome/browser/preferences/website/BraveShieldsContentSettings.java
```diff
@@ -124,14 +124,64 @@ public static void setShields(
...
+                .setForgetFirstPartyStorageEnabled(defaultForgetMe, siteUrl, profile);
+
+        String defaultFingerprinting =
+                BraveShieldsContentSettingsJni.get().getFingerprintingControlType("", profile);
+        BraveShieldsContentSettingsJni.get()
+                .setFingerprintingControlType(defaultFingerprinting, siteUrl, profile);
+
+        String defaultHttpsUpgrade =
+                BraveShieldsContentSettingsJni.get().getHttpsUpgradeControlType("", profile);
+        if (!DEFAULT.equals(defaultHttpsUpgrade)) {
```
SergeyZhukovsky: That skips resetting the per-site HTTPS upgrade override when the global default happens to be "default". Every other setting is unconditionally reset. Is this intentional?

android/java/org/chromium/chrome/browser/shields/BraveUnifiedPanelHandler.java
```diff
@@ -0,0 +1,1760 @@
...
+                popupWindow.showAtLocation(
+                        mAnchorView,
+                        android.view.Gravity.BOTTOM | android.view.Gravity.CENTER_HORIZONTAL,
+                        0,
+                        yOffset);
+            } else {
+                popupWindow.showAsDropDown(mAnchorView);
+            }
+        } catch (Exception e) {
+            return null;
```
SergeyZhukovsky, line 441: maybe an assert here that it triggers in debug with exception description for diagnostic?

android/java/org/chromium/chrome/browser/shields/BraveUnifiedPanelHandler.java
```diff
@@ -0,0 +1,1760 @@
...
+            mWebcompatReporterHandler =
+                    WebcompatReporterServiceFactory.getInstance()
+                            .getWebcompatReporterHandler(mProfile, null);
+        }
+
+        if (mWebcompatReporterHandler != null && mReportContactInput != null) {
+            final TextInputEditText contactInput = mReportContactInput;
+            mWebcompatReporterHandler.getBrowserParams(
+                    (contactInfo, contactInfoSaveFlag, components) -> {
+                        if (contactInfo != null && !contactInfo.isEmpty()) {
```
SergeyZhukovsky, line 876: updates UI without checking on Activity lifecycle.

android/java/org/chromium/chrome/browser/shields/BraveUnifiedPanelHandler.java
```diff
@@ -0,0 +1,1760 @@
...
+                    });
+        }
+
+        if (mWebcompatReporterHandler != null
+                && mReportCategoryDropdown != null
+                && mContext != null) {
+            final AutoCompleteTextView dropdown = mReportCategoryDropdown;
+            mWebcompatReporterHandler.getWebcompatCategories(
+                    (categories) -> {
+                        if (categories == null || categories.length == 0 || mContext == null) {
```
SergeyZhukovsky: updates UI without checking on Activity lifecycle.

android/java/org/chromium/chrome/browser/shields/BraveUnifiedPanelHandler.java
```diff
@@ -0,0 +1,1760 @@
...
+@NullMarked
+public class BraveUnifiedPanelHandler {
+    private static final String TAG = "BraveUnifiedPanel";
+
+    private static class BlockersInfo {
+        public int mAdsBlocked;
+        public int mTrackersBlocked;
+        public int mScriptsBlocked;
+        public int mFingerprintsBlocked;
+        public final ArrayList<String> mBlockerNames = new ArrayList<>();
```
SergeyZhukovsky: is that ever populated?

android/java/brave-res/layout/brave_report_broken_site_panel.xml
```diff
@@ -0,0 +1,296 @@
...
+        <ImageView
+            android:id="@+id/report_back_button"
+            android:layout_width="24dp"
+            android:layout_height="24dp"
+            android:src="@drawable/ic_arrow_small_down"
+            android:rotation="90"
+            android:clickable="true"
+            android:focusable="true"
+            android:background="?android:attr/selectableItemBackgroundBorderless"
+            android:contentDescription="@string/report_broken_site_panel_title"/>
```
SergeyZhukovsky: `contentDescription` says the title instead of "Navigate back"

android/java/brave-res/layout/brave_trackers_ads_panel.xml
```diff
@@ -0,0 +1,182 @@
...
+        <ImageView
+            android:id="@+id/trackers_back_button"
+            android:layout_width="24dp"
+            android:layout_height="24dp"
+            android:src="@drawable/ic_arrow_small_down"
+            android:rotation="90"
+            android:clickable="true"
+            android:focusable="true"
+            android:background="?android:attr/selectableItemBackgroundBorderless"
+            android:contentDescription="@string/block_trackers_ads_title"/>
```
SergeyZhukovsky: `contentDescription` says the title instead of "Navigate back"

android/java/brave-res/layout/brave_cookies_panel.xml
```diff
@@ -0,0 +1,182 @@
...
+        <ImageView
+            android:id="@+id/cookies_back_button"
+            android:layout_width="24dp"
+            android:layout_height="24dp"
+            android:src="@drawable/ic_arrow_small_down"
+            android:rotation="90"
+            android:clickable="true"
+            android:focusable="true"
+            android:background="?android:attr/selectableItemBackgroundBorderless"
+            android:contentDescription="@string/block_cookies"/>
```
SergeyZhukovsky: `contentDescription` says the title instead of "Navigate back"

android/java/brave-res/layout/brave_trackers_ads_panel.xml
```diff
@@ -0,0 +1,182 @@
...
+
+        <!-- Title -->
+        <TextView
+            android:layout_width="0dp"
+            android:layout_height="wrap_content"
+            android:layout_weight="1"
+            android:layout_marginStart="16dp"
+            android:text="@string/block_trackers_ads_title"
+            android:textSize="20sp"
+            android:textColor="@color/shield_text_color"
```
SergeyZhukovsky: I wonder should we remove `shield_text_color` and `text_primary` and use a similar one from nala for both?

android/java/brave-res/values/unified_panel_strings.xml
SergeyZhukovsky: Plus we are missing translator comments here, it will be hard to translate them. Translators won't know what "Shields" or "Shred" means in Brave's context.

android/java/org/chromium/chrome/browser/shields/BraveUnifiedPanelHandler.java
SergeyZhukovsky: A static final String at the top of the class is fine.

android/java/org/chromium/chrome/browser/shields/BraveUnifiedPanelHandler.java
```diff
@@ -0,0 +1,1760 @@
...
+
+        try {
+            mPopupView = inflater.inflate(R.layout.brave_unified_panel_layout, null);
+            setupViews();
+        } catch (Exception e) {
+            return null;
+        }
+
+        float density = mContext.getResources().getDisplayMetrics().density;
+        int marginDp = 16;
```
SergeyZhukovsky: These should be on top of the class as static final
marginDp = 16 — popup margin
screenWidth * 0.50 — landscape width ratio (also in old code)
setElevation(20) — popup elevation (also in old code)
dpToPx(mContext, 24) — favicon icon size
dpToPx(mContext, 6) — favicon padding offset
dpToPx(mContext, -8) — cascading overlap margin
setAlpha(1.0f) — enabled state
setAlpha(0.5f) — disabled state
size() < 50 — max blocked URLs per tab, that's used twice, easy to get out of sync
{".com", ".net", ".org"} — favicon fallback TLDs

android/java/brave-res/layout/brave_cookies_panel.xml
```diff
@@ -0,0 +1,182 @@
...
+
+        <!-- Title -->
+        <TextView
+            android:layout_width="0dp"
+            android:layout_height="wrap_content"
+            android:layout_weight="1"
+            android:layout_marginStart="16dp"
+            android:text="@string/block_cookies"
+            android:textSize="20sp"
```
SergeyZhukovsky: all text in xml files have to use styles from nala. Look for example at https://github.com/brave/brave-core/blob/master/android/java/brave-res/layout/activity_brave_origin_plans.xml#L42

android/java/brave-res/values/unified_panel_strings.xml
Olipro: I'd figured it would be organisationally cleaner to put the strings in a new file. If you're sure that this is considered inferior then I'll migrate them.

SergeyZhukovsky: @mkarolin do you know will be strings from that new file translated?

android/java/brave-res/layout/brave_cookies_panel.xml
```diff
@@ -0,0 +1,182 @@
...
+    <LinearLayout
+        android:layout_width="match_parent"
+        android:layout_height="64dp"
+        android:orientation="horizontal"
+        android:gravity="center_vertical"
+        android:paddingHorizontal="16dp"
+        android:paddingVertical="18dp">
+
+        <!-- Back button -->
+        <ImageView
```
simoarpe, line 25: In general, is preferable to use `ImageButton`s unless you have a specific reason not to. They give you the correct accessibility semantics for free. Something you don't have with `ImageView`s.

Try a direct drop-in replacement with `clickable` and `focusable` removed:
```java
<ImageButton
    android:id="@+id/cookies_back_button"
    android:layout_width="24dp"
    android:layout_height="24dp"
    android:src="@drawable/ic_arrow_small_down"
    android:rotation="90"
    android:background="?android:attr/selectableItemBackgroundBorderless"
    android:contentDescription="@string/block_cookies"/>
```

If the outcome is the same this should be preferred as TalkBack will correctly announce it as a button rather than an image.

android/java/brave-res/layout/brave_cookies_panel.xml
```diff
@@ -0,0 +1,182 @@
...
+                android:layout_height="wrap_content"
+                android:text="@string/cookies_description"
+                android:textSize="14sp"
+                android:textColor="@color/shield_text_color"
+                android:paddingHorizontal="16dp"
+                android:paddingTop="12dp"
+                android:paddingBottom="24dp"/>
+
+            <!-- Radio options -->
+            <RadioGroup
```
simoarpe, line 78: Is this `RadioGroup` actually providing any value here? It can be probably replaced with a plain `LinearLayout`.

android/java/brave-res/layout/brave_https_upgrade_panel.xml
```diff
@@ -0,0 +1,181 @@
...
+    <LinearLayout
+        android:layout_width="match_parent"
+        android:layout_height="64dp"
+        android:orientation="horizontal"
+        android:gravity="center_vertical"
+        android:paddingHorizontal="16dp"
+        android:paddingVertical="18dp">
+
+        <!-- Back button -->
+        <ImageView
```
simoarpe, line 25: Same as above, this can probably become an `ImageButton`.

android/java/brave-res/layout/brave_report_broken_site_panel.xml
```diff
@@ -0,0 +1,296 @@
...
+    <LinearLayout
+        android:layout_width="match_parent"
+        android:layout_height="64dp"
+        android:orientation="horizontal"
+        android:gravity="center_vertical"
+        android:paddingHorizontal="16dp"
+        android:paddingVertical="18dp">
+
+        <!-- Back button -->
+        <ImageView
```
simoarpe, line 26: Same as above, this can probably become an `ImageButton`.

android/java/brave-res/layout/brave_https_upgrade_panel.xml
```diff
@@ -0,0 +1,181 @@
...
+                android:layout_height="wrap_content"
+                android:text="@string/https_upgrade_text"
+                android:textSize="14sp"
+                android:textColor="@color/shield_text_color"
+                android:paddingHorizontal="16dp"
+                android:paddingTop="12dp"
+                android:paddingBottom="24dp"/>
+
+            <!-- Radio option 1: Strict -->
+            <RadioGroup
```
simoarpe, line 78: Same as above, this `RadioGroup` can be probably replaced with a plain `LinearLayout`.

android/java/brave-res/layout/brave_unified_panel_layout.xml
```diff
@@ -0,0 +1,94 @@
...
+
+<FrameLayout
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    android:id="@+id/unified_panel_root"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    android:clipChildren="true"
+    android:clipToPadding="true"
```
simoarpe, line 15: `clipChildren` and `clipToPadding` are `true` by default for `ViewGroup`s so they can be safely removed.

android/java/brave-res/layout/brave_unified_panel_layout.xml
```diff
@@ -0,0 +1,94 @@
...
+    android:background="@drawable/unified_panel_background"
+    android:outlineProvider="background">
+
+    <!-- Main Panel Content -->
+    <ScrollView
+        android:id="@+id/main_panel_container"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:clipChildren="true"
+        android:clipToPadding="true">
```
simoarpe, line 25: Same here, they can be removed.

android/java/brave-res/layout/brave_unified_panel_layout.xml
```diff
@@ -0,0 +1,94 @@
...
+
+        <LinearLayout
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:clipChildren="true"
+            android:clipToPadding="true"
```
simoarpe, line 31: Same here, they can be removed.

android/java/brave-res/layout/brave_unified_panel_layout.xml
```diff
@@ -0,0 +1,94 @@
...
+    android:id="@+id/unified_panel_root"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    android:clipChildren="true"
+    android:clipToPadding="true"
+    android:background="@drawable/unified_panel_background"
+    android:outlineProvider="background">
+
+    <!-- Main Panel Content -->
+    <ScrollView
```
simoarpe, line 20: A `LinearLayout` that wraps a single `<include>` inside a `ScrollView` 👉 unnecessary nesting.

simoarpe (commented):
`brave_cookies_panel.xml`, `brave_https_upgrade_panel.xml`, and `brave_trackers_ads_panel.xml` are structurally identical: same top app bar, same card padding, same radio row template. Only IDs and string refs differ. 
This should be one shared layout inflated with different data, or at minimum the app bar and radio row should be extracted as `<include>` layouts. This alone would eliminate ~300 lines of duplicated XML.

The back-button + title pattern (64dp height, same padding, same structure) is copy-pasted across cookies, HTTPS, trackers, and report panels. Should be a single `<include>` layout.

android/java/brave-res/layout/brave_cookies_panel.xml
Olipro: ~~The RadioGroup provides mutual-exclusion for free. If this was changed, I believe it would then require Java code instead to handle deselection.~~

I see what you mean, ignore the above. Will change.

Olipro:
> `brave_cookies_panel.xml`, `brave_https_upgrade_panel.xml`, and `brave_trackers_ads_panel.xml` are structurally identical: same top app bar, same card padding, same radio row template. Only IDs and string refs differ. This should be one shared layout inflated with different data, or at minimum the app bar and radio row should be extracted as `<include>` layouts. This alone would eliminate ~300 lines of duplicated XML.
> 
> The back-button + title pattern (64dp height, same padding, same structure) is copy-pasted across cookies, HTTPS, trackers, and report panels. Should be a single `<include>` layout.

I think the minimal approach saves precious few lines. Going more full-on would dedupe it but result in punting the problem into Java land. Do you think it's better to implement the handling into Java code to dedupe it? Feels a bit like a _"pick your poison"_ situation so if you're adamant that inflating it from Java land is the superior option, I'll make it so.

simoarpe:
> I think the minimal approach saves precious few lines. Going more full-on would dedupe it but result in punting the problem into Java land. Do you think it's better to implement the handling into Java code to dedupe it? Feels a bit like a _"pick your poison"_ situation so if you're adamant that inflating it from Java land is the superior option, I'll make it so.

If you don't want to touch Java land and keep the refactoring at its minimum you can create a single XML containing the back button only:

```xml
<?xml version="1.0" encoding="utf-8"?>
<!-- Copyright (c) 2026 The Brave Authors. All rights reserved.
     This Source Code Form is subject to the terms of the Mozilla Public
     License, v. 2.0. If a copy of the MPL was not distributed with this file,
     You can obtain one at https://mozilla.org/MPL/2.0/. -->

<ImageButton
    xmlns:android="http://schemas.android.com/apk/res/android"
    android:id="@+id/panel_back_button"
    android:layout_width="24dp"
    android:layout_height="24dp"
    android:src="@drawable/ic_arrow_small_down"
    android:rotation="90"
    android:background="?android:attr/selectableItemBackgroundBorderless"
    android:contentDescription="@string/back"/>
```

And then replace its occurrences as it's replicated in four places (`brave_cookies_panel`, `brave_https_upgrade_panel`, `brave_report_broken_site_panel`, `brave_trackers_ads_panel`).

[You can use the `include` construct](https://developer.android.com/develop/ui/views/layout/improving-layouts/reusing-layouts#Include) and set the ID there, without shifting to Java code.

So for each layout it will become something like this:

`brave_cookies_panel.xml`
```java
        <!-- Back button -->
        <include
            android:id="@+id/cookies_back_button"
            layout="@layout/brave_panel_back_button"/>
```

android/java/org/chromium/chrome/browser/shields/BraveUnifiedPanelHandler.java
```diff
@@ -0,0 +1,1862 @@
...
+        mFaviconHelper.getLocalFaviconImageForURL(
+                mProfile,
+                origin,
+                mFaviconIconSize,
+                (bitmap, iconUrl) -> {
+                    if (mBlockedItemsContainer == null || mContext == null) return;
+                    Activity activity = (Activity) mContext;
+                    if (activity.isFinishing() || activity.isDestroyed()) return;
+
+                    activity.runOnUiThread(
```
SergeyZhukovsky: PostTask

android/java/org/chromium/chrome/browser/shields/BraveUnifiedPanelHandler.java
```diff
@@ -0,0 +1,1862 @@
...
+        mFaviconHelper.getLocalFaviconImageForURL(
+                mProfile,
+                registrableOrigin,
+                mFaviconIconSize,
+                (bitmap, iconUrl) -> {
+                    if (mBlockedItemsContainer == null || mContext == null) return;
+                    Activity activity = (Activity) mContext;
+                    if (activity.isFinishing() || activity.isDestroyed()) return;
+
+                    activity.runOnUiThread(
```
SergeyZhukovsky: PostTask

android/java/org/chromium/chrome/browser/shields/BraveUnifiedPanelHandler.java
```diff
@@ -0,0 +1,1862 @@
...
+                mProfile,
+                candidates.get(index),
+                mFaviconIconSize,
+                (bitmap, iconUrl) -> {
+                    if (mBlockedItemsContainer == null || mContext == null) return;
+                    Activity activity = (Activity) mContext;
+                    if (activity.isFinishing() || activity.isDestroyed()) return;
+
+                    activity.runOnUiThread(
+                            () -> {
```
SergeyZhukovsky: PostTask

android/java/brave-res/layout/brave_cookies_panel.xml
Olipro: ~~I see, problem is, I can't find any text styling in Nala that matches with what we have (which in turn came from the Figma design). Nothing in the XML you linked seems to be a match either?~~

After talking with Agustin, it's turned out that these are fine to style directly from Material - I've tweaked it so that it uses a standard heading style and dynamically downsizes to avoid truncation.

android/java/org/chromium/chrome/browser/shields/BraveUnifiedPanelHandler.java
```diff
@@ -0,0 +1,1902 @@
...
+    private void addBlockOverlay(FrameLayout container, int iconSize) {
+        if (mContext == null) return;
+
+        ImageView overlay = new ImageView(mContext);
+        overlay.setLayoutParams(new FrameLayout.LayoutParams(iconSize, iconSize));
+        overlay.setImageResource(R.drawable.ic_disable_outline);
+        overlay.setScaleType(ImageView.ScaleType.FIT_CENTER);
+        container.addView(overlay);
+    }
+}
```
netzenbot, line 2077: nit: `addObserver` is a public method with a non-obvious single-observer contract — calling it a second time silently replaces the first observer. A brief Javadoc describing the parameter and this replacement behavior would help callers. See [DOC-004](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-004).

browser/ui/android/strings/android_brave_strings.grd
```diff
@@ -626,6 +626,90 @@ This file contains all "about" strings.  It is set to NOT be translated, in tran
...
+      </message>
+      <message name="IDS_REPORT_BROKEN_SITE_PANEL_CONTACT_INFO" desc="Info text about contact data storage.">
+        If you provide contact info, it will be stored for future reports
+      </message>
+      <message name="IDS_REPORT_BROKEN_SITE_PANEL_SUBMIT" desc="Submit button label for the broken site report.">
+        Submit
+      </message>
+      <message name="IDS_REPORT_BROKEN_SITE_VIEW_SCREENSHOT" desc="Link text to view the screenshot attached to a broken site report.">
+        View screenshot
+      </message>
```
netzenbot, line 687: nit: `IDS_SHRED_AUTO_SHRED` is "Auto Shred" (title case) while every other menu item added in this PR uses sentence case — "Shred site data", "Shred site data now", "When site tabs are closed", etc. Per the established Android convention in this PR (confirmed by the designer), this should be "Auto shred". ([L10N-009](https://github.com/brave/brave-core/tree/master/docs/best-practices/localization.md#L10N-009))

android/java/org/chromium/chrome/browser/shields/BraveUnifiedPanelHandler.java
Olipro: This method is called `addBlockOverlay` and is private. This doesn't seem relevant

browser/ui/android/strings/android_brave_strings.grd
Olipro: Fixed

android/java/org/chromium/chrome/browser/shields/BraveUnifiedPanelHandler.java
Olipro: Diff must have been misaligned. Fixed.

android/java/org/chromium/chrome/browser/shields/BraveUnifiedPanelHandler.java
Olipro: The callback now checks this.

android/java/brave-res/values/unified_panel_strings.xml
Olipro: The user would have come to this dialog directly through them disabling shields. This seems redundant.

android/java/brave-res/drawable/ic_unified_shields_disabled.xml
Olipro: Should be fixed.

android/java/brave-res/drawable/ic_unified_shields_enabled.xml
Olipro: Should be fixed.

android/java/brave-res/drawable/ic_blocked_item_overlay.xml
Olipro: Should be fixed.

android/java/brave-res/values/unified_panel_strings.xml
Olipro: Should be fixed.

android/java/brave-res/drawable/ic_brave_unified_arrow_down.xml
Olipro: Should be fixed.

android/java/brave-res/drawable/ic_shred_dropdown_arrows.xml
Olipro: Should be fixed.

android/java/brave-res/values/unified_panel_strings.xml
Olipro: Should be fixed.

Olipro: Should be fixed.

android/java/res/values/brave_colors.xml
Olipro: I couldn't find a singular token that covers both light & dark that would also retain the look properly. If there is one, let me know.

android/java/brave-res/layout/brave_unified_panel_layout.xml
Olipro: I think this is out of date? There is no LinearLayout and the ScrollView prevents clipping in the  event the viewport dimensions are really small.

android/java/org/chromium/chrome/browser/shields/BraveUnifiedPanelHandler.java
Olipro: Should be resolved.

android/java/res/values-night/brave_colors.xml
Olipro: This is taken from Figma. By "there", I guess you mean Nala? I didn't find any consistent singular tokens that could make these redundant. Of course, maybe I'm just a dummy and overlooked them.

android/java/org/chromium/chrome/browser/shields/BraveUnifiedPanelHandler.java
Olipro: Not sure why this isn't showing as outdated, it gets logged now.

android/java/brave-res/layout/brave_cookies_panel.xml
Olipro: I've solved this with a delegate that makes TalkBack think it's a toggle.

android/java/org/chromium/chrome/browser/preferences/website/BraveShieldsContentSettings.java
Olipro: No, should be removed now.

android/java/org/chromium/chrome/browser/shields/BraveUnifiedPanelHandler.java
Olipro: Fixed

android/java/org/chromium/chrome/browser/shields/BraveUnifiedPanelHandler.java
Olipro: fixed

android/java/org/chromium/chrome/browser/shields/BraveUnifiedPanelHandler.java
Olipro: should be fixed

android/java/brave-res/values/unified_panel_strings.xml
Olipro: should be fixed

android/java/res/values/brave_colors.xml
Olipro: Resolving as there's an ongoing discussion in another comment about the same thing.

android/brave_java_sources.gni
```diff
@@ -382,6 +382,7 @@ brave_java_sources = [
   "../../brave/android/java/org/chromium/chrome/browser/shields/BraveShieldsMenuObserver.java",
   "../../brave/android/java/org/chromium/chrome/browser/shields/BraveShieldsScreenshotUtil.java",
   "../../brave/android/java/org/chromium/chrome/browser/shields/BraveShieldsUtils.java",
+  "../../brave/android/java/org/chromium/chrome/browser/shields/BraveUnifiedPanelHandler.java",
```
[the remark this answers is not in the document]
Olipro: I've looked into this same issue before. Due to the dependencies this has, it would be a significant undertaking to refactor how the sources build. Out of scope.

android/java/org/chromium/chrome/browser/shields/BraveUnifiedPanelHandler.java
Olipro: Resolved.

android/java/org/chromium/chrome/browser/shields/BraveUnifiedPanelHandler.java
Olipro: Resolved.

android/java/org/chromium/chrome/browser/shields/BraveUnifiedPanelHandler.java
Olipro: Resolved.

android/java/brave-res/layout/brave_unified_panel_layout.xml
Olipro: Should be resolved.

android/java/brave-res/layout/brave_unified_panel_layout.xml
Olipro: Should be resolved.

android/java/brave-res/layout/brave_unified_panel_layout.xml
Olipro: Should be resolved.

android/java/brave-res/layout/brave_https_upgrade_panel.xml
Olipro: Should be resolved.

android/java/brave-res/layout/brave_report_broken_site_panel.xml
Olipro: Should be resolved.

android/java/brave-res/layout/brave_https_upgrade_panel.xml
Olipro: Should be resolved.

android/java/brave-res/layout/brave_cookies_panel.xml
Olipro: Should be resolved.

android/java/org/chromium/chrome/browser/shields/BraveUnifiedPanelHandler.java
Olipro: Should be resolved.

android/java/res/values-night/brave_colors.xml
aguscruiz: Hey, all the colors from shields are Nala colors, I haven't used any custom ones. 

This one can be found, for example.

- `<color name="shields_container_background">#1F1F23</color>` is -> `Schemes/Surface` which is in this file `<color name="schemes_surface_container">#1F1F23</color>`

But we have others that don't seem to be in this .xml. @SergeyZhukovsky. Is there a reason why we don't import every single Nala colo into this .xml? (and the light mode one) I think we should so we don't find ourselves in this situation where someone doesn't find a color and hard-codes it. 

I can see that `schemes_surface` is a Nala color but it's not here, same with `systemfeedback_warning_text`.

SergeyZhukovsky: I see them both in 
  - Light: `node_modules/@brave/leo/tokens/android/values/colors.xml`
  - Dark: `node_modules/@brave/leo/tokens/android/values-night/colors.xml`
Both `schemes_surface` and `systemfeedback_warning_text` are defined at line 226 and line 61 respectively in each file.
`schemes_surface`
  - Light: #fafafb
  - Dark: #141415

`systemfeedback_warning_text`
  - Light: #745c00
  - Dark: #cfa608

android/java/org/chromium/chrome/browser/shields/BraveUnifiedPanelHandler.java
SergeyZhukovsky: I still see 
`mBlockScriptsSwitch.setAlpha(1.0f)`
`mBlockScriptsSwitch.setAlpha(0.5f)`
`mFaviconIconSize = ViewUtils.dpToPx(mContext, 24);`
`int faviconSize = iconSize - ViewUtils.dpToPx(mContext, 6);`

SergeyZhukovsky:
is there an issue for it? It has to say `Resolves: <issue URL>` in the PR description.

Olipro:
> is there an issue for it? It has to say `Resolves: <issue URL>` in the PR description.

Fixed.

android/java/org/chromium/chrome/browser/shields/BraveUnifiedPanelHandler.java
```diff
@@ -0,0 +1,2029 @@
...
+                            : Math.min(0f, Math.max(dx, getWidth() - bmpW));
+            dy =
+                    bmpH < getHeight()
+                            ? (getHeight() - bmpH) / 2f
+                            : Math.min(0f, Math.max(dy, getHeight() - bmpH));
+            mMatrix.setScale(scale, scale);
+            mMatrix.postTranslate(dx, dy);
+        }
+    }
+}
```
netzenbot, line 2077: The screenshot callback (posted to the UI thread via `ThreadUtils.postOnUiThread`) calls `setEnabled` and `setVisibility` on views without verifying the Activity hasn't been destroyed or finished. Add a check like `if (mContext == null || ((Activity) mContext).isFinishing() || ((Activity) mContext).isDestroyed()) return;` at the top of the lambda. ([Check `isActivityFinishingOrDestroyed()` Before UI Operations in Async Callbacks](https://github.com/brave/brave-core/tree/master/docs/best-practices/android.md#AND-001))

browser/ui/android/strings/android_brave_strings.grd
```diff
@@ -626,6 +626,90 @@ This file contains all "about" strings.  It is set to NOT be translated, in tran
...
+      <message name="IDS_REPORT_BROKEN_SITE_PANEL_SUBMITTED" desc="Confirmation message after submitting a broken site report.">
+        Report submitted. Thank you!
```
netzenbot, line 689: nit: The `desc` attributes for the new `IDS_SHRED_*` strings (e.g. `IDS_SHRED_SITE_DATA_TITLE`, `IDS_SHRED_AUTO_SHRED`) describe the UI element type but don't explain what "shred" means in Brave's context. Translators working only from the `desc` won't know that "shred" means permanently deleting a site's browsing data. Add a brief clarification such as "'Shred' is Brave's feature for permanently deleting a site's cookies and cached data" to the relevant descriptions. ([L10N-003](https://github.com/brave/brave-core/tree/master/docs/best-practices/localization.md#L10N-003))

android/java/org/chromium/chrome/browser/shields/BraveUnifiedPanelHandler.java
Olipro: I think this is out of date, I can't find any code in this file that does this with `mContext`

browser/ui/android/strings/android_brave_strings.grd
Olipro: Considering the pre-existing strings are similar, I don't see this as warranted.

android/java/org/chromium/chrome/browser/shields/BraveUnifiedPanelHandler.java
Olipro: Yeah, first two fixed

I'm not sure how mFaviconIconSize/faviconSize can be made constant since they depend on the runtime `mContext` value.

netzenbot (approved):

SergeyZhukovsky:
> > is there an issue for it? It has to say `Resolves: <issue URL>` in the PR description.
> 
> Fixed.

the issue usually lives here https://github.com/brave/brave-browser/issues

bsclifton:
@SergeyZhukovsky updated top post with new brave-browser issue 😄 
We were missing the issue after the scope had changed

SergeyZhukovsky (approved):
++

bsclifton (approved):
Strings LGTM! Some changes to casing - but matches Figma 🙂++

PR #34550, opened 2026-03-07
[iOS]Autofill: New Design Manage Passwords Settings - Part 2 of 2 Password Detail

### LINKED ISSUE #52211: Update Manage Logins and Passwords Settings UI to match new designs
1. User sees and selects "Passwords" in Autofill section of All Settings
2. User sees Managed Passwords view as per design. If empty, user sees empty list according to design spec
3. If not empty, user sees updated list of sites with saved passwords. User sees password settings above the list

This task applies the new design to what's already implemented.

EliHini:
Resolves [52211](https://github.com/brave/brave-browser/issues/52211)

## Summary
Redesigned Manage Password Detail screen in iOS Settings with a SwiftUI-based UI following Figma spec.
Added a new view for displaying passwords belonging to the same domain (Password Group).
The user is able to delete passwords in the group. 
Also updated authentication challenge to be performed before entering Manage Passwords.
### New SwiftUI Components

- **`ManagePasswordDetailView`** – Displays the contents of a CWVPassword object:
  - Provides menu to copy content on tap
  - Securely copies password into paste board
  - Hides passwords by default
  - Provides a reveal button

- **`ManagePasswordGroupViewMode`** - Data model for showing and modifying passwords in a group.
  - Edit mode for multi-select and bulk delete
  - For future Add password to the group (next PR)
  
- **`ManagePasswordGroupView`** – Displays a group of password items belonging to the same domain.

### Settings Integration
Added AutofillManagement scaffolding to remove code duplications and easy dependency injection
- `ManagePasswordsViewController` is now `ManageAutofillViewController`. It provides the ability to now Manage Password, and in the future Manage Credit Cards and Manage Addresses.
- `AutofillPrivacyLock` wraps `LAContext` for auth along with scene changes.
- Moved authentication challenge using `AutofillPrivacyLock` to the level of Management screens to match [Chromium and Password App](https://bravesoftware.slack.com/archives/C07LWRLAYH5/p1772833498884719)

### Test Password Detail
1. Ensure you have passwords already saved in Saved and Never Saved lists (use a version of the app prior to this)
2. Start up the app for this PR
3. Go to brave://flags
   a. Look for Chromium WebView's Autofill 
   b. Select Enabled
4. Go to all Settings
5. Tap on Manage Passwords
6. Observe the authentication Challenge. Unlock with the available mode (biometric or passcode)
7. Observe that your Saved and Never Saved Passwords are shown
8. Tab on a single password 
9. Observe that it navigates to a Password Detail screen
10. Observe the password contents: Site name, Username, and masked password with a reveal button
11. Show the password by tapping on the reveal button. Observe the revealed password.
12. Re-hide the reveal button to mask the password
13. Tap on the site item in the list. Observe the open menu showing options to Copy the site URL or Open Website
14. Tap on Copy. Paste it somewhere to confirm that the copy was successful
15. Tap on the site item again. Select Open Website. Observe that Brave navigates to the listed site.
16. Tap on the Username item. Observe the menu option to copy.
17. Tap on Copy. Paste it somewhere to confirm.
18. Tap on the Password item. Observe the menu option to copy.
19. Tap on Copy. Paste it somewhere secure to confirm.
20. Wait for 60 seconds. Attempt to paste it again. Observe that the pasteboard is empty.

### Test Password Group To Detail
1. Repeat Test Password Detail Steps 1 to 7.
2. Tap on a Password Item that has more than 1 account.
3. Observe a new screen showing the available passwords in that group.
4. Observe that the screen title is the name of the domain.
5. Tap on any of the password items.
6. Observe that it navigates to the Password Detail Screen.

### Test Password Group Edit
1. Repeat Test Password Group To Detail Steps 1 to 4.
2. Tap on a Password Item that has more than 1 account.
3. Tap on the Edit button.
4. Observe that the list is in selection mode.
5. Tap on any of the password items to select for deletion.
6. Proceed to delete by tapping on the Delete button.
7. Observe the delete confirmation message.
8. If you confirm the delete, observe that if there's only 1 password left, the screen navigates automatically to the details screen to show the detail of the remaining item.

### Test Password Group Swipe to Delete
1. Repeat Test Password Group Edit Steps 1 to 2.
2. Tap on a Password Item that has more than 1 account.
3. Swipe to delete on an item.
4. Observe that the item is deleted. Observe that if there's only 1 password left, the screen navigates automatically to the details screen to show the detail of the remaining item.

### Exclude from Test
1. Tapping on the Add Button (+) does nothing. Add item not implemented.


ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordGroupView.swift
```diff
@@ -0,0 +1,156 @@
...
+        .disabled(isEditMode)
+      }
+
+      if isEditMode {
+        ToolbarItem(placement: .bottomBar) {
+          Button(Strings.Autofill.managePasswordsDeleteCredentialButtonTitle) {
+            isDeleteSelectionDialogPresented = true
+          }
+          .foregroundStyle(
+            selectedCredentialIds.isEmpty ? Color(braveSystemName: .textSecondary) : .red
```
netzenbot: nit: `.red` is a system color. Use a Brave design system color token (e.g., `Color(braveSystemName: .systemfeedbackErrorText)`) for destructive actions, consistent with the rest of the file. [best practice](https://github.com/brave-experiments/brave-core-tools/tree/master/docs/best-practices/ios.md#IOS-027)

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManageAutofillViewController.swift
```diff
@@ -0,0 +1,45 @@
...
+import Preferences
+import SwiftUI
+import UIKit
+
+enum AutofillType {
+  case passwords
+}
+
+/// Hosts an autofill management view (passwords, credit cards, etc.) based on `AutofillType`.
+class ManageAutofillViewController: UIHostingController<AnyView> {
```
kylehickinson: Dont use `AnyView`, make a container view instead

kylehickinson: No reason to generalize this to avoid a few extra hosting controllers (which will go away anyways in the future), please revert to the old version

ios/brave-ios/Sources/Brave/Frontend/Autofill/AutofillManagementContext.swift
```diff
@@ -0,0 +1,34 @@
...
+
+import BraveCore
+import LocalAuthentication
+import SwiftUI
+
+/// Context required by autofill data management views (e.g. `ManagePasswordDetailView`) for authentication,
+/// URL opening, and passcode. Inject via `.environment(\.autofillContext, context)` at
+/// the root of the autofill flow (e.g. `ManagePasswordsView`).
+struct AutofillManagementContext {
+  let settingsDelegate: SettingsDelegate?
```
kylehickinson: Just use the `openURL` environment and in the custom handler send it to the settings delegate at the root level (check `.environment(\.openURL, ...` examples

ios/brave-ios/Sources/Brave/Frontend/Autofill/AutofillManagementContext.swift
```diff
@@ -0,0 +1,34 @@
...
+  let askForAuthentication: (@escaping (Bool, LAError.Code?) -> Void) -> Void
```
kylehickinson: We could probably just get away with using `LAContext` directly instead, never understood why we passed around a special "Ask for authentication" closure

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailView.swift
```diff
@@ -0,0 +1,127 @@
...
+import BraveStrings
+import BraveUI
+import SwiftUI
+import UIKit
+
+struct ManagePasswordDetailView: View {
+  @Environment(\.autofillManagementContext) private var context
+  @Environment(\.dismiss) private var dismiss
+  @State private var isPasswordRevealed = false
+  @State private var isSceneActive = true
```
kylehickinson: I wonder if we could just use the `.privacySensitive` modifier

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailView.swift
```diff
@@ -0,0 +1,127 @@
...
+  @Bindable var viewModel: ManagePasswordDetailViewModel
+
+  private var navigationTitle: String {
+    URL(string: viewModel.site)?.baseDomain ?? ""
+  }
```
kylehickinson: Only used in one place? Can just be inlined

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailView.swift
```diff
@@ -0,0 +1,127 @@
...
+
+  var body: some View {
+    List {
```
kylehickinson: `Form`

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailView.swift
```diff
@@ -0,0 +1,127 @@
...
+            Text(Strings.openWebsite)
+          }
+        } label: {
+          LabeledContent {
+            Text(viewModel.site)
+              .lineLimit(1)
+          } label: {
+            Text(Strings.Login.loginInfoDetailsWebsiteFieldTitle)
+          }
+          .contentShape(Rectangle())
```
kylehickinson: `.rect`

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailViewModel.swift
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) 2026 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import BraveCore
+
+@Observable
+final class ManagePasswordDetailViewModel {
```
kylehickinson: Why do we need a view model for a simple detail form? Just pass in the related information into the View directly.  Edit will have differing logic & views anyways

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailView.swift
```diff
@@ -0,0 +1,127 @@
...
+            Label {
+              Text(Strings.Autofill.managePasswordDetailsRevealPassword)
+            } icon: {
+              Image(braveSystemName: isPasswordRevealed ? "leo.eye.on" : "leo.eye.off")
+                .foregroundStyle(Color(braveSystemName: .textInteractive))
+            }
+            .labelStyle(.iconOnly)
+            .onTapGesture {
+              isPasswordRevealed.toggle()
+            }
```
kylehickinson: Don't use `onTapGesture` on a user facing button, its not the same from an accessibility standpoint. Just use `Button` here with a `plain` button style to avoid it taking over the entire cell's tap target

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailView.swift
```diff
@@ -0,0 +1,127 @@
...
+                  Text(viewModel.passwordValue)
+                    .lineLimit(1)
+                } else {
+                  // allowsHitTesting(false) intentionally prevents editing in View only mode
+                  SecureField(
+                    Strings.Autofill.managePasswordDetailsInputPasswordPlaceholder,
+                    text: $viewModel.passwordValue
+                  )
+                  .lineLimit(1)
+                  .allowsHitTesting(false)
```
kylehickinson: This doesn't stop it from revealing itself to VoiceOver as a text field, may need to, may need adjust accessibility traits

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordGroupView.swift
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  private var isEditMode: Bool {
+    editMode?.wrappedValue == .active
+  }
+
+  private var selectedCredentialsString: String {
+    viewModel.passwords
+      .filter { selectedCredentialIds.contains($0.identifier) }
+      .compactMap { $0.username }
```
kylehickinson: `.map({ $0.username ?? $0.title })`? Especially since this string is used in the confirmation dialog, we shouldn't be silently dropping passwords

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordGroupView.swift
```diff
@@ -0,0 +1,156 @@
...
+      .sorted()
+      .joined(separator: ", ")
```
kylehickinson: Separately, this should be using `ListFormatter`, not explicit joined with a comma. Since you have an array here you could just replace `.joined(...)` with `.formatted()`

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordGroupView.swift
```diff
@@ -0,0 +1,156 @@
...
+    }
+    .overlay {
+      if !isSceneActive {
+        Color(.braveGroupedBackground)
+          .ignoresSafeArea()
+      }
+    }
+    .onChange(of: viewModel.isEmpty) { _, isEmpty in
+      if isEmpty { dismiss() }
+    }
```
kylehickinson: Does this not get triggered anyways when the `NavigationLink` is removed at the root level? Since the parent view model will get the observation when the password list updates and then update its list

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordGroupView.swift
```diff
@@ -0,0 +1,156 @@
...
+      Section {
+        ForEach(viewModel.passwords, id: \.identifier) { password in
+          NavigationLink {
+            ManagePasswordDetailView(
+              viewModel: ManagePasswordDetailViewModel(
+                mode: .view(password),
+                autofillDataManager: viewModel.autofillDataManager
+              )
+            )
+            .environment(\.autofillManagementContext, context)
```
kylehickinson: This will be going away anyways but just fyi the environment should already be passing down to this hierarchy without needing additional `environment` modifiers

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordGroupViewModel.swift
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) 2026 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import BraveCore
+
+@MainActor
+@Observable
+final class ManagePasswordGroupViewModel {
```
kylehickinson: Doesn't seem like this view model is needed? Could probably just pass in the necessary info into the View directly and then handle deletions through the parent model?

kylehickinson (changes requested):
first pass, code only

ios/brave-ios/Sources/BraveStrings/BraveStrings.swift
```diff
@@ -9452,5 +9452,29 @@ extension Strings {
       comment:
         "Label for account when indicating that a domain has multiple accounts"
     )
+    public static let managePasswordDetailsInputUsernamePlaceholder =
```
kylehickinson: Not used anywhere atm

netzenbot (approved):

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManageAutofillViewController.swift
EliHini: I went back on forth on this one

EliHini: Sure but it is a lot of repetitive code

ios/brave-ios/Sources/Brave/Frontend/Autofill/AutofillManagementContext.swift
EliHini: Yeah that was my original go to but I wasn’t sure if we were doing something special with the settings delegate version of it or if it was just convention.

ios/brave-ios/Sources/Brave/Frontend/Autofill/AutofillManagementContext.swift
EliHini: Perfect, I will do that then

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailView.swift
EliHini: Actually that could be sufficient. Let me play with it and see

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailView.swift
EliHini: The plan was to use `List` for the view mode and `Form`for the Edit and Add modes

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailViewModel.swift
EliHini: I’m planning implementing the Edit and Add logics in here too

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailView.swift
EliHini: Good point

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordGroupView.swift
EliHini: I was expecting it to get triggered. I think it should. I may not need that, I had that there before I wired it all up. Good catch

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordGroupView.swift
EliHini: Normally the environment does but there seems to be some kind of quirkiness going on when the root is a UIKit view controller

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordGroupViewModel.swift
EliHini: I’m going to be adding the add password logic in there too

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailView.swift
EliHini: Oh that’s true 💯

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailView.swift
EliHini: Unfortunately `.privacySensitive` and `@Environment(\.redactionReasons) var redactionReasons` don't work either. When hosted inside UIHostingController, SwiftUI's environment is isolated from the UIKit scene graph. The `UIHostingController` creates a detached SwiftUI environment that has no knowledge of the parent UIScene's state transitions. In order to make this work I will need to find a way to manually inject the variables and their values at the call site like what I am doing with `openURL`

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManageAutofillViewController.swift
EliHini: I've removed this all together. Setup now happening at the call site

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailView.swift
EliHini: I've tidied this up

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailView.swift
EliHini: Using `Form` now

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailViewModel.swift
EliHini: Now sharing model from the parent.

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordGroupView.swift
EliHini: Same root cause as in https://github.com/brave/brave-core/pull/34550#discussion_r2916558310

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordGroupViewModel.swift
EliHini: Removed

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailView.swift
kylehickinson: You should be able to still use the modifiers and set `redactionReasons` yourself in a container view based on your original scene observer publishers

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordGroupView.swift
kylehickinson: I dont believe this is true, or you have something else going on. We have plenty of SwiftUI views that live inside of a UIHostingController configuration that use environment values set at the root that propagate properly (`AdvancedShieldsSettingsView` is on example)

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailView.swift
EliHini: I did try that but that didn't work either. Anyways we need more than just the redaction, we also need to authenticate the user before removing the screen privacy occlusion.

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordGroupView.swift
EliHini: Yes that's because in the other views, we are presenting the views as sheets on top of the new navigation stack.
As stated before, the issue is having a UKit NavigationViewController hosting with SwiftUI's version. Using NavigationLink to push the child view surfaces this problem because you've got a new tree. I see the problem if I do the same with `AdvancedShieldsSettingsView`

If we want to avoid this completely then the `ManagePasswordDetailView` and `ManagePasswordGroupView` must be presented in a sheet

EliHini: My latest commit is presents the child views modally so that SwiftUI owns the nav stack onward from there and avoiding the issues discussed above

kylehickinson: I dont think we'll want to present it modally from settings

EliHini: `ManagePasswordsView` is pushed onto the view controller. So nothing changes there. `ManagePasswordDetailView` and `ManagePasswordGroupView` are presented from `ManagedPasswordsView` just like we do with others such as `AdvancedShieldsSettingsView` presenting `ManageWebsiteDataView`

ios/brave-ios/Sources/Brave/Frontend/Autofill/AutofillPrivacyLock.swift
```diff
@@ -0,0 +1,70 @@
...
+/// .overlay { if lock.isLocked { Color(.braveGroupedBackground).ignoresSafeArea() } }
+/// .task { await lock.authenticate(onFailure: dismiss) }
+/// .onReceive(willDeactivate) { lock.lock() }
+/// .onReceive(didActivate)    { Task { await lock.authenticate(onFailure: dismiss) } }
+/// ```
+@MainActor
+@Observable
+final class AutofillPrivacyLock {
+  private(set) var isAuthenticated = false
+  private(set) var isAuthenticating = false
```
kylehickinson: This seems over-complicated, parent container should just be handling the lock/unlock state based on scene activations and passing the current state down the chain (using existing environment value `RedactionReasons`, custom env value, or just pass in directly) -- why would you want each a child view to be able to handle scene observation + lock it manually

ios/brave-ios/Sources/Brave/Frontend/Autofill/AutofillPrivacyLock.swift
```diff
@@ -0,0 +1,70 @@
...
+
+extension EnvironmentValues {
+  /// Shared privacy lock for the autofill management hierarchy.
+  /// The root ManagePasswordsView owns the instance; child views inherit it via environment
+  /// so they show the privacy overlay when backgrounded without running their own auth challenge on appear.
+  var autofillPrivacyLock: AutofillPrivacyLock? {
+    get { self[AutofillPrivacyLockEnvironmentKey.self] }
+    set { self[AutofillPrivacyLockEnvironmentKey.self] = newValue }
+  }
+}
```
kylehickinson: You can just use `@Entry` to avoid all this boilerplate
```suggestion
extension EnvironmentValues {
  @Entry var autofillPrivacyLock: AutofillPrivacyLock?
}
```

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailView.swift
```diff
@@ -0,0 +1,147 @@
...
+import BraveStrings
+import BraveUI
+import SwiftUI
+import UIKit
+
+struct ManagePasswordDetailView: View {
+  @Environment(\.openURL) private var openURL
+  @Environment(\.dismiss) private var dismiss
+  @Environment(\.autofillPrivacyLock) private var privacyLock
+  @Environment(\.presentationMode) @Binding private var presentationMode
```
kylehickinson: `PresentationMode` is deprecated, should just use `dismiss`

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailView.swift
```diff
@@ -0,0 +1,147 @@
...
+    self.password = password
+  }
+
+  var body: some View {
+    Group {
+      if isPresented {
+        NavigationView {
+          content
+        }
+        .navigationViewStyle(.stack)
```
kylehickinson: `NavigationView` is deprecated, use `NavigationStack`

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailView.swift
```diff
@@ -0,0 +1,147 @@
...
+      } else {
+        content
+      }
+    }
+    .overlay {
+      if privacyLock?.isLocked == true { Color(.braveGroupedBackground).ignoresSafeArea() }
+    }
```
kylehickinson: Doesn't hide existing content from being visible in VoiceOver, probably want to apply `accessibilityHidden` based on the locked state to the content

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailView.swift
kylehickinson: Still not fixed?

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordGroupView.swift
```diff
@@ -0,0 +1,163 @@
...
+import UIKit
+
+/// Displays all login credentials for a selected domain.
+/// Each credential is shown as a separate row (username + masked password).
+struct ManagePasswordGroupView: View {
+  @Environment(\.dismiss) private var dismiss
+  @Environment(\.editMode) private var editMode
+  @Environment(\.openURL) private var openURL
+  @Environment(\.autofillPrivacyLock) private var privacyLock
+  @Environment(\.presentationMode) @Binding private var presentationMode
```
kylehickinson: Same as above

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordGroupView.swift
```diff
@@ -0,0 +1,163 @@
...
+  @State private var selectedCredentialIds: Set<String> = []
+  @State private var isDeleteSelectionDialogPresented = false
+
+  private var passwords: [CWVPassword] {
+    (viewModel.allowedGroups.first { $0.domain == domain }
+      ?? viewModel.blockedGroups.first { $0.domain == domain })?.credentials ?? []
+  }
+
+  var body: some View {
+    NavigationView {
```
kylehickinson: Same as above, but I think we may revert anyways to the settings stack

kylehickinson (changes requested):
Not obvious items are tappable anymore, I think overall sheet change was a step back UX/UI wise.  I think we should either revert back to the old navigation hierarchy or we need to switch to the root screen being the sheet

Image

Menu shown when tapping the content in the detail page don't look good, also the password dots should probably be masked to a static length:

Image

Tapping on items while in selection mode opens a sheet instead of selecting them:

https://github.com/user-attachments/assets/d6df5617-ace0-4b37-99b0-b022d7b48d63

ios/brave-ios/Sources/Brave/Frontend/Autofill/AutofillPrivacyLock.swift
EliHini: the child views are not handling scene observations just the root view. This is used by the parent view controller so the child views just observe the property. I am going dow this route because `RedactionReasons` is not firing and it has to do with the whole disassociated hierarchy thing.

ios/brave-ios/Sources/Brave/Frontend/Autofill/AutofillPrivacyLock.swift
EliHini: oh yes that's true.

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailView.swift
EliHini: ah, I was using it because I was trying to determine if the view was presented or pushed. I don't need it anymore

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordGroupView.swift
EliHini: ok, let me revert to the settings stack

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordsView.swift
```diff
@@ -188,12 +182,17 @@ struct ManagePasswordsView: View {
...
+    .toolbar(!privacyLock.isLocked ? .visible : .hidden, for: .navigationBar)
+    .task { await privacyLock.authenticate(onFailure: { dismiss() }) }
     .onReceive(NotificationCenter.default.publisher(for: UIScene.willDeactivateNotification)) { _ in
-      isSceneActive = false
+      privacyLock.lock()
     }
     .onReceive(NotificationCenter.default.publisher(for: UIScene.didActivateNotification)) { _ in
-      isSceneActive = true
+      Task { await privacyLock.authenticate(onFailure: { dismiss() }) }
     }
```
nuo-xu, line 233: have you tested this with multi-window? since UIScene.didActivateNotification fires when ANY scene activates, not just this one

EliHini: great question, I actually got sidetracked when I started testing it. Testing it now

EliHini: seems to work correctly on both windows. I have one minimized and the other open and it applies the lock on the opened one. And if I bring the other one back it has it applied too

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailView.swift
```diff
@@ -29,7 +29,10 @@ struct ManagePasswordDetailView: View {
               Text(Strings.menuItemCopyTitle)
             }
             Button {
-              if let url = URL(string: password.site) {
+              if let url = URL(string: password.site),
+                url.schemeIsValid,
+                url.scheme == "http" || url.scheme == "https"
```
kylehickinson: isWebPage

EliHini:
@kylehickinson security review is completed https://github.com/brave/reviews/issues/2181

ShivanKaul:
@EliHini can we get some screen shots or screen recordings for this PR?

EliHini:
> @EliHini can we get some screen shots or screen recordings for this PR?

Certainly

EliHini:
> > @EliHini can we get some screen shots or screen recordings for this PR?
> 
> Certainly

@ShivanKaul 
https://github.com/user-attachments/assets/13961e02-2ec6-4bd5-8b59-f4c864d28eca

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailView.swift
```diff
@@ -0,0 +1,111 @@
...
+              HStack {
+                Spacer()
+                if isPasswordRevealed {
+                  Text(password.password ?? "")
+                    .lineLimit(1)
+                } else {
+                  Text(String(repeating: "•", count: 8))
+                    .lineLimit(1)
+                    .allowsHitTesting(false)
+                    .accessibility(hidden: !isPasswordRevealed)
```
kylehickinson: This is redundent, `isPasswordRevealed` will always be false here

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailView.swift
```diff
@@ -0,0 +1,111 @@
...
+                Text(Strings.menuItemCopyTitle)
+              }
+            } label: {
+              HStack {
+                Spacer()
+                if isPasswordRevealed {
+                  Text(password.password ?? "")
+                    .lineLimit(1)
+                } else {
+                  Text(String(repeating: "•", count: 8))
```
kylehickinson, line 85: Can you check what other Password managers do with VoiceOver on a hidden password element? Does it read it out anyways?

ios/brave-ios/Sources/Brave/Frontend/Autofill/AutofillPrivacyLock.swift
kylehickinson: I just mean the state can be held in the root view, no need to create a observable class and put it in the environment, you could just put the `isLocked` status in the environment from the root (or just do something like `.environment(\.redactionReasons, isLocked ? [.privacy] : [])` when locked, then use `@Environment(\.redactionReasons) var redactionReasons` in the views to check. The `redactionReasons` environment value is not read-only.  You can even just do a `ViewModifier` to hold the state if you dont want the root view exposed.

Mostly there's no reason to expose `lock` and `authenticate` to children in the environment

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailView.swift
```diff
@@ -0,0 +1,111 @@
...
+                    .frame(maxWidth: .infinity, alignment: .trailing)
+                }
+              }
+              .contentShape(.rect)
+            }
+
+            Button {
+              isPasswordRevealed.toggle()
+            } label: {
+              Image(braveSystemName: isPasswordRevealed ? "leo.eye.on" : "leo.eye.off")
```
kylehickinson: `Label`  w/ iconOnly style with appropriate text for the given state

ios/brave-ios/Sources/BraveStrings/BraveStrings.swift
```diff
@@ -9452,5 +9452,29 @@ extension Strings {
...
+        bundle: .module,
+        value: "Reveal",
+        comment: "Reveal password label"
+      )
+    public static let managePasswordsAuthenticationReason =
+    NSLocalizedString(
+      "autofill.managePasswords.authenticationReason",
+      tableName: "BraveShared",
+      bundle: .module,
+      value: "Authenticate to view.",
```
kylehickinson: Probably need better copy here

kylehickinson (changes requested):

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailView.swift
EliHini: Chrome: "Eight bullets textfield dim" -- Says the number of bullets
Password App: "20 bullets text field, actions available" -- Says the number bullets, and the actions available

ios/brave-ios/Sources/BraveStrings/BraveStrings.swift
EliHini: @rmcfadden3 Any ideas here? "Unlock device to view."

rmcfadden3: @EliHini — can you give full context (or a screenshot) of where use might encounter this string, and what they'd be trying to accomplish? Right now I'm not sure on context so can't quite give copy suggestion.

EliHini:
> I just mean the state can be held in the root view, no need to create a observable class and put it in the environment, you could just put the isLocked status in the environment from the root

The motivation behind this is two folds:

1. I am tracking `isAuthenticating` and `isAuthenticated` states to determine `isLocked`. So in the interest of separating concerns and not have all that state management in the root view I elected to refactor out the implementation into `AutofillPrivacyLock`.
2. This is also forward looking given that I will be implementing the same logic in the root view for Manage Credit Cards and Manage Address 

> or just do something like .environment(\.redactionReasons, isLocked ? [.privacy] : []) when locked, then use @Environment(\.redactionReasons) var redactionReasons in the views to check

I reckoned you were referring to this approach. I did attempt that but went with the environment variable for `AutofillPrivacyLock` since `redactionReasons`is never mutated because of the issue with the Scene and navigation stack (as explained earlier) not being passed down to child views without the manual `.environment(\.redactionReasons, isLocked ? [.privacy] : [])` to propagate it. 

You are right that the child views don't need to be exposed to `lock` and `authenticate`. I will remove their availability by way the environment variables

ios/brave-ios/Sources/BraveStrings/BraveStrings.swift
EliHini: @rmcfadden3 - Yes. This is for the localized reason that app provides the user the first time the challenges the user to authenticate before revealing the manage password settings screens.

netzenbot (approved):

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordsView.swift
```diff
@@ -125,7 +125,7 @@ struct ManagePasswordsView: View {
     .overlay {
       if privacyLock.isLocked { Color(.braveGroupedBackground).ignoresSafeArea() }
     }
-    .environment(\.autofillPrivacyLock, privacyLock)
+    .environment(\.redactionReasons, privacyLock.isLocked ? [.privacy] : [])
```
kylehickinson: You should get the current redaction reasons from the environment and union this value rather than replace it in case Apple ever does fix the bug with it from inside UIHostingControllers (or we eventually move to a SwiftUI root)

ios/brave-ios/Sources/BraveStrings/BraveStrings.swift
rmcfadden3: Okay. I still don't totally understand the context, but would something like `Please verify it's you to manage your passwords` or even `Please authenticate to manage your passwords` or similar work?

EliHini: `Please authenticate to manage your passwords` or `Authenticate to manage your passwords` sound better to me.

The context is that when the user taps on "Manage Passwords" in the Autofill section of the Settings menu, the OS will present them with a reason for requesting biometrics or pincode. This text will be displayed in the reason.

rmcfadden3: Ooh. Okay, yeah, that's helpful context.

I agree that `Please authenticate to manage your passwords` seems fine then.

EliHini: Perfect! That's done.

kylehickinson:
Failing Face ID/auth in general and cancelling causes nav bars to disappear when getting dismissed back to root settings

https://github.com/user-attachments/assets/86ce18b1-8592-4b32-9082-c4d65107092f

kylehickinson:
Delete button in multi-select is not red when items are selected: 
image

kylehickinson:
Same on group screen: 
image

kylehickinson:
Empty toolbar on the details screen should probably be hidden: 
image

kylehickinson:
Missing proper dark mode colors for list background/rows (also somehow nav bar?): 
image

kylehickinson (changes requested):
see comments for bugs

EliHini:
> Missing proper dark mode colors for list background/rows (also somehow nav bar?): image

What are the proper dark mode colors? I am doing anything special for the colors.

kylehickinson:
@EliHini `braveGroupedBackground`/`secondaryBraveGroupedBackground`, see other usages of that across many different settings SwiftUI views

EliHini:
> Empty toolbar on the details screen should probably be hidden: image

This is temporary because the Delete button will go there on the Edit/Add PR

ios/brave-ios/Sources/Brave/Frontend/Autofill/AutofillPrivacyLock.swift
```diff
@@ -54,3 +54,11 @@ final class AutofillPrivacyLock {
     }
   }
 }
+
+struct AutofillPrivacyLockExitOnFailureKey: EnvironmentKey {
+  static let defaultValue: (() -> Void)? = nil
+}
```
kylehickinson: unused

ios/brave-ios/Sources/Brave/Frontend/Autofill/AutofillPrivacyLock.swift
```diff
@@ -54,3 +54,11 @@ final class AutofillPrivacyLock {
...
+
+extension EnvironmentValues {
+  @Entry var autofillPrivacyLockExitOnFailure: (() -> Void)?
+}
```
kylehickinson, line 69: Im pretty sure we've had problems in the past storing straight-up closures in the SwiftUI environment (In wallet? @nuo-xu maybe your remember).  The general pattern to follow instead is create an action struct similar to `OpenURLAction`, `DismissAction`, etc. We have examples of this: `OpenTabURLAction`, `AppRatingRequestAction`, etc

ios/brave-ios/Sources/Brave/Frontend/Settings/SettingsViewController.swift
```diff
@@ -1290,6 +1290,18 @@ class SettingsViewController: TableViewController {
...
+                  // Failed privacy-lock auth must exit the entire navigation flow: `dismiss` only pops the
+                  // top SwiftUI screen, so from second order stack i.e detail/group
+                  // we would not return to settings. `popToViewController` unwinds every `NavigationLink`-pushed host;
+                  // then restore nav bar and toolbar because child SwiftUI can hide them on the shared `UINavigationController`.
+                  .environment(\.autofillPrivacyLockExitOnFailure) { [weak self] in
+                    Task { @MainActor in
+                      guard let self, let nav = self.navigationController else { return }
+                      nav.popToViewController(self, animated: true)
+                      nav.setNavigationBarHidden(false, animated: false)
+                      nav.setToolbarHidden(true, animated: false)
```
kylehickinson: Still seems like somehow you've landed on the wrong solution here if this is necessary

ios/brave-ios/Sources/Brave/Frontend/Autofill/AutofillSettingsHostingController.swift
```diff
@@ -0,0 +1,23 @@
...
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import SwiftUI
+import UIKit
+
+final class AutofillSettingsHostingController<Content: View>: UIHostingController<Content> {
+  override func viewWillAppear(_ animated: Bool) {
+    super.viewWillAppear(animated)
+    // UINavigationController hides the toolbar by default.
+    navigationController?.setNavigationBarHidden(true, animated: animated)
```
kylehickinson: Why are we hiding the nav bar?

ios/brave-ios/Sources/Brave/Frontend/Settings/SettingsViewController.swift
EliHini: How so? What part is wrong? 
`nav.popToViewController(self, animated: true)`: the main issue here is that when lock challenge fails we need to pop back to the settings view. Popping a nested view controller doesn't return here.

`nav.setNavigationBarHidden(false, animated: false)`: Needed to restore the hidden state required in order to have navigation bar on the SwiftUI that is hosted (avoids double nav bars).

`nav.setToolbarHidden(true, animated: false)`: This may be redundant.

ios/brave-ios/Sources/Brave/Frontend/Autofill/AutofillSettingsHostingController.swift
EliHini: that's a typo.

ios/brave-ios/Sources/Brave/Frontend/Settings/SettingsViewController.swift
kylehickinson: - You shouldn't be hiding the navigation controllers nav bar at all in general?
- Why would you have double nav bars? I don't see any NavigationStack's inside the SwiftUI hierarchy?
- You don't have to use UIKit to pop to root, you can just call the DismissAction from the root view?

EliHini: > You shouldn't be hiding the navigation controllers nav bar at all in general?

1. Nav bar is hidden in `ManagePasswordsView` when the lock is active and the auth challenge is presented line on 202 `.toolbar(!privacyLock.isLocked ? .visible : .hidden, for: .navigationBar)`. `isLocked` is a computed variable. If you cancel the auth challenge, then the nav bar remains hidden after the rootView has been dismissed.

> Why would you have double nav bars? I don't see any NavigationStack's inside the SwiftUI hierarchy?
2. That was an error. We've been talking in around this issue for while that I am getting the issues mixed up.

> You don't have to use UIKit to pop to root, you can just call the DismissAction from the root view?
3. I am using `DismissAction` see line  231 in `ManagePasswordsView`. The issue is that the root view's dismiss action has no effect on the Child View in the navigation stack. 
Since  a child view (GroupPassword or Detail) no longer observes `privacyLock.isLocked` the stack has to be unwind at the at the base of the tree while making sure the nav bar is restored as per explanation in 1. 

We can go back to the way it was where each view observes privacyLock and calls dismiss on failure to pop itself.

kylehickinson: > Nav bar is hidden in ManagePasswordsView when the lock is active and the auth challenge is presented line on 202

Dont do that? Worst case you can wrap the toolbar items itself to avoid those being added to the nav bar when locked, but the nav bar itself can remain visible.

EliHini: > Dont do that? Worst case you can wrap the toolbar items itself to avoid those being added to the nav bar when locked, but the nav bar itself can remain visible.

If the bar is not hidden then it will show the bar chrome when the screen is locked. That's the only reason I went with that approach

kylehickinson: Thats fine, we dont care about a bit of chrome showing, the only thing we need to make sure is that users password data doesnt leak into the OS app switcher screenshots.

ios/brave-ios/Sources/Brave/Frontend/Autofill/AutofillPrivacyLock.swift
EliHini: Fixed. using an action struct.

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordsView.swift
```diff
@@ -126,22 +132,23 @@ struct ManagePasswordsView: View {
...
-      if !isSceneActive {
-        Color(.braveGroupedBackground)
-          .ignoresSafeArea()
-      }
+      if privacyLock.isLocked { Color(.braveGroupedBackground).ignoresSafeArea() }
     }
+    .environment(\.redactionReasons, effectiveRedactionReasons)
     .toolbarBackground(.visible, for: .navigationBar)
-    .navigationTitle(Strings.Autofill.managePasswordsTitle)
+    .navigationTitle(privacyLock.isLocked ? "" : Strings.Autofill.managePasswordsTitle)
```
kylehickinson: You dont need to hide the word "Manage Passwords" when locked

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordGroupView.swift
```diff
@@ -0,0 +1,155 @@
...
+
+      if isEditMode {
+        ToolbarItem(placement: .bottomBar) {
+          Button(Strings.Autofill.managePasswordsDeleteCredentialButtonTitle) {
+            isDeleteSelectionDialogPresented = true
+          }
+          .foregroundStyle(
+            Color(
+              braveSystemName: selectedCredentialIds.isEmpty
+                ? .textSecondary : .systemfeedbackErrorVibrant
```
kylehickinson: Use `textDisabled` for disabled text

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordsView.swift
```diff
@@ -150,7 +157,10 @@ struct ManagePasswordsView: View {
             isDeleteSelectionDialogPresented = true
           }
           .foregroundStyle(
-            selectedGroupIds.isEmpty ? Color(braveSystemName: .textSecondary) : .red
+            Color(
+              braveSystemName: selectedGroupIds.isEmpty
+                ? .textSecondary : .systemfeedbackErrorVibrant
```
kylehickinson: same as above

kylehickinson:
We may want to differentiate some by color (textPrimary/textSecondary) for the labeled content in the detail page: 
image

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailView.swift
```diff
@@ -0,0 +1,120 @@
...
+            }
+
+            Button {
+              isPasswordRevealed.toggle()
+            } label: {
+              Label(
+                Strings.Autofill.managePasswordDetailRevealPassword,
+                braveSystemImage: isPasswordRevealed ? "leo.eye.on" : "leo.eye.off"
+              )
+              .foregroundStyle(Color(braveSystemName: .textInteractive))
```
kylehickinson: iconInteractive

kylehickinson:
Not sure how but the nav bar is still the wrong background color in dark mode? 
image

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordGroupView.swift
```diff
@@ -0,0 +1,155 @@
...
+      }
+
+      ToolbarItem(placement: .bottomBar) {
+        EditButton()
+          .disabled(passwords.isEmpty)
+      }
+    }
+    .toolbar(redactionReasons.contains(.privacy) ? .hidden : .visible, for: .bottomBar)
+    .overlay {
+      if redactionReasons.contains(.privacy) { Color(.braveGroupedBackground).ignoresSafeArea() }
```
kylehickinson, line 128: Should probably just apply opacity(0) to the actual list instead

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordsView.swift
```diff
@@ -211,30 +225,51 @@ struct ManagePasswordsView: View {
...
+          // NavigationLink destinations pushed onto a UINavigationController stack run in an isolated
+          // hosting context and do not inherit the parent's environment; Thus the parent's environment variables
+          // must be re-injected explicitly into the child view on the isolated stack.
+          .environment(\.openURL, openURL)
+          .environment(\.redactionReasons, redactionReasons)
+      } else {
+        ManagePasswordGroupView(viewModel: viewModel, domain: domain)
+          .environment(\.openURL, openURL)
+          .environment(\.redactionReasons, redactionReasons)
+      }
```
kylehickinson: you can Group these instead just fyi

```
Group {
  if .. else .. 
}
.environment(...)
```

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordGroupView.swift
EliHini: You want to go back to using the opacity? At one point It was like that, I think in the original draft, and you suggested overlay

EliHini:
> Not sure how but the nav bar is still the wrong background color in dark mode? image

I see the issue. Seems to only affect non-Liquid glass.
needed to set toolbar's background since, I have it visible.

Screenshot 2026-04-02 at 10 24 33 AM

kylehickinson:
Navigation bar color is still wrong 
image

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailView.swift
```diff
@@ -38,9 +38,11 @@ struct ManagePasswordDetailView: View {
           } label: {
             Text(password.site).lineLimit(1)
               .contentShape(.rect)
+              .foregroundStyle(Color(braveSystemName: .textPrimary))
           }
         } label: {
           Text(Strings.Login.loginInfoDetailsWebsiteFieldTitle)
+            .foregroundStyle(Color(braveSystemName: .textSecondary))
```
kylehickinson: These are all backwards, the value should have textSecondary, the label should have textPrimary

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailView.swift
```diff
@@ -110,7 +117,7 @@ struct ManagePasswordDetailView: View {
     .foregroundStyle(Color(braveSystemName: .textPrimary))
     .accessibility(hidden: redactionReasons.contains(.privacy) ? true : false)
     .navigationBarBackButtonHidden(redactionReasons.contains(.privacy))
```
kylehickinson: Dont need to hide the back button

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordGroupView.swift
```diff
@@ -62,7 +62,7 @@ struct ManagePasswordGroupView: View {
     .scrollContentBackground(.hidden)
     .background((Color(.braveGroupedBackground)))
     .navigationBarBackButtonHidden(redactionReasons.contains(.privacy))
-    .navigationTitle(redactionReasons.contains(.privacy) ? "" : domain)
+    .navigationTitle(domain)
```
kylehickinson: maybe want to hide the domain or replace it with `Manage Passwords`

kylehickinson:
Failing the auth shows auth again on the root settings even after being popped back 
image

EliHini:
> Navigation bar color is still wrong image

what color should I be using? I am using `braveGroupedBackground`

kylehickinson:
Second auth still happens when getting kicked back to settings

https://github.com/user-attachments/assets/f664ca91-fc8d-464c-9b07-14345f2a4860

EliHini:
Fixed re-entrant lock challenge.
https://github.com/user-attachments/assets/a17e4698-e42d-4645-9012-bd8e80bc0649

ios/brave-ios/Sources/Brave/Frontend/Autofill/AutofillPrivacyLock.swift
```diff
@@ -53,6 +56,14 @@ final class AutofillPrivacyLock {
...
     }
   }
+
+  /// Permanently disables this lock. After calling, `authenticate(onFailure:)` is a no-op
+  /// and `isLocked` stays true. Use from the dismissal path so a host view controller that
+  /// recycles the SwiftUI view cannot trigger another biometric/passcode prompt.
+  func invalidateSession() {
+    isSessionValid = false
+    lock()
+  }
```
kylehickinson: This doesn't solve the actual issue that you are locking/unlocking based on scene activation state changes when the prompts themselves trigger activation state changes, so you are ending up with `AutofillPrivacyLock.lock` and `AutofillPrivacyLock.authenticate` calls when you shouldn't. You may need to split these out using `UIScene.didEnterBackgroundNotification` and `UIScene.willEnterForegroundNotification` to trigger the actual local authentication invalidation and activation state to control visibility

kylehickinson: If you just swap the `onReceive`'s in `ManagePasswordsView` to use the didEnterBackground/willEnterForeground it works correctly without this commits changes at all, however when you begin backgrounding the app the content wont hide until if fully backgrounds hence likely needing to split things up

EliHini: > This doesn't solve the actual issue that you are locking/unlocking based on scene activation state changes when the prompts themselves trigger activation state changes, so you are ending up with `AutofillPrivacyLock.lock` and `AutofillPrivacyLock.authenticate` calls when you shouldn't. You may need to split these out using `UIScene.didEnterBackgroundNotification` and `UIScene.willEnterForegroundNotification` to trigger the actual local authentication invalidation and activation state to control visibility

I thought the actual issue was to not have the lock challenge after ManagePasswordsView has been dismissed. This fix here simply attenuates the call if the moment that the View is still be held onto by SettingsViewController. 

I had played with `UIScene.willEnterForegroundNotification` and `UIScene.didEnterBackgroundNotification` but they cause other issue.

kylehickinson (changes requested):
Toolbar gets stuck on main settings screen when leaving passwords list:

https://github.com/user-attachments/assets/ea3e0366-0d0e-43e1-a39e-6d3f4dfeed7e

EliHini:
Closing this to fix branch name issue. 
New PR published here https://github.com/brave/brave-core/pull/36449

PR #1196, opened 2019-01-02
Prevent loading brave scheme url from javascript

simonhong:
To fix this, brave scheme is registered to WebSecurityPolicy and
RequestScheme of ChildProcessSecurityPolicy::SecurityState.
Also, makes ProfileIOData::IsHandledProtocol consider brave scheme.

As brave scheme can be loaded only from brave scheme,
it should use chrome scheme instead of brave scheme in anchor tag..
Otherwise, that brave webui loading is prevented in brave webui from renderer.

Fix https://github.com/brave/brave-browser/issues/2861
Fix https://github.com/brave/brave-browser/issues/2849

## Submitter Checklist:

- [x] Submitted a [ticket](https://github.com/brave/brave-browser/issues) for my issue if one did not already exist.
- [x] Used Github [auto-closing keywords](https://help.github.com/articles/closing-issues-via-commit-messages/) in the commit message.
- [ ] Added/updated tests for this change (for new code or code which already has tests).
- Verified that these changes build without errors on
  - [ ] Windows
  - [x] macOS
  - [ ] Linux
- Verified that these changes pass automated tests (`npm test brave_unit_tests && npm test brave_browser_tests`) on
  - [ ] Windows
  - [x] macOS
  - [ ] Linux
- [x] Ran `git rebase master` (if needed).
- [ ] Ran `git rebase -i` to squash commits (if needed).
- [x] Tagged reviewers and labelled the pull request as needed.
- [ ] Request a security/privacy review as needed.
- [x] Add appropriate QA labels (QA/Yes or QA/No) to include the closed issue in milestone

## Test Plan:
Test 1.
  1. Navigate to https://world_languages.gitlab.io/bravepoc/
  2. Click link
  3. Check brave://settings is not loaded in new tab.

Test 2.
  1. Navigate to brave://welcome
  2. Move to `Change your search engine` page in welcome page
  3. Click `Settings` button
  4. Check settings page is opened in new tab
## Reviewer Checklist:

- [ ] New files have MPL-2.0 license header.
- [ ] Request a security/privacy review as needed.
- [ ] Adequate test coverage exists to prevent regressions 
- [ ] Verify test plan is specified in PR before merging to source

PR #34610, opened 2026-03-10
Don't use virtual url for brave/chrome

### LINKED ISSUE #22872: brave://wallet redirects to chrome://wallet
## Description 
`brave://wallet` redirects to `chrome://wallet` 

## Steps to reproduce 

   1. Go to `brave://wallet`
  
## Actual result 
Redirected to `chrome://wallet`

## Expected result
Opening `brave://wallet`

## Issue reproduces how often 
Easily reproduced

## Version/Channel Information:

- Can you reproduce this issue with the current Play Store version? yes
- Can you reproduce this issue with the current Play Store Beta version? yes
- Can you reproduce this issue with the current Play Store Nightly version? yes

## Device details
   - Install type (ARM, x86): ARM
   - Device type (Phone, Tablet, Phablet): Phone, OnePlus 7 Pro
   - Android version: 11
   
## Brave version
Brave 1.38.113, Chromium 101.0.4951.61

### Website problems only
- Does the issue resolve itself when disabling Brave Shields? 
- Does the issue resolve itself when disabling Brave Rewards?
- Is the issue reproducible on the latest version of Chrome?

### LINKED ISSUE #34983: Internal pages showing `chrome://`. They should always show `brave://`
## Description

Example screenshot below. Note how it says `chrome://`. It should say `brave://`. 

IMG_0082

### LINKED ISSUE #53508: chrome vs brave scheme should be display only
We have had an endless stream of issues related to brave vs chrome scheme checks and just recently ran into what looks like a lot more of them on android. The only way to safely resolve this is to make `brave` scheme display only and never part of the actual url. This would also be similar to `about://flags` where `about` gets rewritten to `chrome` without becoming the virtual url. Using `brave` as the virtual url is just too risky in terms of security and I think we need to accept that we may have some display regressions to fix this problem.

### LINKED ISSUE #53549: brave:// -> chrome:// rewrite causes many upstream checks to fail to properly recognize the pages as chrome://
We currently change brave -> chrome scheme in a way that results in the real and virtual url being different. Because some checks use the virtual url and some use the real url, this results in some of them failing to correctly identify the pages as `chrome://`. `brave://`

bridiver:
Use `brave://` for display purposes only. Internally the scheme should always be `chrome://`

Resolves https://github.com/brave/brave-browser/issues/53549
Resolves https://github.com/brave/brave-browser/issues/34983
Resolves https://github.com/brave/brave-browser/issues/22872
Resolves https://github.com/brave/brave-browser/issues/53508

QA:
Verify chrome urls direct to brave
Verify brave urls are not added to history
Test plan from https://github.com/brave/brave-core/pull/1196 partially works, but use brave://about/ to check navigating from one brave "host" to another
check that session restore with brave:// pages works correctly.


android/java/org/chromium/chrome/browser/BraveRewardsNativeWorker.java
```diff
@@ -121,12 +121,8 @@ public void removePublisherObserver(PublisherObserver observer) {
     }
 
     public void onNotifyFrontTabUrlChanged(int tabId, String url) {
-        // Check both chrome:// and brave:// schemes. brave:// URLs (e.g.
-        // brave://version) are not recognized by the publisher lookup and
-        // would leave a stale verified-publisher checkmark from a previous tab.
         boolean internalUrl =
-                url.startsWith(UrlConstants.CHROME_SCHEME)
-                        || url.startsWith(BraveUrlConstants.BRAVE_SCHEME);
+                url.startsWith(UrlConstants.CHROME_SCHEME);
```
bridiver: cc @yrliou

fallaciousreasoning:
This looks great

android/java/org/chromium/chrome/browser/BraveRewardsNativeWorker.java
bridiver: this is fixed now in terms of display, but no idea if that addresses the issue here. We definitely can't use brave as the real or virtual url because that messes up important checks in android

fallaciousreasoning (approved):

browser/ui/toolbar/brave_location_bar_model_delegate.cc
```diff
@@ -100,7 +97,7 @@ const gfx::VectorIcon* BraveLocationBarModelDelegate::GetVectorIconOverride()
 
   // Defer to the Chromium implementation first.
   const gfx::VectorIcon* parent_icon =
-      BrowserLocationBarModelDelegate::GetVectorIconOverride();
+      ChromeLocationBarModelDelegate::GetVectorIconOverride();
```
fallaciousreasoning, line 100: haven't checked the code but could this return icons in more cases than the base class?

bridiver: It's not the base class, Browser subclasses Chrome

bridiver: The only thing different about Browser is that it adds a dependency on tab strip that isn't allowed on android which is why I inverted this relationship

chromium_src/chrome/browser/ui/android/toolbar/location_bar_model_android.h
```diff
@@ -0,0 +1,17 @@
...
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#ifndef BRAVE_CHROMIUM_SRC_CHROME_BROWSER_UI_ANDROID_TOOLBAR_LOCATION_BAR_MODEL_ANDROID_H_
+#define BRAVE_CHROMIUM_SRC_CHROME_BROWSER_UI_ANDROID_TOOLBAR_LOCATION_BAR_MODEL_ANDROID_H_
+
+#include "brave/browser/ui/toolbar/brave_location_bar_model_delegate.h"
+
+#define ChromeLocationBarModelDelegate BraveLocationBarModelDelegate
+
+#include "src/chrome/browser/ui/android/toolbar/location_bar_model_android.h"  // IWYU pragma: export
```
goodov: is this based on an old branch? we don't use `src/` anymore.

patches/components-url_formatter-url_fixer.cc.patch
```diff
@@ -0,0 +1,53 @@
+diff --git a/components/url_formatter/url_fixer.cc b/components/url_formatter/url_fixer.cc
+index 553623e61699757d520699eeeb9405b756872d8f..c4bbd6327970c642532eb02797581babf12839f5 100644
+--- a/components/url_formatter/url_fixer.cc
++++ b/components/url_formatter/url_fixer.cc
+@@ -38,6 +38,7 @@ namespace {
+ 
+ // Hardcode these constants to avoid dependences on //chrome and //content.
+ const char kChromeUIScheme[] = "chrome";
++const char kBraveUIScheme[] = "brave";
```
goodov: I see it's added here, but I don't see it's removed from `content::`. should it be removed?

browser/ui/toolbar/brave_location_bar_model_delegate.cc
```diff
@@ -59,8 +55,9 @@ BraveLocationBarModelDelegate::FormattedStringWithEquivalentMeaning(
 bool BraveLocationBarModelDelegate::GetURL(GURL* url) const {
 #if !BUILDFLAG(IS_ANDROID)
   if (base::FeatureList::IsEnabled(tabs::kBraveSharedPinnedTabs) &&
-      tab_strip_model_->profile()->GetPrefs()->GetBoolean(
-          brave_tabs::kSharedPinnedTab)) {
+      Profile::FromBrowserContext(GetActiveWebContents()->GetBrowserContext())
+          ->GetPrefs()
+          ->GetBoolean(brave_tabs::kSharedPinnedTab)) {
     content::NavigationEntry* entry = GetNavigationEntry();
     if (entry && entry->IsInitialEntry()) {
```
netzenbot, line 63: The new code `Profile::FromBrowserContext(GetActiveWebContents()->GetBrowserContext())` will crash if `GetActiveWebContents()` returns nullptr (e.g., no active tab). The original code used `tab_strip_model_->profile()` which didn't have this issue. Consider null-checking first, similar to line 122 which already does `if (!web_contents)`. [best practice](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-memory.md#CSM-029)

chromium_src/chrome/browser/ui/android/toolbar/location_bar_model_android.h
bridiver: Oops, will fix

patches/components-url_formatter-url_fixer.cc.patch
bridiver: It's in a lot of places just like chrome. I didn't want to add a content dependency here so just hard coded to match chrome

bridiver: oh, you're asking generally, not because of this specific change? I think it's still needed for some cases like converting back to brave for the UI, but I'll double check

bridiver: still used in autocomplete classifier (which may or may not still be needed) and also in `BraveBrowser::UpdateTargetURL` which also may or may not be needed. I'll check those

build.log
simoarpe, line 1: You probably don't want to include the `build.log` here.

rewrite/components/url_formatter/url_fixer.cc.toml
```diff
@@ -0,0 +1,44 @@
+# Copyright (c) 2025 The Brave Authors. All rights reserved.
```
simoarpe:
```suggestion
# Copyright (c) 2026 The Brave Authors. All rights reserved.
```

browser/ui/toolbar/brave_location_bar_model_delegate.cc
```diff
@@ -58,16 +53,19 @@ BraveLocationBarModelDelegate::FormattedStringWithEquivalentMeaning(
 
 bool BraveLocationBarModelDelegate::GetURL(GURL* url) const {
 #if !BUILDFLAG(IS_ANDROID)
+  auto* active_web_contents = GetActiveWebContents();
   if (base::FeatureList::IsEnabled(tabs::kBraveSharedPinnedTabs) &&
-      tab_strip_model_->profile()->GetPrefs()->GetBoolean(
-          brave_tabs::kSharedPinnedTab)) {
+      active_web_contents &&
```
netzenbot, line 58: `Profile::FromBrowserContext(active_web_contents->GetBrowserContext())` can return nullptr, and dereferencing it directly with `->GetPrefs()` will crash if the browser context isn't a Profile. Null-check the result before use, e.g. `if (auto* profile = Profile::FromBrowserContext(...)) { ... }`. [best practice](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-memory.md#CSM-029)

chromium_src/chrome/browser/ui/android/toolbar/location_bar_model_android.h
```diff
@@ -0,0 +1,17 @@
+/* Copyright (c) 2025 The Brave Authors. All rights reserved.
```
netzenbot: nit: Copyright year should be 2026 for new files (currently shows 2025). [best practice](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-012)

patches/components-url_formatter-url_fixer.cc.patch
```diff
@@ -0,0 +1,53 @@
+diff --git a/components/url_formatter/url_fixer.cc b/components/url_formatter/url_fixer.cc
```
netzenbot: This PR adds both `patches/components-url_formatter-url_fixer.cc.patch` (6 hunks) and `rewrite/components/url_formatter/url_fixer.cc.toml` (3 substitutions) for the same file. Is the patch intended to be kept alongside the toml, or is it a leftover? If the toml approach can cover all the changes, consider migrating the remaining patch hunks to additional toml substitutions and removing the patch file entirely. [best practice](https://github.com/brave/brave-core/tree/master/docs/best-practices/chromium-src-overrides.md#CSRC-004)

browser/ui/toolbar/brave_location_bar_model_delegate.cc
bridiver: It only return nullptr if BrowserContext is nullptr, but the BrowserContext for WebContents can never be nullptr  so this is not problem here.

patches/components-url_formatter-url_fixer.cc.patch
fallaciousreasoning: This is incorrect - the patch file is generated from the toml. cc @bbondy

patches/components-url_formatter-url_fixer.cc.patch
bridiver: `UpdateTargetURL` should stay as is since it's about correct display when hovering. Autocomplete classifier should also continue to use `content::kBraveUIScheme`

patches/components-url_formatter-url_fixer.cc.patch
bridiver: Yeah it seems to be confused here, there is always a patch associated with plaster files

patches/components-url_formatter-url_fixer.cc.patch
goodov: should we rename `kBraveUIScheme` to highlight that it shouldn't be used for navigations? The way this var exists is still very tempting to use it instead of the chrome scheme.

`kBraveUISchemeNotForNavigation`

goodov (commented):
please check that session restore with `brave://` pages works correctly.

bridiver:
> please check that session restore with `brave://` pages works correctly.

verified, but I'll add to the QA list anyway

netzenbot (approved):

chromium_src/chrome/browser/ui/android/toolbar/DEPS
```diff
@@ -0,0 +1,3 @@
+include_rules = [
+ "+src/chrome/browser/ui/android/toolbar/location_bar_model_android.h",
```
goodov: `src/` again

goodov (approved):
lgtm overall

patches/content-browser-renderer_host-navigation_controller_impl.cc.patch
```diff
@@ -1,25 +1,12 @@
...
-index ed4e4fe82902a072e99071c918f34b29ab407635..00a7f135a38097b755f1a7e64bbf8cbe1c861bb5 100644
+index ed4e4fe82902a072e99071c918f34b29ab407635..712458113109f3e9e4a3e25939fd07420294d5d0 100644
 --- a/content/browser/renderer_host/navigation_controller_impl.cc
 +++ b/content/browser/renderer_host/navigation_controller_impl.cc
-@@ -4029,16 +4029,17 @@ base::WeakPtr<NavigationHandle> NavigationControllerImpl::NavigateWithoutEntry(
-   // Note: we intentionally leave the pending entry in place for renderer debug
-   // URLs, unlike the cases below where we clear it if the navigation doesn't
-   // proceed.
--  if (blink::IsRendererDebugURL(params.url)) {
-+  const GURL url = params.url.SchemeIs(url::kJavaScriptScheme) ? params.url : pending_entry_->GetURL();
```
bridiver, line 10: I should probably remove this in a separate PR, I'll split it out

patches/content-browser-renderer_host-navigation_controller_impl.cc.patch
```diff
@@ -1,25 +1,12 @@
...
- 
--    HandleRendererDebugURL(node, params.url);
-+    HandleRendererDebugURL(node, url);
-     return nullptr;
-   }
+@@ -452,6 +452,7 @@ void RewriteUrlForNavigation(const GURL& original_url,
+   *url_to_load = *virtual_url = original_url;
+   BrowserURLHandlerImpl::GetInstance()->RewriteURLIfNecessary(
+       url_to_load, browser_context, reverse_on_redirect);
++  MaybeRewriteVirtualURL(url_to_load, virtual_url);
```
bridiver: need to convert this to plaster

chromium_src/chrome/browser/ui/android/toolbar/DEPS
bridiver: this is actually not needed, I think it was only triggering because I had src in the override

yrliou (approved):
wallet-core++

bridiver:
> ### Possible Issues
> 1. **`MaybeRewriteVirtualURL` is overly broad**: In `chromium_src/content/browser/renderer_host/navigation_controller_impl.cc`, the function sets `virtual_url = url_to_load` for _all_ `chrome://` URLs, not just those that were rewritten from `brave://`. This could interfere with Chromium's own virtual URL handling for chrome:// pages that intentionally have different virtual URLs.

Looking into this

> 2. **Test observer rewriting is fragile**: The `test_navigation_observer.cc` override rewrites _any_ `brave://` URL passed as an expected URL to `chrome://`, which modifies the state of the `GURL*` pointer passed in. This could cause unexpected side effects in tests that reuse the URL object after the observer check.

Agreed, but I don't see a good way around this right now other than changing every single test. Maybe in a follow-up

> 3. **`HandleChromeAboutAndChromeSyncRewrite` now returns `true` after rewriting brave→chrome**: The added `return true;` on the brave→chrome rewrite path means it skips the subsequent Chromium `HandleChromeAboutAndChromeSyncRewrite` call. This could skip important logic that Chromium's handler performs for certain URLs.

Not sure this is actually the case after spending time looking at how the return value is used, but I'll double check

> 4. **Browser test `TestWalletPageRoute` was deleted entirely** rather than being updated to test the new behavior. This reduces test coverage for wallet page URL routing on Android.

This test was for a wallet specific workaround and now all android urls work correctly so it's no longer needed

> 5. **Missing `BraveLocationBarModelDelegate` constructor forwarding**: The `BraveLocationBarModelDelegate` now has a default constructor, but `ChromeLocationBarModelDelegate` may require initialization parameters on certain platforms. The Android override in `location_bar_model_android.h` substitutes `ChromeLocationBarModelDelegate` with `BraveLocationBarModelDelegate`, but the construction site may pass arguments that the default constructor doesn't accept.

Not relevant
 
> ### Security Hotspots
> 1. **Extension kill URL check removed** (`chrome-browser-extensions-extension_tab_util.cc.patch` deleted): The patch that added `brave://` scheme checking to `ExtensionTabUtil::IsKillURL` is removed. If `brave://` URLs can still reach this code path before being rewritten, potentially dangerous URLs might not be caught by the kill URL check.
> 2. **WebUI factory no longer checks `kBraveUIScheme`** (`brave_web_ui_controller_factory.cc`): If any code path can reach the factory with an unrewritten `brave://` URL, it would return `nullptr` instead of creating the appropriate WebUI controller, potentially causing navigation failures or unexpected behavior.

These changes are intentional, if anything there should be a CHECK here if the url is ever brave and I might add that instead of outright removing some of these

patches/content-browser-renderer_host-navigation_controller_impl.cc.patch
bridiver: actually this is related to brave://crash so it makes sense to remove it here

patches/third_party-blink-renderer-core-dom-document.cc.patch
```diff
@@ -0,0 +1,28 @@
+diff --git a/third_party/blink/renderer/core/dom/document.cc b/third_party/blink/renderer/core/dom/document.cc
+index dde7e8993cb8e59d0cbc1a15fffaedb8b0a26688..874c34ae74db95ed367524d76bb7ff352e136d81 100644
+--- a/third_party/blink/renderer/core/dom/document.cc
++++ b/third_party/blink/renderer/core/dom/document.cc
+@@ -7561,6 +7561,15 @@ KURL Document::CompleteURL(
+   return CompleteURLWithOverride(url.ToString(), base_url_, preload_status);
+ }
+ 
++KURL MakeChromeURL(const KURL& brave_url) {
```
bridiver: temporary patch for testing

chromium_src/content/public/test/test_navigation_observer.cc
```diff
@@ -0,0 +1,33 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "content/public/browser/browser_url_handler.h"
+#include "content/public/common/url_constants.h"
+#include "url/gurl.h"
+
+namespace {
+
+GURL* RewriteExpectedURLBraveToChrome(GURL* url) {
```
bridiver, line 12: I'm going to add a warning to this because it was only added for backwards compat

bridiver: nvm, this is unavoidable when we want to navigate to `brave://` for testing

brave-browser-releases:
Released in [v1.90.51](https://github.com/brave/brave-browser/releases/tag/v1.90.51)

PR #16466, opened 2022-12-23
fix android deps

bridiver:
Resolves https://github.com/brave/brave-browser/issues/27536

## Submitter Checklist:

- [x] I confirm that no security/privacy review [is needed](https://github.com/brave/brave-browser/wiki/Security-reviews), or that I have [requested](https://github.com/brave/security/issues/new/choose) one
- [x] There is a [ticket](https://github.com/brave/brave-browser/issues) for my issue
- [x] Used Github [auto-closing keywords](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue) in the PR description above
- [x] Wrote a good [PR/commit description](https://google.github.io/eng-practices/review/developer/cl-descriptions.html)
- [x] Squashed any review feedback or "fixup" commits before merge, so that history is a record of what happened in the repo, not your PR
- [ ] Added appropriate labels (`QA/Yes` or `QA/No`; `release-notes/include` or `release-notes/exclude`; `OS/...`) to the associated issue
- [ ] Checked the PR locally:
  * `npm run test -- brave_browser_tests`, `npm run test -- brave_unit_tests` [wiki](https://github.com/brave/brave-browser/wiki/Tests)
  * `npm run lint`, `npm run presubmit` [wiki](https://github.com/brave/brave-browser/wiki/Presubmit-checks), `npm run gn_check`, `npm run tslint`
- [x] Ran `git rebase master` (if needed)

## Reviewer Checklist:

- [ ] A security review [is not needed](https://github.com/brave/brave-browser/wiki/Security-reviews), or a link to one is included in the PR description
- [ ] New files have MPL-2.0 license header
- [ ] Adequate test coverage exists to prevent regressions
- [ ] Major classes, functions and non-trivial code blocks are well-commented
- [ ] Changes in component dependencies are properly reflected in `gn`
- [ ] Code follows the [style guide](https://chromium.googlesource.com/chromium/src/+/HEAD/styleguide/c++/c++.md)
- [ ] Test plan is specified in PR before merging

## After-merge Checklist:

- [ ] The associated issue milestone is set to the smallest version that the
  changes has landed on
- [ ] All relevant documentation has been updated, for instance:
  - [ ] https://github.com/brave/brave-browser/wiki/Deviations-from-Chromium-(features-we-disable-or-remove)
  - [ ] https://github.com/brave/brave-browser/wiki/Proxy-redirected-URLs
[...]

PR #34644, opened 2026-03-11
Upgrade from Chromium 147 to Chromium 148

### LINKED ISSUE #53527: Upgrade from Chromium 147 to Chromium 148
### Description

Upgrade from Chromium 147 to Chromium 148

https://chromium.googlesource.com/chromium/src/+log/147.0.7727.117..148.0.7778.60?pretty=fuller&n=10000

**QA tests:**

**Desktop Affected areas:**
* Verify that bookmark folder works as expected
* Verify that tab search works as expected
* Verify that tabs work as expected (vertical tabs, tab groups, dragging tabs, tab hover card, pinned tabs, etc.)
* Verify that side panel works as expected
* Verify that hamburger menu has expected items
* Verify that split views work as expected
* Verify that default search engine works as expected (including settings page)
* Verify that omnibox autocomplete works as expected
* Verify that Tor is working as expected
* Verify that tests in `Root pinning and HSTS preloading` section pass (https://github.com/brave/brave-browser/wiki/TLS-Policy#root-pinning-and-hsts-preloading)

**iOS Affected areas:**
* Typical iOS testing

**Android Affected areas:**
1. New Tab Page and News - huge code change was here
2. Loyality cards should not be displayed - basically, re-check for https://github.com/brave/brave-browser/issues/52682
3. Re-check Multi-Window, basically this PR https://github.com/brave/brave-core/pull/22862
4. Site-settings should work, have no crash, the order should like before, `Automatically remove permissions` must be at the end
5. Edit of saved passwords should work

### LINKED ISSUE #54829: [CR148] crash on launch when Omaha 4 is enabled
### Description

When testing Chromium 148 PR builds, browser crashed on launch repeatedly when Omaha 4 was enabled.

If updating using Sparkle updater, no crash on launch of C148 builds was observed.

Internal discussion [here](https://bravesoftware.slack.com/archives/CU5UXM3EC/p1776791698773509).

Reproduced by @GeetaSarvadnya @emerick @LaurenWags 

### Steps to reproduce

QA followed upgrade steps similar to those found [here](https://github.com/brave/brave-browser/issues/53944#issuecomment-4134561242) (specifically those found under the "Omaha 4" section).

@emerick found that if you launch `1.91.80` with C148 (PR build) and visit Help -> About Brave, he crashed on relaunch. If Omaha 4 is disabled (ex. by running `--disable-features=BraveUseOmaha4`), then he can relaunch without crashing.

### Actual result

Crashes observed on launch when updating using Omaha 4 in the following cases:
- `1.91.78` with `C147` --> `1.91.84` with `C148`
- `1.91.80` with `C148` --> `1.91.84` with `C148`

In above cases I visited brave://settings/help to get on-demand update, and browser crashed on relaunch repeatedly

callstacks from crash reports were not helpful

Crash Report ID - `d4be0a00-0ddf-950f-0000-000000000000`
Callstack:

<details>

```
[ 00 ] 0x116d5e284 ( Brave Browser Nightly Framework )
[ 01 ] 0x116d5e270 ( Brave Browser Nightly Framework )
[ 02 ] 0x11b36bd04 ( Brave Browser Nightly Framework )
[ 03 ] 0x11b37c624 ( Brave Browser Nightly Framework )
[ 04 ] 0x1169d5214 ( Brave Browser Nightly Framework )
[ 05 ] 0x1169d4e94 ( Brave Browser Nightly Framework )
[ 06 ] 0x1169d4038 ( Brave Browser Nightly Framework )
[ 07 ] 0x1160daca8 ( Brave Browser Nightly Framework )
[ 08 ] 0x11569153c ( Brave Browser Nightly Framework )
[ 09 ] 0x113d157b8 ( Brave Browser Nightly Framework )
[ 10 ] 0x104f1c858 ( Brave Browser Nightly )
[ 11 ] 0x181da2b94 ( dyld )
```

</details>

Crash Report ID - `aecf0a00-0ddf-950f-0000-000000000000`
Callstack:

<details>

```
[ 00 ] 0x11470631c ( Brave Browser Nightly Framework )
[ 01 ] 0x114706308 ( Brave Browser Nightly Framework )
[ 02 ] 0x118d0c94c ( Brave Browser Nightly Framework )
[ 03 ] 0x118d1d29c ( Brave Browser Nightly Framework )
[ 04 ] 0x11437d23c ( Brave Browser Nightly Framework )
[ 05 ] 0x11437cebc ( Brave Browser Nightly Framework )
[ 06 ] 0x11437c060 ( Brave Browser Nightly Framework )
[ 07 ] 0x113a82c98 ( Brave Browser Nightly Framework )
[ 08 ] 0x113039574 ( Brave Browser Nightly Framework )
[ 09 ] 0x1116bd7b8 ( Brave Browser Nightly Framework )
[ 10 ] 0x102834858 ( Brave Browser Nightly )
[ 11 ] 0x1808f6b94 ( dyld )
```

</details>

### Expected result

No crash, as is the case when using Sparkle to update in the above cases

### Reproduces how often

Easily reproduced

C148 PR builds from [here](https://bravesoftware.slack.com/archives/CU5UXM3EC/p1776791698773509)

### Channel information

- [ ] release (stable)
- [ ] beta
- [ ] nightly

### Reproducibility

### Miscellaneous information

cc @mherrmann

cdesouza-chromium:
Resolves https://github.com/brave/brave-browser/issues/53527
Resolves https://github.com/brave/brave-browser/issues/54829


zenparsing (approved):
Rewards 👍

supermassive (approved):
wallet lgtm

kylehickinson (approved):
ios++

components/brave_sync/crypto/crypto_unittest.cc
```diff
@@ -136,13 +136,11 @@ TEST(CryptoTest, Ed25519KeyDerivation) {
   std::vector<uint8_t> info = {0};
   DeriveSigningKeysFromSeed(seed, &HKDF_SALT, &info, &public_key, &private_key);
   EXPECT_EQ("f58ca446f0c33ee7e8e9874466da442b2e764afd77ad46034bdff9e01f9b87d4",
-            base::ToLowerASCII(
-                base::HexEncode(public_key.data(), public_key.size())));
+            base::HexEncodeLower(public_key));
   EXPECT_EQ(
       "b5abda6940984c5153a2ba3653f047f98dfb19e39c3e02f07c8bbb0bd8e8872ef58ca446"
       "f0c33ee7e8e9874466da442b2e764afd77ad46034bdff9e01f9b87d4",
-      base::ToLowerASCII(
```
darkdh, line 144: nit: `base/strings/string_util.h` is no longer needed in this file

darkdh (approved):
sync, wallet, tor ++

components/brave_sync/crypto/crypto_unittest.cc
emerick: Thanks, fixed at https://github.com/brave/brave-core/pull/34644/commits/dc5d289ec37663f28d3e290aca359d2bdf4d1615.

patches/components-search_engines-template_url_starter_pack_data.h.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/components/search_engines/template_url_starter_pack_data.h b/components/search_engines/template_url_starter_pack_data.h
+index a2800658774a8057b4d6b125d205826430e9081d..9ca884cd006465a419a7ae755d10cfdae763db73 100644
+--- a/components/search_engines/template_url_starter_pack_data.h
++++ b/components/search_engines/template_url_starter_pack_data.h
+@@ -34,6 +34,8 @@ enum class StarterPackId {
+   kPage = 5,
+   kAiMode = 6,
+ 
++  // Add an enum value for any Brave-defined search engine starter packs.
```
bridiver: this patch doesn't follow standard patching conventions (comment, etc..)

extensions/BUILD.gn
```diff
@@ -17,6 +17,7 @@ deprecated_visibility_allowed = [
   "//brave/components/psst/browser/core",
   "//brave/components/request_otr/browser",
   "//chrome/browser",
+  "//chrome/browser:browser_public_dependencies",
```
bridiver: this defeats the entire purpose of the visibility list as it effectively allows almost anything in browser to use `//brave/extensions:common`

emerick: cc: @AlexeyBarabash

chromium_src/chrome/browser/notifications/notification_handler_impl.h
```diff
@@ -16,9 +16,10 @@ class NotificationHandler : public NotificationHandler_ChromiumImpl {
     WEB_PERSISTENT = 0,
     WEB_NON_PERSISTENT = 1,
     EXTENSION = 2,
-    SEND_TAB_TO_SELF = 3,
+    // SEND_TAB_TO_SELF = 3,  // Deprecated.
```
bridiver, line 19: why is this stuff here and marked deprecated if it's commented out?

chromium_src/chrome/browser/ui/webui/tab_search/tab_search_page_handler.cc
```diff
@@ -47,11 +47,13 @@ TabSearchPageHandler::TabSearchPageHandler(
...
 #if BUILDFLAG(ENABLE_AI_CHAT)
-  pref_change_registrar_.Add(
+  Profile* profile = Profile::FromWebUI(web_ui_);
+  brave_pref_change_registrar_.Init(profile->GetPrefs());
+  brave_pref_change_registrar_.Add(
       ai_chat::prefs::kBraveAIChatTabOrganizationEnabled,
       base::BindRepeating(
           &TabSearchPageHandler::OnTabOrganizationFeaturePrefChanged,
-          base::Unretained(this), Profile::FromWebUI(web_ui_)));
+          base::Unretained(this), profile));
```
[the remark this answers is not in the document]
bridiver: why is this flagging unchanged code?

patches/chrome-browser-segmentation_platform-android-home_modules_ranking_helper.cc.patch
```diff
@@ -0,0 +1,28 @@
+diff --git a/chrome/browser/segmentation_platform/android/home_modules_ranking_helper.cc b/chrome/browser/segmentation_platform/android/home_modules_ranking_helper.cc
+index da83120c707deb4f32705f43179cfe026186018c..c2fbdbcf4b0c5a8f135ba8369791b42427e45945 100644
+--- a/chrome/browser/segmentation_platform/android/home_modules_ranking_helper.cc
++++ b/chrome/browser/segmentation_platform/android/home_modules_ranking_helper.cc
+@@ -51,6 +51,7 @@ static void JNI_HomeModulesRankingHelper_GetClassificationResult(
+   segmentation_platform::home_modules::HomeModulesCardRegistry* registry =
+       segmentation_platform::SegmentationPlatformServiceFactory::
+           GetHomeModulesCardRegistry(profile);
++  if (!registry) { return; }
```
bridiver, line 9: it's not clear if this is an upstream patch or not?

chromium_presubmit_config.json5
```diff
@@ -176,6 +176,11 @@
     "CheckStableMojomChanges": [
       "components/brave_rewards/core/mojom/.*\\.mojom",
     ],
+    // BraveSigninManager implements the SigninManager interface which defines
```
bridiver: this is just a warning, but did you try using `//  IN-TEST`?

AlexeyBarabash: Yes, `//  IN-TEST` didn't help

patches/third_party-androidx-BUILD.gn.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/third_party/androidx/BUILD.gn b/third_party/androidx/BUILD.gn
+index afe4c234708ca91e3d91b43c8d78916c22d355c8..d90e7e6ff91427e524eaea4c539468c3569a9722 100644
+--- a/third_party/androidx/BUILD.gn
++++ b/third_party/androidx/BUILD.gn
+@@ -750,6 +750,7 @@ if (!limit_android_deps) {
+     info_path = "../androidx/committed/libs/androidx_fragment_fragment/androidx_fragment_fragment.info"
+     enable_bytecode_checks = false
+     deps = [
++      ":androidx_activity_activity_java",
```
bridiver: is this a wip fix or an upstream patch or?

chromium_src/chrome/browser/notifications/notification_handler_impl.h
emerick: I just matched it to how it looks upstream https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/notifications/notification_handler.h;l=30.

The way we override the entire enum seemed weird to me, so I pinged @tmancey about itwhen I made this change and he's going to look into fixing that. Did we have an issue for that work yet @tmancey?

patches/chrome-browser-segmentation_platform-android-home_modules_ranking_helper.cc.patch
emerick: cc: @AlexeyBarabash

patches/third_party-androidx-BUILD.gn.patch
emerick: cc: @AlexeyBarabash

chromium_src/chrome/browser/notifications/notification_handler_impl.h
bridiver: oh yeah, I think I had flagged this duplication before when I came across it here https://github.com/brave/brave-core/pull/16466
This should definitely be fixed properly

tmancey: @aseren can you please pick this up to unblock 148.

extensions/BUILD.gn
AlexeyBarabash: @bridiver , thanks, updated to more close option
https://github.com/brave/brave-core/commit/a3304b2af66123d11d811c65a65337ecadc4d7a5
If looks ok, I'll update the commit message also.

There was a Slack discussion https://bravesoftware.slack.com/archives/CU5UXM3EC/p1774453691869189

bridiver: why does it have to be a public dep?

chromium_presubmit_config.json5
bridiver: I've seen that as well, I think there's a problem with the presubmit check because it seems to flag method definitions as if they are code being called in production. 
```
# We only scan .cc files and the like, as the declaration of
# for-testing functions in header files are hard to distinguish from
# calls to such functions without a proper C++ parser.
```
but clearly they are still flagging method definitions instead of calls to those methods so I think we can just ignore this as a false positive

bridiver: by "ignore" I mean just let presubmit warn about it

patches/base-trace_event-builtin_categories.h.patch
```diff
@@ -1,8 +1,17 @@
...
-index 39168e90fd6ea68e562f0a2c6d8ba1162bc29e71..5c7b2bd8288f24e31e08c091da02aef68e4b8ba1 100644
+index 440349df6c5767fe3f93b51f78b33bf9d3bb5c1a..e8ecc20f5af9110649d3ede401c19cd070931f21 100644
 --- a/base/trace_event/builtin_categories.h
 +++ b/base/trace_event/builtin_categories.h
-@@ -282,6 +282,7 @@ PERFETTO_DEFINE_CATEGORIES_IN_NAMESPACE_WITH_ATTRS(
+@@ -224,7 +224,7 @@ PERFETTO_DEFINE_CATEGORIES_IN_NAMESPACE_WITH_ATTRS(
+       "processes, etc. and their properties.").SetTags("toplevel"),
+     perfetto::Category("persistent_cache"),
+     perfetto::Category("PlatformMalloc"),
+-    perfetto::Category("ppapi"),
```
bridiver: ?

patches/components-search_engines-template_url_starter_pack_data.h.patch
emerick: Fixed here https://github.com/brave/brave-core/pull/34644/commits/43f2a980e970cfeec612954c585855613ba847a4

extensions/BUILD.gn
AlexeyBarabash: I doesn't because it does patching of the target of gn_check failure, I made a mistake.
Fixed at https://github.com/brave/brave-core/commit/743032c0b761e511f38d2d20b6df5af1158cb3df

chromium_presubmit_config.json5
AlexeyBarabash: @bridiver 
There are a lot of places where the upstream uses `ForTesting(); // IN-TEST` at Java
image

Somehow it didn't work for me, I didn't dig deep.

I agree it is false-positive.

Reverted at  https://github.com/brave/brave-core/commit/95cda5bb7bdc5689b814421fe2421de2861be277

patches/base-trace_event-builtin_categories.h.patch
emerick: cc: @cdesouza-chromium

patches/third_party-androidx-BUILD.gn.patch
AlexeyBarabash: @bridiver , I pushed the commit initially as a normal fix, I was sure I need to patch the place where dep was removed, but now after your comment I see it is strange.

 I found a better place to put the dep, updated at e121a0638e30590bba76826daa951be52689381e , thanks

patches/chrome-browser-segmentation_platform-android-home_modules_ranking_helper.cc.patch
AlexeyBarabash: Without this patch Android Brave Browser keeps crashing on a new tab page. Upstream doesn't have such problem. I'll write later what is the difference and why we do need this patch.

browser/brave_browser_process_impl.cc
```diff
@@ -236,6 +236,14 @@ void BraveBrowserProcessImpl::StartTearDown() {
...
 #if BUILDFLAG(ENABLE_BRAVE_AI_CHAT_AGENT_PROFILE)
   ai_chat_agent_profile_manager_.reset();
+#endif
+#if BUILDFLAG(ENABLE_BRAVE_WALLET)
+  // Reset WalletDataFilesInstaller to prevent dangling pointer to
+  // CrxUpdateService. WalletDataFilesInstaller instance is a static
+  // base::NoDestructor<> and makes use of
+  // component_updater::ComponentUpdateService::Observer, so it needs to be
+  // reset before the CrxUpdateService is destroyed.
+  brave_wallet::WalletDataFilesInstaller::GetInstance().Reset();
```
bridiver, line 247: We should create a follow-up issue to fix this correctly. WalletDataFilesInstaller should not be a singleton. Probably it belongs in GlobalFeatures and gets passed to the classes that need it in the constructor. 

Is `Reset` a pattern being used by upstream for this kind of stuff or are they just fixing/removing the problematic singletons?

patches/chrome-browser-segmentation_platform-android-home_modules_ranking_helper.cc.patch
bridiver: if we do need this patch, it does not follow standard practices for patching. That's the reason why I asked it if was an upstream patch.

browser/brave_browser_process_impl.cc
emerick: cc: @mkarolin

mkarolin: Created the follow up issue: https://github.com/brave/brave-browser/issues/54704. `Reset` is not a pattern used upstream (`ResetForTesting` is though).

patches/base-trace_event-builtin_categories.h.patch
emerick: We alerted upstream about this issue here https://issues.chromium.org/issues/498698527. I pushed the following which just bumps the limit without removing any categories https://github.com/brave/brave-core/pull/34644/commits/d68706be0c20b97051493b3b67acc5d8cadd9d27

extensions/BUILD.gn
AlexeyBarabash: After DM updated at https://github.com/brave/brave-core/commit/fd1a0fa4b213a56a096a9a3f21b06cbbf3a36cc9

chromium_src/chrome/browser/resources/side_panel/customize_chrome/app.ts
```diff
@@ -12,6 +12,9 @@ import { CustomizeColorSchemeModeBrowserProxy } from '//resources/cr_components/
 
 export * from './app-chromium.js'
 
+// Suppressing `CrLitElement` linting errors that require the name of the class
+// to match the name of the source, as this is an override.
+// eslint-disable-next-line
 class ClosePanelButtonElement extends CrLitElement {
```
fallaciousreasoning, line 19: I wonder if we should just remove the lint rule that the name of the class should match the name of the source?

Alternatively, I think its possible to disable eslint-rules for a glob (perhaps all of `chromium_src`?

An entry like this in `eslint.config.mjs` should work:

```js
  {
    // Don't check the classname matches the file name in chromium src.
    files: ['chromium_src/**/*.ts'],
    rules: {
      '<rule_name>': 'off',
    },
  },

File: browser/resources/settings/br/page_visibility.ts
Comment (fallaciousreasoning) [inline browser/resources/settings/br/page_visibility.ts:1]: Frankly I don't understand why they've enabled this rule - it seems a bit pointless. Type only imports are just an optimization so that the code doesn't get pulled in if it doesn't need to be (i.e. you only have type only imports).

https://chromium-review.googlesource.com/c/chromium/src/+/7671332/comments/853233ae_26d2b09d

File: browser/resources/settings/brave_clear_browsing_data_dialog/brave_clear_browsing_data_dialog_v2_behavior.ts [hunk 12]
```diff
@@ -16,7 +16,8 @@ import {
   SettingsBraveClearBrowsingDataOnExitPageV2Element
 } from './brave_clear_browsing_data_on_exit_page_v2.js'
 
-import {BrowsingDataType, type UpdateSyncStateEvent} from '../clear_browsing_data_dialog/clear_browsing_data_browser_proxy.js';
+import {BrowsingDataType} from '../clear_browsing_data_dialog/clear_browsing_data_browser_proxy.js';
+import {type UpdateSyncStateEvent} from '../clear_browsing_data_dialog/clear_browsing_data_browser_proxy.js';
```
fallaciousreasoning: nit: If type only imports have to be separate the whole import should be type only:

```suggestion
import type { UpdateSyncStateEvent} from '../clear_browsing_data_dialog/clear_browsing_data_browser_proxy.js';
```

browser/resources/settings/br/page_visibility.ts
```diff
@@ -6,11 +6,12 @@
 import {loadTimeData} from 'chrome://resources/js/load_time_data.js'
 
 import {
-  type PageVisibility,
   pageVisibility as chromiumPageVisibility,
   resetPageVisibilityForTesting
 } from '../page_visibility.js'
 
+import { type PageVisibility } from '../page_visibility.js'
```
fallaciousreasoning: nit:

```suggestion
import type { PageVisibility } from '../page_visibility.js'
```

browser/resources/settings/brave_leo_assistant_page/model_list_section.ts
```diff
@@ -16,12 +16,14 @@ import { BaseMixin } from '../base_mixin.js'
...
-  type BraveLeoAssistantBrowserProxy,
-  type Model,
   OperationResult,
   BraveLeoAssistantBrowserProxyImpl,
   OLLAMA_ENDPOINT
 } from './brave_leo_assistant_browser_proxy.js'
+import {
+  type BraveLeoAssistantBrowserProxy,
+  type Model
+} from './brave_leo_assistant_browser_proxy.js'
```
fallaciousreasoning, line 26: nit:

```suggestion
import type {
  BraveLeoAssistantBrowserProxy,
  Model
} from './brave_leo_assistant_browser_proxy.js'
```

browser/resources/settings/brave_wallet_page/brave_wallet_page.ts
fallaciousreasoning:
```suggestion
import type {
  Option,
  SolanaProvider,
  CardanoProvider
} from './brave_wallet_browser_proxy.js'

File: browser/resources/settings/brave_wallet_page/wallet_networks_list.ts [hunk 15]
```diff
@@ -6,10 +6,8 @@
 import 'chrome://resources/cr_elements/cr_button/cr_button.js'
 import './add_wallet_network_dialog.js'
 
-import {
-  BraveWalletBrowserProxyImpl,
-  type NetworkInfo,
-} from './brave_wallet_browser_proxy.js'
+import {BraveWalletBrowserProxyImpl} from './brave_wallet_browser_proxy.js'
+import {type NetworkInfo} from './brave_wallet_browser_proxy.js'
```
fallaciousreasoning: nit:

```suggestion
import type { NetworkInfo } from './brave_wallet_browser_proxy.js'

File: browser/resources/settings/default_brave_shields_page/brave_adblock_subpage.ts [hunk 16]
```diff
@@ -10,7 +10,9 @@ import './components/brave_adblock_subscribe_dropdown.js'
 import {PrefsMixin} from '/shared/settings/prefs/prefs_mixin.js'
 import {I18nMixin} from 'chrome://resources/cr_elements/i18n_mixin.js'
 import {
-  type DomRepeatEvent,
+  type DomRepeatEvent
+} from 'chrome://resources/polymer/v3_0/polymer/polymer_bundled.min.js'
```
fallaciousreasoning, line 14:
```suggestion
import type {
  DomRepeatEvent
} from 'chrome://resources/polymer/v3_0/polymer/polymer_bundled.min.js'
```

chromium_src/chrome/browser/resources/side_panel/customize_chrome/app.ts
```diff
@@ -5,7 +5,8 @@
 
 import { CustomizeChromeApiProxy } from './customize_chrome_api_proxy.js'
 
-import { CrLitElement, html, css, type CSSResultGroup } from '//resources/lit/v3_0/lit.rollup.js'
+import { CrLitElement, html, css } from '//resources/lit/v3_0/lit.rollup.js'
+import { type CSSResultGroup } from '//resources/lit/v3_0/lit.rollup.js'
```
fallaciousreasoning:
```suggestion
import type { CSSResultGroup } from '//resources/lit/v3_0/lit.rollup.js'
```

chromium_src/ui/webui/resources/cr_elements/cr_button/cr_button.ts
```diff
@@ -5,7 +5,8 @@
 
 import '//resources/brave/leo.bundle.js';
 
-import {CrLitElement, css, type CSSResultGroup, type PropertyValues} from '//resources/lit/v3_0/lit.rollup.js';
+import {CrLitElement, css} from '//resources/lit/v3_0/lit.rollup.js';
+import {type CSSResultGroup, type PropertyValues} from '//resources/lit/v3_0/lit.rollup.js';
```
fallaciousreasoning:
```suggestion
import type {CSSResultGroup, PropertyValues} from '//resources/lit/v3_0/lit.rollup.js';
```

chromium_src/ui/webui/resources/cr_elements/cr_toggle/cr_toggle.ts
```diff
@@ -5,7 +5,8 @@
 
 import '//resources/brave/leo.bundle.js'
 
-import {CrLitElement, css, type CSSResultGroup} from '//resources/lit/v3_0/lit.rollup.js';
+import {CrLitElement, css} from '//resources/lit/v3_0/lit.rollup.js';
+import {type CSSResultGroup} from '//resources/lit/v3_0/lit.rollup.js';
```
fallaciousreasoning:
```suggestion
import type {CSSResultGroup} from '//resources/lit/v3_0/lit.rollup.js';
```

patches/components-webapps-browser-web_app_url_config.cc.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/components/webapps/browser/web_app_url_config.cc b/components/webapps/browser/web_app_url_config.cc
+index 6a94ba48bb02368fe0ac09d8b44050a9b6983324..891734fe39bb05d35d71bf48a82825dfa4017139 100644
+--- a/components/webapps/browser/web_app_url_config.cc
++++ b/components/webapps/browser/web_app_url_config.cc
+@@ -51,6 +51,7 @@ bool IsUrlEligibleForWebApp(const GURL& url) {
+   if (url.SchemeIs(content::kChromeUIScheme) &&
+       (url.GetHost() == password_manager::kChromeUIPasswordManagerHost ||
+        ValidChromeUrlHostsForTesting().contains(url.GetHost()))) {
++    BRAVE_WEB_APP_URL_CONFIG_IS_URL_ELIGIBLE_FOR_WEB_APP
```
fallaciousreasoning: This looks like the patch is wrong - this should either be inside the if condition or its own block outside this if statement.

browser/resources/settings/brave_sync_page/brave_sync_subpage.ts
```diff
@@ -199,29 +199,29 @@ class SettingBraveSyncSubpage extends SettingBraveSyncSubpageBase {
 
   private computeSyncSectionDisabled_(): boolean {
     return this.syncStatus !== undefined &&
-      (!!this.syncStatus.disabled ||
-        !!this.syncStatus.hasSyncWordsDecryptionError ||
-        (!!this.syncStatus.hasError &&
+      ((this.syncStatus.disabled ?? false) ||
+        (this.syncStatus.hasSyncWordsDecryptionError ?? false) ||
+        ((this.syncStatus.hasError ?? false) &&
```
fallaciousreasoning: this seems wrong - these will already be falsey so we shouldn't need the `?? false`

Can we just remove all of this? If not, the best solution is probably to wrap everything in a `!!`

browser/resources/settings/brave_sync_page/brave_sync_subpage.ts
```diff
@@ -199,29 +199,29 @@ class SettingBraveSyncSubpage extends SettingBraveSyncSubpageBase {
...
           this.syncStatus.statusAction !==
           StatusAction.ENTER_PASSPHRASE &&
           this.syncStatus.statusAction !==
           StatusAction.RETRIEVE_TRUSTED_VAULT_KEYS))
   }
 
   private computeSyncDisabledByAdmin_(): boolean {
-    return !!this.syncStatus?.managed
+    return this.syncStatus?.managed ?? false
```
fallaciousreasoning: Hmmm, does this actually fail the rule? It doesn't seem like it should as this is (literally) a necessary conversion

browser/resources/settings/brave_sync_page/brave_sync_subpage.ts
```diff
@@ -199,29 +199,29 @@ class SettingBraveSyncSubpage extends SettingBraveSyncSubpageBase {
...
   }
 
   private computeLockedSafeStorage_(): boolean {
-    return this.syncStatus &&
-           !!this.syncStatus.hasSyncWordsDecryptionError &&
-           !this.syncStatus.isOsEncryptionAvailable
+      return this.syncStatus !== undefined &&
+        (this.syncStatus.hasSyncWordsDecryptionError ?? false) &&
+        !(this.syncStatus.isOsEncryptionAvailable ?? false)
```
fallaciousreasoning: if this compiles then use it:
```suggestion
 this.syncStatus &&
           this.syncStatus.hasSyncWordsDecryptionError &&
           !this.syncStatus.isOsEncryptionAvailable
```

otherwise:
```suggestion
 !!(this.syncStatus &&
           this.syncStatus.hasSyncWordsDecryptionError &&
           !this.syncStatus.isOsEncryptionAvailable)
```

browser/resources/settings/brave_sync_page/brave_sync_subpage.ts
```diff
@@ -199,29 +199,29 @@ class SettingBraveSyncSubpage extends SettingBraveSyncSubpageBase {
...
   }
 
   private computeUnlockedSafeStorageCannotDecryptSeed_(): boolean {
-    return this.syncStatus &&
-      !!this.syncStatus.hasSyncWordsDecryptionError &&
-      !!this.syncStatus.isOsEncryptionAvailable
+    return this.syncStatus !== undefined &&
```
fallaciousreasoning:
```suggestion
   return this.syncStatus &&
      this.syncStatus.hasSyncWordsDecryptionError &&
      this.syncStatus.isOsEncryptionAvailable
```

or 

```suggestion
   return !!(this.syncStatus &&
      this.syncStatus.hasSyncWordsDecryptionError &&
      this.syncStatus.isOsEncryptionAvailable)
```

chromium_src/chrome/browser/resources/certificate_manager/certificate_manager.ts
fallaciousreasoning, line 1: lol, I'm surprised they migrated this given they have `CertificateManagerV2` behind a flag 😆

browser/resources/tab_search/tab_focus/tab_focus_page.ts
```diff
@@ -56,6 +51,13 @@ export class TabFocusPageElement extends CrLitElement {
     }
   }
 
+  protected accessor topics_: string[] = []
+  protected accessor topic: string = ''
+  protected accessor undoTopic_: string = ''
+  protected accessor showFRE_: boolean = loadTimeData.getBoolean(
```
fallaciousreasoning: Is the type not inferred properly here? 

Was the issue that we were missing the `accessor` or the type?

If possible, I'd prefer we don't specify the type on these as its obvious from the rhs

chromium_src/chrome/browser/resources/side_panel/customize_chrome/app.ts
emerick: It seems that the linting occurs on the preprocessed files (see the path in this error message), so that method doesn't seem to help:

```
C:\work\cr148\brave-browser\src\out\Component\gen\chrome\browser\resources\side_panel\customize_chrome\preprocessed\app.ts
  16:1  error  Naming of file/DOM pair app.ts ↔ close-panel-button is inconsistent         @webui-eslint/lit-element-structure
  46:1  error  Naming of file/DOM pair app.ts ↔ brave-darker-theme-toggle is inconsistent  @webui-eslint/lit-element-structure
```

Maybe I'm doing something wrong, though?

fallaciousreasoning: Hmmm, I guess maybe its upstreams linter running over the files too, rather than ours. Okay, let's ignore this for now.

patches/components-webapps-browser-web_app_url_config.cc.patch
emerick: You're right and, in fact, we don't need the patch at all as the override seems to already cover this. Removed at https://github.com/brave/brave-core/pull/34644/commits/2d9468883eafd64983f7189c73b7b0a624e01b58

browser/resources/tab_search/tab_focus/tab_focus_page.ts
emerick: Fixed here https://github.com/brave/brave-core/pull/34644/commits/9784f96f70c5b88f2b7cf610134831c849e41f8c, thanks!

browser/resources/settings/brave_clear_browsing_data_dialog/brave_clear_browsing_data_dialog_v2_behavior.ts
emerick: Fixed here https://github.com/brave/brave-core/pull/34644/commits/20312aa2120af3fc77dcf5a2fd132ace31000610

browser/resources/settings/br/page_visibility.ts
emerick: Fixed here https://github.com/brave/brave-core/pull/34644/commits/20312aa2120af3fc77dcf5a2fd132ace31000610

browser/resources/settings/brave_leo_assistant_page/model_list_section.ts
emerick: Fixed here https://github.com/brave/brave-core/pull/34644/commits/20312aa2120af3fc77dcf5a2fd132ace31000610

browser/resources/settings/brave_wallet_page/brave_wallet_page.ts
emerick: Fixed here https://github.com/brave/brave-core/pull/34644/commits/20312aa2120af3fc77dcf5a2fd132ace31000610

browser/resources/settings/brave_wallet_page/wallet_networks_list.ts
emerick: Fixed here https://github.com/brave/brave-core/pull/34644/commits/20312aa2120af3fc77dcf5a2fd132ace31000610

browser/resources/settings/default_brave_shields_page/brave_adblock_subpage.ts
emerick: Fixed here https://github.com/brave/brave-core/pull/34644/commits/20312aa2120af3fc77dcf5a2fd132ace31000610

chromium_src/chrome/browser/resources/side_panel/customize_chrome/app.ts
emerick: Fixed here https://github.com/brave/brave-core/pull/34644/commits/20312aa2120af3fc77dcf5a2fd132ace31000610

chromium_src/ui/webui/resources/cr_elements/cr_button/cr_button.ts
emerick: Fixed here https://github.com/brave/brave-core/pull/34644/commits/20312aa2120af3fc77dcf5a2fd132ace31000610

chromium_src/ui/webui/resources/cr_elements/cr_toggle/cr_toggle.ts
emerick: Fixed here https://github.com/brave/brave-core/pull/34644/commits/20312aa2120af3fc77dcf5a2fd132ace31000610

browser/resources/settings/brave_sync_page/brave_sync_subpage.ts
emerick: Fixed here https://github.com/brave/brave-core/pull/34644/commits/05cdc2d378bc80199344d37a5c8f31cddb7275b1

browser/resources/settings/brave_sync_page/brave_sync_subpage.ts
emerick: Nope. Fixed here https://github.com/brave/brave-core/pull/34644/commits/05cdc2d378bc80199344d37a5c8f31cddb7275b1

browser/resources/settings/brave_sync_page/brave_sync_subpage.ts
emerick: Fixed here https://github.com/brave/brave-core/pull/34644/commits/05cdc2d378bc80199344d37a5c8f31cddb7275b1

browser/resources/settings/brave_sync_page/brave_sync_subpage.ts
emerick: Fixed here https://github.com/brave/brave-core/pull/34644/commits/05cdc2d378bc80199344d37a5c8f31cddb7275b1

browser/resources/settings/brave_sync_page/brave_sync_subpage.ts
```diff
@@ -199,29 +199,29 @@ class SettingBraveSyncSubpage extends SettingBraveSyncSubpageBase {
 
   private computeSyncSectionDisabled_(): boolean {
     return this.syncStatus !== undefined &&
-      ((this.syncStatus.disabled ?? false) ||
-        (this.syncStatus.hasSyncWordsDecryptionError ?? false) ||
-        ((this.syncStatus.hasError ?? false) &&
-          this.syncStatus.statusAction !==
+      (!!this.syncStatus.disabled ||
+       this.syncStatus.hasSyncWordsDecryptionError ||
+       (!!this.syncStatus.hasError &&
```
fallaciousreasoning, line 204: are these `!!`s still needed?

fallaciousreasoning: if so, let's just add a single `!!` around the whole expression

fallaciousreasoning (approved):
lgtm % mega nit (feel free to resolve the comment)

browser/resources/settings/brave_sync_page/brave_sync_subpage.ts
emerick: It seems so, because those types are defined as optional booleans.

patches/chrome-browser-segmentation_platform-android-home_modules_ranking_helper.cc.patch
AlexeyBarabash: Fixed at https://github.com/brave/brave-core/pull/34644/changes/0e3385f91b6ae2f256c27c835c2afacc59d4bfa2

AlexeyBarabash:
> 1. **`BraveMultiInstanceOrchestratorImpl` coupling**: `setBraveMultiInstanceManager` + deferred `moveTabsToWindowByIdChecked` call is fragile. If `handleMenuOrKeyboardAction` is called without the `move_to_other_window_menu_id` path, `mBraveMultiInstaceManager` stays null, and the `assertNonNull` in `moveTabsToWindowByIdChecked` will crash. The field is also never cleared.

Thanks, you are right, fixed here https://github.com/brave/brave-core/pull/34644/commits/9477df6dd74648f012b4fcb60b7bd3db8fbdf74b

AlexeyBarabash:
> 2\. **`BraveMultiWindowUtils.mergeWindows(mActivity)`**: Comment says "Updated to mActivity to avoid the crash on Merge operation" — but `mActivity` is a field on the parent `MultiInstanceManagerApi31` that could be null if the activity is destroyed. No null guard present.

Here `mActivity` is from `MultiInstanceManagerImpl`, `protected final Activity mActivity;`, it can't be null after the activity is destroyed. 

But I put the guard `!mActivity.isDestroyed() && !mActivity.isFinishing()` to avoid mentioned cases here 
https://github.com/brave/brave-core/pull/34644/commits/9477df6dd74648f012b4fcb60b7bd3db8fbdf74b

AlexeyBarabash:
> 4\. **`BraveNewTabPageLayout.initialize` marked `@Initializer`** but not called from constructor — relies on `BraveNewTabPageCoordinator.initialize()` to call it. If anyone constructs `BraveNewTabPageLayout` without going through the coordinator, `mProfile`/`mNTPBackgroundImagesBridge`/`mWindowAndroid` remain null.

To me it is okay, we are the owners of the `BraveNewTabPageLayout`, it is constructed through `src/brave/android/java/res/layout/new_tab_page_layout.xml`.
And even hypothetically `BraveNewTabPageLayout` will be created proper call of `BraveNewTabPageLayout.initialize` - this will happen on the next major Chromium upgrade, we will see null object access crash at debugger on `mProfile`/`mNTPBackgroundImagesBridge`/`mWindowAndroid` .

AlexeyBarabash:
> 6\. **Typo in `BraveMultiInstanceOrchestratorImpl`**: `mBraveMultiInstaceManager` — missing 'n' in "Instance".

this is a good catch, thanks, fixed at https://github.com/brave/brave-core/pull/34644/changes/1f856fcc7cacf912302323096107ac6e27a2915f

fmarier (approved):
`chromium_src/net/tools/transport_security_state_generator/input_file_parsers.cc` changes :+1:

AlexeyBarabash:
> 7\. **`SyncUserSettingsImpl` rewrite**: The `BraveAdjustPreferredDataTypes` function removes `AUTOFILL_VALUABLE`/`AUTOFILL_VALUABLE_METADATA` but lives in an anonymous namespace in the `.cc` override — the corresponding `toml` rewrite injects a call to it. If the upstream function signature or body structure changes, the regex match will silently fail.

I think, it is ok, a new test `BraveSyncUserSettingsImplTest.PreferredTypesSyncEverything` should catch it.

chromium_src/chrome/browser/bluetooth/chrome_bluetooth_delegate.cc
```diff
@@ -0,0 +1,27 @@
...
+
+#include "chrome/browser/bluetooth/chrome_bluetooth_delegate.h"
+
+#include "base/feature_list.h"
+#include "third_party/blink/public/common/features.h"
+
+// Brave's AllowWebBluetooth checks kBraveWebBluetoothAPI before delegating to
+// the Chromium implementation.
+ChromeBluetoothDelegate::AllowWebBluetoothResult
+ChromeBluetoothDelegate::AllowWebBluetooth(
```
bridiver: seems like we should subclass ChromeBluetoothDelegate and return in BraveContentBrowserClient::GetBluetoothDelegate()

chromium_src/chrome/browser/notifications/notification_handler_impl.h
aseren: as discussed in DM we will fix this issue as a follow-up via https://github.com/brave/brave-browser/issues/54754.
cc @emerick

mkarolin (approved):
`strings`++

chromium_src/chrome/browser/bluetooth/chrome_bluetooth_delegate.cc
emerick: Makes sense, fixed here https://github.com/brave/brave-core/pull/34644/commits/4c22aba8fd5cce47563e6dc210d388337c1ad87b

bridiver (approved):

samartnik (approved):
Android code looks good

components/brave_rewards/content/diagnostic_log.cc
```diff
@@ -180,13 +180,11 @@ bool TrimBeginningOfFile(base::File* file, int keep_num_lines) {
   }
 
   const std::string data(buffer.data(), *n);
-  const int64_t new_size = data.size();
-
-  if (UNSAFE_TODO(file->WriteAtCurrentPos(data.c_str(), new_size)) == -1) {
```
goodov, line 185: technically this passed a null-terminated string, so this is different now, but I don't think it actually matters here.

`const std::string data(buffer.data(), *n);` above should be replaced with a `string_view`. We don't need a copy here.

ios/browser/api/bookmarks/exporter/bookmark_html_writer.cc
```diff
@@ -273,13 +273,14 @@ class Writer : public base::RefCountedThreadSafe<Writer> {
...
-    if (!text.length())
+    if (text.empty()) {
       return true;
-    size_t wrote =
-        UNSAFE_TODO(file_->WriteAtCurrentPos(text.c_str(), text.length()));
-    bool result = (wrote == text.length());
+    }
+    std::optional<size_t> wrote =
+        file_->WriteAtCurrentPos(base::as_bytes(base::span(text)));
+    bool result = wrote.has_value() && (*wrote == text.length());
```
goodov: nit: just `const bool result = wrote == text.length()`. If the optional is empty, this will be false.

components/brave_rewards/core/engine/util/random_util.cc
```diff
@@ -34,7 +34,7 @@ namespace brave_rewards::internal::util {
 std::string GenerateRandomHexString() {
   uint8_t bytes[kLengthHexString];
   crypto::RandBytes(bytes);
-  return base::HexEncode(bytes, sizeof(bytes));
+  return base::HexEncode(base::byte_span_from_ref(bytes));
```
goodov: nit: `base::as_byte_span(bytes)`. `from_ref` is a bit confusing for an array use, I'd say it's main use case for things like:
```
uint64_t var;
base::byte_span_from_ref(var);
```

goodov (approved):

components/brave_rewards/content/diagnostic_log.cc
emerick: Fixed here https://github.com/brave/brave-core/pull/34644/commits/117603684c66d524d8fbc2e447f28f2d282f0a9d.

ios/browser/api/bookmarks/exporter/bookmark_html_writer.cc
emerick: Thanks, fixed here https://github.com/brave/brave-core/pull/34644/commits/96fa3efd10d02c740b7c1f5ce3e9a482ff031d67.

components/brave_rewards/core/engine/util/random_util.cc
emerick: Fixed here https://github.com/brave/brave-core/pull/34644/changes/e2ce1dbcfdeaab8b11d8b3e6d5cee0808932af49.

emerick (approved):

PR #34652, opened 2026-03-11
Cache adblock DAT files to streamline startup

### LINKED ISSUE #27161: Cache adblock DATs clientside for better performance on browser startup
## Test plan
See https://github.com/brave/brave-browser/issues/27161#issuecomment-4315136728

## Description
_Note: The benchmarks below are outdated following the migration of adblock-rust to Flatbuffers, yet the motivation and theory are still valid_

We switched adblock lists to be served as plaintext components rather than serialized DATs in https://github.com/brave/brave-core/pull/15077. Loading adblock lists from plaintext is _slightly_ slower than loading from a serialized DAT, although it leaves us the option of serializing them clientside to the newer uncompressed format which can be deserialized _significantly_ faster. As per https://github.com/brave/brave-core/pull/15077:

> ### Performance analysis
> 
> Benchmarking `adblock-rust` on my own computer shows a minor perf impact related to list parsing at startup, of an additional ~56ms or 32% regression for the entire set of default lists. Note that an additional change can allow adblock-rust to cache compiled DATs in uncompressed format after updates, resulting in a significant improvement to engine load time over the current `master` branch. I've left the caching for a followup to minimize the number of changes in this PR.
> 
> The benchmark results below show the engine load times for the current `master` configuration, the configuration as of this PR, and the configuration with caching, respectively (middle time value is average).
> 
> ```
> from_dat                 time:   [173.66 ms 174.88 ms 176.32 ms]
> from_plaintext           time:   [228.08 ms 230.97 ms 233.52 ms]
> from_uncompressed_dat    time:   [46.191 ms 46.777 ms 47.546 ms]
> ```

Brave should create a cached adblock DAT whenever lists are modified, and load this on startup to reduce load at browser-launch. Also, there is currently no code in place to prevent pages from loading before the adblock engines are ready, so this is important to help ads be caught on startup.

antonok-edm:
Resolves https://github.com/brave/brave-browser/issues/27161

High-level approach:
 - whenever an adblock engine is assembled, serialize it to disk*
 - update `FiltersProvider`s to only trigger engine recompilation when updated filter list text is available (i.e. a download finishes with a new version available)
 - on startup:
   - read any serialized DAT from disk on `AdBlockService` initialization
   - if reading DATs failed for any reason, proceed with initial loading of filter providers immediately

\* DATs are serialized under a new `adblock_cache` directory in the profile dir

### DIFF
--- a/browser/brave_shields/ad_block_service_browsertest.cc
+++ b/browser/brave_shields/ad_block_service_browsertest.cc
@@ -54,6 +54,7 @@
 #include "chrome/browser/content_settings/host_content_settings_map_factory.h"
 #include "chrome/browser/profiles/profile.h"
 #include "chrome/common/chrome_isolated_world_ids.h"
+#include "chrome/common/chrome_switches.h"
 #include "chrome/test/base/chrome_test_utils.h"
 #include "chrome/test/base/platform_browser_test.h"
 #include "components/prefs/pref_change_registrar.h"
@@ -1273,6 +1274,73 @@ IN_PROC_BROWSER_TEST_F(AdBlockServiceTest, NoRemoveparamOnCnameUncloakedUrl) {
       base::BindRepeating(content::FrameHasSourceUrl, redirected_frame_url));
 
   ASSERT_EQ("?test=true", EvalJs(inner_frame, "window.location.search"));
+}
+
+class CachedDATAdBlockServiceTest : public AdBlockServiceTest {
+ public:
+  CachedDATAdBlockServiceTest()
+      : resource_storage_(adblock::new_empty_resource_storage()) {}
+
+ private:
+  rust::Box<adblock::BraveCoreResourceStorage> resource_storage_;
+
+  DATFileDataBuffer CreateAdblockDAT(std::string_view filters) {
+    std::vector<unsigned char> buffer(filters.begin(), filters.end());
+    auto engine = brave_shields::AdBlockEngine(true);
+    engine.Load(false, buffer, *resource_storage_);
+    return engine.Serialize();
+  }
+
+  void SetUpCommandLine(base::CommandLine* command_line) override {
+    InProcessBrowserTest::SetUpCommandLine(command_line);
+
+    base::ScopedAllowBlockingForTesting allow_blocking;
+
+    base::ScopedTempDir user_data_dir;
+    ASSERT_TRUE(user_data_dir.CreateUniqueTempDir());
+
+    const auto profile_path = user_data_dir.Take();
+    const auto adblock_cache_dir =
+        profile_path.AppendASCII("Default").AppendASCII("adblock_cache");
+
+    command_line->AppendSwitchPath(switches::kUserDataDir, profile_path);
+
+    ASSERT_TRUE(base::CreateDirectory(adblock_cache_dir));
+
+    ASSERT_TRUE(base::WriteFile(adblock_cache_dir.AppendASCII("engine0.dat"),
+                                CreateAdblockDAT("scam.png$image\n")));
+    ASSERT_TRUE(base::WriteFile(adblock_cache_dir.AppendASCII("engine1.dat"),
+                                CreateAdblockDAT("cookie_banner.png$image\n")));
+  }
+
+  // Don't run AdBlockServiceTest's override, which installs components
+  void PreRunTestOnMainThread() override {
+    PlatformBrowserTest::PreRunTestOnMainThread();
+  }
+};
+
+IN_PROC_BROWSER_TEST_F(CachedDATAdBlockServiceTest, CachedDATIsUsed) {
+  GURL url = embedded_test_server()->GetURL(kAdBlockTestPage);
+  NavigateToURL(url);
+  content::WebContents* contents = web_contents();
+
+  // sanity check: default component is not loaded
+  ASSERT_EQ(true, EvalJs(contents,
+                         "setExpectations(1, 0, 0, 0);"
+                         "addImage('ad_banner.png')"));
+  EXPECT_EQ(profile()->GetPrefs()->GetUint64(kAdsBlocked), 0ULL);
+
+  // rule from default engine
+  ASSERT_EQ(true, EvalJs(contents,
+                         "setExpectations(1, 1, 0, 0);"
+                         "addImage('scam.png')"));
+  EXPECT_EQ(profile()->GetPrefs()->GetUint64(kAdsBlocked), 1ULL);
+
+  // rule from additional engine
+  ASSERT_EQ(true, EvalJs(contents,
+                         "setExpectations(1, 2, 0, 0);"
+                         "addImage('cookie_banner.png')"));
+  EXPECT_EQ(profile()->GetPrefs()->GetUint64(kAdsBlocked), 2ULL);
 }
 
 class CnameUncloakingFlagDisabledTest : public AdBlockServiceTest {
--- a/browser/brave_shields/ad_block_service_browsertest.cc
+++ b/browser/brave_shields/ad_block_service_browsertest.cc
@@ -276,6 +276,13 @@ void AdBlockServiceTest::UpdateAdBlockResources(const std::string& resources) {
 
 void AdBlockServiceTest::UpdateAdBlockInstanceWithRules(
     const std::string& rules) {
+  // 1ms delay to avoid conflict caching conflicts when reloading rules multiple
+  // times in a row
+  base::RunLoop run_loop;
+  base::SingleThreadTaskRunner::GetCurrentDefault()->PostDelayedTask(
+      FROM_HERE, run_loop.QuitClosure(), base::Milliseconds(1));
+  run_loop.Run();
+
   auto component_path = MakeFileInTempDir("list.txt", rules);
 
   brave_shields::AdBlockService* service =
--- a/components/brave_shields/content/browser/ad_block_engine.cc
+++ b/components/brave_shields/content/browser/ad_block_engine.cc
@@ -11,6 +11,7 @@
 #include <utility>
 #include <vector>
 
+#include "base/containers/to_vector.h"
 #include "base/json/json_reader.h"
 #include "base/logging.h"
 #include "base/metrics/histogram_functions.h"
@@ -369,6 +370,12 @@ void AdBlockEngine::OnDATLoaded(
   }
 
   UpdateAdBlockClient(std::move(client), storage);
+}
+
+DATFileDataBuffer AdBlockEngine::Serialize() {
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+  auto rust_vec = ad_block_client_->serialize();
+  return base::ToVector(rust_vec);
 }
 
 void AdBlockEngine::AddObserverForTest(AdBlockEngine::TestObserver* observer) {
--- a/components/brave_shields/content/browser/ad_block_engine.cc
+++ b/components/brave_shields/content/browser/ad_block_engine.cc
@@ -241,14 +241,14 @@ base::ListValue AdBlockEngine::HiddenClassIdSelectors(
   return list_result;
 }
 
-void AdBlockEngine::Load(bool deserialize,
+bool AdBlockEngine::Load(bool deserialize,
                          const DATFileDataBuffer& dat_buf,
                          const adblock::BraveCoreResourceStorage& storage) {
   DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
   if (deserialize) {
-    OnDATLoaded(dat_buf, storage);
+    return OnDATLoaded(dat_buf, storage);
   } else {
-    OnListSourceLoaded(dat_buf, storage);
+    return OnListSourceLoaded(dat_buf, storage);
   }
 }
 
@@ -310,7 +310,7 @@ void AdBlockEngine::OnFilterSetLoaded(
   UpdateAdBlockClient(std::move(result.value), storage);
 }
 
-void AdBlockEngine::OnListSourceLoaded(
+bool AdBlockEngine::OnListSourceLoaded(
     const DATFileDataBuffer& filters,
     const adblock::BraveCoreResourceStorage& storage) {
   DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
@@ -333,19 +333,20 @@ void AdBlockEngine::OnListSourceLoaded(
   if (result.result_kind != adblock::ResultKind::Success) {
     LOG(ERROR) << "AdBlockEngine::OnListSourceLoaded failed: "
                << result.error_message.c_str();
-    return;
+    return false;
   }
   UpdateAdBlockClient(std::move(result.value), storage);
+  return true;
 }
 
-void AdBlockEngine::OnDATLoaded(
+bool AdBlockEngine::OnDATLoaded(
     const DATFileDataBuffer& dat_buf,
     const adblock::BraveCoreResourceStorage& storage) {
   DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
 
   // An empty buffer will not load successfully.
   if (dat_buf.empty()) {
-    return;
+    return false;
   }
 
   base::ElapsedTimer timer;
@@ -366,10 +367,11 @@ void AdBlockEngine::OnDATLoaded(
 
   if (!result) {
     LOG(ERROR) << "AdBlockEngine::OnDATLoaded deserialize failed";
-    return;
+    return false;
   }
 
   UpdateAdBlockClient(std::move(client), storage);
+  return true;
 }
 
 DATFileDataBuffer AdBlockEngine::Serialize() {
--- a/components/brave_shields/content/browser/ad_block_service.cc
+++ b/components/brave_shields/content/browser/ad_block_service.cc
@@ -106,6 +106,23 @@ void AdBlockService::SourceProviderObserver::OnChanged(bool is_default_engine) {
                                                     std::move(on_loaded_cb));
 }
 
+void AdBlockService::SourceProviderObserver::PreloadCachedDAT(
+    DATFileDataBuffer dat) {
+  task_runner_->PostTaskAndReply(
+      FROM_HERE,
+      base::BindOnce(
+          [](base::WeakPtr<AdBlockEngine> engine, DATFileDataBuffer dat) {
+            if (engine) {
+              // Resources will be loaded when available
+              auto empty_resources = adblock::new_empty_resource_storage();
+              engine->Load(true, std::move(dat), *empty_resources);
+            }
+          },
+          adblock_engine_->AsWeakPtr(), std::move(dat)),
+      base::BindOnce(&AdBlockService::SourceProviderObserver::LoadResources,
+                     weak_factory_.GetWeakPtr()));
+}
+
 void AdBlockService::SourceProviderObserver::OnFilterSetCallbackLoaded(
     base::OnceCallback<void(rust::Box<adblock::FilterSet>*)> cb) {
   task_runner_->PostTaskAndReplyWithResult(
@@ -123,14 +140,18 @@ void AdBlockService::SourceProviderObserver::OnFilterSetCallbackLoaded(
           weak_factory_.GetWeakPtr()));
 }
 
-void AdBlockService::SourceProviderObserver::OnFilterSetCreated(
-    std::unique_ptr<rust::Box<adblock::FilterSet>> filter_set) {
-  TRACE_EVENT("brave.adblock", "OnFilterSetCreated");
-  filter_set_ = std::move(filter_set);
+void AdBlockService::SourceProviderObserver::LoadResources() {
   // multiple AddObserver calls are ignored
   resource_provider_->AddObserver(this);
   resource_provider_->LoadResources(base::BindOnce(
       &SourceProviderObserver::OnResourcesLoaded, weak_factory_.GetWeakPtr()));
+}
+
+void AdBlockService::SourceProviderObserver::OnFilterSetCreated(
+    std::unique_ptr<rust::Box<adblock::FilterSet>> filter_set) {
+  TRACE_EVENT("brave.adblock", "OnFilterSetCreated");
+  filter_set_ = std::move(filter_set);
+  LoadResources();
 }
 
 void AdBlockService::SourceProviderObserver::OnResourcesLoaded(
@@ -434,10 +455,9 @@ void AdBlockService::OnReadCachedDATFiles(
   if (!read_result) {
     ActivateFilterLoading();
   } else {
-    auto empty_resources = adblock::new_empty_resource_storage();
-    default_engine_->Load(true, read_result->first, *empty_resources);
-    additional_filters_engine_->Load(true, read_result->second,
-                                     *empty_resources);
+    default_service_observer_->PreloadCachedDAT(std::move(read_result->first));
+    additional_filters_service_observer_->PreloadCachedDAT(
+        std::move(read_result->second));
   }
 }
 
--- a/components/brave_shields/content/browser/ad_block_service.cc
+++ b/components/brave_shields/content/browser/ad_block_service.cc
@@ -108,19 +108,29 @@ void AdBlockService::SourceProviderObserver::OnChanged(bool is_default_engine) {
 
 void AdBlockService::SourceProviderObserver::PreloadCachedDAT(
     DATFileDataBuffer dat) {
-  task_runner_->PostTaskAndReply(
+  task_runner_->PostTaskAndReplyWithResult(
       FROM_HERE,
       base::BindOnce(
           [](base::WeakPtr<AdBlockEngine> engine, DATFileDataBuffer dat) {
             if (engine) {
               // Resources will be loaded when available
               auto empty_resources = adblock::new_empty_resource_storage();
-              engine->Load(true, std::move(dat), *empty_resources);
+              return engine->Load(true, std::move(dat), *empty_resources);
             }
+            return false;
           },
           adblock_engine_->AsWeakPtr(), std::move(dat)),
-      base::BindOnce(&AdBlockService::SourceProviderObserver::LoadResources,
-                     weak_factory_.GetWeakPtr()));
+      base::BindOnce(
+          &AdBlockService::SourceProviderObserver::OnPreloadCachedDAT,
+          weak_factory_.GetWeakPtr()));
+}
+
+void AdBlockService::SourceProviderObserver::OnPreloadCachedDAT(bool success) {
+  if (success) {
+    LoadResources();
+  } else {
+    OnChanged(adblock_engine_->IsDefaultEngine(), base::Time());
+  }
 }
 
 void AdBlockService::SourceProviderObserver::OnFilterSetCallbackLoaded(
--- a/components/brave_shields/content/browser/ad_block_service.cc
+++ b/components/brave_shields/content/browser/ad_block_service.cc
@@ -44,6 +44,35 @@ namespace brave_shields {
 
 namespace {
 constexpr char kAdblockCacheDir[] = "adblock_cache";
+
+std::optional<std::pair<DATFileDataBuffer, DATFileDataBuffer>>
+ReadCachedDATFiles(base::FilePath cache_dir) {
+  if (!base::DirectoryExists(cache_dir)) {
+    base::CreateDirectory(cache_dir);
+    return std::nullopt;
+  }
+
+  base::FilePath default_engine_dat_file = cache_dir.AppendASCII("engine0.dat");
+  base::FilePath additional_engine_dat_file =
+      cache_dir.AppendASCII("engine1.dat");
+  if (!base::PathExists(default_engine_dat_file) ||
+      !base::PathExists(additional_engine_dat_file)) {
+    return std::nullopt;
+  }
+
+  auto default_engine_dat = base::ReadFileToBytes(default_engine_dat_file);
+  if (!default_engine_dat) {
+    return std::nullopt;
+  }
+  auto additional_engine_dat =
+      base::ReadFileToBytes(additional_engine_dat_file);
+  if (!additional_engine_dat) {
+    return std::nullopt;
+  }
+
+  return std::make_optional(std::make_pair(std::move(*default_engine_dat),
+                                           std::move(*additional_engine_dat)));
+}
 }  // namespace
 
 AdBlockService::SourceProviderObserver::SourceProviderObserver(
@@ -58,7 +87,6 @@ AdBlockService::SourceProviderObserver::SourceProviderObserver(
       cache_dir_(cache_dir),
       task_runner_(owner->GetTaskRunner()) {
   filters_provider_manager_->AddObserver(this);
-  OnChanged(engine_is_default);
 }
 
 AdBlockService::SourceProviderObserver::~SourceProviderObserver() {
@@ -300,6 +328,12 @@ AdBlockService::subscription_service_manager() {
   return subscription_service_manager_.get();
 }
 
+void AdBlockService::ActivateFilterLoading() {
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+  default_service_observer_->OnChanged(true);
+  additional_filters_service_observer_->OnChanged(false);
+}
+
 AdBlockService::AdBlockService(
     PrefService* local_state,
     std::string locale,
@@ -336,6 +370,12 @@ AdBlockService::AdBlockService(
         features::kAdblockOverrideRegexDiscardPolicyDiscardUnusedSec.Get();
     SetupDiscardPolicy(policy);
   }
+
+  base::FilePath cache_dir = profile_dir_.AppendASCII(kAdblockCacheDir);
+  GetTaskRunner()->PostTaskAndReplyWithResult(
+      FROM_HERE, base::BindOnce(&ReadCachedDATFiles, cache_dir),
+      base::BindOnce(&AdBlockService::OnReadCachedDATFiles,
+                     weak_factory_.GetWeakPtr()));
 
   auto default_resource_provider =
       std::make_unique<AdBlockDefaultResourceProvider>(
@@ -374,8 +414,6 @@ AdBlockService::AdBlockService(
             filters_provider_manager_.get());
   }
 
-  base::FilePath cache_dir = profile_dir_.AppendASCII(kAdblockCacheDir);
-  base::CreateDirectory(cache_dir);
   default_service_observer_ =
       std::make_unique<SourceProviderObserver>(this, cache_dir, true);
   additional_filters_service_observer_ =
@@ -387,6 +425,20 @@ AdBlockService::~AdBlockService() {
   // therefore they leak during shutdown.
   ANNOTATE_LEAKING_OBJECT_PTR(default_engine_.get());
   ANNOTATE_LEAKING_OBJECT_PTR(additional_filters_engine_.get());
+}
+
+void AdBlockService::OnReadCachedDATFiles(
+    std::optional<std::pair<DATFileDataBuffer, DATFileDataBuffer>>
+        read_result) {
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+  if (!read_result) {
+    ActivateFilterLoading();
+  } else {
+    auto empty_resources = adblock::new_empty_resource_storage();
+    default_engine_->Load(true, read_result->first, *empty_resources);
+    additional_filters_engine_->Load(true, read_result->second,
+                                     *empty_resources);
+  }
 }
 
 void AdBlockService::EnableTag(const std::string& tag, bool enabled) {
--- a/components/brave_shields/content/browser/ad_block_service.cc
+++ b/components/brave_shields/content/browser/ad_block_service.cc
@@ -78,6 +78,7 @@ ReadCachedDATFiles(base::FilePath cache_dir) {
 AdBlockService::SourceProviderObserver::SourceProviderObserver(
     AdBlockService* owner,
     base::FilePath cache_dir,
+    PrefService* local_state,
     bool engine_is_default)
     : adblock_engine_(engine_is_default
                           ? owner->default_engine_.get()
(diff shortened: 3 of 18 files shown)


components/brave_shields/content/browser/ad_block_engine.cc
```diff
@@ -371,6 +371,12 @@ void AdBlockEngine::OnDATLoaded(
   UpdateAdBlockClient(std::move(client), storage);
 }
 
+DATFileDataBuffer AdBlockEngine::Serialize() {
+  auto rust_vec = ad_block_client_->serialize();
```
bridiver, line 379:
```
rust::Vec<uint8_t> rust_vec = ad_block_client_->serialize();
return base::ToVector(rust_vec);
```

bridiver: actually the way this is being used I think we can avoid a copy here 
```
rust::Vec<uint8_t> rust_vec = ad_block_client_->serialize();
return base::make_span(rust_vec.data(), rust_vec.size());
```

components/brave_shields/core/browser/ad_block_filters_provider.cc
```diff
@@ -49,11 +52,17 @@ void AdBlockFiltersProvider::RemoveObserver(
...
-    observer.OnChanged(is_for_default_engine);
+  if (is_activated_) {
+    for (auto& observer : observers_) {
+      observer.OnChanged(is_for_default_engine);
+    }
   }
 }
 
+void AdBlockFiltersProvider::Activate() {
+  is_activated_ = true;
```
bridiver: I don't understand the purpose of this flag

components/brave_shields/core/browser/ad_block_filters_provider_manager.cc
```diff
@@ -45,6 +45,16 @@ void AdBlockFiltersProviderManager::RemoveProvider(
   NotifyObservers(is_for_default_engine);
 }
 
+void AdBlockFiltersProviderManager::ActivateAll() {
```
bridiver: same as above, what is the reason for "activating" these?

components/brave_shields/content/browser/ad_block_service.cc
```diff
@@ -381,6 +414,33 @@ AdBlockService::~AdBlockService() {
...
+bool AdBlockService::ReadCachedDATFiles() {
+  base::FilePath base = profile_dir_.Append(kAdblockCacheDir);
+  base::FilePath default_engine_dat_file = base.Append("engine0.dat");
+  base::FilePath additional_engine_dat_file = base.Append("engine1.dat");
+  if (!base::PathExists(default_engine_dat_file) ||
+      !base::PathExists(additional_engine_dat_file)) {
+    return false;
+  }
+  const auto default_engine_dat =
+      base::ReadFileToBytes(default_engine_dat_file);
```
bridiver: I'm confused, this appears to be running on the UI thread which is not allowed and should CHECK at runtime

browser/brave_shields/ad_block_service_browsertest.cc
```diff
@@ -214,6 +214,7 @@ void AdBlockServiceTest::AddNewRules(const std::string& rules,
                                      bool first_party_protections) {
   auto source_provider = std::make_unique<brave_shields::TestFiltersProvider>(
       rules, first_party_protections, permission_mask);
+  source_provider->Activate();
```
bridiver: it's not at all clear to me why this is being called here, but we should have a test (ideally a unit test if possible) that writes a cached file and reads it back

components/brave_shields/content/browser/ad_block_service.cc
```diff
@@ -289,6 +299,13 @@ AdBlockService::subscription_service_manager() {
   return subscription_service_manager_.get();
 }
 
+void AdBlockService::ActivateFilterLoading() {
```
bridiver, line 384: I don't understand what's happening here, shouldn't the service just load the dat file into the AdBlockEngine and that's all there is to? I'm not understanding why we need Activate and OnChanged

components/brave_shields/core/browser/ad_block_filters_provider.cc
antonok-edm: Prior to this PR:
- Whenever a new `FiltersProvider` has been fully initialized, it signals its observers that new filters are available
- `AdBlockComponentFiltersProvider` instances sometimes take different lengths of time to be constructed and initialized
- This leads to race conditions at startup where the filters get assembled multiple times as new components are available (although it is eventually consistent)

Defaulting `FiltersProvider`s to a "deactivated" state allows them to initialize in the background, only signaling updates once activated.

This allows us to delay the first `FiltersProvider` updates until after the 15-second delay introduced in this PR, if a cached DAT was available previously. If not, we still activate the providers immediately to load filters as quickly as possible.

components/brave_shields/core/browser/ad_block_filters_provider_manager.cc
antonok-edm: see above

components/brave_shields/content/browser/ad_block_service.cc
antonok-edm: see above as well

components/brave_shields/content/browser/ad_block_service.cc
```diff
@@ -381,6 +414,33 @@ AdBlockService::~AdBlockService() {
...
+  if (!base::PathExists(default_engine_dat_file) ||
+      !base::PathExists(additional_engine_dat_file)) {
+    return false;
+  }
+  const auto default_engine_dat =
+      base::ReadFileToBytes(default_engine_dat_file);
+  if (!default_engine_dat) {
+    return false;
+  }
+  const auto additional_engine_dat =
```
netzenbot: `additional_engine_dat` is populated by reading `default_engine_dat_file` instead of `additional_engine_dat_file` — copy-paste bug. This means the additional engine is always loaded with the default engine's cached data, silently producing wrong filtering behavior.

components/brave_shields/core/browser/ad_block_filters_provider.h
```diff
@@ -62,6 +65,7 @@ class AdBlockFiltersProvider {
   void NotifyObservers(bool is_for_default_engine);
 
  private:
+  bool is_activated_;
```
netzenbot: `bool is_activated_;` has no default initializer in the class definition. The default constructor (`AdBlockFiltersProvider()`) never sets it, and the parameterized constructor only sets it to `true` in one branch — the false branch is skipped entirely. This results in undefined behavior when `IsActivated()` is false or the default constructor is used. Add `= false` at the declaration site.

components/brave_shields/content/browser/ad_block_service.cc
```diff
@@ -381,6 +414,33 @@ AdBlockService::~AdBlockService() {
...
+// cached DATs
+bool AdBlockService::ReadCachedDATFiles() {
+  base::FilePath base = profile_dir_.Append(kAdblockCacheDir);
+  base::FilePath default_engine_dat_file = base.Append("engine0.dat");
+  base::FilePath additional_engine_dat_file = base.Append("engine1.dat");
+  if (!base::PathExists(default_engine_dat_file) ||
+      !base::PathExists(additional_engine_dat_file)) {
+    return false;
+  }
+  const auto default_engine_dat =
```
netzenbot: Copy-paste bug: `additional_engine_dat` reads from `default_engine_dat_file` instead of `additional_engine_dat_file`. Both engines will be initialized with the same (default) engine data on cached startup.

components/brave_shields/content/browser/ad_block_service.cc
```diff
@@ -381,6 +414,33 @@ AdBlockService::~AdBlockService() {
...
 
+// Return true on success, or false if there was a problem reading either of the
+// cached DATs
+bool AdBlockService::ReadCachedDATFiles() {
+  base::FilePath base = profile_dir_.Append(kAdblockCacheDir);
+  base::FilePath default_engine_dat_file = base.Append("engine0.dat");
+  base::FilePath additional_engine_dat_file = base.Append("engine1.dat");
+  if (!base::PathExists(default_engine_dat_file) ||
+      !base::PathExists(additional_engine_dat_file)) {
+    return false;
```
netzenbot: Bug: `additional_engine_dat` is read from `default_engine_dat_file` instead of `additional_engine_dat_file`. The additional filters engine will be initialized with the default engine's cached data, not its own.

components/brave_shields/content/browser/ad_block_service.cc
```diff
@@ -381,6 +414,33 @@ AdBlockService::~AdBlockService() {
...
+  }
+  const auto default_engine_dat =
+      base::ReadFileToBytes(default_engine_dat_file);
+  if (!default_engine_dat) {
+    return false;
```
netzenbot: `additional_engine_dat` is loaded from `default_engine_dat_file` instead of `additional_engine_dat_file`. This means the additional engine always gets the default engine's data, and the actual additional engine cache file is never read.

components/brave_shields/core/browser/ad_block_filters_provider.h
```diff
@@ -62,6 +65,7 @@ class AdBlockFiltersProvider {
   void NotifyObservers(bool is_for_default_engine);
```
netzenbot: `bool is_activated_;` has no default initializer. Per CSA-063, POD-type members must be default-initialized in the header. Any construction path that doesn't explicitly set this field will read garbage, causing unpredictable `NotifyObservers` behavior. Change to `bool is_activated_ = false;`.

components/brave_shields/content/browser/ad_block_service.cc
antonok-edm: yeah, good call. Not sure why it didn't CHECK for me, but I updated it.

bridiver: I think maybe it's a DCHECK and didn't fire because there is no test for loading cached dat file?

components/brave_shields/core/browser/ad_block_filters_provider.cc
bridiver: is this an issue introduced by the cached dat file or is this an attempt to solve the other related issue of doing multiple filter list loads on startup? 

What 15 second delay?

components/brave_shields/content/browser/ad_block_service.cc
```diff
@@ -326,6 +343,12 @@ AdBlockService::AdBlockService(
     SetupDiscardPolicy(policy);
   }
 
+  base::FilePath cache_dir = profile_dir_.Append(kAdblockCacheDir);
+  GetTaskRunner()->PostTaskAndReplyWithResult(
+      FROM_HERE, base::BindOnce(&AdBlockService::ReadCachedDATFiles, cache_dir),
```
bridiver: let's please not add more usages of this class on more than one thread, we still need to fix the ones we already have. This should pass whatever it needs to an anonymous namespace function

components/brave_shields/core/browser/ad_block_filters_provider.cc
bridiver: if we're parsing and loading the dat file on the same sequence used by the normal component list loads then there should be a race between loading the cached dat and loading the filter lists, or is this a problem because it initially tries to load an empty list? Maybe we need to fix that first so it doesn't effectively clear out the newly loaded dat list?

components/brave_shields/content/browser/ad_block_service.cc
bridiver: actually even `CreateDirectory` should trigger the DCHECK so this seems odd, did tests pass?

components/brave_shields/content/browser/ad_block_service.cc
```diff
@@ -381,6 +404,53 @@ AdBlockService::~AdBlockService() {
   ANNOTATE_LEAKING_OBJECT_PTR(additional_filters_engine_.get());
 }
 
+std::optional<std::pair<DATFileDataBuffer, DATFileDataBuffer>>
+AdBlockService::ReadCachedDATFiles(base::FilePath cache_dir) {
+  // Ensure the cache dir exists (likely the first run)
+  base::CreateDirectory(cache_dir);
```
bridiver: should check base::DirectoryExists before creating

components/brave_shields/content/browser/ad_block_engine.cc
antonok-edm: probably not

```
brave/components/brave_shields/content/browser/ad_block_engine.cc:378:10: error: function introduces unsafe buffer manipulation [-Werror,-Wunsafe-buffer-usage]
  378 |   return base::span(rust_vec.data(), rust_vec.size());
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../../brave/components/brave_shields/content/browser/ad_block_engine.cc:378:10: note: See //docs/unsafe_buffers.md for help.
```

(`base::make_span` doesn't appear to be a thing)

components/brave_shields/core/browser/ad_block_filters_provider.cc
antonok-edm: > What 15 second delay?

I explained that part in the top-level PR description

> is this an issue introduced by the cached dat file or is this an attempt to solve the other related issue of doing multiple filter list loads on startup?

sort of both... we need a way to make sure the cached DAT gets loaded first and not immediately overridden by engines created using the first component, then first 2 components, then first 3 components, etc.

That racing already existed before this change, but we didn't have any _better_ alternative source to load filters from until now.

> if we're parsing and loading the dat file on the same sequence used by the normal component list loads then there should be a race between loading the cached dat and loading the filter lists, or is this a problem because it initially tries to load an empty list? Maybe we need to fix that first so it doesn't effectively clear out the newly loaded dat list?

that's exactly why the delay exists :smile:

components/brave_shields/content/browser/ad_block_service.cc
antonok-edm: tests did pass... anyways, I moved both to a `PostTaskAndReplyWithResult`

components/brave_shields/content/browser/ad_block_engine.cc
bridiver: oops, not quite right, but actually we need to copy anyway so the first one

components/brave_shields/content/browser/ad_block_service.cc
antonok-edm: done

components/brave_shields/content/browser/ad_block_service.cc
antonok-edm: done

components/brave_shields/core/browser/ad_block_filters_provider.cc
bridiver: a fixed delay is not a good workaround here, that could be too long or too short. We definitely don't want it to be completely wiped out by an empty list so we should fix that, but not as part of this PR. It should either be a follow-up or merge before this PR. The other issue should again be separate and a 15 second delay is not what we want here.

antonok-edm: > a 15 second delay is not what we want here.

what should we do instead?

bridiver: It's not important right now because it shouldn't be part of this PR and doesn't need to happen as a prerequisite for the OOP filter list loading

bridiver: the only thing we do need to fix is to short circuit instead of trying to load an empty list and that should be a trivial fix

components/brave_shields/content/browser/ad_block_service.cc
```diff
@@ -381,6 +433,26 @@ AdBlockService::~AdBlockService() {
   ANNOTATE_LEAKING_OBJECT_PTR(additional_filters_engine_.get());
 }
 
+void AdBlockService::OnReadCachedDATFiles(
+    std::optional<std::pair<DATFileDataBuffer, DATFileDataBuffer>>
+        read_result) {
+  if (!read_result) {
+    ActivateFilterLoading();
+  } else {
+    auto empty_resources = adblock::new_empty_resource_storage();
```
atuchin-m: why we do this with empty resources? Looks like it results in missing injections of some scripts.

components/brave_shields/core/browser/ad_block_filters_provider.cc
atuchin-m: > engines created using the first component, then first 2 components, then first 3 components, etc.

I would say we need to fix these fundamental issues first and only then adds a .dat caching layer, not vice versa.

components/brave_shields/content/browser/ad_block_service.cc
```diff
@@ -381,6 +433,26 @@ AdBlockService::~AdBlockService() {
...
+    default_engine_->Load(true, read_result->first, *empty_resources);
+    additional_filters_engine_->Load(true, read_result->second,
+                                     *empty_resources);
+    // Permit a short delay since cached filters have already been loaded from
+    // the previous launch
+    base::SequencedTaskRunner::GetCurrentDefault()->PostDelayedTask(
+        FROM_HERE,
+        base::BindOnce(&AdBlockService::ActivateFilterLoading,
```
atuchin-m: @antonok-edm could you clarify, do we always rebuild the .dat files after these delay?
If we can skip it, how we memorize which components and versions were used (to be sure we get the same lists)?

netzenbot (approved):

browser/brave_shields/ad_block_service_browsertest.cc
antonok-edm: ran into some issues setting up a unit test, but I was able to add a browsertest for this

bridiver: what problems specifically? You can dm in slack

components/brave_shields/core/browser/ad_block_filters_provider.cc
bridiver: @atuchin-m I think actually what's missing here is to not try to load the non-cached version at all if we have a cached version and then the rest is a non-issue right now. @antonok-edm maybe track the component version in local state and short circuit the whole thing if we've already loaded the cached version. Then the other issues definitely shouldn't block this.

bridiver: so here's my proposal:
`SourceProviderObserver::OnChanged` also contains the timestamp of the relevant component. Each time we cache the dat file, we write that timestamp to local state. On the next load we no-op in any calls to `SourceProviderObserver::OnChanged` where the timestamp <= the timestamp in local state (which means it's already included in the dat cache. If any single component has bumped, it will only run once to update the dat file again.

components/brave_shields/content/browser/ad_block_service.cc
bridiver: https://github.com/brave/brave-core/pull/34652/files#r2928049095

components/brave_shields/core/browser/ad_block_filters_provider.cc
antonok-edm: done - `AdBlockComponentServiceManager` stores the last-installed version for each component; `AdBlockFiltersProvider::NotifyObservers` is only called when an updated filter list is available (i.e. it is not triggered when re-loading older rules)

components/brave_shields/core/browser/ad_block_filters_provider_manager.cc
antonok-edm: removed all the "activating" logic in favor of the version-comparing approach

components/brave_shields/core/browser/ad_block_component_service_manager.h
```diff
@@ -78,6 +79,22 @@ class AdBlockComponentServiceManager
 
   void RecordP3ACookieListEnabled();
 
+  typedef struct InstalledComponentRecord {
```
boocmp: `typedef` is redundant.

components/brave_shields/core/browser/ad_block_component_service_manager.h
```diff
@@ -78,6 +79,22 @@ class AdBlockComponentServiceManager
...
+    // If the user has explicitly toggled the component, their selection is
+    // recorded here. `nullopt` may be overridden by Griffin policy and/or
+    // default list status.
+    std::optional<bool> enabled_;
```
boocmp: for public members underscore is redundant.

components/brave_shields/core/browser/ad_block_component_service_manager.cc
```diff
@@ -71,8 +71,48 @@ bool IsAdBlockOnlyModeSupportedAndFeatureEnabled(const std::string& locale) {
          brave_shields::IsAdblockOnlyModeSupportedForLocale(locale);
 }
 
+const base::NoDestructor<std::vector<uint32_t>> kVersionZero({0});
```
boocmp: is it better to use `base::NoDestructor<base::Version>`?

components/brave_shields/content/browser/ad_block_service.cc
```diff
@@ -41,17 +42,52 @@
 
 namespace brave_shields {
 
+namespace {
+constexpr char kAdblockCacheDir[] = "adblock_cache";
+
+std::optional<std::pair<DATFileDataBuffer, DATFileDataBuffer>>
+ReadCachedDATFiles(base::FilePath cache_dir) {
+  if (!base::DirectoryExists(cache_dir)) {
```
boocmp, line 50: nit: CreateDirectory checks for existence and will success if it is already exists.

components/brave_shields/core/browser/ad_block_component_service_manager.cc
```diff
@@ -133,6 +173,47 @@ bool AdBlockComponentServiceManager::NeedsLocaleListsMigration(
...
+  return record;
+}
+
+void AdBlockComponentServiceManager::RecordNewComponentVersion(
+    const std::string& uuid,
+    base::Version version) {
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+  DCHECK(!uuid.empty());
+  DCHECK(local_state_);
+  {
```
boocmp: This additional block looks redundant.

browser/brave_shields/ad_block_service_browsertest.cc
```diff
@@ -270,7 +271,8 @@ void AdBlockServiceTest::UpdateAdBlockResources(const std::string& resources) {
   brave_shields::AdBlockService* service =
       g_brave_browser_process->ad_block_service();
 
-  service->default_resource_provider()->OnComponentReady(component_path);
+  service->default_resource_provider()->OnComponentReady(
+      component_path, base::Version(std::vector<uint32_t>({1})));
```
boocmp: `base::Version({1})` and below

browser/brave_shields/ad_block_service_browsertest.cc
```diff
@@ -1275,6 +1279,73 @@ IN_PROC_BROWSER_TEST_F(AdBlockServiceTest, NoRemoveparamOnCnameUncloakedUrl) {
...
 
+class CachedDATAdBlockServiceTest : public AdBlockServiceTest {
+ public:
+  CachedDATAdBlockServiceTest()
+      : resource_storage_(adblock::new_empty_resource_storage()) {}
+
+ private:
+  rust::Box<adblock::BraveCoreResourceStorage> resource_storage_;
+
+  DATFileDataBuffer CreateAdblockDAT(std::string filters) {
```
boocmp: `std::string_view`?

browser/brave_shields/ad_block_service_browsertest.cc
antonok-edm: `base::Version({1})` is ambiguous with the `std::string` constructor, unfortunately

components/brave_shields/content/browser/ad_block_service.cc
antonok-edm: ok, you and @bridiver can sort this one out :joy: 

https://github.com/brave/brave-core/pull/34652#discussion_r2921683726

components/brave_shields/core/browser/ad_block_component_service_manager.h
antonok-edm: huh, TIL C++ does it automatically. Maybe we should have a lint?

components/brave_shields/core/browser/ad_block_component_filters_provider.cc
```diff
@@ -104,14 +112,18 @@ void AdBlockComponentFiltersProvider::UnregisterComponent() {
...
   TRACE_EVENT(
       "brave.adblock", "AdBlockComponentFiltersProvider::OnComponentReady",
       perfetto::TerminatingFlow::FromPointer(this), "path", path.value());
   base::FilePath old_path = component_path_;
   component_path_ = path;
 
-  NotifyObservers(engine_is_default_);
+  if (version > last_known_version_) {
+    NotifyObservers(engine_is_default_);
+    on_new_version_.Run(version);
```
atuchin-m: Looks like we'd better use `NotifyObservers(engine_is_default_, version)`.
`on_new_version_` looks excessive considering we have `AdBlockFiltersProvider::Observer` interface.

components/brave_shields/content/browser/ad_block_service.cc
```diff
@@ -41,17 +42,52 @@
...
+    base::CreateDirectory(cache_dir);
+    return std::nullopt;
+  }
+
+  base::FilePath default_engine_dat_file = cache_dir.AppendASCII("engine0.dat");
+  base::FilePath additional_engine_dat_file =
```
atuchin-m, line 56: I'm not following why we duplicate the filenames here. We already have 
```
    std::string filename = base::StrCat(
        {"engine", adblock_engine_->IsDefaultEngine() ? "0" : "1", ".dat"});
```

components/brave_shields/content/browser/ad_block_service.cc
```diff
@@ -41,17 +42,52 @@
 
 namespace brave_shields {
 
+namespace {
+constexpr char kAdblockCacheDir[] = "adblock_cache";
+
+std::optional<std::pair<DATFileDataBuffer, DATFileDataBuffer>>
```
atuchin-m, line 48: Should we do it independently for the both engines?
Maybe this should be a part of the engine instead of AdblockService.

components/brave_shields/content/browser/ad_block_service.cc
```diff
@@ -41,17 +42,52 @@
...
+    return std::nullopt;
+  }
+  const auto additional_engine_dat =
+      base::ReadFileToBytes(additional_engine_dat_file);
+  if (!additional_engine_dat) {
+    return std::nullopt;
+  }
+
+  return std::make_optional(
+      std::make_pair(*default_engine_dat, *additional_engine_dat));
```
atuchin-m: Let's avoid copying the content. 
Here we need `std::move(*default_engine_dat)` + remove `const` for `default_engine_dat`

components/brave_shields/content/browser/ad_block_service.cc
```diff
@@ -111,16 +147,20 @@ void AdBlockService::SourceProviderObserver::OnResourcesLoaded(
...
+        {"engine", adblock_engine_->IsDefaultEngine() ? "0" : "1", ".dat"});
+    auto cache_file = cache_dir_.AppendASCII(filename);
     auto engine_load_callback = base::BindOnce(
         [](base::WeakPtr<AdBlockEngine> engine,
            std::unique_ptr<rust::Box<adblock::FilterSet>> filter_set,
-           AdblockResourceStorageBox storage) {
+           AdblockResourceStorageBox storage, base::FilePath cache_file) {
           if (engine) {
             engine->Load(std::move(*filter_set.get()), *storage);
+            base::WriteFile(cache_file, engine->Serialize());
```
atuchin-m, line 209: Does it make sense to have some delay/debounce logic?
Especially when we are not sure about how many times we rebuild the engine.

components/brave_shields/core/browser/ad_block_component_service_manager.cc
```diff
@@ -71,8 +71,48 @@ bool IsAdBlockOnlyModeSupportedAndFeatureEnabled(const std::string& locale) {
...
+  if (version) {
+    version_ = base::Version(*version);
+  }
+
+  if (!version_.IsValid()) {
+    version_ = base::Version(*kVersionZero);
+  }
+
+  std::optional<bool> enabled_;
+  if (enabled) {
```
atuchin-m: `enabled.has_value()`?
The whole block looks like it just does `enabled_ = enable`

components/brave_shields/content/browser/ad_block_service.cc
```diff
@@ -381,6 +433,20 @@ AdBlockService::~AdBlockService() {
   ANNOTATE_LEAKING_OBJECT_PTR(additional_filters_engine_.get());
 }
 
+void AdBlockService::OnReadCachedDATFiles(
+    std::optional<std::pair<DATFileDataBuffer, DATFileDataBuffer>>
+        read_result) {
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+  if (!read_result) {
+    ActivateFilterLoading();
```
atuchin-m, line 489: What is the expected order of `OnReadCachedDATFiles()` and `OnComponentReady()` events?
AFAIK, `OnComponentReady()` is posted from another task runner.
If there is no particular order, I'm not following how `ActivateFilterLoading()` works.

components/brave_shields/core/browser/ad_block_component_filters_provider.cc
```diff
@@ -64,7 +66,9 @@ AdBlockComponentFiltersProvider::AdBlockComponentFiltersProvider(
     : AdBlockFiltersProvider(is_default_engine, manager),
       component_id_(component_id),
       permission_mask_(permission_mask),
-      component_updater_service_(cus) {
+      component_updater_service_(cus),
+      last_known_version_(last_known_version),
```
atuchin-m: we set `last_known_version_` without checking the matching .dat file is actual good to load.
In case it isn't `version > last_known_version_` check later suppresses loading the content, doesn't it?

components/brave_shields/core/browser/ad_block_filters_provider.cc
atuchin-m: Storing the version looks like a possible solution, but we need to take into account:
* Storing the version doesn't imply the list is available in `.dat` file. It could be invalid or format could be updated. An invalid or non-existing .dat file implies invalidation of all stored version for the matching engine (better to do it per-engine basis). 
* We have a dynamic logic registering/enabling the components based on base::Features and locales. We should expect .dat file has outdated set of the component.
* AFAIK there is no particular order between `OnComponentReady()` and `OnReadCachedDATFiles()`, we should ensure it works for any order.

ShivanKaul:
@atuchin-m this PR is holding up https://github.com/brave/brave-core/pull/31499 and it doesn't look like the review is converging. What is critical to be resolved? How can we move faster on this?

components/brave_shields/content/browser/ad_block_service.cc
antonok-edm: `OnReadCachedDATFiles` will run as soon as the browser is started; `OnComponentReady` events only occur after there's a component downloaded from the network with a newer version than what is available locally

components/brave_shields/content/browser/ad_block_service.cc
bridiver: This PR has some implementation concerns that I'm discussing in DM right now, but it does not and should not attempt to solve all problems. I think any kind of delay/debounce is out of scope for this PR

components/brave_shields/content/browser/ad_block_service.cc
bridiver: I'm also not clear on this. If there's no cached data file why would we do anything here?

components/brave_shields/core/browser/ad_block_component_filters_provider.cc
bridiver: There was a bit of confusion because I had actually suggested using timestamps and just passing them to the service which eliminates these types of issues. Discussing with @antonok-edm right now

components/brave_shields/content/browser/ad_block_service.cc
atuchin-m: I'm agree with we shouldn't solve all the problems.
The aim is to not make the current problems worse.
After this change we can start write a lot more that usual.

components/brave_shields/content/browser/ad_block_service.cc
atuchin-m: > soon as the browser is started;

In fact we wait until we read a file (with many other IO on startup).
 
`OnComponentReady` is also called on startup when the component is ready (right after it's registered, no network request is needed, [a screenshot](https://github.com/user-attachments/assets/3bda5288-4943-4fc6-b2ef-001e58123aa8))
Also, we trigger the update as soon as the browser is started. So I'm not following why we always have a specific order here.

atuchin-m:
> @atuchin-m this PR is holding up https://github.com/brave/brave-core/pull/31499 and it doesn't look like the review is converging. What is critical to be resolved? How can we move faster on this?

@ShivanKaul we can't merge it as-is, it's likely results in using an outdated rules for some users and performance issues. 

We need to ensure:
1. it works for all cases (any possible order of updates or event);
2. we have tests covering important cases, not just smoke `CachedDATIsUsed`;
3. there are no new performance issues (excessive IO, buffer copying).

We also discussed some more points before https://github.com/brave/brave-core/pull/31499.
At least fixing multiple rebuilds a ruleset, including  the startup empty list case.

components/brave_shields/core/browser/ad_block_component_filters_provider.cc
atuchin-m: WDYT about storing {id, version} of all component we use to construct a particular .dat file?
That way we can make a right decision whether we need to rebuild it from the components or not.
I.e. we should skip only if we get the same set.

netzenbot (approved):

components/brave_shields/core/browser/ad_block_filters_provider.cc
antonok-edm: updated this to be timestamp-based instead of version-based

(the invalid/missing `.dat` case is handled too)

components/brave_shields/content/browser/ad_block_service.cc
antonok-edm: the previous code would load the resources later, but there's now an explicit call to load the resources here

components/brave_shields/core/browser/ad_block_component_filters_provider.cc
antonok-edm: version-related stuff is gone now

components/brave_shields/content/browser/ad_block_service.cc
antonok-edm: what I meant was `OnComponentReady` with a _new version_ will only occur after after a network request, the whole point of the version/timestamp comparison is to skip the no-network case

atuchin-m: I asked about the startup. From my perspective, the order of `OnReadCachedDATFiles()` and `OnComponentReady()` calls was undefined.

browser/brave_shields/ad_block_service_browsertest.cc
```diff
@@ -275,6 +276,13 @@ void AdBlockServiceTest::UpdateAdBlockResources(const std::string& resources) {
 
 void AdBlockServiceTest::UpdateAdBlockInstanceWithRules(
     const std::string& rules) {
+  // 1ms delay to avoid conflict caching conflicts when reloading rules multiple
+  // times in a row
+  base::RunLoop run_loop;
```
bridiver, line 281: `task_environment_->FastForwardBy`

components/brave_shields/content/browser/ad_block_service.cc
bridiver: So the only conflict here that we would care about is if somehow reading the cached dat files happened after a new component version was loaded (which could possibly happen if the component was downloaded, but not cached on the previous session). In that case I think we just need a simple guard to not attempt to load the cached files if a component driven update has already loaded

components/brave_shields/content/browser/ad_block_service.cc
bridiver: not sure what you mean about writing a lot more than usual. With this change as implemented now we should be doing a lot less work

atuchin-m: Currently we building a ruleset multiple times on the startup in some cases: https://github.com/brave/brave-browser/issues/50361
Before this change we don't write anything on disk.
I assume this change make it worse.

browser/brave_shields/ad_block_service_browsertest.cc
```diff
@@ -1275,6 +1283,73 @@ IN_PROC_BROWSER_TEST_F(AdBlockServiceTest, NoRemoveparamOnCnameUncloakedUrl) {
...
+
+    command_line->AppendSwitchPath(switches::kUserDataDir, profile_path);
+
+    ASSERT_TRUE(base::CreateDirectory(adblock_cache_dir));
+
+    ASSERT_TRUE(base::WriteFile(adblock_cache_dir.AppendASCII("engine0.dat"),
+                                CreateAdblockDAT("scam.png$image\n")));
+    ASSERT_TRUE(base::WriteFile(adblock_cache_dir.AppendASCII("engine1.dat"),
+                                CreateAdblockDAT("cookie_banner.png$image\n")));
+  }
```
atuchin-m, line 1321: Hm, where are the timestamps?
I would say we'd better to use  a pre-test to also verify storing logic.

components/brave_shields/content/browser/ad_block_service.cc
bridiver: I don't think this is needed, lists should load as normal while it's reading the cached dat file. This is only additive, but if for some reason it does load new lists first, we shouldn't apply the cached dat file

browser/brave_shields/ad_block_service_browsertest.cc
```diff
@@ -1275,6 +1283,73 @@ IN_PROC_BROWSER_TEST_F(AdBlockServiceTest, NoRemoveparamOnCnameUncloakedUrl) {
   ASSERT_EQ("?test=true", EvalJs(inner_frame, "window.location.search"));
 }
 
+class CachedDATAdBlockServiceTest : public AdBlockServiceTest {
```
bridiver, line 1286: this should be a unit test, I'm working on that now.

bridiver:
moved to https://github.com/brave/brave-core/pull/35069

PR #34727, opened 2026-03-14
[AI Chat] Add id to artifact to allow overwriting existing artifacts

### LINKED ISSUE #53633: [AI Chat] Allow overwriting of artifacts
Allow models to overwrite artifacts generated in the same conversation turn, for the case where the model finds information later on in the conversation turn (i.e. a search was performed after chart generation). Generate & output an ID when an artifact is generated. Add an `existingArtifactID` parameter to the `createLineChart` function to provide a mechanism for overwriting.

DJAndries:
Resolves https://github.com/brave/brave-browser/issues/53633

## Test plan

Prompt Sonnet to "create a chart for mortgage rates over the past 10 years". Ensure only one chart is created.

### DIFF
--- a/browser/ai_chat/code_execution_tool_browsertest.cc
+++ b/browser/ai_chat/code_execution_tool_browsertest.cc
@@ -281,7 +281,8 @@ IN_PROC_BROWSER_TEST_F(AIChatCodeExecutionToolBrowserTest, PluginReturnsValue) {
 
   EXPECT_EQ(output, "done");
   ASSERT_EQ(artifacts.size(), 1u);
-  EXPECT_FALSE(artifacts[0]->id.empty());
+  ASSERT_TRUE(artifacts[0]->id.has_value());
+  EXPECT_FALSE(artifacts[0]->id->empty());
   EXPECT_EQ(artifacts[0]->type, "mock");
   EXPECT_EQ(artifacts[0]->content_json, "\"mock_value\"");
 }
@@ -324,8 +325,8 @@ IN_PROC_BROWSER_TEST_F(AIChatCodeExecutionToolBrowserTest, CreateLineChart) {
   ASSERT_EQ(artifacts.size(), 1u);
 
   const auto& artifact = artifacts[0];
-  EXPECT_FALSE(artifact->id.empty());
-  EXPECT_THAT(output, HasSubstr("Chart created with ID: " + artifact->id));
+  ASSERT_TRUE(artifact->id.has_value());
+  EXPECT_THAT(output, HasSubstr("Chart created with ID: " + *artifact->id));
   EXPECT_EQ(artifact->type, mojom::kLineChartArtifactType);
   EXPECT_THAT(artifact->content_json, base::test::IsJson(R"json({
                 "data": [
--- a/browser/ai_chat/code_execution_tool_browsertest.cc
+++ b/browser/ai_chat/code_execution_tool_browsertest.cc
@@ -281,6 +281,7 @@ IN_PROC_BROWSER_TEST_F(AIChatCodeExecutionToolBrowserTest, PluginReturnsValue) {
 
   EXPECT_EQ(output, "done");
   ASSERT_EQ(artifacts.size(), 1u);
+  EXPECT_FALSE(artifacts[0]->id.empty());
   EXPECT_EQ(artifacts[0]->type, "mock");
   EXPECT_EQ(artifacts[0]->content_json, "\"mock_value\"");
 }
@@ -320,10 +321,11 @@ IN_PROC_BROWSER_TEST_F(AIChatCodeExecutionToolBrowserTest, CreateLineChart) {
   std::vector<mojom::ToolArtifactPtr> artifacts;
   ExecuteCode(script, &output, &artifacts);
 
-  EXPECT_EQ(output, "Chart created successfully");
   ASSERT_EQ(artifacts.size(), 1u);
 
   const auto& artifact = artifacts[0];
+  EXPECT_FALSE(artifact->id.empty());
+  EXPECT_THAT(output, HasSubstr("Chart created with ID: " + artifact->id));
   EXPECT_EQ(artifact->type, mojom::kLineChartArtifactType);
   EXPECT_THAT(artifact->content_json, base::test::IsJson(R"json({
                 "data": [
--- a/components/ai_chat/core/browser/tools/chart_code_plugin.cc
+++ b/components/ai_chat/core/browser/tools/chart_code_plugin.cc
@@ -30,9 +30,12 @@ bool ChartCodePlugin::IsEnabled() {
 }
 
 std::string_view ChartCodePlugin::Description() const {
-  return "Use chartUtil.createLineChart(data, labels) where data is an array "
-         "of objects and labels is an optional map of data keys to display "
-         "labels. You must use 'x' as the key for the x-axis. "
+  return "Use chartUtil.createLineChart(data, labels, existingChartId). "
+         "Returns nothing. data is an array of objects, labels is an optional "
+         "map of data keys to display labels, and existingChartId is an "
+         "optional string ID of a previously created chart to update. You must "
+         "use 'x' as the key for the x-axis. A unique chart ID will be "
+         "provided in the console output. "
          "Example: chartUtil.createLineChart([{x: 'Jan', sales: "
          "100, profit: 30}, {x: 'Feb', sales: 150, profit: 45}], {sales: "
          "'Sales ($)', profit: 'Profit ($)'}).";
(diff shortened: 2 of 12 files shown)


components/ai_chat/core/common/mojom/common.mojom
```diff
@@ -229,6 +229,7 @@ struct SimpleRequestContentBlock {
 
 // Artifact generated from tool call
 struct ToolArtifact {
```
netzenbot, line 278: nit: The new `string id` field should have an explicit default value (`string id = ""`). [ARCH-052](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-052)

components/ai_chat/core/common/mojom/common.mojom
```diff
@@ -229,6 +229,7 @@ struct SimpleRequestContentBlock {
...
+  string id;
```
netzenbot: nit: Since empty string is used as a sentinel for "no id" (checked via `!existing_id->empty()` in C++ and `if (artifact.id)` in TS), this should be `string? id` (nullable) to make the optionality explicit — consistent with other optional fields in this file like `conversation_turn_uuid`, `uuid`, etc. See [CSA-022](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-022).

components/ai_chat/core/browser/tools/code_plugin.h
```diff
@@ -37,6 +37,11 @@ class CodePlugin {
   // |artifact_value| is the parsed JSON.
   virtual std::optional<std::string> ValidateArtifact(
       const base::Value& artifact_value) const;
+
+  // Returns an optional log prefix used when generating artifact creation
+  // log messages (e.g. "Chart"). If nullopt, no log message is emitted for
+  // artifacts of this type.
+  virtual std::optional<std::string_view> ArtifactCreationLogPrefix() const;
```
netzenbot: nit: The example `(e.g. "Chart")` doesn't match what implementations actually return — `ChartCodePlugin` returns `"Chart created with ID: "`, a full sentence prefix. Consider updating the example to reflect the expected format, e.g. `(e.g. "Chart created with ID: ")`. ([Method Documentation Should Describe the Contract](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-004))

browser/ai_chat/code_execution_tool_browsertest.cc
```diff
@@ -320,10 +321,11 @@ IN_PROC_BROWSER_TEST_F(AIChatCodeExecutionToolBrowserTest, CreateLineChart) {
   std::vector<mojom::ToolArtifactPtr> artifacts;
   ExecuteCode(script, &output, &artifacts);
 
-  EXPECT_EQ(output, "Chart created successfully");
   ASSERT_EQ(artifacts.size(), 1u);
 
   const auto& artifact = artifacts[0];
+  EXPECT_FALSE(artifact->id.empty());
+  EXPECT_THAT(output, HasSubstr("Chart created with ID: " + artifact->id));
   EXPECT_EQ(artifact->type, mojom::kLineChartArtifactType);
```
netzenbot, line 330: nit: `"Chart created with ID: "` is duplicated between this test and `ChartCodePlugin::ArtifactCreationLogPrefix()`. If the prefix changes in production, this test won't catch it. Consider exposing it as a testable constant (e.g., a `constexpr` in the header) so the test references the same value. See [TI-009](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-009).

DJAndries: this is probably fine, we can just update the test if the prefix changes. not worth exposing a constant for every single expected string value IMHO

components/ai_chat/core/browser/tools/chart_code_plugin.cc
```diff
@@ -30,9 +30,12 @@ bool ChartCodePlugin::IsEnabled() {
...
 std::string_view ChartCodePlugin::Description() const {
-  return "Use chartUtil.createLineChart(data, labels) where data is an array "
-         "of objects and labels is an optional map of data keys to display "
-         "labels. You must use 'x' as the key for the x-axis. "
+  return "Use chartUtil.createLineChart(data, labels, existingChartId). "
+         "Returns nothing. data is an array of objects, labels is an optional "
+         "map of data keys to display labels, and existingChartId is an "
+         "optional string ID of a previously created chart to update. You must "
+         "use 'x' as the key for the x-axis. A unique chart ID will be "
+         "provided in the console output. "
```
petemill, line 39: nit: do we need to say `console output` instead of simply `output`? Concern about it not knowing how to access the console.

components/ai_chat/core/browser/tools/code_plugin.h
```diff
@@ -37,6 +37,11 @@ class CodePlugin {
...
 };
```
petemill: This seems a bit messy having a prefix created provided by one place, and knowing exactly what something else will append to it. How about inverting it to something like: `string CreateSuccessOutput(std::string_view artifact_id)`?

components/ai_chat/core/common/mojom/common.mojom
petemill: I was thinking this should be `string uuid` since that's the value type, but I like `id` better since we allow duplicates and de-duplication at the UI. Optional is a good recommendation. (more generally, I think the explicit default values is debatable).

components/ai_chat/core/browser/tools/chart_code_plugin.cc
DJAndries: I tried just "output" before, and the model generated code that expected a return value from `createLineChart`. I guess it was too ambiguous.

components/ai_chat/core/browser/tools/code_plugin.h
DJAndries: yeah good idea

components/ai_chat/core/common/mojom/common.mojom
DJAndries: changed to optional

components/ai_chat/core/common/mojom/common.mojom
DJAndries: 4f27e92b6e1

components/ai_chat/core/browser/tools/code_plugin.h
DJAndries: no longer applies

components/ai_chat/core/browser/tools/code_plugin.h
DJAndries: 4f27e92b6e1

components/ai_chat/resources/untrusted_conversation_frame/components/conversation_entries/conversation_entries_utils.ts
```diff
@@ -194,23 +194,33 @@ export const normalizeCitationSpacing = (text: string): string => {
...
+          artifactsById.set(artifact.id, artifact)
+        } else {
+          artifactsWithoutId.push(artifact)
+        }
       }
     }
   }
 
-  return artifacts
+  return [...artifactsWithoutId, ...artifactsById.values()]
```
petemill, line 225: Does preserving the order of artifacts matter? I think they will mostly come out in the order they were added to the hash, but the id-less ones will be returned after the id ones. 

Re-reading the `code_execution_tool.cc`, though, it seems we'll never be in this situation - we'll only have artifacts with IDs (new conversations) or without (old conversations before this PR lands).

components/ai_chat/core/browser/tools/code_plugin.h
```diff
@@ -37,6 +37,11 @@ class CodePlugin {
   // |artifact_value| is the parsed JSON.
   virtual std::optional<std::string> ValidateArtifact(
       const base::Value& artifact_value) const;
+
+  // Returns an optional log message for artifact creation, given the artifact
+  // ID. If nullopt, no output is emitted for artifacts of this type.
+  virtual std::optional<std::string> GetArtifactCreationLog(
+      std::string_view artifact_id) const;
 };
```
petemill, line 45: The fact that status is written as a log is a bit of an implementation leak - we're not actually logging this for the user/developer, but sending it to the AI - would you consider calling this something like `CreateArtifactSuccessMessage`?

components/ai_chat/core/browser/tools/chart_code_plugin.cc
```diff
@@ -30,9 +31,12 @@ bool ChartCodePlugin::IsEnabled() {
...
```
petemill, line 39: Will it know that it can't update a chart created in a previous message? It's gracefully handled, so perhaps it doesn't matter?

petemill (approved):

components/ai_chat/resources/untrusted_conversation_frame/components/conversation_entries/conversation_entries_utils.ts
DJAndries: > we'll only have artifacts with IDs (new conversations) or without (old conversations before this PR lands).

correct. the order does not matter.

components/ai_chat/core/browser/tools/chart_code_plugin.cc
DJAndries: Correct, it doesn't matter. We're only interested in supporting updating artifacts that were created in the same conversation turn, to avoid duplicates.

components/ai_chat/core/browser/tools/code_plugin.h
DJAndries: 1881d539784267cb22a733c3194831f33007056c

netzenbot (approved):

PR #34737, opened 2026-03-14
Add configurable history retention setting

### LINKED ISSUE #29045: Add option to customize browsing history retention time
## Description 
Currently, Brave Browser has a maximum retention time of 3 months for browsing history. While this may be sufficient for some users, others may require a longer retention period for various reasons, such as research or personal preference.

It would be helpful if Brave Browser could provide an option to increase or change the maximum retention time for browsing history. This feature would allow users to choose their preferred retention period based on their individual needs.

## Proposed Solution:

Add an option in the settings menu to allow users to increase or change the maximum retention time for browsing history. Users could select a retention period ranging from 3 months to a year, or even longer if needed.

## Benefits:

- Provides users with more control over their browsing history.
- Allows users to choose a retention period that meets their individual needs.
- Helps users who require a longer retention period for research or personal preference.
- Increases the flexibility and usefulness of Brave Browser.

Thank you for considering this feature request.

MaxMagnor:
## Summary

Adds a configurable history retention setting under brave://settings/privacy.

The setting lets users choose how long Brave keeps local browsing history:
1 day, 1 week, 1 month, 3 months, 6 months, 1 year, 5 years, or forever.

The default remains 90 days. Changing the setting updates the history expiration backend at runtime without requiring a restart.

Resolves brave/brave-browser#29045

## Screenshot 

image

## Test plan

Tested on Windows x64 Debug build.

- Set retention to 1 day, backdated a visit older than 1 day, and verified it expired.
- Set retention to 3 months, backdated a visit by 2 days, and verified it was kept.
- Changed retention from 3 months to 1 day and verified old entries expired.
- Set retention to Forever, backdated a visit by 5000 days, and verified it was still kept.
- Verified the setting appears in English and German.


bsclifton:
Thanks for the PR @MaxMagnor! 

https://github.com/brave/brave-core/pull/33821 was closed because this is changing the default behavior for everybody.

Would you be up for making this value configurable by the user? For example, maybe history length can be changed via brave://settings/privacy

bsclifton:
cc: @ShivanKaul

fallaciousreasoning:
Even if we could add a flag as a medium term solution I think this could get landed (though we wouldn't be able to change the default)

fallaciousreasoning:
Alternatively, if you were feeling brave I think you could add a settings dropdown to brave://settings/privacy

History Retention

with a few options (such as):
- 1 day
- 1 week
- 1 month
- 3 months
- 6 months
- 1 year
- 5 years

MaxMagnor:
Yeah, I will try to implement it as a customizable setting in the form of a drop-down menu.

fallaciousreasoning:
@MaxMagnor our security/privacy team have approved adding a setting or flag for changing the expiry time but not the default expiry, so if you get it working we should be okay to merge :smile:

Ruintar:
I've been following the conversation above and saw that @bsclifton and @fallaciousreasoning suggested making this a configurable setting with a dropdown menu in the privacy page, and that @MaxMagnor will try to implement it that way. Since the direction of the PR has shifted to this, I'd like to drop a reference here that perfectly matches what was requested.

Cromite has already solved this problem properly. Since Chromium v106 it has been inheriting and evolving what Bromite originally started, and with every single one of those 41 Chromium versions the feature has been refined and running in production, from v106 all the way to the current v147. To me that speaks very clearly about the maturity of the solution Cromite has built.

[https://github.com/uazo/cromite/blob/master/build/patches/History-number-of-days-privacy-setting.patch](url)

The approach initially suggested by the author of this pull request was quite different, but since fallaciousreasoning suggested modifying it to have a menu, I believe my suggestion fits perfectly as it delivers exactly what was asked. Instead of replacing a fixed value with another fixed value, Cromite turns **kExpireDaysThreshold** into a variable that can be changed while the browser is actually running, and adds a function in **HistoryBackend** that pushes that change to the expiration engine in real time. If the user picks no history, existing history gets wiped asynchronously in a safe way. If they pick forever, the expiration loop simply does nothing. And all of this kicks in the moment the user changes the setting, no browser restart needed, because the system hooks into a reactive preference listener.

Something Cromite handles carefully is making sure this setting doesn't bleed into parts of the codebase that have nothing to do with it. Metrics components, omnibox shortcuts and the embeddings database all keep 90 days hardcoded because in those places that value serves a different purpose that has nothing to do with what the user wants for their history. This keeps things clean and avoids side effects where they don't belong.

The UI is available on both desktop and Android, directly in the privacy settings page, with options for no history, 1, 7, 15, 30, 60, 90 days, and forever. Below are some images showing Cromite's settings on the desktop browser.





Adapting this for Brave would take considerably less work than figuring out the logic from scratch. The backend code would carry over almost untouched since both projects run on the same Chromium base. The things that would need Brave-specific attention are deciding whether this preference should travel via Brave Sync, wiring up the dropdown using Brave's UI components, and minding the licensing (GPL-3.0/GPL-2.0+), which in practice means using Cromite's architecture as a technical reference to write the final implementation under Brave's MPL-2.0 license. I believe looking at Cromite's code will save @MaxMagnor a lot of development time on this new phase of the PR.

components/constants/pref_names.h
```diff
@@ -104,6 +104,8 @@ inline constexpr char kTabMuteIndicatorNotClickable[] =
     "brave.tabs.mute_indicator_not_clickable";
 inline constexpr char kDontAskForCrashReporting[] =
     "brave.dont_ask_for_crash_reporting";
+// Number of days to retain browsing history. 0 disables saving history.
```
fallaciousreasoning: Maybe comment `-1` is forever here.

patches/components-history-core-browser-expire_history_backend.h.patch
```diff
@@ -0,0 +1,14 @@
+diff --git a/components/history/core/browser/expire_history_backend.h b/components/history/core/browser/expire_history_backend.h
+index dfcab38a3832b385d713bcb38d8eeaea907ad596..6dedfb03d1867ddf6be50bb5726120e9ba31f18d 100644
+--- a/components/history/core/browser/expire_history_backend.h
++++ b/components/history/core/browser/expire_history_backend.h
+@@ -82,6 +82,9 @@ class ExpireHistoryBackend {
+   // will continue until the object is deleted.
+   void StartExpiringOldStuff(base::TimeDelta expiration_threshold);
+ 
++  // Updates the threshold used by already-scheduled expiration work.
++  void SetExpirationThreshold(base::TimeDelta expiration_threshold);
```
fallaciousreasoning: could you try and implement this as a chromium_src override - we try and avoid patches as much as possible because they make rebasing much harder.

patches/components-history-core-browser-expire_history_backend.cc.patch
```diff
@@ -0,0 +1,47 @@
...
+   // reader types (in case we want to expire different types of visits in
+   // different ways), we can make it be populated by creator/owner of
+@@ -407,6 +413,11 @@ void ExpireHistoryBackend::StartExpiringOldStuff(
+   ScheduleExpire();
+ }
+ 
++void ExpireHistoryBackend::SetExpirationThreshold(
++    base::TimeDelta expiration_threshold) {
++  expiration_threshold_ = expiration_threshold;
++}
```
fallaciousreasoning: Once we have the function declared, this can just be implemented in the `chromium_src` override:

chromium_src/components/history/core/browser/expire_history_backend.cc

(you can create it if it doesn't exist)

patches/components-history-core-browser-expire_history_backend.cc.patch
```diff
@@ -0,0 +1,47 @@
...
++
+ void ExpireHistoryBackend::DeleteFaviconsIfPossible(DeleteEffects* effects) {
+   if (!favicon_db_)
+     return;
+@@ -625,6 +636,10 @@ void ExpireHistoryBackend::DoExpireIteration() {
+ 
+ void ExpireHistoryBackend::DoExpireIteration() {
++  if (expiration_threshold_.is_max()) {
++    return;
++  }
```
fallaciousreasoning: I know this is simpler to implement directly as a patch, but instead we should patch in something like `BRAVE_DO_EXPIRE_ITERATION` and then `#define BRAVE_DO_EXPIRE_ITERATION if (expiration_threshold_.is_max()) return;` in the `chromium_src` version of the file.

fallaciousreasoning: Actually, I think we can get rid of this one completely - its only invoked via StartExpiringOldStuff which will also have this check, right?

chromium_src/components/history/core/browser/history_backend.cc
```diff
@@ -7,9 +7,28 @@
...
+  if (days == kKeepHistoryForever) {
+    expirer_.SetExpirationThreshold(base::TimeDelta::Max());
+    return;
+  }
+
+  expirer_.StartExpiringOldStuff(base::Days(days));
+  if (days == 0) {
+    task_runner_->PostTask(
+        FROM_HERE, base::BindOnce(&HistoryBackend::DeleteAllHistory, this));
+  }
```
fallaciousreasoning: Hey, I don't think we need to add this `SetExpirationThreshold` function - `StartExpiringOldStuff` sets it as the first line.
```suggestion
  auto threshold = days == kKeepHistoryForever
    ? base::TimeDelta::Max()
    : base::Days(days);
    
  expirer_.StartExpiringOldStuff(threshold);
  if (days == 0) {
    task_runner_->PostTask(
        FROM_HERE, base::BindOnce(&HistoryBackend::DeleteAllHistory, this));
  }
```

patches/components-history-core-browser-expire_history_backend.cc.patch
```diff
@@ -0,0 +1,47 @@
...
+ 
+-  // Remove all readers, just in case this was method was called before.
++  // Remove all readers and pending work, just in case this method was called
++  // before.
+   readers_.clear();
++  work_queue_ = {};
++
++  if (expiration_threshold_.is_max()) {
++    return;
++  }
```
fallaciousreasoning: Let's use a hook + chromium_src approach here, to reduce the size of the patch:

1. Patch in `BRAVE_START_EXPIRING_OLD_STUFF`
2. Define `BRAVE_START_EXPIRING_OLD_STUFF` in the`.cc` file for this in `chromium_src`

patches/components-history-core-browser-expire_history_backend.cc.patch
fallaciousreasoning: I think we can remove this function

patches/components-history-core-browser-expire_history_backend.cc.patch
fallaciousreasoning: Additionally, I think it'd be good to cancel the timer here, if we can.

patches/chrome-browser-history-history_service_factory.cc.patch
```diff
@@ -0,0 +1,45 @@
...
+   return history_service;
+ }
+ 
+@@ -40,7 +42,9 @@ history::HistoryService* HistoryServiceFactory::GetForProfile(
+     Profile* profile,
+     ServiceAccessType sat) {
+   // If saving history is disabled, only allow explicit access.
+-  if (sat != ServiceAccessType::EXPLICIT_ACCESS &&
+-      profile->GetPrefs()->GetBoolean(prefs::kSavingBrowserHistoryDisabled)) {
++  if (sat != ServiceAccessType::EXPLICIT_ACCESS &&
```
fallaciousreasoning: I think for now, let's remove the `0` option - if the user wants that they can set the `kSavingBrowserHistoryDisabled` pref, and its complicating this PR a bit.

chromium_src/components/history/core/browser/history_service.h
```diff
@@ -6,18 +6,28 @@
...
+  void Cleanup
 
 #define AddRelatedSearchesForVisit                                     \
   GetKnownToSyncCount(                                                 \
       base::OnceCallback<void(history::HistoryCountResult)> callback); \
+  void InitHistoryRetentionPref(PrefService* prefs);                   \
+  void OnHistoryRetentionDaysChanged();                                \
+  raw_ptr<PrefService> active_user_pref_service_ = nullptr;            \
+  std::unique_ptr<PrefChangeRegistrar>                                 \
+      active_user_pref_change_registrar_;                              \
```
fallaciousreasoning: Instead of this, could we add an `IntegerPrefMember` and subscribe to that? Then I think we wouldn't need to store the PrefService here too.

patches/chrome-browser-history-history_service_factory.cc.patch
```diff
@@ -0,0 +1,45 @@
...
++#include "brave/components/constants/pref_names.h"
+ #include "chrome/browser/bookmarks/bookmark_model_factory.h"
+ #include "chrome/browser/history/chrome_history_client.h"
+ #include "chrome/browser/profiles/profile.h"
+@@ -30,6 +30,8 @@ std::unique_ptr<KeyedService> BuildHistoryService(
+           context->GetPath(), chrome::GetChannel()))) {
+     return nullptr;
+   }
++  history_service->InitHistoryRetentionPref(
++      Profile::FromBrowserContext(context)->GetPrefs());
```
fallaciousreasoning: rather than calling this here, could we invoke it from the constructor of the history service?

fallaciousreasoning: then we could completely drop this patch

fallaciousreasoning (commented):
This is looking great @MaxMagnor, thanks for the contribution. I've left a few comments, mostly around how I think the patching approach could be cleaned up (we've got pretty strict requirements around patching so we can stay up to date on Chromium).

The other thing is I think we'll need some tests for this feature, so we can catch if a rebase breaks it in the future.

fallaciousreasoning:
@MaxMagnor please remove all the stuff in `.agents`

patches/chrome-browser-history-history_service_factory.cc.patch
MaxMagnor: I looked into this, but HistoryService’s constructor does not currently receive a Profile or PrefService. The retention listener needs the profile PrefService so IntegerPrefMember can subscribe to the Brave pref.

I reduced this patch so it only initializes the retention pref listener at the factory boundary, where the Profile prefs are already available. Moving this into the constructor would require broader constructor plumbing for what is now a very small patch.

app/resources/brave_generated_resources_de.xtb
fallaciousreasoning, line 1: You can revert this change - it'll get properly localized by the translation team prior to release.

Thanks for thinking about localization though!

patches/chrome-browser-resources-history-app.html.ts.patch
```diff
@@ -1,18 +0,0 @@
-diff --git a/chrome/browser/resources/history/app.html.ts b/chrome/browser/resources/history/app.html.ts
```
fallaciousreasoning, line 1: Did you mean to remove this file? It seems unrelated to the rest of your PR.

patches/chrome-browser-history-history_service_factory.cc.patch
fallaciousreasoning: Good point - okay, sorry I know its a pain but could you patch in a hook here too `BRAVE_INIT_HISTORY_RETENTION_PREF` and define it in the `chromium_src` override - we try to avoid patching in code directly.

chromium_src/components/history/core/browser/history_service.cc
```diff
@@ -3,10 +3,25 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this file,
  * You can obtain one at https://mozilla.org/MPL/2.0/. */
 
+#include "brave/components/constants/pref_names.h"
+#include "components/prefs/pref_service.h"
+
 #include <components/history/core/browser/history_service.cc>
```
fallaciousreasoning: nit sorting should be:

```
"header_file.h"

<triangles>

"normal_imports.h"
```

fallaciousreasoning (commented):
This is looking great - I think you probably need to run a format (`npm run format` and `npm run presubmit -- --fix`).

I'll get the `chromium_src` and patch reviewers teams to have a look.

UI looks good and the feature works nicely
image

Worth adding a few more unit test cases - ideally, we would set the pref and then check history gets cleaned up.

MaxMagnor:
Addressed the review feedback and added unit coverage for the retention backend.

Tests:
- `npm run test -- brave_unit_tests Debug --filter=BraveExpireHistoryBackendTest.*`

Result:
- `BraveExpireHistoryBackendTest.ExpiresOldVisitsForPositiveThreshold`
- `BraveExpireHistoryBackendTest.ForeverCancelsPendingExpiration`
- SUCCESS: all tests passed

fallaciousreasoning:
Hey @MaxMagnor did you push the wrong thing? Seems to be unchanged?

bsclifton (approved):
string reviewers ++

ShivanKaul:
Let's call the setting "Local history retention", description looks good.

app/brave_settings_strings.grdp
```diff
@@ -773,6 +773,37 @@
   <message name="IDS_SETTINGS_PUSH_MESSAGING" desc="Select value">
     Use Google services for push messaging
   </message>
+  <!-- History retention -->
+  <message name="IDS_SETTINGS_HISTORY_RETENTION_LABEL" desc="Label for the setting that controls how long browsing history is retained">
+    History retention
```
fallaciousreasoning:
```suggestion
    Local history retention
```

as per @ShivanKaul's suggestion

patches/chrome-browser-resources-history-app.html.ts.patch
MaxMagnor: Yea, my bad. Fixed it.

MaxMagnor:
Updated this based on the latest review feedback:

- Moved the history service factory initialization behind `BRAVE_INIT_HISTORY_RETENTION_PREF`.
- Removed the manual German `.xtb` translations.
- Restored the unrelated history `app.html.ts` patch file.
- Renamed the setting to “Local history retention”.
- Fixed the include ordering nit.
- Added pref-driven unit coverage for retention changes.

Tested on Windows:

```bat
npm run test -- brave_unit_tests Debug --filter=BraveExpireHistoryBackendTest.*:BraveHistoryRetentionPrefTest.*
```
Result:
```text
[1/4] BraveExpireHistoryBackendTest.ExpiresOldVisitsForPositiveThreshold
[2/4] BraveExpireHistoryBackendTest.ForeverCancelsPendingExpiration
[3/4] BraveHistoryRetentionPrefTest.PrefChangeExpiresOldVisits
[4/4] BraveHistoryRetentionPrefTest.ForeverPrefKeepsOldVisits
SUCCESS: all tests passed.
```
I also ran npm run format and npm run presubmit -- --fix. Presubmit still reports unrelated existing Windows/tooling failures, but the history-related checkdeps issue from this PR is resolved.

app/resources/brave_generated_resources_de.xtb
fallaciousreasoning, line 1: please completely revert this file, so it doesn't show up in the diff:

```
git checkout origin/master -- app/resources/brave_generated_resources_de.xtb
```

fallaciousreasoning (approved):
I've kicked of CI - pinging the remaining reviewers again

components/constants/pref_names.h
```diff
@@ -101,6 +101,8 @@ inline constexpr char kTabMuteIndicatorNotClickable[] =
     "brave.tabs.mute_indicator_not_clickable";
 inline constexpr char kDontAskForCrashReporting[] =
     "brave.dont_ask_for_crash_reporting";
+// Number of days to retain browsing history. -1 keeps history forever.
+inline constexpr char kHistoryRetentionDays[] = "brave.history.retention_days";
```
goodov: please do not add new stuff here. 

instead declare the pref via `chromium_src/components/history/core/common/pref_names.h` as `inline constexpr char k...[] = "...";`

chromium_src/components/history/core/browser/history_backend.cc
```diff
@@ -7,9 +7,21 @@
...
+constexpr int kKeepHistoryForever = -1;
+
+}  // namespace
+
 HistoryCountResult HistoryBackend::GetKnownToSyncCount() {
   int count = 0;
   return {db_ && db_->GetKnownToSyncCount(&count), count};
 }
 
+void HistoryBackend::UpdateExpireDaysThreshold(int days) {
```
goodov: please pass `base::TimeDelay` here and do the value preparation on the caller side.

chromium_src/components/history/core/browser/history_service.cc
```diff
@@ -3,10 +3,27 @@
...
+
 namespace history {
 
+namespace {
+
+constexpr int kDefaultHistoryRetentionDays = 90;
+constexpr int kKeepHistoryForever = -1;
+constexpr char kHistoryRetentionDays[] = "brave.history.retention_days";
+
+int NormalizeHistoryRetentionDays(int days) {
```
goodov: make this return `base::TimeDelta` and maybe name it `HistoryRetentionDaysToTimeDelta`

chromium_src/components/history/core/browser/history_backend.cc
```diff
@@ -7,9 +7,21 @@
...
+  expirer_.StartExpiringOldStuff(days == kKeepHistoryForever
```
goodov: we now call this multiple times (two times at startup and on each pref change). Are you sure this is right?

the expiration logic runs the cleanup loop consistently each 30 seconds or 5 minutes, and we should get the desired result only by changing `expiration_threshold_` value.

The problem right now: there's unconditional `PostTask` in `ExpireHistoryBackend::ScheduleExpire` that queues `ExpireHistoryBackend::DoExpireIteration` on each call, and each call of `StartExpiringOldStuff` will create additional loop of `ScheduleExpire -> DoExpireIteration -> PostTask(ScheduleExpire)`.

app/resources/brave_generated_resources_de.xtb
```diff
@@ -131,12 +131,16 @@ Oder später ändern unter <ph name="SETTINGS_EXTENIONS_LINK"/>.</translation>
 <translation id="8184318863960255706">Weitere Informationen</translation>
 <translation id="780301667611848630">Kein Interesse</translation>
 <translation id="1162706247109707219">Indem Sie auf „OK“ klicken, nehmen Sie zur Kenntnis, dass Ihre Brave-Browser-Installation KEINE kritischen Sicherheitspatches erhalten wird, solange Sie kein Update auf Windows 10 oder höher vornehmen. Diese Warnung wird NICHT wieder angezeigt.</translation>
-<translation id="2971591422768125177">Indem Sie auf „OK“ klicken, nehmen Sie zur Kenntnis, dass Ihre Brave-Browser-Installation KEINE kritischen Sicherheitspatches erhalten wird, solange Sie kein Update auf macOS 10.15 oder höher vornehmen. Diese Warnung wird NICHT wieder angezeigt.</translation>
```
goodov, line 134: this file should not be changed.

chromium_src/components/history/core/browser/history_service.h
```diff
@@ -18,6 +22,9 @@ class BraveHistoryQuickProviderTest;
 #define AddRelatedSearchesForVisit                                     \
   GetKnownToSyncCount(                                                 \
       base::OnceCallback<void(history::HistoryCountResult)> callback); \
+  void InitHistoryRetentionPref(PrefService* prefs);                   \
+  void OnHistoryRetentionDaysChanged();                                \
+  IntegerPrefMember history_retention_days_;                           \
```
goodov, line 29:
```
 private:
  IntegerPrefMember history_retention_days_;  
 public:
```

chromium_src/components/history/core/common/pref_names.h
```diff
@@ -0,0 +1,18 @@
...
+
+#ifndef BRAVE_CHROMIUM_SRC_COMPONENTS_HISTORY_CORE_COMMON_PREF_NAMES_H_
+#define BRAVE_CHROMIUM_SRC_COMPONENTS_HISTORY_CORE_COMMON_PREF_NAMES_H_
+
+#include <components/history/core/common/pref_names.h>  // IWYU pragma: export
+
+namespace prefs {
+
+// Number of days to retain browsing history. -1 keeps history forever.
+inline constexpr char kHistoryRetentionDays[] = "brave.history.retention_days";
```
goodov: let's go with `kBraveHistoryRetentionDays` for clarity that this is Brave-specific pref.

patches/components-history-core-browser-expire_history_backend.cc.patch
```diff
@@ -0,0 +1,9 @@
+diff --git a/components/history/core/browser/expire_history_backend.cc b/components/history/core/browser/expire_history_backend.cc
+index 193014683e670facf7192eccf392c800cb2be879..2351d061567bfe09d5a0421fcc6d0d7bc08e4287 100644
+--- a/components/history/core/browser/expire_history_backend.cc
++++ b/components/history/core/browser/expire_history_backend.cc
+@@ -397,3 +397,4 @@ void ExpireHistoryBackend::StartExpiringOldStuff(
+   // Remove all readers, just in case this was method was called before.
+   readers_.clear();
++  BRAVE_START_EXPIRING_OLD_STUFF
```
goodov: nit: this should be called `BRAVE_EXPIRE_HISTORY_BACKEND_START_EXPIRING_OLD_STUFF`.

patches/chrome-browser-history-history_service_factory.cc.patch
```diff
@@ -0,0 +1,11 @@
+diff --git a/chrome/browser/history/history_service_factory.cc b/chrome/browser/history/history_service_factory.cc
+index f51a9c7b49118cddda8b75d38804c3e9875fef43..b17c0b2f00b86dd07cd46d2a127f75dcf759057f 100644
+--- a/chrome/browser/history/history_service_factory.cc
++++ b/chrome/browser/history/history_service_factory.cc
+@@ -30,5 +30,6 @@ std::unique_ptr<KeyedService> BuildHistoryService(
+           context->GetPath(), chrome::GetChannel()))) {
+     return nullptr;
+   }
++  BRAVE_INIT_HISTORY_RETENTION_PREF
```
goodov: `BRAVE_BUILD_HISTORY_SERVICE`

chromium_src/components/history/core/browser/history_backend.cc
```diff
@@ -12,4 +12,20 @@ HistoryCountResult HistoryBackend::GetKnownToSyncCount() {
...
+    expirer_.weak_factory_.InvalidateWeakPtrs();
+    return;
+  }
+
+  if (expirer_.expiration_threshold_.is_max()) {
+    expirer_.StartExpiringOldStuff(threshold);
+    return;
+  }
+
+  expirer_.expiration_threshold_ = threshold;
```
goodov: make all of this be part of `ExpireHistoryBackend` exposed as a public method, and call it from here. This way you don't need to friend HistoryBackend.

use chromium_src define approach to add new method without git patches into `expire_history_backend.h/cc`:
```cpp
#define ClearOldOnDemandFaviconsIfPossible(...) \
  ClearOldOnDemandFaviconsIfPossible(__VA_ARGS__);
  void UpdateExpireDaysThreshold(base::TimeDelta threshold)
```

chromium_src/components/history/core/browser/history_backend.h
```diff
@@ -6,8 +6,11 @@
 #ifndef BRAVE_CHROMIUM_SRC_COMPONENTS_HISTORY_CORE_BROWSER_HISTORY_BACKEND_H_
 #define BRAVE_CHROMIUM_SRC_COMPONENTS_HISTORY_CORE_BROWSER_HISTORY_BACKEND_H_
 
-#define GetHistoryCount  \
-  GetKnownToSyncCount(); \
+#include "base/time/time.h"
+
+#define GetHistoryCount                                      \
+  GetKnownToSyncCount();                                     \
+  void UpdateExpireDaysThreshold(base::TimeDelta threshold); \
```
goodov: please rename `UpdateExpireDaysThreshold` -> `UpdateExpirationThreshold` here and in expire_history_backend.

chromium_src/components/history/core/browser/expire_history_backend.cc
```diff
@@ -0,0 +1,36 @@
...
+    return;                                                   \
+  }
+
+#include <components/history/core/browser/expire_history_backend.cc>
+
+namespace history {
+
+void ExpireHistoryBackend::UpdateExpireDaysThreshold(
+    base::TimeDelta threshold) {
+  if (threshold.is_max()) {
```
goodov, line 25: can you please add explanation for these conditions and logic inside on why they are laid in this order for future reference?

I had to read upstream code to understand why it's done this way, and it does look correct to me. However in future this might be not that obvious, so comments here would be very valuable.

MaxMagnor:
Updated again based on the latest review:

- Removed the generated German translation file from the PR diff.
- Moved the Brave history retention pref into `chromium_src/components/history/core/common/pref_names.h`.
- Renamed it to `prefs::kBraveHistoryRetentionDays`.
- Made the history service pref member private.
- Renamed the Brave hook macros as requested.
- Moved expiration threshold update logic into `ExpireHistoryBackend::UpdateExpireDaysThreshold(...)`.
- Removed the friend-based `expire_history_backend.h` patch.

Focused unit tests pass:

```bat
npm run test -- brave_unit_tests Debug --filter=BraveExpireHistoryBackendTest.*:BraveHistoryRetentionPrefTest.*
```

goodov:
please:
1. rebase to latest master
2. npm run sync
3. npm run update_patches
4. npm run format
5. npm run build
6. npm run gn_check
7. npm run presubmit

goodov:
```
** Presubmit ERRORS: 1 **
The code requires formatting. Please run: npm run presubmit -- --fix.
--- chromium_src/components/history/core/browser/history_service.cc  (before formatting)
+++ chromium_src/components/history/core/browser/history_service.cc  (after formatting)
@@ -5,10 +5,10 @@
 #include "components/history/core/browser/history_service.h"
-#include <components/history/core/browser/history_service.cc>
-
 #include "components/history/core/common/pref_names.h"
 #include "components/prefs/pref_service.h"
+
+#include <components/history/core/browser/history_service.cc>
```

MaxMagnor:
Fixed the presubmit formatting issue in `chromium_src/components/history/core/browser/history_service.cc` and force-pushed the amended commit.

goodov (approved):

goodov (changes requested):
please rebase to latest master 

Image

goodov:
still see the same error:

```
** Presubmit ERRORS: 1 **
The code requires formatting. Please run: npm run presubmit -- --fix.
--- chromium_src/components/history/core/browser/history_service.cc  (before formatting)
+++ chromium_src/components/history/core/browser/history_service.cc  (after formatting)
@@ -5,10 +5,10 @@
 #include "components/history/core/browser/history_service.h"
-#include <components/history/core/browser/history_service.cc>
-
 #include "components/history/core/common/pref_names.h"
 #include "components/prefs/pref_service.h"
+
+#include <components/history/core/browser/history_service.cc>
 namespace history {
```

MaxMagnor:
Hopefully now everything is fine

fallaciousreasoning:
Looks like CI passed @goodov https://github.com/brave/brave-core/pull/36504

goodov (approved):
lgtm. the approved commit to merge is https://github.com/brave/brave-core/pull/34737/changes/7439145300fb400dcb6f17242c20811111bb4d0e

cc @bsclifton

bsclifton:
Thanks for the contribution, @MaxMagnor! 😄

fallaciousreasoning:
Nice work getting this merged @MaxMagnor!

Ruintar:
Just dropping by to take a moment to thank @MaxMagnor for doing the heavy lifting and making this a reality, as well as everyone else involved in this discussion who helped get this feature built and merged into Brave. Seeing this finally land in v1.93.12 on Chromium 149 is really rewarding. Thanks, everyone!

PR #34784, opened 2026-03-17
[Polkadot] Add polkadot metadata parsing step.

### LINKED ISSUE #53789: [Polkadot] Add Polkadot chain metadata parsing
### Description

Currently we have hardcoded chain metadata for a couple of chains. We need to download metadata for each supported chain and extract pallet information for further extrinsic building.

### Steps to reproduce

-

### Actual result

-

### Expected result

-

### Reproduces how often

Easily reproduced

-

### Channel information

- [ ] release (stable)
- [ ] beta
- [ ] nightly

### Reproducibility

### Miscellaneous information

_No response_

cypt4:
Parse polkadot SCALE encoded metadata response.

Resolves https://github.com/brave/brave-browser/issues/53789


components/brave_wallet/browser/internal/polkadot_chain_metadata.rs
```diff
@@ -0,0 +1,433 @@
...
+///   - `System.Version.spec_version`
+///
+/// References:
+/// - JSON-RPC method: https://github.com/w3f/PSPs/blob/b6d570173146e7a012cf43d270177e02ed886e2e/PSPs/drafts/psp-6.md#1119-state_getmetadata
+/// - Runtime metadata format (Rust): https://docs.rs/frame-metadata/latest/frame_metadata/
+fn parse_chain_metadata_from_scale(bytes: &[u8]) -> Result<CxxPolkadotChainMetadata, Error> {
+    let mut input = bytes;
+    // RuntimeMetadataPrefixed starts with little-endian "meta" magic.
+    let magic: u32 = decode_scale(&mut input)?;
+    if magic != 0x6174_656d {
```
cmazakas: Let's lift this into a named constant.

components/brave_wallet/browser/internal/BUILD.gn
```diff
@@ -112,11 +112,18 @@ rust_static_library("polkadot_extrinsic_rs") {
...
+    "polkadot_extrinsic.rs",
+  ]
+  cxx_bindings = [
+    "polkadot_chain_metadata.rs",
+    "polkadot_extrinsic.rs",
+  ]
 
   deps = [
     "//brave/third_party/rust/blake2b_simd/v1:lib",
+    "//brave/third_party/rust/hex/v0_4:lib",
```
cmazakas: Cross-linking this for future reviewers: https://github.com/brave/brave-core/pull/34760#discussion_r2941381352

components/brave_wallet/browser/polkadot/polkadot_chain_metadata_unittest.cc
```diff
@@ -0,0 +1,124 @@
...
+
+  auto metadata = PolkadotChainMetadata::FromMetadataHex(*metadata_hex);
+  ASSERT_TRUE(metadata);
+  EXPECT_EQ(metadata->GetSs58Prefix(), 42u);
+  EXPECT_EQ(metadata->GetTransferAllowDeathCallIndex(), 0u);
+  EXPECT_EQ(metadata->GetBalancesPalletIndex(), 4u);
+  EXPECT_EQ(metadata->GetSpecVersion(), 1022000u);
+}
+
+}  // namespace brave_wallet
```
netzenbot, line 152: nit: Test utilities from `base/test/values_test_util.h` are preferred for JSON parsing in tests. For reading JSON from a file, `base::test::ParseJsonDictFromFile(fixture_path)` would replace the `ReadFileToString` + `JSONReader::ReadDict` sequence entirely and produce cleaner failure messages. For the string-based parse, `base::test::ParseJsonDict(json_response)` is the idiomatic equivalent. ([✅ Use `base::test::ParseJson` and `base::ExpectDict*` in Tests](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-051))

netzenbot, line 152: nit: Per CSA-054, JSON property key strings should use named constants rather than inline literals — e.g., `constexpr char kResult[] = "result";`. The string `"result"` is used in both fixture tests, so a shared constant would avoid any future typo risk. ([Use Named Constants for JSON Property Keys](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-054))

components/brave_wallet/browser/polkadot/polkadot_chain_metadata.cc
```diff
@@ -0,0 +1,124 @@
...
+
+  return std::nullopt;
+}
+
+const CxxPolkadotChainMetadata& PolkadotChainMetadata::operator*() const {
+  return chain_metadata_;
+}
+
+PolkadotChainMetadata::PolkadotChainMetadata(
+    CxxPolkadotChainMetadata chain_metadata)
```
netzenbot: nit: The private constructor `PolkadotChainMetadata(CxxPolkadotChainMetadata)` is declared in the `private:` section of the header, after all public methods. In the `.cc` it appears between `operator*` and the public getters. Per CS-017, definitions in `.cc` should follow the same order as declarations in `.h`. Move the private constructor definition to after `GetSpecVersion`. ([✅ Function Ordering in `.cc` Should Match `.h`](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-017))

components/brave_wallet/browser/polkadot/polkadot_chain_metadata.h
```diff
@@ -0,0 +1,70 @@
...
+      uint8_t balances_pallet_index,
+      uint8_t transfer_allow_death_call_index,
+      uint16_t ss58_prefix,
+      uint32_t spec_version);
+
+  // Build metadata from a known relay/parachain name returned by system_chain.
+  // Returns std::nullopt for unknown names.
+  static std::optional<PolkadotChainMetadata> FromChainName(
+      std::string_view chain_name);
+
```
netzenbot, line 54: nit: The header doc for `FromChainName` should mention that the returned metadata has an unknown (zero) `spec_version` and callers must populate it from `state_getRuntimeVersion` before use. This restriction is currently only documented inside the `.cc` file, but it belongs in the header where callers see the API. ([✅ Method Documentation Should Describe the Contract](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-004))

components/brave_wallet/browser/polkadot/polkadot_chain_metadata_unittest.cc
netzenbot, line 152: nit: The old `MetadataFromChainName` test included `EXPECT_FALSE(PolkadotChainMetadata::FromChainName(""))` but this case wasn't carried over to the new `FromChainName` test. Per [TI-017](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-017), test coverage should not decrease during refactoring — the empty-string rejection check should be added here.

cmazakas (approved):
Looks like this code is going to be easy to extend in terms of probing for the other pallets, which we're going to have to do soon for transactionpayment which is required for parsing block events and finding the true fee the sender paid.

This is great stuff!

components/brave_wallet/browser/polkadot/polkadot_chain_metadata_unittest.cc
```diff
@@ -0,0 +1,152 @@
...
+  const auto fixture_path =
+      base::PathService::CheckedGet(base::DIR_SRC_TEST_DATA_ROOT)
+          .AppendASCII("brave")
+          .AppendASCII("components")
+          .AppendASCII("test")
+          .AppendASCII("data")
+          .AppendASCII("brave_wallet")
+          .AppendASCII("polkadot")
+          .AppendASCII("chain_metadata")
+          .AppendASCII("state_getMetadata_polkadot.json");
```
kdenhartog, line 102: q: is this state_getMetadata_polkadot.json file just a way to perform standardized test vectors for the external API? Any reason we can't just put it as a string in this test here rather than a separate test data directory?

I've not seen us do that before, so would prefer to keep it consistent with how we do other unit tests if possible.

kdenhartog:
I did some follow up sec review on this afterwards and found these edge cases, so opened a separate PR for them: https://github.com/brave/brave-core/pull/35422

components/brave_wallet/browser/polkadot/polkadot_chain_metadata_unittest.cc
cypt4: It has ~1 mb size, so it is not convenient to keep it in the codebase. We are using test dirs for test vectors in other components, like AI chat.

components/brave_wallet/browser/internal/polkadot_chain_metadata.rs
meisamrezapuran02-collab: Thanks for all

meisamrezapuran02-collab: Thanks for all

PR #34850, opened 2026-03-20
[AI Chat] Fetch model public key for E2EE

### LINKED ISSUE #53818: [AI Chat] Add support for end-to-end encryption with NEAR models
NEAR supports end-to-end encryption with models hosted in their trusted execution environment, allowing for confidential conversations.

Add an "encryption" conversation capability.

If the capability is provided, do the following in the API client:
- Fetch the `ed25519 public signing key associated with the model via a new server endpoint
- Convert it into a x25519 public key (Edwards curve point to Montgomery curve point), perform key exchange
- Derive the encryption key from the shared secret using HKDF SHA256
- Encrypt using XChaCha20Poly1305.

Use existing Rust libraries for cryptography. Vendor x25519-dalek for key exchange.

Enable E2EE for GLM-5 to start.

Attestation/Intel SGX quote verification is currently out of scope.

DJAndries:
Resolves https://github.com/brave/brave-browser/issues/53818


components/ai_chat/core/browser/engine/e2ee_processor.cc
```diff
@@ -0,0 +1,114 @@
...
+    std::move(callback).Run(mojom::APIError::InternalError);
+    return;
+  }
+
+  attestation_cache_[model_name] =
+      Attestation{*model_public_key, base::TimeTicks::Now()};
+  std::move(callback).Run(std::nullopt);
+}
+
+}  // namespace ai_chat
```
netzenbot, line 114: The cache-hit path calls `callback.Run()` synchronously (line 340), while the cache-miss path dispatches an async network request. Per [ARCH-060](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-060), a callback must always be either sync or async — never both. The cache hit should post the callback via `base::SequencedTaskRunner::GetCurrentDefault()->PostTask(...)` to ensure consistent async delivery.

components/ai_chat/core/browser/engine/e2ee_processor_unittest.cc
```diff
@@ -0,0 +1,155 @@
...
+                    ResponseConversionCallback) {
+              std::move(callback).Run(api_request_helper::APIRequestResult(
+                  response_code, body->Clone(), {}, net::OK, GURL()));
+              return api_request_helper::APIRequestHelper::Ticket();
+            });
+  }
+
+  std::optional<mojom::APIError> FetchAndGetError() {
+    std::optional<mojom::APIError> result;
+    base::RunLoop run_loop;
```
netzenbot, line 78: The `FetchAndGetError()` helper uses the verbose `RunLoop` + `BindLambdaForTesting` + `Quit()` pattern that best practices require to avoid. Since `FetchModelAttestation` takes a `base::OnceCallback<void(std::optional<mojom::APIError>)>`, you can use `base::test::TestFuture<std::optional<mojom::APIError>>` directly: `base::test::TestFuture<std::optional<mojom::APIError>> future; processor_->FetchModelAttestation(kModelName, future.GetCallback()); return future.Get();`. See [TA-009](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-async.md#TA-009).

components/ai_chat/core/browser/engine/e2ee_processor.h
```diff
@@ -0,0 +1,73 @@
...
+
+namespace ai_chat {
+
+class E2EEProcessor {
+ public:
+  using FetchModelAttestationCallback =
+      base::OnceCallback<void(std::optional<mojom::APIError>)>;
+
+  explicit E2EEProcessor(
+      scoped_refptr<network::SharedURLLoaderFactory> url_loader_factory);
```
netzenbot, line 37: nit: `SetAPIRequestHelperForTesting` is in the `public:` section, which also triggered the presubmit warning. Per [ARCH-020](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-020), prefer a `private:` method with `FRIEND_TEST_ALL_PREFIXES` declarations rather than a public testing method. The existing `ConversationAPIV2Client` uses `protected` for its testing helpers as a workaround, but the recommended pattern here is `friend class`.

components/ai_chat/core/browser/engine/conversation_api_v2_client.h
```diff
@@ -119,10 +121,20 @@ class ConversationAPIV2Client {
   std::optional<std::string> GetLeoModelKeyFromResponse(
       const base::DictValue& response);
 
+  void OnCredentialFetched(
+      const std::optional<std::string>& attestation_model_name,
+      base::OnceCallback<void(std::optional<CredentialCacheEntry>,
+                              std::optional<mojom::APIError>)> callback,
+      std::optional<CredentialCacheEntry> credential);
+
+  void EnsureE2EEProcessor();
```
netzenbot, line 130: nit: `const std::optional<std::string>&` is flagged by [CSA-031](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-031) — prefer `base::optional_ref<const std::string>` for a non-owning optional reference, or pass by value if the string will be stored.

components/ai_chat/core/browser/engine/e2ee_processor_unittest.cc
```diff
@@ -0,0 +1,155 @@
...
+  // Still within TTL — resolves from cache, no extra Request call needed.
+  task_environment_.FastForwardBy(base::Minutes(55));
+  EXPECT_FALSE(FetchAndGetError().has_value());
+
+  // Past TTL — a fresh network request must be issued.
+  task_environment_.FastForwardBy(base::Minutes(10));
+  EXPECT_FALSE(FetchAndGetError().has_value());
+}
+
+}  // namespace ai_chat
```
netzenbot, line 155: nit: `mock_` is not owned by this test class (ownership is transferred to `processor_`). Per Chromium convention, add `// not owned` comment. See [CSM-002](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-memory.md#CSM-002).

components/ai_chat/core/browser/engine/e2ee_processor.cc
DJAndries: interesting, never heard about this guideline. i don't understand why posting the callback would be better here

components/ai_chat/core/browser/engine/e2ee_processor_unittest.cc
DJAndries: nice thanks, i'll check it out

components/ai_chat/core/browser/engine/conversation_api_v2_client.h
DJAndries: cool

components/ai_chat/core/browser/engine/e2ee_processor_unittest.cc
DJAndries: it's implied

PR #34851, opened 2026-03-20
[AI Chat] Add support for end-to-end encryption with NEAR models

DJAndries:
Resolves https://github.com/brave/brave-browser/issues/53818


components/ai_chat/core/browser/engine/e2ee_processor_unittest.cc
```diff
@@ -152,4 +154,105 @@ TEST_F(E2EEProcessorTest, FetchModelAttestation_ExpiresAfterOneHour) {
...
+
+  // The decrypted content should be the JSON-serialized ListValue.
+  std::string expected_json;
+  base::JSONWriter::Write(content, &expected_json);
+  EXPECT_EQ(*decrypted, expected_json);
+}
+
+TEST_F(E2EEProcessorTest, Encrypt_WithMissingAttestation_Crashes) {
+  base::ListValue content;
+  content.Append(base::Value("hello"));
```
netzenbot, line 200: `Encrypt_WithMissingAttestation_Crashes` uses `EXPECT_DEATH` but the fixture is `E2EEProcessorTest`. Per GoogleTest conventions, death test suite names must end with `DeathTest` (e.g., `E2EEProcessorDeathTest`) so the test runner applies the correct threading behavior. See [CS-061](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-061).

components/ai_chat/core/browser/engine/conversation_api_v2_client_unittest.cc
```diff
@@ -2134,4 +2204,98 @@ TEST_F(ConversationAPIV2ClientUnitTest, OnQueryDataReceived_CompletionChunk) {
   }
```
netzenbot, line 2204: Use `ASSERT_TRUE(public_key)` and `ASSERT_TRUE(messages_list)` instead of `EXPECT_TRUE` + `if`-guard. The `if` block silently skips the inner assertions when the pointer is null, hiding secondary failures. `ASSERT_TRUE` stops the test immediately so failures are obvious. ([Use ASSERT for Preconditions That Subsequent Code Depends On](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-034))

components/ai_chat/core/browser/engine/e2ee_processor.cc
```diff
@@ -99,16 +118,97 @@ void E2EEProcessor::OnFetchModelAttestationComplete(
...
+      model_public_key);
+}
+
+E2EEProcessor::DecryptCallback E2EEProcessor::CreateDecryptCallback(
+    const ClientSecretKey* key) {
+  return base::BindRepeating(
+      [](const ClientSecretKey* secret_key,
+         const std::string& ciphertext_hex) -> std::optional<std::string> {
+        std::string result;
+        std::string_view remaining = ciphertext_hex;
```
netzenbot, line 176: nit: The decrypt lambda body (~30 lines, while loop, multi-branch hex parsing) is complex enough to warrant extracting into a named free function in the anonymous namespace per [CS-052](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-052). A named function like `DecryptCiphertextWithThinkingTags` would improve readability at the call site.

components/ai_chat/core/browser/engine/e2ee_processor.h
```diff
@@ -33,27 +36,63 @@ class E2EEProcessor {
...
+  virtual void ClearCachedModelAttestations();
 
   // Fetches the model attestation from GET /v1/models/{model_name}/attestation.
   // Resolves immediately from cache if a valid entry exists. On success, caches
   // the result and invokes |callback| with nullopt. On failure, invokes
   // |callback| with the error.
-  void FetchModelAttestation(const std::string& model_name,
-                             FetchModelAttestationCallback callback);
+  virtual void FetchModelAttestation(const std::string& model_name,
+                                     FetchModelAttestationCallback callback);
```
netzenbot, line 71: nit: The comment says "Returns empty string on error" but `EncryptCallback` is typed as `base::RepeatingCallback<std::optional<std::string>(...)>` — on failure the implementation returns `std::nullopt`, not an empty string. The doc should say "Returns nullopt on error" to match the actual contract. ([Method Documentation Should Describe the Contract](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-004))

components/ai_chat/core/browser/engine/conversation_api_v2_client_unittest.cc
DJAndries: can't ASSERT in the callback

components/ai_chat/core/browser/engine/e2ee_processor.cc
DJAndries: that would make `CreateDecryptCallback` a one-liner

components/ai_chat/core/browser/engine/e2ee_processor_unittest.cc
DJAndries: See https://source.chromium.org/search?q=expect_death%20path:components&ss=chromium and you'll find many instances of `EXPECT_DEATH` in non-dedicated test suites

PR #34976, opened 2026-03-25
Fix downloads toolbar button toggle close behavior

Baker-Harrison:
Resolves


mkarolin:
@Baker-Harrison thank you for the submission. Could you please file an issue at https://github.com/brave/brave-browser/issues/ explaining what this is meant to fix with the steps to reproduce the problem and link that issue into this PR. Thank you!

PR #35010, opened 2026-03-26
Cleanup in wallet's KeyedServices factories

### LINKED ISSUE #53961: brave_wallet::IsAllowedForContext uses runtime properties to gate returning nullptr for `BraveWalletServiceFactory::GetServiceForContext`
service factories must always return the same instance of the service or nullptr for the lifetime of the browser session. See https://github.com/brave/brave-core/blob/master/docs/keyed_services.md#keyed-services

supermassive:
Resolves https://github.com/brave/brave-browser/issues/53961

- All code external to wallet when it needs check if walled is "allowed" should check if  `BraveWalletServiceFactory::GetServiceForContext` returns a not null service instance. Or a bit shorter `IsBraveWalletServiceAvailable`.
- `BraveWalletServiceFactory::BuildServiceInstanceForBrowserContext` returns null if for given context its state(prefs, incognito, etc) doesn't allow wallet instance. This is done once per context and result is cached. So for a context `BraveWalletService` either available(and fully operational) or not and it can't change. 
- Fixed tests so `kBraveWalletPrivateWindowsEnabled` and `kBraveWalletDisabledByPolicy` prefs take effect only when profile is being created.

### DIFF
--- a/browser/brave_wallet/BUILD.gn
+++ b/browser/brave_wallet/BUILD.gn
@@ -154,6 +154,7 @@ source_set("unit_tests") {
   sources = [
     "asset_discovery_manager_unittest.cc",
     "asset_discovery_task_unittest.cc",
+    "brave_wallet_service_factory_unittest.cc",
     "brave_wallet_service_unittest.cc",
     "eth_allowance_manager_unittest.cc",
     "eth_pending_tx_tracker_unittest.cc",
@@ -176,6 +177,7 @@ source_set("unit_tests") {
     "//brave/components/brave_wallet/common:rust_lib",
     "//brave/components/brave_wallet/common:test_support",
     "//brave/components/resources:strings_grit",
+    "//brave/components/tor/buildflags",
     "//brave/components/version_info",
     "//chrome/browser",
     "//chrome/browser/permissions",
@@ -183,6 +185,14 @@ source_set("unit_tests") {
     "//chrome/test:test_support",
     "//testing/gtest",
   ]
+
+  if (enable_tor) {
+    deps += [
+      "//brave/browser/tor",
+      "//brave/components/tor",
+    ]
+  }
+
   if (!is_android) {
     sources += [
       "brave_wallet_tab_helper_unittest.cc",
--- a/browser/brave_wallet/brave_wallet_context_utils.cc
+++ b/browser/brave_wallet/brave_wallet_context_utils.cc
@@ -5,34 +5,13 @@
 
 #include "brave/browser/brave_wallet/brave_wallet_context_utils.h"
 
-#include "brave/components/brave_wallet/browser/brave_wallet_utils.h"
-#include "brave/components/brave_wallet/browser/pref_names.h"
-#include "chrome/browser/profiles/profile.h"
-#include "components/prefs/pref_service.h"
-#include "components/user_prefs/user_prefs.h"
+#include "brave/browser/brave_wallet/brave_wallet_service_factory.h"
 #include "content/public/browser/browser_context.h"
 
 namespace brave_wallet {
 
-bool IsAllowedForContext(content::BrowserContext* context) {
-  if (!context || context->IsTor()) {
-    return false;
-  }
-
-  auto* prefs = user_prefs::UserPrefs::Get(context);
-  if (!IsAllowed(prefs)) {
-    return false;
-  }
-
-  if (Profile::FromBrowserContext(context)->IsRegularProfile()) {
-    return true;
-  }
-
-  if (Profile::FromBrowserContext(context)->IsIncognitoProfile()) {
-    return prefs->GetBoolean(kBraveWalletPrivateWindowsEnabled);
-  }
-
-  return false;
+bool IsBraveWalletServiceAvailable(content::BrowserContext* context) {
+  return !!BraveWalletServiceFactory::GetServiceForContext(context);
 }
 
 }  // namespace brave_wallet
--- a/browser/brave_wallet/brave_wallet_service_factory.cc
+++ b/browser/brave_wallet/brave_wallet_service_factory.cc
@@ -6,15 +6,13 @@
 #include "brave/browser/brave_wallet/brave_wallet_service_factory.h"
 
 #include <memory>
-#include <utility>
+#include <optional>
 
 #include "base/no_destructor.h"
 #include "brave/browser/brave_wallet/brave_wallet_context_utils.h"
-#include "brave/components/brave_wallet/browser/bitcoin/bitcoin_wallet_service.h"
 #include "brave/components/brave_wallet/browser/brave_wallet_service.h"
 #include "brave/components/brave_wallet/browser/brave_wallet_service_delegate.h"
-#include "brave/components/brave_wallet/browser/json_rpc_service.h"
-#include "brave/components/brave_wallet/browser/tx_service.h"
+#include "brave/components/brave_wallet/browser/brave_wallet_utils.h"
 #include "chrome/browser/browser_process.h"
 #include "chrome/browser/profiles/incognito_helpers.h"
 #include "chrome/browser/profiles/profile.h"
@@ -23,6 +21,24 @@
 #include "content/public/browser/storage_partition.h"
 
 namespace brave_wallet {
+
+namespace {
+std::optional<bool> g_null_in_tests_override;
+
+bool CanBuildWalletServiceInstanceForBrowserContext(
+    content::BrowserContext* context) {
+  if (!context) {
+    return false;
+  }
+
+  auto* profile = Profile::FromBrowserContext(context);
+
+  return CanBuildWalletServiceInstance(
+      user_prefs::UserPrefs::Get(context), profile->IsRegularProfile(),
+      profile->IsIncognitoProfile(), profile->IsTor());
+}
+
+}  // namespace
 
 // static
 BraveWalletServiceFactory* BraveWalletServiceFactory::GetInstance() {
@@ -33,9 +49,6 @@ BraveWalletServiceFactory* BraveWalletServiceFactory::GetInstance() {
 // static
 BraveWalletService* BraveWalletServiceFactory::GetServiceForContext(
     content::BrowserContext* context) {
-  if (!IsAllowedForContext(context)) {
-    return nullptr;
-  }
   return static_cast<BraveWalletService*>(
       GetInstance()->GetServiceForBrowserContext(context, true));
 }
@@ -50,6 +63,10 @@ BraveWalletServiceFactory::~BraveWalletServiceFactory() = default;
 std::unique_ptr<KeyedService>
 BraveWalletServiceFactory::BuildServiceInstanceForBrowserContext(
     content::BrowserContext* context) const {
+  if (!CanBuildWalletServiceInstanceForBrowserContext(context)) {
+    return nullptr;
+  }
+
   return std::make_unique<BraveWalletService>(
       context->GetDefaultStoragePartition()
           ->GetURLLoaderFactoryForBrowserProcess(),
@@ -63,6 +80,10 @@ content::BrowserContext* BraveWalletServiceFactory::GetBrowserContextToUse(
 }
 
 bool BraveWalletServiceFactory::ServiceIsNULLWhileTesting() const {
+  if (g_null_in_tests_override.has_value()) {
+    return g_null_in_tests_override.value();
+  }
+
   // Using BrowserTabStripTracker in BraveWalletServiceDelegate can cause
   // upstream tests to crash if they don't have the correct order of
   // creation/destruction of profiles and calls to
@@ -70,6 +91,13 @@ bool BraveWalletServiceFactory::ServiceIsNULLWhileTesting() const {
   // BrowserTabStripTracker relies on global features. By default, make this
   // service NULL for testing and create it manually in tests when needed.
   return true;
+}
+
+BraveWalletServiceFactory::NotNullForTesting::NotNullForTesting() {
+  g_null_in_tests_override = false;
+}
+BraveWalletServiceFactory::NotNullForTesting::~NotNullForTesting() {
+  g_null_in_tests_override.reset();
 }
 
 }  // namespace brave_wallet
--- a/browser/brave_wallet/notifications/wallet_notification_service_factory.cc
+++ b/browser/brave_wallet/notifications/wallet_notification_service_factory.cc
@@ -28,7 +28,7 @@ WalletNotificationServiceFactory::WalletNotificationServiceFactory()
           "WalletNotificationService",
           BrowserContextDependencyManager::GetInstance()) {
   DependsOn(NotificationDisplayServiceFactory::GetInstance());
-  DependsOn(brave_wallet::BraveWalletServiceFactory::GetInstance());
+  DependsOn(BraveWalletServiceFactory::GetInstance());
 }
 
 WalletNotificationServiceFactory::~WalletNotificationServiceFactory() = default;
@@ -36,6 +36,10 @@ WalletNotificationServiceFactory::~WalletNotificationServiceFactory() = default;
 std::unique_ptr<KeyedService>
 WalletNotificationServiceFactory::BuildServiceInstanceForBrowserContext(
     content::BrowserContext* context) const {
+  if (!IsBraveWalletServiceAvailable(context)) {
+    return nullptr;
+  }
+
   return std::make_unique<WalletNotificationService>(
       BraveWalletServiceFactory::GetServiceForContext(context), context);
 }
@@ -44,11 +48,14 @@ WalletNotificationServiceFactory::BuildServiceInstanceForBrowserContext(
 WalletNotificationService*
 WalletNotificationServiceFactory::GetServiceForContext(
     content::BrowserContext* context) {
-  if (!IsAllowedForContext(context)) {
-    return nullptr;
-  }
   return static_cast<WalletNotificationService*>(
       GetInstance()->GetServiceForBrowserContext(context, true));
+}
+
+content::BrowserContext*
+WalletNotificationServiceFactory::GetBrowserContextToUse(
+    content::BrowserContext* context) const {
+  return context;
 }
 
 }  // namespace brave_wallet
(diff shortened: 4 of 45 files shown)


browser/brave_wallet/android/asset_ratio_service_factory_android.cc
```diff
@@ -16,11 +17,14 @@ static jlong JNI_AssetRatioServiceFactory_GetInterfaceToAssetRatioService(
     JNIEnv* env,
     const base::android::JavaRef<jobject>& profile_android) {
   auto* profile = Profile::FromJavaObject(profile_android);
-  auto pending =
-      brave_wallet::AssetRatioServiceFactory::GetInstance()->GetForContext(
-          profile);
+  if (auto* service =
```
cypt4: What is the case service is not initialized on this call? Maybe we should call CHECK instead?

browser/brave_wallet/brave_wallet_context_utils.cc
goodov, line 32: this is bad. The service should not be runtime-enabled/disabled based on a pref. How about fix this here?

browser/brave_wallet/brave_wallet_context_utils.cc
```diff
@@ -5,34 +5,13 @@
...
-  if (Profile::FromBrowserContext(context)->IsRegularProfile()) {
-    return true;
-  }
-
-  if (Profile::FromBrowserContext(context)->IsIncognitoProfile()) {
-    return prefs->GetBoolean(kBraveWalletPrivateWindowsEnabled);
-  }
-
-  return false;
+bool IsBraveWalletServiceAvailable(content::BrowserContext* context) {
```
bridiver, line 13: Why do we still need this helper? I think it only confuses things and it would be better to just check the factory directly?

browser/brave_wallet/notifications/wallet_notification_service_factory.cc
```diff
@@ -28,14 +28,18 @@ WalletNotificationServiceFactory::WalletNotificationServiceFactory()
...
-  DependsOn(brave_wallet::BraveWalletServiceFactory::GetInstance());
+  DependsOn(BraveWalletServiceFactory::GetInstance());
 }
 
 WalletNotificationServiceFactory::~WalletNotificationServiceFactory() = default;
 
 std::unique_ptr<KeyedService>
 WalletNotificationServiceFactory::BuildServiceInstanceForBrowserContext(
     content::BrowserContext* context) const {
+  if (!IsBraveWalletServiceAvailable(context)) {
```
bridiver, line 39: this logic belongs in GetBrowserContextToUse https://source.chromium.org/chromium/chromium/src/+/main:components/keyed_service/content/browser_context_keyed_service_factory.h;l=134?q=BrowserContextKeyedServiceFactory&ss=chromium

browser/brave_wallet/notifications/wallet_notification_service_factory.cc
```diff
@@ -44,11 +48,14 @@ WalletNotificationServiceFactory::BuildServiceInstanceForBrowserContext(
...
-    return nullptr;
-  }
   return static_cast<WalletNotificationService*>(
       GetInstance()->GetServiceForBrowserContext(context, true));
 }
 
+content::BrowserContext*
+WalletNotificationServiceFactory::GetBrowserContextToUse(
+    content::BrowserContext* context) const {
+  return context;
```
bridiver, line 58: See https://github.com/brave/brave-core/pull/35010/changes#r3100832273 and this is also dangerous and confusing. It makes it look like we want to have a split service where incognito has it's own instance and regular has its own instance. We should follow normal conventions here

browser/brave_wallet/brave_wallet_service_factory.cc
```diff
@@ -72,4 +93,11 @@ bool BraveWalletServiceFactory::ServiceIsNULLWhileTesting() const {
   return true;
 }
 
+BraveWalletServiceFactory::NotNullForTesting::NotNullForTesting() {
```
bridiver, line 96: It should either be null for testing or not null for testing. If it's null for testing and you need it then you need to create it yourself in the test. Please do not create new non-standard conventions for factories.

browser/ui/brave_browser_command_controller_browsertest.cc
```diff
@@ -291,6 +295,17 @@ IN_PROC_BROWSER_TEST_F(BraveBrowserCommandControllerTest,
 #if BUILDFLAG(ENABLE_AI_CHAT)
   EXPECT_FALSE(command_controller->IsCommandEnabled(IDC_TOGGLE_AI_CHAT));
 #endif
+
+#if BUILDFLAG(ENABLE_BRAVE_WALLET)
```
bridiver, line 299: I feel like we're making a lot of changes that should be split up into several PRs and the PR title reflects the ambiguity of this PR. The first change should fix the primary issue which is to split out the pref from check from the creation of the service. Other refactoring or changes that are not strictly necessary (like renaming the helper (or better yet getting rid of it), etc...) should move to follow-up PRs. It's really hard to understand the changes when we're doing so many things at once.

browser/brave_wallet/brave_wallet_service_factory.cc
```diff
@@ -24,6 +22,24 @@
...
+bool CanBuildWalletServiceInstanceForBrowserContext(
+    content::BrowserContext* context) {
+  if (!context) {
+    return false;
+  }
+
+  auto* profile = Profile::FromBrowserContext(context);
+
+  return CanBuildWalletServiceInstance(
+      user_prefs::UserPrefs::Get(context), profile->IsRegularProfile(),
```
bridiver, line 37: this still connects the wallet null/not null to the pref, is the intention to have a follow-up for ios? If so, please create an issue and link it here.

browser/brave_wallet/brave_wallet_context_utils.cc
bridiver: I thought fixing that was the main point of this PR? If not I'm really confused. We need to make this factory follow standard conventions, not make it even more unconventional with new non-standard conventions

browser/brave_wallet/notifications/wallet_notification_service_factory.cc
supermassive: `GetBrowserContextToUse` doesn't cache result.
`BuildServiceInstanceForBrowserContext` also allows null for ineligible context, but  caches result. That's the behavior we want for pref based conditions so mapping `context`->`service?` is resolved once and never changes for lifetime of context
     
Chromium also does that in this way, some of many pref based factories: 
https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/ui/hats/trust_safety_sentiment_service_factory.cc;l=57-94
https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/safe_browsing/gemini_antiscam_protection/gemini_antiscam_protection_service_factory.cc;l=46-77

browser/brave_wallet/notifications/wallet_notification_service_factory.cc
supermassive: why it is dangerous? This is wallet related keyed service and should follow wallet's main service behavior which has its' own instance for incogntio.
In fact it should not be a separate KeyedService, I'll remove it in a followup

bridiver: This is the problem I mentioned in another comment which is that we're changing too many things in one PR and it confuses the issue. If this should be own instance in incognito to match other services, that should be in its own PR and not part of the PR to fix the pref issue

browser/brave_wallet/notifications/wallet_notification_service_factory.cc
bridiver: oic, the fix we're making here is to require restart for those prefs? Maybe we talked about that and I forgot. It's a reasonable short-term fix, but we need a follow-up issue to decouple the pref from service creation entirely

browser/brave_wallet/brave_wallet_service_factory.cc
supermassive: It can't be enabled in tests by default because of upstream tests: https://github.com/brave/brave-core/commit/da604768cd894eb068da65c6f96b33f3fa9e07dc
I don't want copypaste construction.

Chromium does this when it's required:
https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/ash/app_list/app_list_syncable_service_factory.cc;l=110-116
https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/safe_browsing/ohttp_key_service_factory.h;l=58-66
https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/ash/net/alwayson_vpn_pre_connect_url_allowlist_service_factory.cc;l=71-79

browser/brave_wallet/brave_wallet_context_utils.cc
supermassive: Ok, I'll remove

browser/brave_wallet/brave_wallet_service_factory.cc
supermassive: ios change is already included in this PR.

bridiver: I guess I misunderstood and didn't realize that the fix (for now) is to require restart, but if we're still tying the pref to service creation then it seems like we have a lot of unnecessary changes in this PR?

bridiver: if we're requiring restart for the prefs, that is literally the only change that is required here (plus a follow-up issue to fix for real). The rest of the refactoring and other changes should be separate, but the parts flagged as problematic patterns should be left out completely.

browser/ui/brave_browser_command_controller_browsertest.cc
supermassive: This is all very entangled. I can't split it into substantially smaller chunks without losing correctness of code.  
`NotNullForTesting` and `decentralized_dns` could be moved out, but I don't think it will make PR much easier to review.

supermassive: I've already asked goodov (and we already discussed some things locally like `BuildServiceInstanceForBrowserContext`) so maybe it makes sense to delegate review to him

browser/brave_wallet/brave_wallet_service_factory.cc
bridiver: > I don't want copypaste construction.

what is the difference between copypaste construction and copypaste `BraveWalletServiceFactory::NotNullForTesting`? You shouldn't be using factories in unit tests in the first place except for any unit tests for the factory itself

browser/ui/brave_browser_command_controller_browsertest.cc
bridiver: > This is all very entangled. I can't split it into substantially smaller chunks without losing correctness of code.

There are a lot of changes here that are refactoring and a lot of this seems totally unrelated to the original problem. If we're not actually decoupling the pref from the service creation and just working around it by requiring a restart then almost none of this is necessary in this PR

bridiver: things like https://github.com/brave/brave-core/pull/35010#discussion_r3101573181 (if they were necessary for this) should be fixed as PRs before this one, not combined into one PR that fixes many problems.

browser/brave_wallet/brave_wallet_service_factory.cc
supermassive: Lots of code assume that wallet avaliablity can be changed at context's runtime.
This PR changes that prefs are read only once when service is built. So tests are changed not only to set a pref, but also run a new profile

browser/ui/brave_browser_command_controller_browsertest.cc
supermassive: I'll try to extract whatever I can from this PR, but it will not get much smaller

browser/brave_wallet/brave_wallet_service_factory.cc
supermassive: It's this
```
  BraveWalletServiceFactory::NotNullForTesting
      brave_wallet_service_factory_not_null_for_testing_;
```
vs this
```
    DCHECK(g_browser_process->local_state());
    brave_wallet::BraveWalletServiceFactory::GetInstance()->SetTestingFactory(
        browser_context(),
        base::BindLambdaForTesting([&](content::BrowserContext* context)
                                       -> std::unique_ptr<KeyedService> {
          return std::make_unique<brave_wallet::BraveWalletService>(
              test_url_loader_factory_.GetSafeWeakWrapper(),
              brave_wallet::BraveWalletServiceDelegate::Create(context),
              user_prefs::UserPrefs::Get(context),
              g_browser_process->local_state());
        }));
```

browser/net/BUILD.gn
```diff
@@ -113,4 +113,15 @@ source_set("unit_tests") {
     sources += [ "search_ads_header_network_delegate_helper_unittest.cc" ]
     deps += [ "//brave/components/brave_ads/core" ]
   }
+
+  if (enable_brave_wallet) {
```
netzenbot, line 117: Missing `import("//brave/components/brave_wallet/common/buildflags/buildflags.gni")` at the top of this file. `enable_brave_wallet` is not available via any of the existing imports (`brave_ads`, `tor`, `features`), so GN will fail with an undefined identifier error when evaluating `if (enable_brave_wallet)`. ([✅ Unconditional Buildflags Deps with Conditional `deps +=`](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-027))

browser/brave_wallet/BUILD.gn
```diff
@@ -176,13 +177,22 @@ source_set("unit_tests") {
...
     "//brave/components/brave_wallet/common:test_support",
     "//brave/components/resources:strings_grit",
+    "//brave/components/tor/buildflags",
     "//brave/components/version_info",
     "//chrome/browser",
     "//chrome/browser/permissions",
     "//chrome/browser/prefs",
     "//chrome/test:test_support",
     "//testing/gtest",
   ]
```
netzenbot, line 187: Missing `import("//brave/components/tor/buildflags/buildflags.gni")` at the top of this file. `enable_tor` is not transitively available from the existing imports, so GN will fail with an undefined identifier when evaluating `if (enable_tor)` in the `unit_tests` target. ([✅ Unconditional Buildflags Deps with Conditional `deps +=`](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-027))

components/decentralized_dns/content/decentralized_dns_navigation_throttle.h
```diff
@@ -13,6 +13,10 @@
```
netzenbot, line 13: `DecentralizedDnsNavigationThrottle::MaybeCreateAndAdd` now takes `brave_wallet::BraveWalletService*`, adding a direct cross-feature dependency from `decentralized_dns/content` on `brave_wallet/browser`. The pointer is only null-checked, not dereferenced — passing a `bool` (or a simple `std::optional<bool>`) would achieve the same result without coupling the component to the wallet's browser-layer class hierarchy. See [ARCH-053](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-053).

browser/ui/toolbar/brave_app_menu_model_browsertest.cc
```diff
@@ -50,6 +55,16 @@ class BraveAppMenuModelBrowserTest : public InProcessBrowserTest {
 #endif
   }
 
+  void SetUpOnMainThread() override {
+    InProcessBrowserTest::SetUpOnMainThread();
+
+#if BUILDFLAG(ENABLE_BRAVE_WALLET)
+    brave_wallet::BraveWalletServiceFactory::GetServiceForContext(
```
netzenbot, line 62: The return value of `BraveWalletServiceFactory::GetServiceForContext()` is dereferenced directly without a null check. The VPN service in this same file uses `ASSERT_TRUE(!!service)` before dereferencing — apply the same pattern here to get a clear failure message rather than a crash if the service is unexpectedly null. See [CSM-029](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-memory.md#CSM-029).

browser/decentralized_dns/test/decentralized_dns_navigation_throttle_unittest.cc
```diff
@@ -62,11 +64,21 @@ class DecentralizedDnsNavigationThrottleTest : public testing::Test {
         /*create_if_needed=*/true);
   }
 
+  brave_wallet::BraveWalletService* brave_wallet_service() {
+    auto* wallet_service =
+        brave_wallet::BraveWalletServiceFactory::GetServiceForContext(
+            profile());
```
netzenbot, line 70: The `brave_wallet_service()` helper uses `EXPECT_TRUE(wallet_service)` but then returns the (potentially null) pointer. If `wallet_service` is null, the test continues and the null is passed to `MaybeCreateAndAdd`, causing confusing failures downstream. Use `ASSERT_TRUE(wallet_service)` so the helper aborts on null, or add an ASSERT_TRUE check at the call site in the test body instead. ([Use ASSERT for Preconditions That Subsequent Code Depends On](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-034))

browser/brave_wallet/brave_wallet_service_factory.cc
bridiver: and you can't extract the second one into some kind of helper?

browser/brave_wallet/brave_wallet_service_factory.cc
bridiver: > This PR changes that prefs are read only once when service is built.

Yes, and it does so in a way that seems way more complicated than necessary. The entire purpose of doing the workaround for this was to avoid making a lot of changes. This can be done in a much simpler was as discussed in slack

browser/brave_wallet/brave_wallet_service_factory.cc
bridiver: so this is actually a good example of the problem here because this unit test shouldn't be in brave/browser in the first place and it's only there because we're pulling in browser dependencies like the factory https://github.com/brave/brave-core/pull/35010/changes#diff-d1b246fbcd581dda225672687977215b66944d47bf3fff5eb724d921fb283705R80

If you have a browser test you can use SetUpInProcessBrowserTestFixture and unit tests shouldn't be using factories in the first place

PR #35011, opened 2026-03-26
[Wallet] Implement BraveWalletHidChooser for ledger subframe

### LINKED ISSUE #54752: [Wallet] Ledger hw accounts should not require device chooser.
When Ledger js requests access to a device we should grant permissions to a first available device w/o showing chooser dialog.

supermassive:
Resolves https://github.com/brave/brave-browser/issues/54752

Sec review: https://github.com/brave/reviews/issues/2200

Implement `BraveWalletHidChooser` which replaces chromiums `HidChooser` on ledger subframe of wallet webui top origins. `BraveWalletHidChooser` silently grants permissions to first Ledger device available when request comes from `chrome-untrusted://ledger-bridge` subframe of wallet webui mainframe(either page or panel)

## Test plan
Interaction with ledger device(create HW account or sign TX) should not open device picker. First available ledger device is picked automatically.


browser/brave_wallet/brave_wallet_hid_chooser.cc
```diff
@@ -0,0 +1,80 @@
...
+#include <utility>
+
+#include "brave/components/constants/webui_url_constants.h"
+#include "chrome/browser/ui/hid/hid_chooser_controller.h"
+#include "content/public/browser/render_frame_host.h"
+#include "content/public/common/url_constants.h"
+
+namespace {
+
+constexpr uint16_t kLedgerVendorId = 0x2c97;
```
kdenhartog, line 19: out of curiosity how did we get this?

@thypon have you heard of anyone doing PoCs where someone could connect an imitating device like a flipper zero and use it to pass back inputs we'd normal consider semi-safe when they're not?

browser/brave_wallet/brave_wallet_hid_chooser.cc
```diff
@@ -0,0 +1,80 @@
...
+}  // namespace
+
+// static
+bool BraveWalletHidChooser::IsBraveWalletMainFrameOrigin(
+    content::RenderFrameHost* rfh) {
+  if (!rfh) {
+    return false;
+  }
+
+  const auto& origin = rfh->GetMainFrame()->GetLastCommittedOrigin();
```
kdenhartog: comment for myself. Audit with an LLM whether it can come up with an bypasses for this check here.

browser/brave_content_browser_client.h
```diff
@@ -195,6 +197,10 @@ class BraveContentBrowserClient : public ChromeContentBrowserClient {
 
   bool AllowSignedExchange(content::BrowserContext* context) override;
 
+#if !BUILDFLAG(IS_ANDROID)
```
kdenhartog, line 200: did we previously support hardware wallets on Android and are maintaining compatibility here by doing this or just adding it now?

kdenhartog (commented):
Mostly looks good based on static review so far. Going to run some automated checks on it now and dynamically test it then will approve.

browser/brave_content_browser_client.h
supermassive: We don't support HW wallets on android and also this define comes from chromium https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/chrome_content_browser_client.h;l=777-785;drc=8d1a1e17ddf1bc1ec3c6a7db51426be9daf251a7

browser/brave_wallet/brave_wallet_hid_chooser.cc
supermassive: This comes from ledger repo(added link) and is what actually used by their js lib for filtering(see link below)

browser/brave_wallet/BUILD.gn
```diff
@@ -149,6 +149,33 @@ source_set("tab_helper") {
   }
```
netzenbot, line 149: nit: `//brave/components/constants` is a deprecated legacy directory — [BS-050](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-050) says to never add new deps on it. The `kWalletPageHost`/`kWalletPanelHost` constants used in `brave_wallet_hid_chooser.cc` should ideally live in a `brave_wallet/common` target. If moving them is out of scope for this PR, leave a TODO.

browser/brave_wallet/brave_wallet_hid_chooser_browsertest.cc
```diff
@@ -0,0 +1,172 @@
...
+
+  void SetUpOnMainThread() override {
+    InProcessBrowserTest::SetUpOnMainThread();
+
+    embedded_test_server()->ServeFilesFromSourceDirectory("chrome/test/data");
+    ASSERT_TRUE(embedded_test_server()->Start());
+
+    mojo::PendingRemote<device::mojom::HidManager> hid_manager;
+    hid_manager_.Bind(hid_manager.InitWithNewPipeAndPassReceiver());
+
```
netzenbot, line 56: The `base::RunLoop` + `BindLambdaForTesting` + `Quit()` pattern should be replaced with `base::test::TestFuture`. `SetHidManagerForTesting` takes a callback that receives `std::vector<device::mojom::HidDeviceInfoPtr>`, so `TestFuture<std::vector<device::mojom::HidDeviceInfoPtr>> future; ..SetHidManagerForTesting(std::move(hid_manager), future.GetCallback()); future.Get();` is cleaner and the preferred approach. See [TA-009](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-async.md#TA-009).

browser/brave_wallet/brave_wallet_hid_chooser_unittest.cc
```diff
@@ -0,0 +1,214 @@
...
+      delete;
+  ~BraveWalletHidChooserTest() override = default;
+
+  void SetUp() override {
+    ChromeRenderViewHostTestHarness::SetUp();
+
+    mojo::PendingRemote<device::mojom::HidManager> hid_manager;
+    hid_manager_.Bind(hid_manager.InitWithNewPipeAndPassReceiver());
+
+    base::RunLoop run_loop;
```
netzenbot: Same `base::RunLoop` + `BindLambdaForTesting` + `Quit()` pattern as in the browsertest. Replace with `base::test::TestFuture<std::vector<device::mojom::HidDeviceInfoPtr>> future; HidChooserContextFactory::GetForProfile(profile())->SetHidManagerForTesting(std::move(hid_manager), future.GetCallback()); future.Get();` per [TA-009](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-async.md#TA-009).

browser/brave_wallet/BUILD.gn
supermassive: This will be done in a followup

browser/brave_wallet/brave_wallet_hid_chooser_browsertest.cc
supermassive: fixed

browser/brave_wallet/brave_wallet_hid_chooser_unittest.cc
supermassive: fixed

browser/brave_wallet/brave_wallet_hid_chooser.cc
kdenhartog: thanks for the comment. Seems fine for now

browser/brave_wallet/BUILD.gn
```diff
@@ -149,6 +149,33 @@ source_set("tab_helper") {
...
+    sources = [
+      "brave_wallet_hid_chooser.cc",
+      "brave_wallet_hid_chooser.h",
+    ]
+    deps = [
+      "//brave/components/constants",
+      "//chrome/browser/hid",
+      "//chrome/browser/ui/hid",
+      "//services/device/public/cpp/hid",
+    ]
```
netzenbot, line 163: nit: `brave_wallet_hid_chooser.h` publicly inherits from `content::HidChooser` and `permissions::ChooserController::View`, so the targets providing those headers (`//content/public/browser` via `//chrome/browser/ui/hid`, `//components/permissions` via `//chrome/browser/ui/hid`) need to be in `public_deps`, not `deps`. See [BS-024](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-024).

browser/brave_wallet/brave_wallet_hid_chooser.h
```diff
@@ -0,0 +1,61 @@
...
+  void OnAdapterEnabledChanged(bool enabled) override;
+  void OnRefreshStateChanged(bool refreshing) override;
+
+  bool is_valid_ledger_request_ = false;
+  std::unique_ptr<HidChooserController> controller_;
+
+  base::WeakPtrFactory<BraveWalletHidChooser> weak_factory_{this};
+};
+
+#endif  // BRAVE_BROWSER_BRAVE_WALLET_BRAVE_WALLET_HID_CHOOSER_H_
```
netzenbot, line 61: nit: `weak_factory_` is declared but never used — no `GetWeakPtr()` call exists in the .cc file. Remove the unused member and the corresponding `#include "base/memory/weak_ptr.h"` per [CS-001](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-001).

browser/hid/brave_hid_delegate.h
```diff
@@ -0,0 +1,34 @@
...
+
+#ifndef BRAVE_BROWSER_HID_BRAVE_HID_DELEGATE_H_
+#define BRAVE_BROWSER_HID_BRAVE_HID_DELEGATE_H_
+
+#include <memory>
+#include <vector>
+
+#include "chrome/browser/hid/chrome_hid_delegate.h"
+#include "content/public/browser/hid_chooser.h"
+#include "content/public/browser/hid_delegate.h"
```
netzenbot, line 14: nit: `content/public/browser/render_frame_host.h` is included for a pointer-only parameter, but `chrome_hid_delegate.h` (already included as the base class header) already provides the `content::RenderFrameHost` forward declaration — the explicit full include is redundant. See [CS-014](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-014).

browser/brave_wallet/brave_wallet_hid_chooser.h
netzenbot, line 61: nit: `is_valid_ledger_request_` is always set in the constructor's member initializer list, so the `= false` default is misleading — per [CS-054](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-054), drop the in-class default and let the constructor own the initialization.

netzenbot, line 61: nit: `IsBraveWalletMainFrameOrigin` is a public static method with no doc comment. It silently handles `nullptr` and checks both wallet page/panel hosts — worth a one-liner describing the contract (e.g. returns true if the outermost main frame is a Brave Wallet chrome:// origin, safe to call with null). See [DOC-004](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-004).

browser/brave_wallet/brave_wallet_hid_chooser.cc
```diff
@@ -0,0 +1,84 @@
...
+  return origin.scheme() == content::kChromeUIScheme &&
+         (origin.host() == kWalletPageHost ||
+          origin.host() == kWalletPanelHost);
+}
```
kdenhartog, line 58: The current check validates only the main frame's origin, which means any subframe embedded under `chrome://wallet` gets the silent auto-grant (e.g. if we have sub-frames for snaps). Could we tighten this to also validate the requesting frame's own origin? That way only the main frame itself and the known `chrome-untrusted://ledger-bridge/` subframe qualify — preserving the defense-in-depth that the `chrome-untrusted://` boundary is meant to provide.

```suggestion
// static
bool BraveWalletHidChooser::IsBraveWalletMainFrameOrigin(
    content::RenderFrameHost* rfh) {
  if (!rfh) {
    return false;
  }

  const auto& main_origin = rfh->GetMainFrame()->GetLastCommittedOrigin();
  if (main_origin.scheme() != content::kChromeUIScheme ||
      (main_origin.host() != kWalletPageHost &&
       main_origin.host() != kWalletPanelHost)) {
    return false;
  }

  // Allow the wallet main frame itself.
  if (rfh == rfh->GetMainFrame()) {
    return true;
  }

  // Allow only the known ledger-bridge untrusted subframe.
  const auto& frame_origin = rfh->GetLastCommittedOrigin();
  return frame_origin.scheme() == content::kChromeUIUntrustedScheme &&
         frame_origin.host() == kUntrustedLedgerHost;
}
```

kdenhartog (approved):
My last comment is a defense in depth check that we should do, but this is G2G IMO

browser/brave_wallet/brave_wallet_hid_chooser.cc
supermassive: Made it so BraveWalletHidChooser is used for wallet webuis, but will select first device only on ledger subframe, otherwise returns empty device list.

browser/hid/brave_hid_delegate.cc
```diff
@@ -0,0 +1,40 @@
...
+
+std::unique_ptr<content::HidChooser> BraveHidDelegate::RunChooser(
+    content::RenderFrameHost* render_frame_host,
+    std::vector<blink::mojom::HidDeviceFilterPtr> filters,
+    std::vector<blink::mojom::HidDeviceFilterPtr> exclusion_filters,
+    content::HidChooser::Callback callback) {
+  DCHECK(render_frame_host);
+
+#if BUILDFLAG(ENABLE_BRAVE_WALLET)
+  if (BraveWalletHidChooser::IsBraveWalletMainFrameOrigin(render_frame_host)) {
```
goodov: looks like this is called for a subframe. Why this call doesn't include `IsLedgerSubframe` check in this case? Why do we create the wallet hid chooser even if subframe (that we can check right here) is wrong?

browser/brave_wallet/BUILD.gn
```diff
@@ -149,6 +149,37 @@ source_set("tab_helper") {
...
+    public_deps = [
+      "//components/permissions",
+      "//content/public/browser",
+    ]
+  }
+}
+
+source_set("brave_wallet_data_files_installer_delegate_impl") {
```
cypt4: Unrelated changes

browser/brave_wallet/BUILD.gn
```diff
@@ -149,6 +149,37 @@ source_set("tab_helper") {
   }
 }
 
+source_set("brave_wallet_hid_chooser") {
+  if (!is_android) {
```
cypt4: Should checked outside source_set

netzenbot (approved):

goodov (approved):

cypt4 (approved):

PR #35028, opened 2026-03-26
[Sync] [AI Chat] (1 of n) Register AI_CHAT_CONVERSATION data type and UserSelectableType

petemill:
Adds the foundational sync infrastructure for AI Chat conversation history syncing. This registers a new DataType (AI_CHAT_CONVERSATION) and UserSelectableType (kAIChat) in Chromium's sync system, behind the kBraveSyncAIChat feature flag (disabled by default).

Changes:
- New ai_chat_specifics.proto defining AIChatConversationSpecifics and nested message types for entries, events, and associated content
- AI_CHAT_CONVERSATION added to DataType enum (tag 10000001 in EntitySpecifics, which is ~9million higher than chromium's)
- kAIChat added to UserSelectableType (with "Leo AI Conversations" sync settings checkbox in follow-up PR)
- kBraveSyncAIChat feature flag added (disabled by default)
- AI_CHAT_CONVERSATION added to EncryptableUserTypes
- All upstream static_asserts and switch statements updated

Spec:

Series:
- https://github.com/brave/go-sync/pull/427 (server)


chromium_src/components/sync/protocol/proto_visitors.h
```diff
@@ -6,13 +6,56 @@
 #ifndef BRAVE_CHROMIUM_SRC_COMPONENTS_SYNC_PROTOCOL_PROTO_VISITORS_H_
 #define BRAVE_CHROMIUM_SRC_COMPONENTS_SYNC_PROTOCOL_PROTO_VISITORS_H_
 
+#include "brave/components/sync/protocol/ai_chat_specifics.pb.h"
```
netzenbot, line 9: The chromium_src override adds `#include "brave/components/sync/protocol/ai_chat_specifics.pb.h"` directly. Per CSRC-019, chromium_src files must not directly depend on `//brave/components/` targets — the upstream sync protocol target won't have a GN dep on the Brave proto, likely causing `gn check` failures. The dependency should be injected into the upstream target via a `sources.gni` or similar mechanism, rather than included directly from the chromium_src override header. ([chromium_src Must Not Depend on Brave Component Targets](https://github.com/brave/brave-core/tree/master/docs/best-practices/chromium-src-overrides.md#CSRC-019))

patches/components-sync-base-data_type.cc.patch
```diff
@@ -0,0 +1,91 @@
+diff --git a/components/sync/base/data_type.cc b/components/sync/base/data_type.cc
```
netzenbot, line 1: This is a new 91-line multiline patch spanning 8 hunks. Per [PATCH-004](https://github.com/brave/brave-core/tree/master/docs/best-practices/patches.md#PATCH-004), multiline patches should be avoided in favour of `#define` macros or chromium_src overrides. A chromium_src override already exists at `chromium_src/components/sync/base/data_type.cc` — the additional functions modified here should be added there using the `#define` override pattern instead.

patches/components-sync-base-user_selectable_type.cc.patch
```diff
@@ -0,0 +1,38 @@
+diff --git a/components/sync/base/user_selectable_type.cc b/components/sync/base/user_selectable_type.cc
```
netzenbot, line 1: New 38-line multiline patch across 3 hunks. Per [PATCH-004](https://github.com/brave/brave-core/tree/master/docs/best-practices/patches.md#PATCH-004), use `#define` macros or chromium_src overrides instead of multiline patches. A chromium_src override of `components/sync/base/user_selectable_type.cc` would be the correct approach here.

patches/components-sync-service-sync_prefs.cc.patch
```diff
@@ -0,0 +1,31 @@
+diff --git a/components/sync/service/sync_prefs.cc b/components/sync/service/sync_prefs.cc
```
netzenbot, line 1: New 31-line multiline patch across 3 hunks. Per [PATCH-004](https://github.com/brave/brave-core/tree/master/docs/best-practices/patches.md#PATCH-004), use `#define` macros or chromium_src overrides rather than multiline patches. `components/sync/service/sync_prefs.cc` should have a `chromium_src` override for these additions.

chromium_src/components/sync/protocol/proto_visitors.h
petemill: This is already handled — Brave proto sources are appended to the upstream sync_protocol_sources via protocol_sources.gni, in which we've added this `ai_chat_specifics` file in this PR, so the GN dependency exists. The existing BraveSpecificFields
  proto in this same override uses the same pattern. A chromium_src/components/sync/protocol/DEPS with the appropriate include_rules is also included.

patches/components-sync-base-data_type.cc.patch
petemill: I reduced the patch a bit by using chromium_src MACROs for the switch statements. The static_assert and enums I don't think can be done differently.

patches/components-sync-base-user_selectable_type.cc.patch
petemill: I reduced the patch a bit by using chromium_src MACROs for the switch statements. The static_assert and enums I don't think can be done differently.

patches/components-sync-service-sync_prefs.cc.patch
petemill: I reduced the patch a bit by using chromium_src MACROs for the switch statements. The static_assert and enums I don't think can be done differently.

components/sync/protocol/ai_chat_specifics.proto
szilardszaloki, line 1: Why the gaps in the proto field numbers — just grouping/leaving room for future fields?

components/sync/protocol/ai_chat_specifics.proto
```diff
@@ -0,0 +1,86 @@
+// Copyright (c) 2025 The Brave Authors. All rights reserved.
```
szilardszaloki: Nit: 2026 — same for `chromium_src/components/sync/base/user_selectable_type.cc` and `chromium_src/components/sync/service/sync_prefs.cc`.

chromium_src/components/sync/base/user_selectable_type.cc
```diff
@@ -0,0 +1,20 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+// clang-format off
+constexpr char kAIChatTypeName[] = "aiChat";
```
netzenbot, line 6: This new chromium_src override adds behavior to `GetUserSelectableTypeInfo` and `GetUserSelectableTypeFromString` but there's no `chromium_src/components/sync/base/user_selectable_type_unittest.cc` override to cover it. Per [TI-040](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-040), chromium_src overrides should have a corresponding test file that includes the upstream test and adds Brave-specific cases. The upstream `user_selectable_type_unittest.cc` exists and can be included as a base.

chromium_src/components/sync/service/sync_prefs.cc
```diff
@@ -0,0 +1,21 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+// clang-format off
+#define BRAVE_SYNC_PREFS_GET_PREF_NAME_FOR_TYPE \
```
netzenbot: This new chromium_src override modifies `GetPrefNameForType`, `IsTypeSupportedInTransportMode`, and `IsTypeSelectedByDefaultInTransportMode` for `kAIChat`, but there's no `chromium_src/components/sync/service/sync_prefs_unittest.cc` override to verify this behavior. Per [TI-040](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-040), create a chromium_src test override that includes the upstream test file and adds Brave-specific assertions for the new pref name and transport mode behavior.

chromium_src/components/sync/service/sync_prefs.cc
```diff
@@ -0,0 +1,21 @@
...
+    case UserSelectableType::kAIChat: return prefs::internal::kSyncAIChat;
+
+#define BRAVE_SYNC_PREFS_IS_TYPE_SUPPORTED_IN_TRANSPORT_MODE \
+    case UserSelectableType::kAIChat: return true;
```
szilardszaloki: I think "transport mode" here means syncing while signed into a Google account without explicitly enabling Sync. Since Brave Sync uses sync chains, this likely doesn't apply and should be `false`?

chromium_src/components/sync/service/sync_prefs.cc
```diff
@@ -0,0 +1,21 @@
...
+
+#define BRAVE_SYNC_PREFS_IS_TYPE_SELECTED_BY_DEFAULT \
+    case UserSelectableType::kAIChat: return true;
```
szilardszaloki: Given [the above](https://github.com/brave/brave-core/pull/35028/files#r3012291618), I'd expect this to be `false` as well.

patches/components-sync-base-data_type.h.patch
```diff
@@ -0,0 +1,34 @@
...
+--- a/components/sync/base/data_type.h
++++ b/components/sync/base/data_type.h
+@@ -200,7 +200,9 @@ enum DataType {
+   // An accessibility annotation.
+   ACCESSIBILITY_ANNOTATION,
+ 
+-  LAST_USER_DATA_TYPE = ACCESSIBILITY_ANNOTATION,
++  AI_CHAT_CONVERSATION,
++
++  LAST_USER_DATA_TYPE = AI_CHAT_CONVERSATION,
```
szilardszaloki, line 11: Nit: per our patch style, should we collapse this to a single line?

patches/components-sync-base-data_type.h.patch
```diff
@@ -0,0 +1,34 @@
...
+ 
+   // ---- Control Types ----
+   // An object representing a set of Nigori keys.
+@@ -309,7 +311,8 @@ enum class DataTypeForHistograms {
+   kGeminiThread = 77,
+   kThemesIos = 78,
+   kAccessibilityAnnotation = 79,
+-  kMaxValue = kAccessibilityAnnotation,
++  kAIChatConversation = 80,
++  kMaxValue = kAIChatConversation,
```
szilardszaloki: Same here.

chromium_src/components/sync/base/user_selectable_type.cc
szilardszaloki: `UserSelectableTypeTest.GetUserSelectableTypeFromDataType` iterates over `DataTypeSet::All()`, which automatically includes `AI_CHAT_CONVERSATION` once it's added to the `DataType` enum, so the mapping is already covered without any changes. The other tests are just illustrative serialization checks, not meant to be exhaustive. Also, a new syncable type should only be added to `AmbiguousTypes()` if it doesn't map unambiguously to a `UserSelectableType`, so I think we're fine here. If anything, we should enable these tests in Brave.

patches/components-sync-protocol-entity_specifics.proto.patch
```diff
@@ -0,0 +1,20 @@
...
+ import "components/sync/protocol/accessibility_annotation_specifics.proto";
++import "brave/components/sync/protocol/ai_chat_specifics.proto";
+ 
+ message EntitySpecifics {
+   // If a datatype is encrypted, this field will contain the encrypted
+@@ -174,6 +175,7 @@ message EntitySpecifics {
+     SkillSpecifics skill = 1564245;
+     GeminiThreadSpecifics gemini_thread = 1569348;
+     AccessibilityAnnotationSpecifics accessibility_annotation = 1588764;
++    AIChatConversationSpecifics ai_chat_conversation = 2000001;
```
szilardszaloki: Index looks fine, but maybe worth picking a more clearly downstream-only tag while this is still unshipped?

patches/components-sync-protocol-proto_value_conversions_unittest.cc.patch
```diff
@@ -0,0 +1,21 @@
...
+-static_assert(63 == syncer::GetNumDataTypes(),
++static_assert(64 == syncer::GetNumDataTypes(),
+               "When adding a new field, add a DEFINE_SPECIFICS_TO_VALUE_TEST "
+               "for your field below, and optionally a test for the specific "
+               "conversions.");
+@@ -139,6 +139,7 @@ DEFINE_SPECIFICS_TO_VALUE_TEST(contextual_task)
+ DEFINE_SPECIFICS_TO_VALUE_TEST(skill)
+ DEFINE_SPECIFICS_TO_VALUE_TEST(gemini_thread)
+ DEFINE_SPECIFICS_TO_VALUE_TEST(accessibility_annotation)
++DEFINE_SPECIFICS_TO_VALUE_TEST(ai_chat_conversation)
```
szilardszaloki, line 18: Does the test generated from `DEFINE_SPECIFICS_TO_VALUE_TEST(ai_chat_conversation)` run as part of Brave's test suite, or only if we explicitly build `//components:components_unittests`?

patches/components-sync-protocol-nigori_specifics.proto.patch
```diff
@@ -0,0 +1,10 @@
+diff --git a/components/sync/protocol/nigori_specifics.proto b/components/sync/protocol/nigori_specifics.proto
+index 59e1a4d6ee50bdfa6bef1638752e965f0ad8feb4..d090c3034b91329666ffaeab39c149e5e294b8ae 100644
+--- a/components/sync/protocol/nigori_specifics.proto
++++ b/components/sync/protocol/nigori_specifics.proto
+@@ -297,4 +297,5 @@ message NigoriSpecifics {
+   // Current cross user sharing Public-key to expose in server registry
+   // for sharing offers outside account boundary.
+   optional CrossUserSharingPublicKey cross_user_sharing_public_key = 53;
++  optional bool encrypt_ai_chat_conversation = 54;
```
szilardszaloki: `54` works, but since this is an upstream-owned proto, maybe worth picking a value further out to reduce the risk of collisions if upstream adds a field here later?

patches/components-sync-service-sync_user_settings_impl.cc.patch
szilardszaloki, line 1: Should we also remove `AI_CHAT_CONVERSATION` here to opt out of roaming profiles on Windows? This looks like the local-sync unsupported types list.
```cpp
types.Remove(AI_CHAT_CONVERSATION);
```

chromium_src/components/sync/service/sync_prefs.cc
petemill: Yes, transport mode is syncing via just being signed in even when not explicitly enabling sync. It becomes an account-level feature. This does seem to be the default for chromium. For example the new Skills sync type. What I found was that if I return false then I hit a CHECK about the type not being part of kLegacyTypes, which felt deprecated.

For now it seems moot for Brave since we only support 1 type of sync. And we can re-visit if Accounts / Sync v3 affects this. Or if we know we want to use the differentiation then I can patch AI Chat Conversations in to `kLegacyTypes`.

`components/sync/service/data_type_controller.cc`
```
    // New Sync data types must generally support kTransportOnly mode. Only
    // legacy types are still allowed to *not* support it for historical reasons
    // (they will all be either migrated or retired eventually).
    // There are two general ways to go about this:
    // * Single storage: The feature is available to all signed-in users (no
    //   distinction between syncing or non-syncing users), and doesn't support
    //   signed-out users at all - so only a single storage is required.
    //   Examples: SEND_TAB_TO_SELF, SHARING_MESSAGE.
    // * Dual storage: The feature is also available to signed-out users (and
    //   its behavior may or may not differ for syncing vs signed-in-not-syncing
    //   users). Two storages are required to distinguish "local data" from
    //   "account data" and keep them separate. Examples: PASSWORDS, BOOKMARKS.

    // Notes on individual data types:
    // * NIGORI *does* actually support transport-mode, but avoids a separate
    //   delegate, see SyncEngineBackend::LoadAndConnectNigoriController().
    // * BOOKMARKS and READING_LIST: Support is WIP.
    // * PASSWORDS: Already supported on desktop; mobile is WIP.
    // * PREFERENCES in all variants: Support is WIP.
    // * History-related types (HISTORY, HISTORY_DELETE_DIRECTIVES, SESSIONS)
    //   are okay to *not* support transport mode.
    // * APPS/APP_SETTINGS: Deprecated and will eventually be removed.
    // * AUTOFILL/AUTOFILL_PROFILE: Semi-deprecated; will eventually be removed
    //   or replaced by CONTACT_INFO.
    //
    // Note on ChromeOS-Ash: On this platform, the sync machinery always runs in
    // full-sync mode, never transport-mode. So for data types that only exist
    // on this platform, it doesn't matter if they support transport mode or not
    // (this includes PRINTERS, WIFI_CONFIGURATIONS, OS_PREFERENCES,
    // OS_PRIORITY_PREFERENCES, WORKSPACE_DESK, PRINTERS_AUTHORIZATION_SERVERS,
    // COOKIES).
    //
    // All other data types listed here will likely have to be migrated.
    static constexpr DataTypeSet kLegacyTypes = {BOOKMARKS,
                                                 PREFERENCES,
                                                 PASSWORDS,
                                                 AUTOFILL_PROFILE,
                                                 AUTOFILL,
                                                 AUTOFILL_WALLET_METADATA,
                                                 AUTOFILL_WALLET_OFFER,
                                                 THEMES,
                                                 EXTENSIONS,
                                                 SEARCH_ENGINES,
                                                 SESSIONS,
                                                 APPS,
                                                 APP_SETTINGS,
                                                 EXTENSION_SETTINGS,
                                                 HISTORY_DELETE_DIRECTIVES,
                                                 DICTIONARY,
                                                 PRIORITY_PREFERENCES,
                                                 PRINTERS,
                                                 READING_LIST,
                                                 USER_EVENTS,
                                                 WIFI_CONFIGURATIONS,
                                                 WEB_APPS,
                                                 OS_PREFERENCES,
                                                 OS_PRIORITY_PREFERENCES,
                                                 WORKSPACE_DESK,
                                                 HISTORY,
                                                 PRINTERS_AUTHORIZATION_SERVERS,
                                                 NIGORI,
                                                 COOKIES};
```

patches/components-sync-protocol-nigori_specifics.proto.patch
petemill: I'm actually thinking this property isn't required ("Obsolete encryption fields."), since everything is encrypted. I removed it from the sync-side PR and now from here.

patches/components-sync-protocol-proto_value_conversions_unittest.cc.patch
petemill: Yes this runs (and passes) via `npm run test -- components_unittests`. And should run on CI whenever the `CI/run-upstream-tests` PR label is added.

patches/components-sync-protocol-entity_specifics.proto.patch
petemill: By `index` do you mean 2000001 and by tag do you mean `ai_chat_conversation`? I guess I was wondering whether I should prefix everything with `brave_` 😬 ...

szilardszaloki: By both index and tag I meant the tag number (`2000001`). The name looks fine — not sure about adding a `brave_` prefix, but I was just thinking about maybe using a more clearly Brave-specific tag value.

components/sync/protocol/ai_chat_specifics.proto
petemill: Yeah but I don't know really what the best practice is here. I changed it to be sequential

patches/components-sync-service-sync_user_settings_impl.cc.patch
petemill: done

patches/components-sync-protocol-entity_specifics.proto.patch
petemill: What do you suggest? I wanted to leave a big gap to avoid conflicts, but not sure how to do anything more formal than that

patches/chrome-browser-resources-settings_shared-people_page-sync_browser_proxy.ts.patch
```diff
@@ -0,0 +1,32 @@
+diff --git a/chrome/browser/resources/settings_shared/people_page/sync_browser_proxy.ts b/chrome/browser/resources/settings_shared/people_page/sync_browser_proxy.ts
+index 7fa797d298d52b822e048f9a1209e43395799a81..b10c3f377c3b52d172a5e72634588b13bae4fdaf 100644
+--- a/chrome/browser/resources/settings_shared/people_page/sync_browser_proxy.ts
++++ b/chrome/browser/resources/settings_shared/people_page/sync_browser_proxy.ts
+@@ -151,6 +151,9 @@ export interface SyncPrefs {
+   wifiConfigurationsManaged: boolean;
+   wifiConfigurationsRegistered: boolean;
+   wifiConfigurationsSynced: boolean;
++  aiChatManaged: boolean;
```
AlexeyBarabash: Is it possible to make it in one line?

patches/chrome-browser-sync-chrome_sync_controller_builder.cc.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/chrome/browser/sync/chrome_sync_controller_builder.cc b/chrome/browser/sync/chrome_sync_controller_builder.cc
+index 15bc32942d3ce174f8cb0a041006a85c40134fcf..731b6e100fcf1461860e7b4e1e64655c600db945 100644
+--- a/chrome/browser/sync/chrome_sync_controller_builder.cc
++++ b/chrome/browser/sync/chrome_sync_controller_builder.cc
+@@ -411,6 +411,7 @@ ChromeSyncControllerBuilder::Build(syncer::SyncService* sync_service) {
+               sync_service, pref_service_.value()));
+     }
+ #endif  // BUILDFLAG(IS_CHROMEOS)
++    BRAVE_BUILD_CHROME_SYNC_CONTROLLERS
```
AlexeyBarabash: Probably it  should be possible to make this without patch by `chromium_src` as `Build` name is found just once in both `src/chrome/browser/sync/chrome_sync_controller_builder.h`/`src/chrome/browser/sync/chrome_sync_controller_builder.cc`

And btw, where is `BRAVE_BUILD_CHROME_SYNC_CONTROLLERS`? I can't find it at the PR modified files and at the local branch checkout.

patches/chrome-browser-resources-settings_shared-people_page-sync_browser_proxy.ts.patch
AlexeyBarabash: there are also several multiline patches at other places.

patches/components-sync-base-data_type.cc.patch
```diff
@@ -0,0 +1,83 @@
...
+          ACCESSIBILITY_ANNOTATION},
++        BRAVE_SPECIFICS_FIELD_NUMBER_TO_DATA_TYPE
+         // ---- Control Types ----
+         {sync_pb::EntitySpecifics::kNigoriFieldNumber, NIGORI},
+     });
+@@ -318,6 +319,7 @@ void AddDefaultFieldValue(DataType type, sync_pb::EntitySpecifics* specifics) {
+     case ACCESSIBILITY_ANNOTATION:
+       specifics->mutable_accessibility_annotation();
+       break;
++      BRAVE_ADD_DEFAULT_FIELD_VALUE
```
AlexeyBarabash: It seems `BRAVE_ADD_DEFAULT_FIELD_VALUE` can be replaced with a `chromium_src` override, as `AddDefaultFieldValue` has only one occurrence at `src/components/sync/base/data_type.h`/`src/components/sync/base/data_type.cc`

patches/components-sync-base-data_type.cc.patch
```diff
@@ -0,0 +1,83 @@
...
+   }
+ }
+ 
+@@ -460,6 +462,7 @@ int GetSpecificsFieldNumberFromDataType(DataType data_type) {
+       return sync_pb::EntitySpecifics::kGeminiThreadFieldNumber;
+     case ACCESSIBILITY_ANNOTATION:
+       return sync_pb::EntitySpecifics::kAccessibilityAnnotationFieldNumber;
++      BRAVE_GET_SPECIFICS_FIELD_NUMBER_FROM_DATA_TYPE
```
AlexeyBarabash: The same looks for `GetSpecificsFieldNumberFromDataType`/`BRAVE_GET_SPECIFICS_FIELD_NUMBER_FROM_DATA_TYPE`

patches/components-sync-base-data_type.cc.patch
```diff
@@ -0,0 +1,83 @@
...
+       return "Accessibility Annotation";
++      BRAVE_DATA_TYPE_TO_DEBUG_STRING
+   }
+   NOTREACHED();
+ }
+@@ -834,6 +838,7 @@ const char* DataTypeToHistogramSuffix(DataType data_type) {
+       return "GEMINI_THREAD";
+     case ACCESSIBILITY_ANNOTATION:
+       return "ACCESSIBILITY_ANNOTATION";
++      BRAVE_DATA_TYPE_TO_HISTOGRAM_SUFFIX
```
AlexeyBarabash: also should be replacable with `chromium_src` override

patches/components-sync-base-data_type.cc.patch
```diff
@@ -0,0 +1,83 @@
...
+   }
+   // LINT.ThenChange(/tools/metrics/histograms/metadata/sync/histograms.xml:DataTypeHistogramSuffix)
+   NOTREACHED();
+@@ -967,6 +972,7 @@ DataTypeForHistograms DataTypeHistogramValue(DataType data_type) {
+       return DataTypeForHistograms::kGeminiThread;
+     case ACCESSIBILITY_ANNOTATION:
+       return DataTypeForHistograms::kAccessibilityAnnotation;
++      BRAVE_DATA_TYPE_HISTOGRAM_VALUE
```
AlexeyBarabash: this can't be replaced with `chromium_src` override because `DataTypeHistogramValue` is invoked from `DataTypeToStableIdentifier` at the same file

patches/components-sync-base-data_type.cc.patch
```diff
@@ -0,0 +1,83 @@
...
+   }
+   NOTREACHED();
+ }
+@@ -1117,6 +1123,7 @@ const char* DataTypeToStableLowerCaseString(DataType data_type) {
+       return "gemini_thread";
+     case ACCESSIBILITY_ANNOTATION:
+       return "accessibility_annotation";
++      BRAVE_DATA_TYPE_TO_STABLE_LOWER_CASE_STRING
```
AlexeyBarabash: this also can't be replaced with `chromium_src` override, as `DataTypeToStableLowerCaseString` is invoked from `DataTypeToProtocolRootTag`

patches/components-sync-model-model_error.h.patch
```diff
@@ -0,0 +1,14 @@
+diff --git a/components/sync/model/model_error.h b/components/sync/model/model_error.h
+index be8c65925d85dc4382584eaff122dc94f322ed27..a0c09446f516f31e4dc2eae5a9af37b261e7d20a 100644
+--- a/components/sync/model/model_error.h
++++ b/components/sync/model/model_error.h
+@@ -200,7 +200,8 @@ class ModelError {
+     kAutofillValuableMetadataFailedToLoadMetadata = 190,
+     kDataTypeControllerInFailedState = 191,
+     kBookmarksLocalCountExceededLimitAfterInitialMerge = 192,
+-    kMaxValue = kBookmarksLocalCountExceededLimitAfterInitialMerge,
++    kAIChatFailedToLoadMetadata = 193,
```
AlexeyBarabash: probably this patch can be replaced with the approach from https://github.com/brave/brave-core/blob/master/chromium_src/chrome/browser/download/download_commands.h#L13

patches/components-sync-protocol-entity_specifics.proto.patch
szilardszaloki: I don't think there's a formal way to reserve a range here — we can just jump higher (tags go up to [`536,870,911`](https://protobuf.dev/programming-guides/proto2/#assigning)).

components/ai_chat/core/common/features.cc
```diff
@@ -196,4 +196,10 @@ bool IsAIChatDeepResearchEnabled() {
   return base::FeatureList::IsEnabled(features::kAIChatDeepResearch);
 }
 
+BASE_FEATURE(kBraveSyncAIChat, base::FEATURE_DISABLED_BY_DEFAULT);
+
+bool IsBraveSyncAIChatEnabled() {
+  return base::FeatureList::IsEnabled(features::kBraveSyncAIChat);
+}
+
 }  // namespace ai_chat::features
```
netzenbot, line 218: `BASE_FEATURE` requires 3 arguments: the C++ identifier, the string name, and the default state. The existing pattern in this file is `BASE_FEATURE(kAIChatDeepResearch, "AIChatDeepResearch", base::FEATURE_DISABLED_BY_DEFAULT)`. This call is missing the string name argument `"BraveSyncAIChat"` and will not compile.

petemill: This is incorrect. `BASE_FEATURE` has either 2 or 3 arguments and is preferred to use 2 arguments if not using a string that isn't inferred from the feature name.

build/android/bytecode/java/org/brave/bytecode/BraveManageSyncSettingsClassAdapter.java
```diff
@@ -27,5 +27,7 @@ public class BraveManageSyncSettingsClassAdapter extends BraveClassVisitor {
 
         deleteField(sBraveManageSyncSettingsClassName, "mSyncEverything");
         makeProtectedField(sManageSyncSettingsClassName, "mSyncEverything");
+
+        makeProtectedField(sManageSyncSettingsClassName, "mSyncTypeCheckBoxPreferencesMap");
```
AlexeyBarabash: thanks for the Android part

You need also
1. Declare `mSyncTypeCheckBoxPreferencesMap` at `BraveManageSyncSettings` class
2. Update the test around https://github.com/brave/brave-core/blob/master/android/javatests/org/chromium/chrome/browser/BytecodeTest.java#L2313 with `fieldExists` 
3. Update https://github.com/brave/brave-core/blob/master/android/java/apk_for_test.flags#L84

android/java/res/xml/manage_sync_preferences.xml
```diff
@@ -67,6 +67,11 @@
         android:title="@string/sync_apps"
         android:persistent="false"/>
 
+    <org.chromium.components.browser_ui.settings.brave_tricks.checkbox_to_switch.ChromeBaseCheckBoxPreference
+        android:key="sync_ai_chat"
+        android:title="@string/brave_sync_ai_chat"
+        android:persistent="false"/>
```
netzenbot: nit: Add `app:isPreferenceVisible="false"` to the `sync_ai_chat` preference. Since `kBraveSyncAIChat` is `FEATURE_DISABLED_BY_DEFAULT`, the preference will briefly appear before `setVisible(false)` is called in `onCreatePreferences`. ([AND-008](https://github.com/brave/brave-core/tree/master/docs/best-practices/android.md#AND-008))

android/java/org/chromium/chrome/browser/sync/settings/BraveManageSyncSettings.java
```diff
@@ -107,6 +112,17 @@ public void onCreatePreferences(@Nullable Bundle savedInstanceState, String root
             syncPaymentsIntegration.setVisible(false);
```
netzenbot: nit: Add an `assert syncAIChat != null` alongside the null check, consistent with the pattern used for `syncPaymentsIntegration` just above. Since `sync_ai_chat` is Brave-owned XML, a null result would be a Brave-side bug and worth catching early in debug builds. ([AND-009](https://github.com/brave/brave-core/tree/master/docs/best-practices/android.md#AND-009))

browser/resources/settings/br/sync_controls.ts
```diff
@@ -3,17 +3,51 @@
...
       return
     }
-    paymentIntegrationToggle.parentElement.remove()
+    paymentIntegrationToggle.parentElement!.remove()
+
+    // Add AI Chat sync toggle if the feature is enabled.
+    if (loadTimeData.getBoolean('isBraveSyncAIChatEnabled')) {
+      const syncDataTypes =
+        templateContent.querySelector('#sync-data-types')
+      if (syncDataTypes) {
```
netzenbot: nit: The block adding the AI Chat sync toggle is guarded only by `loadTimeData.getBoolean('isBraveSyncAIChatEnabled')` at runtime. Per [BS-044](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-044), WebUI JS code that's specific to an optional feature should also use a `// <if expr="enable_ai_chat">` preprocessor guard to remove it from non-AI-chat builds entirely. Other AI Chat settings code in the same directory (e.g., `settings_menu.ts`, `basic_page.ts`) already follows this pattern.

patches/chrome-browser-sync-chrome_sync_controller_builder.cc.patch
petemill: moved to a different branch where it belongs

patches/components-sync-base-data_type.cc.patch
petemill: removed

patches/components-sync-base-data_type.cc.patch
petemill: removed

patches/components-sync-base-data_type.cc.patch
petemill: removed

build/android/bytecode/java/org/brave/bytecode/BraveManageSyncSettingsClassAdapter.java
petemill: Moved to https://github.com/brave/brave-core/pull/35242 where I addressed these issues

android/java/res/xml/manage_sync_preferences.xml
petemill: Moved to https://github.com/brave/brave-core/pull/35242

android/java/org/chromium/chrome/browser/sync/settings/BraveManageSyncSettings.java
petemill: Moved to https://github.com/brave/brave-core/pull/35242

browser/resources/settings/br/sync_controls.ts
petemill: Moved to https://github.com/brave/brave-core/pull/35242

patches/chrome-browser-resources-settings_shared-people_page-sync_browser_proxy.ts.patch
petemill: Moved to https://github.com/brave/brave-core/pull/35242

chromium_src/components/sync/service/sync_prefs.cc
```diff
@@ -0,0 +1,22 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#define BRAVE_SYNC_PREFS_GET_PREF_NAME_FOR_TYPE \
+  case UserSelectableType::kAIChat:             \
```
netzenbot: nit: This new chromium_src override has no comment explaining its purpose. Per [CSRC-017](https://github.com/brave/brave-core/tree/master/docs/best-practices/chromium-src-overrides.md#CSRC-017), add a brief comment above the `#define` blocks explaining that these macros inject `kAIChat` handling into `GetPrefNameForType`, `IsTypeSupportedInTransportMode`, and `IsTypeSelectedByDefaultInTransportMode`.

chromium_src/components/sync/base/user_selectable_type_unittest.cc
```diff
@@ -0,0 +1,29 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include <components/sync/base/user_selectable_type_unittest.cc>
+
+namespace syncer {
```
netzenbot, line 8: nit: `result` is never modified after initialization — prefer `const auto result` per [CS-064](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-064). Same applies to the `result` variable in `AIChatFromString` below.

chromium_src/components/sync/service/sync_prefs_unittest.cc
```diff
@@ -0,0 +1,18 @@
...
+
+#include <components/sync/service/sync_prefs_unittest.cc>
+
+namespace syncer {
+namespace {
+
+TEST_F(SyncPrefsTest, AIChatPrefName) {
+  EXPECT_STREQ(
+      SyncPrefs::GetPrefNameForTypeForTesting(UserSelectableType::kAIChat),
+      "sync.ai_chat");
```
netzenbot: nit: The string `"sync.ai_chat"` is already defined as `prefs::internal::kSyncAIChat` in `components/sync/base/pref_names.h`, which is already included transitively by the upstream test file. Use the constant instead of duplicating the literal to avoid divergence if the pref name changes. ([TI-009](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-009))

chromium_src/components/sync/service/sync_prefs.cc
petemill: Added comment

chromium_src/components/sync/base/user_selectable_type_unittest.cc
petemill: changed

chromium_src/components/sync/service/sync_prefs_unittest.cc
petemill: replaced with constant

patches/components-sync-base-data_type.h.patch
petemill: condensed to single line

patches/components-sync-base-data_type.h.patch
petemill: condensed to single line

chromium_src/components/sync/service/sync_prefs.cc
petemill: I think this is debatable as per https://github.com/brave/brave-core/pull/35028#discussion_r3013502903 it seems like Chromium expects this to be true moving forward, and prevents adding to `kLegacyTypes`?

patches/components-sync-model-model_error.h.patch
petemill: I'm afraid it won't work. Compile complains that kMaxValue is smaller than the added value. I think it's because `ModelError::Type` is `enum class` whereas `DownloadCommands::Command` is `enum`.

ios/browser/api/sync/driver/brave_sync_profile_service.mm
```diff
@@ -19,7 +19,7 @@
 #error "This file requires ARC support."
 #endif
 
-static_assert(static_cast<NSInteger>(syncer::UserSelectableType::kCookies) ==
+static_assert(static_cast<NSInteger>(syncer::UserSelectableType::kAIChat) ==
                   static_cast<NSInteger>(syncer::UserSelectableType::kLastType),
               "syncer::UserSelectableType has changed in a Chromium update");
```
petemill, line 24: does this avoid the purpose of this assert @brave/chromium-rebase-team? Since we're modifying `UserSelectableType::kLastType` elsewhere, this probably won't be the place to catch it anymore...

patches/build-config-BUILDCONFIG.gn.patch
```diff
@@ -1,8 +1,16 @@
 diff --git a/build/config/BUILDCONFIG.gn b/build/config/BUILDCONFIG.gn
-index 4d1ac59b9b7dc05a20a3bb523b17706a00761cde..3beaf22c51b7cc2b7c0a1a8c4968f68d572b3cd1 100644
+index 4d1ac59b9b7dc05a20a3bb523b17706a00761cde..36c34878684df5f54187144c1d2e1be3b7160fad 100644
 --- a/build/config/BUILDCONFIG.gn
 +++ b/build/config/BUILDCONFIG.gn
-@@ -343,6 +343,7 @@ is_posix = !is_win && !is_fuchsia
+@@ -241,7 +241,6 @@ _default_toolchain = ""
+ 
+ if (target_os == "android") {
```
netzenbot: nit: This patch file now includes two hunks: the existing Brave compiler config addition (line 343), and a new unrelated hunk removing `assert(host_os == "linux", "Android builds are only supported on Linux.")` from the Android toolchain block. Per [ARCH-070](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-070), unrelated changes should be in separate PRs — the Android assertion removal is unrelated to AI Chat sync registration.

chromium_src/components/sync/service/sync_prefs_unittest.cc
```diff
@@ -0,0 +1,18 @@
...
+namespace syncer {
+namespace {
+
+TEST_F(SyncPrefsTest, AIChatPrefName) {
+  EXPECT_STREQ(
+      SyncPrefs::GetPrefNameForTypeForTesting(UserSelectableType::kAIChat),
+      prefs::internal::kSyncAIChat);
+}
+
+}  // namespace
```
netzenbot, line 17: nit: The override in `sync_prefs.cc` adds three behaviors for `kAIChat` (`GetPrefNameForType`, `IsTypeSupportedInTransportMode`, `IsTypeSelectedByDefaultInTransportMode`), but only the pref name is tested here. Consider adding tests for the transport mode methods as well. ([TI-039](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-039))

patches/build-config-BUILDCONFIG.gn.patch
petemill: removed

chromium_src/components/sync/protocol/proto_visitors.h
```diff
@@ -6,13 +6,125 @@
...
-VISIT_PROTO_FIELDS(const sync_pb::BraveSpecificFields& proto) {        \
-  VISIT(is_self_delete_supported);                                     \
-}
+#define BRAVE_VISIT_PROTO_FIELDS_BRAVE_SPECIFIC_FIELD                      \
+  VISIT_PROTO_FIELDS(const sync_pb::BraveSpecificFields& proto) {          \
+    VISIT(is_self_delete_supported);                                       \
+  }                                                                        \
+  VISIT_PROTO_FIELDS(const sync_pb::AIChatCompressibleString& proto) {     \
+    VISIT(raw);                                                            \
+    VISIT(gzipped);                                                        \
```
szilardszaloki: `VISIT_BYTES(gzipped)`?

chromium_src/components/sync/protocol/proto_visitors.h
```diff
@@ -6,13 +6,125 @@
...
+    VISIT(url);                                                            \
+    VISIT(content_type);                                                   \
+    VISIT(content_used_percentage);                                        \
+    VISIT(last_contents);                                                  \
+  }                                                                        \
+  VISIT_PROTO_FIELDS(const sync_pb::AIChatUploadedFile& proto) {           \
+    VISIT(filename);                                                       \
+    VISIT(filesize);                                                       \
+    VISIT(type);                                                           \
+    VISIT(data);                                                           \
```
szilardszaloki: `VISIT_BYTES(data)`?

chromium_src/components/sync/base/user_selectable_type_unittest.cc
```diff
@@ -0,0 +1,29 @@
...
+namespace {
+
+TEST_F(UserSelectableTypeTest, AIChatMapsToUserSelectableType) {
+  const auto result = GetUserSelectableTypeFromDataType(AI_CHAT_CONVERSATION);
+  ASSERT_TRUE(result.has_value());
+  EXPECT_EQ(result.value(), UserSelectableType::kAIChat);
+}
```
szilardszaloki, line 14: This can be simplified to:
```cpp
TEST_F(UserSelectableTypeTest, AIChatMapsToUserSelectableType) {
  EXPECT_EQ(GetUserSelectableTypeFromDataType(AI_CHAT_CONVERSATION),
            UserSelectableType::kAIChat);
}
```

chromium_src/components/sync/base/user_selectable_type_unittest.cc
```diff
@@ -0,0 +1,29 @@
...
+TEST_F(UserSelectableTypeTest, AIChatTypeName) {
+  EXPECT_STREQ(GetUserSelectableTypeName(UserSelectableType::kAIChat),
+               "aiChat");
+}
+
+TEST_F(UserSelectableTypeTest, AIChatFromString) {
+  const auto result = GetUserSelectableTypeFromString("aiChat");
+  ASSERT_TRUE(result.has_value());
+  EXPECT_EQ(result.value(), UserSelectableType::kAIChat);
+}
```
szilardszaloki, line 31: Same here:
```cpp
TEST_F(UserSelectableTypeTest, AIChatFromString) {
  EXPECT_EQ(GetUserSelectableTypeFromString("aiChat"),
            UserSelectableType::kAIChat);
}
```

ios/browser/api/sync/driver/brave_sync_profile_service.mm
szilardszaloki, line 24: Probably add something like:
```cpp
static_assert(
    static_cast<NSInteger>(syncer::UserSelectableType::kCookies) ==
        static_cast<NSInteger>(syncer::UserSelectableType::kAIChat) - 1,
    "...");
```
below the `kLastType` check?

szilardszaloki (approved):
LGTM % a few comments.

chromium_src/components/sync/base/user_selectable_type_unittest.cc
```diff
@@ -0,0 +1,29 @@
...
+
+}  // namespace
```
darkdh, line 33: shouldn't we also need to test `ValueListToUserSelectableTypeSet` and `UserSelectableTypeSetToValueList`?

patches/components-sync-base-pref_names.h.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/components/sync/base/pref_names.h b/components/sync/base/pref_names.h
+index 343d61863e8787db1edf3bc3c311bf31e0dcf598..458def6df0c05459c198b88790e3b517e7425cb0 100644
+--- a/components/sync/base/pref_names.h
++++ b/components/sync/base/pref_names.h
+@@ -89,6 +89,7 @@ inline constexpr char kSyncReadingList[] = "sync.reading_list";
+ inline constexpr char kSyncSavedTabGroups[] = "sync.saved_tab_groups";
+ inline constexpr char kSyncTabs[] = "sync.tabs";
+ inline constexpr char kSyncThemes[] = "sync.themes";
++inline constexpr char kSyncAIChat[] = "sync.ai_chat";
```
darkdh: this simple addition can be done through chromium_src

chromium_src/components/sync/protocol/proto_visitors.h
petemill: thanks - fixed

chromium_src/components/sync/protocol/proto_visitors.h
petemill: thanks - fixed

chromium_src/components/sync/base/user_selectable_type_unittest.cc
petemill: nice - fixed

chromium_src/components/sync/base/user_selectable_type_unittest.cc
petemill: fixed

ios/browser/api/sync/driver/brave_sync_profile_service.mm
petemill: good idea, added

chromium_src/components/sync/base/user_selectable_type_unittest.cc
petemill: Added `AIChatRoundTripsThroughValueList` which tests `UserSelectableTypeSetToValueList` and `ValueListToUserSelectableTypeSet`

patches/components-sync-base-pref_names.h.patch
petemill: good point, moved - https://github.com/brave/brave-core/pull/35028/changes#diff-21817aa27fd12503e46241d20e328b8b24a0908ec3b07196ee6b9f6b6b1a1eb2

patches/components-sync-base-data_type.h.patch
petemill: Changed back to 2 lines because `components/sync/base/data_type.h` is a `GENERATED_JAVA_ENUM_PACKAGE` source feeding `java_cpp_enum("java_enums")`. The generator parses one entry per line and discards anything after the first comma on a line — so `AI_CHAT_CONVERSATION, LAST_USER_DATA_TYPE = …` on one line would silently drop `LAST_USER_DATA_TYPE` (and likewise `kMaxValue`) from the generated Java enum. Splitting the lines preserves both.

patches/components-sync-base-data_type.h.patch
```diff
@@ -0,0 +1,24 @@
...
+ 
+   // ---- Control Types ----
+   // An object representing a set of Nigori keys.
+@@ -314,7 +315,8 @@ enum class DataTypeForHistograms {
+   kThemesIos = 78,
+   kAccessibilityAnnotation = 79,
+   kThemesAndroid = 80,
+-  kMaxValue = kThemesAndroid,
++  kAIChatConversation = 81,
++  kMaxValue = kAIChatConversation,
```
petemill, line 21: note this is a is a `GENERATED_JAVA_ENUM_PACKAGE` source feeding `java_cpp_enum("java_enums")`. The generator parses one entry per line and discards anything after the first comma on a line — so merging to 1 line would silently drop  `kMaxValue` from the generated Java enum. Splitting the lines preserves both.

patches/components-sync-base-user_selectable_type.h.patch
```diff
@@ -0,0 +1,14 @@
...
+index c42bdf12fa2659c61c128b10bc125fbc926f4126..5de574a9e8930a4dd3e0aec5ccc48c36cb922078 100644
+--- a/components/sync/base/user_selectable_type.h
++++ b/components/sync/base/user_selectable_type.h
+@@ -34,7 +34,8 @@ enum class UserSelectableType {
+   kPayments,
+   kProductComparison,
+   kCookies,
+-  kLastType = kCookies
++  kAIChat,
++  kLastType = kAIChat
```
petemill, line 11: note this is a is a `GENERATED_JAVA_ENUM_PACKAGE` source feeding `java_cpp_enum("java_enums")`. The generator parses one entry per line and discards anything after the first comma on a line — so merging to 1 line would silently drop  `kLastType` from the generated Java enum. Splitting the lines preserves both.

chromium_src/components/sync/service/sync_prefs.cc
```diff
@@ -0,0 +1,24 @@
...
+
+// Inject kAIChat handling into GetPrefNameForType,
+// IsTypeSupportedInTransportMode, and IsTypeSelectedByDefaultInTransportMode.
+#define BRAVE_SYNC_PREFS_GET_PREF_NAME_FOR_TYPE \
+  case UserSelectableType::kAIChat:             \
+    return prefs::internal::kSyncAIChat;
+
+#define BRAVE_SYNC_PREFS_IS_TYPE_SUPPORTED_IN_TRANSPORT_MODE \
+  case UserSelectableType::kAIChat:                          \
+    return true;
```
darkdh: I don't think Brave suppport transport mode(user signed in but sync is off) so this should be false.
cc @AlexeyBarabash to double check

chromium_src/components/sync/service/sync_prefs.cc
```diff
@@ -0,0 +1,24 @@
...
+
+#define BRAVE_SYNC_PREFS_IS_TYPE_SELECTED_BY_DEFAULT \
+  case UserSelectableType::kAIChat:                  \
+    return true;
```
darkdh: shouldn't this return if the aichat sync feature is enabled?

chromium_src/components/sync/service/sync_prefs.cc
petemill: I thought it's best to leave this as `true` for the reasons described at https://github.com/brave/brave-core/pull/35028#discussion_r3013502903 

> Yes, transport mode is syncing via just being signed in even when not explicitly enabling sync. It becomes an account-level feature. This does seem to be the default for chromium. For example the new Skills sync type. What I found was that if I return false then I hit a CHECK about the type not being part of kLegacyTypes, which felt deprecated.
>
>For now it seems moot for Brave since we only support 1 type of sync. And we can re-visit if Accounts / Sync v3 affects this. Or if we know we want to use the differentiation then I can patch AI Chat Conversations in to `kLegacyTypes`.
>
>`components/sync/service/data_type_controller.cc`
>
> ```
>     // New Sync data types must generally support kTransportOnly mode. Only
>     // legacy types are still allowed to *not* support it for historical reasons
>     // (they will all be either migrated or retired eventually).
>     // There are two general ways to go about this:
>     // * Single storage: The feature is available to all signed-in users (no
>     //   distinction between syncing or non-syncing users), and doesn't support
>     //   signed-out users at all - so only a single storage is required.
>     //   Examples: SEND_TAB_TO_SELF, SHARING_MESSAGE.
>     // * Dual storage: The feature is also available to signed-out users (and
>     //   its behavior may or may not differ for syncing vs signed-in-not-syncing
>     //   users). Two storages are required to distinguish "local data" from
>     //   "account data" and keep them separate. Examples: PASSWORDS, BOOKMARKS.
> 
>     // Notes on individual data types:
>     // * NIGORI *does* actually support transport-mode, but avoids a separate
>     //   delegate, see SyncEngineBackend::LoadAndConnectNigoriController().
>     // * BOOKMARKS and READING_LIST: Support is WIP.
>     // * PASSWORDS: Already supported on desktop; mobile is WIP.
>     // * PREFERENCES in all variants: Support is WIP.
>     // * History-related types (HISTORY, HISTORY_DELETE_DIRECTIVES, SESSIONS)
>     //   are okay to *not* support transport mode.
>     // * APPS/APP_SETTINGS: Deprecated and will eventually be removed.
>     // * AUTOFILL/AUTOFILL_PROFILE: Semi-deprecated; will eventually be removed
>     //   or replaced by CONTACT_INFO.
>     //
>     // Note on ChromeOS-Ash: On this platform, the sync machinery always runs in
>     // full-sync mode, never transport-mode. So for data types that only exist
>     // on this platform, it doesn't matter if they support transport mode or not
>     // (this includes PRINTERS, WIFI_CONFIGURATIONS, OS_PREFERENCES,
>     // OS_PRIORITY_PREFERENCES, WORKSPACE_DESK, PRINTERS_AUTHORIZATION_SERVERS,
>     // COOKIES).
>     //
>     // All other data types listed here will likely have to be migrated.
>     static constexpr DataTypeSet kLegacyTypes = {BOOKMARKS,
>                                                  PREFERENCES,
>                                                  PASSWORDS,
>                                                  AUTOFILL_PROFILE,
>                                                  AUTOFILL,
>                                                  AUTOFILL_WALLET_METADATA,
>                                                  AUTOFILL_WALLET_OFFER,
>                                                  THEMES,
>                                                  EXTENSIONS,
>                                                  SEARCH_ENGINES,
>                                                  SESSIONS,
>                                                  APPS,
>                                                  APP_SETTINGS,
>                                                  EXTENSION_SETTINGS,
>                                                  HISTORY_DELETE_DIRECTIVES,
>                                                  DICTIONARY,
>                                                  PRIORITY_PREFERENCES,
>                                                  PRINTERS,
>                                                  READING_LIST,
>                                                  USER_EVENTS,
>                                                  WIFI_CONFIGURATIONS,
>                                                  WEB_APPS,
>                                                  OS_PREFERENCES,
>                                                  OS_PRIORITY_PREFERENCES,
>                                                  WORKSPACE_DESK,
>                                                  HISTORY,
>                                                  PRINTERS_AUTHORIZATION_SERVERS,
>                                                  NIGORI,
>                                                  COOKIES};
> ```
>

chromium_src/components/sync/service/sync_prefs.cc
petemill: The feature flag will guard the sync controller creation. Without a sync controller, `AI_CHAT_CONVERSATION` will not be in the list of registered data types. But when the flag is enabled, this is the configuration. I don't think it needs the complication of the flag here because it doesn't get used.

It's also moot because this is only used in transport mode anyway! It determines whether, in transport (aka "signing-in opts-in to sync" mode), is this type enabled by default.

I think this macro can have a better name and be explicit about that. I'll change it to `BRAVE_SYNC_PREFS_IS_TYPE_SELECTED_BY_DEFAULT_IN_TRANSPORT_MODE`.

chromium_src/components/sync/service/sync_prefs.cc
```diff
@@ -0,0 +1,24 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+// Inject kAIChat handling into GetPrefNameForType,
+// IsTypeSupportedInTransportMode, and IsTypeSelectedByDefaultInTransportMode.
+#define BRAVE_SYNC_PREFS_GET_PREF_NAME_FOR_TYPE \
+  case UserSelectableType::kAIChat:             \
+    return prefs::internal::kSyncAIChat;
+
+#define BRAVE_SYNC_PREFS_IS_TYPE_SUPPORTED_IN_TRANSPORT_MODE \
```
cdesouza-chromium: When we use plaster eliminate the use of defines, so these the code below for and other defines can just go straight into the plaster.

rewrite/chrome/browser/android/backup/chrome_backup_agent.cc.yaml
```diff
@@ -0,0 +1,13 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+substitutions:
+  - description:
+      "Bump the kLastType static_assert for Brave's extra UserSelectableType
+      items"
+    re_pattern: 'static_assert\(13\s*==\s*static_cast<int>\(syncer::UserSelectableType::kLastType\),'
+    replace:
+      'static_assert(14 ==
```
cdesouza-chromium: I think we should change this and the other cases to be just `+1` added, and of course without matching whatever value is in there so we don't get patch breakages every time a new key is added to upstream.

```cxx
static_assert(13 + 1 == static_cast<int>(syncer::UserSelectableType::kLastType)
```

I think the same applies to the other plasters.

rewrite/components/browsing_data/core/browsing_data_policies_utils_unittest.cc.yaml
```diff
@@ -0,0 +1,13 @@
+# Copyright (c) 2026 The Brave Authors. All rights reserved.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+substitutions:
+  - description:
+      'Add the kAIChat selectable type to the checked sync types in the
+      AllSyncTypesChecked test.'
+    re_pattern: 'syncer::UserSelectableType::kReadingList\};'
```
cdesouza-chromium: This match is fragile and will break if a new key gets added. Would you mind changing it to match `always_enabled_sync_types = {...}` and inserting at the end?

rewrite/components/sync/base/data_type.cc.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+  - description:
+      "Bump the GetNumDataTypes static_assert for Brave's extra data types"
+    re_pattern: 'static_assert\(GetNumDataTypes\(\)\s*==\s*64,'
+    replace: 'static_assert(GetNumDataTypes() == 65,'
+
+  - description:
+      'Add default field values for Brave data types in AddDefaultFieldValue.'
+    re_pattern: 'specifics->mutable_theme_android\(\);(\n[ \t]*)break;'
+    replace: 'specifics->mutable_theme_android();\1break;\1BRAVE_DATA_TYPE_ADD_DEFAULT_FIELD_VALUE'
```
cdesouza-chromium: This should follow other cases where have for switch insertions. See `rewrite/chrome/browser/enterprise/connectors/device_trust/key_management/browser/metrics_utils.cc.toml`

patches/components-sync-base-data_type.h.patch
cdesouza-chromium: Yeah, I'm not sure there's much we can do here for now. Maybe in the future we can come up with something better.

rewrite/components/sync/base/user_selectable_type.cc.yaml
```diff
@@ -0,0 +1,26 @@
...
+substitutions:
+  - description:
+      "Bump the GetNumDataTypes static_assert for Brave's extra data types"
+    re_pattern: 'static_assert\(64\s*==\s*syncer::GetNumDataTypes\(\),'
+    replace: 'static_assert(65 == syncer::GetNumDataTypes(),'
+
+  - description:
+      'Return UserSelectableTypeInfo for Brave selectable types in
+      GetUserSelectableTypeInfo.'
+    re_pattern: '(\n[ \t]*)return \{kCookiesTypeName, COOKIES, \{COOKIES\}\};'
```
cdesouza-chromium: For this and other cases like this take a look on the example plaster I mentioned in the other thread.

rewrite/components/sync/engine/cycle/data_type_tracker.cc.yaml
```diff
@@ -0,0 +1,19 @@
...
+      types.'
+    re_pattern:
+      '(\n[ \t]*)case THEMES_ANDROID:(\n[ \t]*return
+      kMediumLocalChangeNudgeDelay;)'
+    replace: '\1case THEMES_ANDROID:\1case AI_CHAT_CONVERSATION:\2'
+
+  - description:
+      'Group AI_CHAT_CONVERSATION with the data types that cannot get commits
+      from extensions.'
+    re_pattern: '(\n[ \t]*)case THEMES_ANDROID:(\n[ \t]*return false;)'
```
cdesouza-chromium: I think for both substitutions in this repo we should match the function the switch occurs in first, and then the outcome, or return value only, where we need to add the case key.

rewrite/components/sync/protocol/proto_value_conversions_unittest.cc.yaml
```diff
@@ -0,0 +1,17 @@
...
+  - description:
+      "Bump the GetNumDataTypes static_assert for Brave's extra data types"
+    re_pattern: 'static_assert\(64\s*==\s*syncer::GetNumDataTypes\(\),'
+    replace: 'static_assert(65 == syncer::GetNumDataTypes(),'
+
+  - description:
+      'Add the specifics-to-value test for the AI Chat conversation type.'
+    re_pattern: 'DEFINE_SPECIFICS_TO_VALUE_TEST\(accessibility_annotation\)'
+    replace: |-
+      DEFINE_SPECIFICS_TO_VALUE_TEST(accessibility_annotation)
```
cdesouza-chromium: I feel we can make this a bit more resilient by asking to add this line just after whatever is the last line starting with `DEFINE_SPECIFICS_TO_VALUE_TEST`

rewrite/ios/chrome/browser/settings/manage_sync/coordinator/manage_sync_settings_mediator.mm.yaml
```diff
@@ -0,0 +1,14 @@
...
+
+substitutions:
+  - description:
+      'Handle the kAIChat UserSelectableType in the manage-sync settings switch.'
+    re_pattern:
+      '(\n[ \t]*)case syncer::UserSelectableType::kCookies:(\n[
+      \t]*NOTREACHED\(\);)'
+    replace:
+      '\1case syncer::UserSelectableType::kCookies:\1case
+      syncer::UserSelectableType::kAIChat:\2'
```
cdesouza-chromium: Check that plaster dos and donts PR I had linked before regarding switch cases.

cdesouza-chromium (approved):
LGTM

cdesouza-chromium (approved):
LGTM

darkdh (approved):

petemill:
All tests passed except windows x64 `InteractiveBrowserTestBrowsertest.ScrollIntoView` which is unrelated. I'm disabling upstream tests for the rebuild

PR #35029, opened 2026-03-26
wip [Sync] [AI Chat] (3 of n) Add SyncMetadataStore to AIChatDatabase

petemill:
Implements `syncer::SyncMetadataStore` on `AIChatDatabase` so the sync
infrastructure can persist entity metadata and data type state alongside
conversation data.

- New `ai_chat_sync_metadata` table (`storage_key TEXT PK`, `value BLOB`),
  created in `CreateSchema()`. No schema version bump or migration is
  needed: adding a new table doesn't modify existing tables, so older
  databases get it created via `CREATE TABLE IF NOT EXISTS` on init.
- `GetAllSyncMetadata` / `UpdateEntityMetadata` / `ClearEntityMetadata` /
  `UpdateDataTypeState` / `ClearDataTypeState`
- `DataTypeState` serialized into `sql::MetaTable`
- Unit tests for the metadata store (`AIChatDatabaseSyncTest`)

Spec:

Based on:
- https://github.com/brave/brave-core/pull/35028
- https://github.com/brave/brave-core/pull/35242

Series:
- https://github.com/brave/brave-core/pull/35028
- https://github.com/brave/brave-core/pull/35242
- https://github.com/brave/brave-core/pull/35029
- https://github.com/brave/brave-core/pull/35030
- https://github.com/brave/go-sync/pull/427 (server)

### DIFF
--- a/components/ai_chat/core/browser/BUILD.gn
+++ b/components/ai_chat/core/browser/BUILD.gn
@@ -107,6 +107,7 @@ static_library("browser") {
   public_deps = [
     "//base",
     "//components/os_crypt/async/common",
+    "//components/sync/model",
   ]
 
   deps = [
@@ -134,6 +135,7 @@ static_library("browser") {
     "//components/keyed_service/core",
     "//components/os_crypt/async/browser",
     "//components/prefs",
+    "//components/sync/protocol",
     "//components/update_client",
     "//components/user_prefs",
     "//mojo/public/cpp/bindings",
@@ -216,6 +218,7 @@ source_set("unit_tests") {
     "//components/history/core/test",
     "//components/os_crypt/async/browser:test_support",
     "//components/prefs:test_support",
+    "//components/sync/protocol",
     "//components/sync_preferences:test_support",
     "//net/traffic_annotation:test_support",
     "//services/network:test_support",
--- a/components/ai_chat/core/browser/ai_chat_database.cc
+++ b/components/ai_chat/core/browser/ai_chat_database.cc
@@ -26,6 +26,9 @@
 #include "brave/components/ai_chat/core/common/proto_conversion.h"
 #include "brave/components/ai_chat/core/proto/store.pb.h"
 #include "components/os_crypt/async/common/encryptor.h"
+#include "components/sync/model/metadata_batch.h"
+#include "components/sync/protocol/data_type_state.pb.h"
+#include "components/sync/protocol/entity_metadata.pb.h"
 #include "sql/init_status.h"
 #include "sql/meta_table.h"
 #include "sql/statement.h"
@@ -176,6 +179,9 @@ bool MigrateFrom9to10(sql::Database* db) {
       db->GetUniqueStatement(kAddExtractedTextColumnQuery));
   return statement.is_valid() && statement.Run();
 }
+
+// Key in sql::MetaTable for the serialized DataTypeState.
+const char kAIChatDataTypeStateKey[] = "ai_chat_data_type_state";
 
 }  // namespace
 
@@ -335,7 +341,6 @@ sql::InitStatus AIChatDatabase::InitInternal() {
       }
       current_version = 10;
     }
-
     // Migration unsuccessful, raze the database and re-init
     if (!migration_success) {
       transaction.Rollback();
@@ -1824,7 +1829,142 @@ bool AIChatDatabase::CreateSchema() {
     return false;
   }
 
+  // Sync metadata table for AI Chat sync.
+  static constexpr char kCreateSyncMetadataTableQuery[] =
+      "CREATE TABLE IF NOT EXISTS ai_chat_sync_metadata("
+      "storage_key TEXT PRIMARY KEY NOT NULL,"
+      "value BLOB NOT NULL"
+      ")";
+  CHECK(GetDB().IsSQLValid(kCreateSyncMetadataTableQuery));
+  if (!GetDB().Execute(kCreateSyncMetadataTableQuery)) {
+    return false;
+  }
+
   return true;
+}
+
+bool AIChatDatabase::GetAllSyncMetadata(syncer::MetadataBatch* metadata_batch) {
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+  if (!LazyInit()) {
+    return false;
+  }
+  DCHECK(metadata_batch);
+  if (!GetAllEntityMetadata(metadata_batch)) {
+    return false;
+  }
+  sync_pb::DataTypeState data_type_state;
+  if (!GetDataTypeState(&data_type_state)) {
+    return false;
+  }
+  metadata_batch->SetDataTypeState(data_type_state);
+  return true;
+}
+
+bool AIChatDatabase::ClearAllEntityMetadata() {
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+  if (!LazyInit()) {
+    return false;
+  }
+  sql::Statement s(
+      GetDB().GetUniqueStatement("DELETE FROM ai_chat_sync_metadata"));
+  return s.Run();
+}
+
+bool AIChatDatabase::UpdateEntityMetadata(
+    syncer::DataType data_type,
+    const std::string& storage_key,
+    const sync_pb::EntityMetadata& metadata) {
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+  DCHECK_EQ(data_type, syncer::AI_CHAT_CONVERSATION);
+  DCHECK(!storage_key.empty());
+  if (!LazyInit()) {
+    return false;
+  }
+  sql::Statement s(
+      GetDB().GetUniqueStatement("INSERT OR REPLACE INTO ai_chat_sync_metadata "
+                                 "(storage_key, value) VALUES(?, ?)"));
+  s.BindString(0, storage_key);
+  s.BindString(1, metadata.SerializeAsString());
+  return s.Run();
+}
+
+bool AIChatDatabase::ClearEntityMetadata(syncer::DataType data_type,
+                                         const std::string& storage_key) {
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+  DCHECK_EQ(data_type, syncer::AI_CHAT_CONVERSATION);
+  DCHECK(!storage_key.empty());
+  if (!LazyInit()) {
+    return false;
+  }
+  sql::Statement s(GetDB().GetUniqueStatement(
+      "DELETE FROM ai_chat_sync_metadata WHERE storage_key=?"));
+  s.BindString(0, storage_key);
+  return s.Run();
+}
+
+bool AIChatDatabase::UpdateDataTypeState(
+    syncer::DataType data_type,
+    const sync_pb::DataTypeState& data_type_state) {
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+  DCHECK_EQ(data_type, syncer::AI_CHAT_CONVERSATION);
+  if (!LazyInit()) {
+    return false;
+  }
+  // Store in meta table as the init creates it.
+  sql::MetaTable meta_table;
+  if (!meta_table.Init(&GetDB(), kCurrentDatabaseVersion,
+                       kCompatibleDatabaseVersionNumber)) {
+    return false;
+  }
+  return meta_table.SetValue(kAIChatDataTypeStateKey,
+                             data_type_state.SerializeAsString());
+}
+
+bool AIChatDatabase::ClearDataTypeState(syncer::DataType data_type) {
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+  DCHECK_EQ(data_type, syncer::AI_CHAT_CONVERSATION);
+  if (!LazyInit()) {
+    return false;
+  }
+  sql::MetaTable meta_table;
+  if (!meta_table.Init(&GetDB(), kCurrentDatabaseVersion,
+                       kCompatibleDatabaseVersionNumber)) {
+    return false;
+  }
+  return meta_table.DeleteKey(kAIChatDataTypeStateKey);
+}
+
+bool AIChatDatabase::GetAllEntityMetadata(
+    syncer::MetadataBatch* metadata_batch) {
+  DCHECK(metadata_batch);
+  sql::Statement s(GetDB().GetUniqueStatement(
+      "SELECT storage_key, value FROM ai_chat_sync_metadata"));
+  while (s.Step()) {
+    std::string storage_key = s.ColumnString(0);
+    std::string_view serialized_metadata = s.ColumnStringView(1);
+    auto entity_metadata = std::make_unique<sync_pb::EntityMetadata>();
+    if (!entity_metadata->ParseFromString(serialized_metadata)) {
+      DLOG(WARNING) << "Failed to deserialize AI_CHAT_CONVERSATION "
+                       "sync_pb::EntityMetadata.";
+      return false;
+    }
+    metadata_batch->AddMetadata(storage_key, std::move(entity_metadata));
+  }
+  return true;
+}
+
+bool AIChatDatabase::GetDataTypeState(sync_pb::DataTypeState* state) {
+  sql::MetaTable meta_table;
+  if (!meta_table.Init(&GetDB(), kCurrentDatabaseVersion,
+                       kCompatibleDatabaseVersionNumber)) {
+    return false;
+  }
+  std::string serialized_state;
+  if (!meta_table.GetValue(kAIChatDataTypeStateKey, &serialized_state)) {
+    *state = sync_pb::DataTypeState();
+    return true;
+  }
+  return state->ParseFromString(serialized_state);
 }
 
 }  // namespace ai_chat
--- a/components/ai_chat/core/browser/ai_chat_database.h
+++ b/components/ai_chat/core/browser/ai_chat_database.h
@@ -16,8 +16,13 @@
 #include "brave/components/ai_chat/core/common/mojom/ai_chat.mojom-forward.h"
 #include "brave/components/ai_chat/core/common/mojom/common.mojom-forward.h"
 #include "components/os_crypt/async/common/encryptor.h"
+#include "components/sync/model/sync_metadata_store.h"
 #include "sql/database.h"
 #include "sql/init_status.h"
+
+namespace syncer {
+class MetadataBatch;
+}  // namespace syncer
 
 namespace ai_chat {
 
@@ -30,13 +35,13 @@ extern const int kCurrentDatabaseVersion;
 // should be handled with removal and re-adding so that other classes can make
 // decisions about how it affects the rest of history.
 // All data should be stored encrypted.
-class AIChatDatabase {
+class AIChatDatabase : public syncer::SyncMetadataStore {
  public:
   AIChatDatabase(const base::FilePath& db_file_path,
                  scoped_refptr<os_crypt_async::Encryptor> encryptor);
   AIChatDatabase(const AIChatDatabase&) = delete;
   AIChatDatabase& operator=(const AIChatDatabase&) = delete;
-  virtual ~AIChatDatabase();
+  ~AIChatDatabase() override;
 
   // Gets lightweight metadata for all conversations. No high-memory-consuming
   // data is returned.
@@ -94,6 +99,23 @@ class AIChatDatabase {
   virtual bool DeleteAssociatedWebContent(std::optional<base::Time> begin_time,
                                           std::optional<base::Time> end_time);
 
+  // Reads all sync metadata (entity metadata + data type state) into the batch.
+  bool GetAllSyncMetadata(syncer::MetadataBatch* metadata_batch);
+
+  // Deletes all entity metadata (but not data type state).
+  bool ClearAllEntityMetadata();
+
+  // syncer::SyncMetadataStore:
+  bool UpdateEntityMetadata(syncer::DataType data_type,
+                            const std::string& storage_key,
+                            const sync_pb::EntityMetadata& metadata) override;
+  bool ClearEntityMetadata(syncer::DataType data_type,
+                           const std::string& storage_key) override;
+  bool UpdateDataTypeState(
+      syncer::DataType data_type,
+      const sync_pb::DataTypeState& data_type_state) override;
+  bool ClearDataTypeState(syncer::DataType data_type) override;
+
  private:
   friend class AIChatDatabaseTest;
   friend class AIChatDatabaseMigrationTest;
@@ -110,6 +132,9 @@ class AIChatDatabase {
       std::string_view conversation_id);
   std::vector<mojom::ContentArchivePtr> GetArchiveContentsForConversation(
       std::string_view conversation_uuid);
+
+  bool GetAllEntityMetadata(syncer::MetadataBatch* metadata_batch);
+  bool GetDataTypeState(sync_pb::DataTypeState* state);
 
   std::string DecryptColumnToString(sql::Statement& statement, int index);
   std::optional<std::string> DecryptOptionalColumnToString(
--- a/components/ai_chat/core/browser/ai_chat_database_unittest.cc
+++ b/components/ai_chat/core/browser/ai_chat_database_unittest.cc
@@ -28,6 +28,9 @@
 #include "brave/components/ai_chat/core/common/mojom/ai_chat.mojom.h"
 #include "brave/components/ai_chat/core/common/mojom/common.mojom.h"
 #include "components/os_crypt/async/browser/test_utils.h"
+#include "components/sync/model/metadata_batch.h"
+#include "components/sync/protocol/data_type_state.pb.h"
+#include "components/sync/protocol/entity_metadata.pb.h"
 #include "sql/init_status.h"
 #include "sql/meta_table.h"
 #include "sql/test/test_helpers.h"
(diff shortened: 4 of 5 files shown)


components/ai_chat/core/browser/ai_chat_service.cc
```diff
@@ -707,6 +758,11 @@ void AIChatService::DeleteConversation(const std::string& id) {
     ai_chat_db_
         .AsyncCall(base::IgnoreResult(&AIChatDatabase::DeleteConversation))
         .WithArgs(id);
+    if (sync_bridge_ && db_task_runner_) {
+      db_task_runner_->PostTask(
```
netzenbot: `base::Unretained(sync_bridge_.get())` is posted to `db_task_runner_`, which is a thread pool `SequencedTaskRunner`. If `AIChatService` is destroyed (or `sync_bridge_` is reset) before the posted task runs, this is a use-after-free. Use a `base::WeakPtr` to the bridge or ensure the bridge outlives all pending tasks via sequencing guarantees. The same issue applies to the `OnConversationAdded` and `OnConversationModified` posts. [CSM-010](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-memory.md#CSM-010)

components/ai_chat/core/browser/sync/ai_chat_sync_bridge_unittest.cc
```diff
@@ -0,0 +1,299 @@
...
+      found = true;
+      break;
+    }
+  }
+  EXPECT_TRUE(found) << "Remote conversation not found in database";
+}
+
+TEST_F(AIChatSyncBridgeTest, ApplyIncrementalSyncChangesAdd) {
+  CreateBridge();
+
```
netzenbot: `MergeFullSyncDataAppliesRemoteConversations` asserts the remote conversation exists in the database, but `MergeFullSyncData` only tracks remote UUIDs to skip re-upload — it has `TODO(petemill): Apply remote-only conversations to local database` and never calls any write path. This test will always fail or pass vacuously (if a prior `AddTestConversation` pollutes state). It tests behavior that isn't implemented yet. ([Verify Tests Actually Test What They Claim](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-012))

components/ai_chat/core/browser/sync/ai_chat_sync_bridge_unittest.cc
```diff
@@ -0,0 +1,299 @@
...
+  bool found = false;
+  for (const auto& conv : conversations) {
+    if (conv->uuid == "remote-conv-1") {
+      EXPECT_EQ(conv->title, "Remote Conversation");
+      found = true;
+      break;
+    }
+  }
+  EXPECT_TRUE(found) << "Remote conversation not found in database";
+}
```
netzenbot: `MergeFullSyncDataAppliesRemoteConversations` expects the remote conversation to be written to the database, but the implementation only collects remote UUIDs and has `// TODO(petemill): Apply remote-only conversations to local database.` — it never calls `ApplyRemoteConversation`. This test will fail. Either remove the assertion or implement the TODO first. [TI-039](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-039)

components/ai_chat/core/browser/BUILD.gn
```diff
@@ -130,6 +134,9 @@ static_library("browser") {
     "//components/os_crypt/async/common",
     "//components/os_crypt/sync",
```
netzenbot: `//components/sync/model` is included in public headers (`ai_chat_database.h` inherits from `syncer::SyncMetadataStore`, and `ai_chat_service.h` includes `ai_chat_sync_bridge.h` which pulls in `data_type_sync_bridge.h`). This dep should be in `public_deps`, not `deps`, so consumers get the necessary transitive include paths. See [BS-024](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-024).

patches/components-sync-model-model_error.h.patch
```diff
@@ -0,0 +1,14 @@
...
+index be8c65925d85dc4382584eaff122dc94f322ed27..a0c09446f516f31e4dc2eae5a9af37b261e7d20a 100644
+--- a/components/sync/model/model_error.h
++++ b/components/sync/model/model_error.h
+@@ -200,7 +200,8 @@ class ModelError {
+     kAutofillValuableMetadataFailedToLoadMetadata = 190,
+     kDataTypeControllerInFailedState = 191,
+     kBookmarksLocalCountExceededLimitAfterInitialMerge = 192,
+-    kMaxValue = kBookmarksLocalCountExceededLimitAfterInitialMerge,
++    kAIChatFailedToLoadMetadata = 193,
++    kMaxValue = kAIChatFailedToLoadMetadata,
```
netzenbot: nit: Missing `static_assert` guard. If upstream adds a new enum value at 192 or 193, this will silently produce a collision. Add something like `static_assert(static_cast<int>(syncer::ModelError::Type::kBookmarksLocalCountExceededLimitAfterInitialMerge) == 192, "Upstream added new ModelError::Type values — update kAIChatFailedToLoadMetadata");` before the new value. See [CSRC-021](https://github.com/brave/brave-core/tree/master/docs/best-practices/chromium-src-overrides.md#CSRC-021).

components/ai_chat/core/browser/ai_chat_service.cc
petemill: Fixed — replaced base::Unretained with base::WeakPtr<AIChatSyncBridge> obtained on the db_task_runner_ after bridge creation. The weak pointer is safely copied to the UI thread for PostTask calls and invalidated when the bridge is destroyed.

components/ai_chat/core/browser/sync/ai_chat_sync_bridge_unittest.cc
petemill: Fixed. Moved MergeFullSyncDataAppliesRemoteConversations, ApplyIncrementalSyncChangesAdd, and ApplyRemoteConversation* tests to commit 3 where the implementation lives.

components/ai_chat/core/browser/sync/ai_chat_sync_bridge_unittest.cc
petemill: Fixed. Moved MergeFullSyncDataAppliesRemoteConversations, ApplyIncrementalSyncChangesAdd, and ApplyRemoteConversation* tests to commit 3 where the implementation lives.

components/ai_chat/core/browser/BUILD.gn
petemill: Fixed — moved to public_deps since it's used in public headers (ai_chat_database.h inherits SyncMetadataStore, ai_chat_sync_bridge.h uses DataTypeSyncBridge).

patches/components-sync-model-model_error.h.patch
petemill: The patch itself will fail to apply if upstream modifies the enum around our insertion point, which serves as an equivalent guard. We can add an explicit static_assert if preferred.

components/ai_chat/core/browser/sync/ai_chat_sync_bridge_unittest.cc
```diff
@@ -0,0 +1,239 @@
...
+  AddTestConversation("debug-conv-2", "Debug 2");
+  CreateBridge();
+
+  auto batch = bridge_->GetAllDataForDebugging();
+  ASSERT_NE(batch, nullptr);
+
+  // Consume the batch and count entries.
+  int count = 0;
+  while (batch->HasNext()) {
+    batch->Next();
```
netzenbot: nit: `int count` should be `size_t count` per [CSA-067](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-067) — use `size_t` for counts and indices.

components/ai_chat/core/browser/sync/ai_chat_sync_bridge.h
```diff
@@ -0,0 +1,78 @@
...
+  // syncer::DataTypeSyncBridge:
+  std::unique_ptr<syncer::MetadataChangeList> CreateMetadataChangeList()
+      override;
+  std::optional<syncer::ModelError> MergeFullSyncData(
+      std::unique_ptr<syncer::MetadataChangeList> metadata_change_list,
+      syncer::EntityChangeList entity_data) override;
+  std::optional<syncer::ModelError> ApplyIncrementalSyncChanges(
+      std::unique_ptr<syncer::MetadataChangeList> metadata_change_list,
+      syncer::EntityChangeList entity_changes) override;
+  std::unique_ptr<syncer::DataBatch> GetDataForCommit(
```
netzenbot: nit: `<map>` is included but not used anywhere in this header. No `std::map` appears in the class declaration, method signatures, or data members. Remove it per IWYU. [CS-001](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-001)

components/ai_chat/core/browser/sync/ai_chat_sync_conversions.h
```diff
@@ -0,0 +1,51 @@
...
+#define BRAVE_COMPONENTS_AI_CHAT_CORE_BROWSER_SYNC_AI_CHAT_SYNC_CONVERSIONS_H_
+
+#include <memory>
+#include <string>
+#include <vector>
+
+#include "brave/components/ai_chat/core/common/mojom/ai_chat.mojom-forward.h"
+#include "brave/components/ai_chat/core/common/mojom/common.mojom-forward.h"
+
+namespace sync_pb {
```
netzenbot: nit: `<vector>` is included but none of the function declarations in this header use `std::vector` directly — all parameter and return types are forward-declared or scalar. Move it to the `.cc` file if needed there, or remove it. [CS-001](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-001)

components/ai_chat/core/browser/ai_chat_database.cc
```diff
@@ -1790,7 +1815,142 @@ bool AIChatDatabase::CreateSchema() {
...
+  if (!LazyInit()) {
+    return false;
+  }
+  DCHECK(metadata_batch);
+  if (!GetAllEntityMetadata(metadata_batch)) {
+    return false;
+  }
+  sync_pb::DataTypeState data_type_state;
+  if (!GetDataTypeState(&data_type_state)) {
+    return false;
```
netzenbot, line 1857: nit: `storage_key` and `serialized_metadata` are never modified after assignment — declare them `const` to communicate intent. [CS-064](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-064)

components/ai_chat/core/browser/ai_chat_database_unittest.cc
```diff
@@ -1159,6 +1162,105 @@ TEST_P(AIChatDatabaseTest, DeleteConversationEntryWithAssociatedContent) {
...
+  syncer::MetadataBatch batch;
+  ASSERT_TRUE(db_->GetAllSyncMetadata(&batch));
+  // After clearing, GetDataTypeState returns default (no initial sync).
+  EXPECT_EQ(batch.GetDataTypeState().initial_sync_state(),
+            sync_pb::DataTypeState::INITIAL_SYNC_STATE_UNSPECIFIED);
+}
+
 // Test the migration for each version upgrade
 class AIChatDatabaseMigrationTest : public testing::Test,
                                     public testing::WithParamInterface<int> {
```
netzenbot, line 1278: nit: `sequence_number` is set to 1 in the metadata but never checked after the roundtrip. Per [TI-022](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-022), serialization tests should verify all populated fields — add `EXPECT_EQ(entry->second->sequence_number(), 1)` to ensure the full metadata round-trips correctly.

components/ai_chat/core/browser/BUILD.gn
```diff
@@ -130,6 +137,8 @@ static_library("browser") {
     "//components/os_crypt/async/common",
```
netzenbot: `ai_chat_sync_bridge.cc` and `ai_chat_sync_conversions.cc` both include `brave/components/sync/protocol/ai_chat_specifics.pb.h`, but no corresponding `//brave/components/sync/protocol` (or equivalent brave proto target) dep is added to the `browser` static_library. This will cause build failures unless the dep is covered transitively — verify the proto target is explicitly declared. [BS-012](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-012)

components/ai_chat/core/browser/ai_chat_database.cc
```diff
@@ -161,6 +164,19 @@ bool MigrateFrom8to9(sql::Database* db) {
...
 
+bool MigrateFrom9to10(sql::Database* db) {
+  // Add sync metadata table for AI Chat sync.
+  static constexpr char kCreateSyncMetadataTableQuery[] =
+      "CREATE TABLE IF NOT EXISTS ai_chat_sync_metadata("
+      "storage_key TEXT PRIMARY KEY NOT NULL,"
+      "value BLOB NOT NULL"
+      ")";
+  return db->Execute(kCreateSyncMetadataTableQuery);
+}
```
netzenbot: nit: `const char kAIChatDataTypeStateKey[]` should be `constexpr char kAIChatDataTypeStateKey[]` — the value is known at compile time and the rest of the file (e.g., `kSearchQueriesSeparator` on line 39) uses `constexpr char[]` for string constants in the anonymous namespace. [CSA-025](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-025)

components/ai_chat/core/browser/sync/ai_chat_sync_bridge.h
```diff
@@ -0,0 +1,78 @@
...
+                                   delete_metadata_change_list) override;
+
+  // Called by AIChatService (posted to this sequence) when local data changes.
+  void OnConversationAdded(const std::string& uuid);
+  void OnConversationModified(const std::string& uuid);
+  void OnConversationDeleted(const std::string& uuid);
+
+  base::WeakPtr<AIChatSyncBridge> GetWeakPtr() {
+    return weak_ptr_factory_.GetWeakPtr();
+  }
```
netzenbot: nit: `database_` is a non-owned pointer — add `// not owned` per the Chromium convention. [CSM-002](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-memory.md#CSM-002)

components/ai_chat/core/browser/sync/ai_chat_sync_bridge_unittest.cc
```diff
@@ -0,0 +1,239 @@
...
+    count++;
+  }
+  EXPECT_EQ(count, 2);
+}
+
+TEST_F(AIChatSyncBridgeTest, GetDataForCommit) {
+  AddTestConversation("commit-conv", "Commit Test");
```
netzenbot: nit: `GetDataForCommit` serializes the conversation via `ConversationToSpecifics` but the test only asserts `uuid()`. Add checks for at least `title()` ("Commit Test") and that entries are populated to verify the full roundtrip. [TI-022](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-022)

components/ai_chat/core/browser/ai_chat_service.cc
```diff
@@ -450,13 +481,35 @@ void AIChatService::OnOsCryptAsyncReady(os_crypt_async::Encryptor encryptor) {
...
+                  std::make_unique<syncer::ClientTagBasedDataTypeProcessor>(
+                      syncer::AI_CHAT_CONVERSATION, base::DoNothing()),
+                  db);
+              *weak_out = (*bridge_out)->GetWeakPtr();
+            },
+            database_ptr, &sync_bridge_, &sync_bridge_weak_));
+  }
 }
 
 void AIChatService::OnDataDeletedForDisabledStorage(bool success) {
```
netzenbot, line 463: `sync_bridge_` and `sync_bridge_weak_` are `AIChatService` members (UI thread) but are written via a `PostTask` to `db_task_runner_`. There is no synchronization preventing the UI thread from reading `sync_bridge_` (e.g., in `DeleteConversation`) while the background task is still writing it — this is a data race. Consider using a separate `base::AtomicFlag` or `OnceCallback` to signal readiness, or store the bridge exclusively on the background sequence and expose only a thread-safe handle. ([Thread Safety - Service Method Calls](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-006))

components/ai_chat/core/browser/ai_chat_service.h
```diff
@@ -31,6 +31,7 @@
 #include "brave/components/ai_chat/core/browser/associated_content_delegate.h"
 #include "brave/components/ai_chat/core/browser/conversation_handler.h"
 #include "brave/components/ai_chat/core/browser/engine/engine_consumer.h"
+#include "brave/components/ai_chat/core/browser/sync/ai_chat_sync_bridge.h"
```
netzenbot: nit: `ai_chat_sync_bridge.h` is only needed for the private `sync_bridge_` and `sync_bridge_weak_` members — a forward declaration `class AIChatSyncBridge;` in the header is sufficient, with the full include moved to the `.cc`. This avoids leaking Chromium sync internals (`DataTypeSyncBridge` etc.) to all consumers of `ai_chat_service.h`. [ARCH-022](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-022)

components/ai_chat/core/browser/sync/ai_chat_sync_bridge.cc
```diff
@@ -0,0 +1,234 @@
...
+    syncer::EntityChangeList entity_data) {
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+
+  // Get all local conversations.
+  auto local_conversations = database_->GetAllConversations();
+
+  // Build set of remote UUIDs.
+  std::set<std::string> remote_uuids;
+  for (const auto& change : entity_data) {
+    if (change->type() == syncer::EntityChange::ACTION_DELETE) {
```
netzenbot: nit: `std::set<std::string> remote_uuids` is built once and only used for membership tests — no sorted iteration needed. Per [CSA-005](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-005), use `base::flat_set<std::string>` for write-once/small sorted sets, or `absl::flat_hash_set<std::string>` if sorted order is not required.

components/ai_chat/core/browser/ai_chat_database_unittest.cc
```diff
@@ -1159,6 +1162,105 @@ TEST_P(AIChatDatabaseTest, DeleteConversationEntryWithAssociatedContent) {
...
+
+TEST_F(AIChatDatabaseSyncTest, UpdateAndGetEntityMetadata) {
+  sync_pb::EntityMetadata metadata;
+  metadata.set_creation_time(12345);
+  metadata.set_sequence_number(1);
+
+  EXPECT_TRUE(db_->UpdateEntityMetadata(syncer::AI_CHAT_CONVERSATION,
+                                        "conv-uuid-1", metadata));
+
+  syncer::MetadataBatch batch;
```
netzenbot, line 1214: nit: `EXPECT_EQ(batch.GetAllMetadata().size(), 1u)` should be `ASSERT_EQ` — if the size check fails, the next line calls `begin()->first` on the map, which is UB when the map is empty or has an unexpected size. Use `ASSERT_EQ` to stop the test immediately on failure. [TI-034](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-034)

components/ai_chat/core/browser/ai_chat_database.cc
```diff
@@ -1790,7 +1815,142 @@ bool AIChatDatabase::CreateSchema() {
     return false;
   }
 
+  // Sync metadata table for AI Chat sync.
+  static constexpr char kCreateSyncMetadataTableQuery[] =
+      "CREATE TABLE IF NOT EXISTS ai_chat_sync_metadata("
```
netzenbot, line 1834: nit: `kCreateSyncMetadataTableQuery` is defined twice with identical SQL in the same file — once in the anonymous namespace (for `MigrateFrom9to10`) and again as a `static constexpr` local inside `CreateSchema()`. Since the anonymous-namespace definition is already file-scoped, `CreateSchema` can just use it directly, removing the duplicate. [BS-048](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-048)

components/ai_chat/core/browser/ai_chat_database.h
```diff
@@ -111,6 +133,9 @@ class AIChatDatabase {
   std::vector<mojom::ContentArchivePtr> GetArchiveContentsForConversation(
```
netzenbot, line 133: nit: `GetDataTypeState(sync_pb::DataTypeState* state)` is a new private function that fits the `bool` + out-parameter anti-pattern. Prefer returning `std::optional<sync_pb::DataTypeState>` — returning nullopt on failure and the value on success. [CSA-024](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-024)

components/ai_chat/core/browser/ai_chat_database.cc
```diff
@@ -1790,7 +1815,142 @@ bool AIChatDatabase::CreateSchema() {
...
+  // Store in meta table as the init creates it.
+  sql::MetaTable meta_table;
+  if (!meta_table.Init(&GetDB(), kCurrentDatabaseVersion,
+                       kCompatibleDatabaseVersionNumber)) {
+    return false;
+  }
+  return meta_table.SetValue(kAIChatDataTypeStateKey,
+                             data_type_state.SerializeAsString());
+}
+
```
netzenbot, line 1922: nit: `GetAllEntityMetadata` and `GetDataTypeState` are missing `DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_)` — every other method in `AIChatDatabase`, including all other private helpers, has this check. Add it to both for consistency. [CSM-015](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-memory.md#CSM-015)

components/ai_chat/core/browser/sync/ai_chat_sync_bridge.h
```diff
@@ -0,0 +1,78 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#ifndef BRAVE_COMPONENTS_AI_CHAT_CORE_BROWSER_SYNC_AI_CHAT_SYNC_BRIDGE_H_
+#define BRAVE_COMPONENTS_AI_CHAT_CORE_BROWSER_SYNC_AI_CHAT_SYNC_BRIDGE_H_
+
+#include <map>
```
netzenbot: nit: `database` is DCHECK'd non-null and stored for the lifetime of the bridge — use `AIChatDatabase&` as the constructor parameter and `raw_ref<AIChatDatabase>` as the stored member to communicate non-nullability. [CSA-030](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-030)

components/ai_chat/core/browser/sync/ai_chat_sync_bridge.cc
```diff
@@ -0,0 +1,234 @@
...
+  return std::nullopt;
+}
+
+std::unique_ptr<syncer::DataBatch> AIChatSyncBridge::GetDataForCommit(
+    StorageKeyList storage_keys) {
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+
+  auto batch = std::make_unique<syncer::MutableDataBatch>();
+  for (const auto& key : storage_keys) {
+    auto conversations = database_->GetAllConversations();
```
netzenbot: nit: The inner loop searching for a UUID by equality is a manual `find_if` — use `std::ranges::find_if(conversations, [&](const auto& c) { return c->uuid == key; })`. Also, `GetAllConversations()` is called once per `storage_key` — move it outside the loop to avoid redundant DB fetches. [CSA-042](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-042)

components/ai_chat/core/browser/ai_chat_database.cc
```diff
@@ -1790,7 +1815,142 @@ bool AIChatDatabase::CreateSchema() {
...
+  }
+
+  return true;
+}
+
+bool AIChatDatabase::GetAllSyncMetadata(syncer::MetadataBatch* metadata_batch) {
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+  if (!LazyInit()) {
+    return false;
+  }
```
netzenbot, line 1850: nit: `DCHECK(metadata_batch)` should be `CHECK(metadata_batch)`. The rest of `ai_chat_database.cc` uses `CHECK` for pointer/invariant guards (e.g., `CHECK(!conversation_uuid.empty())`, `CHECK(first_entry)`) — `DCHECK` only appears for sequence checks. A null `metadata_batch` is a programming error that should never happen in any build. [CS-025](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-025)

components/ai_chat/core/browser/ai_chat_database.cc
```diff
@@ -1790,7 +1815,142 @@ bool AIChatDatabase::CreateSchema() {
...
+    const std::string& storage_key,
+    const sync_pb::EntityMetadata& metadata) {
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+  DCHECK_EQ(data_type, syncer::AI_CHAT_CONVERSATION);
+  DCHECK(!storage_key.empty());
+  if (!LazyInit()) {
+    return false;
+  }
+  sql::Statement s(
+      GetDB().GetUniqueStatement("INSERT OR REPLACE INTO ai_chat_sync_metadata "
```
netzenbot, line 1884: nit: `DLOG(WARNING)` here is retained diagnostic logging — per [CS-067](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-067), use `DVLOG(1)` for any logging that stays in the codebase.

components/ai_chat/core/browser/ai_chat_database.h
```diff
@@ -94,6 +99,23 @@ class AIChatDatabase {
   virtual bool DeleteAssociatedWebContent(std::optional<base::Time> begin_time,
                                           std::optional<base::Time> end_time);
 
+  // Reads all sync metadata (entity metadata + data type state) into the batch.
```
netzenbot, line 102: nit: The doc for `GetAllSyncMetadata` doesn't mention that it must be called on the database sequence, or what happens if the batch already contains data before the call. Public methods that have sequence or ownership constraints should document them. [DOC-004](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-004)

petemill:
Converting to draft as am splitting up further

petemill:
Closing because splitting up and the existing noise is not relevant

petemill:
Replaced with https://github.com/brave/brave-core/pull/37537 and follow-ups

PR #35030, opened 2026-03-26
wip [Sync] [AI Chat] (reference branch) - Complete sync data flow: remote apply, controller registration

petemill:
Makes AI Chat sync fully functional end-to-end:

Remote data application:
- SpecificsToConversation() and SpecificsToArchive() reverse conversions to reconstruct mojom types from sync proto
- AIChatDatabase::ApplyRemoteConversation() upserts a full conversation (metadata + associated content + entries) from sync, replacing any existing data for that UUID
- MergeFullSyncData() now applies remote conversations to local DB
- ApplyIncrementalSyncChanges() now applies remote ADD/UPDATE to local DB

Controller registration:
- BRAVE_BUILD_CHROME_SYNC_CONTROLLERS macro in chrome_sync_controller_builder
- Creates DataTypeController for AI_CHAT_CONVERSATION using ProxyDataTypeControllerDelegate from AIChatService
- Gated behind features::IsBraveSyncAIChatEnabled()
- Gets AIChatService via factory from the profile

Also added IsEntityDataValid() to validate incoming entities have a UUID.

Spec:

Based on:
- https://github.com/brave/brave-core/pull/35028
- https://github.com/brave/brave-core/pull/35242
- https://github.com/brave/brave-core/pull/35029

Series:

-----------------------------------

I can also confirm that when the feature flag is disabled and the controller is not registered, that not AI Chat conversations are added to sync

image


components/ai_chat/core/browser/sync/ai_chat_sync_conversions_unittest.cc
```diff
@@ -0,0 +1,286 @@
...
+  auto conv = SpecificsToConversation(specifics);
+  EXPECT_EQ(conv->uuid, "minimal");
+  EXPECT_FALSE(conv->model_key.has_value());
+
+  auto archive = SpecificsToArchive(specifics);
+  EXPECT_TRUE(archive->entries.empty());
+}
+
+}  // namespace
+}  // namespace ai_chat
```
netzenbot, line 869: There's no test for malformed serialized bytes in the event fields (e.g., `web_sources_serialized`, `tool_use_serialized`). The `SpecificsToArchive` code silently drops events when `ParseFromString` fails, but there's no test verifying that behavior. Add a test with corrupted bytes to confirm the entry is still returned with the unparseable event omitted. https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-016

components/ai_chat/core/browser/sync/ai_chat_sync_conversions_unittest.cc
```diff
@@ -0,0 +1,286 @@
...
+  auto roundtripped = SpecificsToArchive(specifics);
+
+  ASSERT_EQ(roundtripped->entries.size(), 1u);
+  const auto& entry = roundtripped->entries[0];
+  EXPECT_EQ(*entry->uuid, "entry-uuid-1");
+  EXPECT_EQ(entry->character_type, mojom::CharacterType::HUMAN);
+  EXPECT_EQ(entry->action_type, mojom::ActionType::QUERY);
+  EXPECT_EQ(entry->text, "What is Brave browser?");
+  EXPECT_EQ(entry->selected_text, "Brave");
+}
```
netzenbot: `entry->uuid` is an optional but it's dereferenced with `*entry->uuid` without first asserting `ASSERT_TRUE(entry->uuid.has_value())`. If the optional is empty, this silently passes a garbage value to the comparison. Use `ASSERT_TRUE(entry->uuid.has_value()); EXPECT_EQ(*entry->uuid, "entry-uuid-1");`. ([Use `ASSERT_TRUE` Before Accessing Optional Values](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-025))

components/ai_chat/core/browser/sync/ai_chat_sync_conversions_unittest.cc
```diff
@@ -0,0 +1,286 @@
...
+
+  auto specifics = ConversationToSpecifics(*conv, *archive);
+  auto roundtripped = SpecificsToArchive(specifics);
+
+  const auto& events = *roundtripped->entries[0]->events;
+  ASSERT_GE(events.size(), 2u);
+  ASSERT_TRUE(events[1]->is_search_queries_event());
+  const auto& queries = events[1]->get_search_queries_event()->search_queries;
+  ASSERT_EQ(queries.size(), 2u);
+  EXPECT_EQ(queries[0], "brave browser");
```
netzenbot: `*roundtripped->entries[0]->events` is dereferenced without a prior `ASSERT_TRUE(roundtripped->entries[0]->events.has_value())`. If events is empty, the dereference causes undefined behavior and the test may crash rather than fail cleanly. The `RoundTripCompletionEvent` test correctly uses `ASSERT_TRUE` — apply the same pattern in all subsequent event tests. ([Use ASSERT for Preconditions That Subsequent Code Depends On](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-034))

components/ai_chat/core/browser/sync/ai_chat_sync_conversions_unittest.cc
petemill: Fixed — added MalformedSerializedBytesAreSkipped test that confirms entries with corrupted serialized bytes in event fields are returned with the unparseable event omitted.

components/ai_chat/core/browser/sync/ai_chat_sync_conversions_unittest.cc
petemill: fixed

components/ai_chat/core/browser/sync/ai_chat_sync_conversions_unittest.cc
petemill: Fixed — added ASSERT_TRUE guards before dereferencing the optional events in all event round-trip tests

chromium_src/chrome/browser/sync/chrome_sync_controller_builder.cc
```diff
@@ -0,0 +1,37 @@
...
+#include "brave/components/ai_chat/core/browser/ai_chat_service.h"
+#include "brave/components/ai_chat/core/common/features.h"
+#include "components/sync/service/data_type_controller.h"
+
+// Register AI Chat sync controller when building Chrome-level sync controllers.
+// Uses the profile from extension_system_profile_ to get AIChatService.
+#define BRAVE_BUILD_CHROME_SYNC_CONTROLLERS                                 \
+  if (ai_chat::features::IsBraveSyncAIChatEnabled()) {                      \
+    auto* ai_chat_service =                                                 \
+        ai_chat::AIChatServiceFactory::GetForBrowserContext(                \
```
netzenbot: The `chromium_src/chrome/browser/sync/DEPS` file only allows `+brave/components/sync/service`, but this new file includes headers from `brave/browser/ai_chat/` and `brave/components/ai_chat/core/`. The `checkdeps` presubmit will likely reject these. Add the required paths to the DEPS file (e.g., `+brave/browser/ai_chat`, `+brave/components/ai_chat/core/browser`, `+brave/components/ai_chat/core/common`). ([Verify chromium_src Header GN Dependencies](https://github.com/brave/brave-core/tree/master/docs/best-practices/chromium-src-overrides.md#CSRC-018))

components/ai_chat/core/browser/sync/ai_chat_sync_conversions.cc
```diff
@@ -128,4 +146,112 @@ std::string GetStorageKeyFromEntitySpecifics(
...
+    }
+    entry->text = entry_proto.entry_text();
+    if (entry_proto.has_prompt()) {
+      entry->prompt = entry_proto.prompt();
+    }
+    entry->character_type =
+        static_cast<mojom::CharacterType>(entry_proto.character_type());
+    entry->action_type =
+        static_cast<mojom::ActionType>(entry_proto.action_type());
+    if (entry_proto.has_selected_text()) {
```
netzenbot: The `has_editing_entry_uuid()` block is a no-op: the condition is checked but `entry_proto.editing_entry_uuid()` is never read or populated into `entry->editing_entry_uuid` before passing to `AddConversationEntry`. The comment says the database method handles it, but there's no code connecting the two. Either populate the field before calling `AddConversationEntry`, or remove the empty block entirely. [Don't Commit Commented-Out/Dead Code](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-066)

components/ai_chat/core/browser/sync/ai_chat_sync_bridge_unittest.cc
```diff
@@ -130,6 +130,65 @@ TEST_F(AIChatSyncBridgeTest, MergeFullSyncDataUploadsLocalConversations) {
...
+                                                    std::move(*entity_data)));
+
+  bridge_->ApplyIncrementalSyncChanges(bridge_->CreateMetadataChangeList(),
+                                       std::move(changes));
+
+  auto conversations = db_->GetAllConversations();
+  bool found = false;
+  for (const auto& conv : conversations) {
+    if (conv->uuid == "incremental-conv-1") {
+      found = true;
```
netzenbot: nit: The `ApplyIncrementalSyncChangesAdd` test uses a manual loop with `break` to check if a conversation exists — this can be replaced with `std::ranges::any_of(conversations, [](const auto& conv) { return conv->uuid == "incremental-conv-1"; })`. https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-042

components/ai_chat/core/browser/sync/ai_chat_sync_conversions_unittest.cc
```diff
@@ -0,0 +1,321 @@
...
+  // silently dropped.
+  EXPECT_EQ(archive->entries[0]->events->size(), 1u);
+  ASSERT_TRUE(archive->entries[0]->events->at(0)->is_completion_event());
+  EXPECT_EQ(
+      archive->entries[0]->events->at(0)->get_completion_event()->completion,
+      "Valid completion");
+}
+
+}  // namespace
+}  // namespace ai_chat
```
netzenbot, line 869: nit: `brave/components/ai_chat/core/common/test_utils.h` is included but none of its symbols are used in this test file. Remove it to keep includes clean. https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-001

components/ai_chat/core/browser/ai_chat_database.h
```diff
@@ -99,6 +99,13 @@ class AIChatDatabase : public syncer::SyncMetadataStore {
   virtual bool DeleteAssociatedWebContent(std::optional<base::Time> begin_time,
                                           std::optional<base::Time> end_time);
 
+  // Applies a remote conversation from sync. Deletes existing conversation
```
netzenbot: nit: The comment reads `|contents| should be empty strings since full text is not synced` but the parameter is named `entries`. Update the comment to reference the correct parameter name. ([Method Documentation Should Describe the Contract](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-004))

components/ai_chat/core/browser/ai_chat_service.cc
```diff
@@ -213,20 +217,25 @@ bool AIChatService::HasActiveConversationHandler(const std::string& uuid) {
...
-            return bridge->change_processor()->GetControllerDelegate();
+            return (*weak_bridge_ptr)
+                ->change_processor()
+                ->GetControllerDelegate();
           },
-          sync_bridge_weak_));
+          &sync_bridge_weak_));
 }
 
 ConversationHandler* AIChatService::CreateConversation() {
```
netzenbot, line 269: `&sync_bridge_weak_` passes a raw pointer to a member of `AIChatService` into a `base::BindRepeating` callback that runs on `db_task_runner_`. If `AIChatService` is destroyed (e.g., during profile shutdown) before the sync framework releases `ProxyDataTypeControllerDelegate`, any pending or queued invocation of this callback will dereference freed memory. The comment only addresses the bridge-not-ready-yet case; it doesn't protect against the outer service being gone. Consider binding `AIChatService::weak_ptr_factory_.GetWeakPtr()` and checking it in the lambda, or using a `base::SequenceBound` wrapper that prevents invocation after destruction. ([Thread Safety - Service Method Calls](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-006))

chromium_src/chrome/browser/sync/chrome_sync_controller_builder.cc
```diff
@@ -0,0 +1,39 @@
...
+            extension_system_profile_.value());                             \
+    if (ai_chat_service) {                                                  \
+      auto full_delegate = ai_chat_service->CreateSyncControllerDelegate(); \
+      auto transport_delegate =                                             \
+          ai_chat_service->CreateSyncControllerDelegate();                  \
+      if (full_delegate && transport_delegate) {                            \
+        controllers.push_back(std::make_unique<syncer::DataTypeController>( \
+            syncer::AI_CHAT_CONVERSATION, /*delegate_for_full_sync_mode=*/  \
+            std::move(full_delegate),     /*delegate_for_transport_mode=*/  \
+            std::move(transport_delegate)));                                \
```
netzenbot: The macro `BRAVE_BUILD_CHROME_SYNC_CONTROLLERS` is defined here but the upstream `chrome/browser/sync/chrome_sync_controller_builder.cc` never calls it — grep confirms no occurrence. The AI Chat sync controller registration block will never execute, meaning AI Chat conversations will never be synced. You need to either place the registration inline (without a macro) or find an expansion point in the upstream file that the macro can hook into. ([✅ Add Explanation Comments in chromium_src Override Files](https://github.com/brave/brave-core/tree/master/docs/best-practices/chromium-src-overrides.md#CSRC-017))

components/ai_chat/core/browser/sync/ai_chat_sync_conversions.cc
```diff
@@ -128,4 +247,108 @@ std::string GetStorageKeyFromEntitySpecifics(
...
+    if (content_proto.has_conversation_entry_uuid()) {
+      content->conversation_turn_uuid = content_proto.conversation_entry_uuid();
+    }
+    conversation->associated_content.push_back(std::move(content));
+  }
+
+  return conversation;
+}
+
+mojom::ConversationArchivePtr SpecificsToArchive(
```
netzenbot: The field `date_unix_epoch_micros` is decoded with `base::Time::FromDeltaSinceWindowsEpoch`, but that function uses the Windows epoch (Jan 1, 1601) not the Unix epoch (Jan 1, 1970). If the upload path in `ConversationToSpecifics` serializes with a different epoch, synced conversation timestamps will be silently corrupted. Verify both the upload and download sides use the same epoch, and if it's Windows epoch, rename the proto field to avoid confusion.

components/ai_chat/core/browser/sync/ai_chat_sync_conversions.cc
```diff
@@ -53,16 +104,86 @@ void ConvertEntryToProto(const mojom::ConversationTurn& entry,
...
+        auto* ws = event_proto->mutable_web_sources();
+        const auto& sources_event = event->get_sources_event();
+        for (const auto& source : sources_event->sources) {
+          ConvertWebSourceToProto(*source, ws->add_sources());
+        }
+        for (const auto& rr : sources_event->rich_results) {
+          ws->add_rich_results(rr);
+        }
+      } else if (event->is_inline_search_event()) {
+        auto* is = event_proto->mutable_inline_search();
```
netzenbot, line 353: The `source->url` and `source->favicon_url` are constructed from proto strings without checking `is_valid()`. The upload path in `ConvertWebSourceToProto` already guards with `if (source.url.is_valid())` before serializing — apply the same check here to avoid storing invalid GURLs from untrusted sync data. https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-060

chromium_src/components/browser_sync/common_controller_builder.cc
```diff
@@ -3,9 +3,35 @@
...
 
+#include "brave/components/ai_chat/core/common/buildflags/buildflags.h"
 #include "brave/components/history/core/browser/sync/brave_history_data_type_controller.h"
 #include "brave/components/history/core/browser/sync/brave_history_delete_directives_data_type_controller.h"
 
+#if BUILDFLAG(ENABLE_AI_CHAT)
+#include "brave/components/ai_chat/core/browser/ai_chat_service.h"
+#include "brave/components/ai_chat/core/common/features.h"
+#include "components/sync/service/data_type_controller.h"
+#endif  // BUILDFLAG(ENABLE_AI_CHAT)
```
netzenbot, line 14: nit: The includes and macro definition are in two back-to-back `#if BUILDFLAG(ENABLE_AI_CHAT)` blocks. Per [BS-035](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-035), merge them into a single guard block.

chromium_src/components/browser_sync/common_controller_builder.h
```diff
@@ -0,0 +1,26 @@
...
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#ifndef BRAVE_CHROMIUM_SRC_COMPONENTS_BROWSER_SYNC_COMMON_CONTROLLER_BUILDER_H_
+#define BRAVE_CHROMIUM_SRC_COMPONENTS_BROWSER_SYNC_COMMON_CONTROLLER_BUILDER_H_
+
+namespace ai_chat {
+class AIChatService;
+}  // namespace ai_chat
+
+#define SetSkillsService(...)                                \
```
netzenbot, line 13: nit: This new chromium_src override file has no explanation comment. Add a brief comment before the `#define` blocks explaining that this injects `SetAIChatService` and `ai_chat_service_` into `CommonControllerBuilder` to support AI Chat sync controller registration. https://github.com/brave/brave-core/tree/master/docs/best-practices/chromium-src-overrides.md#CSRC-017

components/ai_chat/core/browser/sync/ai_chat_sync_conversions.cc
```diff
@@ -129,4 +247,108 @@ std::string GetStorageKeyFromEntitySpecifics(
...
+
+  return conversation;
+}
+
+mojom::ConversationArchivePtr SpecificsToArchive(
+    const sync_pb::AIChatConversationSpecifics& specifics) {
+  auto archive = mojom::ConversationArchive::New();
+
+  for (const auto& entry_proto : specifics.entries()) {
+    auto entry = mojom::ConversationTurn::New();
```
netzenbot: nit: The `extra_snippets`, `rich_results`, and `search_queries` copy loops are direct range-to-vector copies that `base::ToVector` handles cleanly — e.g., `source->extra_snippets = base::ToVector(proto.extra_snippets());`. The same pattern applies to the `rich_results` and `search_queries` loops in `SpecificsToArchive`. https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-020

chromium_src/components/browser_sync/common_controller_builder.h
```diff
@@ -0,0 +1,26 @@
...
+#define skills_service_ \
+  skills_service_;      \
+  SafeOptional<raw_ptr<ai_chat::AIChatService>> ai_chat_service_
+
+#include <components/browser_sync/common_controller_builder.h>  // IWYU pragma: export
+
+#undef skills_service_
+#undef SetSkillsService
+
+#endif  // BRAVE_CHROMIUM_SRC_COMPONENTS_BROWSER_SYNC_COMMON_CONTROLLER_BUILDER_H_
```
netzenbot, line 26: nit: `SafeOptional<raw_ptr<ai_chat::AIChatService>> ai_chat_service_` is missing the `// not owned` annotation. Per Chromium convention, raw pointers to non-owned objects should have this comment. Since this is the last line of the macro (no trailing backslash), you can append `// not owned` directly. ([✅ Comment Non-Owned Raw Pointers](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-memory.md#CSM-002))

components/ai_chat/core/browser/sync/ai_chat_sync_conversions_unittest.cc
```diff
@@ -0,0 +1,320 @@
...
+  // Both events round-trip: the completion and the empty web sources.
+  ASSERT_EQ(archive->entries[0]->events->size(), 2u);
+  EXPECT_TRUE(archive->entries[0]->events->at(0)->is_completion_event());
+  EXPECT_TRUE(archive->entries[0]->events->at(1)->is_sources_event());
+  EXPECT_TRUE(
+      archive->entries[0]->events->at(1)->get_sources_event()->sources.empty());
+}
+
+}  // namespace
+}  // namespace ai_chat
```
netzenbot, line 869: The `RoundTripWebSourcesEvent` test doesn't verify `favicon_url` or `page_content` — both are set in `MakeAssistantEntryWithAllEvents` and serialized by `ConvertWebSourceToProto`. Add `EXPECT_EQ(sources->sources[0]->favicon_url, GURL("https://brave.com/favicon.ico"))` and a page_content check so the full serialization path is covered. https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-022

components/ai_chat/core/browser/sync/ai_chat_sync_conversions_unittest.cc
```diff
@@ -0,0 +1,320 @@
...
+  auto roundtripped = SpecificsToArchive(specifics);
+
+  ASSERT_EQ(roundtripped->entries.size(), 1u);
+  const auto& entry = roundtripped->entries[0];
+  ASSERT_TRUE(entry->uuid.has_value());
+  EXPECT_EQ(*entry->uuid, "entry-uuid-1");
+  EXPECT_EQ(entry->character_type, mojom::CharacterType::HUMAN);
+  EXPECT_EQ(entry->action_type, mojom::ActionType::QUERY);
+  EXPECT_EQ(entry->text, "What is Brave browser?");
+  EXPECT_EQ(entry->selected_text, "Brave");
```
netzenbot: nit: `RoundTripHumanEntry` (and the event round-trip tests) set `created_time` on entries but never verify the value survives the round-trip through `date_unix_epoch_micros`. Use a fixed `base::Time` value and add `EXPECT_EQ(entry->created_time, expected_time)` to confirm the timestamp is serialized and deserialized correctly. https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-022

chromium_src/chrome/browser/sync/sync_service_factory.cc
```diff
@@ -4,10 +4,16 @@
...
+#include "brave/components/ai_chat/core/common/buildflags/buildflags.h"
 #include "brave/components/sync/service/brave_sync_service_impl.h"
 #include "chrome/browser/history/history_service_factory.h"
 #include "chrome/browser/sync/device_info_sync_service_factory.h"
 
+#if BUILDFLAG(ENABLE_AI_CHAT)
+#include "brave/browser/ai_chat/ai_chat_service_factory.h"
+#include "brave/components/ai_chat/core/browser/ai_chat_service.h"
+#endif  // BUILDFLAG(ENABLE_AI_CHAT)
+
```
netzenbot, line 16: The `chromium_src/chrome/browser/sync/DEPS` only allows `+brave/components/sync/service`, but this file now includes `brave/components/ai_chat/core/common/buildflags/buildflags.h`, `brave/browser/ai_chat/ai_chat_service_factory.h`, and `brave/components/ai_chat/core/browser/ai_chat_service.h`. The iOS DEPS (`chromium_src/ios/chrome/browser/sync/DEPS`) was updated correctly — the desktop DEPS needs the same treatment. https://github.com/brave/brave-core/tree/master/docs/best-practices/chromium-src-overrides.md#CSRC-018

petemill:
converting to draft until the rest of the stack lands

PR #35085, opened 2026-03-30
[ads] Multiple tabs can be in focus at the same time for multiple windows

### LINKED ISSUE #38427: [ads] Multiple tabs can be in focus at the same time for multiple windows
### Description

t

### Steps to reproduce

1. 
2. 
3. 

### Actual result

t

### Expected result

t

### Reproduces how often

Easily reproduced

t

### Channel information

- [ ] release (stable)
- [ ] beta
- [ ] nightly

### Reproducibility

### Miscellaneous information

_No response_

tmancey:
OnBrowserSetLastActive and OnBrowserNoLongerActive previously ignored their browser argument so every tab helper responded to every browser focus change, causing all open tabs to simultaneously report as focused when switching windows. The browser argument is now passed through to MaybeSetBrowserIsActive and MaybeSetBrowserIsNoLongerActive, which use a stored TabInterface to compare the owning browser window and skip events from unrelated windows. The constructor is also updated to only mark the browser as active when the tab's own browser window is already active, so tabs opened in background windows start correctly inactive. Browser tests that wait for async notifications are converted from base::RunLoop to base::test::TestFuture so that Wait returns a timeout failure instead of hanging, and notifications that fire inside nested message loops such as CloseBrowserSynchronously are handled correctly regardless of call ordering.

Resolves https://github.com/brave/brave-browser/issues/38427

### DIFF
--- a/browser/brave_ads/tabs/BUILD.gn
+++ b/browser/brave_ads/tabs/BUILD.gn
@@ -35,6 +35,13 @@ source_set("tabs") {
     "//components/sessions",
     "//ui/base",
   ]
+
+  if (!is_android) {
+    deps += [
+      "//chrome/browser/ui/browser_window",
+      "//components/tabs:public",
+    ]
+  }
 }
 
 source_set("browser_tests") {
--- a/browser/brave_ads/tabs/ads_tab_helper.cc
+++ b/browser/brave_ads/tabs/ads_tab_helper.cc
@@ -27,6 +27,9 @@
 #if !BUILDFLAG(IS_ANDROID)
 #include "chrome/browser/ui/browser.h"
 #include "chrome/browser/ui/browser_list.h"
+#include "chrome/browser/ui/browser_window/public/browser_window_interface.h"
+#include "chrome/browser/ui/browser_window/public/browser_window_interface_iterator.h"
+#include "components/tabs/public/tab_interface.h"
 #endif
 
 namespace brave_ads {
@@ -88,9 +91,11 @@ AdsTabHelper::AdsTabHelper(content::WebContents* const web_contents)
 #if !BUILDFLAG(IS_ANDROID)
   // See `application_state_monitor_android.h` for Android.
   BrowserList::AddObserver(this);
-#endif  // !BUILDFLAG(IS_ANDROID)
 
-  MaybeSetBrowserIsActive();
+  MaybeSetInitialBrowserIsActive();
+#else
+  MaybeSetBrowserIsActive(/*browser_window=*/nullptr);
+#endif  // !BUILDFLAG(IS_ANDROID)
 
   OnVisibilityChanged(web_contents->GetVisibility());
 }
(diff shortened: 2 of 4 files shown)


browser/brave_ads/tabs/BUILD.gn
```diff
@@ -20,7 +20,11 @@ source_set("tabs") {
   ]
 
   if (!is_android) {
-    public_deps += [ "//chrome/browser/ui:browser_list" ]
+    public_deps += [
+      "//chrome/browser/ui:browser_list",
+      "//chrome/browser/ui/browser_window",
+      "//components/tabs:public",
```
aseren: do we expose this in headers? if not needs to be moved to deps

tmancey: Fixed

aseren (approved):

browser/brave_ads/tabs/ads_tab_helper.cc
```diff
@@ -91,9 +100,21 @@ AdsTabHelper::AdsTabHelper(content::WebContents* const web_contents)
 #if !BUILDFLAG(IS_ANDROID)
   // See "background_helper_android.h" for Android.
   BrowserList::AddObserver(this);
-#endif  // !BUILDFLAG(IS_ANDROID)
 
+  // Only mark the browser as active if this tab belongs to the currently active
+  // browser window. Tabs opened in background windows must start inactive.
+  if (tabs::TabInterface* const tab =
```
bridiver: please use dependency injection and pass BrowserWindowInterface into the constructor

bridiver: actually I guess the BrowserWindowInterface can change if the tab is moved so maybe the TabInterface?

bridiver: also this logic should probably go in `MaybeSetBrowserIsActive` so we don't duplicate it?

tmancey: We already inject web_contents, so I cannot see any benefit, so changed to `tab_ = tabs::TabInterface::MaybeGetFromContents(web_contents);`.

tmancey: Regarding duplicate code I was going to followup, but have made the change and pushed the fix.

netzenbot (approved):

browser/brave_ads/tabs/ads_tab_helper.cc
tmancey: Recent code in chrome/browser/ssl/https_only_mode_tab_helper.cc and chrome/browser/contextual_tasks/search_ai_mode_promo_tab_helper.cc (2025) does not inject either. I’ll file a follow-up issue to address any missing dependency injection.

browser/brave_ads/tabs/ads_tab_helper.cc
```diff
@@ -22,11 +22,15 @@
...
 #include "net/http/http_status_code.h"
+#include "third_party/abseil-cpp/absl/strings/str_format.h"
 #include "ui/base/page_transition_types.h"
 
 #if !BUILDFLAG(IS_ANDROID)
 #include "chrome/browser/ui/browser.h"
 #include "chrome/browser/ui/browser_list.h"
+#include "chrome/browser/ui/browser_window/public/browser_window_interface.h"
+#include "chrome/browser/ui/browser_window/public/browser_window_interface_iterator.h"
+#include "components/tabs/public/tab_interface.h"
```
bridiver, line 32: neither tab_inteface.h nor browser_window_interface.h are guarded by `!is_android` in gn so they should not be guarded by it here either.

bridiver: same with browser_window_interface_iterator.h

browser/brave_ads/tabs/ads_tab_helper.cc
```diff
@@ -132,7 +144,17 @@ bool AdsTabHelper::IsVisible() const {
   return is_web_contents_visible_ && is_browser_active_.value_or(false);
 }
 
-void AdsTabHelper::MaybeSetBrowserIsActive() {
+void AdsTabHelper::MaybeSetBrowserIsActive(
+    const BrowserWindowInterface* const browser_window) {
+#if !BUILDFLAG(IS_ANDROID)
```
bridiver, line 131: per above why is this code behind !IS_ANDROID? You don't appear to be using any desktop-specific methods here

tmancey:
Closing PR as stale.

PR #35089, opened 2026-03-30
Remote model fetch and parsing

jeremytieman:
Part of https://github.com/brave/brave-browser/issues/53952. This PR contains the base implementation of the remote models fetcher.


components/ai_chat/core/common/features.h
```diff
@@ -63,6 +63,18 @@ extern const base::FeatureParam<bool> kAutomaticModelSupportsTools;
 COMPONENT_EXPORT(AI_CHAT_COMMON)
 extern const base::FeatureParam<bool> kShouldIndentPageContentBlocks;
 
+// Enable remote model fetching from server endpoint
+COMPONENT_EXPORT(AI_CHAT_COMMON)
+BASE_DECLARE_FEATURE(kAIChatRemoteModels);
+
+// Remote models endpoint URL (only used when kAIChatRemoteModels is enabled)
+COMPONENT_EXPORT(AI_CHAT_COMMON)
+extern const base::FeatureParam<std::string> kRemoteModelsEndpoint;
```
netzenbot: nit: Per [CS-022](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-022), comments explaining what a `base::Feature` does should be in the `.cc` file where the feature is defined, not in the `.h` declaration. Move the `// Enable remote model fetching from server endpoint` comment (and the other feature param comments) to `features.cc`.

jeremytieman: Fixed.

netzenbot (approved):

components/ai_chat/core/common/features.cc
```diff
@@ -49,6 +49,17 @@ const base::FeatureParam<bool> kAutomaticModelSupportsTools{
 const base::FeatureParam<bool> kShouldIndentPageContentBlocks{
     &kAIChat, "should_indent_page_content_blocks", true};
 
+// Enable remote model fetching from server endpoint
+BASE_FEATURE(kAIChatRemoteModels, base::FEATURE_DISABLED_BY_DEFAULT);
+
+// Remote models endpoint URL (only used when kAIChatRemoteModels is enabled)
+const base::FeatureParam<std::string> kRemoteModelsEndpoint{
+    &kAIChatRemoteModels, "endpoint_url", ""};
```
DJAndries: Why not use a URL based off the result of `ai_chat::GetEndpointUrl`? If you added this for local dev purposes, you could just use the `--ai-chat-server-url` command-line switch instead.

components/ai_chat/core/common/features.cc
```diff
@@ -49,6 +49,17 @@ const base::FeatureParam<bool> kAutomaticModelSupportsTools{
 const base::FeatureParam<bool> kShouldIndentPageContentBlocks{
     &kAIChat, "should_indent_page_content_blocks", true};
 
+// Enable remote model fetching from server endpoint
+BASE_FEATURE(kAIChatRemoteModels, base::FEATURE_DISABLED_BY_DEFAULT);
```
DJAndries: I recommend renaming this feature to be a little more self-descriptive. i.e. `kAIChatRemoteModelConfig`

components/ai_chat/core/common/pref_names.cc
```diff
@@ -48,6 +48,7 @@ void RegisterProfilePrefs(PrefRegistrySimple* registry) {
     registry->RegisterListPref(kBraveAIChatUserMemories);
     registry->RegisterDictionaryPref(kBraveAIChatSkills);
     registry->RegisterBooleanPref(kBraveAIChatOllamaFetchEnabled, false);
+    registry->RegisterDictionaryPref(kRemoteModelsCache);
```
DJAndries: The model config is pretty large right? Perhaps it would make sense to store the config in a JSON file in the profile directory, and lazy load it when Leo is opened for the first time in a session. Otherwise, the local state file size will be increased substantially.

components/ai_chat/core/common/features.cc
jeremytieman: One idea that was floated is to be able to grab this file from S3 instead of our chat server. Having an entirely separate URL enables that.

components/ai_chat/core/common/pref_names.cc
jeremytieman: Somewhat. The description is a couple of sentences. The rest isn't necessarily any larger than anything else being stored in the registry for AI chat. Based on our other discussion, I'll look into storing this in a JSON file instead of the registry.

components/ai_chat/core/common/features.cc
jeremytieman: Done.

components/ai_chat/core/common/pref_names.cc
jeremytieman: Done.

components/ai_chat/core/common/pref_names.cc
```diff
@@ -56,6 +56,8 @@ void RegisterProfilePrefsForMigration(PrefRegistrySimple* registry) {
   if (ai_chat::features::IsAIChatEnabled()) {
     registry->RegisterBooleanPref(kObseleteBraveChatAutoGenerateQuestions,
                                   false);
+    // Moved from prefs to a JSON file in the profile directory (2026-03).
+    registry->RegisterDictionaryPref(kObsoleteRemoteModelsCache);
```
yrliou: I don't understand, why do we have this? We don't have this yet in the code base right?

components/ai_chat/core/common/features.cc
```diff
@@ -49,6 +49,18 @@ const base::FeatureParam<bool> kAutomaticModelSupportsTools{
...
+
+// Remote models endpoint URL (only used when kAIChatRemoteModelConfig is
+// enabled)
+const base::FeatureParam<std::string> kRemoteModelsEndpoint{
+    &kAIChatRemoteModelConfig, "endpoint_url", ""};
+
+// Cache TTL for remote models in minutes
+const base::FeatureParam<int> kRemoteModelsCacheTTLMinutes{
+    &kAIChatRemoteModelConfig, "cache_ttl_minutes", 1440};
+
```
yrliou, line 54: Naming seems a bit inconsistent (remote model config/remote models).
That being said, it might be actually better to have these reviewed together with actual usages, as each feature flag is basically just 2-3 more lines, it's trivial to look at together. It's a bit hard to conduct a meaningful review without more context in subsequent work. wdyt?

components/ai_chat/core/common/pref_names.cc
jeremytieman: Actually, that should have been pulled out. I'll do so now.

components/ai_chat/core/common/features.cc
jeremytieman: I hadn't noticed the naming inconsistency, but I agree, I'll get that fixed. As far as whether or not to include more context, I would argue it is better to have a too small PR than one too large. In this case, I'm just setting up the endpoint URL and the cache lifetime to be available and configured by Griffin. Is there something unclear about what these settings are for?

yrliou: Adding 2 lines for a feature flag each isn't gonna make any PR meaningfully larger, but I'm okay with very basic feature flags such as skill switches landing first if you prefer.
Please leave cache TTL out at least, that should be reviewed together with the caching.

yrliou: Since we already have this PR opened, probably the easiest is to just simplify this PR to only add enabled feature flag and empty endpoint feature param.

jeremytieman: Since more context seems appropriate, I actually already changed it the other way. I added the feature flag and the remote model fetcher/cache and kept the cache TTL Griffin parameter. This should give a better idea of my approach. It also incorporates Darnell's local JSON file suggestion.

components/ai_chat/core/browser/remote_models_fetcher.cc
```diff
@@ -0,0 +1,598 @@
...
+      background_task_runner_(
+          base::ThreadPool::CreateSequencedTaskRunner({base::MayBlock()})),
+      api_request_helper_(
+          std::make_unique<api_request_helper::APIRequestHelper>(
+              kTrafficAnnotation,
+              url_loader_factory)) {}
+
+RemoteModelsFetcher::~RemoteModelsFetcher() = default;
+
+void RemoteModelsFetcher::FetchModels(const std::string& url,
```
yrliou, line 286: @jeremytieman the PR is getting bigger again (not really due to feature flag), could you please separate it to 1) fetch and response parsing 2) all the caching stuff.

jeremytieman: I have reduced the code to just the fetch and response parsing.

components/ai_chat/core/browser/remote_models_fetcher.cc
```diff
@@ -0,0 +1,385 @@
+// Copyright (c) 2025 The Brave Authors. All rights reserved.
```
yrliou: 2026

components/ai_chat/core/browser/remote_models_fetcher.cc
```diff
@@ -0,0 +1,385 @@
...
+        destination: BRAVE_OWNED_SERVICE
+        internal {
+          contacts {
+            email: "support@brave.com"
+          }
+        }
+        user_data {
+          type: OTHER
+        }
+        last_reviewed: "2025-02-18"
```
yrliou: ?

components/ai_chat/core/browser/remote_models_fetcher.cc
```diff
@@ -0,0 +1,385 @@
...
+      semantics {
+        sender: "AI Chat Remote Models Fetcher"
+        description:
+          "Fetches the list of available AI chat models from a remote "
+          "endpoint. This allows dynamic model configuration without "
+          "requiring browser updates. Sends the browser's UI language "
+          "to allow localized model names."
+        trigger:
+          "Triggered on browser startup or when the model cache expires "
+          "(default: 1 hour TTL)."
```
yrliou: default here seems to be out of date

components/ai_chat/core/browser/remote_models_fetcher.cc
```diff
@@ -0,0 +1,385 @@
...
+  if (!endpoint_url.SchemeIs(url::kHttpsScheme)) {
+    VLOG(1) << "Remote models endpoint must use HTTPS: " << url;
+    std::move(callback).Run({});
+    return;
+  }
+
+  // Add language query parameter
+  std::string lang_code = GetTwoLetterLanguageCode();
+  GURL url_with_lang =
+      net::AppendQueryParameter(endpoint_url, "lang", lang_code);
```
yrliou: What do we use this data for?

components/ai_chat/core/browser/remote_models_fetcher.cc
```diff
@@ -0,0 +1,385 @@
...
+              url_loader_factory)) {}
+
+RemoteModelsFetcher::~RemoteModelsFetcher() = default;
+
+void RemoteModelsFetcher::FetchModels(const std::string& url,
+                                      FetchModelsCallback callback) {
+  GURL endpoint_url(url);
+
+  // Validate URL (HTTPS only)
+  if (!endpoint_url.is_valid() || endpoint_url.is_empty()) {
```
yrliou: Just checking !endpoint_url.is_valid() is enough.

components/ai_chat/core/browser/remote_models_fetcher.cc
```diff
@@ -0,0 +1,385 @@
...
+          "Not implemented yet. If this feature is enabled by default, "
+          "a policy will be added to disable it."
+      })");
+
+mojom::ModelCategory ParseCategory(const std::string& category_str) {
+  if (category_str == "chat") {
+    return mojom::ModelCategory::CHAT;
+  }
+  VLOG(1) << "Unknown model category: " << category_str;
+  return mojom::ModelCategory::CHAT;
```
yrliou: I'm a bit confused here as we are just returning CHAT always, don't we at least have summary already?

components/ai_chat/core/browser/remote_models_fetcher.cc
```diff
@@ -0,0 +1,385 @@
...
+          "fingerprinting signal."
+        destination: BRAVE_OWNED_SERVICE
+        internal {
+          contacts {
+            email: "support@brave.com"
+          }
+        }
+        user_data {
+          type: OTHER
+        }
```
yrliou, line 71: not sure what this one mean

components/ai_chat/core/browser/remote_models_fetcher.cc
```diff
@@ -0,0 +1,385 @@
...
+    net::DefineNetworkTrafficAnnotation("ai_chat_remote_models_fetcher", R"(
+      semantics {
+        sender: "AI Chat Remote Models Fetcher"
+        description:
+          "Fetches the list of available AI chat models from a remote "
+          "endpoint. This allows dynamic model configuration without "
+          "requiring browser updates. Sends the browser's UI language "
+          "to allow localized model names."
+        trigger:
+          "Triggered on browser startup or when the model cache expires "
```
yrliou: We would need to restrict the requests to only be triggered if user's opted into AI Chat too.

components/ai_chat/core/browser/remote_models_fetcher.cc
```diff
@@ -0,0 +1,385 @@
...
+        last_reviewed: "2025-02-18"
+      }
+      policy {
+        cookies_allowed: NO
+        setting:
+          "This feature can be disabled by setting the remote_models_endpoint "
+          "feature parameter to an empty string."
+        policy_exception_justification:
+          "Not implemented yet. If this feature is enabled by default, "
+          "a policy will be added to disable it."
```
yrliou: Do we plan on creating a policy for this? Why?

components/ai_chat/core/browser/remote_models_fetcher.cc
```diff
@@ -0,0 +1,385 @@
...
+constexpr char kDescriptionField[] = "description";
+constexpr char kCategoryField[] = "category";
+constexpr char kAccessField[] = "access";
+constexpr char kMaxAssociatedContentLengthField[] =
+    "max_associated_content_length";
+constexpr char kLongConversationWarningCharacterLimitField[] =
+    "long_conversation_warning_character_limit";
+
+constexpr net::NetworkTrafficAnnotationTag kTrafficAnnotation =
+    net::DefineNetworkTrafficAnnotation("ai_chat_remote_models_fetcher", R"(
```
yrliou, line 49: FYI new network requests need security and privacy review.

components/ai_chat/core/browser/remote_models_fetcher.cc
```diff
@@ -0,0 +1,385 @@
...
+    std::move(callback).Run({});
+    return;
+  }
+
+  // Add language query parameter
+  std::string lang_code = GetTwoLetterLanguageCode();
+  GURL url_with_lang =
+      net::AppendQueryParameter(endpoint_url, "lang", lang_code);
+
+  VLOG(1) << "Fetching remote models from: " << url_with_lang.spec()
```
yrliou: Suggest to use DVLOG(2) instead.

components/ai_chat/core/browser/remote_models_fetcher.cc
```diff
@@ -0,0 +1,385 @@
...
+
+RemoteModelsFetcher::~RemoteModelsFetcher() = default;
+
+void RemoteModelsFetcher::FetchModels(const std::string& url,
+                                      FetchModelsCallback callback) {
+  GURL endpoint_url(url);
+
+  // Validate URL (HTTPS only)
+  if (!endpoint_url.is_valid() || endpoint_url.is_empty()) {
+    VLOG(1) << "Invalid remote models endpoint URL: " << url;
```
yrliou: We don't need to log every error cases, as style guide suggests, most logging should be stripped out before checking in. Use DVLOG(1) for things you think we should be keeping. https://chromium.googlesource.com/chromium/src/+/main/styleguide/c++/c++.md#logging
Same thing applies to almost every log in this file, I think we have excessive logging atm.

components/ai_chat/core/browser/remote_models_fetcher.cc
```diff
@@ -0,0 +1,385 @@
...
+
+  VLOG(1) << "Fetching remote models from: " << url_with_lang.spec()
+          << " (language: " << lang_code << ")";
+
+  // Create a wrapper callback that adapts APIRequestResult to our callback
+  auto result_callback = base::BindOnce(
+      [](base::WeakPtr<RemoteModelsFetcher> fetcher, std::string original_url,
+         FetchModelsCallback callback,
+         api_request_helper::APIRequestResult result) {
+        VLOG(1) << "Anonymous callback";
```
yrliou: pure debugging log for development should be removed

components/ai_chat/core/browser/remote_models_fetcher.cc
```diff
@@ -0,0 +1,385 @@
...
+        if (!fetcher) {
+          return;
+        }
+        fetcher->OnFetchComplete(original_url, std::move(callback),
+                                 std::move(result));
+      },
+      weak_ptr_factory_.GetWeakPtr(), url, std::move(callback));
```
yrliou:
```
  auto result_callback = base::BindOnce(
      &RemoteModelsFetcher::OnFetchComplete,
      weak_ptr_factory_.GetWeakPtr(), url, std::move(callback));
```

components/ai_chat/core/browser/remote_models_fetcher.cc
```diff
@@ -0,0 +1,385 @@
...
+          (leo_opts->access == mojom::ModelAccess::PREMIUM) ? 160000 : 51200;
+      VLOG(1) << "Using default long_conversation_warning_character_limit for "
+              << model->key << ": " << warning_limit.value();
+    }
+
+    leo_opts->max_associated_content_length = max_content_length.value();
+    leo_opts->long_conversation_warning_character_limit = warning_limit.value();
+
+    model->options =
+        mojom::ModelOptions::NewLeoModelOptions(std::move(leo_opts));
```
yrliou: We seem to have duplicate code here and in ValidateModel, why are we doing things twice, why not just  a ParseModel function which returns a model only when it's valid?

components/ai_chat/core/browser/remote_models_fetcher.cc
yrliou: also let's combine checks for HTTPS here to just 1 block

components/ai_chat/core/browser/remote_models_fetcher.cc
jeremytieman: Model description localization.

components/ai_chat/core/browser/remote_models_fetcher.cc
jeremytieman: That seems to be the best option out of the possibilities: https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/privacy/traffic_annotation.proto;l=144?q=traffic_annotation.proto&ss=chromium%2Fchromium%2Fsrc

components/ai_chat/core/browser/remote_models_fetcher.cc
jeremytieman: No, I'll remove that section. Looks like Claude was being a bit proactive here.

components/ai_chat/core/browser/remote_models_fetcher.cc
jeremytieman: We do, and it is in the model definition already. Just missed parsing it here.

components/ai_chat/core/browser/remote_models_fetcher.cc
jeremytieman: Do I need to tag specific people/teams for that?

components/ai_chat/core/browser/remote_models_fetcher.cc
jeremytieman: That is already implemented, I just need to update the comment here.

components/ai_chat/core/browser/remote_models_fetcher.cc
jeremytieman: Fixed.

components/ai_chat/core/browser/remote_models_fetcher.cc
jeremytieman: Updated to today's date.

components/ai_chat/core/browser/remote_models_fetcher.cc
jeremytieman: Fixed.

components/ai_chat/core/browser/remote_models_fetcher.cc
jeremytieman: Done.

components/ai_chat/core/browser/remote_models_fetcher.cc
jeremytieman: Fixed.

components/ai_chat/core/browser/remote_models_fetcher.cc
jeremytieman: Fixed.

components/ai_chat/core/browser/remote_models_fetcher.cc
jeremytieman: Fixed.

components/ai_chat/core/browser/remote_models_fetcher.cc
jeremytieman: Done.

components/ai_chat/core/browser/remote_models_fetcher.cc
jeremytieman: Done, and stripped out some of the excessive logging (and comments).

components/ai_chat/core/browser/remote_models_fetcher.cc
jeremytieman: Done.

components/ai_chat/core/browser/remote_models_fetcher.cc
jeremytieman: Done.

components/ai_chat/core/browser/remote_models_fetcher.cc
jeremytieman: Removed the duplication.

components/ai_chat/core/browser/remote_models_fetcher.cc
yrliou: You can request one via creating an issue in https://github.com/brave/reviews.

jeremytieman: Done.

components/ai_chat/core/browser/remote_models_fetcher.cc
yrliou: How do we plan to use the description, would it replace translated UI strings we currently have when using remote models? My main concerns here is the remote one currently seems to only has English and I'm not sure what's our plan on the localization in the future. For example, we don't even need this at all if we are only gonna have English support.
I'm not sure if the localization part is already hashed out with the team, if it's already discussed and we had a consensus, then that's great. If not, might be good to kick off the discussion with the team.

The reason I brought this up in this PR is to have more context to double check we actually need to pass this data to server, mainly to make sure if we actually needs localization effort server side or would like to continue keeping the localization all on client side.
Or maybe a hybrid approach is better, where we use client side translation whenever available to get better translation support, if not, we can fallback to what we have from the server, so older browser versions can still see a description for newer models, if we're okay to only have english for that case, we don't need to do translation server side either, or if we want to provide a few languages we can support for these cases then we do need this lang_code and some localization support server side.

Anyway, it's bigger than this PR's scope, but seems relevant to me as it would affect whether we need to pass lang_code or not.

yrliou: That being said, I'm going to review as-is for this part, I think it doesn't have to be decided and block this PR anyway, we can just kick off the discussion and create an issue to change things later if we have something decided since we're not enabling the feature yet, and we can talk about this in future PRs when we start using models from remote.

jeremytieman: Yes, it would replace the translated UI strings. We've already had at least one discussion on this in the FE/BE meeting. I think there are already certain descriptions that are not currently translated at all - model names or model makers, I believe, are available only in English and are already currently displayed. So while it is something that needs dealt with, I believe the decision was that its not really worth holding this change up and rather something we should address longer term. I think a question that I am starting to go back-and-forth on is whether it should be an actual parameter as I have it currently coded, or if instead it should use the existing Accept-Language header instead. Both "work" from a practical standpoint, but I wonder if it would be better to stick with the existing standard. Do you have any thoughts or preferences?

yrliou: Existing Accept-Language header seems better to me.

yrliou: Since we don't really do translations yet on server, and we already can have language info in the header, definitely prefer we just don't pass this extra param for now.

components/ai_chat/core/browser/remote_models_fetcher_unittest.cc
```diff
@@ -0,0 +1,538 @@
...
+  fetcher_->FetchModels(
+      kTestEndpoint, base::BindOnce(
+                         [](bool* called, std::vector<mojom::ModelPtr>* models,
+                            std::vector<mojom::ModelPtr> result) {
+                           *called = true;
+                           *models = std::move(result);
+                         },
+                         &callback_called, &fetched_models));
+
+  EXPECT_TRUE(base::test::RunUntil([&]() { return callback_called; }));
```
yrliou, line 187: Please use base::test::TestFuture instead, same for other test cases.

components/ai_chat/core/browser/remote_models_fetcher.cc
```diff
@@ -0,0 +1,295 @@
...
+  }
+
+  const std::string* name = options_dict->FindString(kNameField);
+  if (!name || name->empty()) {
+    DVLOG(1) << "Model options missing or has empty 'name' field: " << *key;
+    return nullptr;
+  }
+
+  std::optional<int> max_content_length =
+      options_dict->FindInt(kMaxAssociatedContentLengthField);
```
ShivanKaul, line 183: should this be a `saturated_cast`? same for the other ints

components/ai_chat/core/browser/remote_models_fetcher.cc
jeremytieman: I've updated the code to use Accept-Language and removed the explicit language code parameter.

components/ai_chat/core/browser/remote_models_fetcher.cc
jeremytieman: Updated.

components/ai_chat/core/browser/remote_models_fetcher.cc
```diff
@@ -0,0 +1,281 @@
...
+  leo_opts->access = *parsed_access;
+
+  if (!max_content_length.has_value() || *max_content_length <= 0) {
+    max_content_length =
+        (leo_opts->access == mojom::ModelAccess::PREMIUM) ? 90000 : 32000;
+  }
+  if (!warning_limit.has_value() || *warning_limit <= 0) {
+    warning_limit =
+        (leo_opts->access == mojom::ModelAccess::PREMIUM) ? 160000 : 51200;
+  }
```
yrliou, line 197: don't need to check <= 0 here as we already checked above while parsing.

components/ai_chat/core/browser/remote_models_fetcher_unittest.cc
```diff
@@ -0,0 +1,538 @@
...
+
+  EXPECT_EQ("test-model-1", fetched_models[0]->key);
+  EXPECT_EQ("Test Model 1", fetched_models[0]->display_name);
+  EXPECT_TRUE(fetched_models[0]->vision_support);
+  EXPECT_FALSE(fetched_models[0]->supports_tools);
+
+  EXPECT_EQ("test-model-2", fetched_models[1]->key);
+  EXPECT_EQ("Test Model 2", fetched_models[1]->display_name);
+  EXPECT_FALSE(fetched_models[1]->vision_support);
+  EXPECT_TRUE(fetched_models[1]->supports_tools);
```
yrliou, line 198: We should be checking exact model with all fields & options, not just these four.
Otherwise the test coverage is insufficient for parsing.

components/ai_chat/core/browser/remote_models_fetcher_unittest.cc
```diff
@@ -0,0 +1,538 @@
...
+                         [](bool* called, std::vector<mojom::ModelPtr>* models,
+                            std::vector<mojom::ModelPtr> result) {
+                           *called = true;
+                           *models = std::move(result);
+                         },
+                         &callback_called, &fetched_models));
+
+  EXPECT_TRUE(base::test::RunUntil([&]() { return callback_called; }));
+  EXPECT_TRUE(fetched_models.empty());
+}
```
yrliou, line 237: Doesn't seem necessary to me to test both 404 and 500, we probably just need to test one?

components/ai_chat/core/browser/remote_models_fetcher_unittest.cc
```diff
@@ -0,0 +1,538 @@
...
+                         &callback_called, &fetched_models));
+
+  EXPECT_TRUE(base::test::RunUntil([&]() { return callback_called; }));
+  ASSERT_EQ(1u, fetched_models.size());
+  ASSERT_TRUE(fetched_models[0]->options->is_leo_model_options());
+  EXPECT_EQ(fetched_models[0]->options->get_leo_model_options()->category,
+            mojom::ModelCategory::SUMMARY);
+}
+
+TEST_F(RemoteModelsFetcherTest, MissingNumericFieldsGetTierDefaults) {
```
yrliou, line 474: We should also test negative value cases.

components/ai_chat/core/browser/remote_models_fetcher.cc
```diff
@@ -0,0 +1,281 @@
...
+      model_dict.FindBool(kVisionSupportField).value_or(false);
+  model->supports_tools =
+      model_dict.FindBool(kSupportsToolsField).value_or(false);
+  model->audio_support =
+      model_dict.FindBool(kAudioSupportField).value_or(false);
+  model->video_support =
+      model_dict.FindBool(kVideoSupportField).value_or(false);
+  model->is_suggested_model =
+      model_dict.FindBool(kIsSuggestedModelField).value_or(false);
+  model->is_near_model = model_dict.FindBool(kIsNearModelField).value_or(false);
```
yrliou, line 170: We're missing supported_capabilities we recently added.

components/ai_chat/core/browser/remote_models_fetcher_unittest.cc
```diff
@@ -0,0 +1,538 @@
...
+#include "testing/gtest/include/gtest/gtest.h"
+
+namespace ai_chat {
+
+namespace {
+
+constexpr char kTestEndpoint[] = "https://example.com/models";
+
+constexpr char kValidModelsJSON[] = R"({
+  "models": [
```
yrliou, line 30: Missing test case for list value, seems like we only test dict.

components/ai_chat/core/browser/remote_models_fetcher_unittest.cc
```diff
@@ -0,0 +1,538 @@
...
+                           *called = true;
+                           *models = std::move(result);
+                         },
+                         &callback_called, &fetched_models));
+
+  EXPECT_TRUE(base::test::RunUntil([&]() { return callback_called; }));
+  EXPECT_TRUE(fetched_models.empty());
+}
+
+TEST_F(RemoteModelsFetcherTest, MissingRequiredField) {
```
yrliou, line 277: We have many required fields it seems, why only testing key?

components/ai_chat/core/browser/remote_models_fetcher_unittest.cc
```diff
@@ -0,0 +1,538 @@
...
+  }
+
+  base::test::TaskEnvironment task_environment_;
+  network::TestURLLoaderFactory test_url_loader_factory_;
+  scoped_refptr<network::SharedURLLoaderFactory> shared_url_loader_factory_;
+  std::unique_ptr<RemoteModelsFetcher> fetcher_;
+};
+
+TEST_F(RemoteModelsFetcherTest, SuccessfulFetch) {
+  SimulateSuccessfulFetch(kValidModelsJSON);
```
yrliou, line 173: Can we also add a test case where there both valid and invalid models are present to verify we have valid models returned in this case?

yrliou (commented):
Please see my comments below, thanks.

components/ai_chat/core/browser/remote_models_fetcher_unittest.cc
```diff
@@ -0,0 +1,538 @@
...
+                           *called = true;
+                           *models = std::move(result);
+                         },
+                         &callback_called, &fetched_models));
+
+  EXPECT_TRUE(base::test::RunUntil([&]() { return callback_called; }));
+  EXPECT_TRUE(fetched_models.empty());
+}
+
+TEST_F(RemoteModelsFetcherTest, SummaryCategory) {
```
yrliou, line 429: In SuccessfulFetch, we can cover both chat and summary, don't need a specific test for summary category here. If we do want to test category specifically in its own test, then we should be testing all possible values (chat, summary, invalid one). So maybe it's better to improve this one to cover all cases to cover entire ParseCategory.

components/ai_chat/core/browser/remote_models_fetcher_unittest.cc
jeremytieman: Done.

components/ai_chat/core/browser/remote_models_fetcher.cc
jeremytieman: Done.

components/ai_chat/core/browser/remote_models_fetcher.cc
jeremytieman: Done.

components/ai_chat/core/browser/remote_models_fetcher_unittest.cc
jeremytieman: Done.

components/ai_chat/core/browser/remote_models_fetcher_unittest.cc
jeremytieman: Done.

components/ai_chat/core/browser/remote_models_fetcher_unittest.cc
jeremytieman: Done.

components/ai_chat/core/browser/remote_models_fetcher_unittest.cc
jeremytieman: Done.

components/ai_chat/core/browser/remote_models_fetcher_unittest.cc
jeremytieman: Done.

components/ai_chat/core/browser/remote_models_fetcher_unittest.cc
jeremytieman: Changed to test all fields.

components/ai_chat/core/browser/remote_models_fetcher_unittest.cc
jeremytieman: Done.

components/ai_chat/core/browser/remote_models_fetcher.cc
```diff
@@ -0,0 +1,307 @@
...
+    for (const auto& capability_value : *capabilities_list) {
+      if (!capability_value.is_string()) {
+        continue;
+      }
+      auto capability = ParseCapability(capability_value.GetString());
+      if (capability.has_value()) {
+        model->supported_capabilities.push_back(*capability);
+      }
+    }
+  }
```
yrliou: Suggest to pull this out for readability:
```
  std::optional<mojom::ConversationCapability> ParseCapability(
      const base::Value& value) {
    if (!value.is_string()) {
      return std::nullopt;
    }
    const std::string& s = value.GetString();
    if (s == "chat") {
      return mojom::ConversationCapability::CHAT;
    }
    if (s == "content_agent") {
      return mojom::ConversationCapability::CONTENT_AGENT;
    }
    if (s == "deep_research") {
      return mojom::ConversationCapability::DEEP_RESEARCH;
    }
    DVLOG(1) << "Unknown conversation capability: " << s;
    return std::nullopt;
  }

  std::vector<mojom::ConversationCapability> ParseCapabilityList(
      const base::DictValue& model_dict) {
    std::vector<mojom::ConversationCapability> capabilities;
    const base::ListValue* list =
        model_dict.FindList(kSupportedCapabilitiesField);
    if (!list) {
      // {CHAT} instead?
      return capabilities;
    }
    for (const auto& value : *list) {
      if (auto capability = ParseCapability(value)) {
        capabilities.push_back(*capability);
      }
    }
    // Should we check here if capabilities list is empty here and fallback to {CHAT}?
    return capabilities;
  }

// update caller here to:
// model->supported_capabilities = ParseCapabilityList(model_dict);
```

yrliou: Few comments:
1. I prefer not using else if but just use if directly when we are just returning things in the if block.
2. Should we fallback to {CHAT} when list is empty or missing? (Test case also needs to be added for both if adding fallbacks).

components/ai_chat/core/browser/remote_models_fetcher.cc
```diff
@@ -0,0 +1,307 @@
...
+  api_request_helper::APIRequestOptions options;
+  options.max_body_size = kMaxResponseSize;
+  options.timeout = base::Seconds(30);
+
+  api_request_helper_->Request(
+      "GET", endpoint_url, "", "", std::move(result_callback),
+      base::flat_map<std::string, std::string>(), options);
+}
+
+std::vector<mojom::ModelPtr> RemoteModelsFetcher::ParseModelsFromJSON(
```
yrliou: nit: this can be in anonymous namespace too like other Parse functions, unless we have other plans.

components/ai_chat/core/browser/remote_models_fetcher.cc
jeremytieman: 1. Sure, I have no strong opinion either way. Your way is less code, so probably better.
2. Until Ocelot, I would have said yes, but given how quick AI models are changing, I feel like making any assumptions in the browser side is probably risky for users. If the wrong model got defaulted to the wrong capability, it could be bad - particularly in an "agent" case. That said, chat is pretty safe. My vote is that no capabilities is safest for the end user, but if there is strong reason to default to chat, I wouldn't push back on that.

yrliou: Currently on client side we don't really expect an empt capability list either in model->support_capabilities or conversation's capabilities.

To be safe, I would propose:
1. Make capability_list a required field from server side model config, if not present, we skip this model client side.
2. If we end up with an empty list parsed, also skip this model, as it doesn't have any known capability that client knows how to support. If we ever have a new capability introduced here for a model, a corresponding client side change would be needed anyway, as model->support_capabilities is only used on client side to gate whether a specific client side tool a model can support or not, so it's actually more client side driven to add new capability types upon needed.

yrliou: btw has the model endpoint be updated to have this list served already?

components/ai_chat/core/browser/remote_models_fetcher.cc
```diff
@@ -0,0 +1,307 @@
...
+  }
+
+  if (!models_list) {
+    return {};
+  }
+
+  std::vector<mojom::ModelPtr> models;
+  for (const auto& model_value : *models_list) {
+    if (!model_value.is_dict()) {
+      continue;
```
netzenbot: Per [ARCH-060](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-060), a callback must be either always sync or always async — never mixed. Here, `callback` is run synchronously when the URL is invalid, but asynchronously via the network stack otherwise. Wrap the early-return path in a `base::SequencedTaskRunner::GetCurrentDefault()->PostTask(...)` to make it consistently async.

components/ai_chat/core/browser/remote_models_fetcher.cc
```diff
@@ -0,0 +1,307 @@
...
+    return mojom::ConversationCapability::CHAT;
+  } else if (capability_str == "content_agent") {
+    return mojom::ConversationCapability::CONTENT_AGENT;
+  } else if (capability_str == "deep_research") {
+    return mojom::ConversationCapability::DEEP_RESEARCH;
+  }
+  DVLOG(1) << "Unknown conversation capability: " << capability_str;
+  return std::nullopt;
+}
+
```
netzenbot, line 115: nit: `ParseCategory` silently falls back to `CHAT` for unknown categories, which is inconsistent with `ParseAccess` and `ParseCapability` both returning `std::nullopt` for unrecognized values. Per [ARCH-032](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-032), unknown types should return an explicit error (e.g., `std::optional<mojom::ModelCategory>`) so callers can decide whether to skip the model rather than silently accepting a wrong category.

components/ai_chat/core/browser/remote_models_fetcher.cc
```diff
@@ -0,0 +1,307 @@
+// Copyright (c) 2026 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#include "brave/components/ai_chat/core/browser/remote_models_fetcher.h"
+
+#include <utility>
```
netzenbot, line 9: nit: `std::optional` is used throughout this file but `<optional>` is not directly included — per [CS-001](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-001), include what you use directly. Add `#include <optional>` alongside the other standard library headers.

components/ai_chat/core/browser/remote_models_fetcher.cc
```diff
@@ -0,0 +1,307 @@
...
+    models_list = json.GetDict().FindList(kModelsKey);
+  } else if (json.is_list()) {
+    models_list = &json.GetList();
+  }
+
+  if (!models_list) {
+    return {};
+  }
+
+  std::vector<mojom::ModelPtr> models;
```
netzenbot: nit: `endpoint_url` is never modified after construction — consider `const GURL endpoint_url(url)` to communicate intent. See [CS-064](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-064).

components/ai_chat/core/browser/remote_models_fetcher.cc
yrliou: You were commenting on the wrong line, I believe it was for:
```
void RemoteModelsFetcher::FetchModels(const std::string& url, 
                                      FetchModelsCallback callback) {
  GURL endpoint_url(url);

  if (!endpoint_url.is_valid() || !endpoint_url.SchemeIs(url::kHttpsScheme)) {
    std::move(callback).Run({});
    return;
  }
```
Though I don't think what you said makes sense here, kicked off the discussion with team on Slack to see if we should remove ARCH-060.

components/ai_chat/core/browser/remote_models_fetcher.cc
jeremytieman: Let me check. If it hasn't, I will make sure it is.

components/ai_chat/core/browser/remote_models_fetcher.cc
yrliou: I think it should be fine to keep it as is in this particular case, we will review the internal caller in future PR.

components/ai_chat/core/browser/remote_models_fetcher.cc
jeremytieman: It hasn't, and I've put together a spec: https://github.com/brave/aichat/wiki/SPEC:-Models-Capabilities-List

components/ai_chat/core/browser/remote_models_fetcher.cc
jeremytieman: Done.

components/ai_chat/core/browser/remote_models_fetcher.cc
jeremytieman: I have implemented the server side changes and this PR has been updated to match.

components/ai_chat/core/browser/remote_models_fetcher.cc
jeremytieman: Done.

components/ai_chat/core/browser/remote_models_fetcher.cc
jeremytieman: Done.

components/ai_chat/core/browser/remote_models_fetcher.cc
jeremytieman: Done.

components/ai_chat/core/browser/remote_models_fetcher_unittest.cc
```diff
@@ -0,0 +1,701 @@
...
+  base::test::TestFuture<std::vector<mojom::ModelPtr>> future;
+  fetcher_->FetchModels(kTestEndpoint, future.GetCallback());
+  const auto& fetched_models = future.Get();
+
+  ASSERT_EQ(1u, fetched_models.size());
+  EXPECT_EQ("test-model-1", fetched_models[0]->key);
+  EXPECT_EQ("Test Model 1", fetched_models[0]->display_name);
+}
+
+TEST_F(RemoteModelsFetcherTest, SkipsUnknownCapabilities) {
```
netzenbot, line 581: The interceptors in `SimulateSuccessfulFetch`, `SimulateHTTPError`, and `SimulateNetworkError` silently ignore requests whose URL does not match `base_url`. Per [TI-037](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-037), add a `NOTREACHED() << request.url.spec()` in the `else` branch so unexpected requests cause an immediate, obvious test failure rather than a silent hang or a misleading "empty result" assertion failure.

components/ai_chat/core/browser/remote_models_fetcher.h
```diff
@@ -0,0 +1,56 @@
...
+
+#include <memory>
+#include <string>
+#include <vector>
+
+#include "base/functional/callback.h"
+#include "base/memory/scoped_refptr.h"
+#include "base/memory/weak_ptr.h"
+#include "brave/components/ai_chat/core/common/mojom/ai_chat.mojom.h"
+
```
netzenbot, line 18: nit: The header only uses `mojom::ModelPtr` in function parameter and return positions, so the full `ai_chat.mojom.h` can be replaced with `ai_chat.mojom-forward.h` here — move the full `#include "brave/components/ai_chat/core/common/mojom/ai_chat.mojom.h"` to `remote_models_fetcher.cc`. Other headers in this component (e.g. `model_service.h`) already follow this pattern. See [CS-070](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-070).

components/ai_chat/core/browser/remote_models_fetcher_unittest.cc
```diff
@@ -0,0 +1,701 @@
...
+  SimulateSuccessfulFetch(kBareListJSON);
+
+  base::test::TestFuture<std::vector<mojom::ModelPtr>> future;
+  fetcher_->FetchModels(kTestEndpoint, future.GetCallback());
+  const auto& fetched_models = future.Get();
+
+  ASSERT_EQ(1u, fetched_models.size());
+  EXPECT_EQ("test-model-1", fetched_models[0]->key);
+  EXPECT_EQ("Test Model 1", fetched_models[0]->display_name);
+}
```
netzenbot, line 579: nit: `MissingKey`, `MissingDisplayName`, `MissingOptions`, `MissingName`, and `MissingAccess` are structurally identical — only the JSON payload differs. Per [TI-023](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-023), consider collapsing them into a single parameterized test (or a local struct array + `SCOPED_TRACE` loop) to avoid repetition and make it easier to add new required-field cases.

components/ai_chat/core/browser/remote_models_fetcher_unittest.cc
```diff
@@ -0,0 +1,701 @@
...
+
```
netzenbot, line 580: nit: The four-line sequence `SimulateSuccessfulFetch` → `TestFuture` → `FetchModels` → `EXPECT_TRUE(future.Get().empty())` is copy-pasted roughly ten times. Per [TI-024](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-024), extract a small helper (e.g., `ExpectEmptyResult(const char* json)`) to keep tests concise and reduce the chance of copy-paste drift.

components/ai_chat/core/browser/remote_models_fetcher_unittest.cc
jeremytieman: Done.

components/ai_chat/core/browser/remote_models_fetcher.h
jeremytieman: Done.

components/ai_chat/core/browser/remote_models_fetcher_unittest.cc
jeremytieman: Done.

components/ai_chat/core/browser/remote_models_fetcher_unittest.cc
jeremytieman: Done.

components/ai_chat/core/browser/remote_models_fetcher.cc
```diff
@@ -0,0 +1,305 @@
...
+    api_request_helper::APIRequestResult result) {
+  if (!result.Is2XXResponseCode()) {
+    std::move(callback).Run({});
+    return;
+  }
+
+  std::move(callback).Run(ParseModelsFromJSON(result.value_body()));
+}
+
+}  // namespace ai_chat
```
netzenbot, line 322: nit: Per [CSM-019](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-memory.md#CSM-019), methods used as `base::BindOnce` targets should include `DCHECK_CALLED_ON_VALID_SEQUENCE`. Add a `SEQUENCE_CHECKER_SEQUENCE_CHECKER(sequence_checker_)` member to the class and call `DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_)` at the top of `OnFetchComplete`.

components/ai_chat/core/browser/remote_models_fetcher.cc
```diff
@@ -168,33 +156,46 @@ mojom::ModelPtr ParseModel(const base::DictValue& model_dict) {
...
+        !has_category) {
+      category = mojom::ModelCategory::SUMMARY;
+      has_category = true;
+    }
+  }
+
+  if (!has_category) {
+    return nullptr;
+  }
+
```
yrliou: I suggest pulling this out similar to below in anonymous namespace:
```
struct ParsedCapabilities {
  std::vector<mojom::ConversationCapability> capabilities;
  mojom::ModelCategory category;
};

// Parses the model's capability list and derives its category from the first
// CHAT or SUMMARY capability in list order. Returns std::nullopt if the list
// is missing or declares neither category capability, in which case the model
// is rejected.
std::optional<ParsedCapabilities> ParseCapabilities(
    const base::DictValue& model_dict) {
  const base::ListValue* capabilities_list =
      model_dict.FindList(kCapabilitiesField);
  if (!capabilities_list) {
    return std::nullopt;
  }

  std::vector<mojom::ConversationCapability> capabilities;
  for (const auto& capability_value : *capabilities_list) {
    if (!capability_value.is_string()) {
      continue;
    }
    if (auto capability = ParseCapability(capability_value.GetString())) {
      capabilities.push_back(*capability);
    }
  }

  for (auto capability : capabilities) {
    if (capability == mojom::ConversationCapability::CHAT) {
      return ParsedCapabilities{std::move(capabilities),
                                mojom::ModelCategory::CHAT};
    }
    if (capability == mojom::ConversationCapability::SUMMARY) {
      return ParsedCapabilities{std::move(capabilities),
                                mojom::ModelCategory::SUMMARY};
    }
  }
  return std::nullopt;
}
```

And in ParseModel:
```
auto parsed_capabilities = ParseCapabilities(model_dict);
if (!parsed_capabilities) {
    return nullptr;
}
...
model->supported_capabilities = std::move(parsed_capabilities->capabilities);
...
leo_opts->category = parsed_capabilities->category;
```

components/ai_chat/core/browser/remote_models_fetcher_unittest.cc
```diff
@@ -228,6 +209,8 @@ class RemoteModelsFetcherTest : public testing::Test {
           if (base::StartsWith(request.url.spec(), base_url)) {
             test_url_loader_factory_.AddResponse(request.url.spec(),
                                                  json_response);
+          } else {
+            NOTREACHED() << "Unexpected request: " << request.url.spec();
```
yrliou: ```ADD_FAILURE() << "Unexpected request: " << request.url.spec();```
same for the other two

components/ai_chat/core/browser/remote_models_fetcher.cc
yrliou: @jeremytieman actually let's follow our documented best practice here.

components/ai_chat/core/browser/remote_models_fetcher.h
```diff
@@ -0,0 +1,59 @@
...
+  explicit RemoteModelsFetcher(
+      scoped_refptr<network::SharedURLLoaderFactory> url_loader_factory);
+  ~RemoteModelsFetcher();
+
+  RemoteModelsFetcher(const RemoteModelsFetcher&) = delete;
+  RemoteModelsFetcher& operator=(const RemoteModelsFetcher&) = delete;
+
+  // Calls callback with parsed models on success, or empty vector on failure.
+  void FetchModels(const std::string& url, FetchModelsCallback callback);
```
netzenbot, line 47: nit: Per [DOC-004](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-004), method docs should describe restrictions. The comment should note that non-HTTPS or invalid URLs are silently rejected and produce an empty result — e.g., `// |url| must be a valid HTTPS URL; non-HTTPS or malformed URLs result in an empty callback.`

yrliou (approved):
LGTM, please make sure @netzenbot's comments are also addressed, thanks!

components/ai_chat/core/browser/remote_models_fetcher_unittest.cc
jeremytieman: Done.

components/ai_chat/core/browser/remote_models_fetcher.cc
jeremytieman: Done.

components/ai_chat/core/browser/remote_models_fetcher.cc
jeremytieman: Done.

components/ai_chat/core/browser/remote_models_fetcher.cc
jeremytieman: Done.

jeremytieman:
> LGTM, please make sure @netzenbot's comments are also addressed, thanks!

Yep, already done.

components/ai_chat/core/browser/remote_models_fetcher.h
jeremytieman: Done.

PR #35143, opened 2026-04-01
[AI Chat]: Fix global sidepanel display

### LINKED ISSUE #55399: [AI Chat] Global Sidepanel needs concept of "main" conversation
When loading a conversation in the sidebar, we show the regular AI Chat header:

Image

When then choosing a conversation from the history, we show the "go back to main conversation" header:

Image

Currently, the concept of "main conversation" is the conversation that's tied to the active Page. That is, when you navigate _back_ to a page, or re-open the sidebar on a page, the last conversation made against that Page is shown.

For Global Sidepanel, there should be no relationship between Page and Conversation. But we should still provide the above feature whereby selecting a conversation from history provides the back arrow to the _main_ conversation.

The _main_ conversation should be defined as follows for Global Sidepanel:

1. When opening a Panel, the f/e binds to a _new_ conversation (instead of asking for the conversation for the active Page as it does currently).
2. The f/e stores that conversation ID as it's _main_ conversation.
3. When clicking the "new conversation" button, the newly-bound conversation becomes the _main_ conversation.
4. When loading a conversation from history, the back arrow is shown to return the the current _main_ conversation.

This supports the following decisions:
- When re-opening the Panel, a new conversation is started instead of the older _main_ conversation retrieved.
- Each browser window has its own conversation.

Test Plan:
1. Enable brave://flags/#brave-ai-chat-global-side-panel
2. Start a new Leo conversation
3. Select a conversation from history

4. Back arrow should appear
5. Click the back arrow should take you back to the same conversation
6. Start a new conversation with the new conversation button
7. Open a conversation from history
8. Clicking the back arrow should take you to the new conversation you started, not the first one.

fallaciousreasoning:
Series https://github.com/brave/brave-browser/issues/53416
Resolves https://github.com/brave/brave-browser/issues/55399

This updates the side panel in non agentic mode to look tab associated

### DIFF
--- a/components/ai_chat/resources/page/state/active_chat_context.test.tsx
+++ b/components/ai_chat/resources/page/state/active_chat_context.test.tsx
@@ -112,7 +112,7 @@ beforeEach(() => {
 
 describe('ActiveChatProvider routing — global panel mode', () => {
   it('binds a fresh conversation on initial load (no selected id)', async () => {
-    await renderProvider({
+    const { detailsRef } = await renderProvider({
       isGlobalPanel: true,
       selectedConversationId: undefined,
     })
@@ -120,14 +120,6 @@ describe('ActiveChatProvider routing — global panel mode', () => {
     // No id -> newConversation -> uiHandler.newConversation in real code.
     expect(mockedBind.newConversation).toHaveBeenCalledTimes(1)
     expect(mockedBind.bindConversation).not.toHaveBeenCalled()
-  })
-
-  it('marks the initial conversation as the main conversation', async () => {
-    const { detailsRef } = await renderProvider({
-      isGlobalPanel: true,
-      selectedConversationId: undefined,
-    })
-
     expect(detailsRef.current!.isMainConversation).toBe(true)
   })
 
@@ -170,22 +162,6 @@ describe('ActiveChatProvider routing — global panel mode', () => {
     // so createNewConversation's tab-sentinel short-circuit doesn't apply and
     // we navigate to '/' (which will mount as a fresh global main on remount).
     expect(window.location.pathname).toBe('/')
-  })
-
-  it('openMainConversation routes back to the captured global main UUID', async () => {
-    const updateSpy = jest.fn()
-    const { detailsRef } = await renderProvider({
-      isGlobalPanel: true,
-      selectedConversationId: undefined,
-      updateSelectedConversationId: updateSpy,
-    })
-
-    act(() => {
-      detailsRef.current!.openMainConversation()
-    })
-
-    // The first global-main UUID handed out by the mock is `new-0`.
-    expect(updateSpy).toHaveBeenCalledWith('new-0')
   })
 })
 
--- a/components/ai_chat/resources/page/state/active_chat_context.test.tsx
+++ b/components/ai_chat/resources/page/state/active_chat_context.test.tsx
@@ -1,0 +1,273 @@
+// Copyright (c) 2026 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import * as React from 'react'
+import { render, act, waitFor } from '@testing-library/react'
+import { clearAllDataForTesting } from '$web-common/api'
+import {
+  ActiveChatProvider,
+  SelectedChatDetails,
+  tabAssociatedChatId,
+  useActiveChat,
+} from './active_chat_context'
+import { MockContext } from './mock_context'
+import * as BindConversation from '../api/bind_conversation'
+import {
+  createMockConversationHandler,
+  defaultConversationState,
+} from '../api/mock_interfaces'
+import createConversationAPI from '../api/conversation_api'
+
+// Replace the real BindConversation helpers with jest mocks so we can:
+// - assert which routing decision the provider made (bind by uuid vs new vs
+//   tab-related bind)
+// - return a working conversation API so getState.fetch() resolves and
+//   `globalConversationId` actually populates inside the provider
+jest.mock('../api/bind_conversation', () => ({
+  __esModule: true,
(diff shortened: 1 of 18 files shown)


Douglashdaniel (approved):

netzenbot (approved):

Douglashdaniel (approved):

components/ai_chat/resources/page/state/active_chat_context.tsx
```diff
@@ -74,40 +77,69 @@ function ActiveChatProvider({
...
+
+  // Whenever we create a new conversation in global panel mode, that becomes the main conversation.
+  React.useEffect(() => {
+    if (isGlobalPanel && !selectedConversationId) {
+      conversationAPI?.api?.getState
+        .fetch()
+        .then((s) => setGlobalConversationId(s.conversationUuid))
+    }
+  }, [isGlobalPanel, conversationAPI])
+
```
petemill, line 88: What happens when we load a specific conversation in the panel, and want to continue with it as the main conversation. I think there are cases where this happens, such as when opening a new conversation from the omnibox. Perhaps this conditional should be

```ts
if (isGlobalPanel && (!selectedConversationId || !globalConversationId)) {
...
```

components/ai_chat/resources/page/state/active_chat_context.tsx
```diff
@@ -163,7 +195,7 @@ function ActiveChatProvider({
 
     // Special case the default conversation - it gets treated specially as
     // the chat is rebound as the tab navigates.
-    if (selectedConversationId === tabAssociatedChatId) return
+    if (isMainConversation) return
```
petemill: This might need to stay as only checking for `tabAssociatedChatId` because what we're doing here really is checking the URL. So it should already be covered by `!selectedConversationId`

components/ai_chat/resources/page/state/active_chat_context.tsx
fallaciousreasoning: ah good point - I think the omnibox ask leo is in fullpage but the context menu tools do that.

components/ai_chat/resources/page/state/active_chat_context.tsx
fallaciousreasoning: done

components/ai_chat/resources/page/state/active_chat_context.test.tsx
```diff
@@ -0,0 +1,273 @@
...
+    expect(mockedBind.bindConversation).not.toHaveBeenCalled()
+  })
+
+  it('marks the initial conversation as the main conversation', async () => {
+    const { detailsRef } = await renderProvider({
+      isGlobalPanel: true,
+      selectedConversationId: undefined,
+    })
+
+    expect(detailsRef.current!.isMainConversation).toBe(true)
```
petemill, line 123: nit: couldn't this just be part of the previous test?

components/ai_chat/resources/page/state/active_chat_context.tsx
```diff
@@ -66,48 +65,78 @@ type ActiveChatContextProps = {
...
+    React.useState<string>('')
+
+  // Whenever we create a new conversation in global panel mode, that becomes the main conversation.
+  React.useEffect(() => {
+    if (isGlobalPanel && (!selectedConversationId || !globalConversationId)) {
+      conversationAPI?.api?.getState
+        .fetch()
+        .then((s) => setGlobalConversationId(s.conversationUuid))
+    }
+  }, [isGlobalPanel, conversationAPI])
```
petemill, line 87: I hate to say it, but do we maybe only need to do any of the "main conversation" stuff if we're `isStandalone`? I suppose it doesn't hurt though, but I can't think that full page would ever need it. Maybe a responsive mode of that where we want to show the back button? (we don't at the moment)

components/ai_chat/resources/page/state/active_chat_context.test.tsx
```diff
@@ -0,0 +1,273 @@
...
+      updateSelectedConversationId: updateSpy,
+    })
+
+    act(() => {
+      detailsRef.current!.openMainConversation()
+    })
+
+    // The first global-main UUID handed out by the mock is `new-0`.
+    expect(updateSpy).toHaveBeenCalledWith('new-0')
+  })
```
petemill: Wouldn't it already be on the main conversation? Should we make it bind to another conversation before navigating back to the main conversation?

components/ai_chat/resources/page/stories/components_panel.tsx
```diff
@@ -487,7 +487,7 @@ function StoryContext(
     selectedConversationId: CONVERSATIONS[0].uuid,
     updateSelectedConversationId: () => {},
     createNewConversation: () => {},
-    isTabAssociated: argsRef.current.isDefaultConversation,
+    isMainConversation: argsRef.current.isDefaultConversation,
```
petemill, line 490: Also needs `openMainConversation`

petemill (approved):
The routing and binding logic look good to me.

I'm seeing some weird bugs in global panel with content not getting attached when it should. And some JS errors with the attachments menu. Can't be related to this PR, but will log some issues and/or discuss with you.

Well done setting up the complex test environment for a React Context that calls external binding functions and reacts to and makes URL changes. You made the actual tests easy to read.
(Although I do have a one minor question on the tests).

components/ai_chat/resources/page/state/active_chat_context.test.tsx
fallaciousreasoning: removed this - it doesn't seem that useful

components/ai_chat/resources/page/state/active_chat_context.tsx
fallaciousreasoning: We probably should - it makes things a bit more complicated though because the `isStandalone` is async too. I'll file a follow up issue for doing this:

fallaciousreasoning: https://github.com/brave/brave-browser/issues/55536

PR #35208, opened 2026-04-02
[Email Aliases] Limit Reached dialog

### LINKED ISSUE #53188: Tell the user that they've reached limit from in-page email aliases creation bubble
If the user has reached the limit of email aliases, we currently still let them choose a new alias (including cycling between aliases), and only block them when they press the `Create alias` button. 

Image

We should not even show them an alias if they've reached their limit.

boocmp:
Fix https://github.com/brave/brave-browser/issues/53188

[Figma](https://www.figma.com/design/N0Xsctr5jt7I3Ots0G355V/%F0%9F%93%A7-Email-aliases?node-id=817-9749&t=j6KO9wfCaeeFPIpN-0) design for the `Limit Reached` state.

image

### DIFF
--- a/components/email_aliases/resources/content/email_alias_limit_reached.tsx
+++ b/components/email_aliases/resources/content/email_alias_limit_reached.tsx
@@ -1,0 +1,96 @@
+// Copyright (c) 2026 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import {
+  color,
+  font,
+  radius,
+  spacing,
+  typography,
+} from '@brave/leo/tokens/css/variables'
+import * as React from 'react'
+import { formatLocale, getLocale } from '$web-common/locale'
+import Alert from '@brave/leo/react/alert'
+import Col from './styles/Col'
+import styled from 'styled-components'
+import { Alias } from 'gen/brave/components/email_aliases/email_aliases.mojom.m'
+import './strings'
+
+const SectionCol = styled(Col)`
+  row-gap: ${spacing['2Xl']};
+`
+
+const LimitAlertTitle = styled.div`
+  font: ${font.heading.h4};
+  letter-spacing: ${typography.letterSpacing.large};
+`
+
+const LimitDescription = styled.div`
+  font: ${font.default.regular};
+`
+
+const ListLabel = styled.div`
+  font: ${font.small.semibold};
+  line-height: ${typography.lineHeight.small};
+  margin: 0;
+  padding: 0 ${spacing.s};
+`
+
+const AliasListBox = styled.div`
+  border: 1px solid ${color.divider.subtle};
+  border-radius: ${radius.xl};
+  overflow: hidden auto;
+  max-height: 208px;
+`
+
+const AliasRow = styled.div<{ showDivider: boolean }>`
+  font-weight: ${font.default.semibold};
+  padding: ${spacing.m} ${spacing.l};
+  gap: 8px;
+  border-bottom: ${(p) =>
+    p.showDivider ? `1px solid ${color.divider.subtle}` : 'none'};
+`
+
+export const EmailAliasLimitReached = ({
+  aliases,
+  aliasLimit,
+}: {
+  aliases: Alias[]
+  aliasLimit: number
+}) => {
+  const count = aliases.length
+  return (
+    <SectionCol>
+      <Alert type='warning'>
+        <LimitAlertTitle>
+          {getLocale(S.SETTINGS_EMAIL_ALIASES_LIMIT_REACHED_ALERT_TITLE)}
+        </LimitAlertTitle>
+        <LimitDescription>
+          {formatLocale(S.SETTINGS_EMAIL_ALIASES_LIMIT_REACHED_ALERT_BODY, {
+            $1: String(aliasLimit),
+          })}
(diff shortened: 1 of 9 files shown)


components/email_aliases/resources/content/email_alias_limit_reached.tsx
```diff
@@ -0,0 +1,96 @@
...
+import { Alias } from 'gen/brave/components/email_aliases/email_aliases.mojom.m'
+import './strings'
+
+const SectionCol = styled(Col)`
+  row-gap: ${spacing['2Xl']};
+`
+
+const LimitAlertTitle = styled.div`
+  font: ${font.heading.h4};
+  letter-spacing: ${typography.letterSpacing.large};
```
fallaciousreasoning: you shouldn't need to set this - it should come from the font property.

components/email_aliases/resources/content/email_alias_limit_reached.tsx
```diff
@@ -0,0 +1,96 @@
...
+`
+
+const LimitDescription = styled.div`
+  font: ${font.default.regular};
+`
+
+const ListLabel = styled.div`
+  font: ${font.small.semibold};
+  line-height: ${typography.lineHeight.small};
```
fallaciousreasoning: ditto

components/email_aliases/resources/content/email_alias_limit_reached.tsx
```diff
@@ -0,0 +1,96 @@
...
+  overflow: hidden auto;
+  max-height: 208px;
+`
+
+const AliasRow = styled.div<{ showDivider: boolean }>`
+  font-weight: ${font.default.semibold};
+  padding: ${spacing.m} ${spacing.l};
+  gap: 8px;
+  border-bottom: ${(p) =>
+    p.showDivider ? `1px solid ${color.divider.subtle}` : 'none'};
```
fallaciousreasoning: Hey, I think this will cause a small layout shift (border 1px to border 0px). Might be better to make the border color transparent

components/email_aliases/resources/content/email_alias_limit_reached.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+const AliasListBox = styled.div`
+  border: 1px solid ${color.divider.subtle};
+  border-radius: ${radius.xl};
+  overflow: hidden auto;
+  max-height: 208px;
+`
+
+const AliasRow = styled.div<{ showDivider: boolean }>`
+  font-weight: ${font.default.semibold};
```
fallaciousreasoning: Hey, so `font.default.semibold` won't work for setting the font-weight - you need to set the `font: font.default.semibold`

components/email_aliases/resources/content/email_alias_limit_reached.tsx
```diff
@@ -0,0 +1,96 @@
...
+  const count = aliases.length
+  return (
+    <SectionCol>
+      <Alert type='warning'>
+        <LimitAlertTitle>
+          {getLocale(S.SETTINGS_EMAIL_ALIASES_LIMIT_REACHED_ALERT_TITLE)}
+        </LimitAlertTitle>
+        <LimitDescription>
+          {formatLocale(S.SETTINGS_EMAIL_ALIASES_LIMIT_REACHED_ALERT_BODY, {
+            $1: String(aliasLimit),
```
fallaciousreasoning: Does this need to be cast to a string? If it does, then maybe `aliasLimit.toString()` would be more idiomatic

components/email_aliases/resources/content/email_alias_limit_reached.tsx
```diff
@@ -0,0 +1,96 @@
...
+          })}
+        </LimitDescription>
+      </Alert>
+      <div>
+        <ListLabel>
+          {formatLocale(S.SETTINGS_EMAIL_ALIASES_YOUR_ALIASES_COUNT_LABEL, {
+            $1: String(count),
+            $2: String(aliasLimit),
```
fallaciousreasoning: ditto - does it not work without the string cast?

components/email_aliases/resources/email_aliases_panel.html
```diff
@@ -16,8 +16,8 @@
     }
     :host, body, #mountPoint {
       display: block;
-      width: 420px;
-      max-width: 420px;
+      width: 512px;
+      max-width: 512px;
```
fallaciousreasoning: This is the one that shows in the dialog, right? Could we set it to be the width of the screen (100vw)?

components/email_aliases/resources/content/email_aliases_modal.tsx
```diff
@@ -314,10 +328,11 @@ export const EmailAliasModal = ({
           {getLocale(S.SETTINGS_EMAIL_ALIASES_BUBBLE_DESCRIPTION)}
         </ModalDescription>
       )}
-      {bubble && limitReached ? (
-        <WarningText>
-          {getLocale(S.SETTINGS_EMAIL_ALIASES_BUBBLE_LIMIT_REACHED)}
-        </WarningText>
+      {limitReached ? (
+        <EmailAliasLimitReached
+          aliases={aliases ?? []}
```
fallaciousreasoning: is it possible that aliases are undefined if we've hit the limit?

boocmp: Yes, it's impossible, I've replaced this with `aliases!`

fallaciousreasoning (approved):

netzenbot (approved):

mkarolin (approved):
`strings`++

PR #35224, opened 2026-04-02
Add disable-model-invocation to all Claude Code skills

bbondy:
## Summary

- Adds `disable-model-invocation: true` to all 14 skill frontmatter files (uplift already had it)
- Skills will only load into context when explicitly invoked via `/<skill-name>`, reducing baseline context token usage by ~1.2k tokens per session

## Test plan

- [x] Start a new Claude Code session and verify `/context` shows reduced skill token usage
- [x] Verify each skill still works when invoked explicitly (e.g. `/preflight`, `/pr`, `/review`)

### DIFF
--- a/.claude/skills/preflight/SKILL.md
+++ b/.claude/skills/preflight/SKILL.md
@@ -3,6 +3,7 @@ name: preflight
 description:
   'Run all preflight checks (format, gn_check, presubmit, build, tests) to make
   sure the current work is ready for review.'
+disable-model-invocation: true
 ---
 
 # Preflight Checks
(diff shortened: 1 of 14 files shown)


petemill (approved):

yrliou (approved):

darkdh (approved):

.claude/skills/preflight/SKILL.md
```diff
@@ -3,6 +3,7 @@ name: preflight
 description:
   'Run all preflight checks (format, gn_check, presubmit, build, tests) to make
   sure the current work is ready for review.'
+disable-model-invocation: true
```
bridiver, line 6: This prevents impl-review from running preflight because it's not visible to the model unless you invoke it directly

bbondy: This was intentional because otherwise it loads all skills into every Claude session and uses tokens.  Chaining officially isn't supported, but does work if you word things saying to manually discover and run the skill.

PR #35226, opened 2026-04-02
[NTP] Don't show the chat input if users have disabled search

### LINKED ISSUE #54177: [NTP] Leo AI input/widget appears when user previously toggled off `Show search widget in new tabs`
### Description

If users have `Show search widget in new tabs` toggled off in the Customize New Tab Page settings, when upgrading to a version or if receiving the griffin seed which has `#brave-ai-chat-show-input-on-new-tab-page` enabled, the `Show Leo AI widget in new tabs` setting should be toggled off.

Image

NTP widget with Leo AI input

Image

Related: https://github.com/brave/brave-browser/issues/48179

### Steps to reproduce

1. Use 1.89.x or higher. 
2. Toggle off `show search widget in the new tabs` setting
3. Enable #brave-ai-chat-show-input-on-new-tab-page and relaunch the browser
4. Notice the Leo AI input widget appears

### Actual result

Appears on new tab page on update when a user previously/explicitly hid the search widget.
Image

### Expected result

No Leo AI widget and `Show Leo AI widget in new tabs` setting toggled off.

### Reproduces how often

Easily reproduced

1.89.x and higher. Target 1.90.x for fix.

### Channel information

- [x] release (stable)
- [x] beta
- [x] nightly

### Reproducibility

### Miscellaneous information

_No response_

zenparsing:
Users who had hidden the search input before the chat input feature was added would unexpectedly see the chat input on upgrade. Default the chat input pref to match the search input pref on first access.

Resolves https://github.com/brave/brave-browser/issues/54177

### DIFF
--- a/browser/ui/webui/brave_new_tab_page_refresh/new_tab_page_initializer.cc
+++ b/browser/ui/webui/brave_new_tab_page_refresh/new_tab_page_initializer.cc
@@ -20,6 +20,7 @@
 #include "brave/components/ai_chat/core/common/buildflags/buildflags.h"
 #include "brave/components/brave_news/common/buildflags/buildflags.h"
 #include "brave/components/brave_rewards/core/buildflags/buildflags.h"
+#include "brave/components/brave_search_conversion/pref_names.h"
 #include "brave/components/brave_talk/buildflags/buildflags.h"
 #include "brave/components/brave_vpn/common/buildflags/buildflags.h"
 #include "brave/components/constants/pref_names.h"
@@ -128,6 +129,35 @@ void NewTabPageInitializer::Initialize() {
 
   content::URLDataSource::Add(GetProfile(),
                               std::make_unique<ThemeSource>(GetProfile()));
+}
+
+// static
+void NewTabPageInitializer::MigrateProfilePrefs(PrefService* prefs) {
+  // Added 2026-03: Migrate "hide all widgets" into individual widget prefs.
+  if (prefs->GetBoolean(kNewTabPageHideAllWidgets)) {
+    prefs->SetBoolean(kNewTabPageHideAllWidgets, false);
+    prefs->SetBoolean(kNewTabPageShowRewards, false);
+#if BUILDFLAG(ENABLE_BRAVE_TALK)
+    prefs->SetBoolean(brave_talk::prefs::kNewTabPageShowBraveTalk, false);
+#endif
+#if BUILDFLAG(ENABLE_BRAVE_VPN)
+    prefs->SetBoolean(kNewTabPageShowBraveVPN, false);
+#endif
+  }
+  prefs->ClearPref(kNewTabPageHideAllWidgets);
+
+  // Added 2026-04: Set chat input visibility to hidden if the user has
+  // explicitly hidden the search input.
+  using brave_search_conversion::prefs::kMigratedNTPChatInputFromSearch;
+  using brave_search_conversion::prefs::kShowNTPChatInput;
+  using brave_search_conversion::prefs::kShowNTPSearchBox;
+  if (!prefs->GetBoolean(kMigratedNTPChatInputFromSearch)) {
+    prefs->SetBoolean(kMigratedNTPChatInputFromSearch, true);
+    if (auto* user_value = prefs->GetUserPrefValue(kShowNTPSearchBox);
+        user_value && !user_value->GetBool()) {
+      prefs->SetBoolean(kShowNTPChatInput, false);
+    }
+  }
 }
 
 Profile* NewTabPageInitializer::GetProfile() {
--- a/browser/ui/webui/brave_new_tab_page_refresh/new_tab_page_initializer.h
+++ b/browser/ui/webui/brave_new_tab_page_refresh/new_tab_page_initializer.h
@@ -20,6 +20,7 @@ namespace regional_capabilities {
 class RegionalCapabilitiesService;
 }
 
+class PrefService;
 class Profile;
 
 namespace brave_new_tab_page_refresh {
@@ -40,6 +41,9 @@ class NewTabPageInitializer {
   NewTabPageInitializer& operator=(const NewTabPageInitializer&) = delete;
 
   void Initialize();
+
+  // Migrates profile prefs associated with the NTP.
+  static void MigrateProfilePrefs(PrefService* prefs);
 
  private:
   Profile* GetProfile();
--- a/chromium_src/chrome/browser/prefs/browser_prefs.cc
+++ b/chromium_src/chrome/browser/prefs/browser_prefs.cc
@@ -18,7 +18,6 @@
 #include "brave/components/brave_shields/content/browser/ad_block_service.h"
 #include "brave/components/brave_shields/core/browser/brave_shields_p3a.h"
 #include "brave/components/brave_sync/brave_sync_prefs.h"
-#include "brave/components/brave_talk/buildflags/buildflags.h"
 #include "brave/components/brave_vpn/common/buildflags/buildflags.h"
 #include "brave/components/brave_wallet/common/buildflags/buildflags.h"
 #include "brave/components/constants/pref_names.h"
@@ -59,12 +58,9 @@
 #include "brave/browser/brave_rewards/rewards_prefs_util.h"
 #endif
 
-#if BUILDFLAG(ENABLE_BRAVE_TALK)
-#include "brave/components/brave_talk/pref_names.h"
-#endif
-
 #if !BUILDFLAG(IS_ANDROID)
 #include "brave/browser/ui/tabs/brave_tab_prefs.h"
+#include "brave/browser/ui/webui/brave_new_tab_page_refresh/new_tab_page_initializer.h"
 #include "brave/browser/ui/webui/welcome_page/brave_welcome_ui_prefs.h"
 #endif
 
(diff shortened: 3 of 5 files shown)


browser/ui/webui/brave_new_tab_page_refresh/new_tab_page_handler.cc
```diff
@@ -207,6 +207,18 @@ void NewTabPageHandler::SetShowSearchBox(bool show_search_box,
...
+
+  // Default the value of the "show AI chat input" pref to the current value of
+  // the "show search input" pref. This ensures that users who have
+  // intentionally hidden the search input will not be shown the chat input
+  // without intentionally enabling it.
+  if (!pref_service_->GetUserPrefValue(kShowNTPChatInput)) {
+    bool should_show = pref_service_->GetBoolean(kShowNTPSearchBox);
+    pref_service_->SetBoolean(kShowNTPChatInput, should_show);
+  }
+
```
petemill: Isn't the problem here that this will occur for any change of the SearchBox pref value?

So if I choose to hide the search box with the intention of keeping the AI Chat box, then when I refresh the NTP I will lose the AI Chat box.

At that point it becomes impossible to hide search but keep AI Chat.

Since the Issue describes wanting users who have previous to the version which introduces the `kShowNTPChatInput` pref have that pref defaulted to their value of `kShowNTPSearchBox` we would normally use a migration function for it, e.g. `ModelService::MigrateProfilePrefs`.
We would ensure the migration only happens once with another pref, e.g. `kHasCreatedDefaulNTPChatInputValue`. And then after a couple versions of release, we would remove the migration and clear the migration tracker pref.

petemill (changes requested):

browser/ui/webui/brave_new_tab_page_refresh/new_tab_page_handler.cc
zenparsing: This is triggered from `MigrateObsoleteProfilePrefs` correct? I considered that option but was discouraged by the fact that this is not in fact an obsolete pref.

petemill: Well yeah I'm not sure why it's called that. But it needs to be done after RefisterProfilePrefs. You can do it in the constructor of a service but I wouldn't do it in the WebUI class.

browser/ui/webui/brave_new_tab_page_refresh/new_tab_page_handler.cc
```diff
@@ -207,6 +207,18 @@ void NewTabPageHandler::SetShowSearchBox(bool show_search_box,
...
 
 void NewTabPageHandler::GetShowChatInput(GetShowChatInputCallback callback) {
+  using brave_search_conversion::prefs::kShowNTPChatInput;
+  using brave_search_conversion::prefs::kShowNTPSearchBox;
+
+  // Default the value of the "show AI chat input" pref to the current value of
+  // the "show search input" pref. This ensures that users who have
+  // intentionally hidden the search input will not be shown the chat input
+  // without intentionally enabling it.
+  if (!pref_service_->GetUserPrefValue(kShowNTPChatInput)) {
```
netzenbot: nit: `should_show` is never modified — prefer `const bool should_show = ...`. ([Prefer `const` for Local Variables Where Possible](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-064))

browser/ui/webui/brave_new_tab_page_refresh/new_tab_page_handler.cc
zenparsing: Thanks for the feedback. I've moved the migration out of the getter and into a `NewTabPageInitializer::MigrateProfilePrefs` static method.

I put it on `NewTabPageInitializer` to mirror how Chromium handles this. Chromium's `NewTabPageUI` exposes pref migration functions, called from `browser_prefs.cc`. `NewTabPageInitializer` makes sense for us because it is already responsible for this kind of work and has a much lighter include footprint than `BraveNewTabPageUI`.

Since we now have a proper place to put these migrations, I also moved the `kNewTabPageHideAllWidgets` migration (added recently) into the same function.

For the guard, I chose to stick with `GetUserPrefValue` rather than introducing a separate tracking pref. Once the pref has been set (either by the migration or by the user), the check is a no-op. A tracking pref would be redundant since the pref's own state already tells us whether the migration has run. Either way, the migration can be removed after a few versions.

browser/ui/webui/brave_new_tab_page_refresh/new_tab_page_initializer.cc
```diff
@@ -96,6 +97,29 @@ std::string_view GetSearchDefaultHost(
   return kBraveSearchHost;
 }
```
netzenbot, line 99: nit: `MigrateProfilePrefs` is placed before the constructor in the `.cc`, but in the `.h` it's declared after `Initialize()`. The `.cc` definition order should match the `.h` declaration order. ([Function Ordering in `.cc` Should Match `.h`](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-017))

browser/ui/webui/brave_new_tab_page_refresh/new_tab_page_initializer.h
```diff
@@ -41,6 +42,8 @@ class NewTabPageInitializer {
 
   void Initialize();
```
netzenbot, line 44: nit: `MigrateProfilePrefs` is a public static method with non-obvious constraints (called once from `MigrateObsoleteProfilePrefs`, idempotent via `GetUserPrefValue` guard). A brief comment describing when to call it and what it migrates would help future readers. ([Method Documentation Should Describe the Contract](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-004))

browser/ui/webui/brave_new_tab_page_refresh/new_tab_page_initializer.cc
```diff
@@ -96,6 +97,29 @@ std::string_view GetSearchDefaultHost(
...
+#endif
+  }
+  prefs->ClearPref(kNewTabPageHideAllWidgets);
+
+  // Added 2026-04: Default chat input visibility to match search input.
+  using brave_search_conversion::prefs::kShowNTPChatInput;
+  using brave_search_conversion::prefs::kShowNTPSearchBox;
+  if (!prefs->GetUserPrefValue(kShowNTPChatInput)) {
+    prefs->SetBoolean(kShowNTPChatInput, prefs->GetBoolean(kShowNTPSearchBox));
+  }
```
petemill: Logically this will have the desired effect but we're always explicitly setting a pref value, and I don't think that's the best solution. If we ever change the default, no users will follow that default. UserPrefValue should be an indication of the user's intent, but here we're setting it even if the user did not explicitly have an intent behind even `kShowNTPSearchBox` (in most cases that will just be the default non-user-set value. Usually we're pretty particular about not setting UserPrefValue unless the user sets it.

For example in `components/brave_wallet/browser/pref_names.cc` there is `RegisterProfilePrefsForMigration`, called from `browser/brave_profile_prefs.cc`'s `RegisterProfilePrefsForMigration` (but then that's just called from `RegisterProfilePrefs`!)

zenparsing: Good points. I've added a tracking pref (`kMigratedNTPChatInputFromSearch`) and the migration now only copies `kShowNTPSearchBox` into `kShowNTPChatInput` when the user has explicitly set it to `false`.

petemill (approved):

netzenbot (approved):

emerick (approved):

LaurenWags:
Verified with

```
Brave | 1.91.132 Chromium: 148.0.7778.60 (Official Build) beta (arm64)
-- | --
Revision | d6e449653b4539ad37c9f4fc186c30427fd7e91a
OS | macOS Version 26.3 (Build 25D125)
```

Reproduced the issue using `1.90.117` :

1. Installed 1.89.145
2. Opened Customize Dashboard from NTP
3. Toggled Search widget off from NTP
4. Closed browser
5. Renamed profile to be for Beta
6. Launched 1.90.117
7. Opened NTP --> the "Leo" widget was visible on the NTP

1.89.x | 1.90.x
--- | ---
Screenshot 2026-04-30 at 5 27 14 PM | Screenshot 2026-04-30 at 5 28 44 PM
Screenshot 2026-04-30 at 5 27 21 PM | Screenshot 2026-04-30 at 5 28 26 PM

Using `1.91.132` and same STR, confirmed after updating from `1.89.x` with the "Search" widget off, to `1.91.132`, the Leo widget was not visible on the NTP:

1. Installed 1.89.145
2. Opened Customize Dashboard from NTP
3. Toggled Search widget off from NTP
4. Closed browser
5. Renamed profile to be for Beta
6. Launched 1.91.132
7. Opened NTP --> the "Leo" widget was not visible on the NTP, confirmed Leo widget was toggled off on the Customize Dashboard modal also

1.89.x | 1.91.x
--- | ---
Screenshot 2026-04-30 at 5 27 14 PM | Screenshot 2026-04-30 at 5 32 19 PM
Screenshot 2026-04-30 at 5 27 21 PM | Screenshot 2026-04-30 at 5 32 04 PM

PR #35229, opened 2026-04-02
[AI Chat] Allow both Active and ActiveDisconnected users to attempt credential refresh

### LINKED ISSUE #54375: [AI Chat] Bug credential refresh for ActiveDisconnected users
### Description

There is an issue where leo premium requests are being routed to non-premium endpoint. This is likely due to credential checking in the browser code for Active and ActiveDisconnected states.

### Steps to reproduce

1. Seems like you need to have an active conversation in non-premium then log in to a premium account then continue the same conversation. @srirambv for more details
2. The request should fail by sending the premium request to the non-premium endpoint

### Actual result

When switching to premium the request should succeed in the same conversation.

### Expected result

When switching to premium the request should succeed in the same conversation.

### Reproduces how often

Intermittent issue

1.89+

### Channel information

- [x] release (stable)
- [x] beta
- [x] nightly

### Reproducibility

### Miscellaneous information

Related PR: https://github.com/brave/brave-core/pull/35229

porteron:
#### Resolves the bug: 
`IsPremiumStatus()` returns true for both `mojom::PremiumStatus::Active` and `mojom::PremiumStatus::ActiveDisconnected`. But `OnGetPremiumStatus()` in `AIChatCredentialManager` only attempts to fetch a credential — and thus only enables the premium URL — when status is strictly Active. For `ActiveDisconnected`, it returns `std::nullopt`, setting `premium_enabled = false`, which causes `GetEndpointUrl` to use the non-premium URL.

#### Solution:
- Allow both Active and ActiveDisconnected users to attempt credential refresh. ActiveDisconnected means the subscription is valid but needs to refresh — PrepareCredentialsPresentation may still succeed, and the UI already treats this state as premium via IsPremiumStatus().

Resolves https://github.com/brave/brave-browser/issues/54375

### DIFF
--- a/components/ai_chat/core/browser/ai_chat_credential_manager.cc
+++ b/components/ai_chat/core/browser/ai_chat_credential_manager.cc
@@ -241,7 +241,8 @@ void AIChatCredentialManager::OnGetPremiumStatus(
         callback,
     mojom::PremiumStatus status,
     mojom::PremiumInfoPtr info) {
-  if (status != mojom::PremiumStatus::Active) {
+  if (status != mojom::PremiumStatus::Active &&
+      status != mojom::PremiumStatus::ActiveDisconnected) {
     std::move(callback).Run(std::nullopt);
     return;
   }


petemill (approved):

netzenbot (approved):

DJAndries (approved):

srirambv:
Verification passed on 

Brave | 1.91.88 Chromium: 147.0.7727.102 (Official Build) nightly (arm64)
-- | --
Revision | 0a347b7e54e42b1945f4da8ee33c9ed23bd57c6b
OS | Windows 11 Version 25H2 (Build 26200.8246)

- Verified switching over from free user to premium user, API end point also switches from `https://ai-chat.bsg.brave.com/` to `https://ai-chat-premium.bsg.brave.com/`
- Verified no network error message is shown when the conversation gets long or when creating a new conversation

PR #35250, opened 2026-04-03
Reland #35004 "Replace kLocalAIModels with upstream kHistoryEmbeddings"

### LINKED ISSUE #53873: Land BraveEmbedder behind flag brave-history-embeddings
Initial attempt to integrate history semantic search with our own embedder

### LINKED ISSUE #53954: Drop kLocalAIModels and reuse history_embeddings::kHistoryEmbedding flag
We currently have `kLocalAIModels` which guarded model download and we don't want users to enable two flags in order to use brave history embedding feature.

darkdh:
Use chromium's existing kHistoryEmbeddings feature flag instead of maintaining a separate kLocalAIModels flag. Replace all kLocalAIModels references with history_embeddings::kHistoryEmbeddings and remove kLocalAIModels.

- Re-expose the upstream history-embeddings flag entry in about:flags (the upstream entry expired at M145)
- Override IsHistoryEmbeddingsEnabledForProfile() to bypass upstream's per-profile OptimizationGuide pref check (ShouldFeatureBeCurrentlyEnabledForUser(kHistorySearch)), which Brave doesn't set up

No need to override IsHistoryEmbeddingsFeatureEnabled(): upstream's implementation only auto-enables the feature for us-en-US profiles when kLaunchedHistoryEmbeddings is enabled, and we already force kLaunchedHistoryEmbeddings to FEATURE_DISABLED_BY_DEFAULT (pre-existing override). That kills upstream's auto-launch fallback, leaving kHistoryEmbeddings as the sole gate for Brave.

Now we have proper handling for cr147 in https://github.com/brave/brave-core/pull/35130 so it is safe to reland https://github.com/brave/brave-core/pull/35004 which was reverted in https://github.com/brave/brave-core/pull/35158

Depends on https://github.com/brave/brave-core/pull/32689

Resolves https://github.com/brave/brave-browser/issues/53954
Resolves https://github.com/brave/brave-browser/issues/53873


components/local_ai/core/features.h
```diff
@@ -6,12 +6,4 @@
...
 
-#include "base/feature_list.h"
-
-namespace local_ai::features {
-
-BASE_DECLARE_FEATURE(kLocalAIModels);
-
-}  // namespace local_ai::features
-
 #endif  // BRAVE_COMPONENTS_LOCAL_AI_CORE_FEATURES_H_
```
netzenbot, line 9: This PR empties `features.h` removing `kLocalAIModels`, but `local_ai_service.cc` (line 100) and `browser/browser_context_keyed_service_factories.cc` (line 151) still include this header and reference `features::kLocalAIModels` / `local_ai::features::kLocalAIModels`. These files will fail to compile — they need to be updated in this PR to use `history_embeddings::kHistoryEmbeddings` instead.

darkdh: you are hallucinating and we don't have that changes anymore, those were old changes.

darkdh: If these changes won't compile, how come the CI results are all green?

chromium_src/chrome/browser/history_embeddings/history_embeddings_utils.cc
```diff
@@ -3,14 +3,34 @@
...
+#undef IsHistoryEmbeddingsFeatureEnabled
+
 namespace history_embeddings {
 
 OVERRIDE_FEATURE_DEFAULT_STATES({{
     {kLaunchedHistoryEmbeddings, base::FEATURE_DISABLED_BY_DEFAULT},
 }});
 
+bool IsHistoryEmbeddingsFeatureEnabled() {
+  return base::FeatureList::IsEnabled(kHistoryEmbeddings);
```
bridiver: I think upstream has changed since the original PR, it has `kLaunchedHistoryEmbeddings` which gates it independently of kHistoryEmbeddings

darkdh: yes, this override is not needed since upstream function is now 
```
bool IsHistoryEmbeddingsFeatureEnabled() {
#if BUILDFLAG(IS_CHROMEOS)
  if (!chromeos::features::IsFeatureManagementHistoryEmbeddingEnabled()) {
    return false;
  }
#endif
  // If the feature is overridden manually or via Finch, return its value.
  if (base::FeatureList::GetStateIfOverridden(kHistoryEmbeddings).has_value()) {
    return base::FeatureList::IsEnabled(kHistoryEmbeddings);
  }
  // Otherwise return true for "us" and "en-US", leaving a Finch hook just in
  // case.
  return IsCountryAndLocale("us", "en-US") &&
         base::FeatureList::IsEnabled(kLaunchedHistoryEmbeddings);
}
```
it will return the kHistoryEmbeddings overridden locally or by Griffin/finch.
If it is not overridden, we don't want to enable the feature automatically for US users which upstream set `kLaunchedHistoryEmbeddings` enabled by default and it should remain disabled by default in our override (we didn't change it in our PR)

bridiver (approved):

PR #35316, opened 2026-04-07
[ads] Support empty data-conversion-url-pattern-value for search result ad JSON-LD

### LINKED ISSUE #37155: [ads] Support empty `data-conversion-url-pattern-value` for search result ad JSON-LD

tmancey:
An empty data-conversion-url-pattern-value in JSON-LD previously caused the entire ad to be rejected because GetRequiredStringValue treats empty strings as invalid. GetStringValue is renamed to GetOptionalStringValue and used for url_pattern so that empty values are accepted and treated as match-all. CreativeSetConversionInfo::IsValid no longer rejects an empty url_pattern and DoesCreativeSetConversionUrlPatternMatchRedirectChain returns true for an empty pattern, meaning the conversion fires regardless of the visited URL.

## Test plan
Need to make sure that parsing of search result ads with conversion is not broken.

* Start the browser
* Open search.brave.com
* Trigger a search result ad with conversion
* Click the search result ad
* Open `brave://ads-internals`
EXPECTATION:
  - Search conversion is presented in `Active Brave Ads conversion URL patterns`

Resolves https://github.com/brave/brave-browser/issues/37155

### DIFF
--- a/components/brave_ads/core/internal/creatives/conversions/creative_set_conversion_builder.cc
+++ b/components/brave_ads/core/internal/creatives/conversions/creative_set_conversion_builder.cc
@@ -18,12 +18,16 @@ FromMojomMaybeBuildCreativeSetConversion(
     return std::nullopt;
   }
 
+  if (!mojom_creative_ad->creative_set_conversion->url_pattern) {
+    return std::nullopt;
+  }
+
   CreativeSetConversionInfo creative_set_conversion;
 
   creative_set_conversion.id = mojom_creative_ad->creative_set_id;
 
   creative_set_conversion.url_pattern =
-      mojom_creative_ad->creative_set_conversion->url_pattern;
+      *mojom_creative_ad->creative_set_conversion->url_pattern;
 
   creative_set_conversion.observation_window =
       mojom_creative_ad->creative_set_conversion->observation_window;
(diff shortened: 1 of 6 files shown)


aseren (approved):

components/brave_ads/core/internal/creatives/conversions/creative_set_conversion_builder.cc
```diff
@@ -18,6 +18,11 @@ FromMojomMaybeBuildCreativeSetConversion(
     return std::nullopt;
   }
 
+  if (mojom_creative_ad->creative_set_conversion->url_pattern.empty()) {
+    // An empty URL pattern means conversion tracking is disabled for this ad.
```
netzenbot: The empty `url_pattern` string is being used as a sentinel value meaning "conversion tracking disabled". Per [CSA-022](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-022), prefer `std::optional<std::string>` over sentinel values like `""`. Consider making `url_pattern` a nullable mojo field (`string? url_pattern`) so that `std::nullopt` (mapped from null) can be used explicitly rather than checking `.empty()`.

tmancey: Fixed.

PR #35339, opened 2026-04-07
Script for pruning obsolete upstream test filters

emerick:
Python-based script for pruning obsolete upstream filters from our test filter files.


tools/cr/prune_test_filters.py
```diff
@@ -0,0 +1,381 @@
+#!/usr/bin/env vpython3
```
netzenbot, line 1: The other scripts in `tools/cr/` all have accompanying `*_test.py` files (e.g., `repository_test.py`, `brockit_test.py`). `prune_test_filters.py` has non-trivial logic — section parsing, protected-filter matching via `fnmatch`, line-ending preservation — and should follow the same pattern. Consider adding a `prune_test_filters_test.py` covering at least the `scan_filter_files` and `list_tests_from_binary` functions. ([PRs Should Include Reasonable Test Coverage](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-039))

tools/cr/prune_test_filters.py
```diff
@@ -0,0 +1,381 @@
...
+
+        flush_section()
+
+        while new_lines and new_lines[-1] == "":
+            if header_blank_line_preserved and len(
+                    new_lines) == len(file_header) + 1:
+                break  # preserve the blank line after file header
+            new_lines.pop()
+
+        # Preserve original line endings
```
netzenbot: nit: BS-029 requires `newline='\n'` for consistent LF line endings when writing files. Using `newline=""` disables translation entirely, which can produce CRLF output on Windows. If preserving the original file's line endings is intentional, a comment explaining why would help. ([Python File Writes: Use `newline='\n'`](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-029))

tools/cr/prune_test_filters.py
```diff
@@ -0,0 +1,381 @@
...
+        print("\nObsolete test filters:")
+        for path, tests in obsolete_by_file.items():
+            print(f"\nFile: {path}")
+            for t in tests:
+                print(f"  {t}")
+    else:
+        print("\nNo obsolete filters found.")
+
+    if args.apply and obsolete_by_file:
+        print("\nFilter files updated.")
```
netzenbot, line 396: nit: `scan_filter_files()` and `collect_tests_per_binary()` are the most complex functions in this script but lack docstrings. A brief description of parameters (`apply`, `sort`, return type `defaultdict`) would help future maintainers — especially since the other functions like `list_tests_from_binary()` and `is_filter_applicable()` do have docstrings. ([DOC-004](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-004))

tools/cr/prune_test_filters.py
emerick: Tests added.

tools/cr/prune_test_filters.py
emerick: newline='\n' added.

tools/cr/prune_test_filters.py
emerick: Docstrings add for all non-trivial functions.

tools/cr/prune_test_filters.py
```diff
@@ -0,0 +1,397 @@
...
+    for plat_key, plat_tag in supported_platforms.items():
+        if platform_info.startswith(plat_key):
+            if plat_tag == current_platform:
+                return True
+            print(
+                f"Skipping {filename} (platform '{plat_key}' does not match '{current_platform}')"
+            )
+            return False
+
+    return True
```
mkarolin: This seems odd: we would only get here if the filter has a name with a `-`, but doesn't match any of our platforms, so this should probably raise an error of unexpected file pattern?

tools/cr/prune_test_filters.py
```diff
@@ -0,0 +1,397 @@
...
+                current_suite = suite_part
+            continue
+
+        main_part = stripped.split("#", 1)[0].strip().rstrip(".")
+        if not main_part:
+            continue
+
+        full_name = f"{current_suite}.{main_part}" if current_suite else main_part
+        allowed_chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789_/."
+        if all(c in allowed_chars for c in full_name):
```
mkarolin: Is this a feasible concern?

tools/cr/prune_test_filters.py
```diff
@@ -0,0 +1,397 @@
...
+        obsolete_by_file map.
+    """
+    obsolete_by_file = defaultdict(list)
+
+    for path in FILTER_DIR.rglob("*.filter"):
+        if not is_filter_applicable(path, current_platform):
+            continue
+
+        binary_name = get_binary_name_from_filter(path)
+        if not binary_name or binary_name not in tests_by_binary:
```
mkarolin, line 230: Seems like `binary_name not in tests_by_binary` should be an error instead of being quietly ignored.

tools/cr/prune_test_filters.py
```diff
@@ -0,0 +1,397 @@
...
+
+    if not FILTER_DIR.exists():
+        sys.stderr.write(
+            f"Error: filter directory does not exist: {FILTER_DIR}\n")
+        sys.exit(1)
+
+    current_platform = get_current_platform_tag()
+    if current_platform == "unknown":
+        sys.stderr.write(
+            "Warning: unknown platform, all filters treated as generic\n")
```
mkarolin: Personally, I think this should just be an error and exit.

tools/cr/prune_test_filters.py
emerick: Good catch, fixed.

tools/cr/prune_test_filters.py
emerick: Nope, removed.

tools/cr/prune_test_filters.py
emerick: Agreed, fixed.

tools/cr/prune_test_filters.py
emerick: Agreed, fixed.

netzenbot (approved):

mkarolin (approved):
++

bbondy:
@emerick this might need adjusting see https://github.com/brave/brave-core/pull/35890

emerick:
> @emerick this might need adjusting see #35890

I think I see the problem I'll get a PR going for that, thank you!

emerick:
> > @emerick this might need adjusting see #35890
> 
> I think I see the problem I'll get a PR going for that, thank you!

Actually, those tests are Linux-only tests (they were disabled on other platforms upstream, due to flakiness). But we had the filter for them in a general (platform-independent) filter file. Since I ran the pruning script on Windows, it literally didn't see those tests when it ran `browser_tests --gtest_list_tests` and thus removed the filter. If the filter were in the correct Linux-specific filter file it wouldn't have been removed. Given that, I think nothing really needs to be adjusted here as this is a bit of an edge case.

PR #35375, opened 2026-04-08
Support Brave Wallet accounts hiding.

### LINKED ISSUE #19761: Add a way to hide / remove accounts
Since accounts are derivation paths.  Accounts can't really be removed but we can give the impression of that in the UI by hiding a "removed" account. We should have a way to get back at a removed account though too.

cypt4:
This PR extends KeyringService with new functions to manage hidden accounts.
And adds corresponding UI to the wallet.
Hidden accounts are stored in a separate preference.
Default account can't be hidden.
Resolves https://github.com/brave/brave-browser/issues/19761
Sec review : https://github.com/brave/reviews/issues/2221

Screenshot 2026-03-24 at 22 07 41

Test cases:
Test cases should be checked on all platforms(desktop, andoroid, ios).
1) Test that we can hide derived accounts except of first account for coin. This function works only for derived accounts.
2) Check that general wallet functionality works when all accounts are hidden. Like portfolio, assets.
3) Check hiding of selected accounts and selected dapp accounts. If selected account is first account should be selected for coin. If selected dapp account is hidden then first coin account should be selected.
4) Check unhiding accounts.
5) dApp connection should not work after account is hiden for SOL, ADA, ETH.
6) Check testnets.

Screenshot 2026-04-15 at 22 30 06


components/brave_wallet/browser/keyring_service.cc
```diff
@@ -2721,6 +2734,102 @@ void KeyringService::SetAccountName(mojom::AccountIdPtr account_id,
...
+  }
+
+  ScopedListPrefUpdate update(profile_prefs_, kBraveWalletHiddenAccounts);
+  if (!std::ranges::contains(*update, base::Value(account_id->unique_key))) {
+    update->Append(account_id->unique_key);
+    NotifyAccountsChanged();
+  }
+
+  std::move(callback).Run(true);
+}
```
netzenbot, line 2839: The call `RevokeHiddenAccountPermisson(account_id.Clone(), base::BindOnce(..., std::move(account_id), ...))` has undefined evaluation order between the two arguments. If the `BindOnce` argument is evaluated first, `account_id` is moved before `.Clone()` is called, producing a clone of a moved-from object. Fix by cloning first: `auto cloned = account_id.Clone(); revoker->RevokeHiddenAccountPermisson(std::move(cloned), base::BindOnce(..., std::move(account_id), ...));` ([❌ Don't Use a Variable After `std::move()`](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-memory.md#CSM-030))

components/brave_wallet/browser/keyring_service.h
```diff
@@ -168,6 +169,14 @@ class KeyringService : public mojom::KeyringService {
   void SetAccountName(mojom::AccountIdPtr account_id,
                       const std::string& name,
                       SetAccountNameCallback callback) override;
+  void InitializeHiddenAccountPermissionRevoker(
```
netzenbot: `InitializeHiddenAccountPermissionRevoker` is a two-phase initialization setter rather than constructor injection. `delegate_` is already available when `keyring_service_` is constructed (declaration order guarantees this), so the revoker could be passed directly to `KeyringService`'s constructor. See [ARCH-025](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-025).

components/brave_wallet/browser/keyring_service.cc
```diff
@@ -2721,6 +2734,102 @@ void KeyringService::SetAccountName(mojom::AccountIdPtr account_id,
   NOTREACHED() << account_id->kind;
 }
 
+void KeyringService::GetHiddenAccounts(GetHiddenAccountsCallback callback) {
+  std::move(callback).Run(GetHiddenAccountsSync());
+}
```
netzenbot, line 2776: Mixed sync/async callback: the early-exit paths call `callback.Run(false)` synchronously (wrong kind, nonexistent account), while the success path is async via `RevokeHiddenAccountPermisson`. Callers cannot reliably reason about when their callback fires. The early exits should post asynchronously or the method should be redesigned to be uniformly async. See [ARCH-060](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-060).

components/brave_wallet/browser/keyring_service.cc
```diff
@@ -2721,6 +2734,102 @@ void KeyringService::SetAccountName(mojom::AccountIdPtr account_id,
...
+    return;
+  }
+
+  CHECK(hidden_account_permission_revoker_);
+  hidden_account_permission_revoker_->RevokeHiddenAccountPermisson(
+      account_id.Clone(),
+      base::BindOnce(&KeyringService::OnHiddenAccountPermissionsRevoked,
+                     weak_ptr_factory_.GetWeakPtr(), std::move(account_id),
+                     std::move(callback)));
+}
```
supermassive: Let's 
1) Check if already hidden(per prefs) and do nothing if already is.
2) Hide in prefs and then go to RevokeHiddenAccountPermisson
3) Ignore result from RevokeHiddenAccountPermisson, most likely it may fail only in edge cases and we can't recover from it. Also there is no way to explain user what exactly they should to make account hidden in case of error.

components/brave_wallet/browser/brave_wallet_constants.h
```diff
@@ -687,6 +687,7 @@ inline constexpr webui::LocalizedString kLocalizedStrings[] = {
     {"braveWalletAddAsset", IDS_BRAVE_WALLET_ADD_ASSET},
     {"braveWalletAccountsExport", IDS_BRAVE_WALLET_ACCOUNTS_EXPORT},
     {"braveWalletAccountsDeposit", IDS_BRAVE_WALLET_ACCOUNTS_DEPOSIT},
+    {"braveWalletAccountsHide", IDS_BRAVE_WALLET_NFTS_TAB_HIDE},
```
supermassive: Is it supposed to be referenced now?

components/brave_wallet/browser/BUILD.gn
```diff
@@ -62,6 +62,8 @@ static_library("browser") {
     "blockchain_list_parser.h",
     "blockchain_registry.cc",
     "blockchain_registry.h",
+    "brave_wallet_hidden_accounts_permissions_revoker.cc",
+    "brave_wallet_hidden_accounts_permissions_revoker.h",
     "brave_wallet_ipfs_service.cc",
```
supermassive: Makes sense to remove brave_wallet prefix

components/brave_wallet/browser/keyring_service_prefs.cc
```diff
@@ -6,6 +6,7 @@
 #include "brave/components/brave_wallet/browser/keyring_service_prefs.h"
 
 #include <string>
+#include <string_view>
 #include <utility>
```
supermassive: ~#include <string_view>~

components/brave_wallet/browser/keyring_service.cc
cypt4: This is fine since this is a mojom method

components/brave_wallet/browser/hidden_accounts_permissions_revoker_unittest.cc
```diff
@@ -0,0 +1,188 @@
...
+  EXPECT_CALL(delegate,
+              ResetPermission(mojom::CoinType::ETH, Eq(expected_origin),
+                              Eq(expected_account_identifier)))
+      .Times(1)
+      .WillOnce(testing::Return(true));
+
+  BraveWalletHiddenAccountsPermissionsRevoker revoker(keyring_service,
+                                                      delegate);
+  keyring_service.NotifyAccountsChangedForTest();
+  task_environment_.RunUntilIdle();
```
netzenbot: The presubmit flagged `task_environment_.RunUntilIdle()` here (and at line 184) as a banned pattern. Replace with an explicit synchronization mechanism: since the Mojo callback is async, use `base::test::TestFuture` or `base::RunLoop::QuitClosure()` to wait only until the expected condition is met. ([❌ NEVER Use RunUntilIdle()](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-async.md#TA-005))

components/brave_wallet/browser/keyring_service.cc
```diff
@@ -2721,6 +2733,82 @@ void KeyringService::SetAccountName(mojom::AccountIdPtr account_id,
...
+    std::move(callback).Run(false);
+    return;
+  }
+
+  ScopedListPrefUpdate update(profile_prefs_, kBraveWalletHiddenAccounts);
+  update->EraseIf([&](const base::Value& value) {
+    const auto* unique_key = value.GetIfString();
+    if (!unique_key || *unique_key != account_id->unique_key) {
+      return false;
+    }
```
netzenbot, line 2819: `NotifyAccountsChanged()` is called even when `EraseIf` removes nothing (the absent-account case). Observers should only fire on actual state changes per [ARCH-021](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-021). Track whether any item was erased (e.g., compare list size before/after, or use a bool set inside the lambda) and only notify if the list was actually modified.

components/brave_wallet/browser/keyring_service_unittest.cc
```diff
@@ -3076,6 +3096,127 @@ TEST_F(KeyringServiceUnitTest, SetAccountName_HardwareAccounts) {
...
+    EXPECT_CALL(observer, AccountsChanged()).Times(0);
+    EXPECT_FALSE(AddHiddenAccount(&service, first_account_id.Clone()));
+    observer.WaitAndVerify();
+
+    hidden_accounts = GetHiddenAccounts(&service);
+    EXPECT_TRUE(hidden_accounts.empty());
+  }
+
+  // Test case: Hide a non-first derived account.
+  {
```
netzenbot, line 3152: `GetSelectedWalletAccount()` can return null (e.g. if all accounts are hidden). Dereferencing its result directly in `EXPECT_EQ` will crash the test process rather than producing a readable failure. Add `ASSERT_TRUE(service.GetSelectedWalletAccount())` before each dereference. Same issue at the check ~10 lines later after `RemoveHiddenAccount`. ([✅ Use ASSERT for Preconditions That Subsequent Code Depends On](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-034))

components/brave_wallet/browser/keyring_service.cc
```diff
@@ -2721,6 +2733,82 @@ void KeyringService::SetAccountName(mojom::AccountIdPtr account_id,
...
+    return true;
+  });
+  NotifyAccountsChanged();
+
+  std::move(callback).Run(true);
+}
```
netzenbot, line 2827: `RemoveHiddenAccount` always calls `NotifyAccountsChanged()` regardless of whether `EraseIf` actually removed anything. This causes unnecessary cache invalidation and observer churn when the account was not hidden to begin with. Check the return value of `EraseIf` (or track whether any element was removed) and skip the notification when nothing changed. ([ARCH-051](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-051))

components/brave_wallet/browser/keyring_service.cc
```diff
@@ -335,6 +336,17 @@ struct ImportedAccountInfo {
   std::optional<uint32_t> bitcoin_next_change_address_index;
 };
 
+bool IsAccountHidden(PrefService* profile_prefs, std::string_view unique_key) {
+  for (const auto& hidden_account :
+       profile_prefs->GetList(kBraveWalletHiddenAccounts)) {
+    const auto* hidden_unique_key = hidden_account.GetIfString();
+    if (hidden_unique_key && *hidden_unique_key == unique_key) {
+      return true;
```
netzenbot: nit: `IsAccountHidden` re-implements what `std::ranges::contains` already provides. `AddHiddenAccount` in this same PR already uses `std::ranges::contains(*update, base::Value(account_id->unique_key))`; the same approach works here: `return std::ranges::contains(profile_prefs->GetList(kBraveWalletHiddenAccounts), base::Value(std::string(unique_key)));`. ([CSA-001](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-001))

components/brave_wallet/browser/keyring_service.cc
```diff
@@ -335,6 +336,17 @@ struct ImportedAccountInfo {
   std::optional<uint32_t> bitcoin_next_change_address_index;
 };
```
netzenbot, line 338: nit: `IsAccountHidden` can be simplified with `std::ranges::any_of` per [CSA-042](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-042). Example: `return std::ranges::any_of(profile_prefs->GetList(kBraveWalletHiddenAccounts), [&](const base::Value& v) { auto* k = v.GetIfString(); return k && *k == unique_key; });`

mkarolin (approved):
`strings`++

components/brave_wallet_ui/components/desktop/popup-modals/add-account-modal/restore-accounts-modal.tsx
```diff
@@ -0,0 +1,156 @@
...
+import { reduceAddress } from '../../../../utils/reduce-address'
+import { getAccountTypeDescription } from '../../../../utils/account-utils'
+import {
+  useGetHiddenAccountsQuery,
+  useRemoveHiddenAccountsMutation,
+} from '../../../../common/slices/api.slice'
+import PopupModal from '..'
+import { Checkbox } from '../../../shared/checkbox/checkbox'
+import { Column, Row, Text } from '../../../shared/style'
+import { CreateAccountIcon } from '../../../shared/create-account-icon/create-account-icon'
```
Douglashdaniel, line 21:
```suggestion
// Utils
import { getLocale } from '../../../../../common/locale'
import { reduceAddress } from '../../../../utils/reduce-address'
import { getAccountTypeDescription } from '../../../../utils/account-utils'

// Types
import { WalletRoutes } from '../../../../constants/types'

// Queries
import {
  useGetHiddenAccountsQuery,
  useRemoveHiddenAccountsMutation,
} from '../../../../common/slices/api.slice'

// Components
import PopupModal from '..'
import { Checkbox } from '../../../shared/checkbox/checkbox'
import { CreateAccountIcon } from '../../../shared/create-account-icon/create-account-icon'

// Styles
import { Column, Row, Text } from '../../../shared/style'
```

components/brave_wallet_ui/components/desktop/popup-modals/add-account-modal/restore-accounts-modal.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+const RestoreAccountsList = styled(Column)`
+  max-height: 320px;
+  overflow: auto;
+  border: 1px solid ${leo.color.divider.subtle};
+  border-radius: 8px;
+  padding: 8px;
+`
```
Douglashdaniel: Styles should be declared in it's own file `restore_accounts_modal.style.ts`

components/brave_wallet_ui/components/desktop/popup-modals/add-account-modal/restore_accounts_modal.tsx
```diff
@@ -0,0 +1,156 @@
...
+    const accountsToRestore = hiddenAccounts.filter((account) =>
+      selectedAccountKeys.has(account.accountId.uniqueKey),
+    )
+    const accountIds = accountsToRestore.map((account) => account.accountId)
+
+    await removeHiddenAccounts(accountIds).unwrap()
+
+    onClose()
+  }, [hiddenAccounts, onClose, removeHiddenAccounts, selectedAccountKeys])
+
```
Douglashdaniel, line 67: Please group with comments

```suggestion
  const history = useHistory()
  
  // Queries
  const { data: hiddenAccounts = [] } = useGetHiddenAccountsQuery()
  const [removeHiddenAccounts] = useRemoveHiddenAccountsMutation()
  
  // State
  const [selectedAccountKeys, setSelectedAccountKeys] = React.useState<
    Set<string>
  >(new Set())

  // Methods
  const onClose = React.useCallback(() => {
    history.push(WalletRoutes.Accounts)
  }, [history])

  const onToggleAccountSelected = React.useCallback(
    (accountUniqueKey: string) => {
      setSelectedAccountKeys((currentKeys) => {
        const nextKeys = new Set(currentKeys)
        if (nextKeys.has(accountUniqueKey)) {
          nextKeys.delete(accountUniqueKey)
        } else {
          nextKeys.add(accountUniqueKey)
        }
        return nextKeys
      })
    },
    [],
  )

  const onRestoreSelectedAccounts = React.useCallback(async () => {
    const accountsToRestore = hiddenAccounts.filter((account) =>
      selectedAccountKeys.has(account.accountId.uniqueKey),
    )
    const accountIds = accountsToRestore.map((account) => account.accountId)

    await removeHiddenAccounts(accountIds).unwrap()

    onClose()
  }, [hiddenAccounts, onClose, removeHiddenAccounts, selectedAccountKeys])

```

components/brave_wallet_ui/components/desktop/popup-modals/add-account-modal/restore_accounts_modal.tsx
Douglashdaniel, line 1: Please use `snake_case` for new files `restore_accounts_modal.tsx`

components/resources/wallet_strings.grdp
```diff
@@ -346,6 +344,8 @@
   <message name="IDS_BRAVE_WALLET_TESTNET_ACCOUNTS" desc="Testnet accounts label">Testnet accounts</message>
   <message name="IDS_BRAVE_WALLET_ACCOUNTS_EXPORT" desc="Account list item Export button">Export</message>
   <message name="IDS_BRAVE_WALLET_ACCOUNTS_DEPOSIT" desc="Account list item Deposit button">Deposit</message>
+  <message name="IDS_BRAVE_WALLET_ACCOUNTS_HIDE" desc="Account list item Hide button">Hide account</message>
+  <message name="IDS_BRAVE_WALLET_ACCOUNTS_RESTORE" desc="Account list item Restore button">Restore</message>
```
Douglashdaniel, line 348: Please add these to `locale.ts` too for storybook

components/brave_wallet/browser/keyring_service.cc
```diff
@@ -3549,6 +3666,10 @@ const std::vector<mojom::AccountInfoPtr>& KeyringService::GetAllAccountInfos() {
         std::make_unique<std::vector<mojom::AccountInfoPtr>>();
     for (const auto& keyring_id : GetEnabledKeyrings()) {
       for (auto& account_info : GetAccountInfosForKeyring(keyring_id)) {
+        if (IsAccountHidden(profile_prefs_,
+                            account_info->account_id->unique_key)) {
+          continue;
+        }
```
supermassive: It is linear search in a loop. Should be a function-scope flat_set of unique_keys

components/brave_wallet/browser/keyring_service.cc
```diff
@@ -2721,6 +2731,113 @@ void KeyringService::SetAccountName(mojom::AccountIdPtr account_id,
...
+
+std::vector<mojom::AccountInfoPtr> KeyringService::GetHiddenAccountsSync() {
+  std::set<std::string> account_unique_keys;
+  for (const auto& hidden_account :
+       profile_prefs_->GetList(kBraveWalletHiddenAccounts)) {
+    if (auto* unique_key = hidden_account.GetIfString()) {
+      account_unique_keys.insert(*unique_key);
+    }
+  }
+
```
supermassive: Should be an anonymous function which returns a flat_set of account_ids(or unique_keys)

components/brave_wallet/browser/keyring_service.h
```diff
@@ -12,6 +12,7 @@
 #include <string>
 #include <vector>
 
+#include "base/functional/callback.h"
```
supermassive: ~#include "base/functional/callback.h"~

components/brave_wallet/browser/keyring_service.cc
```diff
@@ -2721,6 +2731,113 @@ void KeyringService::SetAccountName(mojom::AccountIdPtr account_id,
...
+      return;
+    }
+    unique_keys_to_remove.insert(account_id->unique_key);
+  }
+
+  if (unique_keys_to_remove.empty()) {
+    std::move(callback).Run(true);
+    return;
+  }
+
```
supermassive: We don't need these checks.
`!account_id` mojo guarantess non-null items.
`account_id->unique_key.empty()` this will just never equal to anything in pref
`unique_keys_to_remove.empty()` this will be just a noop below (removed_count==0)

components/brave_wallet/browser/keyring_service.cc
```diff
@@ -1024,6 +1025,17 @@ void MaybeRunPasswordMigrations(PrefService* profile_prefs,
   MaybeMigrateToWalletMnemonic(profile_prefs, password);
 }
 
+std::set<std::string> GetHiddenAccountUniqueKeys(PrefService* profile_prefs) {
```
supermassive, line 1028: `absl::flat_hash_set` or `base::flat_set`

supermassive (approved):
wallet core lgtm

Douglashdaniel (approved):

components/brave_wallet/browser/keyring_service.cc
```diff
@@ -2721,6 +2733,99 @@ void KeyringService::SetAccountName(mojom::AccountIdPtr account_id,
...
+        return *account_info.GetAccountId() == account_id;
+      });
+  if (account_it == derived_accounts.end()) {
+    return false;
+  }
+
+  if (account_it == derived_accounts.begin() &&
+      account_id.coin == mojom::CoinType::ETH &&
+      account_id.keyring_id == mojom::KeyringId::kDefault) {
+    return false;
```
kdenhartog: Do we need this check for the beginning account for each coin type or is the concern about defaults only for ETH?

kdenhartog (commented):
one question while doing some quick testing

components/brave_wallet/browser/keyring_service.cc
cypt4: We need to keep at least 1 account to support "selected account" feature on some platforms.

kdenhartog (approved):
LGTM for sec/privacy

PR #35382, opened 2026-04-08
Add opt-in debug logs for iOS user agent

StephenHeaps:
Adds a User Agent Debug Logs view to Developer Options. Logs are entirely local to device.
    - Developer Options is hidden by default, locked behind password and biometrics.
    - After Developer Options are enabled/exposed, logs are still disabled by default. User must opt-in via toggle in `User Agent Debug` view.
    - Logs are cleared on app launch.
    
User Agent Debug Logs

### Testplan

1. Visit some websites both on and not on [brave-checks.txt](https://github.com/brave/adblock-lists/blob/master/brave-lists/brave-checks.txt)
2. Open Settings > Developer Options > User Agent Debug
3. Verify toggle in navigation bar is disabled (default) and no logs are displayed
4. Enable the toggle
5. Visit some websites both on and not on [brave-checks.txt](https://github.com/brave/adblock-lists/blob/master/brave-lists/brave-checks.txt)
6. Verify logs are now displayed in Settings > Developer Options > User Agent Debug

### DIFF
--- a/ios/brave-ios/Package.swift
+++ b/ios/brave-ios/Package.swift
@@ -532,7 +532,11 @@ var package = Package(
       ],
       plugins: ["LoggerPlugin"]
     ),
-    .target(name: "UserAgent", dependencies: ["Preferences", "BraveCore"]),
+    .target(
+      name: "UserAgent",
+      dependencies: ["Preferences", "BraveCore"],
+      plugins: ["LoggerPlugin"]
+    ),
     .target(
       name: "CredentialProviderUI",
       dependencies: ["BraveCore", "DesignSystem", "BraveShared", "Strings", "BraveUI"]
--- a/ios/brave-ios/Sources/Brave/Frontend/Browser/BrowserViewController/BVC+TabPolicyDecider.swift
+++ b/ios/brave-ios/Sources/Brave/Frontend/Browser/BrowserViewController/BVC+TabPolicyDecider.swift
@@ -495,8 +495,12 @@ extension BrowserViewController {
       )
 
       if let url = modifiedRequest.url {
-        Logger.module.debug(
-          "Cancelled and recreating request to `\(url.absoluteString, privacy: .private)`"
+        Logger(
+          subsystem: "\(Bundle.main.bundleIdentifier ?? "com.brave.ios")",
+          category: "UserAgent"
+        )
+        .info(
+          "getInternalRedirect() - Detected incorrect UA in header. Cancelled and recreating request to `\(url.baseDomain ?? "nil", privacy: .public)` for user agent change from `\(headerUserAgent, privacy: .public)` to `\(userAgentForType, privacy: .public)`."
         )
       }
       return modifiedRequest
--- a/ios/brave-ios/Sources/Brave/Frontend/Browser/BrowserViewController/BVC+UserAgent.swift
+++ b/ios/brave-ios/Sources/Brave/Frontend/Browser/BrowserViewController/BVC+UserAgent.swift
@@ -4,7 +4,9 @@
 // You can obtain one at https://mozilla.org/MPL/2.0/.
 
 import BraveCore
+import BraveShared
 import Foundation
+import OSLog
 import Preferences
 import UserAgent
 import Web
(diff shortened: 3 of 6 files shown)


EliHini (approved):

nuo-xu (approved):

ios/brave-ios/Sources/BraveShared/DebugLogger.swift
```diff
@@ -9,23 +9,36 @@ import UIKit
...
+    case .userAgent:
+      return "userAgentLogs"
+    }
+  }
+
+  var loggingEnabledKey: String {
+    return "\(prefsKey)Enabled"
+  }
+
+  var isLoggingEnabled: Bool {
```
netzenbot: The `isLoggingEnabled` getter has a logic bug: `UserDefaults.standard.object(forKey: loggingEnabledKey) != nil || UserDefaults.standard.bool(forKey: loggingEnabledKey)` returns `true` whenever the key exists, even if the stored value is `false`. Once a user enables logging and then toggles it off, the key is set to `false` — but the getter returns `true` anyway because `object(forKey:) != nil`. The `object != nil` guard is unnecessary; `bool(forKey:)` already returns `false` by default when the key is absent. Replace with just `UserDefaults.standard.bool(forKey: loggingEnabledKey)`.

ios/brave-ios/Sources/BraveShared/DebugLogger.swift
```diff
@@ -81,14 +94,22 @@ public class DebugLogViewController: UIViewController {
...
       action: #selector(shareButtonTapped)
     )
-    navigationItem.rightBarButtonItem = rightBarButtonItem
+
+    let loggingSwitch = UISwitch()
+    loggingSwitch.isOn = loggerType.isLoggingEnabled
+    loggingSwitch.addTarget(self, action: #selector(loggingSwitchToggled(_:)), for: .valueChanged)
+    let switchBarButtonItem = UIBarButtonItem(customView: loggingSwitch)
+
+    navigationItem.rightBarButtonItems = [shareBarButtonItem, switchBarButtonItem]
```
netzenbot, line 92: There's a logic bug here: `UserDefaults.standard.object(forKey: loggingEnabledKey) != nil || UserDefaults.standard.bool(forKey: loggingEnabledKey)`. Once the key has been written (even with value `false`), the first condition is `true` and short-circuits the `||`, so `isLoggingEnabled` always returns `true` — toggling the switch off has no effect. It should just be `return UserDefaults.standard.bool(forKey: loggingEnabledKey)`.

ios/brave-ios/Sources/BraveShared/DebugLogger.swift
```diff
@@ -81,14 +94,22 @@ public class DebugLogViewController: UIViewController {
...
 
     let textView = UITextView()
     textView.translatesAutoresizingMaskIntoConstraints = false
```
netzenbot, line 94: There's a logic bug in `isLoggingEnabled`. The condition `object(forKey:) != nil || bool(forKey:)` returns `true` whenever the key exists in UserDefaults — even when the stored value is `false`. After a user enables logging and then disables it, `object(forKey:) != nil` is still `true`, so `isLoggingEnabled` returns `true` and the toggle can never actually disable logging. Change `||` to `&&`: `return UserDefaults.standard.object(forKey: loggingEnabledKey) != nil && UserDefaults.standard.bool(forKey: loggingEnabledKey)`.

ios/brave-ios/Sources/BraveShared/DebugLogger.swift
```diff
@@ -9,23 +9,36 @@ import UIKit
...
+      return "userAgentLogs"
+    }
+  }
+
+  var loggingEnabledKey: String {
+    return "\(prefsKey)Enabled"
+  }
+
+  var isLoggingEnabled: Bool {
+    get {
```
netzenbot: The `isLoggingEnabled` getter has a bug: `UserDefaults.standard.object(forKey: loggingEnabledKey) != nil` returns `true` whenever the key exists — including when it was stored as `false`. So after a user enables logging then disables it, `isLoggingEnabled` still returns `true` and logging stays on. Replace the whole expression with just `UserDefaults.standard.bool(forKey: loggingEnabledKey)`, which correctly returns `false` by default and reflects the actual stored value.

PR #35385, opened 2026-04-08
Migrate DAU last ping date to local state pref on iOS

### LINKED ISSUE #54317: Store DAU last ping date on iOS in local state instead of Swift preferences
The DAU last ping date is currently stored as a Swift [Int] array (lastLaunchInfo). This makes it inaccessible to C++ and Objective-C platform components and requires fragile manual serialization.
To fix that we need to move the last ping date to local state, so all platform components share one source of truth.

aseren:
The Swift-only lastLaunchInfo preference could not be shared with C++ or Objective-C components, limiting which parts of the codebase could read or update the DAU last ping date.

The PR stores the date as brave.stats.last_ping_date in local state using base::Time so all platform layers share a single typed value.

### Test case 1

* Fresh install
* Verify that Daily usage ping was sent on start with first=true.
* Verify that only one Daily usage ping was sent on start.
* Verify `brave.stats.last_check_ymd` preference in brave://local-state that is set to current date in UTC
* Wait for 10 minutes (need to more that 5 minutes wait because Daily usage ping timer is fired every 5 minutes)
* Verify that Daily usage ping was not sent
~~* Forward clock to 12 hour~~
~~* Verify that Daily usage ping was not sent~~
~~* Forward clock to 11 hours~~
~~* Verify that Daily usage ping was not sent~~
* Forward clock to 1 day
* Verify that Daily usage ping was sent
* Verify `brave.stats.last_check_ymd` preference in brave://local-state that is set to new date
* Forward clock to 1 week
* Verify that weekly usage ping is sent
* Verify `brave.stats.last_check_ymd` preference in brave://local-state that is set to new date
* Forward clock to 1 month
* monthly usage ping is sent
* Verify `brave.stats.last_check_ymd` preference in brave://local-state that is set to new date

### Test case 2

* Fresh install of an old version
* Verify that only one Daily usage ping was sent on start.
* Close the browser
* Install latest version
* Verify  that Daily usage ping is NOT sent on start.
* Verify `brave.stats.last_check_ymd` preference in brave://local-state that is set to current date in UTC
* Forward clock to 1 day
* Verify that Daily usage ping was sent
* Verify `brave.stats.last_check_ymd` preference in brave://local-state that is set to new date

### Test case 3

* Fresh install of an old version
* Verify that only one Daily usage ping was sent on start.
* Close the browser
* Forward clock to 1 day
* Install latest version
* Verify that Daily usage ping was sent
* Verify `brave.stats.last_check_ymd` preference in brave://local-state that is set to current date in UTC

Resolves https://github.com/brave/brave-browser/issues/54317

### DIFF
--- a/ios/brave-ios/Sources/Brave/Migration/Migration.swift
+++ b/ios/brave-ios/Sources/Brave/Migration/Migration.swift
@@ -109,12 +109,26 @@ public class BraveLocalStateMigration {
 
   public func launchMigrations() {
     migrateDAUPingPreference()
+    migrateDAULastLaunchInfoPreference()
     migrateAdsPreferences()
   }
 
   private func migrateDAUPingPreference() {
     Preferences.DeprecatedPreferences.sendUsagePing.migrate { value in
       localState.set(value, forPath: kStatsReportingEnabledPrefName)
+    }
+  }
+
+  private func migrateDAULastLaunchInfoPreference() {
+    Preferences.DeprecatedPreferences.lastLaunchInfo.migrate { value in
+      guard let lastPingTimestamp = value?.first else { return }
+
+      if !Preferences.DAU.firstPingParam.isValueStored {
+        Preferences.DAU.firstPingParam.value = false
+      }
+
+      let date = Date(timeIntervalSince1970: TimeInterval(lastPingTimestamp))
+      localState.set(DAU.dateFormatter.string(from: date), forPath: kLastCheckYMDPrefName)
     }
   }
 
@@ -267,6 +281,12 @@ extension Migration {
 extension Preferences {
   fileprivate final class DeprecatedPreferences {
     static let sendUsagePing = Option<Bool>(key: "dau.send-usage-ping", default: true)
+
+    /// Superseded by `brave.stats.last_ping_date` in local state.
+    static let lastLaunchInfo = Option<[Int]?>(
+      key: "dau.last-launch-info",
+      default: nil
+    )
 
     static let blockAdsAndTracking = Option<Bool>(
       key: "shields.block-ads-and-tracking",
@@ -446,7 +466,6 @@ extension Preferences {
     migrate(key: "braveAdblockUseRegional", to: Preferences.Shields.useRegionAdBlock)
 
     // DAU
-    migrate(key: "dau_stat", to: Preferences.DAU.lastLaunchInfo)
     migrate(key: "week_of_installation", to: Preferences.DAU.weekOfInstallation)
 
     // URP
@@ -462,9 +481,6 @@ extension Preferences {
     migrate(key: "fp_protection", to: Preferences.BlockStats.fingerprintingCount)
     migrate(key: "safebrowsing", to: Preferences.BlockStats.phishingCount)
 
-    // On 1.6 lastLaunchInfo is used to check if it's first app launch or not.
-    // This needs to be translated to our new preference.
-    Preferences.General.isFirstLaunch.value = Preferences.DAU.lastLaunchInfo.value == nil
     Preferences.Migration.completed.value = true
   }
 
--- a/ios/brave-ios/Sources/Growth/DAU.swift
+++ b/ios/brave-ios/Sources/Growth/DAU.swift
@@ -48,7 +48,7 @@ public class DAU {
   }
 
   /// Date formatted used for passing date strings to the DAU server.
-  static let dateFormatter = { () -> DateFormatter in
+  public static let dateFormatter = { () -> DateFormatter in
     let formatter = DateFormatter()
     formatter.dateFormat = "yyyy-MM-dd"
     formatter.calendar = DAU.calendar
@@ -111,7 +111,12 @@ public class DAU {
       Logger.module.debug("DAU ping disabled by the user.")
       return
     }
-    guard let paramsAndPrefs = paramsAndPrefsSetup(for: Date()) else {
+    guard
+      let paramsAndPrefs = paramsAndPrefsSetup(
+        for: Date(),
+        lastPingDate: stats.lastPingDate
+      )
+    else {
       Logger.module.debug("dau, no changes detected, no server ping")
       return
     }
@@ -154,8 +159,12 @@ public class DAU {
       // This preference is set for future DAU pings.
       Preferences.DAU.firstPingParam.value = false
 
-      // This preference is used to calculate whether user used the app in this month and/or day.
-      Preferences.DAU.lastLaunchInfo.value = paramsAndPrefs.lastLaunchInfoPreference
+      // Store the last ping date in local state so all platform components share one source of truth.
+      // Thread hop to main is required because `lastPingDate` sets `brave.stats.last_check_ymd`
+      // Chromium preference, which must be done from the main thread.
+      DispatchQueue.main.async { [weak self] in
+        self?.braveCoreStats?.lastPingDate = paramsAndPrefs.date
+      }
     }
 
     task.resume()
@@ -165,7 +174,7 @@ public class DAU {
   struct ParamsAndPrefs {
     let queryParams: [URLQueryItem]
     let headers: [String: String]
-    let lastLaunchInfoPreference: [Int]
+    let date: Date
   }
 
   func migrateInvalidWeekOfInstallPref() {
@@ -183,8 +192,8 @@ public class DAU {
   }
 
   /// Return params query or nil if no ping should be send to server and also preference values to set
-  /// after a succesful ing.
-  func paramsAndPrefsSetup(for date: Date) -> ParamsAndPrefs? {
+  /// after a succesful ping.
+  func paramsAndPrefsSetup(for date: Date, lastPingDate: Date?) -> ParamsAndPrefs? {
     var params = [channelParam(), versionParam()]
 
     let firstLaunch = Preferences.DAU.firstPingParam.value
@@ -201,7 +210,13 @@ public class DAU {
       migrateInvalidWeekOfInstallPref()
     }
 
-    guard let dauStatParams = dauStatParams(for: date, firstPing: firstLaunch) else {
+    guard
+      let dauStatParams = dauStatParams(
+        for: date,
+        lastPingDate: lastPingDate,
+        firstPing: firstLaunch
+      )
+    else {
       return nil
     }
 
@@ -231,8 +246,6 @@ public class DAU {
       params.append(URLQueryItem(name: "ref", value: referralCode))
     }
 
-    let lastPingTimestamp = [Int((date).timeIntervalSince1970)]
-
     var headers: [String: String] = [:]
 
     if let key = self.apiKey, !key.isEmpty {
@@ -242,7 +255,7 @@ public class DAU {
     return ParamsAndPrefs(
       queryParams: params,
       headers: headers,
-      lastLaunchInfoPreference: lastPingTimestamp
+      date: date
     )
   }
 
@@ -374,7 +387,7 @@ public class DAU {
   /// Returns nil if no dau changes detected.
   func dauStatParams(
     for date: Date,
-    dauStat: [Int?]? = Preferences.DAU.lastLaunchInfo.value,
+    lastPingDate: Date?,
     firstPing: Bool,
     channel: AppBuildChannel = AppConstants.buildChannel
   ) -> [URLQueryItem]? {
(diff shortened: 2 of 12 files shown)


DJAndries:
Why not reuse `kLastCheckYMD`?

tmancey:
@aseren I assume `brave.stats.last_ping_date` matches the same pref path as desktop, if not and it is possible we should keep those pref paths aligned.

EliHini (approved):

aseren:
> Why not reuse `kLastCheckYMD`?
> @aseren I assume `brave.stats.last_ping_date` matches the same pref path as desktop, if not and it is possible we should keep those pref paths aligned.

@DJAndries @tmancey 

Desktop currently lacks a `brave.stats.last_ping_date` `base::Time` preference and instead uses `kLastCheckYMD`, which stores local time as a `YYYY-MM-DD` string. Introducing the new preference addresses following problems:

* The string format is error prone. `kLastCheckYMD` stores local time without a timezone, so users who switch timezones may encounter incorrect behavior. 
* Storing the value as base::Time eliminates the need for manual serialization, reduces the code complexity and removes possible conversion bugs.
* iOS DAU code works with ping dates as timestamps. Converting it to `YYYY-MM-DD` format will require additional code changes which can cause bugs.
* Once Desktop daily pings migrate to UTC, both platforms can use `brave.stats.last_ping_date` and `kLastCheckYMD` can be deprecated.

tmancey (approved):

DJAndries:
> * The string format is error prone. `kLastCheckYMD` stores local time without a timezone, so users who switch timezones may encounter incorrect behavior.

I would think that `base::Time` could be more error prone. As you know, `TimePeriodStorage` is also `base::Time`-based and has had its fair share of DST related issues, both past and present. Both TPS and the stats updater share one thing in common: they are only concerned with the date, and not the time. If we do wish to change types, I think it would make sense to use a date-only type (i.e. one of the icu ones).

Also, users changing timezones seems like an edge case, lmk if you disagree.

> Storing the value as base::Time eliminates the need for manual serialization, reduces the code complexity and removes possible conversion bugs.
> iOS DAU code works with ping dates as timestamps. Converting it to YYYY-MM-DD format will require additional code changes which can cause bugs.

Is generating/parsing a date string not straightforward on iOS?

> Once Desktop daily pings migrate to UTC, both platforms can use brave.stats.last_ping_date and kLastCheckYMD can be deprecated.

I didn't know that UTC pings were back on the menu. Perhaps I missed the memo

aseren:
> I would think that `base::Time` could be more error prone. As you know, `TimePeriodStorage` is also `base::Time`-based and has had its fair share of DST related issues, both past and present. Both TPS and the stats updater share one thing in common: they are only concerned with the date, and not the time. If we do wish to change types, I think it would make sense to use a date-only type (i.e. one of the icu ones).
> 
> Also, users changing timezones seems like an edge case, lmk if you disagree.

I agree this is an edge case but it shows problems with this approach.
I think storing time is fine even if we work only with dates, as a date can be easily calculated from time. 
Also if we move to UTC we won't have any issues with DST you mentioned.

> Is generating/parsing a date string not straightforward on iOS?

It still requires additional code for conversion. We already have such code for Desktop, so we would duplicate it for iOS.
Another thing is that the iOS code which sends pings works with timestamps, but if we use YYYY-MM-DD format then we need to verify that the logic remains the same.

> I didn't know that UTC pings were back on the menu. Perhaps I missed the memo

UTC was used for iOS pings so we are keep using UTC on iOS after migration.
And if we move to UTC on Desktop it can help fixing DST issues with TPS you mentioned.

ios/brave-ios/Sources/Brave/Migration/Migration.swift
```diff
@@ -446,7 +470,7 @@ extension Preferences {
     migrate(key: "braveAdblockUseRegional", to: Preferences.Shields.useRegionAdBlock)
 
     // DAU
-    migrate(key: "dau_stat", to: Preferences.DAU.lastLaunchInfo)
```
kylehickinson, line 449: Just delete entirely, these are super old

ios/brave-ios/Sources/Brave/Migration/Migration.swift
```diff
@@ -464,7 +488,7 @@ extension Preferences {
 
     // On 1.6 lastLaunchInfo is used to check if it's first app launch or not.
     // This needs to be translated to our new preference.
-    Preferences.General.isFirstLaunch.value = Preferences.DAU.lastLaunchInfo.value == nil
+    Preferences.General.isFirstLaunch.value = DeprecatedPreferences.lastLaunchInfo.value == nil
```
kylehickinson: Seems like we may want to address the comment first

ios/brave-ios/Sources/Growth/DAU.swift
```diff
@@ -154,8 +159,10 @@ public class DAU {
       // This preference is set for future DAU pings.
       Preferences.DAU.firstPingParam.value = false
 
-      // This preference is used to calculate whether user used the app in this month and/or day.
-      Preferences.DAU.lastLaunchInfo.value = paramsAndPrefs.lastLaunchInfoPreference
+      // Store the last ping date in local state so all platform components share one source of truth.
+      DispatchQueue.main.async { [weak self] in
```
kylehickinson, line 165: Why thread hop?

ios/brave-ios/Sources/Brave/Migration/Migration.swift
```diff
@@ -265,9 +283,15 @@ extension Migration {
 }
 
 extension Preferences {
-  fileprivate final class DeprecatedPreferences {
+  internal final class DeprecatedPreferences {
```
kylehickinson: Don't really want to expose this

DJAndries:
> Also if we move to UTC we won't have any issues with DST you mentioned.

Yes, but until then, there's potential for bugs.

> It still requires additional code for conversion. We already have such code for Desktop, so we would duplicate it for iOS.

Not sure I understand, don't we just need to call `brave_stats::GetYMDAsDate`  / `brave_stats::GetDateAsYMD` ?

ios/browser/api/brave_stats/brave_stats.mm
```diff
@@ -52,4 +54,18 @@ - (BOOL)isNotificationAdsEnabled {
   return _profilePrefs->GetBoolean(brave_ads::prefs::kOptedInToNotificationAds);
 }
 
+- (nullable NSDate*)lastPingDate {
+  const base::Time last_ping_date =
+      _localPrefs->GetTime(brave_stats::prefs::kLastPingDate);
```
DJAndries: couldn't this simply be `brave_stats::GetYMDAsDate(_localPrefs->GetString(brave_stats::prefs::kLastCheckYMD))` ?

ios/browser/api/brave_stats/brave_stats.mm
```diff
@@ -52,4 +54,18 @@ - (BOOL)isNotificationAdsEnabled {
...
+  if (last_ping_date.is_null()) {
+    return nil;
+  }
+  return last_ping_date.ToNSDate();
+}
+
+- (void)setLastPingDate:(NSDate*)date {
+  _localPrefs->SetTime(brave_stats::prefs::kLastPingDate,
```
DJAndries: couldn't this simply be `_localPrefs->SetString(brave_stats::prefs::kLastCheckYMD, brave_stats::GetDateAsYMD(...))` ?

netzenbot (approved):

ios/brave-ios/Sources/Brave/Migration/Migration.swift
aseren: My understanding is that the comment describes why `Preferences.General.isFirstLaunch` preference was introduced so we don't need to address it. 
The change was done 8 year ago via https://github.com/brave/brave-ios/pull/494

ios/brave-ios/Sources/Growth/DAU.swift
aseren: We set a chromium preference which is needs to be done from main thread

ios/brave-ios/Sources/Brave/Migration/Migration.swift
aseren: okay, this was done to access `Preferences.DeprecatedPreferences.lastLaunchInfo.key` from tests. 
I will convert it to `fileprivate` back and do a private variable in test:
`private let DeprecatedPreferencesLastLaunchInfoKey = "dau.last-launch-info"`

ios/brave-ios/Sources/Brave/Migration/Migration.swift
kylehickinson: Either way it doesn't seem like we should be accessing this pref anymore post migration, that's why `DeprecatedPreferences` is private

kylehickinson: Maybe safe to just fully remove thus migration as well?

ios/browser/api/brave_stats/brave_stats.mm
aseren: GetYMDAsDate is suitable for local time only. Will need to add new utility to support UTC

ios/browser/api/brave_stats/brave_stats.mm
aseren: GetDateAsYMD is suitable for local time only. Will need to add new utility to support UTC

ios/browser/api/brave_stats/brave_stats.mm
DJAndries: we could add an `is_utc` param to the helper func, and swap `FromLocalExploded` with `FromUTCExploded` if true

ios/browser/api/brave_stats/brave_stats.mm
DJAndries: we could add an `is_utc` param to the helper func, and swap `LocalExplode` with `UTCExplode` if true

aseren:
> Not sure I understand, don't we just need to call `brave_stats::GetYMDAsDate` / `brave_stats::GetDateAsYMD` ?

GetDateAsYMD / GetYMDAsDate works with Local Time. For iOS we need to add new utilities which support UTC. 
They will maybe need to be exposed to iOS to be used in migration. I think it would be better to create a helper in swift code.

ios/brave-ios/Sources/Brave/Migration/Migration.swift
aseren: Removed

ios/brave-ios/Sources/Brave/Migration/Migration.swift
aseren: Removed

ios/brave-ios/Sources/Brave/Migration/Migration.swift
aseren: returned to fileprivate back

ios/browser/api/brave_stats/brave_stats.mm
aseren: Done

ios/browser/api/brave_stats/brave_stats.mm
aseren: Done

DJAndries (approved):
cpp/objc changes lgtm, thanks!

ios/brave-ios/Tests/ClientTests/BraveLocalStateMigrationTests.swift
```diff
@@ -0,0 +1,205 @@
...
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import Foundation
+import Preferences
+import Web
+import XCTest
+
+@testable import Brave
+
+private let kLastCheckYMDPrefName = "brave.stats.last_check_ymd"
```
kylehickinson: You can just `import BraveCore` and access this like normal in the tests

ios/brave-ios/Sources/Growth/DAU.swift
kylehickinson: Maybe add a comment with that reasoning so if this is refactored in the future the same requirement is added

ios/brave-ios/Tests/ClientTests/BraveLocalStateMigrationTests.swift
aseren: Fixed

ios/brave-ios/Sources/Growth/DAU.swift
aseren: Added

kylehickinson (approved):

PR #35415, opened 2026-04-09
Upgrade from Chromium 148 to Chromium 149

### LINKED ISSUE #54357: Upgrade from Chromium 148 to Chromium 149
**Description**
Upgrade from Chromium 146 to Chromium 147

https://chromium.googlesource.com/chromium/src/+log/148.0.7778.96..149.0.7819.1?pretty=fuller&n=10000

**QA tests:**

**Desktop Affected areas:**
* TBD

**iOS Affected areas:**
* TBD

**Android Affected areas:**
* TBD

cdesouza-chromium:
Resolves https://github.com/brave/brave-browser/issues/54357


cdesouza-chromium:
> [!WARNING]
> You have got a presubmit warning. Please address it if possible.

```
A banned pattern was used.
    browser/brave_browser_main_parts.cc:146:
      Most code should already be scoped to a Browser. Pass in a Browser* or retrieve from an existing entity with a reference to the Browser.
```

cdesouza-chromium:
> [!WARNING]
> You have got a presubmit warning. Please address it if possible.

```
A banned pattern was used.
    browser/brave_drm_tab_helper.cc:42:
      Most code should already be scoped to a Browser. Pass in a Browser* or retrieve from an existing entity with a reference to the Browser.
```

cdesouza-chromium:
> [!WARNING]
> You have got a presubmit warning. Please address it if possible.

```
A banned pattern was used.
    browser/sync/brave_sync_alerts_service.cc:57:
      Most code should already be scoped to a Browser. Pass in a Browser* or retrieve from an existing entity with a reference to the Browser.
```

cdesouza-chromium:
> [!WARNING]
> You have got a presubmit warning. Please address it if possible.

```
A banned pattern was used.
    components/os_crypt/sync/key_storage_libsecret_unittest.cc:146:
      Do not use UNSAFE_TODO() to write new unsafe code. Use only when removing a pre-existing file-wide allow_unsafe_buffers pragma, or when incrementally converting code off of unsafe interfaces
```

cdesouza-chromium:
> [!WARNING]
> You have got a presubmit warning. Please address it if possible.

```
A banned pattern was used.
    components/os_crypt/sync/key_storage_linux.cc:29:
      Code gated by GOOGLE_CHROME_BRANDING is effectively untested. This is typically wrong. Valid use cases are glue for private modules shipped alongside Chrome, and installation-related logic.
```

cdesouza-chromium:
> [!WARNING]
> You have got a presubmit warning. Please address it if possible.

```
A banned pattern was used.
    components/os_crypt/sync/key_storage_linux_unittest.cc:58:
      Do not RunUntilIdle. If possible, explicitly quit the run loop using run_loop.Quit() or run_loop.QuitClosure() if completion can be observed using a lambda or callback. Otherwise, wait for the condition to be true via base::test::RunUntil().
```

supermassive (approved):
wallet lgtm

fmarier (approved):

components/brave_rewards/content/rewards_service.h
```diff
@@ -262,7 +262,12 @@ class RewardsService : public KeyedService {
   virtual void OnRewardsPageShown() = 0;
 
  protected:
-  base::ObserverList<RewardsServiceObserver> observers_;
+  // TODO(zenparsing): Investigate if reentrancy can be removed.
```
zenparsing: I believe that the only reentrant notification was removed in https://github.com/brave/brave-core/pull/36401.

zenparsing (approved):
Rewards 👍

darkdh (approved):
sync, tor and wallet ++

browser/p3a/p3a_core_metrics.h
```diff
@@ -18,42 +18,50 @@
...
+class BraveWindowTracker : public BrowserCollectionObserver {
  public:
   explicit BraveWindowTracker(PrefService* local_state);
   BraveWindowTracker(const BraveWindowTracker&) = delete;
   BraveWindowTracker& operator=(const BraveWindowTracker&) = delete;
   ~BraveWindowTracker() override;
 
   static void CreateInstance(PrefService* local_state);
+  static bool HasInstance();
+  static void ClearInstance();
```
DJAndries, line 50: what are these methods for?

DJAndries (approved):
p3a++

kylehickinson (approved):
ios++

browser/p3a/p3a_core_metrics.h
mkarolin: Please see the commit message for this change: https://github.com/brave/brave-core/pull/35415/commits/9631f305a4072a4af968299413a6015ac9cefb70 (currently) - `[cr149] Fixes BraveWindowTracker dangling pointer.`

components/brave_rewards/content/rewards_service.h
mkarolin: Thanks @zenparsing, I reverted this change.

components/ai_chat/core/browser/tools/tool_provider.h
```diff
@@ -101,7 +101,12 @@ class ToolProvider {
  protected:
   void NotifyTaskStateChanged();
 
-  base::ObserverList<Observer> observers_;
+  // TODO(petemill): Investigate if reentrancy can be removed.
```
bridiver: this should have an issue link associated with it

components/brave_ads/core/public/ads_client/ads_client_notifier.h
```diff
@@ -67,7 +67,12 @@ class AdsClientNotifier final {
   void NotifyDidSolveAdaptiveCaptcha();
 
  private:
-  base::ObserverList<AdsClientNotifierObserver> observers_;
+  // TODO(tmancey): Investigate if reentrancy can be removed.
```
bridiver: this should have an issue link associated with it

browser/brave_wallet/brave_wallet_service_delegate_impl.cc
```diff
@@ -38,8 +38,9 @@ content::WebContents* GetActiveWebContents() {
     return g_web_contents_for_testing;
   }
 
-  Browser* browser = chrome::FindLastActive();
-  return browser ? browser->tab_strip_model()->GetActiveWebContents() : nullptr;
+  BrowserWindowInterface* browser = chrome::FindLastActive();
```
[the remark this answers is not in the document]
bridiver: do we have an open issue for FindLastActive usage here?

patches/services-network-public-cpp-header_util.cc.patch
```diff
@@ -0,0 +1,22 @@
+diff --git a/services/network/public/cpp/header_util.cc b/services/network/public/cpp/header_util.cc
+index ca0abc83883d9cb03d0fbcb53adf27f9232273a7..6211127ac172c9365b009ab0a8e9f3030036970c 100644
+--- a/services/network/public/cpp/header_util.cc
++++ b/services/network/public/cpp/header_util.cc
+@@ -43,7 +43,8 @@ const char* kUnsafeHeaders[] = {
+     net::HttpRequestHeaders::kHost,
+ 
+     // Trailers are not supported.
+-    "Trailer", "Te",
```
bridiver: Is this unintentional?

rewrite/chrome/browser/ui/page_actions/page_action_metrics_recorder.cc.toml
```diff
@@ -4,21 +4,7 @@
...
-[[substitution]]
-description = 'Replaces RecordIconClick with RecordIconClick_Chromium'
-re_pattern = '(void PageActionPerActionMetricsRecorder::RecordIconClick)\(\)'
-replace = '\1_Chromium()'
-
-[[substitution]]
-description = 'Replaces RecordChipClick with RecordChipClick_Chromium'
-re_pattern = '(void PageActionPerActionMetricsRecorder::RecordChipClick)\(\)'
-replace = '\1_Chromium()'
+description = 'Early return in all Record* functions to suppress Chromium metrics'
```
bridiver: can't we just override this entire class to be a no-op?

chromium_src/chrome/browser/ui/views/permissions/permission_prompt_factory.cc
```diff
@@ -10,6 +10,7 @@
 
 #if BUILDFLAG(ENABLE_BRAVE_WALLET)
 #include "brave/browser/ui/views/permission_bubble/brave_wallet_permission_prompt_impl.h"
+#include "chrome/browser/ui/browser_finder.h"
```
bridiver: this header does not belong behind `ENABLE_BRAVE_WALLET`. guards in cpp should match guards in gn

package-lock.json
```diff
@@ -234,6 +234,7 @@
       "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.29.0.tgz",
```
bridiver, line 234: why is there a package-lock.json change?

patches/tools-typescript-tsconfig_base.json.patch
```diff
@@ -1,12 +1,13 @@
...
+index 638c3a19db70ebe30e89bdc104330edd1104c0ab..1d3eef6c9d0fa1fb65895485e32dd37b12b6f212 100644
 --- a/tools/typescript/tsconfig_base.json
 +++ b/tools/typescript/tsconfig_base.json
-@@ -16,7 +16,6 @@
+@@ -16,7 +16,7 @@
      "noImplicitReturns": true,
      "noPropertyAccessFromIndexSignature": true,
      "noUncheckedIndexedAccess": true,
 -    "noUncheckedSideEffectImports": true,
++    "noUncheckedSideEffectImports": false,
```
bridiver, line 10: `Fix builds` doesn't provide much useful information here? What exactly is the problem?

patches/crypto-subtle_passkey.h.patch
```diff
@@ -0,0 +1,21 @@
+diff --git a/crypto/subtle_passkey.h b/crypto/subtle_passkey.h
+index 50d882eb628271b3d9d29b67e9c94a156bc235cc..7f57b496f09c16ffd9f2b16811d4aadd0408bd87 100644
+--- a/crypto/subtle_passkey.h
++++ b/crypto/subtle_passkey.h
+@@ -42,6 +42,8 @@ namespace webauthn {
+ crypto::SubtlePassKey MakeCryptoPassKey();
+ }  // namespace webauthn
+ 
++class OSCryptImpl;
```
bridiver: `This is a temporary measure until this class is gone.`

when will it be gone?

patches/chrome-android-java-src-org-chromium-chrome-browser-settings-search-SearchIndexProviderRegistry.java.patch
```diff
@@ -1,93 +1,126 @@
...
+-import org.chromium.chrome.browser.settings.MainSettings;
++import org.chromium.chrome.browser.settings.BraveMainPreferencesBase;
+ import org.chromium.chrome.browser.ssl.HttpsFirstModeSettingsFragment;
+ import org.chromium.chrome.browser.sync.settings.AccountManagementFragment;
+ import org.chromium.chrome.browser.sync.settings.GoogleServicesSettings;
+@@ -70,6 +67,31 @@ import org.chromium.components.browser_ui.site_settings.SingleWebsiteSettings;
+ import org.chromium.components.browser_ui.site_settings.SiteSettings;
+ import org.chromium.components.page_info.PageInfoAdPersonalizationSettings;
+ import org.chromium.components.page_info.PageInfoCookiesSettings;
++import org.chromium.brave.browser.custom_search_engines.settings.AddCustomSearchEnginePreferenceFragment;
```
bridiver: can you just use full qualified class names instead of adding all of these imports?

chromium_src/ios/web/webui/crw_web_ui_scheme_handler.mm
```diff
@@ -27,8 +27,20 @@ - (NSHTTPURLResponse*)processResponse:(NSHTTPURLResponse*)response
...
+// Allow a WebUI page to load a different WebUI URL as a child frame. The
+// upstream same-origin guard at the top of `webView:startURLSchemeTask:`
+// rejects any sub-resource whose `SchemeHostPort` differs from the main
+// `webView.URL` unless it is `chrome://resources`. Brave's per-frame `WebUIIOS`
+// support deliberately allows e.g. AI Chat to embed a chrome-untrusted (or
+// sibling-host) WebUI iframe inside a trusted WebUI page, so extend the
+// allow-list to admit any URL that is recognized by a registered
+// `WebUIIOSControllerFactory`. Non-WebUI URLs continue to fall through to the
+// upstream check.
+#define DomainIs(host) DomainIs(host) && GetErrorCodeForUrl(URL) != 0
```
bridiver: it's not clear to me why `GetErrorCodeForUrl != 0` makes this work and I worry that maybe it might allow things we don't want to let through. This seems like it should have some tests associated with it.

bridiver: cc @brave/sec-team

android/java/org/chromium/chrome/browser/BraveIntentHandler.java
```diff
@@ -151,16 +151,17 @@ private static boolean isJavascriptSchemeOrInvalidUrl(String unused_url) {
...
-     * gets rewritten to chrome:// deeper in the navigation stack — too late to protect this guard.
+     * Bytecode-redirected from {@link IntentHandler#intentHasUnsafeUrl}. Defers to the upstream
+     * check (which handles chrome://, chrome-native://, devtools://, distiller://, about://) and
+     * additionally blocks brave://, since it is a display alias for chrome:// and gets rewritten to
+     * chrome:// deeper in the navigation stack — too late to protect this guard.
      */
-    public static boolean intentHasUnsafeInternalScheme(
-            @Nullable String scheme, @Nullable String url, Intent intent) {
-        if (BraveIntentHandlerInternal.intentHasUnsafeInternalScheme(scheme, url, intent)) {
+    public static boolean intentHasUnsafeUrl(@Nullable String url, Intent intent) {
```
bridiver, line 159: as a follow-up is there a way to convert brave -> chrome before it hits this code since we're trying to avoid using `brave://` internally for anything?

components/ai_chat/core/browser/tools/tool_provider.h
mkarolin: Added the issue link in 20671ecdf4c

components/brave_ads/core/public/ads_client/ads_client_notifier.h
mkarolin: Added the issue link in 2d524b79814

patches/chrome-browser-ssl-ask_before_http_dialog_controller.cc.patch
```diff
@@ -0,0 +1,22 @@
...
+-    "https://support.google.com/chrome?p=first_mode";
+ 
+ // Helper to create the prompt body view based on the Ask-before-HTTP warning
+ // type.
+@@ -440,7 +438,7 @@ void AskBeforeHttpDialogController::OnHelpCenterLinkClicked(
+       security_interstitials::MetricsHelper::SHOW_LEARN_MORE);
+ 
+   content::OpenURLParams params(
+-      GURL(kLearnMoreLink), content::Referrer(),
++      GURL(kBraveLearnMoreLink), content::Referrer(),
```
bridiver: why don't we just change the text of `kLearnMoreLink` so we don't need a patch + chromium_src?

patches/services-network-public-cpp-header_util.cc.patch
mkarolin: Yes, unintentional - the victim of format on save :( 
Fixed in 16e7ce611a4

chromium_src/chrome/browser/ui/views/permissions/permission_prompt_factory.cc
mkarolin: Fixed in a12643ea38e

package-lock.json
mkarolin: Hmm, not sure - must have changed due to sync and got accidentally added to the commit. Fixed in 50947326b92

patches/tools-typescript-tsconfig_base.json.patch
mkarolin: cc: @fallaciousreasoning

patches/crypto-subtle_passkey.h.patch
mkarolin: It's already gone - reverted this change in b0a564eb3d4

patches/chrome-android-java-src-org-chromium-chrome-browser-settings-search-SearchIndexProviderRegistry.java.patch
mkarolin: cc: @samartnik

chromium_src/ios/web/webui/crw_web_ui_scheme_handler.mm
mkarolin: cc: @emerick

android/java/org/chromium/chrome/browser/BraveIntentHandler.java
mkarolin: cc: @samartnik

patches/chrome-browser-ssl-ask_before_http_dialog_controller.cc.patch
mkarolin: cc: @cdesouza-chromium

rewrite/components/password_manager/core/browser/password_store/password_store_built_in_backend.cc.toml
```diff
@@ -0,0 +1,17 @@
...
+
+# `UserActionableError::{kNeedsSettingsConfirmation,kUnrecoverableError}` were
+# removed on Android in upstream cr149 (https://chromium-review.googlesource.com/c/chromium/src/+/7772435),
+# but the case labels in this file are still gated only by `!IS_IOS`. Upstream Android doesn't compile
+# this file, but Brave does (`use_login_database_as_backend = true`), so the
+# guard needs `!IS_ANDROID` too.
+
+[[substitution]]
+description = 'Exclude removed UserActionableError values on Android'
+re_pattern = '''#if !BUILDFLAG\(IS_IOS\)\n(    case SyncError::kNeedsSettingsConfirmation:\n    case SyncError::kUnrecoverableError:\n)#endif  // !BUILDFLAG\(IS_IOS\)'''
```
bridiver: this pattern seems fragile, we shouldn't be relying on fixed whitespace here. It also seems unnecessarily specific matching on the entire block verbatim

patches/components-password_manager-core-browser-password_store-password_store_built_in_backend.cc.patch
```diff
@@ -0,0 +1,43 @@
+diff --git a/components/password_manager/core/browser/password_store/password_store_built_in_backend.cc b/components/password_manager/core/browser/password_store/password_store_built_in_backend.cc
+index a142f11d9142096ed7eac2145aa98000131e73b3..3bd42769775b355fce9d4840a825bf43af3d29bc 100644
+--- a/components/password_manager/core/browser/password_store/password_store_built_in_backend.cc
++++ b/components/password_manager/core/browser/password_store/password_store_built_in_backend.cc
+@@ -91,10 +91,10 @@ bool ShouldForwardSyncErrorToStore(
+     case SyncError::kSignInNeedsUpdate:
+     case SyncError::kNeedsTrustedVaultKeyForPasswords:
+     case SyncError::kNeedsTrustedVaultKeyForEverything:
+-#if !BUILDFLAG(IS_IOS)
++#if !BUILDFLAG(IS_IOS) && !BUILDFLAG(IS_ANDROID)
```
bridiver: isn't this backwards? Shouldn't we be removing the android guards in the .h file?

rewrite/components/password_manager/core/browser/password_store/password_store_built_in_backend.cc.toml
mkarolin: cc: @samartnik

patches/components-password_manager-core-browser-password_store-password_store_built_in_backend.cc.patch
mkarolin: cc: @samartnik

chromium_src/ios/web/webui/crw_web_ui_scheme_handler.mm
emerick: `GetErrorCodeForUrl(URL) != 0` is only true for WebUI URLs that aren't registered. So this will allow Brave Wallet and AI Chat but continue to fall through to the upstream check for non-registered UIs.

cc: @kylehickinson to verify my thinking there, I think we chatted about this briefly.

rewrite/chrome/browser/ui/page_actions/page_action_metrics_recorder.cc.toml
mkarolin: Done in https://github.com/brave/brave-core/pull/35415/commits/6f07fcded9897735d26aac0dea48743ff8ce987b

browser/resources/settings/brave_content_page/containers_icon.ts
```diff
@@ -48,7 +48,7 @@ export class SettingsBraveContentContainersIconElement extends CrLitElement {
   }
 
   accessor icon: Icon
-  get leoIcon() {
```
fallaciousreasoning, line 51: Hey not 100% sure but the linked Chromium CL makes it sound like the fix is meant to be to add this to define this in `properties`

fallaciousreasoning: ditto for the others

patches/ui-webui-resources-tools-build_webui.gni.patch
```diff
@@ -27,7 +27,15 @@ index dc93e6525d54e270d5e9e1141e31d8359f24378f..297605755ead5259867c97f255a4a87b
        if (is_ash_webui) {
          enable_web_component_missing_deps = false
        }
-@@ -826,7 +828,7 @@ template("build_webui") {
+@@ -631,6 +633,7 @@ template("build_webui") {
+         "//ui/webui/resources/cr_components/searchbox",
+         "//ui/webui/resources/cr_elements",
+         "//ui/webui/resources/js",
++        "//brave/ui/webui/resources",
```
fallaciousreasoning, line 34: This change doesn't seem that related to the commit message?

patches/tools-typescript-tsconfig_base.json.patch
fallaciousreasoning: Oh yeah, whoops this should have had a more descriptive commit message:

Basically, Chromium has turned on this flag which requires type definitions to be generated for side effect imports (i.e.

```
import './foo'
```

However, the build process we use for generating a bunch of our bundles doesn't emit type definitions. I've got a work in progress branch on master which gets things kind of working but I think I hit a bit of a road block. I'll try and switch back to it.

@mkarolin is it better if I edit the commit message or someone on the rebase team does?

chromium_src/chrome/browser/resources/extensions/item_list.ts
```diff
@@ -5,6 +5,6 @@
 
 import { ExtensionsItemListElement } from './item_list-chromium.js'
 
-(ExtensionsItemListElement.prototype as any).shouldShowEmptyItemsMessage_ = () => true;
+(ExtensionsItemListElement.prototype as unknown as { shouldShowEmptyItemsMessage_: () => boolean }).shouldShowEmptyItemsMessage_ = () => true
```
fallaciousreasoning, line 8: are we able to disable this rule for `chromium_src`? I think that would make more sense than changing each individual one.

browser/resources/settings/getting_started_page/brave_account_row.html.ts
fallaciousreasoning, line 1: cc @szilardszaloki for this

patches/chrome-android-java-src-org-chromium-chrome-browser-settings-search-SearchIndexProviderRegistry.java.patch
samartnik: Changed here 3625afc856484b8ca79c2ad71b68bd80da09d8a7

android/java/org/chromium/chrome/browser/BraveIntentHandler.java
samartnik: created follow up issue to research the possibility https://github.com/brave/brave-browser/issues/55635
cc @SergeyZhukovsky

chromium_src/ios/web/webui/crw_web_ui_scheme_handler.mm
kylehickinson: @bridiver @emerick To clarify: 
- `URL` refers to the request URL (which in Leo WebUI for instance would be `chrome-untrusted://leo-ai-conversation-entries` when it starts loading the iframe). 
- The chromium change introduced blocks requests between different WebUI pages unless its `chrome://resources` which breaks `chrome-untrusted` iframes. 
- There is already a similar `GetErrorCodeForUrl` check above this on the main document URL in cr149. 

This override adds a similar check for the request URL so that requests between different WebUI pages are valid as long as the WebUI function exists (`GetErrorCodeForUrl` eventually calls `GetWebUIIOSFactoryFunction`)

If we wanted to lock this down further and keep the general logic of the Chromium change but allow for `chrome-untrusted` iframes, we could do one or both of the following:
1. Compare `webUIURL` (the main document URL) and `URL` schemes such that they're different (`chrome://` cant load `chrome://`)
2. Scheme check `URL` as `chrome-untrusted://` (only check if a `chrome-untrusted` WebUI exists)

browser/resources/settings/brave_content_page/containers_icon.ts
mkarolin: Based on the other changes in the same upstream crbug, for example https://chromium-review.googlesource.com/c/chromium/src/+/7729512, they removed all `get` and just added protected/private keywords. cc: @cdesouza-chromium

patches/chrome-browser-ssl-ask_before_http_dialog_controller.cc.patch
cdesouza-chromium: Fixed f1cb548397e. Thanks, that was a smart suggestion.

cdesouza-chromium: Small tweak ab557e0efde.

browser/resources/settings/brave_content_page/containers_icon.ts
cdesouza-chromium: This is what I get reverting the change for that source:

```
/home/user/dev/brave-next/src/out/Component/gen/chrome/browser/resources/settings/preprocessed/brave_content_page/containers_icon.html.ts
  17:25  error  'get' syntax in Lit HTML templates is disallowed, encountered a getter for 'leoIcon' in class 'SettingsBraveContentContainersIconElement'  @webui-eslint/lit-reactive-properties

✖ 1 problem (1 error, 0 warnings)
```

cdesouza-chromium: I've added `protected` to the ones that didn't have it b33c10be84a

components/brave_ads/browser/ads_service_impl.cc
```diff
@@ -1055,6 +1055,9 @@ void AdsServiceImpl::ShutdownAdsService() {
 #endif
 
   application_state_monitor_observation_.Reset();
+  // TODO(tmancey): ApplicationStateMonitor is only used by this service, does
+  // it need to be a singleton?
+  ApplicationStateMonitor::GetInstance()->Reset();
```
aseren: Do we need explicitly reset `ApplicationStateMonitor` here rather than just let `ApplicationStateMonitor` destructor does it?

If we really need to call, then `ApplicationStateMonitor::GetInstance()->Reset();` should be called from `AdsServiceImpl::Shutdown()`, because `ShutdownAdsService` can be called multiple times, even if we are not closing the browser right now (for example it can be called when user clears ads data).

cdesouza-chromium: cc @mkarolin

aseren: `Reset()` was added in following commit to fix dangling pointer:
https://github.com/brave/brave-core/pull/35415/changes/95cd419e3e1e8aa5bf249efea245a6acefd4726b

patches/chrome-browser-ssl-ask_before_http_dialog_controller.cc.patch
bridiver: I actually meant 
```
inline constexpr char kLearnMoreLink[] =
    "https://support.brave.app/hc/en-us/articles/15513090104717";
``` 
so we don't need kBraveLearnMoreLink at all

chromium_src/ios/web/webui/crw_web_ui_scheme_handler.mm
bridiver: where is the code that allows `chrome://resources`?

browser/brave_browser_main_extra_parts.cc
```diff
@@ -103,3 +104,14 @@ void BraveBrowserMainExtraParts::PreMainMessageLoopRun() {
 #endif  // !BUILDFLAG(IS_ANDROID)
   g_brave_browser_process->process_misc_metrics()->uptime_monitor()->Init();
 }
+
+void BraveBrowserMainExtraParts::PostDestroyThreads() {
```
bridiver, line 110: as a follow-up we should move this and other p3a/metrics code into BraveBrowserMainExtraPartsP3A or BraveBrowserMainExtraPartsMetrics

chromium_src/ios/web/webui/crw_web_ui_scheme_handler.mm
kylehickinson: @bridiver  Its the `DomainIs` check thats happening in this override: https://source.chromium.org/chromium/chromium/src/+/refs/tags/149.0.7827.9:ios/web/webui/crw_web_ui_scheme_handler.mm;l=67

components/brave_ads/browser/ads_service_impl.cc
mkarolin: Thanks! Moved to Shutdown in https://github.com/brave/brave-core/pull/35415/commits/a101a2e5d95574a9690b0c38d33b2c38acfb73da

browser/brave_browser_main_extra_parts.cc
mkarolin: Created https://github.com/brave/brave-browser/issues/55643

MonotonicObservableSupplier
```diff
@@ -0,0 +1 @@
+b8cb84d2901ae Rename ObservableSupplier -
```
AlexeyBarabash, line 1: looks like a strange leftover

AlexeyBarabash: Please delete it @samartnik

rewrite/components/password_manager/core/browser/password_store/password_store_built_in_backend.cc.toml
samartnik: Removed this plaster in favour of `chromium_src` override https://github.com/brave/brave-core/commit/37733a76d79eb8b4b4985ac4923f862f2c3611ba

AlexeyBarabash (approved):
Android ++, Sync ++, basic sync tests Desktop <=> Android work fine

components/brave_ads/browser/ads_service_impl.cc
aseren: Filed an issue to refactor ApplicationStateMonitor from singleton to AdsServiceImpl member: https://github.com/brave/brave-browser/issues/55645

patches/components-password_manager-core-browser-password_store-password_store_built_in_backend.cc.patch
samartnik: Initially I was following upstream where they exclude both ios and android when handling these values https://source.chromium.org/chromium/chromium/src/+/main:components/sync/service/sync_internals_util.cc;drc=d78abe4e930bc802ef45f8a9a1c5ca2919e6b4bd;l=205
Re-adding them in header is cleaner approach, added it here https://github.com/brave/brave-core/commit/37733a76d79eb8b4b4985ac4923f862f2c3611ba
Just to clarify it, these are essentially dummy values just to fix the build, as in other `cc` their handling is still guarded by `!is_android`.

aseren (approved):
brave ads++

MonotonicObservableSupplier
samartnik: Removed here https://github.com/brave/brave-core/commit/19e6920e3130b962c8d3164691bee69a2dfe4746

AlexeyBarabash: thank you!

chromium_src/ios/web/webui/crw_web_ui_scheme_handler.mm
bridiver: I think including a scheme check for `chrome-untrusted` would probably be good here, but I think either way this needs a @brave/sec-team sign off if they haven't already?

chromium_src/chrome/browser/ui/views/frame/browser_frame_view_win.cc
```diff
@@ -24,14 +23,12 @@
...
 
-#define SupportsTouchableTabStrip(browser)                                  \
-  SupportsTouchableTabStrip(browser);                                       \
-  supports_title =                                                          \
-      supports_title_bar ||                                                 \
-      (WebUITabStripContainerView::SupportsTouchableTabStrip(browser)       \
-           ? WebUITabStripContainerView::SupportsTouchableTabStrip(browser) \
-           : tabs::utils::SupportsBraveVerticalTabs(browser))
-
+// Add titlebar if Brave vertical tabs are used.
```
bridiver: This comment doesn't seem to be consistent with the code. It has supports title bar OR supports vertical tabs

bridiver: actually I think this is a problematic override because it duplicates code in the origin check `!browser_view->GetIsWebAppType() && supports_title_bar`. Also is it correct that it omits

```
// Only show icons if the browser supports title bars.
  if (supports_title_bar) {
    InitThrobberIcons();

    AddChildView(views::Builder<TabIconView>()
                     .CopyAddressTo(&window_icon_)
                     .SetModel(this)
                     .SetID(VIEW_ID_WINDOW_ICON)
                     // Stop the icon from intercepting clicks intended for the
                     // HTSYSMENU region of the window. Even though it does
                     // nothing on click, it will still prevent us from giving
                     // the event back to Windows to handle properly.
                     .SetCanProcessEventsWithinSubtree(false)
                     .Build());
  }
``` 
above? If we want both we could override kFeatureTitleBar to set `supports_title_bar = browser->SupportsWindowFeature(Browser::WindowFeature::kFeatureTitleBar) || tabs::utils::SupportsBraveVerticalTabs(browser)`

mkarolin: In terms of omitting the above, I think @sangwoo108 would have a better answer if it's intentional.

chromium_src/ios/web/webui/crw_web_ui_scheme_handler.mm
emerick: We added a scheme check in https://github.com/brave/brave-core/pull/35415/commits/cf8b725daaaadb3ddf2e3e7edb7c3c432cada92c and also split the allow-list logic into a separate function apart from the macro, so it's easier to see and reason about. Finally, @kylehickinson mentioned that a few additional checks were needed in there for `image` and `favicon2` so added those as well.

fallaciousreasoning (approved):

browser/resources/settings/brave_content_page/containers_icon.ts
fallaciousreasoning: All right, sounds like I'm wrong!

mkarolin:
@fallaciousreasoning 

> @mkarolin is it better if I edit the commit message or someone on the rebase team does?

Please feel free to edit the message via git commit --fixup=amend:<commit_hash>

patches/ui-webui-resources-tools-build_webui.gni.patch
mkarolin: cc: @cdesouza-chromium

chromium_src/chrome/browser/resources/extensions/item_list.ts
mkarolin: I am not sure if we have a per-directory control over such things or just the global eslint.config.mjs. cc: @cdesouza-chromium

chromium_src/chrome/browser/ui/views/frame/browser_frame_view_win.cc
sangwoo108: Yes, it was intended back then. The tab icon and throbber means the small icon on the title bar in the popup window, which we don't want in normal browser.

bridiver: Talk to @cdesouza-chromium about maybe using plaster here or normal patching because it's not appropriate to duplicate upstream code in chromium src as I mentioned above.

patches/chrome-browser-ssl-ask_before_http_dialog_controller.cc.patch
cdesouza-chromium: Fixed in 379d9f5c128.

chromium_src/chrome/browser/resources/extensions/item_list.ts
cdesouza-chromium: @fallaciousreasoning you tell me? I'm not very familiar with this tooling.

patches/ui-webui-resources-tools-build_webui.gni.patch
cdesouza-chromium: If I remove this line, this is what I get:

```
/home/user/dev/brave-next/src/out/Component/gen/brave/ui/webui/resources/preprocessed/lit_overriding.ts
  16:30  error  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any
  22:39  error  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any

/home/user/dev/brave-next/src/out/Component/gen/brave/ui/webui/resources/preprocessed/polymer_overriding.ts
   29:60   error  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any
   41:74   error  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any
   54:69   error  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any
  119:102  error  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any
  ```
  
  So that matches the commit description:
  
  ```
  [cr149][WIP] Fixing @typescript-eslint/no-explicit-any warnings

Chromium changes:
[https://chromium.googlesource.com/chromium/src/+/8f97262a2cc5aa37fcd710c3c141eb7f08c01f92](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)

commit 8f97262a2cc5aa37fcd710c3c141eb7f08c01f92
Author: dpapad <[dpapad@chromium.org](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)>
Date: Wed Apr 15 15:03:48 2026 -0700

WebUI: Enable @typescript-eslint/no-explicit-any where possible.
```

chromium_src/chrome/browser/resources/extensions/item_list.ts
cdesouza-chromium: I'm not very good with TS script, so I won't pretend to be an authority on it. So, my **uninformed** opinion is that I want brave code to be more in line with Chromium code, not less, because my **uninformed** inclination is to think the more we disable certain types of warnings, the more things will break in the future.

chromium_src/chrome/browser/ui/views/frame/browser_frame_view_win.cc
sangwoo108: I think moving `BRAVE_BROWSER_FRAME_VIEW_WIN` up and setting `supports_title_bar = true` would be better as we can avoid copying rest of conditions. 

```diff
+ BRAVE_BROWSER_FRAME_VIEW_WIN
  // If this is a web app window, the window title will be part of the
  // BrowserView and thus we don't need to create another one here.
  if (!browser_view->GetIsWebAppType() && supports_title_bar) {
- BRAVE_BROWSER_FRAME_VIEW_WIN
```

```
#define   BRAVE_BROWSER_FRAME_VIEW_WIN supports_title_bar = supports_title_bar || SupportsBraveVerticalTabs()
``` 

@bridiver , @cdesouza-chromium does it make sense?

cdesouza-chromium: It depends how the `BRAVE_BROWSER_FRAME_VIEW_WIN` is gonna looks like, but yeah, this is becoming a bit hard to understand... I'm having a look.

mkarolin: Replaced it with a direct patch in https://github.com/brave/brave-core/pull/35415/commits/e9e6ad8e656e353a93259281e241db4c52c82809

bridiver (approved):

bsclifton (approved):
string reviewers ++

patches/ui-webui-resources-tools-build_webui.gni.patch
fallaciousreasoning: thats bizzare

chromium_src/chrome/browser/resources/extensions/item_list.ts
fallaciousreasoning: @mkarolin I think we have a bunch of per file/directory rules in `//brave/eslint.config.mjs`

@cdesouza-chromium wrt to diverging from upstream I think there's a case to be made that `chromium_src` is special, and that its expected that we're going to do kind of unconventional things. 

Additionally, I can't see any advantage to doing an `as unknown as { exactlyWhatIneed }`. Its more verbose, more repetitive and no more type safe.

cc @petemill for another opinion.

fallaciousreasoning: that said, I guess its fine not to fix it here - if it ends up being painful we can add the eslint exception later

petemill: Yeah this doesn't improve anything. Either we somehow fix it so typescript knows what type `ExtensionsItemListElement.prototype` is, or we disable the rule IMO because this just adds noise and doesn't verify anything.

browser/resources/settings/getting_started_page/brave_account_row.html.ts
szilardszaloki: Thanks for the heads-up!

So `@webui-eslint/lit-element-template-structure` first made us move the markup into the TS file because of the `createFirstRow()` helper, and now it wants us to move it back into `.html.ts` — nice 🙂. Our options are basically to duplicate the contents, or extract them into a custom element, which is obviously an unreasonable amount of ceremony for this, so let's leave it as-is.

PR #35418, opened 2026-04-10
Add custom attribute support for metrics

### LINKED ISSUE #54377: Add custom attribute support for metrics
Currently, we support a finite list of static attributes which can be included in metric reports (i.e. platform, channel, date of install, etc.). In order to support multi-dimensional P3A metrics, where the answer to the question could be encoded within multiple attributes as opposed to the single `answer_index` attribute, we should:

1. provide the ability to set arbitrary custom attribute values from anywhere within the browser process
2. provide the ability to enable those custom attributes via metric configuration

DJAndries:
Resolves https://github.com/brave/brave-browser/issues/54377


components/p3a/metric_config.cc
```diff
@@ -120,6 +121,27 @@ bool GetOptionalBool(const base::Value* value, std::optional<bool>* field) {
...
+    return false;
+  }
+
+  const auto& list = value->GetList();
+
+  RemoteCustomAttributes attributes;
+
+  for (size_t i = 0; i < list.size() && i < attributes.size(); i++) {
+    if (!list[i].is_string()) {
+      return false;
```
Olipro, line 136: Just a sanity check: You intentionally bail out here instead of a `continue;` right?

components/p3a/p3a_service.h
```diff
@@ -24,6 +24,7 @@
 #include "brave/components/p3a/metric_log_type.h"
 #include "brave/components/p3a/p3a_config.h"
 #include "brave/components/p3a/remote_config_manager.h"
+#include "brave/components/p3a_utils/event_relay.h"
```
netzenbot, line 27: `p3a_service.h` includes `brave/components/p3a_utils/event_relay.h`, making `p3a_utils` a header-visible dependency of the `p3a` target. `//brave/components/p3a_utils` should be moved to `public_deps` in `components/p3a/BUILD.gn` so that targets depending on `:p3a` can also resolve this include. See [BS-024](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-024).

components/p3a/p3a_service.cc
```diff
@@ -342,6 +346,17 @@ void P3AService::OnDefaultBrowserStatusChanged(bool is_default) {
 }
```
netzenbot, line 346: The new `event_relay_observation_` is not reset in `StartTeardown()`, unlike `default_browser_observation_` which is explicitly reset there. After `StartTeardown()`, the pref registrar is removed and the service is in a partially-torn-down state, but `OnCustomAttributeSet` can still fire from the global `EventRelay` singleton and attempt to write prefs. Add `event_relay_observation_.Reset()` in `StartTeardown()` to match the existing pattern. [CSM-017](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-memory.md#CSM-017)

components/p3a/p3a_message_unittest.cc
```diff
@@ -0,0 +1,146 @@
...
+
+  std::string message = GenerateP3AConstellationMessage(
+      kTestMetricName, kTestMetricValue, meta_, kP3AUploadType, &kConfig);
+
+  EXPECT_EQ(message,
+            "metric_name|Brave.Test.Message"
+            ";metric_value|3"
+            ";channel|release"
+            ";dtoi|none");
+
```
netzenbot, line 66: The new `GetCustomAttributes` parser in `metric_config.cc` explicitly handles two invalid-input paths (value is not a list; a list element is not a string), but neither is exercised by any test. Consider adding cases to the remote config tests (or a new `metric_config` test) that pass a non-list value and a list containing a non-string element. https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-016

components/p3a/metric_config.cc
```diff
@@ -120,6 +121,27 @@ bool GetOptionalBool(const base::Value* value, std::optional<bool>* field) {
   return true;
```
netzenbot, line 121: The new `GetCustomAttributes` deserializer has two distinct error paths (`!value->is_list()` and `!list[i].is_string()`) that aren't covered by any test. TI-022 requires testing deserialization of invalid/malformed input. Consider adding cases to `remote_config_manager_unittest.cc` (or a dedicated metric_config test) that supply a non-list value and a list with a non-string element, and verify the field remains unset. https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-022

components/p3a/p3a_service.cc
DJAndries: that's okay, both `EventRelay` and local state will outlive `P3AService`

components/p3a/metric_config.cc
DJAndries: Yes, if the element is not a string, then the metric config does not fit the schema and parsing should fail.

components/p3a/p3a_service.h
DJAndries: 77322433527a3b008fc614895d9c7859da308324

components/p3a/p3a_message_unittest.cc
DJAndries: 77322433527a3b008fc614895d9c7859da308324

components/p3a/metric_config.cc
DJAndries: 77322433527a3b008fc614895d9c7859da308324

components/p3a_utils/event_relay.h
```diff
@@ -0,0 +1,50 @@
...
+#include "base/observer_list.h"
+#include "base/observer_list_types.h"
+
+namespace p3a_utils {
+
+class EventRelay {
+ public:
+  class Observer : public base::CheckedObserver {
+   public:
+    virtual void OnCustomAttributeSet(
```
netzenbot, line 22: nit: The `OnCustomAttributeSet` virtual method doesn't document the nullopt contract. `custom_attributes.h` explains it for callers of `SetCustomAttribute`, but anyone implementing this observer interface directly won't know that `std::nullopt` signals attribute removal. A brief comment here would help. https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-004

components/p3a/p3a_message_unittest.cc
```diff
@@ -0,0 +1,146 @@
...
+#endif
+
+    base::Time install_time;
+    ASSERT_TRUE(base::Time::FromString("2020-01-01", &install_time));
+    meta_.Init(&local_state_, "release", install_time);
+  }
+
+  TestingPrefServiceSimple local_state_;
+  MessageMetainfo meta_;
+};
```
netzenbot, line 47: nit: `Basic` doesn't describe what the test verifies. Per TI-038, names should be descriptive enough to understand the scenario at a glance — something like `GeneratesCorrectMessageWithStandardAttributes` or `SerializesMetricNameValueAndChannel` would be clearer. https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-038

components/p3a_utils/event_relay.h
DJAndries: The meaning of nullopt is implied, and it's also up to the observer impl to decide what to do with a nullopt.

components/p3a/p3a_message_unittest.cc
DJAndries: it seems straightforward to me. it's a basic, general test of the happy path.

netzenbot (approved):

Olipro (approved):
Aside from CI issues, LGTM.

PR #35421, opened 2026-04-10
Ignore unknown attribute names in P3A manifest

DJAndries:
Resolves

### DIFF
--- a/components/p3a/metric_config.cc
+++ b/components/p3a/metric_config.cc
@@ -62,44 +62,27 @@ bool GetMetricLogType(const base::Value* value,
   return true;
 }
 
-bool GetMetricAttributes(const base::Value* value,
-                         std::optional<MetricAttributes>* field) {
+template <size_t N>
+bool GetMetricAttributes(
+    const base::Value* value,
+    std::optional<std::array<std::optional<MetricAttribute>, N>>* field) {
   if (!value || !value->is_list()) {
     return false;
   }
 
-  const auto& list = value->GetList();
-
-  MetricAttributes attributes;
-
-  for (size_t i = 0; i < list.size() && i < attributes.size(); i++) {
-    if (!GetMetricAttribute(&list[i], &attributes[i])) {
-      return false;
+  std::array<std::optional<MetricAttribute>, N> attributes;
+  size_t attr_idx = 0;
+  for (const auto& item : value->GetList()) {
+    if (attr_idx >= N) {
+      break;
     }
+    if (!GetMetricAttribute(&item, &attributes[attr_idx])) {
+      continue;
+    }
+    attr_idx++;
   }
 
   *field = attributes;
-  return true;
-}
-
-bool GetAppendAttributes(const base::Value* value,
-                         std::optional<MetricAttributesToAppend>* field) {
-  if (!value || !value->is_list()) {
-    return false;
-  }
-
-  const auto& list = value->GetList();
-
-  MetricAttributesToAppend attributes;
-
-  for (size_t i = 0; i < list.size() && i < attributes.size(); i++) {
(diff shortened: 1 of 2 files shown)


components/p3a/metric_config.cc
```diff
@@ -62,44 +62,27 @@ bool GetMetricLogType(const base::Value* value,
   return true;
```
netzenbot, line 62: nit: The `continue` here silently skips unknown attribute names, which is the intent of this PR, but a future reader would expect `GetMetricAttribute` failure to be an error. A brief comment like `// Skip unknown attribute names` would make this intentional behavior clear. See [CS-021](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-021).

components/p3a/metric_config.cc
```diff
@@ -62,44 +62,27 @@ bool GetMetricLogType(const base::Value* value,
...
-
-  const auto& list = value->GetList();
-
-  MetricAttributesToAppend attributes;
-
-  for (size_t i = 0; i < list.size() && i < attributes.size(); i++) {
-    if (!GetMetricAttribute(&list[i], &attributes[i])) {
-      return false;
+    if (!GetMetricAttribute(&item, &attributes[attr_idx])) {
+      continue;
```
Olipro, line 83: Sanity check: Is it intentional that you continue the loop but `attr_idx` remains unincremented?

DJAndries: yes, added a comment for clarity

components/p3a/metric_config.cc
DJAndries: https://github.com/brave/brave-core/pull/35421#discussion_r3067141241

netzenbot (approved):

components/p3a/metric_config.cc
```diff
@@ -62,44 +62,30 @@ bool GetMetricLogType(const base::Value* value,
   return true;
 }
 
-bool GetMetricAttributes(const base::Value* value,
-                         std::optional<MetricAttributes>* field) {
+template <size_t N>
+bool GetMetricAttributes(
+    const base::Value* value,
+    std::optional<std::array<std::optional<MetricAttribute>, N>>* field) {
```
Olipro, line 68: A pointer to an optional array of optionals? It's typically considered a code-smell to take a pointer to an optional. What's the use-case here?

DJAndries: `JSONValueConverter::RegisterCustomValueField` requires a function with this signature. see https://source.chromium.org/chromium/chromium/src/+/main:base/json/json_value_converter.h;l=419-421

Olipro: I see. After a bit of investigation, I've discovered that this is one of the reasons it's recommended against using `JSONValueConverter` in favour of using the JSON schema compiler instead.

Any reason why these metrics wouldn't be schema-friendly?

DJAndries: > I've discovered that this is one of the reasons it's recommended against using `JSONValueConverter` in favour of using the JSON schema compiler instead.

Care to share a link to a resource that suggests this?

In any case, if the remote config can be parsed via IDL/schema/whatever, it should be considered orthogonal to this PR. It would require a refactor of `metric_config.cc` which is out of scope here.

Olipro: > Care to share a link to a resource that suggests this?

The resource in this case is a person - I spoke about this with Claudio.

Given the situation, it sounds reasonable that you're working with the way it currently is and it would be an undue burden to refactor the whole thing.

Olipro (approved):
Comments should be resolved. LGTM.

PR #35422, opened 2026-04-10
[Wallet] Security fixes for Polkadot metadata parsing

kdenhartog:
## Summary

Fixes several security issues in the Polkadot chain metadata parser (`polkadot_chain_metadata.rs`) identified during the review of #34784, and adds comprehensive test coverage for each concern.

## Fixes

1. **v14/v15 Map storage format mismatch** — `parse_storage_entry_type` now tries decoding hashers as `Vec<StorageHasher>` (v15) first, and falls back to a single `StorageHasher` byte (v14) on failure. Previously, v14 metadata with certain hasher values (e.g. `Blake2_256`) would fail to parse, causing the wallet to not recognize the chain.

2. **SS58 prefix Compact\<u32\> encoding not handled** — `decode_ss58_prefix` now also tries `Compact<u32>` decoding (after fixed-width `u16`/`u32` to avoid misinterpreting multi-byte encodings). Some runtimes encode SS58Prefix as compact, which would previously be silently skipped, falling through to incorrect values.

3. **Unbounded Vec allocation** — `decode_vec_len` now rejects lengths > 10,000 before allocation, preventing a potential DoS from a malicious RPC response with an enormous Compact length value.

## Tests

Adds 7 targeted unit tests covering each security concern:
- v14 metadata with no storage entries (baseline)
- v14 Map storage with hasher=0 (Blake2_128) — Vec path
- v14 Map storage with hasher=1 (Blake2_256) — single-byte fallback
- v15 Map storage with hasher=1 as Vec\<StorageHasher\>
- SS58Prefix encoded as u16 and as Compact\<u32\>
- Known limitation: trailing bytes not rejected (documented)
- Huge vec length rejected by bounds check

### DIFF
--- a/components/brave_wallet/browser/internal/polkadot_chain_metadata.rs
+++ b/components/brave_wallet/browser/internal/polkadot_chain_metadata.rs
@@ -74,9 +74,15 @@ fn decode_scale<T: Decode>(input: &mut &[u8]) -> Result<T, Error> {
     T::decode(input).map_err(|_| Error::InvalidScale)
 }
 
+const MAX_VEC_LEN: usize = 10_000;
+
 fn decode_vec_len(input: &mut &[u8]) -> Result<usize, Error> {
     let len: Compact<u32> = decode_scale(input)?;
-    usize::try_from(len.0).map_err(|_| Error::InvalidLength)
+    let len = usize::try_from(len.0).map_err(|_| Error::InvalidLength)?;
+    if len > MAX_VEC_LEN {
+        return Err(Error::InvalidLength);
+    }
+    Ok(len)
 }
 
 fn decode_type_id(input: &mut &[u8]) -> Result<u32, Error> {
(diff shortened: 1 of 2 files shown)


cypt4 (approved):

netzenbot (approved):

components/brave_wallet/browser/internal/polkadot_chain_metadata.rs
```diff
@@ -233,10 +239,21 @@ fn parse_storage_entry_type(input: &mut &[u8]) -> Result<(), Error> {
...
                 let _: u8 = decode_scale(input)?;
                 Ok(())
-            })?;
+            })
+            .is_ok()
+            {
+                let _: u32 = decode_type_id(input)?;
+                let _: u32 = decode_type_id(input)?;
+                return Ok(());
+            }
```
cmazakas, line 253: The whitespace for this code makes it a bit difficult to read. It seems like something like this still compiles and passes all the tests:

```rs
            // Most runtimes (including v14 Polkadot) use Vec<StorageHasher>.
            if decode_vec(input, decode_scale::<u8>).is_ok() {
                let _: u32 = decode_type_id(input)?;
                let _: u32 = decode_type_id(input)?;
                return Ok(());
            }
```

Note that we don't _need_ to use a function pointer here and we can replace it with `|input| decode_scale::<u8>(input)` if you'd like.

cmazakas (approved):

cmazakas (approved):
Looks great!

PR #35425, opened 2026-04-10
[Polkadot] Use real chain metadata in polkadot_wallet_service

### LINKED ISSUE #54460: [Polkadot] Add metadata provider and persistent storage
### Description

Currently hardcoded chain metadata is used for tx construction.
We need to add persistent metadata storage and metadata provider so existing code uses fetched metadata.

### Steps to reproduce

-

### Actual result

-

### Expected result

-

### Reproduces how often

Easily reproduced

-

### Channel information

- [ ] release (stable)
- [ ] beta
- [ ] nightly

### Reproducibility

### Miscellaneous information

_No response_

cypt4:
Currently hardcoded chain metadata is used for tx construction.
We need to add persistent metadata storage and metadata provider so existing code uses fetched metadata.
Resolves https://github.com/brave/brave-browser/issues/54460


components/brave_wallet/browser/polkadot/polkadot_metadata_provider.cc
```diff
@@ -0,0 +1,164 @@
...
+void PolkadotMetadataProvider::GetChainMetadata(
+    std::string_view chain_id,
+    GetChainMetadataCallback callback) {
+  CHECK(IsPolkadotNetwork(chain_id));
+
+  auto cache_it = metadata_cache_.find(chain_id);
+  if (cache_it != metadata_cache_.end()) {
+    std::move(callback).Run(base::ok(cache_it->second));
+    return;
+  }
```
cmazakas: This can probably be transformed to:
```cpp
if (const auto* p = base::FindOrNull(metadata_cache_, chain_id)) {
  return std::move(callback).Run(base::ok(*p));
}
```

components/brave_wallet/browser/polkadot/polkadot_wallet_service.cc
```diff
@@ -7,45 +7,28 @@
```
netzenbot, line 7: The `CHECK(profile_prefs)` in the constructor body is too late — `chain_metadata_prefs_(*profile_prefs)` in the initializer list already dereferences the pointer. If `profile_prefs` is null, you get UB before the CHECK fires. Change the parameter type to `PrefService&` to enforce non-null at the call site, which is what `PolkadotChainMetadataPrefs` already takes in its constructor.

components/brave_wallet/browser/polkadot/polkadot_metadata_provider.cc
```diff
@@ -0,0 +1,164 @@
...
+  init_started_ = true;
+
+  RetryInitChainMetadata(mojom::kPolkadotMainnet, /*attempt=*/1);
+  RetryInitChainMetadata(mojom::kPolkadotTestnet, /*attempt=*/1);
+}
+
+void PolkadotMetadataProvider::GetChainMetadata(
+    std::string_view chain_id,
+    GetChainMetadataCallback callback) {
+  CHECK(IsPolkadotNetwork(chain_id));
```
netzenbot, line 59: The callback in `GetChainMetadata` is invoked synchronously when the in-memory cache is populated, but asynchronously when it falls through to `FetchAndParseChainMetadata` or `GetRuntimeVersion`. Mixed sync/async execution means callers cannot reason about when their callback fires and can lead to subtle ordering bugs. Per [ARCH-060](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-060), the cache-hit branch should post the callback asynchronously (e.g., via `base::SequencedTaskRunner::GetCurrentDefault()->PostTask`).

components/brave_wallet/browser/polkadot/polkadot_wallet_service.h
```diff
@@ -38,6 +41,7 @@ class PolkadotWalletService : public mojom::PolkadotWalletService,
   PolkadotWalletService(
       KeyringService& keyring_service,
       NetworkManager& network_manager,
```
netzenbot, line 46: The `profile_prefs` parameter is non-nullable — the constructor immediately `CHECK`s it — so it should be taken as `PrefService&` rather than `PrefService*`. Pointers imply optional nullability; references communicate the non-null invariant at the call site. See [CSA-030](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-030).

components/brave_wallet/browser/polkadot/polkadot_wallet_service.cc
```diff
@@ -7,45 +7,28 @@
...
     scoped_refptr<network::SharedURLLoaderFactory> url_loader_factory)
     : keyring_service_(keyring_service),
       network_manager_(network_manager),
-      polkadot_substrate_rpc_(network_manager, std::move(url_loader_factory)) {
-  keyring_service_->AddObserver(
-      keyring_service_observer_receiver_.BindNewPipeAndPassRemote());
+      polkadot_substrate_rpc_(network_manager, std::move(url_loader_factory)),
+      chain_metadata_prefs_(*profile_prefs),
+      metadata_provider_(chain_metadata_prefs_, polkadot_substrate_rpc_) {
+  CHECK(profile_prefs);
```
netzenbot: The `CHECK(profile_prefs)` in the constructor body fires too late — `*profile_prefs` is already dereferenced in the member initializer list at `chain_metadata_prefs_(*profile_prefs)`. If `profile_prefs` is null, you get UB before the CHECK. Fix by taking `PrefService&` by reference in the constructor signature (matching the pattern used for `KeyringService&` and `NetworkManager&`), which makes the null-check unnecessary. See [CSM-029](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-memory.md#CSM-029).

components/brave_wallet/browser/polkadot/polkadot_chain_metadata_prefs.cc
```diff
@@ -0,0 +1,156 @@
...
+      ReadUint8(*chain_metadata, kTransactionPaymentPalletIndex);
+  auto transfer_allow_death_call_index =
+      ReadUint8(*chain_metadata, kTransferAllowDeathCallIndex);
+  auto transfer_keep_alive_call_index =
+      ReadUint8(*chain_metadata, kTransferKeepAliveCallIndex);
+  auto ss58_prefix = ReadUint16(*chain_metadata, kSs58Prefix);
+  auto spec_version = ReadUint32(*chain_metadata, kSpecVersion);
+  if (!system_pallet_index || !balances_pallet_index ||
+      !transaction_payment_pallet_index || !transfer_allow_death_call_index ||
+      !transfer_keep_alive_call_index || !ss58_prefix || !spec_version) {
```
cmazakas: I think this code will be easier to maintain if we treat it like Go, and put the error handling branches directly after the initialization.

So like:
```cpp
  auto balances_pallet_index = ReadUint8(*chain_metadata, kBalancesPalletIndex);
  if(!balances_pallet_index) { return std::nullopt; }
```

I was almost wondering if we could use the .idl files to generate these JSON bindings for us but we don't gain much here because those generated bindings don't apply the kind of numeric validation we're looking for.

components/brave_wallet/browser/polkadot/polkadot_chain_metadata_prefs.cc
```diff
@@ -0,0 +1,156 @@
...
+std::optional<uint32_t> ReadUint32(const base::DictValue& dict,
+                                   const char* key) {
+  auto value = dict.FindInt(key);
+  uint32_t converted_value = 0;
+  if (!value ||
+      !base::CheckedNumeric<uint32_t>(*value).AssignIfValid(&converted_value)) {
+    return std::nullopt;
+  }
+  return converted_value;
+}
```
cmazakas: I mulled over replacing these 3 with a singular function template like:
```cpp
template <class T>
bool ReadUint(const base::DictValue& dict,
              const char* key,
              T* converted_value) {
  auto value = dict.FindInt(key);
  if (!value ||
      !base::CheckedNumeric<T>(*value).AssignIfValid(converted_value)) {
    return false;
  }
  return true;
}
```
but I'm not sure this really helps all that much.

components/brave_wallet/browser/polkadot/polkadot_chain_metadata_unittest.cc
```diff
@@ -28,53 +28,81 @@ TEST(PolkadotChainMetadataUnitTest, FromFields) {
...
-  EXPECT_EQ(metadata.GetTransferKeepAliveCallIndex(), 4u);
-  EXPECT_EQ(metadata.GetSs58Prefix(), 42u);
-  EXPECT_EQ(metadata.GetSpecVersion(), 1'234'567u);
+  auto expected = PolkadotChainMetadata::FromFields(
+      /*system_pallet_index=*/0, /*balances_pallet_index=*/7,
+      /*transaction_payment_pallet_index=*/0x20,
+      /*transfer_allow_death_call_index=*/2,
+      /*transfer_keep_alive_call_index=*/4,
+      /*ss58_prefix=*/42, /*spec_version=*/1'234'567);
+  EXPECT_EQ(metadata, expected);
```
cmazakas: This change doesn't actually test the same thing. For example, if `FromFields` stores the wrong thing, we won't know here because comparison is going to pass.

I'd keep the original version with all the getters compared directly against the hard-coded values.

components/brave_wallet/browser/polkadot/polkadot_chain_metadata_unittest.cc
```diff
@@ -161,22 +190,23 @@ TEST(PolkadotChainMetadataUnitTest, ParseRealStateGetMetadataResponseWestend) {
...
-  EXPECT_EQ(metadata2->GetBalancesPalletIndex(),
-            metadata->GetBalancesPalletIndex());
-  EXPECT_EQ(metadata2->GetSpecVersion(), 0u);
+  auto expected_from_name = PolkadotChainMetadata::FromFields(
+      /*system_pallet_index=*/0, /*balances_pallet_index=*/4,
+      /*transaction_payment_pallet_index=*/0x1a,
+      /*transfer_allow_death_call_index=*/0,
+      /*transfer_keep_alive_call_index=*/3,
+      /*ss58_prefix=*/42, /*spec_version=*/0);
+  EXPECT_EQ(*metadata2, expected_from_name);
```
cmazakas: This change keeps functional parity of the tests but seems like it can be replaced with just a single:
```cpp
EXPECT_EQ(*metadata2, expected);
```

components/brave_wallet/browser/polkadot/polkadot_test_utils.h
```diff
@@ -6,6 +6,10 @@
 #ifndef BRAVE_COMPONENTS_BRAVE_WALLET_BROWSER_POLKADOT_POLKADOT_TEST_UTILS_H_
 #define BRAVE_COMPONENTS_BRAVE_WALLET_BROWSER_POLKADOT_POLKADOT_TEST_UTILS_H_
 
+#include <string>
+#include <string_view>
+#include <vector>
```
cmazakas: Seems like the only new `#include` we should need here is the `<string_view>`, no?

components/brave_wallet/browser/polkadot/polkadot_wallet_service.h
```diff
@@ -102,30 +106,13 @@ class PolkadotWalletService : public mojom::PolkadotWalletService,
       GetFeeEstimateCallback callback);
 
  private:
-  // KeyringServiceObserverBase:
-  void Unlocked() override;
```
cmazakas, line 106: Last time Anton and I had decided that deferring chain metadata to wallet unlock was the move, instead of service instantiation which if I recall correctly happens before the wallet's unlocked. I'm not sure we should move it into the constructor directly.

cmazakas:
So, it seems like no other RPC service we have in Wallet uses `virtual`s for the sake of mocking. I'm not sure if we should introduce this change here or not.

I recommend seeing what @supermassive thinks about this. In principle, I'm not opposed to updating the `PolkadotSubstrateRpc` to be a `virtual` class but it seems like it's not established practice for us in Wallet.

Edit:

I think making _all_ the members of `PolkadotSubstrateRpc` `virtual` is fine, so this way we can use whatever makes sense for our unittests. I think in some cases, the mock RPC layer we have is largely sufficient and winds up exercising a lot of our codepaths which is nice. But by the same token, being able to mock an RPC function is also useful so I'm okay with everything being `virtual`.

components/brave_wallet/browser/polkadot/polkadot_test_utils.h
cypt4: string is used, also added some missing includes here

components/brave_wallet/browser/polkadot/polkadot_chain_metadata_prefs.cc
cypt4: Done, I think we should migrate to mapping <pallet name, pallet_data> where pallet data is <method name, method index>

components/brave_wallet/browser/polkadot/polkadot_chain_metadata_prefs.cc
cypt4: Done

components/brave_wallet/browser/polkadot/polkadot_chain_metadata_unittest.cc
cypt4: Changed to comparing field values

components/brave_wallet/browser/polkadot/polkadot_wallet_service.h
cypt4: Moved initialization to unlocked

cypt4:
> So, it seems like no other RPC service we have in Wallet uses `virtual`s for the sake of mocking. I'm not sure if we should introduce this change here or not.
> 
> I recommend seeing what @supermassive thinks about this. In principle, I'm not opposed to updating the `PolkadotSubstrateRpc` to be a `virtual` class but it seems like it's not established practice for us in Wallet.
> 
> Edit:
> 
> I think making _all_ the members of `PolkadotSubstrateRpc` `virtual` is fine, so this way we can use whatever makes sense for our unittests. I think in some cases, the mock RPC layer we have is largely sufficient and winds up exercising a lot of our codepaths which is nice. But by the same token, being able to mock an RPC function is also useful so I'm okay with everything being `virtual`.

ZCash uses RPC mocking a lot :)

components/brave_wallet/browser/polkadot/polkadot_chain_metadata_unittest.cc
```diff
@@ -161,22 +190,17 @@ TEST(PolkadotChainMetadataUnitTest, ParseRealStateGetMetadataResponseWestend) {
...
-            metadata->GetTransferAllowDeathCallIndex());
-  EXPECT_EQ(metadata2->GetTransferKeepAliveCallIndex(),
-            metadata->GetTransferKeepAliveCallIndex());
-  EXPECT_EQ(metadata2->GetBalancesPalletIndex(),
-            metadata->GetBalancesPalletIndex());
-  EXPECT_EQ(metadata2->GetSpecVersion(), 0u);
+  EXPECT_EQ(*metadata2, expected);
 }
 
 }  // namespace brave_wallet
```
netzenbot, line 202: This comparison is incorrect — `expected` was built with `spec_version=1'022'000`, but `FromChainName("Westend")` returns a struct with `spec_version=0`. The old code checked `EXPECT_EQ(metadata2->GetSpecVersion(), 0u)` explicitly. You need a separate `expected_from_name` variable with `spec_version=0` here (as was done in the `ParseRealStateGetMetadataResponsePolkadot` test), otherwise this assertion will always fail.

components/brave_wallet/browser/polkadot/polkadot_chain_metadata_unittest.cc
```diff
@@ -161,22 +190,17 @@ TEST(PolkadotChainMetadataUnitTest, ParseRealStateGetMetadataResponseWestend) {
...
-  EXPECT_EQ(metadata2->GetSs58Prefix(), metadata->GetSs58Prefix());
-  EXPECT_EQ(metadata2->GetTransferAllowDeathCallIndex(),
-            metadata->GetTransferAllowDeathCallIndex());
-  EXPECT_EQ(metadata2->GetTransferKeepAliveCallIndex(),
-            metadata->GetTransferKeepAliveCallIndex());
-  EXPECT_EQ(metadata2->GetBalancesPalletIndex(),
-            metadata->GetBalancesPalletIndex());
-  EXPECT_EQ(metadata2->GetSpecVersion(), 0u);
+  EXPECT_EQ(*metadata2, expected);
 }
```
netzenbot, line 201: `metadata2` comes from `FromChainName("Westend")`, which sets `spec_version=0` per the documented invariant. But `expected` was constructed with `spec_version=1'022'000`. This comparison will always fail — the original test explicitly checked `EXPECT_EQ(metadata2->GetSpecVersion(), 0u)`. You need a separate `expected_from_name` with `spec_version=0` here, the same pattern used in the `ParseRealStateGetMetadataResponsePolkadot` test.

components/brave_wallet/browser/polkadot/polkadot_chain_metadata_prefs_unittest.cc
```diff
@@ -0,0 +1,119 @@
...
+  value.Set(kSs58Prefix, 42);
+  // Missing spec version should reject persisted value.
+  update->Set(mojom::kPolkadotMainnet, std::move(value));
+
+  PolkadotChainMetadataPrefs prefs = MakePrefs();
+  EXPECT_FALSE(prefs.GetChainMetadata(mojom::kPolkadotMainnet));
+}
+
+}  // namespace
+}  // namespace brave_wallet
```
netzenbot, line 119: nit: These pref key constants are duplicated from `polkadot_chain_metadata_prefs.cc`. Per [BS-048](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-048), shared constants should live in one place. Consider moving them to a shared header (e.g., a `polkadot_chain_metadata_prefs_internal.h` included by both) so a rename in the impl doesn't silently break the tests.

components/brave_wallet/browser/polkadot/polkadot_test_utils.cc
```diff
@@ -33,6 +54,23 @@ base::DictValue RequestBodyToJsonDict(const network::ResourceRequest& req) {
   return base::test::ParseJsonDict(sv);
 }
 
+std::string ReadMetadataFixtureJson(std::string_view file_name) {
+  return ReadMetadataFixtureJsonImpl(file_name);
+}
+
```
netzenbot, line 60: nit: Prefer `base::test::ParseJsonDict` here for consistency with the rest of this file — `RequestBodyToJsonDict` already uses it. See [CSA-051](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-051).

components/brave_wallet/browser/polkadot/polkadot_test_utils.cc
```diff
@@ -33,6 +54,23 @@ base::DictValue RequestBodyToJsonDict(const network::ResourceRequest& req) {
...
+std::vector<uint8_t> ReadMetadataFixture(std::string_view file_name) {
+  auto json =
+      base::JSONReader::ReadDict(ReadMetadataFixtureJsonImpl(file_name), 0);
```
netzenbot, line 63: nit: Define a named constant for the `"result"` key string rather than using an inline literal, e.g. `constexpr char kResult[] = "result";`. See [CSA-054](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-054).

components/brave_wallet/browser/polkadot/polkadot_chain_metadata_prefs_unittest.cc
cypt4: We don't want tests to silently pass after constant value changes. This could lead to database corruption.

components/brave_wallet/browser/polkadot/polkadot_test_utils.cc
cypt4: This method is used to return string value to rpc response

cmazakas (approved):

components/brave_wallet/browser/polkadot/polkadot_chain_metadata_prefs.cc
cmazakas: Ooh, I like that idea a lot. Let's migrate towards that, yeah.

components/brave_wallet/browser/polkadot/polkadot_chain_metadata_prefs.cc
```diff
@@ -0,0 +1,172 @@
...
+//     "transfer_keep_alive_call_index": int,    // u8
+//     "transfer_all_call_index": int,           // u8
+//     "ss58_prefix": int,                       // u16
+//     "spec_version": int                       // u32
+//   },
+//   ...
+// }
+
+template <class T>
+bool ReadUint(const base::DictValue& dict,
```
kdenhartog, line 46: nice idea for this. Might be worth making this a more generic checking pattern

components/brave_wallet/browser/polkadot/polkadot_metadata_provider.cc
```diff
@@ -0,0 +1,167 @@
...
+
+  if (!chain_metadata_prefs_->SetChainMetadata(chain_id, *parsed)) {
+    std::move(callback).Run(base::unexpected(WalletInternalErrorMessage()));
+    return;
+  }
+  SetCachedMetadata(chain_id, *parsed);
+  std::move(callback).Run(base::ok(std::move(*parsed)));
+}
+
+void PolkadotMetadataProvider::SetCachedMetadata(
```
kdenhartog, line 134: nit: Do we have any way to recover from a corrupted pref state here? For example if they get updated or changed we may want to re-fetch and set them.

kdenhartog (commented):
few nits, but nothing blocking from a sec perspective IMO.

components/brave_wallet/browser/polkadot/polkadot_metadata_provider.cc
cmazakas: We should be handling this case here: https://github.com/brave/brave-core/blob/555d30721fa125bbaf80aed43c5c7b278a8f8905/components/brave_wallet/browser/polkadot/polkadot_metadata_provider.cc#L56-L73

If restoring from the prefs fails here:
```cpp
  auto saved_metadata = chain_metadata_prefs_->GetChainMetadata(chain_id);
  if (!saved_metadata) {
    FetchAndParseChainMetadata(std::string(chain_id), std::move(callback));
    return;
  }
```
we re-fetch and parse again.

I wound up testing this functionality incidentally because I added a new asset_tx_payment field and the prefs turned out to be self-healing. I think we have a test that checks this explicitly, but I can't recall. If we don't, we definitely need one.

PR #35426, opened 2026-04-10
[Wallet] Refactor TxStorageDelegateImpl

### LINKED ISSUE #54435: Wallet transactions db should not be shared with incognito profile
### Description

In additional to in-memory tx list we also use database to persist and load it on browser restart. For incognito we just don't need to persist tx list.

supermassive:
Resolves https://github.com/brave/brave-browser/issues/54435

- Refactored `TxStorageDelegateImpl` to `TxStorage`
- ~For incognito there is now in-memory `TxStorage` is created (see brave_wallet_service.cc) changes.~


browser/brave_wallet/brave_wallet_service_browsertest.cc
```diff
@@ -160,4 +178,73 @@ IN_PROC_BROWSER_TEST_F(BraveWalletServiceTest, IsPrivateWindow) {
...
+  EXPECT_EQ(1u,
+            wallet_service()->tx_service()->GetPendingTransactionsCountSync());
+
+  // No transactions in incognito profile.
+  wallet_service()->SetPrivateWindowsEnabled(true);
+  EXPECT_EQ(0u, incognito_wallet_service()
+                    ->tx_service()
+                    ->GetPendingTransactionsCountSync());
+
+  // Add 2 transactions in incognito profile.
```
cypt4, line 213: Need a case that 2 incognito profiles don't intersect

cypt4 (approved):

netzenbot (approved):

components/brave_wallet/browser/tx_storage.cc
```diff
@@ -0,0 +1,147 @@
...
+
+const base::DictValue& TxStorage::GetTxs() const {
+  return txs_;
+}
+
+void TxStorage::Initialize() {
+  if (!store_) {
+    base::SequencedTaskRunner::GetCurrentDefault()->PostTask(
+        FROM_HERE, base::BindOnce(&TxStorage::OnTxsInitialRead,
+                                  weak_factory_.GetWeakPtr(), std::nullopt));
```
cmazakas, line 73: Because it seems like `TxStorage` is created on the main thread that's already creating all of our other services in the BraveWalletService, is there a reason we need to use PostTask() here?

supermassive: This is done so `TxStorage::TxStorage` ctor and `TxStorage::OnTxsInitialRead` don't appear on the same callstack.  It is not something completely wrong, but might become tricky to support in some tests.  Let's just make it so `TxStorage::OnTxsInitialRead` always comes from task runner.

browser/brave_wallet/brave_wallet_service_browsertest.cc
```diff
@@ -160,4 +178,80 @@ IN_PROC_BROWSER_TEST_F(BraveWalletServiceTest, IsPrivateWindow) {
...
+      tx_add_future.GetCallback());
+  std::tie(success, tx_meta_id, error_message) = tx_add_future.Take();
+  EXPECT_TRUE(success);
+
+  // 2 transactions in incognito profile.
+  EXPECT_EQ(2u, incognito_wallet_service()
+                    ->tx_service()
+                    ->GetPendingTransactionsCountSync());
+  // Still 1 transaction in regular profile.
+  EXPECT_EQ(1u,
```
kdenhartog, line 242: Is this going to mess up our nonce counter on EVM chains? nonce reuse on EVM will be rejected by the chain and could potentially put the regular profile in a bad state.

kdenhartog (commented):
I was talking with @jamesmudgett about this on Friday and I'm thinking that we're probably not gaining much here from a privacy angle since all the transactions are still published on chain. It probably also hurts our ability to keep the local state in sync with the onchain sync as well.

There are cases like bookmarks where state is persisted locally beyond the incognito profile. There's other cases like DNS settings where it kicks back to the regular profile for updates.

So, I think the primary goal here should be focused more on profile separation than on privacy. The tradeoffs we're going to face may actually introduce weird unintended side effects.

browser/brave_wallet/brave_wallet_service_browsertest.cc
supermassive: Fresh nonce is fetched from RPC after transaction is created. We don't store last seen value of nonce anywhere(for the purpose of computing next one).

It is possible (but not practically easy) to submit two transactions with same nonce and one of them gets rejected, and this could happen with incognito vs regular profile as also with two different profiles or two different wallet apps.

kdenhartog: Theoretically could still happen in other cases (user submitting with multiple wallets).

Ok this isn't a blocker

PR #35431, opened 2026-04-10
[Polkadot] [Wallet] add transfer_all functionality

### LINKED ISSUE #54401: Polkadot should support transfer_all
When a user clicks MAX in the Send panel, we should support this transaction natively with the `transfer_all` extrinsic from the Balances pallet: https://github.com/paritytech/polkadot-sdk/blob/de4709b0914c8290f0092c919d841b79249db836/substrate/frame/balances/src/lib.rs#L694-L730

cmazakas:
Resolves https://github.com/brave/brave-browser/issues/54401

Update the code to support the `transfer_all` Balances call. This is triggered when the user clicks MAX in the Send panel, which sets the `sendingMaxAmount` parameter to `true`.

### DIFF
--- a/components/brave_wallet/browser/polkadot/polkadot_extrinsic_unittest.cc
+++ b/components/brave_wallet/browser/polkadot/polkadot_extrinsic_unittest.cc
@@ -39,6 +39,7 @@ PolkadotChainMetadata MakeWestendMetadata() {
       /*transaction_payment_pallet_index=*/0x1a,
       /*transfer_allow_death_call_index=*/0,
       /*transfer_keep_alive_call_index=*/3,
+      /*transfer_all_call_index=*/4,
       /*ss58_prefix=*/42, kSpecVersion);
 }
 
@@ -48,6 +49,7 @@ PolkadotChainMetadata MakePolkadotMetadata() {
       /*transaction_payment_pallet_index=*/0x20,
       /*transfer_allow_death_call_index=*/0,
       /*transfer_keep_alive_call_index=*/3,
+      /*transfer_all_call_index=*/4,
       /*ss58_prefix=*/0, kSpecVersion);
 }
 
@@ -57,6 +59,7 @@ PolkadotChainMetadata MakeWestendAssetHubMetadata() {
       /*transaction_payment_pallet_index=*/0x0b,
       /*transfer_allow_death_call_index=*/0,
       /*transfer_keep_alive_call_index=*/3,
+      /*transfer_all_call_index=*/4,
       /*ss58_prefix=*/42, kSpecVersion);
 }
 
@@ -66,6 +69,7 @@ PolkadotChainMetadata MakePolkadotAssetHubMetadata() {
       /*transaction_payment_pallet_index=*/0x0b,
       /*transfer_allow_death_call_index=*/0,
       /*transfer_keep_alive_call_index=*/3,
+      /*transfer_all_call_index=*/4,
       /*ss58_prefix=*/0, kSpecVersion);
 }
 
@@ -493,9 +497,9 @@ TEST(PolkadotExtrinsics, SignaturePayload) {
       block_hash));
 
   auto encoded = generate_extrinsic_signature_payload(
-      *testnet_metadata, sender_nonce, send_amount_bytes, recipient,
-      spec_version, transaction_version, block_number, genesis_hash,
-      block_hash);
+      *testnet_metadata, sender_nonce, send_amount_bytes,
+      /*transfer_all=*/false, recipient, spec_version, transaction_version,
+      block_number, genesis_hash, block_hash);
 
   constexpr const char kExpected[] =
       R"(0403008eaf04151687736326c9fea17e25fc5287613693c912909cb226aa4794f26a484913750108000061900f001b000000e143f23803ac50e8f6f8e62695d1ce9e4e1d68aa36c1cd2cfd15340213f3423ebdcb3205ee391126e758556ffef5bb0d5a5fd1bbd996c671a079d5b02a67191300)";
@@ -503,7 +507,7 @@ TEST(PolkadotExtrinsics, SignaturePayload) {
   EXPECT_EQ(base::HexEncodeLower(encoded), kExpected);
 }
 
-TEST(PolkadotExtrinsics, SignedExtrinsic) {
+TEST(PolkadotExtrinsics, SignedExtrinsic_TransferKeepAlive) {
   auto testnet_metadata = MakeWestendMetadata();
 
   std::array<uint8_t, 32> recipient = {};
@@ -521,8 +525,6 @@ TEST(PolkadotExtrinsics, SignedExtrinsic) {
   const char genesis_hash_encoded[] =
       R"(0xe143f23803ac50e8f6f8e62695d1ce9e4e1d68aa36c1cd2cfd15340213f3423e)";
 
-  PolkadotUnsignedTransfer transfer(recipient, send_amount);
-
   uint32_t account_index = 0;
 
   auto keypair = HDKeySr25519::GenerateFromSeed(kSchnorrkelSeed);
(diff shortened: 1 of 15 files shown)


components/brave_wallet/browser/polkadot/polkadot_signed_transfer_task.h
```diff
@@ -168,6 +169,7 @@ struct PolkadotSignedTransferTask {
   mojom::AccountIdPtr sender_account_id_;
   std::string chain_id_;
   bool use_dummy_signature_;
```
netzenbot, line 174: nit: `bool transfer_all_` should have a default initializer (`= false`) per [CSA-063](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-063). POD members without default values lead to undefined behavior if ever read before being set.

components/brave_wallet/browser/polkadot/polkadot_extrinsic_unittest.cc
```diff
@@ -533,6 +535,8 @@ TEST(PolkadotExtrinsics, SignedExtrinsic) {
 
   keypair.UseMockRngForTesting();
```
netzenbot, line 537: nit: `const auto transfer_all = false;` (and the corresponding `= true;` in the TransferAll test) — `bool` is short and descriptive enough to spell out explicitly. Per the style guide, avoid `auto` merely to dodge writing a type name when the explicit type adds readability. ([Use `auto` Judiciously — Prefer Explicit Types When Short and Descriptive](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-053))

components/brave_wallet/browser/polkadot/polkadot_wallet_service.h
```diff
@@ -80,13 +80,15 @@ class PolkadotWalletService : public mojom::PolkadotWalletService,
   void GenerateSignedTransferExtrinsic(
       std::string chain_id,
       mojom::AccountIdPtr account_id,
+      bool transfer_all,
```
netzenbot: nit: The doc comment above still says "representing a transfer_allow_death call" but this function now generates either a `transfer_keep_alive` or `transfer_all` call depending on `transfer_all`. Please update the comment to reflect the new behavior. ([Update Documentation Alongside Code Changes](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-001))

components/brave_wallet/browser/polkadot/polkadot_extrinsic_unittest.cc
```diff
@@ -626,6 +630,86 @@ TEST(PolkadotExtrinsics, SignedExtrinsic) {
...
+
+  auto keypair = HDKeySr25519::GenerateFromSeed(kSchnorrkelSeed);
+  keypair = keypair.DeriveHard(base::byte_span_from_cstring("\x1cwestend"));
+  keypair = keypair.DeriveHard(base::byte_span_from_ref(account_index));
+  EXPECT_EQ(base::HexEncodeLower(keypair.GetPublicKey()),
+            "d4f9c4dfa3e6ff57b4e1fdea8699e57b0210cf04afe0281acba187d7d1b49274");
+
+  keypair.UseMockRngForTesting();
+
+  const auto transfer_all = true;
```
netzenbot: nit: `const auto transfer_all = true;` — same issue as in the `TransferKeepAlive` test: `bool` is short and descriptive enough to spell out explicitly. Prefer `const bool transfer_all = true;`. ([CS-053](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-053))

antonok-edm (approved):
rust code

components/brave_wallet/browser/polkadot/polkadot_signed_transfer_task.h
```diff
@@ -77,6 +77,7 @@ struct PolkadotSignedTransferTask {
       mojom::AccountIdPtr sender_account_id,
       std::string chain_id,
       bool use_dummy_signature,
+      bool transfer_all,
```
cypt4: Consider using variant between an amount and transfer_all token

components/brave_wallet/browser/polkadot/polkadot_wallet_service_unittest.cc
```diff
@@ -1946,6 +1955,50 @@ TEST_F(PolkadotWalletServiceUnitTest, SignAndSendTransaction) {
...
+
+  std::array<uint8_t, kPolkadotSubstrateAccountIdSize> recipient_pubkey = {};
+  EXPECT_TRUE(base::HexStringToSpan(kBob, recipient_pubkey));
+
+  polkadot_wallet_service->SignAndSendTransaction(
+      chain_id, polkadot_mainnet_account_->account_id.Clone(),
+      /*transfer_all=*/true, 4321, recipient_pubkey, future.GetCallback());
+
+  auto tx_hash = future.Take();
+  ASSERT_TRUE(tx_hash.has_value());
```
cypt4, line 2004: I believe we also need to verify data that is sent to server

components/brave_wallet/browser/polkadot/polkadot_tx_manager_unittest.cc
```diff
@@ -448,6 +448,95 @@ TEST_F(PolkadotTxManagerUnitTest, ApproveTransaction) {
...
+
+  const auto& txs = tx_service_->GetDelegateForTesting()->GetTxs();
+  const auto* tx = txs.FindDict(tx_meta_id);
+  ASSERT_TRUE(tx);
+
+  auto polkadot_tx = GetPolkadotTxManager()->GetPolkadotTx(tx_meta_id);
+  ASSERT_TRUE(polkadot_tx);
+
+  EXPECT_EQ(polkadot_tx->status(), mojom::TransactionStatus::Submitted);
+  EXPECT_EQ(polkadot_tx->tx()->recipient().ToString().value(), kBob);
```
cypt4, line 501: Can we add == operator for PolkadotTransaction and PolkadotTxMeta to compare? 
If new fields are added it is hard to update all corresponding tests

components/brave_wallet/browser/polkadot/polkadot_wallet_service_unittest.cc
cmazakas: Done! Mock RPC layer now verifies the expected extrinsic.

components/brave_wallet/browser/polkadot/polkadot_signed_transfer_task.h
cmazakas: This is done! It actually wasn't as big a change as I thought it would've been. Note, I didn't update the Rust side because I'm not sure you can even send a variant from C++ -> Rust and also, I'd prefer to not have to get the rust-reviewers involved on this PR unless we really, really need to.

components/brave_wallet/browser/polkadot/polkadot_tx_manager_unittest.cc
cmazakas: In principle I agree. But one problem here is that a lot of these structs don't support being constructed with all parameters at once. Instead, it seems like all the structs in Wallet are DefaultConstructible and then we use the myriad setters.

I'm hesitant to add `operator==` for the sake of tests in this particular PR because it means I'd have to add a factory function (which is also susceptible to missing newly added data members) or I'd have to make a larger change which is add a non-defaulted constructor that takes _all_ the parameters.

Maybe we can revisit this in a CodeHealth PR.

components/brave_wallet/browser/polkadot/polkadot_signed_transfer_task.h
cmazakas: No longer applies because we use a `variant` now.

components/brave_wallet/browser/polkadot/polkadot_extrinsic_unittest.cc
cmazakas: This is done.

components/brave_wallet/browser/polkadot/polkadot_wallet_service.h
cmazakas: This is done.

components/brave_wallet/browser/polkadot/polkadot_extrinsic_unittest.cc
cmazakas: This is done.

netzenbot (approved):

cypt4 (approved):

PR #35435, opened 2026-04-10
Add AIPageContentAgent fallback for page content extraction

### LINKED ISSUE #54407: [AI Chat] Consider using AIPageContentAgent as page content fetching fallback
During the integration of history embeddings, `PageEmbeddingsService` extracting passages from websites to calculate embeddings seems like something we can borrow to enhance our content extraction process

This improves cases like overall GitHub pages including https://gist.github.com/karpathy/442a6bf555914893e9891c11519de94f that we didn't do special handling on it

darkdh:
When the existing page_content_fetcher (AX tree-based) fails to extract any content from a loaded page, fall back to blink's AIPageContentAgent which walks the layout tree directly and may succeed where AX tree extraction fails.

The fallback binds the `blink::mojom::AIPageContentAgent` mojo interface directly on the main frame, calls GetAIPageContent with kDefault mode, and extracts plain text from the returned node tree.

The idea came from https://chromium-review.googlesource.com/c/chromium/src/+/7549570 which will eventually use `AIPageContentAgent` to extract passages for generating embeddings

Resolves https://github.com/brave/brave-browser/issues/54407

# Test plan
1. Navigate to https://gist.github.com/karpathy/442a6bf555914893e9891c11519de94f
2. Ask Leo to summarize it
3. We should get response without capturing screenshots

Screenshot 2026-04-10 at 2 36 45 PM

### DIFF
--- a/components/ai_chat/content/browser/BUILD.gn
+++ b/components/ai_chat/content/browser/BUILD.gn
@@ -35,6 +35,8 @@ static_library("browser") {
     ]
   }
 
+  public_deps = [ "//mojo/public/cpp/bindings" ]
+
   deps = [
     "//base",
     "//brave/components/ai_chat/core/browser",
@@ -96,7 +98,6 @@ source_set("unit_tests") {
     "//components/pdf/common:constants",
     "//content/public/browser",
     "//content/test:test_support",
-    "//mojo/public/cpp/bindings",
     "//net:test_support",
     "//pdf:buildflags",
     "//services/data_decoder/public/cpp",
--- a/components/ai_chat/content/browser/associated_web_contents_content_unittest.cc
+++ b/components/ai_chat/content/browser/associated_web_contents_content_unittest.cc
@@ -262,8 +262,6 @@ TEST_P(AssociatedWebContentsContentUnitTest, GetPageContent_VideoContent) {
 
 TEST_P(AssociatedWebContentsContentUnitTest,
        GetPageContent_PrintPreviewTriggeringURL) {
-  base::MockCallback<AssociatedWebContentsContent::FetchPageContentCallback>
-      callback;
   // A url that triggers print preview extraction - should return empty content
   // to allow autoscreenshots mechanism to handle server-side OCR
   for (const auto& host : kPrintPreviewRetrievalHosts) {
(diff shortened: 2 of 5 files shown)


petemill:
This conflicts with https://github.com/brave/brave-core/pull/35352 which replaces the current content extraction strategy with putting the whole structure from `AIPageContentAgent`, same as the Content Agent conversations do, and never using only the text content of the page's nodes.

That PR will merge that strategy behind a feature flag whilst we optimize the amount of data, and compare its effect to the current strategy, so if we still want to merge this to the current strategy, even though we might quickly remove it, it could be ok. cc @fallaciousreasoning

components/ai_chat/content/browser/BUILD.gn
```diff
@@ -56,13 +56,15 @@ static_library("browser") {
     "//components/strings:components_strings_grit",
     "//components/user_prefs",
     "//content/public/browser",
+    "//mojo/public/cpp/bindings",
```
netzenbot: `//mojo/public/cpp/bindings` and `//third_party/blink/public/mojom:mojom_platform` are used in `associated_web_contents_content.h` (the `mojo::Remote<blink::mojom::AIPageContentAgent>` member declaration), so they must be in `public_deps`, not `deps`. Consumers of this target that include the header need transitive access to those includes. https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-024

components/ai_chat/content/browser/associated_web_contents_content_unittest.cc
```diff
@@ -328,21 +405,24 @@ TEST_P(AssociatedWebContentsContentUnitTest,
     testing::Mock::VerifyAndClearExpectations(&print_preview_extractor_);
```
netzenbot, line 409: The `FetchPageContentFromAIPageContentAgent` method guards against a null or non-live `RenderFrameHost` and falls back to calling `callback.Run("", false, "")`, but none of the new tests cover this code path. Per [TI-016](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-016), invalid/bad input scenarios need test coverage. Consider adding a test that detaches the main frame before calling `GetPageContent` to exercise the early-return guard.

components/ai_chat/content/browser/associated_web_contents_content_unittest.cc
```diff
@@ -262,23 +334,26 @@ TEST_P(AssociatedWebContentsContentUnitTest, GetPageContent_VideoContent) {
```
netzenbot, line 344: nit: This three-line assertion block (`auto [content, is_video, invalidation_token] = future.Get()` + 3 EXPECTs) is repeated verbatim in 5 places in this diff. Consider extracting it into a helper like `ExpectPageContent(future, expected_content)` per [TI-024](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-024).

components/ai_chat/content/browser/BUILD.gn
darkdh: fixed and squashed

components/ai_chat/content/browser/associated_web_contents_content_unittest.cc
darkdh: fixed and squashed

components/ai_chat/content/browser/associated_web_contents_content_unittest.cc
darkdh: fixed and squashed

netzenbot (approved):

darkdh:
> This conflicts with #35352 which replaces the current content extraction strategy with putting the whole structure from `AIPageContentAgent`, same as the Content Agent conversations do, and never using only the text content of the page's nodes.
> 
> That PR will merge that strategy behind a feature flag whilst we optimize the amount of data, and compare its effect to the current strategy, so if we still want to merge this to the current strategy, even though we might quickly remove it, it could be ok. cc @fallaciousreasoning

I think it makes sense to have this fallback in our extraction pipeline and then remove it along with current `mojom::PageContentExtractor` when https://github.com/brave/brave-core/pull/35352 is enable by default.
It serves its purpose while we trial the flag so users won't suffer from not getting context (which often fallback to screenshots)

bbondy (approved):
LGTM. You could add a TODO for removing when brave/brave-core#35352 is enabled by default.  Or mention it in that issue.

darkdh:
> LGTM. You could add a TODO for removing when #35352 is enabled by default. Or mention it in that issue.

comment added in https://github.com/brave/brave-core/pull/35435/files#diff-36e865883e0aaef375834ab6b7f68a136b9b26fae48f45af23b7a37dc678c6a7R156-R157

PR #35438, opened 2026-04-10
[Wallet] update GenerateReceiveAddress for Polkadot

cmazakas:
This update prevents the `BraveWalletService::GenerateReceiveAddress` from hitting a `NOTREACHED()` when it's called for Polkadot accounts.

I didn't realize until I had started writing the tests that we actually already populate the `address` field of the `AccountInfo` structure correctly. I'm not sure exactly how this should be implemented, but we do need to address this because the `NOTREACHED()` will crash the browser during regular testing.

I think how we handle this will tie into how we eventually handle supporting multiple parachains. Because each parachain can have governance of the assets on it (including DOT), we need to somehow tie that directly to the AccountInfo or AccountId structure. To this end, I think maybe relying on `address` for `AccountInfo` is okay as we can add new KeyringIds for each parachain we want to support.

I'm not sure what this kind of change would look like if we wanted to support arbitrary parachains, however. But even in that instance, we'd still need to store an identifier for the parachain regardless as those are usually hosted using their own URL.

### DIFF
--- a/browser/brave_wallet/brave_wallet_service_unittest.cc
+++ b/browser/brave_wallet/brave_wallet_service_unittest.cc
@@ -308,8 +308,11 @@ class BraveWalletServiceUnitTest : public testing::Test {
  protected:
   void SetUp() override {
     scoped_feature_list_.InitWithFeatures(
-        {features::kBraveWalletBitcoinFeature,
-         features::kBraveWalletCardanoFeature},
+        {
+            features::kBraveWalletBitcoinFeature,
+            features::kBraveWalletCardanoFeature,
+            features::kBraveWalletPolkadotFeature,
+        },
         {});
 
     brave_wallet::RegisterLocalStatePrefs(local_state_.registry());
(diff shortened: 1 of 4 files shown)


browser/brave_wallet/brave_wallet_service_unittest.cc
```diff
@@ -2889,14 +2892,16 @@ TEST_F(BraveWalletServiceUnitTest, ConvertFEVMToFVMAddress) {
   }
 }
 
-TEST_F(BraveWalletServiceUnitTest, GenerateReceiveAddress_EthFilSol) {
+TEST_F(BraveWalletServiceUnitTest, GenerateReceiveAddress_EthFilSolDot) {
```
netzenbot, line 2895: The new `!polkadot_wallet_service_` guard in `GenerateReceiveAddress` is only tested via the positive path (feature enabled, service present). Per [TI-026](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-026), consider adding a test case where the Polkadot feature is disabled (service is null) and verifying that `GenerateReceiveAddress` returns an error.

supermassive (approved):

cypt4 (approved):

PR #35444, opened 2026-04-12
Fix crash in BraveTabStrip::UpdateOrientation() with upstream vertical tabs

### LINKED ISSUE #54419: [macOS] Crash after enabling vertical tabs
### Description

The Brave browser on macOS (Sequoia 15.7.4) crashes after enabling vertical tabs, and the Brave browser cannot be opened since.

### Steps to reproduce

1. Download the stable version of Brave browser (v1.89.132)
2.Enable vertical tabs in brave://flags/
3. In the settings, set the appearance to vertical tabs, then close the Brave browser. After that, you will not be able to open the Brave browser.

### Actual result

Brave browser cannot be opened.

### Expected result

The Brave browser can be opened and used normally.

### Reproduces how often

Easily reproduced

v1.89.132

### Channel information

- [x] release (stable)
- [ ] beta
- [ ] nightly

### Reproducibility

- [x] in the latest version of Chrome

### Miscellaneous information

_No response_

simonhong:
When Chromium's vertical tabs feature is enabled, TabStrip::Initialize() is never called
so tab_container_ remains null. This caused a crash in SetAvailableWidthCallback() during startup.
Early return when Brave vertical tabs are not supported.

Resolves https://github.com/brave/brave-browser/issues/54419

### DIFF
--- a/browser/ui/views/frame/vertical_tabs/vertical_tab_strip_browsertest.cc
+++ b/browser/ui/views/frame/vertical_tabs/vertical_tab_strip_browsertest.cc
@@ -2312,3 +2312,27 @@ INSTANTIATE_TEST_SUITE_P(,
                          [](const testing::TestParamInfo<bool>& info) {
                            return info.param ? "Embedded" : "NonEmbedded";
                          });
+
+// Regression test: When Chromium's upstream vertical tabs feature is active,
+// TabStrip::Initialize() is never called so tab_container_ remains null.
+// BraveTabStrip::UpdateOrientation() was crashing by accessing the null
+// tab_container_ via SetAvailableWidthCallback() during startup.
+class UpstreamVerticalTabsCrashTest : public InProcessBrowserTest {
+ public:
+  UpstreamVerticalTabsCrashTest() {
+    scoped_feature_list_.InitAndEnableFeature(tabs::kVerticalTabs);
+  }
+  ~UpstreamVerticalTabsCrashTest() override = default;
+
+ private:
+  base::test::ScopedFeatureList scoped_feature_list_;
+};
+
+IN_PROC_BROWSER_TEST_F(UpstreamVerticalTabsCrashTest, NoCrashOnStartup) {
+  // Simulate the user choosing "Side" in Tab strip position settings,
+  // then opening a new window. Previously this crashed in
+  // BraveTabStrip::UpdateOrientation() because tab_container_ was null.
+  browser()->profile()->GetPrefs()->SetBoolean(prefs::kVerticalTabsEnabled,
+                                               true);
+  CreateBrowser(browser()->profile());
+}
--- a/browser/ui/views/frame/vertical_tabs/vertical_tab_strip_browsertest.cc
+++ b/browser/ui/views/frame/vertical_tabs/vertical_tab_strip_browsertest.cc
@@ -2330,9 +2330,12 @@ class UpstreamVerticalTabsCrashTest : public InProcessBrowserTest {
 
 IN_PROC_BROWSER_TEST_F(UpstreamVerticalTabsCrashTest, NoCrashOnStartup) {
   // Simulate the user choosing "Side" in Tab strip position settings,
-  // then opening a new window. Previously this crashed in
-  // BraveTabStrip::UpdateOrientation() because tab_container_ was null.
+  // then opening a new window. verifies no crash when kVerticalTabs is active
+  // and kVerticalTabsEnabled is set, since tab_container_ may be null in that
+  // configuration.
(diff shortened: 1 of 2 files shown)


browser/ui/views/frame/vertical_tabs/vertical_tab_strip_browsertest.cc
```diff
@@ -2312,3 +2312,27 @@ INSTANTIATE_TEST_SUITE_P(,
...
+  ~UpstreamVerticalTabsCrashTest() override = default;
+
+ private:
+  base::test::ScopedFeatureList scoped_feature_list_;
+};
+
+IN_PROC_BROWSER_TEST_F(UpstreamVerticalTabsCrashTest, NoCrashOnStartup) {
+  // Simulate the user choosing "Side" in Tab strip position settings,
+  // then opening a new window. Previously this crashed in
+  // BraveTabStrip::UpdateOrientation() because tab_container_ was null.
```
netzenbot: nit: The comments say "was crashing" and "Previously this crashed" — these reference historical behavior rather than what the code does now. Prefer something like "Verifies no crash when kVerticalTabs is active and kVerticalTabsEnabled is set, since tab_container_ may be null in that configuration." ([Comments Must Make Sense to Future Readers of the Codebase](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-019))

browser/ui/views/tabs/brave_tab_strip.cc
```diff
@@ -305,6 +305,13 @@ bool BraveTabStrip::ShouldShowPinnedTabsInGrid() const {
 }
 
 void BraveTabStrip::UpdateOrientation() {
+  // When Chromium's upstream vertical tabs feature is active,
```
netzenbot: nit: This is a workaround for upstream behavior, but there's no link to a Chromium bug. Per [CS-023](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-023), please file an upstream issue (or find an existing one) and add the link here so this guard can be removed once the upstream fix lands.

browser/ui/views/frame/vertical_tabs/vertical_tab_strip_browsertest.cc
```diff
@@ -2312,3 +2312,27 @@ INSTANTIATE_TEST_SUITE_P(,
...
+  browser()->profile()->GetPrefs()->SetBoolean(prefs::kVerticalTabsEnabled,
+                                               true);
+  CreateBrowser(browser()->profile());
+}
```
netzenbot, line 2341: nit: The test verifies no crash by surviving `CreateBrowser()`, but per TI-029, browser tests should verify at least one observable side effect. Consider asserting the returned `Browser*` is non-null or that the tab strip model has the expected tab count. ([Verify Observable Side Effects in Browser Tests](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-029))

browser/ui/views/tabs/brave_tab_strip.cc
simonhong: It's not an upstream bug. TabStrip::Initialize() is not used by vertical tab mode.
Updated that comment.

browser/ui/views/frame/vertical_tabs/vertical_tab_strip_browsertest.cc
simonhong: Fixed.

browser/ui/views/frame/vertical_tabs/vertical_tab_strip_browsertest.cc
simonhong: Updated comments.

browser/ui/views/tabs/brave_tab_strip.cc
```diff
@@ -305,6 +305,14 @@ bool BraveTabStrip::ShouldShowPinnedTabsInGrid() const {
...
+  // TabStrip::Initialize() is never called because it's only used by
+  // horizontal tab mode so tab_container_ remains null.
+  // This method must bail out early to avoid crashing when accessing it.
+  if (!tabs::utils::SupportsBraveVerticalTabs(GetBrowserWindowInterface())) {
+    return;
+  }
```
sangwoo108: If there's no call sites except AddedToWidget(), how about early return there and make this CHECK()?

sangwoo108 (approved):
LGTM

browser/ui/views/tabs/brave_tab_strip.cc
simonhong: Updated in that way. thanks!

netzenbot (approved):

PR #35450, opened 2026-04-13
[Containers] Show menu on partitioned storage page action view

### LINKED ISSUE #53350: [Partitioned Storage Tab] What should be done when user clicks the page action in location bar?
Currently nothing haapens when page action icon for containers in location bar is clicked.

Need discussion cc @rebron @aguscruiz

sangwoo108:
* Show container menu on partitioned storage page action view, when
  it's right clicked.

Resolves https://github.com/brave/brave-browser/issues/53350

### DIFF
--- a/browser/containers/containers_browsertest.cc
+++ b/browser/containers/containers_browsertest.cc
@@ -45,6 +45,13 @@
 #include "services/network/public/cpp/network_switches.h"
 #include "third_party/abseil-cpp/absl/strings/str_format.h"
 #include "third_party/skia/include/core/SkColor.h"
+#include "ui/actions/actions.h"
+#include "ui/events/base_event_utils.h"
+#include "ui/events/event_constants.h"
+#include "ui/events/keycodes/keyboard_codes.h"
+#include "ui/views/controls/button/button.h"
+#include "ui/views/controls/menu/menu_controller.h"
+#include "ui/views/test/button_test_api.h"
 #include "ui/views/test/views_test_utils.h"
 #include "ui/views/view.h"
 #include "url/gurl.h"
@@ -1253,6 +1260,57 @@ IN_PROC_BROWSER_TEST_F(ContainersBrowserTest,
   tab_strip_model->ActivateTabAt(1);
   RunScheduledLayouts();
   EXPECT_TRUE(partitioned_storage_view->GetVisible());
+}
+
+IN_PROC_BROWSER_TEST_F(
+    ContainersBrowserTest,
+    PartitionedStoragePageActionRightClickSetsActionItemShowingBubble) {
+  actions::ActionItem* const partitioned_storage_action =
+      actions::ActionManager::Get().FindAction(kActionShowPartitionedStorage);
+  ASSERT_NE(nullptr, partitioned_storage_action);
+
+  IconLabelBubbleView* const partitioned_storage_view =
+      browser()->GetBrowserView().toolbar_button_provider()->GetPageActionView(
+          kActionShowPartitionedStorage);
+  ASSERT_NE(nullptr, partitioned_storage_view);
+
+  const GURL url("https://a.test/simple.html");
+  ASSERT_TRUE(ui_test_utils::NavigateToURL(browser(), url));
+
+  auto container = containers::mojom::Container::New();
+  container->id = "test-container";
+  container->name = "Test Container";
+  container->icon = containers::mojom::Icon::kSocial;
+  container->background_color = SK_ColorYELLOW;
+
+  brave::OpenUrlInContainer(browser(), url, container);
+  ASSERT_TRUE(content::WaitForLoadStop(
+      browser()->tab_strip_model()->GetActiveWebContents()));
+  RunScheduledLayouts();
+
+  ASSERT_TRUE(partitioned_storage_view->GetVisible());
+  EXPECT_FALSE(partitioned_storage_action->GetIsShowingBubble());
+
+  const gfx::Point click_location =
+      partitioned_storage_view->GetLocalBounds().CenterPoint();
+  const ui::MouseEvent click_event(ui::EventType::kMousePressed, click_location,
+                                   click_location, ui::EventTimeForNow(),
+                                   ui::EF_RIGHT_MOUSE_BUTTON,
+                                   ui::EF_RIGHT_MOUSE_BUTTON);
+  views::test::ButtonTestApi(views::Button::AsButton(partitioned_storage_view))
+      .NotifyClick(click_event);
+
+  EXPECT_TRUE(base::test::RunUntil(
+      [&]() { return partitioned_storage_action->GetIsShowingBubble(); }));
+
+  if (views::MenuController* menu_controller =
+          views::MenuController::GetActiveInstance()) {
+    ui::KeyEvent escape_event(ui::EventType::kKeyPressed, ui::VKEY_ESCAPE,
+                              ui::EF_NONE);
+    menu_controller->OnWillDispatchKeyEvent(&escape_event);
+  }
+  EXPECT_TRUE(base::test::RunUntil(
+      [&]() { return !partitioned_storage_action->GetIsShowingBubble(); }));
 }
 
 IN_PROC_BROWSER_TEST_F(ContainersBrowserTest, GetStoragePartitionKeyToRestore) {
--- a/browser/containers/containers_browsertest.cc
+++ b/browser/containers/containers_browsertest.cc
@@ -1289,7 +1289,7 @@ IN_PROC_BROWSER_TEST_F(
   RunScheduledLayouts();
 
   ASSERT_TRUE(partitioned_storage_view->GetVisible());
-  EXPECT_FALSE(partitioned_storage_action->GetIsShowingBubble());
+  ASSERT_FALSE(partitioned_storage_action->GetIsShowingBubble());
 
   const gfx::Point click_location =
       partitioned_storage_view->GetLocalBounds().CenterPoint();
--- a/browser/ui/browser_commands.cc
+++ b/browser/ui/browser_commands.cc
@@ -21,7 +21,6 @@
 #include "base/i18n/time_formatting.h"
 #include "base/logging.h"
 #include "base/memory/scoped_refptr.h"
-#include "base/notimplemented.h"
 #include "base/notreached.h"
 #include "base/path_service.h"
 #include "base/strings/utf_string_conversions.h"
@@ -128,8 +127,11 @@
 #endif
 
 #if BUILDFLAG(ENABLE_CONTAINERS)
+#include "brave/browser/ui/tabs/public/brave_tab_features.h"
+#include "brave/browser/ui/views/page_action/partitioned_storage_page_action_controller.h"
 #include "brave/components/containers/content/browser/storage_partition_utils.h"
 #include "brave/components/containers/core/mojom/containers.mojom.h"
+#include "components/tabs/public/tab_interface.h"
 #endif
 
 using content::WebContents;
@@ -1221,9 +1223,27 @@ void OpenUrlWithoutContainer(BrowserWindowInterface* browser_window,
   Navigate(&params);
 }
 
-void OpenContainerMenuOnPageActionView(BrowserWindowInterface* browser_window) {
-  // TODO(https://github.com/brave/brave-browser/issues/53350)
-  NOTIMPLEMENTED();
+void OpenContainerMenuOnPageActionView(BrowserWindowInterface* browser_window,
+                                       actions::ActionItem* item) {
+  if (!browser_window) {
+    LOG(ERROR) << "Browser window is not valid";
+    return;
+  }
+
+  tabs::TabInterface* tab = browser_window->GetActiveTabInterface();
+  if (!tab) {
+    LOG(ERROR) << "Tab is not valid";
+    return;
+  }
+
+  tabs::BraveTabFeatures* brave_tab_features =
+      tabs::BraveTabFeatures::FromTabFeatures(tab->GetTabFeatures());
+  CHECK(brave_tab_features);
+
+  page_actions::PartitionedStoragePageActionController* const controller =
+      brave_tab_features->partitioned_storage_page_action_controller();
+  CHECK(controller);
+  controller->ExecuteAction(item);
 }
 #endif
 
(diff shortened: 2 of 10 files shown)


browser/ui/views/page_action/partitioned_storage_page_action_controller.cc
```diff
@@ -116,4 +142,74 @@ void PartitionedStoragePageActionController::UpdatePageAction() {
...
+          kActionShowPartitionedStorage);
+  CHECK_DEREF(anchor_view).SetHighlighted(false);
+
+  menu_runner_.reset();
+  menu_model_adapter_.reset();
+  containers_menu_model_.reset();
+  containers_menu_delegate_.reset();
+}
+
 }  // namespace page_actions
```
netzenbot, line 222: `tab_->GetBrowserWindowInterface()` can return null (e.g., when the tab is detached from its window), and `BrowserView::GetBrowserViewForBrowser(nullptr)` will crash on `browser->GetWindow()`. `ExecuteAction` guards both `bwi` and `browser_view` with null checks, but `OnPartitionedStorageMenuClosed` skips them entirely — add the same guards here.

browser/ui/config.gni
```diff
@@ -55,7 +55,10 @@ if (toolkit_views) {
   }
 
   if (enable_containers) {
-    brave_ui_allow_circular_includes_from += [ "//brave/browser/ui/containers" ]
+    brave_ui_allow_circular_includes_from += [
+      "//brave/browser/ui/containers",
+      "//brave/browser/ui/views/page_action",
```
netzenbot, line 60: Adding `//brave/browser/ui/views/page_action` to `brave_ui_allow_circular_includes_from` introduces a new circular dependency: `//brave/browser/ui` already depends on `//brave/browser/ui/views/page_action` (toolkit_views block), and now `page_action` headers include back into `//brave/browser/ui`. Per [ARCH-038](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-038), component dependencies must form a tree — no circular deps. Consider breaking the cycle with a delegate interface or by moving the shared type to a `common/` or `public/` target.

browser/containers/containers_browsertest.cc
```diff
@@ -1224,6 +1231,57 @@ IN_PROC_BROWSER_TEST_F(ContainersBrowserTest,
   EXPECT_TRUE(partitioned_storage_view->GetVisible());
```
netzenbot, line 1231: The test for this UI feature is added to `ContainersBrowserTest` which derives from `InProcessBrowserTest`. Per ARCH-062, UI CUJ tests should use `InteractiveBrowserTest` (e.g., `RunTestSequence(PressButton(...), WaitForShow(...))`) rather than raw event injection. Consider adding an `InteractiveBrowserTest`-based test for this right-click menu flow. ([Every UI Feature Needs a CUJ Test](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-062))

browser/ui/views/page_action/partitioned_storage_page_action_controller.cc
```diff
@@ -56,7 +70,19 @@ PartitionedStoragePageActionController::PartitionedStoragePageActionController(
 }
```
netzenbot, line 71: The destructor should call `weak_ptr_factory_.InvalidateWeakPtrs()` before resetting `menu_runner_` and the other members. The WeakPtrFactory is used to bind `OnPartitionedStorageMenuClosed` — without invalidation first, a pending close callback could fire on a partially-destroyed object. https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-memory.md#CSM-006

browser/ui/views/page_action/partitioned_storage_page_action_controller.cc
```diff
@@ -116,4 +142,74 @@ void PartitionedStoragePageActionController::UpdatePageAction() {
...
+  menu_runner_ = std::make_unique<views::MenuRunner>(
+      std::move(root),
+      views::MenuRunner::HAS_MNEMONICS | views::MenuRunner::CONTEXT_MENU);
+  menu_runner_->RunMenuAt(
+      anchor_view->GetWidget(), nullptr, anchor_view->GetAnchorBoundsInScreen(),
+      views::MenuAnchorPosition::kTopLeft, ui::mojom::MenuSourceType::kMouse);
+}
+
+void PartitionedStoragePageActionController::OnPartitionedStorageMenuClosed() {
+  CHECK_DEREF(action_item_for_menu_).SetIsShowingBubble(false);
```
netzenbot, line 200: `OnPartitionedStorageMenuClosed` dereferences `bwi` and `browser_view` without null checks, unlike `ExecuteAction` which guards both. If the tab is detached from its browser window before the menu closes, `BrowserView::GetBrowserViewForBrowser(bwi)` or `browser_view->toolbar_button_provider()` will crash. Add null guards consistent with `ExecuteAction`. See https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-memory.md#CSM-029

browser/ui/config.gni
sangwoo108: @goodov , Is this something we can address?

browser/containers/containers_browsertest.cc
sangwoo108: This is non-applicable. We don't have non-sharded interactive browser tests env and this makes it impossible to use mouse simulation - which is required to simulate the right click.

browser/ui/views/page_action/partitioned_storage_page_action_controller.cc
```diff
@@ -116,4 +143,79 @@ void PartitionedStoragePageActionController::UpdatePageAction() {
...
+  action_item_for_menu_ = nullptr;
+
+  auto* bwi = tab_->GetBrowserWindowInterface();
+  if (bwi) {
```
netzenbot, line 204: `BrowserView::GetBrowserViewForBrowser(bwi)` can still return null even when `bwi` is non-null (e.g., when `bwi->GetWindow()` or its native window is null during construction or tests — see the comment in the implementation). `ExecuteAction` handles this with `if (!browser_view) { return; }`, but `OnPartitionedStorageMenuClosed` uses `CHECK(browser_view)` which crashes in those cases. Replace `CHECK(browser_view)` with a null guard and early return, consistent with `ExecuteAction`. https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-memory.md#CSM-029 ([Don't Store Window-Scoped State on Tab-Scoped Objects](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-058))

browser/ui/views/page_action/partitioned_storage_page_action_controller.cc
```diff
@@ -116,4 +143,79 @@ void PartitionedStoragePageActionController::UpdatePageAction() {
...
+    // In case of the tab is detached from the browser, the browser window might
+    // be null.
+    auto* browser_view = BrowserView::GetBrowserViewForBrowser(bwi);
+    CHECK(browser_view);
+    auto* anchor_view =
+        browser_view->toolbar_button_provider()->GetPageActionView(
```
netzenbot: `BrowserView::GetBrowserViewForBrowser(bwi)` can return null even when `bwi` is non-null — the upstream implementation explicitly guards `!browser->GetWindow()` and returns `nullptr`. Using `CHECK(browser_view)` here will crash in those cases. Use a null guard consistent with `ExecuteAction` instead. https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-027

browser/containers/containers_browsertest.cc
```diff
@@ -1224,6 +1231,57 @@ IN_PROC_BROWSER_TEST_F(ContainersBrowserTest,
...
+
+  const GURL url("https://a.test/simple.html");
+  ASSERT_TRUE(ui_test_utils::NavigateToURL(browser(), url));
+
+  auto container = containers::mojom::Container::New();
+  container->id = "test-container";
+  container->name = "Test Container";
+  container->icon = containers::mojom::Icon::kSocial;
+  container->background_color = SK_ColorYELLOW;
+
```
netzenbot, line 1254: `EXPECT_FALSE` here checks a precondition — if the bubble is unexpectedly already showing, the subsequent `RunUntil([&]() { return GetIsShowingBubble(); })` passes immediately, giving a false positive. Use `ASSERT_FALSE` to abort the test when the precondition isn't met. ([Use ASSERT for Preconditions That Subsequent Code Depends On](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-034))

sangwoo108:
Hi @fallaciousreasoning @goodov , friendly reminder :)

browser/ui/browser_commands.h
```diff
@@ -21,6 +21,10 @@
 
 #if BUILDFLAG(ENABLE_CONTAINERS)
 #include "brave/components/containers/core/mojom/containers.mojom-forward.h"
+
+namespace actions {
```
netzenbot, line 22: nit: The `namespace actions { class ActionItem; }` forward declaration doesn't need to be inside `#if BUILDFLAG(ENABLE_CONTAINERS)`. Per [BS-051](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-051), forward declarations cost nothing at compile time when unused — move it outside the guard and keep only the `#include` and actual usages inside.

browser/ui/views/page_action/partitioned_storage_page_action_controller.h
```diff
@@ -29,14 +50,27 @@ class PartitionedStoragePageActionController {
...
+  std::unique_ptr<containers::ContainersMenuModel> containers_menu_model_;
+  std::unique_ptr<views::MenuModelAdapter> menu_model_adapter_;
+  std::unique_ptr<views::MenuRunner> menu_runner_;
+  raw_ptr<actions::ActionItem> action_item_for_menu_ = nullptr;
+
+  base::WeakPtrFactory<PartitionedStoragePageActionController>
+      weak_ptr_factory_{this};
 };
 
 }  // namespace page_actions
```
netzenbot, line 76: nit: `action_item_for_menu_` is a non-owning pointer (the `ActionItem` is owned by `actions::ActionManager`). Per Chromium convention, add a `// not owned` comment. https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-memory.md#CSM-002

browser/ui/views/page_action/partitioned_storage_page_action_controller.cc
```diff
@@ -116,4 +143,80 @@ void PartitionedStoragePageActionController::UpdatePageAction() {
       kActionShowPartitionedStorage, ui::EF_RIGHT_MOUSE_BUTTON);
```
netzenbot: nit: Per [CS-017](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-017), function definitions in `.cc` should follow the same order as their declarations in `.h`. `ExecuteAction` is declared as a public method before the private `UpdatePageAction` in the header, so it should appear before `UpdatePageAction` in the `.cc` as well.

browser/ui/browser_commands.cc
```diff
@@ -1221,9 +1223,27 @@ void OpenUrlWithoutContainer(BrowserWindowInterface* browser_window,
   Navigate(&params);
 }
 
-void OpenContainerMenuOnPageActionView(BrowserWindowInterface* browser_window) {
-  // TODO(https://github.com/brave/brave-browser/issues/53350)
-  NOTIMPLEMENTED();
+void OpenContainerMenuOnPageActionView(BrowserWindowInterface* browser_window,
+                                       actions::ActionItem* item) {
+  if (!browser_window) {
+    LOG(ERROR) << "Browser window is not valid";
```
netzenbot: nit: Per CS-067, `LOG(ERROR)` should not be left in production code. These null guards are defensive and the error strings add binary bloat in release builds. Use `DVLOG(1)` instead, or just `return;` with no logging if the condition is truly unexpected. https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-067

browser/ui/views/page_action/partitioned_storage_page_action_controller.h
sangwoo108: This is incorrect. The BP will be removed

browser/ui/views/page_action/partitioned_storage_page_action_controller.h
```diff
@@ -29,14 +50,27 @@ class PartitionedStoragePageActionController {
...
   base::CallbackListSubscription did_activate_subscription_;
   base::CallbackListSubscription did_become_visible_subscription_;
   base::CallbackListSubscription will_discard_contents_subscription_;
+
+  std::unique_ptr<brave::ContainersTabMenuModelDelegate>
+      containers_menu_delegate_;
+  std::unique_ptr<containers::ContainersMenuModel> containers_menu_model_;
+  std::unique_ptr<views::MenuModelAdapter> menu_model_adapter_;
+  std::unique_ptr<views::MenuRunner> menu_runner_;
+  raw_ptr<actions::ActionItem> action_item_for_menu_ = nullptr;  // not owned
```
fallaciousreasoning: I think you can remove the `not owned` now, after your discussion on Slack 😄

fallaciousreasoning (approved):
lgtm

browser/ui/views/page_action/partitioned_storage_page_action_controller.h
sangwoo108: Oops, thanks :)

browser/ui/config.gni
goodov: tbh I don't quite understand why this reverse dependency is required, but splitting into public/impl seems reasonable as it's already done upstream in `chrome/browser/ui/views/page_action/BUILD.gn`.

sangwoo108: partitioned_storage_page_action_controller is including `browser_view.h` . Um, let me check if we can inject what we need from outside.

sangwoo108: Okay, injection worked!

sangwoo108:
@goodov Do you have any other comments?

netzenbot (approved):

goodov (approved):

PR #35451, opened 2026-04-13
Improved sidebar/panel bounds calc for sidebar v2

### LINKED ISSUE #54275: Update browser layout - locates panel first and sidebar UI
Sub class of https://github.com/brave/brave-browser/issues/51462

With https://github.com/brave/brave-core/pull/35285, side panel is buildable and opened but its position is wrong.
In LTR, put panel and then sidebar UI.

simonhong:
Resolves https://github.com/brave/brave-browser/issues/54275

image

**Root cause**

`CalculateSideBarLayout` used `contents_layout->bounds` (which in V2 is
already reduced by the upstream side panel) to position the sidebar
control, placing it between contents and panel instead of outside the
panel.  The desired order is:

  LTR (sidebar right): [vtab_left] [contents] [panel] [sidebar] [vtab_right]
  LTR (sidebar left):  [vtab_left] [sidebar] [panel] [contents] [vtab_right]

When the vertical tab is on the same side as the sidebar it must be the
outermost element.

**Layout fix (CalculateSideBarLayout)**

- Compute the outer edge from `browser_view->GetLocalBounds()` minus any
  vertical-tab width on the same side, instead of from `contents_bounds`.
  This gives the correct outermost anchor regardless of V1/V2.
- Introduce `adjust_panel` lambda (V2 only; no-op in V1 because the
  upstream panel is re-parented into `sidebar_container_view_` and is
  therefore absent from the top-level `ProposedLayout`) that shifts
  `toolbar_height_side_panel` / `contents_height_side_panel` inward by
  shifting `panel.x` by `±sidebar_width`.  Shifting by the sidebar width
  preserves the upstream slide animation offset rather than snapping to
  the final position.
- The fix is unconditional (no `#if BUILDFLAG(ENABLE_SIDEBAR_V2)`);
  the `adjust_panel` path is structurally a no-op in V1.

**Extracted static helpers (for testability)**

`ComputeSidebarBounds` and `ComputeAdjustedPanelBounds` are extracted as
private static methods and exposed for testing via a friend class.  They
contain the pure geometry arithmetic so it can be unit-tested without any
view or browser infrastructure.

**Unit tests (brave_browser_view_tabbed_layout_impl_unittest.cc)**

- `ComputeSidebarBoundsOnRight/Left` – sidebar placed at outer-right /
  outer-left edge.
- `ComputeSidebarBoundsOn{Right,Left}WithVtabOnSameSide` – outer edge
  inset when vertical tab is on the same side.
- `ComputeAdjustedPanelBoundsSidebarOnRight/LeftFullyOpen` – panel flush
  with sidebar edge when fully open.
- `ComputeAdjustedPanelBoundsSidebarOnRight/LeftAnimating` – animation
  offset is preserved (correct number of pixels visible to the user).

**Browser test (sidebar_browsertest.cc, ENABLE_SIDEBAR_V2 only)**

`SidebarV2PanelPositionTest` opens the upstream side panel and verifies:
- Sidebar on right (default LTR): `panel.right == sidebar.x`
- Sidebar on left (pref toggled): `sidebar.right == panel.x`

### DIFF
--- a/browser/ui/sidebar/sidebar_browsertest.cc
+++ b/browser/ui/sidebar/sidebar_browsertest.cc
@@ -2051,7 +2051,9 @@ IN_PROC_BROWSER_TEST_F(SidebarBrowserTest, SidebarV2PanelPositionTest) {
 
   // Panel sits immediately left of the sidebar control:
   //   [contents] [panel] [sidebar_control]
-  EXPECT_EQ(panel->bounds().right(), sidebar->bounds().x());
+  EXPECT_EQ(panel->bounds().right(), sidebar->bounds().x())
+      << "panel=" << panel->bounds().ToString()
+      << " sidebar=" << sidebar->bounds().ToString();
 
   // --- Sidebar on left (kSidePanelHorizontalAlignment = false)
   prefs->SetBoolean(prefs::kSidePanelHorizontalAlignment, false);
@@ -2061,7 +2063,9 @@ IN_PROC_BROWSER_TEST_F(SidebarBrowserTest, SidebarV2PanelPositionTest) {
 
   // Panel sits immediately right of the sidebar control:
   //   [sidebar_control] [panel] [contents]
-  EXPECT_EQ(sidebar->bounds().right(), panel->bounds().x());
+  EXPECT_EQ(sidebar->bounds().right(), panel->bounds().x())
+      << "sidebar=" << sidebar->bounds().ToString()
+      << " panel=" << panel->bounds().ToString();
 }
 #endif  // BUILDFLAG(ENABLE_SIDEBAR_V2)
 
--- a/browser/ui/sidebar/sidebar_browsertest.cc
+++ b/browser/ui/sidebar/sidebar_browsertest.cc
@@ -19,6 +19,7 @@
 #include "base/test/scoped_feature_list.h"
 #include "brave/app/brave_command_ids.h"
 #include "brave/browser/ui/browser_commands.h"
+#include "brave/browser/ui/sidebar/buildflags/buildflags.h"
 #include "brave/browser/ui/sidebar/features.h"
 #include "brave/browser/ui/sidebar/sidebar_controller.h"
 #include "brave/browser/ui/sidebar/sidebar_model.h"
@@ -30,7 +31,6 @@
 #include "brave/browser/ui/views/frame/split_view/brave_contents_container_view.h"
 #include "brave/browser/ui/views/frame/split_view/brave_multi_contents_view.h"
 #include "brave/browser/ui/views/frame/split_view/brave_multi_contents_view_mini_toolbar.h"
-#include "brave/browser/ui/views/side_panel/side_panel.h"
 #include "brave/browser/ui/views/side_panel/side_panel_resize_widget.h"
 #include "brave/browser/ui/views/sidebar/sidebar_container_view.h"
 #include "brave/browser/ui/views/sidebar/sidebar_control_view.h"
@@ -67,6 +67,7 @@
 #include "chrome/browser/ui/ui_features.h"
 #include "chrome/browser/ui/views/frame/multi_contents_view.h"
 #include "chrome/browser/ui/views/frame/toolbar_button_provider.h"
+#include "chrome/browser/ui/views/side_panel/side_panel.h"
 #include "chrome/common/chrome_switches.h"
 #include "chrome/common/pref_names.h"
 #include "chrome/test/base/in_process_browser_test.h"
@@ -135,7 +136,11 @@ class SidebarBrowserTest : public InProcessBrowserTest {
   }
 
   views::Widget* GetSidePanelResizeWidget() {
+#if BUILDFLAG(ENABLE_SIDEBAR_V2)
+    NOTREACHED() << "No resize widget in v2";
+#else
     return GetSidePanel()->resize_widget_->GetWidget();
+#endif
   }
 
   raw_ptr<SidebarItemsContentsView> GetSidebarItemsContentsView(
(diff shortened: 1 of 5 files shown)


browser/ui/views/frame/layout/brave_browser_view_tabbed_layout_impl.h
```diff
@@ -55,6 +56,26 @@ class BraveBrowserViewTabbedLayoutImpl : public BrowserViewTabbedLayoutImpl {
...
+  // the inner edge of |sidebar_bounds|.
+  static gfx::Rect ComputeAdjustedPanelBounds(bool sidebar_on_left,
+                                              const gfx::Rect& sidebar_bounds,
+                                              const gfx::Rect& panel_bounds);
+
+  friend class BraveBrowserViewTabbedLayoutImplSidebarTest;
+
 #if BUILDFLAG(IS_MAC)
   gfx::Insets AddFrameBorderInsets(const gfx::Insets& insets) const;
   gfx::Insets AddVerticalTabFrameBorderInsets(const gfx::Insets& insets) const;
```
netzenbot, line 81: nit: Per [CS-015](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-015), `friend` declarations should go immediately after `private:`, before any member functions or variables. Move `friend class BraveBrowserViewTabbedLayoutImplSidebarTest;` to the top of the `private:` block.

browser/ui/views/frame/layout/brave_browser_view_tabbed_layout_impl.cc
```diff
@@ -325,6 +325,41 @@ void BraveBrowserViewTabbedLayoutImpl::CalculateBraveVerticalTabStripLayout(
   layout.AddChild(views().vertical_tab_strip_host, vertical_tab_strip_bounds);
 }
 
+// static
```
netzenbot: nit: Per [CS-017](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-017), `.cc` definitions should appear in the same order as their `.h` declarations. `ComputeSidebarBounds` and `ComputeAdjustedPanelBounds` are declared after `GetInsetsConsideringVerticalTabHost` in the header but defined much earlier in the `.cc` (between `CalculateBraveVerticalTabStripLayout` and `CalculateSideBarLayout`). Either move the `.h` declarations to match the `.cc` order, or move the `.cc` definitions to after `GetInsetsConsideringVerticalTabHost`.

browser/ui/views/frame/layout/brave_browser_view_tabbed_layout_impl.cc
```diff
@@ -337,32 +372,86 @@ void BraveBrowserViewTabbedLayoutImpl::CalculateSideBarLayout(
...
+  //   [vertical_tab] [sidebar] [panel] [contents]
+  //   [sidebar] [panel] [contents] [vertical_tab]
+  //
+  // In V2, sidebar_container holds only the control view and the upstream
+  // side panels (toolbar/contents_height_side_panel) are direct children of
+  // browser_view positioned separately.  In V1, sidebar_container wraps both
+  // the control and the side panel, and those upstream panel pointers point
+  // back into the container (so they are NOT in the proposed layout as top-
+  // level entries — layout.GetLayoutFor returns null).  This means
+  // adjust_panel below is a no-op in V1 and meaningful only in V2.
```
netzenbot: nit: `adjust_panel below` uses a positional reference. Reference it explicitly, e.g. `The adjust_panel lambda in this function is a no-op in V1 and meaningful only in V2.` ([❌ Don't Use Positional Terms in Code Comments](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-024))

browser/ui/views/frame/layout/brave_browser_view_tabbed_layout_impl.h
```diff
@@ -55,6 +56,26 @@ class BraveBrowserViewTabbedLayoutImpl : public BrowserViewTabbedLayoutImpl {
...
+  // between the contents area and the sidebar control view.  The panel is
+  // shifted inward (toward the centre of the browser) until it is flush with
+  // the inner edge of |sidebar_bounds|.
+  static gfx::Rect ComputeAdjustedPanelBounds(bool sidebar_on_left,
+                                              const gfx::Rect& sidebar_bounds,
+                                              const gfx::Rect& panel_bounds);
+
+  friend class BraveBrowserViewTabbedLayoutImplSidebarTest;
+
 #if BUILDFLAG(IS_MAC)
```
netzenbot, line 79: nit: `ComputeSidebarBounds` and `ComputeAdjustedPanelBounds` are pure geometry functions with no side effects. Making them `public static` would remove the need for `friend class BraveBrowserViewTabbedLayoutImplSidebarTest` and let tests call them directly without special access. https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-011

browser/ui/sidebar/sidebar_browsertest.cc
```diff
@@ -2023,4 +2028,41 @@ INSTANTIATE_TEST_SUITE_P(
...
+
+  // Panel sits immediately left of the sidebar control:
+  //   [contents] [panel] [sidebar_control]
+  EXPECT_EQ(panel->bounds().right(), sidebar->bounds().x());
+
+  // --- Sidebar on left (kSidePanelHorizontalAlignment = false)
+  prefs->SetBoolean(prefs::kSidePanelHorizontalAlignment, false);
+  RunScheduledLayouts();
+
+  ASSERT_TRUE(sidebar->sidebar_on_left());
```
netzenbot, line 2062: nit: The layout assertions (e.g., `EXPECT_EQ(panel->bounds().right(), sidebar->bounds().x())`) don't include diagnostic context. Adding `<< panel->bounds().ToString() << " vs " << sidebar->bounds().ToString()` would make failures actionable without re-running under a debugger. https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-014

simonhong: Updated

browser/ui/views/frame/layout/brave_browser_view_tabbed_layout_impl.cc
simonhong: Updated.

browser/ui/views/frame/layout/brave_browser_view_tabbed_layout_impl.cc
simonhong: Updated.

browser/ui/views/frame/layout/brave_browser_view_tabbed_layout_impl.h
simonhong: Updated.

browser/ui/views/frame/layout/brave_browser_view_tabbed_layout_impl.h
simonhong: Updated.

browser/ui/views/frame/layout/brave_browser_view_tabbed_layout_impl.h
```diff
@@ -10,6 +10,10 @@
 
 #include "chrome/browser/ui/views/frame/layout/browser_view_tabbed_layout_impl.h"
 
+namespace gfx {
+class Rect;
+}  // namespace gfx
```
sangwoo108: Q. I'm wondering if we should have #include instead of forward declartion as we're returing gfx::Rect as value in the static methods.

sangwoo108 (approved):
LGTM

browser/ui/views/frame/layout/brave_browser_view_tabbed_layout_impl.h
simonhong: Updated in that way. thanks!

netzenbot (approved):

PR #35455, opened 2026-04-13
Using brave_account::AuthenticationObserver instead of prefs observer.

boocmp:
As far as Brave Account service provides the account state observer, Email aliases service doesn't need to access the prefs directly.


szilardszaloki (commented):
Moving away from pref observers is definitely the right call, but we have an unnecessary abstraction layer.

Right now:
- `BraveAccountService` emits `OnAccountStateChanged()` via `brave_account::mojom::AuthenticationObserver`
- `EmailAliasesAuth` caches the state and fires `EmailAliasesService::OnAuthChanged()`
- `EmailAliasesService` then re-broadcasts the exact same event via its own observer (`email_aliases::mojom::EmailAliasesServiceObserver`'s `OnAuthStateChanged()`)

So this is effectively just wrapping and forwarding the Account observer.

You should use `brave_account::mojom::AuthenticationObserver` directly in the Email Aliases UI instead (which would remove a lot of boilerplate and avoid duplicating the same signal).

netzenbot (approved):

browser/email_aliases/email_aliases_browsertest.cc
```diff
@@ -111,36 +161,101 @@ class EmailAliasesBrowserTestBase : public InProcessBrowserTest {
...
+        base::expected<brave_account::mojom::LoginInitializeResultPtr,
+                       brave_account::mojom::LoginErrorPtr>>
+        login_init_future;
+    authentication->LoginInitialize(
+        brave_account::mojom::Service::kEmailAliases, email, "serialized_ke1",
+        login_init_future.GetCallback());
+    const auto login_init_result = login_init_future.Take();
+    ASSERT_TRUE(login_init_result.has_value());
+
+    base::test::TestFuture<
```
netzenbot, line 211: nit: SetBraveAccountLoggedIn (and SetBraveAccountLoggedOut) wrap several ASSERT_TRUE checks and are called from multiple tests. Add SCOPED_TRACE(base::Location::Current()) at the top so a failure points to the calling test rather than this helper. ([Use SCOPED_TRACE with base::Location in Browser Test Helpers](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-015))

browser/ui/webui/email_aliases/email_aliases_panel_ui.h
```diff
@@ -6,6 +6,7 @@
 #ifndef BRAVE_BROWSER_UI_WEBUI_EMAIL_ALIASES_EMAIL_ALIASES_PANEL_UI_H_
 #define BRAVE_BROWSER_UI_WEBUI_EMAIL_ALIASES_EMAIL_ALIASES_PANEL_UI_H_
 
+#include "brave/components/brave_account/mojom/brave_account.mojom.h"
 #include "brave/components/email_aliases/email_aliases.mojom.h"
 #include "chrome/browser/ui/webui/constrained_web_dialog_ui.h"
 #include "content/public/browser/webui_config.h"
```
netzenbot, line 12: nit: this header only references `brave_account::mojom::Authentication` as a `mojo::PendingReceiver<>` parameter, so `brave_account.mojom-forward.h` is sufficient here and the full `brave_account.mojom.h` include can move to the .cc. ([Use *.mojom-forward.h in Headers When Only Forward Declarations Are Needed](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-070))

browser/brave_content_browser_client.cc
```diff
@@ -802,7 +802,8 @@ void BraveContentBrowserClient::RegisterTrustedWebUIInterfaceBrokers(
   if (email_aliases::features::IsEmailAliasesEnabled()) {
     registry.ForWebUI<EmailAliasesPanelUI>()
         .Add<email_aliases::mojom::EmailAliasesService>()
-        .Add<email_aliases::mojom::EmailAliasesPanelHandler>();
+        .Add<email_aliases::mojom::EmailAliasesPanelHandler>()
+        .Add<brave_account::mojom::Authentication>();
```
szilardszaloki: Super nit: keep alphabetical order.

components/email_aliases/email_aliases_service.cc
szilardszaloki, line 1: We'd want to avoid consulting a potentially stale cached account state, so ideally we'd get rid of synchronous "am I logged in?" checks altogether. In this case, `!auth_->IsAuthenticated()` in `EmailAliasesService::RefreshAliases()` isn't really needed, since `GetServiceToken()` is already safe to call regardless of state and is the authoritative check anyway. If the user is logged out, it ends up in `brave_account::StateBase::GetServiceToken()`, which just synchronously returns `GetServiceTokenClientErrorCode::CalledInWrongState` without ever touching the network.

szilardszaloki: I see you've now built up quite a bit of machinery around synchronously querying account state. A reasonable simplification: cache the email, and that's it. Brave Account hard-asserts that a logged-in state always carries a (non-empty) email address (see the `CHECK(email && !email->empty())` guarding `LoggedInState` construction in `brave_account_state_prefs.cc`), so we can lean on that invariant directly. Cache the email off the `OnAccountStateChanged()` push, if the cached string is empty, the user is logged out, otherwise they're logged in.

boocmp: Ok, switched to use the cached email. In some of follow ups, I think I'll get rid of the EmailAliasesAuth class completely by inlining its functionality in the service.

components/email_aliases/email_aliases_auth.cc
```diff
@@ -68,17 +49,19 @@ void EmailAliasesAuth::GetServiceToken(
...
 void EmailAliasesAuth::OnDisconnect() {
   brave_account_auth_.reset();
+  auth_email_.clear();
   on_changed_.Run();
 }
 
-void EmailAliasesAuth::OnPrefChanged(const std::string& pref_name) {
+void EmailAliasesAuth::OnAccountStateChanged(
+    brave_account::mojom::AccountStatePtr state) {
+  if (state->is_logged_in()) {
```
szilardszaloki: Nit:
```cpp
CHECK(state);
```

szilardszaloki: Also, if you want it a bit terser:
```cpp
auth_email_ = state->is_logged_in() ? state->get_logged_in()->email : "";
```

components/email_aliases/email_aliases_auth.h
```diff
@@ -6,51 +6,44 @@
...
-class EmailAliasesAuth {
+class EmailAliasesAuth : public brave_account::mojom::AuthenticationObserver {
  public:
   using OnChangedCallback = base::RepeatingClosure;
 
   explicit EmailAliasesAuth(
-      PrefService& prefs_service,
       mojo::PendingRemote<brave_account::mojom::Authentication>
           brave_account_auth,
       OnChangedCallback on_changed = base::DoNothing());
```
szilardszaloki: Is this default ever used?

components/email_aliases/email_aliases_auth.cc
```diff
@@ -68,17 +49,19 @@ void EmailAliasesAuth::GetServiceToken(
...
 }
 
-void EmailAliasesAuth::SetAuthEmailForTesting(const std::string& email) {
-  auth_email_for_testing_ = email;
-  on_changed_.Run();
-}
-
 void EmailAliasesAuth::OnDisconnect() {
   brave_account_auth_.reset();
+  auth_email_.clear();
```
szilardszaloki, line 54: Just noting that this conflates "pipe died" with "user logged out".

components/email_aliases/email_aliases_service.cc
szilardszaloki: Agreed that this class should be folded into the service.

components/email_aliases/email_aliases_service.cc
```diff
@@ -208,8 +196,6 @@ void EmailAliasesService::AddObserver(
   auto id = observers_.Add(std::move(observer));
   auto* remote = observers_.Get(id);
   if (remote) {
-    remote->OnAuthStateChanged(
-        mojom::AuthState::New(GetCurrentStatus(), GetAuthEmail()));
     RefreshAliases();
   }
```
szilardszaloki: `observers_.Add(std::move(observer))` should always return a valid ID, and `observers_.Get(id)` immediately after with that same ID should always return non-`nullptr`. Also, `remote` is now unused, so this should likely just be:
```cpp
observers_.Add(std::move(observer));
RefreshAliases();
```

components/email_aliases/email_aliases_service.cc
```diff
@@ -112,7 +112,7 @@ EmailAliasesService::EmailAliasesService(
   CHECK(base::FeatureList::IsEnabled(email_aliases::features::kEmailAliases));
   CHECK(brave_account_auth);
 
-  auth_.emplace(pref_service_.get(), std::move(brave_account_auth),
+  auth_.emplace(std::move(brave_account_auth),
                 base::BindRepeating(&EmailAliasesService::OnAuthChanged,
                                     weak_factory_.GetWeakPtr()));
```
szilardszaloki, line 117: You can bind `EmailAliasesService::RefreshAliases()` directly instead and remove `EmailAliasesService::OnAuthChanged()`.

components/email_aliases/email_aliases_service.cc
```diff
@@ -252,7 +238,7 @@ void EmailAliasesService::OnEditAliasResponse(
 
 void EmailAliasesService::RefreshAliases() {
   CHECK(auth_);
-  if (observers_.empty() || !auth_->IsAuthenticated()) {
+  if (observers_.empty()) {
     return;
   }
   auth_->GetServiceToken(
```
szilardszaloki, line 241:
```cpp
CHECK_DEREF(auth_).GetServiceToken(...);
```
in case you prefer.

netzenbot (approved):

browser/email_aliases/email_aliases_browsertest.cc
```diff
@@ -113,21 +164,41 @@ class EmailAliasesBrowserTestBase : public InProcessBrowserTest {
...
   void SetUpOnMainThread() override {
+    brave_account::BraveAccountEncryption::SetOSCryptCallbacksForTesting(
+        base::BindRepeating([](const std::string& in, std::string* out) {
+          *out = in;
+          return true;
+        }),
+        base::BindRepeating([](const std::string& in, std::string* out) {
+          *out = in;
+          return true;
+        }));
```
szilardszaloki: I don't think you need this, `TestOSCryptAsync` already gives you a working encryptor.

browser/email_aliases/email_aliases_browsertest.cc
```diff
@@ -141,15 +212,60 @@ class EmailAliasesBrowserTestBase : public InProcessBrowserTest {
       user_prefs::UserPrefs::Get(browser()->profile())
           ->SetBoolean(prefs::kPromoShown, true);
     }
+  }
+
+  void TearDownOnMainThread() override {
+    brave_account::BraveAccountEncryption::SetOSCryptCallbacksForTesting(
+        base::NullCallback(), base::NullCallback());
+    InProcessBrowserTest::TearDownOnMainThread();
+  }
```
szilardszaloki: And then you won't need this either.

browser/email_aliases/email_aliases_browsertest.cc
```diff
@@ -56,6 +69,43 @@ namespace email_aliases {
...
+         body.serialized_ke2 = "serialized_ke2";
+         return body;
+       }()});
+
+  brave_account::endpoint_client::MockResponseFor<LoginFinalize>(
+      test_url_loader_factory,
+      {.net_error = net::OK, .status_code = net::HTTP_OK, .body = [email] {
+         LoginFinalize::Response::SuccessBody body;
+         body.auth_token = kAuthenticationToken;
+         body.email = email;
```
szilardszaloki: `email` is always `kSuccessEmail` in your tests, in case you want to simplify:
```cpp
body.email = kSuccessEmail;
```
and remove the function parameter.

browser/email_aliases/email_aliases_browsertest.cc
```diff
@@ -56,6 +69,43 @@ namespace email_aliases {
 namespace {
 constexpr char kSuccessEmail[] = "success@domain.com";
 
+constexpr char kAuthenticationToken[] = "authentication_token";
+constexpr char kCachedServiceToken[] = "cached_service_token";
+constexpr char kLoginToken[] = "login_token";
```
szilardszaloki: These are only referenced once, if you wanted to inline them.

browser/email_aliases/email_aliases_browsertest.cc
```diff
@@ -141,15 +212,60 @@ class EmailAliasesBrowserTestBase : public InProcessBrowserTest {
...
+    ASSERT_TRUE(login_finalize_future.Get().has_value());
+
+    base::test::TestFuture<
+        base::expected<brave_account::mojom::GetServiceTokenResultPtr,
+                       brave_account::mojom::GetServiceTokenErrorPtr>>
+        service_token_future;
+    authentication->GetServiceToken(
+        brave_account::mojom::Service::kEmailAliases,
+        service_token_future.GetCallback());
+    ASSERT_TRUE(service_token_future.Get().has_value());
```
szilardszaloki: Why call `GetServiceToken()` here? It shouldn't be necessary — `LoginFinalize()` will log you in.

browser/email_aliases/email_aliases_browsertest.cc
```diff
@@ -141,15 +212,60 @@ class EmailAliasesBrowserTestBase : public InProcessBrowserTest {
...
+
+    ASSERT_TRUE(base::test::RunUntil(
+        [this]() { return email_aliases_service()->IsAuthenticated(); }));
+  }
+
+  void SetBraveAccountLoggedOut() {
+    mojo::Remote<brave_account::mojom::Authentication> authentication =
+        BindAuthentication();
+    authentication->LogOut();
+    authentication.FlushForTesting();
```
szilardszaloki: `authentication.FlushForTesting();` shouldn't be necessary — `base::test::RunUntil()` should be sufficient on its own.

browser/email_aliases/email_aliases_browsertest.cc
```diff
@@ -308,6 +424,13 @@ class EmailAliasesBrowserTestBase : public InProcessBrowserTest {
         browser()->profile());
   }
 
+  mojo::Remote<brave_account::mojom::Authentication> BindAuthentication() {
+    mojo::Remote<brave_account::mojom::Authentication> authentication;
+    brave_account::BraveAccountServiceFactory::GetFor(browser()->profile())
+        ->BindInterface(authentication.BindNewPipeAndPassReceiver());
+    return authentication;
+  }
```
szilardszaloki: You could also just store a
```cpp
mojo::Remote<mojom::Authentication> authentication_;
```
and bind it once during setup.

browser/email_aliases/email_aliases_browsertest.cc
```diff
@@ -498,20 +612,20 @@ IN_PROC_BROWSER_TEST_F(EmailAliasesBrowserTest, ContextMenuAuthorizedCancel) {
   EXPECT_TRUE(AwaitText("#type-email", ""));  // text not changed
 }
 IN_PROC_BROWSER_TEST_F(EmailAliasesBrowserTest, LogInLogOut) {
-  // Prepare auth token
-  email_aliases_service()->GetAuth()->SetAuthEmailForTesting(kSuccessEmail);
+  SetBraveAccountLoggedIn(kSuccessEmail);
+  EXPECT_TRUE(email_aliases_service()->IsAuthenticated());
```
szilardszaloki: This seems redundant: `SetBraveAccountLoggedIn()` already `ASSERT`s `RunUntil(IsAuthenticated())` before returning.

browser/email_aliases/email_aliases_browsertest.cc
```diff
@@ -498,20 +612,20 @@ IN_PROC_BROWSER_TEST_F(EmailAliasesBrowserTest, ContextMenuAuthorizedCancel) {
...
-  // Reset token
-  email_aliases_service()->GetAuth()->SetAuthEmailForTesting("");
+  SetBraveAccountLoggedOut();
+  EXPECT_FALSE(email_aliases_service()->IsAuthenticated());
   WaitDisappear("#create-new-item-button");  // Settings in sing-in state.
 
-  // Logged-in again
-  email_aliases_service()->GetAuth()->SetAuthEmailForTesting(kSuccessEmail);
+  SetBraveAccountLoggedIn(kSuccessEmail);
+  EXPECT_TRUE(email_aliases_service()->IsAuthenticated());
```
szilardszaloki: Same here.

browser/email_aliases/email_aliases_browsertest.cc
```diff
@@ -498,20 +612,20 @@ IN_PROC_BROWSER_TEST_F(EmailAliasesBrowserTest, ContextMenuAuthorizedCancel) {
...
 
   // Settings in logged-in state
   Navigate(GURL("chrome://settings/email-aliases"));
   InjectHelpers(ActiveWebContents());
   Wait("#create-new-item-button");
 
-  // Reset token
-  email_aliases_service()->GetAuth()->SetAuthEmailForTesting("");
+  SetBraveAccountLoggedOut();
+  EXPECT_FALSE(email_aliases_service()->IsAuthenticated());
```
szilardszaloki: Same here with `SetBraveAccountLoggedOut()` and `RunUntil(!IsAuthenticated())`.

browser/email_aliases/email_aliases_browsertest.cc
```diff
@@ -113,21 +164,41 @@ class EmailAliasesBrowserTestBase : public InProcessBrowserTest {
...
+        }));
+
+    using brave_account::endpoints::AuthValidate;
+    brave_account::endpoint_client::MockResponseFor<AuthValidate>(
+        test_url_loader_factory_,
+        {.net_error = net::OK, .status_code = net::HTTP_OK, .body = [] {
+           AuthValidate::Response::SuccessBody body;
+           body.email = kSuccessEmail;
+           return body;
+         }()});
```
szilardszaloki: I don't think you need this either. Without it, the background `AuthValidate` poll just hangs pending forever, which should be fine for these tests.

components/email_aliases/email_aliases_service_unittest.cc
```diff
@@ -168,19 +165,38 @@ class EmailAliasesAPITest : public ::testing::Test {
...
+
+    // Default list response so login-triggered refresh does not notify an
+    // error.
     AddRefreshResponseFor();
 
+    email_aliases_auth_observer_remote_->OnAccountStateChanged(
+        brave_account::mojom::AccountState::NewLoggedIn(
+            brave_account::mojom::LoggedInState::New("test@login.com",
+                                                     nullptr)));
+    email_aliases_auth_observer_remote_.FlushForTesting();
```
szilardszaloki, line 191: This shouldn't be necessary here either — `RunUntil(IsAuthenticated())` should have you covered below.

components/email_aliases/email_aliases_auth.h
boocmp: For now, no.

browser/email_aliases/email_aliases_browsertest.cc
boocmp: Removed

components/email_aliases/email_aliases_service.cc
boocmp: I'd prefer it as is

szilardszaloki (approved):
Browser backend changes LGTM.

components/brave_account/mojom/BUILD.gn
```diff
@@ -24,6 +24,8 @@ mojom("mojom") {
   }
 
   generate_java = true
+  generate_legacy_js_bindings = true
```
fallaciousreasoning, line 27: Why do we need to generate the legacy bindings now? If possible we should try and get the new style bindings working.

components/email_aliases/resources/content/use_email_aliases.ts
```diff
@@ -4,54 +4,73 @@
...
-const emptyAliasesUpdate = (): AliasesUpdate => ({
+const emptyAliasesUpdate = {
   aliases: [],
   error: undefined,
-})
+}
+
+export function isAccountLoggedIn(
+  accountState: AccountState | undefined,
+): boolean {
```
fallaciousreasoning, line 24:
```suggestion
export function isAccountLoggedIn(
  accountState: AccountState | undefined,
): accountState is Required<AccountState> {
```

and then we can get rid of all the `!` in the usages :smile:

components/email_aliases/resources/BUILD.gn
```diff
@@ -34,7 +34,7 @@ transpile_web_ui("email_aliases_ui") {
   resource_name = "email_aliases"
 
   deps = [
-    "//brave/components/brave_account/mojom:mojom_ts__generator",
+    "//brave/components/brave_account/mojom:mojom_js",
```
fallaciousreasoning, line 37: Are we sure we need this? It'd be good if we could keep using the new TS generator, as I think Chromium will switch off the legacy one eventually

fallaciousreasoning (commented):
this looks like a pretty significant improvement, but I don't think we should move to the legacy JS bindings if we can help it.

components/email_aliases/resources/BUILD.gn
boocmp: Discussed in DM, we'll try to switch to TS generator in the follow up.

szilardszaloki: Yeah, I suggested using the legacy JS bindings. The TS generator emits `.ts` files that import each other with `.js` extensions. If `transpile_web_ui` passes those `.ts` files straight to webpack, it won't resolve the `.js` imports on its own. We'll probably have to compile them first.

PR #35457, opened 2026-04-13
[iOS] Increase sync device limit to 50 to match backend

### LINKED ISSUE #54452: Sync chain device limit is incorrect on iOS
iOS currently limits the number of devices that can be added to the sync chain to 10 which is not up to date with backend (50)

Community post: https://community.brave.app/t/brave-sync-limit-10-devices-not-updating-after-removing-devices/570318/

kylehickinson:
The current client-side limit of 10 was out of date

Resolves https://github.com/brave/brave-browser/issues/54452

### DIFF
--- a/ios/brave-ios/Sources/Brave/Frontend/Sync/SyncWelcomeViewController.swift
+++ b/ios/brave-ios/Sources/Brave/Frontend/Sync/SyncWelcomeViewController.swift
@@ -464,7 +464,7 @@ extension SyncWelcomeViewController: SyncPairControllerDelegate {
             )
             present(alert, animated: true, completion: nil)
           case .blocked:
-            // Devices 10 and more - add alert to block and prevent sync
+            // Devices 50 and more - add alert to block and prevent sync
             let alert = UIAlertController(
               title: Strings.genericErrorTitle,
               message: Strings.Sync.maximumDeviceReachedErrorDescription,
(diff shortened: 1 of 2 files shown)


ios/brave-ios/Sources/Brave/Frontend/Sync/SyncWelcomeViewController.swift
```diff
@@ -510,9 +510,9 @@ extension SyncWelcomeViewController: SyncPairControllerDelegate {
     switch devices.count {
     case 1...4:
       deviceLimitLevel = .safe
-    case 5...9:
+    case 5..<50:
       deviceLimitLevel = .approvalNeeded
-    case 10...:
+    case 50...:
```
StephenHeaps, line 515: Is there a constant in core that could be used so this only needs to be updated once for all platforms?

kylehickinson: AFAIK iOS was the only one to implement this client side check, so nothing shared to use from desktop/android. The limit is here: https://github.com/brave/go-sync/blob/master/command/device_limit.go#L9

Perhaps we just remove it entirely but I'll let @fmarier weigh in on that

fmarier: Yes, we may as well rely on the server-enforced limit without enforcing an additional client-side limit.

kylehickinson: @fmarier Lets open a follow-up issue for removing it in that case, as it has some other product related changes that probably will want more focus

fmarier (approved):

StephenHeaps (approved):

ios/brave-ios/Sources/Brave/Frontend/Sync/SyncWelcomeViewController.swift
StephenHeaps: Sounds good, feel free to resolve comment to merge 🙂

kylehickinson: Opened https://github.com/brave/brave-browser/issues/54453

PR #35460, opened 2026-04-13
Open origin settings on first purchase

### LINKED ISSUE #54236: Brave Origin upgrade case should load brave://settings/origin when purchase state is loaded

bbondy:
Fix https://github.com/brave/brave-browser/issues/54236


components/brave_origin/brave_origin_service.cc
```diff
@@ -71,6 +72,14 @@ BraveOriginService::BraveOriginService(
   // Eagerly check purchase state on startup so the cached value is available.
   CheckPurchaseState(base::DoNothing());
 
+  // Re-check purchase state whenever SKU credentials change (e.g. after
+  // the user completes a purchase on account.brave.com).
+  skus_pref_registrar_.Init(local_state_);
+  skus_pref_registrar_.Add(
+      skus::prefs::kSkusState,
+      base::BindRepeating(&BraveOriginService::OnSkusStateChanged,
+                          base::Unretained(this)));
```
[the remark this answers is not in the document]
bbondy: The registrar is an owned member, and `Shutdown()` calls `skus_pref_registrar_.RemoveAll()` before nulling any state, so the callback cannot fire after teardown.

components/brave_origin/brave_origin_service.h
```diff
@@ -69,6 +71,17 @@ class BraveOriginService : public KeyedService {
...
+    // Called once when a purchase is first detected, after credentials
+    // have been verified.
+    virtual void OpenOriginSettings() = 0;
+  };
+
+  void SetDelegate(std::unique_ptr<Delegate> delegate);
+
  protected:
   // Local state and profile preferences this state is associated with
   raw_ptr<PrefService> local_state_;
```
netzenbot, line 84: The delegate is always set immediately after construction in the factory — consider passing it via the constructor instead. [ARCH-025](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-025) prefers constructor injection over setter methods to make dependencies explicit and avoid ambiguous initialization ordering.

browser/ui/brave_origin/BUILD.gn
```diff
@@ -0,0 +1,13 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+source_set("brave_origin") {
+  sources = [ "brave_origin_navigation.cc" ]
+
+  deps = [
+    "//brave/browser/brave_origin",
+    "//chrome/browser/profiles",
+  ]
```
netzenbot, line 14: `brave_origin_navigation.cc` includes `chrome/browser/ui/chrome_pages.h`, but the `deps` list only has `//brave/browser/brave_origin` and `//chrome/browser/profiles` — neither of which provides that header. Add `//chrome/browser/ui` (or the appropriate sub-target) to avoid relying on transitive deps that can break. See [BS-012](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-012).

browser/brave_origin/BUILD.gn
```diff
@@ -16,6 +16,7 @@ import("//build/config/features.gni")
 
 source_set("brave_origin") {
   sources = [
+    "brave_origin_navigation.h",
```
netzenbot, line 19: `brave_origin_navigation.h` includes `brave/components/brave_origin/brave_origin_service.h`, making `//brave/components/brave_origin` a header-visible dependency of this target. Move it from `deps` to `public_deps` so downstream targets (e.g. `//brave/browser/ui/brave_origin`) can find the transitive include. See [BS-024](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-024).

browser/ui/brave_origin/BUILD.gn
bbondy: can't be added directly because it would create a dependency cycle

cdesouza-chromium: So, ideally, the way to work with this is to something similar to the type of componentisation that is being done in Chromium, where you have a `foo` for the interfaces and `foo:impl` for the implementation to eventually break the cycle. Here's an example of this being done for `chrome/browser/push_messaging` just yesterday:

https://chromium-review.googlesource.com/c/chromium/src/+/6839344

Interestingly, your cycle is being caused by an inclusion of `chrome/browser/ui/chrome_pages.h`.  Here's an example of `//chrome/browser/ui/passwords` being componentised to allow fixing that in the future:

https://chromium-review.googlesource.com/c/chromium/src/+/6839344

Notice that this does require us to add your `*:impl` target to `allow_circular_includes_from` in `//chrome/browser/ui`, which we use `brave_chrome_browser_ui_allow_circular_includes_from` in brave for that.

browser/ui/config.gni
```diff
@@ -10,7 +10,7 @@ import("//brave/components/containers/buildflags/buildflags.gni")
 import("//brave/components/speedreader/common/buildflags/buildflags.gni")
 import("//build/config/ui.gni")
 
-brave_ui_allow_circular_includes_from = []
+brave_ui_allow_circular_includes_from = [ "//brave/browser/ui/brave_origin" ]
```
cdesouza-chromium: nit: maybe add a comment documenting why this is necessary like upstream does? For example, this is in the upstream gn file:

```
    # TODO(crbug.com/437811373): Remove this circular dependency when the following
    # headers get componentized:
    # - c/b/ui/chrome_pages.h and browser.h
    # - c/b/ui/webui/password_manager/password_manager_ui.h
    "//chrome/browser/ui/passwords:impl",
```

components/brave_origin/brave_origin_service.cc
```diff
@@ -227,6 +251,14 @@ void BraveOriginService::OnCredentialSummary(
   if (local_state_ && purchased) {
     local_state_->SetBoolean(kOriginPoliciesWereEnforced, true);
   }
+
+  // On first purchase detection, open the settings page so the user
+  // can configure Origin policies.
+  if (purchased && !was_previously_purchased && delegate_) {
+    delegate_->OpenOriginSettings();
+    delegate_.reset();
```
cdesouza-chromium: It is not important, but as this delegate has only one thunk, and we reset the delegate immediately after calling `OpenOriginSettings`, I'm wondering if just having a simple `base::OnceCallback<void()> on_open_origin_settings_` would simplify things to the reader. A `OnceCallback` would naturally move itself to a null/disengaged state.

goodov: the delegate is right approach, but resetting it here is wrong. It should only be removed in `::Shutdown()`.

detecting whether the settings were opened or not is part of the service logic. Making it equal to the delegate owning state is wrong.

`SkusServiceGetter` callback should become part of the delegate: `SkusService* GetSkusService()`.

tl;dr:

1. keep delegate
2. add boolean to track whether `OpenOriginSettings` was called for this profile (if you need to).
3. drop delegate only at `Shutdown()`.
4. replace `SkusServiceGetter` callback with delegate method.

browser/brave_origin/brave_origin_navigation.h
```diff
@@ -0,0 +1,31 @@
...
+// Constructed with the Profile so it can locate the browser window.
+class BraveOriginNavigationDelegate : public BraveOriginService::Delegate {
+ public:
+  explicit BraveOriginNavigationDelegate(Profile* profile);
+  ~BraveOriginNavigationDelegate() override;
+
+  void OpenOriginSettings() override;
+
+ private:
+  raw_ptr<Profile> profile_;
```
goodov: nit: `raw_ref`

browser/ui/brave_origin/BUILD.gn
cdesouza-chromium: Just unresolving this in case there was a misunderstanding, to clarify the CLs above:

We are nearly there as is at the moment, but to clarify what upstream is doing in cases like this to make it recognisable pattern once all these cases are fixed, as linked in the CLs above: I think this `gn` file should look something like the same pattern:

```
source_set("brave_origin") {
  sources = [ "brave_origin_navigation.h" ]

  # ...
}

source_set("impl") {
  sources = [ "brave_origin_navigation.cc" ]

  # ...
}
```

At the moment we have `browser/brave_origin/brave_origin_navigation.h` and `browser/ui/brave_origin/brave_origin_navigation.cc`. It would be better if `.h`, and `.cc` were living as siblings under the same directory, possibly `browser/ui/brave_origin`, similar `//chrome/browser/ui/passwords/settings` (https://chromium-review.googlesource.com/c/chromium/src/+/6839344).


Also, nit: the comments below explaining the circular dep can be removed, as a notice has been added to `brave_ui_allow_circular_includes_from`, and that's where people will end up at to understand the circular dep.

browser/ui/config.gni
```diff
@@ -17,6 +17,9 @@ if (enable_ai_chat) {
       [ "//brave/browser/ui/webui/ai_chat" ]
 }
 if (!is_android) {
+  # TODO(bbondy): Remove this circular dependency when
+  # chrome/browser/ui/chrome_pages.h gets componentized.
+  brave_ui_allow_circular_includes_from += [ "//brave/browser/ui/brave_origin" ]
```
cdesouza-chromium: Any particular reason for duplicating `brave_ui_allow_circular_includes_from` rather than just add a line to the array below?

cdesouza-chromium: Mentioned in DMs, but it is better to add this dep exclusion to `brave_chrome_browser_ui_allow_circular_includes_from`, to make exception for a circular dep with `//chrome/browser/ui` unless we do need a circular dep with `//brave/browser/ui` too.

cdesouza-chromium (approved):
LGTM % changes regarding the circular dep.

browser/ui/brave_origin/brave_origin_navigation.cc
```diff
@@ -0,0 +1,28 @@
...
+
+BraveOriginNavigationDelegate::~BraveOriginNavigationDelegate() = default;
+
+void BraveOriginNavigationDelegate::OpenOriginSettings() {
+  chrome::ShowSettingsSubPageForProfile(&*profile_, "origin");
+}
+
+mojo::PendingRemote<skus::mojom::SkusService>
+BraveOriginNavigationDelegate::GetSkusService() {
+  return skus::SkusServiceFactory::GetForContext(&*profile_);
```
goodov, line 34: `base::to_address`

components/brave_origin/brave_origin_service_unittest.cc
```diff
@@ -42,6 +43,36 @@ constexpr char kTestBrowserNotUserSettablePolicyKey[] =
...
+ public:
+  void OpenOriginSettings() override {}
+  mojo::PendingRemote<skus::mojom::SkusService> GetSkusService() override {
+    return {};
+  }
+};
+
+// Delegate backed by a FakeSkusService, with an observable flag for
+// OpenOriginSettings calls.
+class TestDelegate : public BraveOriginService::Delegate {
```
goodov, line 57: vars and mock classes in tests should be in anonymous ns. The test fixture should NOT be in anonymous ns.

ios/browser/brave_origin/brave_origin_service_factory.mm
```diff
@@ -163,7 +163,7 @@
       user_prefs::UserPrefs::Get(profile), profile_id,
       profile->GetPolicyConnector()->GetPolicyService(),
       GetApplicationContext()->GetBrowserPolicyConnector()->GetPolicyService(),
-      std::move(skus_service_getter));
+      std::move(skus_service_getter), nullptr);
```
goodov: is this wrong now?

goodov (approved):

netzenbot (approved):

browser/ui/brave_origin/brave_origin_navigation.cc
```diff
@@ -0,0 +1,37 @@
...
+#if !BUILDFLAG(IS_ANDROID)
+  chrome::ShowSettingsSubPageForProfile(&*profile_, "origin");
+#else
+  NOTIMPLEMENTED();
+#endif
+}
+
+mojo::PendingRemote<skus::mojom::SkusService>
+BraveOriginNavigationDelegate::GetSkusService() {
+  return skus::SkusServiceFactory::GetForContext(&*profile_);
```
bridiver, line 34: you should be using dependency injection for SkusService. Just because we happen to pass in Profile doesn't mean that we should abandon dependency injection. See https://chromium.googlesource.com/chromium/src/+/lkgr/docs/chrome_browser_design_principles.md#simple-browser-feature for a specific example that includes Profile

bridiver: actually that example doesn’t include Profile and also things that can be derived from Profile, I mixed up two different examples, but it’s still not correct to use Profile as a lookup key even if you need Profile anyway

bridiver: for instance unit tests for BraveOriginNavigationDelegate will now depend on SkusServiceFactory and there is no good way to mock or override it

PR #35465, opened 2026-04-13
Fix default browser detection for Origin branded builds

bbondy:
## Summary

Brave Origin was thinking it was default when Brave was default.

- **macOS**: Override `GetDefaultBrowser()` and `IsDefaultHandlerForUTType()` with Brave-aware brand comparison that correctly distinguishes Origin (`com.brave.Browser.origin.*`) from regular Brave (`com.brave.Browser.*`). The upstream `IsAnotherChromeChannel()` only compares the first 3 bundle ID components, which incorrectly treats Origin as "another channel" of regular Brave.
- **Linux**: Scope `IsAnyBraveBrowserDefaultBrowser()` to search for the current brand's desktop file prefix (`brave-origin` for Origin, `brave-browser` for regular) instead of always searching for `brave-browser`.

## Test plan
- [ ] On macOS with regular Brave as default, open Brave Origin — verify the "set as default browser" infobar appears
- [ ] On macOS with Brave Origin as default, open regular Brave — verify the infobar appears
- [ ] On macOS with Brave Beta as default, open Brave Stable — verify the infobar does NOT appear (same brand, different channel)
- [ ] On Linux, repeat the above scenarios using the appropriate desktop files

### DIFF
--- a/chromium_src/chrome/browser/shell_integration_linux.cc
+++ b/chromium_src/chrome/browser/shell_integration_linux.cc
@@ -3,6 +3,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this file,
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+#include "brave/components/brave_origin/buildflags/buildflags.h"
+
 #define GetDefaultBrowser GetDefaultBrowser_ChromiumImpl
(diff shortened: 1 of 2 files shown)


emerick (approved):
LGTM, but there's an unused function warning that needs to be addressed for Mac.

PR #35470, opened 2026-04-13
[iOS] Only restore Origin when using restoring in Origin paywall

kylehickinson:
This avoids attempts to restore other Leo & VPN services when tapping the restore button on the Origin paywall


nuo-xu (approved):
nit: need to update the comments inside `BraveStoreSDK`'s`private func updateSkusPurchaseState(for product: BraveStoreProduct) async throws` 
it currently says it only supports Leo

kylehickinson:
I'll get that in a separate PR as it doesn't really relate to this change specifically

PR #35473, opened 2026-04-14
Add optional horizontal tab strip scroll buttons

### LINKED ISSUE #54307: [Feature Request] Provide clickable buttons at the side of the tabstrip for scrolling tabs - for people with accessibility issues
### Description

As suggested by @sangwoo108 in issue #53101 and stemming from current discussions in issue #52045, people with chronic pain and motor accessibility issues had relied on the side buttons at the end of the tabstrip. They were used to scroll the tabs (when the tabstrip is in scroll mode, horizonatally and vertically) or to change the active tab (when in fixed mode).
With current implementation provided in response to #53101, this need was overlooked.
It would be great if a solution could be implemented to address this need.

### Steps to reproduce

1. Set the tabstrip in scroll mode
2. Adjust the tab size according to own preference (using a nightly build for now)
3. Try to scroll tabs without using the mouse wheel

### Actual result

Currently scrolling the tabs without the mouse wheel is not possible.
Tab dragging will only change the position of the tab in among the other tabs.
There are no buttons at the side of the tabstrip that could allow scrolling the tabs by simply clicking (remember, we operate as a disabled user).

### Expected result

Two buttons should be present (according to a parameter to enable/disable their presence).
These buttons should allow a disabled user to scroll tabs left and right (or up and down, depending on the tabstrip orientation).

### Reproduces how often

Easily reproduced

Any nightly after 1.91.x

### Channel information

- [ ] release (stable)
- [ ] beta
- [x] nightly

### Reproducibility

### Miscellaneous information

This request adds easier to use UI elements for user with motor skills impairment oar users with chronic pain.

sangwoo108:
Introduce a profile preference (Appearance → Tabs) to show leading and trailing scroll controls when the horizontal unpinned tab area overflows. Expose the pref through settings WebUI, extension settings_private, and the incognito allowlist where appropriate.

BraveHorizontalTabStripRegionView inserts TabStripControlButton instances with flex-friendly child ordering and visibility tied to overflow. Scroll offset changes are observed via BraveTabContainer’s RegisterHorizontalScrollOffsetChangedCallback subscription so the controls enable/disable correctly at the scroll extents (not only on layout or tab model events).

Update brave_tab_prefs, BraveTabStrip helpers, localized strings, and browser tests for the new behavior.

Resolves https://github.com/brave/brave-browser/issues/54307


browser/ui/views/frame/brave_tab_strip_region_view.cc
```diff
@@ -35,8 +44,109 @@ ui::DropTargetEvent ConvertRootLocation(views::View* view,
...
+
+  const bool show = container && container->ShouldShowHorizontalScrollButton();
+  tab_scroll_previous_button_->SetVisible(show);
+  tab_scroll_next_button_->SetVisible(show);
+  if (show) {
+    tab_scroll_previous_button_->SetEnabled(container->CanScrollTabsStart());
+    tab_scroll_next_button_->SetEnabled(container->CanScrollTabsEnd());
+  }
+}
+
```
netzenbot, line 120: nit: `container` from `strip->GetBraveTabContainer()` is dereferenced directly here and in `OnScrollNextPressed` without a null check, while `UpdateScrollButtonsVisibility` guards with `container &&`. In practice `BraveTabStrip` always has a `BraveTabContainer`, but the inconsistency is worth fixing for defensive correctness per [CSM-029](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-memory.md#CSM-029).

browser/ui/views/tabs/brave_tab_container.h
```diff
@@ -135,6 +161,12 @@ class BraveTabContainer : public TabContainerImpl,
                            MaxScrollOffsetZeroWithPinnedAndUnpinnedTab);
   FRIEND_TEST_ALL_PREFIXES(HorizontalScrollableTabStripBrowserTest,
                            AddingNewTabShouldScrollToBeVisible);
+  FRIEND_TEST_ALL_PREFIXES(HorizontalScrollableTabStripBrowserTest,
+                           HorizontalScrollButtonsVisibleWhenPrefOnAndOverflow);
```
netzenbot: nit: Of the three new `FRIEND_TEST_ALL_PREFIXES` entries, only `HorizontalScrollButtonsVisibleWhenPrefOnAndOverflow` accesses the private `scroll_offset_` member directly. The other two tests only call public methods and don't need friend access. Per [TI-011](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-011), consider a `scroll_offset_for_testing()` accessor for the one case that needs private access, and drop the unnecessary declarations for the other two.

browser/ui/views/frame/brave_tab_strip_region_view.cc
sangwoo108: This should be never nullptr.

browser/ui/views/tabs/brave_tab_container_browsertest.cc
```diff
@@ -273,3 +281,92 @@ IN_PROC_BROWSER_TEST_F(HorizontalScrollableTabStripBrowserTest,
   EXPECT_EQ(1, model->active_index())
```
netzenbot, line 284: nit: The tests cover the pref guard (on/off) and the vertical-tab guard, but not the overflow guard — there's no test that enables the pref, skips the tab-flooding loop, and asserts the buttons are still hidden when `GetMaxScrollOffset() == 0`. Per [TI-026](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-026), both sides of each guard condition should be covered.

browser/ui/views/tabs/brave_tab_container_browsertest.cc
```diff
@@ -273,3 +281,92 @@ IN_PROC_BROWSER_TEST_F(HorizontalScrollableTabStripBrowserTest,
...
+
+  // Grow overflow with scroll buttons off first (same pattern as
+  // HorizontalScrollButtonsHiddenWhenPrefOff). Adding many tabs while the
+  // scroll-button chrome is visible can change active-tab sizing so much that
+  // the strip no longer reports overflow, which makes ScrollTabsBy a no-op.
+  browser()->profile()->GetPrefs()->SetBoolean(
+      brave_tabs::kShowHorizontalTabScrollButtons, false);
+  while (container->max_scroll_offset_for_testing() == 0) {
+    AppendTab();
+    StopAnimatingAndLayout();
```
netzenbot, line 347: nit: The tests verify button visibility but not enabled state. `CanScrollTabsStart()` and `CanScrollTabsEnd()` are new, and the buttons being disabled at the scroll ends is an observable side effect that isn't covered. Consider adding assertions on `GetEnabled()` for the leading/trailing buttons after scrolling to each extreme. Per [TI-039](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-039).

netzenbot (approved):

browser/ui/views/frame/brave_tab_strip_region_view.cc
```diff
@@ -48,30 +163,45 @@ void BraveHorizontalTabStripRegionView::Layout(PassKey) {
...
+                    GetLayoutConstant(LayoutConstant::kToolbarDividerSpacing);
+      new_tab_button_->SetBoundsRect(gfx::Rect(gfx::Point(x, 0), button_size));
     }
     return;
   }
 
   // in vertical tabs mode, we make tab strip's height is the same with this
   // view's height to avoid extra gaps.
   tab_strip_->SetBoundsRect(gfx::Rect(0, 0, width(), height()));
+  UpdateScrollButtonsVisibility();
```
simonhong: Maybe we can call this once at the start of this method?

browser/ui/views/frame/brave_tab_strip_region_view.cc
```diff
@@ -35,8 +44,114 @@ ui::DropTargetEvent ConvertRootLocation(views::View* view,
...
+      l10n_util::GetStringUTF16(IDS_ACCNAME_TAB_SCROLL_TRAILING));
+
+  tab_scroll_previous_button_->SetVisible(false);
+  tab_scroll_next_button_->SetVisible(false);
+}
+
+void BraveHorizontalTabStripRegionView::UpdateScrollButtonsVisibility() {
+  if (!tab_scroll_previous_button_ || !tab_scroll_next_button_) {
+    return;
+  }
```
simonhong, line 126: How about having another method something like `CanUpdateScrollButtons()` ?

We have many duplicated check code and one is checking both and the other is checking one.
Having that method could avoid confusing also.
```
 if (!tab_scroll_previous_button_ || !tab_scroll_next_button_) {
    return;
  }
```
or
```
 if (!tab_scroll_next_button_) {
    return;
  }
```

browser/ui/views/frame/brave_tab_strip_region_view.cc
```diff
@@ -103,6 +233,53 @@ void BraveHorizontalTabStripRegionView::UpdateTabStripMargin() {
   tab_strip_->SetProperty(views::kMarginsKey, margins);
 }
 
+void BraveHorizontalTabStripRegionView::UpdateTrailingScrollButtonMargin(
```
simonhong, line 256: Curious this method need to know vertical tab condition?
I think we can call when `!vertical_tabs`?
Or need to clear margin explicitely even its invisible?

browser/ui/views/frame/brave_tab_strip_region_view.cc
```diff
@@ -35,8 +44,114 @@ ui::DropTargetEvent ConvertRootLocation(views::View* view,
 
 }  // namespace
 
-BraveHorizontalTabStripRegionView::~BraveHorizontalTabStripRegionView() =
-    default;
+BraveHorizontalTabStripRegionView::~BraveHorizontalTabStripRegionView() {}
```
simonhong: q) any reason to change this?

browser/ui/views/tabs/brave_tab_container.cc
```diff
@@ -1797,5 +1813,69 @@ bool BraveTabContainer::IsPinned(const Tab* tab) const {
   return tab_index && *tab_index < pinned_tab_count;
 }
 
+void BraveTabContainer::OnHorizontalTabScrollButtonsPrefChanged() {
+  PreferredSizeChanged();
+  InvalidateLayout();
+  if (views::View* p = parent()) {
+    p->InvalidateLayout();
+  }
```
simonhong: I think we don't need to call all these InvalidateLayout() because PreferredSizeChanged() does it already?

browser/ui/views/tabs/brave_tab_container.cc
```diff
@@ -99,13 +99,18 @@ BraveTabContainer::BraveTabContainer(
...
     scrollable_horizontal_tab_strip_.Init(
         brave_tabs::kScrollableHorizontalTabStrip, prefs,
         base::BindRepeating(
             &BraveTabContainer::OnScrollableHorizontalTabStripPrefChanged,
             base::Unretained(this)));
+    show_horizontal_tab_scroll_buttons_.Init(
+        brave_tabs::kShowHorizontalTabScrollButtons, prefs,
+        base::BindRepeating(
+            &BraveTabContainer::OnHorizontalTabScrollButtonsPrefChanged,
+            base::Unretained(this)));
```
simonhong: I think region view is more proper place for monitoring this prefs change instead of container.
because region view controls scroll buttons. WDYT?

simonhong:
please check with RTL mode also. Run with `--force-ui-direction=rtl`

browser/ui/views/frame/brave_tab_strip_region_view.cc
sangwoo108: Nope, it's left over . Thanks for catching.

browser/ui/views/frame/brave_tab_strip_region_view.cc
sangwoo108: Great. We can dedupe the calls inside the branch above.

browser/ui/views/frame/brave_tab_strip_region_view.cc
sangwoo108: Yes, in order to clear the margin.

browser/ui/views/tabs/brave_tab_container.cc
sangwoo108: Sounds good.

browser/ui/views/tabs/brave_tab_container.cc
sangwoo108: Oh yes, good point.

sangwoo108:
THanks, checking RTL

sangwoo108:
Flipped icon for buttons in RTL

browser/ui/views/frame/brave_tab_strip_region_view.cc
```diff
@@ -38,24 +51,161 @@ ui::DropTargetEvent ConvertRootLocation(views::View* view,
...
+  tab_scroll_previous_button_->SetVisible(false);
+  tab_scroll_next_button_->SetVisible(false);
+}
+
+void BraveHorizontalTabStripRegionView::
+    OnShowHorizontalTabScrollButtonsChanged() {
+  InvalidateLayout();
+  auto* strip = views::AsViewClass<BraveTabStrip>(tab_strip_);
+  CHECK(strip);
+  strip->InvalidateTabContainerLayout();
```
simonhong: Can you add detailed comment about why need to invalidate tab container layout explicitely from here?
and needs to call this class' InvalidateLayout() explicitely? I thought it's called from child view's invalidation.

simonhong (approved):
++

browser/ui/views/frame/brave_tab_strip_region_view.cc
sangwoo108: Thanks, I gave it a second thought and we can remove this. As the container's size would be changed, the ideal bounds of tabs would be regenerated.

browser/ui/views/frame/brave_tab_strip_region_view.cc
```diff
@@ -150,6 +346,30 @@ void BraveHorizontalTabStripRegionView::Initialize() {
...
+
+    auto* bwi = tab_strip_->GetBrowserWindowInterface();
+    CHECK(bwi);
+    PrefService* prefs = bwi->GetProfile()->GetPrefs();
+    CHECK(prefs);
+    show_horizontal_tab_scroll_buttons_.Init(
+        brave_tabs::kShowHorizontalTabScrollButtons, prefs,
+        base::BindRepeating(&BraveHorizontalTabStripRegionView::
+                                OnShowHorizontalTabScrollButtonsChanged,
+                            base::Unretained(this)));
```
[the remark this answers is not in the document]
sangwoo108: safe to have unretainef here, as it's RAII style pref registrar

mkarolin (approved):
`strings`++

browser/ui/views/frame/brave_tab_strip_region_view.cc
```diff
@@ -38,24 +51,158 @@ ui::DropTargetEvent ConvertRootLocation(views::View* view,
...
   if (!tabs::utils::ShouldShowBraveVerticalTabs(
           tab_strip_->GetBrowserWindowInterface())) {
     LayoutSuperclass<HorizontalTabStripRegionView>(this);
 
-    // Ensure that the new tab button is positioned after the last tab, with the
-    // correct amount of padding.
-    if (new_tab_button_) {
-      new_tab_button_->SetX(
-          tab_strip_->bounds().right() +
-          GetLayoutConstant(LayoutConstant::kTabStripPadding));
```
simonhong, line 58: @sangwoo108  I think this change made regression when scroll button is hidden)
Latest master doesn't have padding between last tab and newtab button.
Image

sangwoo108: Thanks, filed here https://github.com/brave/brave-browser/issues/54971

PR #35478, opened 2026-04-14
Re-check PiP icon visibility when document load is complete

### LINKED ISSUE #48679: sometimes the `PiP` icon doesn't appear within the URL bar/omnibox
### Description

Sometimes the `PiP` icon doesn't appear within the URL bar/omnibox. I've only seen this happen ~twice when restoring a session and a YT video is set as the active tab. Refreshing the page fixes the issue.

### Steps to reproduce

Quick notice that the following isn't very easy to reproduce.

1. launch `1.83.76 Chromium: 140.0.7339.26` (with both  https://github.com/brave/brave-core/pull/30548 & https://github.com/brave/brave-core/pull/30578)
2. once launched, visit YT and select any videoYou'll notice that sometimes the `PiP` icon doesn't appear within the URL bar/omnibox. However, this seems to be pretty rare and I've only seen it happen ~twice.

### Actual result

Image

### Expected result

Image

### Reproduces how often

Easily reproduced

### Brave version

```
Brave | 1.83.76 Chromium: 140.0.7339.26 (Official Build) canary (64-bit)
--- | ---
Revision | 21db11da15145499c9eb15e549c04a3c52bd3f02
```

### Device

- Brand/model: `Samsung S25 Ultra`
- Android version: `Android 15`

------------

- Brand/model: `Pixel 6`
- Android version: `Android 16`

### Channel information

- [ ] release (stable)
- [x] beta
- [x] nightly

### Reproducibility

### Miscellaneous information

_No response_

simoarpe:
## Resolves https://github.com/brave/brave-browser/issues/48679

Fix an intermittent bug where the YouTube PiP icon doesn't appear in the toolbar.

The root cause was a timing gap between navigation commit and document load completion. 
`showYouTubePipIcon` is only called from `onDidFinishNavigationInPrimaryMainFrame`, but the native `IsPictureInPictureAvailable()` checks `IsDocumentOnLoadCompletedInPrimaryMainFrame()`, which is still false at navigation commit time. By the time the document load completes (and the check would return true), there was no callback to re-evaluate the icon visibility.

## Test plan

1. Navigate to a YouTube video and verify PiP icon appears.
2. Restore a session with a YouTube video as the active tab and verify PiP icon appears.
3. Navigate away from YouTube and verify PiP icon disappears.
4. Disable Brave Shields while on a YouTube video and verify PiP icon remains visible.

### DIFF
--- a/android/java/org/chromium/chrome/browser/toolbar/top/BraveToolbarLayoutImpl.java
+++ b/android/java/org/chromium/chrome/browser/toolbar/top/BraveToolbarLayoutImpl.java
@@ -568,6 +568,14 @@ public abstract class BraveToolbarLayoutImpl extends ToolbarLayout
                                     && !mBraveShieldsHandler.isShowing()) {
                                 checkForTooltip(tab);
                             }
+
+                            // Re-check PiP icon visibility now that the document
+                            // load is complete. The earlier check in
+                            // onDidFinishNavigationInPrimaryMainFrame may have
+                            // returned false because
+                            // IsDocumentOnLoadCompletedInPrimaryMainFrame was not
+                            // yet true at navigation commit time.
+                            showYouTubePipIcon(tab);
                         }
 
                         String countryCode = Locale.getDefault().getCountry();


SergeyZhukovsky (approved):
++

PR #35484, opened 2026-04-14
Enable Brave Origin by default and add chrome://flags entry

bbondy:
## Summary
- Enable the `kBraveOrigin` feature flag by default on all platforms (previously only enabled for branded builds)
- Add a `chrome://flags#brave-origin` entry on desktop and Android so the feature can be toggled for testing or reverted

## Test plan
- [ ] Verify `chrome://flags#brave-origin` is visible and defaults to Enabled on desktop and Android
- [ ] Toggling the flag to Disabled hides the Origin settings page
- [ ] Branded builds are unaffected


SergeyZhukovsky (approved):
++

netzenbot (approved):

PR #35522, opened 2026-04-15
Fix mojo event loss in createInterfaceApi QueryObserver subscribe

### LINKED ISSUE #54502: AI Chat: "Current Tab" button in attachment menu fails to appear when page context flag is disabled
### Description

When the `brave://flags/#brave-ai-chat-web-content-association-default` flag is set to **Disabled**, the "Current tab contents" button in the AI Chat attachment menu fails to appear approximately 80-90% of the time on initial sidebar open. A hard refresh (Cmd+Shift+R / Ctrl+Shift+R) makes it appear.

The root cause is a race condition in the `createInterfaceApi` event subscription system (`components/common/api/create_interface_api.ts`). When a mojo event (e.g., `OnNewDefaultConversation`) fires before the React `useEffect` hook subscribes to it, the event data is stored in the Tanstack Query cache. However, when the `QueryObserver` is later created and `.subscribe()` is called, the observer's constructor already reads the cached data into `#currentResult`. The subsequent `onSubscribe → updateResult()` detects no change via `shallowEqualObjects` and **silently drops the notification** — the subscriber callback never fires.

This means `defaultTabContentId` is never set in React state, and the "Current tab contents" button never appears.

### Steps to reproduce

1. Set `brave://flags/#brave-ai-chat-web-content-association-default` to **Disabled**
2. Navigate to any webpage (e.g., a YouTube video)
3. Open the AI Chat sidebar
4. Click the attachment button (paperclip icon)
5. Observe the "Current tab contents" option is missing
6. Press Cmd+Shift+R (or Ctrl+Shift+R) to hard-refresh the page
7. Open AI Chat sidebar again and click attachment button
8. "Current tab contents" now appears

### Actual result

The "Current tab contents" button is missing from the attachment menu ~80-90% of the time on initial sidebar open.

### Expected result

The "Current tab contents" button should always appear in the attachment menu when a tab is active, regardless of the flag state.

### Reproduces how often

Intermittent issue

Current nightly

### Channel information

- [x] nightly

### Reproducibility

_No response_

### Miscellaneous information

**Root cause:** `QueryObserver` subscribe deduplication in `components/common/api/create_interface_api.ts`. The `subscribeToMyEvent` function creates a `QueryObserver` whose constructor eagerly reads cached data. When `.subscribe(callback)` is called, `onSubscribe → updateResult()` finds no change and skips the notification. Events that arrive before the React effect mounts are silently lost.

**Fix:** After calling `.subscribe()`, check if the handler was invoked synchronously. If not, check `observer.getCurrentResult()` for pre-existing data and manually deliver it to the handler. This is a general fix in `create_interface_api.ts` that could affect other mojo event subscriptions that had similar silent failures.

thypon:
Resolves https://github.com/brave/brave-browser/issues/54502

## Summary

- Fix a race condition in `createInterfaceApi` where mojo events arriving before React `useEffect` subscribes are silently dropped by Tanstack Query's `QueryObserver` deduplication logic
- After `.subscribe()`, check if the handler was invoked synchronously; if not, manually deliver any pre-existing cached event data

## Root Cause

When a mojo event (e.g., `OnNewDefaultConversation`) fires before the React component mounts and subscribes via `useEffect`, the event data is stored in the Tanstack Query cache by `emitEvent`. However, when the `QueryObserver` is later created in `subscribeToMyEvent`, its constructor eagerly reads the cached data into `#currentResult`. The subsequent `.subscribe()` call triggers `onSubscribe → updateResult()`, which compares the new result against `#currentResult` via `shallowEqualObjects` — finds no change — and **never invokes the listener callback**.

This causes the AI Chat "Current tab contents" button in the attachment menu to fail to appear ~80-90% of the time when `brave://flags/#brave-ai-chat-web-content-association-default` is Disabled, because `defaultTabContentId` is never set in React state.

## Fix

After calling `.subscribe()`, track whether the handler was called synchronously. If it wasn't (the dedup case), check `observer.getCurrentResult()` for pre-existing data and manually invoke the handler. This ensures events that arrived before subscription are always delivered.

### DIFF
--- a/components/common/api/create_interface_api.test.tsx
+++ b/components/common/api/create_interface_api.test.tsx
@@ -960,4 +960,121 @@ describe('createInterfaceApi', () => {
     expect(secondHookResult.result.current.data).toEqual(['5', 5])
     expect(secondHookResult.result.current.hasEmitted).toBe(true)
   })
+
+  it('delivers pre-existing event data to late subscribers', () => {
+    const api = createInterfaceApi({
+      actions: {},
+      endpoints: {},
+      events: {
+        myEvent: event<[], [string, number]>(() => {}),
+      },
+    })
+
+    // Emit event BEFORE any subscriber exists — this is the scenario where
+    // the QueryObserver constructor reads cached data into #currentResult
+    // and .subscribe() skips notification due to shallowEqualObjects.
+    api.emitEvent('myEvent', ['hello', 42])
+
+    // Now subscribe after the event was already emitted
+    const observer = jest.fn()
+    const unsubscribe = api.subscribeToMyEvent(observer)
+
+    // The subscriber should receive the pre-existing data immediately
+    expect(observer).toHaveBeenCalledTimes(1)
+    expect(observer).toHaveBeenCalledWith('hello', 42)
+
+    // Subsequent events should still work normally
+    api.emitEvent('myEvent', ['world', 99])
+    expect(observer).toHaveBeenCalledTimes(2)
+    expect(observer).toHaveBeenLastCalledWith('world', 99)
+
+    unsubscribe()
+  })
+
+  it('delivers pre-existing void event data to late subscribers', () => {
+    const api = createInterfaceApi({
+      actions: {},
+      endpoints: {},
+      events: {
+        myVoidEvent: event<[], []>(() => {}),
+      },
+    })
+
+    // Emit void event before subscriber exists
+    api.emitEvent('myVoidEvent', [])
+
+    const observer = jest.fn()
+    const unsubscribe = api.subscribeToMyVoidEvent(observer)
+
+    // Late subscriber should still be notified of the pre-existing event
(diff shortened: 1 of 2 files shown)


petemill:
Needs a unit test please

petemill (commented):
My issue with this proposal is it changes the nature of `api.useOnMyEvent()` whereby it will call a new handler immediately, with the latest emitted event, if the event has previously been emitted _at any time_. This will affect any subscriptions at any point during the app lifetime, and not just in-between mojo subscription and React mounting.

It _is_ an issue to look out for but I think, in this issue at least, it is solved with the existing `api.useCurrentOnMyEvent()` which does similar: provides the latest emitted event and updates with any future emitted event. That fix is coincidentally provided with https://github.com/brave/brave-core/pull/35540

netzenbot (approved):

thypon:
Closing it since it looks fixed with another commit

PR #35536, opened 2026-04-15
Force Origin onboarding to use dark theme buttons

### LINKED ISSUE #54514: [Origin] Force desktop onboarding to always load screens in dark mode
### Description

Since the onboarding is dark in styles, we need to force the onboarding screens for origin to also load always in dark mode to avoid contrast issues.

The problems are specifically noticeable when you are in the import screen, which has a disabled button, that is unreadable when it loads in light mode.

### Actual result
See how the browser is in light mode and loads the onboarding using light mode

Image

Image

Image

### Expected result
See how the browser is in light mode and loads the onboarding using dark mode

Image

Image

Image

### Reproduces how often

Easily reproduced

-

### Channel information

- [ ] release (stable)
- [ ] beta
- [x] nightly

### Reproducibility

### Miscellaneous information

_No response_

aguscruiz:
Resolves https://github.com/brave/brave-browser/issues/54514

## Summary

Brave Origin onboarding uses light hero imagery. Leo buttons and brave-ui / `styled-components` theming still followed the **system** color scheme, so on a light OS theme controls stayed in the **light** palette and looked wrong on the art.

For **`is_brave_origin_branded`** builds only, the welcome WebUI forces **dark** theming from the app root.

## Approach

Per review, this replaces per-screen `data-theme` wrappers with a **single root-level** control:

1. **`BraveCoreThemeProvider`** (`components/common/BraveCoreThemeProvider.tsx`)
   - New optional prop: **`forceDarkMode`**.
   - When `true`:
     - Always selects the **dark** brave-ui theme for `styled-components` (ignores `prefers-color-scheme`).
     - Skips registering media-query / visibility listeners for theme toggling.
     - Uses **`useLayoutEffect`** to set **`document.documentElement.dataset.theme = 'dark'`** so **Leo** components resolve dark tokens (with cleanup that restores the previous `data-theme` when the provider unmounts or `forceDarkMode` turns off). No extra DOM wrapper around the tree.

2. **`brave_welcome.tsx`**
   - Uses the usual `// <if expr="is_brave_origin_branded">` preprocessor pattern **on `BraveCoreThemeProvider`**: only Origin builds include the JSX boolean shorthand **`forceDarkMode`** (equivalent to `forceDarkMode={true}`). Standard Brave omits the prop so onboarding still follows **system** light/dark.

3. **Removed** the previous `OriginLeoButtonTheme` helper and all per-component wrappers; welcome screens use plain `Button` again.

**Review follow-ups (latest push):** no intermediate `forceDarkMode` variable in `App`; no unrelated indentation change in `select-browser`.

## How to test

1. Build **Brave Origin** (`is_brave_origin_branded`).
2. Open welcome / onboarding with the OS in **light** mode.
3. Confirm Leo buttons (and any brave-ui themed bits) match **dark** styling across steps (welcome, import, theme, help improve, etc.).
4. Build **standard Brave**: onboarding should still follow **system** light/dark (no forced dark).

## Screenshots

Browser in light mode with onboarding using dark controls:

Image

Image

Image


Douglashdaniel (commented):
Hey, instead of having all these individual wrappers, would adding this to the root work?
https://github.com/brave/brave-core/blob/master/components/brave_welcome_ui/brave_welcome.tsx

```
function App() {
  return (
    <DataContextProvider>
      <BraveCoreThemeProvider
        // <if expr="is_brave_origin_branded">
        forceDarkMode
        // </if>
      >
        <MainContainer />
      </BraveCoreThemeProvider>
    </DataContextProvider>
  )
}
```

components/common/BraveCoreThemeProvider.tsx
```diff
@@ -36,14 +42,24 @@ export default function LightDarkThemeProvider (props: React.PropsWithChildren<P
...
   return (
     <ThemeProvider theme={selectedTheme}>
-      {React.Children.only(props.children)}
+      {props.forceDarkMode ? (
+        <span data-theme='dark' style={{ display: 'contents' }}>
+          {child}
+        </span>
+      ) : (
+        child
+      )}
```
Douglashdaniel: We should avoid wrapping the entire UI in a `span` should just return `{React.Children.only(props.children)}` here, perhaps we can instead set the theme in a `useLayoutEffect`?

```
  React.useLayoutEffect(() => {
    if (!forceDarkMode) {
      return
    }
    const html = document.documentElement
    const previous = html.dataset.theme
    html.dataset.theme = 'dark'
    return () => {
      if (previous === undefined) {
        delete html.dataset.theme
      } else {
        html.dataset.theme = previous
      }
    }
  }, [forceDarkMode])
```

components/brave_welcome_ui/components/select-browser/index.tsx
```diff
@@ -168,14 +168,14 @@ function SelectBrowser () {
...
-            {getLocale('braveWelcomeImportButtonLabel')}
-          </Button>
+        <Button
+          kind="filled"
+          isDisabled={!hasSelectedBrowser}
+          onClick={handleImport}
+          size="large"
+        >
+          {getLocale('braveWelcomeImportButtonLabel')}
+        </Button>
```
Douglashdaniel: Can we undo this?

components/brave_welcome_ui/brave_welcome.tsx
```diff
@@ -15,9 +15,16 @@ import MainContainer from './main_container'
 import DataContextProvider from './state/data-context-provider'
 
 function App () {
+  let forceDarkMode: boolean
+  // <if expr="is_brave_origin_branded">
+  forceDarkMode = true
+  // <else>
+  forceDarkMode = false
+  // </if>
+
```
Douglashdaniel: Shouldn't need to declare this.

components/brave_welcome_ui/brave_welcome.tsx
```diff
@@ -15,9 +15,16 @@ import MainContainer from './main_container'
...
   return (
     <DataContextProvider>
-      <BraveCoreThemeProvider>
+      <BraveCoreThemeProvider forceDarkMode={forceDarkMode}>
```
Douglashdaniel: Can we just do this instead

```suggestion
      <BraveCoreThemeProvider
        // <if expr="is_brave_origin_branded">
        forceDarkMode
        // </if>
       >
```

components/brave_welcome_ui/brave_welcome.tsx
```diff
@@ -17,7 +17,11 @@ import DataContextProvider from './state/data-context-provider'
 function App () {
   return (
     <DataContextProvider>
-      <BraveCoreThemeProvider>
+      <BraveCoreThemeProvider
+        // <if expr="is_brave_origin_branded">
+        forceDarkMode
```
fallaciousreasoning: I don't think this exists - we probably don't need this anyway

fallaciousreasoning: oh wait, I see - this context is just for the old brave-ui, not Leo

components/common/BraveCoreThemeProvider.tsx
```diff
@@ -36,11 +42,29 @@ export default function LightDarkThemeProvider (props: React.PropsWithChildren<P
...
+      } else {
+        html.dataset.theme = previous
+      }
+    }
+  }, [props.forceDarkMode])
+
+  const selectedTheme =
+    props.forceDarkMode || isDarkMode
+      ? props.dark || DefaultDarkTheme
+      : props.light || DefaultTheme
```
fallaciousreasoning: I think we should just do the `is_origin` check here

fallaciousreasoning (approved):
lgtm % formatting

petemill (approved):

Douglashdaniel (approved):

PR #35537, opened 2026-04-15
[toolchain] `fetch` the minimum necessary

cdesouza-chromium:
This change adds `--nohistory`, and `--nohooks` to `fetch`, to make
cloning Chromium leaner.

Bug: https://github.com/brave/brave-browser/issues/54478

### DIFF
--- a/tools/rust/build_rust_toolchain_standalone.py
+++ b/tools/rust/build_rust_toolchain_standalone.py
@@ -438,7 +438,11 @@ class ToolchainBuilder:
         self._bootstrap_depot_tools()
 
         self.chromium_src.parent.mkdir(parents=True, exist_ok=True)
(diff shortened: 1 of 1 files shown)


goodov (approved):

netzenbot (approved):

cdesouza-chromium:
I'm holding from merging this PR yet because I'm not sure we get better performance at the end. I need to measure this.

goodov:
> I'm holding from merging this PR yet because I'm not sure we get better performance at the end. I need to measure this.

`--nohistory` shouldn't matter, because we use git cache. It only make sense if you fetch from git repo directly.

`--nohooks` can speedup things, that's true, but it also may prevent correct toolchain building, because some things may be missing.

cdesouza-chromium:
This PR has been superseded for another one.

PR #35540, opened 2026-04-15
Fix current tab contents button not showing on initial side panel open

### LINKED ISSUE #54520: [AI Chat] "Current Tab contents" button only appears in attachment menu after reload/navigation when page context flag is disabled
### Description

When the "brave://flags/#brave-ai-chat-web-content-association-default" flag is disabled, the "Current Tab contents" button should be visible in the Leo AI attachment menu immediately. However, the button doesn't appear initially, it only shows up after reloading the page or clicking a link while the Leo AI side panel is open.

### Steps to reproduce

1. Disable this flag: brave://flags/#brave-ai-chat-web-content-association-default
2. Go to any site (eg: youtube.com)
3. Open Leo AI on the side panel
4. Click the attachment button ( "Current Tab contents" button is not shown)
5. Reload the site, or click any link on the page while Leo AI side panel is open
6. "Current Tab contents" is shown now

### Actual result

The "Current Tab contents" button does not appear in the attachment menu on initial page load, even though the flag is disabled. It only becomes visible after a page reload or navigation event.

https://github.com/user-attachments/assets/d7d41d0a-4ca8-46f9-a0c4-38f16ca1b679

### Expected result

The "Current Tab contents" button should be visible in the Leo AI attachment menu immediately upon opening it, without requiring a page reload or navigation, when the brave://flags/#brave-ai-chat-web-content-association-default flag is disabled.

### Reproduces how often

Easily reproduced

Brave | 1.89.132 Chromium: 147.0.7727.56 (Official Build) (arm64)
Revision | d491a927d068805831350ec9d7e5afda454e83ce
OS | macOS Version 26.3.1 (a) (Build 25D771280a)

### Channel information

- [x] release (stable)
- [x] beta
- [x] nightly

### Reproducibility

### Miscellaneous information

_No response_

shisirbasel:
Fix https://github.com/brave/brave-browser/issues/54520

### DIFF
--- a/components/ai_chat/resources/page/state/ai_chat_context.tsx
+++ b/components/ai_chat/resources/page/state/ai_chat_context.tsx
@@ -51,12 +51,6 @@ export default function useProvideAIChatContext(props: AIChatContextProps) {
   const { getConversationsData, isPlaceholderData: isConversationsLoading } =
     api.useGetConversations()
 
-  const [defaultTabContentId, setDefaultTabContentId] = React.useState<number>()
-
-  api.useOnNewDefaultConversation((contentId) => {
-    setDefaultTabContentId(contentId)
-  })
-
   const store = {
     api: props.api,
     initialized:
@@ -163,7 +157,7 @@ export default function useProvideAIChatContext(props: AIChatContextProps) {
      */
     openURL: api.uiHandler.openURL,
 
-    defaultTabContentId,
+    defaultTabContentId: api.useCurrentOnNewDefaultConversation().data?.[0],
     editingConversationId,
     setEditingConversationId,
     deletingConversationId,


components/ai_chat/resources/page/api/ai_chat_api.ts
```diff
@@ -187,8 +189,9 @@ export default function createAIChatApi(
       ...eventsFor(
         Mojom.ChatUIInterface,
         {
-          // We don't need to do any handling here, just pass through
-          onNewDefaultConversation(contentId: number) {},
+          onNewDefaultConversation(contentId: number) {
+            api.defaultTabContentId.update(contentId)
+          },
```
petemill: We don't need a state type purely to pass through the "current" event arg. We can simply have interested components call `api.useCurrentOnNewDefaultConversation()` to get the latest event call data, and this file needs no changes from master.

components/ai_chat/resources/page/state/ai_chat_context.tsx
```diff
@@ -163,7 +157,7 @@ export default function useProvideAIChatContext(props: AIChatContextProps) {
      */
     openURL: api.uiHandler.openURL,
 
-    defaultTabContentId,
+    defaultTabContentId: api.useDefaultTabContentId().data,
```
petemill: Since we always want the current value, this could simply be `api.useCurrentOnNewDefaultConversation()`

petemill (changes requested):
This solution somewhat conflicts with https://github.com/brave/brave-core/pull/35522 but I think this simple fix, with the suggested simplification, will solve the issue. There is another call to `useOnNewDefaultConversation`, in `active_chat_context.tsx` but I don't think we need to change that because it needs to be an event and is less susceptible to a race condition - there would need to be something else happening (the target WebContents would have to be mid-navigation when the panel is opened).

netzenbot (approved):

petemill (approved):

PR #35541, opened 2026-04-15
[Polkadot] [Wallet] add transaction status tracking task

### LINKED ISSUE #54523: [Polkadot] [Wallet] Add transaction status tracking routines
We need a way to probe the block chain for the status of an extrinsic. This involves crawing the block chain within the specified mortality window for an extrinsic and searching each block's `extrinsics` array. The array index where the extrinsic is found is what's used during event parsing, which gives us the ultimate information about whether or not the extrinsic was successful or not and what the true fee used by the chain was.

In addition to success or failure states, an extrinsic can also be entirely absent from the block chain in which case we consider it to be dropped.

cmazakas:
Resolves https://github.com/brave/brave-browser/issues/54523

Add a PolkadotTransactionStatusTask that can be used to crawl the block chain, probing each block for the provided extrinsic. Assuming the extrinsic is found, we then parse the events of the block to determine the final status of the extrinsic along with the true fee withdrawn from the sender's account.

Status tracking can yield 4 different states. If we start crawling too quickly and outpace the finalized head, we return early. If we probe the entire mortality window of the extrinsic, we inform the caller that the extrinsic was not found and should be considered dropped. Otherwise, if we find the extrinsic within the mortality window, we can accurately determine if it was successful or failed.


components/brave_wallet/browser/polkadot/polkadot_test_utils.cc
```diff
@@ -440,4 +489,126 @@ bool PolkadotMockRpc::HandlePaymentInfoRequest(
   return false;
 }
 
+bool PolkadotMockRpc::HandleBlockHashRequest(
+    const network::ResourceRequest& req,
+    const base::DictValue& req_body) {
```
netzenbot, line 521: `FindString` returns `nullptr` if the key is absent — dereferencing `method` unconditionally here is a null-pointer dereference. The same bug exists in `HandleBlockRequest` (line ~511) and `HandleEventsRequest` (line ~537). Follow the existing pattern used in `HandleAuthorSubmitExtrinsic`: `if (const auto* method = req_body.FindString("method")) { if (*method == "chain_getBlockHash") { ... } }`.

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task.cc
```diff
@@ -0,0 +1,183 @@
...
+      ::rust::Slice<const uint8_t>(events.value()), *extrinsic_idx_, *pubkey,
+      **chain_metadata_, fee);
+
+  std::move(callback_).Run(
+      base::ok(std::pair(was_successful ? PolkadotTransactionStatus::kSuccess
+                                        : PolkadotTransactionStatus::kFailed,
+                         base::bit_cast<uint128_t>(fee))));
+}
+
+}  // namespace brave_wallet
```
netzenbot, line 245: nit: `InitRequests()` calls `GetBlockHash()` whose callback eventually calls `OnGetBlockForStatus()`, which calls `InitRequests()` again — a deliberate re-entrant async loop. Per ARCH-061, if re-entrancy is unavoidable it should be clearly documented. A comment above `InitRequests()` explaining this iterative block-walking pattern would help future readers. ([Avoid Re-entrancy; Localize Control Flow](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-061))

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task.h
```diff
@@ -0,0 +1,99 @@
...
+  kNotFinalized,
+};
+
+class PolkadotTransactionStatusTask {
+ public:
+  using GetTransactionStatusCallback = base::OnceCallback<void(
+      base::expected<std::pair<PolkadotTransactionStatus, uint128_t>,
+                     std::string>)>;
+
+  PolkadotTransactionStatusTask(PolkadotWalletService& polkadot_wallet_service,
```
netzenbot: nit: Per [CSA-042](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-042), prefer `std::ranges` algorithms over manual loops with break conditions. `Position` could be replaced with `auto it = std::ranges::find(r, val); if (it == std::ranges::end(r)) return std::nullopt; return static_cast<size_t>(std::ranges::distance(std::ranges::begin(r), it));`.

components/brave_wallet/browser/polkadot/polkadot_test_utils.cc
```diff
@@ -440,4 +489,126 @@ bool PolkadotMockRpc::HandlePaymentInfoRequest(
...
+  if (const auto* method = req_body.FindString("method");
+      *method == "chain_getBlockHash") {
+    if (const auto* params_list = req_body.FindList("params")) {
+      CHECK(!params_list->empty());
+
+      uint32_t block_num = 0;
+      CHECK(
```
netzenbot, line 528: nit: Use `base::FindOrNull(block_hash_map_, block_num)` instead of the manual `find` + `end()` check. This pattern also appears in `HandleBlockRequest` (where there's additionally a redundant `CHECK` and `if` on the same condition) and `HandleEventsRequest`. Requires adding `#include "base/containers/map_util.h"`. ([Use `base::FindOrNull()` for Map Lookups](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-049))

components/brave_wallet/browser/polkadot/polkadot_test_utils.h
```diff
@@ -27,10 +27,27 @@ struct PolkadotMockRpc {
...
   void RejectAccountInfoRequest();
   void SetSenderPubKey(
       base::span<uint8_t, kPolkadotSubstrateAccountIdSize> pubkey);
+  void SetFinalizedBlockHeader(std::string_view json_str);
+
+  // Used to map requests for a block hash given a block's number.
+  void SetBlockHashMap(base::flat_map<uint32_t, std::string> block_hash_map);
+
+  // Used to map requests for an entire block given its hash.
+  void SetBlockMap(base::flat_map<std::string, PolkadotBlock> block_map);
```
netzenbot, line 42: nit: `SetBadBlockMapKey` is defined inline in the header. Per [CS-013](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-013), method definitions belong in `.cc` files — headers should only contain declarations.

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task.h
```diff
@@ -0,0 +1,100 @@
...
+      std::optional<std::array<uint8_t, kPolkadotBlockHashSize>> block_hash,
+      std::optional<std::string> err_str);
+
+  void OnGetBlockForStatus(std::optional<PolkadotBlock> block,
+                           std::optional<std::string> err_str);
+
+  void OnGetEvents(base::expected<std::vector<uint8_t>, std::string> events);
+
+  void HandleError(std::string err_str);
+
```
netzenbot, line 101: nit: The `extrinsic` constructor parameter is `std::vector<uint8_t>` but the constructor only reads it to produce a hex string. Per [CSA-017](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-017), prefer `base::span<const uint8_t>` for read-only byte buffer parameters — this avoids forcing callers to own a `std::vector` and accepts any contiguous container.

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task.h
```diff
@@ -0,0 +1,100 @@
...
+  base::raw_ref<PolkadotWalletService> polkadot_wallet_service_;
+  base::raw_ref<KeyringService> keyring_service_;
+  std::optional<PolkadotChainMetadata> chain_metadata_;
+
+  mojom::AccountIdPtr sender_account_id_;
+  std::string chain_id_;
+  std::string extrinsic_hex_;
+  uint32_t block_num_;
+  uint32_t curr_block_num_;
```
netzenbot: nit: `block_num_`, `curr_block_num_`, and `mortality_period_` are POD types without default initializers in the header. Per [CSA-063](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-063), add `= 0` to each. (`finalized_block_num_` already does this correctly.)

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task.cc
```diff
@@ -0,0 +1,192 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "brave/components/brave_wallet/browser/polkadot/polkadot_transaction_status_task.h"
```
netzenbot, line 6: nit: `// IWYU pragma: export` has no effect in a `.cc` file; this pragma is only meaningful in headers that re-export includes to their consumers. Remove it. ([Always Include What You Use](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-001))

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task.h
```diff
@@ -0,0 +1,100 @@
...
+  uint32_t mortality_period_;
+  uint32_t finalized_block_num_ = 0;
```
supermassive: `= 0;`

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task.h
```diff
@@ -0,0 +1,100 @@
...
+  size_t i = 0;
+  for (const auto& v : r) {
+    if (v == val) {
+      return i;
+    }
+    ++i;
+  }
+  return std::nullopt;
+}
+
```
supermassive: should be in .cc

components/brave_wallet/browser/polkadot/polkadot_test_utils.h
```diff
@@ -27,10 +27,27 @@ struct PolkadotMockRpc {
...
+
+  // Used to simulate a network failure request for chain_getBlock. Should match
+  // a key present in block_map_.
+  void SetBadBlockMapKey(std::string bad_block_map_key) {
+    bad_block_map_key_ = bad_block_map_key;
```
supermassive: std::move or string_view

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task.cc
```diff
@@ -0,0 +1,192 @@
...
+
+#include "base/strings/strcat.h"  // IWYU pragma: export
```
supermassive: Should be `keep`.
`export` is to be used in headers.

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task.cc
```diff
@@ -0,0 +1,192 @@
...
+  InitRequests();
+}
+
+void PolkadotTransactionStatusTask::InitRequests() {
+  if (curr_block_num_ > finalized_block_num_) {
+    return std::move(callback_).Run(
+        base::ok(std::pair(PolkadotTransactionStatus::kNotFinalized, 0)));
+  }
+
+  if (curr_block_num_ >= (block_num_ + mortality_period_)) {
```
supermassive: Let's use safe math here

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task.cc
```diff
@@ -0,0 +1,192 @@
...
+    return std::move(callback_).Run(
+        base::ok(std::pair(PolkadotTransactionStatus::kNotFound, 0)));
+  }
+
+  GetPolkadotRpc()->GetBlockHash(
+      chain_id_, curr_block_num_,
+      base::BindOnce(&PolkadotTransactionStatusTask::OnGetBlockHashForStatus,
+                     weak_ptr_factory_.GetWeakPtr()));
+
+  ++curr_block_num_;
```
supermassive: - Let's increment just before `GetBlockHash` calls. Just to minimize chances of possible inifinte recursion when `GetBlockHash` synchronously calls `InitRequests` and used same `curr_block_num_`.
- Need some comments in code on what are upper limits of number of `InitRequests()` iteratios. Maybe need some hardcoded limits for this. So it doesn't become 1000 requests or so in production.
- Let's do safe math for increment just in case.

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task.cc
```diff
@@ -0,0 +1,192 @@
...
+    extrinsic_idx_ = idx;
+    GetPolkadotRpc()->GetEvents(
+        chain_id_, block->header.GetHash(),
+        base::BindOnce(&PolkadotTransactionStatusTask::OnGetEvents,
+                       weak_ptr_factory_.GetWeakPtr()));
+
+    return;
+  }
+
+  InitRequests();
```
supermassive: Needs beter name. Somehting like `FetchCurrentBlock`.
Also makes sense to actually do `++curr_block_num_` here

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task_unittest.cc
```diff
@@ -0,0 +1,1206 @@
...
+            "0x0442454546840385b89b5af093f2269fac4180942cd2f3f3814783f33ee572b3c6df4570e21fc1",
+            "0x054241424501017c737c5a03844ffdf92efbe9a8fa2c47f40d9b3b4a040ef99c1ed3ba4529916a59611765c8d78e4d4ba9c8eb4866aa89e5bdf9959583de8a073696134d107782"
+          ]
+        }
+      }
+    }
+  )");
+
+  {
+    base::flat_map<uint32_t, std::string> block_hash_map;
```
cypt4, line 1011: Please add a comment describing what we are doing here

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task_unittest.cc
```diff
@@ -0,0 +1,1206 @@
...
+      *polkadot_wallet_service, *keyring_service_,
+      polkadot_testnet_account_->account_id->Clone(), mojom::kPolkadotTestnet,
+      std::move(extrinsic), block_num, mortality_period);
+
+  base::test::TestFuture<base::expected<
+      std::pair<PolkadotTransactionStatus, uint128_t>, std::string>>
+      future;
+
+  polkadot_mock_rpc->SetFinalizedBlockHeader(R"(
+    {
```
cypt4, line 131: What do you think if we make this more human-readable? For ex. add human-redable methods that return this string constants.
For example - GetValidBlockHeader_0x1ce354f

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task.cc
```diff
@@ -0,0 +1,192 @@
...
+  if (!block_header) {
+    return HandleError(WalletInternalErrorMessage());
+  }
+
+  finalized_block_num_ = block_header->block_number;
+  InitRequests();
+}
+
+void PolkadotTransactionStatusTask::InitRequests() {
+  if (curr_block_num_ > finalized_block_num_) {
```
cypt4, line 146: Please add comments describing what happens here

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task.cc
```diff
@@ -0,0 +1,192 @@
...
+
+  if (!block) {
+    return HandleError(WalletInternalErrorMessage());
+  }
+
+  if (extrinsic_idx_.has_value()) {
+    return;
+  }
+
+  auto idx = Position(block->extrinsics, extrinsic_hex_);
```
cypt4, line 194: FindPosition?

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task.h
cmazakas: Done!

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task.h
cmazakas: Done!

components/brave_wallet/browser/polkadot/polkadot_test_utils.h
cmazakas: I chose `std::move()`, as from one your previous reviews, I like being transparent in taking an existing allocation instead of implicitly performing one.

This is done.

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task.cc
cmazakas: Oh, my bad. I guess that's what blind copy-paste does, huh?

This has been addressed.

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task.cc
cmazakas: Alright, this should all be addressed.
I updated Start() to support only being called once and I also updated the code to now fallibly create task if the provided mortality period is too large. Internally, we're already safe-guarded from infinite looping so it was just about making sure we're not exploitable from the outside due to programmer bug.

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task.cc
cmazakas: I wound up doing exactly this.

components/brave_wallet/browser/polkadot/polkadot_test_utils.cc
```diff
@@ -463,4 +516,126 @@ bool PolkadotMockRpc::HandlePaymentInfoRequest(
   return false;
```
netzenbot, line 516: This null dereference is still present in `HandleBlockHashRequest`, `HandleBlockRequest`, and `HandleEventsRequest`. `FindString` returns `nullptr` when the key is absent — dereferencing `*method` without a prior null check crashes if any request arrives without a `"method"` field. The correct pattern, used by `HandleAuthorSubmitExtrinsic` and `HandlePaymentInfoRequest`, is: `if (const auto* method = req_body.FindString("method")) { if (*method == "chain_getBlockHash") { ... } }`.

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task_unittest.cc
```diff
@@ -0,0 +1,1391 @@
...
+  polkadot_mock_rpc->AddGetFinalizedBlockHeader();
+  polkadot_mock_rpc->FinalizeSetup();
+
+  task->Start(future.GetCallback());
+
+  auto result = future.Take();
+  ASSERT_TRUE(result.has_value());
+
+  const auto [status, fee] = *result;
+  EXPECT_EQ(status, PolkadotTransactionStatus::kNotFinalized);
```
netzenbot, line 831: nit: `extrinsic_hex` is duplicated in every test function with the same value. Per [TI-009](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-009), pull this into a file-scope `constexpr char[]` or a `PolkadotTransactionStatusTaskUnitTest` fixture member to avoid silent divergence if the value ever changes.

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task_unittest.cc
```diff
@@ -0,0 +1,1391 @@
...
+  const uint32_t mortality_period = 64;
+
+  auto task = PolkadotTransactionStatusTask::Create(
+      *polkadot_wallet_service, *keyring_service_,
+      polkadot_testnet_account_->account_id->Clone(), mojom::kPolkadotTestnet,
+      std::move(extrinsic), block_num, mortality_period);
+  ASSERT_TRUE(task);
+
+  base::test::TestFuture<base::expected<
+      std::pair<PolkadotTransactionStatus, uint128_t>, std::string>>
```
netzenbot, line 968: nit: This test is named `FindExtrinsicFailed` but it expects `PolkadotTransactionStatus::kInvalidResponse`, not `kFailed` — the `kFailed` case is actually covered by `InconsistentFees`. A name like `FindExtrinsicWithZeroFeeReturnsInvalidResponse` would match what is actually tested. ([Use Descriptive Test Method Names](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-038))

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task.cc
cmazakas: Done!

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task_unittest.cc
cmazakas: Done! I did this for the majority of the cases. Some had some test-specific logic so I don't pull those out but in general, the test suite is a lot cleaner now.

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task_unittest.cc
cmazakas: Done!

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task.cc
cmazakas: Done!

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task.cc
cmazakas: Done!

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task.h
```diff
@@ -0,0 +1,125 @@
...
+  GetTransactionStatusCallback callback_;
+
+  bool initiated_ = false;
+
+  base::WeakPtrFactory<PolkadotTransactionStatusTask> weak_ptr_factory_{this};
+};
+
+}  // namespace brave_wallet
+
+#endif  // BRAVE_COMPONENTS_BRAVE_WALLET_BROWSER_POLKADOT_POLKADOT_TRANSACTION_STATUS_TASK_H_
```
netzenbot, line 128: nit: `Start()` needs a contract comment — it's the main public API and has non-obvious constraints: single-call-only (enforced by `initiated_`), asynchronous execution, and requires the task to outlive the callback. Per [DOC-004](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-004), public method docs should cover behavior, restrictions, and gotchas.

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task.cc
```diff
@@ -0,0 +1,260 @@
...
+      chain_id_(std::move(chain_id)),
+      extrinsic_hex_(base::StrCat({"0x", base::HexEncodeLower(extrinsic)})),
+      block_num_{block_num},
+      mortality_period_{mortality_period},
+      curr_block_num_{block_num} {
+  // Use saturating addition here for the sake of simplicity. While it's
+  // unlikely, a block produced near the end of the chain can still have an
+  // extrinsic within its mortality window even if the true end of the mortality
+  // window exceeds numeric limits for a uint32_t.
+  max_block_num_ = base::ClampAdd(block_num_, mortality_period_).RawValue();
```
supermassive: Not sure if RawValue() must be there. Picked randomly some uses, and they all seem to cast value automatically

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task.cc
```diff
@@ -0,0 +1,260 @@
...
+        base::BindOnce(&PolkadotTransactionStatusTask::OnGetEvents,
+                       weak_ptr_factory_.GetWeakPtr()));
+
+    return;
+  }
+
+  base::CheckedNumeric<uint32_t> curr_block_num{curr_block_num_};
+  curr_block_num += 1;
+  if (!curr_block_num.AssignIfValid(&curr_block_num_)) {
+    return std::move(callback_).Run(
```
supermassive: It could be a class field
`base::CheckedNumeric<uint32_t> curr_block_num_;`
and checked for validiy only in `FetchCurrentBlock`.

Otherwsie a bit shorter one
``` 
if (!base::CheckAdd(curr_block_num_, 1).AssignIfValid(&curr_block_num_)) {
  ... 
}
```

supermassive (approved):

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task_unittest.cc
```diff
@@ -0,0 +1,988 @@
...
+  polkadot_mock_rpc->FinalizeSetup();
+
+  task->Start(future.GetCallback());
+
+  auto result = future.Take();
+  ASSERT_TRUE(result.has_value());
+
+  const auto [status, fee] = *result;
+  EXPECT_EQ(status, PolkadotTransactionStatus::kFailed);
+  EXPECT_EQ(fee, uint128_t{15937408476});
```
cypt4, line 298: Why we compare fee if task failed?

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task.h
```diff
@@ -0,0 +1,125 @@
...
+// from there is guaranteed to be on-chain within mortality_period blocks
+// relative to the signing block.
+// At a maximum performs mortality_period block fetches.
+// Only finalized blocks are considered.
+class PolkadotTransactionStatusTask {
+ public:
+  using PassKey = base::PassKey<PolkadotTransactionStatusTask>;
+
+  using GetTransactionStatusCallback = base::OnceCallback<void(
+      base::expected<std::pair<PolkadotTransactionStatus, uint128_t>,
```
cypt4, line 48: base::expected<std::pair<PolkadotTransactionStatus, uint128_t> =>
base::expected<std::pair<PolkadotTransactionStatus, std::optional<uint128_t>> ?

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task_unittest.cc
```diff
@@ -0,0 +1,988 @@
...
+  base::test::TestFuture<base::expected<
+      std::pair<PolkadotTransactionStatus, uint128_t>, std::string>>
+      future;
+
+  RegisterDefaultFinalizedHeader(polkadot_mock_rpc.get());
+
+  // For this test, we'll just reuse the same blockhash across all requests and
+  // we'll return an empty block each time. This guarantees the extrinsic won't
+  // be found within the mortality period.
+  {
```
cypt4: Can we reuse FindExtrinsicSucceeds data and just use small mortality period value?

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task_unittest.cc
cmazakas: Hmm, maybe I'm using the wrong enum value then. This is the case where an extrinsic was included in a finalized block but it failed, i.e. ExtrinsicFailed so a fee was incurred from the sender.

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task.h
cmazakas: Hmm. Not a bad idea, actually. Maybe we could use some massaging here. The error arm of this particular `expected` is used for other failures, such as network ones or general RPC parsing errors.

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task.cc
cmazakas: Ah, nice. I wound up going with the `CheckAdd` for now.

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task_unittest.cc
cmazakas: We can! This is a cool exercise because it tests the code in a new way too: using a different mortality period value.

Done!

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task.cc
cmazakas: Removed!

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task.h
cmazakas: Done! Misread the comment originally. Updated code to return an optional fee.

cypt4 (approved):

components/brave_wallet/browser/polkadot/polkadot_transaction_status_task_unittest.cc
cmazakas: Resolved on slack!

netzenbot (approved):

PR #35542, opened 2026-04-15
Add SerpMetrics ObjC bridge

aseren:
Defines an ObjC protocol wrapping the C++ SerpMetrics class so that Swift consumers can read per-engine search counts and clear stored history.

This is a pre-requisite of https://github.com/brave/brave-browser/issues/53912

### DIFF
--- a/ios/BUILD.gn
+++ b/ios/BUILD.gn
@@ -121,6 +121,8 @@ brave_core_public_headers = [
   "//brave/ios/browser/safari_data_import/safari_data_import_client_delegate.h",
   "//brave/ios/browser/safari_data_import/safari_data_importer_bridge.h",
   "//brave/ios/browser/safari_data_import/safari_data_importer_coordinator.h",
+  "//brave/ios/browser/serp_metrics/serp_metrics_bridge.h",
+  "//brave/ios/browser/serp_metrics/serp_metrics_service_factory_bridge.h",
   "//brave/ios/browser/playlist/playlist_pref_names_bridge.h",
   "//brave/ios/browser/ui/l10n_utils_typed_bridge.h",
   "//ui/base/l10n/l10n_util_mac_bridge.h",
--- a/ios/browser/serp_metrics/BUILD.gn
+++ b/ios/browser/serp_metrics/BUILD.gn
@@ -5,6 +5,11 @@
 
 source_set("serp_metrics") {
   sources = [
+    "serp_metrics_bridge.h",
+    "serp_metrics_bridge_impl.h",
+    "serp_metrics_bridge_impl.mm",
+    "serp_metrics_service_factory_bridge.h",
+    "serp_metrics_service_factory_bridge.mm",
     "serp_metrics_service_factory_ios.h",
     "serp_metrics_service_factory_ios.mm",
     "serp_metrics_tab_helper.h",
@@ -21,6 +26,7 @@ source_set("serp_metrics") {
     "//brave/components/serp_metrics",
     "//brave/components/serp_metrics:features",
     "//brave/components/serp_metrics/time_period_storage",
+    "//brave/ios/browser/keyed_service",
     "//components/keyed_service/core",
     "//components/pref_registry",
     "//components/search_engines",
--- a/ios/browser/serp_metrics/serp_metrics_bridge.h
+++ b/ios/browser/serp_metrics/serp_metrics_bridge.h
@@ -1,0 +1,30 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#ifndef BRAVE_IOS_BROWSER_SERP_METRICS_SERP_METRICS_BRIDGE_H_
+#define BRAVE_IOS_BROWSER_SERP_METRICS_SERP_METRICS_BRIDGE_H_
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+/// Records and aggregates search engine usage counts.
+NS_SWIFT_NAME(SerpMetrics)
+@protocol SerpMetricsBridge
+/// Brave Search engine usage counts recorded yesterday.
+@property(readonly) NSInteger braveSearchCountForYesterday;
+/// Google Search engine usage counts recorded yesterday.
+@property(readonly) NSInteger googleSearchCountForYesterday;
+/// Non-Brave, non-Google search engine usage counts recorded yesterday.
+@property(readonly) NSInteger otherSearchCountForYesterday;
+/// Search engine usage counts recorded for the stale period.
+@property(readonly) NSInteger searchCountForStalePeriod;
+/// Clears all stored search engine usage counts.
+- (void)clearHistory;
+@end
+
+NS_ASSUME_NONNULL_END
+
+#endif  // BRAVE_IOS_BROWSER_SERP_METRICS_SERP_METRICS_BRIDGE_H_
--- a/ios/browser/serp_metrics/serp_metrics_bridge_impl.h
+++ b/ios/browser/serp_metrics/serp_metrics_bridge_impl.h
@@ -1,0 +1,24 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#ifndef BRAVE_IOS_BROWSER_SERP_METRICS_SERP_METRICS_BRIDGE_IMPL_H_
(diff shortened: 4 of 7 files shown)


ios/app/brave_profile_controller.h
```diff
@@ -31,6 +31,7 @@ OBJC_EXPORT
 @property(readonly) BraveTabGeneratorAPI* tabGeneratorAPI;
 @property(readonly) BraveWalletAPI* braveWalletAPI;
 @property(readonly) BraveStats* braveStats;
+@property(readonly) SerpMetricsImpl* serpMetrics;
```
kylehickinson, line 34: `id<SerpMetrics>`, the Impl should not be exposed

ios/browser/api/serp_metrics/BUILD.gn
```diff
@@ -0,0 +1,19 @@
+# Copyright (c) 2026 The Brave Authors. All rights reserved.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+static_library("serp_metrics") {
```
kylehickinson, line 6: We are no longer actively adding new files to `ios/browser/api`, please just add these new files to `ios/browser/serp_metrics`

ios/browser/api/serp_metrics/headers.gni
```diff
@@ -0,0 +1,7 @@
+# Copyright (c) 2026 The Brave Authors. All rights reserved.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+serp_metrics_public_headers =
+    [ "//brave/ios/browser/api/serp_metrics/serp_metrics.h" ]
```
kylehickinson, line 7: You can just add this header directly to the `ios/BUILD.gn` list for now directly

ios/browser/api/serp_metrics/serp_metrics.h
kylehickinson, line 1: `serp_metrics_bridge.h`

ios/browser/api/serp_metrics/serp_metrics+private.h
kylehickinson, line 1: `serp_metrics_bridge_impl.h`

ios/browser/api/serp_metrics/serp_metrics.mm
kylehickinson, line 1: `serp_metrics_bridge_impl.mm`

ios/browser/api/serp_metrics/serp_metrics.h
```diff
@@ -0,0 +1,36 @@
...
+
+#ifndef BRAVE_IOS_BROWSER_API_SERP_METRICS_SERP_METRICS_H_
+#define BRAVE_IOS_BROWSER_API_SERP_METRICS_SERP_METRICS_H_
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+/// SerpMetrics records and aggregates search engine usage counts.
+@protocol SerpMetrics <NSObject>
```
kylehickinson, line 14: 1. Name it `SerpMetricsBridge` and then add `NS_SWIFT_NAME(SerpMetrics)`
2. Drop `<NSObject>`

ios/browser/api/serp_metrics/serp_metrics.h
```diff
@@ -0,0 +1,36 @@
...
+/// Non-Brave, non-Google search engine usage counts recorded yesterday.
+@property(readonly) NSInteger otherSearchCountForYesterday;
+/// Search engine usage counts recorded for the stale period.
+@property(readonly) NSInteger searchCountForStalePeriod;
+/// Clears all stored search engine usage counts.
+- (void)clearHistory;
+@end
+
+OBJC_EXPORT
+@interface SerpMetricsImpl : NSObject <SerpMetrics>
```
kylehickinson, line 28: Doesn't need to be exposed at all in this header, can be put in `serp_metrics_bridge_impl.h`

ios/browser/api/serp_metrics/serp_metrics.mm
```diff
@@ -0,0 +1,60 @@
...
+#include "brave/ios/browser/api/serp_metrics/serp_metrics+private.h"
+#include "brave/ios/browser/serp_metrics/serp_metrics_service_factory_ios.h"
+#include "ios/chrome/browser/shared/model/profile/profile_ios.h"
+
+@implementation SerpMetricsImpl {
+  raw_ptr<serp_metrics::SerpMetrics> _serp_metrics;
+}
+
+- (instancetype)initWithProfile:(ProfileIOS*)profile {
+  if ((self = [super init])) {
```
kylehickinson, line 22: The impl object should just accept a plain `serp_metrics::SerpMetrics` (or whatever its bridging), it shouldn't accept a profile or access it via the factory directly

ios/browser/api/serp_metrics/serp_metrics.mm
```diff
@@ -0,0 +1,60 @@
...
+        serp_metrics::SerpMetricsServiceFactoryIOS::GetForProfile(profile);
+    if (serp_metrics_service) {
+      _serp_metrics = serp_metrics_service->Get();
+    }
+  }
+  return self;
+}
+
+- (NSInteger)braveSearchCountForYesterday {
+  CHECK(_serp_metrics);
```
kylehickinson, line 33: raw_ref the main ptr and do a CHECK_DEREF in the init/caller side instead

ios/app/brave_profile_controller.h
kylehickinson: Actually you should really just be bridging/exposing the factory instead and using that to get this instead of adding new things to `BraveProfileController`

aseren: Changed to `id<SerpMetrics>`. Bridging and exposing the factory gives no benefit here but will add more complexity because will require additional bridge code, so I would go the current approach if possible.

ios/browser/api/serp_metrics/BUILD.gn
aseren: done

ios/browser/api/serp_metrics/headers.gni
aseren: done

ios/browser/serp_metrics/serp_metrics.h
aseren: changed

ios/browser/api/serp_metrics/serp_metrics+private.h
aseren: changed

ios/browser/api/serp_metrics/serp_metrics.mm
aseren: changed

ios/browser/api/serp_metrics/serp_metrics.h
aseren: fixed

ios/browser/api/serp_metrics/serp_metrics.h
aseren: moved to `serp_metrics_bridge_impl.h`

ios/browser/api/serp_metrics/serp_metrics.mm
aseren: Done. Had to use raw_ptr for _serpMetrics member because ObjC ivars cannot be raw_ref. Added a comment for it:
```
@implementation SerpMetricsBridgeImpl {
  // `raw_ref` cannot be used for here because ObjC ivars are zero-initialized
  // before `init` runs, but `raw_ref` cannot represent a null or default value.
  raw_ptr<serp_metrics::SerpMetrics> _serpMetrics;  // Not owned.
}
```

ios/browser/api/serp_metrics/serp_metrics.mm
aseren: fixed, it accepts `serp_metrics::SerpMetrics&`

ios/app/brave_profile_controller.h
kylehickinson: @aseren It definitely gives benefit: the benefit of being able to pass in any profile when attempting to fetch the service. `BraveProfileController` will go away in the future and we shouldn't be adding new things to it as it doesn't support multiple profiles

ios/BUILD.gn
```diff
@@ -150,6 +150,8 @@ brave_core_public_headers +=
 brave_core_public_headers += browser_api_https_upgrades_public_headers
 brave_core_public_headers += browser_debounce_public_headers
 brave_core_public_headers += brave_stats_public_headers
+brave_core_public_headers +=
+    [ "//brave/ios/browser/serp_metrics/serp_metrics_bridge.h" ]
```
kylehickinson: Can just add the header directly to the `brave_core_public_headers` assignment above, rather than here

aseren: Moved, thanks!

ios/app/brave_profile_controller.h
aseren: Added SerpMetricsServiceFactoryBridge

kylehickinson (approved):

PR #35554, opened 2026-04-15
[ads][CodeHealth] Use base::span for redirect_chain params in URL and conversion utilities

tmancey:
No behavioral changes.

Resolves

### DIFF
--- a/components/brave_ads/core/internal/creatives/conversions/creative_set_conversion_util.h
+++ b/components/brave_ads/core/internal/creatives/conversions/creative_set_conversion_util.h
@@ -9,8 +9,8 @@
 #include <cstddef>
 #include <map>
 #include <string>
-#include <vector>
 
+#include "base/containers/span.h"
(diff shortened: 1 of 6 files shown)


aseren (approved):

components/brave_ads/core/internal/creatives/conversions/creative_set_conversion_util.h
```diff
@@ -9,12 +9,11 @@
 #include <cstddef>
 #include <map>
 #include <string>
-#include <vector>
 
+#include "base/containers/span.h"
 #include "brave/components/brave_ads/core/internal/creatives/conversions/creative_set_conversion_info.h"
 #include "brave/components/brave_ads/core/internal/user_engagement/ad_events/ad_event_info.h"
-
```
aseren, line 16: forward declaration is enough, will be fixed in a follow-up

tmancey: Fixed

PR #35556, opened 2026-04-15
Disable Chromium's default PDF viewer infobar

bbondy:
For some reason this is happening more on Origin builds, but it can also happen on normal Brave builds.

Screenshot 2026-04-15 at 2 28 35 PM

## Summary
- Disables the `kPdfInfoBar` feature flag on Windows and macOS
- This Chromium feature prompts users to set Chrome as the default PDF viewer, which is not appropriate for Brave
- The infobar was stacking with the existing default browser infobar on startup (due to an upstream bug in the pin infobar controller not propagating `another_infobar_shown` through the callback chain)

## Test plan
- [ ] Verify the "Set Brave as your default PDF viewer" infobar no longer appears on startup (macOS, Windows)
- [ ] Verify the "Brave isn't your default browser" infobar still works correctly

### DIFF
--- a/chromium_src/chrome/browser/ui/ui_features.cc
+++ b/chromium_src/chrome/browser/ui/ui_features.cc
@@ -15,6 +15,10 @@ OVERRIDE_FEATURE_DEFAULT_STATES({{
 #if BUILDFLAG(IS_WIN) || BUILDFLAG(IS_MAC) || BUILDFLAG(IS_LINUX)
     {kFewerUpdateConfirmations, base::FEATURE_DISABLED_BY_DEFAULT},
 #endif
+#if BUILDFLAG(IS_WIN) || BUILDFLAG(IS_MAC)
+    // PDF infobar is only used on Windows and macOS upstream.
+    {kPdfInfoBar, base::FEATURE_DISABLED_BY_DEFAULT},
+#endif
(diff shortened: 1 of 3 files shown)


cdesouza-chromium (approved):
LGTM

chromium_src/chrome/browser/ui/ui_features.cc
```diff
@@ -14,6 +14,10 @@ namespace features {
 OVERRIDE_FEATURE_DEFAULT_STATES({{
 #if BUILDFLAG(IS_WIN) || BUILDFLAG(IS_MAC) || BUILDFLAG(IS_LINUX)
     {kFewerUpdateConfirmations, base::FEATURE_DISABLED_BY_DEFAULT},
+#endif
+#if BUILDFLAG(IS_WIN) || BUILDFLAG(IS_MAC)
+    // PDF infobar is only used on Windows and macOS upstream.
+    {kPdfInfoBar, base::FEATURE_DISABLED_BY_DEFAULT},
```
emerick, line 20: Would be good to add an entry for this to `//brave/app/feature_defaults_unittest.cc`.

bbondy: Added in f88e1cb6b8b. Thanks!

emerick (approved):

PR #35570, opened 2026-04-15
[iOS] Bridge autofill popup suggestions to BraveWebView / Swift

EliHini:
## Summary
When `kUseChromiumWebViewsAutofill` is enabled, Adds a pre-popup hook so Brave can observe autofill suggestion metadata from the Chromium autofill client path and forwards it to Swift (via `CWVAutofillController` and an associated-object block).

Contributes to [52216](https://github.com/brave/brave-browser/issues/52216)

### DIFF
--- a/ios/browser/api/web_view/BUILD.gn
+++ b/ios/browser/api/web_view/BUILD.gn
@@ -15,6 +15,7 @@ config("config") {
 
 source_set("web_view") {
   sources = [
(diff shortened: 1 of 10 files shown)


EliHini:
Closing because the `CWVAutofillController` forwards suggestion via `CWVAutofillControllerDelegate` which is sufficient for what we need in the UI.

PR #35573, opened 2026-04-15
Fix webpack config for web-discovery-project

### LINKED ISSUE #54545: Extension-based WDP not loading
The extension background page logs this error:

```
[web-discovery] ReferenceError: module is not defined
    at ./brave/web-discovery-project/index.js (brave_web-discovery-…_js.chunk.js:1:167)
```

This is likely a regression caused by a webpack upgrade.

DJAndries:
Resolves https://github.com/brave/brave-browser/issues/54545

## Test plan

0. Start with fresh profile
1. Enable WDP
2. Go to `brave://inspect`, click the "inspect" link under the Brave extension
3. Ensure there are no errors in the DevTools console
4. Execute `window.WDP`, ensure it returns an object like so:

Screenshot-2026-04-15_20-36

### DIFF
--- a/browser/web_discovery/BUILD.gn
+++ b/browser/web_discovery/BUILD.gn
@@ -56,10 +56,6 @@ source_set("browser_tests") {
 
     sources = []
 
-    if (!is_android) {
-      sources += [ "web_discovery_browsertest.cc" ]
-    }
-
     deps = [
       "//base/test:test_support",
       "//brave/components/constants",
@@ -77,6 +73,15 @@ source_set("browser_tests") {
       "//testing/gmock",
       "//testing/gtest",
     ]
+
+    if (!is_android) {
+      sources += [ "web_discovery_browsertest.cc" ]
+      deps += [
+        "//extensions:test_support",
+        "//extensions/browser",
+        "//extensions/common:common_constants",
+      ]
+    }
 
     if (enable_web_discovery_native) {
       sources += [ "content_scraper_browsertest.cc" ]
--- a/components/brave_extension/extension/brave_extension/background/webDiscoveryProject.ts
+++ b/components/brave_extension/extension/brave_extension/background/webDiscoveryProject.ts
@@ -6,6 +6,18 @@ export {}
 
 declare let window: any
 
+let wdpStartedCallbackForTesting: (() => void) | undefined
+
+function setWDPStartedCallbackForTesting(callback: () => void) {
+  if (window.WDP !== undefined) {
+    callback()
+    return
+  }
+  wdpStartedCallbackForTesting = callback
+}
+
+window.setWDPStartedCallbackForTesting = setWDPStartedCallbackForTesting
+
 async function startWDP() {
   if (window.WDP === undefined) {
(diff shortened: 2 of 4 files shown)


components/webpack/webpack.config.js
```diff
@@ -209,6 +209,13 @@ module.exports = async function (env, argv) {
           test: /\.(ttf|eot|ico|svg|png|jpg|jpeg|gif|webp)(\?v=[0-9]\.[0-9]\.[0-9])?$/,
           loader: 'file-loader'
         },
+        // web-discovery-project is built as CommonJS but may be classified
+        // as ESM by webpack. Force auto-detection for correct CJS handling.
+        {
+          test: /\.js$/,
+          include: /web-discovery-project/,
+          type: 'javascript/auto'
```
fallaciousreasoning, line 217: Hey, just been reading up on this - it sounds like `auto` is the default type? 

It seems like `javascript/dynamic` should force common js?

browser/web_discovery/BUILD.gn
```diff
@@ -72,6 +72,8 @@ source_set("browser_tests") {
       "//components/prefs",
       "//content/public/browser",
       "//content/test:test_support",
+      "//extensions/browser",
```
fallaciousreasoning: do we need `//extensions/test:test_support` or something like that for `extensions/test/extension_background_page_waiter.h"`

components/brave_extension/extension/brave_extension/background/webDiscoveryProject.ts
```diff
@@ -6,6 +6,16 @@ export {}
 
 declare let window: any
 
+function setWDPStartedCallbackForTesting(callback: () => void) {
+  if (window.WDP !== undefined) {
+    callback()
+    return
+  }
+  window.__wdpStartedCallbackForTesting = callback
```
fallaciousreasoning: mega nit, but there's no reason to expose this on `window` - this could just be a local variable.

fallaciousreasoning (approved):
lgtm % nits & a question

components/webpack/webpack.config.js
DJAndries: it will, but it will break any transitive imports of non-cjs modules used by WDP (just tried it out, tons of errors)

szilardszaloki (approved):
I figured explicitly setting `chunkLoading` would solve the issue, but on its own it wouldn't fix webpack's module classification. Too bad we can't hint webpack here though, instead of having to add an explicit rule for WDP.

kjozwiak:

```
Brave | 1.91.66 Chromium: 147.0.7727.102 (Official Build) nightly (64-bit)
-- | --
Revision | 311811faa906d8bc382ffbb94df981d0e5f36885
OS | Windows 11 Version 25H2 (Build 26200.8246)
```

Using the STR/Cases outlined via https://github.com/brave/brave-core/pull/35573#issue-4272239071, ensured that `window.WDP` returns an object as per the following:

**`Example`** | **`Example`**
---------------|-----------------
Screenshot 2026-04-16 151125 | Screenshot 2026-04-16 151227

PR #35575, opened 2026-04-16
[Tree Tabs] Don't access the opener collection if it's not in the same window.

### LINKED ISSUE #54334: Crash Report
IMPORTANT: Your crash has already been automatically reported to our crash system. Please file this bug only if you can provide more information about it.

Brave Version: 1.91.25 Chromium: 147.0.7727.56
Operating System: Mac OS X 26.5.0

URL (if applicable) where crash occurred:

Can you reproduce this crash?
Clicking on link in a slack PWA with Tree style tabs flag on and checked. 

What steps will reproduce this crash? (If it's not reproducible, what were you doing just before the crash?)
1. 
2.
3.

****DO NOT CHANGE BELOW THIS LINE****
Crash ID: crash/a66a0000-fca6-8c0f-0000-000000000000

sangwoo108:
When a link is opened from a PWA window, the opener collection is not in the same window. - link can be opened in normal window type from PWA window. In this case, we should not try accessing the opener collection.

Resolves https://github.com/brave/brave-browser/issues/54334

### DIFF
--- a/browser/ui/tabs/brave_tree_tab_strip_collection_delegate.cc
+++ b/browser/ui/tabs/brave_tree_tab_strip_collection_delegate.cc
@@ -302,6 +302,12 @@ BraveTreeTabStripCollectionDelegate::TryAddTabToSameTreeAsOpener(
 
   tabs::TreeTabNodeTabCollection* opener_collection = nullptr;
   if (opener) {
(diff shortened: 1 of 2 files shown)


netzenbot (approved):

simonhong (approved):
++

PR #35586, opened 2026-04-16
Add logic to load and parse query filter rule file

### LINKED ISSUE #54395: Add support to load and parse the query filter json file
Once we have added the support for fetching the query-filter component (https://github.com/brave/brave-browser/issues/54393), we should add a singleton class which would parse the json and convert it into a meaningful data-structure, which can be later consumed by the query_filter clients for lookup.

r0hit0303:
Adds a singleton QueryFilterData class to load, and then parse the downloaded rules of [query-filter component](https://github.com/brave/brave-browser/issues/54393). The `QueryFilterData` class then store these rules in memory via a `vector` of new data structure `QueryFilterRule` which is exposed to public via the member `rules()` method. 

Client integration to consume the query-filter would be followed-up in the next change. This change also includes unit tests, and was tested manually to ensure the component was downloaded from server and was parsed correctly.
e2e integration tests / browser tests would be followed up later once we have the client integration. 

This change is behind the feature flag [kQueryFilterComponent](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/components/query_filter/common/features.cc?L13) which is disabled by default. 

Resolves https://github.com/brave/brave-browser/issues/54395

## Checklist:

- [x] Review design docs
  [Browser design principles](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/docs/chrome_browser_design_principles.md)
  [Style guide](https://chromium.googlesource.com/chromium/src/+/main/styleguide/c++/c++.md)
  [Core principles](https://www.chromium.org/developers/core-principles/)
-  [x] Ensure there are (tests)[https://www.chromium.org/developers/testing/]. Unit test as much as possible (including edge cases), but also include browser tests covering high level functionality.
- [x] Ensure that there are comments explaining what classes/methods are/do. The "why" is often more important than the "what" in comments. Also update any relevant docs (moving docs from wiki to brave-core if necessary).
- Request security or other review (third-party libraries, rust code, etc...) if applicable [security/privacy review is needed](https://github.com/brave/brave-browser/wiki/Security-reviews) [other review](https://github.com/brave/reviews/issues/new/choose)
  Also see [adding third-party libraries](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/docs/adding_to_third_party.md) for general guidelines on using third party code
- [x] Make sure there is a [ticket](https://github.com/brave/brave-browser/issues) for your issue
- [x] Use Github [auto-closing keywords](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue) in the PR description above
- [x] Write a good [PR/commit description](https://google.github.io/eng-practices/review/developer/cl-descriptions.html)
- Squash any review feedback or "fixup" commits before merge, so that history is a record of what happened in the repo, not your PR
- [x] Add appropriate labels (`QA/Yes` or `QA/No`; `release-notes/include` or `release-notes/exclude`; `OS/...`) to the associated issue
- [x] Checked the PR locally:
  * `npm run test -- brave_browser_tests`, `npm run test -- brave_unit_tests` [wiki](https://github.com/brave/brave-browser/wiki/Tests)
  * `npm run presubmit` [wiki](https://github.com/brave/brave-browser/wiki/Presubmit-checks), `npm run gn_check`, `npm run tslint`
- Run `git rebase master` (if needed)


r0hit0303:
Hi @fmarier and @cdesouza-chromium, tagging you for visibility on this upcoming review to share any early thoughts. Thanks!

components/query_filter/browser/query_filter_service_unittest.cc
```diff
@@ -0,0 +1,178 @@
...
+
+TEST_F(QueryFilterServiceTest, LoadsAndParsesQueryFilterFileCorrectly) {
+  AssertCurrentRulesMatchDefaultRules();
+}
+
+// Parse rules json specific tests below.
+TEST_F(QueryFilterServiceTest,
+       ParseRulesJsonChecks_EmptyJson_ProducesNoNewRules) {
+  TryNewRulesUpdate("");
+  AssertCurrentRulesMatchDefaultRules();
```
netzenbot: Per GoogleTest conventions, test suites containing death tests must end with `DeathTest` (e.g., `QueryFilterServiceDeathTest`) so GoogleTest applies the correct threading behavior. The `EXPECT_DEATH_IF_SUPPORTED` call in `QueryFilterServiceTest` violates this. See [CS-061](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-061).

components/query_filter/browser/query_filter_service_unittest.cc
```diff
@@ -0,0 +1,178 @@
...
+    base::ScopedAllowBlockingForTesting allow_blocking;
+    ASSERT_TRUE(component_install_dir_.CreateUniqueTempDir());
+    ASSERT_TRUE(base::WriteFile(
+        component_install_dir_.GetPath().AppendASCII("query-filter.json"),
+        kSampleQueryFilterJson));
+    service()->OnComponentReady(ComponentInstallDir());
+    EXPECT_TRUE(base::test::RunUntil(
+        [&]() { return service()->rules().size() == 2u; }));
+  }
+
```
netzenbot, line 67: The tests that call `TryNewRulesUpdate` (which forwards to the private `ParseRulesJson`) are testing an internal implementation detail rather than the public API. Consider driving these tests through `OnComponentReady` with different file contents, or expose a `ParseRulesJson_for_testing()` method if direct invocation is necessary — using `friend class` to reach private methods couples the tests to the implementation. https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-002

components/query_filter/browser/query_filter_service_unittest.cc
```diff
@@ -0,0 +1,178 @@
...
+                testing::ElementsAre("__hsfp", "gclid", "fbclid"));
+    EXPECT_THAT(service()->rules()[1].include,
+                testing::ElementsAre("*://*.youtube.com/*", "*://youtube.com/*",
+                                     "*://youtu.be/*"));
+    EXPECT_TRUE(rules[1].exclude.empty());
+    EXPECT_THAT(rules[1].params, testing::ElementsAre("si"));
+  }
+
+ protected:
+  QueryFilterService* service() const {
```
netzenbot, line 69: `GetInstance()` has a feature-flag guard that returns `nullptr` when `kQueryFilterComponent` is disabled, but no test covers that path. Per [TI-026](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-026), both sides of a guard condition should be tested — add a test with the feature disabled that asserts `GetInstance()` returns `nullptr`.

components/query_filter/browser/query_filter_service_unittest.cc
r0hit0303: The tests that call ` TryNewRulesUpdate` are all prefixed as `ParseRulesJsonChecks` because the intent of the test was to check the parser (which is internal). Relying on `OnComponentReady` felt like a mis-direction to the reader in this context. 

I'm think carving out the parsing functionality outside of the class and make it as a static free method which would be a better approach, as it would shift the responsibility of the parser from the service, which the latter than can rely on it to get the right ruleset for the clients, for a given json string.

components/query_filter/browser/query_filter_service.cc
```diff
@@ -0,0 +1,135 @@
...
+    }
+  };
+
+  // Parse each rule entry in the list.
+  for (const base::Value& entry : *list) {
+    const base::DictValue* rule_dict = entry.GetIfDict();
+    if (!rule_dict) {
+      DLOG(ERROR) << "query-filter.json: non dict rule entry found";
+      continue;
+    }
```
netzenbot, line 114: nit: `install_dir` is stored into a member, so it's a sink parameter — take it by value (`base::FilePath install_dir`) to allow callers to move in a temporary. https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-047

components/query_filter/browser/query_filter_service.cc
```diff
@@ -0,0 +1,135 @@
...
+  std::vector<QueryFilterRule> rules;
+
+  // Helper to insert valid strings from a list into a vector.
+  auto valid_string_inserter = [](const base::ListValue* lv,
+                                  std::vector<std::string>& output) {
+    if (!lv || lv->empty()) {
+      return;
+    }
+    // Resize the output vector in advance for faster insertion and doing only
+    // one time vector memory allocation.
```
netzenbot: nit: `std::count_if` returns `ptrdiff_t` (a signed type), so `precomputed_output_size` is deduced as signed. Prefer `size_t precomputed_output_size = static_cast<size_t>(std::count_if(...))` to avoid an implicit signed-to-unsigned conversion when calling `output.resize()`. See https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-067.

components/query_filter/browser/query_filter_service.cc
```diff
@@ -0,0 +1,135 @@
...
+    QueryFilterRule rule;
+    valid_string_inserter(rule_dict->FindList(kQueryFilterRulesIncludeKey),
+                          rule.include);
+    valid_string_inserter(rule_dict->FindList(kQueryFilterRulesExcludeKey),
+                          rule.exclude);
+    valid_string_inserter(rule_dict->FindList(kQueryFilterRulesParamsKey),
+                          rule.params);
+    rules.push_back(std::move(rule));
+  }
+
```
netzenbot, line 138: nit: `OnRulesJsonLoaded` is a `base::BindOnce` callback target but lacks a `DCHECK_CALLED_ON_VALID_SEQUENCE` guard. Per [CSM-019](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-memory.md#CSM-019), methods used as bind targets should assert the correct sequence.

components/query_filter/browser/query_filter_service.cc
```diff
@@ -0,0 +1,135 @@
...
+  // Parse each rule entry in the list.
+  for (const base::Value& entry : *list) {
+    const base::DictValue* rule_dict = entry.GetIfDict();
+    if (!rule_dict) {
+      DLOG(ERROR) << "query-filter.json: non dict rule entry found";
+      continue;
+    }
+
+    QueryFilterRule rule;
+    valid_string_inserter(rule_dict->FindList(kQueryFilterRulesIncludeKey),
```
netzenbot, line 130: nit: The thread pool task only specifies `base::MayBlock()` but omits an explicit `base::TaskPriority` and `base::TaskShutdownBehavior`. Per [CSM-022](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-memory.md#CSM-022), both should be specified — e.g., `{base::MayBlock(), base::TaskPriority::BEST_EFFORT, base::TaskShutdownBehavior::SKIP_ON_SHUTDOWN}` for non-urgent, abandonment-safe work like reading a component file.

components/query_filter/browser/query_filter_service.h
```diff
@@ -0,0 +1,64 @@
...
+
+#include "base/files/file_path.h"
+#include "base/memory/singleton.h"
+#include "base/memory/weak_ptr.h"
+
+namespace query_filter {
+
+struct QueryFilterRule {
+  QueryFilterRule();
+  ~QueryFilterRule();
```
netzenbot, line 22: nit: `<string_view>` is used in the `ParseRulesJson` declaration but not included. Add `#include <string_view>` to the header. https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-001

components/query_filter/browser/query_filter_service.cc
```diff
@@ -0,0 +1,142 @@
...
+
+std::vector<QueryFilterRule> ParseRulesJson(const std::string_view contents) {
+  if (contents.empty()) {
+    return {};
+  }
+
+  auto parsed = base::JSONReader::ReadAndReturnValueWithError(
+      contents, base::JSON_PARSE_CHROMIUM_EXTENSIONS);
+  if (!parsed.has_value()) {
+    DLOG(ERROR) << "query-filter.json parse error: " << parsed.error().message;
```
r0hit0303, line 77: I feel like a DCHECK here and in other places below would be more appropriate but wanted to get your thoughts.

components/query_filter/browser/query_filter_service.cc
```diff
@@ -0,0 +1,142 @@
...
+  // Parse each rule entry in the list.
+  for (const base::Value& entry : *root_list) {
+    const base::DictValue* rule_dict = entry.GetIfDict();
+    if (!rule_dict) {
+      DLOG(ERROR) << "query-filter.json: non dict rule entry found";
+      continue;
+    }
+
+    if (!rule_dict->FindList(kQueryFilterRulesIncludeKey) ||
+        !rule_dict->FindList(kQueryFilterRulesExcludeKey) ||
```
netzenbot, line 117: nit: `OnComponentReady` writes to `install_dir_` but lacks `DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_)`, while `OnRulesJsonLoaded` has it. Per [CSM-015](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-memory.md#CSM-015), sequence checking should be applied consistently to all methods that access shared state.

components/query_filter/browser/query_filter_service.h
```diff
@@ -0,0 +1,71 @@
...
+
+// Singleton responsible for getting rules from the query filter component.
+// (query-filter.json). See brave/adblock-lists/brave-lists/query-filter.json
+class QueryFilterService {
+ public:
+  QueryFilterService(const QueryFilterService&) = delete;
+  QueryFilterService& operator=(const QueryFilterService&) = delete;
+  ~QueryFilterService();
+
+  static QueryFilterService* GetInstance();
```
netzenbot, line 40: nit: `GetInstance()` returns `nullptr` when `kQueryFilterComponent` is disabled, but this is not documented. Callers who miss the null-check will crash. Per [DOC-004](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-004), the method contract — including possible null return — should be documented: `// Returns nullptr if the kQueryFilterComponent feature is disabled.`

components/query_filter/browser/query_filter_service_unittest.cc
```diff
@@ -0,0 +1,177 @@
...
+INSTANTIATE_TEST_SUITE_P(/** no prefix */,
+                         QueryFilterServiceTest,
+                         ::testing::Values(false, true),
+                         [](const ::testing::TestParamInfo<bool>& info) {
+                           return info.param ? "QueryFilterComponentEnabled"
+                                             : "QueryFilterComponentDisabled";
+                         });
+
+// Parse rules json specific tests below.
+class QueryFilterParseRulesJsonTest : public testing::Test {};
```
netzenbot, line 113: nit: `QueryFilterParseRulesJsonTest` adds no members, `SetUp`, or `TearDown`, so there's no reason for a fixture — use `TEST()` instead of `TEST_F()`. https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-036

components/query_filter/browser/query_filter_component_installer.cc
```diff
@@ -95,8 +96,10 @@ void QueryFilterComponentInstallerPolicy::ComponentReady(
     const base::Version& version,
     const base::FilePath& path,
     base::DictValue manifest) {
-  // TODO(https://github.com/brave/brave-browser/issues/54395): Pass the path to
-  // query filter service when it is implemented.
+  // Notify the service that the component is ready.
+  if (QueryFilterService* service = QueryFilterService::GetInstance()) {
```
cdesouza-chromium: Please, consider implementing this the same way it is implemented here (which means moving the file parsing to this translation unit, etc):

https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/component_updater/commerce_heuristics_component_installer.cc;l=148

Make sure to check Chromium code for similar implementations, as there are hundreds of examples in Chromium of how to approach this.

components/query_filter/browser/query_filter_service.cc
```diff
@@ -0,0 +1,142 @@
...
+QueryFilterRule& QueryFilterRule::operator=(const QueryFilterRule& other) =
+    default;
+
+// QueryFilterService implementation.
+// static
+QueryFilterService* QueryFilterService::GetInstance() {
+  if (!base::FeatureList::IsEnabled(features::kQueryFilterComponent)) {
+    return nullptr;
+  }
+  return base::Singleton<QueryFilterService>::get();
```
cdesouza-chromium: Is there any particular reason for the use of `base::Singleton` here? if not, take a look at:

https://source.chromium.org/chromium/chromium/src/+/main:components/commerce/core/commerce_heuristics_data.cc;l=39

components/query_filter/browser/query_filter_service.cc
```diff
@@ -0,0 +1,142 @@
...
+
+void QueryFilterService::OnRulesJsonLoaded(const std::string& contents) {
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+  auto rules = ParseRulesJson(contents);
+  if (!rules.empty()) {
+    rules_ = std::move(rules);
+  }
+}
+
+std::vector<QueryFilterRule> ParseRulesJson(const std::string_view contents) {
```
cdesouza-chromium: As recommended in the other comment, move this code to `components/query_filter/browser/query_filter_component_installer.cc` and make sure this function is called with the appropriate task traits, meaning, `{base::MayBlock(), base::TaskPriority::BEST_EFFORT}`, as this correct signals that the task is allow to called blocking functions, and should only be loaded once we have resources for doing disk reads.

components/query_filter/browser/query_filter_service.cc
cdesouza-chromium: As in the example from upstream I linked above, you can also find some inspiration there:

https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/component_updater/commerce_heuristics_component_installer.cc;l=67

If you notice that function, it has lines like these:

```cxx
  if (!base::ReadFileToString(commerce_hint_file_path,
                              &commerce_hint_json_data)) {
    LOG(WARNING) << "Failed reading from " << commerce_hint_file_path.value();
    return;
  }
```

I'm not sure what else we can do, if the data is broken. Certainly crashing the browser is not the way.

In general, the `DCHECK` macro is not encouraged, because it has no effect in release mode. Please review this document for more details: https://chromium.googlesource.com/chromium/src/+/HEAD/styleguide/c++/checks.md

components/query_filter/browser/query_filter_component_installer.cc
r0hit0303: Thanks, I have updated my change according to that. I think in the upstream there's a slight chance of hitting a race condition. They are posting a task on thread [LoadHeuristicFilesFromDisk](https://source.chromium.org/chromium/chromium/src/+/main:components/commerce/core/commerce_heuristics_data.cc;l=57), which not only loads the file but also touches the `CommerceHeuristicsData` [singleton]( https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/component_updater/commerce_heuristics_component_installer.cc;l=114;bpv=0) to populate the contents of the file to data structures. This is done w/o ensuring any checks on sequence_checker, so we could have situations where the consumers of `CommerceHeuristicsData` may access the underlying data structure from another thread while it's being modified.  I've taken care of validating with the sequence_checker in my latest change.

components/query_filter/browser/query_filter_service.cc
r0hit0303: Done

components/query_filter/browser/query_filter_service.cc
r0hit0303: Done

components/query_filter/browser/query_filter_service.cc
r0hit0303: Crashing on debug test builds made sense to me to catch bad pushes to query-filter.json early, but I'm also ok with what you suggested to instead LOG them.

components/query_filter/browser/query_filter_component_installer.h
```diff
@@ -6,15 +6,54 @@
...
+class FilePath;
+class Version;
+}  // namespace base
+
 namespace component_updater {
 class ComponentUpdateService;
-}  // namespace component_updater
 
-namespace query_filter {
+class QueryFilterComponentInstallerPolicy
```
r0hit0303, line 24: @cdesouza-chromium I had to move the declaration back to the header file in order to facilitate testing the ComponentReady method.  Please see `QueryFilterComponentInstallerTest#TestComponentReady` in `components/query_filter/browser/query_filter_component_installer_unittest.cc`

components/query_filter/browser/query_filter_component_installer.cc
```diff
@@ -8,64 +8,58 @@
...
+void OnQueryFilterFileRead(const base::Version& version,
+                           const std::string& json_data) {
+  if (json_data.empty()) {
+    return;
+  }
+  auto* data = query_filter::QueryFilterData::GetInstance();
+  if (!data) {
+    LOG(WARNING) << "QueryFilterData instance is not available";
+    return;
+  }
```
cdesouza-chromium:
```cxx
auto* data = query_filter::QueryFilterData::GetInstance();
CHECK(data);
```

For this particular case we should CHECK because if we got here `features::kQueryFilterComponent` has to be set, as per `RegisterQueryFilterComponent`, so something is wrong at this point if we don't have a valid instance.

components/query_filter/browser/query_filter_component_installer.h
```diff
@@ -6,15 +6,54 @@
...
+  update_client::InstallerAttributes GetInstallerAttributes() const override;
+  bool IsBraveComponent() const override;
+};
+
 // Registers the Query Filter component with the component updater.
 void RegisterQueryFilterComponent(
     component_updater::ComponentUpdateService* cus);
-
-}  // namespace query_filter
+}  // namespace component_updater
```
cdesouza-chromium, line 57: Did you mean to change the namespace here from `query_filter` to `component_updater`?

components/query_filter/browser/query_filter_component_installer.h
cdesouza-chromium: Ah, that's unfortunate, but it may be better for a separation of concerns.

components/query_filter/browser/query_filter_data.cc
```diff
@@ -0,0 +1,141 @@
...
+
+const std::string QueryFilterData::GetVersion() const {
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+  if (!version_.IsValid()) {
+    return std::string();
+  }
+  return version_.GetString();
+}
+
+bool QueryFilterData::PopulateDataFromComponent(std::string_view json_data) {
```
cdesouza-chromium, line 62: Prefer having this being:

```cxx
bool QueryFilterData::PopulateDataFromComponent(const base::ListValue& rules)
```

components/query_filter/browser/query_filter_data.h
```diff
@@ -0,0 +1,72 @@
...
+  const std::vector<QueryFilterRule>& rules() const;
+
+ private:
+  friend class base::NoDestructor<QueryFilterData>;
+  QueryFilterData();
+
+  std::vector<QueryFilterRule> rules_;
+  base::Version version_;
+
+  SEQUENCE_CHECKER(sequence_checker_);
```
cdesouza-chromium: We don't need this sequence checker anymore, now that the disk op is being handled elsewhere.

components/query_filter/browser/query_filter_component_installer_unittest.cc
```diff
@@ -7,46 +7,94 @@
...
 #include "testing/gtest/include/gtest/gtest.h"
-
-namespace query_filter {
-
-inline constexpr char kQueryFilterComponentId[] =
-    "cemdlagocoimleflkfkjoihojfainiho";
+namespace {
+// Sample query filter JSON which would be written to a file during setup
+// and then read by the query filter service to prepopulate the default rules.
+constexpr char kSampleQueryFilterJson[] = R"json(
```
cdesouza-chromium, line 27: Where are we getting this list from ultimately? Do we own the format of this JSON schema?

components/query_filter/browser/query_filter_component_installer.cc
r0hit0303: Done

components/query_filter/browser/query_filter_component_installer.h
r0hit0303: Yes, indeed it was intentional. Most of the upstream clients like [1](https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/component_updater/commerce_heuristics_component_installer.h;l=22), [2](https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/component_updater/cros_component_installer_chromeos.h;l=59), [3](https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/component_updater/pki_metadata_component_installer_policy.h;l=19), [4](https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/component_updater/file_type_policies_component_installer.h;l=27) relied on scoping their feature specific component installer policy within `component_updater` namespace. I think I might have also seen one or two clients in Brave doing the same. Therefore, I went ahead with this update to put the `QueryFilterComponentInstallerPolicy`, within the `component_updater` namespace.

components/query_filter/browser/query_filter_component_installer_unittest.cc
r0hit0303: We are getting this list from [here](https://sourcegraph.com/r/github.com/brave/adblock-lists/-/blob/brave-lists/query-filter.json). Yes, we own the format of this schema.

components/query_filter/browser/query_filter_data.cc
r0hit0303: I'm not sure if I completely follow the full scope of the change asked here. Currently, `PopulateDataFromComponent` takes in a `std::string_view` indicating to the readers that the contents of the file are raw and more importantly un-treated. I might be wrong here, but changing the method signature to take in a `base::ListValue&` would slightly erode that indication of rawness of the json content. 

Was there anything specific you were thinking of with the asked change that the current change is not capturing right? I want to make sure I understand your concerns better.

components/query_filter/browser/BUILD.gn
```diff
@@ -7,20 +7,27 @@ static_library("browser") {
...
 
+  public_deps = [ "//base" ]
+
   deps = [
-    "//base",
     "//brave/components/brave_component_updater/browser:public",
     "//brave/components/query_filter/common",
     "//components/component_updater",
+    "//components/update_client",
   ]
```
netzenbot, line 23: `//components/update_client` is included in the public header (`query_filter_component_installer.h`) and its types appear directly in the class declaration — that makes it a header-visible dependency. It should be in `public_deps`, not `deps`, so consumers of this target also get the transitive include path. See https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-024

components/query_filter/browser/query_filter_data.cc
```diff
@@ -0,0 +1,136 @@
...
+  rules_ = std::move(rules);
+  return true;
+}
+
+void QueryFilterData::ResetRulesForTesting() {
+  rules_.clear();
+  version_ = base::Version();
+}
+
+}  // namespace query_filter
```
netzenbot, line 140: nit: `std::count_if` and `std::for_each` are used in `valid_string_inserter` but `<algorithm>` is not included. Per [CS-001](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-001), always include what you use — don't rely on transitive includes.

components/query_filter/browser/query_filter_component_installer_unittest.cc
```diff
@@ -58,24 +106,54 @@ TEST_F(QueryFilterComponentInstallerTest,
...
+  base::ScopedTempDir component_install_dir_;
+  ASSERT_TRUE(component_install_dir_.CreateUniqueTempDir());
+  ASSERT_TRUE(base::WriteFile(component_install_dir_.GetPath().AppendASCII(
+                                  query_filter::kQueryFilterJsonFile),
+                              kSampleQueryFilterJson));
+
+  EXPECT_TRUE(component_updater::QueryFilterComponentInstallerPolicy()
+                  .VerifyInstallation(base::DictValue(),
+                                      component_install_dir_.GetPath()));
 }
```
netzenbot, line 135: `TestComponentReady` populates `QueryFilterData::GetInstance()` with version and rules but `QueryFilterComponentInstallerTest` has no `TearDown` that calls `ResetRulesForTesting()`. The singleton state will leak into any test that runs afterward and checks initial conditions (e.g., the `EXPECT_EQ("", ...)` pre-check at the top of this test would fail if test ordering changes). Add a `TearDown` override that calls `query_filter::QueryFilterData::GetInstance()->ResetRulesForTesting()`. https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-013

components/query_filter/browser/query_filter_component_installer_unittest.cc
```diff
@@ -58,24 +106,54 @@ TEST_F(QueryFilterComponentInstallerTest,
...
+  // Target version to populate with
+  base::Version version("1.0.0");
+
+  // Verify the version is empty before the component is ready
+  EXPECT_EQ("", query_filter::QueryFilterData::GetInstance()->GetVersion());
+  // Verify also the rules are empty before the component is ready
+  EXPECT_TRUE(query_filter::QueryFilterData::GetInstance()->rules().empty());
+
+  // Initiate component ready which would load the json file and populate the
+  // query filter data.
```
netzenbot, line 155: `QueryFilterData::GetInstance()` returns `nullptr` when the feature is disabled. Even though the fixture enables the feature, add `ASSERT_NE(nullptr, query_filter::QueryFilterData::GetInstance())` before dereferencing it in `TestComponentReady` — if the guard ever fails the test will crash rather than report a clean failure. https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-034

components/query_filter/browser/query_filter_data.cc
r0hit0303: Interesting this only catches now. `/review local` had passed the change locally for me earlier. 
image

components/query_filter/browser/query_filter_data.cc
cdesouza-chromium: Explained in a call. I particularly think it would be better if JSON parsing was moved into `components/query_filter/browser/query_filter_component_installer.cc`, and this function contract was narrower to take only the a list, but I don't feel particularly strong about it, leaving it up to you to decide (and resolve this comment)

components/query_filter/browser/query_filter_data.cc
```diff
@@ -0,0 +1,138 @@
...
+  const base::ListValue* root_list = parsed->GetIfList();
+  if (!root_list) {
+    LOG(WARNING) << "query-filter.json: root must be an array";
+    return false;
+  }
+
+  std::vector<QueryFilterRule> rules;
+
+  // Helper to insert valid strings from a list into a vector.
+  auto valid_string_inserter = [](const base::ListValue* lv,
```
cdesouza-chromium, line 82: As discussed in a call, all this parsing code should go, and we should use the JSON schema compiler here. Log an issue for a follow up and have a comment added here for:

```
TODO(issuelink): Code to be replaced, etc
```

cdesouza-chromium (approved):
LGTM % comments

components/query_filter/browser/query_filter_data.cc
r0hit0303: Done

components/query_filter/browser/query_filter_data.cc
r0hit0303: Ack

netzenbot (approved):

components/query_filter/browser/query_filter_data_unittest.cc
```diff
@@ -0,0 +1,200 @@
...
+  EXPECT_EQ(rules.size(), 2u);
+  EXPECT_THAT(rules[0].include, testing::ElementsAre("*://a/*"));
+  EXPECT_TRUE(rules[0].exclude.empty());
+  EXPECT_THAT(rules[0].params, testing::ElementsAre("x"));
+  EXPECT_THAT(rules[1].include, testing::ElementsAre("*://b/*"));
+  EXPECT_TRUE(rules[1].exclude.empty());
+  EXPECT_THAT(rules[1].params, testing::ElementsAre("y"));
+}
+
+TEST_F(QueryFilterDataTest, TestNonStringListItemsAreIgnored) {
```
fmarier, line 101: In this case, we should probably ignore the rule entirely, not just the exclude line. Otherwise we could end up accidentally applying a rule more broadly then intended.

r0hit0303: Thanks Francois. As discussed offline your proposition makes sense to me. I'll make sure to keep in mind to ignore the whole rule when I implement @cdesouza-chromium [proposal](https://github.com/brave/brave-browser/issues/54775).

PR #35587, opened 2026-04-16
[ads][CodeHealth] Remove RunUntilIdle from ads tests

tmancey:
Remove all RunUntilIdle calls from AdsClientNotifierForTesting, test setup, and test bodies. Resource loading tests use RunUntil with an IsLoaded condition since LoadAndParseResourceComponent is async. SetUpUnitTest uses a BarrierClosure and RunLoop to wait for database creation, token state loading, and client state loading without advancing the mock clock. SetUpIntegrationTest flushes only 0-delay replies. Mobile notification ad tests avoid nested RunLoops by separating serve from event trigger and using TestFuture for the async callback wait.

No behavioral changes.

Resolves

### DIFF
--- a/components/brave_ads/core/internal/ad_units/notification_ad/notification_ad_for_mobile_test.cc
+++ b/components/brave_ads/core/internal/ad_units/notification_ad/notification_ad_for_mobile_test.cc
@@ -7,6 +7,7 @@
 #include "base/test/gmock_callback_support.h"
 #include "base/test/mock_callback.h"
 #include "base/test/scoped_feature_list.h"
+#include "base/test/test_future.h"
 #include "brave/components/brave_ads/core/internal/catalog/catalog_url_request_builder_util.h"
 #include "brave/components/brave_ads/core/internal/common/platform/platform_helper.h"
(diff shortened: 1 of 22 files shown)


aseren (approved):

netzenbot (approved):

tmancey:
The count is correct and the comment above it names both async paths explicitly. There's no way to derive the count from the call sites without restructuring the function, and converting to sequential chaining would change the semantics since the two paths are genuinely concurrent. Left as-is; a code review should catch any future changes to the step count.

tmancey:
Fixed. For TextClassificationResource, the pipeline is created synchronously before the async parse begins, so ASSERT_TRUE(resource_->IsLoaded()) was added to the Arrange section to confirm the load was non-vacuously initiated before the RunUntil(!IsLoaded()) wait. For AntiTargetingResource and PurchaseIntentResource, the load is fully async with no intermediate loaded state, so ASSERT_TRUE(resource_->GetManifestVersion()) was added instead to confirm the component notification was processed and the load dispatched.

PR #35592, opened 2026-04-16
[CodeHealth] Make `post_presubmit.py` report `gh` failures

cdesouza-chromium:
This PR changes the script for posting presubmit reports to github, to
print the `stderr` output from process that have exited with an error
code. This will let us see why a particular API call failed at a given
moment.


mkarolin (approved):
++

netzenbot (approved):

PR #35606, opened 2026-04-16
[Android] Fix Origin paywall stuck during Play Store purchase restore

### LINKED ISSUE #54637: [Android][Origin] Brave Origin paywall stuck when restoring Play Store purchase after device change
When a user who previously purchased Brave Origin on Android switches to a new device, opening Settings → Brave Origin can show the paywall screen indefinitely instead of restoring the purchase.

At Brave startup, `BraveOriginSubscriptionPrefs.verifyPurchase()` queries Google Play for the existing purchase, finds it, stores the purchase token, and kicks off the SKUs SDK `createOrderFromReceipt` → `fetchOrderCredentials` chain to obtain an order ID. Until that chain completes, `requestCredentialSummary` returns inactive (no order ID yet), so `handleOriginPreferenceClick` routes the user to `BraveOriginPlansActivity` (the purchase screen) — even though a Play Store purchase exists and is mid-restore.

Steps to Reproduce                                                                                                                                                                                                                                                                                                                                  
  1. Purchase Brave Origin on Android device A.
  2. Install/sign into Brave on Android device B using the same Play account (or alternatively just uninstall Brave on device A and install again if you have one device only)
  3. Launch Brave on device B, immediately open Settings → Brave Origin.

Expected: Settings → Brave Origin screen opens with a spinner while credentials finish fetching (same UX as the post-purchase flow on the buying device), then transitions to the restart prompt.
                                                                                                                                                                                                                                                                                                                                                      
Actual: The purchase screen is shown and does not go away, even though the purchase exists and the restore is in progress.

SergeyZhukovsky:
What that PR added:
  1. In `BraveMainPreferencesBase.handleOriginPreferenceClick`, check `BraveOriginSubscriptionPrefs.isFetchingCredentials(profile)` before calling `requestCredentialSummary`. If true, navigate directly to `BraveOriginPreferences` (which renders the fetching spinner via its existing `setCredentialsFetchedCallback` path) instead of launching the purchase screen.
  2. Harden `BraveOriginSubscriptionPrefs.setCredentialsFetchedCallback` against a lost-notification race: after registering the callback, re-check `isFetchingCredentials` and synthesize a success notification if the fetch already completed (e.g., between the caller's state check and this registration). Order-ID-set is the only path that flips the state to false, so synthesizing success is safe.

Resolves: https://github.com/brave/brave-browser/issues/54637

### DIFF
--- a/android/java/org/chromium/chrome/browser/brave_origin/BraveOriginSubscriptionPrefs.java
+++ b/android/java/org/chromium/chrome/browser/brave_origin/BraveOriginSubscriptionPrefs.java
@@ -76,9 +76,19 @@ public class BraveOriginSubscriptionPrefs {
      * fetchOrderCredentials finishes. Any previously registered callback is replaced.
      *
      * @param callback Called with true on success, false on failure.
+     * @param profile Used to re-check fetch state and guard against a race where the fetch
+     *     completed between the caller's isFetchingCredentials() check and this call.
      */
-    public static void setCredentialsFetchedCallback(@Nullable Callback<Boolean> callback) {
+    public static void setCredentialsFetchedCallback(
+            @Nullable Callback<Boolean> callback, @Nullable Profile profile) {
         sCredentialsFetchedCallback = callback;
+        // If the fetch completed between the caller's isFetchingCredentials() check and this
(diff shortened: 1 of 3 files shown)


samartnik (approved):
++

PR #35611, opened 2026-04-16
[ads][CodeHealth] Use raw_ref for prefs_ and local_state_ members

tmancey:
Replace raw_ptr with raw_ref for all prefs_ and local_state_ members across VirtualPrefProvider, AdsServiceImpl, AdsServiceDelegate, AdsServiceImplIOS, and BraveStatsHelper. These members are always non-null at construction and are now typed to express that invariant. Update the VirtualPrefProvider free functions IsSurveyPanelist and GetSkus to take const PrefService& to eliminate conversion at call sites. Simplify the CHECK_IS_TEST guard in AdsServiceImpl to check only the nullable history_service_ pointer.

No behavioral changes.

Resolves


aseren (approved):

browser/brave_ads/ads_service_factory.cc
```diff
@@ -130,16 +131,17 @@ AdsServiceFactory::BuildServiceInstanceForBrowserContext(
       HostContentSettingsMapFactory::GetForProfile(profile);
 
   auto http_client = std::make_unique<HttpClient>(
-      *local_state,
+      CHECK_DEREF(local_state),
```
cdesouza-chromium: I'm not particularly fond of these uses of `CHECK_DEREF` because `g_browser_process->local_state()` can never be `nullptr` and the same goes for `profile->GetPrefs()`, so I feel these redundant checks give the wrong impression to the reader.

Maybe a second opinion should be sought on this, beyond mine. My reasoning with this is that although we do check invariants with these macros, I think there's a certain amount of reasonable assumptions we can rely on otherwise the whole application is broken.

If we were receiving these pointers from unknown sources, then I would understand the use of `CHECK_DEREF`, but we do know that these functions don't return `nullptr`.

tmancey: Thanks let me remove those

browser/brave_browser_process_impl.cc
```diff
@@ -484,7 +485,7 @@ brave_stats::BraveStatsUpdater* BraveBrowserProcessImpl::brave_stats_updater() {
 brave_ads::BraveStatsHelper* BraveBrowserProcessImpl::ads_brave_stats_helper() {
   if (!brave_stats_helper_) {
     brave_stats_helper_ = std::make_unique<brave_ads::BraveStatsHelper>(
-        local_state(), profile_manager());
+        CHECK_DEREF(local_state()), profile_manager());
```
cdesouza-chromium: Same here.

ios/browser/brave_ads/ads_service_factory_ios.mm
```diff
@@ -36,7 +37,7 @@
 std::unique_ptr<KeyedService> AdsServiceFactoryIOS::BuildServiceInstanceFor(
     ProfileIOS* profile) const {
   CHECK(profile);
-  return std::make_unique<AdsServiceImplIOS>(profile->GetPrefs());
+  return std::make_unique<AdsServiceImplIOS>(CHECK_DEREF(profile->GetPrefs()));
```
cdesouza-chromium, line 40: Same here.

cdesouza-chromium (approved):
LGTM % certain checks.

netzenbot (approved):

browser/brave_ads/ads_service_factory.cc
tmancey: Fixed

browser/brave_browser_process_impl.cc
tmancey: Fixed

ios/browser/brave_ads/ads_service_factory_ios.mm
tmancey: Fixed

kylehickinson (approved):
ios++

PR #35614, opened 2026-04-17
Jest - use guessConfig to use brave-core config and env variables

petemill:
Remove hard-coded list of possible platforms and build type configurations. Instead rely on the same mechanics as regular build (and storybook)

### DIFF
--- a/build/commands/lib/guessConfig.js
+++ b/build/commands/lib/guessConfig.js
@@ -34,16 +34,19 @@ let outputPath = config.outputDir
 function getBuildOutputPathList() {
   if (os.platform() === 'win32') {
     return buildConfigs.flatMap((config) => [
-      path.win32.resolve(__dirname, `..\\..\\..\\..\\out\\${config}`),
+      path.win32.resolve(import.meta.dirname, `..\\..\\..\\..\\out\\${config}`),
       ...extraArchitectures.map((arch) =>
-        path.win32.resolve(__dirname, `..\\..\\..\\..\\out\\${config}_${arch}`),
+        path.win32.resolve(
+          import.meta.dirname,
+          `..\\..\\..\\..\\out\\${config}_${arch}`,
+        ),
       ),
     ])
   } else {
     return buildConfigs.flatMap((config) => [
-      path.resolve(__dirname, `../../../../out/${config}`),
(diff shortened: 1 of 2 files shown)


fallaciousreasoning (approved):
this lgtm if it works

netzenbot (approved):

diracdeltas (approved):
also lgtm if it works

build/commands/lib/guessConfig.js
```diff
@@ -34,16 +34,19 @@ let outputPath = config.outputDir
 function getBuildOutputPathList() {
   if (os.platform() === 'win32') {
     return buildConfigs.flatMap((config) => [
-      path.win32.resolve(__dirname, `..\\..\\..\\..\\out\\${config}`),
+      path.win32.resolve(import.meta.dirname, `..\\..\\..\\..\\out\\${config}`),
```
goodov: import.meta.dirname is not very supported currently.. since this helper is already importing config.ts, use predefined paths from it:
```
(config.srcDir, `out\\...`)
```

petemill: It was added in node 20 and 21. And we enforce `"node": ">=24.11.1 <25.0.0"`. That ok? Just want to make one change at a time since a few things use this.

goodov: the problem is not Node, it's `jest`+`ts-jest` combo. In my experience it didn't work right for `meta.dirname` no matter what I tried.

petemill: Yeah I didn't realize guessConfig was used inside a test too (lit mangler) so I've changed this. Will see what happens with CI.

build/commands/lib/guessConfig.js
```diff
@@ -5,7 +5,7 @@
 
 import config from './config.ts'
 import os from 'node:os'
-import path from 'node:path/posix'
+import path from 'node:path'
```
petemill, line 7: We don't remember why we're using `path/posix` and `path/win32` instead of just `path` in this module, so I've pushed this to run on all CI platforms

goodov (approved):

goodov:
let's not try to solve this with other workarounds. We need to remove guessConfig entirely, I'll try to do this in near future.

PR #35646, opened 2026-04-17
[ads][CodeHealth] Remove redundant callback type aliases

tmancey:
Removes duplicate aliases from ads_service_callback.h and replaces remaining redundant aliases throughout the component with ResultCallback.

No behavioral changes.

Resolves

### DIFF
--- a/components/brave_ads/core/browser/service/BUILD.gn
+++ b/components/brave_ads/core/browser/service/BUILD.gn
@@ -3,6 +3,8 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at https://mozilla.org/MPL/2.0/.
 
+import("//brave/components/brave_ads/buildflags/buildflags.gni")
+
 # The target provides the AdsService interface and related types without
 # depending on Brave Ads implementation targets, allowing it to be used even
 # when Brave Ads is disabled by build flag.
--- a/components/brave_ads/core/browser/service/ads_service.h
+++ b/components/brave_ads/core/browser/service/ads_service.h
@@ -85,7 +85,7 @@ class AdsService : public KeyedService {
 
   // Called to clear ads data. The callback takes one argument - `bool`
   // is set to `true` if successful otherwise `false`
-  virtual void ClearData(ClearDataCallback callback) = 0;
+  virtual void ClearData(ResultCallback callback) = 0;
 
   // Called to add an ads observer.
   virtual void AddBatAdsObserver(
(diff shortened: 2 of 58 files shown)


components/brave_ads/core/browser/service/BUILD.gn
```diff
@@ -21,6 +23,10 @@ source_set("service") {
     "//components/keyed_service/core",
     "//mojo/public/cpp/bindings",
   ]
+
+  if (enable_brave_ads) {
```
netzenbot, line 27: The `service` target is explicitly designed to compile when `enable_brave_ads=false` (see the BUILD.gn comment), but this change makes `ads_service.h` unconditionally include `ads_callback.h`, which has `static_assert(BUILDFLAG(ENABLE_BRAVE_ADS))`. The conditional `public_deps` in BUILD.gn doesn't fix this — the static_assert fires before any dep resolution helps. The fix is to define `ResultCallback` in `ads_service_callback.h` (which is already in the always-built `:service` target) instead of pulling it from the ads-only `ads_callback.h`. ([Prefer Internal Feature Guards Over External Ifdefs](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-004))

components/brave_ads/core/browser/service/ads_service.h
```diff
@@ -16,6 +16,7 @@
 #include "brave/components/brave_ads/core/browser/service/ads_service_callback.h"
 #include "brave/components/brave_ads/core/browser/service/ads_service_observer.h"
 #include "brave/components/brave_ads/core/mojom/brave_ads.mojom-forward.h"
+#include "brave/components/brave_ads/core/public/ads_callback.h"
```
netzenbot, line 19: `ads_callback.h` contains `static_assert(BUILDFLAG(ENABLE_BRAVE_ADS))`, so including it unconditionally here breaks the guarantee stated in the class comment ("This header can be included even when Brave Ads is disabled by build flag"). The BUILD.gn correctly gates the dep on `enable_brave_ads`, but the `#include` also needs to be wrapped in `#if BUILDFLAG(ENABLE_BRAVE_ADS)` — or alternatively, `ResultCallback` and the other types used in method signatures need to be moved to a header that doesn't assert on the build flag. See [BS-012](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-012).

tmancey: Fixed

aseren (approved):

components/brave_ads/core/browser/service/BUILD.gn
tmancey: Fixed.

tmancey: Fixed

PR #35651, opened 2026-04-17
[Android] Honor Brave Rewards disabled-by-policy pref

### LINKED ISSUE #54703: [Android][Origin] Ads service keeps running and sponsored NTP images / Rewards retention notifications still render
On Android, the `brave.rewards.disabled_by_policy` managed pref is not honored by `brave_rewards::IsSupported()`. Brave Origin writes this pref for subscribers (via `BraveOriginServiceFactory`), but `brave_rewards::IsDisabledByPolicy()` is hardcoded to return false on Android inside `components/brave_rewards/core/rewards_util.cc`. As a result, any Android call site that gates on `IsSupported()` (as    
opposed to the Java-side `BraveRewardsPolicy.isDisabledByPolicy`) fails to suppress for Origin subscribers.

Java-side Rewards UI surfaces are unaffected — they already gate on `BraveRewardsPolicy.isDisabledByPolicy(profile)` directly (toolbar button, Rewards preferences, etc.). The leaks are in the paths that bottom out on the native IsSupported() helper.

SergeyZhukovsky:
`brave_rewards::IsDisabledByPolicy()` was hardcoded to return false on Android regardless of the managed `kDisabledByPolicy` pref, so every native call site that gates on `brave_rewards::IsSupported()` failed to suppress for Brave Origin subscribers (and enterprise admins setting the `BraveRewardsDisabled` policy).

Symptoms fixed:
- Ads service kept initializing on Android for Origin profiles (`AdsServiceFactory::GetForProfile` gates on `IsSupported`).
- NTP sponsored background images and sponsored rich-media takeover rendered (downstream of the ads service via ViewCounterService).
- DAY_10 / DAY_30 / DAY_35 Rewards retention notifications fired (`RetentionNotificationPublisher` gates on `rewardsNativeWorker .isSupported`).

Java-side Rewards UI surfaces were already correctly gated via `BraveRewardsPolicy.isDisabledByPolicy(profile)` and are unaffected.

Also decouples `BraveVpnUtils.isRegionSupported()` from `BraveRewardsNativeWorker.isSupported()` — otherwise the guard flip would cause VPN UI to hide for any profile where Rewards is policy-disabled as a side effect. Added `BraveVpnNativeWorker.isSupportedRegion` that does the OFAC region check directly via `brave_l10n`.

Resolves: https://github.com/brave/brave-browser/issues/54703

### DIFF
--- a/android/java/org/chromium/chrome/browser/vpn/BraveVpnNativeWorker.java
+++ b/android/java/org/chromium/chrome/browser/vpn/BraveVpnNativeWorker.java
@@ -176,6 +176,10 @@ public class BraveVpnNativeWorker {
         return BraveVpnNativeWorkerJni.get().isPurchasedUser(mNativeBraveVpnNativeWorker);
     }
 
+    public boolean isSupportedRegion() {
+        return BraveVpnNativeWorkerJni.get().isSupportedRegion(mNativeBraveVpnNativeWorker);
+    }
+
     public void getSubscriberCredentialV12() {
         BraveVpnNativeWorkerJni.get().getSubscriberCredentialV12(mNativeBraveVpnNativeWorker);
     }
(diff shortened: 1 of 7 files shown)


netzenbot (approved):

emerick (approved):
rewards++

PR #35657, opened 2026-04-18
fix: autoplay lock

### LINKED ISSUE #53976: Unable to play video when `disable autoplay` is enabled
### Description

If a user `Block sites from autoplaying videos` enabled in brave://settings/content/autoplay, they are still unable to play the video.

### Steps to reproduce

1. Go to brave://settings/content/autoplay 
2. Select `Block sites from autoplaying videos`
3. Go to https://brave.com/features/ or https://www.bulgari.com/en-us/
4. Try to play video

### Actual result

Videos do not play.
You get a prompt that says `Autoplay is blocked on this page`
Image

Clicking on the prompt gives you site settings but no controls or ability to enable autoplay on the site.
Image

To adjust the global setting or add a domain, a user has to dig pretty deep into the site settings 
`brave://settings` -> `Privacy and security` (brave://settings/privacy) -> `Site and Shield settings` (brave://settings/content) -> `Permissions` -> `Additional Permissions` -> `Autoplay` (brave://settings/content/autoplay)

### Expected result

Videos should be able to be played on user click.

_Suggestions:
Include a link to brave://settings/content/autoplay when the user clicks on the prompt or include options for users to allow autoplay 
ex._ 
Image

### Reproduces how often

Easily reproduced

1.87.192 Chromium: 145.0.7632.160 (Official Build) (arm64) 
1.89.118 Chromium: 146.0.7680.80 (Official Build) beta (arm64) 

### Channel information

- [x] release (stable)
- [x] beta
- [ ] nightly

### Reproducibility

### Miscellaneous information

related: https://github.com/brave/brave-browser/issues/34784

cdev-h:
Fixes https://github.com/brave/brave-browser/issues/53976
### About
---
Both BRAVE_AUTOPLAY_POLICY_IS_GESTURE_NEEDED_FOR_PLAYBACK and BRAVE_GET_AUTOPLAY_POLICY_FOR_DOCUMENT were causing issues when a user would click on a video who was relying solely on HTML5 (with no JS interaction, like Youtube) for autoplay. Also in very specific cases, like in Brave's website, the video itself was being display via an iframe, meaning the click event (in the case of a desktop interaction) would only propagate to the webpage itself, not the content of the iframe, which in turn would cause the video to be stuck even after the user's click.

I simply walk the frame's tree and check if there's an immediate action from the user or if there's a "sticky" action, if so I'm allowing for the video to be played regardless of the autoplay settings (they have already clicked on the video, thus giving us permissions to play it.)
### AI Disclosure
---
AI has been used to discover where the autoplay policies were being set and which functions to use to check on any possible interaction from the user on a specific frame.


chromium_src/third_party/blink/renderer/core/html/media/autoplay_policy.cc
```diff
@@ -6,13 +6,37 @@
...
 #include "third_party/blink/renderer/core/html/media/html_media_element.h"
+#include "third_party/blink/renderer/platform/wtf/casting.h"
 
 namespace blink {
 namespace {
 
+bool TreeHasUserActivation(Frame* frame) {
+  for (Frame* frame_video = frame; frame_video;
+       frame_video = frame_video->Tree().Parent()) {
+    if (frame_video->HasStickyUserActivation()) {
```
darkdh, line 17: Sticky activation is too broad without a gate: any gesture (scroll, link click, banner dismiss) sets it for the document's lifetime, so one click anywhere unlocks autoplay for every video on the page. As mentioned in [#discussion_r3150497761](https://github.com/brave/brave-core/pull/35657#discussion_r3150497761) upstream uses the same sticky walk in `AutoplayPolicy::IsDocumentAllowedToPlay` with gated Permissions Policy `kAutoplay`.

chromium_src/third_party/blink/renderer/core/html/media/autoplay_policy.cc
```diff
@@ -6,13 +6,37 @@
 #include "third_party/blink/public/platform/web_content_settings_client.h"
 #include "third_party/blink/renderer/core/frame/local_frame.h"
 #include "third_party/blink/renderer/core/html/media/html_media_element.h"
+#include "third_party/blink/renderer/platform/wtf/casting.h"
 
 namespace blink {
 namespace {
 
+bool TreeHasUserActivation(Frame* frame) {
+  for (Frame* frame_video = frame; frame_video;
```
darkdh, line 15: This duplicates `AutoplayPolicy::IsDocumentAllowedToPlay` ([autoplay_policy.cc;l=100](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/html/media/autoplay_policy.cc;l=100;drc=9c8071511cb8bb15fa3fa4845fb38e8490aa057a)), minus the Permissions Policy `kAutoplay` gate that makes the walk safe. Without that gate, once the user has interacted with the page **once** (sticky activation persists for the document's lifetime), a third-party iframe can poll `video.play()` indefinitely:

```js
// inside cross-origin ad iframe
setInterval(() => { video.play(); }, 1000);
```

— so the autoplay-block setting becomes a no-op after any gesture.

We should delegate this to upstream instead of re-implementing:

```cpp
bool IsAutoplayAllowedForFrame(LocalFrame* frame, bool play_requested) {
  if (!frame) return false;
  if (auto* document = frame->GetDocument();
      document && AutoplayPolicy::IsDocumentAllowedToPlay(*document)) {
    return true;
  }
  if (auto* settings_client = frame->GetContentSettingsClient()) {
    ...
  }
  return true;
}
```

Fixes the reported same-origin / `allow="autoplay"` cases, keeps Brave's block meaningful for ad iframes, and stays in sync with upstream automatically.

chromium_src/third_party/blink/renderer/core/html/media/autoplay_policy.cc
```diff
@@ -6,13 +6,37 @@
...
+  return false;
+}
+
 bool IsAutoplayAllowedForFrame(LocalFrame* frame, bool play_requested) {
-  if (!frame)
+  if (!frame) {
     return false;
+  }
+
+  if (TreeHasUserActivation(frame)) {
```
darkdh, line 36: The early return skips `frame->ClearUserActivation()`, which the original code uses to defensively reset activation after a block ("in order to block media when refresh or navigate" — see comment in the code below). Without it, activation can carry across refresh via `HadStickyUserActivationBeforeNavigation()`, weakening the autoplay-block setting post-reload. Note: the suggested `IsDocumentAllowedToPlay` delegation has the same exposure since that function also checks `HadStickyUserActivationBeforeNavigation`. Please add browser tests for: same-origin parent click → embedded play allowed; cross-origin iframe `.play()` without `allow="autoplay"` → blocked; reload after a click with autoplay-block on → blocked.

chromium_src/third_party/blink/renderer/core/html/media/autoplay_policy.cc
```diff
@@ -6,13 +6,37 @@
 #include "third_party/blink/public/platform/web_content_settings_client.h"
 #include "third_party/blink/renderer/core/frame/local_frame.h"
 #include "third_party/blink/renderer/core/html/media/html_media_element.h"
+#include "third_party/blink/renderer/platform/wtf/casting.h"
 
 namespace blink {
 namespace {
 
+bool TreeHasUserActivation(Frame* frame) {
```
darkdh: nit: this only walks ancestors, not the whole frame tree — `AncestorChainHasUserActivation` is more accurate. The loop variable `frame_video` is also misleading; it's just the current ancestor (`ancestor` would read better).

chromium_src/third_party/blink/renderer/core/html/media/autoplay_policy.cc
bridiver: also tests are definitely needed for this

PR #35661, opened 2026-04-19
Add one-liner horizontal tab layout setting

cmdr-chara:
## Summary
- Adds a new `brave.tabs.one_liner_enabled` pref for horizontal tabs.
- Adds an Appearance > Tabs toggle for the one-liner layout.
- Targets a 60px one-liner top chrome row by laying out the horizontal tab strip and toolbar on the same row when enabled and wide enough.

## Testing
- `git diff --check`
- Not run: full Brave build. A local Windows build setup was started, but the first Chromium sync/build was aborted because it would take hours on the available machine (I only have 16 GB RAM).

## Notes
This is an experimental first implementation intended to make the one-liner behavior testable. It may need visual follow-up for caption buttons, toolbar width allocation, and narrow-window behavior after a full build can be tested.


cmdr-chara:
Closing in favor of the clean-branch replacement PR: https://github.com/brave/brave-core/pull/35662.

PR #35680, opened 2026-04-20
[Android] Hide send_web_discovery toggle when managed by policy

### LINKED ISSUE #54751: [Android][Origin] Web Discovery Project toggle remains visible when managed by policy
On Android, the `Settings → Search engines → Web Discovery Project` toggle is supposed to be hidden when the `kWebDiscoveryEnabled` pref is managed by policy. It's not at the moment.

SergeyZhukovsky:
`BraveSearchEnginesPreferences.removePreferenceIfPresent` was calling `getPreferenceScreen().removePreference(pref)` unconditionally. That only removes direct children of `PreferenceScreen` — when the target preference is nested inside a `PreferenceCategory` (as `send_web_discovery` is, inside `currently_used_search_engines_category`), the remove call silently returns false and the toggle stays visible.
That PR removes from the preference's actual parent group. `preference.getParent()` returns the containing `PreferenceGroup` — the `PreferenceCategory` for nested prefs, the `PreferenceScreen` itself for top-level prefs. Works for both cases.

Other `PREF_CUSTOM_SEARCH_ENGINES_CATEGORY` caller is a top-level `PreferenceCategory`; new behavior is equivalent for that case.

Resolves: https://github.com/brave/brave-browser/issues/54751

### DIFF
--- a/android/java/org/chromium/chrome/browser/settings/BraveSearchEnginesPreferences.java
+++ b/android/java/org/chromium/chrome/browser/settings/BraveSearchEnginesPreferences.java
@@ -12,6 +12,7 @@ import android.view.View;
 
 import androidx.preference.Preference;
 import androidx.preference.PreferenceCategory;
+import androidx.preference.PreferenceGroup;
 
 import org.chromium.base.BraveFeatureList;
 import org.chromium.base.supplier.MonotonicObservableSupplier;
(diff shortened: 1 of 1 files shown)


samartnik (approved):
++

netzenbot (approved):

PR #35683, opened 2026-04-20
[iOS] Update Brave Origin product copy name

kylehickinson:
"Origin" → "Brave Origin"


kylehickinson:
More copy changes than first expected

PR #35686, opened 2026-04-20
[cr149] Removing uses of `OSCrypt` sync backend

### LINKED ISSUE #54755: Stop using sync backend for `OSCrypt`
The sync backend is being deleted in M149 (https://crrev.com/c/7765593), and we should migrate to the async backend all cases where we are using the synced one.

cdesouza-chromium:
This PR changes several places where there is still use of the sync
backend for `OSCrytp` to rely on the async backend. This is being done
in preparation for the deletion of the sync code in upstream
(https://crrev.com/c/7765593).

Resolves https://github.com/brave/brave-browser/issues/54755


components/brave_sync/brave_sync_prefs_unittest.cc
```diff
@@ -60,23 +59,16 @@ class BraveSyncPrefsTest : public testing::Test {
 // On macOS expected to see decryption failure when reading seed on
 // locked keyring
 TEST_F(BraveSyncPrefsTest, ValidPassphraseKeyringLocked) {
-  OSCryptMocker::SetUp();
-
   brave_sync_prefs()->SetSeed(kValidSyncCode);
 
   bool failed_to_decrypt = false;
-  OSCryptMocker::SetBackendLocked(true);
   EXPECT_EQ(brave_sync_prefs()->GetSeed(&failed_to_decrypt), "");
```
netzenbot: This test removes `OSCryptMocker::SetBackendLocked(true)` but `brave_sync_prefs.cc` still uses `OSCrypt::sync` for `GetSeed`/`SetSeed`. Without locking the backend, decryption succeeds and `failed_to_decrypt` stays `false`, so `EXPECT_TRUE(failed_to_decrypt)` will fail on Apple. Either keep the backend lock or update the test expectations to match the new behavior.

components/brave_sync/brave_sync_prefs_unittest.cc
```diff
@@ -60,23 +59,16 @@ class BraveSyncPrefsTest : public testing::Test {
 // On macOS expected to see decryption failure when reading seed on
 // locked keyring
 TEST_F(BraveSyncPrefsTest, ValidPassphraseKeyringLocked) {
-  OSCryptMocker::SetUp();
-
   brave_sync_prefs()->SetSeed(kValidSyncCode);
```
netzenbot, line 66: This test is broken by the change. `brave_sync_prefs.cc` still uses the sync `OSCrypt::DecryptString`, so `SetSeed` will encrypt and `GetSeed` will decrypt successfully — `failed_to_decrypt` will be `false`, failing `EXPECT_TRUE(failed_to_decrypt)`. Removing `OSCryptMocker::SetBackendLocked(true)` is only valid if `brave_sync_prefs.cc` is also migrated to the async interface in this PR, which it isn't.

components/brave_sync/brave_sync_prefs_unittest.cc
```diff
@@ -60,23 +59,16 @@ class BraveSyncPrefsTest : public testing::Test {
...
 
   bool failed_to_decrypt = false;
-  OSCryptMocker::SetBackendLocked(true);
   EXPECT_EQ(brave_sync_prefs()->GetSeed(&failed_to_decrypt), "");
   EXPECT_TRUE(failed_to_decrypt);
```
netzenbot: The test comment says it expects decryption failure on a locked keyring, but `OSCryptMocker::SetBackendLocked(true)` is removed while `brave_sync_prefs.cc` still uses `OSCrypt::DecryptString`. Without locking the backend, `GetSeed()` should succeed, causing `EXPECT_TRUE(failed_to_decrypt)` to fail. If `brave_sync_prefs.cc` is being migrated to the async encryptor in a follow-up, this test needs to be updated (or deleted) to reflect what the new failure path looks like — otherwise it will either fail or silently stop covering the locked-keyring scenario.

components/brave_sync/brave_sync_prefs_unittest.cc
cdesouza-chromium: Fixed.

components/brave_sync/brave_sync_prefs_unittest.cc
cdesouza-chromium: Fixed.

components/brave_sync/brave_sync_prefs_unittest.cc
cdesouza-chromium: Fixed.

netzenbot (approved):

PR #35688, opened 2026-04-20
[ads] Fix unsafe buffer access in DaysInMonth

tmancey:
Removes unsafe memory access from the month-length lookup and documents the proleptic Gregorian calendar rule in IsLeapYear.

No behavioral changes.

Resolves


cdesouza-chromium (approved):
LGTM

aseren (commented):
LGTM

PR #35689, opened 2026-04-20
[toolchain] Remove legacy rust WASM toolchain builder

### LINKED ISSUE #54478: [toolchain] Generate rust toolchain using vanilla Chromium
This issue tracks the work to have the rust toolchain for brave to be generated using vanilla Chromium, and with regular checks to make sure we the job starts on its own in CI.

Currently, to generate the toolchain, we use `npm run build_rust_toolchain_aux`. This command has some draw backs though:

- its brave-specific parts and checkout expectations make it harder to understand failures when they occur.
- Additional support for tags had to be added in a work around way to allows to generate toolchain for a tag that we still don't have functioning brave check out, which is every new tag where the toolchain has been updated.
- The tag workaround does not work on windows, due to issues relating to the use of `git reset --hard` by `build_rust_toolchain_aux`.

The proposal is to have a standalone python script for the toolchain building, that works with vanilla Chromium alone, and roll that out in CI.

Additionally, we should use the logic present in `brockit` for the rust toolchain to have regular checks for the latest toolchain.

cdesouza-chromium:
The new builder has been rolled out to all pipelines.

Resolves https://github.com/brave/brave-browser/issues/54478


diracdeltas (approved):
unblocking js deps review

netzenbot (approved):

goodov (approved):

PR #35694, opened 2026-04-20
[ads][CodeHealth] Remove redundant static_assert from internal headers

tmancey:
Internal headers are already gated by BUILD.gn deps so the ENABLE_BRAVE_ADS static_assert is redundant and causes build failures when the header is transitively included from an always-built target.

No behavioral changes.

Resolves

### DIFF
--- a/components/brave_ads/core/internal/account/confirmations/confirmation_type.h
+++ b/components/brave_ads/core/internal/account/confirmations/confirmation_type.h
@@ -8,10 +8,7 @@
 
 #include <string_view>
 
-#include "brave/components/brave_ads/buildflags/buildflags.h"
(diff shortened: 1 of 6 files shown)


netzenbot (approved):

aseren:
These static_assert help to identify such cases when we include ads only header from always-built target transitively.
If we need to do it, then we should consider moving ads only header to some always-built target.

tmancey:
> These static_assert help to identify such cases when we include ads only header from always-built target transitively. If we need to do it, then we should consider moving ads only header to some always-built target.

If this is the case this fix is correct because these files are internal. Lets merge this and raise a PR that moves internal files so they cannot be included.

aseren (approved):

PR #35698, opened 2026-04-20
[ads] Fix AdsServiceFactory returning nullptr when Rewards unavailable

tmancey:
AdsServiceFactory::GetForProfile returned nullptr when IsSupported
was false, which includes a managed pref that can change during a
session. The guard is removed from the factory so AdsService is
always created for regular profiles. BAT ads startup is now gated
in CanStartBatAdsService instead. AdsServiceImpl and
AdsServiceDelegate also handle null service dependencies gracefully.

A follow-up will add an ads-specific group policy to allow
independent control of ads without relying on the Rewards policy.

Resolves

### DIFF
--- a/browser/brave_adaptive_captcha/brave_adaptive_captcha_service_factory.cc
+++ b/browser/brave_adaptive_captcha/brave_adaptive_captcha_service_factory.cc
@@ -103,11 +103,17 @@ BraveAdaptiveCaptchaServiceFactory::BuildServiceInstanceForBrowserContext(
     content::BrowserContext* context) const {
   auto url_loader_factory = context->GetDefaultStoragePartition()
                                 ->GetURLLoaderFactoryForBrowserProcess();
+#if BUILDFLAG(ENABLE_BRAVE_REWARDS)
+  auto* rewards_service = brave_rewards::RewardsServiceFactory::GetForProfile(
+      Profile::FromBrowserContext(context));
+  if (!rewards_service) {
+    return nullptr;
+  }
+#endif
   return std::make_unique<BraveAdaptiveCaptchaService>(
       user_prefs::UserPrefs::Get(context), std::move(url_loader_factory),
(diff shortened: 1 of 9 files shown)


netzenbot (approved):

aseren (approved):

zenparsing (approved):
Rewards 👍

cdesouza-chromium (approved):
LGTM

browser/brave_adaptive_captcha/brave_adaptive_captcha_service_factory.cc
```diff
@@ -103,11 +103,17 @@ BraveAdaptiveCaptchaServiceFactory::BuildServiceInstanceForBrowserContext(
     content::BrowserContext* context) const {
   auto url_loader_factory = context->GetDefaultStoragePartition()
                                 ->GetURLLoaderFactoryForBrowserProcess();
+#if BUILDFLAG(ENABLE_BRAVE_REWARDS)
+  auto* rewards_service = brave_rewards::RewardsServiceFactory::GetForProfile(
```
bridiver, line 107: it's not clear to me how this is related to ads service factory?

bridiver: fyi according to @supermassive the rewards factory return value is stable over the browser session (the managed pref is cached per @DJAndries work and the "allow in private" pref requires a restart)

tmancey: `AdsServiceFactory` previously `CHECK`'d that `BraveAdaptiveCaptchaService` was non-null. Once the `IsSupported` guard was removed from the ads factory, ads can now be created for profiles where Rewards is unavailable, so the captcha service may not exist. The captcha factory change ensures it returns `nullptr` cleanly in that case, and the delegate was updated to handle a `null` captcha service rather than crashing.

PR #35703, opened 2026-04-20
[AI Chat]: Better editing approach

### LINKED ISSUE #54766: [AI Chat]: Associate content with edited turns
At the moment, content is only associated with the initial entry. As a stepping stone to being able to edit content associated with turns, each edit should get its own version of the associated content.

**Note:** This does not make it possible to edit the attached content (that will happen in a follow up).

fallaciousreasoning:
Resolves https://github.com/brave/brave-browser/issues/54766

### DIFF
--- a/components/ai_chat/core/browser/associated_content_manager.cc
+++ b/components/ai_chat/core/browser/associated_content_manager.cc
@@ -21,10 +21,12 @@
 #include "base/memory/weak_ptr.h"
 #include "base/strings/string_util.h"
 #include "base/strings/utf_string_conversions.h"
+#include "base/uuid.h"
 #include "brave/components/ai_chat/core/browser/associated_archive_content.h"
 #include "brave/components/ai_chat/core/browser/associated_content_delegate.h"
 #include "brave/components/ai_chat/core/browser/conversation_handler.h"
 #include "brave/components/ai_chat/core/browser/model_service.h"
+#include "brave/components/ai_chat/core/common/utils.h"
 
 namespace ai_chat {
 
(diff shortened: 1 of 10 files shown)


components/ai_chat/core/browser/conversation_handler.cc
```diff
@@ -748,6 +748,12 @@ void ConversationHandler::ModifyConversation(const std::string& entry_uuid,
     return;
   }
 
+  // Capture the source UUID before adding a new edit to the list.
+  const std::string source_turn_uuid = (turn->edits && !turn->edits->empty() &&
```
fallaciousreasoning: surely this is just the id of the turn we're editing?

fallaciousreasoning: its not - edited turn is the first one

components/ai_chat/core/browser/associated_content_manager.cc
```diff
@@ -215,11 +217,40 @@ void AssociatedContentManager::AssociateUnsentContentWithTurn(
...
+  std::vector<AssociatedContentDelegate*> source_delegates;
+  for (auto* delegate : content_delegates_) {
+    auto it = content_uuid_to_conversation_turns_.find(delegate->uuid());
+    if (it != content_uuid_to_conversation_turns_.end() &&
+        it->second == source_turn_uuid) {
+      source_delegates.push_back(delegate);
+    }
+  }
+
+  for (auto* delegate : source_delegates) {
```
netzenbot, line 244: nit: `new_uuid` is never mutated after initialization — `const auto new_uuid = ...` would communicate that intent. See [CS-064](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-064).

components/ai_chat/core/browser/associated_content_manager.cc
```diff
@@ -215,11 +217,40 @@ void AssociatedContentManager::AssociateUnsentContentWithTurn(
     const mojom::ConversationTurnPtr& turn) {
   CHECK(turn->uuid.has_value());
```
yrliou, line 218: Should we update this CHECK since we might be using edit turn's UUID?

components/ai_chat/core/browser/conversation_handler.cc
```diff
@@ -769,6 +770,11 @@ void ConversationHandler::ModifyConversation(const std::string& entry_uuid,
       std::back_inserter(erased_turn_ids),
       [](mojom::ConversationTurnPtr& turn) { return turn->uuid; });
   turn->edits->emplace_back(std::move(edited_turn));
+  // Clone content from the previous version of this turn to the new edit so
+  // the edit carries forward the same associated content while preserving the
+  // original association for potential revert.
+  associated_content_manager_->CloneContentForTurnEdit(source_turn_uuid,
```
yrliou, line 776: As mentioned in the DM, let's explore a way to share the actual contents between edits, maybe reference through IDs, instead of direct clones.

petemill: I think that's going to add a lot of complexity that we don't need. Adding a many:many relationship is not necessary given we're not displaying previous versions of the edit. And even if we were, we are deleting the original sibling entries. If we add UI for previous versions of entries, we'll have to have a DB design around threads, and can plan at that point what relationship we want between the original entry attachments and the new thread's attachments.

I'd be fine if we either:
- Delete the base entry content and move the references to the edit
- Keep the base entry content and an edit will always have the base entry's content attached.

Given that there is no attachment button for edits, I think that's fine, and is the same as what happens now.

When we go to re-design the edit to be in the parent frame, we can give it an attachment icon and decide what to do (probably the first option: delete base attachments, move to edit).

When we want to display the base thread and the edit threads, then we might need a many:many or similar design - although I'd probably be fine if we cloned since that's what happens at the moment with the same content attached to multiple conversations.

yrliou: I'm fine as long as we don't have multiple copies for all edits we made. The fact that we're not displaying and let user go switch between edits at the moment is exactly what makes multiple copies in all edits more wasteful without real benefits and should be avoided for attachments as it might be big.
I'm fine with "Delete the base entry content and move the references to the edit".
For "Keep the base entry content and an edit will always have the base entry's content attached.", does it work? I thought we want to allow user being able to delete base attachments.

petemill: Yes eventually when we add an attachment button to edits (and edits are in the parent frame), we need to do the first option (move the references to the edit) so that the user can remove or add to the list. But as an iteration step, to unblock inline mentions of content (https://github.com/brave/brave-core/pull/32717), we can do the second option since the user can't remove or add attachments during the edit anyway.

PR #35723, opened 2026-04-21
[Wallet] remove restricted address error messaging

cmazakas:
Because we're interested in not leaking information about an address being restricted, we remove the helper function WalletRestrictedAddressErrorMessage and in its stead we just use WalletInternalErrorMessage.


supermassive (approved):

mkarolin (approved):
`strings`++

PR #35727, opened 2026-04-21
[CodeHealth] Remove unused API from SerpMetricsTimePeriodStorage

aseren:
The deprecated PrefService-based constructors and five methods with no production callsites were carried over from TimePeriodStorage. The UTC and DST-offset modes were always disabled in every caller, and the internal clock was always the system clock, leaving them dead code. All are removed; tests are migrated to the store-based constructor and a plain fixture.

This is a pre-requisite for https://github.com/brave/brave-browser/issues/54758.

Resolves

### DIFF
--- a/components/serp_metrics/serp_metrics.cc
+++ b/components/serp_metrics/serp_metrics.cc
@@ -49,9 +49,7 @@ BuildTimePeriodStorages(
     time_period_storages.emplace(
         type,
         std::make_unique<SerpMetricsTimePeriodStorage>(
-            std::move(time_period_store), kSerpMetricsTimePeriodInDays.Get(),
-            /*should_use_utc=*/false,
-            /*should_offset_dst=*/false));
+            std::move(time_period_store), kSerpMetricsTimePeriodInDays.Get()));
   }
 
   return time_period_storages;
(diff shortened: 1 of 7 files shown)


tmancey (approved):

aseren:
> ### Possible Issues
> * **`base::Time::Now()` in production code** — removed `base::Clock` injection means `FilterToPeriod()` and `GetPeriodSum()` call `base::Time::Now()` directly. Tests rely on `TaskEnvironment::AdvanceClock` which controls `base::Time::Now()`, so existing tests pass, but any future test needing a custom clock must use a different pattern (e.g. `TaskEnvironment` time source). This is a testability regression.

Current tests didn't use the constructor with base::Clock, so constructor with `base::Clock` injection was a dead code.
Future tests will use TaskEnvironment for tests.

> * **`GetPeriodSumInTimeRange` range boundary change** — the DST offset padding (`kPotentialDSTOffset`) was removed from the accumulator lambda. Callers that previously relied on the ±1 hour tolerance around bucket boundaries will now see different (smaller) sums near DST transitions. If any external caller depended on this, metrics could silently drop a bucket.

kPotentialDSTOffset is not used for SERP metrics

PR #35740, opened 2026-04-21
[iOS] Enable smooth scrolling in BraveWebClient

### LINKED ISSUE #54808: White line showing at bottom of browser when top bar is enabled (Beta build)
### Description

When using the top address bar on iOS, a thin white line appears at the bottom of the browser. Possibly caused by cr147 update. User is notably seeing this in the Beta build.

Image

### Steps to reproduce

1. Update Brave to latest iOS build
2. Enable top bar view
3. Scroll the page so that the toolbar collapses

### Actual result

White line/bar appears on bottom of browser

### Expected result

No line should appear

### Reproduces how often

Easily reproduced

### Brave version

v1.88

### Device/iOS version

iOS 26.4.1 (23E254) on iPhone 17
Brave Version 1.90.111

### Affected browser versions

- [ ] latest AppStore
- [x] latest TestFlight
- [ ] previous TestFlight

### Reproducibility

### Miscellaneous information

https://community.brave.app/t/white-bottom-bar-brave-beta/652047/4

kylehickinson:
Brave handles the web views frame while scrolling on the Swift side so we need to return true for the `IsSmoothScrollingSupported` web client method so that it doesn't automatically inset the web view by the safe area

```
Chromium change:
commit 0b9166658b79a5f788860e3e215165d8d26d73b1
Author: Gauthier Ambard <gambard@chromium.org>
Date:   Mon Mar 9 08:40:04 2026 -0700

    [iOS] Replace use of SmoothScrolling flag with the provider

    This is how it is supposed to be used. Make sure to move all the uses of
    the smooth scrolling flag at the same time to have a state where half
    the app is using the flag and the other app the provider.

    Also, updates the test to ensure they only run when the provider is
    returning true.

    The download manager test is updated as the tap was failing.

    Bug: 483998779
    Change-Id: Ic81a977ecbc0ebc131c49bd992511a97eb916abb
    Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/7648409
    Reviewed-by: Aliona Dangla <alionadangla@chromium.org>
    Commit-Queue: Gauthier Ambard <gambard@chromium.org>
    Cr-Commit-Position: refs/heads/main@{#1596357}
```

Resolves https://github.com/brave/brave-browser/issues/54808

### DIFF
--- a/ios/browser/web/brave_web_client.h
+++ b/ios/browser/web/brave_web_client.h
@@ -62,6 +62,7 @@ class BraveWebClient : public ChromeWebClient {
 
   void DidResetConfiguration(web::BrowserState* browser_state,
                              WKWebViewConfiguration* configuration) override;
+  bool IsSmoothScrollingSupported() const override;
 
  private:
   std::string legacy_user_agent_;
--- a/ios/browser/web/brave_web_client.mm
+++ b/ios/browser/web/brave_web_client.mm
@@ -267,3 +267,7 @@ void BraveWebClient::DidResetConfiguration(web::BrowserState* browser_state,
     BraveWebView.didResetConfiguration(profile_bridge, config);
   }
 }
+
+bool BraveWebClient::IsSmoothScrollingSupported() const {
+  return true;
+}


StephenHeaps (approved):

Uni-verse:
Verified on `iPhone 12` running `iOS 26.3.1` using version `1.91 (95)`

Reproduce the issue based on the STR in 1.91 (91) and the upgraded to 1.91 (95) and verified that the white bar is not shown when using the steps to reproduce in https://github.com/brave/brave-browser/issues/54808#issue-4305254546
   - Additionally verified that the issue does not occur when enabling `Reader Mode`

**`1.91 (91)`** | **`1.91 (95)`** | **`1.91 (95)`**
--|--|--
IMG_0227 3 | IMG_0228 3 | IMG_0229 2

PR #35741, opened 2026-04-21
Vendor BoringTun 3p dependency

### LINKED ISSUE #54590: Vendor BoringTun 3p dependency

alekspop26:
Resolves https://github.com/brave/brave-browser/issues/54590

Add boringtun dependency to the brave-core, to be used later in the VPN 2.0 architecture by a privileged helper on desktop OSes. Add a dependency to the existing VPN to ensure the dependency actually gets built with the browser (but not distributed). Builds only for Windows and macOS, since VPN is not enabled on Linux. Tested on Linux locally.

A separate commit was created for "vendor" directory since the crates were auto-generated by `npm run update_boringtun_resources` command and license creation script ("script/boringtun_create_licenses.py"). No manual intervention into that commit.

### DIFF
--- a/browser/BUILD.gn
+++ b/browser/BUILD.gn
@@ -39,6 +39,14 @@ source_set("browser_process") {
     "//chrome/browser:browser_process",
     "//extensions/buildflags",
   ]
+
+  # Make sure boringtun keeps getting built even if we don't use
+  # and don't distribute it (for now). Keep it as browser level
+  # dependency, so we also build for Linux (where Brave VPN is not
+  # yet enabled).
+  if (is_mac || is_win || is_linux) {
+    deps += [ "//brave/third_party/boringtun" ]
+  }
 }
 
 source_set("brave_global_features") {
--- a/browser/brave_vpn/BUILD.gn
+++ b/browser/brave_vpn/BUILD.gn
@@ -34,10 +34,4 @@ source_set("brave_vpn") {
   if (is_mac) {
     deps += [ "//brave/browser/brave_vpn/mac" ]
   }
-
-  # Make sure boringtun keeps getting built even if we don't use
-  # and don't distribute it (for now).
-  if (is_mac || is_win || is_linux) {
-    deps += [ "//brave/third_party/boringtun" ]
-  }
 }
--- a/browser/brave_vpn/BUILD.gn
+++ b/browser/brave_vpn/BUILD.gn
@@ -34,4 +34,10 @@ source_set("brave_vpn") {
   if (is_mac) {
     deps += [ "//brave/browser/brave_vpn/mac" ]
   }
+
+  # Make sure boringtun keeps getting built even if we don't use
+  # and don't distribute it (for now).
+  if (is_mac || is_win || is_linux) {
+    deps += [ "//brave/third_party/boringtun" ]
+  }
 }
--- a/build/commands/scripts/updateBoringtunResources.ts
+++ b/build/commands/scripts/updateBoringtunResources.ts
@@ -1,0 +1,15 @@
+// Copyright (c) 2026 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import path from 'node:path'
+
+import config from '../lib/config.ts'
+import util from '../lib/util.js'
+
+util.run(
+  'vpython3',
+  [path.join(config.srcDir, 'brave', 'third_party', 'boringtun', 'update.py')],
+  config.defaultOptions,
+)
--- a/third_party/boringtun/.cargo/config.toml
+++ b/third_party/boringtun/.cargo/config.toml
@@ -1,0 +1,44 @@
+# Cargo configuration for the BoringTun vendored build.
+#
+# CARGO_HOME and CARGO_TARGET_DIR are set by build.py to isolated,
+# Brave-specific locations -- this build never reads or writes Chromium's
+# shared Rust caches.
+
+# -----------------------------------------------------------------------------
+# Per-target rustflags
+# -----------------------------------------------------------------------------
+# Disable AVX and wider SIMD to ensure compatibility with the widest
+# possible range of hardware. "x86-64" is the baseline x86_64 feature set
+# (SSE2 only). For ARM, "generic" avoids any microarch-specific tuning.
+
+[target.i686-pc-windows-msvc]
+rustflags = ["-C", "target-feature=+crt-static", "-C", "target-cpu=pentium4", "-C", "target-feature=-avx,-avx2,-avx512f"]
+
+[target.x86_64-pc-windows-msvc]
+rustflags = ["-C", "target-feature=+crt-static", "-C", "target-cpu=x86-64", "-C", "target-feature=-avx,-avx2,-avx512f"]
+ 
+[target.aarch64-pc-windows-msvc]
+rustflags = ["-C", "target-feature=+crt-static", "-C", "target-cpu=generic"]
+
+[target.x86_64-unknown-linux-gnu]
+rustflags = ["-C", "target-cpu=x86-64", "-C", "target-feature=-avx,-avx2,-avx512f"]
+
+[target.aarch64-unknown-linux-gnu]
+rustflags = ["-C", "target-cpu=generic"]
+
+[target.x86_64-apple-darwin]
+rustflags = ["-C", "target-cpu=x86-64", "-C", "target-feature=-avx,-avx2,-avx512f", "-C", "link-arg=-Wl,-undefined,error"]
+
+[target.aarch64-apple-darwin]
+rustflags = ["-C", "target-cpu=apple-m1", "-C", "link-arg=-Wl,-undefined,error"]
+
+[source.crates-io]
+replace-with = "vendored-sources"
+
+[source."git+https://github.com/cloudflare/boringtun.git?rev=cdf3b24558441ff4dc62bf5992484c5457a36760"]
+git = "https://github.com/cloudflare/boringtun.git"
+rev = "cdf3b24558441ff4dc62bf5992484c5457a36760"
+replace-with = "vendored-sources"
+
+[source.vendored-sources]
+directory = "vendor/"
--- a/third_party/boringtun/.cargo/config.toml
+++ b/third_party/boringtun/.cargo/config.toml
@@ -12,25 +12,63 @@
 # (SSE2 only). For ARM, "generic" avoids any microarch-specific tuning.
 
 [target.i686-pc-windows-msvc]
-rustflags = ["-C", "target-feature=+crt-static", "-C", "target-cpu=pentium4", "-C", "target-feature=-avx,-avx2,-avx512f"]
+rustflags = [
+  "-C", "prefer-dynamic=false",
+  "-C", "target-feature=+crt-static",
+  "-C", "target-cpu=pentium4",
+  "-C", "target-feature=-avx,-avx2,-avx512f"
+]
 
 [target.x86_64-pc-windows-msvc]
-rustflags = ["-C", "target-feature=+crt-static", "-C", "target-cpu=x86-64", "-C", "target-feature=-avx,-avx2,-avx512f"]
+rustflags = [
+  "-C", "prefer-dynamic=false",
+  "-C", "target-feature=+crt-static",
+  "-C", "target-cpu=x86-64",
+  "-C", "target-feature=-avx,-avx2,-avx512f"
+]
  
 [target.aarch64-pc-windows-msvc]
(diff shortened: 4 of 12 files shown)


fmarier (approved):
`script/brave_license_helper.py` changes :+1:

third_party/boringtun/update.py
szilardszaloki, line 1: Since this is mostly a copy of [third_party/wasm/update.py](https://github.com/brave/brave-core/blob/master/third_party/wasm/update.py), and it looks like we're starting to add vendored Rust workspaces beyond just WASM, I'm wondering if this would be a good time to pull the shared bits into a reusable helper.

third_party/boringtun/create_licenses.py
szilardszaloki, line 1: This seems mostly generic apart from a few BoringTun-specific bits. I think we should parameterize those and share it across vendored Rust workspaces instead of adding a one-off script.

third_party/boringtun/build.py
szilardszaloki, line 1: `build.py` writes into the source tree during the GN action — `brave/third_party/boringtun/bin/` and `brave/third_party/boringtun/include/` (and these outputs aren't declared to GN either).

alekspop26: Great points; I think we actually don't need to write to the source tree. Will fix.

third_party/boringtun/update.py
alekspop26: I'm not sure we're literally **starting** to add vendored out-of-tree dependencies. Boringtun is an exception here: we're doing this since we don't want to ship binaries only for VPN, which will make our security/privacy audit much more complicated. We don't want to pollute the rest of Brave Rust crates as well. Hence, I'd prefer to keep this update script scoped to Boringtun.
But I agree, that if we decide that this is a common way to vendor new dependencies, we should revisit the scripting framework to make it shared.

third_party/boringtun/create_licenses.py
alekspop26: Agree, but since it's more of an improvement than a blocker, I'd rather fix it in a follow-up task, for which I can cut a new ticket. Would that work?

alekspop26:
Here are Boringtun build times collected from the CI runs:
| CI task | Platform | Builds stdlib? | Time, seconds |
| --- | --- | --- | ---: |
| continuous-integration/windows-x86/pr-head | win x86 | Yes | 11.7 |
| continuous-integration/windows-x64/pr-head | win x64 | No | 11.1 |
| continuous-integration/windows-arm64/pr-head | win arm64 | Yes | 12.1 |
| continuous-integration/origin-windows-x64/pr-head | win x64 | No | 11.2 |
| continuous-integration/origin-linux-x64/pr-head | linux x64 | No | 5.1 |
| continuous-integration/macos-x64/pr-head | mac x64 | No | 19.8 |
| continuous-integration/macos-arm64/pr-head | mac arm64 | No | 5.1 |
| continuous-integration/origin-macos-arm64/pr-head | mac arm64 | No | 11.7 |
| PR (component) | linux x64 | No | 7.6 |
| PR (component) | mac x64 | Yes | 6.2 |
| PR (component) | win x64 | Yes | 9.9 |
| PR (component) | linux-nodchecks x64 | No | 6.1 |
| PR linux (x64) | linux x64 | No | 7.5 |
| PR linux (arm64) | linux arm64 | Yes | 7.6 |

alekspop26:
# BoringTun vendoring: reviewer notes

## Cross-arch + bundled toolchain mismatch

Brave's bundled Rust toolchain ships only host-arch rust-std. For non-host triples we use `-Zbuild-std` to compile std from rust-src (with `RUSTC_BOOTSTRAP=1`). This requires merging two cargo vendor sources — the project's `vendor/` and rust-src's `library/vendor/` — into one directory cargo can read. We do this via symlinks under `target/.merged-vendor/`, ephemeral, never committed.

## Per-target rustflags don't reach host build scripts

When `--target` is passed, cargo applies `[target.<X>].rustflags` to the target artifacts but **not** to host build script binaries. Linker config (`CARGO_TARGET_<TRIPLE>_LINKER`) does propagate. So we put cross-cutting linker flags in a wrapper script, not in rustflags — the wrapper is what cargo invokes for both target and host artifacts.

## Linker wrapper

`make_linker_wrapper` writes a per-platform `.cmd`/`.sh` that prepends fixed flags (`-fuse-ld=lld`, `--target=`, `--sysroot=`, `-mmacosx-version-min=`) before forwarding to the real linker. Lives under `target/.tool-wrappers/`. Naming matters: filenames containing `ld` or `lld` trigger rustc/lld-link argv-flavor heuristics; we use neutral names (`clang-driver.sh`, `msvc-link.cmd`).

## `-fuse-ld=lld` everywhere except Windows

- **Mac**: clang's default linker discovery runs `xcodebuild -find ld`, which fails on minimal-Xcode CI runners.
- **Linux**: Chromium's bundled host ld (newer binutils) can't link against the pinned old debian sysroot's libs (versioned-symbol mismatches). lld is more permissive and matches Chromium's own C++ build.
- **Windows**: rustc invokes `lld-link.exe` directly with MSVC syntax — `-fuse-ld` is a clang-driver concept that doesn't apply.

`lld-link.exe` tolerates a leading `-flavor link` from rustc only when it's the first arg, so the Windows linker wrapper appends our injected flags **after** `%*`, not before.

## Windows clang shim

Brave's bundled LLVM on Windows ships only `clang-cl.exe`, not `clang.exe`. ring's build.rs explicitly invokes bare `clang` for `aarch64-pc-windows-msvc` (it forces clang-driver mode for that target's GNU-style C). We link `clang-cl.exe` → `clang.exe` in a shim dir on PATH; argv[0] basename selects driver mode (cl vs gcc), so the same binary serves both personalities.

## Windows sysroot — split channel for the two compiler personalities

ring's compiler override inherits cc-rs's accumulated CFLAGS, but the two personalities accept different sysroot flags:

- **clang-cl** (cc-rs's normal path): `/winsysroot:<path>` — works.
- **clang** (ring's override): `/winsysroot:` is treated as an input filename — fails.

Putting either form in `CFLAGS_<triple>` breaks one path. Solution: per-personality wrappers. `clang-cl.cmd` and `clang.cmd` each inject the sysroot in the dialect their target compiler understands (`/winsysroot:` vs `-Xmicrosoft-windows-sys-root`). CFLAGS stays minimal — just `--target=`. Linker side stays simple: rustc → lld-link with `/winsysroot:` (linker isn't subject to ring's override).

## Hermetic env

`os.environ.copy()` then mutate the copy. The cargo subprocess sees our `CC_<triple>`, `AR_<triple>`, `CARGO_TARGET_<TRIPLE>_LINKER`, etc.; nothing leaks back to the parent process or sibling GN actions. Other GN actions don't read these vars anyway (Chromium's C++ build uses GN-generated ninja rules with explicit paths).

atuchin-m:
The PR adds +1.7M new lines of Rust code.
Is it okay to commit all that code to `brave-core` or it make sense to use some other way like DEPS?

It's not a new issue and currently `brave-core` mostly consist of bundled rust code (~80% of total lines in source code) But this PR will add extra 20%.
```
mikhail@am-macbook brave % git ls-files '*.cc' '*.cpp' '*.cxx' '*.h' '*.hh' '*.hpp' '*.hxx' '*.inc' '*.ts' '*.js' '*.py' '*.java' '*.go' '*.cs' '*.swift' '*.m' '*.mm' | xargs cat | wc -l 
 2114194
 
mikhail@am-macbook brave % git ls-files git ls-files '*.rs' | xargs cat | wc -l
 9811857
```

third_party/boringtun/.cargo/config.toml
```diff
@@ -0,0 +1,82 @@
+# Cargo configuration for the BoringTun vendored build.
+#
+# CARGO_HOME and CARGO_TARGET_DIR are set by build.py to isolated,
+# Brave-specific locations -- this build never reads or writes Chromium's
+# shared Rust caches.
+
+# -----------------------------------------------------------------------------
+# Per-target rustflags
+# -----------------------------------------------------------------------------
+# Disable AVX and wider SIMD to ensure compatibility with the widest
```
bridiver, line 10: I don't think this is necessary, this only needs to be compatible with the same hardware that Brave is compatible with and most of these flags are not passed in the chromium build. I think we want to generate the rustflags in gn and pass them directly to cargo instead of having a static config.toml because the flags also depend on the build type (debug, official build, etc...)

alekspop26: I assume that we always build release boringtun regardless of the Brave build. It acutally makes sense since we're not currently going to debug boringtun library itself, and if we want to do that locally - we can always rebuild using the custom flags.
As for hardware, by default, rust tends to choose the best available processor features which once led to the dll crashing on @bsclifton 's machine. So I decided to force the most compatible feature set here.

build/commands/scripts/updateBoringtunResources.ts
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) 2026 The Brave Authors. All rights reserved.
```
bridiver, line 1: I guess wasm update also has an npm wrapper, but it really doesn't seem necessary. Since we're just calling a python script we can just call it directly

third_party/boringtun/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+description = "Internal pin of cloudflare/boringtun for Brave's vendored cdylib build."
+
+[lib]
+path = "src/lib.rs"
+crate-type = ["rlib"]
+
+[dependencies]
+boringtun = { git = "https://github.com/cloudflare/boringtun.git", rev = "cdf3b24558441ff4dc62bf5992484c5457a36760", default-features = false, features = ["ffi-bindings"] }
+
+[profile.release]
```
bridiver, line 38: same as https://github.com/brave/brave-core/pull/35741/changes#r3174304889 let's generate these in gn and pass them to cargo as args

browser/brave_vpn/BUILD.gn
```diff
@@ -34,4 +34,10 @@ source_set("brave_vpn") {
   if (is_mac) {
     deps += [ "//brave/browser/brave_vpn/mac" ]
   }
+
+  # Make sure boringtun keeps getting built even if we don't use
+  # and don't distribute it (for now).
+  if (is_mac || is_win || is_linux) {
```
cdesouza-chromium: This needs a gn build flag, and I would recommend to also have a flag to enable it in our build... I suspect this can be in the codebase but not part of the main build for now, so all this work should be behind a build flag.

third_party/boringtun/BUILD.gn
```diff
@@ -0,0 +1,78 @@
...
+  } else if (is_mac) {
+    args += [
+      "--mac-min-version",
+      mac_deployment_target,
+    ]
+  }
+}
+
+group("boringtun") {
+  public_deps = [ ":build_boringtun" ]
```
bridiver, line 77: what's the reason for having a group target that only contains a single public dep?

browser/BUILD.gn
```diff
@@ -39,6 +39,14 @@ source_set("browser_process") {
     "//chrome/browser:browser_process",
     "//extensions/buildflags",
   ]
+
+  # Make sure boringtun keeps getting built even if we don't use
```
bridiver, line 43: This is going to waste a lot of time and resources to build something that isn't being used. You should add it to the `all` target instead.

third_party/boringtun/.cargo/config.toml
bridiver: building release is generally slower due to optimizations and loses things like symbols that we probably want in dev. Also regardless of what options we use, they should be maintained in gn, not in config.toml

browser/BUILD.gn
bsclifton: That's a good point - we do at least need to make sure it doesn't add dependency for Origin build

build/commands/scripts/updateBoringtunResources.ts
alekspop26: Both wasm script and this script have dependencies that can't get resolved when one runs the script standalone. For example, wasm's update script prints this:
```
Traceback (most recent call last):
  File "/Users/apopov/work/src/brave/third_party/wasm/update.py", line 17, in <module>
    import toml
ModuleNotFoundError: No module named 'toml'
```
My script prints:
```
Traceback (most recent call last):
  File "/Users/apopov/work/src/brave/third_party/boringtun/update.py", line 33, in <module>
    import brave_chromium_utils
ModuleNotFoundError: No module named 'brave_chromium_utils'
```
So apparently running it from npm ensures the right environment.

netzenbot (approved):

build/commands/scripts/updateBoringtunResources.ts
alekspop26: The npm wrapper has been removed.

alekspop26:
Replaced by multi-part changes:
1. https://github.com/brave/brave-core/pull/36102
2. https://github.com/brave/brave-core/pull/36142
3. https://github.com/brave/brave-core/pull/36192
4. https://github.com/brave/brave-core/pull/36229
5. https://github.com/brave/brave-core/pull/36261

PR #35742, opened 2026-04-21
Remove `sensitive` flag on `MetricsReportingEnabled` policy

### LINKED ISSUE #49357: `MetricsReportingEnabled` policy isn't hiding crash report checkbox on brave://welcome
### Description

Found while verifying https://github.com/brave/brave-browser/issues/48403

When testing the above issue the `BraveP3AEnabled` policy works as expected, but the `MetricsReportingEnabled` policy does not. The crash reporting checkbox is still displayed on brave://welcome with `MetricsReportingEnabled` policy set (I used a value of `false`). I verified on macOS and @GeetaSarvadnya confirmed she sees the same on Windows.

Interestingly, on brave://policy when `MetricsReportingEnabled` is configured the "status" shows `Error, Ignored` which doesn't seem correct.

### Steps to reproduce

1. Clean up from any other policy testing
2. Set the `MetricsReportingEnabled` policy 
 - on macOS I ran `sudo /usr/libexec/PlistBuddy -c "Add :MetricsReportingEnabled bool false" /Library/Managed\ Preferences/com.brave.Browser.beta.plist` and `sudo killall cfprefsd`
 - on Windows since the policy doesn't appear in the list as mentioned via the [Group Policy help page](https://support.brave.app/hc/en-us/articles/360039248271-Group-Policy), @GeetaSarvadnya set the policy via REGEDIT 
3. Launch Brave/Relaunch Brave
4. Navigate to brave://policy to ensure the policy is set (note the "Status" field)
5. Navigate to brave://welcome/ and go through onboarding until the final slide

Expected Result: Ensure P3A opt-in checkbox is displayed and the Crash reporting checkbox is hidden because `MetricsReportingEnabled` policy is set
Actual Result: With the above steps I'm seeing the crash reporting checkbox still displayed on brave://welcome/

Note for @brave/qa-team - when verifying this is fixed, this case from the [original test plan](https://github.com/brave/brave-core/pull/30635#issue-3320044685) also needs to be verified:
> Ensure that the opt-in step is completely skipped if both (`BraveP3AEnabled` and `MetricsReportingEnabled`) policies are set.

### Actual result

macOS:

https://github.com/user-attachments/assets/6f01c916-289d-4da7-82dd-706b44771d5b

Win:

Image

### Expected result

Crash reporting checkbox should be hidden from brave://welcome 

### Reproduces how often

Easily reproduced

Brave | 1.83.99 Chromium: 140.0.7339.133 (Official Build) beta (x86_64)
-- | --
Revision | a69d256e386747e16fe8979aab14fdc1446c83ea
OS | macOS Version 15.6.1 (Build 24G90)

### Channel information

- [ ] release (stable)
- [x] beta
- [ ] nightly

### Reproducibility

### Miscellaneous information

cc @DJAndries

DJAndries:
Resolves https://github.com/brave/brave-browser/issues/49357

Docs on the sensitive policy flag: https://source.chromium.org/chromium/chromium/src/+/main:components/policy/resources/new_policy_templates/policy.yaml;l=110-114
```
# Optional field that is `false` by default.
# When set to true, this policy is considered only if the user is part of a an
# an Active Directory domain on Windows, managed on the Mac, or enrolled in
# Chrome Enterprise Core.
```

Removing this restriction so users can set the policy value without needing a fully managed machine, like we already allow for our usage ping and P3A policies.

### DIFF
--- a/patches/components-policy-resources-templates-policy_definitions-Miscellaneous-MetricsReportingEnabled.yaml.patch
+++ b/patches/components-policy-resources-templates-policy_definitions-Miscellaneous-MetricsReportingEnabled.yaml.patch
@@ -1,0 +1,12 @@
+diff --git a/components/policy/resources/templates/policy_definitions/Miscellaneous/MetricsReportingEnabled.yaml b/components/policy/resources/templates/policy_definitions/Miscellaneous/MetricsReportingEnabled.yaml
+index 31df634c993eda0a026459655dc7c2a4605ce694..2e40c7c2546f458a2f42b022bc5d5d2060d22427 100644
+--- a/components/policy/resources/templates/policy_definitions/Miscellaneous/MetricsReportingEnabled.yaml
++++ b/components/policy/resources/templates/policy_definitions/Miscellaneous/MetricsReportingEnabled.yaml
+@@ -27,7 +27,6 @@ owners:
+ - zmin@chromium.org
+ schema:
+   type: boolean
+-sensitive: true
+ supported_on:
+ - chrome.*:8-
+ - ios:88-


mkarolin (approved):
++

netzenbot (approved):

PR #35746, opened 2026-04-21
[AI Chat]: Make Code Block Header Sticky

### LINKED ISSUE #53704: [AI Chat] Make the code block header with `Copy` button sticky when scrolling
### Description

When a code block is generated, the `Copy` button remains on top and need to scroll up to copy it

### Steps to reproduce

1. Ask Leo to write a code block
2. Ensure the code block is big so that it scrolls out of view
3. To copy the code, you need to scroll back up and then select

### Actual result

Code block can't be copied when you have scrolled down to the bottom

### Expected result

Code block hearder with `Copy` button should be sticky so that you don't need to scroll up and just copy as long as the code block is visible

### Reproduces how often

Easily reproduced

All

### Channel information

- [x] release (stable)
- [x] beta
- [x] nightly

### Reproducibility

### Miscellaneous information

cc: @yrliou @mattmcalister @petemill

Douglashdaniel:
## Description 

Makes Code Block `Headers` sticky when scrolling through the chat.

Resolves <https://github.com/brave/brave-browser/issues/53704>

https://github.com/user-attachments/assets/9677fbbd-6509-4770-9865-4c117daa6df6

### DIFF
--- a/components/ai_chat/resources/untrusted_conversation_frame/components/code_block/style.module.scss
+++ b/components/ai_chat/resources/untrusted_conversation_frame/components/code_block/style.module.scss
@@ -3,7 +3,8 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this file,
 // you can obtain one at https://mozilla.org/MPL/2.0/.
 
-.container {
+.inline,
+.blockRoot {
   overflow: auto;
   background: var(--leo-color-page-background);
   border: 1px solid var(--leo-color-divider-subtle);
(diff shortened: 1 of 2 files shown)


components/ai_chat/resources/untrusted_conversation_frame/components/code_block/style.module.scss
```diff
@@ -26,8 +26,43 @@
   font-size: inherit;
 }
 
+// No overflow on this shell so the toolbar can use `position: sticky` against
+// the conversation scroller.
+.blockRoot {
```
fallaciousreasoning, line 30: could we share most of the styles still with 

```css
.inline, .blockRoot {
// shared
}

.blockRoot {
// customizations
}
```

Douglashdaniel: Good catch, updated!

fallaciousreasoning (approved):

PR #35750, opened 2026-04-22
[plaster] `DownloadToolbarUIController` migration

### LINKED ISSUE #54814: Provide plaster for `DownloadToolbarUIController`
This class has a now unnecessary use of inheritance that is causing problems in M149 due to https://crrev.com/c/7766617.

cdesouza-chromium:
This class has a few customisations that were awkward with macro
replacements and therefore the derived function. With plaster though, we
can just add a `ChromiumImpl` variant of a function, and then only
replace the declartion in the translation unit, which makes a lot of the
complexity present go away.

This change drops the derived class for `DownloadToolbarUIController`,
and just adds a call to our custom function at the end of the relevant
function updating the icon appearance.

This particular class was running into problems in M149
(https://crrev.com/c/7766617).

Resolves https://github.com/brave/brave-browser/issues/54814


chromium_src/chrome/browser/ui/views/download/bubble/download_toolbar_ui_controller.cc
bridiver, line 61: I feel like this could just be an anonymous namespace function like GetIconColor that takes `DownloadBubbleUpdateService` as a parameter

bridiver: actually the same goes for `UpdateIcon` passing in browser_view_ and action_item_ so we don't have to touch the header at all and we can just call these methods from the original just like GetIconColor

cdesouza-chromium: Fixed. (I do think though that dropping the use of `UpdateIcon_ChromiumImpl` makes sense slightly harder to understand, to barely any gain, as I don't see the patch on the header here as a draw back, as both would always break together.)

netzenbot (approved):

patches/chrome-browser-ui-views-download-bubble-download_toolbar_ui_controller.cc.patch
```diff
@@ -0,0 +1,21 @@
...
+ 
++  icon_color = GetIconColor(state_, active_, browser_view_->GetColorProvider());
+   action_item_->SetImage(ui::ImageModel::FromVectorIcon(*new_icon, icon_color));
+ 
+   // Update the toolbar button's tooltip.
+@@ -723,6 +724,8 @@ void DownloadToolbarUIController::UpdateIcon() {
+     return;
+   }
+   progress_ring->SetDownloading(progress_info_.progress_percentage);
++
```
bridiver: we should still avoid empty lines even plaster I think

bridiver (approved):

patches/chrome-browser-ui-views-download-bubble-download_toolbar_ui_controller.cc.patch
cdesouza-chromium: Done.

cdesouza-chromium:
Force merging, as CI is failing on something unrelated:

```
AdBlockDATCacheCorruptBrowserTest.FallsBackToFilterList
```

PR #35752, opened 2026-04-22
[AI Chat]: Fix scroll height issues with notices

### LINKED ISSUE #54813: [AI Chat] Fresh profile first conversation entry is scrolled past the top of the viewport
The response is allocating too much space above the "Conversation history" notice. We should respond to these notices and when they're dismissed when reserving space.

Probably a regression with https://github.com/brave/brave-core/pull/35447 which is only v.1.91.36+

## Steps to reproduce

This requires a fresh profile because it's caused by the notice appearing. Other notices may cause the same issue.

0. Fresh profile (global browser or in-browser profile)
1. Open sidebar
2. Submit any prompt, press enter

Expected:
Submitted prompt is shown at top of viewport and answer starts to appear below it

Actual:
Submitted prompt and answer are hidden - they exist above the top of the viewport

Image

fallaciousreasoning:
Resolves https://github.com/brave/brave-browser/issues/54813

### DIFF
--- a/components/ai_chat/resources/untrusted_conversation_frame/components/conversation/index.tsx
+++ b/components/ai_chat/resources/untrusted_conversation_frame/components/conversation/index.tsx
@@ -161,10 +161,12 @@ function Conversation(props: ConversationProps) {
       return
     }
     const observer = new ResizeObserver(() => {
-      document.body.style.setProperty(
-        '--suggested-questions-height',
-        `calc(${el.offsetHeight}px + var(--leo-spacing-2xl) * 2)`,
-      )
+      requestAnimationFrame(() => {
+        document.body.style.setProperty(
+          '--suggested-questions-height',
+          `calc(${el.getBoundingClientRect().height}px + var(--leo-spacing-2xl) * 2)`,
+        )
+      })
     })
     observer.observe(el)
     return () => {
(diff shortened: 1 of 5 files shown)


netzenbot (approved):

components/ai_chat/resources/untrusted_conversation_frame/components/conversation_entries/style.module.scss
```diff
@@ -19,7 +19,7 @@
 }
 
 .hasGenerated .entryPair:last-child {
-  min-height: calc(100vh - var(--suggested-questions-height) * 2);
```
fallaciousreasoning, line 22: I'm surprised this was working at all - I had the padding * 2 being subtracted here originally

Douglashdaniel (approved):

components/ai_chat/resources/untrusted_conversation_frame/components/conversation/index.tsx
```diff
@@ -145,31 +145,41 @@ function Conversation(props: ConversationProps) {
   const { scrollToBottomContinuously, hasScrollableContent } =
     useScrollToBottom(scrollElementRef, contentRef)
 
+  const scrollToBottom = React.useCallback(
+    () => scrollToBottomContinuously(),
+    [scrollToBottomContinuously],
+  )
+
```
petemill, line 152: nit: shouldn't `useScrollToBottom` return a `useCallback`'ed `scrollToBottomContinuously` instead of having to wrap it here?

petemill (approved):
Tested working well!

petemill:
failing macos browser test is unrelated

components/ai_chat/resources/untrusted_conversation_frame/components/conversation/index.tsx
fallaciousreasoning: oooooh that seems like it would be much more sensible.

fallaciousreasoning: Ah actually, the reason it needs a separate useCallback is 'cause the function takes an argument, and I don't want the `HTMLElement` to be passed into the boolean param.

PR #35753, opened 2026-04-22
Try fixing crash when closing a tab with accent icon.

### LINKED ISSUE #54773: [Crash] Browser crashes when closing split container tab via keyboard shortcut
IMPORTANT: Your crash has already been automatically reported to our crash system. Please file this bug only if you can provide more information about it.

Brave Version: 1.91.78 Chromium: 147.0.7727.102
Operating System: Windows NT 10.0.26200

URL (if applicable) where crash occurred:

Can you reproduce this crash?
Yes

What steps will reproduce this crash? (If it's not reproducible, what were you doing just before the crash?)
1. Open link in a container
2. Open a second link in split view from the container 
3. Focus the tab that is opened in spilt view (not the original tab)
4. Press Ctrl+W to close the tab
5. Browser crashes

****DO NOT CHANGE BELOW THIS LINE****
Crash ID: crash/7fd10800-0ddf-950f-0000-000000000000

cc: @brave/legacy_qa @rebron

sangwoo108:
It could be a timing issue where small accent icon view isn't hidden yet after a tab is being closed. When the web contents is not available anymore, the container can be invalid, while the tab accent icon isn't hidden yet.

Resolves https://github.com/brave/brave-browser/issues/54773

### DIFF
--- a/browser/ui/views/tabs/brave_tab.cc
+++ b/browser/ui/views/tabs/brave_tab.cc
@@ -56,6 +56,11 @@ BraveTab::SmallAccentIconView::~SmallAccentIconView() = default;
 
 void BraveTab::SmallAccentIconView::OnPaint(gfx::Canvas* canvas) {
   auto* tab = static_cast<BraveTab*>(parent());
+  if (!tab->ShouldPaintTabAccent()) {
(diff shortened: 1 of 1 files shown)


simonhong (approved):
++

PR #35754, opened 2026-04-22
Update package_and_channel assignment logic

bondeabhijeet:
Fixed Linux installer branding metadata so package_and_channel now only omits the channel suffix for stable builds, and correctly includes -<channel> for non-stable builds (brave-origin-nightly). This fixes the desktop integration names with channel-specific installs.

Resolves
https://github.com/brave/brave-browser/issues/54734


bondeabhijeet:
@bbondy this PR would solve https://github.com/brave/brave-browser/issues/54734

bbondy:
Why is the fix generic and not only specific to Brave Origin?

bbondy:
We're going to go with this fix instead: https://github.com/brave/brave-core/pull/36145
If you could please re-test it after a build is made that contains it. Usually you should see a comment that says which version it is released in.

PR #35778, opened 2026-04-22
Compact mode for Desktop

### LINKED ISSUE #54722: Please add compact Browser layout like Helium
### Platforms

all

### Description

Image

i am opening this issue on behalf of [User from reddit](https://www.reddit.com/r/brave_browser/comments/1sp3zuz/feature_request_please_add_compact_browser_layout/)  .

it will also be great to have a "one-liner" mode just like in helium for brave

Image

aguscruiz:
Resolves https://github.com/brave/brave-browser/issues/54722

## Screenshots

image

image

## Summary

Introduces a single source of truth for **compact horizontal tabs** metrics in `chromium_src/chrome/browser/ui/tabs/brave_compact_horizontal_tabs_layout.h` and wires those constants through Brave layout helpers (`chromium_src/chrome/browser/ui/layout_constants.cc` / `.h`).

Compact metrics apply only when **`#brave-compact-horizontal-tabs`** is enabled **and** touch UI is off, via `tabs::ShouldUseCompactHorizontalTabsForNonTouchUI()` (also exposed in the public header for consumers outside `layout_constants.cc`).

Goal: a tighter tab strip + toolbar row on the order of **~60 DIP** while keeping touch targets at upstream defaults when touch UI is active.

### Tab/toolbar overlap vs control placement

Upstream uses `LayoutConstant::kTabstripToolbarOverlap` for both **tab/toolbar geometry** (tab height, strip height, frame top-area math, etc.) and, in some translation units, **vertical placement of tab-strip controls** (new-tab / combo / tab-search buttons) and **caption buttons**. Those roles want **different** numbers in compact mode.

This PR keeps the **large** overlap (**8 DIP**, `kTabstripToolbarOverlap`) centralized in `GetBraveLayoutConstant(LayoutConstant::kTabstripToolbarOverlap)` for geometry everywhere.

For **control alignment only**, it exposes `tabs::GetHorizontalTabControlsDelta()`, backed by `kTabStripControlsHeightDelta` (**−5** DIP compact, **−4** default). Consumers that need the smaller value use TU-local **`GetLayoutConstant` wrappers** in `chromium_src` that redirect only `kTabstripToolbarOverlap` to `GetHorizontalTabControlsDelta()` while including upstream `.cc` files—no upstream patch required:

- `chromium_src/chrome/browser/ui/views/frame/horizontal_tab_strip_region_view.cc` (tab strip control buttons / `UpdateButtonBorders()` path)
- `chromium_src/chrome/browser/ui/views/frame/browser_frame_view_win.cc` (Windows caption button vertical centering)

Elsewhere, deltas are applied explicitly (Mac traffic lights, Linux caption buttons, Brave NTB layout, opaque frame top area, etc.).

## Platform / UI notes

- **macOS**
  - `BraveBrowserFrameViewMac::GetCaptionButtonBounds()`: shifts traffic lights by `GetHorizontalTabControlsDelta()`; in compact mode **zeros** top/bottom margins on the caption bounds so the tab strip can align with the lights (see comments referencing Helium / caption-button bounds).
  - `BraveBrowserNativeWidgetMac::GetWindowFrameTitlebarHeight()`: in compact mode shrinks the titlebar height AppKit sees by **8 DIP** so centered traffic lights move up with the shorter row.
  - `BraveBrowserViewTabbedLayoutImpl::GetHorizontalTabStripLeadingMargin()`: when compact is active, tucks the first tab closer to leading caption buttons (`TabStyle` bottom corner radius / 2), preserving upstream behavior for split-view `kInternalPaddingKey` paths.
- **Windows**
  - `BraveBrowserFrameViewWin::LayoutCaptionButtons()`: applies `GetHorizontalTabControlsDelta()` to caption button Y (pairs with the `browser_frame_view_win.cc` chromium_src wrapper).
- **Linux**
  - `browser_frame_view_layout_linux.cc`: vertical adjustment for window caption buttons when overlapping toolbar (vertical-tabs path), using `GetHorizontalTabControlsDelta()` where needed.
- **Tab strip region (Brave)**
  - `BraveHorizontalTabStripRegionView::Layout`: NTB vertical position uses `GetHorizontalTabControlsDelta()`; horizontal placement after scroll-next or after last tab as appropriate.
- **Chrome UI polish**
  - `BraveOpaqueBrowserFrameView::GetTopAreaHeight()`: tabbed windows factor in `GetHorizontalTabControlsDelta()` for consistent top chrome height.
  - Split-view tile corner radius in `brave_tab_container.cc` uses **`kHorizontalSplitViewTileCornerRadiusDip` (9)** in compact horizontal mode vs **12** DIP default.
  - `tab_style.cc`: compact drag-handle extension (**2** vs **4** DIP default).
  - `tab_group_style.cc`: title chip vertical offset uses **`GetHorizontalTabVerticalSpacing()`** instead of the raw constant so compact spacing is respected.

## Constants header highlights

Key compact (non-touch) values documented in `brave_compact_horizontal_tabs_layout.h`, including tab visual height (**26**), vertical spacing (**2**), tabstrip–toolbar overlap (**8**), toolbar/location bar dimensions, omnibox padding, tab/group insets, and defaults when the flag is off or touch UI is on.

## Tests

`browser/ui/views/tabs/brave_compact_horizontal_tabs_layout_unittest.cc` verifies compact vs default layout constants and `GetHorizontalTabControlsDelta()` behavior across:

- `SelectsCompactLayoutConstants`
- `FlagDisabledReturnsDefaultLayoutConstants`
- `TouchUiReturnsDefaultLayoutConstants`

Shared expectations live in `ExpectDefaultLayoutConstants()`.

## Related upstream / merge

Branch includes merge with current `master` (e.g. Chromium uplift work **148** as of the merge commit on this branch).


chromium_src/chrome/browser/ui/layout_constants.cc
```diff
@@ -134,24 +172,47 @@ bool UseCompact() {
```
netzenbot, line 170: This PR modifies user-visible layout (URL bar height, toolbar interior margin, tab strip padding, drag handle extension) when `kBraveCompactHorizontalTabs` is enabled, but no CUJ or browser test is added. Per [ARCH-062](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-062), every feature that modifies user-visible UI needs at least one `InteractiveBrowserTest`. Consider adding a test that enables the flag and verifies at least one of the changed layout constants (e.g. `kLocationBarHeight`, `kTabStripPadding`) returns the compact value.

browser/ui/tabs/brave_compact_horizontal_tabs_layout.h
```diff
@@ -0,0 +1,83 @@
...
+inline constexpr int kTabstripToolbarOverlap = 8;
+
+// `LayoutConstant::kTabStripHeight` (compact, non-touch). Keep in sync when
+// editing the three values above (no separate magic number).
+inline constexpr int kTabStripLayoutHeightTarget =
+    kTabVisualHeight + 2 * kTabVerticalSpacing + kTabstripToolbarOverlap;
+
+// Horizontal padding inside the tab label area (see GetHorizontalTabPadding()).
+inline constexpr int kTabHorizontalPadding = 4;
+
```
netzenbot, line 55: nit: Avoid positional terms in comments — "three values above" is fragile if code is inserted between those constants and `kTabStripLayoutHeightTarget`. List them by name instead, e.g. `// Keep in sync with kTabVisualHeight, kTabVerticalSpacing, and kTabstripToolbarOverlap.` ([CS-024](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-024))

aguscruiz:
@sangwoo108 I have a problem with this implemenation that maybe you can help me out with: 

In `chromium_src/chrome/browser/ui/tabs/brave_compact_horizontal_tabs_layout.h` i created a set of overrides for the different sizing of things, right? 

The way it's setup now, this line
`inline constexpr int kTabstripToolbarOverlap = 8;` achieves the goal of getting the tab strip to be the required height. 

However, if I do that, the traffic lights (in macos) and the new tab button get missaligned as you can see here:
image

If I change the value to 
`inline constexpr int kTabstripToolbarOverlap = 4;` they become centered again, but now the tab bar gains 4 extra pixels of empty space beneath the tabs.

I can't figure out how else to get rid of those 4 extra pixels without this override. Maybe you can take a look and have a better idea of what the solution could be? 

image


Here's figma for reference
https://www.figma.com/design/T1vmHIPLFZ3dS1EZWOBDvN/%F0%9F%92%BB-General-desktop-design?node-id=4006-2655

chromium_src/chrome/browser/ui/layout_constants.cc
```diff
@@ -5,29 +5,57 @@
...
 #include "ui/gfx/geometry/insets.h"
 
+static_assert(tabs::kHorizontalTabVerticalSpacing ==
+              tabs::compact_horizontal_tabs_layout::kTabVerticalSpacingDefault);
+
 namespace {
 
 using tabs::HorizontalTabsUpdateEnabled;
 
+bool BraveCompactHorizontalTabsMetricsActive() {
```
sangwoo108: I think ShouldUseCompactHorizontalTabsForNonTouchUI() would be a better name

browser/ui/views/tabs/brave_compact_horizontal_tabs_interactive_ui_test.cc
sangwoo108, line 1: I think this test can be simple unittest instead of interactive ui test

chromium_src/chrome/browser/ui/layout_constants.cc
sangwoo108: Can you move this up to near BraveCompactHorizontalTabsMetricsActive() so that we can see both together?

chromium_src/chrome/browser/ui/layout_constants.cc
```diff
@@ -134,24 +172,47 @@ bool UseCompact() {
...
+    return tabs::compact_horizontal_tabs_layout::kTabVerticalSpacingTouch;
+  }
+  return UseCompact()
+             ? tabs::compact_horizontal_tabs_layout::kTabVerticalSpacing
+             : tabs::compact_horizontal_tabs_layout::kTabVerticalSpacingDefault;
+}
+
 int GetHorizontalTabHeight() {
-  return UseCompact() ? 28 : 32;
+  if (ui::TouchUiController::Get()->touch_ui()) {
```
sangwoo108: I think checking `ui::TouchUiController::Get()->touch_ui()` here should be replaced with ShouldUseCompactHorizontalTabsForNonTouchUI().

chromium_src/chrome/browser/ui/layout_constants.cc
aguscruiz: Addressed by replacing the interactive UI test with `browser/ui/brave_compact_horizontal_tabs_layout_unittest.cc`, which enables `kBraveCompactHorizontalTabs`/`kBraveHorizontalTabsUpdate` and verifies the compact values returned for `kLocationBarHeight`, `kTabStripPadding`, and `kTabstripToolbarOverlap`. I also reran presubmit after the change.

browser/ui/tabs/brave_compact_horizontal_tabs_layout.h
aguscruiz: Updated the comment to name the constants explicitly instead of referring to positional context. It now says to keep `kTabStripLayoutHeightTarget` in sync with `kTabVisualHeight`, `kTabVerticalSpacing`, and `kTabstripToolbarOverlap`.

chromium_src/chrome/browser/ui/layout_constants.cc
aguscruiz: Done. I moved the related helper so it sits next to `ShouldUseCompactHorizontalTabsForNonTouchUI()` in `layout_constants.cc` for readability.

chromium_src/chrome/browser/ui/layout_constants.cc
aguscruiz: Renamed to `ShouldUseCompactHorizontalTabsForNonTouchUI()` in the anonymous namespace at the top of `chromium_src/chrome/browser/ui/layout_constants.cc`, and updated all five call sites (insets + `kTabstripToolbarOverlap`, `kLocationBarHeight`, and the location-bar padding family, plus the `tabs::GetHorizontalTab*` getters). Reflected in commit e95a754.

browser/ui/views/tabs/brave_compact_horizontal_tabs_interactive_ui_test.cc
aguscruiz: Replaced with a unit test at `browser/ui/views/tabs/brave_compact_horizontal_tabs_layout_unittest.cc` (`BraveCompactHorizontalTabsLayoutTest.SelectsCompactLayoutConstants`). It uses `ui::TouchUiController::TouchUiScoperForTesting` to force non-touch mode and asserts the compact values of `kLocationBarHeight`, `kTabStripPadding`, and `kTabstripToolbarOverlap`. The `interactive_ui_tests` source_set was removed from `browser/ui/views/tabs/BUILD.gn` (and its dep in `test/BUILD.gn`), and the new unittest is now part of the existing `unit_tests` target. Reflected in commit e95a754.

chromium_src/chrome/browser/ui/layout_constants.cc
aguscruiz: Done — the old duplicate `UseCompact()` inside `namespace tabs` was removed, and `UseCompactHorizontalTabs()` now sits right next to `ShouldUseCompactHorizontalTabsForNonTouchUI()` in the anonymous namespace at the top of the file, so both helpers are visible together. The `tabs::GetHorizontalTab*` getters below use those file-local helpers directly. Reflected in commit e95a754.

chromium_src/chrome/browser/ui/layout_constants.cc
aguscruiz: Done. All raw `ui::TouchUiController::Get()->touch_ui()` checks in the tab-strip getters (`GetHorizontalTabVerticalSpacing`, `GetHorizontalTabHeight`, `GetHorizontalTabPadding`, `GetTabGroupTitleHorizontalInset`) now gate on `ShouldUseCompactHorizontalTabsForNonTouchUI()`, matching the PR intent that compact metrics only apply when the flag is on and touch UI is off. This also let me drop the unused `kTabVisualHeightTouchCompact` and `kTabVerticalSpacingTouch` constants from `brave_compact_horizontal_tabs_layout.h`. Reflected in commit e95a754.

browser/ui/views/tabs/brave_compact_horizontal_tabs_layout_unittest.cc
```diff
@@ -0,0 +1,44 @@
...
+
+  EXPECT_EQ(compact_horizontal_tabs_layout::kLocationBarHeight,
+            GetLayoutConstant(LayoutConstant::kLocationBarHeight));
+  EXPECT_EQ(compact_horizontal_tabs_layout::kTabVerticalSpacing,
+            GetLayoutConstant(LayoutConstant::kTabStripPadding));
+  EXPECT_EQ(compact_horizontal_tabs_layout::kTabstripToolbarOverlap,
+            GetLayoutConstant(LayoutConstant::kTabstripToolbarOverlap));
+}
+
+}  // namespace tabs
```
netzenbot, line 90: The test only covers the compact-active path. Per [TI-026](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-026), both sides of the guard condition should be tested — add a test (or parameterized case) that disables `kBraveCompactHorizontalTabs` (or enables touch UI) and verifies the default values like `kLocationBarHeightDefault` (32) and `kTabVerticalSpacingDefault` (4) are returned.

chromium_src/chrome/browser/ui/layout_constants.cc
```diff
@@ -126,32 +168,39 @@ int GetLayoutConstant(LayoutConstant constant) {
```
netzenbot, line 157: nit: `GetHorizontalTabVerticalSpacing()` is defined here before `GetHorizontalTabHeight()`, but in the header it is declared after `GetHorizontalTabHeight()`. Per [CS-017](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-017), function definitions in `.cc` files should appear in the same order as their declarations in the `.h`. Move `GetHorizontalTabVerticalSpacing()` after `GetHorizontalTabHeight()` to match.

chromium_src/chrome/browser/ui/layout_constants.h
```diff
@@ -24,15 +24,21 @@ namespace tabs {
 
 // The visual height of tabs in horizontal tabs mode. Note that the height of
 // the view may be greater than the visual height of the tab shape. See also
-// `kHorizontalTabVerticalSpacing`.
+// `GetHorizontalTabVerticalSpacing()`.
 int GetHorizontalTabHeight();
 
+// Spacing between the top and bottom of tabs and the tab strip bounds. Use this
+// instead of `kHorizontalTabVerticalSpacing` for layout; it reflects compact
+// mode and touch UI.
```
netzenbot: nit: The comment for this new public method says it "reflects compact mode and touch UI" but doesn't say what values it returns under each condition. Per [DOC-004](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-004), public method docs in headers should describe the contract including return values — e.g., note that it returns `compact_horizontal_tabs_layout::kTabVerticalSpacing` when the compact flag is on and touch UI is off, and `kHorizontalTabVerticalSpacing` otherwise.

browser/ui/views/tabs/brave_compact_horizontal_tabs_layout_unittest.cc
aguscruiz: Addressed. Split the test into three `TEST_F` cases covering both sides of the guard:

- `SelectsCompactLayoutConstants` — flag on + touch UI off → compact metrics.
- `FlagDisabledReturnsDefaultLayoutConstants` — flag off → defaults (`kLocationBarHeightDefault` = 32, `kTabVerticalSpacingDefault` = 4, `kTabstripToolbarOverlapDefault`).
- `TouchUiReturnsDefaultLayoutConstants` — flag on + touch UI on → defaults (the `!touch_ui()` branch).

The `ScopedFeatureList` and `TouchUiScoperForTesting` are now per-test so each case controls its own state.

chromium_src/chrome/browser/ui/layout_constants.cc
aguscruiz: Addressed — swapped the order of `GetHorizontalTabVerticalSpacing()` and `GetHorizontalTabHeight()` in `layout_constants.cc` so it matches the declaration order in `layout_constants.h` (`GetHorizontalTabHeight()` first, then `GetHorizontalTabVerticalSpacing()`).

chromium_src/chrome/browser/ui/layout_constants.h
aguscruiz: Addressed — expanded the doc comment for `GetHorizontalTabVerticalSpacing()` in `layout_constants.h` to describe the return contract under each condition:

> Returns `compact_horizontal_tabs_layout::kTabVerticalSpacing` (2) when the `#brave-compact-horizontal-tabs` flag is enabled and touch UI is off. Returns `compact_horizontal_tabs_layout::kTabVerticalSpacingDefault` (which equals `kHorizontalTabVerticalSpacing`, 4) otherwise (flag disabled or touch UI active).

sangwoo108:
> @sangwoo108 I have a problem with this implemenation that maybe you can help me out with:
> 
> In `chromium_src/chrome/browser/ui/tabs/brave_compact_horizontal_tabs_layout.h` i created a set of overrides for the different sizing of things, right?
> 
> The way it's setup now, this line `inline constexpr int kTabstripToolbarOverlap = 8;` achieves the goal of getting the tab strip to be the required height.
> 
> However, if I do that, the traffic lights (in macos) and the new tab button get missaligned as you can see here: image
> 
> If I change the value to `inline constexpr int kTabstripToolbarOverlap = 4;` they become centered again, but now the tab bar gains 4 extra pixels of empty space beneath the tabs.
> 
> I can't figure out how else to get rid of those 4 extra pixels without this override. Maybe you can take a look and have a better idea of what the solution could be?
> 
> image
> Here's figma for reference https://www.figma.com/design/T1vmHIPLFZ3dS1EZWOBDvN/%F0%9F%92%BB-General-desktop-design?node-id=4006-2655

Let's take another round for this one. It needs mac specific changes.

chromium_src/chrome/browser/ui/layout_constants.cc
```diff
@@ -5,29 +5,61 @@
...
 #include <optional>
 
+#include "base/feature_list.h"
+#include "chrome/browser/ui/tabs/brave_compact_horizontal_tabs_layout.h"
 #include "chrome/browser/ui/tabs/features.h"
 #include "ui/base/pointer/touch_ui_controller.h"
 #include "ui/gfx/geometry/insets.h"
 
+static_assert(tabs::kHorizontalTabVerticalSpacing ==
+              tabs::compact_horizontal_tabs_layout::kTabVerticalSpacingDefault);
```
sangwoo108: If we want these two to be same, I think we could assign the value in chromium_src/chrome/browser/ui/tabs/brave_compact_horizontal_tabs_layout.h with the upstream value.

chromium_src/chrome/browser/ui/views/frame/horizontal_tab_strip_region_view.cc
```diff
@@ -16,11 +17,24 @@
...
+int GetLayoutConstantForBraveTabStripControls(LayoutConstant constant) {
+  if (constant == LayoutConstant::kTabstripToolbarOverlap) {
+    return tabs::GetHorizontalTabControlOverlap();
+  }
+  return GetLayoutConstant(constant);
+}
+
+}  // namespace
+
+#define GetLayoutConstant GetLayoutConstantForBraveTabStripControls
```
sangwoo108: Can we remove these duplicated override and deal with kTabStripToolbarOverlap from std::optional<int> GetBraveLayoutConstant(LayoutConstant constant) in chromium_src/.../layout_constants.cc ?

browser/ui/views/frame/brave_opaque_browser_frame_view.cc
```diff
@@ -142,7 +143,13 @@ int BraveOpaqueBrowserFrameView::GetTopAreaHeight() const {
     return layout_->NonClientTopHeight(false);
   }
 
-  return OpaqueBrowserFrameView::GetTopAreaHeight();
+  const bool is_tabbed = GetBrowserView()->GetIsNormalType();
+  const auto info = GetClientFrameElementInfo();
+  const auto frame_height = FrameBorderInsets(false).top();
```
netzenbot: nit: `FrameBorderInsets(false).top()` and `layout_->NonClientTopHeight(false)` both return `int` — a short, descriptive type. Per [CS-053](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-053), prefer spelling out `int` instead of `auto` here so the type is immediately visible at the declaration site.

browser/ui/views/tabs/brave_compact_horizontal_tabs_layout_unittest.cc
```diff
@@ -0,0 +1,92 @@
...
+// Compact flag on + touch UI off: `GetLayoutConstant` should return the
+// compact metrics from `brave_compact_horizontal_tabs_layout.h`.
+TEST_F(BraveCompactHorizontalTabsLayoutTest, SelectsCompactLayoutConstants) {
+  scoped_feature_list_.InitWithFeatures(
+      {tabs::kBraveCompactHorizontalTabs, tabs::kBraveHorizontalTabsUpdate},
+      {});
+  ui::TouchUiController::TouchUiScoperForTesting touch_ui_scoper(
+      /*touch_ui_enabled=*/false);
+
+  ASSERT_FALSE(ui::TouchUiController::Get()->touch_ui());
```
netzenbot, line 47: nit: `FlagDisabledReturnsDefaultLayoutConstants` and `TouchUiReturnsDefaultLayoutConstants` repeat the same four `EXPECT_EQ` assertions. Per [TI-024](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-024), extract them into a helper like `ExpectDefaultLayoutConstants()` to avoid the duplication.

chromium_src/chrome/browser/ui/layout_constants.cc
aguscruiz: Addressed. I updated `chromium_src/chrome/browser/ui/tabs/brave_compact_horizontal_tabs_layout.h` to include `chrome/browser/ui/layout_constants.h` and set `kTabVerticalSpacingDefault = kHorizontalTabVerticalSpacing` so it stays tied to the upstream value. Since this is now directly aliased, I also removed the redundant `static_assert` from `layout_constants.cc`.

chromium_src/chrome/browser/ui/views/frame/horizontal_tab_strip_region_view.cc
aguscruiz: Addressed. I removed the duplicated `GetLayoutConstant` override shim from `chromium_src/chrome/browser/ui/views/frame/horizontal_tab_strip_region_view.cc`. `kTabstripToolbarOverlap` now remains handled centrally in `GetBraveLayoutConstant(LayoutConstant)` in `layout_constants.cc`. For the control-placement path in `UpdateButtonBorders()`, I updated the rewrite/patch to use `tabs::GetHorizontalTabControlOverlap()` directly, which preserves the intended control-specific behavior without redefining `GetLayoutConstant` in that translation unit.

browser/ui/views/tabs/brave_compact_horizontal_tabs_layout_unittest.cc
aguscruiz: Addressed. I extracted the repeated default-branch assertions into a local helper `ExpectDefaultLayoutConstants()` in `browser/ui/views/tabs/brave_compact_horizontal_tabs_layout_unittest.cc`, then reused it from both `FlagDisabledReturnsDefaultLayoutConstants` and `TouchUiReturnsDefaultLayoutConstants`.

browser/ui/views/frame/brave_opaque_browser_frame_view.cc
aguscruiz: Addressed. In `browser/ui/views/frame/brave_opaque_browser_frame_view.cc` I replaced `const auto` with explicit `const int` for both `frame_height` (`FrameBorderInsets(false).top()`) and `top_height` (`layout_->NonClientTopHeight(false)`) per CS-053.

browser/ui/views/frame/brave_browser_frame_view_win.cc
```diff
@@ -155,4 +155,7 @@ void BraveBrowserFrameViewWin::LayoutCaptionButtons() {
             ? 0
             : width() - caption_button_container_->width());
   }
+
+  caption_button_container_->SetY(caption_button_container_->y() +
+                                  tabs::GetHorizontalTabControlOverlap());
```
sangwoo108: I'm curious why we don't adjust height instead. Becacause it might make caption button overlaps toolbar area on some platforms like Windows and Linux.

chromium_src/chrome/browser/ui/tabs/brave_compact_horizontal_tabs_layout.h
```diff
@@ -0,0 +1,83 @@
...
+// ---------------------------------------------------------------------------
+
+// --- Tab strip (non-touch, compact) ----------------------------------------
+inline constexpr int kTabVisualHeight = 26;
+inline constexpr int kTabVerticalSpacing = 2;
+inline constexpr int kTabstripToolbarOverlap = 8;
+// Control-specific overlap compensation used by navigation/caption controls.
+// Keep this separate from `kTabstripToolbarOverlap` so tab rendering overlap
+// can be tuned independently from control placement.
+inline constexpr int kTabstripToolbarControlsOverlap = -5;
```
sangwoo108: If this and kTabstripToolbarControlsOverlapDefault should be relative values, I don't think it's proper name. Maybe kTabStripHeightDelta or kTabStripInsetY could be a better name. GetHorizontalTabControlOverlap() also should be something like GetHorizontalTabControlsDelta() could be better.

chromium_src/chrome/browser/ui/tabs/tab_style.cc
```diff
@@ -58,7 +61,11 @@ class BraveTabStyle : public TabStyle {
     // The "drag handle extension" is the amount of space in DIP at the top of
     // inactive tabs where mouse clicks are treated as clicks in the "caption"
     // area, i.e. the draggable part of the window frame.
-    return 4;
+    if (base::FeatureList::IsEnabled(tabs::kBraveCompactHorizontalTabs) &&
+        !ui::TouchUiController::Get()->touch_ui()) {
```
sangwoo108: If we exposes ShouldUseCompactHorizontalTabsForNonTouchUI() in layout_constants.cc to the header, we can dedupe this line.

rewrite/chrome/browser/ui/views/frame/horizontal_tab_strip_region_view.cc.toml
```diff
@@ -8,3 +8,9 @@ description = "Replace TabStrip construction with BraveTabStrip"
 pattern = "std::make_unique<TabStrip>"
 replace = "std::make_unique<BraveTabStrip>"
 count = 1
+
+[[substitution]]
+description = "Tab strip button border math uses control overlap (see GetHorizontalTabControlOverlap) instead of tab/toolbar overlap from GetLayoutConstant"
+pattern = "GetLayoutConstant(LayoutConstant::kTabstripToolbarOverlap)"
+replace = "tabs::GetHorizontalTabControlOverlap()"
+count = 2
```
sangwoo108: I'm not sure if it's good approach.

I found that we're already kTabstripToolbarOverlap overriding in GetBraveLayoutConstnats(). So should we reconsider replacement with tabs::GetHorizontalTabControlOverlap() from callsites and the implementation? One reason I'm not sure is that we might have regression regarding the GetLayoutConstants(kTabstripToolbarOverlap)

chromium_src/chrome/browser/ui/tabs/brave_compact_horizontal_tabs_layout.h
aguscruiz: Renamed in 50e2... (incoming push):

- `kTabstripToolbarControlsOverlap` → `kTabStripControlsHeightDelta`
- `kTabstripToolbarControlsOverlapDefault` → `kTabStripControlsHeightDeltaDefault`
- `tabs::GetHorizontalTabControlOverlap()` → `tabs::GetHorizontalTabControlsDelta()`

Updated all callsites (header + .cc, `layout_constants.h`/`.cc`, the rewrite TOML, the regenerated patch, the unit test, plus the `brave_browser_frame_view_*`, `brave_opaque_browser_frame_view`, `brave_tab_strip_region_view`, and Linux/Win frame view layout chromium_src files). I went with `kTabStripControlsHeightDelta` rather than your suggested `kTabStripHeightDelta`/`kTabStripInsetY` because the value applies specifically to control-button vertical placement (`UpdateButtonBorders()`), not the tab strip height as a whole — happy to switch to your wording verbatim if you prefer.

chromium_src/chrome/browser/ui/tabs/tab_style.cc
aguscruiz: Done. Exposed `tabs::ShouldUseCompactHorizontalTabsForNonTouchUI()` in [`chromium_src/chrome/browser/ui/tabs/brave_compact_horizontal_tabs_layout.h`](https://github.com/brave/brave-core/blob/master/chromium_src/chrome/browser/ui/tabs/brave_compact_horizontal_tabs_layout.h):

- Moved the definition out of the anonymous namespace in `layout_constants.cc` into `namespace tabs { … }`, alongside the other `tabs::` getters.
- Inlined `UseCompactHorizontalTabs()` into it (it was the only caller).
- In-file callers (`GetBraveLayoutInsets`, `GetBraveLayoutConstant`, the tab getters) now go through the public symbol via a `using tabs::ShouldUseCompactHorizontalTabsForNonTouchUI;` in the anonymous namespace.
- `BraveTabStyle::GetDragHandleExtension` now calls the helper directly, and I dropped the now-unused `#include "base/feature_list.h"` and `#include "ui/base/pointer/touch_ui_controller.h"` (kept `features.h` — still used).

rewrite/chrome/browser/ui/views/frame/horizontal_tab_strip_region_view.cc.toml
aguscruiz: Good question — I audited every `kTabstripToolbarOverlap` consumer before keeping the substitution. Short version: the split is intentional, no regressions.

`LayoutConstant::kTabstripToolbarOverlap` plays two distinct roles in compact mode:

1. **Tab/toolbar geometry** — `kTabHeight`/`kTabStripHeight` math, tab rendering (`tab_style_views.cc`, `tab.cc`), tab group underline (`tab_group_*.cc`), `BrowserRootView::OnPaintBackground()`, frame views (`browser_frame_view*.cc`, `opaque_browser_frame_view.cc`). All of these want the geometry value (compact = 8, default = 1).
2. **Control button vertical placement** inside the tab strip (`UpdateButtonBorders()`: new tab button, combo button, tab search button, …). These need a different value (compact = -5, default = -4); using 8 here is exactly what produces the misalignment you saw in the original screenshot thread.

So the central `GetBraveLayoutConstant(kTabstripToolbarOverlap)` override stays in place for role (1) — every other consumer of `kTabstripToolbarOverlap` continues to flow through `GetLayoutConstant()` and picks up the override. The rewrite substitution intentionally targets only the two `UpdateButtonBorders()` callsites in this TU (hence `count = 2`) to satisfy role (2).

To make the rationale explicit (and prevent someone collapsing the two later), I added:

- A block comment in `brave_compact_horizontal_tabs_layout.h` above `kTabStripControlsHeightDelta` documenting the two-role split.
- An expanded description on the rewrite substitution calling out that it targets only `UpdateButtonBorders()` and that the central `kTabstripToolbarOverlap` override remains for tab/toolbar geometry.

Lmk if you'd still prefer a different shape (e.g. a dedicated `LayoutConstant` enum value for control placement instead of the rewrite substitution); I'm happy to refactor if you have a strong preference.

chromium_src/chrome/browser/ui/views/frame/browser_frame_view_win.cc
```diff
@@ -30,7 +31,20 @@
            ? WebUITabStripContainerView::SupportsTouchableTabStrip(browser) \
            : tabs::utils::SupportsBraveVerticalTabs(browser))
 
+namespace {
+
```
netzenbot, line 36: nit: The new `#define GetLayoutConstant GetLayoutConstantForBraveWindowControls` wrapper added here has no comment explaining why the `kTabstripToolbarOverlap` interception is necessary for Win caption buttons. Per [CSRC-017](https://github.com/brave/brave-core/tree/master/docs/best-practices/chromium-src-overrides.md#CSRC-017), chromium_src overrides should include a comment explaining the purpose — something like "Intercept `kTabstripToolbarOverlap` in caption button layout to use the control-placement delta instead of the tab/toolbar geometry value, which is too large for button centering in compact mode."

rewrite/chrome/browser/ui/views/frame/horizontal_tab_strip_region_view.cc.toml
```diff
@@ -8,3 +8,20 @@ description = "Replace TabStrip construction with BraveTabStrip"
...
+
+# `LayoutConstant::kTabstripToolbarOverlap` is used by upstream for two
+# distinct roles in compact mode: (a) tab/toolbar geometry (e.g. `kTabHeight`,
+# `kTabStripHeight`, tab/group rendering, frame view top-area math) and (b)
+# control-button placement inside the tab strip (`UpdateButtonBorders()`,
+# which positions the new tab button, combo button, tab search button, etc.).
+# Compact mode requires different values for those two roles, so we override
+# (a) centrally in `GetBraveLayoutConstant(kTabstripToolbarOverlap)` and split
+# off (b) here. The substitution targets only the two `UpdateButtonBorders()`
+# callsites in this translation unit (hence `count = 2`); all other consumers
```
netzenbot: nit: `pattern = "GetLayoutConstant(LayoutConstant::kTabstripToolbarOverlap)"` is a compound expression, not a simple symbol name. Per [PLSTR-002](https://github.com/brave/brave-core/tree/master/docs/best-practices/plaster.md#PLSTR-002), use `pattern_re` when the match includes parentheses or additional context — e.g. `pattern_re = 'GetLayoutConstant\(LayoutConstant::kTabstripToolbarOverlap\)'` — so the substitution stays robust if upstream reformats the argument spacing.

browser/ui/views/frame/layout/brave_browser_view_tabbed_layout_impl.cc
```diff
@@ -302,6 +304,35 @@ BraveBrowserViewTabbedLayoutImpl::GetTopSeparatorType() const {
...
+                             static_cast<float>(internal_padding->left()));
+  }
+
+  // Compact mode default: tuck the first tab pill closer to the trailing edge
+  // of the caption-button cluster so the tab strip and the traffic lights
+  // read as a single visually centred row. Halving the corner-radius mirrors
+  // imputnet/helium @ b6e5b77e (`fix-caption-button-bounds.patch`, the
+  // companion to the `GetCaptionButtonBounds()` margin zeroing in
+  // `BraveBrowserFrameViewMac::GetCaptionButtonBounds()`).
+  return TabStyle::Get()->GetBottomCornerRadius() / 2;
```
sangwoo108, line 343: I guess the reason why this isn't gaurded with the OS_MAC flag is that this is applied only when there's leading caption buttons by upstream logic? If I'm correct, please leave it to the comment and otherwise, it should be guarded.

rewrite/chrome/browser/ui/views/frame/horizontal_tab_strip_region_view.cc.toml
sangwoo108: In that case, Could we make (2) cases handle compensate it from other places? We consider the patching strategy as the last resort, so we should explore. Maybe we can make it happen by overriding UpdateButtonBorders from BraveTabStripRegionView.

aguscruiz: Refactored to drop the rewrite/patch entirely. Done in b8bc95f9f4e.

Inside `UpdateButtonBorders()` the body still references private members of `HorizontalTabStripRegionView` (`tab_strip_action_container_`, `combo_button_`, `new_tab_button_`, `unfocus_button_`, `tab_search_button_` plus the static `UpdateBorderInsetsIfNeeded`), so a clean virtual override from `BraveHorizontalTabStripRegionView` would have meant duplicating the entire upstream body and keeping it in sync with future changes — which is exactly the kind of drift we want to avoid.

Instead, I applied the same chromium_src `#define GetLayoutConstant <wrapper>` pattern that's already used in `chromium_src/chrome/browser/ui/views/frame/browser_frame_view_win.cc`. The new wrapper in `chromium_src/chrome/browser/ui/views/frame/horizontal_tab_strip_region_view.cc` redirects `kTabstripToolbarOverlap` calls in this TU to `tabs::GetHorizontalTabControlsDelta()` and lets every other `LayoutConstant` flow through the central `GetBraveLayoutConstant()` override. Net effect:

- `rewrite/.../horizontal_tab_strip_region_view.cc.toml` no longer carries the second substitution.
- `patches/chrome-browser-ui-views-frame-horizontal_tab_strip_region_view.cc.patch` was regenerated and is now just the `BraveTabStrip` swap (the `UpdateButtonBorders()` hunk is gone).
- Upstream `horizontal_tab_strip_region_view.cc` is no longer modified for `kTabstripToolbarOverlap`.

I also documented this in the comment on `kTabStripControlsHeightDelta` (`brave_compact_horizontal_tabs_layout.h`) and added the corresponding CSRC-017 comment on the existing `GetLayoutConstantForBraveWindowControls` wrapper.

Happy to revisit if you'd still rather pursue the virtual-override route (e.g. via a chromium_src header `#define UpdateButtonBorders virtual UpdateButtonBorders` plus a duplicated body) — just wanted to flag the duplication trade-off before going there.

browser/ui/views/frame/layout/brave_browser_view_tabbed_layout_impl.cc
aguscruiz: You're right — kept it as a comment (no `IS_MAC` guard).

Tracing the call chain: `leading_margin` flows into upstream's `BrowserViewLayoutImpl::GetBoundsWithExclusion`, which only consumes it via `params.leading_exclusion.ContentWithPaddingAndInsets(leading_margin, …)`. When `params.leading_exclusion` is empty (the typical Windows/Linux case with no leading caption buttons), `ContentWithPaddingAndInsets` returns `(0, 0)` and `leading_margin` has no observable effect on the resulting bounds. So the half-radius value only takes effect on platforms that surface leading caption buttons (Mac traffic lights, plus Linux DEs configured with left-side controls), which is exactly the case it's intended for.

Added that rationale to the comment in `BraveBrowserViewTabbedLayoutImpl::GetHorizontalTabStripLeadingMargin` (b8bc95f9f4e) so the next reader doesn't have to re-derive it.

chromium_src/chrome/browser/ui/views/frame/browser_frame_view_win.cc
aguscruiz: Done in b8bc95f9f4e. Added a CSRC-017 block comment above `GetLayoutConstantForBraveWindowControls` in `chromium_src/chrome/browser/ui/views/frame/browser_frame_view_win.cc` that explains:

- why `kTabstripToolbarOverlap` is intercepted (caption-button vertical centering wants the smaller `tabs::GetHorizontalTabControlsDelta()` value, not the tab/toolbar geometry value);
- why this is scoped to the TU instead of folded into the central `GetBraveLayoutConstant()` (every other consumer of `kTabstripToolbarOverlap` does want the geometry value);
- a cross-reference to the matching wrapper in `horizontal_tab_strip_region_view.cc` so the two stay discoverable together.

rewrite/chrome/browser/ui/views/frame/horizontal_tab_strip_region_view.cc.toml
aguscruiz: Moot — addressed by removing the substitution entirely (b8bc95f9f4e). The `kTabstripToolbarOverlap` rewrite/patch is gone; the only substitution left in `rewrite/.../horizontal_tab_strip_region_view.cc.toml` is the `std::make_unique<TabStrip>` → `std::make_unique<BraveTabStrip>` swap, which is a simple-symbol pattern that `pattern` handles correctly. The control-overlap retargeting now lives in a chromium_src `#define GetLayoutConstant` wrapper instead, so `pattern_re` is not needed.

chromium_src/chrome/browser/ui/views/frame/browser_frame_view_layout_linux.cc
```diff
@@ -53,4 +54,6 @@ void BrowserFrameViewLayoutLinux::SetBoundsForButton(
     static_cast<views::FrameCaptionButton*>(button)->SetInkDropCornerRadius(
         views::kCaptionButtonInkDropDefaultCornerRadius);
   }
+
+  button->SetY(button->y() + tabs::GetHorizontalTabControlsDelta());
```
netzenbot: nit: This `SetY` call sits inside the vertical-tabs code path (the function returns early when `should_window_caption_buttons_overlap_toolbar` is false), so it always shifts Linux vertical-tab caption buttons by `GetHorizontalTabControlsDelta()` — which returns -4 even when compact horizontal tabs is disabled. If compact horizontal tabs and vertical tabs are mutually exclusive, this is an unconditional -4px change to vertical-tab button placement on Linux. Per [CSRC-017](https://github.com/brave/brave-core/tree/master/docs/best-practices/chromium-src-overrides.md#CSRC-017), a comment explaining why this delta is correct here (or a `ShouldUseCompactHorizontalTabsForNonTouchUI()` guard if it is not intentional in the default case) would prevent future confusion.

aguscruiz: Good catch — fixed by dropping the `SetY` adjustment entirely (commit incoming).

Tracing the call chain: this entire function is the vertical-tabs Linux path (it returns early unless `ShouldShowBraveVerticalTabs(browser) && !ShouldShowWindowTitleForVerticalTabs(browser)`). `tabs::GetHorizontalTabControlsDelta()` is meant for *horizontal*-tab control/caption alignment — and since vertical tabs and `#brave-compact-horizontal-tabs` are visually mutually exclusive, the delta has no business here. Pre-PR this code path was just `button->SetY(toolbar_origin.y())` with no extra offset, and that was already correctly aligned, so the default `-4` was an unintentional regression.

Removed the `button->SetY(button->y() + tabs::GetHorizontalTabControlsDelta())` line (and the now-unused `chrome/browser/ui/layout_constants.h` include). The Mac/Win/Opaque callsites (`brave_browser_frame_view_mac.mm`, `brave_browser_frame_view_win.cc`, `brave_opaque_browser_frame_view.cc`) keep `GetHorizontalTabControlsDelta()` because they sit in horizontal-tab frame paths where `-4` is the existing baseline and `-5` covers compact mode — that asymmetry with the Linux vertical-tab path is the point of the fix.

sangwoo108 (approved):
LGTM

browser/ui/tabs/brave_compact_horizontal_tabs_layout.h
```diff
@@ -0,0 +1,113 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
```
goodov, line 1: chromium_src is for chromium overrides only. that's a new brave-specific file, it shouldn't be here.

aguscruiz: @goodov ok updated, would you mind reviewing again please? Thanks

chromium_src/chrome/browser/ui/views/frame/browser_frame_view_win.cc
```diff
@@ -30,7 +33,63 @@
            ? WebUITabStripContainerView::SupportsTouchableTabStrip(browser) \
            : tabs::utils::SupportsBraveVerticalTabs(browser))
 
+namespace brave {
+
+static thread_local int g_win_caption_geometry_tabstrip_overlap_depth = 0;
+
+ScopedWinCaptionLayoutUsesGeometryTabstripOverlap::
```
goodov: that's a bit unusual that this is declared in brave files, but implemented in the chromium override.. This might make sense in some rare cases, but I'd say here it just adds confusion for no good reason. Or is there a good reason?

how about implement it in //brave/**.cc file, move the thread_local static there and expose it via a static function in the scoped class? like 
```
class ScopedWinCaptionLayoutUsesGeometryTabstripOverlap {
...
public:
 static int GetCurrentWinCaptionGeometryTabstripOverlapDepth();
```

aguscruiz: Thanks for the feedback — agreed that splitting declaration vs TLS implementation across Brave headers and `chromium_src` was awkward.

**What we changed**

- Moved the `thread_local` depth counter and the `ScopedWinCaptionLayoutUsesGeometryTabstripOverlap` ctor/dtor into `browser/ui/views/frame/brave_win_caption_layout.cc` under `//brave`.
- Added `ScopedWinCaptionLayoutUsesGeometryTabstripOverlap::GetCurrentWinCaptionGeometryTabstripOverlapDepth()` on the scoped class so consumers don’t touch an internal global.
- `chromium_src/.../browser_frame_view_win.cc` now only calls that static getter inside `GetLayoutConstantForBraveWindowControls`; no Brave TLS or RAII bodies live there anymore.
- Wired `brave_win_caption_layout.{cc,h}` into `browser/ui/BUILD.gn` for Windows.

Behavior is unchanged; only where the implementation lives and how depth is queried.

goodov (approved):

browser/ui/commander/commander_service_browsertest.cc
```diff
@@ -42,6 +43,7 @@ class CommanderServiceBrowserTest : public InProcessBrowserTest {
   void SetUp() override { InProcessBrowserTest::SetUp(); }
   void TearDownOnMainThread() override {
     commander()->Hide();
```
netzenbot, line 45: `content::RunAllTasksUntilIdle()` is banned under [TA-005](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-async.md#TA-005) — it shares the same flakiness characteristics as `RunLoop::RunUntilIdle()` and can return too early when events span different task queues. The `WaitUntil(!IsShowing())` call that follows already provides the necessary synchronization; this extra flush is a timing pad, not a root-cause fix. Remove it and rely solely on `WaitUntil`.

PR #35795, opened 2026-04-22
Account: migrate `BraveAccountService` to async `OSCrypt`

### LINKED ISSUE #54817: Account: migrate `BraveAccountService` from synchronous `OSCrypt` to asynchronous `OSCrypt`

szilardszaloki:
Resolves https://github.com/brave/brave-browser/issues/54817.

Follow-ups:
- https://github.com/brave/brave-browser/issues/54998


kylehickinson (approved):
ios++

components/brave_account/brave_account_service.cc
```diff
@@ -129,29 +143,61 @@ BraveAccountService::BraveAccountService(
...
+            return encryptor.EncryptString(plaintext, ciphertext);
+          },
+          base::OwnedRef(std::move(encryptors[0]))),
+      base::BindRepeating(
+          [](const os_crypt_async::Encryptor& encryptor,
+             const std::string& ciphertext, std::string* plaintext) {
+            return encryptor.DecryptString(ciphertext, plaintext);
+          },
+          base::OwnedRef(std::move(encryptors[1]))));
+}
```
cdesouza-chromium, line 170: Do you think we can remove these thunks? In general, we are incentivised to not use callbacks when there are multiple of them, but rather a delegate. They have a shared refptr inside of them, and they are heavier than a simple delegate for dispatching. In this particular case we don't even need a delegate because in production code we don't need this type of dispatching either, so the production code is taking a path that is more complex than it needs to be.

Can we just make the encryptor a member, and have the tests rely on the test machinery from `os_crypt_async`?

cdesouza-chromium (commented):
Thanks Szilard for this. It really helps. Nearly there, just one question that I had brought in DM before.

components/brave_account/brave_account_service.cc
szilardszaloki: Yeah, I checked, and we can't drop these callbacks, unfortunately. The test-only constructor is what lets tests force encrypt/decrypt failures and cover user-visible error mappings (`kVerificationTokenEncryptionFailed`, `kVerificationTokenDecryptionFailed`, `kAuthenticationTokenEncryptionFailed`, `kAuthenticationTokenDecryptionFailed`, etc.). Those paths handle `OSCrypt` platform failures.

cdesouza-chromium: I see, so you so the test encryptor does not provide us ways to override that behaviour it seems (components/os_crypt/async/common/test_encryptor.h)

I'm not gonna block you on this then, but I'll leave something in here for thought that I think would improve things, specially with efficiency and memory consumption in production code:

When it comes to test code paths by Chromium practice, we are incentivised to think in terms of customisation points that do not require incur into additional load into production code. For instance, in some cases things can be kept to a minimum, through some global control that is only used when testing the code:

For example, in cases like this throughout the application we tend to do something like:

```cxx
namespace {
EncryptCallback* encryt_callback_for_testing = nullptr;
DecryptCallback* decrypt_callback_for_testing = nullptr;
}
```

We have some discrete storage for the test only code/ Then we offer a `ForTesting` method, that can only be called in tests

```cxx
  void MyClass::SetEncryptCallbacksForTesting(...) {
    CHECK_IS_TEST();
    encryt_callback_for_testing = ...
   decrypt_callback_for_testing = ...
  }
```

and in the places where you would use the encryptor, you would just do:

```cxx
if (encryt_callback_for_testing) {
  CHECK_IS_TEST();
  encryt_callback_for_testing(...);
  return;
}

encryptor_->EncryptString(...);
```

Chromium does not subscribe to the principle that test code has to be the same as production code, because there is a performance and memory cost to that, and in reality they aren't the same, but of course this is predicated on the exceptions being easy to read, and hard to get wrong. This also creates an added benefit where we can easily see what is testing what is not.

I'm gonna leave it to you to decide, but it would be nice if we could drop these callbacks from this type.

cdesouza-chromium (approved):

components/brave_account/brave_account_service.cc
```diff
@@ -110,16 +111,29 @@ auto MakeMojomError(int status_code, ErrorBody error_body) {
 
 BraveAccountService::BraveAccountService(
     PrefService* pref_service,
-    scoped_refptr<network::SharedURLLoaderFactory> url_loader_factory)
-    : BraveAccountService(pref_service,
-                          std::move(url_loader_factory),
-                          base::BindRepeating(&OSCrypt::EncryptString),
-                          base::BindRepeating(&OSCrypt::DecryptString)) {}
+    scoped_refptr<network::SharedURLLoaderFactory> url_loader_factory,
+    os_crypt_async::OSCryptAsync* os_crypt_async)
```
cdesouza-chromium, line 115: nit: `os_crypt_async::OSCryptAsync&`

components/brave_account/brave_account_service.cc
szilardszaloki: Thanks! For this PR I'd rather keep the scope to the migration and not mix in extra refactoring, but I see what you're getting at — I'll give it some thought and circle back.

components/brave_account/brave_account_service.cc
szilardszaloki: I had it as a reference initially, but upstream consistently passes `OSCryptAsync` as a pointer, so I followed that pattern for consistency.

components/brave_account/brave_account_service.cc
szilardszaloki: https://github.com/brave/brave-browser/issues/54998

PR #35797, opened 2026-04-22
Upgrade from Chromium 147.0.7727.102 to Chromium 147.0.7727.117 (1.90.x)

emerick:
Resolves https://github.com/brave/brave-browser/issues/54850


bsclifton:
Going to admin merge here - we have some audit_deps failures which are known are are OK as they should not impact the final binary (only impacts dev tooling). Employees can see more on Slack in https://bravesoftware.slack.com/archives/C7VLGSR55/p1776860633403999?thread_ts=1776849498.996009&cid=C7VLGSR55

PR #35804, opened 2026-04-22
[ads] Fix idle notification dropped by IDLE_STATE_UNKNOWN transitions

tmancey:
When the OS returns IDLE_STATE_UNKNOWN, the previous code updated last_idle_state_ even though no notification was sent, corrupting the last known good state. An ACTIVE to UNKNOWN to ACTIVE sequence dropped the idle notification entirely; an IDLE to UNKNOWN to IDLE sequence fired a duplicate idle notification. Returning early on UNKNOWN preserves the last known state. Adds FakeBatAdsClientNotifier and TestingAdsServiceImpl with two regression tests covering both failure modes.

Resolves

### DIFF
--- a/components/brave_ads/browser/ads_service_impl_unittest.cc
+++ b/components/brave_ads/browser/ads_service_impl_unittest.cc
@@ -13,6 +13,7 @@
 #include "base/test/run_until.h"
 #include "base/test/scoped_feature_list.h"
 #include "base/test/task_environment.h"
+#include "base/time/time.h"
 #include "base/values.h"
 #include "brave/components/brave_ads/browser/test/fake_ads_service_delegate.h"
 #include "brave/components/brave_ads/browser/test/fake_ads_tooltips_delegate.h"
@@ -31,6 +32,7 @@
 #include "components/prefs/testing_pref_service.h"
 #include "components/variations/pref_names.h"
 #include "testing/gtest/include/gtest/gtest.h"
+#include "ui/base/idle/idle.h"
 
 #if BUILDFLAG(ENABLE_BRAVE_REWARDS)
 #include "brave/components/brave_ads/browser/test/fake_rewards_service.h"
@@ -107,11 +109,13 @@ class BraveAdsAdsServiceImplTest : public testing::Test {
 
   void Shutdown() { ads_service_->Shutdown(); }
 
-#if BUILDFLAG(IS_ANDROID)
-  base::test::ScopedFeatureList scoped_feature_list_;
-#endif  // BUILDFLAG(IS_ANDROID)
+  void SimulateIdleState(ui::IdleState idle_state, base::TimeDelta idle_time) {
+    ads_service_->ProcessIdleState(idle_state, idle_time);
+  }
 
   base::test::TaskEnvironment task_environment_;
+
+  base::test::ScopedFeatureList scoped_feature_list_;
 
   base::ScopedTempDir profile_dir_;
 
@@ -376,7 +380,6 @@ TEST_F(
 }
 #endif  // BUILDFLAG(ENABLE_BRAVE_REWARDS)
 
-#if !BUILDFLAG(IS_ANDROID)
 TEST_F(BraveAdsAdsServiceImplTest,
        ServiceDoesNotStartForSearchResultAdsWhenRewardsIsDisabledByPolicy) {
   // Arrange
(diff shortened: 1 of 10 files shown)


aseren (approved):

aseren (approved):

SergeyZhukovsky (approved):
++

components/brave_ads/browser/ads_service_impl_unittest.cc
```diff
@@ -32,6 +30,16 @@
 #include "components/variations/pref_names.h"
 #include "testing/gtest/include/gtest/gtest.h"
 
+#if BUILDFLAG(IS_ANDROID)
+#include "base/test/scoped_feature_list.h"
+#include "brave/components/brave_rewards/core/features.h"
+#endif  // BUILDFLAG(IS_ANDROID)
+
+#if !BUILDFLAG(IS_ANDROID)
+#include "base/time/time.h"
```
bridiver: these includes are not desktop specific so they should not be guarded by !IS_ANDROID

components/brave_ads/browser/ads_service_impl_unittest.cc
```diff
@@ -32,6 +30,16 @@
 #include "components/variations/pref_names.h"
 #include "testing/gtest/include/gtest/gtest.h"
 
+#if BUILDFLAG(IS_ANDROID)
+#include "base/test/scoped_feature_list.h"
```
bridiver: these includes are not android specific so they shouldn't be guarded by IS_ANDROID

components/brave_ads/browser/ads_service_impl_unittest.cc
```diff
@@ -107,12 +115,18 @@ class BraveAdsAdsServiceImplTest : public testing::Test {
 
   void Shutdown() { ads_service_->Shutdown(); }
 
+#if !BUILDFLAG(IS_ANDROID)
+  void SimulateIdleState(ui::IdleState idle_state, base::TimeDelta idle_time) {
```
bridiver, line 112: why is this !IS_ANDROID?

components/brave_ads/browser/test/fake_bat_ads_service_factory.h
```diff
@@ -9,13 +9,19 @@
...
 #include "brave/components/brave_ads/browser/bat_ads_service_factory.h"
+#include "build/build_config.h"
 #include "mojo/public/cpp/bindings/remote.h"
 
 namespace brave_ads::test {
 
 class FakeBatAdsService;
 
+#if !BUILDFLAG(IS_ANDROID)
+class FakeBatAdsClientNotifier;
```
bridiver, line 18: unless I'm missing something none of this BatAdsClientNotifier stuff appears to be behind !IS_ANDROID  so it's not clear why it's behind the flag here?

tmancey:
```
void AdsServiceImpl::CheckIdleStateAfterDelay() {
#if !BUILDFLAG(IS_ANDROID)
  idle_state_timer_.Stop();

  idle_state_timer_.Start(FROM_HERE, base::Seconds(1), this,
                          &AdsServiceImpl::CheckIdleState);
#endif
}
```
We do not use idle detection to deliver notification ads on Android, this applies only to desktop. On mobile, notifications are served at regular intervals.

However, I have made a few changes around IS_ANDROID

bridiver: for follow-up I would convert this to `brave_ads_idle_detection_enabled = !is_android` or something and put the entire method behind the flag. OS flags should be for OS-specific code, not OS enabled features.

components/brave_ads/browser/ads_service_impl_unittest.cc
bridiver: above https://github.com/brave/brave-core/pull/35804/changes#r3130641960 the flag is inside the method so shouldn't it be the same here and then change them both together later to be around the method? This is what makes it especially confusing imo

tmancey: Fixed by removing guards.

components/brave_ads/browser/test/fake_bat_ads_service_factory.h
tmancey: Fixed by removing guards.

components/brave_ads/browser/ads_service_impl_unittest.cc
tmancey: Fixed by removing guards.

components/brave_ads/browser/ads_service_impl_unittest.cc
tmancey: Fixed by removing guards.

bridiver (approved):

PR #35815, opened 2026-04-23
[CodeHealth] Fix `Prefs::GetSeed` interface

cdesouza-chromium:
This PR corrects `GetSeed` to be easier to use and less bug prone. This
is being done in preparation to the `os_crypt_sync` migration. An
interface change was the addition of `IsEncryptionAvailable()`, which
will be useful when this type is given an encryptor.

This PR also removes a lot of the os crypt setup and tearr down for
tests, to make it easier to have them removed once they are not needed
anymore.

### DIFF
--- a/browser/infobars/sync_cannot_run_infobar_delegate.cc
+++ b/browser/infobars/sync_cannot_run_infobar_delegate.cc
@@ -23,17 +23,6 @@
 #include "ui/base/l10n/l10n_util.h"
 #include "ui/views/vector_icons.h"
 
-namespace {
-
-bool SeedDecryptionFailed(raw_ptr<brave_sync::Prefs> brave_sync_prefs) {
-  CHECK_NE(brave_sync_prefs, nullptr);
-  bool failed_to_decrypt = false;
-  std::string seed = brave_sync_prefs->GetSeed(&failed_to_decrypt);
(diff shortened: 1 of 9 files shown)


AlexeyBarabash (approved):
looks better than it was

emerick (approved):

browser/infobars/sync_cannot_run_infobar_delegate.cc
```diff
@@ -44,7 +33,8 @@ void SyncCannotRunInfoBarDelegate::Create(
     return;
   }
```
netzenbot, line 35: This logic is inverted. `GetSeed()` returns `std::nullopt` on decryption failure, so `!has_value()` is `true` when decryption fails — and the early `return` suppresses the infobar in exactly the case it should appear. The original code returned early when `!SeedDecryptionFailed()` (i.e., when decryption succeeded), showing the infobar only on failure. The condition here should be `if (brave_sync_prefs.GetSeed().has_value())` to preserve that behavior. The comment `// Decription failed.` (also a typo) is on the wrong branch.

cdesouza-chromium: Fixed

netzenbot (approved):

PR #35816, opened 2026-04-23
[AI Chat]: Use `<input type="file">` for uploading files rather than custom C++ implementation

### LINKED ISSUE #55107: [AI Chat]: Refactor file uploads to use `<input type=file>`
Currently, we have two largely independent paths for uploading files.
1. Upload: We trigger a dialog from C++ and upload the file before passing it back to the frontend
2. DragNDrop/Paste: The user drops or pastes a file before we hand it off to C++ for processing and making it available as an attachment.

We should unify these two paths by using the existing WebAPI for uploading a file, rather than handling it over mojom.

Test Plan:
we want to end up with the same functionality as we currently have, just with less code. With that in mind:
1. It should be possible to upload files to a Leo conversation by going through the attachments menu and selecting `Attach files` (it should trigger a dialog).
2. On desktop it should be possible to pick any sort of file (though some files, such as executables should fail to attach).
3. On Android, it should be possible to pick Pdf or Images (no text).
4. It should be possible to paste files into the input
5. It should be possible to drag and drop files onto the conversation
6. When submitting a text or PDF file to Leo, it should be able to see the file contents.

fallaciousreasoning:
Resolves https://github.com/brave/brave-browser/issues/55107

**Note:** There should be no functional changes here, it just reuses the existing upload pattern on the web for picking files.


browser/ui/webui/ai_chat/ai_chat_ui_page_handler.cc
```diff
@@ -220,133 +255,11 @@ void AIChatUIPageHandler::ShowSoftKeyboard() {
...
-#endif
-        CHECK(extractor);
-        auto* extractor_ptr = extractor.get();
-        extractors_.push_back(std::move(extractor));
-        extractor_ptr->ExtractText(
-            profile_, info.path,
-            base::BindOnce(&AIChatUIPageHandler::OnSingleFileExtracted,
-                           weak_ptr_factory_.GetWeakPtr(), extractor_ptr,
-                           info.index, barrier));
-      }
```
darkdh, line 298: you can't just delete pdf text extraction and also sandboxed text extraction

darkdh (changes requested):

components/ai_chat/resources/page/utils/file_utils.ts
```diff
@@ -27,78 +27,34 @@ export class ImageProcessingError extends Error {
...
+  const type = fileTypeToUploadedFileType(file)
+  return {
+    filename: file.name,
+    filesize: file.size,
+    data: Array.from(new Uint8Array(await file.arrayBuffer())),
+    type,
+    extractedText:
+      type === Mojom.UploadedFileType.kText ? await file.text() : undefined,
+  } as Mojom.UploadedFile
+}
```
darkdh: you also stripped image sanitation flow and we are now not doing any at all. You should call `processImageFile` for `image/*` here

components/ai_chat/resources/page/components/attachment_button_menu/index.tsx
```diff
@@ -85,7 +97,16 @@ export default function AttachmentButtonMenu(props: Props) {
           </leo-menu-item>
         )}
         {props.isMobile && (
-          <leo-menu-item onClick={() => props.uploadFile([true])}>
+          <leo-menu-item
+            onClick={async () => {
+              const files = await pickFiles({
+                capture: 'camera',
```
darkdh: we need to restrict android accept type as 
`accept="image/png,image/jpeg,image/webp,application/pdf,text/*"`

components/ai_chat/resources/page/utils/file_utils.ts
```diff
@@ -27,78 +27,34 @@ export class ImageProcessingError extends Error {
...
-
-    return processedFile
+    return Mojom.UploadedFileType.kImage
   } else {
-    // Everything else is treated as a text file. The renderer handles
-    // MIME sniffing and will render the file as text if possible.
-    // If extraction fails (e.g. binary file), null is returned and
-    // the file is not attached.
-    if (processTextFile) {
-      const processedFile = await processTextFile([
```
darkdh, line 94: non-image/non-PDF files need `processTextFile` to reject binary files (.zip, exe, .tar... etc.)

components/ai_chat/resources/page/utils/file_utils.ts
```diff
@@ -27,78 +27,34 @@ export class ImageProcessingError extends Error {
...
-  })
-
-  const uint8Array = new Uint8Array(arrayBuffer)
-
-  // Check file type and handle accordingly
+const fileTypeToUploadedFileType = (file: File): Mojom.UploadedFileType => {
   const mimeType = file.type.toLowerCase()
   if (mimeType === 'application/pdf') {
-    if (processPdfFile) {
-      const processedFile = await processPdfFile([
```
darkdh, line 58: we now lost basic pdf validation (`printing::LooksLikePdf()` on any .pdf file that contains invalid headers) and instead relies on `MIME application/pdf` and completely trust those bytes. So we basically accept a pdf that was renamed from a zip which is not a pdf file at all, not to mention malicious pdf files.
It was done in `processPdfFile` which validates via `PdfTextExtractor` internally (malformed PDFs fail extraction)

darkdh (changes requested):
This is not "no functional changes" at all, you just deleted many important protections in the file upload flow

components/ai_chat/resources/page/utils/file_utils.test.ts
```diff
@@ -3,370 +3,151 @@
...
 
-    it('handles different file types', async () => {
-      const file = createMockFile('screenshot.jpeg', 'image/jpeg')
-      const mockArrayBuffer = new ArrayBuffer(16)
-      const mockProcessedFile = {
-        filename: 'screenshot.jpeg',
-        filesize: 1000,
-        data: [5, 6, 7, 8],
-        type: Mojom.UploadedFileType.kScreenshot,
+    expect(mockAttachFiles).toHaveBeenCalledWith([
```
netzenbot: The deleted `upload_file_helper_unittest.cc` contained tests for invalid PDF headers, binary file rejection (zip, exe), and undersized PDFs. These checks must be relocated to the appropriate new test location rather than simply removed. Per TI-017, test coverage should never decrease during refactoring — if those validations are no longer enforced, the tests should document why (and if they still are enforced, they need tests). https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-017

components/common/uploadFile.ts
```diff
@@ -0,0 +1,67 @@
...
+  multiple?: boolean
+  /**
+   * Comma-separated accept list (HTML `accept` attribute), e.g. `"image/*"`,
+   * `"application/pdf"`, or `".pdf,.docx"`.
+   */
+  accept?: string
+  /**
+   * Passed through to the input `capture` attribute when using file inputs for
+   * camera/microphone where supported (typically mobile).
+   */
```
netzenbot, line 17: The `pickFiles` function in `components/common/uploadFile.ts` is a shared utility with non-trivial logic: a `settled` guard against double-resolution, two event listeners, DOM cleanup, and a cancel path. There are no tests for it. The `components/common/` directory already has `*.test.ts` files for other utilities (e.g., `mojomUtils.test.ts`, `classnames.test.ts`), so adding a `uploadFile.test.ts` with jsdom-based tests covering the cancel path, multi-file selection, and cleanup would fit naturally. ([PRs Should Include Reasonable Test Coverage](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-039))

browser/ui/webui/ai_chat/ai_chat_ui_page_handler.cc
```diff
@@ -17,6 +17,7 @@
 #include "base/functional/bind.h"
 #include "base/functional/callback_forward.h"
 #include "base/strings/utf_string_conversions.h"
+#include "base/task/thread_pool.h"
```
netzenbot, line 20: nit: The `UploadFile` method (the only user of `ChromeSelectFilePolicy`) is deleted in this PR, but `#include "chrome/browser/ui/chrome_select_file_policy.h"` is not removed. After this PR, that include is unused — it should be dropped per IWYU. ([✅ Always Include What You Use (IWYU)](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-001))

components/ai_chat/resources/page/utils/file_utils.ts
```diff
@@ -27,78 +27,34 @@ export class ImageProcessingError extends Error {
...
+}
+
+// Utility function to convert File objects to UploadedFile format
+export const processFiles = async (
+  files: File[],
+  attachFiles: ConversationContext['attachFiles'],
+) => {
+  const uploadedFiles = await Promise.all(files.map(toUploadedFile))
+  await attachFiles(uploadedFiles)
+}
```
netzenbot: nit: `FileReadError` and `ImageProcessingError` are no longer used or imported anywhere after `convertFileToUploadedFile` is removed. Per [CS-068](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-068), dead code should be cleaned up when removing features. These classes can be deleted along with the old function.

browser/ui/views/side_panel/ai_chat/ai_chat_side_panel_web_view.h
```diff
@@ -48,6 +57,10 @@ class AIChatSidePanelWebView : public SidePanelWebUIViewT<AIChatUI> {
       base::OnceCallback<void(content::NavigationHandle&)>
           navigation_handle_callback) override;
 
+  void RunFileChooser(content::RenderFrameHost* render_frame_host,
+                      scoped_refptr<content::FileSelectListener> listener,
+                      const blink::mojom::FileChooserParams& params) override;
+
  private:
   // This callback is invoked multiple times, so we need to ensure that
```
netzenbot, line 66: nit: Consider adding a brief comment explaining why this override is needed — e.g. that the side panel's WebContents doesn't have a browser-level file picker by default and must delegate to `browser_view->browser()`. Without it, the intent of the delegation isn't obvious. ([Method Documentation Should Describe the Contract](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-004))

components/ai_chat/resources/page/utils/file_utils.ts
fallaciousreasoning: Sorry, I actually misread the function in `conversation_context, and thought this was all happening there (the function is called `processFiles`) so I thought we were still hitting the C++ side.

components/ai_chat/resources/page/utils/file_utils.ts
fallaciousreasoning: In my head this was happening in the `processFile` function in `attachFiles` too but it definitely isn't.

I've moved the `processXXX` logic into there

components/ai_chat/resources/page/components/attachment_button_menu/index.tsx
fallaciousreasoning: Do we also need to restrict the desktop `accept` too?

fallaciousreasoning: I've gone with this accept for both:
`image/*,application/pdf,text/*`

If we fail to process on of the image types because we don't support decoding it the user will just see an error which I think is fine.

darkdh: desktop currently accepts everything, and special processing for image and pdf them delegate rest of the files to a renderer to do extracting text. So if you upload a dmg, we will show error on UI.
android doesn't support that so that is the reason we need to keep it as it is now

browser/ui/webui/ai_chat/ai_chat_ui_page_handler.cc
fallaciousreasoning: I think this is still handled through `ProcessXXX`. This code path was only being executed when files were uploaded.

I'm not 100% sure but this is how I thought the two different paths were working. Its pretty confusing though so I could be wrong:

**UploadFile**
Frontend ==> UploadFile() ==> C++ shows UI ==> C++ processes uploaded files (this deleted code) ==> C++ tells frontend it has the files ready ==> Front end gets the files ==> stages the attached files

**DragNDrop**
Frontend receives `File` ==> calls `ProcessXXX` ==> Gets back an `UploadedFile` ==> stages the attached files

Now `UploadFile` goes through the same flow as drag and drop, so we don't need this anymore. `ProcessXXX` still does the `PdfTextExtraction` and `SandboxedTextExtraction`, and I don't think `DragNDrop` ever went through this.

That said, very possible I'm still misunderstanding something - the old `UploadFile` flow was pretty complicated.

components/ai_chat/resources/page/components/attachment_button_menu/index.tsx
darkdh: so just `image/*,application/pdf` for android and accept everything from desktop

browser/ui/webui/ai_chat/ai_chat_ui_page_handler.cc
darkdh: You're right that `<input type="file">` should share the drag/drop path — one File-handling pipeline for both is clearly the correct shape, and that's a nice cleanup.

The thing is, this PR simplified the drag/drop path too: in master, drag/drop ran files through `processImageFile` / `processPdfFile` / `processTextFile`, but `toUploadedFile` no longer calls any of them. So after the PR both paths skip backend processing — PDF extraction and the sandboxed `TextFileExtractor` (view-source in a scriptless/originless WebContents with download-detection) are dead code for both.

The unified path should keep calling `processPdfFile` / `processTextFile` per MIME, so both `<input type="file">` and drag/drop benefit from the extractors. That preserves the architectural win here without dropping the sandboxed extraction.

As a follow-up once this is resolved: the path-based `FileTextExtractorBase::ExtractText(BrowserContext*, const FilePath&, callback)` overload has no production callers after the deletion of `OnFilesUploaded` — only tests use it. Worth a cleanup PR to drop it along with `LoadInWebContents(BrowserContext*, const FilePath&)`.

browser/ui/webui/ai_chat/ai_chat_ui_page_handler_browsertest.cc
```diff
@@ -316,33 +316,6 @@ IN_PROC_BROWSER_TEST_F(AIChatUIPageHandlerBrowserTest, ProcessPdfFile) {
...
-  page_handler->OnFilesUploaded(future.GetCallback(),
-                                std::make_optional(std::move(files)));
-
-  auto result = future.Take();
-  ASSERT_TRUE(result.has_value());
-  ASSERT_EQ(result->size(), 1u);
-  EXPECT_EQ((*result)[0]->filename, "dummy.pdf");
-  ASSERT_TRUE((*result)[0]->extracted_text.has_value());
-  EXPECT_EQ(*(*result)[0]->extracted_text, kExpectedPdfText);
-}
```
darkdh, line 345: We can't just discard this without equivalent replacement(same as `OnFilesUploaded_WithText`), these are the cases to make sure renderer extraction works. And they are not unit testable, which is why they are in browser test.

browser/ai_chat/upload_file_helper_unittest.cc
```diff
@@ -1,507 +0,0 @@
...
-  // Browser::OpenFile works. The renderer handles MIME sniffing.
-  ASSERT_TRUE(dialog_params_.file_types);
-  EXPECT_TRUE(dialog_params_.file_types->extensions.empty());
-#if BUILDFLAG(IS_ANDROID)
-  // Android doesn't support view-source for text file extraction, so only
-  // image and PDF MIME types are accepted.
-  EXPECT_THAT(dialog_params_.accept_types,
-              testing::UnorderedElementsAre(u"image/png", u"image/jpeg",
-                                            u"image/webp", u"application/pdf"));
-#endif
```
darkdh, line 148: we need a equivalent test to make sure these are restricted on android

browser/ui/webui/ai_chat/ai_chat_ui_page_handler.cc
fallaciousreasoning: Yup, I know - I thought the `processUploadedFiles` function in `ConversationContext` was calling `ProcessXXXFile` but its not. I've updated ti to unify all the `ProcessXXX` calls in one place.

components/ai_chat/resources/page/utils/file_utils.test.ts
fallaciousreasoning: Added separate tests in `ai_chat_page_ui_handler`

browser/ui/webui/ai_chat/ai_chat_ui_page_handler.cc
darkdh: Please open the follow-up issue as I mentioned and posted in this thread so I can resolve this conversation.

netzenbot (approved):

components/ai_chat/resources/page/utils/file_utils.test.ts
darkdh: don't resolve the comment if you didn't push changes to address it.

browser/ui/webui/ai_chat/ai_chat_ui_page_handler_browsertest.cc
fallaciousreasoning: FWIW, I think these tests still existed in the ProcessXXXFile tests. I've removed those too though, in favor of a more complete test `ProcessUploadedFilesFromFrontend`

browser/ai_chat/upload_file_helper_unittest.cc
fallaciousreasoning: Will it be sufficient to check the correct `accept` is being passed to `pickFiles`? I'll add a typescript unit test for this.

browser/ui/views/side_panel/ai_chat/ai_chat_side_panel_web_view.cc
```diff
@@ -137,6 +138,18 @@ content::WebContents* AIChatSidePanelWebView::AddNewContents(
   return params.navigated_or_inserted_contents;
 }
 
+void AIChatSidePanelWebView::RunFileChooser(
+    content::RenderFrameHost* render_frame_host,
+    scoped_refptr<content::FileSelectListener> listener,
+    const blink::mojom::FileChooserParams& params) {
+  auto* browser_view = BrowserView::GetBrowserViewForNativeWindow(
+      GetWidget()->GetNativeWindow());
```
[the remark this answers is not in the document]
fallaciousreasoning: /fp - I think this is actually what we want here.

browser/ui/webui/ai_chat/ai_chat_ui_page_handler.cc
fallaciousreasoning: https://github.com/brave/brave-browser/issues/55017

browser/ui/webui/ai_chat/ai_chat_ui_page_handler.cc
fallaciousreasoning: removed

components/ai_chat/resources/page/utils/file_utils.ts
fallaciousreasoning: removed

browser/ui/views/side_panel/ai_chat/ai_chat_side_panel_web_view.h
fallaciousreasoning: done

browser/ai_chat/upload_file_helper_unittest.cc
fallaciousreasoning: added

browser/ui/webui/ai_chat/ai_chat_ui_page_handler_browsertest.cc
fallaciousreasoning: Added an end to end test

components/ai_chat/resources/page/components/attachment_button_menu/attachment_button_menu.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+undefined
+// </if>
+
+describe('attachment button menu', () => {
+  beforeEach(() => {
+    mockPickFiles.mockResolvedValue([])
+  })
+
+  it('upload menu item passes correct accept to pickFiles', async () => {
+    const { container } = render(<AttachmentButtonMenu {...defaultProps} />)
```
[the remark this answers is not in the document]
fallaciousreasoning: This open grep rule should only apply to raw HTML elements (lower case components) rather than any React component.

components/ai_chat/resources/page/components/attachment_button_menu/attachment_button_menu.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+    expect(mockPickFiles).toHaveBeenCalledWith({
+      multiple: true,
+      accept: expectedAccept,
+    })
+  })
+
+  it('camera menu item passes correct accept to pickFiles', async () => {
+    const { container } = render(
+      <AttachmentButtonMenu
+        {...defaultProps}
```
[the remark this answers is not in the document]
fallaciousreasoning: ditto

browser/ui/webui/ai_chat/ai_chat_ui_page_handler_browsertest.cc
```diff
@@ -256,191 +273,150 @@ IN_PROC_BROWSER_TEST_F(AIChatUIPageHandlerBrowserTest,
   ai_chat_contents->Close();
 }
 
-IN_PROC_BROWSER_TEST_F(AIChatUIPageHandlerBrowserTest, ProcessImageFile) {
```
fallaciousreasoning, line 259: There were two close-to-duplicated test flows here for Upload and Process (the paths for which have been unified).

I've also removed the ProcessXXXFile tests in favor of an approach which kicks off from the frontend to get a more complete end-to-end test.

browser/ui/webui/ai_chat/ai_chat_ui_page_handler_unittest.cc
```diff
@@ -128,94 +128,4 @@ TEST_F(AIChatUIPageHandlerTest, AssociateUrlContent_InvalidConversation) {
   // Should not crash
 }
 
-TEST_F(AIChatUIPageHandlerTest, FinishUpload_StripsPathToBasename) {
```
fallaciousreasoning, line 131: These `Upload` related tests have been removed

components/ai_chat/resources/page/utils/file_utils.test.ts
fallaciousreasoning: sorry, was resolving so I could track what I'd done locally - can definitely see how it'd be confusing though

browser/ui/views/side_panel/ai_chat/ai_chat_side_panel_web_view.h
```diff
@@ -48,6 +57,11 @@ class AIChatSidePanelWebView : public SidePanelWebUIViewT<AIChatUI> {
       base::OnceCallback<void(content::NavigationHandle&)>
           navigation_handle_callback) override;
 
+  // Allow the file chooser to be user from the AI Chat side panel.
```
darkdh: s/user/used/

components/ai_chat/resources/page/components/attachment_button_menu/index.tsx
```diff
@@ -85,7 +104,17 @@ export default function AttachmentButtonMenu(props: Props) {
...
```
darkdh: why are we using `camera` as capture value? It is not a valid value and capture will be ignored then fallback to default behavior
https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/capture

browser/ai_chat/upload_file_helper_unittest.cc
```diff
@@ -1,507 +0,0 @@
...
-  // Should return a stub entry (empty data, kText) so the frontend can
-  // detect the rejection and show an error.
-  ASSERT_TRUE(result);
-  ASSERT_EQ(1u, result->size());
-  EXPECT_TRUE((*result)[0]->data.empty());
-  EXPECT_EQ((*result)[0]->type, mojom::UploadedFileType::kText);
-  EXPECT_FALSE((*result)[0]->extracted_text.has_value());
-}
-
-TEST_F(UploadFileHelperTest, PdfFileTooSmall) {
```
darkdh, line 306: We should not lose basic pdf validation. There is no migration for existing use of `LooksLikePdf` and these two tests to test it.

browser/ai_chat/upload_file_helper_unittest.cc
```diff
@@ -1,507 +0,0 @@
...
-#if BUILDFLAG(IS_ANDROID)
-  // Android doesn't support view-source for text file extraction, so only
-  // image and PDF MIME types are accepted.
-  EXPECT_THAT(dialog_params_.accept_types,
-              testing::UnorderedElementsAre(u"image/png", u"image/jpeg",
-                                            u"image/webp", u"application/pdf"));
-#endif
-}
-
-TEST_F(UploadFileHelperTest, ImageRead) {
```
darkdh, line 151: you moved the image sanitization and didn't move this unit test case to test it.
especially 
- empty PNG → null — tested DataDecoder's empty-input rejection
- 2048→1024×768 scale-down integration

components/ai_chat/resources/page/components/attachment_button_menu/attachment_button_menu.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+  conversationStarted: false,
+}
+
+// Mirrors the accept constant in the component — must match its preprocessor block exactly.
+const expectedAccept: string | undefined =
+  // <if expr="is_android">
+  'image/*,application/pdf'
+// <else>
+undefined
+// </if>
```
darkdh, line 36: how does these get preprocessed?

components/ai_chat/resources/page/utils/file_utils.ts
darkdh: this is now in `processUploadedFiles` of `components/ai_chat/resources/page/state/conversation_context.tsx`

components/ai_chat/resources/page/utils/file_utils.ts
darkdh: pdf extractor path is ok now but we still miss basic pdf validation like I mentioned in https://github.com/brave/brave-core/pull/35816#discussion_r3155927851

components/ai_chat/resources/page/state/conversation_context.tsx
```diff
@@ -418,19 +425,68 @@ export function useProvideConversationContext(props: ConversationContextProps) {
...
+        return null
+      }
+    }
+
+    const processedFiles = await Promise.all(files.map(processFile))
+    return attachUploadedFiles(processedFiles)
+  }
+
+  // Let the test know about processUploadedFiles, so it can be used.
+  ;(window as any).__onTestProcessFilesReady__?.(processUploadedFiles)
```
darkdh: we are exporting `window.__onTestProcessFilesReady__`  and `window.___onTestFilesChanged__` in both production and test. I don't think e2e test should be the scope of this PR and it shouldn't even blend with normal browser tests.
Also @brave/sec-team needs to approve this move.

browser/ui/webui/ai_chat/ai_chat_ui_page_handler_browsertest.cc
darkdh: This e2e test should be removed. Test the C++ handlers by calling `ProcessImageFile` / `ProcessPdfFile` / `ProcessTextFile` directly — the way the deleted browsertests did. That's the only test we want here.

  The e2e approach forced three pieces of test-only machinery into production code (`__onTestProcessFilesReady__`, `__onTestFilesChanged__`, and an `uploadAttemptCount` state whose
  comment admits it exists solely to drive the hook), and the test itself is concatenated `R"js(...)"` + `#if` + `BytesToJsArray()` through `EvalJs` — hard to read, one failure aborts everything. None of that is necessary: direct calls into `page_handler->ProcessXxxFile(bytes, name, future.GetCallback())` cover each case in five lines with proper per-case granularity.

  Bring back the deleted per-handler browsertests, add back the edge cases the old tests had and drop the e2e test along with the production-side test hooks.

components/ai_chat/resources/page/state/conversation_context.tsx
```diff
@@ -392,18 +393,24 @@ export function useProvideConversationContext(props: ConversationContextProps) {
...
+
     // Show error when some files were dropped: either text extraction
     // failed, or unsupported files were included (the backend returns
     // empty stubs for unsupported types like zip so they are filtered
     // out here, while cancellation returns null and skips this path).
-    if (validFiles.length < files.length) {
+    if (validFiles.length < uploadedFiles.length) {
       showAlert({
         type: 'error',
         content: getLocale(S.CHAT_UI_FILE_UPLOAD_ERROR),
```
darkdh, line 415: We missed testing partial failures showing error on multi-select which is not introduced by this migration so not necessary needs to be done in this PR

darkdh (commented):
production code migration is better now but still missed one pdf validation
And there are still some works to do on not losing original test coverage we had

components/ai_chat/resources/page/components/attachment_button_menu/attachment_button_menu.test.tsx
fallaciousreasoning: At build time, via `ifdef-loader.ts`

darkdh: please elaborate, is this leverage some upstream tool or we have our custom flow?

components/ai_chat/resources/page/state/conversation_context.tsx
fallaciousreasoning: Why does @brave/sec-team need to approve this? We regularly add test only functions and hooks (see `ForTesting` in C++).

I think an e2e test makes sense for this PR - its where we found we didn't have any tests checking that attaching a file runs through the processing logic (and I was bitten by it here).

We have a number of e2e tests which exist in browsertest as they are, literally, browser tests:
browser/ui/webui/brave_new_tab_page_refresh/ntp_ai_chat_browsertest.cc
browser/ui/webui/brave_new_tab_page_refresh/brave_new_tab_page_ui_browsertest.cc
browser/ui/webui/brave_wallet/wallet_panel_ui_browsertest.cc
browser/ai_chat/ai_chat_agent_profile_browsertest.cc
browser/brave_rewards/test/util/rewards_browsertest_context_util.cc

I can separate this into another file if you prefer, but I think its worth having and doing in this PR.

components/ai_chat/resources/page/components/attachment_button_menu/index.tsx
fallaciousreasoning: The things you learn! Apparently `capture=camera` has been deprecated as of HTML 5.1
https://stackoverflow.com/a/49615759/3260044

fallaciousreasoning: Do we have a preference as to which one should be used? I guess environment?

Perhaps it makes sense to let it use the default behavior anyway?

components/ai_chat/resources/page/state/conversation_context.tsx
darkdh: > Why does @brave/sec-team need to approve this? We regularly add test only functions and hooks (see ForTesting in C++).

Because you can't just export objects through window.* to the public.

> I can separate this into another file if you prefer, but I think its worth having and doing in this PR.

Then you add it to a new file and don't delete the original `Process*` test cases in that browser tests.

components/ai_chat/resources/page/components/attachment_button_menu/attachment_button_menu.test.tsx
fallaciousreasoning: Its a webpack plugin we've created. Its analogous to the GRIT preprocessing upstream does on their Web files.

`components/webpack/plugins/README.md`

components/ai_chat/resources/page/components/attachment_button_menu/index.tsx
darkdh: Let's use `capture="environment"`.

The old `SetUseMediaCapture(true)` was just a boolean — Blink reduces `<input capture>` to a boolean too: it only checks attribute presence (`input.FastHasAttribute(html_names::kCaptureAttr)` — see [file_input_type.cc;l=220-221](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/html/forms/file_input_type.cc;l=220-221;drc=e5a001f1eb7c4bc046a03168d1dcd778a69d62dc)) and `file_chooser.mojom` carries a `bool use_media_capture`. So `capture="camera"` works in Chromium by accident — value is ignored, only presence matters.

`environment` matches the old boolean-true behavior on Android (default rear camera), is honored correctly on iOS Safari, and matches the "take a picture of something" intent of this menu item. Dropping `capture` entirely would regress the UX — the picker would no longer offer the camera shortcut.

components/ai_chat/resources/page/state/conversation_context.tsx
darkdh: `*ForTesting` cpp functions are completely different than exporting a window object in js and then later call in c++ broweser tests. Those functions often have `CHECK_IS_TEST` to prevent misused in production codes

There is only one from the tests you mentioned `browser/resources/settings/brave_wallet_page/add_wallet_network_dialog.ts` that exports objects but it is doing `window.testing.*` which is not unconditionally export (guarded by `shouldExposeElementsForTesting`)

fallaciousreasoning: This is running on a `chrome://leo-ai` so is not exposed to the public. Its available on the `window` object for that WebUI only. Arguing that exposing something on `window` is a security risk doesn't really make sense - anything able to call that function could as easily rewrite `Array.prototype.push` or replace the URL constructor.

fallaciousreasoning: I'll migrate back to the `Process*` tests as you seem quite opposed to this.

browser/ai_chat/upload_file_helper_unittest.cc
darkdh: Bring back `LooksLikePdf` validation — fast-path rejection of fake/too-small PDFs before spinning up the headless viewer. Same place the deleted `DetermineFileType` did it.
So it would be something like 
  ```c++
  void AIChatUIPageHandler::ProcessPdfFile(const std::vector<uint8_t>& file_data,
                                           const std::string& filename,
                                           ProcessPdfFileCallback callback) {
  #if BUILDFLAG(ENABLE_PDF)
    if (!printing::LooksLikePdf(file_data)) {
      std::move(callback).Run(nullptr);
      return;
    }
    ExtractAndProcessFile(std::make_unique<PdfTextExtractor>(), file_data,
                          FILE_PATH_LITERAL("pdf"), filename,
                          mojom::UploadedFileType::kPdf, std::move(callback));
  #else
     ... // remain the same
  #endif  // BUILDFLAG(ENABLE_PDF)
  }
  ```

components/ai_chat/resources/page/state/conversation_context.tsx
diracdeltas: for future reference we have guidelines on exposing new window APIs: https://github.com/brave/internal/wiki/Pull-request-security-audit-checklist#window-api-injections

browser/ui/views/side_panel/ai_chat/ai_chat_side_panel_web_view.cc
```diff
@@ -137,6 +138,18 @@ content::WebContents* AIChatSidePanelWebView::AddNewContents(
...
+  if (browser_view) {
+    static_cast<content::WebContentsDelegate*>(browser_view->browser())
```
netzenbot, line 148: When `browser_view` is null, the `listener` is silently dropped without calling either `FileSelected()` or `FileSelectionCanceled()`. Per the `content::FileSelectListener` interface contract (see `file_select_listener.h`), one of these must always be called — in debug builds `FileChooserImpl::FileSelectListenerImpl` DCHECKs this. Add `listener->FileSelectionCanceled()` in the else branch. ([Callback Execution Must Be Consistently Sync or Async](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-060))

browser/ui/webui/ai_chat/ai_chat_ui_page_handler_browsertest.cc
```diff
@@ -256,191 +273,150 @@ IN_PROC_BROWSER_TEST_F(AIChatUIPageHandlerBrowserTest,
...
-  ASSERT_TRUE(result.has_value());
-  ASSERT_EQ(result->size(), 1u);
-  EXPECT_EQ((*result)[0]->filename, "dummy.txt");
-  ASSERT_TRUE((*result)[0]->extracted_text.has_value());
-  EXPECT_EQ(*(*result)[0]->extracted_text, kExpectedTextContent);
+  auto result = content::EvalJs(ai_chat_contents, script);
+  ASSERT_TRUE(result.is_ok()) << result;
+  EXPECT_EQ(result.ExtractString(), "success");
 }
-#endif  // !BUILDFLAG(IS_ANDROID)
```
netzenbot, line 339: The `content::EvalJs(ai_chat_contents, script)` call runs in the main world. Per [TJ-003](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-javascript.md#TJ-003), test JavaScript should use `ISOLATED_WORLD_ID_BRAVE_INTERNAL` to avoid interfering with page scripts. Note: this test intentionally communicates through `window.__onTestProcessFilesReady__` and `window.__onTestFilesChanged__`, which requires main-world access — but that design itself is part of what makes this test approach problematic (production code carries test hooks). The fix is to drop this e2e test and restore the per-handler `ProcessXxxFile` browsertests as the reviewer has requested.

components/ai_chat/resources/page/components/attachment_button_menu/attachment_button_menu.test.tsx
```diff
@@ -0,0 +1,77 @@
+// Copyright (c) 2025 The Brave Authors. All rights reserved.
```
netzenbot: nit: This is a new file but the copyright year is 2025; the current year is 2026. Per [CS-012](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-012), new files must use the current year in the copyright header.

components/ai_chat/resources/page/components/attachment_button_menu/attachment_button_menu.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+  setAttachmentsDialog: jest.fn(),
+  associateDefaultContent: undefined,
+  unassociatedTabs: [],
+  isMobile: false,
+  conversationStarted: false,
+}
+
+// Mirrors the accept constant in the component — must match its preprocessor block exactly.
+const expectedAccept: string | undefined =
+  // <if expr="is_android">
```
netzenbot, line 32: nit: `expectedAccept` duplicates the `accept` constant from `index.tsx` — if the production value changes, this test silently drifts out of sync. Per [TI-009](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-009), export `accept` from `index.tsx` and import it here instead of re-defining the preprocessor block.

fallaciousreasoning: The test won't silently drift out of sync, it will fail. This is intended as a check for when the Android accept is updated.

browser/ui/webui/ai_chat/ai_chat_ui_page_handler_browsertest.cc
fallaciousreasoning: no longer using the end to end test.

components/ai_chat/resources/page/state/conversation_context.tsx
fallaciousreasoning: Bit of a moot point now I've removed this, but I think that document is referring to adding new browser APIs available to web pages, not adding a field inside a Brave internal page.

components/ai_chat/resources/page/state/conversation_context.tsx
fallaciousreasoning: Let's leave this as a follow up :smile:

darkdh: yeah, please open an issue for tighten up UI flow coverage

browser/ui/webui/ai_chat/ai_chat_ui_page_handler_browsertest.cc
```diff
@@ -263,45 +265,88 @@ IN_PROC_BROWSER_TEST_F(AIChatUIPageHandlerBrowserTest, ProcessImageFile) {
...
+  }
+
+  // LooksLikePdf early exit: sufficient size but wrong header.
+  {
+    std::vector<uint8_t> wrong_header(50, 'x');
+    base::test::TestFuture<ai_chat::mojom::UploadedFilePtr> future;
+    page_handler->ProcessPdfFile(wrong_header, "fake.pdf",
+                                 future.GetCallback());
+    EXPECT_FALSE(future.Take());
+  }
```
darkdh: These new added cases should be unit tests in `ai_chat_ui_page_handler_unittest.cc` — they don't need a real renderer. Image processing works in-process via `data_decoder::test::InProcessDataDecoder` (same pattern as the deleted `upload_file_helper_unittest.cc`), and the `LooksLikePdf` early-exit cases return null before touching `PdfTextExtractor`.

Only the extraction-based cases (`ProcessPdfFile` valid PDF, `ProcessTextFile`, `ProcessHtmlFile`) actually need browsertest because they spawn sandboxed `WebContents` for the PDF viewer / `view-source:`. Keep those here.

Browsertests cost ~1–3s each vs ~10ms for unit tests; the savings add up across CI, and unit-test failures are easier to diagnose without `WaitForLoadStop` / chrome:// flakiness.

browser/ui/webui/ai_chat/ai_chat_ui_page_handler_browsertest.cc
```diff
@@ -263,45 +265,88 @@ IN_PROC_BROWSER_TEST_F(AIChatUIPageHandlerBrowserTest, ProcessImageFile) {
...
 #if BUILDFLAG(ENABLE_PDF)
 IN_PROC_BROWSER_TEST_F(AIChatUIPageHandlerBrowserTest, ProcessPdfFile) {
   auto* page_handler = GetPageHandler(web_contents());
   ASSERT_TRUE(page_handler);
 
+  // LooksLikePdf early exit: data shorter than 50 bytes, even with valid
+  // PDF header.
+  {
+    std::vector<uint8_t> too_small(std::begin("%PDF-1.4"),
+                                   std::end("%PDF-1.4") - 1);
```
darkdh:
```
auto bytes = base::byte_span_from_cstring("%PDF-1.4")
std::vector<uint8_t> too_small(bytes.begin(), bytes.end());
```

components/ai_chat/resources/page/components/attachment_button_menu/attachment_button_menu.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+  'image/*,application/pdf'
+// <else>
+undefined
+// </if>
+
+describe('attachment button menu', () => {
+  beforeEach(() => {
+    mockPickFiles.mockResolvedValue([])
+  })
```
netzenbot, line 41: The test expects `capture: 'camera'` but `index.tsx` passes `capture: 'environment'` to `pickFiles`. These must match or the test will fail.

components/ai_chat/resources/page/components/attachment_button_menu/attachment_button_menu.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+      accept: expectedAccept,
+    })
+  })
+
+  it('camera menu item passes correct accept to pickFiles', async () => {
+    const { container } = render(
+      <AttachmentButtonMenu
+        {...defaultProps}
+        isMobile={true}
+      />,
```
netzenbot, line 62: The camera menu-item test asserts `capture: 'camera'`, but `index.tsx` now passes `capture: 'environment'` (updated following review discussion). This mismatch means the test will fail. Update the expectation to `'environment'`.

components/ai_chat/resources/page/components/attachment_button_menu/attachment_button_menu.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+const expectedAccept: string | undefined =
+  // <if expr="is_android">
+  'image/*,application/pdf'
+// <else>
+undefined
+// </if>
+
+describe('attachment button menu', () => {
+  beforeEach(() => {
+    mockPickFiles.mockResolvedValue([])
```
netzenbot, line 40: The camera menu-item assertion expects `capture: 'camera'` but `index.tsx` calls `pickFiles({ capture: 'environment', accept })`. These values don't match, so this test will always fail. Update the expectation to `capture: 'environment'`.

components/ai_chat/resources/page/components/attachment_button_menu/attachment_button_menu.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+    // With isMobile=true and no associated content, menu order is:
+    // [0] upload, [1] camera, [2] bookmarks, [3] history
+    const menuItems = container.querySelectorAll<HTMLElement>('leo-menu-item')
+    await act(async () => {
+      menuItems[1].click()
+    })
+
+    expect(mockPickFiles).toHaveBeenCalledWith({
+      capture: 'camera',
+      accept: expectedAccept,
```
netzenbot, line 74: The test expects `capture: 'camera'` but the production code in `index.tsx` now passes `capture: 'environment'` (updated per reviewer feedback). This mismatch means the test either fails or verifies the wrong behavior. Update the expected value to `'environment'` to match the actual production code. ([TI-012](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-012))

browser/ai_chat/BUILD.gn
```diff
@@ -33,8 +33,6 @@ static_library("ai_chat") {
     "tab_tracker_service_factory.h",
     "tools/code_execution_tool.cc",
     "tools/code_execution_tool.h",
-    "upload_file_helper.cc",
-    "upload_file_helper.h",
```
darkdh, line 39: With these files removed (including the unit test), several deps in this target are now unused. 

**`static_library("ai_chat")`** — six unused, all came in via the deleted `upload_file_helper.cc`:
- `//ui/shell_dialogs` (`SelectFileDialog` / `SelectFilePolicy`)
- `//services/data_decoder/public/cpp` (`DataDecoder`, `decode_image` — moved to `browser/ui/webui/ai_chat`)
- `//printing:printing_base` (`printing/printing_utils.h::LooksLikePdf` — moved)
- `//net` (`net/base/mime_util.h`)
- `//third_party/blink/public/common` (`blink::IsSupportedImageMimeType`)
- `//ui/base`

**`source_set("unit_tests")`**
- `//chrome/browser/ui/select_file_policy` (was for `ChromeSelectFilePolicy` in deleted `upload_file_helper_unittest.cc`)

components/ai_chat/resources/page/state/conversation_context.tsx
fallaciousreasoning: https://github.com/brave/brave-browser/issues/55106

components/ai_chat/resources/page/components/attachment_button_menu/attachment_button_menu.test.tsx
fallaciousreasoning: resolved all but one of these as they're duplicates

browser/ui/webui/ai_chat/ai_chat_ui_page_handler_browsertest.cc
fallaciousreasoning: I moved these here originally because there were already test cases for Images and PDFs here - The 1 to 3 seconds is for test start up, and adding another case to the inside of an existing browser test isn't going to slow things down.

I'ved move the `ProcessImageFile` test into the unittest (that test was not introduced in this PR, it just had cases added to it), and I've split the `LooksLikePdf` tests out into their own unittest. I'd argue that it makes sense to keep all the Pdf checks in the same place but I don't have strong feelings about it.

browser/ui/webui/ai_chat/ai_chat_ui_page_handler_browsertest.cc
fallaciousreasoning: not sure why Claude decided to get the data like that, its pretty odd.

browser/ui/webui/ai_chat/ai_chat_ui_page_handler_browsertest.cc
darkdh: anything unit test can test should belong in unit tests. Those cases in browser tests were a mistake.
You can either move new one or existing cases all together.

browser/ai_chat/BUILD.gn
fallaciousreasoning: Nice catch - completely forgot to clean up the deps

browser/ui/webui/ai_chat/ai_chat_ui_page_handler_browsertest.cc
darkdh: https://bravesoftware.slack.com/archives/C7VLGSR55/p1749156496316139

darkdh (approved):
lgtm now and nice cleanup

kylehickinson (approved):
ios++ 

tested on device, all paths work well now, only suggestion for UI improvement would be to unify the "upload file…" and "take a picture" options as they show the same menu now

components/ai_chat/resources/page/state/conversation_context.tsx
```diff
@@ -439,43 +492,15 @@ export function useProvideConversationContext(props: ConversationContextProps) {
...
-  const isAttachedFilesProcessing =
-    aiChat.api.useCurrentOnUploadFilesSelected().hasEmitted
-
   // Is uploading is a combination of if files are being retrieved
   // or if chosen files are being processed.
-  const isUploadingFiles =
-    screenshotsMutation.isPending
-    || uploadFileMutation.isPending
-    || isAttachedFilesProcessing
+  const isUploadingFiles = screenshotsMutation.isPending
```
petemill, line 504: I think the problem with this change is that it's stops the progress indicator showing when files are being processed. I think we can bring back using the mutation hook instead of the direct mojo interface call. I'll submit it.

Logged https://github.com/brave/brave-browser/issues/55694

PR #35817, opened 2026-04-23
Use upstream function for getting P3A locale-based country code

### LINKED ISSUE #54888: Use upstream function for getting P3A locale-based country code
The current default engine and default engine switch metrics are configured to report a country code extracted from the user's locale. However, the function used for extracting the country code is different from the one used for determining the user's initial default engine, leading to misleading metric results when filtering by new users.

Update the locale-based country code extraction to use the same function that upstream uses to determine the initial default engine.

DJAndries:
Resolves https://github.com/brave/brave-browser/issues/54888


netzenbot (approved):

components/p3a/p3a_message.cc
```diff
@@ -307,7 +307,8 @@ void MessageMetainfo::Init(PrefService* local_state,
 
   country_code_from_timezone_raw_ =
       base::ToUpperASCII(base::CountryCodeForCurrentTimezone());
-  country_code_from_locale_raw_ = brave_l10n::GetDefaultISOCountryCodeString();
+  country_code_from_locale_raw_ =
+      std::string(country_codes::GetCurrentCountryID().CountryCode());
```
Olipro, line 311: I'm curious why you had to explicitly declare the `std::string` prvalue around it.

DJAndries: to convert the string_view to a string

Olipro (approved):
LGTM

PR #35818, opened 2026-04-23
[toolchain] Clean up Rust WASM script initialisation

cdesouza-chromium:
This PR does a series of clean ups to this script on the initialisation
process, by removing complex env and const initialisation from the
global scope, and keeping them closer to where they are used, and behind
the conditionals that guard their use.

This change also removes all code from `ToolchainBuilder` that is not
strictly involved with field value initialisations, so we we are left
with guarantees that all member data fields are initialised by the time
they are supposed to be used.

There are removal for unnecessary checks for `mkdir` operations that
will fail anyway if they cannot perform, as well as for checks if a
directory is a file, which can be avoided by checking an expected file
under the directory.

Finally, this change corrects our use of `depot_tools`, by making sure
we deploy a copy of it as a sibling path to `src`, rather than using the
one under `src/third_party` as that one should be left pinned to a
particular version chose by Chromium, and using its `gclient`
innevitably leads to it self-updating.

### DIFF
--- a/tools/rust/build_rust_toolchain_standalone.py
+++ b/tools/rust/build_rust_toolchain_standalone.py
@@ -72,6 +72,7 @@ import subprocess
 import sys
 import tarfile
 from types import ModuleType
+from typing import Optional
 
 # Filename of the LLVM linker binary produced by the Chromium LLVM build.
 LLD = 'lld' + ('.exe' if sys.platform == 'win32' else '')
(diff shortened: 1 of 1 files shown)


tools/rust/build_rust_toolchain_standalone.py
```diff
@@ -545,11 +494,58 @@ def run(self, use_ref: str = None):
...
+            clone_chromium: Whether the builder should operate in a mode that
+                allows cloning Chromium automatically, if necessary.
         """
+        self._bootstrap_depot_tools()
+
+        if not self._has_valid_chromium_path():
+            if clone_chromium:
+                logging.info('Chromium src not found at %s, cloning...',
+                             self.chromium_src)
+                self._clone_chromium()
```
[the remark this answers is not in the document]
cdesouza-chromium: Fixed.

tools/rust/build_rust_toolchain_standalone.py
```diff
@@ -192,70 +180,42 @@ class ToolchainBuilder:
...
             chromium_src: Path to the Chromium `src/` directory.  Must exist
                 and contain a `tools/rust/` subdirectory.
             out_dir: Directory where the output `.tar.xz` archive is written.
                 Created (including parents) if it does not already exist.
-            clone_chromium: Whether the builder should operate in a mode that
-                allows cloning Chromium automatically, if necessary.
 
         Raises:
             RuntimeError: If `chromium_src` does not exist or does not look
                 like a Chromium src tree, or if `out_dir` exists but is not a
```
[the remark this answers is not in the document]
cdesouza-chromium: Fixed

tools/rust/build_rust_toolchain_standalone.py
```diff
@@ -594,20 +585,28 @@ def main():
     logging.basicConfig(level=logging.DEBUG if args.verbose else logging.INFO)
 
     if args.with_git_cache:
-        if not GIT_CACHE_PATH.is_dir():
+        # This mirrors a bit what is being done in our infra, where the cache
+        #  is usually baked under the home directory as `cache/`.
+        if sys.platform == 'win32':
+            git_cache_path = Path(
+                os.environ.get('USERPROFILE', str(Path.home()))) / 'cache'
```
goodov: this doesn't look right. The git cache path can be different on CI and in fact it should already be set by the pipeline environment.

my point about git cache in slack discussion was about ensuring that we use git cache. The way it should be used is by setting it as CI environment as we have it for all other pipelines.

hardcoding these paths here is a no-no.

goodov: To cleanup this:
1. convert this flag to accept a string and override GIT_CACHE_PATH env var if the flag is passed
2. set GIT_CACHE_PATH on CI via environment or pass it via the flag.

tools/rust/build_rust_toolchain_standalone.py
```diff
@@ -545,11 +486,61 @@ def run(self, use_ref: str = None):
             use_ref: Optional Git reference (branch, tag, or commit) to check
                 out before building. If provided, calls `_checkout_chromium_ref`
                 first.
+            clone_chromium: Whether the builder should operate in a mode that
+                allows cloning Chromium automatically, if necessary.
         """
```
netzenbot, line 501: nit: `run()` can raise `RuntimeError` in several paths (invalid chromium src, missing `tools/rust/` dir, missing `sh.exe` on Windows) but the docstring has no `Raises:` section. The old `__init__` documented these — now that validation moved here, the contract should be documented here too. https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-004

tools/rust/build_rust_toolchain_standalone.py
cdesouza-chromium: To be very clear: **we are NOT using this flag in CI, and I do not plan to use this flag ever in the pipeline**.

About the way I'm assuming the value of path, this is how the pipeline sets the CI value for `GIT_CACHE_PATH`.

```
def getGitCachePath() {
    if (isWin()) {
        return "${USERPROFILE}\\cache"
    } else {
        return "${HOME}/cache"
    }
}
```

I added this flag merely for when I want to spin up a windows node, and remote connect to it for debugging problems. In those cases, I do use `--with-git-cache`, and I assume whatever we are currently using in the method above that CI uses. I'm not particularly concerned about this flag, because it is only a convenience during debugging.

cdesouza-chromium: > 1. convert this flag to accept a string and override GIT_CACHE_PATH env var if the flag is passed

I have done this though, because it makes sense, and it doesn't hurt, but I default it to `<home>/cache` if no value is provided.

netzenbot (approved):

tools/rust/build_rust_toolchain_standalone.py
goodov: > we are NOT using this flag in CI, and I do not plan to use this flag ever in the pipeline

ah, that's good then! I did assume it was used on CI, my bad.

> def getGitCachePath() {
>     if (isWin()) {
>         return "${USERPROFILE}\\cache"
>     } else {
>         return "${HOME}/cache"
>     }
> }

the thing with this is that this cache precalc on jenkins may be replaced with workspace-relative cache dir, so these defaults in the script may become obsolete and only make sense in a very special case as you described. I think it's fine though.

goodov (approved):

PR #35821, opened 2026-04-23
Workspaces - phase 1

bsclifton:
Closed in favor of https://github.com/brave/brave-core/pull/36059

## Description of the feature
The code here will allow the user to save a set of windows and tabs as a "workspace". At the moment, this save is "on demand", meaning that the workspace won't automatically save itself. Every time a person makes a change, they would need to manually save the workspace.  The windows and tabs are serialized as a list of Chromium [`sessions::SessionCommand`](https://source.chromium.org/chromium/chromium/src/+/main:components/sessions/core/session_command.h;l=40;drc=364c522b1ac0314bd967d423d6fc05f18e745f74) objects. 

Examples of what are saved:
- Windows (including position and size, and which window was active)
- Tab groups (including name, color, collapsed state)
- Tabs (including the navigation history)
- Pinned tabs

Once saved, a person could load that workspace on demand. Existing windows and tabs are not closed - but the "workspace" that was saved is loaded. Any windows and tabs contained in that workspace are restored by running the deserialized Chromium `SessionCommand` objects.

## Overview of the session code

When it's time to save a workspace, the code will create an empty list of  `SessionCommand` objects. The code can then iterate through each window to get the tabs.

All open windows are currently retrieved using `BrowserWindowInterface`. For each window (`Browser*`) found, a method I introduced called `AppendBrowserSessionCommands` is called. This will then start appending `SessionCommand` objects to the list.
- commands to restore the window size and position are appended
- tabs groups are then iterated and commands to restore them are appended
- the tabs themselves are iterated and commands to create them are appended
- the navigation history for the tabs is iterated and commands to restore are appended (referencing tab ID)
- selected tab selection is appended

You can see all of this logic in `AppendBrowserSessionCommands` (which takes the `Browser*`, one Chromium window).

After the `SessionCommand` objects are created for all given windows, the code will then append commands to restore the active window and then pass this list of serialized `SessionCommands` to a new `BraveWorkspaceService` keyed service.

Examples of what a `SessionCommand` are [can be found here](https://source.chromium.org/chromium/chromium/src/+/main:components/sessions/core/session_service_commands.h). 

The `BraveWorkspaceService` serializes the commands to disk using [`sessions::CommandStorageBackend`](https://source.chromium.org/chromium/chromium/src/+/main:components/sessions/core/command_storage_backend.h;l=56;drc=d345c21d2c52cb8754e1e3d896c57756ba74d3a8).

## Using the feature

1. Visit brave://flags/#brave-workspace to enable the feature
2. Once enabled, two new options will show up under Hamburger menu > `More tools`
    <kbd>image</kbd>
3. You can save your open tabs by picking `Save tabs as workspace...`
    - When you save it, the file is written to disk in your profile folder. 
    - For example, on Windows when building from source this would be `%LOCALAPPDATA%\BraveSoftware\Brave-Browser-Development\User Data\Default\workspaces`. 
    - A profile preference `kWorkspacesMetadataPref` for the given workspace name is created which has metadata as a `base::DictValue` (#tabs, #windows)
    - A folder is created with the name you specified.
    - Inside that is a folder `Sessions` which has the serialized `SessionCommand` objects. If you save the same workspace multiple times, it will have multiple files. Only the most recent one is loaded.
5. At any point that you want to load a workspace, you can pick `Open workspace...`
    - Data which was previously saved is loaded.
    - Workspace names are in the profile preference `kWorkspacesMetadataPref`.
    
The name of the workspace is currently hardcoded. The UI has been split out to a separate branch.


browser/ui/browser_commands.cc
```diff
@@ -1163,6 +1349,104 @@ void ForcePasteInWebContents(content::WebContents* web_contents) {
...
+  if (!service) {
+    return;
+  }
+
+  // Collect session commands on the UI thread, then write to disk on a
+  // background task (WriteWorkspaceToDisk does blocking file I/O).
+  std::vector<std::unique_ptr<sessions::SessionCommand>> commands;
+
+  auto browsers =
+      chrome::FindAllTabbedBrowsersWithProfile(calling_browser->profile());
```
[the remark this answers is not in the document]
bsclifton: Just as a heads up - I do realize this is bad and I do feel bad for using it 😓 

It was removed in Chromium 149 with https://source.chromium.org/chromium/chromium/src/+/4650246ccb09055bd278e075fb26055343174087

browser/ui/BUILD.gn
```diff
@@ -666,6 +667,10 @@ source_set("ui") {
       "views/view_shadow.h",
       "views/window_closing_confirm_dialog_view.cc",
       "views/window_closing_confirm_dialog_view.h",
+      "views/workspace/open_workspace_dialog.cc",
+      "views/workspace/open_workspace_dialog.h",
+      "views/workspace/save_workspace_dialog.cc",
+      "views/workspace/save_workspace_dialog.h",
```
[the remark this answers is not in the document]
bsclifton: addressed

browser/ui/BUILD.gn
```diff
@@ -666,6 +667,10 @@ source_set("ui") {
       "views/view_shadow.h",
       "views/window_closing_confirm_dialog_view.cc",
       "views/window_closing_confirm_dialog_view.h",
+      "views/workspace/open_workspace_dialog.cc",
+      "views/workspace/open_workspace_dialog.h",
+      "views/workspace/save_workspace_dialog.cc",
```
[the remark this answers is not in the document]
bsclifton: addressed

browser/ui/BUILD.gn
```diff
@@ -666,6 +667,10 @@ source_set("ui") {
       "views/view_shadow.h",
       "views/window_closing_confirm_dialog_view.cc",
       "views/window_closing_confirm_dialog_view.h",
+      "views/workspace/open_workspace_dialog.cc",
+      "views/workspace/open_workspace_dialog.h",
```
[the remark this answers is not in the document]
bsclifton: addressed

browser/ui/BUILD.gn
```diff
@@ -666,6 +667,10 @@ source_set("ui") {
       "views/view_shadow.h",
       "views/window_closing_confirm_dialog_view.cc",
       "views/window_closing_confirm_dialog_view.h",
+      "views/workspace/open_workspace_dialog.cc",
```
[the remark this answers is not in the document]
bsclifton: addressed

simonhong:
Can you write about what this PR does (architecture overview, UX and etc..)?
Having big picture before reading the code will be really helpful to understand the code. :)

simonhong:
First question is - why another storage (`info.json`)?
We have per-profile preference. I think workspace could be saved/restored there.
Is there any reason about this?

bsclifton:
@simonhong good question - it could definitely be a preference 🙂 I just didn't consider that at the time. If it's a preference, we don't need a lot of extra folders. The `workspaces` folder can just have all the serialized session command files.

I'll look at making this change 😄 

**edit: RESOLVED! 😄**

simonhong:
Second question is - Workspace really needs new KeyedService?
I think Workspace feature doesn't need runtime per-profile state.
It loads and stores workspace data on-demand. I think we can achieve more simply with core apis(or object). WDYT?

browser/ui/browser_commands.cc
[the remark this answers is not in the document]
bsclifton: Fixed

bsclifton:
> Second question is - Workspace really needs new KeyedService?
> I think Workspace feature doesn't need runtime per-profile state.
> It loads and stores workspace data on-demand. I think we can achieve more simply with core apis(or object). WDYT?

Good question - let me think about it more. I was originally thinking this needs to depend on other services. But it might not. I don't think it's too big of a deal to change if it's not needed

chromium_src/ui/views/window/dialog_delegate.h
```diff
@@ -36,6 +38,8 @@ class BraveVpnDnsSettingsNotificiationDialogView;
 #define CreatePassKey                                                 \
   CreatePassKey_Unused();                                             \
   friend class ::CrashReportPermissionAskDialogView;                  \
+  friend class ::OpenWorkspaceDialog;                                 \
+  friend class ::SaveWorkspaceDialog;                                 \
```
simonhong: We should not add another class here.
As chromium deprecated `DialogDelegateView`, using `DialogDelegate` + `View` is preferred way.

```
If you need a highly custom UI that DialogModel can't handle, you now split the implementation:

  1. Create a class that inherits from views::DialogDelegate.

  2. Create a separate views::View for your custom content.

  3. In your delegate, override GetContentsView() to return your custom view.
```

browser/ui/brave_browser_command_controller.cc
```diff
@@ -358,6 +360,17 @@ void BraveBrowserCommandController::InitBraveCommandState() {
...
+  UpdateCommandEnabled(
+      IDC_SAVE_WORKSPACE,
+      features::IsBraveWorkspaceEnabled() && browser_->is_type_normal() &&
+          BraveWorkspaceServiceFactory::GetForProfile(browser_->profile()) !=
+              nullptr);
+  UpdateCommandEnabled(
+      IDC_OPEN_WORKSPACE,
+      features::IsBraveWorkspaceEnabled() && browser_->is_type_normal() &&
+          BraveWorkspaceServiceFactory::GetForProfile(browser_->profile()) !=
+              nullptr);
```
simonhong: I think we can make `IDC_OPEN_WORKSPACE` disabled when workspace list is empty.

browser/ui/browser_commands.h
```diff
@@ -154,6 +154,11 @@ void ForcePasteInBrowser(Browser* browser);
 // Force pastes into the web contents if focused.
 void ForcePasteInWebContents(content::WebContents* contents);
 
+void ShowSaveWorkspaceDialog(Browser* browser);
+void ShowOpenWorkspaceDialog(Browser* browser);
+void SaveWorkspace(Browser* browser, const std::string& name);
+void RestoreWorkspace(Browser* browser, const std::string& name);
```
simonhong: Can we pass `BrowserWindowInterface*` here instead of `Browser*`?

browser/workspace/brave_workspace_service.cc
```diff
@@ -0,0 +1,196 @@
...
+const char kWorkspaceTabCount[] = "number-of-tabs";
+const char kWorkspaceModifiedAt[] = "modifed-at";
+
+}  // namespace
+
+// static
+void BraveWorkspaceService::RegisterProfilePrefs(
+    user_prefs::PrefRegistrySyncable* registry) {
+  registry->RegisterDictionaryPref(kWorkspacesMetadataPref);
+}
```
simonhong: I think we don't need this static method.
`BraveWorkspaceServiceFactory::RegisterProfilePrefs()` can call this directly.
 `registry->RegisterDictionaryPref(kWorkspacesMetadataPref);`

browser/workspace/brave_workspace_service.cc
```diff
@@ -0,0 +1,196 @@
...
+
+  // AppendCommands() is synchronous: it does file I/O directly on the calling
+  // thread.  Pass a null error callback so the backend never attempts to
+  // PostTask back to callback_task_runner_.
+  backend->AppendCommands(std::move(commands), /*truncate=*/true,
+                          base::OnceClosure());
+
+  // IsFileOpenForTesting() is the only public success indicator: the file
+  // stays open on success and is closed (open_file_ = null) on any I/O error.
+  bool success = backend->IsFileOpenForTesting();
```
simonhong: we can't use `XXXForTesting()` method from production code.

browser/ui/views/workspace/open_workspace_dialog.cc
```diff
@@ -0,0 +1,329 @@
...
+    auto* name_btn_raw = row->AddChildView(std::move(name_btn));
+    row_layout->SetFlexForView(name_btn_raw, 1);
+
+    auto del_btn = views::CreateVectorImageButtonWithNativeTheme(
+        base::BindRepeating(&OpenWorkspaceDialog::OnDeleteClicked,
+                            base::Unretained(this), static_cast<int>(i)),
+        kTrashCanIcon);
+    del_btn->SetTooltipText(
+        l10n_util::GetStringUTF16(IDS_WORKSPACE_OPEN_DIALOG_ROW_DELETE_BUTTON));
+    row->AddChildView(std::move(del_btn));
```
simonhong: I think these whole things can be done by `WorkspaceRowView`.
It's not common to set a View's layout manager from outside.
Adding child views to a View from outside is also not commonly used pattern.
If `workspaces_[i]` is passed to `WorkspaceRowView`'s ctor, it could do all by itself.

browser/ui/views/workspace/open_workspace_dialog.cc
```diff
@@ -0,0 +1,329 @@
...
+  UpdateRowSelection();
+}
+
+void OpenWorkspaceDialog::UpdateRowSelection() {
+  if (!list_container_) {
+    return;
+  }
+  const auto& children = list_container_->children();
+  for (size_t i = 0; i < children.size(); ++i) {
+    static_cast<WorkspaceRowView*>(children[i].get())
```
simonhong: we have `views::AsViewClass()`

browser/ui/views/workspace/open_workspace_dialog.h
```diff
@@ -0,0 +1,66 @@
...
+  void UpdateRowSelection();
+  void OnWorkspaceSelected(int index);
+  void OnAccept();
+  void OnDeleteClicked(int index);
+  void OnDeleteConfirmed(std::string name, chrome::MessageBoxResult result);
+  void OnDeleteCompleted(std::string name, bool success);
+
+  raw_ptr<Browser> browser_;
+  std::vector<WorkspaceInfo> workspaces_;
+  int selected_index_ = -1;
```
simonhong: if we use `optional<size_t>` for `selected_index_`, maybe we can avoid unnecessary `static_cast`?

browser/ui/browser_commands.h
simonhong: How about moving these methods' implementation to `BraveWorkspaceService()`?
Something like this for gathering workspace related logics all to its service.
```
void RestoreWorkspace(BrowserWindowInterface* bwi, const std::string& name) {
  auto* service =
      BraveWorkspaceServiceFactory::GetForProfile(bwi-GetProfile());
  if (!service) { 
    return;
  }

  service->RestoreWorkspace(name);
```

browser/workspace/BUILD.gn
```diff
@@ -0,0 +1,24 @@
+# Copyright (c) 2026 The Brave Authors. All rights reserved.
```
simonhong, line 1: If we can separate workspace logic in /components and /views/, it would be best for modern chromium modularization.

browser/ui/views/workspace/open_workspace_dialog.cc
```diff
@@ -0,0 +1,329 @@
...
+    if (selected_ == selected) {
+      return;
+    }
+    selected_ = selected;
+    UpdateBackground();
+  }
+
+  void OnMouseEntered(const ui::MouseEvent& event) override {
+    views::View::OnMouseEntered(event);
+    hovered_ = true;
```
simonhong: `View::IsMouseHovered()` is not sufficient?

browser/ui/views/workspace/open_workspace_dialog.cc
```diff
@@ -0,0 +1,329 @@
...
+constexpr int kDialogWidth = 420;
+constexpr int kListHeight = 240;
+constexpr int kPadding = 16;
+constexpr int kRowSpacing = 2;
+constexpr int kRowHeight = 36;
+
+// A workspace list row that highlights its background on hover and shows a
+// darker tint when selected.  SetNotifyEnterExitOnChild propagates mouse
+// enter/exit from child buttons up to this view so the whole row responds.
+class WorkspaceRowView : public views::View {
```
simonhong: It would be better to use separated files for this view.

browser/ui/views/workspace/open_workspace_dialog.cc
bsclifton: Fixed!

browser/ui/views/workspace/open_workspace_dialog.cc
bsclifton: Fixed!

browser/workspace/brave_workspace_service.cc
bsclifton: Fixed 😄

browser/ui/views/workspace/open_workspace_dialog.cc
bsclifton: Fixed!

browser/ui/browser_commands.h
bsclifton: Fixed! Used `Profile*` instead as `Browser*` not needed

bsclifton: This is a great one - I'll do this next!

bsclifton: Had some challenges here - raised in a DM

browser/ui/brave_browser_command_controller.cc
bsclifton: Fixed!

browser/workspace/BUILD.gn
bsclifton: I broke the UI apart from this PR - let me know if we should still do something here

chromium_src/ui/views/window/dialog_delegate.h
bsclifton: split out to another PR

browser/ui/views/workspace/open_workspace_dialog.cc
bsclifton: split out to another PR

browser/workspace/brave_workspace_service.cc
bsclifton: I didn't even notice factory was just calling static method 🤣 Good catch! Removed

PR #35732, opened 2026-04-21
Add enable_playlist buildflag gated on is_brave_origin_branded

bbondy:
Fix https://github.com/brave/brave-browser/issues/54129

- New `enable_playlist` GN buildflag, defaults to `!is_brave_origin_branded`.
- Follows the same pattern as `enable_brave_wallet` / `enable_ai_chat` — gates all Playlist code out of origin-branded desktop builds.
- `enable_playlist_webui` now chains through `enable_playlist`.
- Desktop-only: iOS and Android already assert `!is_brave_origin_branded`, so no guards on those platforms.

PR #35823, opened 2026-04-23
Gate Playlist UI on BravePlaylistEnabled policy

### LINKED ISSUE #54128: Add ability to disable playlist on upgrade case for Brave Origin

bbondy:
## Summary

Similar to #35732 which was for `is_brave_origin_branded=true` builds. This disables all playlist functionality for the admin policy on Desktop and also uses that policy for Brave Origin.

Fixes https://github.com/brave/brave-browser/issues/54128

## Test plan

- [x] Without the `BravePlaylistEnabled` policy set: Playlist sidebar item, side panel, toolbar action, URL-bar icon, and settings section are present (default behavior unchanged).
- [x] Set `BravePlaylistEnabled=false` via enterprise policy (or `chrome://policy`) before launch: all Playlist surfaces above are hidden, `chrome-untrusted://playlist` is unreachable.
- [x] With user pref toggled off via `brave://settings` (unmanaged): the sidebar item, toolbar action, and URL-bar icon remain reachable (no managed-policy short-circuit).
- [x] Origin-branded build still has no playlist
- [x] `brave_unit_tests --gtest_filter=*Sidebar*:*Playlist*` passes.
- [x] iOS behavior unchanged (iOS already uses `PrefService.isPlaylistAvailable`).

### Disabling Playlist via admin policy:

sudo /usr/libexec/PlistBuddy -c "Add :BravePlaylistEnabled bool false" /Library/Managed\ Preferences/com.brave.Browser.nightly.plist
sudo killall cfprefsd

### Reverting back to the normal state:

sudo /usr/libexec/PlistBuddy -c "Delete :BravePlaylistEnabled" /Library/Managed\ Preferences/com.brave.Browser.nightly.plist
sudo killall cfprefsd


mkarolin (approved):
`strings`++

browser/ui/webui/brave_settings_ui.cc
```diff
@@ -270,12 +270,16 @@ void BraveSettingsUI::AddResources(content::WebUIDataSource* html_source,
...
                                   brave_rewards::prefs::kDisabledByPolicy));
 #if BUILDFLAG(ENABLE_PLAYLIST)
+  const bool playlist_feature_enabled =
+      base::FeatureList::IsEnabled(playlist::features::kPlaylist);
   html_source->AddBoolean(
       "isPlaylistAllowed",
-      base::FeatureList::IsEnabled(playlist::features::kPlaylist) &&
+      playlist_feature_enabled &&
           profile->GetPrefs()->GetBoolean(playlist::kPlaylistEnabledPref));
+  html_source->AddBoolean("isPlaylistFeatureEnabled", playlist_feature_enabled);
```
bridiver: seems weird that we're passing the combination of feature and pref in `isPlaylistAllowed` (which doesn't sound right anyway because isn't kPlaylistEnabledPref a user toggle?) and then separately adding just the feature flag value. Shouldn't we just be accessing the pref directly in the settings UI so we can also detect the managed state and only pass in the feature flag because it's not otherwise available to settings?

bridiver: Also this logic is slightly different form policy::IsPlaylistEnabled because it doesn't include the managed state. I feel like maybe we need a PR before this to refactor a bit because the terminology seems a bit inconsistent. Here we have `isPlaylistAllowed` which maps to the pref && feature flag, but `playlist::IsPlaylistEnabled` maps to the pref && feature flag && managed state and that seems backwards. I would think allowed would map to managed pref state && feature flag and enabled would map to just the pref. 

Also here `isPlaylistEnabled` doesn't mean the same thing as `playlist::IsPlaylistEnabled`

components/playlist/core/browser/utils.cc
```diff
@@ -0,0 +1,31 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "brave/components/playlist/core/common/utils.h"
+
+#include "base/check.h"
+#include "base/feature_list.h"
+#include "brave/components/playlist/core/common/features.h"
+#include "brave/components/playlist/core/common/pref_names.h"
+#include "components/prefs/pref_service.h"
```
bridiver, line 12: this is a layering violation, PrefService can only be access in the browser process so this code belongs in browser, not common

components/playlist/core/browser/DEPS
```diff
@@ -0,0 +1,3 @@
+include_rules = [
+  "+components/prefs",
```
bridiver, line 2: see https://github.com/brave/brave-core/pull/35823/changes#r3134171824 this is not allowed because common can be access by any process

netzenbot (approved):

components/playlist/core/browser/utils.cc
bbondy: Moved to `//brave/components/playlist/core/browser target` and posted a best practice to avoid that in the future: //brave/components/playlist/core/browser target

components/playlist/core/browser/DEPS
bbondy: The file is now in `core/browser/DEPS`

browser/ui/webui/brave_settings_ui.cc
bbondy: Took a pass at this, let me know what you think.

zenparsing (approved):

bbondy:
if there are any follow ups @bridiver just comment and I'll do them in a future PR. Thanks.

PR #35826, opened 2026-04-23
[iOS] [AI Chat] Use Skia to process images pasted into the input field

### LINKED ISSUE #54899: Brave crashes when pasting an image into the input field in Leo WebUI
### Description

The app will crash if you paste an image directly into the input field instead of choose it through the file picker

### Steps to reproduce

1. Copy an image into your clipboard
2. Open Leo WebUI
3. Long press the input box
4. Tap Paste

### Actual result

Crashes

### Expected result

Should link the image to your prompt

### Reproduces how often

Easily reproduced

### Brave version

1.88+

### Device/iOS version

n/a

### Affected browser versions

- [x] latest AppStore
- [x] latest TestFlight
- [ ] previous TestFlight

### Reproducibility

### Miscellaneous information

This is caused because we cannot bind an `ImageDecoder` to `DataDecoder` on non-blink platforms, the `AIChatUIPageHandler::ProcessImageFile` implementation will have to be rewritten to avoid the use of `DataDecoderService`

kylehickinson:
This replaces the DataDecoder service which is not available to use and hits a fatal CHECK on iOS when not building with blink, fixing the crash when pasting images into the input field on Leo WebUI

Resolves https://github.com/brave/brave-browser/issues/54899

### DIFF
--- a/ios/browser/ui/webui/ai_chat/ai_chat_ui_page_handler.mm
+++ b/ios/browser/ui/webui/ai_chat/ai_chat_ui_page_handler.mm
@@ -50,9 +50,7 @@
 #include "ios/web/public/web_state_observer.h"
 #include "ios/web_view/internal/cwv_web_view_internal.h"
 #include "net/base/apple/url_conversions.h"
-#include "services/data_decoder/public/cpp/data_decoder.h"
-#include "services/data_decoder/public/mojom/data_decoder_service.mojom.h"
-#include "services/data_decoder/public/mojom/image_decoder.mojom.h"
+#include "skia/ext/skia_utils_ios.h"
 #include "third_party/skia/include/core/SkBitmap.h"
 #include "ui/base/l10n/l10n_util.h"
 #include "ui/base/page_transition_types.h"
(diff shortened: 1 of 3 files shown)


ios/browser/ui/webui/ai_chat/ai_chat_ui_page_handler.mm
```diff
@@ -80,41 +78,6 @@ void OnImageDecoded(
...
-  data_decoder::mojom::ImageDecoder* raw_decoder = decoder.get();
-  raw_decoder->DecodeImage(
-      encoded_bytes, codec, shrink_to_fit, max_size_in_bytes,
-      desired_image_frame_size,
-      base::IgnoreArgs<base::TimeDelta>(std::move(callback_pair.second)));
-}
-
 }  // namespace
 
 AIChatUIPageHandler::ChatContextObserver::ChatContextObserver(
```
netzenbot, line 83: nit: `bitmaps` is only read after initialization (`empty()` and `front()` are both const-qualified), so it could be `const std::vector<SkBitmap>`. Same applies to `ns_data` — the pointer is never reassigned. ([Prefer `const` for Local Variables Where Possible](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-064))

ios/browser/ui/webui/ai_chat/ai_chat_ui_page_handler.mm
```diff
@@ -194,24 +157,26 @@ void DecodeImageUsingServiceProcess(
...
     const std::string& filename,
     ProcessImageFileCallback callback) {
-  DecodeImageUsingServiceProcess(
-      &data_decoder_, file_data, data_decoder::mojom::ImageCodec::kDefault,
-      /*shrink_to_fit=*/true, kDataDecoderDefaultMaxSizeInBytes, gfx::Size(),
+  NSData* ns_data = [NSData dataWithBytes:file_data.data()
+                                   length:file_data.size()];
+  const std::vector<SkBitmap> bitmaps = skia::ImageDataToSkBitmaps(ns_data);
+  const SkBitmap decoded_bitmap =
+      bitmaps.empty() ? SkBitmap() : bitmaps.front();
```
nuo-xu: do we still need to decode when `bitmaps.empty()`?
`OnImageDecoded` immediately checks decoded_bitmap.drawsNothing(), and the default `SkBitmap()`'s `drawsNothing()` is probably always true.

nuo-xu (approved):

ios/browser/ui/webui/ai_chat/ai_chat_ui_page_handler.mm
kylehickinson: Hmm I suppose we could just run the callback early yeah

Uni-verse:
Verified on `iPhone 15` & `iPhone 12` running `iOS 18.6.2` & `iOS 26.3.1` respectively using version `1.91 (108)`

- ensured that pasting an image or multiple images in the chat field does not cause a crash. 
- ensured also images can be added using `Upload a file` method.

example | example
--|--
IMG_1336 | IMG_1337

PR #35832, opened 2026-04-23
[ads] Fix bugs in ObliviousHttpKeyConfig and ObliviousHttpClientImpl

### LINKED ISSUE #54903: [ads] Fix bugs in ObliviousHttpKeyConfig and ObliviousHttpClientImpl
When key rotation triggers Refetch, the cached key config is cleared so that requests fail fast rather than reaching the relay with an invalid config. However, any fetch callback still in flight from before the rotation is not cancelled, so it can fire after the clear and silently repopulate the config with the now-invalid key config. Requests made after rotation but before the fresh fetch completes then use the stale config and reach the relay with an invalid key.

Steps to Reproduce:

- Not reproducible manually. Key rotation is triggered server-side and cannot be initiated by a user or QA engineer.

tmancey:
Refetch cancels in-flight key config callbacks via InvalidateWeakPtrs before starting a new fetch so a stale response cannot overwrite the cleared config. OnCompleted uses an exhaustive switch on the mojo union tag with no default case. Unit tests cover both classes including happy and sad paths and the in-flight cancellation.

Resolves https://github.com/brave/brave-browser/issues/54903

### DIFF
--- a/components/brave_ads/core/browser/network/oblivious_http_client_impl_unittest.cc
+++ b/components/brave_ads/core/browser/network/oblivious_http_client_impl_unittest.cc
@@ -1,0 +1,118 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "brave/components/brave_ads/core/browser/network/oblivious_http_client_impl.h"
+
+#include <utility>
+
+#include "base/test/test_future.h"
+#include "brave/components/brave_ads/core/mojom/brave_ads.mojom.h"
+#include "net/base/net_errors.h"
+#include "net/http/http_response_headers.h"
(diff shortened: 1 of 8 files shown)


components/brave_ads/core/browser/network/oblivious_http_client_impl_unittest.cc
```diff
@@ -0,0 +1,118 @@
...
+#include "testing/gmock/include/gmock/gmock.h"
+#include "testing/gtest/include/gtest/gtest.h"
+#include "url/gurl.h"
+
+// npm run test -- brave_unit_tests --filter=BraveAds*
+
+namespace brave_ads {
+
+namespace {
+constexpr char kUrl[] = "https://example.com/resource";
```
aseren, line 25: std::string_view

aseren (approved):

components/brave_ads/core/browser/network/oblivious_http_client_impl_unittest.cc
tmancey: Because CI is nearly finished I will followup with std::string_view code health across several files.

PR #35833, opened 2026-04-23
Adblock intermittent test failures

bridiver:
Fixes intermittent adblock related test failures due to a race condition with adblock startup before attaching observers


cdesouza-chromium (approved):
LGTM

PR #35834, opened 2026-04-23
[ads][CodeHealth] Optimize SQLite indexes and queries across database tables

tmancey:
The reconcile query in the transactions database table had a dead
`OR creative_instance_id IN (?)` clause whose placeholder was never
bound, causing SQLite to evaluate it as NULL and never match any row.
Remove the clause. The PurgeExpired query in the ad_events database
table ran two separate DELETE statements that each evaluated the same
`creative_set_conversions` subquery; merge them into one statement so
the subquery runs once.

Drop the now-unused `creative_instance_id` index and the redundant `id`
index from the transactions table (PRIMARY KEY already implies one), and
drop the redundant `creative_instance_id` index from the deposits table
for the same reason. Replace the standalone `reconciled_at` index on
transactions with a composite `(reconciled_at, created_at)` index that
covers both columns filtered by PurgeExpired in a single scan. Bump the
schema to version 57 to apply these index changes to existing databases.

Resolves


components/brave_ads/core/internal/account/transactions/transactions_database_table.cc
```diff
@@ -191,8 +191,23 @@ void MigrateToV43(const mojom::DBTransactionInfoPtr& mojom_db_transaction) {
...
+  DropTableIndex(mojom_db_transaction,
+                 "transactions_creative_instance_id_index");
+
+  // Redundant with the implicit index on the PRIMARY KEY.
+  DropTableIndex(mojom_db_transaction, "transactions_id_index");
+
+  // Replaced with a composite index that also covers `created_at`, optimizing
+  // `PurgeExpired` which filters on both columns.
+  DropTableIndex(mojom_db_transaction, "transactions_reconciled_at_index");
   CreateTableIndex(mojom_db_transaction, /*table_name=*/"transactions",
```
aseren, line 211: MigrateToV43 now doesn't do:
```
  CreateTableIndex(mojom_db_transaction,                    
                    /*table_name=*/"transactions",
                   /*columns=*/{"creative_instance_id"});
```

components/brave_ads/core/internal/user_engagement/ad_events/ad_events_database_table.cc
```diff
@@ -572,28 +572,13 @@ void AdEvents::PurgeExpired(ResultCallback callback) const {
   mojom::DBTransactionInfoPtr mojom_db_transaction =
       mojom::DBTransactionInfo::New();
 
-  size_t days;
+  size_t days = UserHasJoinedBraveRewards() ? 90 : 30;
```
aseren: const size_t days

tmancey: it is assigned a few lines below.

components/brave_ads/core/internal/account/transactions/transactions_database_table.cc
tmancey: Great catch, fixed.

components/brave_ads/core/internal/user_engagement/ad_events/ad_events_database_table.cc
tmancey: Improved code to make it cleerer, thanks

aseren (approved):

PR #35846, opened 2026-04-23
Let review-prs skill come back after approval when re-requested

bbondy:
## Summary
- Once the review-prs skill approved a PR, the PR number was added to `_approved` in the local cache and `fetch-prs.py` permanently skipped it — even after new commits and an explicit re-request of the bot as a reviewer. This caused e.g. #35089 to never get re-reviewed after the initial approval.
- First commit lets an approved PR fall through when the bot is in `reviewRequests`, mirroring the existing reviewer-priority escape hatch used for the SHA-cache check below it.
- Second commit clears the stale `_approved` entry when the override fires on a *new* SHA, so `check-can-approve.py` (which also gates on the local `_approved` set) no longer blocks a fresh approval. Adds a `save_cache` helper and `cache_dirty` tracking so the clear persists.

Same-SHA re-requests still go to `skipped_approved` — nothing new to review, and re-approving the exact same commit would be a duplicate.

## Test plan
- [x] Run `/review-prs #35089 open auto reviewer-priority` — bot should now pick up the PR instead of skipping as `skipped_approved`.
- [x] Confirm on a clean re-review that a fresh APPROVE is submitted at the new SHA and `_approved` is re-populated in the cache.
- [x] Confirm that re-requesting the bot on a same-SHA approved PR still skips (no duplicate approval).

### DIFF
--- a/.claude/skills/review-prs/fetch-prs.py
+++ b/.claude/skills/review-prs/fetch-prs.py
@@ -191,6 +191,11 @@ def load_cache():
         return {}
 
 
+def save_cache(cache):
+    with open(CACHE_PATH, "w") as f:
+        json.dump(cache, f, indent=2)
+
+
 def load_org_members():
     """Load Brave org member logins from the cached file."""
     if not os.path.isfile(ORG_MEMBERS_PATH):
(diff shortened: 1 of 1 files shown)


netzenbot (approved):

bbondy:
This skill isn't actually used yet, the one from brave-dev-bot is used. But it exists because Mihai is in the process of moving the netzenbot reviews to be inside CI. It just syncs up to the skill there.

yrliou (approved):

PR #35847, opened 2026-04-23
Use liquid glass macOS app icons for Brave Origin

### LINKED ISSUE #54739: [Origin] For macOS 26, use the liquid glass app icons that were prepared.
### Description

I noticed that on macOS 26 we're still using the pre-liquid glass icons. We should be using the liquid glass ones. This applies to all releases (dev / nightly / beta / stable)

Here are the resources needed, it's an .icon file per platform
https://drive.google.com/drive/folders/1t5vgeJt61Xm7yMMdQWhUhy31REpkWTXX?usp=sharing

### Actual result

Image

### Expected result

Image

### Reproduces how often

Easily reproduced

-

### Channel information

- [x] release (stable)
- [x] beta
- [x] nightly

### Reproducibility

### Miscellaneous information

-

bbondy:
Resolves https://github.com/brave/brave-browser/issues/54739

## Summary

Compiles the Origin `.icon` packages provided by design into the channel `Assets.car` files so macOS 26 uses the liquid glass rendition while older macOS continues to use the bundled PNG fallbacks.

- Regenerated `app/theme/brave_origin/mac/{,beta/,nightly/}Assets.car` from the designer-provided `.icon` packages, running `src/tools/mac/icons/compile_car.py` (Xcode 26.3 / `actool` 24506) with the existing PNG bitmaps packaged as an `Assets.xcassets` fallback.
- Each new `.car` now contains the liquid glass iconstack (3 × AppIcon renditions — light/dark/tinted), the layer groups, vector SVG layers and color fills from the designer's `icon.json`, plus the 7 fallback PNG bitmaps (16/32/64/128/256/512/1024) for macOS < 26.
- The existing `app.icns` files are intentionally left untouched: post-Xcode-26.0 `actool` emits a minimal `.icns` with only 4 sizes, which would regress rendering on macOS < 26 vs. the existing 10-size bitmap set (see `src/tools/mac/icons/compile_car.py:98-115`).

## Scope

Applied to the Origin macOS channels that ship: release, beta, nightly. `development/` is left as-is (local dev builds only; no `.icon` source was provided for it), and Origin does not ship a `dev` channel.

## Test plan

- [x] Build Brave Origin Mac release/beta/nightly with `is_brave_origin_branded=true` and confirm the new liquid glass app icon appears on macOS 26.
- [x] Confirm the bitmap icon still renders correctly on macOS < 26 (icon in Dock, Finder, About dialog, LaunchServices).
- [x] `assetutil --info` spot-check that each shipped `.car` contains both an `AppIcon` iconstack rendition and the `Icon Image` PNG fallbacks.


SergeyZhukovsky (approved):
++

netzenbot (approved):

PR #35848, opened 2026-04-23
[ads] Add BraveAdsDisabled administrator policy and BraveOrigin registration

### LINKED ISSUE #54917: [ads] Allow administrators to disable Brave Ads independently of Brave Rewards via group policy across all profiles
Brave Ads currently relies on the Brave Rewards group policy to block the ads service, which means administrators cannot disable ads without also disabling Rewards or vice versa. This change introduces a dedicated `SponsoredAdsEnabled` group policy that maps to a `brave.brave_ads.sponsored.enabled` preference gating the sponsored ads directly, decouples ad service startup from Rewards policy, and registers the `brave.brave_ads.sponsored.enabled` preference with `BraveOriginService` so the policy is correctly enforced across all browser profiles.

`brave.brave_ads.sponsored.enabled` preference was introduced in https://github.com/brave/brave-browser/issues/57204

Done:
- [x] https://github.com/brave/brave-core/pull/35872

tmancey:
Adds a BraveAdsDisabled group policy and registers it with BraveOrigin
on desktop, Android, and iOS. The BraveOrigin registration disables
ads by default in that build, and the policy allows administrators to
disable ads independently of Brave Rewards. Previously the ads service
had no dedicated policy gate of its own. Adds browser tests for both
the enabled and disabled policy states.

Resolves https://github.com/brave/brave-browser/issues/54917


components/policy/resources/templates/policy_definitions/BraveSoftware/BraveAdsDisabled.yaml
```diff
@@ -0,0 +1,38 @@
...
+  value: true
+- caption: Allow the user to decide
+  value: null
+owners:
+- tmancey@brave.com
+- aseren@brave.com
+schema:
+  type: boolean
+supported_on:
+- ios:139-
```
aseren: is this supported on iOS?

components/policy/resources/templates/policy_definitions/BraveSoftware/BraveAdsDisabled.yaml
```diff
@@ -0,0 +1,38 @@
...
+- chrome.*:105-
+- android:142-
```
aseren: need to update with proper versions when policy is introduced

components/policy/resources/templates/policy_definitions/BraveSoftware/BraveAdsDisabled.yaml
tmancey: Fixed

components/policy/resources/templates/policy_definitions/BraveSoftware/BraveAdsDisabled.yaml
tmancey: Fixed

components/brave_ads/browser/ads_service_impl.cc
```diff
@@ -450,6 +460,20 @@ void AdsServiceImpl::InitializeBatAdsCallback(bool success) {
...
+}
+
+bool AdsServiceImpl::IsInitialized() const {
+  return is_bat_ads_initialized_;
+}
+
+void AdsServiceImpl::NotifyAdsServiceIneligibleToStart() {
+  was_marked_ineligible_to_start_ = true;
+
+  observers_.Notify(&AdsServiceObserver::OnAdsServiceIneligibleToStart);
```
netzenbot, line 474: nit: `NotifyAdsServiceIneligibleToStart` fires observers even when `was_marked_ineligible_to_start_` is already `true`. Since `MaybeStartBatAdsService` is called on every pref change, this can fire `OnAdsServiceIneligibleToStart` repeatedly for the same ineligibility state. Guard with `if (!was_marked_ineligible_to_start_)` before notifying. See https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-051

browser/brave_origin/brave_origin_service_factory.cc
```diff
@@ -125,6 +130,14 @@ constexpr auto kBraveOriginProfileMetadata =
...
+        // Brave Ads preferences
+        {brave_ads::prefs::kDisabledByPolicy,
+         BraveOriginServiceFactory::BraveOriginPrefMetadata(
+             true,
+             /*user_settable=*/false)},
+#endif
+
         // Brave Rewards preferences
         {brave_rewards::prefs::kDisabledByPolicy,
          BraveOriginServiceFactory::BraveOriginPrefMetadata(
```
netzenbot, line 143: nit: This `#if` block is >3 lines, so per the [#endif comment rule](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-045) it should have a comment: `#endif  // BUILDFLAG(ENABLE_BRAVE_ADS)`.

browser/brave_ads/test/ads_policy_browsertest_base.cc
```diff
@@ -0,0 +1,48 @@
...
+      &configuration_policy_provider_);
+  policy::PolicyMap policies;
+  policies.Set(policy::key::kBraveAdsDisabled, policy::POLICY_LEVEL_MANDATORY,
+               policy::POLICY_SCOPE_USER, policy::POLICY_SOURCE_PLATFORM,
+               base::Value(disable_ads),
+               /*external_data_fetcher=*/nullptr);
+  configuration_policy_provider_.UpdateChromePolicy(policies);
+}
+
+}  // namespace brave_ads
```
netzenbot, line 48: nit: `POLICY_LEVEL_MANDATORY`, `POLICY_SCOPE_USER`, and `POLICY_SOURCE_PLATFORM` are declared in `components/policy/core/common/policy_types.h`. That header is only reachable here transitively through `policy_map.h`, not included directly. Add `#include "components/policy/core/common/policy_types.h"` per [IWYU](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-001).

components/brave_ads/core/browser/service/ads_service.h
```diff
@@ -66,7 +66,13 @@ class AdsService : public KeyedService {
   void AddObserver(AdsServiceObserver* observer);
   void RemoveObserver(AdsServiceObserver* observer);
 
-  // Returns true if a browser upgrade is required to serve ads.
+  // Returns `true` if the service determined it was ineligible to start.
+  virtual bool IsIneligibleToStart() const = 0;
```
netzenbot, line 70: nit: `IsIneligibleToStart()` is a boolean method using the negative form. Per the [coding standards](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-003), prefer the positive form — e.g. `IsEligibleToStart()` returning `false` when ineligible, consistent with the existing `CanStartBatAdsService()` pattern.

PR #35859, opened 2026-04-23
[cr149][presubmit] Fix reading of gh paginated results

cdesouza-chromium:
The way this utiltiy was parsing the initial reply with all the comments
was not taking into account that paginated results break the pages per
line.


netzenbot (approved):

AlexeyBarabash (approved):
lgtm

PR #35862, opened 2026-04-24
Fix Playlist showing downloaded indicator for missing cached items

### LINKED ISSUE #54342: Playlist showing circled tick for items but when played it is not cached / downloaded and has to download again
### Description

Recently I've been having issues with a lot of my previously downloaded playlist items have somehow required to be all be redownloaded again. Even though they have a circle tick on the item (which I think means it is downloaded / cached versus a cloud icon and a down arrow), when I go to play it in CarPlay it causes an error because the media isn't there (see related #54341 ). In the iOS app on the phone, when I view the item there is a circled tick, but when I click to play, it prepares and downloads the item, and then starts playing.

I expect if there's a circled tick, then the item has already been downloaded and is cached, and when I go to play it will play immediately from the cache either in the iOS or CarPlay player and does not attempt to download/redownload.

### Steps to reproduce

N/A - could have just been an issue with my playlists, not sure how to reproduce

### Actual result

N/A

### Expected result

N/A

### Reproduces how often

Easily reproduced

### Brave version

iOS 1.88 (137)

### Device/iOS version

iOS 26.4

### Affected browser versions

- [x] latest AppStore
- [ ] latest TestFlight
- [ ] previous TestFlight

### Reproducibility

### Miscellaneous information

_No response_

EliHini:
Resolves https://github.com/brave/brave-browser/issues/54342

## Summary

On iOS, a Playlist item could display the "downloaded" checkmark while the underlying media file no longer existed on disk. Tapping the item would then silently fall back to streaming / re-downloading on the phone, and on CarPlay it would surface as an error.

1. `PlaylistItemList` rendered `.completed` whenever `PlaylistItem.cachedData` (a file bookmark) was non-empty, without verifying the file still existed. Even after the async `downloadState(for:)` lookup completed and returned `.invalid`, the closure fell through to the `cachedData` fallback and still returned `.completed` — so the tick persisted even once the authoritative check had contradicted it.
2. `deleteDanglingManagedAssets()` only cleaned up files-without-references. Items whose bookmarks were stale, or pointed to files iOS had reclaimed (HLS `.movpkg` cleanup, iCloud restore, storage pressure), were never cleared, so the tick persisted across cold starts.
3. `CarPlayController` read from synchronous `state(for:)` (which does verify file existence), while the phone list used the looser `cachedData` fallback. The two surfaces reported different offline states for the same item.

### Areas changed 
| Area | State | Source of truth |
| --- | --- | --- |
| Badge | Correct after wipe (no checkmark for missing file) | `downloadState(for:)` (`fileExists` check) |
| Context menu | "Save offline data" after wipe | Same `downloadState` dictionary as the badge |
| Player | Streams when there's no local file | `prepareToPlaySelectedItem` → `mediaStreamer` fallback |
| Re-tap recovery | Re-prepares when player is empty | `hasCurrentItem` check  |

## Changes

- **`PlaylistItemList.swift`** — derive the offline indicator exclusively from the async `PlaylistManager.downloadState(for:)` map. Drop the `cachedData` fallback.
- **`PlaylistManager.swift`**
  - `downloadState(for:)` now clears stale `cachedData` and emits `.invalid` when the bookmark resolves to a missing file, so the list refreshes promptly.
  - `deleteDanglingManagedAssets()` now reconciles the reverse direction as well — items whose bookmark is stale or whose file is missing have `cachedData` cleared and an `.invalid` state event fired. The reconciliation now also feeds the "valid paths" set used when deleting orphan files, so a single pass handles both directions.
- **`CarPlayController.swift`** — read from the same async `downloadState(for:)` path (cached per-refresh) so phone + CarPlay cannot disagree. `itemListTemplate(for:)`, the folder-list handler, and the `downloadStateChanged` / `objectWillChange` subscribers are now async-safe.

### Playing downloaded content offline
1. Navigate to a video on YouTube. Save a video to the playlist. Make sure that it is done downloading. 
2. Then go to another video and try saving it but before it is done turn off all network connections so that you have a partial download.
 - (Turn off wifi, Disable cellular data for Brave if testing with a physical device, you use the link conditioner). 
3. Restart the app (with no internet connection), go to the the playlist and observe that the fully downloaded video is what's in the playlist.
4. Attempt playing the downloaded video. Observe that it plays while the connection is still off.
5. Observe that you cannot play partially downloaded items.

### Test Stale or dangling data
1. Save a Playlist item for offline. Confirm the tick is present on the item.
2. Go to "All Settings" → "Playlist Debug" → "Manage Playlist Files".
3. Observe a list of files that represent the playlist items shown in the Playlist.
4. Delete all files, or swipe to delete a specific playlist file
5. Go back to the Playlist View in Brave
6. Observe that the delete playlist items do not have the tick items
7. Tap on any playlist item to play
8. Observe that it will begin to stream
9. Long press a playlist item and tap on "Save Offline Data "
10. Observe that it is downloading. Observe that when done, it has the tick mark to signify it has been saved for offline playback
11. Turn of all network connections and playback the item
12. Observe that it plays.

### DIFF
--- a/ios/brave-ios/Sources/Brave/Frontend/Browser/Playlist/Debug/PlaylistDebugFilesView.swift
+++ b/ios/brave-ios/Sources/Brave/Frontend/Browser/Playlist/Debug/PlaylistDebugFilesView.swift
@@ -1,0 +1,177 @@
+// Copyright 2026 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import SwiftUI
+
+struct PlaylistDebugFilesView: View {
+  @Environment(\.dismiss) private var dismiss
+  @State private var files: [URL] = []
+  @State private var errorMessage: String?
+  @State private var deleteError: String?
+  @State private var shouldDeleteAll = false
+  var body: some View {
+    NavigationStack {
+      Group {
+        if let errorMessage {
+          ContentUnavailableView(
+            "Couldn't load files",
+            systemImage: "exclamationmark.triangle",
+            description: Text(errorMessage)
+          )
+        } else if files.isEmpty {
+          ContentUnavailableView(
+            "No files",
+            systemImage: "tray",
+            description: Text("The Playlist folder is empty.")
+          )
+        } else {
+          List {
+            ForEach(files, id: \.self) { url in
+              HStack {
+                Image(systemName: "doc")
+                VStack(alignment: .leading) {
+                  Text(url.lastPathComponent)
+                  if let size = fileSize(url) {
+                    Text(size)
+                      .font(.caption)
+                      .foregroundStyle(.secondary)
+                  }
+                }
+              }
+              .swipeActions(edge: .trailing, allowsFullSwipe: true) {
+                Button(role: .destructive) {
+                  delete(url)
+                } label: {
+                  Label("Delete", systemImage: "trash")
+                }
+              }
+            }
+            .onDelete(perform: deleteAt)
+          }
+        }
+      }
+      .navigationTitle("Playlist File")
+      .toolbar {
+        ToolbarItem(placement: .cancellationAction) {
+          Button("Done") { dismiss() }
+        }
+        ToolbarItem(placement: .primaryAction) {
+          Button {
+            loadFiles()
+          } label: {
+            Image(systemName: "arrow.clockwise")
+          }
+        }
+
+        ToolbarItem(placement: .bottomBar) {
+          Spacer()
+        }
+        ToolbarItem(placement: .bottomBar) {
+          Button {
+            shouldDeleteAll = true
+          } label: {
+            Label("Delete All", systemImage: "trash")
+              .foregroundStyle(Color(braveSystemName: .systemfeedbackErrorIcon))
+          }
+          .confirmationDialog(
+            "Delete all files?",
+            isPresented: $shouldDeleteAll,
+            titleVisibility: .visible
+          ) {
+            Button("Delete \(files.count) File\(files.count == 1 ? "" : "s")", role: .destructive) {
+              deleteAll()
+            }
+            Button("Cancel", role: .cancel) {}
+          } message: {
+            Text(
+              "This will permanently remove every file in the Playlist folder. This action cannot be undone."
+            )
+          }
+        }
+      }
+      .onAppear(perform: loadFiles)
+    }
+  }
+
+  private func loadFiles() {
+    do {
+      let fm = FileManager.default
+      let appSupport = try fm.url(
+        for: .applicationSupportDirectory,
+        in: .userDomainMask,
+        appropriateFor: nil,
+        create: true
+      )
+      let playlistURL = appSupport.appendingPathComponent("Playlist", isDirectory: true)
+
+      if !fm.fileExists(atPath: playlistURL.path) {
+        try fm.createDirectory(at: playlistURL, withIntermediateDirectories: true)
+      }
+
+      let contents = try fm.contentsOfDirectory(
+        at: playlistURL,
+        includingPropertiesForKeys: [.isRegularFileKey, .fileSizeKey],
+        options: [.skipsHiddenFiles]
+      )
+
+      files = contents.sorted {
+        $0.lastPathComponent.localizedCompare($1.lastPathComponent) == .orderedAscending
+      }
+      errorMessage = nil
+    } catch {
+      errorMessage = error.localizedDescription
+      files = []
+    }
+  }
+
+  private func fileSize(_ url: URL) -> String? {
+    guard let size = try? url.resourceValues(forKeys: [.fileSizeKey]).fileSize else { return nil }
+    return ByteCountFormatter.string(fromByteCount: Int64(size), countStyle: .file)
+  }
+
+  private func delete(_ url: URL) {
+    do {
+      try FileManager.default.removeItem(at: url)
+      files.removeAll { $0 == url }
+    } catch {
+      deleteError = error.localizedDescription
+    }
+  }
+
+  private func deleteAt(_ offsets: IndexSet) {
+    for index in offsets {
+      let url = files[index]
+      do {
+        try FileManager.default.removeItem(at: url)
+      } catch {
+        deleteError = error.localizedDescription
+        return
+      }
+    }
+    files.remove(atOffsets: offsets)
+  }
+
+  private func deleteAll() {
+    do {
+      let fm = FileManager.default
+      let appSupport = try fm.url(
+        for: .applicationSupportDirectory,
+        in: .userDomainMask,
+        appropriateFor: nil,
+        create: false
+      )
+      let playlistURL = appSupport.appendingPathComponent("Playlist", isDirectory: true)
+      try fm.removeItem(at: playlistURL)
+      try fm.createDirectory(at: playlistURL, withIntermediateDirectories: true)
+      files = []
+    } catch {
+      deleteError = error.localizedDescription
+    }
+  }
+}
+
+#Preview {
+  PlaylistDebugFilesView()
+}
--- a/ios/brave-ios/Sources/Brave/Frontend/Browser/Playlist/Debug/PlaylistDebugFilesView.swift
+++ b/ios/brave-ios/Sources/Brave/Frontend/Browser/Playlist/Debug/PlaylistDebugFilesView.swift
@@ -52,7 +52,8 @@ struct PlaylistDebugFilesView: View {
           }
         }
       }
-      .navigationTitle("Playlist File")
+      .navigationTitle("Playlist Files")
+      .navigationBarTitleDisplayMode(.inline)
       .toolbar {
         ToolbarItem(placement: .cancellationAction) {
           Button("Done") { dismiss() }
@@ -92,6 +93,18 @@ struct PlaylistDebugFilesView: View {
         }
       }
       .onAppear(perform: loadFiles)
+      .alert(
+        "Couldn't delete file",
+        isPresented: Binding(
+          get: { deleteError != nil },
+          set: { if !$0 { deleteError = nil } }
+        ),
+        presenting: deleteError
+      ) { _ in
+        Button("OK", role: .cancel) {}
+      } message: { message in
+        Text(message)
+      }
     }
   }
 
@@ -172,6 +185,8 @@ struct PlaylistDebugFilesView: View {
   }
 }
 
+#if DEBUG
 #Preview {
   PlaylistDebugFilesView()
 }
+#endif
--- a/ios/brave-ios/Sources/Playlist/PlaylistManager.swift
+++ b/ios/brave-ios/Sources/Playlist/PlaylistManager.swift
@@ -259,10 +259,10 @@ public class PlaylistManager: NSObject {
       return .inProgress
     }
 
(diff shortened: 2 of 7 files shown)


ios/brave-ios/Sources/Playlist/PlaylistManager.swift
```diff
@@ -263,6 +263,7 @@ public class PlaylistManager: NSObject {
       if await AsyncFileManager.default.fileExists(atPath: assetUrl.path) {
         return .downloaded
       }
+      await invalidateCachedState(for: itemId)
```
nuo-xu: invalidate an item in a getter type of func. we either need to rename this func or put invalidation logic somewhere else. cant this be in `deleteDanglingManagedAssets()`?

ios/brave-ios/Sources/Playlist/PlaylistManager.swift
```diff
@@ -596,40 +597,69 @@ public class PlaylistManager: NSObject {
...
+    guard let playlistFolderPath = await PlaylistDownloadManager.playlistDirectory else {
+      return
+    }
+
+    let items = PlaylistItem.all().compactMap {
+      item -> (itemId: String, pageSrc: String, cachedData: Data)? in
+      guard let itemId = item.uuid, let cachedData = item.cachedData, !cachedData.isEmpty else {
+        return nil
+      }
+      return (itemId: itemId, pageSrc: item.pageSrc, cachedData: cachedData)
```
nuo-xu: tuples with more than 2 elements are bit hard to read. can be a `typealias`, a new struct or here can just keep the `PlaylistItem`

ios/brave-ios/Sources/Playlist/PlaylistManager.swift
```diff
@@ -596,40 +597,69 @@ public class PlaylistManager: NSObject {
...
+          validCachedPaths.insert(url.path)
         }
+        return (validCachedPaths, staleCachedItems)
+      }.value
+
+    for item in staleCachedItems {
+      cancelDownload(itemId: item.itemId)
+      PlaylistItem.updateCache(uuid: item.itemId, pageSrc: item.pageSrc, cachedData: nil)
+      onDownloadStateChanged(id: item.itemId, state: .invalid, displayName: nil, error: nil)
+    }
```
nuo-xu: same logic of `invalidateCachedState`?

ios/brave-ios/Sources/PlaylistUI/CarPlayController.swift
```diff
@@ -276,6 +287,18 @@ public class CarPlayController {
...
     return template
   }
+
+  @MainActor private func refreshDownloadStates(for items: [PlaylistItem]) async {
+    var itemIDs = Set<String>()
+    for item in items {
+      guard let uuid = item.uuid else { continue }
+      itemIDs.insert(uuid)
+      downloadStates[uuid] = await PlaylistManager.shared.downloadState(for: uuid)
+    }
```
nuo-xu: this is O(N) sequential awaits. Use withTaskGroup to parallelize

ios/brave-ios/Sources/PlaylistUI/PlaylistItemList.swift
```diff
@@ -55,19 +55,17 @@ struct PlaylistItemList: View {
...
-                if state == .inProgress {
-                  // PlaylistDownloadManager reports percent as 0...100 not 0...1
-                  let percentCompleted = downloadProgress[uuid, default: 0.0] / 100.0
-                  return .downloading(percentComplete: percentCompleted)
-                }
+              guard let uuid = item.uuid, let state = downloadStates[uuid] else {
+                return nil
               }
-              if let cachedData = item.cachedData, !cachedData.isEmpty {
+              if state == .downloaded {
```
nuo-xu, line 61: we dont need to check this condition anymore?

nuo-xu:
@EliHini how do i test this? what is the expected behaviour. the issue doesn't specify the `actual result` or `expected result`

netzenbot (approved):

ios/brave-ios/Sources/PlaylistUI/PlaylistItemList.swift
EliHini: no because the cache data can become stale.

EliHini:
> @EliHini how do i test this? what is the expected behaviour. the issue doesn't specify the `actual result` or `expected result`

I will add some test procedures to the description.

1. Essentially, save a video to the playlist. Make sure that it is done downloading. 
2. Then go to another video and try saving it but before it is done turn off all network connections  (so that you have a partial download). 
3. Restart the app (with no internet connection), go to the the playlist and observe that the fully downloaded video is what's in the playlist.
4. Attempt playing the downloaded video. Observe that it plays while the connection is still off.

ios/brave-ios/Sources/Playlist/PlaylistManager.swift
EliHini: Good call; moved the side effect out. `downloadState(for:)` is now a pure getter, and the only caller of `invalidateCachedState` is `deleteDanglingManagedAssets`, which is the right place for cleanup. Since the UI now derives the tick from `downloadState(for:)` instead of `cachedData`, having a stale `cachedData` row in CoreData isn't user-visible -- it just gets reaped on the next periodic reconciliation pass.

ios/brave-ios/Sources/Playlist/PlaylistManager.swift
EliHini: Done. The detached resolution work only needs `(itemId, cachedData)` now, since `invalidateCachedState` re-fetches the `PlaylistItem` to read `pageSrc`. The remaining 2-tuple felt fine; happy to bump it to a small struct if you'd still prefer.

ios/brave-ios/Sources/Playlist/PlaylistManager.swift
EliHini: Yep, collapsed. The loop is now just `invalidateCachedState(for: itemId)`. I also folded `cancelDownload` into `invalidateCachedState` so the two call sites stay in sync.

ios/brave-ios/Sources/PlaylistUI/CarPlayController.swift
EliHini: That's fair. Switched to `withTaskGroup` so the per-item `downloadState(for:)` calls run concurrently. As a bonus we no longer need the dangling-key sweep; replacing `downloadStates` wholesale with the fresh dictionary drops anything that's no longer in the folder.

ios/brave-ios/Sources/Brave/Frontend/Browser/Playlist/Debug/PlaylistDebugFilesView.swift
```diff
@@ -0,0 +1,177 @@
...
+      files = []
+    } catch {
+      deleteError = error.localizedDescription
+    }
+  }
+}
+
+#Preview {
+  PlaylistDebugFilesView()
+}
```
nuo-xu, line 191: put preview inside DEBUG macro

ios/brave-ios/Sources/PlaylistUI/PlaylistItemList.swift
```diff
@@ -80,13 +78,13 @@ struct PlaylistItemList: View {
           PlaylistManager.shared.getAssetDuration(item: .init(item: item)) { _ in }
         }
         .contextMenu {
-          if let cachedData = item.cachedData, !cachedData.isEmpty {
+          // Mirror the badge's source of truth: only offer "Remove offline data" when the cached file
+          // actually exists on disk. Otherwise we'd offer to remove a phantom download for an item whose bookmark
+          // survived but whose file was wiped (e.g. by iOS reclaiming `com.apple.UserManagedAssets*`).
+          if downloadStates[item.uuid ?? ""] == .downloaded {
```
nuo-xu: guard `item.uuid` to match the rest of file uses proper optional binding consistently

ios/brave-ios/Sources/Brave/Frontend/Browser/Playlist/Debug/PlaylistDebugFilesView.swift
```diff
@@ -0,0 +1,177 @@
...
+        in: .userDomainMask,
+        appropriateFor: nil,
+        create: false
+      )
+      let playlistURL = appSupport.appendingPathComponent("Playlist", isDirectory: true)
+      try fm.removeItem(at: playlistURL)
+      try fm.createDirectory(at: playlistURL, withIntermediateDirectories: true)
+      files = []
+    } catch {
+      deleteError = error.localizedDescription
```
nuo-xu, line 183: nit: error is set with no show

ios/brave-ios/Sources/Brave/Frontend/Browser/Playlist/Debug/PlaylistDebugFilesView.swift
```diff
@@ -0,0 +1,177 @@
...
+                } label: {
+                  Label("Delete", systemImage: "trash")
+                }
+              }
+            }
+            .onDelete(perform: deleteAt)
+          }
+        }
+      }
+      .navigationTitle("Playlist File")
```
nuo-xu: nit: Playlist Files

ios/brave-ios/Sources/Brave/Frontend/Browser/Playlist/Debug/PlaylistDebugFilesView.swift
```diff
@@ -0,0 +1,177 @@
...
+  private func loadFiles() {
+    do {
+      let fm = FileManager.default
+      let appSupport = try fm.url(
+        for: .applicationSupportDirectory,
+        in: .userDomainMask,
+        appropriateFor: nil,
+        create: true
+      )
+      let playlistURL = appSupport.appendingPathComponent("Playlist", isDirectory: true)
```
nuo-xu, line 120: can just use `await PlaylistDownloadManager.playlistDirectory`

ios/brave-ios/Sources/PlaylistUI/PlaylistItemList.swift
EliHini: guard doesn't work in `contextMenu` builder which is why I am using this pattern. But I will use `if let` to guard it

nuo-xu: that's fine. just use `if let uuid = item.uuid, downloadStates[uuid] == .downloaded`

ios/brave-ios/Sources/Brave/Frontend/Browser/Playlist/Debug/PlaylistDebugFilesView.swift
EliHini: Since this is a test harness, I prefer this to be standalone as much as possible so as not introduce module dependency from features under test

StephenHeaps (approved):
LGTM, please wait for @nuo-xu approval

ios/brave-ios/Sources/Brave/Frontend/Browser/Playlist/Debug/PlaylistDebugFilesView.swift
```diff
@@ -0,0 +1,191 @@
...
+                } label: {
+                  Label("Delete", systemImage: "trash")
+                }
+              }
+            }
+            .onDelete(perform: deleteAt)
+          }
+        }
+      }
+      .navigationTitle("Playlist Files")
```
StephenHeaps, line 55: nit: add `.navigationBarTitleDisplayMode(.inline)`

nuo-xu (approved):

PR #35873, opened 2026-04-24
[ads] Add BraveAdsEnabled administrator policy

tmancey:
Adds the BraveAdsEnabled policy that lets administrators enable or
disable Brave Ads via MDM. When set to false the policy forces all ad
types off and prevents users from overriding them individually.

Part of https://github.com/brave/brave-browser/issues/54917.

Resolves


ios/browser/policy/BUILD.gn
```diff
@@ -24,6 +25,9 @@ source_set("policy") {
     "//components/policy/core/browser",
     "//ios/chrome/browser/policy/model",
   ]
```
netzenbot, line 31: The `brave_simple_policy_map_ios.h` includes `brave/components/brave_ads/buildflags/buildflags.h` unconditionally (needed for the `#if BUILDFLAG(ENABLE_BRAVE_ADS)` check), but there's no unconditional `"//brave/components/brave_ads/buildflags"` in this target's `deps`. Only the conditional `headers` dep is added. Per [BS-027](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-027), the buildflags dep must be unconditional so `gn check` passes when `enable_brave_ads` is false. Add `"//brave/components/brave_ads/buildflags"` to the unconditional `deps` list.

browser/brave_ads/ads_disabled_by_policy_browsertest.cc
```diff
@@ -0,0 +1,24 @@
...
+    SetUpBraveAdsPolicy(/*disable_ads=*/true);
+  }
+};
+
+IN_PROC_BROWSER_TEST_F(BraveAdsAdsDisabledByPolicyTest, ServiceDoesNotStart) {
+  test::AdsServiceWaiter waiter(GetAdsService());
+  waiter.WaitForOnAdsServiceIneligibleToStart();
+}
+
+}  // namespace brave_ads
```
netzenbot: The test confirms the service is ineligible to start but never explicitly asserts `IsInitialized()` is false. Per TI-021, failure-path tests should verify side effects that should NOT occur. Consider adding `EXPECT_FALSE(GetAdsService().IsInitialized())` after the waiter call to make the intent clear and catch regressions where the service becomes ineligible but still initializes. ([Verify Negative Expectations in Tests](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-021))

browser/brave_ads/BUILD.gn
```diff
@@ -165,13 +165,27 @@ source_set("browser_tests") {
   deps = [
     ":brave_ads",
     ":impl",
+    "//brave/components/brave_ads/core/public:headers",
     "//chrome/browser",
+    "//chrome/browser/profiles:profile",
+    "//chrome/browser/ui",
     "//chrome/test:test_support",
```
netzenbot, line 174: nit: The policy-related deps (`//components/policy:generated`, `//components/policy/core/browser`, `//components/policy/core/common:test_support`) are added to the top-level `deps` but the sources that use them are only in the `else` block. These deps should move inside the `else` block alongside their source files to keep sources and deps grouped together. See [BS-005](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-005).

browser/brave_ads/ads_disabled_by_policy_browsertest.cc
```diff
@@ -0,0 +1,24 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "brave/browser/brave_ads/ads_service_waiter.h"
+#include "brave/browser/brave_ads/test/ads_policy_browsertest_base.h"
+#include "content/public/test/browser_test.h"
+
+namespace brave_ads {
+
+class BraveAdsAdsDisabledByPolicyTest
```
netzenbot: nit: The existing `BraveRewardsPolicyTest` and `BraveWalletPolicyTest` both use `WithParamInterface<bool>` + `INSTANTIATE_TEST_SUITE_P` to cover the enabled/disabled policy states in a single parameterized test. Two separate fixture classes across two files for what is essentially `disable_ads=true` vs `disable_ads=false` diverges from that pattern. Consider collapsing into one `BraveAdsPolicyBrowserTest : public PlatformBrowserTest, public ::testing::WithParamInterface<bool>` following the same style. ([Use Parameterized Tests for Similar Scenarios](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-023))

ios/browser/policy/BUILD.gn
tmancey: Fixed

browser/brave_ads/BUILD.gn
tmancey: Fixed.

browser/brave_ads/ads_disabled_by_policy_browsertest.cc
tmancey: Fixed.

browser/brave_ads/ads_disabled_by_policy_browsertest.cc
tmancey: Fixed.

netzenbot (approved):

components/brave_ads/core/public/prefs/pref_names.h
```diff
@@ -12,6 +12,10 @@ static_assert(BUILDFLAG(ENABLE_BRAVE_ADS));
 
 namespace brave_ads::prefs {
 
+// Policy prefs. Not prefixed with `brave.brave_ads` so they are not cleared
+// when the user resets ads data.
+inline constexpr char kDisabledByPolicy[] = "brave.ads.disabled_by_policy";
```
bbondy: 2 things about this:
- Prefs should be named enabled and not disabled. Working with double negatives can get confusing.
-  I believe this would be better named something like brave.ads.enabled. And then the pref can be managed by policy (or not). We do have a couple cases of prefs with `_by_policy` but we'd rather not have them.

components/policy/resources/templates/policy_definitions/BraveSoftware/BraveAdsDisabled.yaml
```diff
@@ -0,0 +1,38 @@
+caption: Disable Brave Ads
+default: null
+desc: |-
+  Disable Brave Ads in Brave.
```
bbondy: Reverse this please

components/brave_ads/core/public/prefs/pref_names.h
bridiver: I believe there's already a pref for ads so we don't need to add one? Just make the existing pref managed.

tmancey: @bridiver

`kOptedInToNotificationAds` ("brave.brave_ads.enabled") applies only to notification ads. However, the ads service can still start independently of this preference in several cases:

- when users have joined Brave Rewards (regardless of per-unit opt-ins),
- for search result ads,
- and for new tab page ads.

As a result, setting `kOptedInToNotificationAds` to `false` via policy does not fully prevent the service from starting. A separate `kEnabledByPolicy` preference is required to properly gate the entire service startup.

Additionally, because `kOptedInToNotificationAds` is scoped under `brave.brave_ads`, it is cleared when users reset their ads data, making it unsuitable for use as a policy preference.

components/brave_ads/core/public/prefs/pref_names.h
```diff
@@ -12,6 +12,10 @@ static_assert(BUILDFLAG(ENABLE_BRAVE_ADS));
 
 namespace brave_ads::prefs {
 
+// Policy pref. Not prefixed with `brave.brave_ads` so it is not cleared when
+// the user resets ads data.
+inline constexpr char kEnabledByPolicy[] = "brave.ads.enabled";
```
bbondy: Can we just name this kBraveAdsEnabled?
The pref itself does not mean it is enabled by policy, so the pref name is kind of wrong.

components/brave_ads/core/internal/prefs/obsolete_pref_util.cc
```diff
@@ -125,6 +126,14 @@ void MigrateObsoleteProfilePrefs(PrefService* const prefs) {
   // Added 04/2026.
   prefs->ClearPref(kObsoleteHasMigratedConfirmationStateV8);
   prefs->ClearPref(kObsoleteHasMigratedStateV2);
+
+  if (prefs->IsManagedPreference(brave_rewards::prefs::kDisabledByPolicy) &&
+      prefs->GetBoolean(brave_rewards::prefs::kDisabledByPolicy) &&
+      !prefs->HasPrefPath(prefs::kEnabledByPolicy)) {
+    // On first run, preserve ads-disabled state for admins who used
+    // `BraveRewardsDisabled` before `BraveAdsEnabled` existed.
+    prefs->SetBoolean(prefs::kEnabledByPolicy, false);
```
bbondy: These users can never re-enable ads unless they use the admin policy for ads is that correct?

components/policy/resources/templates/policy_definitions/BraveSoftware/BraveAdsEnabled.yaml
```diff
@@ -0,0 +1,38 @@
...
+  value: true
+- caption: Allow the user to decide
+  value: null
+owners:
+- tmancey@brave.com
+- aseren@brave.com
+schema:
+  type: boolean
+supported_on:
+- chrome.*:147-
```
aseren: this should match current Chromium version which is 148

tmancey: Ah yes the trains left yesterday :-)

components/brave_ads/core/public/prefs/pref_names.h
tmancey: Discussed with @bridiver and we can remove this pref and just use existing prefs for NTP, Search, and Notification Ads.

bridiver: as discussed in DM adding another pref on top of the three current prefs (notification ads, search ads and ntt ads) makes every check even more fragile, error prone and hard to understand (existing code that added `ByPolicy` prefs demonstrates this complexity). So there are two separate things here (over two PRs)
1. Make the existing 3 prefs managed instead of adding a 4th
2. Gate the "start" of the ads service on those 3 prefs whether they are managed or not. If they are all disabled the ads service should do nothing

tmancey: Great idea, I am away on PTO after tomorrow, but let me see if I can fix first thing tomorrow.

aseren:
Opened `Desktop/Android` policy change to a separate PR: https://github.com/brave/brave-core/pull/36595

PR #35875, opened 2026-04-24
Add cache_dir var in `build/commands`.

### LINKED ISSUE #54933: Support universal cache dir in `build/commands`
CI need to set a single cache dir var and brave-core need to put everything inside it.

goodov:
Add `cache_dir` env var (unset by default). When it's set, git, siso and vpython caches will all be placed inside.

Resolves https://github.com/brave/brave-browser/issues/54933

### DIFF
--- a/build/commands/lib/config.ts
+++ b/build/commands/lib/config.ts
@@ -66,6 +66,8 @@ export class Config {
   notary_password: string | undefined
   channel: string
   isBraveOriginBranded: boolean | undefined
+  // Parent cache directory for all internal caches (git, siso, vpython).
+  readonly cacheDir: string | undefined
   gitCachePath: string | undefined
   rbeService: string
   rbeTlsClientAuthCert: string | undefined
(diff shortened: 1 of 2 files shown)


build/commands/lib/config.ts
```diff
@@ -198,8 +200,11 @@ export class Config {
     this.isBraveOriginBranded = envConfig.getBoolean([
       'is_brave_origin_branded',
     ])
+    this.cacheDir = envConfig.getPath(['cache_dir'])
     this.gitCachePath =
-      envConfig.getPath(['git_cache_path']) || process.env.GIT_CACHE_PATH
+      envConfig.getPath(['git_cache_path'])
+      || this.resolveCacheDir('git')
+      || process.env.GIT_CACHE_PATH
```
atuchin-m, line 207: nit: should `process.env.GIT_CACHE_PATH` be the second, not the third?

atuchin-m (approved):

build/commands/lib/config.ts
goodov: the idea here is "config values prioritized", so `process.env` still goes last.

but tbh I'm not sure about the original impl here. To me `process.env` variables should actually be a priority and override the config, at least that's what I expect in general in other apps.

PR #35876, opened 2026-04-24
Added policy for Email Aliases.

### LINKED ISSUE #55150: Email Aliases shouldn't be enabled on Origin by default
We don't want to enable Email Aliases on Origin builds by default.
Users should be able to upgrade to it.

Email Aliases is disabled by default for now but will soon by enabled by default on Nightly.

boocmp:
Resolves https://github.com/brave/brave-browser/issues/55150


components/email_aliases/pref_names.h
```diff
@@ -8,6 +8,9 @@
 
 namespace email_aliases::prefs {
 
+inline constexpr char kEmailAliasesDisabledByPolicy[] =
+    "brave.email_aliases.disabled_by_policy";
```
bbondy: The prefs like this should just be named `brave.email_aliases.enabled`
and pref constant name `kEmailAliasesEnabled`. The pref itself has nothing to do with if it is managed or not. And ideally we don't want to have prefs that are only for managed things separate from normal disabling.

browser/email_aliases/email_aliases_service_factory.cc
```diff
@@ -45,6 +47,21 @@ EmailAliasesServiceFactory* EmailAliasesServiceFactory::GetInstance() {
   return instance.get();
 }
 
+// static
+bool EmailAliasesServiceFactory::IsServiceDisabled(
```
bbondy: Can we name things Enabled and not Disabled?

components/policy/resources/templates/policy_definitions/BraveSoftware/EmailAliasesDisabled.yaml
```diff
@@ -0,0 +1,36 @@
+caption: Disable Email Aliases
```
bbondy: This should be Enabled and not Disabled.

bbondy (changes requested):
There should also be a toggle on the Brave Origin settings page. Up to you but you could easily split this like:
PR 1: Build flag
PR 2: Admin policy
PR 3: Origin specific things

ios/browser/policy/brave_simple_policy_map_ios.h
```diff
@@ -79,6 +84,13 @@ inline constexpr PolicyToPreferenceMapEntry kBraveSimplePolicyMap[] = {
         playlist::kPlaylistEnabledPref,
         base::Value::Type::BOOLEAN,
     },
+#if BUILDFLAG(ENABLE_EMAIL_ALIASES)
+    {
```
netzenbot, line 88: This pref is added to the iOS policy map but never registered in the iOS pref registry. On iOS there is no `EmailAliasesService`, so `EmailAliasesService::RegisterProfilePrefs` is never called. Add `registry->RegisterBooleanPref(email_aliases::prefs::kEmailAliasesDisabledByPolicy, false)` inside a `#if BUILDFLAG(ENABLE_EMAIL_ALIASES)` guard in `ios/browser/shared/prefs/browser_prefs_impl.mm`, following the same pattern as `brave_talk::prefs::kDisabledByPolicy`. See https://github.com/brave/brave-core/tree/master/docs/best-practices/ios.md#IOS-046

browser/email_aliases/email_aliases_service_factory.cc
```diff
@@ -45,6 +47,21 @@ EmailAliasesServiceFactory* EmailAliasesServiceFactory::GetInstance() {
...
+    content::BrowserContext* context) {
+  if (!features::IsEmailAliasesEnabled()) {
+    return true;
+  }
+  const auto* pref_service = user_prefs::UserPrefs::Get(context);
```
netzenbot: nit: `IsServiceDisabled` only uses `context` to call `user_prefs::UserPrefs::Get(context)`, so it should accept `PrefService*` directly rather than `BrowserContext*`. See [ARCH-003](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-003).

browser/email_aliases/email_aliases_service_factory.cc
```diff
@@ -45,6 +47,21 @@ EmailAliasesServiceFactory* EmailAliasesServiceFactory::GetInstance() {
   return instance.get();
 }
 
+// static
+bool EmailAliasesServiceFactory::IsServiceDisabled(
+    content::BrowserContext* context) {
+  if (!features::IsEmailAliasesEnabled()) {
+    return true;
+  }
```
netzenbot: nit: `PrefService*` is a short, descriptive type — prefer `const PrefService* pref_service` over `const auto*` here. See https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-053

browser/email_aliases/email_aliases_service_factory_unittest.cc
```diff
@@ -77,6 +72,39 @@ TEST_F(EmailAliasesServiceFactoryTest, NoServiceWhenFeatureDisabled) {
   EXPECT_EQ(service, nullptr);
 }
 
+TEST_F(EmailAliasesServiceFactoryTest, ServiceWithPolicy) {
+  scoped_feature_list_.Reset();
+  scoped_feature_list_.InitWithFeatures(
+      {brave_account::features::BraveAccountFeatureForTesting(),
+       features::kEmailAliases},
```
netzenbot, line 79: nit: `ServiceWithPolicy` doesn't tell you what to expect without reading the body. Compare with `NoServiceWhenFeatureDisabled` in the same file. Something like `ServiceDisabledWhenPolicyDisablesIt` or `PolicyCanDisableService` would be clearer per [TI-038](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-038).

mkarolin (approved):
`strings`++

ios/browser/brave_origin/brave_origin_service_factory.mm
```diff
@@ -146,6 +151,14 @@ explicit BraveOriginDelegateIOS(SkusServiceGetter skus_service_getter)
          BraveOriginServiceFactory::BraveOriginPrefMetadata(
              false,
              /*user_settable=*/false)},
+
+#if BUILDFLAG(ENABLE_EMAIL_ALIASES)
+        // Email Aliases preferences
+        {email_aliases::prefs::kEmailAliasesDisabledByPolicy,
```
bbondy: Isn't this a compiling error on iOS? Not sure why CI is passing.

ios/browser/brave_origin/brave_origin_service_factory.mm
```diff
@@ -146,6 +151,14 @@ explicit BraveOriginDelegateIOS(SkusServiceGetter skus_service_getter)
...
+         BraveOriginServiceFactory::BraveOriginPrefMetadata(
+             true,
```
bbondy: Once it is changed to the enabled variant, this should also switch to false

browser/resources/settings/br/page_visibility.ts
```diff
@@ -71,6 +72,7 @@ function getPageVisibility () {
       // </if>
       surveyPanelist: false,
       braveTor: false,
+      emailAliaes: false,
```
bbondy: should be: `emailAliases`

browser/email_aliases/email_aliases_service_factory.cc
```diff
@@ -45,6 +47,19 @@ EmailAliasesServiceFactory* EmailAliasesServiceFactory::GetInstance() {
...
+bool EmailAliasesServiceFactory::IsServiceEnabled(
+    content::BrowserContext* context) {
+  if (!features::IsEmailAliasesEnabled()) {
+    return false;
+  }
+  const PrefService* pref_service = user_prefs::UserPrefs::Get(context);
+  if (pref_service && pref_service->GetBoolean(prefs::kEmailAliasesEnabled)) {
+    return true;
+  }
+  return false;
```
bbondy: nit: Just `return pref_service && pref_service->GetBoolean(prefs::kEmailAliasesEnabled)`

components/policy/resources/templates/policy_definitions/BraveSoftware/EmailAliasesEnabled.yaml
```diff
@@ -0,0 +1,36 @@
+caption: Enable Email Aliases
+default: null
+desc: |-
+  Enable Email Aliases in Brave.
+
+  Email Aliases is a feature that allows users to sign up for services anonymously and keep their emails free from spam.
```
bbondy, line 6: nit: I think the margin spacing here is inconsistent.

netzenbot (approved):

ios/browser/brave_origin/brave_origin_service_factory.mm
boocmp: Because this buildflag is disabled for iOS for now. Fixed.

components/policy/resources/templates/policy_definitions/BraveSoftware/EmailAliasesEnabled.yaml
boocmp: I'm using the same structure as AI Chat.

bbondy: That doesn't make it right does it? You can ignore if you really want to as it's a nit.

bbondy (approved):

browser/email_aliases/email_aliases_service_factory.cc
```diff
@@ -22,7 +24,7 @@ namespace email_aliases {
 // static
 EmailAliasesService* EmailAliasesServiceFactory::GetServiceForProfile(
     Profile* profile) {
-  if (!features::IsEmailAliasesEnabled()) {
+  if (!IsServiceEnabled(profile)) {
```
goodov: why this is done on a getter level rather than on a construction level? This will break when you add `ServiceIsCreatedWithBrowserContext` (which you should do).

boocmp: Yep, it could. Lets move checks to the build func.

ios/browser/shared/prefs/browser_prefs_impl.mm
```diff
@@ -86,6 +91,10 @@ void RegisterBrowserStatePrefs(user_prefs::PrefRegistrySyncable* registry) {
                                 false);
 #endif
 
+#if BUILDFLAG(ENABLE_EMAIL_ALIASES)
+  email_aliases::EmailAliasesService::RegisterProfilePrefs(registry);
```
goodov: shouldn't we do this via a factory method instead?
```
class EmailAliasesServiceFactory {
  void RegisterProfilePrefs(user_prefs::PrefRegistrySyncable* registry) final;
}
```

goodov (approved):

ios/browser/shared/prefs/browser_prefs_impl.mm
boocmp: We could proxy the service's registration with a new method, but I don't see any benefits. Having this in the service allows us to perform registration in the component's unit tests

boocmp: Ah, you mean the virtual method, yep, it makes sense. Will update.

PR #35148, opened 2026-04-01
Use base::SequenceBound in AdBlockService

atuchin-m:
Resolves https://github.com/brave/brave-browser/issues/54257
The PR:
* rewrites test `EngineTestObserver` to avoid direct UI => task_runner access ;
* use `base::SequenceBound` instead of manual posting;

PR #35881, opened 2026-04-24
Remove unnecessary racy check on adblock service startup

bridiver:
Resolves


bridiver:
This isn't actually racy because it has to post back the UI thread before the engine loads. It's not great, but will defer to https://github.com/brave/brave-core/pull/35148 for fix

PR #35885, opened 2026-04-24
[ads][CodeHealth] Rename PlatformHelper to OperatingSystem

tmancey:
Renames the platform folder and all PlatformHelper/PlatformType
identifiers to OperatingSystem/OperatingSystemType to better reflect
what the abstraction models. The old name implied mobile versus desktop
awareness, but the class only identifies the host operating system.

IsMobile is a platform concept rather than an operating system identity
question, so it is extracted as a free function in operating_system_util
derived from OperatingSystemType. This removes the virtual method and
all per-platform overrides that simply hardcoded true or false based on
the type already returned by GetType.

Also removes the NoDestructor friend declarations from subclass headers
by making their constructors public, which is all NoDestructor requires.

No behavioral changes.

Resolves


aseren (approved):

tmancey:
1. Lives inside namespace brave_ads so no ambiguity with base::SysInfo or anything else
2. Fixed. FakeOperatingSystem now installs itself via SetForTesting(this) in its constructor and resets to nullptr in its destructor
3. Intentional and correct. NoDestructor is an implementation detail of GetInstance() and has no place in the public header

PR #35888, opened 2026-04-24
[Polkadot] [Wallet] update PolkadotTxManager to track transaction status

### LINKED ISSUE #54944: [Polkadot] [Wallet] transaction status should be updated with block tracking
We can sign and send Polkadot DOT transfers, but currently lack the block tracking routines required to update the pending transactions.

cmazakas:
Resolves https://github.com/brave/brave-browser/issues/54944

Update the PolkadotTxManager to track blocks and update transaction status using finalized data.

# Test Plan

Confirm a DOT send transaction in the UI and simply wait, the status should eventually come in over the course of 30s to 1 minute.

* Confirm MAX can successfully send a transfer_all (inspect this via block explorer).
* Confirm normal send amounts are successful
* Confirm we see errors by attempting a normal send that'd cause the account to dip below its "existential deposit", i.e. take account's DOT, subtract approximation of its fee, then attempt normal send (don't hit MAX). This should error. See https://westend.subscan.io/tx/0x1c6cc512fae5984d98bc670787554c80897de85539960f1c4c7ce88654bcd113 as an example.

### DIFF
--- a/components/brave_wallet/browser/polkadot/polkadot_block_tracker.cc
+++ b/components/brave_wallet/browser/polkadot/polkadot_block_tracker.cc
@@ -5,23 +5,28 @@
 
 #include "brave/components/brave_wallet/browser/polkadot/polkadot_block_tracker.h"
 
-#include "base/notimplemented.h"
-#include "brave/components/brave_wallet/browser/json_rpc_service.h"
 #include "brave/components/brave_wallet/browser/polkadot/polkadot_substrate_rpc.h"
+#include "brave/components/brave_wallet/common/common_utils.h"
 
 namespace brave_wallet {
 
-PolkadotBlockTracker::PolkadotBlockTracker() : weak_factory_(this) {}
+PolkadotBlockTracker::PolkadotBlockTracker(PolkadotSubstrateRpc& polkadot_rpc)
+    : polkadot_rpc_(polkadot_rpc) {}
 
 PolkadotBlockTracker::~PolkadotBlockTracker() = default;
 
 void PolkadotBlockTracker::Start(const std::string& chain_id,
                                  base::TimeDelta interval) {
-  NOTIMPLEMENTED_LOG_ONCE();
-}
+  CHECK(IsPolkadotNetwork(chain_id));
 
-void PolkadotBlockTracker::Stop(const std::string& chain_id) {
-  NOTIMPLEMENTED_LOG_ONCE();
+  auto& timer = timers_[chain_id];
+  if (!timer) {
+    timer = std::make_unique<base::RepeatingTimer>();
+  }
+
+  timer->Start(FROM_HERE, interval,
+               base::BindRepeating(&PolkadotBlockTracker::GetLatestBlock,
+                                   weak_ptr_factory_.GetWeakPtr(), chain_id));
 }
 
 void PolkadotBlockTracker::AddObserver(Observer* observer) {
@@ -33,14 +38,42 @@ void PolkadotBlockTracker::RemoveObserver(Observer* observer) {
 }
 
 void PolkadotBlockTracker::GetLatestBlock(const std::string& chain_id) {
-  NOTIMPLEMENTED_LOG_ONCE();
+  polkadot_rpc_->GetFinalizedHead(
+      chain_id, base::BindOnce(&PolkadotBlockTracker::OnGetFinalizedHead,
+                               weak_ptr_factory_.GetWeakPtr(), chain_id));
 }
 
-void PolkadotBlockTracker::OnGetLatestBlock(const std::string& chain_id,
-                                            uint64_t block_num,
-                                            mojom::ProviderError error,
-                                            const std::string& error_message) {
-  NOTIMPLEMENTED_LOG_ONCE();
+void PolkadotBlockTracker::OnGetFinalizedHead(
+    const std::string& chain_id,
+    std::optional<std::array<uint8_t, kPolkadotBlockHashSize>> block_hash,
+    std::optional<std::string> err_str) {
+  if (err_str || !block_hash) {
+    return;
+  }
+
+  auto& latest_hash = latest_block_hashes_map_[chain_id];
+  if (block_hash == latest_hash) {
+    return;
+  }
+
+  latest_hash = block_hash.value();
+  polkadot_rpc_->GetBlockHeader(
+      chain_id, latest_hash,
+      base::BindOnce(&PolkadotBlockTracker::OnGetFinalizedBlockHeader,
+                     weak_ptr_factory_.GetWeakPtr(), chain_id));
+}
+
+void PolkadotBlockTracker::OnGetFinalizedBlockHeader(
+    const std::string& chain_id,
+    std::optional<PolkadotBlockHeader> block_header,
+    std::optional<std::string> err_str) {
+  if (err_str || !block_header) {
+    return;
+  }
+
+  for (auto& observer : observers_) {
+    observer.OnLatestBlock(chain_id, block_header->block_number);
+  }
 }
 
 }  // namespace brave_wallet
--- a/components/brave_wallet/browser/polkadot/polkadot_block_tracker.cc
+++ b/components/brave_wallet/browser/polkadot/polkadot_block_tracker.cc
@@ -71,9 +71,8 @@ void PolkadotBlockTracker::OnGetFinalizedBlockHeader(
     return;
   }
 
-  for (auto& observer : observers_) {
-    observer.OnLatestBlock(chain_id, block_header->block_number);
-  }
+  observers_.Notify(&Observer::OnLatestBlock, chain_id,
+                    block_header->block_number);
 }
 
 }  // namespace brave_wallet
--- a/components/brave_wallet/browser/polkadot/polkadot_block_tracker.h
+++ b/components/brave_wallet/browser/polkadot/polkadot_block_tracker.h
@@ -6,20 +6,22 @@
 #ifndef BRAVE_COMPONENTS_BRAVE_WALLET_BROWSER_POLKADOT_POLKADOT_BLOCK_TRACKER_H_
 #define BRAVE_COMPONENTS_BRAVE_WALLET_BROWSER_POLKADOT_POLKADOT_BLOCK_TRACKER_H_
 
-#include <memory>
 #include <string>
 
 #include "base/memory/weak_ptr.h"
 #include "base/observer_list.h"
 #include "base/time/time.h"
 #include "brave/components/brave_wallet/browser/block_tracker.h"
-#include "brave/components/brave_wallet/common/brave_wallet.mojom.h"
+#include "brave/components/brave_wallet/browser/polkadot/polkadot_block_header.h"
+#include "brave/components/brave_wallet/browser/polkadot/polkadot_utils.h"
 
 namespace brave_wallet {
 
+class PolkadotSubstrateRpc;
+
 class PolkadotBlockTracker : public BlockTracker {
  public:
-  PolkadotBlockTracker();
+  explicit PolkadotBlockTracker(PolkadotSubstrateRpc& polkadot_rpc);
   ~PolkadotBlockTracker() override;
 
   PolkadotBlockTracker(const PolkadotBlockTracker&) = delete;
@@ -27,30 +29,32 @@ class PolkadotBlockTracker : public BlockTracker {
 
   class Observer : public base::CheckedObserver {
    public:
-    // Fires for each latest block check
+    // Fires for each latest block check.
     virtual void OnLatestBlock(const std::string& chain_id,
-                               uint64_t block_num) = 0;
-    // Only fires when there is a new block
-    virtual void OnNewBlock(const std::string& chain_id,
-                            uint64_t block_num) = 0;
+                               uint32_t block_num) = 0;
   };
 
   // BlockTracker
   void Start(const std::string& chain_id, base::TimeDelta interval) override;
-  void Stop(const std::string& chain_id) override;
 
   void AddObserver(Observer* observer);
   void RemoveObserver(Observer* observer);
 
  private:
   void GetLatestBlock(const std::string& chain_id);
-  void OnGetLatestBlock(const std::string& chain_id,
-                        uint64_t block_num,
-                        mojom::ProviderError error,
-                        const std::string& error_message);
+  void OnGetFinalizedHead(
+      const std::string& chain_id,
+      std::optional<std::array<uint8_t, kPolkadotBlockHashSize>>,
+      std::optional<std::string>);
+  void OnGetFinalizedBlockHeader(const std::string& chain_id,
+                                 std::optional<PolkadotBlockHeader>,
+                                 std::optional<std::string>);
 
+  std::map<std::string, std::array<uint8_t, kPolkadotBlockHashSize>>
+      latest_block_hashes_map_;
   base::ObserverList<Observer> observers_;
-  base::WeakPtrFactory<PolkadotBlockTracker> weak_factory_{this};
+  raw_ref<PolkadotSubstrateRpc> polkadot_rpc_;
+  base::WeakPtrFactory<PolkadotBlockTracker> weak_ptr_factory_{this};
 };
 
 }  // namespace brave_wallet
--- a/components/brave_wallet/browser/polkadot/polkadot_block_tracker.h
+++ b/components/brave_wallet/browser/polkadot/polkadot_block_tracker.h
@@ -8,6 +8,7 @@
 
 #include <string>
 
+#include "base/containers/flat_map.h"
 #include "base/memory/weak_ptr.h"
 #include "base/observer_list.h"
 #include "base/time/time.h"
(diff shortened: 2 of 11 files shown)


components/brave_wallet/browser/polkadot/polkadot_block_tracker.h
```diff
@@ -6,51 +6,55 @@
...
-                        uint64_t block_num,
-                        mojom::ProviderError error,
-                        const std::string& error_message);
-
+  void OnGetFinalizedHead(
+      const std::string& chain_id,
+      std::optional<std::array<uint8_t, kPolkadotBlockHashSize>>,
+      std::optional<std::string>);
+  void OnGetFinalizedBlockHeader(const std::string& chain_id,
+                                 std::optional<PolkadotBlockHeader>,
```
netzenbot: nit: `latest_block_hashes_map_` will hold at most a handful of chain IDs. Per the container guidelines, `base::flat_map` is preferred over `std::map` for small or write-once sorted containers — `std::map` is only justified for large, frequently-mutated collections. ([Use the Right Associative Container](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-005))

components/brave_wallet/browser/polkadot/polkadot_block_tracker.h
```diff
@@ -6,51 +6,55 @@
...
+                                 std::optional<std::string>);
+
+  std::map<std::string, std::array<uint8_t, kPolkadotBlockHashSize>>
+      latest_block_hashes_map_;
```
netzenbot, line 56: nit: The parameters for `OnGetFinalizedHead` and `OnGetFinalizedBlockHeader` are unnamed in the header. Per [CSA-059](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-059), all function parameters in header declarations should be named to match the `.cc` file (e.g., `block_hash`, `err_str`, `block_header`).

components/brave_wallet/browser/polkadot/polkadot_tx_manager.h
```diff
@@ -87,14 +91,22 @@ class PolkadotTxManager : public TxManager,
       base::expected<std::pair<std::string, PolkadotExtrinsicMetadata>,
                      std::string> tx_hash_metadata_pair);
 
+  void OnUpdatePendingTransactions(
+      PolkadotTransactionStatusTask* task,
+      std::unique_ptr<PolkadotTxMeta> polkadot_tx,
+      base::expected<
```
netzenbot, line 97: nit: The `base::expected<...>` parameter in `OnUpdatePendingTransactions` is unnamed in the header. Per [CSA-059](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-059), parameters should be named to match the `.cc` file (e.g., `result`).

components/brave_wallet/browser/polkadot/polkadot_test_utils.cc
```diff
@@ -508,6 +481,68 @@ bool PolkadotMockRpc::HandleGetAccountInfoRequest(
...
 
+bool IsCommand(const base::DictValue& req_body, std::string_view method) {
+  if (const auto* json_method = req_body.FindString("method");
+      json_method && *json_method == method) {
+    return true;
+  }
+  return false;
+}
+
+const base::ListValue* FindParamsOrNull(const base::DictValue& req_body) {
```
netzenbot: nit: `IsCommand` and `FindParamsOrNull` are file-internal helpers and should go in the existing anonymous namespace at the top of the file, not directly in `namespace brave_wallet`. See [CS-016](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-016).

components/brave_wallet/browser/polkadot/polkadot_block_tracker.cc
```diff
@@ -33,14 +38,42 @@ void PolkadotBlockTracker::RemoveObserver(Observer* observer) {
...
+void PolkadotBlockTracker::OnGetFinalizedBlockHeader(
+    const std::string& chain_id,
+    std::optional<PolkadotBlockHeader> block_header,
+    std::optional<std::string> err_str) {
+  if (err_str || !block_header) {
+    return;
+  }
+
+  for (auto& observer : observers_) {
+    observer.OnLatestBlock(chain_id, block_header->block_number);
```
netzenbot: nit: Per [CS-056](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-056), prefer `observers_.Notify(&Observer::OnLatestBlock, chain_id, block_header->block_number)` over the manual for-loop. The `FilBlockTracker` sibling uses the same pattern, but the documented best practice is to use the `Notify` helper.

components/brave_wallet/browser/polkadot/polkadot_tx_manager.cc
```diff
@@ -281,17 +286,113 @@ mojom::CoinType PolkadotTxManager::GetCoinType() const {
...
+  }
+
+  tx_state_manager().AddOrUpdateTx(*polkadot_tx);
 }
 
-void PolkadotTxManager::OnNewBlock(const std::string& chain_id,
-                                   uint64_t block_num) {
-  NOTIMPLEMENTED_LOG_ONCE();
+void PolkadotTxManager::OnLatestBlock(const std::string& chain_id,
+                                      uint32_t block_num) {
```
cypt4, line 417: Looks like block_num is unused

components/brave_wallet/browser/polkadot/polkadot_tx_manager.cc
```diff
@@ -281,17 +286,113 @@ mojom::CoinType PolkadotTxManager::GetCoinType() const {
...
+                               std::move(polkadot_tx)));
+  }
+
+  CheckIfBlockTrackerShouldRun(pending_chain_ids);
 }
 
-void PolkadotTxManager::OnLatestBlock(const std::string& chain_id,
-                                      uint64_t block_num) {
-  NOTIMPLEMENTED_LOG_ONCE();
+void PolkadotTxManager::OnUpdatePendingTransactions(
```
cypt4: OnUpdatePendingTransactions -> OnUpdatePendingTransaction since it handles a single transaction status update.
Maybe even change to smth like OnPendingTransactionStatusResolved

components/brave_wallet/browser/polkadot/polkadot_tx_manager.cc
```diff
@@ -281,17 +286,113 @@ mojom::CoinType PolkadotTxManager::GetCoinType() const {
...
+      CHECK(fee_paid.has_value());
+      adjust_transfer_all_amount(polkadot_tx->tx());
+      polkadot_tx->set_status(mojom::TransactionStatus::Confirmed);
+      polkadot_tx->set_confirmed_time(base::Time::Now());
+      polkadot_tx->tx()->set_fee(*fee_paid);
+      break;
+
+    case PolkadotTransactionStatus::kFailed:
+      CHECK(fee_paid.has_value());
+      adjust_transfer_all_amount(polkadot_tx->tx());
```
cypt4: Could tx be failed while fee_paid is nullopt?

components/brave_wallet/browser/polkadot/polkadot_tx_manager.cc
```diff
@@ -281,17 +286,113 @@ mojom::CoinType PolkadotTxManager::GetCoinType() const {
...
+      polkadot_tx->set_status(mojom::TransactionStatus::Dropped);
+      polkadot_tx->tx()->set_fee(0);
+      break;
+
+    case PolkadotTransactionStatus::kInvalidResponse:
+      // Don't clear our estimated fee here. This case only occurs when we've
+      // failed to parse the events blob or there was a failed integrity check
+      // within it. Regardless, the user's extrinsic is still included inside of
+      // a finalized block which means a fee was incurred.
+      polkadot_tx->set_status(mojom::TransactionStatus::Error);
```
cypt4: Why invalid response is a tx error?

components/brave_wallet/browser/polkadot/polkadot_tx_manager.cc
```diff
@@ -281,17 +286,113 @@ mojom::CoinType PolkadotTxManager::GetCoinType() const {
...
+
+    auto* task = task_ptr.get();
+    polkadot_transaction_status_tasks_.insert(std::move(task_ptr));
+
+    task->Start(base::BindOnce(&PolkadotTxManager::OnUpdatePendingTransactions,
+                               weak_ptr_factory_.GetWeakPtr(), task,
+                               std::move(polkadot_tx)));
+  }
+
+  CheckIfBlockTrackerShouldRun(pending_chain_ids);
```
cypt4: If UpdatePendingTransactions is called with exact chain_id looks like timer will be stopped for other chains.

components/brave_wallet/browser/polkadot/polkadot_block_tracker.h
```diff
@@ -6,51 +6,55 @@
...
-                               uint64_t block_num) = 0;
-    // Only fires when there is a new block
-    virtual void OnNewBlock(const std::string& chain_id,
-                            uint64_t block_num) = 0;
+                               uint32_t block_num) = 0;
   };
 
   // BlockTracker
   void Start(const std::string& chain_id, base::TimeDelta interval) override;
-  void Stop(const std::string& chain_id) override;
```
cypt4, line 40: Why we don't have stop method?

components/brave_wallet/browser/polkadot/polkadot_tx_manager.cc
cmazakas: Yeah, I was thinking about that. I tried to copy what the other block trackers were doing and it seemed like we didn't necessarily always use these kinds of values. I can update this call signature to no longer take a `block_num` and I can even skip the intermediate call to actually get the `block_num` itself.

components/brave_wallet/browser/polkadot/polkadot_tx_manager.cc
cmazakas: For this case, we should only be hitting when we see `system(ExtrinsicFailed)` in the events log for the block, which means the extrinsic was processed and included in a finalized  block which for send transactions should always incur a fee.

components/brave_wallet/browser/polkadot/polkadot_tx_manager.cc
cmazakas: I actually wasn't sure how else to store this information, ha ha.

If we receive an invalid response here, should we just leave the transaction untouched? I guess that makes a ton of sense.

components/brave_wallet/browser/polkadot/polkadot_tx_manager.cc
cmazakas: Alright, this is done.

I added a test for this that covers two chains. I also added a test that checks to make sure also stop block tracking for a given chain if there's no more Submitted transactions. This way, we only block track when there's a Submitted transaction.

components/brave_wallet/browser/polkadot/polkadot_block_tracker.h
cmazakas: We shouldn't need one here. We get `Stop()` from the base class which removes the timer from the `std::set`.

components/brave_wallet/browser/polkadot/polkadot_tx_manager.cc
```diff
@@ -281,17 +286,122 @@ mojom::CoinType PolkadotTxManager::GetCoinType() const {
...
+    new_pending_chain_ids.insert(polkadot_tx->chain_id());
+
+    auto task_ptr = PolkadotTransactionStatusTask::Create(
+        *polkadot_wallet_service_, keyring_service(),
+        polkadot_tx->from()->Clone(), polkadot_tx->chain_id(),
+        extrinsic_metadata->extrinsic(), extrinsic_metadata->block_num(),
+        extrinsic_metadata->mortality_period());
+
+    auto* task = task_ptr.get();
+    polkadot_transaction_status_tasks_.insert(std::move(task_ptr));
```
netzenbot, line 328: `PolkadotTransactionStatusTask::Create` returns `nullptr` when `mortality_period > 1024` (see the factory implementation). Here `task_ptr` is never checked before `task->Start(...)` is called, so an oversized mortality period causes a null dereference crash. Add a null check and `continue` (or log an error) if `task_ptr` is null. ([CSM-029](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-memory.md#CSM-029))

components/brave_wallet/browser/polkadot/polkadot_block_tracker_unittest.cc
```diff
@@ -0,0 +1,126 @@
...
+  tracker_->Start(mojom::kPolkadotTestnet, base::Seconds(2));
+  EXPECT_CALL(observer, OnLatestBlock(mojom::kPolkadotTestnet, 0x1c06355))
+      .Times(0);
+  EXPECT_CALL(observer, OnLatestBlock(mojom::kPolkadotMainnet, 0x1c06355))
+      .Times(0);
+  task_environment_.FastForwardBy(base::Seconds(5));
+  EXPECT_TRUE(testing::Mock::VerifyAndClearExpectations(&observer));
+}
+
+}  // namespace brave_wallet
```
netzenbot, line 130: The new `OnGetFinalizedHead` implementation has a deduplication guard — if the finalized head hash hasn't changed, it returns early and skips observer notification. Neither `GetLatestBlock` nor `GetLatestBlockInvalidResponseJSON` tests this path: neither advances the mock clock a second time with the same hash and asserts `OnLatestBlock` is NOT called again. Consider adding a test that fires the timer twice with an identical finalized head and asserts the observer fires only once. ([TI-039](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-039))

components/brave_wallet/browser/polkadot/polkadot_block_tracker_unittest.cc
```diff
@@ -0,0 +1,126 @@
...
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "brave/components/brave_wallet/browser/polkadot/polkadot_block_tracker.h"
+
+#include <memory>
+#include <optional>
+#include <string>
+
+#include "base/json/json_writer.h"
+#include "base/scoped_observation.h"
```
netzenbot, line 12: nit: `base/json/json_writer.h`, `base/strings/string_number_conversions.h`, and `brave/.../polkadot_wallet_service.h` are not used in this file — `PolkadotWalletService` is never instantiated and no JSON writing or number conversion functions appear in the test. Per [CS-001](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-001), remove unused includes.

components/brave_wallet/browser/polkadot/polkadot_tx_manager_unittest.cc
```diff
@@ -10,22 +10,29 @@
...
 #include "base/functional/callback_forward.h"
+#include "base/scoped_observation.h"
+#include "base/strings/strcat.h"
 #include "base/strings/string_util.h"
+#include "base/test/gmock_callback_support.h"
 #include "base/test/scoped_feature_list.h"
 #include "base/test/task_environment.h"
 #include "brave/components/api_request_helper/api_request_helper.h"
 #include "brave/components/brave_wallet/browser/account_resolver_delegate_impl.h"
 #include "brave/components/brave_wallet/browser/blockchain_registry.h"
```
netzenbot, line 21: nit: `brave/components/brave_wallet/browser/brave_wallet_constants.h` is added but nothing from it appears to be used in this test. Per [CS-001](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-001), remove unused includes.

components/brave_wallet/browser/polkadot/polkadot_tx_manager.cc
cmazakas: Done!

components/brave_wallet/browser/polkadot/polkadot_tx_manager_unittest.cc
```diff
@@ -402,20 +509,265 @@ TEST_F(PolkadotTxManagerUnitTest, ApproveTransaction) {
   const auto* tx = txs.FindDict(tx_meta_id);
   ASSERT_TRUE(tx);
```
netzenbot, line 537: nit: `ApproveTransaction_Confirmed` and `ApproveTransaction_Failed` both copy-paste the same 7-assertion pre-change block verbatim. Per [TI-024](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-024), extract this into a helper like `ExpectSubmittedTxState(polkadot_tx.get(), tx_meta_id, kExpectedExtrinsic)` to avoid the repetition.

components/brave_wallet/browser/polkadot/polkadot_tx_manager.cc
cypt4: How this works if rpc node is down? I believe we should keep tx status as pending if we can't resolve tx status.

cmazakas: In the case of the network being unavailable, we'll hit the error arm of the `base::expected` which is handled here:
```cpp
  if (!result.has_value()) {
    return;
  }
```
The `kInvalidResponse` handling was only for when e.g. the events blob was corrupted. But has now been updated to leave the transaction as Submitted.

cypt4 (approved):

PR #35900, opened 2026-04-24
Bump postcss from 8.5.3 to 8.5.10 [uplift to beta]

diracdeltas:
uplift of https://github.com/brave/brave-core/pull/35879


mihaiplesa:
@diracdeltas `0.90.x` is a used/protected branch name and TeamCity builds will fail. Why was that used here?

PR #35907, opened 2026-04-25
[plaster] PermissionContextBase migration

### LINKED ISSUE #54952: Migrate `PermissionContextBase` to plaster
This class override would be more safer if we did use brockit, and it would avoid macros in several places.

cdesouza-chromium:
The introduction of plasters for `PermissionContextBase` gets rid of a
lot of replacement cruft that was really difficult to make sense of.
This change also tweaks the interface for
`BraveCanBypassEmbeddingOriginCheck`, so it can make the plaster simpler
for this substitution too.

The core of this change is to move `PermissionContextBase` in upstream
into `chromium_impl::`, and derive our own implementatiom from that,
which simplifies a lot of the issues with naming replacement everywhere.

Resolves https://github.com/brave/brave-browser/issues/54952


components/permissions/permissions_client_unittest.cc
```diff
@@ -42,23 +42,24 @@ TEST_F(PermissionsClientUnitTest, BraveCanBypassEmbeddingOriginCheck) {
             "4n2qrud2xlv9fgj53n6ds3t8cs4fvzs05yzmz:123"),
        ContentSettingsType::BRAVE_CARDANO}};
```
netzenbot, line 44: nit: The loop counter `i` was previously `size_t` and is now `int`. Per [CSA-067](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-067), prefer `size_t` for element counts. It's only used for diagnostic output here so the risk is minimal, but it's trivial to keep it `size_t`.

patches/components-permissions-permission_context_base.cc.patch
```diff
@@ -0,0 +1,37 @@
...
+ namespace {
+ 
+ using PermissionStatus = blink::mojom::PermissionStatus;
+@@ -509,7 +509,7 @@ bool PermissionContextBase::IsPermissionAvailableToOrigins(
+     // possible as this corresponds to the requirements of the secure contexts
+     // spec and matches what is implemented in blink. Right now we just check
+     // the top level and requesting origins.
+-    if (!PermissionsClient::Get()->CanBypassEmbeddingOriginCheck(
++    if (!PermissionsClient::Get()->BraveCanBypassEmbeddingOriginCheck(content_settings_type_, 
+             requesting_origin, embedding_origin) &&
```
netzenbot, line 20: nit: There's trailing whitespace after `content_settings_type_, ` on the wrapped call lines (this appears twice, at the two `BraveCanBypassEmbeddingOriginCheck` replacements). Per [PATCH-001](https://github.com/brave/brave-core/tree/master/docs/best-practices/patches.md#PATCH-001), patches should not introduce unnecessary whitespace changes. The TOML substitution pattern could avoid this by keeping the replacement on one line or trimming the trailing space.

rewrite/components/permissions/permission_context_base.h.toml
```diff
@@ -0,0 +1,33 @@
...
+re_flags = ['MULTILINE']
+replace = '''
+namespace chromium_impl {
+\1
+}  // namespace chromium_impl\2'''
+
+[[substitution]]
+description = 'Make PermissionDecided virtual to allow Brave override in derived class'
+pattern = '  void PermissionDecided('
+replace = '  virtual void PermissionDecided('
```
netzenbot: nit: `pattern = '  void PermissionDecided('` includes leading whitespace, which will break if upstream changes indentation. Per [PLSTR-002](https://github.com/brave/brave-core/tree/master/docs/best-practices/plaster.md#PLSTR-002), use `re_pattern = '(^\s+)(void PermissionDecided\()'` with `replace = '\1virtual void PermissionDecided('` to handle flexible whitespace.

components/permissions/permissions_client_unittest.cc
cdesouza-chromium: Fixed.

patches/components-permissions-permission_context_base.cc.patch
cdesouza-chromium: This white space is appropriate, as there are more arguments after.

rewrite/components/permissions/permission_context_base.h.toml
cdesouza-chromium: Fixed.

netzenbot (approved):

goodov (approved):
`namespace chromium_impl` is an interesting approach. nice!

PR #35909, opened 2026-04-25
Translate selected text using Brave Translate (issue #47798)

### LINKED ISSUE #47798: Use Brave Translate for translate selected text via the context menu
### Description

As it is now, Brave Translate will only translate full web pages. In order to translate selected lines of text/sections of a web page, the translation has to go through Google and, further, Google Search must be selected as the default search engine in order for this option to appear at all:

Image

We should allow users to translate selected text using the same Brave Translate implementation we use for full page translations. This would allow users to translate selected text without having to have Google Search selected as the default engine and without sending any data through Google. 

Selected text that gets translated should appear and leverage the same drop-down box that appears in the address bar, but get run through Brave's implementation rather than Googles, as well as displaying our own icon rather than showing Google's in the dropdown (this branding change [has been mentioned before)](https://github.com/brave/brave-browser/issues/30721):

Image

### Steps to reproduce

1. Select text on a page with Brave Search (or other search engine not Google) selected as default
2. Right-click and observe no Translate option available in context menu
3. Change default search engine to Google, repeat
4. Observe translation option in context menu -- selecting this will run translation through Google

### Actual result

Selected text is translated through Google rather than Brave Translate, shows Google icon/branding

### Expected result

Selected text should go through Brave Translate and show Brave branding in dropdown.

### Reproduces how often

Easily reproduced

v1.80.122 (all versions of brave behave this way)

### Channel information

- [x] release (stable)
- [x] beta
- [x] nightly

### Reproducibility

### Miscellaneous information

Some relevant reports/issues:

- https://community.brave.com/t/when-at-last-will-we-be-able-to-mark-and-translate-just-words-phrases-etc-on-a-given-website/633934/13
- https://community.brave.com/t/translate-a-selected-text/508222/8
- https://community.brave.com/t/translator-on-right-click/70061
- https://community.brave.com/t/option-for-english-translation/261162/3
- https://github.com/brave/brave-browser/issues/47723

modhtom:
Implements "Translate selected text" using Brave Translate, resolving #47798.

Added a new context menu item "Translate Selection with Brave" that:
- Appears whenever text is selected, regardless of the default search engine.
- Routes translation through translate.brave.com.
- Reuses the existing translate bubble UI (address-bar dropdown).

Closes https://github.com/brave/brave-browser/issues/47798


mkarolin:
cc: @atuchin-m would you mind taking a look since you are more familiar with this area?

chromium_src/chrome/browser/renderer_context_menu/render_view_context_menu.cc
```diff
@@ -919,3 +931,66 @@ void RenderViewContextMenu::BuildEmailAliasesMenu() {
...
+  return true;
+}
+
+void RenderViewContextMenu::BuildBraveTranslateSelectionItem() {
+  if (!IsBraveTranslateSelectionEnabled()) {
+    return;
+  }
+
+  std::optional<size_t> google_index =
+      menu_model_.GetIndexOfCommandId(IDC_CONTENT_CONTEXT_PARTIAL_TRANSLATE);
```
atuchin-m, line 958: It's not clear why we can't just reuse the google entry.

Image

chromium_src/chrome/browser/renderer_context_menu/render_view_context_menu.cc
```diff
@@ -919,3 +931,66 @@ void RenderViewContextMenu::BuildEmailAliasesMenu() {
...
 #endif  // BUILDFLAG(ENABLE_EMAIL_ALIASES)
+
+bool RenderViewContextMenu::IsBraveTranslateSelectionEnabled() const {
+  if (params_.selection_text.empty()) {
+    return false;
+  }
+  if (!translate::IsBraveTranslateGoAvailable()) {
+    return false;
+  }
+  if (GetProfile()->IsOffTheRecord()) {
```
atuchin-m, line 942: Why we need this?

atuchin-m:
@modhtom is the PR is AI-generated, please add `ai-generated` label.
Also, please provide screenshots or a screencast how it works.

Please note, the according our policies we reject AI-generated PRs that aren't verified by human.

browser/translate/brave_translate_browsertest.cc
```diff
@@ -406,4 +408,31 @@ IN_PROC_BROWSER_TEST_F(BraveTranslateBrowserGoogleRedirectTest,
...
+
+  ASSERT_TRUE(
+      menu.IsItemPresent(IDC_CONTENT_CONTEXT_BRAVE_TRANSLATE_SELECTION));
+  ASSERT_TRUE(
+      menu.IsCommandIdEnabled(IDC_CONTENT_CONTEXT_BRAVE_TRANSLATE_SELECTION));
+
+  menu.ExecuteCommand(IDC_CONTENT_CONTEXT_BRAVE_TRANSLATE_SELECTION, 0);
+  auto* controller = TranslateBubbleController::From(browser());
+  ASSERT_TRUE(controller);
+  EXPECT_TRUE(controller->GetTranslateBubble());
```
atuchin-m, line 435: This is only about the bubble is showing.
Does it make sense to verify it has the translated text?

chromium_src/chrome/browser/renderer_context_menu/render_view_context_menu.cc
```diff
@@ -919,3 +931,66 @@ void RenderViewContextMenu::BuildEmailAliasesMenu() {
...
+  }
+
+  translate::TranslateManager* manager =
+      chrome_translate_client->GetTranslateManager();
+
+  if (!manager) {
+    return;
+  }
+
+  manager->ShowTranslateUI(true, true);
```
atuchin-m, line 995: What exactly this call does?
Looks like it basically shows a **suggest** to translate the **whole** page

atuchin-m (changes requested):

modhtom:
Hi @atuchin-m , this PR was assisted by AI, but I have reviewed, and tested the code manually to ensure it works.
I tried adding labels but I don't think I have the permissions to add labels to PRs in this repository. 

Here are the screenshots showing how the translation feature works:

selectionMenu
popup
popupFail
The last screenshot shows an error because my local developer build receives an expected 403 Forbidden from translate.brave.com, but DevTools verifies the routing to the Brave backend is working perfectly.

chromium_src/chrome/browser/renderer_context_menu/render_view_context_menu.cc
modhtom: I created a new ID because the existing one is hidden unless Google is the default search engine. I can refactor it to reuse the ID and override the search engine check if you want.

chromium_src/chrome/browser/renderer_context_menu/render_view_context_menu.cc
modhtom: To match the existing rules for full-page translation.
The first condition is to see if the user's settings if they turned off Brave Translate.
The second condition is to stop using the feature if the user is using Private/Tor window, preventing page content from leaking.

browser/translate/brave_translate_browsertest.cc
modhtom: Yeah, I need to update this test to mock the response and verify the translated text.

chromium_src/chrome/browser/renderer_context_menu/render_view_context_menu.cc
modhtom: Yes this will trigger the full-page translation bubble. I used it because the upstream PartialTranslateManager sends a request to Google bypassing Brave's proxy. Using ShowTranslateUI route the request to translate.brave.com.

What do you think is the preferred architectural path here?

atuchin-m:
@modhtom from my perspective that PR remove "Translate selected text" feature and just call the page translation (it's another feature).
That is definitely not what we what to.

image


> but DevTools verifies the routing to the Brave backend is working perfectly.

the requests to translate the selected text aren't visible in DevTools. What you see is definitely page translation.

modhtom:
Okay, Thank you for clarifying.
what do you think of this plan to fix it?
I will drop my command ID and reuse `IDC_CONTENT_CONTEXT_PARTIAL_TRANSLATE`, then I will bypass the upstream `search::DefaultSearchProviderIsGoogle()` check so the option appears for all search engines.
And I will make a override for `StartPartialTranslate` to bypass the `ContextualSearchDelegate` then I will use a `network::SimpleURLLoader` to hit Brave's translation endpoint directly, parse the response, and feed the translated text back into the PartialTranslateCallback.

PR #35921, opened 2026-04-25
[ads][CodeHealth] General code health

tmancey:
Remove unnecessary includes across the ads component. Remove verbose per-method doc comments from BatAdsClientNotifierImpl, replacing them with a class-level comment and an override group label. Delete copy and move special member declarations from BatAdsObserver that were never defined and cannot work with a non-copyable mojo::Remote. Fix the AdsServiceDelegate class comment and remove the redundant explicit from its constructor. Collapse trivial Act and Assert phases into a single // Act & Assert expression in value util tests.

No behavioral changes.

Resolves

### DIFF
--- a/browser/brave_ads/ads_service_delegate.h
+++ b/browser/brave_ads/ads_service_delegate.h
@@ -22,15 +22,14 @@ class BraveAdaptiveCaptchaService;
 
 namespace brave_ads {
 
-// Singleton that owns all AdsService and associates them with Profiles.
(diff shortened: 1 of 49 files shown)


tmancey:
BatAdsObserver is always heap-allocated via std::make_unique and passed around as std::unique_ptr, the object itself is never moved.

netzenbot (approved):

aseren (approved):

PR #35925, opened 2026-04-26
Account: unblocks `AuthenticationObserver` usage in iOS Settings

### LINKED ISSUE #51003: Add `RegisterErrorCode`/`LoginErrorCode` enumerators for empty server response fields.

### LINKED ISSUE #54943: Account: unblock `AuthenticationObserver` usage in iOS Settings
When implementing password reset, prefs alone will no longer be enough to determine the account state or decide which UI should be shown.

Previously, the `verification` state implicitly meant a `verification_token` with the registration intent, so it could be modeled as its own distinct state. With password reset (and later password change), different intents can exist across different states, so that separation no longer holds.

There is already a Mojom-based `AuthenticationObserver` used across Account components, but iOS Settings (Swift-based), still relies on pref listeners. This gap exists because `brave_account.mojom` defines nullable primitives (such as `int32?` and `Service?`) which `ios_objc_mojom_wrappers`/`mojom_objc_generator.py` cannot handle.

To support a clean iOS implementation, `brave_account.mojom` needs to be reworked so its Objective-C++ bridges can be generated, allowing iOS Settings to move to `AuthenticationObserver` instead of pref listeners.

### LINKED ISSUE #54974: Account: use `CHECK` for invalid inputs passed to `mojom::Authentication`
Methods on the `mojom::Authentication` interface currently handle invalid caller-provided inputs as soft errors. Since these values come from Mojo callers, invalid input should be treated as a programming error instead.

szilardszaloki:
Resolves https://github.com/brave/brave-browser/issues/54943, resolves https://github.com/brave/brave-browser/issues/51003, resolves https://github.com/brave/brave-browser/issues/54974.

Follow-ups:
- https://github.com/brave/brave-browser/issues/54976 (cc: @boocmp)
- https://github.com/brave/brave-browser/issues/55014

### DIFF
--- a/browser/resources/settings/getting_started_page/brave_account_row.ts
+++ b/browser/resources/settings/getting_started_page/brave_account_row.ts
@@ -20,8 +20,10 @@ import {
   AccountState,
   AccountStateFieldTags,
   ResendConfirmationEmailError,
-  ResendConfirmationEmailErrorCode,
+  ResendConfirmationEmailErrorFieldTags,
+  ResendConfirmationEmailServerErrorCode,
   whichAccountState,
+  whichResendConfirmationEmailError,
 } from '../brave_account.mojom-webui.js'
 import { getCss } from './brave_account_row.css.js'
 import { getHtml } from './brave_account_row.html.js'
@@ -105,8 +107,10 @@ export class SettingsBraveAccountRowElement extends I18nMixinLit(CrLitElement) {
         error = e as ResendConfirmationEmailError
       } else {
         console.error('Unexpected error:', e)
-        error = { netErrorOrHttpStatus: null, errorCode: null }
+        return
       }
+    } finally {
+      this.isResendingConfirmationEmail = false
     }
 
     leoShowAlert({
@@ -122,8 +126,6 @@ export class SettingsBraveAccountRowElement extends I18nMixinLit(CrLitElement) {
               BraveAccountSettingsStrings
                    .BRAVE_ACCOUNT_RESEND_CONFIRMATION_EMAIL_SUCCESS)
     }, 30000)
-
-    this.isResendingConfirmationEmail = false
   }
 
   protected onCancelRegistrationButtonClicked() {
@@ -236,46 +238,42 @@ export class SettingsBraveAccountRowElement extends I18nMixinLit(CrLitElement) {
       : stateHtml[whichAccountState(this.state)]()
   }
 
-  private getErrorMessage(details: ResendConfirmationEmailError): string {
-    const ERROR_STRINGS: Partial<
-      Record<ResendConfirmationEmailErrorCode, string>
+  private getErrorMessage(error: ResendConfirmationEmailError): string {
+    const SERVER_ERROR_STRINGS: Partial<
+      Record<ResendConfirmationEmailServerErrorCode, string>
     > = {
-      [ResendConfirmationEmailErrorCode.kMaximumEmailSendAttemptsExceeded]:
-        this.i18n(
-            BraveAccountSettingsStrings
-                 .BRAVE_ACCOUNT_RESEND_CONFIRMATION_EMAIL_MAXIMUM_SEND_ATTEMPTS_EXCEEDED),
-      [ResendConfirmationEmailErrorCode.kEmailAlreadyVerified]:
-        this.i18n(
-            BraveAccountSettingsStrings
-                 .BRAVE_ACCOUNT_RESEND_CONFIRMATION_EMAIL_ALREADY_VERIFIED),
+      [ResendConfirmationEmailServerErrorCode
+          .kMaximumEmailSendAttemptsExceeded]:
+        BraveAccountSettingsStrings
+             .BRAVE_ACCOUNT_RESEND_CONFIRMATION_EMAIL_MAXIMUM_SEND_ATTEMPTS_EXCEEDED,
+      [ResendConfirmationEmailServerErrorCode.kEmailAlreadyVerified]:
+        BraveAccountSettingsStrings
+             .BRAVE_ACCOUNT_RESEND_CONFIRMATION_EMAIL_ALREADY_VERIFIED,
     }
 
-    const { netErrorOrHttpStatus, errorCode } = details
+    const errorLabel = this.i18n(
+        BraveAccountSettingsStrings.BRAVE_ACCOUNT_ERROR)
 
-    if (netErrorOrHttpStatus == null) {
-      // client-side error
+    if (whichResendConfirmationEmailError(error)
+            === ResendConfirmationEmailErrorFieldTags.CLIENT_ERROR) {
       return this.i18n(
           BraveAccountSettingsStrings.BRAVE_ACCOUNT_CLIENT_ERROR,
-          errorCode != null
-            ? ` (${this.i18n(
-                       BraveAccountSettingsStrings
-                            .BRAVE_ACCOUNT_ERROR)}=${errorCode})`
-            : '',
+          ` (${errorLabel}=${error.clientError!.errorCode})`,
       )
     }
 
-    // server-side error
-    return (
-      (errorCode != null ? ERROR_STRINGS[errorCode] : null)
-      ?? this.i18n(
+    const serverError = error.serverError!
+    const stringId = SERVER_ERROR_STRINGS[serverError.errorCode]
+    if (stringId) {
+      return this.i18n(stringId)
+    }
+
+    return this.i18n(
         BraveAccountSettingsStrings.BRAVE_ACCOUNT_SERVER_ERROR,
-        `${netErrorOrHttpStatus > 0 ? 'HTTP' : 'NET'}=${netErrorOrHttpStatus}`,
-        errorCode != null
-          ? `, ${this.i18n(
-                     BraveAccountSettingsStrings
-                          .BRAVE_ACCOUNT_ERROR)}=${errorCode}`
-          : '',
-      )
+        `${serverError.netErrorOrHttpStatus > 0 ? 'HTTP' : 'NET'}=${
+          serverError.netErrorOrHttpStatus
+        }`,
+        `, ${errorLabel}=${serverError.errorCode}`,
     )
   }
 
--- a/browser/resources/settings/getting_started_page/brave_account_row.ts
+++ b/browser/resources/settings/getting_started_page/brave_account_row.ts
@@ -19,6 +19,7 @@ import { BraveAccountSettingsStrings } from '../brave_components_webui_strings.j
 import {
   AccountState,
   AccountStateFieldTags,
+  ResendConfirmationEmailClientErrorCode,
   ResendConfirmationEmailError,
   ResendConfirmationEmailErrorFieldTags,
   ResendConfirmationEmailServerErrorCode,
@@ -107,10 +108,12 @@ export class SettingsBraveAccountRowElement extends I18nMixinLit(CrLitElement) {
         error = e as ResendConfirmationEmailError
       } else {
         console.error('Unexpected error:', e)
-        return
+        error = {
+          clientError: {
+            errorCode: ResendConfirmationEmailClientErrorCode.kUnexpected,
+          },
+        }
       }
-    } finally {
-      this.isResendingConfirmationEmail = false
     }
 
     leoShowAlert({
@@ -126,6 +129,8 @@ export class SettingsBraveAccountRowElement extends I18nMixinLit(CrLitElement) {
               BraveAccountSettingsStrings
                    .BRAVE_ACCOUNT_RESEND_CONFIRMATION_EMAIL_SUCCESS)
     }, 30000)
+
+    this.isResendingConfirmationEmail = false
   }
 
   protected onCancelRegistrationButtonClicked() {
--- a/components/brave_account/mojom/brave_account.mojom
+++ b/components/brave_account/mojom/brave_account.mojom
@@ -35,11 +35,12 @@ struct RegisterVerifyResult {
 };
 
 enum RegisterClientErrorCode {
-  kOpaqueError                         = -5,
-  kVerificationTokenEncryptionFailed   = -4,
-  kVerificationTokenDecryptionFailed   = -3,
-  kUserNotInTheVerificationState       = -2,
-  kAuthenticationTokenEncryptionFailed = -1,
+  kUnexpected,
+  kOpaqueError,
+  kVerificationTokenEncryptionFailed,
+  kVerificationTokenDecryptionFailed,
+  kUserNotInTheVerificationState,
+  kAuthenticationTokenEncryptionFailed,
 };
 
 struct RegisterClientError {
@@ -101,10 +102,11 @@ struct LoginFinalizeResult {
 };
 
 enum LoginClientErrorCode {
-  kOpaqueError                         = -4,
-  kLoginTokenEncryptionFailed          = -3,
-  kLoginTokenDecryptionFailed          = -2,
-  kAuthenticationTokenEncryptionFailed = -1,
+  kUnexpected,
+  kOpaqueError,
+  kLoginTokenEncryptionFailed,
+  kLoginTokenDecryptionFailed,
+  kAuthenticationTokenEncryptionFailed,
 };
 
 struct LoginClientError {
@@ -155,8 +157,9 @@ struct ResendConfirmationEmailResult {
 };
 
 enum ResendConfirmationEmailClientErrorCode {
-  kUserNotInTheVerificationState     = -2,
-  kVerificationTokenDecryptionFailed = -1,
+  kUnexpected,
+  kUserNotInTheVerificationState,
+  kVerificationTokenDecryptionFailed,
 };
 
 struct ResendConfirmationEmailClientError {
@@ -209,10 +212,11 @@ struct GetServiceTokenResult {
 };
 
 enum GetServiceTokenClientErrorCode {
-  kUserNotLoggedIn                     = -4,
-  kAuthenticationTokenDecryptionFailed = -3,
-  kAuthenticationSessionChanged        = -2,
-  kServiceTokenEncryptionFailed        = -1,
+  kUnexpected,
+  kUserNotLoggedIn,
+  kAuthenticationTokenDecryptionFailed,
+  kAuthenticationSessionChanged,
+  kServiceTokenEncryptionFailed,
 };
 
 struct GetServiceTokenClientError {
--- a/components/brave_account/mojom/brave_account.mojom
+++ b/components/brave_account/mojom/brave_account.mojom
@@ -34,15 +34,26 @@ struct RegisterVerifyResult {
   // Currently empty, serves as a placeholder for success.
 };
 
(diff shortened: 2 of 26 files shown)


boocmp (approved):

SergeyZhukovsky (approved):
Android ++

kylehickinson (approved):
ios++

browser/resources/settings/getting_started_page/brave_account_row.ts
```diff
@@ -105,8 +107,10 @@ export class SettingsBraveAccountRowElement extends I18nMixinLit(CrLitElement) {
         error = e as ResendConfirmationEmailError
       } else {
         console.error('Unexpected error:', e)
-        error = { netErrorOrHttpStatus: null, errorCode: null }
+        return
```
fallaciousreasoning: Hey, should we not still be showing an `Alert`? I think a generic "Something went wrong" is probably better than failing silently.

browser/resources/settings/getting_started_page/brave_account_row.ts
```diff
@@ -236,46 +238,42 @@ export class SettingsBraveAccountRowElement extends I18nMixinLit(CrLitElement) {
...
+             .BRAVE_ACCOUNT_RESEND_CONFIRMATION_EMAIL_ALREADY_VERIFIED,
     }
 
-    const { netErrorOrHttpStatus, errorCode } = details
+    const errorLabel = this.i18n(
+        BraveAccountSettingsStrings.BRAVE_ACCOUNT_ERROR)
 
-    if (netErrorOrHttpStatus == null) {
-      // client-side error
+    if (whichResendConfirmationEmailError(error)
```
fallaciousreasoning, line 262: So the reason the above strings map is partial is because we need to handle this one separately? Perhaps it should be a map of `() => string` for generating the localized strings. 

Otherwsie its going to be easy to miss this when we add new error codes.

components/brave_account/mojom/brave_account.mojom
```diff
@@ -34,15 +34,26 @@ struct RegisterVerifyResult {
...
-enum RegisterErrorCode {
-  // Client-side errors:
-  kOpaqueError                             = -5,
-  kVerificationTokenEncryptionFailed       = -4,
-  kVerificationTokenDecryptionFailed       = -3,
-  kUserNotInTheVerificationState           = -2,
-  kAuthenticationTokenEncryptionFailed     = -1,
-  // Server-side errors:
-  kMiscServerError                         = 0,
+enum RegisterClientErrorCode {
```
fallaciousreasoning, line 37: Hey, do we gain enough from separating the server/client errors here to make it worth it? 

It seems like:

```mojom
struct RegisterClientError {
    int32? net_or_http_code = std::nullopt;
    RegisterClientErrorCode code;
};
```

would work fine.

Going one step further, there's a lot of overlap in error codes for different message types. I think things would be a bit simpler if we just had one error type, which we could specify the specific errors for different endpoints within.

Then, we could just have a single formatter on the Typescript side and let all places which handle errors call into it, and they wouldn't need to add their own error handling.

Might not be for this PR, but I think its worth thinking about and possibly changing.

components/brave_account/resources/brave_account_browser_proxy.ts
```diff
@@ -23,7 +23,7 @@ export interface BraveAccountBrowserProxy {
   dialog_controller: DialogControllerInterface
   password_strength_meter: PasswordStrengthMeterInterface
   closeDialog: () => void
-  getInitiatingService: () => Service | null
+  getInitiatingService: () => Service
 }
```
fallaciousreasoning, line 27: Hey, so this interface seems to exist so people can only see public methods. I don't see why we wouldn't get rid of this interface, and just make the fields we want to hide private.

Otherwise, we're specifying the same thing in two places for no reason (or benefit).

components/brave_account/resources/brave_account_common.ts
```diff
@@ -22,87 +25,97 @@ export type Error =
   | { kind: 'register'; details: RegisterError }
   | { kind: 'resendConfirmationEmail'; details: ResendConfirmationEmailError }
 
-const LOGIN_ERROR_STRINGS: Partial<Record<LoginErrorCode, string>> = {
-  [LoginErrorCode.kEmailNotVerified]:
+const LOGIN_SERVER_ERROR_STRINGS: Partial<
```
fallaciousreasoning, line 28: fwiw, I'm a bit worried about these all being partial, as its going to make it easy for someone to forget to handle them here.

Perhaps this should be a `switch` where you need to specify all the cases (and people need to explicitly add a case fallthrough for the default handling)?

components/brave_account/resources/brave_account_common.ts
```diff
@@ -22,87 +25,97 @@ export type Error =
...
 > = {
-  [ResendConfirmationEmailErrorCode.kMaximumEmailSendAttemptsExceeded]:
+  [ResendConfirmationEmailServerErrorCode.kMaximumEmailSendAttemptsExceeded]:
     BraveAccountStrings.BRAVE_ACCOUNT_RESEND_CONFIRMATION_EMAIL_MAXIMUM_SEND_ATTEMPTS_EXCEEDED,
-  [ResendConfirmationEmailErrorCode.kEmailAlreadyVerified]:
+  [ResendConfirmationEmailServerErrorCode.kEmailAlreadyVerified]:
     BraveAccountStrings.BRAVE_ACCOUNT_RESEND_CONFIRMATION_EMAIL_ALREADY_VERIFIED,
 }
 
 function getErrorMessageImpl<
```
fallaciousreasoning, line 65: coming back to my earlier point from the mojom, I don't think we're gaining anything from having the error codes separatated into client/server and by request type. 

Everything funnels through the same method anyway, so its just introducing more boilerplate.

components/brave_account/resources/brave_account_common.ts
```diff
@@ -22,87 +25,97 @@ export type Error =
...
+      return getErrorMessageImpl(LOGIN_SERVER_ERROR_STRINGS, error.details)
     case 'register':
-      return getErrorMessageImpl(REGISTER_ERROR_STRINGS, error.details)
+      return getErrorMessageImpl(REGISTER_SERVER_ERROR_STRINGS, error.details)
     case 'resendConfirmationEmail':
       return getErrorMessageImpl(
-        RESEND_CONFIRMATION_EMAIL_ERROR_STRINGS,
+        RESEND_CONFIRMATION_EMAIL_SERVER_ERROR_STRINGS,
         error.details,
       )
```
fallaciousreasoning, line 120: similarly, we could completely remove this branching

components/brave_account/resources/brave_account_create_dialog.ts
```diff
@@ -72,12 +72,11 @@ export class BraveAccountCreateDialogElement extends CrLitElement {
...
       } else if (typeof e === 'string') {
         error = {
-          netErrorOrHttpStatus: null,
-          errorCode: RegisterErrorCode.kOpaqueError,
+          clientError: { errorCode: RegisterClientErrorCode.kOpaqueError },
         }
       } else {
         console.error('Unexpected error:', e)
-        error = { netErrorOrHttpStatus: null, errorCode: null }
+        return
```
fallaciousreasoning: I think we still need to show something to the user if the error wasn't something we know about. Otherwise whatever action they take is going to fail silently.

components/brave_account/resources/brave_account_otp_dialog.ts
```diff
@@ -44,7 +44,7 @@ export class BraveAccountOtpDialogElement extends CrLitElement {
         error = e as RegisterError
       } else {
         console.error('Unexpected error:', e)
-        error = { netErrorOrHttpStatus: null, errorCode: null }
+        return
```
fallaciousreasoning: ditto

components/brave_account/resources/brave_account_sign_in_dialog.ts
```diff
@@ -72,12 +75,11 @@ export class BraveAccountSignInDialogElement extends CrLitElement {
...
       } else if (typeof e === 'string') {
         error = {
-          netErrorOrHttpStatus: null,
-          errorCode: LoginErrorCode.kOpaqueError,
+          clientError: { errorCode: LoginClientErrorCode.kOpaqueError },
         }
       } else {
         console.error('Unexpected error:', e)
-        error = { netErrorOrHttpStatus: null, errorCode: null }
+        return
```
fallaciousreasoning: ditto

browser/resources/settings/getting_started_page/brave_account_row.ts
szilardszaloki: The map is partial because it only handles server error codes — the branch at line `257` splits on a different axis (client vs. server), not on a missing entry in the map.

components/brave_account/mojom/brave_account.mojom
szilardszaloki: Hey! 🙂

So the main goal of this PR is to get rid of nullable primitives — `ios_objc_mojom_wrappers` can't handle them, so adding back the `int32?` would not work.

For the codes, they overlap numerically, but each endpoint only returns its own subset, so unifying them would make call sites deal with codes they can never actually get. Also, the same code can mean different things depending on the flow (e.g. `kEmailAlreadyVerified` is unexpected in verify-complete, but user-facing in resend), so a shared string map would be wrong. Per-endpoint enums keep that distinction strongly typed and let consumers switch over a small, flow-specific set.

components/brave_account/resources/brave_account_browser_proxy.ts
szilardszaloki: Yeah, this interface predates this PR. Agreed this is worth handling as a follow-up — I'll file an issue for it.

components/brave_account/resources/brave_account_common.ts
szilardszaloki: The partial-ness is intentional — many server codes are internal (see the comments in `brave_account.mojom`) and deliberately fall through to the generic `BRAVE_ACCOUNT_SERVER_ERROR` formatter.

components/brave_account/mojom/brave_account.mojom
```diff
@@ -65,13 +76,17 @@ enum RegisterErrorCode {
...
-struct RegisterError {
-  // null: client-side error
+struct RegisterServerError {
   // < 0: network error code (net_error_list.h)
   // >= 0: HTTP status code (http_status_code_list.h) if available,
   //       otherwise net::OK
-  int32? netErrorOrHttpStatus;
-  RegisterErrorCode? errorCode;
+  int32 net_error_or_http_status;
+  RegisterServerErrorCode error_code;
```
petemill, line 85: nit: http status is a code too, so shouldn't they both have the `_code` prefix 😄

components/brave_account/resources/brave_account_common.ts
szilardszaloki: The split is for two reasons outside this function:
1. `ios_objc_mojom_wrappers` can't handle nullable primitives, which is what started this rework. Collapsing client/server back into one struct would bring those back.
2. per-endpoint enums let each call site only handle codes it can actually get, and let the UX differ by flow for the same numeric code. For example, `kEmailAlreadyVerified` turns into the message `Email already verified.`, but is an unexpected internal state during verify-complete.

The TS funnel is intentional too — the boilerplate stays in the types, the runtime path is shared.

browser/resources/settings/getting_started_page/brave_account_row.ts
fallaciousreasoning: I see. Perhaps we should have a switch for the error codes so we can be explicit about which server errors should use the fallback (and so people know to handle new error codes here):

```ts
switch (errorCode) {
 case  ResendConfirmationEmailServerErrorCode
          .kMaximumEmailSendAttemptsExceeded:
    return BraveAccountSettingsStrings.BRAVE_ACCOUNT_RESEND_CONFIRMATION_EMAIL_MAXIMUM_SEND_ATTEMPTS_EXCEEDED

case ResendConfirmationEmailServerErrorCode
          .kDefault1:
case ResendConfirmationEmailServerErrorCode
          .kDefault2:
  return DefaultErrorCodeHandling()
```

optional though :smile:

components/brave_account/resources/brave_account_common.ts
szilardszaloki: Not really — each `error.details` is a different Mojom union (`LoginError`/`RegisterError`/`ResendConfirmationEmailError`) with its own server enum, and the string maps are typed against those. TS needs the discriminant to pick the right map. Collapsing this means unifying the Mojom errors first, which brings back the same trade-offs.

components/brave_account/mojom/brave_account.mojom
fallaciousreasoning: Ah, always forget about the lack of optional primitives on iOS. 

The overlapping error codes does make things more complicated - guess there's no point changing things now :)

components/brave_account/resources/brave_account_common.ts
fallaciousreasoning: Yeah the error code overlap on the server would make this hard - okay, fine to leave it as is

components/brave_account/resources/brave_account_common.ts
fallaciousreasoning: Right - the side effect of that though is that new error codes are going to be internal by default, unless people know to update these maps. Happy to leave this as is in this PR though as its already like this.

components/brave_account/resources/brave_account_browser_proxy.ts
szilardszaloki: https://github.com/brave/brave-browser/issues/55014

netzenbot (approved):

components/brave_account/mojom/brave_account.mojom
szilardszaloki: Haha, fair, but then `net_error` should be `net_error_code` too, and now we've summoned `net_error_code_or_http_status_code`. 😄

components/brave_account/resources/brave_account_common.ts
szilardszaloki: Yeah, I think that's fine — whether a code is user-facing comes down to whether we want to show dedicated copy for it. Once you've registered the string in `brave_account_strings.grdp`, wiring up the mapping should be the natural next step.

browser/resources/settings/getting_started_page/brave_account_row.ts
szilardszaloki: That's definitely an alternative — though a bit more maintenance-heavy, since every new enumerator would need its own case to keep exhaustiveness, even if we'd only route it to the fallback anyway. It's all tradeoffs... anyway, let's keep this focused on the nullable primitives for now. 🙂

DJAndries (approved):
it seems fine for the purposes of unblocking iOS issues, but i would highly reconsider not splitting client and server errors between different enums/types, either in this PR or a future one. it makes the type definitions, construction and identification more complex than it needs to be. perhaps you could make `net_error_or_http_status` 0 or -1 for client-side errors, as a workaround for the lack of nullability.

szilardszaloki:
> it seems fine for the purposes of unblocking iOS issues, but i would highly reconsider not splitting client and server errors between different enums/types, either in this PR or a future one. it makes the type definitions, construction and identification more complex than it needs to be. perhaps you could make `net_error_or_http_status` 0 or -1 for client-side errors, as a workaround for the lack of nullability.

@DJAndries Appreciate the suggestion, but the `net_error_or_http_status = 0`/`-1` workaround doesn't seem to quite work — `0` is `net::OK`, and `-1` is `net::ERR_IO_PENDING`, both of which are real `net::Error` values. Happy to revisit unification in a follow-up if we find a cleaner answer.

browser/resources/settings/getting_started_page/brave_account_row.ts
szilardszaloki: https://github.com/brave/brave-core/pull/35925/commits/5c4fe9e8bcb7f43dbf1876dfbf115541837975c6

components/brave_account/resources/brave_account_create_dialog.ts
szilardszaloki: https://github.com/brave/brave-core/pull/35925/commits/5c4fe9e8bcb7f43dbf1876dfbf115541837975c6

components/brave_account/resources/brave_account_otp_dialog.ts
szilardszaloki: https://github.com/brave/brave-core/pull/35925/commits/5c4fe9e8bcb7f43dbf1876dfbf115541837975c6

components/brave_account/resources/brave_account_sign_in_dialog.ts
szilardszaloki: https://github.com/brave/brave-core/pull/35925/commits/5c4fe9e8bcb7f43dbf1876dfbf115541837975c6

fallaciousreasoning (approved):

fallaciousreasoning:
Just wanted to add a :+1: to @DJAndries suggestion to unify client/server errors - its much harder to follow things with the client/server error split.

Maybe we could have a sentinel value like -int16 for `No Error`

DJAndries:
> @DJAndries Appreciate the suggestion, but the `net_error_or_http_status = 0`/`-1` workaround doesn't seem to quite work — `0` is `net::OK`, and `-1` is `net::ERR_IO_PENDING`, both of which are real `net::Error` values. Happy to revisit unification in a follow-up if we find a cleaner answer.

If the request resulted in a `net::OK`, then wouldn't the field contain a status code rather than a net error? If so, it seems that using 0 as a substitute for a null value wouldn't collide.

szilardszaloki:
@DJAndries @fallaciousreasoning Every `*ServerError`'s `net_error_or_http_status` comes from `brave_account::endpoint_client::Response<>`'s `net_error`/`status_code` (see [brave/components/brave_account/endpoint_client/response.h](https://github.com/brave/brave-core/blob/master/components/brave_account/endpoint_client/response.h)).

`status_code` is assigned [here](https://github.com/brave/brave-core/blob/master/components/brave_account/endpoint_client/response.h#L47):
```cpp
response.status_code = headers->response_code();
```

Per `net::HttpResponseHeaders::response_code()`'s [documentation](https://source.chromium.org/chromium/chromium/src/+/main:net/http/http_response_headers.h;l=427-430), the return value can be `0` if the response code text exists but is unparseable. So `0` can legitimately end up in `net_error_or_http_status` for a real server error, and using it as a "this is a client error" sentinel would silently misclassify that case.

Also worth noting: the unified client/server representation isn't a new idea — that's basically what the types we're moving away from in this PR did, and the whole point of the rework is to move away from it for `ios_objc_mojom_wrappers`.

No strong preference between the two shapes on my end, and I'm open to unifying them again if we find a decent representation that also works cleanly with `ios_objc_mojom_wrappers`. But encoding a type distinction (client vs. server) as a magic value on a numeric field isn't something I'd go for.

FWIW, I think the client/server split stands on its own merit — my 2 cents.

PR #35932, opened 2026-04-27
[query-filter] Add schema integration for query-filter parsing

### LINKED ISSUE #54775: [query-filter] Explore JSON Compiler and IDL to parse query-filter.json
### Description
Investigate and follow-up on updating the query_filter.json [schema](https://sourcegraph.com/r/github.com/brave/adblock-lists/-/blob/brave-lists/query-filter.json) to:
```
{
   "rules": [
      {
          "include": [],
          "exclude": [],
          "params": [],
      },
      {
          "include": [],
          "exclude": [],
          "params": [],
      }
   ]
}
```
and to have a query_filter_schema.idl in the client side: 
```
// Schema types by the query filter
[generate_error_messages, none_as_absent_optional]
namespace query_filter {
  // Desc
  dictionary Rule {
    // Desc
    DOMString[] include;
    // Desc
    DOMString[] exclude;
    // Desc
    DOMString[] params;
  };

  // Desc
  dictionary Ruleset {
    // Desc
    Rule[] rules;
  };
};
```
which would allow us to have stronger type checks on the client side, and easier parsing. 

Examples of idl at Brave can be found [here](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/brave/brave-core%24+f:idl%24+&patternType=keyword&sm=0)

r0hit0303:
This change adds a new schema for the [query-filter.json](https://sourcegraph.com/r/github.com/brave/adblock-lists/-/blob/brave-lists/query-filter.json) which streamlines the parsing logic in QueryFilterData by relying on the internal JSON compiler.  

Resolves https://github.com/brave/brave-browser/issues/54775

- [x] Review design docs
  [Browser design principles](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/docs/chrome_browser_design_principles.md)
  [Style guide](https://chromium.googlesource.com/chromium/src/+/main/styleguide/c++/c++.md)
  [Core principles](https://www.chromium.org/developers/core-principles/)
- [x] Ensure there are (tests)[https://www.chromium.org/developers/testing/]. Unit test as much as possible (including edge cases), but also include browser tests covering high level functionality.
- [x] Ensure that there are comments explaining what classes/methods are/do. The "why" is often more important than the "what" in comments. Also update any relevant docs (moving docs from wiki to brave-core if necessary).
- [x] Request security or other review (third-party libraries, rust code, etc...) if applicable [security/privacy review is needed](https://github.com/brave/brave-browser/wiki/Security-reviews) [other review](https://github.com/brave/reviews/issues/new/choose)
  Also see [adding third-party libraries](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/docs/adding_to_third_party.md) for general guidelines on using third party code
- [x] Make sure there is a [ticket](https://github.com/brave/brave-browser/issues) for your issue
- [x] Use Github [auto-closing keywords](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue) in the PR description above
- [x] Write a good [PR/commit description](https://google.github.io/eng-practices/review/developer/cl-descriptions.html)
- [x] Squash any review feedback or "fixup" commits before merge, so that history is a record of what happened in the repo, not your PR
- [x] Add appropriate labels (`QA/Yes` or `QA/No`; `release-notes/include` or `release-notes/exclude`; `OS/...`) to the associated issue
- [x] Checked the PR locally:
  * `npm run test -- brave_browser_tests`, `npm run test -- brave_unit_tests` [wiki](https://github.com/brave/brave-browser/wiki/Tests)
  * `npm run presubmit` [wiki](https://github.com/brave/brave-browser/wiki/Presubmit-checks), `npm run gn_check`, `npm run tslint`
- Run `git rebase master` (if needed)


components/query_filter/browser/query_filter_data.cc
```diff
@@ -57,83 +42,48 @@ const std::string QueryFilterData::GetVersion() const {
...
-      LOG(WARNING) << "query-filter.json: missing required fields";
-      continue;
-    }
-
-    if (rule_dict->size() != 3) {
-      LOG(WARNING) << "query-filter.json: rule entry must have exactly 3 "
-                      "fields: include, exclude and params";
-      continue;
+  // We can't rely on schema::Ruleset::FromValue as a "single" bad rule entry
+  // will result in all the rules getting dropped. Therefore, we iterate over
```
r0hit0303: @cdesouza-chromium Do you have any suggestions on how we can still rely on schema::Ruleset::FromValue w/o losing the whole list of rules if one of the rule is not parsed properly?

components/query_filter/browser/query_filter_component_installer_unittest.cc
```diff
@@ -158,9 +168,44 @@ TEST_F(QueryFilterComponentInstallerTest, TestComponentReady) {
...
-  }));
-  EXPECT_EQ(2U, query_filter::QueryFilterData::GetInstance()->rules().size());
+  EXPECT_TRUE(base::test::RunUntil(
+      [&]() { return GetQueryFilterVersion() == "1.0.0"; }));
+  const auto& new_rules = GetQueryFilterRules();
+  EXPECT_EQ(2U, new_rules.size());
+}
+
+// TODO( https://github.com/brave/brave-browser/issues/10188): Update this test
+// to pass a callback and remove the RunUntil.
```
r0hit0303, line 180: @cdesouza-chromium  @fmarier 
I'll need to make some changes in the component_installer file in-order to allow passing a callback so that it can be fired from [OnQueryFilterFileRead](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/components/query_filter/browser/query_filter_component_installer.cc?L44). That way we can reliably wait inside the test. For now I've added a TODO to capture that, and kept this test anyways.  I'll follow up on the change in another PR to the concern separate.

components/query_filter/browser/query_filter_data_unittest.cc
```diff
@@ -82,36 +85,45 @@ TEST_F(QueryFilterDataTest, TestNonDictRuleEntryIsIgnored) {
...
   EXPECT_THAT(rules[1].include, testing::ElementsAre("*://b/*"));
-  EXPECT_TRUE(rules[1].exclude.empty());
+  EXPECT_THAT(rules[1].exclude, testing::ElementsAre("foubah"));
   EXPECT_THAT(rules[1].params, testing::ElementsAre("y"));
 }
 
-TEST_F(QueryFilterDataTest, TestNonStringListItemsAreIgnored) {
+TEST_F(QueryFilterDataTest, TestOneBadRuleEntryItemIgnoresThatEntries) {
+  // In 1st entry we have couple of integers which doesn't match the expected
+  // schema.
```
fmarier:
```suggestion
  // schema and so that rule will be ignored. Only the second will be loaded.
```

components/query_filter/browser/BUILD.gn
```diff
@@ -13,6 +13,7 @@ static_library("browser") {
 
   public_deps = [
     "//base",
+    "//brave/components/query_filter/common:generated_query_filter_schema",
```
cdesouza-chromium: Could we use just `:schema` here? I think it is pretty self-explanatory.

components/query_filter/browser/query_filter_component_installer_unittest.cc
```diff
@@ -158,9 +168,44 @@ TEST_F(QueryFilterComponentInstallerTest, TestComponentReady) {
...
-           "1.0.0";
-  }));
-  EXPECT_EQ(2U, query_filter::QueryFilterData::GetInstance()->rules().size());
+  EXPECT_TRUE(base::test::RunUntil(
+      [&]() { return GetQueryFilterVersion() == "1.0.0"; }));
+  const auto& new_rules = GetQueryFilterRules();
+  EXPECT_EQ(2U, new_rules.size());
+}
+
+// TODO( https://github.com/brave/brave-browser/issues/10188): Update this test
```
cdesouza-chromium: nit:
```cxx
// TODO(https://github.com/brave/brave-browser/issues/10188): Update this test
```

components/query_filter/browser/query_filter_component_installer_unittest.cc
cdesouza-chromium: Ack

components/query_filter/browser/query_filter_component_installer_unittest.cc
```diff
@@ -158,9 +168,44 @@ TEST_F(QueryFilterComponentInstallerTest, TestComponentReady) {
...
+// to pass a callback and remove the RunUntil.
+TEST_F(QueryFilterComponentInstallerTest,
+       TestComponentReady_WithBadJson_DoesNotUpdateVersion) {
+  // Setup: create a temporary directory and write the bad JSON.
+  base::ScopedTempDir component_install_dir_;
+  ASSERT_TRUE(component_install_dir_.CreateUniqueTempDir());
```
cdesouza-chromium: I've noticed every test is doing:

```cxx
ASSERT_TRUE(component_install_dir_.CreateUniqueTempDir());
```

This call should be moved to `SetUp`, as there's no need to repeat it in every test.

components/query_filter/browser/query_filter_data.cc
```diff
@@ -57,83 +42,48 @@ const std::string QueryFilterData::GetVersion() const {
...
+  // base::JSONReader::ReadList doesn't expose an error message.
   auto parsed = base::JSONReader::ReadAndReturnValueWithError(
       json_data, base::JSON_PARSE_CHROMIUM_EXTENSIONS);
-  if (!parsed.has_value()) {
-    LOG(WARNING) << "query-filter.json parse error: " << parsed.error().message;
+  if (!parsed.has_value() || !parsed->is_list()) {
+    std::string error_message = !parsed.has_value()
+                                    ? parsed.error().message
+                                    : "Root should be base::List type";
+    LOG(WARNING) << "query-filter.json parse error: " << error_message;
```
cdesouza-chromium: `LOG(ERROR)`?

components/query_filter/browser/query_filter_data.cc
cdesouza-chromium: Is that something you absolutely have to support? If it is, then only the way you are doing here. This is obviously not great, but it is still useful that the types are generated by the schema compiler, as we have less code to maintain then by having our own types.

However, I will say that if we are gonna to parse like this, we potentially should delete `Ruleset` from the schema, because there's no use for it, really, when you think about it. The whole point of `Ruleset` was to wrap this loop here, so we may as well just have a field like this:

```cxx
std::vector<schema::Rule> rules_;
```

components/query_filter/browser/query_filter_data.cc
```diff
@@ -57,83 +42,48 @@ const std::string QueryFilterData::GetVersion() const {
...
+  // will result in all the rules getting dropped. Therefore, we iterate over
+  // each rule manually and rely on schema::Rule::FromValue as that
+  // facilitates dropping only the bad rule in case we get a parse error.
+  std::vector<schema::Rule> new_rules;
+  for (const base::Value& rule_raw : parsed->GetList()) {
+    auto rule = schema::Rule::FromValue(rule_raw);
+    if (!rule.has_value()) {
+      LOG(WARNING) << "query-filter.json parse error: " << rule.error();
+    } else {
+      new_rules.push_back(std::move(rule.value()));
```
cdesouza-chromium: nit: `emplate_back`

components/query_filter/browser/query_filter_data.h
```diff
@@ -46,21 +34,26 @@ class QueryFilterData {
...
+
+  // A test only method to reset the |rules_set_| and |version_|.
+  void ResetRulesForTesting() noexcept;
 
  private:
   friend class base::NoDestructor<QueryFilterData>;
   QueryFilterData();
 
-  std::vector<QueryFilterRule> rules_;
+  schema::Ruleset rules_set_;
```
cdesouza-chromium: Nit: doc comments for these fields.

components/query_filter/browser/query_filter_data.cc
```diff
@@ -46,8 +31,8 @@ void QueryFilterData::UpdateVersion(base::Version version) {
   version_ = std::move(version);
 }
 
-const std::vector<QueryFilterRule>& QueryFilterData::rules() const {
-  return rules_;
+const schema::Ruleset& QueryFilterData::rules_set() const {
```
cdesouza-chromium: looking at the places that use this function, they always do this:

```cxx
rules_set().rules
```

We may as well just have

```
const std::vector<schema::Rule>& rules() const { return rules_set_.rules(); }
```

This also make clear that rather than having a member `rule_set_`, we possibly should for now have only a `rules` member.

I address though in another comment the possibility of just deleting `Ruleset` entirely, so check that out first.

components/query_filter/browser/query_filter_data.h
```diff
@@ -46,21 +34,26 @@ class QueryFilterData {
...
   bool PopulateDataFromComponent(std::string_view json_data);
 
-  // Resets the |rules_| to an empty vector and the |version_| to empty.
-  void ResetRulesForTesting();
-
   // Returns the current set of query filter rules.
-  // Returns an empty vector if no rules are available.
-  const std::vector<QueryFilterRule>& rules() const;
+  // Returns an empty ruleset if no rules are available.
+  const schema::Ruleset& rules_set() const;
```
cdesouza-chromium: Accessors of this nature should be inlined, meaning their implementation is placed in the header.

cdesouza-chromium (commented):
Nearly there. Thanks for providing this PR so readily.

components/query_filter/browser/query_filter_data.h
r0hit0303: Done

components/query_filter/browser/query_filter_data.cc
r0hit0303: Done

components/query_filter/browser/query_filter_data.h
r0hit0303: Done

components/query_filter/browser/query_filter_data.cc
r0hit0303: Done

components/query_filter/browser/query_filter_data.cc
r0hit0303: Done

components/query_filter/browser/query_filter_component_installer_unittest.cc
r0hit0303: Done

components/query_filter/browser/query_filter_component_installer_unittest.cc
r0hit0303: Done

components/query_filter/browser/BUILD.gn
r0hit0303: Done

components/query_filter/browser/query_filter_data_unittest.cc
r0hit0303: Done

components/query_filter/browser/query_filter_data.cc
r0hit0303: Thanks for the clarification. Deleted the Ruleset and relying on `std::vector<schema::Rule> rules_`. 
Done

r0hit0303:
Thanks @fmarier and @cdesouza-chromium  for your thorough review. I've added all the requested changes in the latest patch-set. (Please also ignore the last few messages, I somehow managed to screw-up the rebasing process but everything should be fine now)

components/query_filter/browser/query_filter_component_installer_unittest.cc
```diff
@@ -120,47 +145,62 @@ TEST_F(QueryFilterComponentInstallerTest, NoRegisterWhenCUSIsNull) {
...
-  EXPECT_EQ("", query_filter::QueryFilterData::GetInstance()->GetVersion());
+  EXPECT_EQ("", GetQueryFilterVersion());
   // Verify also the rules are empty before the component is ready
-  EXPECT_TRUE(query_filter::QueryFilterData::GetInstance()->rules().empty());
+  const auto& rules = GetQueryFilterRules();
+  EXPECT_TRUE(rules.empty());
 
   // Initiate component ready which would load the json file and populate the
   // query filter data.
   component_updater::QueryFilterComponentInstallerPolicy policy;
```
netzenbot, line 169: This `RunUntil([&]() { return GetQueryFilterVersion() == ""; })` passes immediately because the version is already empty before `ComponentReady` runs — it doesn't actually wait for the async task to complete. The code comment even acknowledges this: "the RunUntil may exit before the ComponentReady finished the async task inside." A proper fix per [TA-003](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-async.md#TA-003) is to add a callback to `ComponentReady`/`OnQueryFilterFileRead` so the test can synchronize on actual completion, as captured in the TODO. Until that's done, the assertions after the `RunUntil` are racing with the background task.

components/query_filter/common/query_filter_schema.idl
```diff
@@ -0,0 +1,27 @@
...
+  dictionary Rule {
+    // The list of domains (wildcard supported) to consider to remove the query params |params|.  
+    DOMString[] include;
+    // The list of domains (wildcard supported) on which NOT to exclude the query params |params|. 
+    DOMString[] exclude;
+    // The list of query parameters which are subjected to query filtering.
+    DOMString[] params;
+  };
+
+  // The main collection of Rule objects to be enforced in the client.
```
netzenbot: The `Ruleset` dictionary is still present in the IDL but isn't used anywhere — only `schema::Rule` is used directly in `query_filter_data.h` and `.cc`. @cdesouza-chromium asked you to delete it and you replied "Done", but it's still here. Please remove it to avoid dead generated code. ([Delete Dead Documentation](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-002))

components/query_filter/browser/query_filter_data_unittest.cc
```diff
@@ -197,4 +219,84 @@ TEST_F(QueryFilterDataTest, TestCheckGeneralRulesPopulation) {
...
+  {
+    "include": ["*://*/*"],
+    "exclude": [],
+    "params": ["__hsfp", "gclid", "fbclid"]
+  }
+]
+)json";
+
+  EXPECT_TRUE(instance()->PopulateDataFromComponent(kJson1));
+  const auto& rules = GetQueryFilterRules();
```
netzenbot, line 234: Use `ASSERT_EQ(1UL, rules.size())` instead of `EXPECT_EQ` here and in the other new tests (`TestSequentialRulesUpdate_FirstGood_SecondBad_KeepsFirst`, `TestOneBadRuleEntryIgnoresThatRule`, `TestOnNonSupportedFields_OnlyNonSupportedEntryItemIsIgnored`, `TestCheckGeneralRulesPopulation_IsCorrect`). `EXPECT_EQ` is non-fatal, so the test continues to `rules[0]` even if the size check fails, causing a bounds-overread crash rather than a clean failure. [TI-034](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-034)

components/query_filter/browser/query_filter_component_installer_unittest.cc
r0hit0303: Ack. This will be taken care in the next CL as stated to limit the scope of change to this CL.

components/query_filter/common/query_filter_schema.idl
r0hit0303: Missed this. Done

components/query_filter/browser/query_filter_data_unittest.cc
r0hit0303: Done

components/query_filter/common/schema.idl
```diff
@@ -0,0 +1,21 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
```
cdesouza-chromium, line 1: As I mentioned before, rename this source to `schema`. No need to repeat `query_filter` twice in the path.

cdesouza-chromium (approved):
LGTM % name change.

components/query_filter/common/schema.idl
r0hit0303: Done

components/query_filter/common/schema.idl
```diff
@@ -0,0 +1,21 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
```
fmarier:
```suggestion
 * You can obtain one at https://mozilla.org/MPL/2.0/. */
```

fmarier (approved):

netzenbot (approved):

PR #35951, opened 2026-04-27
[iOS] Hide shortcut button for features disabled by Origin

### LINKED ISSUE #55006: [iOS] Shortcut Button displays shortcuts for features that may be disabled by admin policies
If you disable a feature via admin policy the shortcut button in the toolbar can still display a shortcut for it even if said shortcut does nothing.

kylehickinson:
This updates hides both the options that are disabled in the shortcut button picker as well as the shortcut button itself if the option was selected prior to disabling the feature with Origin

Resolves https://github.com/brave/brave-browser/issues/55006

### Test
- Set the shortcut button to an Origin feature such as Leo
- Verify shortcut button list is typical prior to purchasing Brave Origin
- Purchase Origin and keep all features disabled
- Verify shortcut button disappears from toolbar since it was originally set to Leo
- Re-enable Leo in Origin settings and re-launch
- Verify shortcut button appears again with Leo
- Long press shortcut button in toolbar and verify only enabled features are listed
- Visit Shortcut Button picker in settings and verify only enabled features are listed

### DIFF
--- a/ios/brave-ios/Sources/Brave/Extensions/WidgetShortcutExtension.swift
+++ b/ios/brave-ios/Sources/Brave/Extensions/WidgetShortcutExtension.swift
@@ -3,24 +3,44 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this file,
 // You can obtain one at https://mozilla.org/MPL/2.0/.
 
+import BraveCore
 import BraveStrings
 import BraveWidgetsModels
 import DesignSystem
 import Foundation
+import OrderedCollections
 import Strings
 import UIKit
 
 extension WidgetShortcut {
-  static let eligibleButtonShortcuts: [WidgetShortcut] = [
-    .bookmarks,
-    .history,
-    .downloads,
-    .playlist,
-    .wallet,
-    .braveNews,
-    .braveLeo,
-    .askBrave,
-  ]
+  static func eligibleButtonShortcuts(
+    prefs: any PrefService,
+    isWalletAvailable: Bool
+  ) -> OrderedSet<WidgetShortcut> {
+    var options = OrderedSet<WidgetShortcut>([
+      .bookmarks,
+      .history,
+      .downloads,
+      .playlist,
+      .wallet,
+      .braveNews,
+      .braveLeo,
+      .askBrave,
+    ])
+    if !prefs.isPlaylistAvailable {
+      options.remove(.playlist)
+    }
+    if !prefs.isBraveNewsAvailable {
+      options.remove(.braveNews)
+    }
+    if !isWalletAvailable {
+      options.remove(.wallet)
+    }
+    if !AIChatUtils.isAIChatEnabled(for: prefs) {
+      options.remove(.braveLeo)
+    }
+    return options
+  }
 
   var displayString: String {
     switch self {
--- a/ios/brave-ios/Sources/Brave/Frontend/Browser/Toolbars/UrlBar/TopToolbarView.swift
+++ b/ios/brave-ios/Sources/Brave/Frontend/Browser/Toolbars/UrlBar/TopToolbarView.swift
@@ -8,6 +8,7 @@ import BraveWidgetsModels
 import Combine
 import Data
 import DesignSystem
+import OrderedCollections
 import Preferences
 import Shared
 import SnapKit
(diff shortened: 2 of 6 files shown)


netzenbot (approved):

ios/brave-ios/Sources/Brave/Frontend/Settings/Features/ShortcutButtonPickerView.swift
```diff
@@ -14,12 +16,19 @@ struct ShortcutButtonPickerView: View {
   @ObservedObject private var selectedShortcut = Preferences.General.toolbarShortcutButton
   @Environment(\.dismiss) var dismiss
```
netzenbot, line 18: nit: `profile` is only used for `profile.prefs` here. Prefer passing `prefs: any PrefService` directly instead of the whole `Profile` object. See [ARCH-003](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-003).

ios/brave-ios/Sources/Brave/Frontend/Browser/Toolbars/UrlBar/TopToolbarView.swift
```diff
@@ -68,7 +72,11 @@ class TopToolbarView: UIView, ToolbarProtocol {
 
   var helper: ToolbarHelper?
 
-  weak var delegate: TopToolbarDelegate?
+  weak var delegate: TopToolbarDelegate? {
```
netzenbot, line 75: nit: The `delegate` `didSet` calls `updateViewsForOverlayModeAndToolbarChanges()` unconditionally, including when the delegate is set to nil or reassigned to the same value. Consider guarding with `guard newValue !== oldValue else { return }` to avoid unnecessary UI updates. ([ARCH-051](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-051))

ios/brave-ios/Sources/Brave/Extensions/WidgetShortcutExtension.swift
```diff
@@ -3,24 +3,46 @@
...
+      .bookmarks,
+      .history,
+      .downloads,
+      .playlist,
+      .wallet,
+      .braveNews,
+      .braveLeo,
+      .askBrave,
+    ])
+    if FeatureList.kBraveOrigin.enabled {
```
kylehickinson: I should actually probably remove this Origin check as technically standard admin policies can disable these features even without Origin

nuo-xu: what is standard admin policies? i guess will these get misaligned by any chance?

nuo-xu (approved):

ios/brave-ios/Sources/Brave/Extensions/WidgetShortcutExtension.swift
kylehickinson: Managed preferences/enterprise policies, which is what Origin uses under the hood

kjozwiak:

```
Brave | 1.92.29 Chromium: 148.0.7778.96 (Official Build)nightly (64-bit) 
--- | ---
Revision | acc7e8e98d87...
OS | iOS
```

Using the STR/Cases outlined via https://github.com/brave/brave-core/pull/35951#issue-4338653265, ensured that `Shortcut` icons that belong to features gated by `Brave Origin` are removed if they're being currently used as per the following:

* ensured that the `shortcut` menu/setting's page is being updated correctly as features are enabled/disabled
* ensured that `shortcuts` are being removed when a corresponding feature is disabled via `Origin`
* ensured that the previous `shortcut` is retained/restores if a feature is re-enabled and a new `shortcut` wasn't selected
* ensured that a previous `shortcut` corresponding to a `Origin` doesn't replace current `shortcut` when feature enabled

There's an issue where a restart is needed to remove the `shortcut` from the toolbar that's associated with a `Origin` feature that was disabled. @kylehickinson is working on a PR to add a `Restart Required` message similar to `Android` as per https://github.com/brave/brave-core/pull/36220.

https://github.com/user-attachments/assets/3f8511dd-86bc-4782-8a68-fff43d1c94a5

PR #35953, opened 2026-04-27
[AI Chat] Migrate `ModelService` to async `OSCrypt`

### LINKED ISSUE #55011: [AI Chat] Migrate `ModelService` to async `OSCrypt`
We are using `OSCrypt::EncryptString` and `OSCrypt::DecryptString` currently which is deprecated in upcoming Chromium upgrade (C149).

yrliou:
`ModelService` now takes an `os_crypt_async::OSCryptAsync*` in its
constructor and obtains its `Encryptor` asynchronously via
`GetInstance()`, replacing the synchronous `OSCrypt::EncryptString` /
`OSCrypt::DecryptString` free functions which are deprecated.

The model list is loaded synchronously in the constructor so callers
(e.g. `ConversationHandler::InitEngine`) can resolve a default model
immediately. Custom-model API keys decrypt to empty strings until
`OnEncryptorReady()` arrives; that callback then refreshes the
`api_key` field on each already-loaded custom model in place via
`RefreshCustomModelApiKeys()` and notifies observers via
`OnModelListUpdated()`, which propagates to engines through
`ConversationHandler::OnModelListUpdated()` ->
`engine_->UpdateModelOptions()`. Requests without encryptor ready will
go out with an empty API key which server would reject. Subsequent
requests succeed once the engine refreshes.

The desktop factory moves from `components/ai_chat/content/browser/`
to `browser/ai_chat/` because `g_browser_process->os_crypt_async()`
requires `//chrome/browser:browser_process`, which `//brave/components`
cannot depend on. The iOS factory uses
`GetApplicationContext()->GetOSCryptAsync()`.

Existing unit tests inject a sync test instance via
`GetTestOSCryptAsyncForTesting(/*is_sync_for_unittests=*/true)`,
matching the convention used across upstream Chromium. A new
`ModelServiceAsyncEncryptorTest` exercises the empty -> populated
api_key transition explicitly using the async test-instance variant.

Note that there's an existing issue where we do not handle error during encrypt/decrypt but fail silently, this PR does not change this, it is out of our migration scope here.

Resolves https://github.com/brave/brave-browser/issues/55011


yrliou:
This PR triggers 
```
# mojo interfaces should be implemented directly by the underlying classes
browser/ai_chat/ai_chat_settings_helper.* 
```

code owner rule, but I'm merely updating an existing header include path in ai_chat_settings_helper.h.

components/ai_chat/core/browser/model_service.cc
```diff
@@ -558,23 +560,43 @@ const std::vector<mojom::ModelPtr>& GetLeoModels() {
...
+  // `encryptor_` is unset between `ModelService` construction and
+  // `OnEncryptorReady()` firing; `InitModels()` runs synchronously in the
+  // constructor and reaches here, so we tolerate the missing-encryptor case
+  // and return an empty key. Once the encryptor arrives,
+  // `OnEncryptorReady()` refreshes keys in place via
+  // `RefreshCustomModelApiKeys()` so callers see the real value. An empty
+  // `encoded_api_key` (e.g. the user never set one) also short-circuits.
+  if (!encryptor_ || encoded_api_key.empty()) {
     return std::string();
   }
```
petemill, line 590: So because we are expected to always have a model list, we're ok having a model with an invalid config (i.e. without an api key value) because we don't expect any requests to go out before `OnModelListUpdated` is called, which will provide any engines with the valid decrypted API key once the encryptor is ready? If so, it's a little brittle but fine for now because we probably need to make the whole model-fetching more async-aware when we move to fetching the built-in models from an endpoint. Whereby perhaps we would not allow engines to be created before the models are done initializing? 

If that's all true, and you agree it's a little brittle, we could leave a comment here about that, along with a github issue.

components/ai_chat/core/browser/BUILD.gn
```diff
@@ -131,7 +131,6 @@ static_library("browser") {
     "//components/keyed_service/core",
```
netzenbot, line 134: `model_service.h` now includes `components/os_crypt/async/common/encryptor.h` and stores `std::optional<os_crypt_async::Encryptor>` as a member, making `//components/os_crypt/async/common` a header-visible dependency. Per [BS-024](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-024), it should be moved from `deps` to `public_deps` in `components/ai_chat/core/browser/BUILD.gn`.

components/ai_chat/core/browser/model_service_unittest.cc
```diff
@@ -579,4 +606,97 @@ TEST_F(ModelServiceTest, GetCustomModels) {
...
+  run_loop.Run();
+
+  // Post-encryptor: each model's api_key matches expectation. Order matches
+  // insertion order in both prefs and `models_`.
+  auto after = service->GetCustomModels();
+  ASSERT_EQ(after.size(), 3u);
+  EXPECT_EQ(after[0]->options->get_custom_model_options()->model_request_name,
+            "model-alpha");
+  EXPECT_EQ(after[0]->options->get_custom_model_options()->api_key,
+            "key-alpha");
```
netzenbot, line 695: The post-encryptor section dereferences `get_custom_model_options()` on each element without first calling `ASSERT_TRUE(after[N]->options->is_custom_model_options())`. If the option variant is ever wrong (e.g. a regression changes the type), the test crashes rather than failing cleanly. Add `ASSERT_TRUE` guards matching the pattern used in the pre-encryptor loop. [TI-034](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-034)

components/ai_chat/core/browser/model_service_unittest.cc
```diff
@@ -579,4 +606,97 @@ TEST_F(ModelServiceTest, GetCustomModels) {
...
+// `OnModelListUpdated()` fires. Includes a model with an empty `api_key` to
+// cover the no-op refresh path.
+TEST_F(ModelServiceAsyncEncryptorTest,
+       CustomModelApiKeyRefreshedAfterEncryptor) {
+  PreloadCustomModelsInPrefs({
+      {"model-alpha", "key-alpha"},
+      {"model-beta", std::string()},  // empty key — should remain empty
+      {"model-gamma", "key-gamma"},
+  });
+
```
netzenbot, line 669: nit: `service_` is always non-null (dereferenced unconditionally in the destructor) — prefer `raw_ref<ModelService>` over `raw_ptr`. Production code in `ollama_model_fetcher.h` already uses `raw_ref<ModelService>` for this pattern. [CSM-036](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-memory.md#CSM-036)

components/ai_chat/core/browser/model_service.cc
```diff
@@ -558,23 +560,43 @@ const std::vector<mojom::ModelPtr>& GetLeoModels() {
   return *kModels;
```
netzenbot, line 560: nit: The private method definitions in the `.cc` don't follow the declaration order in `.h`. The header lists `OnEncryptorReady` first, then `InitModels`, `RefreshCustomModelApiKeys`, `EncryptAPIKey`, `DecryptAPIKey`, `CustomModelToPrefDict`; but the `.cc` defines `EncryptAPIKey`/`DecryptAPIKey`/`CustomModelToPrefDict` before the constructor and `OnEncryptorReady`/`RefreshCustomModelApiKeys` after it. Please reorder to match the header. See [CS-017](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-017).

components/ai_chat/core/browser/model_service.cc
```diff
@@ -636,10 +664,47 @@ ModelService::ModelService(PrefService* prefs_service)
...
+  CHECK_DEREF(os_crypt_async)
+      .GetInstance(base::BindOnce(&ModelService::OnEncryptorReady,
+                                  weak_ptr_factory_.GetWeakPtr()));
 }
 
 ModelService::~ModelService() = default;
 
+void ModelService::OnEncryptorReady(os_crypt_async::Encryptor encryptor) {
+  encryptor_.emplace(std::move(encryptor));
+  // Refresh custom-model API keys in-place. Notifies observers via
```
netzenbot, line 674: nit: Prefer `observers_.Notify(&Observer::OnModelListUpdated)` over the manual `for (auto& obs : observers_)` loop. See [CS-056](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-056).

components/ai_chat/core/browser/model_service.cc
yrliou: Yes, our current expectation is we always have a model list, this PR intentionally does not change everything to async design because it isn't really the scope for migrating to async OSCrypt.
And as mentioned in the model_service.h comment, in theory there is only a very brief window between ModelService construction and OnEncryptorReady() that if a request ever goes out in this short period of window (unlikely in practice because it usually triggered by UI actions but we don't really enforce this won't ever happen), it would just not have a valid API key and gets rejected by remote, but back to normal once encryptor is ready and triggered updates of in-memory model list.
I think it's a safe thing to do and keep the migration scope small.

Is the comment in the model_service.h not enough for documentation purpose? I think it describes ModelService's current behavior.

components/ai_chat/core/browser/model_service.cc
```diff
@@ -558,23 +560,43 @@ const std::vector<mojom::ModelPtr>& GetLeoModels() {
...
+std::string ModelService::EncryptAPIKey(const std::string& api_key) const {
+  // `DecryptAPIKey()` tolerates a missing encryptor because the empty result
+  // gets refreshed in place by `RefreshCustomModelApiKeys()` once the
+  // encryptor arrives. The write path has no analogous refresh: pref entries
+  // are written once at `AddCustomModel()` / `SaveCustomModel()` time (via
+  // `CustomModelToPrefDict()`) and never re-encrypted. So a write before the
+  // encryptor is ready is a programmer error (callers should never reach
+  // here in that state) and we CHECK to surface it loudly rather than fall
+  // back to a silent no-op.
+  CHECK(encryptor_);
```
darkdh: This assumes users won't be able to add custom model before `OnEncryptorReady`. 
But what about `OllamaModelFetcher::OnModelDetailsFetched` which doesn't need API keys but it still calls `AddCustomModel` which then call `EncryptAPIKey`? There is a race window between "Ollama HTTP response arrives" and "OnEncryptorReady fires" especially on a slow keychain unlock ex. locked macOS keychain.

darkdh: Even users manually adding model flow would hit this CHECK if they need to unlock macOS keychain every time. We can keep this current silent fail as it-is without usingCHECK and open a follow-up issue to report failure since this PR is focusing on migration.

components/ai_chat/core/browser/model_service.cc
yrliou: Intentionally not following the declaration order so the diff on existing methods (Encrypt/Decrypt/CustomModelToPrefDict) is clear.

components/ai_chat/core/browser/model_service.cc
yrliou: https://github.com/brave/brave-core/pull/35953/commits/7ba01d50e5959bf3b9f56c8412ff313ce4bfe00f
Move the encryptor_ check to be together with existing encrypt/decrypt result check.
In theory I think it doesn't make things any worse than our existing behavior, we are already silently failing before if we ever can't encrypt/decrypt properly. (API key would just be empty)
Existing behavior is definitely not great, hence we have the existing TODO comment in the code, created https://github.com/brave/brave-browser/issues/55033 to track this, this PR would only focus on migrating to async OSCrypt.

components/ai_chat/core/browser/model_service.cc
yrliou: https://github.com/brave/brave-core/pull/35953/commits/7ba01d50e5959bf3b9f56c8412ff313ce4bfe00f

components/ai_chat/core/browser/BUILD.gn
yrliou: https://github.com/brave/brave-core/pull/35953/commits/7dc54155452b97aac48c3f72233f40ca9f3176ae

components/ai_chat/core/browser/model_service.cc
yrliou: https://github.com/brave/brave-core/pull/35953/commits/6c6c69452daa17c6c0062b2f173d1387fa6c058c

components/ai_chat/core/browser/model_service_unittest.cc
yrliou: https://github.com/brave/brave-core/pull/35953/commits/d349e38e306ed39b3b037eadeb93d69ead544e06

components/ai_chat/core/browser/model_service_unittest.cc
yrliou: https://github.com/brave/brave-core/pull/35953/commits/d349e38e306ed39b3b037eadeb93d69ead544e06

components/ai_chat/core/browser/model_service_unittest.cc
```diff
@@ -579,4 +607,99 @@ TEST_F(ModelServiceTest, GetCustomModels) {
...
+  ASSERT_TRUE(after[0]->options->is_custom_model_options());
+  EXPECT_EQ(after[0]->options->get_custom_model_options()->model_request_name,
+            "model-alpha");
+  EXPECT_EQ(after[0]->options->get_custom_model_options()->api_key,
+            "key-alpha");
+  ASSERT_TRUE(after[1]->options->is_custom_model_options());
+  EXPECT_EQ(after[1]->options->get_custom_model_options()->model_request_name,
+            "model-beta");
+  EXPECT_EQ(after[1]->options->get_custom_model_options()->api_key,
+            std::string());
```
darkdh: nit: just EXPECT_TRUE(after[1]->options->get_custom_model_options()->api_key.empty())

components/ai_chat/core/browser/model_service_unittest.cc
```diff
@@ -579,4 +607,99 @@ TEST_F(ModelServiceTest, GetCustomModels) {
...
+  // subject reads them at construction. Uses a temporary `ModelService` on
+  // the same `OSCryptAsync` to write the entries (encryption requires the
+  // encryptor). Each input is `(model_request_name, api_key)`.
+  void PreloadCustomModelsInPrefs(
+      const std::vector<std::pair<std::string, std::string>>& models) {
+    auto service =
+        std::make_unique<ModelService>(&pref_service_, os_crypt_async_.get());
+    base::RunLoop run_loop;
+    ScopedModelListReadyObserver observer(*service, run_loop.QuitClosure());
+    run_loop.Run();
```
darkdh, line 635: can we add a comment explain why we need to wait for `OnModelListUpdated` here?

darkdh (approved):

components/ai_chat/core/browser/model_service_unittest.cc
yrliou: https://github.com/brave/brave-core/pull/35953/commits/8411c0e8ce654f3e425476aa67b4bbb77ebac55f

components/ai_chat/core/browser/model_service_unittest.cc
yrliou: https://github.com/brave/brave-core/pull/35953/commits/8411c0e8ce654f3e425476aa67b4bbb77ebac55f

petemill (approved):

netzenbot (approved):

goodov (approved):
factory changes lgtm

PR #35954, opened 2026-04-27
[brockit] Provide logs for `temp_dir.cleanup()` failure

cdesouza-chromium:
This chnage is adding logs to this `temp_dir.cleanup()` call, to provide
more detailed errors when this call fails. This call has started failing
recently on TeamCity, and it is not very clear why.

Bug https://github.com/brave/brave-browser/issues/55013


mkarolin (approved):
LGTM

PR #35958, opened 2026-04-28
[plaster] Migrate `DeviceInfo` override

### LINKED ISSUE #55018: Migrate `DeviceInfo` to a plaster
Migrate `BraveDeviceInfo` to a plaster, as the current way as it works is very laborious to maintain.

cdesouza-chromium:
This PR rewrites our overrides relating to `DeviceInfo` specifically
into plaster. This involves dropping `BraveDeviceInfo` and implementing
the functionality in terms of extending `DeviceInfo`

Resolves https://github.com/brave/brave-browser/issues/55018


netzenbot (approved):

rewrite/components/sync_device_info/device_info.h.toml
```diff
@@ -0,0 +1,54 @@
...
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+[[substitution]]
+description = '''Add is_self_delete_supported to the ctor arg list
+
+Adding `is_self_delete_supported` so it can be used to initialise
+`is_self_delete_supported_`.
+
+This regex matches with the constructor declaration, by checking that there is
+a `std::string` somewhere in the list, and adds `is_self_delete_supported` as
```
AlexeyBarabash, line 13: > checking that there is a `std::string` somewhere in the list
this is required because there are also deleted constructors

chromium_src/components/sync_device_info/device_info_sync_bridge.cc
```diff
@@ -59,7 +58,7 @@ std::unique_ptr<BraveDeviceInfo> BraveSpecificsToModel(
   }
   // The code is duplicated from SpecificsToModel by intent to avoid use of
   // extra patch
-  return std::make_unique<BraveDeviceInfo>(
+  return std::make_unique<DeviceInfo>(
```
AlexeyBarabash, line 61: It looks to me `BraveSpecificsToModel` can be dropped also in a favor of plaster/patch which adds 
```
 specifics.has_brave_fields() &&
          specifics.brave_fields().has_is_self_delete_supported() &&
          specifics.brave_fields().is_self_delete_supported());
```
into the original
`DeviceInfo SpecificsToModel(const DeviceInfoSpecifics& specifics)`

The purpose of `BraveSpecificsToModel` was to avoid raw patches as possible.

The difference at the top of `BraveSpecificsToModel` vs `SpecificsToModel` is the result of the fact I missed when it changed at the upstream.

AlexeyBarabash (approved):
++

chromium_src/components/sync_device_info/device_info_sync_bridge.cc
AlexeyBarabash: can be done as a follow up PR to keep the current small

bridiver (approved):

PR #35963, opened 2026-04-28
[Polkadot] Add parachains support

### LINKED ISSUE #44750: [Polkadot] Polkadot parachains support
### Description

Core part for parachains support

### Steps to reproduce

-

### Actual result

-

### Expected result

-

### Reproduces how often

Easily reproduced

-

### Channel information

- [ ] release (stable)
- [ ] beta
- [ ] nightly

### Reproducibility

### Miscellaneous information

_No response_

cypt4:
Add Polkadot mainnet\testnet parachains.
Add supportedKeyrings field to persist for NetworkInfo.
This is required since each polkadot keyring can have multiple
parachains.
Updated Deposit modal so showed address uses patachain ss58 prefix.
Resolves https://github.com/brave/brave-browser/issues/44750
Sec review: https://github.com/brave/reviews/issues/2192

### DIFF
--- a/components/brave_wallet/browser/blockchain_list_parser.cc
+++ b/components/brave_wallet/browser/blockchain_list_parser.cc
@@ -565,8 +565,7 @@ bool ParseChainList(const std::string& json, ChainList* result) {
       continue;
     }
     network->coin = mojom::CoinType::ETH;
-    network->supported_keyrings =
-        GetSupportedKeyringsForNetwork(network->coin, network->chain_id);
+    network->supported_keyrings = {mojom::KeyringId::kDefault};
 
     result->push_back(std::move(network));
   }
--- a/components/brave_wallet/browser/keyring_service.cc
+++ b/components/brave_wallet/browser/keyring_service.cc
@@ -1280,6 +1280,10 @@ void KeyringService::CreateKeyrings(const KeyringSeed& keyring_seed) {
         base::span(keyring_seed.polkadot_seed).first<kPolkadotSeedSize>();
     polkadot_mainnet_keyring_ = std::make_unique<PolkadotKeyring>(
         polkadot_seed, KeyringId::kPolkadotMainnet, is_address_allowed);
+  }
+  if (IsKeyringEnabled(KeyringId::kPolkadotTestnet)) {
+    auto polkadot_seed =
+        base::span(keyring_seed.seed).first<kPolkadotSeedSize>();
     polkadot_testnet_keyring_ = std::make_unique<PolkadotKeyring>(
         polkadot_seed, KeyringId::kPolkadotTestnet, is_address_allowed);
   }
@@ -1876,7 +1880,7 @@ mojom::AccountInfoPtr KeyringService::ImportPolkadotAccountSync(
     const std::string& password,
     const std::string& network) {
   if (account_name.empty() || json_export.empty() || password.empty() ||
-      IsLockedSync() || !IsPolkadotNetwork(network)) {
+      IsLockedSync() || !IsPolkadotRelayNetwork(network)) {
     return nullptr;
   }
   CHECK(encryptor_);
@@ -2182,7 +2186,6 @@ std::optional<std::string> KeyringService::AddHDAccountForKeyringInternal(
   if (auto* keyring = GetKeyring<PolkadotKeyring>(keyring_id)) {
     return keyring->AddNewHDAccount(index);
   }
-
   return std::nullopt;
 }
 
--- a/components/brave_wallet/browser/keyring_service.cc
+++ b/components/brave_wallet/browser/keyring_service.cc
@@ -1283,7 +1283,7 @@ void KeyringService::CreateKeyrings(const KeyringSeed& keyring_seed) {
   }
   if (IsKeyringEnabled(KeyringId::kPolkadotTestnet)) {
     auto polkadot_seed =
-        base::span(keyring_seed.seed).first<kPolkadotSeedSize>();
+        base::span(keyring_seed.polkadot_seed).first<kPolkadotSeedSize>();
     polkadot_testnet_keyring_ = std::make_unique<PolkadotKeyring>(
         polkadot_seed, KeyringId::kPolkadotTestnet, is_address_allowed);
   }
--- a/components/brave_wallet/browser/network_manager.cc
+++ b/components/brave_wallet/browser/network_manager.cc
@@ -65,27 +65,30 @@ constexpr auto kEip1559ForKnownChains =
 
 constexpr auto kChainSubdomains =
     base::MakeFixedFlatMap<std::string_view, std::string_view>(
-        {{mojom::kMainnetChainId, "ethereum-mainnet"},
-         {mojom::kSepoliaChainId, "ethereum-sepolia"},
-         {mojom::kPolygonMainnetChainId, "polygon-mainnet"},
-         {mojom::kOptimismMainnetChainId, "optimism-mainnet"},
-         {mojom::kBaseMainnetChainId, "base-mainnet"},
-         {mojom::kAvalancheMainnetChainId, "avalanche-mainnet"},
-         {mojom::kBnbSmartChainMainnetChainId, "bsc-mainnet"},
+        {
+            {mojom::kMainnetChainId, "ethereum-mainnet"},
+            {mojom::kSepoliaChainId, "ethereum-sepolia"},
+            {mojom::kPolygonMainnetChainId, "polygon-mainnet"},
+            {mojom::kOptimismMainnetChainId, "optimism-mainnet"},
+            {mojom::kBaseMainnetChainId, "base-mainnet"},
+            {mojom::kAvalancheMainnetChainId, "avalanche-mainnet"},
+            {mojom::kBnbSmartChainMainnetChainId, "bsc-mainnet"},
 
-         // SVM chains.
-         {mojom::kSolanaMainnet, "solana-mainnet"},
+            // SVM chains.
+            {mojom::kSolanaMainnet, "solana-mainnet"},
 
-         // Bitcoin chains.
-         {mojom::kBitcoinMainnet, "bitcoin-mainnet"},
+            // Bitcoin chains.
+            {mojom::kBitcoinMainnet, "bitcoin-mainnet"},
 
-         // Cardano chains.
-         {mojom::kCardanoMainnet, "cardano-mainnet"},
-         {mojom::kCardanoTestnet, "cardano-preprod"},
+            // Cardano chains.
+            {mojom::kCardanoMainnet, "cardano-mainnet"},
+            {mojom::kCardanoTestnet, "cardano-preprod"},
 
-         // Polkadot chains.
-         {mojom::kPolkadotMainnet, "polkadot-mainnet"},
-         {mojom::kPolkadotTestnet, "polkadot-westend"}},
+            // Polkadot mainnet chains.
+            {mojom::kPolkadotMainnet, "polkadot-mainnet"},
+            // Polkadot testnet chains.
+            {mojom::kPolkadotTestnet, "polkadot-westend"},
+        },
         CaseInsensitiveCompare());
 
 constexpr char kGanacheLocalhostURL[] = "http://localhost:7545/";
@@ -174,7 +177,7 @@ const mojom::NetworkInfo* GetEthMainnet() {
        "Ethereum",
        18,
        coin,
-       GetSupportedKeyringsForNetwork(coin, chain_id)});
+       GetSupportedKeyringsForKnownNetwork(coin, chain_id)});
   return network_info.get();
 }
 
@@ -193,7 +196,7 @@ const mojom::NetworkInfo* GetPolygonMainnet() {
        "MATIC",
        18,
        coin,
-       GetSupportedKeyringsForNetwork(coin, chain_id)});
+       GetSupportedKeyringsForKnownNetwork(coin, chain_id)});
   return network_info.get();
 }
 
@@ -212,7 +215,7 @@ const mojom::NetworkInfo* GetBscMainnet() {
        "BNB",
        18,
        coin,
-       GetSupportedKeyringsForNetwork(coin, chain_id)});
+       GetSupportedKeyringsForKnownNetwork(coin, chain_id)});
   return network_info.get();
 }
 
@@ -231,7 +234,7 @@ const mojom::NetworkInfo* GetAvalancheMainnet() {
        "Avalanche",
        18,
        coin,
-       GetSupportedKeyringsForNetwork(coin, chain_id)});
+       GetSupportedKeyringsForKnownNetwork(coin, chain_id)});
   return network_info.get();
 }
 
@@ -250,7 +253,7 @@ const mojom::NetworkInfo* GetOptimismMainnet() {
        "Ether",
        18,
        coin,
-       GetSupportedKeyringsForNetwork(coin, chain_id)});
+       GetSupportedKeyringsForKnownNetwork(coin, chain_id)});
   return network_info.get();
 }
 
@@ -269,7 +272,7 @@ const mojom::NetworkInfo* GetBaseMainnet() {
        "Ether",
        18,
        coin,
-       GetSupportedKeyringsForNetwork(coin, chain_id)});
+       GetSupportedKeyringsForKnownNetwork(coin, chain_id)});
   return network_info.get();
 }
 
@@ -288,7 +291,7 @@ const mojom::NetworkInfo* GetNeonEVMMainnet() {
        "Neon",
        18,
        coin,
-       GetSupportedKeyringsForNetwork(coin, chain_id)});
+       GetSupportedKeyringsForKnownNetwork(coin, chain_id)});
   return network_info.get();
 }
 
@@ -307,7 +310,7 @@ const mojom::NetworkInfo* GetSepoliaTestNetwork() {
        "Ethereum",
        18,
        coin,
-       GetSupportedKeyringsForNetwork(coin, chain_id)});
+       GetSupportedKeyringsForKnownNetwork(coin, chain_id)});
   return network_info.get();
 }
 
@@ -326,7 +329,7 @@ const mojom::NetworkInfo* GetEthLocalhost() {
        "Ethereum",
        18,
        coin,
-       GetSupportedKeyringsForNetwork(coin, chain_id)});
+       GetSupportedKeyringsForKnownNetwork(coin, chain_id)});
   return network_info.get();
 }
 
@@ -345,7 +348,7 @@ const mojom::NetworkInfo* GetFilecoinEthereumMainnet() {
        "Filecoin",
        18,
        coin,
-       GetSupportedKeyringsForNetwork(coin, chain_id)});
+       GetSupportedKeyringsForKnownNetwork(coin, chain_id)});
   return network_info.get();
 }
 
@@ -364,7 +367,7 @@ const mojom::NetworkInfo* GetFilecoinEthereumTestnet() {
        "Filecoin",
        18,
        coin,
-       GetSupportedKeyringsForNetwork(coin, chain_id)});
+       GetSupportedKeyringsForKnownNetwork(coin, chain_id)});
   return network_info.get();
 }
 
@@ -403,7 +406,7 @@ const mojom::NetworkInfo* GetSolMainnet() {
        "Solana",
        9,
        coin,
-       GetSupportedKeyringsForNetwork(coin, chain_id)});
+       GetSupportedKeyringsForKnownNetwork(coin, chain_id)});
   return network_info.get();
 }
 
@@ -422,7 +425,7 @@ const mojom::NetworkInfo* GetSolTestnet() {
        "Solana",
        9,
        coin,
-       GetSupportedKeyringsForNetwork(coin, chain_id)});
+       GetSupportedKeyringsForKnownNetwork(coin, chain_id)});
   return network_info.get();
 }
 
@@ -441,7 +444,7 @@ const mojom::NetworkInfo* GetSolDevnet() {
        "Solana",
        9,
        coin,
-       GetSupportedKeyringsForNetwork(coin, chain_id)});
+       GetSupportedKeyringsForKnownNetwork(coin, chain_id)});
   return network_info.get();
 }
 
@@ -461,7 +464,7 @@ const mojom::NetworkInfo* GetSolLocalhost() {
        "Solana",
        9,
        coin,
-       GetSupportedKeyringsForNetwork(coin, chain_id)});
+       GetSupportedKeyringsForKnownNetwork(coin, chain_id)});
   return network_info.get();
 }
 
@@ -492,7 +495,7 @@ const mojom::NetworkInfo* GetFilMainnet() {
        "Filecoin",
        18,
        coin,
-       GetSupportedKeyringsForNetwork(coin, chain_id)});
+       GetSupportedKeyringsForKnownNetwork(coin, chain_id)});
   return network_info.get();
 }
 
@@ -511,7 +514,7 @@ const mojom::NetworkInfo* GetFilTestnet() {
        "Filecoin",
        18,
        coin,
-       GetSupportedKeyringsForNetwork(coin, chain_id)});
+       GetSupportedKeyringsForKnownNetwork(coin, chain_id)});
   return network_info.get();
 }
 
@@ -530,7 +533,7 @@ const mojom::NetworkInfo* GetFilLocalhost() {
        "Filecoin",
        18,
        coin,
-       GetSupportedKeyringsForNetwork(coin, chain_id)});
+       GetSupportedKeyringsForKnownNetwork(coin, chain_id)});
   return network_info.get();
 }
 
@@ -620,7 +623,7 @@ const mojom::NetworkInfo* GetBitcoinMainnet() {
        "Bitcoin",
        8,
        coin,
-       GetSupportedKeyringsForNetwork(coin, chain_id)});
+       GetSupportedKeyringsForKnownNetwork(coin, chain_id)});
   return network_info.get();
 }
 
@@ -639,7 +642,7 @@ const mojom::NetworkInfo* GetBitcoinTestnet() {
        "Bitcoin",
        8,
        coin,
-       GetSupportedKeyringsForNetwork(coin, chain_id)});
+       GetSupportedKeyringsForKnownNetwork(coin, chain_id)});
   return network_info.get();
 }
 
@@ -658,7 +661,7 @@ const mojom::NetworkInfo* GetZCashMainnet() {
        "Zcash",
        8,
        coin,
-       GetSupportedKeyringsForNetwork(coin, chain_id)});
+       GetSupportedKeyringsForKnownNetwork(coin, chain_id)});
   return network_info.get();
 }
 
@@ -677,7 +680,7 @@ const mojom::NetworkInfo* GetZCashTestnet() {
        "Zcash",
        8,
        coin,
-       GetSupportedKeyringsForNetwork(coin, chain_id)});
+       GetSupportedKeyringsForKnownNetwork(coin, chain_id)});
   return network_info.get();
 }
 
@@ -696,7 +699,7 @@ const mojom::NetworkInfo* GetCardanoMainnet() {
        "Cardano",
        6,
        coin,
-       GetSupportedKeyringsForNetwork(coin, chain_id)});
+       GetSupportedKeyringsForKnownNetwork(coin, chain_id)});
   return network_info.get();
 }
 
@@ -715,7 +718,7 @@ const mojom::NetworkInfo* GetCardanoTestnet() {
        "Cardano",
        6,
        coin,
-       GetSupportedKeyringsForNetwork(coin, chain_id)});
+       GetSupportedKeyringsForKnownNetwork(coin, chain_id)});
   return network_info.get();
 }
 
@@ -726,7 +729,7 @@ const mojom::NetworkInfo* GetPolkadotMainnet() {
   static base::NoDestructor<mojom::NetworkInfo> network_info(
       {chain_id,
        "Polkadot Mainnet",
-       {"https://polkadot.statescan.io/"},
+       {"https://polkadot.subscan.io/"},
        {},
        0,
        {PolkadotMainnetRpcUrl()},
@@ -734,7 +737,7 @@ const mojom::NetworkInfo* GetPolkadotMainnet() {
        "Polkadot",
        10,
        coin,
-       GetSupportedKeyringsForNetwork(coin, chain_id)});
+       GetSupportedKeyringsForKnownNetwork(coin, chain_id)});
   return network_info.get();
 }
 
@@ -750,10 +753,10 @@ const mojom::NetworkInfo* GetPolkadotTestnet() {
        0,
        {PolkadotTestnetRpcUrl()},
        "WND",
-       "Polkadot",
+       "Westend",
        12,
        coin,
-       GetSupportedKeyringsForNetwork(coin, chain_id)});
+       GetSupportedKeyringsForKnownNetwork(coin, chain_id)});
   return network_info.get();
 }
 
@@ -997,6 +1000,10 @@ mojom::NetworkInfoPtr NetworkManager::GetChain(std::string_view chain_id,
   }
 
   return nullptr;
+}
+
+bool NetworkManager::IsPolkadotChain(std::string_view chain_id) {
+  return !!GetChain(chain_id, mojom::CoinType::DOT);
 }
 
 std::vector<mojom::NetworkInfoPtr> NetworkManager::GetAllCustomChains(
--- a/components/brave_wallet/browser/network_manager.cc
+++ b/components/brave_wallet/browser/network_manager.cc
@@ -1003,7 +1003,7 @@ mojom::NetworkInfoPtr NetworkManager::GetChain(std::string_view chain_id,
 }
 
 bool NetworkManager::IsPolkadotChain(std::string_view chain_id) {
-  return !!GetChain(chain_id, mojom::CoinType::DOT);
+  return !(GetChain(chain_id, mojom::CoinType::DOT).is_null());
 }
 
 std::vector<mojom::NetworkInfoPtr> NetworkManager::GetAllCustomChains(
--- a/components/brave_wallet/browser/network_manager_unittest.cc
+++ b/components/brave_wallet/browser/network_manager_unittest.cc
@@ -18,7 +18,6 @@
 #include "base/test/values_test_util.h"
 #include "base/values.h"
 #include "brave/components/brave_wallet/browser/pref_names.h"
-#include "brave/components/brave_wallet/browser/test_utils.h"
 #include "brave/components/brave_wallet/common/brave_wallet.mojom.h"
 #include "brave/components/brave_wallet/common/common_utils.h"
 #include "brave/components/brave_wallet/common/features.h"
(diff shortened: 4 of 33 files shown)


components/brave_wallet/common/common_utils.cc
```diff
@@ -14,6 +14,7 @@
 #include "base/notreached.h"
 #include "base/strings/string_number_conversions.h"
 #include "base/strings/string_util.h"
+#include "brave/components/brave_wallet/common/brave_wallet.mojom-forward.h"
```
supermassive: ~#include "brave/components/brave_wallet/common/brave_wallet.mojom-forward.h"~

components/brave_wallet/common/common_utils.cc
```diff
@@ -291,7 +292,7 @@ bool IsPolkadotImportKeyring(mojom::KeyringId keyring_id) {
          keyring_id == mojom::KeyringId::kPolkadotImportTestnet;
 }
 
-bool IsPolkadotNetwork(std::string_view network_id) {
+bool IsPolkadotRelayNetwork(std::string_view network_id) {
   return network_id == mojom::kPolkadotMainnet ||
```
supermassive, line 303: Why it needs to have Relay in its name?

components/brave_wallet/common/value_conversion_utils.cc
```diff
@@ -147,6 +144,20 @@ mojom::NetworkInfoPtr ValueToNetworkInfo(const base::Value& value) {
...
   }
 
+  if (const auto* keyringIdsValue =
+          params_dict->FindList("supportedKeyrings")) {
+    for (const auto& entry : *keyringIdsValue) {
+      if (!entry.is_int()) {
+        continue;
+      }
+      chain.supported_keyrings.push_back(
+          static_cast<mojom::KeyringId>(entry.GetInt()));
```
supermassive, line 164: LEt's also validate with mojom::IsKnownEnumValue

components/brave_wallet/browser/network_manager.cc
```diff
@@ -65,27 +65,35 @@ constexpr auto kEip1559ForKnownChains =
...
+            // Polkadot mainnet chains.
+            {mojom::kPolkadotMainnet, "polkadot-mainnet"},
+            {mojom::kAcalaMainnet, "acala-mainnet"},
+            {mojom::kMoonbeamMainnet, "moonbeam-mainnet"},
+            {mojom::kBifrostMainnet, "bifrost-mainnet"},
+            // Polkadot testnet chains.
+            {mojom::kPolkadotTestnet, "polkadot-westend"},
+            {mojom::kAssetsHubTestnet, "assets-hub-testnet"},
+            {mojom::kCollectivesTestnet, "collectives-testnet"},
+        },
```
cmazakas, line 91: There's a typo here, it should be `kAssetHubTestnet`, not "assets".

Also, we'll need AssetHub mainnet support as well, as it has the most active holders of any of the chains I've seen at around ~1.78 million active holders: https://assethub-polkadot.subscan.io/

components/brave_wallet/browser/polkadot/polkadot_wallet_service.cc
```diff
@@ -198,6 +206,15 @@ void PolkadotWalletService::SignAndSendTransaction(
                      std::move(callback)));
 }
 
+bool PolkadotWalletService::IsPolkadotChain(std::string_view chain_id) {
+  auto network = network_manager_->GetChain(chain_id, mojom::CoinType::DOT);
+  if (!network) {
```
netzenbot: `IsPolkadotChain` checks for `kPolkadotMainnet` in `supported_keyrings`, but testnet chains carry `kPolkadotTestnet` instead — so this returns `false` for every testnet chain and will silently reject all testnet transactions via `AddUnapprovedPolkadotTransaction`. The check should accept any valid DOT network keyring (or simply test that `network` is non-null), not gate on mainnet specifically. [ARCH-032](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-032)

components/brave_wallet/browser/polkadot/polkadot_wallet_service.cc
```diff
@@ -198,6 +206,15 @@ void PolkadotWalletService::SignAndSendTransaction(
                      std::move(callback)));
 }
 
+bool PolkadotWalletService::IsPolkadotChain(std::string_view chain_id) {
+  auto network = network_manager_->GetChain(chain_id, mojom::CoinType::DOT);
```
netzenbot: `IsPolkadotChain` checks for `kPolkadotMainnet` in `supported_keyrings`, but testnet chains (e.g. `kPolkadotTestnet`, `kAssetsHubTestnet`, `kCollectivesTestnet`) only contain `kPolkadotTestnet`/`kPolkadotImportTestnet`. This means any testnet transaction sent through `AddUnapprovedPolkadotTransaction` will be rejected. If the intent is to validate that the chain is a known Polkadot chain, the `network_manager_->GetChain` check on the line above already does that — the `contains` check should either be removed or extended to also accept testnet keyrings.

components/brave_wallet/browser/polkadot/polkadot_metadata_provider.cc
```diff
@@ -56,7 +58,7 @@ void PolkadotMetadataProvider::Init() {
 void PolkadotMetadataProvider::GetChainMetadata(
```
netzenbot, line 58: This downgrades `CHECK` to `DCHECK`, weakening an invariant that should hold in all builds. If an invalid `chain_id` reaching here is truly a programmer error (impossible at runtime), `CHECK` is correct per [CS-025](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-025). If you want graceful handling instead, remove the assert and propagate the error up like `GetChainMetadata` in the wallet service does.

components/brave_wallet/common/common_utils.cc
cypt4: It is used in import flow where we select relay chain

browser/resources/settings/brave_wallet_page/wallet_networks_subpage.html
```diff
@@ -36,7 +36,7 @@
   <template is="dom-if" if="[[isPolkadotEnabled]]" restamp>
     <wallet-networks-list id="dot-networks-section"
                           coin="[[dotCoin]]"
-                          list-title="$i18n{walletCardanoNetworksListTitle}">
+                          list-title="$i18n{walletPolkadotNetworksListTitle}">
```
cmazakas: This change almost looks like it should be its own separate PR so we can merge it independently, no? It'd be nice to get this fix in sooner, rather than later.

components/brave_wallet/common/brave_wallet.mojom
```diff
@@ -1548,10 +1549,23 @@ const string kZCashMainnet = "zcash_mainnet";
...
 const string kZCashTestnet = "zcash_testnet";
 
+// Hardcoded Polkadot mainnet parachains.
 const string kPolkadotMainnet = "polkadot_mainnet";
 
+const string kAcalaMainnet = "acala_mainnet";
+
+const string kMoonbeamMainnet = "moonbeam_mainnet";
+
+const string kBifrostMainnet = "bifrost_mainnet";
```
cmazakas: The most important mainnet to support here would be AssetHub, which seems to be missing.

cmazakas:
So, we've learned that parachains can have extra fields in the signature payload (AssetHub includes the `assetId` after the `tip` field).

I think the gist of this PR is good but we need to refactor this and split it out over several different PRs.

We need to first update the runtime metadata code to probe for the UnceckedExtrinsic signature payload and determine if we have any extra fields we need to encode during the extrinsic signing process. For AssetHub, this is required. We'll probably just need a PR that adds a field to the Rust signing routines like `use_asset_id` for `make_signed_extrinsic`.

Once we have that done, I think we should just add support AssetHub first. This is probably where the majority of the changes from this PR can be recycled.

And then from there we can add Acala, which iirc is the most popular for dApps, and the rest of the parachains. But we need to audit each chain individually and make sure that we all speak the same language. Because I know right for a fact that any AssetHub testnet transactions for this PR will fail, because the asset id was never encoded.

browser/resources/settings/brave_wallet_page/wallet_networks_subpage.html
cypt4: Ok, reverted

components/brave_wallet/browser/polkadot/polkadot_block_tracker.h
```diff
@@ -56,6 +58,7 @@ class PolkadotBlockTracker : public BlockTracker {
       latest_block_hashes_map_;
   base::ObserverList<Observer> observers_;
   raw_ref<PolkadotSubstrateRpc> polkadot_rpc_;
```
netzenbot, line 60: `PolkadotBlockTracker` only needs to check if a `chain_id` is a known DOT chain — it calls nothing else on `polkadot_wallet_service_`. `NetworkManager` is already available at all construction sites and is the minimal dependency needed for that check. Passing the full `PolkadotWalletService` is a "bag of stuff" antipattern per [ARCH-003](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-003).

components/brave_wallet/browser/polkadot/polkadot_wallet_service.cc
```diff
@@ -36,7 +38,11 @@ PolkadotWalletService::~PolkadotWalletService() = default;
 void PolkadotWalletService::GetChainMetadata(
     std::string_view chain_id,
     GetChainMetadataCallback callback) {
-  CHECK(IsPolkadotNetwork(chain_id));
+  auto network = network_manager_->GetChain(chain_id, mojom::CoinType::DOT);
+  if (!network) {
```
netzenbot: `GetChainMetadata` now has a mixed execution model: the early-return error path fires `callback` synchronously, while the happy path delegates to `metadata_provider_.GetChainMetadata` which is asynchronous. Callers cannot reliably reason about when their callback runs. Wrap the synchronous error return in a `PostTask` to make all paths consistently async. [ARCH-060](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-060)

components/brave_wallet/browser/network_manager_unittest.cc
```diff
@@ -372,16 +372,21 @@ TEST_F(NetworkManagerUnitTest, GetAllChainsTest) {
...
+  EXPECT_EQ(polkadot_chains[1]->chain_id, mojom::kAcalaMainnet);
+  EXPECT_EQ(polkadot_chains[2]->chain_id, mojom::kMoonbeamMainnet);
+  EXPECT_EQ(polkadot_chains[3]->chain_id, mojom::kBifrostMainnet);
+  EXPECT_EQ(polkadot_chains[4]->chain_id, mojom::kAssetHubMainnet);
+  EXPECT_EQ(polkadot_chains[5]->chain_id, mojom::kPolkadotTestnet);
+  EXPECT_EQ(polkadot_chains[6]->chain_id, mojom::kAssetHubTestnet);
+  EXPECT_EQ(polkadot_chains[7]->chain_id, mojom::kCollectivesTestnet);
+  for (const auto& chain : polkadot_chains) {
+    EXPECT_THAT(chain->supported_keyrings,
+                ElementsAreArray(GetSupportedKeyringsForKnownNetwork(
```
netzenbot: The old code explicitly asserted `supported_keyrings` for each Polkadot chain against hardcoded expected values (e.g., `{kPolkadotMainnet, kPolkadotImport}`). Replacing those with `ElementsAreArray(GetSupportedKeyringsForKnownNetwork(...))` is circular — it compares the output against the very function that produced it, so a bug in that function can't be caught here. Per [TI-017](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-017), test coverage must not decrease during refactoring. Add explicit per-chain assertions using hardcoded expected keyring values for each of the 8 chains.

components/brave_wallet/browser/polkadot/polkadot_wallet_service_unittest.cc
```diff
@@ -415,6 +448,18 @@ TEST_F(PolkadotWalletServiceUnitTest, GetAddress) {
       *keyring_service_, *network_manager_, prefs_,
```
netzenbot, line 436: `GetAddress` now fetches chain metadata to determine the SS58 prefix — the critical path for parachain address support — but no test exercises this for any of the new parachain chain IDs (e.g., `kAcalaMainnet`). A test verifying that a mainnet account produces the correct address for a parachain (with the parachain-specific SS58 prefix from metadata) would cover this new behavior. [TI-039](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-039)

components/brave_wallet/common/value_conversion_utils.cc
```diff
@@ -147,6 +144,30 @@ mojom::NetworkInfoPtr ValueToNetworkInfo(const base::Value& value) {
...
+      if (!mojom::IsKnownEnumValue(
+              static_cast<mojom::KeyringId>(entry.GetInt()))) {
+        return nullptr;
+      }
+      chain.supported_keyrings.push_back(
+          static_cast<mojom::KeyringId>(entry.GetInt()));
+    }
+  } else {
+    chain.supported_keyrings =
+        GetSupportedKeyringsForKnownNetwork(chain.coin, chain.chain_id);
```
supermassive, line 168: Can this hit NOTREACHED?

components/brave_wallet/browser/network_manager.cc
```diff
@@ -65,27 +65,36 @@ constexpr auto kEip1559ForKnownChains =
...
+            {mojom::kAcalaMainnet, "acala-mainnet"},
+            {mojom::kMoonbeamMainnet, "moonbeam-mainnet"},
+            {mojom::kBifrostMainnet, "bifrost-mainnet"},
+            {mojom::kAssetHubMainnet, "asset-hub-mainnet"},
+            // Polkadot testnet chains.
+            {mojom::kPolkadotTestnet, "polkadot-westend"},
+            {mojom::kAssetHubTestnet, "asset-hub-testnet"},
+            {mojom::kCollectivesTestnet, "collectives-testnet"},
+        },
         CaseInsensitiveCompare());
```
supermassive, line 92: Can we have a `polkadot-` prefix for all of them so they don't get mixed with evm networks?

supermassive (approved):

components/brave_wallet/common/value_conversion_utils.cc
cypt4: This is not a NOTREACHED case, this is a migration case

components/brave_wallet/browser/network_manager.cc
```diff
@@ -750,10 +893,46 @@ const mojom::NetworkInfo* GetPolkadotTestnet() {
...
+  return network_info.get();
+}
+
+const mojom::NetworkInfo* GetAssetHubTestnet() {
+  const auto coin = mojom::CoinType::DOT;
+  const auto* chain_id = mojom::kAssetHubTestnet;
+  static base::NoDestructor<mojom::NetworkInfo> network_info(
+      {chain_id,
+       "Asset Hub Testnet",
+       {"https://assets-hub.subscan.io/"},
```
cmazakas: This URL should be: https://assethub-westend.subscan.io/

And it seems the canonical spelling is `"AssetHub"`, instead of `"Asset Hub"`.

Edit: Actually seems like "Asset Hub" and "AssetHub" are both used. Hmm. Hard to know how to stay consistent here but maybe "Asset Hub" is the way to go because it's what's returned from the `system_chain` RPC call:
```
{"jsonrpc":"2.0","id":1,"result":"Westend Asset Hub"}
```

components/brave_wallet/browser/network_manager.cc
```diff
@@ -750,10 +893,46 @@ const mojom::NetworkInfo* GetPolkadotTestnet() {
...
+  return network_info.get();
+}
+
+const mojom::NetworkInfo* GetCollectivesTestnet() {
+  const auto coin = mojom::CoinType::DOT;
+  const auto* chain_id = mojom::kCollectivesTestnet;
+  static base::NoDestructor<mojom::NetworkInfo> network_info(
+      {chain_id,
+       "Collectives Testnet",
+       {"https://collectives.subscan.io/"},
```
cmazakas: This URL doesn't seem to load for me. subscan says the network isn't found.

I'm not sure, does Collectives even have a testnet?

Edit: Ah, it does:
```
D:\brave-browser\src\brave>curl.exe   -H "Content-Type: application/json"   -d "{\"id\":1, \"jsonrpc\":\"2.0\", \"method\": \"system_chain\", \"params\": []}"   https://westend-collectives-rpc.polkadot.io 
{"jsonrpc":"2.0","id":1,"result":"Westend Collectives"}
```

components/brave_wallet/browser/network_manager.cc
```diff
@@ -734,7 +803,81 @@ const mojom::NetworkInfo* GetPolkadotMainnet() {
...
+}
+
+const mojom::NetworkInfo* GetMoonbeamMainnet() {
+  const auto coin = mojom::CoinType::DOT;
+  const auto* chain_id = mojom::kMoonbeamMainnet;
+  // Moonbeam is a chain ID; it uses the Polkadot mainnet keyring.
+  static base::NoDestructor<mojom::NetworkInfo> network_info(
+      {chain_id,
+       "Moonbeam",
+       {"https://moonscan.io/"},
```
cmazakas: The other URLs are all subscan-based and https://moonbeam.subscan.io/ exists. Is moonscan.io dramatically better and worth breaking consistency over?

components/brave_wallet/common/common_utils_unittest.cc
```diff
@@ -391,16 +391,16 @@ TEST(CommonUtils, IsPolkadotImportKeyring) {
...
-  EXPECT_FALSE(IsPolkadotNetwork(mojom::kBitcoinMainnet));
-  EXPECT_FALSE(IsPolkadotNetwork(""));
-  EXPECT_FALSE(IsPolkadotNetwork("abc"));
+  EXPECT_FALSE(IsPolkadotRelayNetwork(mojom::kMainnetChainId));
+  EXPECT_FALSE(IsPolkadotRelayNetwork(mojom::kFilecoinMainnet));
+  EXPECT_FALSE(IsPolkadotRelayNetwork(mojom::kSolanaMainnet));
+  EXPECT_FALSE(IsPolkadotRelayNetwork(mojom::kBitcoinMainnet));
+  EXPECT_FALSE(IsPolkadotRelayNetwork(""));
+  EXPECT_FALSE(IsPolkadotRelayNetwork("abc"));
 }
```
cmazakas, line 404: The `EXPECT_FALSE` case should probably include our parachains, shouldn't it?

components/brave_wallet/common/common_utils_unittest.cc
```diff
@@ -565,83 +565,95 @@ TEST(CommonUtils, GetEnabledKeyrings) {
...
-  EXPECT_THAT(GetSupportedKeyringsForNetwork(mojom::CoinType::DOT,
-                                             mojom::kPolkadotTestnet),
+  EXPECT_THAT(GetSupportedKeyringsForKnownNetwork(mojom::CoinType::DOT,
+                                                  mojom::kPolkadotTestnet),
               ElementsAreArray({mojom::KeyringId::kPolkadotTestnet,
                                 mojom::KeyringId::kPolkadotImportTestnet}));
-  EXPECT_THAT(GetSupportedKeyringsForNetwork(mojom::CoinType::DOT,
-                                             "any non mainnet chain"),
+  EXPECT_THAT(GetSupportedKeyringsForKnownNetwork(mojom::CoinType::DOT,
+                                                  mojom::kAssetHubTestnet),
```
cmazakas: Shouldn't we include all parachains here as well?

components/brave_wallet/browser/network_manager.cc
```diff
@@ -750,10 +893,46 @@ const mojom::NetworkInfo* GetPolkadotTestnet() {
...
+  const auto coin = mojom::CoinType::DOT;
+  const auto* chain_id = mojom::kAssetHubTestnet;
+  static base::NoDestructor<mojom::NetworkInfo> network_info(
+      {chain_id,
+       "Asset Hub Testnet",
+       {"https://assets-hub.subscan.io/"},
+       {},
+       0,
+       {AssetHubTestnetRpcUrl()},
+       "AUSD",
```
cmazakas: It seems like Westend Asset Hub (https://assethub-westend.subscan.io/) still uses WND as the currency symbol. Where did AUSD come from? I also couldn't find it in their Asset symbols download either: https://assethub-westend.subscan.io/asset_token

cmazakas:
This PR has some issues but more importantly, it seems to enable broken functionality.

If I circumvent the UI and do a deposit into an account outside of the wallet, I can start the Send flow which doesn't work currently. For example, AssetHub doesn't work at all.

I'm hesitant to approve and merge something that we know is broken. I know we have an entry in `brave://flags` for this but still, I'm not stoked on merging functionality that we don't fully understand or have tested yet.

Because we reuse keypairs for all parachains and users can deposit into those accounts from outside our wallet, I'd really only feel comfortable merging this PR if we've proved we can transact on each parachain by adding subscan.io links to the PR description, as an example.

Also, this PR description could've used some extra helpers. For example, I had to manually hunt down RPC URLs to accurately test these changes:
```
npm run start -- ^
    --enable-logging=stderr ^
    --v=0 ^
    --enable-features=BraveWalletPolkadot ^
    --polkadot-mainnet-rpc-url=https://rpc.polkadot.io ^
    --acala-mainnet-rpc-url=https://acala-rpc.aca-api.network ^
    --moonbeam-mainnet-rpc-url=https://rpc.api.moonbeam.network ^
    --bifrost-mainnet-rpc-url=https://public-01.mainnet.bifrostnetwork.com/rpc ^
    --asset-hub-mainnet-rpc-url=https://polkadot-asset-hub-rpc.polkadot.io ^
    --polkadot-testnet-rpc-url=https://westend-rpc.polkadot.io ^
    --asset-hub-testnet-rpc-url=https://westend-asset-hub-rpc.polkadot.io
```
This should've been included in the PR description, not just for me but also for the sake of QA.

components/brave_wallet/browser/network_manager.cc
cypt4: Yep, this should be WND

components/brave_wallet/browser/network_manager.cc
cypt4: Changed collectives -> coretime since didn't find working collectives block explorer.

Douglashdaniel (approved):
++ Wallet Front-end

cmazakas:
To reiterate, I don't think we should pushing even half-baked testnets into master.

This PR attempts to do too many things at once.

We should just remove all the additions of untested parachain testnets.

This code lays the foundation for such additions, which is what we really want here. We don't want to put code into master that's half-baked. Removing the untested testnets from the UI and wallet is better, because now we can incrementally add them one at a time and test each individual one thoroughly.

AssetHub taught us that we have no idea what using a parachain is going to be like until we actually attempt to use it. Going forward, I would say we only merge a parachain into master if we have proof that our wallet works via a block explorer link showing one of our extrinsics in a finalized block.

kdenhartog:
++ to what @cmazakas is suggesting of adding parachain support separately. Going to start reviewing this though

components/brave_wallet/browser/polkadot/polkadot_wallet_service_unittest.cc
```diff
@@ -356,6 +356,31 @@ TEST_F(PolkadotWalletServiceUnitTest, ConcurrentChainNameFetches) {
...
+  EXPECT_TRUE(
+      polkadot_wallet_service->IsPolkadotChain(mojom::kPolkadotTestnet));
+
+  // Custom DOT chain should also be considered Polkadot.
+  auto custom_dot_chain = network_manager_->GetKnownChain(
+      mojom::kPolkadotMainnet, mojom::CoinType::DOT);
+  ASSERT_TRUE(custom_dot_chain);
+  custom_dot_chain->chain_id = "custom_dot_chain";
+  network_manager_->AddCustomNetwork(*custom_dot_chain);
+  EXPECT_TRUE(polkadot_wallet_service->IsPolkadotChain("custom_dot_chain"));
```
kdenhartog: q: does polkadot use string syntax rather than hex numbers to identify their parachains by ID?

components/brave_wallet/common/brave_wallet.mojom
```diff
@@ -1548,10 +1549,25 @@ const string kZCashMainnet = "zcash_mainnet";
 
 const string kZCashTestnet = "zcash_testnet";
 
+// Hardcoded Polkadot mainnet parachains.
 const string kPolkadotMainnet = "polkadot_mainnet";
 
+const string kPolkadotAcalaMainnet = "polkadot_acala_mainnet";
```
kdenhartog: ideally we set these in smaller individual PRs for each parachain. That way we can have clear understanding of changes that may need to be made per chain to easily come back and find it through the git log later.

components/brave_wallet_ui/common/slices/endpoints/wallet.endpoints.ts
```diff
@@ -590,6 +593,11 @@ async function createDefaultAccounts({
   const networksWithUniqueKeyrings = []
 
   for (const net of visibleNetworks) {
+    if (net.coin === BraveWallet.CoinType.DOT) {
```
kdenhartog: nit: I don't think we use this pattern very often. Should we use `&&~ here rather than 2 separate if statements?

kdenhartog (commented):
few comments based on my own manual review to start. Going to do some LLM scanning next.

Also +1 to separating out parachain additions except into separate PRs otherwise let's test them and publish the transaction IDs here to show we've got it working as expected.

components/brave_wallet/browser/polkadot/polkadot_wallet_service_unittest.cc
cypt4: This is our internal strings, they have parachain identifiers, not sure if we should use them

components/brave_wallet/common/brave_wallet.mojom
cypt4: I've added a feature flag that is not shown on brave://flags. We need to have parachains to test different types parachain-management stuff, for example proper receive address resolution, recipient address selection using valid chain id, etc..

components/brave_wallet/browser/polkadot/polkadot_block_tracker.h
cypt4: done

netzenbot (approved):

components/brave_wallet/common/value_conversion_utils.cc
```diff
@@ -147,6 +144,30 @@ mojom::NetworkInfoPtr ValueToNetworkInfo(const base::Value& value) {
         GetFirstValidChainURLIndex(chain.rpc_endpoints);
   }
 
+  auto* supportedKeyringsValue = params_dict->Find("supportedKeyrings");
```
cmazakas: nit: isn't chromium style to use snake_case for variable names?

components/brave_wallet/browser/polkadot/polkadot_tx_manager.cc
```diff
@@ -154,10 +154,9 @@ void PolkadotTxManager::AddUnapprovedPolkadotTransaction(
     mojom::NewPolkadotTransactionParamsPtr params,
     AddUnapprovedPolkadotTransactionCallback callback) {
   auto chain_id = params->chain_id;
-  if (chain_id != GetNetworkForPolkadotAccount(params->from)) {
+  if (!polkadot_wallet_service_->IsPolkadotChain(chain_id)) {
```
kdenhartog: This seems like it makes it so that a user thinks they're signing a testnet transaction, but it's actually a diffent mainnet or parachain they're actually signing and it gets submitted elsewhere.

Please add a test to make sure that we don't accidentally introduce a chain confusion issue here.

components/brave_wallet/browser/polkadot/polkadot_wallet_service.cc
```diff
@@ -304,4 +323,25 @@ void PolkadotWalletService::OnGetChainMetadataForValidateAddress(
...
+    GetAddressCallback callback,
+    base::expected<PolkadotChainMetadata, std::string> metadata) {
+  if (!metadata.has_value()) {
+    std::move(callback).Run(std::nullopt, WalletInternalErrorMessage());
+    return;
+  }
+
+  PolkadotAddress polkadot_address;
+  polkadot_address.pubkey = pubkey;
+  polkadot_address.ss58_prefix = metadata->GetSs58Prefix();
```
kdenhartog, line 337: nit: whats the expected behavior if a user chooses to load a malicious RPC? seems like a lot more of the data here for parachains is dependent on theses RPCs, so probably worth considering.

kdenhartog (commented):
few further questions

components/brave_wallet/browser/polkadot/polkadot_tx_manager.cc
cmazakas: Hmm, I'm not sure if these lines are interchangeable.

Right now, we grab the chain_id for a given account and then compare that to what was passed. The old code was incorrect because we never verified if the chain_id was correct. This new code solves that and verifies the chain_id but now it doesn't verify the account.

I wonder if we need both here or something like that.

components/brave_wallet/browser/polkadot/polkadot_block_tracker_unittest.cc
```diff
@@ -57,16 +56,15 @@ class PolkadotBlockTrackerUnitTest : public testing::Test {
     brave_wallet::RegisterLocalStatePrefs(local_state_.registry());
     brave_wallet::RegisterProfilePrefs(prefs_.registry());
     network_manager_ = std::make_unique<NetworkManager>(&prefs_);
-    keyring_service_ =
-        std::make_unique<KeyringService>(nullptr, &prefs_, &local_state_);
```
cmazakas, line 61: Ah! Thank you for removing this.

components/brave_wallet/browser/polkadot/polkadot_wallet_service.cc
```diff
@@ -37,7 +40,13 @@ PolkadotWalletService::~PolkadotWalletService() = default;
...
     std::string_view chain_id,
     GetChainMetadataCallback callback) {
-  CHECK(IsPolkadotNetwork(chain_id));
+  if (!network_manager_->IsPolkadotChain(chain_id)) {
+    base::SequencedTaskRunner::GetCurrentDefault()->PostTask(
+        FROM_HERE,
+        base::BindOnce(std::move(callback),
+                       base::unexpected(WalletInternalErrorMessage())));
+    return;
+  }
```
cmazakas, line 49: Out of curiosity, is there a reason why we're weakening this from a CHECK()? I think the change is fine in either case. I'm not sure I saw an updated test in polkadot_wallet_service_unittest.cc.

components/brave_wallet/browser/blockchain_list_parser.cc
```diff
@@ -565,8 +565,7 @@ bool ParseChainList(const std::string& json, ChainList* result) {
       continue;
     }
     network->coin = mojom::CoinType::ETH;
-    network->supported_keyrings =
-        GetSupportedKeyringsForNetwork(network->coin, network->chain_id);
+    network->supported_keyrings = {mojom::KeyringId::kDefault};
```
cmazakas, line 568: I don't actually understand this change. Would you mind helping me understand the context?

components/brave_wallet/browser/keyring_service.cc
```diff
@@ -1267,6 +1267,10 @@ void KeyringService::CreateKeyrings(const KeyringSeed& keyring_seed) {
         base::span(keyring_seed.seed).first<kPolkadotSeedSize>();
     polkadot_mainnet_keyring_ = std::make_unique<PolkadotKeyring>(
         polkadot_seed, KeyringId::kPolkadotMainnet, is_address_allowed);
+  }
+  if (IsKeyringEnabled(KeyringId::kPolkadotTestnet)) {
```
cmazakas, line 1284: Ha, good catch!

components/brave_wallet/browser/network_manager.cc
```diff
@@ -999,6 +1002,10 @@ mojom::NetworkInfoPtr NetworkManager::GetChain(std::string_view chain_id,
   return nullptr;
 }
 
+bool NetworkManager::IsPolkadotChain(std::string_view chain_id) {
+  return !!GetChain(chain_id, mojom::CoinType::DOT);
```
cmazakas: nit: seems like Brave code favors `!GetChain().is_null()` but I'm not sure that's necessarily a huge win here.

components/brave_wallet/common/common_utils_unittest.cc
```diff
@@ -565,83 +565,87 @@ TEST(CommonUtils, GetEnabledKeyrings) {
...
+  EXPECT_THAT(GetSupportedKeyringsForKnownNetwork(mojom::CoinType::DOT,
+                                                  mojom::kPolkadotMainnet),
               ElementsAreArray({mojom::KeyringId::kPolkadotMainnet,
                                 mojom::KeyringId::kPolkadotImport}));
-  EXPECT_THAT(GetSupportedKeyringsForNetwork(mojom::CoinType::DOT,
-                                             mojom::kPolkadotTestnet),
-              ElementsAreArray({mojom::KeyringId::kPolkadotTestnet,
-                                mojom::KeyringId::kPolkadotImportTestnet}));
-  EXPECT_THAT(GetSupportedKeyringsForNetwork(mojom::CoinType::DOT,
-                                             "any non mainnet chain"),
```
cmazakas, line 644: Maybe I'm rushing through code review too much but it's not clear to me why this was removed.

components/brave_wallet/browser/blockchain_list_parser.cc
cypt4: "GetSupportedKeyringsForNetwork" method was renamed so it is considered to work with known chains.
Since for ETH we have only a single keyring we just use it here. Since chain can be custom in this place.

components/brave_wallet/browser/polkadot/polkadot_wallet_service.cc
cypt4: It could be a problem if wrong pallet info \ method id provided, but i do not have an idea how we can protect user from subscribing on malicious RPC

components/brave_wallet/common/common_utils_unittest.cc
cypt4: This method is considered to be used with hardcoded networks

components/brave_wallet/browser/polkadot/polkadot_wallet_service.cc
cypt4: We have a check in PolkadotMetadataProvider, so here we have an error since it is for more public use than PolkadotMetadataProvider. Added a test for the error handling.

components/brave_wallet/browser/polkadot/polkadot_tx_manager.cc
cypt4: Added supported keyring validation

cypt4: Added supported account id validation.

components/brave_wallet/browser/polkadot/polkadot_tx_manager.cc
```diff
@@ -153,11 +157,16 @@ void PolkadotTxManager::OnApprovePolkadotTransaction(
...
-  auto chain_id = params->chain_id;
-  if (chain_id != GetNetworkForPolkadotAccount(params->from)) {
+  if (!params || !params->from) {
     return std::move(callback).Run(false, "", WalletInternalErrorMessage());
   }
 
+  auto chain_id = params->chain_id;
+  auto network = network_manager_->GetChain(chain_id, mojom::CoinType::DOT);
+  if (!network || !std::ranges::contains(network->supported_keyrings,
+                                         params->from->keyring_id)) {
```
cmazakas, line 167: This `contains()` pattern is used at least 3 times in this PR. I think at this point, it's safe to say we need to abstract it into a dedicated named helper function.

cmazakas (approved):
Looks okay to me!

kdenhartog (approved):
blocker addressed. This is all set for sec review

PR #35967, opened 2026-04-28
[DO-NO-SUBMIT][gerrit] Use our deps mirror for `chromium/src/`

cdesouza-chromium:
This is just an experiment to see what happens in CI.

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,7 +78,7 @@
         "dir": "src",
         "tag": "148.0.7778.60",
         "repository": {
-          "url": "https://chromium.googlesource.com/chromium/src.git"
+          "url": "https://gerrit.brave.com/googlesource/chromium"
         },
         "custom_deps": {
           "src/testing/libfuzzer/fuzzers/wasm_corpus": null,


netzenbot (approved):

package.json
```diff
@@ -78,7 +78,7 @@
         "dir": "src",
         "tag": "148.0.7778.60",
         "repository": {
-          "url": "https://chromium.googlesource.com/chromium/src.git"
+          "url": "https://gerrit.brave.com/googlesource/chromium"
```
goodov, line 81: this line [1] in devops prevents this url update to do anything. You need to remove it first.

1. https://github.com/brave/devops/blob/153d8e5aae5077163c9630a22a847ebb98559349/jenkins/lib/vars/utils.groovy#L1004

cdesouza-chromium: I assume we don't need that line any longer there, as this was introduced the day of the outage, is that right?

goodov: it was added long before, the reason: always use googlesource on CI and rely heavily on git cache for `chromium/src` when we had github chromium mirror in package.json.

goodov: we can safely remove it now from devops as we fetch from googlesource now in all branches (nightly, beta, release).

PR #35979, opened 2026-04-28
[gerrit] `refresh_mirrors.py` for mirror CI pipeline

cdesouza-chromium:
This script will be used in the pipeline for sync gerrit mirrors. This
is a very basic script that merely creates `bare` repos to serve as a
trampoline to push things into gerrit.

This first implementation only contains settings for `chromium/src/`.

```sh
python3 refresh_mirrors.py --mirrors-path=./mirrors [--verbose]
```

Bug: https://github.com/brave/brave-browser/issues/55044

### DIFF
--- a/tools/cr/refresh_mirrors.py
+++ b/tools/cr/refresh_mirrors.py
@@ -1,0 +1,213 @@
+#!/usr/bin/env python3
+# Copyright (c) 2026 The Brave Authors. All rights reserved.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+"""Refresh git mirrors from upstream sources into our Gerrit instance.
+
+This script fetches each configured upstream repository and pushes it to its
+corresponding Gerrit mirror. It is designed to run standalone in CI without
+any third-party dependencies.
+
+Usage:
+    python3 refresh_mirrors.py --mirrors-path=./mirrors [--verbose]
+
+Or directly from GitHub:
+    curl -sSLf \\
+        "https://raw.githubusercontent.com/brave/brave-core/master/tools/cr/refresh_mirrors.py" \\
+        | python3 - --mirrors-path=./mirrors
+"""
+
+import argparse
+import logging
+import subprocess
+from dataclasses import dataclass, replace
+from pathlib import Path
+from typing import List, Optional
+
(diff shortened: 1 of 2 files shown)


tools/cr/refresh_mirrors_test.py
```diff
@@ -0,0 +1,214 @@
...
+        result = subprocess.run(
+            ['git', '-C', str(self.mirror_target), 'show-ref'],
+            capture_output=True,
+            text=True,
+            check=False)
+        self.assertEqual(result.stdout.strip(), '')
+
+
+if __name__ == '__main__':
+    unittest.main()
```
netzenbot, line 214: The check `self.assertEqual(result.stdout.strip(), '')` on `self.mirror_target` is vacuously true — this mirror's remote points to `BRAVE_GERRIT_URL`, not `self.mirror_target`, so `mirror_target` is always empty here. Use `_setup_mirror()` to redirect the push remote to `mirror_target` first, then verify after a failed fetch; or check the local bare repo for the absence of refs. [TI-012](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-012)

tools/cr/refresh_mirrors.py
```diff
@@ -0,0 +1,213 @@
...
+    for mirror in MIRRORS:
+        try:
+            replace(mirror, mirrors_path=mirrors_path,
+                    gerrit_url=gerrit_url).sync()
+        except subprocess.CalledProcessError:
+            logging.error('Mirror %s failed; skipping.', mirror.mirror)
+
+
+if __name__ == '__main__':
+    main()
```
netzenbot, line 213: nit: Other scripts in `tools/cr/` (e.g. `plaster.py`, `post_presubmit.py`) use `sys.exit(main())` per the [BS-004 convention](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-004). Change to `sys.exit(main())` so non-zero exit codes from `main()` propagate correctly if a return value is added later.

tools/cr/refresh_mirrors_test.py
netzenbot, line 214: nit: `TestMirrorSetup` tests `_setup()` directly, but `_setup()` is an internal method (Python `_` prefix). Prefer exercising it through the public `sync()` interface to avoid coupling tests to implementation details. [TI-002](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-002)

emerick (approved):

PR #35984, opened 2026-04-28
[Wallet] fix accesses to non-existent data member

cmazakas:
We no longer use a TxStorageDelegate in Wallet. This stray data member access was left in because asan builds aren't common for developers or CI on PRs. Updates the code to use the new proper data member which carries the same member function.


yrliou (approved):

cdesouza-chromium (approved):
LGTM

PR #35991, opened 2026-04-29
Add Commander.js extra-typings module.

goodov:
Add type support for commander.js actions. The module is type-only addition.

Related https://github.com/brave/brave-browser/issues/54990

### DIFF
--- a/build/commands/lib/commander.d.ts
+++ b/build/commands/lib/commander.d.ts
@@ -1,0 +1,9 @@
+// Copyright (c) 2026 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+// Extend the commander module with extra typings.
+declare module 'commander' {
+  export * from '@commander-js/extra-typings'
+}
--- a/build/commands/scripts/commands.js
+++ b/build/commands/scripts/commands.js
@@ -6,7 +6,7 @@
 // Check environment before doing anything.
 import '../lib/checkEnvironment.js'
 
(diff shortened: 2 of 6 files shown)


build/commands/lib/commander.d.ts
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) 2026 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+// Extend the commander module with extra typings.
+declare module 'commander' {
+  export * from '@commander-js/extra-typings'
```
goodov, line 8: this enables type support to `tsc` without replacing all imports, which makes `npm run` commands work as is without triggering another `npm install`.

build/commands/scripts/commands.js
```diff
@@ -100,24 +100,30 @@ program
     '--symlink_dir <symlink_dir>',
     'symlink that points to the actual build directory',
   )
-  .option(
-    '--target_os <target_os_type>',
-    'target OS type',
-    /^(host_os|ios|android)$/i,
```
goodov, line 106: regex matcher is deprecated and doesn't exist in typed version. The new approach is to use `choices()`.

atuchin-m (approved):

netzenbot (approved):

goodov:
@brave/js-deps-reviewers any concerns on adding a type-only dev-package here? It has zero deps, don't think it needs a 3p-sec-review.

bsclifton (approved):
More information in https://github.com/commander-js/extra-typings

This is a type-only package and it's devDependency, not a proper dependency.

LGTM

PR #35992, opened 2026-04-29
Wallet notifications

### LINKED ISSUE #54389: Add wallet-specific notification handler type
### Description

In cr148, `NotificationHandler::Type::SEND_TAB_TO_SELF` was removed here https://chromium.googlesource.com/chromium/src/+/577a926419a0a7350d43af2cbeae968c7b87555e. Our wallet code was taking advantage of that type and it's associated click handler to show notifications.

In order to continue compiling cr148, we temporarily switched the wallet code to use `NotificationHandler::Type::ANNOUNCEMENT` instead, but that click handler navigates to https://www.google.com/chrome/terms/ which we don't want. We should instead add a wallet-specific notification type and associated handler (we have one for `BRAVE_ADS`, so this can be modeled after that).

supermassive:
Resolves https://github.com/brave/brave-browser/issues/54389
- Added notification type BRAVE_WALLET and handler `BraveWalletNotificationHandler` to support wallet TX notifications. Previously we used chromium's `SEND_TAB_TO_SELF` which was removed.
- Removed yet another `KeyedService` (`WalletNotificationService`). Now `BraveWalletService` tracks tx status changes and displays notifications through `BraveWalletServiceDelegate`


patches/chrome-browser-notifications-notification_display_service_impl.cc.patch
```diff
@@ -6,7 +6,7 @@ index c185575c57d0f0e72edc3c9bcfc5874e4c2a63dd..0632480d7827cf594eb696db0237068e
                               std::make_unique<NearbyNotificationHandler>());
      }
  #endif
-+    BRAVE_ADD_BRAVE_ADS_NOTIFICATION_HANDLER;
++    BRAVE_ADD_BRAVE_ADS_NOTIFICATION_HANDLER; BRAVE_ADD_BRAVE_WALLET_NOTIFICATION_HANDLER;
```
goodov: just replace this with a single define named `BRAVE_NOTIFICATION_DISPLAY_SERVICE_IMPL_NOTIFICATION_DISPLAY_SERVICE_IMPL` or `BRAVE_NOTIFICATION_DISPLAY_SERVICE_IMPL_CONSTRUCTOR` and add new things into it.

browser/brave_wallet/notifications/brave_wallet_notification_handler.cc
```diff
@@ -0,0 +1,49 @@
...
+  ForEachCurrentBrowserWindowInterfaceOrderedByActivation(
+      [profile, &browser](BrowserWindowInterface* bwi) {
+        if (bwi->GetProfile() == profile) {
+          browser = bwi;
+          return false;
+        }
+        return true;
+      });
+
+  brave::ShowBraveWalletTxNotificationUrl(browser, origin);
```
goodov: if (browser)

patches/chrome-browser-notifications-notification_display_service_impl.cc.patch
supermassive: fixed

browser/brave_wallet/notifications/brave_wallet_notification_handler.cc
supermassive: fixed

goodov (approved):

browser/sources.gni
```diff
@@ -354,6 +353,10 @@ if (enable_brave_wallet) {
     "//brave/components/decentralized_dns/content",
     "//brave/components/decentralized_dns/core",
   ]
+  if (!is_android) {
```
cypt4, line 356: Shouldn't we also check for enable_brave_wallet  here?

browser/notifications/BUILD.gn
```diff
@@ -4,6 +4,7 @@
 # You can obtain one at https://mozilla.org/MPL/2.0/.
 
 import("//brave/components/brave_ads/buildflags/buildflags.gni")
+import("//brave/components/brave_wallet/common/buildflags/buildflags.gni")
```
cypt4: Where is this used?

browser/sources.gni
supermassive: Yes. It is already in `enable_brave_wallet` scope

browser/notifications/BUILD.gn
supermassive: removed

netzenbot (approved):

cypt4 (approved):

PR #35996, opened 2026-04-29
[canary] Add git_cr.py commit wrapper and hook manager

### LINKED ISSUE #55078: Make using `commit-message.py` more ergonomic and reliable
The inputs for this hook are provided via env variables. This doesn't work well to allow users to know when they mistype things, and make the whole interface a bit mysterious. We should have some wrapper to `git commit` to set these variables with proper validation that they will have the desired effect.

cdesouza-chromium:
Before this change, we had to pass the environment variables manually
before every commit:

```
tags=WIP issue=1234 git commit -m "message"
```

This was fragile in several ways: typos like `tgas=test` are
silently swallowed by the shell and the hook never sees them,
`key=value` prefixes break shell autocompletion from the common
command-line starter, and the hook was traditionally copied into
.git/hooks/ where it went stale whenever commit-msg.py changed
upstream.

git_cr.py fixes all three problems with a single script that
registers as `git cr` via `git config --local alias.cr`:

```
git cr commit --tagged WIP --issue 1234 -m "My change"
```

The --tagged, --issue, and --culprit flags are parsed and
injected as the environment variables the hook expects ($tags,
$issue, $culprit). Unknown flags are forwarded verbatim to
git commit so git's own error messages are preserved.

  git cr install-hook  — creates a symlink on POSIX (or a bash
                         shim on Windows) so the hook tracks
                         changes to commit-msg.py automatically.
  git cr setup-alias   — writes alias.cr to .git/config without
                         touching any shell config files.

cmd_commit performs a sanity check before running git commit:
if the hook is absent or not executable it prints a clear
message pointing the user to `git cr install-hook`.

- Global hooks for `git-secrets`

In cases where `git-secrets` is installed, there is a global hook.

```
git config --global core.hooksPath ~/.git-templates/git-secrets/hooks
```

When core.hooksPath is set globally, git looks exclusively in that
directory for hooks across all repositories, silently ignoring the
per-repo .git/hooks/.  This means the brave-core commit-msg hook
was never invoked even when placed in `brave-core`.

install-hook works around this by doing the following:

1. Sets `core.hooksPath` locally (`.git/config`) to `.git/hooks/`, so git
   uses the per-repo directory for brave-core regardless of the
   global setting.

2. Writes a chaining shim to `.git/hooks/commit-msg` instead of a
   plain symlink.  The shim re-reads core.hooksPath at run time, so
   it adapts automatically if the engineer's global config changes
   without requiring a re-install.  Execution order is global hook
   first (git-secrets), then commit-msg.py, matching the expectation
   that secrets scanning blocks before message validation.

Resolved: https://github.com/brave/brave-browser/issues/55078

### DIFF
--- a/tools/cr/git_cr.py
+++ b/tools/cr/git_cr.py
@@ -1,0 +1,267 @@
+#!/usr/bin/env python3
+# Copyright (c) 2026 The Brave Authors. All rights reserved.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+"""git_cr — git commit wrapper for brave-core.
+
+Provides ergonomic flags for the brave-core commit-msg hook and manages hook
+installation.
+
+Usage
+-----
+  git cr commit [--tagged TAG[,TAG…]] [--issue NUM[,NUM…]]
+                [--culprit HASH[,HASH…]] [<git-commit-args> …]
+
+  git cr commit -m "Fix login button alignment"
+  git cr commit --tagged WIP -m "Work in progress"
+
+  git cr install-hook   Install commit-msg hook from tools/cr/commit-msg.py
+  git cr setup-alias    Register the git cr alias in .git/config
+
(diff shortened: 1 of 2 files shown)


AlexeyBarabash (approved):
LGTM

PR #35997, opened 2026-04-29
Fix race condition in DAU ping on iOS

### LINKED ISSUE #55149: Race condition in DAU ping on iOS
After a successful DAU ping, `processingPing`, `lastPingDate` flags and `lastPingDate` were written on different threads, creating a race condition where a new ping could start before the previous one's state was fully saved.

The race was introduced in https://github.com/brave/brave-core/pull/35385 which was released in v1.91.40.

## Test plan

The test plan is specified in DAU ping issue at https://github.com/brave/brave-core/pull/35385. Added verification step that no duplicate DAU pings are sent on browser start.

aseren:
After a successful DAU ping, `processingPing`, `lastPingDate` flags and `lastPingDate` were written on different threads, creating a race condition where a new ping could start before the previous one's state was fully saved.
To fix this writes now happen together on the main thread.

This is a follow-up to https://github.com/brave/brave-core/pull/35385.

Resolves https://github.com/brave/brave-browser/issues/55149

### DIFF
--- a/ios/brave-ios/Sources/Growth/DAU.swift
+++ b/ios/brave-ios/Sources/Growth/DAU.swift
@@ -9,6 +9,23 @@ import Preferences
 import Shared
 import os.log
 
+/// Abstracts `BraveStats` so tests can inject a fake implementation.
+// TODO(https://github.com/brave/brave-browser/issues/55105):
+// Change this to use a protocol exposed from the `BraveStats` bridge.
+public protocol BraveCoreStats: AnyObject {
+  /// Reads the `brave.stats.reporting_enabled` Chromium preference,
+  // so must be accessed on the main thread.
+  var isStatsReportingEnabled: Bool { get }
+  /// Reads the `brave.brave_ads.enabled` Chromium preference,
+  // so must be accessed on the main thread.
+  var isNotificationAdsEnabled: Bool { get }
+  /// Reads and writes the `brave.stats.last_check_ymd` Chromium preference,
+  // so must be accessed on the main thread.
+  var lastPingDate: Date? { get set }
+}
+
+extension BraveStats: BraveCoreStats {}
+
 public class DAU {
 
   /// Default installation date for legacy woi version.
@@ -65,11 +82,11 @@ public class DAU {
   }
 
   private let apiKey: String?
-  private let braveCoreStats: BraveStats?
+  private let braveCoreStats: (any BraveCoreStats)?
   private let serpMetrics: (any SerpMetrics)?
 
   public init(
-    braveCoreStats: BraveStats?,
+    braveCoreStats: (any BraveCoreStats)?,
     serpMetrics: (any SerpMetrics)?
   ) {
     self.braveCoreStats = braveCoreStats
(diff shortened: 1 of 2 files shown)


ios/brave-ios/Sources/Growth/DAU.swift
```diff
@@ -158,15 +165,19 @@ public class DAU {
...
+        // Store the last ping date in local state so all platform components
+        // share one source of truth.
+        // Thread hop to main is required because `lastPingDate` sets
+        // `brave.stats.last_check_ymd` Chromium preference, which must be done
+        // from the main thread.
+        stats.lastPingDate = paramsAndPrefs.date
+
+        // Ping was successful, next ping should be sent with `first` parameter set to false.
+        // This preference is set for future DAU pings.
+        Preferences.DAU.firstPingParam.value = false
```
aseren: moved setting `Preferences.DAU.firstPingParam.value` to the same thread where we set stats.lastPingDate to avoid possible race condition.

ios/brave-ios/Sources/Growth/DAU.swift
```diff
@@ -9,6 +9,14 @@ import Preferences
...
 import os.log
 
+/// Abstracts `BraveStats` so tests can inject a fake implementation.
+public protocol BraveCoreStats: AnyObject {
+  var isStatsReportingEnabled: Bool { get }
+  var lastPingDate: Date? { get set }
+}
+
+extension BraveStats: BraveCoreStats {}
+
```
netzenbot, line 28: nit: `lastPingDate` has a non-obvious threading constraint — the implementation requires it be set on the main thread (see the `DispatchQueue.main.async` wrap at the call site). A brief doc comment on the property noting this requirement would help anyone implementing the protocol. [DOC-004](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-004)

aseren: Added a comment to the protocol

ios/brave-ios/Sources/Growth/DAU.swift
```diff
@@ -9,6 +9,21 @@ import Preferences
 import Shared
 import os.log
 
+/// Abstracts `BraveStats` so tests can inject a fake implementation.
+public protocol BraveCoreStats: AnyObject {
```
kylehickinson, line 15: This should just be done on `BraveStats` itself in the core side rather than introduce a protocol here

kylehickinson: Fine to be done as a follow-up

ios/brave-ios/Tests/GrowthTests/DAUTests.swift
```diff
@@ -522,6 +558,49 @@ class DAUTests: XCTestCase {
...
+      for: NSPredicate { _, _ in
+        !Preferences.DAU.firstPingParam.value && mockStats.lastPingDate != nil
+      },
+      evaluatedWith: nil
+    )
+    waitForExpectations(timeout: 2)
+
+    XCTAssertFalse(Preferences.DAU.firstPingParam.value)
+    XCTAssertNotNil(mockStats.lastPingDate)
+    _ = dau
```
kylehickinson: does DAU need to stay alive post expectations? It doesnt seem like its used anywhere

aseren: was added do avoid DAU object deletion while doing thread hops inside sendPingToServer(). But it is actually not needed now, because we no longer pass self as a weak. I will remove `_ = dau` line.

aseren: Removed

ios/brave-ios/Sources/Growth/DAU.swift
aseren: Raised an issue https://github.com/brave/brave-browser/issues/55105 and added TODO comment in code

kylehickinson (approved):

Uni-verse:
Verified on `iPhone 12` running `iOS 26.3.1` using version `1.92 (46)`

Test Plan used https://github.com/brave/brave-core/pull/35385#issue-4227072326

**`Test Case 1`** 
   - ensured that daily usage ping is sent on first launch which contains first=true in the query params
   - ensured that `brave.stats_last_check_ymd` in brave://local-state has the correct time in UTC
   - ensured that usage ping is sent and time in the `brave.stats_last_check_ymd` metric is updated when moving time by 1 day, 1 week and also 1 month. 

first launch | 1 day | 1 week | 1 month
--|--|--|--
IMG_0290 | IMG_0291 | IMG_0292 | IMG_0293 2
Screenshot 2026-05-11 at 2 40 47 PM | Screenshot 2026-05-11 at 3 41 42 PM | Screenshot 2026-05-11 at 3 43 08 PM | Screenshot 2026-05-11 at 3 44 32 PM

**`Test Case 2`** 
   - ensured that daily usage ping is sent on launch when installing an older version of Brave but not after upgrading to `1.92 (46)` 
   - ensured that correct date is reflected in the `brave.stats_last_check_ymd` metric when upgrading / changing system time. 
   
old version | clock 1 day ahead 
--|--
Screenshot 2026-05-11 at 4 08 44 PM | Screenshot 2026-05-11 at 4 12 37 PM
IMG_0295 2 | IMG_0296 2



**`Test Case 3`**
   - ensured that daily usage ping is sent and `brave.stats_last_check_ymd` metric was updated after moving the system time ahead and then upgrading the app version
   
example | example
--|--
IMG_0297 2 | Screenshot 2026-05-11 at 4 22 00 PM

PR #36002, opened 2026-04-29
Fix network-audit default args passing after commander.js bump.

goodov:
Commander.js changed the way it handles arguments in "allowUnknownOption" mode, so in such call:
```
npm run test brave_network_audit_tests -- --ui-test-action-timeout=320000 --test-launcher-timeout=2200000 Release --target_arch=x64 --output_xml
```
`Release` argument put right after `--test-launcher-timeout=2200000` is not treated as a "build config" anymore.

The fix is to avoid such call by putting default args into a test runner instead of `package.json`.

Related https://github.com/brave/brave-browser/issues/54990

### DIFF
--- a/build/commands/lib/test.js
+++ b/build/commands/lib/test.js
@@ -58,6 +58,18 @@ const buildTests = async (testsToRun, config) => {
   await util.buildTargets(config.buildTargets, config.defaultOptions)
 }
 
+const defaultTestSuiteArgs = (testSuite) => {
+  switch (testSuite) {
+    case 'brave_network_audit_tests':
+      return [
+        '--ui-test-action-timeout=320000',
+        '--test-launcher-timeout=2200000',
+      ]
+    default:
+      return []
+  }
+}
+
 const runTests = async (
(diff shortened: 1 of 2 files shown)


petemill (approved):
So with commander we can no longer `npm run commandA` from `commandB` if we have `-- [args]` and want the caller to also specify args? That's unfortunate.

Or perhaps we no longer need `--` - did you try that?

netzenbot (approved):

goodov:
> So with commander we can no longer `npm run commandA` from `commandB` if we have `-- [args]` and want the caller to also specify args? That's unfortunate.

we can, but only when all options are declared in a command and command doesn't allow "unknown" options.

In commands where we allow unknown options (mainly `test` and `start`), this becomes tricky - we need to pass arguments and known options first, and only then we are allowed to pass unknown flags.

> Or perhaps we no longer need `--` - did you try that?

that's `npm`-specific, so not related here, but we will drop that after switch to pnpm.

PR #36003, opened 2026-04-29
[Wallet] Remove buildflag guard to fix notreached crash for iOS

### LINKED ISSUE #55085: [Wallet] Crash when trying to copy private key
### Description

Crash when trying to copy private key 

### Steps to reproduce

1. Open Export option for an ETH account
2. Enter password to view the private key
3. Tap on the private key to copy, browser crashes

### Actual result

Browser crash when trying to copy private key

### Expected result

No crash. Should be able to copy private key to clipboard

### Reproduces how often

Easily reproduced

### Brave version

All

### Device/iOS version

iPhone 13 (iOS 26.4)

### Affected browser versions

- [x] latest AppStore
- [x] latest TestFlight
- [x] previous TestFlight

### Reproducibility

### Miscellaneous information

cc:: @nuo-xu @Douglashdaniel

nuo-xu:
Resolves https://github.com/brave/brave-browser/issues/55085

### DIFF
--- a/components/brave_wallet/browser/brave_wallet_service.cc
+++ b/components/brave_wallet/browser/brave_wallet_service.cc
@@ -2137,12 +2137,6 @@ void BraveWalletService::SetTransactionSimulationOptInStatus(
 
 void BraveWalletService::WriteToClipboard(const std::string& text,
                                           bool is_sensitive) {
-  // We manually disable the iOS builds here because of an upstream bug in how
-  // Chromium is adding sources to the clipboard component. It only
-  // conditionally adds the iOS sources when use_blink=true, which unfortunately
-  // leads to a whole slew of unresolved symbols during linking.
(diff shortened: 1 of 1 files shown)


cmazakas (approved):

kjozwiak:

```
Brave | 1.92.4 Chromium: 148.0.7778.96(Official Build)nightly (64-bit) 
--- | ---
Revision | 0be82570025e...
OS | iOS
```

Reproduced the original issue using the STR/Cases outlined via https://github.com/brave/brave-browser/issues/55085#issue-4351411929 as per the following:

**`Example`** | **`Example`**
----------------|------------------
IMG_1603 | IMG_1602

Using the same STR/Cases mentioned above, ensured that Brave doesn't crash when copying a private key via `Export` while using `Wallet WebUI`.

PR #36006, opened 2026-04-29
Skip opening origin settings on first purchase for branded builds

### LINKED ISSUE #55037: Brave Origin Separate build new profile opens chrome://settings
There's a report of this:
> Perhaps expected, but when I provided my purchase code to launch the browser for the first time, I did notice that both [brave://newtab](chrome://newtab/) and [brave://settings](chrome://settings/) both appeared, and in their own windows.

That's not expected.

bbondy:
On `is_brave_origin_branded=true` builds we should not open the settings page when Brave Origin is purchased. 

QA plan:
- Brave Origin branded build on startup after activating does NOT open the settings page
- Brave build being upgraded to Origin does open the settings page after purchase.

Fix https://github.com/brave/brave-browser/issues/55037


SergeyZhukovsky (approved):
++

PR #36012, opened 2026-04-29
[ads][CodeHealth] Extract initialization pipeline from AdsImpl into AdsInitializer

tmancey:
Moves the multi-step startup sequence (database open, client state migration and load, confirmation state migration and load) into a dedicated AdsInitializer class. AdsImpl retains ownership of is_initialized_ and the task queue and receives a single bool result via InitializeCallback. Trace events remain in AdsImpl so that TRACE_ID_LOCAL(this) stays anchored to the AdsImpl address. Adds ads_impl_unittest.cc and ads_impl_not_initialized_unittest.cc covering the Initialize and Shutdown public API paths.

No behavioral changes.

Resolves

### DIFF
--- a/components/brave_ads/core/internal/BUILD.gn
+++ b/components/brave_ads/core/internal/BUILD.gn
@@ -311,6 +311,8 @@ static_library("internal") {
     "ads_core/ads_core_util.h",
     "ads_impl.cc",
     "ads_impl.h",
+    "ads_initializer.cc",
+    "ads_initializer.h",
     "ads_internals/ads_internals_util.cc",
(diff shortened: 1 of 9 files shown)


tmancey:
This warning is pre-existing in `AdsImpl`. The `NESTABLE_ASYNC` trace macros were introduced with the original `AdsImpl::Initialize` implementation and are not new to this branch. This PR extracts the initialization pipeline into `AdsInitializer` without modifying the tracing code. The macros can be migrated to `TRACE_EVENT_BEGIN/END` with `perfetto::Track` in a follow-up.

aseren (approved):

PR #36015, opened 2026-04-29
[ads][CodeHealth] Refactor AdsServiceImpl constructor and startup

tmancey:
Moves MaybeInitNotificationHelper and all service observations into InitializeBatAdsCallback so they are only active once the service is fully initialized, and moves the device ID fetch into MaybeStartBatAdsService with lazy caching so subsequent restarts skip the fetch. Removes the redundant CanStartBatAdsService call from the constructor, removes IsBatAdsServiceBound in favour of checking is_bound directly, fixes the IsObserving and GetSource patterns that caused InitializeRewardsWallet to always fall through to the null-wallet path, renames the resource component registration methods for clarity, and corrects observation setup and teardown ordering.

No behavioral changes.

## Test plan

Test toggling different ad types on/off for non-origin cases where the service should run as long as at least one ad type is enabled. Verify that disabling all ad types shuts down the ads service, and confirm it restarts when re-enabling each ad type individually. Also confirm observers are working for NTT component updates after browser startup, as well as after shutdown and restart.

Also test toggling the last remaining switch which on/off fast. To see if any crashes due to race conditions when shutting down/starting up the service.

Resolves

### DIFF
--- a/components/brave_ads/browser/ads_service_impl.cc
+++ b/components/brave_ads/browser/ads_service_impl.cc
@@ -143,6 +143,7 @@ AdsServiceImpl::AdsServiceImpl(
           std::move(virtual_pref_provider_delegate))),
       http_client_(std::move(http_client)),
       channel_name_(channel_name),
+      resource_component_(resource_component),
       history_service_(history_service),
       host_content_settings_map_(host_content_settings_map),
       ads_tooltips_delegate_(std::move(ads_tooltips_delegate)),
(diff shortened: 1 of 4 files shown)


components/brave_ads/browser/ads_service_impl.cc
```diff
@@ -737,17 +730,18 @@ void AdsServiceImpl::InitializeSearchResultAdsPrefChangeRegistrar() {
...
   }
 
-  if (IsBatAdsServiceBound() && UserHasOptedInToNotificationAds() &&
+  if (bat_ads_service_remote_.is_bound() && UserHasOptedInToNotificationAds() &&
       path == prefs::kOptedInToNotificationAds) {
     // Register language resource components if the user has joined Brave
     // Rewards, opted into notification ads, and the Bat Ads Service has
     // already started.
-    RegisterResourceComponentsForDefaultLanguageCode();
+    delegate_->MaybeInitNotificationHelper();
```
aseren: the comment above should belong to `RegisterLanguageResourceComponent()`

aseren: Fixed

aseren (approved):

PR #36016, opened 2026-04-29
Upgrade from Chromium 148.0.7778.60 to Chromium 148.0.7778.96 (1.91.x)

emerick:
Resolves https://github.com/brave/brave-browser/issues/55095


kjozwiak (approved):
Uplift into `1.91.x` approved 👍

PR #36023, opened 2026-04-29
[Origin] Apply profile policies before pref store reads

### LINKED ISSUE #54991: Android Origin (1.91.89) showing Brave Search API NTP Ad after upgrading
### Description

From the user: 
```

Hello Brave Support team,

I bought a license for brave origin and am still seeing ads on the Android app and I want to know why.

Thank you for your time on this matter
```

Image

Image

Image

### Steps to reproduce

1. Pay for Origin on Brave Nightly 
2. Complete upgrade 
3. NTP appears 

### Actual result

Android Origin still showing Brave Search API NTP Ad after upgrading. 

### Expected result

Expectation is that even house NTP ads would be turned off after upgrading. 

### Reproduces how often

Easily reproduced

### Brave version

Nightly 1.91.89

### Device

- Brand/model: 
- Android version: 16 

### Channel information

- [ ] release (stable)
- [ ] beta
- [x] nightly

### Reproducibility

### Miscellaneous information

_No response_

SergeyZhukovsky:
BraveOrigin profile policies (e.g. BraveRewardsDisabled) were not landing in the managed pref store before AdsService's CanStartBatAdsService gate ran on Android and desktop. The race had two parts:

1. BraveOriginPolicyManager was initialized lazily from the per-profile keyed-service factory, which on Android's startup-data path runs after the profile policy connector is already built.
2. BraveProfilePolicyProvider::SetProfileID was called late (from BRAVE_PROFILE_IMPL_TAKE_PREFS_FROM_STARTUP_DATA), so OnBravePoliciesReady couldn't trigger RefreshPolicies until well after PolicyServiceImpl had been constructed with our empty bundle.

Two changes shift the policy push earlier in startup so the async merge posted by PolicyServiceImpl::OnUpdatePolicy has time to propagate before consumers read the pref store:

- BraveBrowserProcessImpl::Init eagerly initializes BraveOriginPolicyManager with local_state().

- BraveProfilePolicyProvider gets a thread-local profile-path stash that is populated by callers immediately before profile policy connector creation: chromium_src wrappers for both CreateProfilePolicyConnectorForBrowserContext (desktop) and StartupData::CreateServicesInternal (Android, via two new macro hooks in upstream chrome/browser/startup_data.cc). The BRAVE_PROFILE_POLICY_CONNECTOR_INIT macro consumes the stash and calls SetProfileID before provider->Init() runs, so OnBravePoliciesReady fires synchronously inside Observe() with profile_id_ already set and triggers RefreshPolicies in time.

Also adds an early-return in SetProfileID so legacy SetBraveProfilePolicyProviderProfileID call sites don't re-fire RefreshPolicies for the already-set id.

Resolves: https://github.com/brave/brave-browser/issues/54991


mkarolin (approved):
`chromium_src`, `patches`, `deps` - LGTM

PR #36024, opened 2026-04-29
[AI Chat] Pass uploaded files through skill submission path

### LINKED ISSUE #51185: Leo Skills can't see uploaded images
### Description

When using a Skill for analyzing an attached photo, Leo seems to lose the image and can't see it when processing your prompt. In my example, I set up a Skill called /where to tell me the location seen in a photo. This skill works when I'm viewing a website with a photo but not when I attach a photo. I also run into the issue if trying to view a photo through its local file path (eg brave://background-wallpaper/vedant-sonani.jpg)

The issue seems to persist whether I'm using Automatic or I switch to a "Vision" model.

### Steps to reproduce

1. Create a skill about analyzing images
2. Attach an image
3. Select the image analysis skill for your prompt

### Actual result

The example skill, /where:
Image

Attaching the image to the prompt:
Image

Leo attempting to answer the prompt:
Image

### Expected result

Leo should be able to view the image and provide a guess at the location of the photo. The image would remain visible to the user in the conversation as well.

Here's the correct result when typing out the prompt manually and attaching the image:
Image

### Reproduces how often

Easily reproduced

Brave: 1.87.33 Chromium: 143.0.7499.40 (Official Build) nightly (64-bit)
Revision: 50aedac435134da4b4de013df68978b7b6e56f3b
OS: Windows 11 Version 24H2 (Build 26100.7171)

### Channel information

- [ ] release (stable)
- [ ] beta
- [x] nightly

### Reproducibility

### Miscellaneous information

_No response_

yrliou:
The skill submission path in the conversation UI never forwarded the pending uploaded files to the browser, so attaching an image and invoking a skill caused Leo to answer as if no image were present. Add an uploaded_files parameter to SubmitHumanConversationEntryWithSkill across the mojom interface and the handler, pass the staged files from the frontend, and run the existing vision-model auto-switch on this path so an unmodeled (or non-vision) skill upgrades to a vision model when an image is attached.

Resolves https://github.com/brave/brave-browser/issues/51185

Test plan:
1. Create a new skill /where with prompt "where is this photo taken"
2. Drag an image of a place into a conversation with Leo and type /where to invoke the skill
3. Submit /where skill with the image of a place
4. Leo should be able to tell the place of where this photo is taken.

Screenshot 2026-04-29 at 2 22 42 PM

### DIFF
--- a/components/ai_chat/core/browser/conversation_handler.cc
+++ b/components/ai_chat/core/browser/conversation_handler.cc
@@ -72,6 +72,18 @@ using ai_chat::mojom::ConversationTurn;
 
 constexpr size_t kDefaultSuggestionsCount = 4;
 
+bool HasUploadedImageOrScreenshot(
+    const std::optional<std::vector<mojom::UploadedFilePtr>>& uploaded_files) {
+  if (!uploaded_files || uploaded_files->empty()) {
+    return false;
+  }
+  return std::ranges::any_of(
+      *uploaded_files, [](const mojom::UploadedFilePtr& file) {
+        return file->type == mojom::UploadedFileType::kImage ||
+               file->type == mojom::UploadedFileType::kScreenshot;
+      });
+}
+
 }  // namespace
 
 ConversationHandler::Suggestion::Suggestion(std::string title)
@@ -652,24 +664,28 @@ void ConversationHandler::SubmitHumanConversationEntryWithSkill(
   // Get Skill from prefs and convert to SkillEntry object
   auto skill = prefs::GetSkillFromPrefs(*prefs_, skill_id);
 
+  // Decide the intended target: skill's pinned model when set, else the
+  // current model. With an image attached, the target must support vision —
+  // fall back if it doesn't. Switch once if we're not already there.
+  std::string target = (skill && skill->model && !skill->model->empty())
+                           ? *skill->model
+                           : GetCurrentModel().key;
+  if (HasUploadedImageOrScreenshot(uploaded_files)) {
+    target = GetVisionCapableModelKey(target);
+  }
+  if (target != GetCurrentModel().key) {
+    ChangeModel(target);
+  }
+
   mojom::SkillEntryPtr skill_entry = nullptr;
 
   if (skill) {
-    // Switch model if specified and different from current model
-    if (skill->model && !skill->model->empty() &&
-        *skill->model != GetCurrentModel().key) {
-      ChangeModel(*skill->model);
-    }
-
     // Create Skill entry
     skill_entry = mojom::SkillEntry::New(skill->shortcut, skill->prompt);
 
     // Update last_used time
     prefs::UpdateSkillLastUsedInPrefs(skill_id, *prefs_);
   }
-
-  // Auto-switch to vision model if needed
-  MaybeSwitchToVisionModel(uploaded_files);
 
   mojom::ConversationTurnPtr turn = mojom::ConversationTurn::New(
       std::nullopt, CharacterType::HUMAN, mojom::ActionType::QUERY, input,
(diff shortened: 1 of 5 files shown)


components/ai_chat/core/browser/conversation_handler.cc
```diff
@@ -663,11 +664,14 @@ void ConversationHandler::SubmitHumanConversationEntryWithSkill(
     prefs::UpdateSkillLastUsedInPrefs(skill_id, *prefs_);
   }
 
+  // Auto-switch to vision model if needed
+  MaybeSwitchToVisionModel(uploaded_files);
```
darkdh: Shouldn't we only do auto switch when the skill has no pinned model or the pinned model has no vision support?

yrliou: There's an existing block doing model switches based on skill settings above this.
So this only switches if the pinned model or current model (when no pinned model) has no vision support.

yrliou: drawback is it's possible that we'll switch model twice if user configure their image related skill with a model without vision support

darkdh: so ^ that is the missing test coverage in unit test, we only test model switching without pinned model but lack testing pinned model that has no vision support

darkdh: I guess to prevent the double switching, we can simply check if it has `vision_support` when `uploaded_files` which contains images is presented before `ChangeModel`

darkdh: that leaves one scenario unchecked: current globally selected model support vision->skill pinned model has no vision support->we won't do any switches when images presented

I think this is acceptable and should be covered by test if we implemented the above check.

netzenbot (approved):

components/ai_chat/core/browser/conversation_handler.cc
yrliou: should be fixed by https://github.com/brave/brave-core/pull/36024/commits/a892c353dd761aa5afc792ca4b99b6be8c7c3c59

components/ai_chat/core/browser/conversation_handler.cc
```diff
@@ -642,23 +654,31 @@ void ConversationHandler::SubmitHumanConversationEntryWithAction(
...
+  // fall back if it doesn't. Switch once if we're not already there.
+  std::string target = (skill && skill->model && !skill->model->empty())
+                           ? *skill->model
+                           : GetCurrentModel().key;
+  if (HasUploadedImageOrScreenshot(uploaded_files)) {
+    target = GetVisionCapableModelKey(target);
+  }
+  if (target != GetCurrentModel().key) {
+    ChangeModel(target);
+  }
```
darkdh: This can be merged into `MaybeSwitchToVisionModel` if we can pass model key as optional

components/ai_chat/core/browser/conversation_handler_unittest.cc
```diff
@@ -5019,34 +5019,80 @@ TEST_F(ConversationHandlerUnitTest,
...
 }
 
-TEST_F(ConversationHandlerUnitTest,
-       SubmitHumanConversationEntryWithSkill_WithUploadedFiles) {
+struct SkillImageUploadScenario {
+  const char* test_name;
+  const char* starting_model;
+  std::optional<const char*> skill_pinned_model;
+  const char* expected_final_model_key;
+  int expected_model_changes_during_submit;
```
darkdh: does this need to be int? It seems like bool would be more appropriate since we don't expected changes more than once

components/ai_chat/core/browser/conversation_handler_unittest.cc
```diff
@@ -5019,34 +5019,80 @@ TEST_F(ConversationHandlerUnitTest,
...
+class ConversationHandlerSkillImageUploadTest
+    : public ConversationHandlerUnitTest,
+      public testing::WithParamInterface<SkillImageUploadScenario> {};
+
+// Covers model resolution and file plumbing when a skill submission carries
+// an image attachment. Each case starts on a specific model and may have a
+// pinned skill model; the test asserts the final model key and the number
+// of ChangeModel calls during submit.
+TEST_P(ConversationHandlerSkillImageUploadTest, ResolvesModelAndPlumbsImage) {
+  const auto& tc = GetParam();
```
darkdh: what is tc?

yrliou: renamed to test_params in https://github.com/brave/brave-core/pull/36024/commits/ae8019110d9733742ec8f5fac1196883ff8bfcc6

components/ai_chat/core/browser/conversation_handler_unittest.cc
yrliou: good point, updated to bool in https://github.com/brave/brave-core/pull/36024/commits/6760db222126e95010f614069fe0edf17203bcc2

components/ai_chat/core/browser/conversation_handler.cc
yrliou: https://github.com/brave/brave-core/pull/36024/commits/699d0ce287a7d92f1a0f54e5c2cb63d9e71412a0

darkdh (approved):

netzenbot (approved):

PR #36033, opened 2026-04-30
[iOS] Playlist: Remove the ability to add homepage and search result page as playlist items

### LINKED ISSUE #55010: Playlist: Capture allowed from non‑playable pages produces permanently broken items
### Description

### Steps to reproduce

1. Navigate to you tube

### Actual result

### Expected result

3. The video is playable.

### Reproduces how often

Easily reproduced

### Brave version

1.89

### Device/iOS version

iPhone 15 Pro Max
iPhone 17

iOS 26.3

### Affected browser versions

- [x] latest AppStore
- [x] latest TestFlight
- [x] previous TestFlight

### Reproducibility

### Miscellaneous information

_No response_

EliHini:
## Summary
Updates `PlaylistScriptHandler` so playlist items store a `pageSrc` that `LivePlaylistWebLoader` can reload, and avoids offering captures that would persist as broken entries on YouTube listing-style URLs.

## Problem
Adding to Playlist from contexts where **`pageSrc` was wrong or not reloadable** (always forcing the visible tab URL when the script already reported the correct page, or allowing captures from YouTube feeds/search/home where there is no stable video page to resolve later) produced **playlist items that could not be played back reliably**.

- **YouTube guard** — **`canResolvePageSrcLater`** returns `false` for `youtube.com` paths that are not `/watch`, `/shorts/…`, `/embed/…`, or `/v/…`, so the normal add flow is not entered. **`youtu.be/&lt;id&gt;`** remains allowed when the path segment is non-empty. Other domains are unchanged (**allowed by default**).
- **Early exit** — Resets the playlist URL bar to **`.none`**, consistent with other rejected paths.

Resolves  https://github.com/brave/brave-browser/issues/55010

### DIFF
--- a/ios/brave-ios/Sources/Brave/Frontend/UserContent/UserScripts/Scripts_Dynamic/ScriptHandlers/Paged/PlaylistScriptHandler.swift
+++ b/ios/brave-ios/Sources/Brave/Frontend/UserContent/UserScripts/Scripts_Dynamic/ScriptHandlers/Paged/PlaylistScriptHandler.swift
@@ -130,11 +130,24 @@ class PlaylistScriptHandler: NSObject, TabContentScript, TabObserver {
       return
     }
 
-    // Copy the item but use the web-view's title and location instead, if available
-    // This is due to a iFrames security
-    item.pageSrc = tab.visibleURL?.absoluteString ?? item.pageSrc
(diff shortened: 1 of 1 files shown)


netzenbot (approved):

EliHini:
Canceling because of CI errors -- unable to resolve branch.
Re-opened as https://github.com/brave/brave-core/pull/36086

PR #36038, opened 2026-04-30
[query-filter] Skip version update on failed parsing

### LINKED ISSUE #55124: Skip version update on failed parsing and improve testing
### Description

We should skip the version update if we were unable to [parse the json contents](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/components/query_filter/browser/query_filter_component_installer.cc?L53). 

We should add support to pass a callback to this method [OnQueryFilterFileRead](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/components/query_filter/browser/query_filter_component_installer.cc?L44), which would allow to better test the ComponentReady functionality [here](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/components/query_filter/browser/query_filter_component_installer_unittest.cc?L181-206) 

@fmarier @cdesouza-chromium for visibility on this issue.

r0hit0303:
This change updates the query filter component installer to skip updating version if the json file parsing failed. 

This change also adds logic to supply test only callback which gets fired after the component installer attempted to parse the json. This facilitates testing the ComponentReady functionality e2e. 

Resolves https://github.com/brave/brave-browser/issues/55124

### DIFF
--- a/components/query_filter/browser/query_filter_component_installer.cc
+++ b/components/query_filter/browser/query_filter_component_installer.cc
@@ -10,6 +10,7 @@
 #include <array>
 #include <vector>
 
+#include "base/check_is_test.h"
 #include "base/feature_list.h"
 #include "base/files/file_path.h"
 #include "base/files/file_util.h"
@@ -22,9 +23,11 @@
 #include "brave/components/query_filter/common/features.h"
 #include "components/component_updater/component_installer.h"
 #include "components/component_updater/component_updater_service.h"
+#include "third_party/abseil-cpp/absl/cleanup/cleanup.h"
 
 namespace {
 inline constexpr char kQueryFilterComponentName[] = "Query Filter";
+base::OnceClosure* g_on_file_loaded_callback_for_testing_ = nullptr;
 
 // This is the SHA-256 of the query-filter component's public key.
 inline constexpr std::array<uint8_t, 32> kQueryFilterComponentPublicKeySHA256 =
@@ -43,6 +46,14 @@ std::string ReadQueryFilterFile(const base::FilePath& path) {
 
 void OnQueryFilterFileRead(const base::Version& version,
                            const std::string& json_data) {
+  absl::Cleanup cleanup_for_test_only = []() {
+    if (g_on_file_loaded_callback_for_testing_) {
+      CHECK_IS_TEST();
+      CHECK(!g_on_file_loaded_callback_for_testing_->is_null());
+      std::move(*g_on_file_loaded_callback_for_testing_).Run();
+    }
+  };
+
   if (json_data.empty()) {
     return;
   }
@@ -98,11 +109,6 @@ void QueryFilterComponentInstallerPolicy::ComponentReady(
   VLOG(1) << "Component ready, version " << version.GetString() << " in "
           << install_dir.value();
 
-  base::OnceClosure on_file_loaded_callback =
-      on_file_loaded_callback_for_testing_.is_null()
-          ? base::DoNothing()
-          : std::move(on_file_loaded_callback_for_testing_);
-
   base::ThreadPool::PostTaskAndReplyWithResult(
       FROM_HERE,
       {base::MayBlock(), base::TaskShutdownBehavior::SKIP_ON_SHUTDOWN,
@@ -110,8 +116,7 @@ void QueryFilterComponentInstallerPolicy::ComponentReady(
       base::BindOnce(
           &ReadQueryFilterFile,
           install_dir.AppendASCII(query_filter::kQueryFilterJsonFile)),
-      base::BindOnce(&OnQueryFilterFileRead, version)
-          .Then(std::move(on_file_loaded_callback)));
+      base::BindOnce(&OnQueryFilterFileRead, version));
 }
 
 base::FilePath QueryFilterComponentInstallerPolicy::GetRelativeInstallDir()
@@ -136,6 +141,11 @@ QueryFilterComponentInstallerPolicy::GetInstallerAttributes() const {
 
 bool QueryFilterComponentInstallerPolicy::IsBraveComponent() const {
   return true;
+}
+
+void QueryFilterComponentInstallerPolicy::SetOnFileLoadedCallbackForTesting(
+    base::OnceClosure* callback) {
+  g_on_file_loaded_callback_for_testing_ = callback;
 }
 
 void RegisterQueryFilterComponent(
--- a/components/query_filter/browser/query_filter_component_installer.cc
+++ b/components/query_filter/browser/query_filter_component_installer.cc
@@ -27,6 +27,8 @@
 
 namespace {
 inline constexpr char kQueryFilterComponentName[] = "Query Filter";
+
+// IN-TEST
 base::OnceClosure* g_on_file_loaded_callback_for_testing_ = nullptr;
 
 // This is the SHA-256 of the query-filter component's public key.
@@ -143,8 +145,9 @@ bool QueryFilterComponentInstallerPolicy::IsBraveComponent() const {
   return true;
 }
 
-void QueryFilterComponentInstallerPolicy::SetOnFileLoadedCallbackForTesting(
-    base::OnceClosure* callback) {
+void QueryFilterComponentInstallerPolicy::
+    SetOnFileLoadedCallbackForTesting(  // IN-TEST
+        base::OnceClosure* callback) {
   g_on_file_loaded_callback_for_testing_ = callback;
 }
 
--- a/components/query_filter/browser/query_filter_component_installer.cc
+++ b/components/query_filter/browser/query_filter_component_installer.cc
@@ -51,6 +51,7 @@ void OnQueryFilterFileRead(const base::Version& version,
 
   if (!data->PopulateDataFromComponent(json_data)) {
     LOG(WARNING) << "Failed to populate data from component";
+    return;
   }
   data->UpdateVersion(version);
 }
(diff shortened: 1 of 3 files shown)


components/query_filter/browser/query_filter_component_installer.cc
```diff
@@ -97,14 +98,20 @@ void QueryFilterComponentInstallerPolicy::ComponentReady(
   VLOG(1) << "Component ready, version " << version.GetString() << " in "
           << install_dir.value();
 
+  base::OnceClosure on_file_loaded_callback =
+      on_file_loaded_callback_for_testing_.is_null()
```
cdesouza-chromium: Don't make this a callback that is a member of this class. Callbacks have a shared ptr inside, which involves control blocks, so we are wasting memory. Is it the case that during tests we don't instantiate more than one instance of `QueryFilterComponentInstallerPolicy` at a time? If so, we can do just:

```cxx
namespace {
// Used by tests to receive a failure signal
base::OnceClosure* g_on_file_loaded_callback_for_testing_ = nullptr;
}  // namespace
```

Then have:
```cxx
  void QueryFilterComponentInstallerPolicy::SetOnFileLoadedCallbackForTesting(base::OnceClosure& callback) {
    g_on_file_loaded_callback_for_testing_ = &callback;
  }
```

Then in `OnQueryFilterFileRead` do:

```
void OnQueryFilterFileRead(const base::Version& version,
                           const std::string& json_data) {
  // ...
  // ...
  if (g_on_file_loaded_callback_for_testing_) {
     CHECK_IS_TEST();
     CHECK(!g_on_file_loaded_callback_for_testing_.is_null());
     std::move(*g_on_file_loaded_callback_for_testing_).Run();
  }
}
```

This way we have only a single pointer that is guaranteed to be on the stack.

components/query_filter/browser/query_filter_component_installer_unittest.cc
```diff
@@ -166,18 +166,18 @@ TEST_F(QueryFilterComponentInstallerTest, TestComponentReady) {
 
   // Initiate component ready which would load the json file and populate the
   // query filter data.
+  base::RunLoop run_loop;
   component_updater::QueryFilterComponentInstallerPolicy policy;
+  policy.set_on_file_loaded_callback_for_testing(run_loop.QuitClosure());
```
cdesouza-chromium: Use something like this class in your tests:

```cxx
namespace {
class ScopedFileLoadedCallbackForTesting {                                                                                                                                                                             
   public:                                                                                                                                                                                                                 
    ScopedFileLoadedCallbackForTesting(                                                                                                                                                                                    
        QueryFilterComponentInstallerPolicy& policy,                                                                                                                                                                       
        base::OnceClosure callback)                                                                                                                                                                                        
        : policy_(policy), callback_(std::move(callback)) {                                                                                                                                                                
      policy_->SetOnFileLoadedCallbackForTesting(&callback_);
    }                                                                                                                                                                                                                      
                  
    ~ScopedFileLoadedCallbackForTesting() {
      policy_->SetOnFileLoadedCallbackForTesting(nullptr);
    }                                                                                                                                                                                                                      
   
    ScopedFileLoadedCallbackForTesting(                                                                                                                                                                                    
        const ScopedFileLoadedCallbackForTesting&) = delete;
    ScopedFileLoadedCallbackForTesting& operator=(
        const ScopedFileLoadedCallbackForTesting&) = delete;                                                                                                                                                               
   
   private:                                                                                                                                                                                                                
    raw_ref<QueryFilterComponentInstallerPolicy> policy_;
    base::OnceClosure callback_;
  };
}  namespace
```
This will mange the pointer assignment and making sure we won't get it wrong, and it will also own the callback.

components/query_filter/browser/query_filter_component_installer_unittest.cc
```diff
@@ -166,18 +166,18 @@ TEST_F(QueryFilterComponentInstallerTest, TestComponentReady) {
 
   // Initiate component ready which would load the json file and populate the
   // query filter data.
+  base::RunLoop run_loop;
```
cdesouza-chromium: We are encouraged to not use `base::RunLoop`. Let's just use:
```cxx
base::test::TestFuture<void> future; 
component_updater::QueryFilterComponentInstallerPolicy policy;
ScopedFileLoadedCallbackForTesting scoped(policy_, future.GetCallback());
policy.ComponentReady(version, GetInstallDirectoryPath(), base::DictValue());
ASSERT_TRUE(future.Wait());

**Review by cdesouza-chromium — COMMENTED** [review]
Thanks Rohit.. we are nearly there. Just a few things that you were not aware of.

File: components/query_filter/browser/query_filter_component_installer_unittest.cc (hunk 3 above)
Comment (r0hit0303) [reply components/query_filter/browser/query_filter_component_installer_unittest.cc]: Sorry missed this. I was under the impression that this was fine because we were passing the callback explicitly to make the wait more deterministic, and that RunUntillde was not permitted (because it waited only for the current message loop to be over, and our test/callback could arrive after undeterministically). 

However, indeed as you said from the testing best practices it seems future is [preferable](https://github.com/brave/brave-core/blob/master/docs/best-practices/testing-async.md#-always-use-basetesttestfuture-over-runloop-for-callbacks). I missed it but I'll make sure to keep it mind for future. Thanks! :)

File: components/query_filter/browser/query_filter_component_installer.cc (hunk 1 above)
Comment (r0hit0303) [reply components/query_filter/browser/query_filter_component_installer.cc]: > Callbacks have a shared ptr inside, which involves control blocks, so we are wasting memory. 

I might be wrong here but I don't think that's the case i.e having an overhead of a control block(like std::shared_ptr for internal ref_counts and weak_ptr counts) with base::OnceClosure (which is an alias for base::OnceCallback). There could be situations though that the functor itself which is binded to could be arbitrarily large (in contrast with lambdas) but that would never be the case for us here. Going through the codebase we have:

1) base::OnceCallback internally holds one member variable of type [internal::BindStateHolder](https://source.chromium.org/chromium/chromium/src/+/main:base/functional/callback.h;drc=30b305c5f320177757a739df0d0b60820684fa17;bpv=1;bpt=1;l=276) which intern holds a [scoped_refptr\<BindStateBase\>](https://source.chromium.org/chromium/chromium/src/+/main:base/functional/callback_internal.h;drc=30b305c5f320177757a739df0d0b60820684fa17;l=158) and that in turn is already a [nullptr](https://source.chromium.org/chromium/chromium/src/+/main:base/memory/scoped_refptr.h;drc=30b305c5f320177757a739df0d0b60820684fa17;l=366). 

2) To be 100% sure I also added logs to check the size in real time and indeed found that base::OnceCallback had allocated only 8 bytes. 

```
[51481:13966464:ERROR:brave/components/query_filter/browser/query_filter_component_installer.cc:106] sizeof: on_file_loaded_callback8
[51481:13966464:ERROR:brave/components/query_filter/browser/query_filter_component_installer.cc:108] sizeof: on_file_loaded_callback pointer8
```
Therefore, I don't think there's any difference memory wise doing the current way or the way you are proposing. 

-- 
> Then in OnQueryFilterFileRead do:

I think that code wouldn't take into account the early returns we have in place which would result in not firing our callback at all.  I think a cleaner way could be to rely on the [Then](https://chromium.googlesource.com/chromium/src/+/HEAD/docs/callback.md#chaining-callbacks) functionality to chain the callback, which would then unconditionally call any registered callbacks after the  `OnQueryFilterFileRead` : 

```
void FireCallbackIfNeeded() {
  if (g_on_file_loaded_callback_for_testing_) {
    CHECK_IS_TEST();
    CHECK(!g_on_file_loaded_callback_for_testing_->is_null());
    std::move(*g_on_file_loaded_callback_for_testing_).Run();
  }
}

  base::ThreadPool::PostTaskAndReplyWithResult(
      FROM_HERE,
      {base::MayBlock(), base::TaskShutdownBehavior::SKIP_ON_SHUTDOWN,
       base::TaskPriority::BEST_EFFORT},
      base::BindOnce(
          &ReadQueryFilterFile,
          install_dir.AppendASCII(query_filter::kQueryFilterJsonFile)),
      base::BindOnce(&OnQueryFilterFileRead, version)
          .Then(base::BindOnce(&FireCallbackIfNeeded)));
```
If there was no other reason to use global variable I feel having it as a member variable and making `FireCallbackIfNeeded` as a member method would be more comprehensible, unless you have strong objections in which case I'm happy to change.

cdesouza-chromium: `.Then` is a no go there, because we only need that for test. Production code should not have these types of hoops for test only code. We should NOT be binding `DoNothing` there in production.

Regarding the pointer, I guess `scoped_refptr` only attaches the control block once engaged, as `shared_ptr` in chromium is shared from this, so in disengaged mode it has only a pointer.

However, the pointer is still better. It is either a pointer, or a `NoDestructor` for that value to be global, and I don't see any gain in that if we are using it only for tests, and with something managing the setting of the pointer. If we use a `NoDestructor` here, it will cause cache syncs, etc, and we don't want that, so let's use just the pointer.

components/query_filter/browser/query_filter_component_installer_unittest.cc
r0hit0303: Done

components/query_filter/browser/query_filter_component_installer_unittest.cc
r0hit0303: Done

components/query_filter/browser/query_filter_component_installer.cc
r0hit0303: Done. Switched to using a global variable like you suggested.

cdesouza-chromium (approved):
LGTM

r0hit0303:
Thanks Claudio for the review.

PR #36039, opened 2026-04-30
Tag Brave Search from Shortcuts widget with source=ios-widget

EliHini:
Resolves https://github.com/brave/internal/issues/1714

## Summary
Adds Brave Search attribution for queries started from the **home screen Shortcuts widget** so QPD and related metrics can separate widget traffic from other iOS omnibox usage.

## Implementation notes
- Brave’s OpenSearch plugin already sends `source=ios`; `URL.updatingSourceWithWidgetAttribution()` performs the host-safe (`search.brave.com`) query rewrite for `source=ios` only.
- This mirrors the implementation on Android  (`source=android` → `source=android-widget` for widget intents).
- Widget initiated search happens on the new tab created for it, or the first empty tab of the tab collection. 
- Subsequent searches after the first tab are regular searches
- Once the user navigates away from the widget search tab, subsequent searches on the tab are regular searches


ios/brave-ios/Sources/Brave/Frontend/Browser/BrowserViewController/BrowserViewController.swift
```diff
@@ -114,6 +114,8 @@ public class BrowserViewController: UIViewController {
   var searchController: SearchViewController?
   var favoritesController: FavoritesViewController?
 
+  private var tabIdPendingShortcutsWidgetBraveSearchAttribution: UUID?
```
nuo-xu: why this whole thing is associated with tab.id? feel like it adds complexity but doesn't provide meaningful protection over a plain Bool. 
it also has logic flaws: widget opens → user presses cancel → user manually opens URL bar on the same tab → searches → flag is still set, tab.id still matches → wrongly attributes.

ios/brave-ios/Sources/Brave/Frontend/Browser/BrowserViewController/BrowserViewController.swift
```diff
@@ -3088,10 +3096,15 @@ extension BrowserViewController {
...
+  }
+
   public func submitSearchText(_ text: String, isBraveSearchPromotion: Bool = false) {
-    var engine = profile.searchEngines.defaultEngine(
-      forType: privateBrowsingManager.isPrivateBrowsing ? .privateMode : .standard
-    )
+    let engineType: DefaultEngineType =
+      privateBrowsingManager.isPrivateBrowsing ? .privateMode : .standard
+    let defaultEngine = profile.searchEngines.defaultEngine(forType: engineType)
+    var engine = defaultEngine
```
nuo-xu: they are the same

ios/brave-ios/Sources/Brave/Frontend/Browser/BrowserViewController/BrowserViewController.swift
```diff
@@ -3107,10 +3120,19 @@ extension BrowserViewController {
...
+          url = url.updatingSourceWithWidgetAttribution()
+        }
+        tabIdPendingShortcutsWidgetBraveSearchAttribution = nil
+      }
+      finishEditingAndSubmit(url)
     } else {
       // We still don't have a valid URL, so something is broken. Give up.
+      if tabManager.selectedTab?.id == tabIdPendingShortcutsWidgetBraveSearchAttribution {
+        tabIdPendingShortcutsWidgetBraveSearchAttribution = nil
+      }
```
nuo-xu: logic flaws as i mentioned above

ios/brave-ios/Sources/Brave/Frontend/Browser/BrowserViewController/BrowserViewController.swift
```diff
@@ -3107,10 +3120,19 @@ extension BrowserViewController {
       text,
       isBraveSearchPromotion: isBraveSearchPromotion
     ) {
-      // We couldn't find a matching search keyword, so do a search query.
-      finishEditingAndSubmit(searchURL)
+      var url = searchURL
+      if tabManager.selectedTab?.id == tabIdPendingShortcutsWidgetBraveSearchAttribution {
+        if defaultEngine?.shortName == OpenSearchEngine.EngineNames.brave {
+          url = url.updatingSourceWithWidgetAttribution()
```
nuo-xu: update search url can be done inside `searchURLForQuery(_ query: String, locale: Locale = Locale.current, isBraveSearchPromotion: Bool = false)`. probably similar to how we handle `isBraveSearchPromotion`

nuo-xu (changes requested):

ios/brave-ios/Sources/Shared/Extensions/URLExtensions.swift
```diff
@@ -93,6 +93,32 @@ extension URL {
     return components.url!
   }
 
+  /// Rewrites Brave Search URLs that use the app’s default `source=ios` (OpenSearch template) to `source=ios-widget`.
+  /// To be called after the browser has already determined the omnibox search came from the Shortcuts widget.
+  public func updatingSourceWithWidgetAttribution() -> URL {
```
kylehickinson: Don't add new Brave code to `Shared`, use `BraveShared`

kylehickinson: This should just exist in an isolated tab helper though

kylehickinson (changes requested):
Refactor this to be isolated to a tab helper, setting some state inside of it when a widget creates said tab, ensure that you reset that state if the user navigates to other pages before searching, use that state to either cancel & re-submit the request in ~decidePolicy~ shouldAllowRequest with the attribution when it needs to add the attribution or read from submission/searchURLForQuery

ios/brave-ios/Sources/Shared/Extensions/URLExtensions.swift
```diff
@@ -93,6 +93,32 @@ extension URL {
...
+    else { return self }
+
+    components.queryItems = items.map { item in
+      (item.name == BraveSearchSource.queryKey && item.value == BraveSearchSource.appValue)
+        ? URLQueryItem(name: BraveSearchSource.queryKey, value: BraveSearchSource.widgetValue)
+        : item
+    }
+    return components.url ?? self
+  }
+
```
netzenbot: `updatingSourceWithWidgetAttribution()` is Brave Search-specific logic and doesn't belong in the generic `URL` extension in `Shared`. Per [IOS-045](https://github.com/brave/brave-core/tree/master/docs/best-practices/ios.md#IOS-045), feature-specific helpers should be created in a feature-specific location rather than added to generic utility extensions.

ios/brave-ios/Sources/Shared/Extensions/URLExtensions.swift
```diff
@@ -93,6 +93,32 @@ extension URL {
...
   public func getQuery() -> [String: String] {
```
netzenbot, line 96: nit: The doc comment describes the rewrite case but doesn't mention that the URL is returned unchanged when the guard conditions aren't met (non-Brave-Search host, missing `source=ios` param, etc.). Per [DOC-004](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-004), public method docs should describe the full contract including fallback return behavior.

ios/brave-ios/Sources/Brave/Frontend/Browser/BrowserViewController/BrowserViewController.swift
EliHini: I am fixing the complexity.

> it also has logic flaws: widget opens → user presses cancel → user manually opens URL bar on the same tab → searches → flag is still set, tab.id still matches → wrongly attributes.

Why isn't that a valid widget search. When the user hits the cancel button are they really meaning to cancel the widget search? 
cc @timchilds

ios/brave-ios/Sources/Brave/Frontend/Browser/BrowserViewController/BrowserViewController.swift
EliHini: I elected to not do it inside of `searchURLForQuery(...)` to separate the concerns, and because `WidgetSearchTabHelper` tears itself down after it has applied the source attribution. That logic needs to stay with the helper

ios/brave-ios/Sources/Shared/Extensions/URLExtensions.swift
EliHini: Done

ios/brave-ios/Sources/Brave/Frontend/Browser/Search/WidgetSearchTabHelper.swift
```diff
@@ -0,0 +1,96 @@
...
+  static func applyWidgetAttribution(_ url: URL) -> URL {
+    guard url.host?.lowercased() == BraveSearchSource.host,
+      var components = URLComponents(url: url, resolvingAgainstBaseURL: false),
+      let items = components.queryItems,
+      items.contains(where: {
+        $0.name == BraveSearchSource.queryKey && $0.value == BraveSearchSource.appValue
+      })
+    else { return url }
+
+    components.queryItems = items.map { item in
```
netzenbot: nit: The doc comment for `applyWidgetAttribution` describes the rewrite case but doesn't mention that the URL is returned unchanged when the guard conditions aren't satisfied (non-Brave-Search host, missing `source=ios` param, etc.). Per [DOC-004](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-004), method docs should describe the full contract including fallback return behavior.

ios/brave-ios/Sources/Brave/Frontend/Browser/Search/WidgetSearchTabHelper.swift
```diff
@@ -0,0 +1,96 @@
...
+  /// Removes this helper from its tab. After this call the helper is no longer reachable
+  /// via `TabDataValues` and will be deallocated once the current call stack unwinds.
+  private func detach() {
+    tab?.widgetSearchTabHelper = nil
+  }
+
+  // MARK: - URL rewriting
+
+  /// Rewrites `source=ios` to `source=ios-widget` on `search.brave.com`.
+  static func applyWidgetAttribution(_ url: URL) -> URL {
```
kylehickinson: Feels like this logic should just be part of `finalize`? Why is it separate?

ios/brave-ios/Sources/Brave/Frontend/Browser/Search/SearchViewController.swift
```diff
@@ -849,11 +851,15 @@ public class SearchViewController: UIViewController, LoaderListener {
     let offset = dataSource.isAIChatAvailable ? 1 : 0  // offset for the Leo button
     let engine = dataSource.quickSearchEngines[index - offset]
     let localSearchQuery = dataSource.searchQuery.lowercased()
-    guard let url = engine.searchURLForQuery(localSearchQuery) else {
+    guard var url = engine.searchURLForQuery(localSearchQuery) else {
```
kylehickinson: Is there a reason we can't pass in like "isTabCreatedByWidget" or something into `searchURLForQuery` instead so that the logic of adding the attribution is tied to the engine instead similar to `isBraveSearchPromotion`?

ios/brave-ios/Sources/Brave/Frontend/Browser/Search/SearchViewController.swift
```diff
@@ -209,6 +210,7 @@ public class SearchViewController: UIViewController, LoaderListener {
   )
 
   weak var searchDelegate: SearchViewControllerDelegate?
+  weak var associatedTab: (any TabState)?
```
kylehickinson: Probably just want to use `SearchViewControllerDelegate` to ask if the tab is created by a widget

EliHini: I've removed the directed dependence no `associatedTab`. `SearchViewController` asks `SearchViewControllerDelegate` via `searchViewControllerIsWidgetInitiatedSearchSession` (browser answers from `selectedTab?.widgetSearchTabHelper != nil`) and `searchViewControllerFinalizePendingWidgetSearch` after committing a URL built with widget attribution when that was true. 

I named it widget-initiated search session rather than literally “tab created by widget” because the widget search shortcut can reuse an existing NTP tab without opening a new one.

ios/brave-ios/Sources/Brave/Frontend/Browser/Search/SearchViewController.swift
EliHini: Done. I added `isWidgetSearchAttribution` to `searchURLForQuery` / `getURLFromTemplate`, parallel to `isBraveSearchPromotion`.

ios/brave-ios/Sources/Brave/Frontend/Browser/Search/WidgetSearchTabHelper.swift
EliHini: To stay consistent with `BraveSearchPromotion` I've relocated this logic into `OpenSearchEngine`

ios/brave-ios/Sources/Brave/Frontend/Browser/Search/SearchViewController.swift
```diff
@@ -849,11 +859,19 @@ public class SearchViewController: UIViewController, LoaderListener {
...
+      searchDelegate?.searchViewControllerIsWidgetInitiatedSearchSession(self) ?? false
+    guard
+      let url = engine.searchURLForQuery(localSearchQuery, isWidgetSearchAttribution: widgetAttr)
+    else {
       assertionFailure()
       return
     }
 
+    if widgetAttr {
+      searchDelegate?.searchViewControllerFinalizePendingWidgetSearch(self)
```
kylehickinson: won't this detach anyways once you submit the request and the navigation commits?

ios/brave-ios/Sources/Brave/Frontend/Browser/BrowserViewController/BrowserViewController.swift
```diff
@@ -3124,11 +3132,16 @@ extension BrowserViewController {
...
+    let hasPendingWidgetSearch = tabManager.selectedTab?.widgetSearchTabHelper != nil
     if let searchURL = engine?.searchURLForQuery(
       text,
-      isBraveSearchPromotion: isBraveSearchPromotion
+      isBraveSearchPromotion: isBraveSearchPromotion,
+      isWidgetSearchAttribution: hasPendingWidgetSearch
     ) {
       // We couldn't find a matching search keyword, so do a search query.
+      if hasPendingWidgetSearch {
+        tabManager.selectedTab?.widgetSearchTabHelper?.finalize()
```
kylehickinson: same here

ios/brave-ios/Sources/Brave/Frontend/Browser/Search/SearchViewController.swift
EliHini: For the happy path yes—`tabDidCommitNavigation` clears the helper once we commit off the NTP.

The explicit finalize() is there to drop `widgetSearchTabHelper` immediately when we submit, so we don’t leave a window where the helper is still present before navigation commits (quick second search / another URL build could still see “widget session” and apply `isWidgetSearchAttribution` again).

kylehickinson: Can you clarify what path you're worried about? 

Should it not be for the first happy path? If a user somehow submits another search between submitting the first and the navigation committing then they still never actually performed a search yet. And its going to be tied to a specific tab anyways not window so the only path thats not happy is the user cancels their request/internet fails and then they search again in which case its still the same tab coming from a widget so seems fine.

kylehickinson (changes requested):
The previous solution was better/more straightforward, please revert latest commit

ios/brave-ios/Sources/Brave/Frontend/Browser/Search/OpenSearch.swift
```diff
@@ -119,16 +126,41 @@ class OpenSearchEngine: NSObject, NSSecureCoding {
   func searchURLForQuery(
     _ query: String,
     locale: Locale = Locale.current,
-    isBraveSearchPromotion: Bool = false
+    isBraveSearchPromotion: Bool = false,
+    isWidgetSearchAttribution: Bool = false
   ) -> URL? {
     return getURLFromTemplate(
       searchTemplate,
```
netzenbot, line 133: nit: The doc comment describes the rewrite case but doesn't mention that the URL is returned unchanged when conditions aren't met (non-Brave-Search host, missing `source=ios` param, `nil` input, etc.). Per [DOC-004](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-004), public method docs should describe the full contract including fallback return behavior.

ios/brave-ios/Sources/Brave/Frontend/Browser/Search/SearchViewController.swift
kylehickinson: You never answered me here and the code still calls exposes & calls `finalize` here: https://github.com/brave/brave-core/pull/36039/changes#r3242727361

kylehickinson (changes requested):
Still havent answered this question: https://github.com/brave/brave-core/pull/36039/changes#r3249289838

ios/brave-ios/Sources/Brave/Frontend/Browser/Search/SearchViewController.swift
EliHini: The explicit `finalize()` grew out of @nuo-xu's earlier feedback on the `tabIdPendingShortcutsWidgetBraveSearchAttribution approach` from https://github.com/brave/brave-core/pull/36039#discussion_r3168571685: 

> widget opens → user cancels → user manually searches on the same tab → the flag still matches → wrongful attribution.

The `WidgetSearchTabHelper` refactor is meant to replace that with tab-scoped state that tears down on navigation commit.

I added finalize() on submit because I was worried about a lingering-attribution window in that same spirit: after the first search URL is built/submitted, the helper stays attached until navigation commits, so a second URL build on that tab would also see `widgetSearchTabHelper != nil`.
I treated that as incorrect double-attribution. _Sounds like this is a wrong assumption: A user should be able to use the same tab of searches and it should be attributed as a widget search._

Thinking over again, clearing on submit isn't the right fix as per the following scenario that convinced me (which was my initial pushback in my response but I didn't hear back):

Widget opens NTP, user submits `"tab"`, then quickly corrects to `"tap"` and submits again before the first navigation commits. With finalize() on the first submit, the helper is cleared immediately, so the search that actually commits `"tap"` loses widget attribution even though the user never left the widget-initiated flow.

Since the helper is tab-scoped and `tabDidCommitNavigation` handles teardown once something real commits, the explicit `finalize()` calls aren't needed. I'll remove them.

ios/brave-ios/Sources/Brave/Frontend/Browser/BrowserViewController/BrowserViewController.swift
EliHini: Removing it

kylehickinson (approved):

ios/brave-ios/Sources/Brave/Frontend/Browser/BrowserViewController/BrowserViewController.swift
EliHini: Fixed

ios/brave-ios/Sources/Brave/Frontend/Browser/Search/WidgetSearchTabHelper.swift
```diff
@@ -0,0 +1,57 @@
...
+    tab.removeObserver(self)
+  }
+
+  // MARK: - Teardown
+
+  /// Removes this helper from its tab. After this call the helper is no longer reachable
+  /// via `TabDataValues` and will be deallocated once the current call stack unwinds.
+  private func detach() {
+    tab?.widgetSearchTabHelper = nil
+  }
```
nuo-xu: nit: can just inline it. its a one line function and not being called anywhere else.

nuo-xu (approved):

ios/brave-ios/Sources/Brave/Frontend/Browser/Search/WidgetSearchTabHelper.swift
EliHini: ah yes, the other callers are gone so there's no need for it to be a stand alone function

netzenbot (approved):

PR #36048, opened 2026-04-30
[Polkadot] Validate recepient ss58 prefix

### LINKED ISSUE #55134: [Polkadot] Validate recipient address format when typed
### Description

Polkadot recipient address should be validated over valid prefix and address format. Error should be shown if not.

### Steps to reproduce

-

### Actual result

-

### Expected result

-

### Reproduces how often

Easily reproduced

-

### Channel information

- [ ] release (stable)
- [ ] beta
- [ ] nightly

### Reproducibility

### Miscellaneous information

_No response_

cypt4:
Resolves https://github.com/brave/brave-browser/issues/55134

Adds validation steps for types\selected addresses to match chain id format.

Test plan:
Validate polkadot send transaction recipient address validation. Validate error is displayed when wrong ss58 prefix is used(prefix can be changed using https://ss58.org/) or wrong address is constructed(symbols are missing for ex.).

### DIFF
--- a/components/brave_wallet/browser/polkadot/polkadot_wallet_service.cc
+++ b/components/brave_wallet/browser/polkadot/polkadot_wallet_service.cc
@@ -14,6 +14,7 @@
 #include "brave/components/brave_wallet/browser/keyring_service.h"
 #include "brave/components/brave_wallet/browser/network_manager.h"
 #include "brave/components/brave_wallet/common/common_utils.h"
+#include "brave/components/brave_wallet/common/encoding_utils.h"
 
 namespace brave_wallet {
 
@@ -136,6 +137,23 @@ void PolkadotWalletService::GetAccountBalance(
 
   polkadot_substrate_rpc_.GetAccountBalance(chain_id, *pubkey,
                                             std::move(callback));
+}
+
+void PolkadotWalletService::ValidateAddressForTransaction(
+    const std::string& chain_id,
+    const std::string& address,
+    ValidateAddressForTransactionCallback callback) {
+  auto network = network_manager_->GetChain(chain_id, mojom::CoinType::DOT);
+  if (!network) {
+    std::move(callback).Run(mojom::PolkadotAddressError::kInvalidAddressFormat);
+    return;
+  }
+
+  metadata_provider_.GetChainMetadata(
+      chain_id,
+      base::BindOnce(
+          &PolkadotWalletService::OnGetChainMetadataForValidateAddress,
+          weak_ptr_factory_.GetWeakPtr(), address, std::move(callback)));
 }
 
 void PolkadotWalletService::GenerateSignedTransferExtrinsicImpl(
@@ -263,6 +281,25 @@ void PolkadotWalletService::OnEstimatedFee(
     GetFeeEstimateCallback callback,
     base::expected<uint128_t, std::string> partial_fee) {
   std::move(callback).Run(std::move(partial_fee));
+}
+
+void PolkadotWalletService::OnGetChainMetadataForValidateAddress(
+    const std::string& address,
+    ValidateAddressForTransactionCallback callback,
+    base::expected<PolkadotChainMetadata, std::string> metadata) {
+  if (!metadata.has_value()) {
+    std::move(callback).Run(mojom::PolkadotAddressError::kInvalidAddressFormat);
+    return;
+  }
+
+  auto parsed_address =
+      ParsePolkadotAccount(address, metadata->GetSs58Prefix());
+  if (!parsed_address.has_value()) {
+    std::move(callback).Run(parsed_address.error());
+    return;
+  }
+
+  std::move(callback).Run(mojom::PolkadotAddressError::kNoError);
 }
 
 }  // namespace brave_wallet
--- a/components/brave_wallet/browser/polkadot/polkadot_wallet_service.cc
+++ b/components/brave_wallet/browser/polkadot/polkadot_wallet_service.cc
@@ -288,7 +288,8 @@ void PolkadotWalletService::OnGetChainMetadataForValidateAddress(
     ValidateAddressForTransactionCallback callback,
     base::expected<PolkadotChainMetadata, std::string> metadata) {
   if (!metadata.has_value()) {
-    std::move(callback).Run(mojom::PolkadotAddressError::kInvalidAddressFormat);
+    std::move(callback).Run(
+        mojom::PolkadotAddressError::kFailedToFetchMetadata);
     return;
   }
 
--- a/components/brave_wallet/browser/polkadot/polkadot_wallet_service.cc
+++ b/components/brave_wallet/browser/polkadot/polkadot_wallet_service.cc
@@ -145,7 +145,8 @@ void PolkadotWalletService::ValidateAddressForTransaction(
     ValidateAddressForTransactionCallback callback) {
   auto network = network_manager_->GetChain(chain_id, mojom::CoinType::DOT);
   if (!network) {
-    std::move(callback).Run(mojom::PolkadotAddressError::kInvalidAddressFormat);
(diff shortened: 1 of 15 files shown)


cmazakas:
This PR might be duplicating some functionality a little bit. We already validate the recipient address during transaction creation, at the unapproved transaction stage.

Namely, we have:
https://github.com/brave/brave-core/blob/59690c19274f1b2e73ad8b5b9fc9351b40c9afec/components/brave_wallet/browser/polkadot/polkadot_utils.cc#L235-L260

which is called here:
https://github.com/brave/brave-core/blob/59690c19274f1b2e73ad8b5b9fc9351b40c9afec/components/brave_wallet/browser/polkadot/polkadot_tx_manager.cc#L172-L184

This PR seems more focused on the UI component which makes me wonder if it makes more sense to maybe just update:
```cpp
  auto recipient =
      ParsePolkadotAccount(params->to, chain_metadata->GetSs58Prefix());
  if (!recipient) {
    return std::move(callback).Run(false, "", WalletInternalErrorMessage());
  }
```
to return a proper error message, instead of just `WalletInternalErrorMessage()`.

That, or we should unify the checks here between the PolkadotWalletService and the PolkadotTxManager.

components/brave_wallet_ui/page/screens/send/components/select_address_modal/select_address_modal.tsx
```diff
@@ -294,6 +307,7 @@ export const SelectAddressModal = React.forwardRef<HTMLDivElement, Props>(
           coinType: fromAccountId?.coin ?? BraveWallet.CoinType.ETH,
```
netzenbot, line 310: `getPolkadotAddressValidationResult` is referenced inside this `useMemo` call but is missing from the dependency array (lines 305-320). This means `addressMessageId` will stale-cache and never update when the Polkadot address validation result changes — the prefix error will silently not appear. Add `getPolkadotAddressValidationResult` to the dependency array.

components/brave_wallet_ui/page/screens/send/components/select_address_modal/select_address_modal.tsx
```diff
@@ -294,6 +307,7 @@ export const SelectAddressModal = React.forwardRef<HTMLDivElement, Props>(
...
           token: selectedAsset,
           zcashAddressError: getZCashTransactionTypeResult.error,
+          polkadotAddressError: getPolkadotAddressValidationResult,
           fullTokenList,
```
netzenbot, line 314: `getPolkadotAddressValidationResult` is passed to `processAddressOrUrl` inside the `useMemo` for `addressMessageId`, but it is missing from the dependency array. This means `addressMessageId` won't recompute when the Polkadot validation result changes, so the UI will show stale address errors. Add `getPolkadotAddressValidationResult` to the dependency array.

components/brave_wallet/common/brave_wallet.mojom
```diff
@@ -2269,6 +2269,12 @@ struct PolkadotAccountInfo {
   PolkadotAccountBalance data;
 };
 
+enum PolkadotAddresError {
+  kNoError = 0,
+  kInvalidPrefix = 1,
+  kInvalidAddressFormat = 2,
```
netzenbot, line 2275: `PolkadotAddresError` is missing an 's' — should be `PolkadotAddressError`. This typo then propagates through all the C++ and TypeScript code that references it. ([✅ Grammatical Correctness](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-006))

components/brave_wallet/browser/polkadot/polkadot_wallet_service_unittest.cc
```diff
@@ -476,6 +476,49 @@ TEST_F(PolkadotWalletServiceUnitTest, GetAddress) {
...
+          .spec();
+  std::string mainnet_url =
+      network_manager_
+          ->GetKnownChain(mojom::kPolkadotMainnet, mojom::CoinType::DOT)
+          ->rpc_endpoints.front()
+          .spec();
+  AddValidMetadataResponses(url_loader_factory_, testnet_url, mainnet_url);
+
+  {
+    base::test::TestFuture<mojom::PolkadotAddresError> future;
```
netzenbot: The implementation's first check is `if (!network)` when the chain_id is unknown, returning `kInvalidAddressFormat` — but no test covers this path. The existing `GetAddress` test already has an equivalent case using `"asd"` as the chain_id. See [TI-016](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-016).

components/brave_wallet/browser/polkadot/polkadot_wallet_service_unittest.cc
```diff
@@ -476,6 +476,49 @@ TEST_F(PolkadotWalletServiceUnitTest, GetAddress) {
...
+  }
+
+  {
+    base::test::TestFuture<mojom::PolkadotAddresError> future;
+    polkadot_wallet_service->ValidateAddressForTransaction(
+        mojom::kPolkadotMainnet, "not-an-address", future.GetCallback());
+    EXPECT_EQ(future.Get(), mojom::PolkadotAddresError::kInvalidAddressFormat);
+  }
+}
+
```
netzenbot, line 533: Two branches in `ValidateAddressForTransaction` and `OnGetChainMetadataForValidateAddress` are not exercised: the early-return when `network_manager_->GetChain()` returns null (invalid chain ID), and the path when metadata fetch fails. Per [TI-039](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-039), these reachable error paths should have corresponding test cases.

mkarolin (approved):
`strings`++

components/brave_wallet/common/brave_wallet.mojom
```diff
@@ -2280,6 +2286,9 @@ interface PolkadotWalletService {
 
   GetAccountBalance(AccountId account, string chain_id)
       => (PolkadotAccountInfo? account, string? error_message);
+
+  ValidateAddressForTransaction(string chain_id, string address)
+      => (PolkadotAddresError error);
```
cmazakas: Is it possible to instead return `(PolkadotAddressError? error)` here?

The approach we have works but I'm not stoked on the naming of `PolkadotAddressError` if its first member is `kNoError`. Maybe it's better to just rename this enum to `PolkadotAddressStatus` or something along those lines.

components/brave_wallet/browser/polkadot/polkadot_wallet_service.cc
```diff
@@ -265,4 +283,27 @@ void PolkadotWalletService::OnEstimatedFee(
...
+  if (!metadata.has_value()) {
+    std::move(callback).Run(mojom::PolkadotAddresError::kInvalidAddressFormat);
+    return;
+  }
+
+  auto ss58_address = Ss58Address::Decode(address);
+  if (ss58_address && ss58_address->prefix != metadata->GetSs58Prefix()) {
+    std::move(callback).Run(mojom::PolkadotAddresError::kInvalidPrefix);
+    return;
+  }
```
cmazakas: This seems unnecessary because `ParsePolkadotAccount` already does this same check, unless I'm mistaken.

cmazakas:
I added some console logging and it seems like we fire 2 checks per keystroke when typing in an address manually. I guess this doesn't really matter but I tried looking into the code to see where the event handlers were being attached and got lost in the maze, ha ha.

It'd be nice to eventually get that worked out and cleaned up as well, but that's most likely completely out of scope for this PR.

Douglashdaniel (approved):
++ Wallet Front-end

cmazakas (approved):
Code LGTM but the netzenbot comment about missing a test path or two should probably be addressed.

PR #36060, opened 2026-04-30
Add temporary containers.

### LINKED ISSUE #55114: [Containers] Add support for temporary containers
Basic implementation with local-only containers with BIP-39 names, random colors and random icons (from the list of predefined ones for now).

Next step is to make icons be fully randomized glyph-like images.

goodov:
Add ability to add local-only temporary containers.

Temporary containers use random names generated from two BIP-39 words, icons and colors are chosen randomly from existing list we use for normal containers. Temporary containers are not syncable.

Resolves https://github.com/brave/brave-browser/issues/55114


mkarolin (approved):
`strings`, `chromium_src` ++

components/containers/core/browser/temporary_container.cc
```diff
@@ -0,0 +1,107 @@
...
+
+// Picks a random container background color from the available colors.
+SkColor PickTemporaryContainerBackground() {
+  const auto* color_provider =
+      ui::ColorProviderManager::Get().GetColorProviderFor(
+          ui::ColorProviderKey());
+  CHECK(color_provider);
+
+  // Keep in sync with
+  // brave/browser/resources/settings/brave_content_page/background_colors.ts.
```
sangwoo108, line 76: Thinking out loud: I'm wondering if we could have something like
  // LINT.IfChange(Foo)
  // Lint.ThenChange(Bar)
from upstream, which is checked by gerrit(i guess).

Maybe the netzenbot can understand this?

sangwoo108 (approved):
Nice work! LGTM

netzenbot (approved):

components/containers/core/browser/temporary_container.cc
goodov: I was thinking about a presubmit step for this, but turns out there's no ready implementation for this in chromium codebase. I think using some AI to generate a script to run such check doesn't sound that bad. The scenarios are clear, the logic should also be not that bad..

PR #36067, opened 2026-04-30
[Origin] Disable brave://getting-started in Brave Origin branded builds

### LINKED ISSUE #55035: Brave Origin should exclude WebUI page brave://getting-started
This WebUI page doesn't make sense in Brave Origin as most of the features on that page do not apply to Brave Origin. This shouldn't show up in Brave Origin ideally.

bbondy:
Most features promoted on `brave://getting-started` don't apply to Brave Origin/

Defaults `enable_brave_education` to off when `is_brave_origin_branded` is true, matching other buidl flags.

Resolves https://github.com/brave/brave-browser/issues/55035

## Test plan

- [ ] brave://getting-started should be available in normal builds
- [ ] brave://getting-started should NOT be available in Brave Origin builds

### DIFF
--- a/browser/ui/webui/welcome_page/welcome_dom_handler.cc
+++ b/browser/ui/webui/welcome_page/welcome_dom_handler.cc
@@ -16,8 +16,7 @@
 #include "base/values.h"
 #include "brave/browser/brave_browser_features.h"
 #include "brave/common/importer/importer_constants.h"
-#include "brave/components/brave_education/education_urls.h"
-#include "brave/components/brave_education/features.h"
+#include "brave/components/brave_education/buildflags.h"
 #include "brave/components/constants/pref_names.h"
 #include "brave/components/p3a/pref_names.h"
 #include "brave/components/web_discovery/buildflags/buildflags.h"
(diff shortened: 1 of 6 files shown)


browser/ui/webui/welcome_page/welcome_dom_handler.cc
```diff
@@ -208,6 +219,9 @@ void WelcomeDOMHandler::HandleGetWelcomeCompleteURL(
       brave_education::EducationPageType::kGettingStarted,
       base::BindOnce(&WelcomeDOMHandler::OnGettingStartedServerCheck,
                      weak_ptr_factory_.GetWeakPtr(), callback_id));
+#else
+  OnGettingStartedServerCheck(callback_id, /* available */ false);
+#endif
```
aseren: nit:
#endif   // BUILDFLAG(ENABLE_BRAVE_EDUCATION)

aseren: fine to ignore nit and can add comments as a follow-up, because other #endif in file don't have comments

aseren (approved):

netzenbot (approved):

bbondy:
I don't think we need to uplift this one to 1.91.x

PR #36076, opened 2026-04-30
[Brave Origin] Add switch to skip startup dialog on Linux

### LINKED ISSUE #55151: Provide command line option to skip Origin startup dialog for Linux
@henri  asks:
> Thank you for ongoing efforts to Brave. Also, thank you for offering Brave Origin as a donation on LINUX :) very cool.
> Would you please consider adding a command line option to skip the startup screen on Brave Origin (asking for donation)? An alternative approach could be that when starting a private window the pop-up is skipped by default?
> Having any pop-ups displayed when starting a browser Incognito mode is not desirable.

This dialog is only shown once. But it's ok to add a command line to skip it because it shows clear user intent.
This would be for Linux only where payment is not needed. The reason for this is maybe there's some kind of automation at play that would be more difficult because of it.

bbondy:
## Summary

- Adds a new `--skip-origin-startup-dialog` command-line switch.
- Honored only on Linux (where Brave Origin is free), where it makes `BraveOriginStartupView::ShouldShowDialog()` return false so the donation/startup dialog is bypassed.
- Useful for automation and scripted launches where the dialog would otherwise block startup, as discussed in the linked issue.

Resolves https://github.com/brave/brave-browser/issues/55151

## Test plan

- [x] On a Linux is_brave_origin_branded build with a fresh profile (dialog would normally appear), launch with `--skip-origin-startup-dialog` and confirm the browser starts directly without showing the dialog.
- [x] Launch the same build without the switch and confirm the dialog still appears as before.
- [x] On macOS / Windows is_brave_origin_branded builds, confirm the switch has no effect (dialog still appears since payment is required).
- [x] Run `brave_unit_tests --gtest_filter=BraveOriginStartupViewTest.*` and verify the new `ShouldNotShowDialogWhenSkipSwitchPresentOnLinux` test passes on Linux.

### DIFF
--- a/browser/ui/views/brave_origin/brave_origin_startup_view.cc
+++ b/browser/ui/views/brave_origin/brave_origin_startup_view.cc
@@ -36,6 +36,10 @@
 #include "url/gurl.h"
 #include "url/url_constants.h"
 
+#if BUILDFLAG(IS_LINUX)
+#include "brave/components/brave_origin/switches.h"
+#endif
+
 namespace {
 
 constexpr int kDialogWidth = 540;
@@ -85,6 +89,12 @@ bool BraveOriginStartupView::ShouldShowDialog(PrefService* local_state) {
   }
 
 #if BUILDFLAG(IS_LINUX)
+  if (base::CommandLine::ForCurrentProcess()->HasSwitch(
+          brave_origin::switches::kSkipOriginStartupDialog)) {
+    // Persist acceptance so future launches without the switch also skip
+    // the dialog.
+    local_state->SetBoolean(brave_origin::kOriginFreeTierAccepted, true);
+  }
   if (local_state->GetBoolean(brave_origin::kOriginFreeTierAccepted)) {
     return false;
   }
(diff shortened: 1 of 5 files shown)


browser/ui/views/brave_origin/brave_origin_startup_view.cc
```diff
@@ -15,6 +15,7 @@
 #include "brave/brave_domains/service_domains.h"
 #include "brave/browser/ui/webui/brave_origin_startup/brave_origin_startup_ui.h"
 #include "brave/components/brave_origin/pref_names.h"
+#include "brave/components/brave_origin/switches.h"
```
netzenbot: nit: `switches.h` is only used inside the `#if BUILDFLAG(IS_LINUX)` block, so the include should be inside that guard too. See [BS-034](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-034).

bbondy: Done in b23137fde02 — moved the include inside the `#if BUILDFLAG(IS_LINUX)` guard in both the .cc and the unittest.

SergeyZhukovsky (approved):
++

PR #36080, opened 2026-04-30
[AI Chat]: Add WebMCP support to Leo

fallaciousreasoning:
Series https://github.com/brave/brave-browser/issues/55232

Security/Privacy Review:
https://github.com/brave/reviews/issues/2243

### DIFF
--- a/browser/ai_chat/web_mcp_browsertest.cc
+++ b/browser/ai_chat/web_mcp_browsertest.cc
@@ -6,6 +6,8 @@
 #include <set>
 #include <string>
 
+#include "base/files/file_path.h"
+#include "base/path_service.h"
 #include "base/test/scoped_feature_list.h"
 #include "base/test/test_future.h"
 #include "brave/browser/ai_chat/ai_chat_service_factory.h"
@@ -16,6 +18,7 @@
 #include "brave/components/ai_chat/core/browser/associated_content_manager.h"
 #include "brave/components/ai_chat/core/browser/conversation_handler.h"
 #include "brave/components/ai_chat/core/browser/tools/tool.h"
+#include "brave/components/constants/brave_paths.h"
 #include "chrome/browser/profiles/profile.h"
 #include "chrome/browser/ui/browser.h"
 #include "chrome/test/base/in_process_browser_test.h"
@@ -25,8 +28,6 @@
 #include "content/public/test/content_mock_cert_verifier.h"
 #include "net/dns/mock_host_resolver.h"
 #include "net/test/embedded_test_server/embedded_test_server.h"
-#include "net/test/embedded_test_server/http_request.h"
-#include "net/test/embedded_test_server/http_response.h"
 #include "testing/gmock/include/gmock/gmock.h"
 #include "testing/gtest/include/gtest/gtest.h"
 #include "third_party/blink/public/common/features.h"
@@ -35,42 +36,12 @@ namespace ai_chat {
 
 namespace {
 
-// HTML that registers two tools via navigator.modelContext. The feature is
-// only available on secure contexts when blink::features::kWebMCPTesting is
-// enabled (which implies kWebMCP).
-constexpr char kPageWithTools[] = R"HTML(
-<!doctype html>
-<html><head><title>WebMCP test</title></head>
-<body>
-<script>
-  window.__webmcpRegisterError = null;
-  try {
-    navigator.modelContext.registerTool({
-      execute: async ({text}) => text,
-      name: "echo",
-      description: "Echo input back",
-      inputSchema: {
-        type: "object",
-        properties: {
-          text: { type: "string", description: "Value to echo" }
-        },
-        required: ["text"]
-      },
-    });
-    navigator.modelContext.registerTool({
-      execute: async () => "ok",
-      name: "ping",
-      description: "Returns ok",
-    });
-  } catch (e) {
-    window.__webmcpRegisterError = String(e);
-  }
-</script>
-</body></html>)HTML";
-
-constexpr char kPageWithoutTools[] =
-    "<!doctype html><html><head><title>plain</title></head>"
-    "<body>no tools here</body></html>";
+// Test pages served from test/data/leo/. The tool-registering page uses
+// navigator.modelContext, which is only available on secure contexts when
+// blink::features::kWebMCPTesting is enabled (which implies kWebMCP).
+constexpr char kPageWithToolsPath[] = "/web_mcp_tools.html";
+// A basic existing page that registers no tools.
+constexpr char kPageWithoutToolsPath[] = "/dummy.html";
 
 }  // namespace
 
@@ -92,8 +63,9 @@ class WebMcpBrowserTest : public InProcessBrowserTest {
     mock_cert_verifier_.mock_cert_verifier()->set_default_result(net::OK);
     host_resolver()->AddRule("*", "127.0.0.1");
 
-    https_server_.RegisterRequestHandler(base::BindRepeating(
-        &WebMcpBrowserTest::HandleRequest, base::Unretained(this)));
+    base::FilePath test_data_dir =
+        base::PathService::CheckedGet(brave::DIR_TEST_DATA);
+    https_server_.ServeFilesFromDirectory(test_data_dir.AppendASCII("leo"));
     ASSERT_TRUE(https_server_.Start());
   }
 
@@ -140,21 +112,6 @@ class WebMcpBrowserTest : public InProcessBrowserTest {
   net::EmbeddedTestServer* https_server() { return &https_server_; }
 
  private:
-  std::unique_ptr<net::test_server::HttpResponse> HandleRequest(
-      const net::test_server::HttpRequest& request) {
-    auto response = std::make_unique<net::test_server::BasicHttpResponse>();
-    response->set_code(net::HTTP_OK);
-    response->set_content_type("text/html");
-    if (request.relative_url == "/with-tools") {
-      response->set_content(kPageWithTools);
-    } else if (request.relative_url == "/no-tools") {
-      response->set_content(kPageWithoutTools);
-    } else {
-      return nullptr;
-    }
-    return response;
-  }
-
   base::test::ScopedFeatureList scoped_feature_list_;
   content::ContentMockCertVerifier mock_cert_verifier_;
   net::EmbeddedTestServer https_server_;
@@ -168,7 +125,7 @@ class WebMcpBrowserTest : public InProcessBrowserTest {
 IN_PROC_BROWSER_TEST_F(WebMcpBrowserTest,
                        AssociatedWebContentsContent_DirectFetch) {
   ASSERT_TRUE(ui_test_utils::NavigateToURL(
-      browser(), https_server()->GetURL("a.com", "/with-tools")));
+      browser(), https_server()->GetURL("a.com", kPageWithToolsPath)));
   ASSERT_EQ(
       2, content::EvalJs(GetActiveWebContents(),
                          "navigator.modelContextTesting.listTools().length"));
@@ -184,7 +141,7 @@ IN_PROC_BROWSER_TEST_F(WebMcpBrowserTest,
 IN_PROC_BROWSER_TEST_F(WebMcpBrowserTest,
                        AssociatedContentManager_SeesRegisteredTools) {
   ASSERT_TRUE(ui_test_utils::NavigateToURL(
-      browser(), https_server()->GetURL("a.com", "/with-tools")));
+      browser(), https_server()->GetURL("a.com", kPageWithToolsPath)));
 
   // Make sure the WebMCP runtime feature is actually present in the renderer
   // and the page-side registration didn't throw; without this the rest of the
@@ -214,22 +171,26 @@ IN_PROC_BROWSER_TEST_F(WebMcpBrowserTest,
   auto tools = RefreshAndGetTools(manager);
   ASSERT_EQ(2u, tools.size());
 
-  // Names are prefixed with the sanitized host ("a.com" → "a_com") to
-  // disambiguate page-defined tools across origins.
+  // Names are prefixed with the sanitized host and path
+  // ("a.com/web_mcp_tools.html" → "a_com_web_mcp_tools_html") to disambiguate
+  // page-defined tools across origins and pages.
   std::set<std::string> tool_names;
   for (const auto& tool : tools) {
     ASSERT_TRUE(tool);
     tool_names.insert(std::string(tool->Name()));
   }
   EXPECT_THAT(tool_names,
-              ::testing::UnorderedElementsAre("a_com_echo", "a_com_ping"));
+              ::testing::UnorderedElementsAre("a_com_web_mcp_tools_html_echo",
+                                              "a_com_web_mcp_tools_html_ping"));
 
   // Sanity check on metadata for the richer tool.
   for (const auto& tool : tools) {
-    if (tool->Name() == "a_com_echo") {
-      // The description embeds the host and the page-provided description.
+    if (tool->Name() == "a_com_web_mcp_tools_html_echo") {
+      // The description embeds the full page URL and the page-provided
+      // description.
       std::string description(tool->Description());
       EXPECT_NE(description.find("a.com"), std::string::npos);
+      EXPECT_NE(description.find(kPageWithToolsPath), std::string::npos);
       EXPECT_NE(description.find("Echo input back"), std::string::npos);
       ASSERT_TRUE(tool->InputProperties().has_value());
       EXPECT_TRUE(tool->InputProperties()->contains("text"));
@@ -243,7 +204,7 @@ IN_PROC_BROWSER_TEST_F(WebMcpBrowserTest,
 IN_PROC_BROWSER_TEST_F(WebMcpBrowserTest,
                        AssociatedContentManager_NoToolsWhenPageHasNone) {
   ASSERT_TRUE(ui_test_utils::NavigateToURL(
-      browser(), https_server()->GetURL("a.com", "/no-tools")));
+      browser(), https_server()->GetURL("a.com", kPageWithoutToolsPath)));
 
   auto* content = GetActiveContent();
   ASSERT_TRUE(content);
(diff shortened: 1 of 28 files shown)


browser/ai_chat/tools/script_tool.cc
```diff
@@ -0,0 +1,116 @@
...
+void ScriptTool::UseTool(const std::string& input_json,
+                         UseToolCallback callback) {
+  content::RenderFrameHost* rfh = rfh_.AsRenderFrameHostIfValid();
+  if (!rfh) {
+    std::move(callback).Run({}, {});
+    return;
+  }
+
+  mojo::AssociatedRemote<chrome::mojom::ChromeRenderFrame> render_frame;
+  rfh->GetRemoteAssociatedInterfaces()->GetInterface(&render_frame);
```
fallaciousreasoning: I was wondering I should introduce our own interface here which would let us communicate with the renderer from the `//brave/components/ai_chat/content` layer, in which case we could get rid of the `ScriptToolFactory`

browser/ai_chat/tools/script_tool.cc
```diff
@@ -0,0 +1,116 @@
...
+
+  auto invocation = actor::mojom::ToolInvocation::New();
+  // Use a sentinel task ID since this invocation is outside any actor task.
+  invocation->task_id = actor::TaskId::FromUnsafeValue(1);
```
fallaciousreasoning: cc @petemill I'm unsure about this. It feels a little wrong that I'm using Chromium's actor framework outside of agentic mode. This would also be fixed by using our own mojom interface.

browser/brave_tab_helpers.cc
```diff
@@ -185,11 +186,16 @@ void AttachTabHelpers(content::WebContents* web_contents) {
...
+                })),
 #else
-        nullptr
+        nullptr,
 #endif
-    );
+        base::BindRepeating(
+            [](const blink::mojom::ScriptTool& script_tool,
+               content::WeakDocumentPtr rfh) -> std::unique_ptr<ai_chat::Tool> {
+              return std::make_unique<ai_chat::ScriptTool>(script_tool,
```
fallaciousreasoning: Its unfortunate that executing a script tool depends on `//chrome` so this can't live inside `associated_web_contents_content` where it (arguably) belongs

components/ai_chat/core/browser/associated_content_manager.h
```diff
@@ -110,6 +117,10 @@ class AssociatedContentManager : public AssociatedContentDelegate::Observer {
   // The number of content delegates.
   size_t GetContentDelegateCount() const;
 
+  // ToolProvider:
+  void OnNewGenerationLoop() override;
```
fallaciousreasoning: Hey @petemill, I was thinking of making this take a callback, so `ToolProvider`s can do async work here and block generation until it completes (so that we can collect ScriptTools from all associated content). WDYT?

components/ai_chat/core/browser/conversation_handler.cc
```diff
@@ -151,6 +151,7 @@ ConversationHandler::ConversationHandler(
   BuildCapabilitiesSet();
 
   // Observe tool providers
+  associated_content_manager_->AddObserver(this);
```
fallaciousreasoning: for now, I haven't added this to the `ToolProvider`s list, but I' like to before it merges. The main thing blocking this is not having an async version of `OnNewGenerationLoop`

components/ai_chat/core/browser/conversation_handler.cc
```diff
@@ -616,9 +618,15 @@ void ConversationHandler::SubmitHumanConversationEntry(
     engine_->SanitizeInput(*latest_turn->selected_text);
   }
 
-  // If it's a suggested question, replace with the prompt.
-  std::string prompt = latest_turn->prompt.value_or(latest_turn->text);
+  // Collect current script tools from attached content before generating, then
+  // continue submission once tools are available.
+  associated_content_manager_->GetScriptTools(
```
fallaciousreasoning: **Note:** with an async `OnNewGenerationLoop` this could be removed (and the corresponding `OnScriptToolsCollected`).

components/ai_chat/core/browser/associated_content_manager.cc
```diff
@@ -345,6 +345,30 @@ void AssociatedContentManager::GetContent(base::OnceClosure callback) {
   }
 }
 
+void AssociatedContentManager::GetScriptTools(base::OnceClosure callback) {
+  tools_.clear();
```
fallaciousreasoning: Note to self - this isn't reentrant at all, and we'll get into a very weird state if you call it twice without waiting for it to finish.

browser/ai_chat/tools/script_tool.cc
petemill: I think one issue is that the actor framework isn't built on android (last time I checked) and I don't see why this feature wouldn't be available on Android. Are any of the other MCP interfaces or classes gated on android?

components/ai_chat/core/browser/associated_content_manager.h
petemill: Seems sensible. Should it be more intentional like `GetToolForNewGenerationLoop()` or is this ok?

browser/ai_chat/tools/script_tool.cc
fallaciousreasoning: added our own function for ExecuteTool so this should work on Android

components/ai_chat/core/browser/associated_content_manager.h
fallaciousreasoning: https://github.com/brave/brave-core/pull/36156

components/ai_chat/core/browser/conversation_handler.cc
fallaciousreasoning: https://github.com/brave/brave-core/pull/36156

components/ai_chat/core/common/mojom/page_content_extractor.mojom
```diff
@@ -44,4 +44,8 @@ interface PageContentExtractor {
   // continue-with-leo HTML anchor element (open Leo button). These two
   // nonces value must be the same, otherwise nullopt will be returned.
   GetOpenAIChatButtonNonce() => (string? nonce);
+
+  // Execute a script tool registered on the page by name. Returns the tool
+  // result string, or null on failure.
+  ExecuteScriptTool(string name, string input_json) => (string? result);
```
fallaciousreasoning: Dunno if this is the right place for this - maybe if I rename it (separately) to something else...

fallaciousreasoning: cc @petemill

netzenbot (approved):

components/ai_chat/core/browser/associated_content_delegate.h
```diff
@@ -86,6 +87,12 @@ class AssociatedContentDelegate {
   virtual void GetScreenshots(
       mojom::ConversationHandler::GetScreenshotsCallback callback);
 
+  // Fetches the current set of script tools from the content. The default
+  // implementation calls |callback| with an empty list immediately.
+  using GetScriptToolsCallback =
+      base::OnceCallback<void(std::vector<std::unique_ptr<Tool>>)>;
+  virtual void GetScriptTools(GetScriptToolsCallback callback);
+
```
petemill: I like the concept of Content being able to expose Tools, but I'm not sure they need to be explicitly called _Script_ Tools here, which is more of an implementation detail? Should we just call it `GetTools` or `GetContentTools`?

components/ai_chat/core/browser/associated_content_manager.cc
```diff
@@ -477,6 +482,41 @@ void AssociatedContentManager::OnTitleChanged(
...
 
+void AssociatedContentManager::UpdateToolsForNewGenerationLoop(
+    base::OnceClosure on_updated) {
+  tools_.clear();
+  if (content_delegates_.empty()) {
+    std::move(on_updated).Run();
+    return;
+  }
+  auto barrier =
+      base::BarrierClosure(content_delegates_.size(), std::move(on_updated));
```
petemill, line 493: I believe `content_delegates_` is _all_ content attached to all messages, whereas as discussed just now the UX we're aiming for is that:
- both Content and Tools for a given AssociatedContent get attachment UI elements, either one can be unattached independently
- we should only be attaching tools for Tools of AssociatedContent that is currently staged
- we re-stage Tools for any AssociatedContent that haven't been unattached already on the next message

I'm fine if we address that altogether in a follow-up since it might make sense to land that data model change separately.

Changing this now to only consider staged content would mean we can't experiment with this feature for follow-ups because we cannot stage content that has already been attached to a message.

Since this feature is behind a default-disabled flag, I'm fine landing like this but let's check in with the team.

components/ai_chat/core/browser/conversation_handler.cc
```diff
@@ -1180,6 +1180,8 @@ void ConversationHandler::ProcessPermissionChallenge(
   // Find the tool and notify it
   base::WeakPtr<Tool> tool_ptr;
   for (auto& tool : GetTools()) {
+    LOG(ERROR) << "Tool name: " << tool->Name()
+               << " tool use name: " << tool_use->tool_name;
```
petemill: needs removing

components/ai_chat/core/browser/associated_content_manager.cc
petemill: We also need to consider how to select for edited submissions.

components/ai_chat/renderer/page_content_extractor.cc
```diff
@@ -340,6 +342,32 @@ void PageContentExtractor::OnJSTranscriptUrlResult(
...
+             std::unique_ptr<blink::WebScriptToolDeclaration> declaration,
+             base::expected<blink::WebString, blink::WebScriptToolError>
+                 result) {
+            if (result.has_value()) {
+              std::move(cb).Run(result->Utf8());
+            } else {
+              std::move(cb).Run(std::nullopt);
+            }
+          },
+          std::move(wrapped)));
```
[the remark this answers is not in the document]
petemill: /fp document.ExecuteScriptTool is different than content scripts. Still something to review, but this is to do with AI WebMCP tools, and not content scripts.

components/ai_chat/core/browser/associated_content_delegate.h
fallaciousreasoning: Renaming to ContentTools

components/ai_chat/core/browser/associated_content_manager.cc
fallaciousreasoning: I guess if the content is still live, it will work. If its not, it won't?

components/ai_chat/content/browser/script_tool.cc
```diff
@@ -0,0 +1,97 @@
...
+  auto* extractor_ptr = extractor.get();
+  extractor_ptr->ExecuteScriptTool(
+      name_, input_json,
+      base::BindOnce(
+          [](UseToolCallback cb, mojo::Remote<mojom::PageContentExtractor>,
+             const std::optional<std::string>& result) {
+            std::move(cb).Run(CreateContentBlocksForText(result.value_or("")),
+                              {});
+          },
+          std::move(callback), std::move(extractor)));
```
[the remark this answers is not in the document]
fallaciousreasoning: /fp this is a different ExecuteScriptTOol

fallaciousreasoning:
@petemill this will ask for permission every time you try to execute a script tool (even the same tool multiple times in a row). Does that sound okay for an initial version?

It'd be good to get it in so we can start iterating.

diracdeltas:
adding @stoletheminerals to sign off / verify the security hotspots

browser/ai_chat/web_mcp_browsertest.cc
```diff
@@ -0,0 +1,282 @@
...
+#include "third_party/blink/public/common/features.h"
+
+namespace ai_chat {
+
+namespace {
+
+// HTML that registers two tools via navigator.modelContext. The feature is
+// only available on secure contexts when blink::features::kWebMCPTesting is
+// enabled (which implies kWebMCP).
+constexpr char kPageWithTools[] = R"HTML(
```
petemill: Would it be nicer to have these response constants be .html files at `test/data/leo` (should be `ai_chat` but is `leo` already), along with `https_server_.ServeFilesFromDirectory(test_data_dir)`?

components/ai_chat/content/browser/content_tool.cc
```diff
@@ -0,0 +1,133 @@
...
+
+namespace ai_chat {
+
+ContentTool::ContentTool(const blink::mojom::ScriptTool& script_tool,
+                         content::WeakDocumentPtr rfh)
+    : rfh_(std::move(rfh)), internal_tool_name_(script_tool.name) {
+  // Name of the ContentTool is {origin}_tool_name.
+  name_ = base::StrCat(
+      {rfh_.AsRenderFrameHostIfValid()->GetLastCommittedURL().host(), "_",
+       script_tool.name});
```
petemill: I'm thinking we should do the full URL and path here or we may end up with duplicates? At least until we have a smarter selection strategy for only including tools from pages with a unique "authentication context" or similar. Any other ideas to avoid duplicates?

components/ai_chat/content/browser/content_tool.cc
```diff
@@ -0,0 +1,133 @@
...
+       "page."});
+
+  if (!script_tool.input_schema) {
+    return;
+  }
+  auto schema = base::JSONReader::ReadDict(
+      *script_tool.input_schema, base::JSON_PARSE_CHROMIUM_EXTENSIONS);
+  if (!schema) {
+    return;
+  }
```
petemill, line 61: Maybe we need a plan to avoid sending bad tools to the LLM. For this constructor to identify that it's bad and shouldn't be used. Is there any validation in the blink layer?

components/ai_chat/core/browser/associated_content_manager.cc
petemill: Yeah but we still need a way for the user to know that tools will be attached to the next message. Especially since there is no UI difference between live and archived content. And a way for them to detach the tools for content that is attached to previous messages. I know this is obvious now that we've discussed it, so I'm just writing it down.

Let's make sure we create an issue or add to a tracking issue.

components/ai_chat/core/browser/conversation_handler.h
```diff
@@ -493,6 +491,12 @@ class ConversationHandler : public mojom::ConversationHandler,
   // (e.g. a KeyedService, or global singleton).
   std::vector<std::unique_ptr<ToolProvider>> tool_providers_;
 
+  // Pointer into one of the entries in `tool_providers_` above. Declared after
+  // `tool_providers_` so that it is destroyed first — otherwise the unique_ptr
+  // owning the AssociatedContentManager is freed before this raw_ptr,
+  // triggering BackupRefPtr's dangling-pointer check.
+  raw_ptr<AssociatedContentManager> associated_content_manager_;
```
petemill, line 514: It's a bit awkward that we have a tool provider that is also a special class that we want to be treated differently, because if we do something special with the tool provider collection in the future, e.g. clear it, then we'll have side effects for attached content. But I can't immediately think of anything better, aside a GetToolProviders internal method which combines `tool_providers_` with `associated_content_manager_`. Maybe you had other ideas, or don't feel like this is a big concern?

components/ai_chat/core/browser/associated_content_manager.cc
fallaciousreasoning: https://github.com/brave/brave-browser/issues/56082

components/ai_chat/content/browser/content_tool.cc
fallaciousreasoning: There is a bit yeah - has to meet some criteria and have a description. I don't know if this check is strictly needed but lets keep it for now

components/ai_chat/core/browser/conversation_handler.h
fallaciousreasoning: Yeah, I wasn't super thrilled either - my other idea was a separate array which just holds a pointer to the tools. But lets cross that bridge when we get to it. Should be pretty obvious something has gone wrong when the associated content manager gets deleted :laughing:

petemill (approved):

stoletheminerals (approved):
Approving this PR specifically, but the security/privacy issue will stay open while this feature is not fully implemented and while it's off by default and behind the flag. The feature needs UI changes, which are out of scope for this PR. Please don't enable it by default until the security review is closed

components/ai_chat/content/browser/associated_web_contents_content.cc
```diff
@@ -337,6 +338,48 @@ void AssociatedWebContentsContent::GetOpenAIChatButtonNonce(
...
+  if (!rfh || !rfh->IsRenderFrameLive()) {
+    std::move(callback).Run({});
+    return;
+  }
+  // Use a local Remote rather than the `ai_page_content_agent_` member: the
+  // member is also used by FetchPageContentFromAIPageContentAgent (the
+  // GetContent fallback path). If GetContent's fallback was in flight when
+  // GetContentTools is called, the shared remote gets reset and the in-flight
+  // response is dropped — and vice versa. Owning the remote in the bound
+  // callback keeps each call independent.
```
goodov: this all look like a questionable workaround instead of persisting the remote as a member of a class. Basically this is how you get memory leaks even with smart pointers.

```
If GetContent's fallback was in flight when
  // GetContentTools is called, the shared remote gets reset and the in-flight
  // response is dropped — and vice versa.
```

why the existing remote is reset in the first place? I looked into the existing code and it feels like llm tried to do something smart without a real reason, but maybe I miss something?

can you look at `ai_page_content_agent_` closer and maybe just make it right?

fallaciousreasoning: Good point, I was definitely a bit lazy here - I just assumed there was some setup on the other side so we could just make one call. Looking at the existing code, I'd say its because the remote is bound to a `RenderFrame` while this class is tied to a `WebContents`, and its meant to ensure that the remote is never pointing to a different `RenderFrame`.

I've updated it to reset the remote `OnNewPage` which should have a similar result.

cc @darkdh to confirm I'm not misunderstanding what the reset was for.

darkdh: yes, the current `ai_page_content_agent_` reset before calling `GetAIPageContent` introduced in https://github.com/brave/brave-core/pull/35435/ is to make sure it doesn't get outdated one which bound to previous `RenderFrame` which got navigated away.

In hindsight, we don't have to keep the remote as member, we can do it like what `components/optimization_guide/content/browser/page_content_proto_provider.cc` does in [GetAIPageContent()](https://source.chromium.org/chromium/chromium/src/+/main:components/optimization_guide/content/browser/page_content_proto_provider.cc;l=656-667;drc=7b56828f51975088b9ad114b20dad1865bbdcd810) which moves the remote along with the callback

goodov: ok, I don't like this idea of shoving the mojo remote in a callback, but it seems like an acceptable approach, although it needs to be wrapped with `WrapCallbackWithDefaultInvokeIfNotRun` to make it reliable.

lgtm to use it in both places for now.

A better approach is to look into `DocumentUserData` to make the remote be bound to the lifetime of a document. This should solve both problems - persistent **single** remote and auto-invalidation on document change.

fallaciousreasoning: Awesome, converted both usages over to my original approach :smile:

fallaciousreasoning:
cc @goodov should be ready for another look :smile:

goodov (approved):

PR #36083, opened 2026-05-01
[Cleanup] Remove unnecessay IPH suppress in runtime

simonhong:
Resolves - simple cleanup issue.

We can set disable `kIPHSideBySidePinnableFeature/kIPHSideBySideTabSwitchFeature` as default state.
So, don't need to suppress them in rumtime.
All deleted and add state overriding to `chromium_src/components/feature_engagement/public/feature_constants.cc`

It was added by https://github.com/brave/brave-core/pull/32937


mkarolin (approved):
++

PR #36086, opened 2026-05-01
Playlist: Remove the ability to add homepage and search result page as playlist items

### LINKED ISSUE #55010: Playlist: Capture allowed from non‑playable pages produces permanently broken items

EliHini:
## Summary
Updates `PlaylistScriptHandler` so playlist items store a `pageSrc` that `LivePlaylistWebLoader` can reload, and avoids offering captures that would persist as broken entries on YouTube listing-style URLs.

## Problem
Adding to Playlist from contexts where **`pageSrc` was wrong or not reloadable** (always forcing the visible tab URL when the script already reported the correct page, or allowing captures from YouTube feeds/search/home where there is no stable video page to resolve later) produced **playlist items that could not be played back reliably**.

## Testplan

| # | Steps | Expected result |
|---|---------------|-----------------|
| 1 | Visit YouTube home page that contains feed  | The top video in view should begin playing. **Note** that there is no add playlist button in the address bar. |
| 2 | Select any video to go to its page. | The add playlist button is visible and enabled. |
| 3 | Add the video to the playlist. If onboarding hasn't happened yet, follow the instructions to add the playlist. | The video is added to the playlist and is playable. The thumbnail is filled in with a frame from the video |

Resolves  https://github.com/brave/brave-browser/issues/55010

### DIFF
--- a/ios/brave-ios/Sources/Brave/Frontend/UserContent/UserScripts/Scripts_Dynamic/ScriptHandlers/Paged/PlaylistScriptHandler.swift
+++ b/ios/brave-ios/Sources/Brave/Frontend/UserContent/UserScripts/Scripts_Dynamic/ScriptHandlers/Paged/PlaylistScriptHandler.swift
@@ -130,11 +130,24 @@ class PlaylistScriptHandler: NSObject, TabContentScript, TabObserver {
       return
     }
 
-    // Copy the item but use the web-view's title and location instead, if available
-    // This is due to a iFrames security
-    item.pageSrc = tab.visibleURL?.absoluteString ?? item.pageSrc
+    // The script reports `window.top.location.href` for same-origin frames (which already equals `tab.visibleURL`)
+    // and falls back to the iframe's own URL when blocked by SOP. We only want to override in that cross-origin case so
+    // the stored `pageSrc` matches the page the user thinks they added the item from, instead of an opaque embedder URL.
+    if let visibleURL = tab.visibleURL,
+      let pageURL = URL(string: item.pageSrc),
+      pageURL.host != visibleURL.host
+    {
+      item.pageSrc = visibleURL.absoluteString
+    }
     item.pageTitle = tab.title ?? item.pageTitle
     item.lastPlayedOffset = 0.0
+
+    guard Self.canResolvePageSrcLater(item.pageSrc) else {
+      DispatchQueue.main.async {
+        handler.delegate?.updatePlaylistURLBar(tab: tab, state: .none, item: nil)
+      }
+      return
+    }
 
     Self.queue.async { [weak handler] in
       guard let handler = handler else { return }
@@ -165,6 +178,31 @@ class PlaylistScriptHandler: NSObject, TabContentScript, TabObserver {
           }
         }
       }
+    }
+  }
+
+  /// Returns `true` when `pageSrc` is the kind of URL we can reasonably re-resolve later by re-loading it in a `LivePlaylistWebLoader`.
+  /// For known sites whose feed/home/search pages have no stable `<video>` element to extract from, returns `false` so we don't capture
+  /// items that are guaranteed to fail resolution. Unknown sites fall through and are trusted; we only special-case sites we have evidence for.
+  private static func canResolvePageSrcLater(_ pageSrc: String) -> Bool {
+    guard let url = URL(string: pageSrc), let baseDomain = url.baseDomain else {
+      return true
+    }
+    let path = url.path
+    switch baseDomain {
+    case "youtube.com":
+      // Only watch/shorts/embed pages have a stable video at a stable URL. Homepage,
+      // search results (`/results`), feed (`/feed/*`), channel (`/@handle`, `/c/*`,
+      // `/channel/*`, `/user/*`) and similar listings do not.
+      return path.hasPrefix("/watch")
+        || path.hasPrefix("/shorts/")
+        || path.hasPrefix("/embed/")
+        || path.hasPrefix("/v/")
+    case "youtu.be":
+      // Short links of the form `youtu.be/<videoId>`.
+      return url.pathComponents.count >= 2 && !url.pathComponents[1].isEmpty
+    default:
+      return true
     }
   }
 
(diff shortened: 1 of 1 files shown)


netzenbot (approved):

ios/brave-ios/Sources/Brave/Frontend/UserContent/UserScripts/Scripts_Dynamic/ScriptHandlers/Paged/PlaylistScriptHandler.swift
```diff
@@ -168,6 +181,31 @@ class PlaylistScriptHandler: NSObject, TabContentScript, TabObserver {
     }
   }
 
+  /// Returns `true` when `pageSrc` is the kind of URL we can reasonably re-resolve later by re-loading it in a `LivePlaylistWebLoader`.
+  /// For known sites whose feed/home/search pages have no stable `<video>` element to extract from, returns `false` so we don't capture
```
StephenHeaps: What does `stable` mean here (I'm a bit unfamiliar with the playlist setup). Don't we only show the playlist button when there is a `video` tag? Does the tag get removed when it scrolls off screen, causing the failed to add issue?

StephenHeaps (commented):
I think the main concern with this is that we are hardcoding the paths of YouTube pages we support, if these ever changed then users need an app update to support Playlist with YouTube?

ios/brave-ios/Sources/Brave/Frontend/UserContent/UserScripts/Scripts_Dynamic/ScriptHandlers/Paged/PlaylistScriptHandler.swift
EliHini: That's a great question. Stable here refers to `pageSrc` after a full reload, not about whether a `<video>` exists in the current viewport. When the direct src can’t be streamed anymore, we fall back to loading `item.pageSrc` in a `LivePlaylistWebLoader` and rely on the same playlist user script to find media again and produce a new playable URL (see `PlaylistMediaStreamer.streamingFallback`, which loads `URL(string: item.pageSrc)`).

This means that we need `pageSrc` to be a URL where, on a cold navigation, the page still represents one primary piece of media we can rediscover. 

On YouTube home / search / feed / channel listings, that isn’t true in the same way: reload doesn’t give you “the same video page,” so re-resolution is unreliable. Watch / shorts / embed / v/ URLs are the ones where reload leads to the same watch experience where we can find a `<video>` again.

ios/brave-ios/Sources/Brave/Frontend/UserContent/UserScripts/Scripts_Dynamic/ScriptHandlers/Paged/PlaylistScriptHandler.swift
```diff
@@ -168,6 +181,31 @@ class PlaylistScriptHandler: NSObject, TabContentScript, TabObserver {
...
+      // Only watch/shorts/embed pages have a stable video at a stable URL. Homepage,
+      // search results (`/results`), feed (`/feed/*`), channel (`/@handle`, `/c/*`,
+      // `/channel/*`, `/user/*`) and similar listings do not.
+      return path.hasPrefix("/watch")
+        || path.hasPrefix("/shorts/")
+        || path.hasPrefix("/embed/")
+        || path.hasPrefix("/v/")
+    case "youtu.be":
+      // Short links of the form `youtu.be/<videoId>`.
+      return url.pathComponents.count >= 2 && !url.pathComponents[1].isEmpty
```
StephenHeaps: nit:
```suggestion
      return url.pathComponents[safe: 1]?.isEmpty == false
```

EliHini:
> I think the main concern with this is that we are hardcoding the paths of YouTube pages we support, if these ever changed then users need an app update to support Playlist with YouTube?

That's a fair concern. The context around my approach is that YouTube's watch-style paths have been stable for a long time. So I focused the fix on a narrow YouTube-specific allowlist for pageSrc so we don’t offer add-to-playlist when re-loading that URL later is unlikely to reproduce a single watchable item; thereby guarding a failure mode in the page reload fallback. There are other path for YouTube in Brave.

Alternatively, we could follow up with remote config for path patterns if we want to adjust without a store release, or loosen the heuristic (e.g. default allow and only block known-bad prefixes like /results, /feed) if we prefer fewer false negatives at the cost of more bad pageSrcs until we tighten again.

ios/brave-ios/Sources/Brave/Frontend/UserContent/UserScripts/Scripts_Dynamic/ScriptHandlers/Paged/PlaylistScriptHandler.swift
EliHini: Good call — switched to pathComponents[safe: 1]?.isEmpty == false so we don’t index past the end and it matches our usual Array pattern.

StephenHeaps:
> > I think the main concern with this is that we are hardcoding the paths of YouTube pages we support, if these ever changed then users need an app update to support Playlist with YouTube?
> 
> That's a fair concern. The context around my approach is that YouTube's watch-style paths have been stable for a long time. So I focused the fix on a narrow YouTube-specific allowlist for pageSrc so we don’t offer add-to-playlist when re-loading that URL later is unlikely to reproduce a single watchable item; thereby guarding a failure mode in the page reload fallback. There are other path for YouTube in Brave.
> 
> Alternatively, we could follow up with remote config for path patterns if we want to adjust without a store release, or loosen the heuristic (e.g. default allow and only block known-bad prefixes like /results, /feed) if we prefer fewer false negatives at the cost of more bad pageSrcs until we tighten again.

Would it not be better to exclude the results & feed paths? The PR title sort of implies that, but instead hardcodes support to specific paths. Assuming those are the two main pages we're trying to fix the add to playlist button from showing on

StephenHeaps (approved):

nuo-xu (approved):

EliHini:
Closing this PR for now. Implementing a component based system to dynamically get the deniedlist

PR #36087, opened 2026-05-01
Unify YouTube Picture-in-Picture flow into a dedicated controller

### LINKED ISSUE #56382: Skip fullscreen JS injection for unsupported Android versions
Brave's YouTube Picture in Picture flow currently prepares the page for PiP even on Android versions where Brave will not attempt PiP entry. Android Picture in Picture is intentionally disabled for Android versions older than R, because Chromium and Brave avoid a platform crash when entering PiP immediately after leaving it. The Java entry path already checks `PictureInPicture.isEnabled(activity)`, but the native YouTube script injector should mirror that gate before enabling YouTube PiP support or attempting to enter PiP after fullscreen is reported.

## Problem

* The page can receive the YouTube PiP flag override even though the device cannot enter PiP.
* The native fullscreen callback can continue toward the PiP entry path on a version where the Java layer will reject the request.
* The unsupported platform path creates avoidable page mutation, unnecessary work, and a mismatch between the toolbar availability check and the native script injector.

## Expected behavior

* On Android R and newer, Brave keeps the existing YouTube PiP setup and entry flow.
* On Android versions older than R, Brave does not inject the YouTube PiP flag override and does not call into the PiP entry path from the fullscreen callback.
* The native gate should match the Java `PictureInPicture.isEnabled` gate used by the toolbar and entry helper.

### LINKED ISSUE #56383: Improve fullscreen script to inject when PIP is triggered
The YouTube PiP toolbar flow depends on a small script that puts the mobile YouTube player into fullscreen before Brave asks Android to enter PiP. That step is necessary because Android PiP captures the fullscreen video surface. The existing script is too dependent on one YouTube DOM shape and on timing events that are not stable during cold loads, slow networks, and YouTube player hydration.

When the user taps the toolbar PiP button, Brave should make the fullscreen request promptly if the player is ready, but it also needs to keep watching for the controls when YouTube has not rendered them yet. The improved script should use a deterministic sequence of attempts, resolve only once, avoid toggling fullscreen off when it is already active, and avoid racing two fullscreen mechanisms against each other.

## Current issues

* It relies on a narrow set of selectors, so a YouTube player variant can fail even when the video and fullscreen control are present.
* It waits on `canplay` in a way that ties fullscreen entry to video decode readiness instead of the availability of the player control.
* It can miss the moment when YouTube inserts or reveals the fullscreen button after the script already ran.
* It may resolve too early with missing elements during a slow player load, leaving the toolbar PiP request with no fullscreen transition.
* It does not have a single shared completion guard, so multiple asynchronous paths can compete after the user has already moved on.
* It does not first check both document fullscreen and YouTube player fullscreen state, so fallback logic can accidentally call a toggle API from the wrong state.
* It does not have a broad enough fallback strategy for YouTube player variants where the visible button is delayed or unavailable.

## Proposed change

* Replace the old entry script with a staged fullscreen algorithm inside `kYoutubeFullscreen`.
* Start immediately after `DOMContentLoaded` when needed, or immediately if the document is already ready.
* First detect whether the document or YouTube player is already fullscreen and return without toggling it back out.
* Try the fast path by clicking the YouTube fullscreen control when the video and control are already present.
* Tap the player once to reveal controls, then retry the fast path.
* Fall back to YouTube's own `toggleFullscreen` API only when the player is not already fullscreen.
* Fall back to the standard `requestFullscreen` API on the player, player container, or video element when no YouTube control path is available.
* Use a `MutationObserver` on the player subtree, or the document body as a last resort, so the script reacts as soon as YouTube inserts or reveals the fullscreen control.
* Keep a 30 second timeout so the observer cannot live forever on a broken or unexpected page.
* Resolve through a single `resolveOnce` function so late callbacks cannot report conflicting results.
* Keep the script in `ISOLATED_WORLD_ID_BRAVE_INTERNAL` and execute it with user activation, matching the existing Brave controlled injection model.

## Expected result

* Tapping the toolbar PiP button should enter fullscreen quickly when YouTube is already ready.
* Cold cache loads and slow networks should still work once the player controls appear.
* YouTube player UI changes should be less likely to break PiP entry.
* The flow should avoid entering PiP with no fullscreen video content.
* Already fullscreen pages should not be toggled back out by the fallback path.
* The entry script should be a clearer, more maintainable sequence with explicit state checks and bounded async work.

### LINKED ISSUE #56384: Reject PIP for inner WebContents
Brave should only expose YouTube PiP for the outer tab WebContents. The YouTube PiP controller and script injector assume they are operating on the tab level primary main frame. If the same availability check succeeds for an inner WebContents, such as an embedded guest view, the toolbar or entry path could target the wrong frame tree and bypass the tab lifecycle that tracks fullscreen state, navigation, visibility, and PiP session ownership.

## Problem

* Inner WebContents do not represent the top level tab that owns the browser UI and Android PiP session.
* Running the availability check on an inner WebContents can report a YouTube video even though the tab level controller cannot safely manage that embedded page.
* The fullscreen request, script injector binding, and PiP lifecycle cleanup are designed around the outer tab WebContents.
* Allowing inner WebContents creates confusing ownership between the embedded content and the browser activity.

## Expected behavior

* `IsPictureInPictureAvailable` should return false when the WebContents has an outer WebContents.
* Brave should only show or trigger YouTube PiP from the outermost tab WebContents.
* The controller should only manage fullscreen and PiP state for a tab it can cleanly observe across navigation and lifecycle events.

## Proposed change

* Add an explicit `web_contents()` null guard before evaluating availability.
* Reject any WebContents where `GetOuterWebContents()` is not null.
* Keep the existing YouTube watch URL, feature flag, and page load checks for the valid tab WebContents path.

### LINKED ISSUE #56386: Prevent race condition during fullscreen injection
There is a race in the YouTube Picture in Picture toolbar flow when fullscreen script injection happens after navigation or frame changes.

`YouTubeScriptInjectorTabHelper` keeps a `script_injector_remote_` used to send the fullscreen script to the renderer. That remote is associated with a specific `RenderFrameHost`. If the tab navigates after the remote has been bound, a later fullscreen request should not assume that the previous remote is still valid for the current page.

## Problem

* `script_injector_remote_` is bound to a specific `RenderFrameHost`.
* The helper can keep that remote across navigation or RFH changes.
* A later fullscreen request may rely on stale renderer binding state.
* The fullscreen request should always target the current primary main frame.
* Stale fullscreen setup state should not affect a later committed page.

## Expected behavior

* `MaybeSetFullscreen()` should use the current primary main frame when requesting fullscreen script execution.
* The script injector remote should be rebound before sending a fullscreen request.
* The cached script injector remote should be cleared when the RFH it was bound to is replaced.
* Fullscreen requested state should continue to reset on page and frame changes.
* A fullscreen request after navigation should operate on the new current page, not on stale renderer state from the previous page.

## Proposed change

* Reset any previous `script_injector_remote_` before binding for a fullscreen request.
* Bind `script_injector_remote_` to the current primary main frame in `MaybeSetFullscreen()`.
* Track the RFH the remote is bound to.
* Reset the cached remote from `RenderFrameHostChanged()` when the bound RFH is replaced.
* Add browser test coverage for navigation after a pending fullscreen request, then verify fullscreen can still be triggered on the new page.

simoarpe:
## Summary

This PR hardens the Brave handled YouTube Picture in Picture entry flow on Android. The main change is a rewrite of the fullscreen script that runs when the toolbar PiP button is pressed. The new script is more tolerant of YouTube player hydration, slow network conditions, and player DOM variants, while keeping the request bounded and tied to the current YouTube tab.

The PR also brings the native script injector in line with the Java PiP availability gate, rejects inner WebContents, and closes a navigation race where a pending fullscreen request could be delivered after the tab had already left YouTube.

## What changed

* Reworked `kYoutubeFullscreen` into a staged algorithm that first checks whether the document or YouTube player is already fullscreen.
* Added a fast path for the normal case where the video and fullscreen control are already available.
* Added a player tap and retry path so controls that render only after interaction can still be used.
* Added fallback paths through YouTube's player fullscreen API and the standard `requestFullscreen` API.
* Replaced fragile readiness timing with a `MutationObserver` that reacts when YouTube inserts or reveals the fullscreen control.
* Added a single `resolveOnce` completion path and a 30 second timeout so asynchronous work cannot keep running indefinitely.
* Expanded player and container detection so Brave is less dependent on one mobile YouTube DOM shape.
* Added native Android SDK gating so PiP support script injection and the final PiP entry call are skipped on Android versions older than R.
* Added a committed YouTube domain check inside `MaybeSetFullscreen()` so the script is not injected after the tab navigates away from YouTube.
* Rejected inner WebContents in `IsPictureInPictureAvailable()` so PiP remains owned by the outer tab WebContents and its lifecycle controller.

## Why

The previous script worked for the common loaded player state, but it was fragile when YouTube had not yet hydrated its controls, when video readiness lagged behind UI readiness, when the player used a different fullscreen control, or when the tab navigated between the toolbar check and renderer execution. Those failures could leave the user tapping PiP with no Android PiP entry, could attempt work on unsupported Android versions, or could target a WebContents that the tab level PiP flow cannot own safely.

## User impact

* YouTube PiP from the toolbar should start more reliably on slow loads and cold player states.
* Android versions where Brave cannot safely enter PiP no longer receive unnecessary YouTube PiP setup work.
* PiP availability is limited to the top level tab WebContents, avoiding incorrect embedded content ownership.
* A quick navigation after tapping PiP no longer lets the fullscreen script run in a non YouTube document.

## Test plan (Check both mobile and tablet)

* On Android R or newer, open `m.youtube.com/watch?v=...`, wait for the player to load, then tap the toolbar PiP button. The video should enter Android PiP.
* Repeat immediately after the watch page opens, before the player controls have fully rendered. The video should still enter Android PiP once the control appears.
* Repeat on a slow network or cold cache page load.
* Start from an already fullscreen YouTube player and verify the Brave flow does not toggle the player back out of fullscreen.
* Tap the toolbar PiP button and immediately navigate away from YouTube. The script should not be injected into the new document.
* On Android 10 API 29, verify the toolbar flow remains unavailable and Brave does not perform the YouTube PiP support injection or final PiP entry call.
* Verify a normal YouTube watch page still shows PiP availability only for the outer tab WebContents.

### DIFF
--- a/browser/android/youtube_script_injector/youtube_script_injector_tab_helper.cc
+++ b/browser/android/youtube_script_injector/youtube_script_injector_tab_helper.cc
@@ -100,83 +100,169 @@ constexpr char16_t kYoutubeFullscreen[] =
 (function() {
   return new Promise((resolve) => {
     const videoPlaySelector = "video.html5-main-video";
-    const fullscreenSelector = "button.fullscreen-icon";
-    function triggerFullscreen() {
-      // Check if the video is not in fullscreen mode already.
-      if (!document.fullscreenElement) {
-        var fullscreenBtn = document.querySelector(fullscreenSelector);
-        var videoPlayer = document.querySelector(videoPlaySelector);
-        // Check if fullscreen button and video are available.
-        if (fullscreenBtn && videoPlayer) {
-         requestFullscreen(fullscreenBtn, resolve, videoPlayer);
-        } else {
-          // When fullscreen button is not available
-          // clicking the movie player resume the UI.
-          var playerContainer = document.getElementById("player-container-id");
-          if (videoPlayer && playerContainer) {
-            let observerTimeout;
-            // Create a MutationObserver to watch for changes in the DOM.
-            const observer = new MutationObserver(
-            (_mutationsList, observer) => {
-              var fullscreenBtn = document.querySelector(fullscreenSelector);
-              var videoPlayer = document.querySelector(videoPlaySelector);
-              if (fullscreenBtn && videoPlayer) {
-                clearTimeout(observerTimeout);
-                observer.disconnect()
-                requestFullscreen(fullscreenBtn, resolve, videoPlayer);
-              }
-            });
-            // Auto-disconnect the observer after 30 seconds,
-            // a reasonable duration picked after some testing.
-            observerTimeout = setTimeout(() => {
-              observer.disconnect();
-              resolve('timeout');
-            }, 30000);
-            // Start observing the DOM.
-            observer.observe(playerContainer, {
-              childList: true, subtree: true
-            });
-            // Make sure the player is in focus or responsive.
-            videoPlayer.click();
-          } else {
-            // No fullscreen elements found, resolve immediately
-            resolve('no_elements');
-          }
-        }
-      } else {
-        // Already in fullscreen, resolve immediately
-        resolve('already_fullscreen');
+    const fullscreenSelector = "button.fullscreen-icon, "
+        + "button.ytp-fullscreen-button, .ytp-fullscreen-button";
+    const playerSelector = "#movie_player, .html5-video-player";
+    const playerContainerSelector = "#player-container-id, ytm-player, #player";
+    const maxAttempts = 6;
+    const retryDelayMs = 350;
+    function isFullscreen(player) {
+      player = player || document.querySelector(playerSelector);
+      const isPlayerFullscreen = player?.isFullscreen?.();
+      return !!document.fullscreenElement || !!isPlayerFullscreen;
+    }
+    function hasFullscreenClass(player) {
+      return !!(player && player.classList.contains("ytp-fullscreen"));
+    }
+    function revealControls(videoPlayer, player, playerContainer) {
+      if (videoPlayer) {
+        videoPlayer.click();
+      } else if (player) {
+        player.click();
+      } else if (playerContainer) {
+        playerContainer.click();
       }
     }
-    // Attempts to request fullscreen mode for the given movie player element.
-    // Resolves with 'fullscreen_triggered' if successful, or
-    // 'requestFullscreen_failed' if the request fails.
-    function requestFullscreen(fullscreenBtn, resolve, videoPlayer) {
+    function triggerFullscreen(attempt) {
+      const player = document.querySelector(playerSelector);
+      if (isFullscreen(player)) {
+        resolve('already_fullscreen');
+        return;
+      }
+
+      const fullscreenBtn = document.querySelector(fullscreenSelector);
+      const videoPlayer = document.querySelector(videoPlaySelector);
+      const playerContainer = document.querySelector(playerContainerSelector);
+      if (videoPlayer) {
+        const playResult = videoPlayer.play?.();
+        if (playResult) {
+          playResult.catch(() => {});
+        }
+      }
+
+      if (fullscreenBtn && videoPlayer) {
+        requestFullscreen(fullscreenBtn, videoPlayer, attempt);
+        return;
+      }
+
+      // YouTube sometimes hides the fullscreen button until the player
+      // receives a tap (e.g. on fresh load). Reveal controls then retry.
+      revealControls(videoPlayer, player, playerContainer);
+      if (requestYoutubeFullscreen(player, attempt)) {
+        return;
+      }
+      if (requestElementFullscreen(player || playerContainer || videoPlayer,
+              attempt)) {
+        return;
+      }
+      scheduleRetry(attempt);
+    }
+    function requestFullscreen(fullscreenBtn, videoPlayer, attempt) {
       if (videoPlayer.readyState >= 3) {
         videoPlayer.click();
-        clickFullscreenButton(fullscreenBtn, resolve);
+        clickFullscreenButton(fullscreenBtn, attempt);
       } else {
         videoPlayer.addEventListener("canplay", () => {
           videoPlayer.click();
-          clickFullscreenButton(fullscreenBtn, resolve);
+          clickFullscreenButton(fullscreenBtn, attempt);
         }, { once: true });
+        scheduleRetry(attempt);
       }
     }
-    function clickFullscreenButton(fullscreenBtn, resolve) {
-      if (fullscreenBtn && !document.hidden) {
+    function clickFullscreenButton(fullscreenBtn, attempt) {
+      if (isFullscreen()) {
+        resolve('already_fullscreen');
+        return;
+      }
+      if (fullscreenBtn) {
         fullscreenBtn.click();
         resolve('fullscreen_triggered');
       } else {
-        resolve('requestFullscreen_failed');
+        scheduleRetry(attempt);
       }
+    }
+    function requestYoutubeFullscreen(player, attempt) {
+      if (isFullscreen(player)) {
+        resolve('already_fullscreen');
+        return true;
+      }
+      if (!player || !player.toggleFullscreen || hasFullscreenClass(player)) {
+        return false;
+      }
+
+      try {
+        player.toggleFullscreen();
+        if (requestElementFullscreen(player, attempt)) {
+          return true;
+        }
+        waitForFullscreen(attempt);
+        return true;
+      } catch (e) {
+        return false;
+      }
+    }
+    function waitForFullscreen(attempt) {
+      setTimeout(() => {
+        if (isFullscreen()) {
+          resolve('fullscreen_triggered');
+          return;
+        }
+        scheduleRetry(attempt);
+      }, retryDelayMs);
+    }
+    function requestElementFullscreen(element, attempt) {
+      if (isFullscreen()) {
+        resolve('already_fullscreen');
+        return true;
+      }
+
+      if (!element?.requestFullscreen) {
+        return false;
+      }
+
+      try {
+        const result = element.requestFullscreen();
+        if (result && result.then) {
+          result.then(() => resolve('fullscreen_triggered'))
+              .catch(() => scheduleRetry(attempt));
+        } else {
+          resolve('fullscreen_triggered');
+        }
+        return true;
+      } catch (e) {
+        return false;
+      }
+    }
+    function scheduleRetry(attempt) {
+      if (attempt >= maxAttempts) {
+        resolve('no_elements');
+        return;
+      }
+      setTimeout(() => triggerFullscreen(attempt + 1), retryDelayMs);
     }
     if (document.readyState === "loading") {
       // Loading hasn't finished yet.
       document.addEventListener("DOMContentLoaded",
-      triggerFullscreen, { once: true });
+      () => triggerFullscreen(0), { once: true });
     } else {
       // `DOMContentLoaded` has already fired.
-      triggerFullscreen();
+      triggerFullscreen(0);
+    }
+  });
+}());
+)";
+
+constexpr char16_t kYoutubeExitFullscreen[] =
+    uR"(
+(function() {
+  if (!document.fullscreenElement || !document.exitFullscreen) {
+    return;
+  }
+
+  document.exitFullscreen().catch(() => {
+    const fullscreenBtn = document.querySelector("button.fullscreen-icon");
+    if (fullscreenBtn && document.fullscreenElement) {
+      fullscreenBtn.click();
     }
   });
 }());
@@ -269,6 +355,16 @@ void YouTubeScriptInjectorTabHelper::MaybeSetFullscreen() {
       base::BindOnce(
           &YouTubeScriptInjectorTabHelper::OnFullscreenScriptComplete,
           weak_factory_.GetWeakPtr(), rfh->GetGlobalFrameToken()));
+}
+
+void YouTubeScriptInjectorTabHelper::MaybeExitFullscreen() {
+  content::RenderFrameHost* rfh = web_contents()->GetPrimaryMainFrame();
+  if (!rfh || !rfh->IsRenderFrameLive()) {
+    return;
+  }
+
+  content::RenderFrameHost::AllowInjectingJavaScript();
+  rfh->ExecuteJavaScript(kYoutubeExitFullscreen, base::NullCallback());
 }
 
 bool YouTubeScriptInjectorTabHelper::IsYouTubeDomain(bool mobileOnly) const {
--- a/browser/android/youtube_script_injector/youtube_script_injector_tab_helper.cc
+++ b/browser/android/youtube_script_injector/youtube_script_injector_tab_helper.cc
@@ -156,6 +156,14 @@ constexpr char16_t kYoutubeFullscreen[] =
     }
 
     function tryYoutubeApi() {
+      // Bail if the player or document already reports fullscreen. toggleFullscreen()
+      // is YouTube's undocumented API; calling it from a fullscreen state would flip
+      // us back out, which on the entry path would silently land us in PiP with no
+      // fullscreen content. The classList check alone is not enough: the player can
+      // be in document-level fullscreen without the ytp-fullscreen class set.
+      if (isFullscreen()) {
+        return false;
+      }
       const player = document.querySelector(playerSelector);
       if (!player?.toggleFullscreen
           || player.classList.contains('ytp-fullscreen')) {
--- a/browser/android/youtube_script_injector/youtube_script_injector_tab_helper.cc
+++ b/browser/android/youtube_script_injector/youtube_script_injector_tab_helper.cc
@@ -505,9 +505,19 @@ void YouTubeScriptInjectorTabHelper::OnFullscreenScriptComplete(
 }
 
 bool YouTubeScriptInjectorTabHelper::IsPictureInPictureAvailable() const {
+  if (!web_contents()) {
+    return false;
+  }
+  // Reject inner WebContents (guest views, fenced frames, portals). PiP must be
+  // driven from the outermost tab WebContents; offering it on an embedded view
+  // would target the wrong frame tree and bypass the tab level lifecycle the
+  // controller depends on.
+  if (web_contents()->GetOuterWebContents() != nullptr) {
+    return false;
+  }
   return base::FeatureList::IsEnabled(
              preferences::features::kBravePictureInPictureForYouTubeVideos) &&
-         IsYouTubeVideo(true) && web_contents() &&
+         IsYouTubeVideo(true) &&
          web_contents()->IsDocumentOnLoadCompletedInPrimaryMainFrame();
 }
 
--- a/browser/android/youtube_script_injector/youtube_script_injector_tab_helper.cc
+++ b/browser/android/youtube_script_injector/youtube_script_injector_tab_helper.cc
@@ -406,8 +406,11 @@ void YouTubeScriptInjectorTabHelper::MaybeExitFullscreen() {
     return;
   }
 
-  content::RenderFrameHost::AllowInjectingJavaScript();
-  rfh->ExecuteJavaScript(kYoutubeExitFullscreen, base::NullCallback());
+  EnsureBound(rfh);
+  script_injector_remote_->RequestAsyncExecuteScript(
+      ISOLATED_WORLD_ID_BRAVE_INTERNAL, kYoutubeExitFullscreen,
+      blink::mojom::UserActivationOption::kActivate,
+      blink::mojom::PromiseResultOption::kDoNotWait, base::NullCallback());
 }
 
 bool YouTubeScriptInjectorTabHelper::IsYouTubeDomain(bool mobileOnly) const {
--- a/browser/android/youtube_script_injector/youtube_script_injector_tab_helper.cc
+++ b/browser/android/youtube_script_injector/youtube_script_injector_tab_helper.cc
@@ -247,45 +247,6 @@ constexpr char16_t kYoutubeFullscreen[] =
 }());
 )";
 
-constexpr char16_t kYoutubeExitFullscreen[] =
-    uR"(
-(function() {
-  // Mirror the entry-side selectors so we can fall back to whichever
-  // fullscreen button this YouTube variant is rendering.
-  const fullscreenSelector = "button.fullscreen-icon, "
-      + "button.ytp-fullscreen-button, .ytp-fullscreen-button";
-  const playerSelector = "#movie_player, .html5-video-player";
-  const player = document.querySelector(playerSelector);
-  const isPlayerFullscreen = player?.isFullscreen?.() === true;
-
-  if (document.fullscreenElement && document.exitFullscreen) {
-    document.exitFullscreen().catch(() => {
-      const fullscreenBtn = document.querySelector(fullscreenSelector);
-      if (fullscreenBtn && document.fullscreenElement) {
-        fullscreenBtn.click();
-      }
-    });
-    return;
-  }
-
-  if (isPlayerFullscreen && player?.toggleFullscreen) {
-    try {
-      player.toggleFullscreen();
-      return;
-    } catch (e) {
-      // Fall through to the button-based fallback below.
-    }
-  }
-
-  if (isPlayerFullscreen) {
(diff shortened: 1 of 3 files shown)


android/brave_java_sources.gni
```diff
@@ -239,6 +239,7 @@ brave_java_sources = [
   "../../brave/android/java/org/chromium/chrome/browser/local_database/TopSiteTable.java",
   "../../brave/android/java/org/chromium/chrome/browser/media/BraveFullscreenVideoPictureInPictureController.java",
   "../../brave/android/java/org/chromium/chrome/browser/media/BravePictureInPictureActivity.java",
+  "../../brave/android/java/org/chromium/chrome/browser/media/BraveYouTubePictureInPictureController.java",
```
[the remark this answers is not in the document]
simoarpe: I'm having a look at this and see if we can avoid circular dependency issues as it currently requires `BraveActivity`.

android/java/org/chromium/chrome/browser/media/BraveFullscreenVideoPictureInPictureController.java
```diff
@@ -16,18 +16,81 @@ public class BraveFullscreenVideoPictureInPictureController {
      */
     protected boolean mDismissPending;
 
-    void dismissActivityIfNeeded(Activity activity, /*MetricsEndReason*/ int reason) {
-        if (reason == 8 /*MetricsEndReason.START*/ || reason == 0 /*MetricsEndReason.RESUME*/) {
+    // LINT.IfChange(brave_fullscreen_video_pip_dismiss_hook)
```
AlexeyBarabash: what is `brave_fullscreen_video_pip_dismiss_hook`?

android/java/org/chromium/chrome/browser/media/BraveYouTubePictureInPictureController.java
```diff
@@ -0,0 +1,554 @@
...
+
+    private @Nullable WebContents mWebContents;
+    private boolean mActive;
+    private boolean mExiting;
+    private boolean mInterruptedByScreenLock;
+    private int mSessionId;
+    private @Nullable BroadcastReceiver mScreenStateReceiver;
+
+    /**
+     * True iff a PiP entry should resume the media session once the system reports the activity is
```
AlexeyBarabash: nit: seems a typo `iff`

android/java/org/chromium/chrome/browser/media/BraveFullscreenVideoPictureInPictureController.java
simoarpe: It's a leftover from a previous commit, I'll remove it!

android/java/org/chromium/chrome/browser/media/BraveYouTubePictureInPictureController.java
simoarpe: I wanted to use it as an abbreviation for "If and only if", but I agree is not very clear: I'll change it.

simoarpe: There are also a couple of other occurrences of this, I'll change all of them.

AlexeyBarabash (approved):
LGTM

android/java/org/chromium/chrome/browser/media/BraveYouTubePictureInPictureController.java
AlexeyBarabash: I didn't know about this abbreviation

rewrite/chrome/android/java/src/org/chromium/chrome/browser/fullscreen/FullscreenHtmlApiHandlerBase.java.toml
```diff
@@ -0,0 +1,70 @@
...
+# LINT.IfChange(brave_fullscreen_html_api_handler_base)
+# Keep this file in sync with the substitution targets it relies on:
+#   * BraveFullscreenHtmlApiHandlerBase.shouldPreservePersistentFullscreenForPictureInPicture
+#   * BraveFullscreenHtmlApiHandlerBase.exitPersistentFullscreenModeForPictureInPicture
+# in src/brave/android/java/org/chromium/chrome/browser/fullscreen/
+# BraveFullscreenHtmlApiHandlerBase.java.
+
+[[substitution]]
+description = 'Preserve fullscreen state while Brave-managed YouTube PiP is active when a tab is hidden'
+re_pattern = '''(public void onHidden\(Tab tab, @TabHidingType int reason\)\s+\{\n)([\s\S]*?\n)( +)mFullscreenManagerDelegate\.onExitFullscreen\(null\);'''
```
AlexeyBarabash: at this place you capture indentation with `\3`  to re-use, but in other places at this file an other files you don't use such way, and just use the space chars for indentation.

Is here some strong reason to capture `\3`?

simoarpe: Good catch! There's no strong reason. 
I checked and this is the only .toml file that captures indentation with ( +), every other plaster file just hardcodes the spaces. The capture would only buy us resilience to upstream re-indentation, which is rare and would be cleaned up by git cl format anyway. I've switched substitutions to literal spaces for consistency in [a5ec6fe](https://github.com/brave/brave-core/pull/36087/commits/a5ec6fe89b4cb0166afb8be0e74822f675145365) . I confirm the regenerated patch is byte-identical to before.

AlexeyBarabash: thanks!

android/brave_java_sources.gni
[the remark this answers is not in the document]
AlexeyBarabash: this PR is big by itself, @simoarpe,  please create a follow-up issue to investigate if we can put all `brave/android/java/org/chromium/chrome/browser/media/*.java` into a Java library

netzenbot (approved):

diracdeltas:
re: the security hotspots, are they actually new or just refactored as part of this PR? if new could you please explain what they're doing?

simoarpe:
@diracdeltas the script tries `document.exitFullscreen()`, falls back to YouTube's undocumented `player.toggleFullscreen()`, and then to clicking the player's fullscreen button. It's purely the inverse of the entry script in `kYoutubeFullscreen`.

It runs in `ISOLATED_WORLD_ID_BRAVE_INTERNAL`, the same world as the existing entry script. The script can't be observed by page JS and shares no JS heap with the page world.
I've used `kActivate`  same as the entry call, so YouTube treats it as a user initiated exit. 
`kDoNotWait` and `NullCallback` are the only differences from the entry call's `kYoutubeFullscreen` as we don't need the script's return value because the controller now waits for the renderer's `didToggleFullscreenModeForTab(false, ...)`instead of the JS promise.

It's a symmetric counterpart to the existing entry injection, same isolated world ID, same hardcoded-script logic, no new user-input surface. Just inverting the operation at session teardown.

diracdeltas:
> @diracdeltas the script tries `document.exitFullscreen()`, falls back to YouTube's undocumented `player.toggleFullscreen()`, and then to clicking the player's fullscreen button. It's purely the inverse of the entry script in `kYoutubeFullscreen`.
> 
> It runs in `ISOLATED_WORLD_ID_BRAVE_INTERNAL`, the same world as the existing entry script. The script can't be observed by page JS and shares no JS heap with the page world. I've used `kActivate` same as the entry call, so YouTube treats it as a user initiated exit. `kDoNotWait` and `NullCallback` are the only differences from the entry call's `kYoutubeFullscreen` as we don't need the script's return value because the controller now waits for the renderer's `didToggleFullscreenModeForTab(false, ...)`instead of the JS promise.
> 
> It's a symmetric counterpart to the existing entry injection, same isolated world ID, same hardcoded-script logic, no new user-input surface. Just inverting the operation at session teardown.

thanks for the detailed info. @bridiver any concerns?

browser/android/youtube_script_injector/youtube_script_injector_tab_helper.cc
```diff
@@ -271,6 +387,19 @@ void YouTubeScriptInjectorTabHelper::MaybeSetFullscreen() {
...
 }
 
+void YouTubeScriptInjectorTabHelper::MaybeExitFullscreen() {
+  content::RenderFrameHost* rfh = web_contents()->GetPrimaryMainFrame();
+  if (!rfh || !rfh->IsRenderFrameLive()) {
+    return;
+  }
+
+  EnsureBound(rfh);
+  script_injector_remote_->RequestAsyncExecuteScript(
```
diracdeltas: does this need to check `IsYouTubeDomain` before injecting or is that done elsewhere? if done elsewhere, what if the webcontents is navigated between that check and the script injection?

simoarpe: Good catch. The domain check happens in the toolbar when tapping the PiP icon via `BraveYouTubeScriptInjectorNativeHelper.isPictureInPictureAvailable(webContents)` (which calls `IsYouTubeVideo` → `IsYouTubeDomain`), and the exit path inherits that guarantee implicitly from the fact that the controller only ever calls `exitFullscreen` against a `WebContents` it bound to an active Brave-managed YT PiP session.

But you're right that there's a navigation race in the [caller check] → [JNI → Mojo → renderer execution] window. It's more relevant for the exit path: the exit script first action is `document.exitFullscreen()`, which would be observable on any non-YouTube page that happened to be in DOM fullscreen when the navigation committed during that window. The entry script selectors just miss harmlessly on a non-YT page.

Pushed `IsYouTubeDomain()` checks at the top of both `MaybeSetFullscreen` and `MaybeExitFullscreen` so the C++ helper itself is the source of truth instead of depending on the call-site assumption holding across the IPC 👉 [`87afd54` (this PR)](https://github.com/brave/brave-core/pull/36087/commits/87afd545515e985c8185c88ec40ffb9bb44e438d)

android/brave_java_sources.gni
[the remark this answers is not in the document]
simoarpe: Created follow up issue: https://github.com/brave/brave-browser/issues/55338

browser/android/youtube_script_injector/youtube_script_injector_tab_helper.cc
```diff
@@ -246,16 +361,22 @@ void YouTubeScriptInjectorTabHelper::MediaEffectivelyFullscreenChanged(
...
       ::youtube_script_injector::EnterPictureInPicture(web_contents());
     }
   }
 }
 
 void YouTubeScriptInjectorTabHelper::MaybeSetFullscreen() {
   content::RenderFrameHost* rfh = web_contents()->GetPrimaryMainFrame();
-  // Check if fullscreen has already been requested for this page.
-  if (!rfh || !rfh->IsRenderFrameLive() || HasFullscreenBeenRequested()) {
+  // Check if fullscreen has already been requested for this page. The
```
bridiver, line 334: This sounds like a separate issue that should be in it's own PR? I feel like there are maybe several distinct changes here that have all been rolled together and that makes it harder to understand. There is this, then also putting the JS in grd that we discussed in dm, adding the method to exit fullscreen and the unified controller all seem like distinct concerns that should be split into smaller PRs?

browser/android/youtube_script_injector/youtube_script_injector_tab_helper.cc
```diff
@@ -378,9 +518,19 @@ void YouTubeScriptInjectorTabHelper::OnFullscreenScriptComplete(
 }
 
 bool YouTubeScriptInjectorTabHelper::IsPictureInPictureAvailable() const {
+  if (!web_contents()) {
+    return false;
+  }
+  // Reject inner WebContents (guest views, fenced frames, portals). PiP must be
```
bridiver, line 466: this is another one that seems like it should be a separate PR? We don't want to roll a bunch of bug fixes into a larger refactoring PR

rewrite/chrome/android/java/src/org/chromium/chrome/browser/media/FullscreenVideoPictureInPictureController.java.toml
```diff
@@ -0,0 +1,81 @@
...
+re_pattern = '''(\s+Log\.e\(TAG, "Error entering PiP with bounds %s", bounds, e\);\n)'''
+replace = '''\1            if (braveActivity != null) {
+                braveActivity.onYouTubePictureInPictureEnterFailed();
+            }
+'''
+count = 1
+
+[[substitution]]
+description = 'Add a helper to detect Brave-managed YouTube PiP candidates for manual fullscreen-video PiP'
+re_pattern = '''(    private boolean canStartPipBasedOnRecentTasks\(\) \{\n\s+return AndroidTaskUtils\.getRecentAppTasksMatchingComponentNames\(\n\s+mActivity, NO_PIP_COMPONENT_NAMES\)\n\s+\.isEmpty\(\);\n\s+\}\n\n)(    /\*\*)'''
```
bridiver: patterns shouldn't rely on fixed whitespace `(    private ...`  and `(      /\*\*()`

rewrite/chrome/android/java/src/org/chromium/chrome/browser/media/FullscreenVideoPictureInPictureController.java.toml
```diff
@@ -0,0 +1,81 @@
...
+[[substitution]]
+description = 'Import the Brave YouTube PiP helper'
+pattern = 'import org.chromium.chrome.browser.tab.Tab;'
+replace = '''import org.chromium.chrome.browser.tab.Tab;
+import org.chromium.chrome.browser.youtube_script_injector.BraveYouTubeScriptInjectorNativeHelper;'''
+count = 1
+
+[[substitution]]
+description = 'Let Brave handle YouTube PiP dismiss signals using upstream enum names'
+re_pattern = '''(    private void dismissActivityIfNeeded\(Activity activity, @MetricsEndReason int reason\) \{\n)'''
```
bridiver: see below re fixed whitespace

patches/chrome-android-java-src-org-chromium-chrome-browser-media-FullscreenVideoPictureInPictureController.java.patch
```diff
@@ -0,0 +1,83 @@
...
+ import org.chromium.chrome.browser.fullscreen.FullscreenManager;
+ import org.chromium.chrome.browser.infobar.InfoBarContainer;
+ import org.chromium.chrome.browser.notifications.NotificationIntentInterceptor;
+ import org.chromium.chrome.browser.tab.EmptyTabObserver;
+ import org.chromium.chrome.browser.tab.Tab;
++import org.chromium.chrome.browser.youtube_script_injector.BraveYouTubeScriptInjectorNativeHelper;
+ import org.chromium.chrome.browser.util.AndroidTaskUtils;
+ import org.chromium.content_public.browser.MediaSession;
+ import org.chromium.content_public.browser.WebContents;
+@@ -246,6 +248,16 @@ public class FullscreenVideoPictureInPictureController {
```
bridiver: can we subclass FullscreenVideoPictureInPictureController instead of patching in new methods? Or use a patch to rename the `FullscreenVideoPictureInPictureController` and then create our own `FullscreenVideoPictureInPictureController` that subclasses the original? I think we want to avoid a lot of code to existing classes and we should still try to find alternatives. It's not a free pass to just insert whatever we want into upstream code.

browser/android/youtube_script_injector/youtube_script_injector_tab_helper.cc
```diff
@@ -271,6 +377,19 @@ void YouTubeScriptInjectorTabHelper::MaybeSetFullscreen() {
...
+  content::RenderFrameHost* rfh = web_contents()->GetPrimaryMainFrame();
+  if (!rfh || !rfh->IsRenderFrameLive()) {
+    return;
+  }
+
+  EnsureBound(rfh);
+  script_injector_remote_->RequestAsyncExecuteScript(
+      ISOLATED_WORLD_ID_BRAVE_INTERNAL, kYoutubeExitFullscreen,
+      blink::mojom::UserActivationOption::kActivate,
+      blink::mojom::PromiseResultOption::kDoNotWait, base::NullCallback());
```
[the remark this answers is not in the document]
simoarpe: Code removed from this PR.
I'm resolving this.

browser/android/youtube_script_injector/youtube_script_injector_tab_helper.cc
```diff
@@ -271,6 +377,19 @@ void YouTubeScriptInjectorTabHelper::MaybeSetFullscreen() {
...
 
+void YouTubeScriptInjectorTabHelper::MaybeExitFullscreen() {
+  content::RenderFrameHost* rfh = web_contents()->GetPrimaryMainFrame();
+  if (!rfh || !rfh->IsRenderFrameLive()) {
+    return;
+  }
+
+  EnsureBound(rfh);
+  script_injector_remote_->RequestAsyncExecuteScript(
+      ISOLATED_WORLD_ID_BRAVE_INTERNAL, kYoutubeExitFullscreen,
```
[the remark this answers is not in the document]
simoarpe: Code removed from this PR.
I'm resolving this.

browser/android/youtube_script_injector/youtube_script_injector_tab_helper.cc
```diff
@@ -271,6 +385,19 @@ void YouTubeScriptInjectorTabHelper::MaybeSetFullscreen() {
...
+      blink::mojom::UserActivationOption::kActivate,
+      blink::mojom::PromiseResultOption::kDoNotWait, base::NullCallback());
```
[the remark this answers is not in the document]
simoarpe: Code removed from this PR.
I'm resolving this.

browser/android/youtube_script_injector/youtube_script_injector_tab_helper.cc
```diff
@@ -271,6 +385,19 @@ void YouTubeScriptInjectorTabHelper::MaybeSetFullscreen() {
...
 
+void YouTubeScriptInjectorTabHelper::MaybeExitFullscreen() {
+  content::RenderFrameHost* rfh = web_contents()->GetPrimaryMainFrame();
+  if (!rfh || !rfh->IsRenderFrameLive()) {
+    return;
+  }
+
+  EnsureBound(rfh);
+  script_injector_remote_->RequestAsyncExecuteScript(
+      ISOLATED_WORLD_ID_BRAVE_INTERNAL, kYoutubeExitFullscreen,
```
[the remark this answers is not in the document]
simoarpe: Code removed from this PR.
I'm resolving this.

rewrite/chrome/android/java/src/org/chromium/chrome/browser/media/FullscreenVideoPictureInPictureController.java.toml
simoarpe: Code removed from this PR.
I'm resolving this.

patches/chrome-android-java-src-org-chromium-chrome-browser-media-FullscreenVideoPictureInPictureController.java.patch
simoarpe: Code removed from this PR.
I'm resolving this.

rewrite/chrome/android/java/src/org/chromium/chrome/browser/media/FullscreenVideoPictureInPictureController.java.toml
simoarpe: Code removed from this PR.
I'm resolving this.

simoarpe:
Discussed this PR in DMs. Closing this in favor of smaller PRs so we can tell which part fixes what and can easily rollback just one part of it.

- https://github.com/brave/brave-core/pull/37370
- https://github.com/brave/brave-core/pull/37366
- https://github.com/brave/brave-core/pull/37367
- https://github.com/brave/brave-core/pull/37371

PR #36088, opened 2026-05-01
Fix for `--3way` apply crash due to warnings

### LINKED ISSUE #55170: `brockit` crashes when `git apply --3way` fails but doesn't print `error:` in the first line
I had a crash recently as brockit called:

```
 git -C .. apply --3way --ignore-space-change --ignore-whitespace brave/patches/chrome-browser-ui-views-page_action-page_action_controller.h.patch 
```

And the output was:

```
brave/patches/chrome-browser-ui-views-page_action-page_action_controller.h.patch:27: trailing whitespace.
  virtual void SetOverrideTriggerableEvent(actions::ActionId action_id, 
error: chrome/browser/ui/views/page_action/page_action_controller.h: does not exist in index
```

cdesouza-chromium:
The way `brockit` was handling `--3way` apply failures was always
assuming that the first line would have error. This has proven wrong,
and caused a crash when whitespace warnings were print first. This
change corrects the logic to look for the error line.

Resolves https://github.com/brave/brave-browser/issues/55170

### DIFF
--- a/tools/cr/patchfile.py
+++ b/tools/cr/patchfile.py
@@ -182,8 +182,11 @@ class Patchfile:
                                         patch=replace(
                                             self,
                                             source_from_git=conflicted_file))
-            if e.stderr.startswith('error:'):
-                [_, reason] = e.stderr.strip().split(': ', 1)
+            error_line = next(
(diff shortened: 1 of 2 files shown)


AlexeyBarabash (approved):
lgtm

PR #36094, opened 2026-05-01
[iOS] Add GPC JavaScriptFeature

### LINKED ISSUE #55175: Add support for GPC when using single profile web view configuration
Add support for querying `navigator.globalPrivacyControl` from pages when the `UseProfileWebViewConfiguration` feature flag is enabled

kylehickinson:
This adds a JavaScriptFeature that adds `navigator.globalPrivacyControl` to the page when the `UseProfileWebViewConfiguration` is enabled.

Resolves https://github.com/brave/brave-browser/issues/55175

### Test
- Enable the `UseProfileWebViewConfiguration` feature flag
- Verify `navigator.globalPrivacyControl` is true on the page (through any test page or Safari web inspector)
- Go to settings > shields & privacy and disable Global Privacy Control 
- Refresh the page
- Verify `navigator.globalPrivacyControl` is false on the page (through any test page or Safari web inspector)

<details>
<summary>Personal Test Page</summary>

```
<!DOCTYPE html>
<html>
<head><title>GPC Check</title></head>
<body>
<div id="output"></div>
<script>
  const output = document.getElementById('output');
  const total = 100;
  const interval = 2000 / total;
  let attempt = 0;

  function check() {
    attempt++;
    const p = document.createElement('p');
    const ts = performance.now().toFixed(2);
    p.textContent = 'Attempt ' + attempt + ' [' + ts + 'ms]: navigator.globalPrivacyControl = ' + navigator.globalPrivacyControl;
    output.appendChild(p);
    if (attempt < total) setTimeout(check, interval);
  }

  check();
</script>
</body>
</html>
```

</details>

### DIFF
--- a/ios/browser/global_privacy_control/gpc_javascript_feature.mm
+++ b/ios/browser/global_privacy_control/gpc_javascript_feature.mm
@@ -1,0 +1,73 @@
+// Copyright (c) 2026 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#include "brave/ios/browser/global_privacy_control/gpc_javascript_feature.h"
+
+#include "brave/components/global_privacy_control/pref_names.h"
+#include "components/prefs/pref_service.h"
+#include "ios/chrome/browser/shared/model/profile/profile_ios.h"
+#include "ios/web/public/browser_state.h"
+#include "ios/web/public/js_messaging/java_script_feature.h"
+#include "ios/web/public/js_messaging/web_frame.h"
+#include "ios/web/web_state/ui/wk_web_view_configuration_provider.h"
+
+namespace {
+constexpr char kGPCJavaScriptFeatureKeyName[] = "gpc_java_script_feature";
+constexpr char kScriptName[] = "gpc";
+}  // namespace
+
+GPCJavaScriptFeature::GPCJavaScriptFeature(ProfileIOS* profile)
+    : JavaScriptFeature(web::ContentWorld::kPageContentWorld,
+                        /*feature_scripts=*/{}),
+      profile_(profile) {
+  prefs_change_registrar_.Init(profile_->GetPrefs());
+  prefs_change_registrar_.Add(
+      global_privacy_control::kGlobalPrivacyControlEnabled,
+      base::BindRepeating(&GPCJavaScriptFeature::GPCPrefUpdated,
+                          base::Unretained(this)));
+}
+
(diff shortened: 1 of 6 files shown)


ios/browser/global_privacy_control/gpc_javascript_feature.mm
```diff
@@ -0,0 +1,73 @@
...
+  bool is_gpc_enabled = profile_->GetPrefs()->GetBoolean(
+      global_privacy_control::kGlobalPrivacyControlEnabled);
+  return {FeatureScript::CreateWithFilename(
+      kScriptName, FeatureScript::InjectionTime::kDocumentStart,
+      FeatureScript::TargetFrames::kAllFrames,
+      FeatureScript::ReinjectionBehavior::kInjectOncePerWindow,
+      base::BindRepeating(
+          [](bool is_gpc_enabled) -> FeatureScript::PlaceholderReplacements {
+            return @{
+              @"window.gBraveGPCEnabledPlaceholder" : is_gpc_enabled ? @"true"
```
bridiver: is this exposed to the page? I think we want to integrate this into whatever needs to read it or put it in an isolated world?

bridiver: I think this can be done without polluting the global namespace and anything we do add that is page accessible needs privacy and possibly security review

kylehickinson: Im not sure I follow. Do you mean `window.gBraveGPCEnabledPlaceholder` specifically? That is just a placeholder which is valid TypeScript, its never actually exposed. You can find the same patterns used in Chromium: https://source.chromium.org/search?q=%22(window%20as%20any)%22%20filepath:ios

bridiver: there is some kind of script messaging used by upstream features like NavigationJavaScriptFeature that I think can be used to communicate when the value changes and we can remove the static script entirely and generate the initial property define dynamically?

kylehickinson: > anything we do add that is page accessible needs privacy and possibly security review

Yeah we have this ongoing security review for all migrated page scripts: https://github.com/brave/reviews/issues/2201

bridiver: if `window.gBraveGPCEnabledPlaceholder` is never exposed to the page then I guess this is ok, but seems a bit odd.

bridiver: if we're going to do this though lets be consistent with upstream naming and use `gCrWebPlaceholder` prefix?

StephenHeaps (approved):

netzenbot (approved):

stoletheminerals (approved):

PR #35906, opened 2026-04-24
[History Embeddings] Move internal logic into BraveBatchPassageEmbedder

darkdh:
Resolves https://github.com/brave/brave-browser/issues/54763

  Thins `BravePassageEmbeddingsService` to match upstream's
  `PassageEmbeddingsService` shape: it now just owns the background
  WebContents, reads model files from disk, and hands them to
  `BraveBatchPassageEmbedder`, which owns the full per-load lifecycle
  (factory init, receiver binding, disconnect).

  Split across 4 commits so each is independently reviewable:

  1. **Guard against concurrent contents creation** — in-flight flag so
     successive `MaybeCreateBackgroundContents` calls don't double-spawn
     WebContents before the factory callback returns.
     _Addresses [r3113112620](https://github.com/brave/brave-core/pull/32689#discussion_r3113112620)._

  2. **Replace model-load barrier with `base::OneShotEvent`s** — drops
     the `BarrierClosure(2)` + `component_ready_counted_` flag (needed
     because `AddObserver` re-fires `OnLocalModelsReady` synchronously).
     `OneShotEvent::Signal()` is idempotent, so the guard flag
     disappears.
     _Addresses [r3114239143](https://github.com/brave/brave-core/pull/32689#discussion_r3114239143),
  [r3113758418](https://github.com/brave/brave-core/pull/32689#discussion_r3113758418)._

  3. **Move factory/init orchestration into `BraveBatchPassageEmbedder`**
     — removes `pending_loads_` queue, `model_ready_` flag,
     `FulfillPendingLoads`/`FailPendingLoads`/`OnFactoryInitDone` state
     machine from the service. Replaces with a single
     `std::optional<PendingLoad>` + `MaybeStartLoad` poll; uses
     `base::ScopedClosureRunner` to collapse `OnLocalModelFilesLoaded`
     failure paths.
     _Addresses [r3113855476](https://github.com/brave/brave-core/pull/32689#discussion_r3113855476),
  [r3114149489](https://github.com/brave/brave-core/pull/32689#discussion_r3114149489)._

  4. **Extract `BraveBatchPassageEmbedder` to its own .h/.cc** — pure
     move, no code changes, now that the class is a standalone unit.

  5. **Add tests for `BraveBatchPassageEmbedder` and service-level edge cases**
     — a dedicated `brave_batch_passage_embedder_unittest.cc` covers
     the BatchEmbedder's surface (init success/failure, batch
     ordering, sequential serialization, empty/out-of-range renderer
[...]

PR #36104, opened 2026-05-01
[History Embeddings] Guard against concurrent contents creation

darkdh:
Part of https://github.com/brave/brave-browser/issues/54763. Issue resolved by the final PR in the stack (#36107).

## Summary

`MaybeCreateBackgroundContents` had no in-flight flag, so successive `BindPassageEmbedder` calls before `OnBackgroundContentsCreated` fired would each kick off an additional factory run. Track the pending creation and short-circuit while it is outstanding.

Addresses [r3113112620](https://github.com/brave/brave-core/pull/32689#discussion_r3113112620) on the original PR.

## Stack

This PR is part 1 of 4 splitting the original follow-up (#35906):

1. **This PR** — Guard against concurrent contents creation
2. #36105 — Replace barrier with `base::OneShotEvent`s
3. #36106 — Move factory/init into `BraveBatchPassageEmbedder`
4. #36107 — Extract `BraveBatchPassageEmbedder` to its own files

### DIFF
--- a/browser/history_embeddings/brave_passage_embeddings_service.cc
+++ b/browser/history_embeddings/brave_passage_embeddings_service.cc
@@ -368,9 +368,10 @@ void BravePassageEmbeddingsService::OnBackgroundContentsDestroyed(
 }
 
 void BravePassageEmbeddingsService::MaybeCreateBackgroundContents() {
-  if (background_web_contents_) {
+  if (background_web_contents_ || background_web_contents_creating_) {
     return;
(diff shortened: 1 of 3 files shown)


netzenbot (approved):

darkdh:
Superseded — the in-flight flag is redundant with the fast-fail introduced in #36106. Replaced by an orphan-renderer guard folded into #36106 as a separate commit.

PR #36111, opened 2026-05-01
Account: collapse 3-state model to 2 states with verification intent

### LINKED ISSUE #55109: Account: collapse 3-state model to 2 states with verification intent
Replace `AccountState` (`LoggedOutState`/`VerificationState`/`LoggedInState`) with a 2-arm union (`LoggedOutState`/`LoggedInState`) where each state optionally carries a verification intent ("registration"/"reset password" for logged-out users, "change password" for logged-in users). Migrate the `kBraveAccountVerificationToken` string pref to a `kBraveAccountVerification` dict pref holding both token and intent, and update all WebUI/desktop/iOS/Android dispatch sites accordingly.

szilardszaloki:
Resolves https://github.com/brave/brave-browser/issues/55109.

Follow-ups:
- https://github.com/brave/brave-browser/issues/55264
- https://github.com/brave/brave-browser/issues/55265
- https://github.com/brave/brave-browser/issues/55179


SergeyZhukovsky (approved):
android ++

kylehickinson (approved):

components/brave_account/brave_account_state_prefs.cc
```diff
@@ -0,0 +1,154 @@
...
+                          .FindString(prefs::keys::kAuthenticationToken);
+  return token ? *token : "";
+}
+
+std::string GetCachedServiceToken(const PrefService& pref_service,
+                                  const std::string& service_name) {
+  CHECK(!service_name.empty());
+
+  const auto* service_tokens = pref_service.GetDict(prefs::kBraveAccountState)
+                                   .FindDict(prefs::keys::kServiceTokens);
```
DJAndries, line 111: consider having multiple prefs instead of jam-packing all the values into one dict (i.e. service tokens dict pref, verification dict pref, etc.)

components/brave_account/brave_account_state_prefs.cc
```diff
@@ -0,0 +1,154 @@
...
+  const auto last_fetched_time = base::ValueToTime(*last_fetched_value);
+  if (!last_fetched_time) {
+    return "";
+  }
+
+  if (base::Time::Now() - *last_fetched_time >= kServiceTokenMaxAge) {
+    return "";
+  }
+
+  return *encrypted_service_token;
```
DJAndries, line 135: consider moving the encrypt/decrypt functions into this new state file in order to further decrease the size of the service class, and returning a decrypted token from this function

components/brave_account/brave_account_state.cc
```diff
@@ -0,0 +1,154 @@
...
+}
+
+void UpdateEmail(PrefService& pref_service, const std::string& email) {
+  CHECK(!email.empty());
+
+  ScopedDictPrefUpdate(&pref_service, prefs::kBraveAccountState)
+      ->Set(prefs::keys::kEmail, email);
+}
+
+void CacheServiceToken(PrefService& pref_service,
```
DJAndries: these functions frequently reference pref_service. consider creating a class that holds a pref service, and converting these functions into methods.

components/brave_account/brave_account_state_prefs.cc
```diff
@@ -0,0 +1,154 @@
...
+  const auto* verification = account_state.FindDict(prefs::keys::kVerification);
+  const auto intent =
+      verification ? verification->FindInt(prefs::keys::kVerificationIntent)
+                   : std::nullopt;
+
+  if (kind && *kind == prefs::state_kinds::kLoggedIn) {
+    const auto* email = account_state.FindString(prefs::keys::kEmail);
+    CHECK(email && !email->empty());
+    return mojom::AccountState::NewLoggedIn(mojom::LoggedInState::New(
+        *email, MakeVerification<mojom::LoggedInVerificationPtr>(intent)));
```
DJAndries, line 93: the only call sites of `MakeVerification` are in this function. consider removing `MakeVerification`, and moving its contents just above the if block.

components/brave_account/BUILD.gn
```diff
@@ -49,6 +48,26 @@ source_set("brave_account") {
   }
 }
 
+source_set("internal") {
+  visibility = [
+    ":*",
+    "//brave/components/email_aliases:service",
```
DJAndries, line 54: it really seems odd that email aliases should have access to this "internal" target. consider calling an accounts service method from email aliases to retrieve the email address, or moving pref_names.h to a separate target at the very least.

components/brave_account/brave_account_state.h
```diff
@@ -0,0 +1,75 @@
...
+                 const std::string& encrypted_authentication_token);
+
+mojom::AccountStatePtr GetAccountState(const PrefService& pref_service);
+
+std::string GetAuthenticationToken(const PrefService& pref_service);
+
+template <typename Intent>
+  requires std::same_as<Intent, mojom::LoggedOutVerificationIntent> ||
+           std::same_as<Intent, mojom::LoggedInVerificationIntent>
+std::string GetVerificationToken(const PrefService& pref_service,
```
DJAndries: it's a little brutal that we need a function template for this...

please consider combining the two verification intent enums to remove such complexities

DJAndries (approved):

components/brave_account/BUILD.gn
szilardszaloki: Yep, that's what the follow-up issue is for in the description (https://github.com/brave/brave-browser/issues/55179).

components/brave_account/brave_account_state_prefs.cc
szilardszaloki: I've thought about that. `Encrypt()`/`Decrypt()` both operate on `os_crypt_async::Encryptor`, which is asynchronously populated in `BraveAccountService`. There's also test wiring dependent on `EncryptCallbackForTesting()`/`DecryptCallbackForTesting()`. Moving them to `brave_account_state.{h,cc}` would pull the encryptor lifetime and the test-callback machinery into a module that's currently just pref schema, which is different concerns.

components/brave_account/brave_account_state_prefs.cc
szilardszaloki: The single dict is intentional (I did start out with multiple prefs originally btw). Each state transition replaces the whole pref in one `SetDict()`, so observers only ever see a complete state, not some halfway mix like `kind == "logged-in"` before the email pref has been updated, etc.

Splitting it into multiple prefs adds extra churn, for example coordinating several `Set()` calls with inconsistent intermediate states, or adding some transaction-like layer. Replacing the whole dict also makes it harder to accidentally keep stale fields from the previous state.

components/brave_account/brave_account_state.h
szilardszaloki: We can't really avoid the complexity entirely, unfortunately. Each state should only carry the intents that make sense for it, e.g. logged-out shouldn't be able to hold "password-change", and logged-in shouldn't be able to hold "registration". If we merge them, we lose that type-level guard and have to introduce runtime checks.

netzenbot (approved):

components/brave_account/brave_account_state.h
```diff
@@ -0,0 +1,75 @@
...
+void SetLoggedOutWithVerification(
+    PrefService& pref_service,
+    const std::string& encrypted_verification_token,
+    mojom::LoggedOutVerificationIntent intent);
+
+void SetLoggedIn(PrefService& pref_service,
+                 const std::string& email,
+                 const std::string& encrypted_authentication_token);
+
+mojom::AccountStatePtr GetAccountState(const PrefService& pref_service);
```
boocmp: It could be usefull to move `GetAccountState` from `internal` and make it public.

components/brave_account/brave_account_state_prefs.cc
szilardszaloki: That's right, but the two call sites use different template arguments (`LoggedInVerificationPtr` vs. `LoggedOutVerificationPtr`) for the two return paths. Inlining would require duplicating the intent-validation + construction logic twice. A template lambda would work but the call syntax (`make_verification.template operator()<mojom::LoggedInVerificationPtr>()`) is uglier than `MakeVerification<mojom::LoggedInVerificationPtr>()`.

components/brave_account/brave_account_state.cc
szilardszaloki: Good idea — will push a fix-up commit.

components/brave_account/brave_account_state.h
szilardszaloki: If another component needs to react to account state, the existing pattern is to bind `mojom::Authentication`, call `AddObserver()`, and handle `OnAccountStateChanged()` — that gets you both the initial state and notifications on change, atomically. A synchronous `GetAccountState()` reader would go stale unless paired with subscription, so it'd usually be paired with the observer anyway. I don't really see the point of bypassing that.

boocmp (approved):

components/brave_account/brave_account_state.cc
szilardszaloki: https://github.com/brave/brave-core/pull/36111/commits/94a144630c2130de299cd91472c2713219cd5f4d

PR #35813, opened 2026-04-22
[History Embeddings] Clarify zero data collection for history semantic search

darkdh:
Resolves https://github.com/brave/brave-browser/issues/54798

  Brave's history semantic search runs entirely on-device, but the inherited
  Chromium UI still implies a cloud round-trip. This PR reconciles the UI
  with reality:

  - Reword the history/omnibox disclaimer text — no data leaves the device,
    no human reviewers — by adding Brave-owned strings
    (`IDS_BRAVE_HISTORY_EMBEDDINGS_DISCLAIMER`,
    `IDS_BRAVE_OMNIBOX_HISTORY_EMBEDDINGS_DISCLAIMER_IPH`) and swapping the
    IDS in the chromium_src overrides for `history_embeddings_utils.cc` and
    `featured_search_provider.cc`. Owning the strings is resilient to
    upstream wording tweaks, and neither override needs a new GN dep —
    `brave/grit/brave_generated_resources.h` is reachable via the existing
    `//chrome/app:generated_resources` transitive dep, and
    `components/grit/brave_components_strings.h` via the existing
    `brave_components_omnibox_browser_deps` injection.
  - Remove UI that implies a cloud feature: thumbs up/down feedback buttons,
    the "Learn more" link on chrome://history and the history side panel,
    the IPH promo's "Manage your history search setting" link, and the
    omnibox disclaimer IPH "Learn more" link.
  - Hide `chrome://settings/ai` by forcing `IsHistoryEmbeddingsSettingVisible()`
    to false; with History Search being the only AI feature Brave surfaces,
    the entire AI settings section drops out.

### Side note
If you accidentally dismiss promo in brave://history, you can open dev tools console and 
```
localStorage.removeItem('history-embeddings-promo');
localStorage.removeItem('history-embeddings-answers-promo');
```
then reload the page

### Steps
1. Enable  brave://flags/#brave-history-embeddings and relaunch
2. Navigate to https://www.teslarati.com/tesla-robotaxi-appears-heading-new-u-s-city/ 
3. Wait for a few moments (embeddings are calculated in the background)
4. Navigate to brave://history/ and check disclaimer
Screenshot 2026-04-22 at 14 08 38
5. Search for "self driving" and check result, there should be no thumb up/down button
Screenshot 2026-04-22 at 14 08 55
6. Try to search on omnibox with `@history` with "self driving" and check disclaimer
Screenshot 2026-04-22 at 14 09 11
7. Go to setting, and check there is no "AI innovations" section
[...]

PR #36112, opened 2026-05-01
[History Embeddings] Hide omnibox thumbs up/down and fix disclaimer trailing space

### LINKED ISSUE #55178: Remove feedback buttons from urlbar for history embeddings
### Description

Remove feedback buttons from url bar when brave://flags/#brave-history-embeddings is enabled.
Image

Follow-up to: https://github.com/brave/brave-browser/issues/54798

### Steps to reproduce

1. brave://flags/#brave-history-embeddings
2. Browse a few pages to have some sites in history
3. Go into the url bar and type @history and a query that will trigger a match in history an AI result
4. Notice the thumbs up/down button in the url bar

### Actual result

See STR

### Expected result

No feedback buttons in the url bar.

### Reproduces how often

Easily reproduced

1.92.x

### Channel information

- [ ] release (stable)
- [ ] beta
- [x] nightly

### Reproducibility

### Miscellaneous information

_No response_

darkdh:
Resolves https://github.com/brave/brave-browser/issues/55178

Follow-up to #35813 (brave-browser#54798). The previous PR stripped the WebUI feedback buttons on `chrome://history` and the history side panel, but missed the native Views omnibox: thumbs up/down still appeared on `HISTORY_EMBEDDINGS` matches when the user typed `@history`.

## Summary

- **Hide omnibox thumbs up/down.** Plaster `OmniboxPopupSelection::IsControlPresentOnMatch` to inject a short-circuit `if ((true)) return false;` immediately after the `FOCUSED_BUTTON_THUMBS_UP` / `FOCUSED_BUTTON_THUMBS_DOWN` case labels. Matching only the case labels (not the existing return expression) keeps the plaster resilient if Chromium tweaks the body. With these selections never available, `OmniboxResultView::UpdateFeedbackButtonsVisibility` keeps the buttons hidden, keyboard traversal skips them, and `OmniboxEditModel::UpdateFeedbackOnMatch` / `ChromeOmniboxClient::ShowFeedbackPage` are unreachable from the omnibox.
- **Fix disclaimer DCHECK crash.** Combine the existing `featured_search_provider.cc` IDS swap with a drop of the upstream `+ u" "` separator into a single plaster substitution. With the link text already swapped to `IDS_BRAVE_EMPTY_STRING` in #35813, the separator left `match.contents` with a trailing space that tripped `AutocompleteMatch::SanitizeString` in `AutocompleteResult::AppendMatches`.

```
[FATAL:autocomplete_result.cc:287] DCHECK failed:
  AutocompleteMatch::SanitizeString(match.contents) == match.contents
```

## Test plan

1. Enable `brave://flags/#brave-history-embeddings` and relaunch.
2. Browse a few pages so history has matchable content.
3. Focus the URL bar, type `@history` and a query that matches.
4. Confirm:
   - No thumbs up/down buttons appear on the result row (hover and keyboard-tab through the row).
   - The disclaimer IPH renders as the last row with no trailing whitespace and no DCHECK crash.
5. Repeat in `chrome://history` semantic search to confirm #35813 behavior is unchanged.

### DIFF
--- a/rewrite/components/omnibox/browser/featured_search_provider.cc.toml
+++ b/rewrite/components/omnibox/browser/featured_search_provider.cc.toml
@@ -16,6 +16,14 @@ re_pattern = '\bIDS_OMNIBOX_HISTORY_EMBEDDINGS_DISCLAIMER_IPH_LINK_TEXT\b'
 replace = 'IDS_BRAVE_EMPTY_STRING'
 
 [[substitution]]
-description = 'Swap IDS_OMNIBOX_HISTORY_EMBEDDINGS_DISCLAIMER_IPH to the Brave-owned string.'
-re_pattern = '\bIDS_OMNIBOX_HISTORY_EMBEDDINGS_DISCLAIMER_IPH\b'
-replace = 'IDS_BRAVE_OMNIBOX_HISTORY_EMBEDDINGS_DISCLAIMER_IPH'
+description = '''Swap IDS_OMNIBOX_HISTORY_EMBEDDINGS_DISCLAIMER_IPH to the
+Brave-owned string and drop the upstream " " separator that follows.
+
+Without dropping the separator, match.contents would have a trailing
+space (the link text was swapped to IDS_BRAVE_EMPTY_STRING above) that
+fails AutocompleteMatch::SanitizeString in
+AutocompleteResult::AppendMatches.
+'''
+re_pattern = 'IDS_OMNIBOX_HISTORY_EMBEDDINGS_DISCLAIMER_IPH(\))\s*\+\s*u" "'
+replace = 'IDS_BRAVE_OMNIBOX_HISTORY_EMBEDDINGS_DISCLAIMER_IPH\1'
+count = 1
--- a/rewrite/components/omnibox/browser/omnibox_popup_selection.cc.toml
+++ b/rewrite/components/omnibox/browser/omnibox_popup_selection.cc.toml
@@ -1,0 +1,24 @@
+# Copyright (c) 2026 The Brave Authors. All rights reserved.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+# Brave's history semantic search runs entirely on-device and never sends
+# data anywhere. Suppress the omnibox thumbs up/down feedback buttons that
+# upstream attaches to HISTORY_EMBEDDINGS matches so no feedback UI is
+# reachable from the URL bar. With these selections never available,
+# OmniboxResultView::UpdateFeedbackButtonsVisibility keeps the buttons
+# hidden and keyboard/selection traversal skips over them.
+
+[[substitution]]
+description = '''Never expose thumbs up/down feedback buttons on omnibox
+HISTORY_EMBEDDINGS matches.
+
+Inject a short-circuit `return false;` immediately after the case
+labels so we don't depend on the exact form of the upstream return
+expression. This keeps the plaster resilient if Chromium tweaks the
+condition (e.g. swaps the match type check or refactors the body).
+'''
+re_pattern = '(case FOCUSED_BUTTON_THUMBS_DOWN:\n)(\s+)return'
+replace = '\1\2if ((true)) return false;\n\2return'
+count = 1
--- a/patches/components-omnibox-browser-featured_search_provider.cc.patch
+++ b/patches/components-omnibox-browser-featured_search_provider.cc.patch
@@ -1,14 +1,14 @@
 diff --git a/components/omnibox/browser/featured_search_provider.cc b/components/omnibox/browser/featured_search_provider.cc
-index 766588861b85c3a4ef69ed85522761aaf41cb620..8543840c0640d303ddd2b59934a743551480604c 100644
+index 766588861b85c3a4ef69ed85522761aaf41cb620..2cf5c1a4b3b24f47fca2fc6a469ec31952b73f44 100644
 --- a/components/omnibox/browser/featured_search_provider.cc
 +++ b/components/omnibox/browser/featured_search_provider.cc
-@@ -611,10 +611,10 @@ bool FeaturedSearchProvider::ShouldShowHistoryEmbeddingsDisclaimerIphMatch()
+@@ -611,10 +611,9 @@ bool FeaturedSearchProvider::ShouldShowHistoryEmbeddingsDisclaimerIphMatch()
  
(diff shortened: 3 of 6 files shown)


netzenbot (approved):

rewrite/components/omnibox/browser/featured_search_provider.cc.toml
```diff
@@ -15,6 +15,11 @@ description = 'Swap IDS_OMNIBOX_HISTORY_EMBEDDINGS_DISCLAIMER_IPH_LINK_TEXT to a
 re_pattern = '\bIDS_OMNIBOX_HISTORY_EMBEDDINGS_DISCLAIMER_IPH_LINK_TEXT\b'
 replace = 'IDS_BRAVE_EMPTY_STRING'
 
+[[substitution]]
+description = 'Drop the upstream " " separator appended after the disclaimer text. With the trailing link text swapped to IDS_BRAVE_EMPTY_STRING, the separator would leave match.contents with a trailing space that fails AutocompleteMatch::SanitizeString in AutocompleteResult::AppendMatches.'
```
cdesouza-chromium: Use ''' to have multiline comments, and try to keep this to 80 characters. I guess we need better docs about this, but it is like in python, where you have a title, and optionally a body:

Kinda like this:

```py
def _repair_plaster_files(old_chromium: Path,
                          new_chromium: Path,
                          no_git: bool,
                          run_plaster: bool = True) -> None:
    '''Moves the plaster file and deletes the corresponding patch file.

    Plaster file convention: chromium path A/foo.h lives at
    rewrite/A/foo.h.toml. If no plaster file exists, this is a no-op.
    The old patch file for that plaster gets automatically deleted.
    '''
```

rewrite/components/omnibox/browser/featured_search_provider.cc.toml
```diff
@@ -15,6 +15,11 @@ description = 'Swap IDS_OMNIBOX_HISTORY_EMBEDDINGS_DISCLAIMER_IPH_LINK_TEXT to a
...
+re_pattern = '(IDS_OMNIBOX_HISTORY_EMBEDDINGS_DISCLAIMER_IPH\))\s*\+\s*u" "'
+replace = '\1'
```
cdesouza-chromium: Add `count =1`

rewrite/components/omnibox/browser/featured_search_provider.cc.toml
```diff
@@ -15,6 +15,11 @@ description = 'Swap IDS_OMNIBOX_HISTORY_EMBEDDINGS_DISCLAIMER_IPH_LINK_TEXT to a
...
+
 [[substitution]]
 description = 'Swap IDS_OMNIBOX_HISTORY_EMBEDDINGS_DISCLAIMER_IPH to the Brave-owned string.'
```
cdesouza-chromium: add `count = 2`

rewrite/components/omnibox/browser/omnibox_popup_selection.cc.toml
```diff
@@ -0,0 +1,18 @@
...
+
+# Brave's history semantic search runs entirely on-device and never sends
+# data anywhere. Suppress the omnibox thumbs up/down feedback buttons that
+# upstream attaches to HISTORY_EMBEDDINGS matches so no feedback UI is
+# reachable from the URL bar. With these selections never available,
+# OmniboxResultView::UpdateFeedbackButtonsVisibility keeps the buttons
+# hidden and keyboard/selection traversal skips over them.
+
+[[substitution]]
+description = 'Never expose thumbs up/down feedback buttons on omnibox HISTORY_EMBEDDINGS matches.'
```
cdesouza-chromium: same about line length.

cdesouza-chromium: So, this one maybe can be made a bit more resilient, it depends what level of safety we want here. Currently the code is:

```cxx
    case FOCUSED_BUTTON_THUMBS_UP:
    case FOCUSED_BUTTON_THUMBS_DOWN:
      return match.type == AutocompleteMatchType::HISTORY_EMBEDDINGS;
```

And with this plaster we are getting:
```
    case FOCUSED_BUTTON_THUMBS_UP:
    case FOCUSED_BUTTON_THUMBS_DOWN:
      return false;
```

But if it is the case we never case that we don't ever care about the `return` line in the upstream file changing to something else, then we could make this plaster a bit more resilient this way:

```
    case FOCUSED_BUTTON_THUMBS_UP:
    case FOCUSED_BUTTON_THUMBS_DOWN:
+    if ((true)) return false;
       return match.type == AutocompleteMatchType::HISTORY_EMBEDDINGS;
```

In this case, we only match the keys , but not he expression bellow, so there's less that could cause a break to the expression. However, if breaking the plaster is intentional whenever the return expression changes, then that's okay to leave at is, but it may help to have something added to the body of the description explaining what the concerns are.

rewrite/components/omnibox/browser/omnibox_popup_selection.cc.toml
```diff
@@ -0,0 +1,18 @@
...
+re_pattern = 'case FOCUSED_BUTTON_THUMBS_UP:\s+case FOCUSED_BUTTON_THUMBS_DOWN:\s+return match\.type == AutocompleteMatchType::HISTORY_EMBEDDINGS;'
+replace = '''case FOCUSED_BUTTON_THUMBS_UP:
```
cdesouza-chromium: Add `count = 1`

cdesouza-chromium (approved):
LGTM % changes requested.

I'm leaving it already approved, so you can go ahead solve the comments, and not be blocked, unless you have some question about any request of course.

rewrite/components/omnibox/browser/featured_search_provider.cc.toml
darkdh: formated

rewrite/components/omnibox/browser/featured_search_provider.cc.toml
darkdh: added

rewrite/components/omnibox/browser/featured_search_provider.cc.toml
darkdh: I combined this two substitutions into one and with count

rewrite/components/omnibox/browser/omnibox_popup_selection.cc.toml
darkdh: agree, injected `if ((true)) return false;`

rewrite/components/omnibox/browser/omnibox_popup_selection.cc.toml
darkdh: added

PR #36117, opened 2026-05-02
[brockit] Support for `git apply` for renamed patches

### LINKED ISSUE #55188: Add `brockit` support for `--3way` reapply to renamed patches
We should do a modest implementation to experiment with this, by at least apply the patch to the destination and letting continue force the user to commit the culprit for such cases.

cdesouza-chromium:
This change adds a modest support to have `git apply --3way` attempted
to patches that have been renamed. The user report has been updated to
provide clues about the result of the attempt, and the routine
differentiates between a applying cleanly, conflicts, and broken
apply, and handles them accordingly.

A second pass on this feature may simplify things by instantiating a
`Patchfile` for the reapply operation.

Resolves https://github.com/brave/brave-browser/issues/55188

### DIFF
--- a/tools/cr/brockit.py
+++ b/tools/cr/brockit.py
@@ -972,10 +972,6 @@ class Upgrade(Versioned):
         When running brockit in a vscode terminal, this method will open any
         files that need attention in the editor session.
         """
-        # This is a flat list of patches that cannot be applied due to their
-        # source file being deleted.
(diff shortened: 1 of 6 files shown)


emerick (approved):

netzenbot (approved):

cdesouza-chromium:
This is the wrong approach. Abandoning this in favour of the follow renames tooling.

PR #36107, opened 2026-05-01
[History Embeddings] Extract BraveBatchPassageEmbedder to its own files

darkdh:
Resolves https://github.com/brave/brave-browser/issues/54763.

## Summary

Pure file-split — `BraveBatchPassageEmbedder` now owns factory init, the receiver, and the renderer pipe for a single load (per the preceding PR), so it is a standalone unit rather than a helper of `BravePassageEmbeddingsService`. Splitting the files makes the two classes independently readable and enables direct unit testing of the BatchEmbedder.

This PR also adds `brave_batch_passage_embedder_unittest.cc` covering the BatchEmbedder's surface (init success/failure, batch ordering, sequential serialization, empty/out-of-range renderer responses, receiver/factory/renderer disconnects).

Together with #36106, addresses [r3113855476](https://github.com/brave/brave-core/pull/32689#discussion_r3113855476) on the original PR.

## Stack

This PR is part 3 of 3 splitting the original follow-up (#35906):

1. #36124 — Replace model-load barrier with state machine
2. #36106 — Move factory/init into `BraveBatchPassageEmbedder` (+ orphan-renderer guard)
3. **This PR** — Extract `BraveBatchPassageEmbedder` to its own files

PR #36106, opened 2026-05-01
[History Embeddings] Move model lifecycle into controller, renderer-side into BatchEmbedder

darkdh:
Resolve https://github.com/brave/brave-browser/issues/54763

## Summary

`BravePassageEmbeddingsService` held the full embedder lifecycle — the background WebContents that hosts the WASM worker, the `local_ai::mojom::LocalAIService` receiver set, the renderer-side factory remote, a queue of pending loads, and a state machine that drove `factory_->Init` in addition to file loading — which made the set of reachable states hard to reason about and left room for the factory remote, the background contents, and the batch embedder to drift out of sync.

Restructure ownership to mirror upstream's `PassageEmbeddingsServiceController` / `PassageEmbeddingsService` split, where the controller owns the model lifecycle and the service is a thin factory.

**Renderer-side lifecycle into `BraveBatchPassageEmbedder`:**
* the background WebContents (delegate impl + owning ptr),
* the `LocalAIService` receiver set,
* the `PassageEmbedderFactory` remote and `RegisterPassageEmbedderFactory` handler (including the duplicate-registration guard),
* the `factory->Init` + `factory->Bind` handshake and the renderer-side `PassageEmbedder` pipe.

**Model-file loading into `BravePassageEmbeddingsServiceController`:**
The controller observes `LocalModelsUpdaterState` and posts the disk read for the five EmbeddingGemma files. `EmbedderReady()` returns true iff the component is installed; `OnLocalModelsReady` fires `EmbedderMetadataUpdated` so `SchedulingEmbedder` retries. `GetEmbeddings` short-circuits with `kModelUnavailable` when not ready, otherwise hands the loaded `ModelFilesPtr` to the service through `BindPassageEmbedder(receiver, files, cb)`.

**`BraveBatchPassageEmbedder` receives `ModelFilesPtr` in its ctor** and binds the caller-side `PassageEmbedder` receiver only from `OnInitDone(true)`. Mojo queues `GenerateEmbeddings` calls on the pipe until the receiver binds, mirroring upstream's utility-process embedder. The phase machine collapses to `kCreatingContents → kAwaitingFactory → kInitializing → kReady`; `ProcessNext`'s "renderer not bound" branch becomes a `CHECK`.

`BravePassageEmbeddingsService` shrinks to a thin factory: `BindPassageEmbedder(receiver, model_files, cb)` constructs a `BraveBatchPassageEmbedder` around the supplied files. No observer plumbing, no file-load state.

[...]

PR #36124, opened 2026-05-02
[History Embeddings] Replace model-load barrier with state machine

darkdh:
Part of https://github.com/brave/brave-browser/issues/54763. Issue resolved by the final PR in the stack (#36107).

## Summary

Two signals gate the EmbeddingGemma model load: the component being installed and the renderer-side `PassageEmbedderFactory` registering. A `BarrierClosure(2)` paired with a `component_ready_counted_` flag was needed because `LocalModelsUpdaterState::AddObserver` synchronously re-fires `OnLocalModelsReady` when `install_dir` is already set, which would otherwise tick the barrier twice.

Replace the barrier with a `LoadPhase` state machine (`kWaiting → kLoading → kInitializing → kReady`) driven by a single `MaybeLoadLocalModelFiles` entry point, called from `OnLocalModelsReady`, `RegisterPassageEmbedderFactory`, and `BindPassageEmbedder`. It re-enters only when `phase_ == kWaiting` and both `factory_.is_bound()` and `updater_state_->GetInstallDir()` are present, so the `component_ready_counted_` guard disappears and `AddObserver`'s sync re-fire is naturally a no-op.

Addresses [r3114239143](https://github.com/brave/brave-core/pull/32689#discussion_r3114239143) and [r3113758418](https://github.com/brave/brave-core/pull/32689#discussion_r3113758418) on the original PR.

## Stack

This PR is part 1 of 3 splitting the original follow-up (#35906):

1. **This PR** — Replace model-load barrier with state machine
2. #36106 — Move factory/init into `BraveBatchPassageEmbedder` (with orphan-renderer guard)
3. #36107 — Extract `BraveBatchPassageEmbedder` to its own files

Related (not in stack): #36184 — Reject duplicate `RegisterPassageEmbedderFactory`.
(#36104 was closed as superseded — its in-flight flag was redundant with the fast-fail in part 2; an orphan-renderer guard now lives in part 2 instead.)

### DIFF
--- a/browser/history_embeddings/brave_passage_embeddings_service.cc
+++ b/browser/history_embeddings/brave_passage_embeddings_service.cc
@@ -7,7 +7,6 @@
 
 #include <utility>
 
-#include "base/barrier_closure.h"
 #include "base/check.h"
 #include "base/containers/flat_map.h"
 #include "base/feature_list.h"
@@ -16,6 +15,7 @@
 #include "base/logging.h"
 #include "base/no_destructor.h"
 #include "base/numerics/safe_math.h"
+#include "base/one_shot_event.h"
 #include "base/task/thread_pool.h"
 #include "components/history_embeddings/core/history_embeddings_features.h"
 #include "content/public/browser/web_contents.h"
@@ -266,10 +266,12 @@ BravePassageEmbeddingsService::BravePassageEmbeddingsService(
       updater_state_(updater_state) {
   CHECK(base::FeatureList::IsEnabled(history_embeddings::kHistoryEmbeddings));
   CHECK(updater_state_);
-  // Arm the barrier FIRST so CountComponentReadyOnce can route through
-  // it. AddObserver's immediate-notify (when install_dir is already
-  // set) will invoke OnLocalModelsReady synchronously; the flag guard
-  // inside CountComponentReadyOnce keeps that from counting twice.
+  // Arm the events FIRST so AddObserver's immediate-notify (when
+  // install_dir is already set) has a live component_ready_event_ to
+  // signal. OneShotEvent::Signal() is idempotent under the
+  // is_signaled() guard, so the sync install_dir check inside
+  // MaybeWaitForLocalModelFilesReady and the observer callback can
+  // both fire without double-counting.
   MaybeWaitForLocalModelFilesReady();
   updater_state_->AddObserver(this);
 }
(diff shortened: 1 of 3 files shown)


browser/history_embeddings/brave_passage_embeddings_service.cc
```diff
@@ -397,21 +401,30 @@ void BravePassageEmbeddingsService::CloseBackgroundContents() {
...
 
 void BravePassageEmbeddingsService::MaybeWaitForLocalModelFilesReady() {
-  component_ready_counted_ = false;
-  model_load_barrier_ = base::BarrierClosure(
-      2, base::BindOnce(&BravePassageEmbeddingsService::LoadLocalModelFiles,
-                        weak_ptr_factory_.GetWeakPtr()));
+  component_ready_event_ = std::make_unique<base::OneShotEvent>();
+  factory_registered_event_ = std::make_unique<base::OneShotEvent>();
+  // Chain: once the component is ready, wait for the factory
+  // registration and then load the model files. OneShotEvent::Post
```
netzenbot: nit: "matching the pre-existing barrier's fire-once semantics" references code that no longer exists. A future reader has no context for what the 'pre-existing barrier' was. Drop the historical comparison and just describe the current behavior, e.g. "OneShotEvent::Post runs the closure asynchronously regardless of whether the event fires before or after Post is called." ([✅ Comments Must Make Sense to Future Readers of the Codebase](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-019))

browser/history_embeddings/brave_passage_embeddings_service_unittest.cc
```diff
@@ -267,6 +276,22 @@ TEST_F(BravePassageEmbeddingsServiceTest, BackgroundContentsDestroyedCloses) {
...
+  // re-fires OnLocalModelsReady synchronously when install_dir is
+  // already set, exercising the OneShotEvent's idempotent Signal()
+  // alongside the explicit signal in MaybeWaitForLocalModelFilesReady.
+  SetUpModelFiles();
+  RecreateService();
+
+  auto load = IssueLoad();
+  RegisterFactory();
+  ASSERT_TRUE(
+      base::test::RunUntil([&] { return fake_factory_.init_count() > 0; }));
```
netzenbot: This test is meant to exercise the idempotent Signal() scenario (install_dir already set at construction time triggers both the sync check and the observer callback). Asserting `init_count() > 0` won't catch a regression where `LoadLocalModelFiles` fires twice. Add `EXPECT_EQ(1, fake_factory_.init_count())` to verify the negative side effect — that the init was not triggered more than once. ([Verify Negative Expectations in Tests](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-021))

browser/history_embeddings/brave_passage_embeddings_service.cc
darkdh: not applicable with state machine changes

browser/history_embeddings/brave_passage_embeddings_service_unittest.cc
darkdh: fixed in fa2c43709690590d4f0b499a10d28e23690b9bd4 with exactly count == 1

bridiver (approved):

PR #36128, opened 2026-05-03
Add best practice: TabStripModelObserver teardown (no manual RemoveObserver)

sangwoo108:
We don't need to call TabStripModel::RemoveObserver, as TabStripModelObserver::~TabStripModelObserver() is doing this.


simonhong (approved):
++

netzenbot (approved):

PR #36129, opened 2026-05-04
Fix caption button overlap with toolbar in vertical tabs Qt theme

### LINKED ISSUE #54471: Missing Window Controls (Min/Max/Close) in Top-Right Corner when Vertical Tabs are Enabled (Custom Title Bar)
### Description

Environment:

- OS: Fedora 43
- Desktop Environment: KDE Plasma 6.6
- Display Server: Wayland
- Theme Engine: QT Theme applied

When vertical tabs are enabled in Brave Browser on Linux, the window control buttons (Minimize, Maximize, Close) disappear completely from the top-right corner of the browser window. This occurs specifically when the system title bar is disabled (using Brave's custom title bar) and a QT theme is in use under Wayland.

### Steps to reproduce

1. Launch Brave Browser on Fedora 43 (KDE Plasma/Wayland).
2. Ensure "Use system title bar and borders" is disabled in Brave settings.
3. Ensure a QT theme is being applied to the browser environment.
4. Enable "Vertical Tabs" (Right-click on a tab -> select "Use vertical tabs").
5. Observe the top-right corner of the browser window

### Actual result

The window controls are completely missing. Only browser-internal icons (Shields, Extensions, Media Controls, Sidebar, Wallet, etc.) are rendered.

Image

### Expected result

The window control buttons (Minimize, Maximize, Close) should be visible and functional in the top-right corner, integrating properly with the custom title bar alongside the extension icons.

### Reproduces how often

Easily reproduced

- Brave Version: 1.89.132 (Official Build) (64-bit)
- Chromium Version: 147.0.7727.56

### Channel information

- [x] release (stable)
- [ ] beta
- [ ] nightly

### Reproducibility

- [x] in the latest version of Chrome

### Miscellaneous information

_No response_

simonhong:
When vertical tabs with hidden title bar is enabled, window control buttons overlapped the toolbar on Linux with the Qt theme. GTK theme worked correctly.

The root cause was in GetLeadingTrailingCaptionButtonWidth(): it branched on UsingSystemTheme() which returns true for both GTK and Qt. The GTK path cast the frame view to BraveBrowserFrameViewLinuxNative to read actual button bounds. However, Qt doesn't use the native frame view (QtUi::CreateNavButtonProvider() returns nullptr), so the cast failed silently and returned zero-width exclusions, causing overlap.

Replace the custom GetLeadingTrailingCaptionButtonWidth() with upstream Chromium's BrowserFrameView::GetBrowserLayoutParams(), which reads actual button positions post-layout from whichever frame view is active. This works correctly for GTK, Qt, Classic, Windows, and macOS without platform-specific branching.

Also add a Layout(PassKey) override in BraveOpaqueBrowserFrameView to call UpdateHorizontalPadding() after buttons are positioned, fixing the overlap that occurred after theme switches before the next manual window resize.

Qt:
image

Classic:
image

Gtk:
image

Windows:
image

macOS:
image

Resolves https://github.com/brave/brave-browser/issues/54471

### DIFF
--- a/browser/ui/views/tabs/vertical_tab_utils.cc
+++ b/browser/ui/views/tabs/vertical_tab_utils.cc
@@ -95,7 +95,7 @@ std::pair<int, int> GetLeadingTrailingCaptionButtonWidth(
   if (!frame_view) {
     return {};
   }
-  const auto params = frame_view->GetBrowserLayoutParams();
+  const BrowserLayoutParams params = frame_view->GetBrowserLayoutParams();
   return {
       base::ClampCeil(params.leading_exclusion.ContentWithPadding().width()),
       base::ClampCeil(params.trailing_exclusion.ContentWithPadding().width())};
--- a/browser/ui/views/tabs/vertical_tab_utils.cc
+++ b/browser/ui/views/tabs/vertical_tab_utils.cc
@@ -16,28 +16,8 @@
 #include "chrome/browser/ui/browser_window/public/browser_window_features.h"
 #include "chrome/browser/ui/browser_window/public/browser_window_interface.h"
 #include "chrome/browser/ui/tabs/features.h"
-#include "chrome/browser/ui/views/frame/browser_native_widget.h"
 #include "components/prefs/pref_service.h"
 
-#if !BUILDFLAG(IS_MAC)
-#include "chrome/browser/ui/frame/window_frame_util.h"
-#include "chrome/browser/ui/views/frame/browser_view.h"
-#include "chrome/browser/ui/views/frame/opaque_browser_frame_view_layout.h"
-#include "ui/base/theme_provider.h"
-#include "ui/views/resources/grit/views_resources.h"
-#endif
-
-#if BUILDFLAG(IS_LINUX)
-#include "brave/browser/ui/views/frame/brave_browser_frame_view_linux_native.h"
-#include "chrome/browser/themes/theme_service.h"
-#include "chrome/browser/themes/theme_service_factory.h"
-#include "chrome/browser/ui/layout_constants.h"
-#include "chrome/common/pref_names.h"
-#include "ui/linux/linux_ui.h"
-#include "ui/views/view_utils.h"
-#include "ui/views/window/caption_button_layout_constants.h"
-#include "ui/views/window/window_button_order_provider.h"
-#endif
 
 namespace tabs::utils {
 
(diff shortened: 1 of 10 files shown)


browser/ui/views/tabs/vertical_tab_utils.cc
```diff
@@ -104,69 +87,18 @@ bool IsVerticalTabOnRight(const BrowserWindowInterface* browser) {
...
-  const ui::ThemeProvider* tp = frame->GetThemeProvider();
-  DCHECK(tp);
-  for (auto image_id : {IDR_MINIMIZE, IDR_MAXIMIZE, IDR_CLOSE}) {
-    if (const gfx::ImageSkia* image = tp->GetImageSkiaNamed(image_id)) {
-      width += image->width();
-    }
-  }
-  return {0, width};
+  const auto params = frame_view->GetBrowserLayoutParams();
+  return {
```
netzenbot, line 99: nit: `const auto params = frame_view->GetBrowserLayoutParams()` — `BrowserLayoutParams` is short and descriptive; prefer spelling it out explicitly per [CS-053](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-053).

browser/ui/views/toolbar/brave_toolbar_view_browsertest.cc
```diff
@@ -605,3 +606,45 @@ IN_PROC_BROWSER_TEST_F(BraveToolbarViewTest,
   prefs->SetBoolean(brave_tabs::kVerticalTabsEnabled, false);
```
netzenbot, line 606: nit: These assertions transition through multiple states; adding `<< "after enabling vertical tabs, hiding title bar"` would pinpoint which state fails without needing a debugger. See [TI-014](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-014).

browser/ui/views/tabs/vertical_tab_utils.cc
simonhong: Done.

browser/ui/views/toolbar/brave_toolbar_view_browsertest.cc
simonhong: Updated.

browser/ui/views/toolbar/brave_toolbar_view.cc
```diff
@@ -496,13 +496,22 @@ void BraveToolbarView::UpdateHorizontalPadding() {
...
-        tabs::utils::GetLeadingTrailingCaptionButtonWidth(
-            browser_view_->browser_widget());
-    container_view->SetBorder(views::CreateEmptyBorder(
-        gfx::Insets().set_left(leading).set_right(trailing)));
+    return;
+  }
+
+  const gfx::Insets current_insets =
+      container_view->GetBorder() ? container_view->GetBorder()->GetInsets()
+                                  : gfx::Insets();
```
sangwoo108: Wonder if we can just use container_view->GetInsets(), which would return gfx::Insets() in case there's no border

browser/ui/views/tabs/vertical_tab_utils.cc
```diff
@@ -104,69 +87,18 @@ bool IsVerticalTabOnRight(const BrowserWindowInterface* browser) {
...
-  for (auto image_id : {IDR_MINIMIZE, IDR_MAXIMIZE, IDR_CLOSE}) {
-    if (const gfx::ImageSkia* image = tp->GetImageSkiaNamed(image_id)) {
-      width += image->width();
-    }
-  }
-  return {0, width};
+  const BrowserLayoutParams params = frame_view->GetBrowserLayoutParams();
+  return {
+      base::ClampCeil(params.leading_exclusion.ContentWithPadding().width()),
+      base::ClampCeil(params.trailing_exclusion.ContentWithPadding().width())};
```
sangwoo108, line 101: Clever 👍

sangwoo108 (approved):
LGTM 👍

browser/ui/views/toolbar/brave_toolbar_view.cc
simonhong: It's better :) Done.

PR #36135, opened 2026-05-04
Display runtime container in the context menu where applicable.

### LINKED ISSUE #55253: [Containers] Always display runtime containers in the context menu
When a tab is contained, but a container doesn't exist in the list of user-editable containers (a removed one, but still in use), we need to display it in the context menu the same way we display icon and accent color in the tab.

This is both UX improvement for the current impl and a precondition to add temporary containers.

goodov:
Display runtime container in the context menu in most scenarios.
Remove current container display from the context menu on bookmark click.

Resolves https://github.com/brave/brave-browser/issues/55253

### DIFF
--- a/browser/ui/containers/containers_bookmark_menu_model_delegate.cc
+++ b/browser/ui/containers/containers_bookmark_menu_model_delegate.cc
@@ -40,6 +40,8 @@ void ContainersBookmarkMenuModelDelegate::OnNoContainerSelected() {
 
 base::flat_set<std::string>
 ContainersBookmarkMenuModelDelegate::GetCurrentContainerIds() {
+  // The bookmark menu is not associated with any tab or link, so there are no
+  // current container IDs.
   return {};
 }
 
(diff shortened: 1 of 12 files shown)


browser/ui/containers/containers_bookmark_menu_model_delegate.cc
```diff
@@ -45,29 +40,7 @@ void ContainersBookmarkMenuModelDelegate::OnNoContainerSelected() {
...
-
-  const auto storage_partition_config = contents->GetSiteInstance()
-                                            ->GetSecurityPrincipal()
-                                            .GetStoragePartitionConfig();
-  if (!IsContainersStoragePartition(storage_partition_config)) {
-    return {};
-  }
-
-  return {storage_partition_config.partition_name()};
+  return {};
```
sangwoo108, line 45: Could you leave why we return empty set here?

sangwoo108 (approved):
LGTM 👍

browser/ui/containers/containers_bookmark_menu_model_delegate.cc
goodov: This context menu is not bound to a tab nor a link inside the tab. Having a currently selected container displayed in the menu doesn't make sense here.

goodov: added as a comment to code

PR #36139, opened 2026-05-04
Open new contained tab as "typed" to open as last tabs.

### LINKED ISSUE #55255: [Containers] New contained tab should be put last
Left click on a new tab button -> new tab is created as the last one no matter what current tab is selected.

Right click on a new tab button, select container -> new tab is created after a currently selected one, not as the last one.

goodov:
Chromium New tab action uses `PAGE_TRANSITION_TYPED`. We should use the same transition for scenarios when a new tab is created to create such tab at a correct position.

Resolves https://github.com/brave/brave-browser/issues/55255


sangwoo108 (approved):
LGTM 👍

PR #36142, opened 2026-05-04
[VPN 2.0] BoringTun 3p dependency: vendored crates

### LINKED ISSUE #54590: Vendor BoringTun 3p dependency

alekspop26:
This is a multi-part change to add BoringTun dependency to the brave-core, to be used later in the VPN 2.0 architecture by a privileged helper on desktop OSes. The dependency is added to the existing VPN to ensure it actually gets built with the browser, but not yet distributed.

Part 2: vendored crates that BoringTun depends on, plus some ignore rules for linters to make sure builds are not broken; following the similar ignores for wasm. There is also a Cargo.lock file with the list of the dependencies and versions.

All the vendored crates were created by running `cargo vendor`; the source files of the crates are unmodified, and **DO NOT NEED** any manual review.

Implements part of https://github.com/brave/brave-browser/issues/54590

Resolves


thypon (approved):
.semgrepignore fine

.gitignore
```diff
@@ -45,6 +45,7 @@ patches/**/*.patchinfo
 *.VC.VC.opendb
 .vs/
 .vscode/
+!third_party/boringtun/vendor/**/.vscode
```
bridiver: are these actually needed or did you just copy from wasm?

alekspop26: *.key and *.envrc files inclusion was actually needed, the rest I just copied from wasm to be on the safe side.
```
>find vendor -type f \( -name "*.vscode" -o -name "*.envrc" -o -name "*.key" -o -name "CMakeFiles.txt" \)

vendor/untrusted/mk/llvm-snapshot.gpg.key
vendor/sharded-slab/.envrc
```

alekspop26: Actually, good catch here.
I'll keep only those affecting the current vendor.
And I'm also going to modify the update script that checks if there are some files excluded by `.gitignore`, and prints the list of those files, so the exclusion rules can be revisited on demand.

cdesouza-chromium (approved):
LGTM

bridiver (approved):

PR #36152, opened 2026-05-04
[plaster][History Embeddings] Hide omnibox thumbs up/down and fix disclaimer trailing space

cdesouza-chromium:
Follow-up to #35813 / brave-browser#54798. The previous PR removed the
WebUI feedback buttons on chrome://history and the side panel, but the
native Views omnibox still surfaced thumbs up/down on HISTORY_EMBEDDINGS
matches when the user typed `@history`
(brave-browser#55178).

- Plaster `OmniboxPopupSelection::IsControlPresentOnMatch` to return
  false for FOCUSED_BUTTON_THUMBS_UP/DOWN. With the selection never
  available, OmniboxResultView::UpdateFeedbackButtonsVisibility keeps
  the buttons hidden, keyboard traversal skips them, and the
  feedback page can no longer be reached from the omnibox.

- Extend the existing featured_search_provider.cc plaster to drop the
  upstream `+ u\" \"` separator appended after the disclaimer text.
  With the link text already swapped to IDS_BRAVE_EMPTY_STRING, the
  separator left match.contents with a trailing space that tripped the
  AutocompleteMatch::SanitizeString DCHECK in
  AutocompleteResult::AppendMatches.


netzenbot (approved):

cdesouza-chromium:
This PR was created only for test to check why plaster was failing.

PR #36155, opened 2026-05-05
[toolchain] Rust build script under rebase tooling

cdesouza-chromium:
This change moves the script used to generate the rust WASM toolchain
under `tools/cr/toolchains`, as tooling packaging is a Chromium rebase
concern, and it is better kept under the rebase's team purview. A
separate PR has been submitted to fix CI.


mkarolin (approved):
++

PR #36156, opened 2026-05-05
[AI Chat]: Rename `OnNewGenerationLoop` to `UpdateToolsForNewGeneration` and accept a callback.

fallaciousreasoning:
Series https://github.com/brave/brave-browser/issues/55232

This allows tools to do asynchronous initialization, which will be a requirement for WebMCP (which needs to collect tools from AssociatedContent).

No functional changes in this PR.

### DIFF
--- a/components/ai_chat/core/browser/conversation_handler.cc
+++ b/components/ai_chat/core/browser/conversation_handler.cc
@@ -15,6 +15,7 @@
 #include <variant>
 #include <vector>
 
+#include "base/barrier_closure.h"
 #include "base/check.h"
 #include "base/check_op.h"
 #include "base/containers/fixed_flat_set.h"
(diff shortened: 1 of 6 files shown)


petemill (commented):
This seems fine architecturally but we probably need at least a test that verifies this async behavior. I think we have Mock Tool Providers for both unit and browser tests, so could be either.

netzenbot (approved):

components/ai_chat/core/browser/conversation_handler.cc
```diff
@@ -1185,15 +1186,20 @@ void ConversationHandler::AddToConversationHistory(
...
+
   // We can reset any already-created tools that don't want state to
   // survive between loops (i.e. when a new user message is received).
+  // Wait for all providers to finish before continuing.
+  auto barrier =
+      base::BarrierClosure(tool_providers_.size(), std::move(on_updated));
   for (auto& tool_provider : tool_providers_) {
-    tool_provider->OnNewGenerationLoop();
+    tool_provider->UpdateToolsForNewGenerationLoop(barrier);
   }
```
petemill, line 1202: I think this is fine because the tool providers are owned by ConversationHandler, and so will only get destroyed if this ConversationHandler is destroyed, so nothing will be dangling?

petemill (approved):

components/ai_chat/core/browser/tools/tool_provider.h
```diff
@@ -44,12 +45,12 @@ class ToolProvider {
   // should only maintain state within the tool loop of a single set of
   // responses. For example a TODO tool would only be applicable during 1 task,
   // but not a whole conversation.
-  virtual void OnNewGenerationLoop() {}
```
petemill, line 47: I could have sworn `OnNewGenerationLoop` was used somewhere! But maybe it's in a different branch.

fallaciousreasoning: I could've sworn I reviewed a PR using it :laughing:

components/ai_chat/core/browser/conversation_handler.cc
fallaciousreasoning: Yup, they're all owned in a vector of std::unique_ptr

PR #36157, opened 2026-05-05
[presubmit] Patches should use `--default-prefix` and similar opts

### LINKED ISSUE #55231: Add a presubmit to make sure patches are created with the expected index options
This presubmit check is important, as every now and then someone submits a manually generated patch, and tools like `plaster` and `git-cr-follow-renames` make assumptions about which flags are passed to `git diff` when creating patches.

cdesouza-chromium:
The patching machinery in Brave usually creates patches with the
equivalent of:

```
git -C .. diff --src-prefix=a/ --dst-prefix=b/ --default-prefix --full-index --ignore-space-at-eol browser/foo.cc
```

This is not enforced officially, and every now and then someone
generates patches manually and gets them merged with some other type of
header, which ends up causing issues when syncing, or using other types
of tooling that assume certain expectations from these patches. This PR
prevents that.

Another check being added is to validate that a patch added to this path
always have one file only in it, as this is an invariant of our patching
system.

Resolves https://github.com/brave/brave-browser/issues/55231

### DIFF
--- a/patches/PRESUBMIT.py
+++ b/patches/PRESUBMIT.py
@@ -5,14 +5,24 @@
 
 import chromium_presubmit_overrides
 import os
+import re
+
+# Full 40-char SHA hashes as produced by `git diff --full-index`.
+_FULL_INDEX_RE = re.compile(r'^index ([0-9a-f]{40})\.\.([0-9a-f]{40})( \d+)?$')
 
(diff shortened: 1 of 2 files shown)


AlexeyBarabash (approved):
++

emerick (approved):

PR #36168, opened 2026-05-05
Show an error message instead of an empty email aliases list in case of an error.

### LINKED ISSUE #46246: Email Aliases: Implement OnRefreshAliasesResponse error handling
### Platforms

Android, Linux, macOS, Windows

### Description

We need `OnAliasesUpdated(array<Alias> aliases, string? error_message)` to handle the case where the C++ code is unable to retrieve the list of aliases remotely. This could be due to the internet connection being disconnected, or possibly other errors.

boocmp:
Resolves https://github.com/brave/brave-browser/issues/46246

### DIFF
--- a/components/email_aliases/email_aliases_service.cc
+++ b/components/email_aliases/email_aliases_service.cc
@@ -89,6 +89,17 @@ auto MakeRequest(const std::string& bearer_token) {
   return request;
 }
 
+void NotifyObserversAliasesUpdated(
+    mojo::RemoteSet<mojom::EmailAliasesServiceObserver>& observers,
+    mojom::AliasesUpdatePtr update) {
+  if (observers.empty()) {
+    return;
+  }
+  for (auto& observer : observers) {
+    observer->OnAliasesUpdated(mojo::Clone(update));
+  }
+}
+
 }  // namespace
 
 EmailAliasesService::EmailAliasesService(
@@ -242,6 +253,10 @@ void EmailAliasesService::RefreshAliasesWithToken(TokenResult token) {
         url_loader_factory_, std::move(request),
         base::BindOnce(&EmailAliasesService::OnRefreshAliasesResponse,
                        weak_factory_.GetWeakPtr()));
+  } else {
+    NotifyObserversAliasesUpdated(
+        observers_, mojom::AliasesUpdate::NewError(l10n_util::GetStringUTF8(
+                        IDS_SETTINGS_EMAIL_ALIASES_AUTH_ERROR)));
   }
 }
 
@@ -322,34 +337,33 @@ void EmailAliasesService::DeleteAliasWithToken(const std::string& alias_email,
 
 void EmailAliasesService::OnRefreshAliasesResponse(
     endpoints::AliasList::Response response) {
-  // TODO(https://github.com/brave/brave-browser/issues/48959):
-  // In this function, when an error happens, we should show
-  // an error message to the user (requires design work)
   if (!response.body) {
-    LOG(ERROR) << "Email Aliases service error: No response body";
+    NotifyObserversAliasesUpdated(
+        observers_, mojom::AliasesUpdate::NewError(l10n_util::GetStringUTF8(
+                        IDS_EMAIL_ALIASES_SERVICE_ERROR_NO_RESPONSE_BODY)));
     return;
   }
   if (!response.body->has_value()) {
-    LOG(ERROR) << "Email Aliases service error: Invalid response format";
+    NotifyObserversAliasesUpdated(
+        observers_, mojom::AliasesUpdate::NewError(l10n_util::GetStringFUTF8(
+                        IDS_EMAIL_ALIASES_SERVICE_REPORTED_ERROR,
+                        base::UTF8ToUTF16(response.body->error().message))));
     return;
   }
 
   EmailAliasesNotes notes(pref_service_.get(), GetAuthEmail());
   notes.RemoveNotesForDeletedAliases(response.body.value()->result);
 
-  if (!observers_.empty()) {
-    std::vector<email_aliases::mojom::AliasPtr> aliases;
-    for (const auto& entry : response.body.value()->result) {
-      auto alias_obj = email_aliases::mojom::Alias::New();
-      alias_obj->email = entry.alias;
-      alias_obj->note = notes.GetNote(entry.alias);
-      aliases.push_back(std::move(alias_obj));
-    }
-
-    for (auto& observer : observers_) {
-      observer->OnAliasesUpdated(mojo::Clone(aliases));
-    }
+  std::vector<email_aliases::mojom::AliasPtr> aliases;
+  for (const auto& entry : response.body.value()->result) {
+    auto alias_obj = email_aliases::mojom::Alias::New();
+    alias_obj->email = entry.alias;
+    alias_obj->note = notes.GetNote(entry.alias);
+    aliases.push_back(std::move(alias_obj));
   }
+
+  NotifyObserversAliasesUpdated(
+      observers_, mojom::AliasesUpdate::NewAliases(std::move(aliases)));
 }
 
 }  // namespace email_aliases
--- a/components/email_aliases/email_aliases_service.cc
+++ b/components/email_aliases/email_aliases_service.cc
@@ -256,7 +256,7 @@ void EmailAliasesService::RefreshAliasesWithToken(TokenResult token) {
   } else {
     NotifyObserversAliasesUpdated(
         observers_, mojom::AliasesUpdate::NewError(l10n_util::GetStringUTF8(
-                        IDS_SETTINGS_EMAIL_ALIASES_AUTH_ERROR)));
+                        IDS_SETTINGS_EMAIL_ALIASES_INFO_ERROR_MESSAGE)));
   }
 }
 
@@ -340,7 +340,7 @@ void EmailAliasesService::OnRefreshAliasesResponse(
   if (!response.body) {
     NotifyObserversAliasesUpdated(
         observers_, mojom::AliasesUpdate::NewError(l10n_util::GetStringUTF8(
-                        IDS_EMAIL_ALIASES_SERVICE_ERROR_NO_RESPONSE_BODY)));
+                        IDS_SETTINGS_EMAIL_ALIASES_INFO_ERROR_MESSAGE)));
     return;
   }
   if (!response.body->has_value()) {
--- a/components/email_aliases/email_aliases_service.cc
+++ b/components/email_aliases/email_aliases_service.cc
@@ -92,9 +92,6 @@ auto MakeRequest(const std::string& bearer_token) {
 void NotifyObserversAliasesUpdated(
     mojo::RemoteSet<mojom::EmailAliasesServiceObserver>& observers,
     mojom::AliasesUpdatePtr update) {
-  if (observers.empty()) {
-    return;
-  }
   for (auto& observer : observers) {
     observer->OnAliasesUpdated(mojo::Clone(update));
   }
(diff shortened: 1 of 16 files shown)


netzenbot (approved):

components/email_aliases/resources/content/email_aliases_manage_page.tsx
```diff
@@ -35,22 +36,27 @@ const BraveAccountSignIn = () => {
 
 export const ManagePage = ({
   aliasesState,
+  aliasesError,
   authState,
   emailAliasesService,
 }: {
   aliasesState: Alias[]
+  aliasesError: string | null
```
fallaciousreasoning: potentially this could take an `AliasesUpdate`, as we only show:
1. An error
2. Aliases, if we're authenticated.

components/email_aliases/resources/content/email_aliases_manage_page.tsx
```diff
@@ -35,22 +36,27 @@ const BraveAccountSignIn = () => {
...
-    {authState.status === AuthenticationStatus.kAuthenticated && (
-      <AliasList
-        aliases={aliasesState}
-        authEmail={authState.email}
-        emailAliasesService={emailAliasesService}
-      />
-    )}
+    {authState.status === AuthenticationStatus.kAuthenticated
+      && (aliasesError ? (
+        <Alert type='error'>{aliasesError}</Alert>
```
fallaciousreasoning: this feels like we should be showing the error even if we aren't authenticated (i.e. auth failed).

components/email_aliases/resources/content/email_aliases_manage_page.tsx
fallaciousreasoning: Even if not the `AliasesUpdate` it does feel like it could take some kind of union

components/email_aliases/resources/content/use_email_aliases_observer.ts
```diff
@@ -0,0 +1,59 @@
...
+  Alias,
+  EmailAliasesServiceObserverInterface,
+} from 'gen/brave/components/email_aliases/email_aliases.mojom.m'
+
+/**
+ * Subscribes to EmailAliasesService observer callbacks and exposes auth state,
+ * alias list, and list-load error. Pass a stable `bindObserver` from the host
+ * (e.g. one created when the Mojo pipes are set up).
+ */
+export function useEmailAliasesObserver(
```
fallaciousreasoning: not sure if you've seen it or not (and not for this PR) but Pete Mill has designed this new API for talking to mojom and it makes stuff like this very easy:

https://github.com/brave/brave-core/blob/master/components/common/api/readme.md

You don't have to use it, but its worth having a look at, especially if you start any new pages.

components/email_aliases/resources/tests/email_aliases.test.tsx
```diff
@@ -389,4 +397,20 @@ describe('ManagePageConnected', () => {
     await authenticate(service, AuthenticationStatus.kAuthenticated)
     await expectAliasesNotVisible()
   })
+
+  it('shows error alert instead of alias list when refresh fails', async () => {
```
fallaciousreasoning, line 401: could we add some tests that:
1. It clears the error when unauthed
2. Doesn't update the error when no authed
3. Doesn't display errors/aliases when no auth

components/email_aliases/resources/content/use_email_aliases_observer.ts
fallaciousreasoning: nit: also, I think this should be called `useEmailAliases` as we're not using the observer externally - we just want to use the up to date email aliases/ error

fallaciousreasoning: might be worth adding some tests directly for this hook - fwiw, claude is quite good at generating them

components/email_aliases/resources/email_aliases.tsx
```diff
@@ -25,37 +23,13 @@ export const ManagePageConnected = ({
...
-        setAuthState(state)
-        if (status !== AuthenticationStatus.kAuthenticated) {
-          setAliasesState([])
-        }
-      },
-    }
-    return bindObserver(observer)
-  }, [])
+  const { authState, aliasesState, aliasesError } =
+    useEmailAliasesObserver(bindObserver)
```
fallaciousreasoning: 👍 much cleaner

components/email_aliases/resources/email_aliases_panel.tsx
```diff
@@ -34,30 +32,7 @@ export const EmailAliasesPanelConnected = ({
...
-        status = state.status
-        setAuthState(state)
-        if (status !== AuthenticationStatus.kAuthenticated) {
-          setAliasesState([])
-        }
-      },
-    }
-    return bindObserver(observer)
-  }, [])
+  const { authState, aliasesState } = useEmailAliasesObserver(bindObserver)
```
fallaciousreasoning: 👍

components/email_aliases/email_aliases_service.cc
```diff
@@ -87,6 +87,17 @@ auto MakeRequest(const std::string& bearer_token) {
...
 }
 
+void NotifyObserversAliasesUpdated(
+    mojo::RemoteSet<mojom::EmailAliasesServiceObserver>& observers,
+    mojom::AliasesUpdatePtr update) {
+  if (observers.empty()) {
+    return;
+  }
+  for (auto& observer : observers) {
+    observer->OnAliasesUpdated(mojo::Clone(update));
```
fallaciousreasoning, line 96: mega nit: I think `update->Clone()` is what I've normally seen

components/email_aliases/email_aliases_service.cc
```diff
@@ -87,6 +87,17 @@ auto MakeRequest(const std::string& bearer_token) {
   return request;
 }
 
+void NotifyObserversAliasesUpdated(
+    mojo::RemoteSet<mojom::EmailAliasesServiceObserver>& observers,
+    mojom::AliasesUpdatePtr update) {
+  if (observers.empty()) {
+    return;
+  }
```
fallaciousreasoning: any point in this check? We just won't iterate if there's nothing in the list

components/email_aliases/email_aliases_service.cc
```diff
@@ -239,6 +250,10 @@ void EmailAliasesService::RefreshAliasesWithToken(TokenResult token) {
         url_loader_factory_, std::move(request),
         base::BindOnce(&EmailAliasesService::OnRefreshAliasesResponse,
                        weak_factory_.GetWeakPtr()));
+  } else {
+    NotifyObserversAliasesUpdated(
```
fallaciousreasoning, line 254: I'm probably way off base here, but if we don't have a token, aren't we probably not logged in?

components/email_aliases/email_aliases_service_unittest.cc
```diff
@@ -49,9 +50,23 @@ class AliasObserver : public mojom::EmailAliasesServiceObserver {
  public:
   void OnAuthStateChanged(mojom::AuthStatePtr) override {}
 
-  void OnAliasesUpdated(std::vector<mojom::AliasPtr> aliases) override {
+  void OnAliasesUpdated(mojom::AliasesUpdatePtr update) override {
     ++alias_updates;
-    last_aliases = std::move(aliases);
+    if (!update) {
```
fallaciousreasoning: I think rather than doing this, we could just store `mojom::AliasesUpdatePtr last_update_` on the class. Then we wouldn't need to worry about resetting whichever the other is, and we could dp `last_update_ = std::move(update);`

components/email_aliases/email_aliases_service_unittest.cc
```diff
@@ -70,9 +85,20 @@ class AliasObserver : public mojom::EmailAliasesServiceObserver {
     return last_aliases;
   }
 
+  const std::optional<std::string>& get_last_error() const {
+    return last_error;
```
fallaciousreasoning: this could access the update field, if we need both of these accessors - might be neater to expose a `const mojom::AliasUpdatePtr& last_update() const { return last_update_; }` though

fallaciousreasoning (commented):
mostly lgtm with some mostly nitty comments

components/email_aliases/resources/content/email_aliases_manage_page.tsx
boocmp: I plan to show auth errors in the follow up. I firstly have to switch to the brave account state instead of custom email aliases status

components/email_aliases/resources/content/use_email_aliases_observer.ts
boocmp: Yep, I want to switch to the new API in the future.

components/email_aliases/email_aliases_service.cc
boocmp: Yep, this is a generic clone. At this place it doesn't matter what to use, but generic clone also can clone typemapped mojo types (and any copyable types), so it is the way to clone even if type doesn't support Clone method.

components/email_aliases/email_aliases_service.cc
boocmp: There are two tokens: the BraveAccount token (logged-in, logged-out, etc.) and a specific short-lived email aliases token used to make calls to the EA backend. One possible reason for not having this token is being in a logged-out state.

components/email_aliases/email_aliases_service.cc
boocmp: Copy-pasted and missed. Removed.

components/email_aliases/resources/tests/email_aliases.test.tsx
boocmp: Added

components/email_aliases/resources/content/email_aliases_manage_page.tsx
boocmp: Changed to use AliasesUpdate

components/email_aliases/resources/content/email_aliases_manage_page.tsx
```diff
@@ -35,25 +35,23 @@ const BraveAccountSignIn = () => {
...
-        <Alert type='error'>{aliasesError}</Alert>
+      && (aliasesUpdate.error ? (
+        <Alert type='error'>{aliasesUpdate.error}</Alert>
       ) : (
         <AliasList
-          aliases={aliasesState}
+          aliases={aliasesUpdate.aliases!}
           authEmail={authState.email}
           emailAliasesService={emailAliasesService}
         />
```
fallaciousreasoning, line 57: optional, but possibly nicer:

```suggestion
{authState.status === AuthenticationStatus.kAuthenticated && <>
  {aliasesUpdate.error &&  <Alert type='error'>{aliasesUpdate.error}</Alert>}
  {aliasesUpdate.aliases && <AliasList
          aliases={aliasesUpdate.aliases!}
          authEmail={authState.email}
          emailAliasesService={emailAliasesService}
        />}
</>}

File: components/email_aliases/resources/content/use_email_aliases.ts [hunk 13]
```diff
@@ -8,24 +8,29 @@ import {
...
   AuthenticationStatus,
   AuthState,
-  Alias,
   EmailAliasesServiceObserverInterface,
 } from 'gen/brave/components/email_aliases/email_aliases.mojom.m'
 
+const emptyAliasesUpdate = (): AliasesUpdate => ({
+  aliases: [],
+  error: undefined,
+})
```
fallaciousreasoning, line 17: nit: This can just be a constant

```suggestion
const emptyAliasesUpdate: : AliasesUpdate = {
  aliases: [],
  error: undefined,
}
```

fallaciousreasoning (approved):
lgtm % some minor nits - nice cleanup extracting that logic to the `useEmailAliases` hook

mkarolin (approved):
`strings`++

PR #36188, opened 2026-05-05
Fix chrome url on restored tabs

### LINKED ISSUE #53821: `chrome://<host>` is displayed on restored brave://<host> tabs when they are not loaded yet
https://github.com/user-attachments/assets/1519c949-f29d-43af-87b1-fd2939d81e0d

bridiver:
## Summary
- Override `NavigationEntryImpl::GetTitleForDisplay()` to convert `chrome://` to `brave://` in the display title fallback for restored but unloaded tabs

Resolves https://github.com/brave/brave-browser/issues/53821

## Root Cause
Brave uses `chrome://` URLs internally and converts to `brave://` in the UI layer (location bar, hover cards, etc.). When a restored tab hasn't loaded yet and has no saved page title, `GetTitleForDisplay()` falls back to formatting the virtual URL. Since the virtual URL is `chrome://settings` (not `brave://settings`), the tab displays "chrome://settings" to the user.

## Fix
Override `GetTitleForDisplay()` in Brave's existing `NavigationEntryImpl` chromium_src override. When the real title is empty (fallback path), check if the cached display title starts with `chrome://` and replace with `brave://`. This catches all display paths (tab strip, window title, back/forward menu, task manager, etc.) since they all go through `GetTitleForDisplay()`.

### DIFF
--- a/chromium_src/content/browser/renderer_host/navigation_entry_impl_unittest.cc
+++ b/chromium_src/content/browser/renderer_host/navigation_entry_impl_unittest.cc
@@ -63,8 +63,8 @@ TEST_F(BraveNavigationEntryTest,
   entry = CreateEntry(GURL("http://chrome.com"));
   EXPECT_EQ(u"chrome.com", entry->GetTitleForDisplay());
 
-  entry = CreateEntry(GURL("http://example.com?chrome://settings"));
-  EXPECT_EQ(u"example.com?chrome://settings", entry->GetTitleForDisplay());
+  entry = CreateEntry(GURL("http://example.com/?chrome://settings"));
+  EXPECT_EQ(u"example.com/?chrome://settings", entry->GetTitleForDisplay());
 }
 
 }  // namespace content
--- a/chromium_src/content/browser/renderer_host/navigation_entry_impl_unittest.cc
+++ b/chromium_src/content/browser/renderer_host/navigation_entry_impl_unittest.cc
@@ -1,0 +1,54 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "content/public/browser/navigation_controller.h"
+#include "content/public/browser/navigation_entry.h"
+#include "content/public/test/browser_task_environment.h"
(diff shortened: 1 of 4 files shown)


chromium_src/content/browser/renderer_host/navigation_entry_impl_unittest.cc
```diff
@@ -0,0 +1,67 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
```
netzenbot, line 1: The upstream `navigation_entry_impl_unittest.cc` already has tests for `GetTitleForDisplay` (lines 93–159) that should run against this override. Add `#include <content/browser/renderer_host/navigation_entry_impl_unittest.cc>` at the top and move the Brave-specific tests after it, following the pattern in `chromium_src/content/browser/renderer_host/renderer_sandboxed_process_launcher_delegate_unittest.cc`. See [TI-040](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-040).

chromium_src/content/browser/renderer_host/navigation_entry_impl_unittest.cc
```diff
@@ -0,0 +1,67 @@
...
+  EXPECT_EQ(u"chrome-untrusted://settings", entry->GetTitleForDisplay());
+
+  entry = CreateEntry(GURL("http://chrome.com"));
+  EXPECT_EQ(u"chrome.com", entry->GetTitleForDisplay());
+
+  entry = CreateEntry(GURL("http://example.com?chrome://settings"));
+  EXPECT_EQ(u"example.com", entry->GetTitleForDisplay());
+}
+
+}  // namespace content
```
netzenbot, line 70: nit: `shared_url_loader_factory.h` appears unused here — `navigation_controller.h` already forward-declares `network::SharedURLLoaderFactory`, and passing `nullptr` for a `scoped_refptr<T>` parameter doesn't require the full type definition. https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-001

goodov (approved):

chromium_src/content/browser/renderer_host/navigation_entry_impl_unittest.cc
bridiver: This is not a valid best practice as it would result in either only running with upstream tests or including private deps (//content/browser)

netzenbot (approved):

PR #36209, opened 2026-05-06
[git-cr] Rewrite `gn` references of moved targets

### LINKED ISSUE #55297: `git-cr-mv` and `git-cr-follow-renames` should rewrite `gn` references
We should be able to get a full build just doing:

```sh
git cr mv components/api_request_helper/ components/api_foo
npm run build -- --target=brave:all
```

cdesouza-chromium:
This PR introduces support for `cr mv` and `cr follow-renames` to
correct `gn` references across the projects when a `BUILD.gn` path
changes. This is a modest introduction to references path correction.
The rewrite does handle relative references though. In source references
are only corrected for the main target, i.e `:basename`, which gets
corrected to the new path base name.

This new approach also attempts to correct gn references to moved
sources, but only for `//` root reference paths. In the future we could
potentially introduce some mechanism handles file renames.

The baseline for this feature was to get the following to build with no
errors:

```
git cr mv components/api_request_helper/ components/api_foo
npm run build -- --target=brave:all
```

As a small detour, this PR adds `npm run format` to these commands, when
wrapping up.

Resolves https://github.com/brave/brave-browser/issues/55297

### DIFF
--- a/tools/cr/alias/README.md
+++ b/tools/cr/alias/README.md
@@ -1,0 +1,196 @@
+# `git cr` alias for Chromium rebasing tooling
+
+This alias is maintained by the Chromium Rebase team to simplify many of the
+rebase tasks. This directory provides the `git cr` git alias and the supporting
+commit-msg hook used when committing to `brave-core`.
+
+## Quick start
+
+From the `brave-core` repository root, install this alias with:
+
+```sh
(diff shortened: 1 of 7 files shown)


AlexeyBarabash (approved):
LGTM!

PR #36218, opened 2026-05-06
[Polkadot] [Wallet] add asset_tx_payment field to chain metadata

### LINKED ISSUE #55307: [Polkadot] [Wallet] AssetHub requires extra signing fields
AssetHub is unique in that it requires an `asset_id` passed after the `tip` in order for an extrinsic to be well-formed.

This is because AssetHub permits users to pay transaction fees in a custom asset. For our needs, we will default to None for the AssetId which becomes `0x00` when SCALE-encoded.

This is required to be part of the signature payload when assembling the extrinsic body and also for generating the signature as well.

cmazakas:
Resolves https://github.com/brave/brave-browser/issues/55307

Update chain metadata parsing to handle probing for extrinsic extensions, which tells us if we need to supply an asset_id during the extrinsic signing process.

Note, adding extra/new fields to the Polkadot chain metadata doesn't require manual migrations. If new fields are added then the initial parse from the Prefs will fail which causes the metadata to be re-fetched which will then build the new metadata properly and cache it, so there should be no issues in this regard.

In Polkadot, extrinsics may require specific signing extensions. This code updates the metadata parsing routines to parse this data from the metadata blobs we get from the remote and then persist the new field to the prefs.

Follow-up PR will update the extrinsic signing code to use this new field.


components/brave_wallet/browser/polkadot/polkadot_chain_metadata.h
```diff
@@ -71,6 +72,10 @@ class PolkadotChainMetadata {
   uint16_t GetSs58Prefix() const;
   uint32_t GetSpecVersion() const;
```
netzenbot, line 74: nit: CS-005 says to avoid `Get*` for bool queries. `GetAssetTxPayment()` returns `bool` — consider `HasAssetTxPayment()` or `UsesAssetTxPayment()` to signal it's a boolean predicate rather than a value retrieval. ([✅ Use Conventional Method Prefixes](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-005))

components/brave_wallet/browser/polkadot/polkadot_chain_metadata_prefs_unittest.cc
```diff
@@ -126,6 +162,7 @@ TEST_F(PolkadotChainMetadataPrefsUnitTest, MissingRequiredFieldRejected) {
     value.Set(kTransferAllowDeathCallIndex, 1);
```
netzenbot, line 165: nit: `MissingRequiredFieldRejected` tests missing `spec_version` but has no case for missing `asset_tx_payment`. Since `ReadBool` returns `std::nullopt` when the field is absent, a test case omitting `kAssetTxPayment` would verify the new required field is actually enforced by the deserialization path. ([Unit Tests Required for Bad/Invalid Input](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-016))

components/brave_wallet/browser/polkadot/polkadot_chain_metadata_prefs_unittest.cc
```diff
@@ -61,6 +62,39 @@ TEST_F(PolkadotChainMetadataPrefsUnitTest, SetAndGetChainMetadataRoundTrip) {
   EXPECT_EQ(loaded->GetTransferAllCallIndex(), 5u);
```
netzenbot: nit: `SetAndGetChainMetadataRoundTrip_AssetHub` is nearly identical to `SetAndGetChainMetadataRoundTrip` — both do a full set/get round-trip and verify all the same fields, differing only in `asset_tx_payment`. Consider collapsing these into a single parameterized test (e.g., with a `TestCase` struct holding the two varying inputs and a `SCOPED_TRACE`) to avoid duplicating all the assertion lines. ([✅ Use Parameterized Tests for Similar Scenarios](https://github.com/brave/brave-core/blob/master/docs/testing-isolation.md#TI-023))

netzenbot (approved):

cypt4 (approved):

antonok-edm (approved):

PR #36220, opened 2026-05-06
[iOS] Display toast on Origin settings after changing a feature state

### LINKED ISSUE #55313: Show toast when toggling features in Origin settings
Add a toast that displays when a user enables or disables a feature from the Origin settings panel telling them that they must restart the app for the settings to work correctly

Figma: https://www.figma.com/design/IwYu4Yjt3wYFG59zcwWSGu/%F0%9F%A6%81-Brave-Origin?node-id=2419-24131&t=4XrRSgMMuvtk0LzZ-4

kylehickinson:
This adds a new toast that displays until the user dismisses it/swipes it away that explains to users that the app needs to be restarted in order to apply Origin changes

Resolves https://github.com/brave/brave-browser/issues/55313

### DIFF
--- a/ios/brave-ios/Sources/Origin/OriginSettingsView.swift
+++ b/ios/brave-ios/Sources/Origin/OriginSettingsView.swift
@@ -114,24 +114,23 @@ public struct OriginSettingsView: View {
     .navigationBarTitleDisplayMode(.inline)
     .scrollContentBackground(.hidden)
     .background(Color(.braveGroupedBackground))
-    .safeAreaInset(edge: .bottom) {
+    .overlay(alignment: .bottom) {
       if viewModel.isRestartToastVisible {
         FeatureEnabledToastView {
-          viewModel.isRestartToastVisible = false
+          withAnimation(.toast) {
+            viewModel.isRestartToastVisible = false
+          }
         }
-        .transition(.move(edge: .bottom).combined(with: .opacity))
+        .transition(.move(edge: .bottom).combined(with: .blurReplace))
       }
     }
-    .animation(
-      .spring(response: 0.3, dampingFraction: 0.725),
-      value: viewModel.isRestartToastVisible
-    )
   }
 
   private struct FeatureEnabledToastView: View {
     var dismiss: () -> Void
     @State private var dragOffset: CGFloat = 0
     @State private var height: CGFloat = 0
+    @AccessibilityFocusState(for: .voiceOver) private var isAccessibilityFocused: Bool
 
     var body: some View {
       HStack(alignment: .firstTextBaseline) {
@@ -177,6 +176,16 @@ public struct OriginSettingsView: View {
           }
       )
       .padding(8)
+      .dynamicTypeSize(.xSmall..<DynamicTypeSize.accessibility2)
+      .accessibilityElement(children: .combine)
+      .accessibilityAddTraits(.isModal)
+      .accessibilityFocused($isAccessibilityFocused)
+      .accessibilityAction {
+        dismiss()
+      }
+      .onAppear {
+        isAccessibilityFocused = true
+      }
     }
   }
 }
@@ -213,6 +222,12 @@ extension Bool {
   fileprivate var inversed: Bool {
     get { !self }
     set { self = !newValue }
+  }
+}
+
+extension Animation {
+  static var toast: Animation {
+    .spring(response: 0.3, dampingFraction: 0.725)
   }
 }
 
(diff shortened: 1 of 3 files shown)


netzenbot (approved):

ios/brave-ios/Sources/Origin/OriginSettingsView.swift
```diff
@@ -114,6 +114,70 @@ public struct OriginSettingsView: View {
...
+        FeatureEnabledToastView {
+          viewModel.isRestartToastVisible = false
+        }
+        .transition(.move(edge: .bottom).combined(with: .opacity))
+      }
+    }
+    .animation(
+      .spring(response: 0.3, dampingFraction: 0.725),
+      value: viewModel.isRestartToastVisible
+    )
```
nuo-xu: this animation should only apply to the toast content. i think adding this to the entire form will cause weird effects since the safe area inset growing/shrinking with a scroll view

ios/brave-ios/Sources/Origin/OriginSettingsView.swift
```diff
@@ -114,6 +114,70 @@ public struct OriginSettingsView: View {
...
+  }
```
EliHini: I think this should be scoped to toast view itself either `withAnimation` block around the state change, or `.animation(_:value:)` on the `FeatureEnabledToastView` rather than the parent

ios/brave-ios/Sources/Origin/OriginSettingsView.swift
```diff
@@ -114,6 +114,70 @@ public struct OriginSettingsView: View {
     .navigationBarTitleDisplayMode(.inline)
     .scrollContentBackground(.hidden)
     .background(Color(.braveGroupedBackground))
+    .safeAreaInset(edge: .bottom) {
+      if viewModel.isRestartToastVisible {
+        FeatureEnabledToastView {
+          viewModel.isRestartToastVisible = false
+        }
+        .transition(.move(edge: .bottom).combined(with: .opacity))
```
EliHini: Consider adding `.accessibilityAddTraits(.isModal)` or an announcement so screen reader users know a notification appeared. 
Maybe we already have accessibility done somewhere automatically

EliHini (changes requested):

ios/brave-ios/Sources/Origin/OriginSettingsView.swift
kylehickinson: I've tried it both and it the little flicker still occurs, just a bug in SwiftUI + `safeAreaInset` API. The other option is using an `overlay` instead of `safeAreaInset` but then toast will occlude the list when scrolling. And the newer scoped animations don't work with transitions like this.  If I was to swap them it would basically just use `withAnimation` in 2 spots instead of this

ios/brave-ios/Sources/Origin/OriginSettingsView.swift
kylehickinson: https://github.com/brave/brave-core/pull/36220#discussion_r3204406566

ios/brave-ios/Sources/Origin/OriginSettingsView.swift
kylehickinson: Good call, added that trait along with focusing it immediately here: https://github.com/brave/brave-core/pull/36220/commits/5582e6d17b414b4577bc26b8ae4da3ac4902b430

ios/brave-ios/Sources/Origin/OriginSettingsView.swift
kylehickinson: Fixed in https://github.com/brave/brave-core/pull/36220/commits/5582e6d17b414b4577bc26b8ae4da3ac4902b430

nuo-xu (approved):
code lgtm. just core wont let me toggle :(

EliHini (approved):

kjozwiak:

```
Brave | 1.92.46 Chromium: 148.0.7778.96 (Official Build)nightly (64-bit)
--- | ---
Revision | 1c36cb7869a0...
OS : iOS`
```

Using the STR/Cases and design outlined via https://github.com/brave/brave-browser/issues/55313#issue-4393735290, ensured that we're displaying a `Restarted needed` toast via the `Brave Origin` settings whenever either enabling/disabling settings as per the following:

* ensured that the toast appears whenever you're either enabling/disabling a setting within `Brave Origin` settings
* ensured that tapping on `X` closes the toast without any issues/failures
* ensured that the toast is dismissed if you close the applications while it's opened
* ensured that swiping down on the toast closes the message/notification without any issues/failures

IMG_1620

PR #36222, opened 2026-05-06
[query-filter] Enable query filter component integration

### LINKED ISSUE #55305: Expose the downloaded rules to query filter clients
We now have support for downloading query filter rules from remote to the client via component updater. The rules are stored inside [QueryFilterData](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/components/query_filter/browser/query_filter_data.h?L58).  

We should expose the rules vector inside [utils.cc](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/components/query_filter/browser/utils.cc?L203) and update the stripping logic to consume these set of rules instead of the hard-coded ones. The new stripping logic should be behind the [QueryFilterComponent](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/components/query_filter/common/features.h?L14) feature flag.

r0hit0303:
This change updates the existing client to now allow stripping the query parameters with the newly added query filter component. 

This change also enables the `QueryFilterComponent` flag **enabled** by default and removes all references to the old code. This means that turning off the flag would also turn off the whole query filter feature explicitly as discussed [here](https://github.com/brave/brave-core/pull/36222#discussion_r3197754497). 

Manual testing [video](https://github.com/user-attachments/assets/22df2285-a077-4881-be96-22a38d29e180).

Resolves https://github.com/brave/brave-browser/issues/55305


components/query_filter/browser/utils_unittest.cc
```diff
@@ -107,9 +291,36 @@ TEST(BraveQueryFilter, FilterQueryTrackers) {
...
+  // Conditional filtering, test nothing filtered. `h_slt` param with `/email/`
+  // present in url.
+  EXPECT_FALSE(query_filter::MaybeApplyQueryStringFilter(
+      GURL(), GURL("https://brave.com"),
+      GURL("https://test.com/email/?h_slt=123"), "GET", false));
+  // Conditional filtering, test nothing filtered. `mkt_tok` param with
+  // `Unsubscribe` present in url.
+  EXPECT_FALSE(query_filter::MaybeApplyQueryStringFilter(
+      GURL(), GURL("https://brave.com"),
+      GURL("https://Unsubscribe.com/?mkt_tok=123"), "GET", false));
```
r0hit0303, line 292: @fmarier `GURL("https://Unsubscribe.com/?mkt_tok=123"), "GET", false));` 
I think in this case we should have removed the `mkt_tok` parameter no? If so should we modify the regex [here](https://sourcegraph.com/r/github.com/brave/brave-core@master/-/blob/components/query_filter/browser/utils.cc?L143) to enclose it with "//" or something?

components/query_filter/browser/utils.cc
```diff
@@ -152,6 +161,8 @@ static constexpr auto kConditionalQueryStringTrackers =
 // `-Wexit-time-destructors` violation and `[[clang::no_destroy]]` has been
 // added in the meantime to fix the build error. Remove this attribute and
 // provide a proper fix.
+// TODO(https://github.com/brave/brave-browser/issues/10188): Remove this
```
fmarier: Let's migrate in this PR. No need to be overly cautious about this transition. We can just have a kill switch to turn the query filter OFF if it's broken. Instead of reverting to the old implementation, we can simply expedite a fix to the new one.

fmarier:
Make sure to test this new implementation against https://fmarier.github.io/brave-testing/query-filter.html

components/query_filter/browser/utils_unittest.cc
fmarier: I think it's fine to keep the parameter in this case. That's not how it's used to track users in the wild.

components/query_filter/browser/utils.cc
r0hit0303: > Instead of reverting to the old implementation, we can simply expedite a fix to the new one.

Wouldn't we end up in situations where we turn "on" the query filter flag again globally, say after we have landed a fix, but the older clients would continue to break as we have just turned "on" the query filter flag back again but they haven't yet updated Brave. In this case they would not even have the original hardcoded fallback and query filtering will continue to break.

I feel we should either a) continue with the ff OR b) get rid of any flag completely so we don't end up in situations like above. wdyt?

fmarier: I believe we have a `min_version` in griffin for that purpose: https://bravesoftware.slack.com/archives/C071UDBFFT6/p1778107480600319?thread_ts=1778086724.617479&cid=C071UDBFFT6

components/query_filter/browser/schema_utils_unittest.cc
```diff
@@ -0,0 +1,171 @@
+// Copyright (c) 2026 The Brave Authors. All rights reserved.
```
r0hit0303, line 2: This test file is completely LLM generated. I went through the tests, removed a few but otherwise everything lgtm.

components/query_filter/browser/utils.cc
```diff
@@ -182,69 +47,79 @@ static constexpr auto kExemptedHostnames =
...
-  }
-
-  return false;
-}
-
 // Remove tracking query parameters from a GURL, leaving all
 // other parts untouched.
 std::optional<std::string> StripQueryParameter(std::string_view query,
                                                std::string_view spec) {
+  // Note that when disabled this would also remove support for
```
r0hit0303: @fmarier, I just wanted to highlight this subtle difference here. Removing the conditional tracking query parameter is independent from removing scoped and simple query params, and since technically our component only supports the latter I could have plugged it in the flow. However, given that we would remove most of the query filter functionality if the flag is disabled, it only made sense to me to also decouple removing conditional trackers functionality. 
I've also added a test to check we have indeed decoupled removing conditional query trackers in the utils_unittest.cc file in BraveQueryFilter_ComponentDisabled#NoStrippingOccurs.

components/query_filter/browser/utils.cc
```diff
@@ -182,69 +47,79 @@ static constexpr auto kExemptedHostnames =
...
+  // |kConditionalQueryStringTrackers|.
+  if (!base::FeatureList::IsEnabled(
+          query_filter::features::kQueryFilterComponent)) {
+    return std::nullopt;
+  }
+
+  // Set of all the params which are blocked for |spec| from our rules.
+  const base::flat_set<std::string> blocked_params_set =
```
r0hit0303: @fmarier  @cdesouza-chromium 

We could have relied on `absl::flat_hash_set` here since we don't require the ordering but that would have meant adding the whole absl dependency to our build target (which is also a dep to other clients). So, I went with the slightly less-efficient "ordered" O(log(N)) container like `base::flat_set` which is already part of our target to avoid inflating it further. However, given the small # of query parameters and reliance on a flat structure (which is cache friendly), I don't suspect any visible performance issues.

components/query_filter/browser/utils.cc
```diff
@@ -182,69 +47,79 @@ static constexpr auto kExemptedHostnames =
...
-      ++disallowed_count;
-    } else {
-      output_kv_strings.push_back(kv_string);
+  std::vector<std::string_view> output_tokens;
+  bool did_strip = false;
+
+  // Iterate over all tokens sequentially maintaining their order in the URL.
+  for (const auto token : tokens) {
+    // Try to parse the token as param=value.
+    const auto param_value = base::SplitStringOnce(token, "=");
```
r0hit0303: The previous code relied on `SplitStringPiece` which I thought didn't make sense in this context where we are only interested in separating lhs and rhs around a *single* "=". So, I've switched to `SplitStringOnce` which seemed more appropriate.

r0hit0303:
@cdesouza-chromium @fmarier I'll break this change into smaller parts and send the individual reviews your way. I've left a few comments in this change so PTAL and lmk if things look good overall.

r0hit0303:
@fmarier and @cdesouza-chromium could you PTAL at this change? I think everything looks in order. This is the main change that would enable the query filter component. Thanks!

netzenbot (approved):

components/query_filter/browser/utils.cc
fmarier: It's fine to disable all query string filtering based on that features. However, let's move that check to `MaybeApplyQueryStringFilter()` where the other `return std::nullopt` live.

components/query_filter/browser/utils.cc
```diff
@@ -182,69 +48,79 @@ static constexpr auto kExemptedHostnames =
...
+
   // We are using custom query string parsing code here. See
   // https://github.com/brave/brave-core/pull/13726#discussion_r897712350
   // for more information on why this approach was selected.
   //
   // Split query string by ampersands, remove tracking parameters,
   // then join the remaining query parameters, untouched, back into
   // a single query string.
-  const std::vector<std::string_view> input_kv_strings =
+  const std::vector<std::string_view> tokens =
```
fmarier: Maybe we should avoid renaming these two variables so that the diff is clearer. This `tokens` variable is storing the same info as before, no? Similar to `key` v. `param`.

I'm not opposed to changing the names, but maybe we can do that in a follow-up so that it's more obvious that this PR doesn't change the logic at all.

components/query_filter/browser/utils.cc
fmarier: How do the two compare when it comes to parameters like `foo==1234`? And does it preserve whitespace like the old code?

components/query_filter/browser/utils.cc
```diff
@@ -182,69 +48,79 @@ static constexpr auto kExemptedHostnames =
...
       SplitStringPiece(query, "&", base::KEEP_WHITESPACE, base::SPLIT_WANT_ALL);
-  std::vector<std::string_view> output_kv_strings;
-  int disallowed_count = 0;
```
fmarier, line 227: Same here. Let's stick to `disallowed_count` which I agree is unnecessary (a bool is better) and then follow up with a simple change  right after this PR.

components/query_filter/browser/utils.cc
```diff
@@ -182,69 +48,79 @@ static constexpr auto kExemptedHostnames =
...
-  for (const auto& kv_string : input_kv_strings) {
-    const std::vector<std::string_view> pieces = SplitStringPiece(
-        kv_string, "=", base::KEEP_WHITESPACE, base::SPLIT_WANT_NONEMPTY);
-    const std::string_view key = pieces.empty() ? "" : pieces[0];
-    if (pieces.size() >= 2 &&
```
fmarier, line 232: It's very nice to be able to move away from this messy if statement :)

components/query_filter/browser/utils.cc
fmarier: It may be worth adding a test case for a double equal sign like this if we don't already check for this corner case.

components/query_filter/browser/utils.cc
r0hit0303: Your suggestion to follow-up on improving the code in the next change sgtm. 

> This tokens variable is storing the same info as before, no? Similar to key v. param. 

My reason to update the variable `|input_kv_strings|` to `|tokens|` was because I don't think it holds a kv string all the time. We try to to split the main string into chunks where the separator is "&" and assume the chunks are of of form `kv` which may not be necessarily true. 
For example: in this URL "example.com?q=a&aa", we would have the chunks "q=a" and "aa". That's why changing the name to `tokens` felt more right.

components/query_filter/browser/utils.cc
r0hit0303: > does it preserve whitespace like the old code? 

From the implementation of base::SplitStringOnce it does seem to preserve the whitespace [1](https://source.chromium.org/chromium/chromium/src/+/main:base/strings/string_split.cc;l=27?q=string_split.cc).  

> How do the two compare when it comes to parameters like foo==1234?

We would strip the query parameter in this case which matches the existing behaviour. I've filed an issue to track that separately [here](https://github.com/brave/brave-browser/issues/55593).

> It may be worth adding a test case for a double equal sign like this if we don't already check for this corner case.

I didn't see any test for this corner case. I'll add this once we resolve the [issue]( https://github.com/brave/brave-browser/issues/55593) as I feel we shouldn't remove the param but want to double check with you. Thanks!

r0hit0303:
**Update**: Going with minimal refactor approach after discussing offline with fmarier@. Will follow-up with the  refactoring of `StripQueryParameter` method in 1 week (>= May 28th) targeting the next release.  

---
Hi François, thank you for the review. 

I'm actually a bit confused on how to proceed with your request for renaming variables at [1](https://github.com/brave/brave-core/pull/36222/changes#r3229770742) and [2](https://github.com/brave/brave-core/pull/36222/changes#r3229667125) as a future change. 

The reason being, I feel renaming these variables is a much smaller change that I introduced when compared with [3](https://github.com/brave/brave-core/pull/36222/changes#r3229774518) and if you are ok with [3] and then surely [1] and [2] should also be fine given that you already agree that they make the code more comprehensible without altering the behaviour.  Moreover, both the existing and the new tests (+manual test) passes on this change so I'm wondering if you could maybe give a second thought on whether the change in [1] and [2] is really needed in the context of this PR.  

Alternatively, I can try and ditch [3] and then add the new rules based parsing [here](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/components/query_filter/browser/utils.cc?L232) 

```
 if (pieces.size() >= 2 &&
        (**ShouldRemoveParam(key, spec)** ||
         (kConditionalQueryStringTrackers.count(key) == 1 &&
          !re2::RE2::PartialMatch(
              spec, kConditionalQueryStringTrackers.at(key).data())))) {
```

to keep the changes minimal and then follow up with the refactoring that the current PR is proposing, but I really don't see much benefit in doing it this way either, if we both already are on the same page that the current refactoring is fine (given [1], [2], and [3]) and it doesn't change the existing logic.. wdyt?

components/query_filter/browser/utils.cc
r0hit0303: Good suggestion. That makes sense to me.

components/query_filter/browser/utils.cc
r0hit0303: Please un-resolve this if needed. Thanks!

components/query_filter/browser/utils.cc
r0hit0303: Thank you for resolving the [issue](https://github.com/brave/brave-browser/issues/55593). I've added tests in the `brave_site_hacks_network_delegate_helper_unittest.cc` to check we remove the param for "fbclid==x" and keep for "fbclid==".

r0hit0303:
@fmarier could you ptal at the latest change? I've removed the refactoring in the latest change and will follow-up in the next change. Thanks!

fmarier (approved):

cdesouza-chromium (approved):
LGTM

r0hit0303:
Thank you for the review

PR #36225, opened 2026-05-06
[iOS] fix reader mode button disappears after coming back from an reader mode active page

### LINKED ISSUE #55309: [iOS] ReaderMode button disappears after tapping back button from ReaderMode active page
### Description

ReaderMode button disappears after tapping back button from ReaderMode active page

https://github.com/user-attachments/assets/d401c605-0a4c-4385-9824-ecda48d8e138

### Steps to reproduce

1. Enable single web view configuration per profile flag in brave://flags
2. visit a page that has reader mode available (e.x wiki page)
3. tap reader mode button in omnibox 
4. verify page is reloaded and reader mode is activated 
5. tap back button

### Actual result

navigate to previous page with no reader mode button in omnibox 

### Expected result

navigate to previous page with with reader mode button not activated in omnibox 

### Reproduces how often

Intermittent issue

### Brave version

nightly

### Device/iOS version

all

### Affected browser versions

- [ ] latest AppStore
- [ ] latest TestFlight
- [ ] previous TestFlight

### Reproducibility

### Miscellaneous information

_No response_

nuo-xu:
When using back/forward button to navigate page, WKWebView would using caches to restore pages which will make most CWV navigation callbacks very early/quick while the JavaScript execution context is reinitialized asynchronously. 
This will make the reader mode readability check returns wrong status since web main frame is not yet ready. 
The solution is to add another readability check when the main frame did become available. 

Resolves https://github.com/brave/brave-browser/issues/55309

### DIFF
--- a/ios/brave-ios/Sources/Brave/Frontend/Browser/BrowserViewController/BVC+TabObserver.swift
+++ b/ios/brave-ios/Sources/Brave/Frontend/Browser/BrowserViewController/BVC+TabObserver.swift
@@ -392,26 +392,19 @@ extension BrowserViewController: TabObserver {
     // reader mode page, most of the `TabObserver` methods are fired before
     // web main frame is ready,
     // For this case, we need to check readability after main frame did became available
-    guard frame.isMainFrame,
+    guard FeatureList.kUseProfileWebViewConfiguration.enabled,
+      frame.isMainFrame,
+      let readerMode = tab.readerMode,
       let url = tab.visibleURL,
       !url.isNewTabURL,
       !InternalURL.isValid(url: url) || url.isInternalURL(for: .readermode),
       !url.isFileURL
     else { return }
-    if FeatureList.kUseProfileWebViewConfiguration.enabled {
-      if let readerMode = tab.readerMode {
-        Task {
-          await readerMode.checkReadability()
-          if tabManager.selectedTab === tab {
-            topToolbar.updateReaderModeState(readerMode.state)
-          }
-        }
+    Task {
+      await readerMode.checkReadability()
+      if tabManager.selectedTab === tab {
+        topToolbar.updateReaderModeState(readerMode.state)
       }
-    } else {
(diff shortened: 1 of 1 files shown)


netzenbot (approved):

ios/brave-ios/Sources/Brave/Frontend/Browser/BrowserViewController/BVC+TabObserver.swift
```diff
@@ -386,6 +386,34 @@ extension BrowserViewController: TabObserver {
...
+      if let readerMode = tab.readerMode {
+        Task {
+          await readerMode.checkReadability()
+          if tabManager.selectedTab === tab {
+            topToolbar.updateReaderModeState(readerMode.state)
+          }
+        }
+      }
+    } else {
+      tab.evaluateJavaScript(
```
kylehickinson: You can probably just ignore this path entirely and add `FeatureList.kUseProfileWebViewConfiguration.enabled` to the guard above since we only require that the frame be available when using Chromium JS features

ios/brave-ios/Sources/Brave/Frontend/Browser/BrowserViewController/BVC+TabObserver.swift
```diff
@@ -386,6 +386,34 @@ extension BrowserViewController: TabObserver {
...
+    // web main frame is ready,
+    // For this case, we need to check readability after main frame did became available
+    guard frame.isMainFrame,
+      let url = tab.visibleURL,
+      !url.isNewTabURL,
+      !InternalURL.isValid(url: url) || url.isInternalURL(for: .readermode),
+      !url.isFileURL
+    else { return }
+    if FeatureList.kUseProfileWebViewConfiguration.enabled {
+      if let readerMode = tab.readerMode {
```
kylehickinson: Also may as well unwrap this in the guard too

kylehickinson (approved):

PR #35768, opened 2026-04-22
[Sidebar V2] Manage side panel border and resize area based on rounded corners pref

simonhong:
Resolves https://github.com/brave/brave-browser/issues/54735

With this PR, side panel's padding is toggled per rounded corners setting.

Introduces BraveSidePanelResizeArea, a SidePanelResizeArea subclass
that adapts its layout based on whether the panel has a border:

 - With border (rounded corners on): delegates to the upstream layout
  so the resize strip sits inside the existing border gap.
 - Without border (rounded corners off): positions itself as a 5 px
   strip at the inner panel edge, overlapping the content area and
   sitting above it in z-order to win hit-testing.

Adds SidePanel::SetBorderEnabled() via chromium_src macro injection
to toggle the border at runtime. When disabled, the panel's views::Border
is removed so content fills the full panel area. UpdateBorder() centralises
the state management and is called from a wrapped Open() to undo the
unconditional border_view_ re-show that upstream's UpdateVisibility() performs.
The border insets are also corrected: upstream's negative top inset
(meant to overlap the toolbar) is zeroed out because Brave does not need that overlap.

Fixes the side panel's vertical position: the upstream layout places
contents_height_side_panel 1 px above visual_client_area.y() to overlap the toolbar separator.
CalculateSideBarLayout now aligns the panel's y and height with the contents container
so the top separator remains fully visible when rounded corners are off.

BraveBrowserView::UpdateSidebarBorder() calls SetBorderEnabled() based on
ShouldUseBraveWebViewRoundedCornersForContents(), and BraveSidePanelResizeArea
is installed via SetResizeArea() during sidebar setup.

TEST=SidebarBrowserTest.SidebarV2ResizeAreaPositionMatchesBorderState

**NOTE:** Applying Brave's specific border style (corner radius, shadow or border padding) is left for a follow-up issue.

image

image

PR #36240, opened 2026-05-07
[Sidebar V2] Fix border not updating when switching between panels

simonhong:
When switching between sidebar panels while the sidebar stays open, the border style set 
by SetRoundedBorderEnabled(false) was being reset to visible.

The previous fix overrode VisibilityChanged() to reapply the border after upstream's
UpdateVisibility() reset it. But VisibilityChanged() is never called when switching from one open
panel to another — the sidebar container stays visible the whole time, so no visibility event fires.

Override Open()/Close() instead. Both are called for every panel transition, including panel
switches. Each wrapper calls the upstream implementation then UpdateBorder() to restore the
correct border state, overriding whatever UpdateVisibility() did inside.

TEST=SidebarBrowserTest.SidebarV2ResizeAreaPositionMatchesBorderState

Resolves: f/u PR for https://github.com/brave/brave-core/pull/35768

### DIFF
--- a/browser/ui/sidebar/sidebar_browsertest.cc
+++ b/browser/ui/sidebar/sidebar_browsertest.cc
@@ -2214,6 +2214,21 @@ IN_PROC_BROWSER_TEST_F(SidebarBrowserTest,
   RunScheduledLayouts();
   EXPECT_TRUE(side_panel->GetInsets().IsEmpty());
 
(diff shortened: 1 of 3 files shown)


netzenbot (approved):

mkarolin (approved):
LGTM

PR #36245, opened 2026-05-07
Upgrade from Chromium 149 to Chromium 150

### LINKED ISSUE #55302: Upgrade from Chromium 149 to Chromium 150
### Description

Upgrade from Chromium 149 to Chromium 150

https://chromium.googlesource.com/chromium/src/+log/149.0.7827.103..150.0.7871.13?pretty=fuller&n=10000

**QA tests:**

**Desktop Affected areas:**
* Verify that infobars work and look as expected (testing one or two different ones is probably sufficient - please check WDP and search conversion infobars as well)
* Verify that autocomplete works as expected
* Verify that bookmarks work as expected
* Verify that tab-related functionality (tabs, tab groups, dragging/splitting tabs, vertical tabs, etc.) works as expected
* Verify that omnibox icons look correct, particularly the Brave icon when visiting brave://settings, etc.
* Verify that brave://settings/autofill doesn't have unexpected visual/functional changes
* Verify that hamburger menu has expected menu items
* Veirfy that bookmark menu items continue to work as expected (including items added when containers feature flag is enabled)
* Verify that supported MV2 extensions continue to work as expected
* Verify side panel continues to work as expected

**iOS Affected areas:**
* Standard checks

**Android Affected areas:**
1. Password manager: ensure that can save/delete passwords 
2. Must not crash when granting geolocation permissions at maps.google.com
3. Ensure that Brave News are displayed
4. Ensure that possible to open a New Private tab on tablet device
5. Ensure that animation for `Open in new tab` works fine
6. Ensure can save/remove bookmark with a toolbar button

### LINKED ISSUE #56439: [CR150]  Bookmarks are not shown in the bookmarks sidepanel upon upgrade from CR 149-CR 150
### Description

Found while testing https://github.com/brave/brave-browser/issues/56422.  

Could not reproduce in `1.93.67 Chromium: 149.0.7827.115 (Official Build) nightly (64-bit)` --> `1.93.73 Chromium: 149.0.7827.115 (Official Build) nightly (64-bit)`

### Steps to reproduce

1. Installed `1.93.67 Chromium: 149.0.7827.115 (Official Build) nightly (64-bit) `
2. launched Brave
3. bookmarked few tabs to show on bookmarks bar and bookmarks side panel
4. upgrade to CR 150 via `brave://settings/help`

### Actual result

Bookmarks are not shown in the bookmarks side panel. 
However, they are shown in the bookmarks bar and `brave://bookmarks`

https://github.com/user-attachments/assets/0005e13b-e2e9-4907-aca5-0eea967b23c5

### Expected result

Bookmarks should be shown in the sidepanel

Image

### Reproduces how often

Easily reproduced

Brave | 1.93.70 Chromium: 150.0.7871.13 (Official Build) nightly (64-bit)
-- | --
Revision | aca2afe5ac16d7796435b5e27c058e7019173707

### Channel information

- [ ] release (stable)
- [ ] beta
- [x] nightly

### Reproducibility

### Miscellaneous information

cc: @rebron @emerick

cdesouza-chromium:
Resolves https://github.com/brave/brave-browser/issues/55302
Resolves https://github.com/brave/brave-browser/issues/56439


emerick:
Temporarily removed `CI/run-perf-smoke-tests` until https://issues.chromium.org/issues/512709653 is addressed upstream.

patches/tools-grit-grit-gather-policy_json.py.patch
```diff
@@ -1,9 +1,10 @@
...
-index c708c0a037b01d6a6517ab6022a579fdb29c607f..de2ca976e0ca2e5ae0bc4b0f034bfd3449304f4a 100644
+index c708c0a037b01d6a6517ab6022a579fdb29c607f..6a0b13a27a58b629b78de229c761bdcd64654444 100644
 --- a/tools/grit/grit/gather/policy_json.py
 +++ b/tools/grit/grit/gather/policy_json.py
 @@ -350,3 +350,4 @@ class PolicyJson(skeleton_gatherer.SkeletonGatherer):
        }
      else:
        raise Exception('Unknown build')
 +from brave_chromium_utils import inline_chromium_src_override; inline_chromium_src_override(globals(), locals())
+\ No newline at end of file
```
mkarolin: @cdesouza-chromium is this intentional?

zenparsing (approved):
Rewards 👍

tmancey (approved):

darkdh (approved):
sync, tor and wallet ++

browser/brave_news/rss_links_fetcher.cc
```diff
@@ -0,0 +1,102 @@
+// Copyright (c) 2026 The Brave Authors. All rights reserved.
```
fallaciousreasoning, line 1: Do we also need to include the chromium license header here as we're taking the file from upstream?

browser/ui/views/frame/brave_tab_strip_region_view.cc
fallaciousreasoning, line 1: For these icon renames I think we might need to add these to `leo_overrides.json` too. The `kXXXOldIcon` icons should point to the same Nala icon as the non-old version

browser/ui/views/overlay/brave_video_overlay_window_views.cc
fallaciousreasoning, line 1: ditto here sorry, for icons that are mentioned in `leo_overrides.json`

browser/ui/views/toolbar/bookmark_button.cc
fallaciousreasoning, line 1: and here 😢

browser/ui/views/download/bubble/download_toolbar_ui_controller_browsertest.cc
fallaciousreasoning, line 1: need relevant mentions in `leo_overrides` to be updated too

browser/ui/views/split_view/split_view_browsertest.cc
fallaciousreasoning, line 1: should this commit just land in master? This test fails semi-regularly there too

build/commands/lib/branding.js
```diff
@@ -203,10 +203,21 @@ const update = () => {
...
+  // destinations.
   const omniboxIconsBranding = config.isBraveOriginBranded
     ? 'brave_origin'
     : 'brave'
-  for (const iconName of ['product.icon', 'product_chrome_refresh.icon']) {
+  /** @type {[string, string][]} */
+  const omniboxIconMappings = [
+    ['product.icon', 'product_old.icon'],
+    ['product.icon', 'chrome_product.icon'],
+    ['product_chrome_refresh.icon', 'product_chrome_refresh_old.icon'],
```
fallaciousreasoning, line 218: potentially we should just override these icons with `leo_overrides.json` instead?

DJAndries (approved):
p3a++

supermassive (approved):
wallet lgtm

fallaciousreasoning (commented):
mostly LGTM, but I think we need to override the old+non-old icons in `leo_overrides` where they're mentioned

browser/brave_news/rss_links_fetcher.cc
emerick: Absolutely! Added the Chromium license header to the relevant files here https://github.com/brave/brave-core/pull/36245/commits/0c6eae8927985bbc55010a9db7094b7d46e728ae.

browser/ui/views/frame/brave_tab_strip_region_view.cc
emerick: Great catch, thank you for noticing that! Included all of the renamed icons here https://github.com/brave/brave-core/pull/36245/commits/4ded99b8ba499521539538a125cc57f78a5c7112.

browser/ui/views/overlay/brave_video_overlay_window_views.cc
emerick: Fixed in https://github.com/brave/brave-core/pull/36245/commits/4ded99b8ba499521539538a125cc57f78a5c7112.

browser/ui/views/toolbar/bookmark_button.cc
emerick: Fixed in https://github.com/brave/brave-core/pull/36245/commits/4ded99b8ba499521539538a125cc57f78a5c7112.

browser/ui/views/download/bubble/download_toolbar_ui_controller_browsertest.cc
emerick: Fixed in https://github.com/brave/brave-core/pull/36245/commits/4ded99b8ba499521539538a125cc57f78a5c7112.

browser/ui/views/split_view/split_view_browsertest.cc
emerick: 150 is going to land in master very shortly, so will just let this ride along if that's OK.

build/commands/lib/branding.js
emerick: These are branding related and have impact on Brave Origin, so I think it probably makes sense to leave them here. Definitely open to discussing it, but if we want to change this further it should probably be outside of the scope of the rebase.

kylehickinson (approved):
ios++

browser/ui/views/frame/brave_tab_strip_region_view.cc
fallaciousreasoning: thanks, sorry they're such a pain

fallaciousreasoning (approved):
lgtm, thanks for those changes @emerick

browser/brave_news/sources.gni
```diff
@@ -19,15 +19,21 @@ if (enable_brave_news) {
     "//brave/browser/brave_news/brave_news_tab_helper.h",
     "//brave/browser/brave_news/direct_feed_fetcher_delegate_impl.cc",
     "//brave/browser/brave_news/direct_feed_fetcher_delegate_impl.h",
+    "//brave/browser/brave_news/rss_links_fetcher.cc",
```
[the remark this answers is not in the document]
bridiver: yeah this is bad, we can add them for this bump, but this needs to be fixed in a follow-up

browser/sources.gni
```diff
@@ -787,6 +787,14 @@ brave_chrome_browser_ui_allow_circular_includes_from = [ "//brave/browser/ui" ]
 brave_chrome_browser_ui_allow_circular_includes_from +=
     brave_ui_allow_circular_includes_from
 
+# //brave/browser/ui is folded into //chrome/browser/ui, which has a circular
+# dependency with //chrome/browser:core. Mirror that relationship so brave UI
+# sources can include headers from //chrome/browser:core (e.g.
+# chrome/browser/defaults.h, chrome/browser/file_select_helper.h,
+# chrome/browser/command_observer.h).
+brave_chrome_browser_core_allow_circular_includes_from =
+    [ "//brave/browser/ui" ]
```
bridiver: shouldn't this (and most everything else) now be removed from `brave_chrome_browser_allow_circular_includes_from`?

git-log-300.txt
```diff
@@ -0,0 +1,8579 @@
+commit d90907784a4f422acce9cc6937e58c3db2076285
```
bridiver: what is this file?

browser/brave_news/sources.gni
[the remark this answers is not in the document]
bridiver: here's a start https://github.com/brave/brave-core/pull/37177

git-log-300.txt
AlexeyBarabash: this is wrongly committed file, removed at https://github.com/brave/brave-core/pull/36245/changes/b13b46b1d08c37ae895d82a1a692f1c9d043ec5a

browser/sources.gni
emerick: This is `brave_chrome_browser_core_allow_circular_includes_from` (for the new `:core` target), not `brave_chrome_browser_allow_circular_includes_from` (for `:browser`). It's needed here at this point in the commits, but it does go away in a later commit in cr150 (c5532c45abf), which deletes this block and replaces it with direct deps (`//chrome/browser:defaults`, `:file_select_helper`, etc.) once those headers get their own granular targets.

It seems to me that the other `allow_circular_includes_from` entries are still needed at this point until upstream componentizes further.

samartnik (approved):
android ++

fmarier (approved):

mkarolin (approved):
`strings`++

browser/sources.gni
bridiver: but I thought almost everything went into `core` so why would so deps still need to be in `brave_chrome_browser_allow_circular_includes_from`?

tools/cr/alias/source_rewrite.py
```diff
@@ -160,17 +160,24 @@
...
+    new_include = f'#include <{new_chromium_path.as_posix()}>'
+    new_content = content.replace(old_include, new_include)
+
+    # For .h shadow files, add IWYU pragma if not already present.
+    if path.suffix == '.h':
+        pragma_line = f'{new_include}  // IWYU pragma: export'
+        if pragma_line not in new_content:
+            new_content = new_content.replace(new_include, pragma_line)
+
+    path.write_text(new_content, encoding='utf-8', newline='\n')
```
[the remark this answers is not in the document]
bridiver: @brave/devops-team I thought we disabled sonarcloud for brave-core? These are not relevant

browser/permissions/permission_lifetime_manager_unittest.cc
```diff
@@ -300,8 +300,14 @@ TEST_F(PermissionLifetimeManagerTest, DifferentTypePermissions) {
 
   browser_task_environment_.FastForwardBy(kOneSecond);
 
+  // Use MEDIASTREAM_MIC rather than GEOLOCATION: on Android,
```
bridiver: These comments won't make sense without the context of the change here. I would just leave them out and put them in the commit comments instead

patches/chrome-browser-ui-page_action-page_action_model.cc.patch
```diff
@@ -0,0 +1,18 @@
+diff --git a/chrome/browser/ui/page_action/page_action_model.cc b/chrome/browser/ui/page_action/page_action_model.cc
+index c0dc2f8169e196d830389b2f0a970b417daaf364..6181f9e87b29c0ad12ebba19367010753f91a777 100644
+--- a/chrome/browser/ui/page_action/page_action_model.cc
++++ b/chrome/browser/ui/page_action/page_action_model.cc
+@@ -16,6 +16,7 @@ namespace {
+ using ::actions::ActionItem;
+ }  // namespace
+ 
++namespace chromium_impl {
```
bridiver, line 9: as discussed I think we want to avoid using namespace for this stuff

rewrite/content/browser/renderer_host/navigation_request.cc.yaml
```diff
@@ -0,0 +1,41 @@
...
+
+      Anchored on `    begin_params_->headers = headers.ToString();`
+      (4-space indent, only the in-`if` occurrence in this file) plus
+      the closing `}` and blank line that follow, so the insertion lands
+      exactly where the upstream block ends.
+
+      `GetTopDocumentGURL` comes from the chromium_src shadow file's
+      anonymous namespace.
+    re_pattern:
+      '(    begin_params_->headers = headers\.ToString\(\);\n  \}\n\n)'
```
bridiver: please avoid fixed whitespace

patches/chrome-browser-ui-webui-BUILD.gn.patch
```diff
@@ -1,12 +1,20 @@
...
- source_set("webui") {
++import("//brave/browser/ui/webui/sources.gni")
+ import("//chrome/common/features.gni")
+ import("//chromeos/dbus/config/use_real_dbus_clients.gni")
+ import("//components/enterprise/buildflags/buildflags.gni")
+@@ -90,6 +91,7 @@ source_set("configs") {
+     "//third_party/abseil-cpp:absl",
+     "//ui/webui:buildflags",
+   ]
++  deps += brave_browser_ui_webui_configs_deps
```
bridiver, line 10: I would still keep the import on the same line, I think there's no benefit to putting it at the top and it makes the replacement more complicated with two rules instead of one

patches/ui-color-BUILD.gn.patch
```diff
@@ -1,20 +1,44 @@
...
-   }
-+  import("//brave/ui/color/sources.gni") sources += brave_ui_color_mixer_sources public_deps += brave_ui_color_mixer_public_deps
- }
- 
- if (is_android) {
+@@ -2,6 +2,7 @@
+ # Use of this source code is governed by a BSD-style license that can be
+ # found in the LICENSE file.
+ 
++import("//brave/ui/color/sources.gni")
```
bridiver: same as previous one, I feel like this is just making the replacement more complex with no real benefit by adding more lines instead of putting the import on the same line

rewrite/chrome/browser/ui/webui/cr_components/searchbox/searchbox_omnibox_client.cc.yaml
```diff
@@ -0,0 +1,17 @@
...
+substitutions:
+  - description: |
+      Tag NTP searchbox Brave Search navigations with `source=newtab`.
+
+      Wrap the params handed to the destination `web_contents_->OpenURL()` call
+      within `OnAutocompleteAccept()` with `MaybeOverrideURLParams()` (defined
+      in the chromium_src shadow), which appends the source marker for Brave
+      Search keyword matches. Anchored on `OnAutocompleteAccept` so the
+      identically-shaped `OpenURL()` call in `OpenUrl()` is left untouched.
+    re_pattern: '(SearchboxOmniboxClient::OnAutocompleteAccept\(.*?web_contents_->OpenURL\(\s*)(content::OpenURLParams\(destination_url,.*?false\))'
```
bridiver: do we need to be this specific with individual params, etc...? Are there multiple calls and we need to pick out a specific one?

browser/permissions/permission_lifetime_manager_unittest.cc
emerick: cc: @AlexeyBarabash

patches/components-sync_device_info-device_info_sync_bridge.h.patch
```diff
@@ -0,0 +1,19 @@
...
+   ~DeviceInfoSyncBridge() override;
+ 
++  void DeleteDeviceInfo(const std::string& client_id,
++                        base::OnceClosure callback) override;
++  std::vector<std::unique_ptr<DeviceInfo>> GetAllBraveDeviceInfo()
++      const override;
++  void RefreshLocalDeviceInfoIfNeeded_ChromiumImpl();
++  void OnDeviceInfoDeleted(const std::string& client_id, const int attempt,
++                           base::OnceClosure callback);
++
```
bridiver: we should still avoid adding blank lines I think

patches/chrome-browser-ui-page_action-page_action_model.cc.patch
emerick: cc: @cdesouza-chromium

rewrite/content/browser/renderer_host/navigation_request.cc.yaml
emerick: cc: @cdesouza-chromium

patches/chrome-browser-ui-webui-BUILD.gn.patch
emerick: cc: @cdesouza-chromium

patches/ui-color-BUILD.gn.patch
emerick: cc: @cdesouza-chromium

rewrite/chrome/browser/ui/webui/cr_components/searchbox/searchbox_omnibox_client.cc.yaml
emerick: cc: @cdesouza-chromium

patches/components-sync_device_info-device_info_sync_bridge.h.patch
emerick: cc: @cdesouza-chromium

tools/cr/toolchains/install_extra_deps.py
```diff
@@ -0,0 +1,336 @@
...
+    meantime, we leave it to the archive's internal layout the determinatio of
+    where extracted files should go in the existing path.
+
+    Extraction uses tarfile's `data` filter (PEP 706) so a malformed archive
+    cannot escape `dest_dir` via absolute paths, `..` traversal, or links
+    pointing outside it. The archive is only extracted though based on the
+    verified sha256, so this is just extra defense.
+    """
+    with tarfile.open(archive_path, mode='r:*') as tar:
+        tar.extractall(path=dest_dir, filter='data')
```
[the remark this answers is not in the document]
bridiver: this rule is not relevant and should be removed from brave-core

browser/ui/views/tabs/brave_tab.cc
```diff
@@ -182,11 +182,16 @@ void BraveTab::MaybeStartObservingFullscreenChanges() {
     return;
   }
 
-  fullscreen_observation_.Observe(fullscreen_controller);
+  // Observe changes to fullscreen state.
+  fullscreen_subscription_ =
+      browser->GetExclusiveAccessManager()
+          ->fullscreen_controller()
+          ->RegisterOnFullscreenStateChanged(base::BindRepeating(
+              &BraveTab::OnFullscreenStateChanged, base::Unretained(this)));
```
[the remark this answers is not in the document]
bridiver: @thypon this is still flagging unchanged lines

browser/browsing_data/brave_browsing_data_remover_delegate.cc
```diff
@@ -161,8 +165,23 @@ void BraveBrowsingDataRemoverDelegate::RemoveEmbedderData(
     // Only clear the storage when the remove mask covers data types that live
     // on a storage partition.
     if (container_remove_mask) {
+      // Skip orphaned containers (locally used but no longer in the synced
+      // list). Their on-disk data is wiped wholesale by
+      // ContainersService::ScheduleOrphanedContainersCleanup; issuing a
+      // RemoveWithFilter here would instantiate the storage partition
+      // (can_create=true inside BrowsingDataRemoverImpl) and let its backends
+      // lazily recreate the directory we are trying to delete.
+      base::flat_set<std::string> synced_ids;
```
bridiver: test?

build/commands/lib/branding.js
bridiver: we are actively trying to get rid of things in this file so if there's a better alternative we should definitely consider that

rewrite/components/permissions/request_type.cc.toml
```diff
@@ -72,7 +72,12 @@ replace = '''\1
 #endif
     case RequestType::kBraveOpenAIChat:
       return ContentSettingsType::BRAVE_OPEN_AI_CHAT;\2
-    case RequestType::kWidevine:\3'''
+    case RequestType::kWidevine:
+#if !BUILDFLAG(ENABLE_BRAVE_WALLET)
```
bridiver, line 76: is the buildflag in deps?

browser/ui/views/side_panel/playlist/playlist_contents_wrapper.cc
```diff
@@ -68,7 +68,14 @@ void PlaylistContentsWrapper::EnterFullscreenModeForTab(
...
 
-  fullscreen_observation_.Observe(fullscreen_controller);
+  // Observe changes to fullscreen state.
+  fullscreen_subscription_ =
+      browser_view_->browser()
+          ->GetExclusiveAccessManager()
+          ->fullscreen_controller()
+          ->RegisterOnFullscreenStateChanged(base::BindRepeating(
+              &PlaylistContentsWrapper::OnFullscreenStateChanged,
+              base::Unretained(this)));
```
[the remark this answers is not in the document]
bridiver: This doesn't look safe, can't the browser window outlive the playlist sidepanel?

rewrite/third_party/rust/regex_automata/v0_4/BUILD.gn.toml
```diff
@@ -0,0 +1,20 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+[[substitution]]
+description = 'Allow regex_automata to be used outside of test targets'
+pattern = 'testonly = true'
+replace = 'testonly = false'
+
+[[substitution]]
+# Without this, chrome.dll links two `aho_corasick` rlibs with identical crate
```
bridiver: we should remove `//brave/third_party/rust/aho_corasick/v1` if it exists in upstream

patches/components-omnibox-browser-BUILD.gn.patch
```diff
@@ -1,20 +1,36 @@
...
-index e3769976d70686c8ce20f5f0ee13c1a31d895ec4..dec10c90d0dc35230a750721aa7a798b1a571ede 100644
+index e3769976d70686c8ce20f5f0ee13c1a31d895ec4..cbd9858b1f18464f374d591264c7ddce5ed76888 100644
 --- a/components/omnibox/browser/BUILD.gn
 +++ b/components/omnibox/browser/BUILD.gn
-@@ -404,6 +404,7 @@ static_library("browser") {
+@@ -330,6 +330,7 @@ static_library("browser") {
+     "zero_suggest_verbatim_match_provider.cc",
+     "zero_suggest_verbatim_match_provider.h",
+   ]
++  sources += brave_components_omnibox_browser_sources
```
bridiver, line 9: please still keep these on one line to simplify the replacement. The more complex we make these, the more difficult they will be to maintain and understand

browser/sources.gni
emerick: My understanding is that `core` did absorb most of Chromium's own code. But this is tied to where we inject our sources and we still put everything into `:browser` not `:core`, so there are circular deps between `:browser` and our targets that stick around no matter how slim `:browser` gets. The one entry the split actually lets us drop is `//brave/browser/ui`, and that happens in c5532c4. I think the rest go away one at a time as upstream finishes componentizing each header.

I can open up a followup issue if we want to go through these in a little more depth and see if there are any other deps we can eliminate.

@cdesouza-chromium may have a better handle on this one, feel free to chime in.

rewrite/third_party/rust/regex_automata/v0_4/BUILD.gn.toml
emerick: cc: @cdesouza-chromium

patches/components-omnibox-browser-BUILD.gn.patch
emerick: cc: @cdesouza-chromium

browser/browsing_data/brave_browsing_data_remover_delegate.cc
emerick: Great point, added one here https://github.com/brave/brave-core/pull/36245/commits/209737a79ecabb610491a7fa40c16e9f5b42dc3e

bridiver: I was thinking more along the lines of a unit test for BraveBrowsingDataRemoverDelegate, maybe we can look at converting it in a follow-up?

browser/sources.gni
bridiver: It definitely seems like there should be a follow up issue if we don't handle it here because if upstream moved most of their code into core then we should too

browser/browsing_data/brave_browsing_data_remover_delegate.cc
```diff
@@ -161,8 +165,23 @@ void BraveBrowsingDataRemoverDelegate::RemoveEmbedderData(
...
+      if (auto* service = ContainersServiceFactory::GetForProfile(profile_)) {
+        for (const auto& container : service->GetContainers()) {
+          synced_ids.insert(container->id);
+        }
+      }
       for (const content::StoragePartitionConfig& config :
            containers::GetUsedContainerStoragePartitionConfigs(profile_)) {
+        if (!synced_ids.contains(config.partition_name())) {
+          continue;
+        }
```
goodov: thanks for trying to fix this, but the fix doesn't look right. We need to keep the cleanup flow as it is, but instead adjust `GetUsedContainerStoragePartitionConfigs` to skip containers that are currently being deleted.

just disable the failing test, I'll land a fix separately in https://github.com/brave/brave-browser/issues/56319

goodov: it is also not clear: was the failure triggered by some cr150 change?

patches/components-sync_device_info-device_info_sync_bridge.h.patch
AlexeyBarabash: as I remember, this patch file  is backed by plaster

browser/permissions/permission_lifetime_manager_unittest.cc
AlexeyBarabash: Removed comments from code  at https://github.com/brave/brave-core/pull/36245/commits/f71b35a12f787a7c3cca8eea945dc5c7edb9ece9 . Commit message had the description.

browser/ui/views/side_panel/playlist/playlist_contents_wrapper.cc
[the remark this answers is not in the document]
cdesouza-chromium: It is safe. `FullscreenController` is owned by `BrowserWindowInterface::exclusive_access_manager_`, and that instance is destroyed at `BrowserWindowFeatures::TearDownPreBrowserWindowDestruction`, and after it is destroyed, there are n

https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/ui/browser_window/internal/browser_window_features.cc;l=1145

Our `PlaylistContentsWrapper` is owned by `PlaylistSidePanelCoordinator`, and we destroy `BrowserWindowFeatures::playlist_side_panel_coordinator_` in `BrowserWindowFeatures::TearDownPreBrowserWindowDestruction`, after the destruction of `exclusive_access_manager_`. 

https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/browser/ui/browser_window/internal/browser_window_features.cc?L165

Also, `FullscreenController` doesn't post the callback to a scheduler, but rather runs it sequentially when notifications occur.

patches/tools-grit-grit-gather-policy_json.py.patch
cdesouza-chromium: Fixed, but I've also added a plaster, because we shouldn't have got a merge conflict there. e410fd0a99b

browser/sources.gni
cdesouza-chromium: @emerick can we long a follow up issue to start doing some componentisation and readjusting some of these gn values to reflect the names of the new targets?

Regarding why we still need this, I have not dug deeper yet on this, but without it we would be getting this type of cycle:

```
ERROR Dependency cycle:
  //chrome/browser/ui:ui ->
  //brave/browser/ui:ui ->
```

So yeah, we do need a follow-up to better where the cycle really is, because it is not necessarily in core. There has been quite a lot of componentisation in upstream, so we may be over-permitting some of these cycles too, but I'm not sure which cases would those be, so we would have to go over each entry one by one.

patches/chrome-browser-ui-page_action-page_action_model.cc.patch
cdesouza-chromium: @bridiver I can correct that, but that feels a bit out of scope here, because `cr150` is just moving these files from `c/b/u/page_actions/` -> `c/b/u/page_action/`. See https://crrev.com/c/7818759

I'm gonna do these corrections on `master` once `cr150` is merged.

browser/browsing_data/brave_browsing_data_remover_delegate.cc
emerick: Sorry, I meant to ping you on that change, actually. I reverted it and disabled the test separately here https://github.com/brave/brave-core/pull/36245/commits/721ef2d43a2e54ad56dc951602283a48771d002f.

Looking at this a bit closer, I suspect it could be related to this upstream change https://source.chromium.org/chromium/chromium/src/+/ca762005442f0963c6211f86a46300b41327d5f6.

browser/sources.gni
emerick: Sure, I captured the details in https://github.com/brave/brave-browser/issues/56332.

rewrite/content/browser/renderer_host/navigation_request.cc.yaml
cdesouza-chromium: Fixed in 63eb90a417a0c280c5e4f88e7c0ca6e08be68cf1

rewrite/chrome/browser/ui/webui/cr_components/searchbox/searchbox_omnibox_client.cc.yaml
cdesouza-chromium: Done on a5abb4a754d, but maybe now that we are using plaster, we could change this to be simpler.

rewrite/components/permissions/request_type.cc.toml
emerick: Yes, it's included via `brave_components_permissions_deps` and that's patched into the appropriate `BUILD.gn`, etc.

patches/components-omnibox-browser-BUILD.gn.patch
bridiver: sorry as discussed it doesn't have to be one line anymore, but please continue to keep them grouped together for now

rewrite/third_party/rust/regex_automata/v0_4/BUILD.gn.toml
emerick: Removed here https://github.com/brave/brave-core/pull/36245/commits/452822af1fb53dd68181ac1f14a079ee543a3c36

patches/chrome-browser-ui-webui-BUILD.gn.patch
bridiver: sorry, not the same line, just grouped together

chromium_src/chrome/browser/google/sources.gni
```diff
@@ -0,0 +1,24 @@
...
+
+# Brave builds the Google Update integration sources for official Windows builds
+# even though we are not is_chrome_branded. Upstream only adds them to the
+# //chrome/browser/google target under is_chrome_branded, so we add them here
+# instead of injecting them into the //chrome/browser:browser monolith. Keeping
+# them in //chrome/browser/google lets modularized consumers such as
+# //chrome/browser/feedback:feedback_impl reach the headers through their
+# existing dependency on this target.
+if (is_win && is_official_build && !is_chrome_branded) {
+  brave_chrome_browser_google_sources += [
```
bridiver, line 18: I think it's probably better here to just patch out `is_chrome_branded` with plaster?

patches/components-sync_device_info-device_info_sync_bridge.h.patch
bridiver: but the empty line doesn't appear to serve any useful purpose without more context? It make more sense i we did something like `// Brave` or something at the beginning

rewrite/components/sync_device_info/device_info_sync_bridge.cc.yaml
```diff
@@ -0,0 +1,89 @@
...
+  - description: |
+      Rename upstream `RefreshLocalDeviceInfoIfNeeded` to `_ChromiumImpl`.
+
+      The `chromium_src` bails out before calling into it when `all_data_` no
+      longer contains the local device guid. This avoids a crash that happened
+      when the user closed sync settings before `OnDeviceInfoDeleted`
+      acknowledged.
+    re_pattern:
+      '(void DeviceInfoSyncBridge::)RefreshLocalDeviceInfoIfNeeded(\(\))'
+    replace: '\1RefreshLocalDeviceInfoIfNeeded_ChromiumImpl\2'
```
bridiver: this seems like unnecessary added complexity. Why don't we just insert our changes into the original method?

patches/components-sync_device_info-device_info_sync_bridge.cc.patch
```diff
@@ -1,24 +1,48 @@
...
++void DeviceInfoSyncBridge::RefreshLocalDeviceInfoIfNeeded_ChromiumImpl() {
+   // Device info cannot be synced if the provider is not initialized. When it
+   // gets initialized, local device info will be sent.
+   if (!local_device_info_provider_->GetLocalDeviceInfo()) {
+@@ -548,17 +549,19 @@ std::optional<ModelError> DeviceInfoSyncBridge::ApplyIncrementalSyncChanges(
    for (const std::unique_ptr<EntityChange>& change : entity_changes) {
      const std::string guid = change->storage_key();
  
-+    BRAVE_DEVICE_INFO_SYNC_BRIDGE_APPLY_SYNC_CHANGES_SKIP_NEXT_IF
++#if 0  // Brave: re-upload defeats removing other devices in the sync chain
```
bridiver, line 47: I think it would be better to patch in `false &&` into the `if` statement so we only have a single insertion

rewrite/third_party/rust/regex_automata/v0_4/BUILD.gn.toml
bridiver: or am I misunderstanding the problem here?

tools/cr/alias/source_rewrite.py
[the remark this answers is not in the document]
mihaiplesa: It is completely disabled but this is an old PR/comment.

chromium_src/chrome/browser/google/sources.gni
emerick: I was about to make that change, but they've continued modifying this file upstream in cr151 (https://source.chromium.org/chromium/chromium/src/+/ec5782b198a8aa274141d6f47e86267c9b01ba12). I think the best approach might be to leave this as-is here and I'll reach out to @mkarolin and let him know he can revert this in cr151.

patches/components-sync_device_info-device_info_sync_bridge.cc.patch
emerick: cc: @cdesouza-chromium

rewrite/components/sync_device_info/device_info_sync_bridge.cc.yaml
emerick: cc: @cdesouza-chromium

rewrite/third_party/rust/regex_automata/v0_4/BUILD.gn.toml
emerick: OK to remove our crate @cdesouza-chromium?

patches/components-sync_device_info-device_info_sync_bridge.h.patch
emerick: I don't think we added a line at the end, though. We added a line at the beginning after the replacement and I guess that throws the diff off?

```
    replace: |-
      \1\2

      \1void DeleteDeviceInfo(const std::string& client_id,
      \1                      base::OnceClosure callback) override;
      \1std::vector<std::unique_ptr<DeviceInfo>> GetAllBraveDeviceInfo()
      \1    const override;
      \1void RefreshLocalDeviceInfoIfNeeded_ChromiumImpl();
      \1void OnDeviceInfoDeleted(const std::string& client_id, const int attempt,
      \1                         base::OnceClosure callback);
```

I guess we could delete that blank line, though.

emerick: Made that change here https://github.com/brave/brave-core/pull/36245/commits/cfbb9177b70246c6df6768a7c9322307e7bb4d61.

patches/chrome-browser-ui-webui-BUILD.gn.patch
bridiver: Just to be clear there are several reasons I'm asking for these changes:
1. I'm really a struggling to review all of these plaster conversions when we're simultaneously converting them to plaster and also making changes to how the patches work. It's the same thing with PRs if a file is moved and also changed. First we have a PR to move the file and then one to change it so it doesn't show up as a deleted file and then a new file. 
2. This is not unambiguously better. There are trade-offs because splitting them makes it harder to see all the changes we have made when you're looking at the source file. Grouping them together makes it easer to see that we're added sources and deps, etc...
3. Related to 1, we should make changes incrementally so we don't drop a completely new way to do everything on devs and so we can validate that we're moving in the right direction before we start pulling more levers.

patches/components-sync_device_info-device_info_sync_bridge.h.patch
bridiver: Having the blank line is useful if we're clearly designating these as Brave methods with a comment, but yes, I guess git doesn't really differentiate which new line is blank

rewrite/third_party/rust/regex_automata/v0_4/BUILD.gn.toml
cdesouza-chromium: Fixed here 8f5fe914b38

patches/chrome-browser-ui-webui-BUILD.gn.patch
cdesouza-chromium: I've bundled them together in 3df9f423c52. Do you want me to insert at the end of the target?

patches/ui-color-BUILD.gn.patch
cdesouza-chromium: Fixed in a0f8efff5e2. I'm adding at the end of each target, and I've reverted back to repeat the import in each case, as the patch was before.

patches/components-omnibox-browser-BUILD.gn.patch
cdesouza-chromium: Fixed in eac03fb27a3636675831bfe7c1bd6a353d808037. I've also migrated the plaster to yaml.

rewrite/components/sync_device_info/device_info_sync_bridge.cc.yaml
cdesouza-chromium: Fixed in 3020e541673.

patches/components-sync_device_info-device_info_sync_bridge.cc.patch
cdesouza-chromium: @bridiver that's not safe. Imagine `if (false && true)` becomes `if (false && true || false)`. Also notice that that is being done by a single substitution, no two.

patches/chrome-browser-ui-webui-BUILD.gn.patch
cdesouza-chromium: Changed it in 88c0a9f4a5c to always insert as the last line of the target.

rewrite/components/sync_device_info/device_info_sync_bridge.cc.yaml
cdesouza-chromium: Further fix in 81838dd1289

patches/components-sync_device_info-device_info_sync_bridge.cc.patch
bridiver: Ah right because it includes the original as part of the substitution. Ok

patches/chrome-browser-sharing_hub-sharing_hub_model.cc.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/chrome/browser/sharing_hub/sharing_hub_model.cc b/chrome/browser/sharing_hub/sharing_hub_model.cc
+index cccf8f87396595f3117a7b63080eeab7fb177091..a1889c1158d22d5eeccf08de413f861abd582643 100644
+--- a/chrome/browser/sharing_hub/sharing_hub_model.cc
++++ b/chrome/browser/sharing_hub/sharing_hub_model.cc
+@@ -96,7 +96,7 @@ void SharingHubModel::PopulateFirstPartyActions() {
+   DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+ 
+   first_party_action_list_.emplace_back(
+-      IDC_COPY_URL, l10n_util::GetStringUTF16(IDS_SHARING_HUB_COPY_LINK_LABEL),
++      IDC_COPY_URL, l10n_util::GetStringUTF16(MaybeAddCopyCleanLinkItem(context_, &first_party_action_list_)),
```
bridiver, line 10: For follow-up I think we should change the way this works because this method works very strangely as a result of originally being a chromium src override. We should just patch in a call to add COPY_CLEAN_LINK before `first_party_action_list_.emplace_back`

patches/chrome-browser-ui-browser_window-internal-browser_window_features.cc.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/chrome/browser/ui/browser_window/internal/browser_window_features.cc b/chrome/browser/ui/browser_window/internal/browser_window_features.cc
+index 6276b2ba443087cc34f3a332a961ec8e0530fd92..08a884534388c4a9bd3c62643bcba7db09bba9f0 100644
+--- a/chrome/browser/ui/browser_window/internal/browser_window_features.cc
++++ b/chrome/browser/ui/browser_window/internal/browser_window_features.cc
+@@ -287,7 +287,7 @@ void BrowserWindowFeatures::Init(BrowserWindowInterface* browser) {
+           *browser, fullscreen_controller_.get(), app_browser_controller_.get(),
+           browser->GetType(),
+           browser->GetBrowserForMigrationOnly()->is_trusted_source(),
+-          browser->GetUnownedUserDataHost());
++          browser->GetUnownedUserDataHost(), *browser);
```
bridiver, line 10: I think we need to fix the underlying problem here in ShouldShowBraveVerticalTabs and SupportsBraveVerticalTabs which is that they shouldn't take `BrowserWindowInterface` as a param in the first place. What they actually need are the profile prefs and browser type. We're creating so many patches to keep a reference to a thing that we're only using to get references to the things we actually need.

bridiver: This could potentially be a follow-up, but it should be high priority because this isn't great imo

browser/brave_adaptive_captcha/brave_adaptive_captcha_service_factory.cc
```diff
@@ -46,7 +46,8 @@ class CaptchaDelegate
     // isn't associated with any particular `Browser` instance and we can use
```
bridiver, line 46: We should open an issue for this because it's dangerous among other issues. The browser window we pick could be destroyed while the captcha is still open (and the rewards panel coordinator along with it). The architecture here seems problematic and we need to rethink how this works. cc @tmancey

chromium_src/components/search_engines/template_url.cc
```diff
@@ -12,6 +12,11 @@
   case template_url_starter_pack_data::StarterPackId::kAskBraveSearch: \
     return KEYWORD_MODE_STARTER_PACK_ASK_BRAVE_SEARCH
 
+// Suppress Chrome-specific source attribution (chrome.ob/chrome.rb) in Google
+// search.
+#define BRAVE_TEMPLATE_URL_REPLACE_PARAMETER return true;
```
bridiver, line 17: it's very strange that the method returns a bool that is always true, but this still feels fragile. Can we just remove `GOOGLE_SEARCH_SOURCE` after it gets added or remove `google:searchSource` from the prepopulated engines configuration?

patches/chrome-browser-sharing_hub-sharing_hub_model.cc.patch
emerick: Created follow-up issue here https://github.com/brave/brave-browser/issues/56389.

patches/chrome-browser-ui-browser_window-internal-browser_window_features.cc.patch
emerick: Created a followup here https://github.com/brave/brave-browser/issues/56390 and pinged @sangwoo108, we'll make sure it gets fixed once cr150 merges.

browser/brave_adaptive_captcha/brave_adaptive_captcha_service_factory.cc
emerick: Created https://github.com/brave/brave-browser/issues/56391 to track this.

chromium_src/components/search_engines/template_url.cc
emerick: Good call, fixed here https://github.com/brave/brave-core/pull/36245/commits/c78dd572a6c2e456035cc5188af5fcaba289cd23.

bridiver (approved):

PR #36247, opened 2026-05-07
[query_filter] Update BraveQueryFilter tests.

### LINKED ISSUE #55341: Add more tests and also update BraveQueryFilter tests
### Description

The [BraveQueryFilter](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/components/query_filter/browser/utils_unittest.cc), is currently missing tests on checking the [conditional filtering logic](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/components/query_filter/browser/utils.cc?L135). 

The test suite can also be improved by updating the `[IsScopedTracker](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/components/query_filter/browser/utils_unittest.cc?L112)` tests to test the actual [MaybeApplyQueryStringFilter](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/components/query_filter/browser/utils.cc?L260) public API.

r0hit0303:
This change: 
- Adds missing tests on conditional query filtering logic. 
- Updates the tests around scoped query tracking to make the tests test directly the public API `MaybeApplyQueryStringFilter`. The `IsScopedTracker` and the hardcoded rules would also be [removed](https://github.com/brave/brave-core/pull/36222#discussion_r3197754497) once `QueryFilterComponent` flag is enabled.  
- Adds a failing wildcard test in `BraveQueryFilter#ScopedQueryTrackingTest` with a TODO. As, the current existing logic doesn't have support for it. This would be updated with the `QueryFilterComponent`. 

Resolves https://github.com/brave/brave-browser/issues/55341

### DIFF
--- a/components/query_filter/browser/utils.cc
+++ b/components/query_filter/browser/utils.cc
@@ -19,6 +19,11 @@
 
 namespace {
 
+// Allows to replace the |kScopedQueryStringTrackers| with custom values during
+// testing.
+const query_filter::ScopedQueryTrackerType* g_scoped_trackers_for_testing_ =
+    nullptr;
+
 static constexpr auto kSimpleQueryStringTrackers =
     base::MakeFixedFlatSet<std::string_view>(
         base::sorted_unique,
@@ -153,7 +158,7 @@ static constexpr auto kConditionalQueryStringTrackers =
 // added in the meantime to fix the build error. Remove this attribute and
 // provide a proper fix.
 [[clang::no_destroy]] static const auto kScopedQueryStringTrackers =
-    std::map<std::string_view, std::vector<std::string_view>>({
+    query_filter::ScopedQueryTrackerType({
         // https://github.com/brave/brave-browser/issues/35094
         {"igsh", {"instagram.com"}},
         // https://github.com/brave/brave-browser/issues/11580
@@ -176,10 +181,12 @@ static constexpr auto kExemptedHostnames =
             "urldefense.com",
         });
 
-bool IsScopedTracker(
-    std::string_view param_name,
-    std::string_view spec,
-    const std::map<std::string_view, std::vector<std::string_view>>& trackers) {
+bool IsScopedTracker(std::string_view param_name, std::string_view spec) {
+  query_filter::ScopedQueryTrackerType trackers =
+      (g_scoped_trackers_for_testing_ != nullptr)
+          ? *g_scoped_trackers_for_testing_
+          : kScopedQueryStringTrackers;
+
   if (!trackers.contains(param_name)) {
     return false;
   }
@@ -219,7 +226,7 @@ std::optional<std::string> StripQueryParameter(std::string_view query,
     const std::string_view key = pieces.empty() ? "" : pieces[0];
     if (pieces.size() >= 2 &&
         (kSimpleQueryStringTrackers.count(key) == 1 ||
-         IsScopedTracker(key, spec, kScopedQueryStringTrackers) ||
+         IsScopedTracker(key, spec) ||
          (kConditionalQueryStringTrackers.count(key) == 1 &&
           !re2::RE2::PartialMatch(
               spec, kConditionalQueryStringTrackers.at(key).data())))) {
@@ -301,10 +308,8 @@ std::optional<GURL> MaybeApplyQueryStringFilter(
   return ApplyQueryFilter(request_url);
 }
 
-bool IsScopedTrackerForTesting(  // IN-TEST
-    std::string_view param_name,
-    std::string_view spec,
-    const std::map<std::string_view, std::vector<std::string_view>>& trackers) {
-  return IsScopedTracker(param_name, spec, trackers);
+void SetScopedTrackerForTesting(  // IN-TEST
+    const ScopedQueryTrackerType* trackers) {
+  g_scoped_trackers_for_testing_ = trackers;
 }
 }  // namespace query_filter
--- a/components/query_filter/browser/utils.cc
+++ b/components/query_filter/browser/utils.cc
@@ -10,6 +10,7 @@
 #include <string_view>
 #include <vector>
 
+#include "base/check_is_test.h"
 #include "base/containers/fixed_flat_map.h"
 #include "base/containers/fixed_flat_set.h"
 #include "base/strings/string_split.h"
@@ -182,10 +183,14 @@ static constexpr auto kExemptedHostnames =
         });
 
 bool IsScopedTracker(std::string_view param_name, std::string_view spec) {
-  query_filter::ScopedQueryTrackerType trackers =
-      (g_scoped_trackers_for_testing_ != nullptr)
-          ? *g_scoped_trackers_for_testing_
-          : kScopedQueryStringTrackers;
+  query_filter::ScopedQueryTrackerType trackers;
+
+  if (g_scoped_trackers_for_testing_) {
+    CHECK_IS_TEST();
+    trackers = *g_scoped_trackers_for_testing_;
+  } else {
+    trackers = kScopedQueryStringTrackers;
+  }
 
   if (!trackers.contains(param_name)) {
     return false;
--- a/components/query_filter/browser/utils.h
+++ b/components/query_filter/browser/utils.h
@@ -38,7 +38,8 @@ std::optional<GURL> MaybeApplyQueryStringFilter(
     const std::string& request_method,
     const bool internal_redirect);
 
-// A method to override the scoped trackers maps for testing purposes.
+// Overrides the scoped trackers used by IsScopedTracker() for testing. Pass
+// nullptr to restore the default production data.
 void SetScopedTrackerForTesting(const ScopedQueryTrackerType* trackers);
 }  // namespace query_filter
 #endif  // BRAVE_COMPONENTS_QUERY_FILTER_BROWSER_UTILS_H_
--- a/components/query_filter/browser/utils.h
+++ b/components/query_filter/browser/utils.h
@@ -6,17 +6,17 @@
 #ifndef BRAVE_COMPONENTS_QUERY_FILTER_BROWSER_UTILS_H_
 #define BRAVE_COMPONENTS_QUERY_FILTER_BROWSER_UTILS_H_
 
+#include <map>
 #include <optional>
 #include <string>
 #include <string_view>
-#include <unordered_map>
 #include <vector>
 
 #include "url/gurl.h"
 
 namespace query_filter {
 using ScopedQueryTrackerType =
-    std::unordered_map<std::string_view, std::vector<std::string_view>>;
+    std::map<std::string_view, std::vector<std::string_view>>;
 
 std::optional<GURL> ApplyQueryFilter(const GURL& original_url);
 
--- a/components/query_filter/browser/utils.h
+++ b/components/query_filter/browser/utils.h
@@ -6,14 +6,18 @@
 #ifndef BRAVE_COMPONENTS_QUERY_FILTER_BROWSER_UTILS_H_
 #define BRAVE_COMPONENTS_QUERY_FILTER_BROWSER_UTILS_H_
 
-#include <map>
 #include <optional>
 #include <string>
+#include <string_view>
+#include <unordered_map>
 #include <vector>
 
 #include "url/gurl.h"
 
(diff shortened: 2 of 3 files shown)


r0hit0303:
hello @fmarier and @cdesouza-chromium, could you PTAL at this small refactoring change? Thanks!

components/query_filter/browser/utils.h
```diff
@@ -6,14 +6,18 @@
...
+#include <unordered_map>
 #include <vector>
 
 #include "url/gurl.h"
 
 namespace query_filter {
+using ScopedQueryTrackerType =
+    std::unordered_map<std::string_view, std::vector<std::string_view>>;
+
 std::optional<GURL> ApplyQueryFilter(const GURL& original_url);
```
netzenbot, line 22: `std::unordered_map` is banned per Chromium container guidelines ([CSA-005](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-005)). Use `absl::flat_hash_map<std::string_view, std::vector<std::string_view>>` instead.

components/query_filter/browser/utils.cc
```diff
@@ -19,6 +19,11 @@
```
netzenbot, line 20: The `g_scoped_trackers_for_testing_` global and the conditional branch in `IsScopedTracker` are solely for test support (`_for_testing` suffix, `// IN-TEST` marker, reset in `TearDown`). This adds a null-check on every production call. The CSA-035 exception is for thin accessors exposing internalized features, not for overriding a production data map. Consider using a seam (e.g., a virtual method or a dependency-injected accessor) that doesn't touch the production code path, or keep the test-only helper in a `_test_support` target. https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-035

components/query_filter/browser/utils.cc
```diff
@@ -176,10 +181,12 @@ static constexpr auto kExemptedHostnames =
             "urldefense.com",
         });
 
-bool IsScopedTracker(
-    std::string_view param_name,
-    std::string_view spec,
-    const std::map<std::string_view, std::vector<std::string_view>>& trackers) {
+bool IsScopedTracker(std::string_view param_name, std::string_view spec) {
```
netzenbot, line 185: nit: The non-null branch for `g_scoped_trackers_for_testing_` is a test-only path — `CHECK_IS_TEST()` should precede the ternary (or guard the branch) to make that explicit. See [ARCH-024](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-024).

components/query_filter/browser/utils_unittest.cc
```diff
@@ -109,47 +126,131 @@ TEST(BraveQueryFilter, FilterQueryTrackers) {
...
-  EXPECT_TRUE(query_filter::IsScopedTrackerForTesting(
-      "igshid", "http://www.instagram.com/", trackers));
-  EXPECT_FALSE(query_filter::IsScopedTrackerForTesting(
-      "igshid", "https://example.com/", trackers));
+  EXPECT_EQ(query_filter::MaybeApplyQueryStringFilter(
+                GURL("https://brave.com"), GURL(),
+                GURL("https://instagram.com/?igshid=123"), "GET", false),
+            GURL("https://instagram.com/"));
+  EXPECT_EQ(query_filter::MaybeApplyQueryStringFilter(
+                GURL("https://brave.com"), GURL(),
```
netzenbot, line 148: nit: `std::unordered_map` is used directly here but `<unordered_map>` isn't included in this file — it comes transitively from `utils.h`. Per [CS-001](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-001), include what you use directly. Either add `#include <unordered_map>` or use the `query_filter::ScopedQueryTrackerType` alias (which also avoids repeating the type).

components/query_filter/browser/utils_unittest.cc
```diff
@@ -10,7 +10,24 @@
...
+// fixture on the QueryFilterComponent feature flag once the support to parse
+// the new downloaded rules is added.
+class BraveQueryFilter : public testing::Test {
+ public:
+  void SetUp() override {}
+
+  void TearDown() override {
+    query_filter::SetScopedTrackerForTesting(nullptr);
+  }
+
```
netzenbot, line 19: nit: `void SetUp() override {}` does nothing — don't override it at all. The `TearDown` already handles cleanup; `SetUp` has no purpose here. See [CS-030](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-030).

components/query_filter/browser/utils.cc
r0hit0303: Given that |g_scoped_trackers_for_testing_| would be removed in this [change](https://github.com/brave/brave-core/pull/36222#discussion_r3197754497). I think doing this way should be fine.

components/query_filter/browser/utils_unittest.cc
r0hit0303: Removed

components/query_filter/browser/utils.h
r0hit0303: Reverted back to using std::map. It's just that having a sorted container was not needed here in the first place, which had an extra O(logN) complexity for insertion and deletion. We will also get rid of the map anyways in the next change.

components/query_filter/browser/utils_unittest.cc
```diff
@@ -109,47 +122,130 @@ TEST(BraveQueryFilter, FilterQueryTrackers) {
...
+  EXPECT_FALSE(query_filter::MaybeApplyQueryStringFilter(
+      GURL("https://example.com"), GURL(),
+      GURL("https://brave.com/?sample4=123"), "GET", false));
+
+  // Wildcard case for the domain *://*.facebook.com/*.
+  // TODO(https://github.com/brave/brave-browser/issues/55305): The new rules
+  // would have URL wildcard support on domains. Currently we don't have that
+  // support so query filter is not applied. When that's supported update this
+  // test to EXPECT_EQ.
+  EXPECT_FALSE(query_filter::MaybeApplyQueryStringFilter(
```
fmarier, line 195: This one will always be false because it's a same-site navigation (which are exempt from query string filtering). You'll have to change the first facebook.com to example.com to make it true with wildcard support.

components/query_filter/browser/utils_unittest.cc
```diff
@@ -109,47 +122,130 @@ TEST(BraveQueryFilter, FilterQueryTrackers) {
...
+      GURL("https://facebook.com"), GURL(),
+      GURL("https://mobile.facebook.com/evil=666"), "GET", false));
+}
+
+TEST_F(BraveQueryFilter, ConditionalFilteringTest) {
```
fmarier, line 200: We could avoid the duplicate comments in this test by grouping the test cases related to the same param together:
```
  // Conditional filtering, `ck_subscriber_id` param gets removed when
  // `unsubscribe` not present in url.
  EXPECT_EQ(query_filter::MaybeApplyQueryStringFilter(
                GURL(), GURL("https://brave.com"),
                GURL("https://test.com/?ck_subscriber_id=123"), "GET", false),
                GURL("https://test.com/"));
  EXPECT_FALSE(query_filter::MaybeApplyQueryStringFilter(
      GURL(), GURL("https://brave.com"),
      GURL("https://unsubscribe.com/?ck_subscriber_id=123"), "GET", false));
```

fmarier (changes requested):

components/query_filter/browser/utils_unittest.cc
```diff
@@ -109,47 +122,130 @@ TEST(BraveQueryFilter, FilterQueryTrackers) {
...
+  // `Unsubscribe` present in url.
+  EXPECT_FALSE(query_filter::MaybeApplyQueryStringFilter(
+      GURL(), GURL("https://brave.com"),
+      GURL("https://Unsubscribe.com/?mkt_tok=123"), "GET", false));
+  // Conditional filtering, test nothing filtered. `mkt_tok` param with
+  // `emailWebview` present in url.
+  EXPECT_FALSE(query_filter::MaybeApplyQueryStringFilter(
+      GURL(), GURL("https://brave.com"),
+      GURL("https://test.com/emailWebview?mkt_tok=123"), "GET", false));
 }
```
netzenbot, line 241: The URL `GURL("https://mobile.facebook.com/evil=666")` has no query string — `evil=666` is in the path, not after `?`. `MaybeApplyQueryStringFilter` returns early at `request_url.has_query()` before the scoped tracker check is reached, so this test proves nothing about wildcard domain matching. The TODO expects to flip to `EXPECT_EQ` once wildcard support lands, but that also won't work unless the URL is corrected to `https://mobile.facebook.com/?evil=666`. See [TI-012](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-012).

components/query_filter/browser/utils.h
```diff
@@ -34,9 +38,7 @@ std::optional<GURL> MaybeApplyQueryStringFilter(
     const std::string& request_method,
     const bool internal_redirect);
 
-bool IsScopedTrackerForTesting(
-    std::string_view param_name,
-    std::string_view spec,
-    const std::map<std::string_view, std::vector<std::string_view>>& trackers);
+// A method to override the scoped trackers maps for testing purposes.
+void SetScopedTrackerForTesting(const ScopedQueryTrackerType* trackers);
 }  // namespace query_filter
```
netzenbot, line 44: nit: The comment `// A method to override the scoped trackers maps for testing purposes.` omits that passing `nullptr` resets the trackers to the default production data — which is how `TearDown` uses it. Per [DOC-004](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-004), method docs should describe the contract. Consider: `// Overrides the scoped trackers used by IsScopedTracker() for testing. Pass nullptr to restore the default production data.`

components/query_filter/browser/utils_unittest.cc
```diff
@@ -109,47 +122,130 @@ TEST(BraveQueryFilter, FilterQueryTrackers) {
...
-  EXPECT_FALSE(query_filter::IsScopedTrackerForTesting(
-      "ref_src", "https://z.com/", trackers));
+  EXPECT_EQ(query_filter::MaybeApplyQueryStringFilter(
+                GURL("https://brave.com"), GURL(),
+                GURL("https://twitter.com/?ref_src=123"), "GET", false),
+            GURL("https://twitter.com/"));
+  EXPECT_EQ(query_filter::MaybeApplyQueryStringFilter(
+                GURL("https://brave.com"), GURL(),
+                GURL("https://x.com/?ref_src=123"), "GET", false),
+            GURL("https://x.com/"));
```
netzenbot, line 163: nit: `ScopedQueryTrackingTest` repeats the same `MaybeApplyQueryStringFilter` call pattern for each tracker/domain pair. Per [TI-023](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-023), consider a table-driven loop with `SCOPED_TRACE` — it would halve the line count and make adding new cases trivial.

components/query_filter/browser/utils_unittest.cc
r0hit0303: Oops. My bad, I missed that little detail. Thanks for pointing it out. Done.

components/query_filter/browser/utils_unittest.cc
r0hit0303: Done

components/query_filter/browser/utils_unittest.cc
r0hit0303: Ack. I tried to update the tests locally with what the bot proposed but it made the tests very DRY. I can try and follow-up on this separately if required.

components/query_filter/browser/utils_unittest.cc
r0hit0303: I missed to notice that the "?" was omitted. I've added it in the latest patch.

fmarier (approved):

r0hit0303:
Thanks François for the review and moving @cdesouza-chromium to cc as discussed offline before merging the change. Thank you all!

PR #36255, opened 2026-05-07
[FocusMode] Introduce TabStripPlacementCoordinator

zenparsing:
Adds a TabStripPlacementCoordinator that owns the parenting of the horizontal tab strip region view and selects an appropriate parent based on browser state (vertical tabs, focus mode, default). Callers register parents per PlacementKind and trigger reassignment via UpdatePlacement.

Migrates BraveVerticalTabStripRegionView off its ad-hoc reparenting onto the coordinator.

The kTopContainer placement is wired up but left unused; Focus Mode will register a parent for it in a follow-up.

### DIFF
--- a/browser/ui/BUILD.gn
+++ b/browser/ui/BUILD.gn
@@ -731,7 +731,7 @@ source_set("ui") {
       "//brave/browser/ui/side_panel",
       "//brave/browser/ui/sidebar/buildflags",
       "//brave/browser/ui/views:layout_provider",
-      "//brave/browser/ui/views/frame",
+      "//brave/browser/ui/views/frame:frame_impl",
       "//brave/browser/ui/views/location_bar",
       "//brave/browser/ui/views/page_action",
       "//brave/browser/ui/views/tabs/accent_color",
--- a/browser/ui/views/frame/brave_browser_view.cc
+++ b/browser/ui/views/frame/brave_browser_view.cc
@@ -36,6 +36,7 @@
 #include "brave/browser/ui/views/frame/brave_contents_view_util.h"
 #include "brave/browser/ui/views/frame/split_view/brave_contents_container_view.h"
 #include "brave/browser/ui/views/frame/split_view/brave_multi_contents_view.h"
+#include "brave/browser/ui/views/frame/tab_strip_placement_coordinator.h"
 #include "brave/browser/ui/views/frame/vertical_tabs/vertical_tab_strip_region_view.h"
 #include "brave/browser/ui/views/frame/vertical_tabs/vertical_tab_strip_widget_delegate_view.h"
 #include "brave/browser/ui/views/location_bar/brave_location_bar_view.h"
@@ -306,6 +307,10 @@ bool BraveBrowserView::ShouldUseBraveWebViewRoundedCornersForContents(
 }
 
 BraveBrowserView::BraveBrowserView(Browser* browser) : BrowserView(browser) {
+  tab_strip_placement_ = std::make_unique<TabStripPlacementCoordinator>(
+      base::PassKey<BraveBrowserView>(), browser,
+      horizontal_tab_strip_region_view_);
+
   // Need this background view always as we have contents margin/rounded corners
   // when split view is active regardless of rounded corners feature.
   contents_background_view_ =
--- a/browser/ui/views/frame/tab_strip_placement_coordinator.cc
+++ b/browser/ui/views/frame/tab_strip_placement_coordinator.cc
@@ -1,0 +1,84 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "brave/browser/ui/views/frame/tab_strip_placement_coordinator.h"
+
+#include "base/check_deref.h"
+#include "brave/browser/ui/focus_mode/focus_mode_utils.h"
+#include "brave/browser/ui/views/tabs/vertical_tab_utils.h"
+#include "ui/views/view.h"
+
+TabStripPlacementCoordinator::TabStripPlacementCoordinator(
+    base::PassKey<BraveBrowserView>,
+    BrowserWindowInterface* browser_window_interface,
+    views::View* tab_strip_region_view)
+    : browser_window_interface_(CHECK_DEREF(browser_window_interface)),
+      tab_strip_region_view_(CHECK_DEREF(tab_strip_region_view)) {
+  auto* original_parent = tab_strip_region_view_->parent();
+  CHECK(original_parent);
+
+  auto original_index =
+      original_parent->GetIndexOf(base::to_address(tab_strip_region_view_));
+
+  SetPlacement(PlacementKind::kDefault, original_parent, original_index);
+}
+
+TabStripPlacementCoordinator::~TabStripPlacementCoordinator() {
+  // On destruction, restore the tabstrip to its original placement so that
+  // browser view teardown runs as expected. Clear all placements other than the
+  // default placement and trigger an update.
+  for (auto& [kind, placement] : placements_) {
+    if (kind != PlacementKind::kDefault) {
+      placement = {};
+    }
+  }
+  UpdatePlacement();
+}
+
+void TabStripPlacementCoordinator::SetPlacement(PlacementKind kind,
+                                                views::View* parent,
+                                                std::optional<size_t> index) {
+  placements_[kind] = {.parent = parent, .index = index};
+}
+
+void TabStripPlacementCoordinator::ClearPlacement(PlacementKind kind) {
+  placements_[kind] = {};
+}
+
+void TabStripPlacementCoordinator::UpdatePlacement() {
+  auto* browser = base::to_address(browser_window_interface_);
+
+  auto get_placement = [&]() -> const Placement& {
+    if (tabs::utils::ShouldShowBraveVerticalTabs(browser)) {
+      auto& placement = placements_[PlacementKind::kVerticalTabStrip];
+      if (placement.parent) {
+        return placement;
+      }
+    }
+    if (IsFocusModeEnabled(browser)) {
+      auto& placement = placements_[PlacementKind::kTopContainer];
+      if (placement.parent) {
+        return placement;
+      }
+    }
+    return placements_[PlacementKind::kDefault];
+  };
+
+  auto placement = get_placement();
+  auto* parent = placement.parent.get();
+
+  if (parent && parent != tab_strip_region_view_->parent()) {
+    // The following remove-then-add sequence is required in order to trigger
+    // AddedToWidget in BraveTabStrip, which calls SetAvailableWidthCallback as
+    // appropriate for the current tab strip orientation.
+    if (tab_strip_region_view_->parent()) {
+      tab_strip_region_view_->parent()->RemoveChildView(
+          base::to_address(tab_strip_region_view_));
+    }
+    size_t max_index = parent->children().size();
+    size_t add_at = std::min(placement.index.value_or(max_index), max_index);
+    parent->AddChildViewAt(base::to_address(tab_strip_region_view_), add_at);
+  }
+}
--- a/browser/ui/views/frame/tab_strip_placement_coordinator.h
+++ b/browser/ui/views/frame/tab_strip_placement_coordinator.h
@@ -1,0 +1,74 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#ifndef BRAVE_BROWSER_UI_VIEWS_FRAME_TAB_STRIP_PLACEMENT_COORDINATOR_H_
+#define BRAVE_BROWSER_UI_VIEWS_FRAME_TAB_STRIP_PLACEMENT_COORDINATOR_H_
+
+#include <optional>
+
+#include "base/containers/flat_map.h"
+#include "base/memory/raw_ptr.h"
+#include "base/memory/raw_ref.h"
+#include "base/types/pass_key.h"
+
+class BraveBrowserView;
+class BrowserWindowInterface;
+
+namespace views {
+class View;
+}  // namespace views
+
+// Coordinates parenting of the tab strip view within the browser view
+// hierarchy, which can depend upon various browser states (e.g. vertical tabs
+// or focus mode). Placement must be triggered explicitly with a call to
+// `UpdatePlacement`.
+class TabStripPlacementCoordinator {
+ public:
+  // Creates an instance for the specified browser view and tab strip region
+  // view. Instances may only be created and owned by BraveBrowserView.
+  TabStripPlacementCoordinator(base::PassKey<BraveBrowserView>,
+                               BrowserWindowInterface* browser_window_interface,
+                               views::View* tab_strip_region_view);
+
+  TabStripPlacementCoordinator(const TabStripPlacementCoordinator&) = delete;
+  TabStripPlacementCoordinator& operator=(const TabStripPlacementCoordinator&) =
+      delete;
+
+  ~TabStripPlacementCoordinator();
+
+  enum class PlacementKind {
+    kDefault = 0,
+    kTopContainer = 1,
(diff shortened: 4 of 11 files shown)


browser/ui/views/frame/tab_strip_placement_coordinator_browsertest.cc
```diff
@@ -0,0 +1,93 @@
...
+  Browser* second = CreateBrowser(browser()->profile());
+  ASSERT_TRUE(second);
+
+  auto* second_view =
+      BraveBrowserView::From(BrowserView::GetBrowserViewForBrowser(second));
+  ASSERT_TRUE(second_view->vertical_tab_strip_widget_delegate_view()->Contains(
+      second_view->tab_strip_view()));
+
+  CloseBrowserSynchronously(second);
+}
```
netzenbot, line 82: `second_view` is dereferenced before any null guard — add `ASSERT_TRUE(second_view)` before the `ASSERT_TRUE(second_view->...)` line. If `BraveBrowserView::From(...)` returns null the test crashes rather than failing gracefully. See [TI-034](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-034).

browser/ui/views/frame/tab_strip_placement_coordinator.h
```diff
@@ -0,0 +1,78 @@
...
+
+ private:
+  struct Placement {
+    raw_ptr<views::View> parent = nullptr;
+    std::optional<size_t> index;
+  };
+
+  bool enabled_ = false;
+  raw_ptr<BrowserWindowInterface> browser_window_interface_;
+  raw_ptr<views::View> tab_strip_region_view_;
```
netzenbot: nit: Both `raw_ptr` members lack explicit `= nullptr` initialization. The existing codebase consistently initializes `raw_ptr` members with `= nullptr` (see surrounding files in this directory), and `raw_ptr`'s default constructor may leave the pointer uninitialized on some build configurations. [CSA-063](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-063)

browser/ui/views/frame/tab_strip_placement_coordinator.cc
```diff
@@ -0,0 +1,99 @@
...
+    }
+    return placements_[PlacementKind::kDefault];
+  };
+
+  auto placement = get_placement();
+  auto* parent = placement.parent.get();
+
+  if (parent && parent != tab_strip_region_view_->parent()) {
+    // The following remove-then-add sequence is currently required in order to
+    // trigger AddedToWidget in BraveTabStrip, which calls
```
netzenbot: nit: The comment describes this remove-then-add sequence as "currently required" to trigger `AddedToWidget`, which implies it's a workaround that could be removed if `BraveTabStrip`'s behavior changes. Per [CSA-053](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-053), temporary workarounds need a `TODO(https://github.com/brave/brave-browser/issues/<id>)` tracking issue. If this is instead a permanent design decision, rephrase the comment to make that clear.

browser/ui/views/frame/tab_strip_placement_coordinator.h
```diff
@@ -0,0 +1,78 @@
...
+  // moving the view under a new parent.
+  void UpdatePlacement();
+
+ private:
+  struct Placement {
+    raw_ptr<views::View> parent = nullptr;
+    std::optional<size_t> index;
+  };
+
+  bool enabled_ = false;
```
netzenbot: nit: `browser_window_interface_` and `tab_strip_region_view_` are verified non-null via `CHECK` in the constructor and never cleared, so `raw_ref<T>` (with `CHECK_DEREF` at the call site) would better express that invariant. See [CSM-036](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-memory.md#CSM-036).

browser/ui/views/frame/tab_strip_placement_coordinator.cc
```diff
@@ -0,0 +1,99 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "brave/browser/ui/views/frame/tab_strip_placement_coordinator.h"
+
+#include "brave/browser/ui/focus_mode/focus_mode_utils.h"
+#include "brave/browser/ui/views/tabs/vertical_tab_utils.h"
+#include "chrome/browser/ui/browser_window/public/browser_window_features.h"
+#include "chrome/browser/ui/browser_window/public/browser_window_interface.h"
```
netzenbot: nit: `browser_window_features.h` doesn't appear to be used in this file. `BrowserWindowInterface` is covered by `browser_window_interface.h`, and no `BrowserWindowFeatures` symbols are referenced. Per [CS-001](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-001), remove includes you don't actually use.

browser/ui/views/frame/tab_strip_placement_coordinator.h
```diff
@@ -0,0 +1,79 @@
...
+
+  TabStripPlacementCoordinator(const TabStripPlacementCoordinator&) = delete;
+  TabStripPlacementCoordinator& operator=(const TabStripPlacementCoordinator&) =
+      delete;
+
+  ~TabStripPlacementCoordinator();
+
+  // Enables or disables tabstrip placement behavior.
+  void SetEnabled(bool enabled);
+
```
netzenbot: nit: The doc for `SetEnabled` doesn't capture its full contract — re-enabling immediately triggers `UpdatePlacement()`, whereas disabling freezes subsequent updates without moving the tab strip. Callers who toggle this flag during teardown or immersive transitions may be surprised. [DOC-004](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-004)

browser/ui/views/frame/tab_strip_placement_coordinator.cc
```diff
@@ -0,0 +1,99 @@
...
+  };
+
+  auto placement = get_placement();
+  auto* parent = placement.parent.get();
+
+  if (parent && parent != tab_strip_region_view_->parent()) {
+    // The following remove-then-add sequence is required in order to trigger
+    // AddedToWidget in BraveTabStrip, which calls SetAvailableWidthCallback as
+    // appropriate for the current tab strip orientation.
+    if (tab_strip_region_view_->parent()) {
```
sangwoo108, line 76: Q. When can the current parent of tab strip region view be null?

sangwoo108 (commented):
LGTM 👍

browser/ui/views/frame/tab_strip_placement_coordinator.cc
zenparsing: It should never be null. I have a slight preference for the defensive `if` branch, but let me know if you prefer a `CHECK`.

sangwoo108: No problem. I don't think it adds much values making this CHECK() and bringing up the case to the surface.

browser/ui/views/frame/brave_browser_view.cc
```diff
@@ -306,6 +307,11 @@ bool BraveBrowserView::ShouldUseBraveWebViewRoundedCornersForContents(
 }
 
 BraveBrowserView::BraveBrowserView(Browser* browser) : BrowserView(browser) {
+  tab_strip_placement_ = std::make_unique<TabStripPlacementCoordinator>(
+      base::PassKey<BraveBrowserView>(), this,
+      horizontal_tab_strip_region_view_);
+  tab_strip_placement_->SetEnabled(true);
```
simonhong: nit: if we want to start with enabled, how about set `true` to TabStripPlacementCoordinator::enabled_ by default?

browser/ui/views/frame/brave_browser_view.cc
```diff
@@ -1030,6 +1036,7 @@ void BraveBrowserView::ReparentTopContainerForEndOfImmersive() {
   }
 
   BrowserView::ReparentTopContainerForEndOfImmersive();
+  tab_strip_placement_->SetEnabled(true);
```
simonhong: It's necessary here?
Curious why `tab_strip_placement_->SetEnabled()` is called from some immersive related calls?
It would be great to have comment why enabled state should be changed. Here and other places.

and don't need to call it from `ReparentTopContainerForStartOfImmersive()`?

browser/ui/views/frame/vertical_tabs/vertical_tab_strip_region_view.cc
```diff
@@ -686,33 +702,24 @@ void BraveVerticalTabStripRegionView::OnBrowserPanelsMoved() {
...
+      return views::LayoutOrientation::kVertical;
     }
+    return views::LayoutOrientation::kHorizontal;
+  };
 
-    static_cast<views::FlexLayout*>(original_region_view_->GetLayoutManager())
-        ->SetOrientation(views::LayoutOrientation::kHorizontal);
-  }
+  static_cast<views::FlexLayout*>(original_region_view_->GetLayoutManager())
+      ->SetOrientation(get_layout_orientation());
```
simonhong: How about simply like this?
Curious we need to check region view's parent view for setting its orientation.
Some place calls `tabs::utils::ShouldShowBraveVerticalTabs(browser_)` and some places checks parent view to know target mode. it would be good to use one source for it?

```suggestion
  const bool show_vertical_tab = tabs::utils::ShouldShowBraveVerticalTabs(browser_);
  if (show_vertical_tab) {
    ReorderChildView(resize_area_, children().size() - 1);
  }

  static_cast<views::FlexLayout*>(original_region_view_->GetLayoutManager())
      ->SetOrientation(show_vertical_tab ? views::LayoutOrientation::kVertical : views::LayoutOrientation:kHorizontal);
```

browser/ui/views/frame/tab_strip_placement_coordinator.cc
```diff
@@ -0,0 +1,95 @@
...
+    BrowserView* browser_view,
+    views::View* tab_strip_region_view)
+    : browser_window_interface_(CHECK_DEREF(browser_view->browser())),
+      tab_strip_region_view_(CHECK_DEREF(tab_strip_region_view)) {
+  CHECK(browser_view->Contains(base::to_address(tab_strip_region_view_)));
+
+  auto* original_parent = tab_strip_region_view_->parent();
+  CHECK(original_parent);
+
+  auto original_index =
```
simonhong, line 22: Q) Before this PR, origianl index was updated whenever region view is reparented to container.
This would not add any regression?

browser/ui/views/frame/tab_strip_placement_coordinator.h
```diff
@@ -0,0 +1,80 @@
...
+      delete;
+
+  ~TabStripPlacementCoordinator();
+
+  // Enables or disables tabstrip placement behavior. When disabled, calls to
+  // UpdatePlacement will have no effect.
+  void SetEnabled(bool enabled);
+
+  enum class PlacementKind {
+    kDefault = 0,
```
simonhong, line 42: How about using `kHorizontal` for more clear?

browser/ui/BUILD.gn
```diff
@@ -641,6 +606,8 @@ source_set("ui") {
       "views/frame/brave_tab_strip_region_view.h",
       "views/frame/brave_window_frame_graphic.cc",
       "views/frame/brave_window_frame_graphic.h",
+      "views/frame/tab_strip_placement_coordinator.cc",
+      "views/frame/tab_strip_placement_coordinator.h",
```
simonhong: How about adding this file to `//brave/browser/ui/views/frame` target?
Upstream(and we also) is going to avoid giant `//chrome/browser/ui` target.

browser/ui/views/frame/tab_strip_placement_coordinator.cc
```diff
@@ -0,0 +1,95 @@
...
+
+#include "base/check_deref.h"
+#include "brave/browser/ui/focus_mode/focus_mode_utils.h"
+#include "brave/browser/ui/views/tabs/vertical_tab_utils.h"
+#include "chrome/browser/ui/views/frame/browser_view.h"
+#include "ui/views/view.h"
+
+TabStripPlacementCoordinator::TabStripPlacementCoordinator(
+    base::PassKey<BraveBrowserView>,
+    BrowserView* browser_view,
```
simonhong: If we remove `BrowserView` dependency, maybe we can have unit test for this coordinator?

browser/ui/views/frame/brave_browser_view.cc
zenparsing: Done

browser/ui/views/frame/brave_browser_view.cc
zenparsing: This call (and the other calls from the immersive functions) is not necessary at this time; I've removed them.

browser/ui/views/frame/vertical_tabs/vertical_tab_strip_region_view.cc
zenparsing: Done

browser/ui/views/frame/tab_strip_placement_coordinator.cc
zenparsing: Switching between various browser states (horizontal/vertical/fullscreen/etc.) I haven't yet seen a regression.

Once we introduce focus mode, we can have a transition like:

```
horizontal (default) -> top container (focus mode) -> vertical
```

Tracking the original index at the time of reparenting will then become quite complex.

browser/ui/views/frame/tab_strip_placement_coordinator.h
zenparsing: I think `kDefault` is more appropriate, because in focus mode we may still be using horizontal tabs, but not the original placement.

browser/ui/BUILD.gn
zenparsing: Adding to the `frame` target will create a circular dependency between 

```
//brave/browser/ui/views/frame/vertical_tabs
```

and

```
//brave/browser/ui/views/frame
```

Is there a separate issue to clean up the `frame` target?

browser/ui/views/frame/tab_strip_placement_coordinator.cc
zenparsing: I've removed `BrowserView*` from the constructor, but since this is essentially an **integration** component, I think that a browser test has more ROI. If we mock out the various parts for unit testing, we'll be left testing mostly trivial code.

netzenbot (approved):

browser/ui/views/frame/tab_strip_placement_coordinator_browsertest.cc
```diff
@@ -0,0 +1,96 @@
...
+    return brave_browser_view()->tab_strip_view();
+  }
+
+  views::View* tab_strip_region_view_parent() {
+    return tab_strip_region_view()->parent();
+  }
+
+  void SetVerticalTabsEnabled(bool enabled) {
+    browser()->profile()->GetPrefs()->SetBoolean(
+        brave_tabs::kVerticalTabsEnabled, enabled);
```
simonhong, line 42: Recommend to call `RunScheduledLayouts()` after toggle to make UI ready before returning from this method.

browser/ui/views/frame/tab_strip_placement_coordinator_browsertest.cc
```diff
@@ -0,0 +1,96 @@
...
+  EXPECT_NE(default_parent, tab_strip_region_view_parent());
+  EXPECT_TRUE(
+      brave_browser_view()->vertical_tab_strip_widget_delegate_view()->Contains(
+          tab_strip_region_view()));
+
+  SetVerticalTabsEnabled(false);
+  EXPECT_EQ(default_parent, tab_strip_region_view_parent());
+}
+
+// SetEnabled(false) freezes placement: pref changes do not move the tab strip.
```
simonhong: Q) do we need this `SetEnabled()` method? Curious when we need to call it with false.
If we don't have that scenario, it would be better to remove that method. WDYT?

browser/ui/views/frame/tab_strip_placement_coordinator_browsertest.cc
zenparsing: Done.

browser/ui/views/frame/tab_strip_placement_coordinator_browsertest.cc
zenparsing: Removed.

simonhong (approved):
++

PR #36261, opened 2026-05-07
Alekspop26 vpn v2 vendor boringtun part5

### LINKED ISSUE #54590: Vendor BoringTun 3p dependency

alekspop26:
This is a multi-part change to add BoringTun dependency to the brave-core, to be used later in the VPN 2.0 architecture by a privileged helper on desktop OSes. The dependency is added to the existing VPN to ensure it actually gets built with the browser, but not yet distributed.

Part 5 (final): build script that actually builds BoringTun dependency in an isolated bundled Rust/C++ environment; BUILD.gn that passes the necessary arguments to the script; GN buildflag 'enable_brave_vpn2_deps' that controls whether VPN 2.0 dependencies are getting built; off by default until we actually start using the dependency.

Resolves https://github.com/brave/brave-browser/issues/54590

### DIFF
--- a/third_party/boringtun/BUILD.gn
+++ b/third_party/boringtun/BUILD.gn
@@ -5,6 +5,7 @@
 
 import("//brave/components/brave_vpn/common/buildflags/buildflags.gni")
 import("//build/config/clang/clang.gni")
+import("//build/config/rust.gni")
 import("//build/config/sysroot.gni")
 
 assert(enable_brave_vpn2_deps,
@@ -44,10 +45,12 @@ action("boringtun") {
     target_cpu,
     "--cargo-target-dir",
     rebase_path("$target_gen_dir/cargo-target", root_build_dir),
+    "--rust-sysroot",
+    rebase_path(rust_sysroot, root_build_dir),
     "--cc-binpath",
     rebase_path("$clang_base_path/bin", root_build_dir),
     "--output-lib",
-    rebase_path("$root_out_dir", root_build_dir),
+    rebase_path(root_out_dir, root_build_dir),
     "--output-headers",
     rebase_path("$target_gen_dir/include", root_build_dir),
     "--depfile",
@@ -71,7 +74,7 @@ action("boringtun") {
     import("//build/config/win/visual_studio_version.gni")
     args += [
       "--cc-sysroot",
-      rebase_path("$visual_studio_path", root_build_dir),
+      rebase_path(visual_studio_path, root_build_dir),
     ]
   } else if (is_mac) {
     args += [
--- a/third_party/boringtun/BUILD.gn
+++ b/third_party/boringtun/BUILD.gn
@@ -1,0 +1,82 @@
+# Copyright (c) 2026 The Brave Authors. All rights reserved.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import("//brave/components/brave_vpn/common/buildflags/buildflags.gni")
+import("//build/config/clang/clang.gni")
+import("//build/config/sysroot.gni")
+
+assert(enable_brave_vpn2_deps,
+       "BoringTun is only buildable when Brave VPN 2.0 deps are enabled")
+
+action("boringtun") {
+  script = "build.py"
+
+  inputs = [
+    ".cargo/config.toml",
+    "Cargo.lock",
+    "Cargo.toml",
+    "src/lib.rs",
+  ]
+
+  if (is_win) {
+    _bt_out_filename = "boringtun.dll"
+  } else if (is_linux) {
+    _bt_out_filename = "libboringtun.so"
+  } else if (is_mac) {
+    _bt_out_filename = "libboringtun.dylib"
+  }
+
+  outputs = [
+    "$root_out_dir/$_bt_out_filename",
+    "$target_gen_dir/include/wireguard_ffi.h",
+    "$target_gen_dir/boringtun.stamp",
+  ]
+
+  depfile = "$target_gen_dir/boringtun.d"
+
+  args = [
+    "--quiet-until-error",
+    "--target-os",
+    target_os,
+    "--target-cpu",
+    target_cpu,
+    "--cargo-target-dir",
+    rebase_path("$target_gen_dir/cargo-target", root_build_dir),
+    "--cc-binpath",
+    rebase_path("$clang_base_path/bin", root_build_dir),
+    "--output-lib",
+    rebase_path("$root_out_dir", root_build_dir),
+    "--output-headers",
+    rebase_path("$target_gen_dir/include", root_build_dir),
+    "--depfile",
+    rebase_path("$target_gen_dir/boringtun.d", root_build_dir),
+    "--stamp",
+    rebase_path("$target_gen_dir/boringtun.stamp", root_build_dir),
+  ]
+
+  if (is_debug) {
+    args += [ "--debug" ]
+  }
+
+  if (!is_win && sysroot != "") {
+    args += [
+      "--cc-sysroot",
+      rebase_path(sysroot, root_build_dir),
+    ]
+  }
+
+  if (is_win) {
+    import("//build/config/win/visual_studio_version.gni")
+    args += [
+      "--cc-sysroot",
+      rebase_path("$visual_studio_path", root_build_dir),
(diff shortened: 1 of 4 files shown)


alekspop26:
Here are Boringtun build times collected from the CI runs:
| CI task | Platform | Builds stdlib? | Time, seconds |
| --- | --- | --- | ---: |
| continuous-integration/windows-x86/pr-head | win x86 | Yes | 11.7 |
| continuous-integration/windows-x64/pr-head | win x64 | No | 11.1 |
| continuous-integration/windows-arm64/pr-head | win arm64 | Yes | 12.1 |
| continuous-integration/origin-windows-x64/pr-head | win x64 | No | 11.2 |
| continuous-integration/origin-linux-x64/pr-head | linux x64 | No | 5.1 |
| continuous-integration/macos-x64/pr-head | mac x64 | No | 19.8 |
| continuous-integration/macos-arm64/pr-head | mac arm64 | No | 5.1 |
| continuous-integration/origin-macos-arm64/pr-head | mac arm64 | No | 11.7 |
| PR (component) | linux x64 | No | 7.6 |
| PR (component) | mac x64 | Yes | 6.2 |
| PR (component) | win x64 | Yes | 9.9 |
| PR (component) | linux-nodchecks x64 | No | 6.1 |
| PR linux (x64) | linux x64 | No | 7.5 |
| PR linux (arm64) | linux arm64 | Yes | 7.6 |

third_party/boringtun/build.py
```diff
@@ -0,0 +1,742 @@
...
+               manifest,
+               triple,
+               target_os,
+               libname,
+               env,
+               *,
+               is_debug,
+               locked,
+               build_std,
+               merged_vendor=None):
```
netzenbot, line 424: nit: `path.write_text(..., encoding='ascii')` on the POSIX branch is missing `newline='\n'`. Without it, Python's default text mode may produce CRLF on Windows hosts, which would break the generated `.sh` wrapper scripts. See [BS-029](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-029).

third_party/boringtun/build.py
```diff
@@ -0,0 +1,744 @@
...
+        cargo_lib = target_dir / triple / profile / lib_filename
+        out = Path(args.output_lib)
+        out.parent.mkdir(parents=True, exist_ok=True)
+        built_lib = out / lib_filename
+        shutil.copy2(cargo_lib, built_lib)
+        log(f'staged -> {built_lib}')
+
+    if args.output_headers:
+        # Extract the FFI header. Vendored sources put it at a fixed path.
+        header = boringtun / 'vendor' / 'boringtun' / 'src' / 'wireguard_ffi.h'
```
netzenbot, line 766: nit: `depfile_path.write_text(..., encoding='utf-8')` is missing `newline='\n'`. On Windows hosts, Python's default text mode will produce CRLF line endings in the depfile, which could cause issues with Make-style depfile parsers. See [BS-029](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-029).

third_party/boringtun/build.py
```diff
@@ -0,0 +1,744 @@
...
+        else:
+            print(line, file=sys.stderr, flush=True)
+
+    def flush(self):
+        if self._buffer:
+            print('\n'.join(self._buffer), file=sys.stderr, flush=True)
+            self._buffer.clear()
+
+
+log = _Log()
```
cdesouza-chromium: Have you considered using python's `logging`?  It is a well known idiom, and you have log levels (error, warning, info, debug, etc).

third_party/boringtun/build.py
```diff
@@ -0,0 +1,744 @@
...
+
+
+def _exe_name(name: str) -> str:
+    """Append '.exe' to name on Windows hosts; return as-is elsewhere."""
+    return name + '.exe' if sys.platform == 'win32' else name
```
cdesouza-chromium: I wonder it would simplify to declare these file names in a consistent way like this:

```python
_ext = ".exe" if sys.platform == 'win32' else ""
CLANG    = f"clang{_ext}"
CLANG_CL = f"clang-cl{_ext}"  # Note: usually clang-cl uses a hyphen
LDD_LINK = f"ldd-link{_ext}"
CARGO    = f"cargo{_ext}"
RUSTC    = f"rustc{_ext}"
```

This way, you can't make a mistake of where to apply the `.exe` suffix.

third_party/boringtun/build.py
```diff
@@ -0,0 +1,744 @@
...
+        return ['-C', 'target-feature=+crt-static']
+    if target_os == 'linux':
+        return ['-C', 'link-arg=-Wl,--no-undefined']
+    if target_os == 'mac':
+        return [
+            '-C',
+            'link-arg=-Wl,-undefined,error',
+            '-C',
+            f'link-arg=-Wl,-install_name,@rpath/{libname}',
+        ]
```
cdesouza-chromium, line 312: These flags belong in the `.gn` file for the action.

cdesouza-chromium (commented):
Thanks for these PRs. We are nearly there. I've left some comments.

third_party/boringtun/build.py
alekspop26: Moving configuration to GN makes sense. But these aren't build configuration in the "knob someone might tune" sense; they're correctness requirements for the artifact:
* `+crt-static` is how the Windows DLL avoids a vcruntime redistributable dependency;
* `--no-undefined` and `-undefined,error` make link-time symbol resolution actually fail rather than defer;
* `@rpath/...` is how the dylib stays relocatable.

No GN config (Origin, debug, official, etc.) would want different values here.

`build.py` is also intentionally self-contained and can be invoked directly outside GN, which is useful for testing vendored-dep bumps and for quick local checks. Keeping these flags in the script means manual invocations produce the same artifact as the GN-driven build.

Hence, I'd prefer to keep these particular flag inside the build script.

third_party/boringtun/build.py
alekspop26: Even with logging, I have to implement that custom "quiet" logic, and the immediate benefit is not clear. However, I switched to the logging module, since the `it's a stdlib idiom` argument makes sense.

third_party/boringtun/build.py
alekspop26: Done.

third_party/boringtun/build.py
cdesouza-chromium: Ack

third_party/boringtun/build.py
cdesouza-chromium: Acknowledged.

cdesouza-chromium (approved):
LGTM, however check with @bridiver too on this one, because I think he had comments open in the previous PR.

third_party/boringtun/build.py
```diff
@@ -0,0 +1,786 @@
...
+
+    if bin_path:
+        extra = str(bin_path)
+        if args.target_os == 'win':
+            shim_dir = _ensure_win_clang_shim(bin_path, wrappers_dir)
+            extra = str(shim_dir) + os.pathsep + extra
+        env['PATH'] = extra + os.pathsep + env['PATH']
+
+    log.info(f'cargo:            {cargo}')
+    log.info(f'rustc:            {env.get("RUSTC", "(default, via PATH)")}')
```
bridiver: `PATH` should not be the default. We should pass the `cargo` and `rustc` paths from chromium. Builds should not require any local tools to be installed except for the ones specific in the brave-core setup

third_party/boringtun/BUILD.gn
```diff
@@ -0,0 +1,82 @@
...
+
+  outputs = [
+    "$root_out_dir/$_bt_out_filename",
+    "$target_gen_dir/include/wireguard_ffi.h",
+    "$target_gen_dir/boringtun.stamp",
+  ]
+
+  depfile = "$target_gen_dir/boringtun.d"
+
+  args = [
```
bridiver, line 40: as mentioned in the build.py script we need to pass the chromium rust toolchain executable paths. This is also necessary to ensure we're always building with the same rust toolchain/version and not just whatever local rustc someone has installed

third_party/boringtun/build.py
alekspop26: Fixed

third_party/boringtun/BUILD.gn
alekspop26: Fixed

bridiver (approved):

PR #36239, opened 2026-05-07
Hide Brave Account UI if Email Aliases is disabled by policy.

boocmp:
We don't want to show the Brave Account UI in Brave Origin if BA feature is not explicitly enabled and Email Aliases is turned off by the preference toggle

PR #36262, opened 2026-05-07
Account/Email Aliases: add profile-scoped enabled predicates

szilardszaloki:
Add per-profile `IsBraveAccountEnabledForProfile()`/`IsEmailAliasesEnabledForProfile()` helpers and switch `brave://settings` to use them, so the Brave Account section hides when the Email Aliases pref is off.

See details [here](https://bravesoftware.slack.com/archives/C071UDBFFT6/p1778134478488459?thread_ts=1778086724.617479&cid=C071UDBFFT6) and [here](https://github.com/brave/brave-core/pull/36239#pullrequestreview-4247575802).

### DIFF
--- a/components/brave_account/features.cc
+++ b/components/brave_account/features.cc
@@ -17,13 +17,26 @@ namespace {
 BASE_FEATURE(kBraveAccount, base::FEATURE_DISABLED_BY_DEFAULT);
 }  // namespace
 
-// Brave Account is enabled when:
-// - the explicit kBraveAccount feature flag is enabled (dev/testing), OR
-// - a dependent feature (e.g. Email Aliases) requires it
+// Process-wide: true if kBraveAccount is on, or a dependent feature (e.g.
+// Email Aliases) is built and feature-flagged on. Use when per-profile state
+// shouldn't or can't influence the answer.
 bool IsBraveAccountEnabled() {
   return base::FeatureList::IsEnabled(kBraveAccount)
 #if BUILDFLAG(ENABLE_EMAIL_ALIASES)
          || email_aliases::features::IsEmailAliasesEnabled()
+#endif
+      ;
+}
+
+// Per-profile: "is Brave Account active for this profile?".
+// Use for UI gating and other per-profile decisions. The result can differ
+// across profiles because dependent features (Email Aliases) consult a
+// per-profile pref.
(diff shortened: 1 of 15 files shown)


components/brave_account/features.cc
```diff
@@ -28,6 +28,19 @@ bool IsBraveAccountEnabled() {
       ;
 }
 
+// Per-profile: "is Brave Account active for this profile?".
+// Use for UI gating and other per-profile decisions. The result can differ
+// across profiles because dependent features (Email Aliases) consult a
+// per-profile pref.
+bool IsBraveAccountEnabledForProfile(const PrefService& pref_service) {
```
ShivanKaul, line 35: We would still need to check against this in browser/ui/webui/brave_settings_ui.cc for Origin, right?

szilardszaloki: The existing `IsBraveAccountEnabledForProfile()` call in `brave_settings_ui.cc` already covers Origin — Origin disables Email Aliases via the same per-profile pref the helper reads (https://github.com/brave/brave-core/blob/master/browser/brave_origin/brave_origin_service_factory.cc#L210-L213).

ShivanKaul (approved):

netzenbot (approved):

boocmp (approved):

goodov (approved):

components/email_aliases/features.h
```diff
@@ -8,11 +8,14 @@
 
 #include "base/feature_list.h"
 
+class PrefService;
+
 namespace email_aliases::features {
 
 BASE_DECLARE_FEATURE(kEmailAliases);
 
 bool IsEmailAliasesEnabled();
```
goodov, line 17: @boocmp as discussed in person, please consider cleaning this up in a follow up. Only one `IsEmailAliasesEnabled*` method should be public, otherwise it will lead to confusion, misuse and bugs.

szilardszaloki: @goodov Could you sketch how you see that working? A lot of the call sites don't have a `Profile` in scope, so I'm not immediately sure how we'd plumb that through cleanly.

goodov: You just check the feature at those places explicitly and that's all.

PR #36194, opened 2026-05-05
Pre-profile adblock service initialization

antonok-edm:
Resolves https://github.com/brave/brave-browser/issues/55278

PR #36263, opened 2026-05-07
Migrate per-profile social media blocking settings

### LINKED ISSUE #55277: Social Media Blocking settings should not be per-profile
The `Social media blocking` settings in `brave://settings/shields` are stored on a per-profile basis, yet they are set globally when toggled. This leads to subtle issues when using multiple profiles - if any settings differ across profiles, the settings of the last-opened profile will be used on all of them.

These settings will need to be migrated into `Local State`. It's unclear which settings to use for the migration if any of them are different, so it was decided to use the exact same logic to select the final settings to be migrated.

Users are encouraged to confirm their settings match what they expected after this migration occurs.

### Pre-upgrade

1. In two or more profiles:
  a. Visit `brave://settings/shields`
  b. Modify some or all of the `Social media blocking` toggles at the bottom of the page
2. Decide on a profile to be the "primary" profile, and note down the settings you toggled for this profile

### Post-upgrade

3. Launch the designated "primary" profile
4. Visit `brave://settings/shields`
5. Confirm that the toggles are the same as what you selected previously
6. Open `brave://settings/shields` in other profiles and confirm the settings match the primary profile
7. Changing any of these settings in 1 profile should change them across all profiles

### Behavioral check

8. In any profile, visit https://example.com
9. Execute the following in the devtools console:
   ```js
   await Promise.all([
     fetch('https://graph.facebook.com'),
     fetch('https://platform.twitter.com'),
     fetch('https://platform.linkedin.com'),
   ])
   ```
10. You should see `net::ERR_BLOCKED_BY_CLIENT` statuses corresponding to platforms for which the toggle is **off**, and `4xx` statuses corresponding to platforms for which the toggle is **on**.

antonok-edm:
Resolves https://github.com/brave/brave-browser/issues/55277

With this PR, we still have `AdBlockPrefService`, but it no longer handles adblock prefs. The relevant code was moved to `AdBlockService` to prevent it from running for each profile. The pref service will be removed entirely in a followup.

may depend on https://github.com/brave/brave-core/pull/36194

### DIFF
--- a/components/brave_shields/content/browser/ad_block_pref_service.cc
+++ b/components/brave_shields/content/browser/ad_block_pref_service.cc
@@ -17,10 +17,12 @@
 
 namespace brave_shields {
 
-AdBlockPrefService::AdBlockPrefService(PrefService* prefs,
+AdBlockPrefService::AdBlockPrefService(bool is_regular_profile,
+                                       PrefService* prefs,
                                        PrefService* local_state,
                                        const std::string& locale) {
-  if (!local_state->GetBoolean(
+  if (is_regular_profile &&
+      !local_state->GetBoolean(
           prefs::kMigratedAdblockSocialMediaBlockingSettings)) {
     // Yes, we're reading these on the first profile launched, and migrating
     // them to Local State. It's not ideal, but that's how the settings have
--- a/components/brave_shields/content/browser/ad_block_pref_service.cc
+++ b/components/brave_shields/content/browser/ad_block_pref_service.cc
@@ -8,42 +8,18 @@
 #include <utility>
 
 #include "base/check.h"
-#include "base/functional/bind.h"
-#include "brave/components/brave_shields/content/browser/ad_block_service.h"
 #include "brave/components/brave_shields/core/browser/brave_shields_locale_utils.h"
-#include "brave/components/brave_shields/core/common/brave_shield_constants.h"
 #include "brave/components/brave_shields/core/common/pref_names.h"
 #include "components/prefs/pref_change_registrar.h"
 #include "components/prefs/pref_service.h"
 #include "components/proxy_config/pref_proxy_config_tracker.h"
 #include "components/user_prefs/user_prefs.h"
-#include "content/public/browser/browser_context.h"
 
 namespace brave_shields {
 
-namespace {
-
-std::string GetTagFromPrefName(const std::string& pref_name) {
-  if (pref_name == prefs::kFBEmbedControlType) {
-    return brave_shields::kFacebookEmbeds;
-  }
-  if (pref_name == prefs::kTwitterEmbedControlType) {
-    return brave_shields::kTwitterEmbeds;
-  }
-  if (pref_name == prefs::kLinkedInEmbedControlType) {
-    return brave_shields::kLinkedInEmbeds;
-  }
-  return "";
-}
-
-}  // namespace
-
-AdBlockPrefService::AdBlockPrefService(AdBlockService* ad_block_service,
-                                       PrefService* prefs,
+AdBlockPrefService::AdBlockPrefService(PrefService* prefs,
                                        PrefService* local_state,
-                                       const std::string& locale)
-    : ad_block_service_(ad_block_service), prefs_(prefs) {
-  pref_change_registrar_.reset(new PrefChangeRegistrar());
+                                       const std::string& locale) {
   if (!local_state->GetBoolean(
           prefs::kMigratedAdblockSocialMediaBlockingSettings)) {
     // Yes, we're reading these on the first profile launched, and migrating
@@ -59,24 +35,6 @@ AdBlockPrefService::AdBlockPrefService(AdBlockService* ad_block_service,
     local_state->SetBoolean(prefs::kMigratedAdblockSocialMediaBlockingSettings,
                             true);
   }
-  pref_change_registrar_->Init(prefs_);
-  pref_change_registrar_->Add(
-      prefs::kFBEmbedControlType,
-      base::BindRepeating(&AdBlockPrefService::OnPreferenceChanged,
-                          base::Unretained(this), prefs::kFBEmbedControlType));
-  pref_change_registrar_->Add(
-      prefs::kTwitterEmbedControlType,
-      base::BindRepeating(&AdBlockPrefService::OnPreferenceChanged,
-                          base::Unretained(this),
-                          prefs::kTwitterEmbedControlType));
-  pref_change_registrar_->Add(
-      prefs::kLinkedInEmbedControlType,
-      base::BindRepeating(&AdBlockPrefService::OnPreferenceChanged,
-                          base::Unretained(this),
-                          prefs::kLinkedInEmbedControlType));
-  OnPreferenceChanged(prefs::kFBEmbedControlType);
-  OnPreferenceChanged(prefs::kTwitterEmbedControlType);
-  OnPreferenceChanged(prefs::kLinkedInEmbedControlType);
 
   ManageAdBlockOnlyModeByLocale(local_state, locale);
 }
@@ -107,8 +65,6 @@ AdBlockPrefService::GetLatestProxyConfig(
 }
 
 void AdBlockPrefService::Shutdown() {
-  pref_change_registrar_.reset();
-
   // `pref_proxy_config_tracker_` has a reference to `proxy_config_service_`,
   // therefore detach `pref_proxy_config_tracker_` first, to prevent the
   // reference from dangling.
@@ -122,15 +78,6 @@ void AdBlockPrefService::Shutdown() {
     proxy_config_service_->RemoveObserver(this);
     proxy_config_service_.reset();
   }
-}
-
-void AdBlockPrefService::OnPreferenceChanged(const std::string& pref_name) {
-  std::string tag = GetTagFromPrefName(pref_name);
-  if (tag.length() == 0) {
-    return;
-  }
-  bool enabled = prefs_->GetBoolean(pref_name);
-  ad_block_service_->EnableTag(tag, enabled);
 }
 
 void AdBlockPrefService::OnProxyConfigChanged(
--- a/components/brave_shields/content/browser/ad_block_pref_service.cc
+++ b/components/brave_shields/content/browser/ad_block_pref_service.cc
@@ -25,11 +25,15 @@ AdBlockPrefService::AdBlockPrefService(PrefService* prefs,
     // Yes, we're reading these on the first profile launched, and migrating
     // them to Local State. It's not ideal, but that's how the settings have
     // existed for a long time.
-    for (const auto& pref :
-         {prefs::kFBEmbedControlType, prefs::kTwitterEmbedControlType,
-          prefs::kLinkedInEmbedControlType}) {
-      if (prefs->FindPreference(pref)->HasUserSetting()) {
-        local_state->SetBoolean(pref, prefs->GetBoolean(pref));
+    for (const auto& [old_pref, new_pref] :
+         {std::pair{prefs::kLegacyFBEmbedControlType,
+                    prefs::kFBEmbedControlType},
+          std::pair{prefs::kLegacyTwitterEmbedControlType,
+                    prefs::kTwitterEmbedControlType},
+          std::pair{prefs::kLegacyLinkedInEmbedControlType,
+                    prefs::kLinkedInEmbedControlType}}) {
+      if (prefs->FindPreference(old_pref)->HasUserSetting()) {
+        local_state->SetBoolean(new_pref, prefs->GetBoolean(old_pref));
       }
     }
     local_state->SetBoolean(prefs::kMigratedAdblockSocialMediaBlockingSettings,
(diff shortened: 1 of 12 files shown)


components/brave_shields/content/browser/ad_block_service.cc
```diff
@@ -173,6 +191,26 @@ AdBlockService::AdBlockService(
   // Initializes adblock-rust's domain resolution implementation
   adblock::set_domain_resolver();
 
+  pref_change_registrar_.reset(new PrefChangeRegistrar());
+  pref_change_registrar_->Init(local_state_);
+  pref_change_registrar_->Add(
+      prefs::kFBEmbedControlType,
+      base::BindRepeating(&AdBlockService::OnPreferenceChanged,
+                          base::Unretained(this), prefs::kFBEmbedControlType));
```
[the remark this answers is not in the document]
antonok-edm: this is just moved code

antonok-edm:
> `Discovered possible removal of preference registrations.`

Presubmit warning is flagging a renamed pref variable, it's still registered

components/brave_shields/content/browser/ad_block_service.cc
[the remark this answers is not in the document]
bridiver: @thypon @goodov maybe we should remove this from semgrep and add it as a best practice since the bot can be a bit smarter about whether this is ok in context? wdyt @bbondy ?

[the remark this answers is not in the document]
bridiver: this is one of the "ok" examples above for PrefChangeRegistrar so it's generating a lot of noise and I don't know that anyone is actually following up on these?

components/brave_shields/content/browser/ad_block_pref_service.cc
```diff
@@ -8,60 +8,40 @@
...
-  OnPreferenceChanged(prefs::kFBEmbedControlType);
-  OnPreferenceChanged(prefs::kTwitterEmbedControlType);
-  OnPreferenceChanged(prefs::kLinkedInEmbedControlType);
+    : prefs_(prefs) {
+  if (is_regular_profile &&
+      !local_state->GetBoolean(
+          prefs::kMigratedAdblockSocialMediaBlockingSettings)) {
+    // Yes, we're reading these on the first profile launched, and migrating
+    // them to Local State. It's not ideal, but that's how the settings have
+    // existed for a long time.
```
netzenbot, line 39: The migration moves these settings from profile prefs to `local_state`, making them global across all profiles. For a multi-profile setup this means changing social media blocking in one profile affects all others. The code comment itself says "it's not ideal" — if these are genuinely per-profile settings, they should remain profile-scoped per [ARCH-057](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-057). Is there a plan to properly scope these per-profile, or is the global behavior intentional?

components/brave_shields/content/browser/ad_block_pref_service.cc
```diff
@@ -8,60 +8,40 @@
...
+      if (prefs->FindPreference(old_pref)->HasUserSetting()) {
+        local_state->SetBoolean(new_pref, prefs->GetBoolean(old_pref));
+      }
+    }
+    local_state->SetBoolean(prefs::kMigratedAdblockSocialMediaBlockingSettings,
+                            true);
+  }
 
   ManageAdBlockOnlyModeByLocale(local_state, locale);
 }
```
netzenbot, line 55: The migration code that reads legacy profile prefs and writes them to local state is new logic with no test coverage. Since this migration runs only once per user and is irreversible, a bug here would silently reset users' custom settings to defaults. Consider adding a browser test that sets an old pref (e.g., `kLegacyFBEmbedControlType`) to a non-default value, triggers migration, and verifies the new pref (`kFBEmbedControlType` in local state) reflects the old value. ([PRs Should Include Reasonable Test Coverage](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-039))

components/brave_shields/content/browser/ad_block_pref_service.cc
antonok-edm: global behavior is required because the adblock engine is globally scoped

components/brave_shields/content/browser/ad_block_service.cc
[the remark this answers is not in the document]
bridiver: /ar PrefChangeRegistrar

components/brave_shields/content/browser/ad_block_service.cc
```diff
@@ -173,6 +191,26 @@ AdBlockService::AdBlockService(
...
+  pref_change_registrar_->Add(
+      prefs::kTwitterEmbedControlType,
+      base::BindRepeating(&AdBlockService::OnPreferenceChanged,
+                          base::Unretained(this),
+                          prefs::kTwitterEmbedControlType));
+  pref_change_registrar_->Add(
+      prefs::kLinkedInEmbedControlType,
+      base::BindRepeating(&AdBlockService::OnPreferenceChanged,
+                          base::Unretained(this),
```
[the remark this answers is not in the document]
bridiver: /ar PrefChangeRegistrar

components/brave_shields/content/browser/ad_block_pref_service.cc
```diff
@@ -8,60 +8,40 @@
...
+    // Yes, we're reading these on the first profile launched, and migrating
+    // them to Local State. It's not ideal, but that's how the settings have
+    // existed for a long time.
+    for (const auto& [old_pref, new_pref] :
+         {std::pair{prefs::kLegacyFBEmbedControlType,
+                    prefs::kFBEmbedControlType},
+          std::pair{prefs::kLegacyTwitterEmbedControlType,
+                    prefs::kTwitterEmbedControlType},
+          std::pair{prefs::kLegacyLinkedInEmbedControlType,
+                    prefs::kLinkedInEmbedControlType}}) {
```
netzenbot, line 45: `PrefService::FindPreference()` returns `nullptr` if the preference is not registered. Calling `->HasUserSetting()` directly on the result without a null check will crash if any of the legacy prefs aren't registered in a profile. Consider using `if (auto* pref = prefs->FindPreference(old_pref); pref && pref->HasUserSetting())` instead. See [CSM-029](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-memory.md#CSM-029).

components/brave_shields/content/browser/ad_block_service.cc
```diff
@@ -363,6 +401,15 @@ void AdBlockService::RemoveObserver(Observer* observer) {
   observers_.RemoveObserver(observer);
```
netzenbot, line 402: nit: `OnPreferenceChanged` is bound via `base::BindRepeating` but is missing `DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_)`, which every other non-trivial method in this class has. Per [CSM-019](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-memory.md#CSM-019), callback targets should include the sequence check.

components/brave_shields/content/browser/ad_block_service.h
```diff
@@ -34,6 +34,7 @@
 #include "brave/components/brave_shields/core/browser/ad_block_list_p3a.h"
 #include "brave/components/brave_shields/core/browser/ad_block_resource_provider.h"
 #include "brave/components/brave_shields/core/common/adblock/rs/src/lib.rs.h"
+#include "components/prefs/pref_change_registrar.h"
```
netzenbot: nit: `PrefChangeRegistrar` is only held as `std::unique_ptr<PrefChangeRegistrar>` in this header, and `~AdBlockService()` is defined in the `.cc` file, so a forward declaration (`class PrefChangeRegistrar;`) in the header and the full `#include "components/prefs/pref_change_registrar.h"` in the `.cc` file would be sufficient. ([CS-014](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-014))

components/brave_shields/content/browser/ad_block_pref_service.cc
bridiver: can we use `MigrateObsoleteProfilePrefs` for this?

components/brave_shields/core/common/pref_names.h
```diff
@@ -27,11 +27,21 @@ inline constexpr char kAdBlockListSubscriptions[] =
...
+    "brave.migrated_adblock_social_media_blocking_settings";
+// Migrated settings (local state)
+inline constexpr char kFBEmbedControlType[] = "brave.shields.fb_embed_default";
 inline constexpr char kTwitterEmbedControlType[] =
-    "brave.twitter_embed_default";
+    "brave.shields.twitter_embed_default";
 inline constexpr char kLinkedInEmbedControlType[] =
+    "brave.shields.linkedin_embed_default";
+// Legacy settings (profile preferences)
+inline constexpr char kLegacyFBEmbedControlType[] = "brave.fb_embed_default";
```
bridiver: it's better to define these only where we need to migrate them

components/brave_shields/content/browser/ad_block_pref_service.cc
```diff
@@ -8,60 +8,40 @@
...
+    : prefs_(prefs) {
+  if (is_regular_profile &&
+      !local_state->GetBoolean(
+          prefs::kMigratedAdblockSocialMediaBlockingSettings)) {
+    // Yes, we're reading these on the first profile launched, and migrating
+    // them to Local State. It's not ideal, but that's how the settings have
+    // existed for a long time.
+    for (const auto& [old_pref, new_pref] :
+         {std::pair{prefs::kLegacyFBEmbedControlType,
+                    prefs::kFBEmbedControlType},
```
netzenbot: nit: The phrase `"that's how the settings have existed for a long time"` references historical prior behavior rather than explaining current code. Per [CS-019](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-019), comments should be meaningful to someone reading the code for the first time with no knowledge of the PR or change history. Consider replacing with something that explains the architectural reason: e.g., "The adblock engine is globally scoped, so we use the first regular profile's settings as the global default."

components/brave_shields/content/browser/ad_block_pref_service.h
```diff
@@ -21,12 +21,10 @@ class PrefProxyConfigTracker;
...
 
-class AdBlockService;
-
 class AdBlockPrefService : public KeyedService,
                            public net::ProxyConfigService::Observer {
  public:
-  explicit AdBlockPrefService(AdBlockService* ad_block_service,
+  explicit AdBlockPrefService(bool is_regular_profile,
                               PrefService* prefs,
                               PrefService* local_state,
```
netzenbot, line 31: nit: After the pref-change logic moves to `AdBlockService`, both `pref_change_registrar_` and `prefs_` become dead in `AdBlockPrefService` — `pref_change_registrar_` is never initialized, and `prefs_` is stored but never read. Consider removing both members (and the corresponding `raw_ptr<PrefService> prefs_` member initialization in the constructor initializer list). ([CS-068](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-068))

components/brave_shields/content/browser/ad_block_pref_service.cc
antonok-edm: I think this missed the point of CS-019... this is not explaining removed code, it's explaining migration logic which inherently requires historical context

components/brave_shields/content/browser/ad_block_pref_service.cc
antonok-edm: I don't think so... we can't do the migration until we encounter a profile for which `Profile::IsRegularProfile()` is true. `Profile` isn't accessible from `MigrateObsoleteProfilePrefs`.

bridiver: `MigrateObsoleteProfilePrefs` only gets called for regular profiles.

components/brave_shields/content/browser/ad_block_pref_service.cc
```diff
@@ -8,60 +8,48 @@
...
-#include "brave/components/brave_shields/core/common/brave_shield_constants.h"
 #include "brave/components/brave_shields/core/common/pref_names.h"
+#include "components/pref_registry/pref_registry_syncable.h"
 #include "components/prefs/pref_change_registrar.h"
 #include "components/prefs/pref_service.h"
 #include "components/proxy_config/pref_proxy_config_tracker.h"
 #include "components/user_prefs/user_prefs.h"
-#include "content/public/browser/browser_context.h"
-
-namespace brave_shields {
```
netzenbot, line 18: nit: The anonymous `namespace {` block with the legacy pref name constants is placed outside `namespace brave_shields` in the new file structure. Per [CSA-025](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-025), constants in anonymous namespaces should be placed inside the component's namespace. Consider moving the `namespace {` block inside `namespace brave_shields {` (or defining the constants directly in the `brave_shields` namespace as file-local `constexpr` after the namespace opens).

bridiver (approved):

components/brave_shields/content/browser/ad_block_pref_service.cc
antonok-edm: we usually don't do this for prefs since they are registered and can be assumed-initialized

components/brave_shields/content/browser/ad_block_pref_service.cc
antonok-edm: for posterity: it does also get called on the system profile, which is problematic here

PR #36265, opened 2026-05-07
[FocusMode] Add FocusModeTitleBar

zenparsing:
Small title bar shown at the top of the browser window in Focus Mode. Displays the active tab's favicon and a formatted host string sourced from `TabUIHelper`.

Security review: https://github.com/brave/reviews/issues/2215

Screenshot 2026-05-07 at 6 57 37 PM

### DIFF
--- a/browser/ui/views/frame/BUILD.gn
+++ b/browser/ui/views/frame/BUILD.gn
@@ -7,6 +7,7 @@ source_set("frame") {
   sources = [
     "brave_contents_view_util.h",
     "brave_side_panel_shadow_overlay_view.h",
+    "focus_mode_title_bar_view.h",
     "layout/brave_browser_view_layout_delegate_impl.cc",
     "layout/brave_browser_view_layout_delegate_impl.h",
     "layout/brave_browser_view_tabbed_layout_impl.cc",
@@ -16,6 +17,7 @@ source_set("frame") {
 
   public_deps = [
     "//base",
+    "//components/tabs:public",
     "//ui/views",
   ]
 
@@ -32,6 +34,7 @@ source_set("frame_impl") {
   sources = [
     "brave_contents_view_util.cc",
     "brave_side_panel_shadow_overlay_view.cc",
+    "focus_mode_title_bar_view.cc",
     "tab_strip_placement_coordinator.cc",
   ]
 
@@ -46,13 +49,16 @@ source_set("frame_impl") {
     "//chrome/common:constants",
     "//components/prefs",
     "//components/split_tabs",
-    "//components/tabs:public",
   ]
 }
 
 source_set("unit_tests") {
   testonly = true
-  sources = [ "layout/brave_browser_view_tabbed_layout_impl_unittest.cc" ]
+
+  sources = [
+    "focus_mode_title_bar_view_unittest.cc",
+    "layout/brave_browser_view_tabbed_layout_impl_unittest.cc",
+  ]
 
   if (is_linux) {
     sources += [ "browser_frame_view_layout_linux_unittest.cc" ]
@@ -60,11 +66,19 @@ source_set("unit_tests") {
 
   deps = [
     ":frame",
+    ":frame_impl",
     "//base",
     "//base/test:test_support",
     "//chrome/browser/ui",
+    "//chrome/test:test_support",
+    "//components/tabs:test_support",
+    "//content/test:test_support",
     "//testing/gmock",
     "//testing/gtest",
+    "//ui/base",
+    "//ui/gfx",
+    "//ui/views:test_support",
+    "//url",
   ]
 }
 
--- a/browser/ui/views/frame/focus_mode_title_bar_view.cc
+++ b/browser/ui/views/frame/focus_mode_title_bar_view.cc
@@ -1,0 +1,143 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "brave/browser/ui/views/frame/focus_mode_title_bar_view.h"
+
+#include <string>
+
+#include "base/functional/bind.h"
+#include "brave/browser/ui/brave_scheme_utils.h"
+#include "brave/ui/color/nala/nala_color_id.h"
+#include "chrome/browser/ui/color/chrome_color_id.h"
+#include "chrome/browser/ui/tab_ui_helper.h"
+#include "components/url_formatter/url_formatter.h"
+#include "ui/base/metadata/metadata_impl_macros.h"
+#include "ui/base/models/image_model.h"
+#include "ui/color/color_provider.h"
+#include "ui/gfx/font_list.h"
+#include "ui/gfx/geometry/size.h"
+#include "ui/gfx/image/image_skia_operations.h"
+#include "ui/views/background.h"
+#include "ui/views/controls/image_view.h"
+#include "ui/views/controls/label.h"
+#include "ui/views/layout/box_layout.h"
+#include "url/gurl.h"
+
+namespace {
+
+constexpr int kTitleBarHeight = 20;
+constexpr int kFaviconSize = 12;
+constexpr int kFaviconLabelSpacing = 4;
+constexpr int kLabelFontSize = 11;
+
+gfx::FontList GetLabelFont() {
+  const gfx::FontList base;
+  return base.DeriveWithSizeDelta(kLabelFontSize - base.GetFontSize());
+}
+
+}  // namespace
+
+FocusModeTitleBarView::FocusModeTitleBarView() {
+  auto* layout = SetLayoutManager(std::make_unique<views::BoxLayout>(
+      views::BoxLayout::Orientation::kHorizontal, gfx::Insets(),
+      kFaviconLabelSpacing));
+  SetBackground(views::CreateSolidBackground(kColorToolbar));
+
+  layout->set_main_axis_alignment(views::BoxLayout::MainAxisAlignment::kCenter);
+  layout->set_cross_axis_alignment(
+      views::BoxLayout::CrossAxisAlignment::kCenter);
+
+  SetPreferredSize(gfx::Size(0, kTitleBarHeight));
+
+  favicon_image_ = AddChildView(std::make_unique<views::ImageView>());
+  favicon_image_->SetImageSize(gfx::Size(kFaviconSize, kFaviconSize));
+  favicon_image_->SetVisible(false);
+
+  domain_label_ = AddChildView(std::make_unique<views::Label>(
+      u"", views::Label::CustomFont(GetLabelFont())));
+  domain_label_->SetDirectionalityMode(
+      gfx::DirectionalityMode::DIRECTIONALITY_AS_URL);
+  domain_label_->SetElideBehavior(gfx::ELIDE_TAIL);
+  domain_label_->SetEnabledColor(nala::kColorTextTertiary);
+  domain_label_->SetAutoColorReadabilityEnabled(false);
+}
+
+FocusModeTitleBarView::~FocusModeTitleBarView() = default;
+
+void FocusModeTitleBarView::SetTab(tabs::TabInterface* tab) {
+  tab_ui_updated_subscription_ = {};
+  tab_will_detach_subscription_ = {};
+  tab_ = tab;
+
+  if (tab_) {
+    tab_will_detach_subscription_ =
+        tab_->RegisterWillDetach(base::BindRepeating(
+            &FocusModeTitleBarView::OnTabWillDetach, base::Unretained(this)));
+    if (auto* helper = TabUIHelper::From(tab_)) {
+      tab_ui_updated_subscription_ =
+          helper->AddTabUIChangeCallback(base::BindRepeating(
+              &FocusModeTitleBarView::Update, base::Unretained(this)));
+    }
+  }
+
+  Update();
+}
+
+bool FocusModeTitleBarView::IsFaviconVisibleForTesting() const {
+  return favicon_image_->GetVisible();
+}
+
+void FocusModeTitleBarView::Update() {
+  auto* tab_ui_helper = tab_ ? TabUIHelper::From(tab_) : nullptr;
+  if (!tab_ui_helper) {
+    favicon_image_->SetImage(ui::ImageModel());
+    favicon_image_->SetVisible(false);
+    domain_label_->SetText(u"");
+    return;
+  }
+
+  GURL domain_url = tab_ui_helper->GetVisibleURL();
+  std::u16string domain;
+  if (tab_ui_helper->ShouldDisplayURL()) {
+    domain = url_formatter::FormatUrl(
+        domain_url,
+        (url_formatter::kFormatUrlOmitDefaults &
+         ~url_formatter::kFormatUrlOmitHTTP) |
+            url_formatter::kFormatUrlOmitTrivialSubdomains |
+            url_formatter::kFormatUrlOmitHTTPS |
+            url_formatter::kFormatUrlTrimAfterHost,
+        base::UnescapeRule::SPACES, nullptr, nullptr, nullptr);
+    brave_utils::ReplaceChromeToBraveScheme(domain);
+  }
+
+  domain_label_->SetText(domain);
+
+  if (ui::ImageModel favicon = tab_ui_helper->GetFavicon();
+      !favicon.IsEmpty() && !domain.empty()) {
+    bool themify_favicon = tab_ui_helper->ShouldThemifyFavicon();
+    if (auto* provider = GetColorProvider(); provider && themify_favicon) {
+      SkColor favicon_color = provider->GetColor(kColorBookmarkFavicon);
+      if (favicon_color != SK_ColorTRANSPARENT) {
+        favicon = ui::ImageModel::FromImageSkia(
+            gfx::ImageSkiaOperations::CreateColorMask(
+                favicon.Rasterize(provider), favicon_color));
+      }
+    }
+    favicon_image_->SetImage(favicon);
+    favicon_image_->SetVisible(true);
+  } else {
+    favicon_image_->SetImage(ui::ImageModel());
+    favicon_image_->SetVisible(false);
+  }
+}
+
+void FocusModeTitleBarView::OnTabWillDetach(
+    tabs::TabInterface* tab,
+    tabs::TabInterface::DetachReason reason) {
+  SetTab(nullptr);
+}
+
+BEGIN_METADATA(FocusModeTitleBarView)
+END_METADATA
--- a/browser/ui/views/frame/focus_mode_title_bar_view.h
+++ b/browser/ui/views/frame/focus_mode_title_bar_view.h
@@ -1,0 +1,53 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#ifndef BRAVE_BROWSER_UI_VIEWS_FRAME_FOCUS_MODE_TITLE_BAR_VIEW_H_
+#define BRAVE_BROWSER_UI_VIEWS_FRAME_FOCUS_MODE_TITLE_BAR_VIEW_H_
+
+#include "base/callback_list.h"
+#include "base/memory/raw_ptr.h"
+#include "components/tabs/public/tab_interface.h"
+#include "ui/base/metadata/metadata_header_macros.h"
+#include "ui/gfx/text_constants.h"
+#include "ui/views/view.h"
+
+namespace views {
+class ImageView;
+class Label;
+}  // namespace views
+
+// Small title bar shown at the top of the browser window when Focus Mode is
+// enabled. Displays the active tab's favicon followed by the formatted host
+// (omits https://, trivial subdomains, and trims after the host).
+class FocusModeTitleBarView : public views::View {
+  METADATA_HEADER(FocusModeTitleBarView, views::View)
+
+ public:
+  FocusModeTitleBarView();
+  FocusModeTitleBarView(const FocusModeTitleBarView&) = delete;
+  FocusModeTitleBarView& operator=(const FocusModeTitleBarView&) = delete;
+  ~FocusModeTitleBarView() override;
+
+  // Points this view at the given tab. `tab` may be null, in which case the
+  // display is cleared. The view automatically clears itself if `tab` is
+  // destroyed.
+  void SetTab(tabs::TabInterface* tab);
+
+  bool IsFaviconVisibleForTesting() const;
+  const views::Label* domain_label_for_testing() const { return domain_label_; }
+
+ private:
+  void Update();
+  void OnTabWillDetach(tabs::TabInterface* tab,
+                       tabs::TabInterface::DetachReason reason);
+
+  base::CallbackListSubscription tab_will_detach_subscription_;
+  base::CallbackListSubscription tab_ui_updated_subscription_;
+  raw_ptr<views::ImageView> favicon_image_ = nullptr;
+  raw_ptr<views::Label> domain_label_ = nullptr;
+  raw_ptr<tabs::TabInterface> tab_ = nullptr;
+};
+
+#endif  // BRAVE_BROWSER_UI_VIEWS_FRAME_FOCUS_MODE_TITLE_BAR_VIEW_H_
--- a/browser/ui/views/frame/focus_mode_title_bar_view_unittest.cc
+++ b/browser/ui/views/frame/focus_mode_title_bar_view_unittest.cc
@@ -1,0 +1,164 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "brave/browser/ui/views/frame/focus_mode_title_bar_view.h"
+
+#include <memory>
+#include <string>
+#include <utility>
+
+#include "base/callback_list.h"
+#include "chrome/browser/profiles/profile.h"
+#include "chrome/browser/ui/tab_ui_helper.h"
+#include "chrome/browser/ui/tabs/public/tab_features.h"
+#include "chrome/browser/ui/views/chrome_layout_provider.h"
+#include "chrome/test/base/testing_profile.h"
+#include "chrome/test/views/chrome_views_test_base.h"
+#include "components/tabs/public/mock_tab_interface.h"
+#include "components/tabs/public/tab_interface.h"
+#include "content/public/browser/web_contents.h"
+#include "content/public/test/test_renderer_host.h"
+#include "content/public/test/web_contents_tester.h"
+#include "testing/gmock/include/gmock/gmock.h"
+#include "testing/gtest/include/gtest/gtest.h"
+#include "ui/base/unowned_user_data/unowned_user_data_host.h"
+#include "ui/gfx/geometry/size.h"
+#include "ui/gfx/text_constants.h"
+#include "ui/views/controls/label.h"
+#include "ui/views/test/views_test_utils.h"
+#include "url/gurl.h"
+
+namespace {
+
+class FakeTab : public tabs::MockTabInterface {
+ public:
+  explicit FakeTab(Profile* profile)
+      : tab_features_(std::make_unique<tabs::TabFeatures>()),
+        contents_(content::WebContentsTester::CreateTestWebContents(profile,
+                                                                    nullptr)) {
+    using ::testing::Return;
+    using ::testing::ReturnRef;
+
+    ON_CALL(*this, GetContents()).WillByDefault(Return(contents_.get()));
+    ON_CALL(*this, GetTabFeatures()).WillByDefault(Return(tab_features_.get()));
+    ON_CALL(*this, GetUnownedUserDataHost()).WillByDefault(ReturnRef(host_));
+    ON_CALL(*this, RegisterWillDetach(::testing::_))
+        .WillByDefault([this](tabs::TabInterface::WillDetach callback) {
+          return will_detach_callbacks_.Add(std::move(callback));
+        });
+
+    tab_features_->SetTabUIHelperForTesting(
+        std::make_unique<TabUIHelper>(*this));
+  }
+
+  ~FakeTab() override {
+    will_detach_callbacks_.Notify(this,
+                                  tabs::TabInterface::DetachReason::kDelete);
+  }
+
+  void NavigateAndCommit(const GURL& url) {
+    content::WebContentsTester::For(contents_.get())->NavigateAndCommit(url);
+  }
+
+ private:
+  ui::UnownedUserDataHost host_;
+  std::unique_ptr<tabs::TabFeatures> tab_features_;
+  std::unique_ptr<content::WebContents> contents_;
+  base::RepeatingCallbackList<void(tabs::TabInterface*,
+                                   tabs::TabInterface::DetachReason)>
+      will_detach_callbacks_;
+};
+
+class FocusModeTitleBarViewTest : public ChromeViewsTestBase {
+ public:
+  void SetUp() override {
+    ChromeViewsTestBase::SetUp();
+    view_ = std::make_unique<FocusModeTitleBarView>();
+  }
+
+  void TearDown() override {
+    view_.reset();
+    ChromeViewsTestBase::TearDown();
+  }
+
+ protected:
+  std::unique_ptr<FakeTab> AddTab(const GURL& url) {
+    auto tab = std::make_unique<FakeTab>(&profile_);
+    tab->NavigateAndCommit(url);
+    return tab;
+  }
+
+  std::u16string_view GetDomainText() {
+    return view_->domain_label_for_testing()->GetText();
+  }
+
+  std::u16string_view AddTabAndGetDomainText(const GURL& gurl) {
+    auto tab = AddTab(gurl);
+    view_->SetTab(tab.get());
+    return GetDomainText();
+  }
+
+  std::unique_ptr<FocusModeTitleBarView> view_;
+  TestingProfile profile_;
+  content::RenderViewHostTestEnabler render_view_host_test_enabler_;
+};
+
+TEST_F(FocusModeTitleBarViewTest, UpdatesOnNavigation) {
+  auto tab = AddTab(GURL("https://www.example.com/some/path?q=1"));
+  view_->SetTab(tab.get());
+  EXPECT_EQ(GetDomainText(), u"example.com");
+
+  tab->NavigateAndCommit(GURL("https://docs.example.com/title1.html"));
+  EXPECT_EQ(GetDomainText(), u"docs.example.com");
+
+  tab->NavigateAndCommit(GURL("http://insecure.example.com/"));
+  EXPECT_EQ(GetDomainText(), u"http://insecure.example.com");
+
+  tab->NavigateAndCommit(GURL("chrome://abc"));
+  EXPECT_EQ(GetDomainText(), u"brave://abc");
+
+  tab->NavigateAndCommit(GURL("chrome-untrusted://print"));
+  EXPECT_EQ(GetDomainText(), u"chrome-untrusted://print");
+}
+
+TEST_F(FocusModeTitleBarViewTest, ClearsWhenTabDetaches) {
+  auto tab = AddTab(GURL("https://www.example.com/title1.html"));
+  view_->SetTab(tab.get());
+  ASSERT_EQ(GetDomainText(), u"example.com");
+
+  tab.reset();
+  EXPECT_TRUE(GetDomainText().empty());
+  EXPECT_FALSE(view_->IsFaviconVisibleForTesting());
+}
+
+TEST_F(FocusModeTitleBarViewTest, ElidesDomainFromTail) {
+  auto tab = AddTab(GURL("https://very-very-long-name.example.com/"));
+  view_->SetTab(tab.get());
+  ASSERT_EQ(GetDomainText(), u"very-very-long-name.example.com");
+
+  view_->SetSize(gfx::Size(60, 20));
+  views::test::RunScheduledLayout(view_.get());
+
+  std::u16string_view displayed =
+      view_->domain_label_for_testing()->GetDisplayTextForTesting();
+  EXPECT_TRUE(displayed.ends_with(u"\u2026"));
+  EXPECT_TRUE(displayed.starts_with(u"very"));
+}
+
+TEST_F(FocusModeTitleBarViewTest, UpdatesOnTabSwitch) {
+  auto first = AddTab(GURL("https://first.test/title1.html"));
+  view_->SetTab(first.get());
+  EXPECT_EQ(GetDomainText(), u"first.test");
+
+  auto second = AddTab(GURL("https://second.test/title1.html"));
+  view_->SetTab(second.get());
+  EXPECT_EQ(GetDomainText(), u"second.test");
+
+  view_->SetTab(nullptr);
+  EXPECT_TRUE(GetDomainText().empty());
+  EXPECT_FALSE(view_->IsFaviconVisibleForTesting());
+}
+
+}  // namespace


browser/ui/views/frame/focus_mode_title_bar_view_browsertest.cc
```diff
@@ -0,0 +1,140 @@
...
+  view_->SetWebContents(active_contents());
+  ASSERT_FALSE(view_->GetDomainTextForTesting().empty());
+
+  browser()->tab_strip_model()->CloseWebContentsAt(
+      browser()->tab_strip_model()->active_index(), TabCloseTypes::CLOSE_NONE);
+
+  view_->SetWebContents(nullptr);
+  EXPECT_TRUE(view_->GetDomainTextForTesting().empty());
+  EXPECT_FALSE(view_->IsFaviconVisibleForTesting());
+}
```
netzenbot: ARCH-062 requires new user-visible UI to have at least one CUJ test using `InteractiveBrowserTest` with `RunTestSequence`. These tests use `InProcessBrowserTest` instead, which doesn't fulfill the CUJ requirement. Consider adding an `InteractiveBrowserTest` that verifies the title bar's domain label shows when Focus Mode is active — e.g. `WaitForShow` on the view and `CheckViewProperty` on the label text. See [ARCH-062](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-062).

browser/ui/views/frame/focus_mode_title_bar_view.cc
```diff
@@ -0,0 +1,152 @@
...
+  std::u16string domain;
+  if (tab_ui_helper->ShouldDisplayURL()) {
+    domain = FormatDomain(domain_url);
+  }
+
+  domain_label_->SetText(domain);
+  domain_label_->SetElideBehavior(domain_url.IsStandard() &&
+                                          !domain_url.SchemeIsFile() &&
+                                          !domain_url.SchemeIsFileSystem()
+                                      ? gfx::ELIDE_HEAD
```
netzenbot: `TabUIHelper::From(interface)` can return null — `UnownedUserDataHost::Get()` explicitly documents this. `Update()` correctly guards with `CHECK(tab_ui_helper)`, but `SetWebContents()` dereferences the result directly. Add a null check (or `CHECK`) here too before calling `->AddTabUIChangeCallback(...)`. ([Null-Check Return Values Before Dereferencing](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-memory.md#CSM-029))

browser/ui/views/frame/focus_mode_title_bar_view.cc
```diff
@@ -0,0 +1,152 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "brave/browser/ui/views/frame/focus_mode_title_bar_view.h"
+
+#include "base/strings/utf_string_conversions.h"
+#include "brave/browser/ui/brave_scheme_utils.h"
```
netzenbot, line 11: nit: `base/strings/utf_string_conversions.h` appears unused here — the file uses `std::u16string` and `u""` literals directly but calls no function from that header. Remove it per IWYU. https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-001

browser/ui/views/frame/focus_mode_title_bar_view.h
```diff
@@ -0,0 +1,62 @@
...
+ private:
+  void Update();
+
+  raw_ptr<views::ImageView> favicon_image_ = nullptr;
+  raw_ptr<views::Label> domain_label_ = nullptr;
+  base::WeakPtr<content::WebContents> web_contents_;
+  base::CallbackListSubscription tab_ui_updated_subscription_;
+};
+
+#endif  // BRAVE_BROWSER_UI_VIEWS_FRAME_FOCUS_MODE_TITLE_BAR_VIEW_H_
```
netzenbot, line 53: nit: Per [CS-054](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-054), omit the `= nullptr` defaults for `favicon_image_` and `domain_label_` — both are unconditionally assigned in the constructor via `AddChildView()`, so the defaults are misleading.

browser/ui/views/frame/focus_mode_title_bar_view_unittest.cc
```diff
@@ -0,0 +1,53 @@
...
+}
+
+TEST_F(FocusModeTitleBarViewTest, DoesNotRewriteChromeUntrustedScheme) {
+  const std::u16string formatted =
+      FocusModeTitleBarView::FormatDomain(GURL("chrome-untrusted://print"));
+  EXPECT_NE(formatted.find(u"chrome-untrusted"), std::u16string::npos);
+  EXPECT_EQ(formatted.find(u"brave"), std::u16string::npos);
+}
+
+}  // namespace
```
netzenbot, line 164: nit: `FormatDomain` is a public static method but the unit tests only cover well-formed URLs. Per [TI-016](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-016), bad/invalid input should be covered — consider adding a case for an empty/invalid `GURL()` to confirm the function handles it gracefully.

browser/ui/BUILD.gn
```diff
@@ -606,6 +606,8 @@ source_set("ui") {
       "views/frame/brave_tab_strip_region_view.h",
       "views/frame/brave_window_frame_graphic.cc",
       "views/frame/brave_window_frame_graphic.h",
+      "views/frame/focus_mode_title_bar_view.cc",
```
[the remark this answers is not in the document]
zenparsing: Moving this into the `brave/browser/ui/views/frame` target would create a circular dependency due to the include of `brave/browser/ui/brave_scheme_utils.h`.

browser/ui/BUILD.gn
```diff
@@ -606,6 +606,8 @@ source_set("ui") {
...
+      "views/frame/focus_mode_title_bar_view.h",
```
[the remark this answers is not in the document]
zenparsing: Moving this into the `brave/browser/ui/views/frame` target would create a circular dependency due to the include of `brave/browser/ui/brave_scheme_utils.h`.

browser/ui/views/frame/focus_mode_title_bar_view.cc
```diff
@@ -0,0 +1,152 @@
...
+    favicon_image_->SetImage(ui::ImageModel());
+    favicon_image_->SetVisible(false);
+    domain_label_->SetText(u"");
+    return;
+  }
+
+  auto* tab_ui_helper = TabUIHelper::From(interface);
+  CHECK(tab_ui_helper);
+
+  GURL domain_url = tab_ui_helper->GetVisibleURL();
```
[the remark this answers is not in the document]
zenparsing: This view is intended to show the same information as the Omnibox (i.e. the visible URL). Additionally, `tab_ui_helper->ShouldDisplayURL()` uses the visible URL.

Also, using the committed URL results in displaying the URL sometimes when it should not be shown (i.e. when the user is transitioning from the NTP to another page).

browser/ui/views/frame/focus_mode_title_bar_view.cc
```diff
@@ -0,0 +1,153 @@
...
+  tab_ui_updated_subscription_ = {};
+  web_contents_.reset();
+  if (web_contents) {
+    web_contents_ = web_contents->GetWeakPtr();
+  }
+
+  if (auto* helper = GetTabUIHelper(web_contents_.get())) {
+    tab_ui_updated_subscription_ =
+        helper->AddTabUIChangeCallback(base::BindRepeating(
+            &FocusModeTitleBarView::Update, base::Unretained(this)));
```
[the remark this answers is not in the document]
zenparsing: This callback is guarded by a callback subscription.

diracdeltas:
@zenparsing did this have security design review? what is shown for HTTP sites or sites with security errors? you can find test cases on https://badssl.com/

diracdeltas:
i cannot seem to assign him as a reviewer but please wait on @thypon approval

t2micro:
Is a feature flag already availabe for this change? Is it going to be available by default? @zenparsing

browser/ui/views/frame/focus_mode_title_bar_view.cc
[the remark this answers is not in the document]
t2micro: gtg

browser/ui/views/frame/focus_mode_title_bar_view.cc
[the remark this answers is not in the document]
t2micro: Cc @bridiver

browser/ui/BUILD.gn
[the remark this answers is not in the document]
bridiver: Why was this resolved without any changes?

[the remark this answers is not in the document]
bridiver: I mean ignore the ridiculous suggestion to just comment it out, but these shouldn't just be ignored in general.

simonhong:
Curious when this title bar is visible? Always?
I think window should have caption buttons(maximum, close and etc..)
We'll not have caption button in focus mode?

browser/ui/BUILD.gn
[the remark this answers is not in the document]
zenparsing: @simonhong Do we have an issue yet to clean up the `frame` target?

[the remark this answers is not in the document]
zenparsing: Prereq: https://github.com/brave/brave-core/pull/36324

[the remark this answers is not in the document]
zenparsing: Why are these messages being duplicated?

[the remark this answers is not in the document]
bridiver: It is normal and expected to have circular dependencies when modularizing browser/ui code. Just add it to `brave_ui_allow_circular_includes_from` until the circular deps are resolved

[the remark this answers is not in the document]
bridiver: An "experiment" was running that shouldn't have been. It is being disabled.

browser/ui/views/frame/focus_mode_title_bar_view.cc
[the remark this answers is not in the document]
bridiver: I'm more concerned with edge cases where this might not update properly and result in url spoofing. I only see that it updates when calling `SetWebContents`, but it's not clear when that gets called. @simonhong this same issue came up before with splitview, what did we do there again? This needs some very careful attention because it's potentially a very serious security issue so we need to look at this carefully. Also it should be behind a feature flag so we can disable by griffin if there are any issues with spoofing.

browser/ui/views/frame/focus_mode_title_bar_view.cc
```diff
@@ -0,0 +1,153 @@
...
+  domain_label_->SetEnabledColor(nala::kColorTextTertiary);
+  domain_label_->SetAutoColorReadabilityEnabled(false);
+}
+
+FocusModeTitleBarView::~FocusModeTitleBarView() = default;
+
+void FocusModeTitleBarView::SetWebContents(content::WebContents* web_contents) {
+  tab_ui_updated_subscription_ = {};
+  web_contents_.reset();
+  if (web_contents) {
```
bridiver: if `web_contents` is nullptr we just continue this method? That doesn't seem right, when would this be null?

browser/ui/views/frame/focus_mode_title_bar_view_unittest.cc
```diff
@@ -0,0 +1,58 @@
...
+#include <string>
+
+#include "testing/gtest/include/gtest/gtest.h"
+#include "url/gurl.h"
+
+namespace {
+
+using FocusModeTitleBarViewTest = ::testing::Test;
+
+TEST_F(FocusModeTitleBarViewTest, StripsHttpsAndTrivialSubdomain) {
```
bridiver: This appears to only test `FormatDomain`, but there are lot of edge cases around `SetWebContents`

browser/ui/views/frame/focus_mode_title_bar_view.cc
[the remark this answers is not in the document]
zenparsing: Great questions. `SetWebContents` will be called from:

* `BrowserView::OnActiveTabChanged` (with the new web contents)
* `UpdateFocusModeState` (with `browser()->tab_strip_model()->GetActiveWebContents()` if focus mode is enabled).

This view is not currently hooked up. When it is, it will only be added to the browser view when the focus mode feature flag is enabled.

https://github.com/brave/brave-core/pull/36025 contains a prototype where this title bar view is wired up.

browser/ui/views/frame/focus_mode_title_bar_view.cc
zenparsing: `GetTabUIHelper` gracefully deals with `nullptr`, and `Update` must be called in either case. Does that work?

browser/ui/views/frame/focus_mode_title_bar_view.cc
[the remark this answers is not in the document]
bridiver: The code should CHECK that the flag is enabled, but `OnActiveTabChanged` is not sufficient because it needs to update whenever the page navigates. We dealt with this same issue with split view and re-used existing logic, but I don't recall off hand what we did exactly so waiting to hear back from @simonhong.

browser/ui/views/frame/focus_mode_title_bar_view_unittest.cc
zenparsing: There are several tests in `focus_mode_title_bar_view_browsertest.cc` that exercise `SetWebContents`. If there are any missing cases we should add them there.

browser/ui/views/frame/focus_mode_title_bar_view.cc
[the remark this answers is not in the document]
zenparsing: Good idea on the `CHECK`.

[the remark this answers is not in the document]
bridiver: also TabListInterfaceObserver should be preferred over BrowserView::OnActiveTabChanged.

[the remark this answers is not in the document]
zenparsing: A callback is registered with `helper->AddTabUIChangeCallback`. It looks like this callback is called for the various WebContents state changes (including navigation).

browser/ui/views/frame/focus_mode_title_bar_view.cc
bridiver: but when would it be null? `absl::MakeCleanup` should be used for code that always needs to run at the end of the method even if it returns early

browser/ui/views/frame/focus_mode_title_bar_view.cc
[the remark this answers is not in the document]
bridiver: I see, I'll take a look. Was there some other code you used as a model for this?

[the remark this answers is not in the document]
bridiver: in any case as I mentioned below we need to test all of these edge cases in the unit test. Also is there a reason we're dealing with the WebContents instead of just setting the TabInterface directly? See https://bravesoftware.slack.com/archives/C09JE6H9QP4/p1775047837172039

browser/ui/views/frame/focus_mode_title_bar_view_unittest.cc
bridiver: We should be unit testing where possible and there are plenty of cases here that can and should be unit tested.

browser/ui/views/frame/focus_mode_title_bar_view.cc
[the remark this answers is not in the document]
zenparsing: See `chrome/browser/ui/views/frame/multi_contents_view_mini_toolbar.h`.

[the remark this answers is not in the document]
bridiver: actually why is `SetWebContents` a public method? The class should be observing changes and updating, not having changes pushed to it?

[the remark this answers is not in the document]
bridiver: that class uses 
```
GURL domain_url = tab_ui_helper->GetVisibleURL();
  if (tab_ui_helper->GetLastCommittedURL().is_valid()) {
    domain_url = tab_ui_helper->GetLastCommittedURL();
  }
```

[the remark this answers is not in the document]
zenparsing: Yes - that was the code originally used in this view. However, it was found that that resulted in a brief display of `brave://newtab` as the tab navigated from the NTP to the next page. The reason: `tab_ui_helper->ShouldDisplayURL()` only looks at the visible URL and we rely on that to hide the host display for the NTP.

Regarding `SetWebContents` being public, it seemed a better design to have it simply driven by the owner (the browser view), rather than made aware of the tabstrip. Suggestions welcome.

browser/ui/views/frame/focus_mode_title_bar_view_unittest.cc
zenparsing: Please advise if there are any concrete missing cases.

browser/ui/views/frame/focus_mode_title_bar_view.h
netzenbot, line 53: nit: `GetDomainTextForTesting()`, `IsFaviconVisibleForTesting()`, and `GetElideBehaviorForTesting()` are public only for test access. Per [ARCH-020](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-020), these should be `private` with `friend class FocusModeTitleBarViewBrowserTest` and `friend class FocusModeTitleBarViewTest` declarations instead.

browser/ui/views/frame/focus_mode_title_bar_view.h
```diff
@@ -0,0 +1,62 @@
...
+  // trivial subdomains, anything after the host, and rewrites a `chrome://`
+  // scheme to `brave://`.
+  static std::u16string FormatDomain(const GURL& url);
+
+  std::u16string_view GetDomainTextForTesting() const;
+  bool IsFaviconVisibleForTesting() const;
+  gfx::ElideBehavior GetElideBehaviorForTesting() const;
+
+ private:
+  void Update();
```
netzenbot, line 42: nit: The comment notes `web_contents` may be null but doesn't describe what happens in that case. Per DOC-004, the contract should cover behavior — e.g., that passing null clears the domain label, hides the favicon, and cancels the tab-UI subscription. See https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-004.

browser/ui/views/frame/focus_mode_title_bar_view.h
zenparsing: `*ForTesting` methods should be public.

https://chromium.googlesource.com/chromium/src/+/HEAD/docs/patterns/fortesting-methods.md.

browser/ui/views/frame/focus_mode_title_bar_view_browsertest.cc
zenparsing: CUJ tests will be added for focus mode in later PRs.

browser/ui/views/frame/focus_mode_title_bar_view.h
zenparsing: This rule should only apply when the member is always initialized in a constructor member initializer list, otherwise the member will contain garbage in the constructor body.

browser/ui/views/frame/focus_mode_title_bar_view.cc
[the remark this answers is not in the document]
simonhong: When we had our own mini urlbar for split view, we used `LocationBarModelImpl::GetURLForDisplay()`. Then, it's gone as we migrated to upstream's split view impl. As @zenparsing mentioned above, upstream handles visibile url in `MultiContentsViewMiniToolbar::UpdateContents()`

browser/ui/views/frame/focus_mode_title_bar_view_browsertest.cc
```diff
@@ -0,0 +1,140 @@
...
+#include "content/public/browser/web_contents.h"
+#include "content/public/test/browser_test.h"
+#include "net/dns/mock_host_resolver.h"
+#include "net/test/embedded_test_server/embedded_test_server.h"
+#include "ui/base/window_open_disposition.h"
+#include "ui/gfx/text_constants.h"
+#include "ui/views/widget/widget.h"
+#include "url/gurl.h"
+
+class FocusModeTitleBarViewBrowserTest : public InProcessBrowserTest {
```
simonhong: Not sure but I think this browser test could be done in unittest?
* Creating web contents by using `TestWebContentsFactory::CreateWebContents()`
* Navigating by using `NavigationSimulator::NavigateAndCommitFromBrowser`
* Prepare title view by using `CreateTestWidget()` and set title view as widget's contents view(`Widget::SetContentsView())`

browser/ui/views/frame/focus_mode_title_bar_view.cc
simonhong: It's difficult to know how `FocusModeTitleBarView::SetWebContents()` works because no one calls this in production code. Only test code uses it in this PR.
I assume that maybe another controller(focus mode controller?) will call it with active web contents whenever active tab is changed. Am I right?

simonhong: I recommend that it's better to make UI dumb as much as possible something like caller just set favicon or url directly to this.

zenparsing: Updated to use `TabInterface`. The owner (`BraveBrowserView`) is expected to call `SetTab` as the active tab changes.

Recommend against making this view a `TabListInterfaceObserver`, as that would couple the view to a window-level concept ("the active tab") which it doesn't control or need to know about.

Recommend against having the owning view (`BraveBrowserView`) push favicons and URLs: this would bloat an already large, complex class with focus-mode related display policy.

browser/ui/views/frame/focus_mode_title_bar_view_browsertest.cc
zenparsing: I think that works for views that consume a `WebContents` directly, but this view uses `TabInterface` and reads through `TabUIHelper`, which requires a real `TabFeatures`. In order to make unit testing work, we'd need to mock/abstract out the `TabUIHelper` usage. Suggest we stick to browser tests unless/until the testing surface grows extensively.

browser/ui/views/frame/focus_mode_title_bar_view.cc
```diff
@@ -0,0 +1,152 @@
...
+  tab_ = tab;
+
+  if (tab_) {
+    tab_will_detach_subscription_ =
+        tab_->RegisterWillDetach(base::BindRepeating(
+            &FocusModeTitleBarView::OnTabWillDetach, base::Unretained(this)));
+    if (auto* helper = TabUIHelper::From(tab_)) {
+      tab_ui_updated_subscription_ =
+          helper->AddTabUIChangeCallback(base::BindRepeating(
+              &FocusModeTitleBarView::Update, base::Unretained(this)));
```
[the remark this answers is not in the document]
zenparsing: Guarded by subscription.

browser/ui/views/frame/focus_mode_title_bar_view.cc
```diff
@@ -0,0 +1,152 @@
...
+
+void FocusModeTitleBarView::SetTab(tabs::TabInterface* tab) {
+  tab_ui_updated_subscription_ = {};
+  tab_will_detach_subscription_ = {};
+  tab_ = tab;
+
+  if (tab_) {
+    tab_will_detach_subscription_ =
+        tab_->RegisterWillDetach(base::BindRepeating(
+            &FocusModeTitleBarView::OnTabWillDetach, base::Unretained(this)));
```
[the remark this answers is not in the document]
zenparsing: Guarded by subscription.

browser/ui/views/frame/focus_mode_title_bar_view.cc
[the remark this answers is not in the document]
zenparsing: Guarded by subscription.

browser/ui/views/frame/BUILD.gn
```diff
@@ -6,6 +6,7 @@
 source_set("frame") {
   sources = [
     "brave_contents_view_util.h",
+    "focus_mode_title_bar_view.h",
```
netzenbot, line 10: `focus_mode_title_bar_view.h` is listed in the `frame` source_set and its public method `SetTab(tabs::TabInterface*)` includes `components/tabs/public/tab_interface.h` — a header-visible dependency. Per [BS-024](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-024), `//components/tabs:public` must be in `public_deps` of `frame` (not just in `frame_impl`'s `deps`) so consumers of `:frame` that include this header get the transitive include path.

browser/ui/views/frame/focus_mode_title_bar_view.h
```diff
@@ -0,0 +1,63 @@
...
+                       tabs::TabInterface::DetachReason reason);
+
+  raw_ptr<views::ImageView> favicon_image_ = nullptr;
+  raw_ptr<views::Label> domain_label_ = nullptr;
+  raw_ptr<tabs::TabInterface> tab_ = nullptr;
+  base::CallbackListSubscription tab_will_detach_subscription_;
+  base::CallbackListSubscription tab_ui_updated_subscription_;
+};
+
+#endif  // BRAVE_BROWSER_UI_VIEWS_FRAME_FOCUS_MODE_TITLE_BAR_VIEW_H_
```
netzenbot, line 53: nit: Per [CSM-013](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-memory.md#CSM-013), unowned `raw_ptr<>` members should come after owning members. `favicon_image_`, `domain_label_`, and `tab_` are unowned, but they're declared before the owning `CallbackListSubscription` members — reverse the order.

browser/ui/views/frame/focus_mode_title_bar_view.cc
```diff
@@ -0,0 +1,152 @@
...
+    bool themify_favicon = tab_ui_helper->ShouldThemifyFavicon();
+    if (auto* provider = GetColorProvider(); provider && themify_favicon) {
+      SkColor favicon_color = provider->GetColor(kColorBookmarkFavicon);
+      if (favicon_color != SK_ColorTRANSPARENT) {
+        favicon = ui::ImageModel::FromImageSkia(
+            gfx::ImageSkiaOperations::CreateColorMask(
+                favicon.Rasterize(provider), favicon_color));
+      }
+    }
+    favicon_image_->SetImage(favicon);
```
netzenbot, line 128: nit: The `.cc` definition order doesn't match the `.h` declaration order. The header declares `Update` then `OnTabWillDetach` in the private section, and all private methods after the public ones; the `.cc` places `OnTabWillDetach` before `FormatDomain` and `Update` at the end. Per [CS-017](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-017), keep `.cc` order consistent with `.h`.

browser/ui/views/frame/focus_mode_title_bar_view.cc
```diff
@@ -0,0 +1,152 @@
...
+      kFaviconLabelSpacing));
+  layout->set_main_axis_alignment(views::BoxLayout::MainAxisAlignment::kCenter);
+  layout->set_cross_axis_alignment(
+      views::BoxLayout::CrossAxisAlignment::kCenter);
+
+  SetPreferredSize(gfx::Size(0, kTitleBarHeight));
+
+  favicon_image_ = AddChildView(std::make_unique<views::ImageView>());
+  favicon_image_->SetImageSize(gfx::Size(kFaviconSize, kFaviconSize));
+  favicon_image_->SetVisible(false);
```
simonhong, line 56: need to control `favicon_image_`'s visibility in this class?
Not sure but maybe it's invisible when it doesn't have image as its size becomes zero.

browser/ui/views/frame/focus_mode_title_bar_view.cc
```diff
@@ -0,0 +1,152 @@
...
+    }
+  }
+
+  Update();
+}
+
+void FocusModeTitleBarView::OnTabWillDetach(
+    tabs::TabInterface* tab,
+    tabs::TabInterface::DetachReason reason) {
+  if (reason == tabs::TabInterface::DetachReason::kDelete) {
```
simonhong: don't need to handle to `DetachReason::kInsertIntoOtherWindow`?

browser/ui/views/frame/focus_mode_title_bar_view.cc
simonhong: Setting `TabInterface` seems much better.

browser/ui/views/frame/focus_mode_title_bar_view.cc
zenparsing: Since we set the image size to 12x12 to match designs, I believe that the preferred size is retained even for a null image.

browser/ui/views/frame/focus_mode_title_bar_view.cc
zenparsing: Eventually it would be handled by the owner calling `SetTab` when the active tab changes, but handling `kInsertIntoOtherWindow` here is more correct. Thanks!

browser/ui/views/frame/focus_mode_title_bar_view_browsertest.cc
```diff
@@ -0,0 +1,136 @@
...
+ protected:
+  tabs::TabInterface* active_tab() {
+    return browser()->tab_strip_model()->GetActiveTab();
+  }
+
+  raw_ptr<FocusModeTitleBarView> view_ = nullptr;
+  std::unique_ptr<views::Widget> widget_;
+};
+
+IN_PROC_BROWSER_TEST_F(FocusModeTitleBarViewBrowserTest, ReflectsActiveTab) {
```
simonhong: nit: I think launching each browser test case is quite resource consuming.
If some of test cases could be unified, it would be great.

simonhong (approved):
++

t2micro:
Is there a downloadable build already?

zenparsing:
@t2micro The prototype branch at https://github.com/brave/brave-core/pull/36461 has CI builds and can be used to demo the feature as a whole.

browser/ui/views/frame/focus_mode_title_bar_view_browsertest.cc
bridiver: Shouldn't this be a unit test with ChromeViewsTestBase? This isn't actually a proper browser test anyway because we're creating FocusModeTitleBarView directly

browser/ui/views/frame/focus_mode_title_bar_view_browsertest.cc
```diff
@@ -0,0 +1,136 @@
...
+      browser(), embedded_test_server()->GetURL("second.test", "/title1.html"),
+      WindowOpenDisposition::NEW_FOREGROUND_TAB,
+      ui_test_utils::BROWSER_TEST_WAIT_FOR_LOAD_STOP |
+          ui_test_utils::BROWSER_TEST_WAIT_FOR_TAB));
+  view_->SetTab(active_tab());
+  EXPECT_TRUE(view_->GetDomainTextForTesting().contains(u"second.test"));
+}
+
+IN_PROC_BROWSER_TEST_F(FocusModeTitleBarViewBrowserTest,
+                       ChromeSchemeRewrittenToBrave) {
```
bridiver: This is definitely not something we should be testing in a browser test, but it's also a change detector test so I think we should drop it entirely per chromium design guidelines

browser/ui/views/frame/focus_mode_title_bar_view_unittest.cc
bridiver: one that is definitely missing is calling SetTab with nullptr

browser/ui/views/frame/focus_mode_title_bar_view_browsertest.cc
bridiver: also shouldn't there be an interactive ui test? We should testing most of this behavior in unit tests and reserve interactive ui tests for just one or two basic cases

browser/ui/views/frame/focus_mode_title_bar_view_unittest.cc
zenparsing: Covered by `FocusModeTitleBarViewBrowserTest.SwitchesBetweenTabs`.

browser/ui/views/frame/focus_mode_title_bar_view_browsertest.cc
zenparsing: From a comment above:

> ...this view uses TabInterface and reads through TabUIHelper, which requires a real TabFeatures. In order to make unit testing work, we'd need to mock/abstract out the TabUIHelper usage. Suggest we stick to browser tests unless/until the testing surface grows extensively.

Interestingly, `MultiContentsViewMiniToolbar` doesn't have any tests at all, so there's isn't a good upstream pattern.

Options:

1. Keep the browser tests.
2. Drop the browser tests and introduce browser/interactive UI tests when the other parts land.
3. Introduce a delegate interface, which defaults to wrapping `TabUIHelper`. The unit test would implement with a fake.

IMO, 3 is overkill, but your thoughts?

browser/ui/views/frame/focus_mode_title_bar_view_browsertest.cc
zenparsing: It's covered in the unit tests - I can remove it from the browser tests.

zenparsing: Browser test removed.

browser/ui/views/frame/focus_mode_title_bar_view_browsertest.cc
zenparsing: Per @simonhong 's suggestion, I've consolidated the browser tests; we now have just 2 browser tests. Between the unit tests exercising the domain formatting and the browser tests validating the integration with tabs, I think we have good coverage. Agreed?

netzenbot (approved):

browser/ui/views/frame/focus_mode_title_bar_view_browsertest.cc
```diff
@@ -0,0 +1,106 @@
...
+  std::unique_ptr<views::Widget> widget_;
+};
+
+IN_PROC_BROWSER_TEST_F(FocusModeTitleBarViewBrowserTest, UpdatesOnNavigation) {
+  // Verify that the title bar displays correctly for the page.
+  ASSERT_TRUE(ui_test_utils::NavigateToURL(
+      browser(),
+      embedded_test_server()->GetURL("www.example.com", "/title1.html")));
+  view_->SetTab(active_tab());
+  ASSERT_TRUE(view_->GetDomainTextForTesting().contains(u"example.com"));
```
bridiver: I don't understand why you need `GetElideBehaviorForTesting`, why aren't we we testing that the title is displayed as expected? We shouldn't be testing internal details that aren't part of the "output"

zenparsing: Removed, and added a test for the "elide front" behavior.

browser/ui/views/frame/focus_mode_title_bar_view_unittest.cc
```diff
@@ -0,0 +1,198 @@
...
+ private:
+  ui::UnownedUserDataHost host_;
+  std::unique_ptr<tabs::TabFeatures> tab_features_;
+  std::unique_ptr<content::WebContents> contents_;
+  base::RepeatingCallbackList<void(tabs::TabInterface*,
+                                   tabs::TabInterface::DetachReason)>
+      will_detach_callbacks_;
+};
+
+class FocusModeTitleBarViewTest : public ChromeRenderViewHostTestHarness {
```
bridiver: Why is this a ChromeRenderViewHostTestHarness subclass? We're not doing anything with the renderer process. Maybe you intended to use ChromeViewsTestBase here?

browser/ui/views/frame/focus_mode_title_bar_view_unittest.cc
```diff
@@ -0,0 +1,198 @@
...
+#include "ui/views/controls/label.h"
+#include "ui/views/test/views_test_utils.h"
+#include "url/gurl.h"
+
+namespace {
+
+using FocusModeTitleBarViewFormatTest = ::testing::Test;
+
+TEST_F(FocusModeTitleBarViewFormatTest, StripsHttpsAndTrivialSubdomain) {
+  EXPECT_EQ(FocusModeTitleBarView::FormatDomain(
```
bridiver: These tests seem like they're actually testing `url_formatter::FormatUrl` which isn't necessary here. Seems like this is only static/public for unit tests which is generally not desirable, this should just be an internal method. The place to test this is just to check the full display value in FocusModeTitleBarViewTests

browser/ui/views/frame/focus_mode_title_bar_view_unittest.cc
```diff
@@ -0,0 +1,198 @@
...
+    tabs_.clear();
+    ChromeRenderViewHostTestHarness::TearDown();
+  }
+
+ protected:
+  FakeTab* AddTab(const GURL& url) {
+    auto tab = std::make_unique<FakeTab>(profile());
+    tab->NavigateAndCommit(url);
+    FakeTab* tab_ptr = tab.get();
+    tabs_.push_back(std::move(tab));
```
bridiver: I don't understand the purpose of `tabs_`, it seems like it just complicates things vs returning the unique_ptr directly?

browser/ui/views/frame/focus_mode_title_bar_view_unittest.cc
```diff
@@ -0,0 +1,166 @@
...
+  auto tab = AddTab(GURL("https://very-long-spoofing.example.com/"));
+  view_->SetTab(tab.get());
+  ASSERT_EQ(GetDomainText(), u"very-long-spoofing.example.com");
+
+  view_->SetSize(gfx::Size(60, 20));
+  views::test::RunScheduledLayout(view_.get());
+
+  std::u16string_view displayed =
+      view_->domain_label_for_testing()->GetDisplayTextForTesting();
+  EXPECT_TRUE(displayed.starts_with(u"\u2026"));
```
bridiver: we know exactly what this should be, why not check the full display value?

zenparsing: It turns out the display text is not the same across platforms, possibly due to font differences.

browser/ui/views/frame/focus_mode_title_bar_view.cc
```diff
@@ -0,0 +1,142 @@
...
+
+FocusModeTitleBarView::~FocusModeTitleBarView() = default;
+
+void FocusModeTitleBarView::SetTab(tabs::TabInterface* tab) {
+  tab_ui_updated_subscription_ = {};
+  tab_will_detach_subscription_ = {};
+  tab_ = tab;
+
+  if (tab_) {
+    tab_will_detach_subscription_ =
```
bridiver, line 75: shouldn't we also have an attach subscription? Doesn't look like we currently have any tests for attach/detach behavior?

zenparsing: I don't see what the purpose would be. The owner is responsible for only calling `SetTab` with tabs attached to the current window. Detach path is tested via `FocusModeTitleBarViewTest.ClearsWhenTabDetaches`.

bridiver: > The owner is responsible for only calling SetTab with tabs attached to the current window

Is that contract stated or checked anywhere? How is this enforced or at least communicated? Maybe there should be a check that BrowserWindowInterface is not null?

zenparsing: We could add a check against `GetBrowserWindowInterface`, but according to the documentation [here](https://source.chromium.org/chromium/chromium/src/+/main:components/tabs/public/tab_interface.h;l=80;drc=2e070bfa61f747388fddaafd8eaf9519f89e80de;bpv=1;bpt=1), it will almost always be non-null in any case. I'm curious, can you explain why this is important?

zenparsing: I took a deeper look at this, and I think that the tab's attach/detach state is not a concern of this view. This view simply displays information for whatever tab it is associated with. The detach subscription is primarily there to ensure memory-safety - that the `raw_ptr<TabInterface>` does not outlive the tab.

browser/ui/views/frame/focus_mode_title_bar_view_unittest.cc
zenparsing: `tabs_` removed.

browser/ui/views/frame/focus_mode_title_bar_view_unittest.cc
zenparsing: Formatting tests have been moved into `FocusModeTitleBarViewTest`.

browser/ui/views/frame/focus_mode_title_bar_view_unittest.cc
zenparsing: Fixed.

thypon (approved):

PR #36267, opened 2026-05-08
Try to minimize plaster related to PageActionController/Model

sangwoo108:
In order to make patching strategy more robust agains upstream chagnes, we should prefer
`chromium_src override without #define macro' to 'complex plaster'.


chromium_src/chrome/browser/ui/views/page_action/page_action_controller.h
```diff
@@ -0,0 +1,51 @@
...
+                          std::optional<SkColor> override_foreground_color);
+  void ClearOverrideChipColors(actions::ActionId action_id);
+  void SetOverrideHeight(actions::ActionId action_id, int height);
+  void ClearOverrideHeight(actions::ActionId action_id);
+  void SetOverrideTriggerableEvent(actions::ActionId action_id,
+                                   std::optional<int> event_flags);
+
+  // PageActionController_Chromium:
+  std::unique_ptr<PageActionModelInterface> CreateModel(
+      actions::ActionId action_id,
```
netzenbot, line 38: nit: The override comment says `// PageActionController_Chromium:` but the actual renamed class is `PageActionControllerImpl_Chromium`. Future readers looking for the base class will search for the wrong name. ([Comments Must Make Sense to Future Readers of the Codebase](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-019))

cdesouza-chromium (approved):
LGTM

PR #36271, opened 2026-05-08
[plaster][brockit] `PlasterFile` does not handle errors

cdesouza-chromium:
This change removes uses of `sys.exit(1)` and error message prints from
`PlasterFile`, as this interferes with other tools using this class,
like `cr-mv`, `cr-follow-renames`, and `brockit`.


mkarolin (approved):
++

PR #36151, opened 2026-05-04
[Ads][Origin] Defer ads service start until policy service is initialized

SergeyZhukovsky:
The ads gate reads `kDisabledByPolicy` via `IsManagedPreference`, so if `AdsServiceImpl` is constructed before the BraveOrigin policy bundle has been merged into the managed pref store, the gate evaluates as unmanaged and the service starts. NTT ads can then serve despite `BraveRewardsDisabled` being active.

Defer the initial `MaybeStartBatAdsService()` call until `PolicyService::IsInitializationComplete(POLICY_DOMAIN_CHROME)` is true. Adds `[AdsGate]` diagnostics behind the existing "Verbose Logs for Ad-Rewards" QA toggle.

### Test case 1:
Make sure that following issue cannot be reproduced:
https://github.com/brave/brave-browser/issues/54991

### Test case 2:
1. Start the browser (non Brave Origin)
2. Make sure that NTTs can be served
3. Open any web page
4. Restart the browser
5. Open NTP (without much a delay)
EXPECTATION: NTT is displayed

### Test case 3:
1. Start the browser (non Brave Origin)
2. Enable Rewards
3. Open search.brave.com
4. Trigger a search result ad
5. Restart the browser
6. Click the "search result ad" on restored tab with SERP (without much a delay)
7. Open [brave://ads-internals](chrome://ads-internals/)
EXPECTATION: Search result ad clicked event is recorded.

Resolves: https://github.com/brave/brave-browser/issues/55228
Resolves: https://github.com/brave/brave-browser/issues/54991

PR #36276, opened 2026-05-08
[Origin] Don't start ads service when Rewards is disabled by Origin policy

bbondy:
## Summary

Fixes ads service starting (and showing NTP ads / ad notifications) for Brave Origin users despite Rewards being policy-disabled.

The chromium policy → managed-pref pipeline races against eager AdsService construction at profile init: `bat_ads` launches before `BraveRewardsDisabled` lands as a managed pref, and the existing `brave_rewards::IsSupported()` check sees the pref unmanaged.

Wallet and AI Chat are not affected by the same race because their factory-level gate is re-evaluated on every `GetForBrowserContext()` call. AdsService is unique in committing to a launched utility process at construction without re-checking.

## Fix

Read the Origin policy directly from local state (`kBraveOriginPolicies` dict via `BraveOriginPolicyManager`) at the top of `CanStartBatAdsService()`. The dict honors user overrides (e.g. user toggled Rewards back on under Origin), so the gate respects both the policy default and the override.

## Test plan

- [ ] Brave Origin purchased, default: ads service does not start.
- [ ] Brave Origin purchased, user toggles Rewards back on: ads service starts.
- [ ] Non-Origin user, no Rewards opt-in: ads service starts (NTP/SearchResult branch).
- [ ] Non-Origin user, opted into Rewards: ads service starts (joined-Rewards branch).
- [ ] Same matrix on Android.

### DIFF
--- a/components/brave_ads/browser/ads_service_impl.cc
+++ b/components/brave_ads/browser/ads_service_impl.cc
@@ -44,6 +44,8 @@
 #include "brave/components/brave_ads/core/public/history/site_history.h"
 #include "brave/components/brave_ads/core/public/prefs/pref_names.h"
 #include "brave/components/brave_ads/core/public/user_attention/user_idle_detection/user_idle_detection_feature.h"
+#include "brave/components/brave_origin/brave_origin_policy_manager.h"
+#include "brave/components/brave_origin/profile_id.h"
 #include "brave/components/brave_rewards/core/buildflags/buildflags.h"
 #include "brave/components/brave_rewards/core/mojom/rewards.mojom.h"
 #include "brave/components/brave_rewards/core/pref_names.h"
@@ -55,6 +57,7 @@
 #include "components/content_settings/core/common/content_settings.h"
 #include "components/content_settings/core/common/content_settings_pattern.h"
 #include "components/metrics/metrics_pref_names.h"
+#include "components/policy/policy_constants.h"
 #include "components/prefs/pref_service.h"
 #include "components/variations/pref_names.h"
 #include "content/public/browser/browser_context.h"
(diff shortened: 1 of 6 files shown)


components/brave_ads/browser/ads_service_impl.cc
```diff
@@ -247,6 +252,25 @@ bool AdsServiceImpl::UserHasOptedInToSearchResultAds() const {
 }
 
 bool AdsServiceImpl::CanStartBatAdsService() const {
+  // Read the Brave Origin policy state directly from local state. The chromium
```
bridiver: This seems concerning, if managed prefs aren't correct at profile init then we can have all kinds of potential issues with this. I think this is something we need to look at because if it's a problem here then it's also a problem in other places, no?

bridiver: Also this logic doesn't seem like it belongs in AdsServiceImpl and should be encapsulated in BraveOriginPolicyManager so you just pass in the relevant pref, but I'm really concerned that we have to do this at all.

aseren (approved):

components/brave_ads/browser/ads_service_impl.cc
bridiver: I guess this is just a temporary workaround, but I'll still prefer to expose it to ads service as just a simple helper method on BraveOriginPolicyManager or some util function or something that passes in kBraveRewardsDisabled

bbondy: Done in 7c357b0 — extracted into `BraveOriginPolicyManager::GetEnforcedPolicyValue(policy_key, profile_id)` and the ads call site is now a single `value_or(false)` check. Added a `TEMPORARY WORKAROUND` block on the helper documenting why it exists and the proper fix pointing at `chrome/browser/extensions/forced_extensions/force_installed_tracker.cc` (observe `policy::PolicyService::IsInitializationComplete(POLICY_DOMAIN_CHROME)` before doing irreversible work). PR description updated to call out temporary nature and the followup.

bridiver (approved):

bbondy:
Closing in favor of https://github.com/brave/brave-core/pull/36151

PR #36280, opened 2026-05-08
Add "No container" item display condition and change position.

### LINKED ISSUE #54770: [Containers] Hide "No container" item menu when there's no container in the context
In container-related context menus when `GetCurrentContainerIds` returns an empty list, hide "No container" item. The item can stay in New tab right click menu, but in other menus it should be removed (right click on a link, right click on a tab, right click on a bookmark).

goodov:
Changes:

* Do not display "No container" in "New tab" context menu and in "Open bookmark" context menus.
* Do not display it if there's no active container in the context.
* When displayed, put it at the end of containers list.

| Open in | Create new tab|
|--------|--------|
| image | image |

Resolves https://github.com/brave/brave-browser/issues/54770


aguscruiz (commented):
Why was it requested to be placed as the last option instead of the first one? 

I had placed it as the first option because a user might have a large number of containers, and putting "No container" first makes it more readily available than having to go down to the last item of the list.

goodov:
> Why was it requested to be placed as the last option instead of the first one?
> 
> I had placed it as the first option because a user might have a large number of containers, and putting "No container" first makes it more readily available than having to go down to the last item of the list.

Do you think this item is used very frequently? To me this is very rarely used item that takes the first place for no good reason, and in "New Tab" and "Open bookmark" context menus this item shouldn't exist. The idea is to not display it *consistently*, but if it's displayed, then it shouldn't mess with the order of containers (first item is always "Personal" container, for ex.).

Or do you think having "No container" option in "New tab" and "Open bookmark" context menus is right? I personally just see it as the noise that I always have to ignore in those menus.

aguscruiz:
I think I'm seeing it from the side of, if I already have a tab open in a container, if I right click I would expect the first item to be "No container" so I can escape the container quickly. That's why. I think having it as the first option adds more predictiability and muscle memory for the user rather than having to visually scan for what the last item on the list is (but not last LAST, the last one before the other section begins). 

Like you said, if the option has no use for you then your mind just skips it, but if it's relevant you always will know it's the first option on the list.

image

browser/ui/containers/containers_menu_model_unittest.cc
```diff
@@ -134,7 +150,6 @@ TEST_F(ContainersMenuModelUnitTest, GetCurrentContainerIdsAreChecked) {
     ContainersMenuModel model =
         test::ContainersMenuModelTestApi::CreateContainersMenuModel(
             delegate, GetContainers());
-    EXPECT_TRUE(model.IsCommandIdChecked(IDC_OPEN_IN_CONTAINER_NO_CONTAINER));
     EXPECT_FALSE(model.IsCommandIdChecked(
         test::ContainersMenuModelTestApi::GetCommandIdFromItemIndex(model, 0)));
     EXPECT_FALSE(model.IsCommandIdChecked(
```
netzenbot, line 155: nit: Add `<< " at index " << i` to the `EXPECT_NE` so failures identify which iteration failed. Without it, a failure just says the command ID matched without telling you where. ([✅ Include Diagnostic State in Test Assertions](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-014))

goodov:
I think I'm gonna close this. The current impl is reasonable, I don't have strong arguments against it. Maybe I'd remove a checkbox from "No container" item when it doesn't make sense, but that's a very minor issue.

browser/ui/containers/containers_menu_model_unittest.cc
netzenbot, line 155: nit: Add `<< " at index " << i` to the `EXPECT_NE` so failures identify which iteration failed. Without it, a failure just says the command ID matched without telling you where. ([✅ Include Diagnostic State in Test Assertions](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-014))

PR #36282, opened 2026-05-08
[ads] Fix AdsClientNotifier observer list reentrancy in tests

### LINKED ISSUE #55314: [ads] Address `ObserverList` reentrancy in `AdsClientNotifier` for Brave Ads tests
The `AdsClientNotifierObserver` observer list in `AdsClientNotifier` is currently being iterated recursively in Brave Ads unit tests which causes a DCHECK if reentrancy policy is set to kDisallowReentrancy.
We need to preferably refactor `AdsClientNotifier` and tests to eliminate the reentrancy or use ReentrantObserverList.

The DCHECK happens in tests only, because in production `AdsClientNotifierObserver` events are triggered asynchronously via mojom communication and cannot be reentered on the same thread.

aseren:
The test mock for SetProfilePref and SetLocalStatePref called NotifyPrefDidChange synchronously from within a pref change handler, causing reentrant observer notifications. The changes post the notification asynchronously and update affected tests to wait for async delivery matching production behavior.

IMPORTANT: The PR is in draft and needs additional changes because some tests can now pass without expectation verification.

Resolves https://github.com/brave/brave-browser/issues/55314

### DIFF
--- a/components/brave_ads/core/internal/account/user_rewards/user_rewards_util_unittest.cc
+++ b/components/brave_ads/core/internal/account/user_rewards/user_rewards_util_unittest.cc
@@ -6,6 +6,7 @@
 #include "brave/components/brave_ads/core/internal/account/user_rewards/user_rewards_util.h"
 
 #include "base/scoped_observation.h"
+#include "base/test/run_until.h"
 #include "brave/components/brave_ads/core/internal/account/issuers/issuers_info.h"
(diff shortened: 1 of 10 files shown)


aseren:
As discussed with @tmancey, we are closing the PR as won't fix because it makes tests to be executed asynchronously which leads tests to miss some expectations. Need to find another approach keeping tests to be executed synchronously.

PR #36283, opened 2026-05-08
[Polkadot] [Wallet] add asset id to extrinsic signing

### LINKED ISSUE #55152: [Polkadot] [Wallet] Polkadot signing payloads are chain-dependent
When it comes to generating signed extrinsics, the signed payload that forms an extrinsic is chain-dependent. The most readily available example is AssetHub.

The signature payload for AssetHub requires:
```json
      "def": {
        "composite": {
          "fields": [
            {
              "name": "tip",
              "type": 46,
              "typeName": "BalanceOf<T>",
              "docs": []
            },
            {
              "name": "asset_id",
              "type": 357,
              "typeName": "Option<T::AssetId>",
              "docs": []
            }
          ]
        }
      },
```

An asset id is required after the tip. This asset id controls which asset is used to pay the transaction fee. For our purposes, an asset id of 0x00 for DOT should be sufficient.

The definition of an extrinsic can be seen here:
https://github.com/polkadot-js/api/blob/dd76109c422f4f872d25c4939f11a3638b531eea/packages/types-support/src/metadata/v15/asset-hub-polkadot-types.json#L73884

The Extra field lists the required types which leads to:
https://github.com/polkadot-js/api/blob/dd76109c422f4f872d25c4939f11a3638b531eea/packages/types-support/src/metadata/v15/asset-hub-polkadot-types.json#L73707-L73740

We need to update chain metadata parsing to probe for this field and determine if we need to add an asset id or anything else required by the parachain. Though most parachains should be exempt from this requirement and just AssetHub is an outlier because it manages many different assets.

cmazakas:
Resolves https://github.com/brave/brave-browser/issues/55152

Update extrinsic signing routines to add an asset id denoting that the chain's native token (DOT) will be used to pay the transaction fees.

The asset id is specified to appear after the tip. Tests include signed extrinsics accepted by AssetHub Westend.

### DIFF
--- a/components/brave_wallet/browser/internal/polkadot_extrinsic.rs
+++ b/components/brave_wallet/browser/internal/polkadot_extrinsic.rs
@@ -295,6 +295,21 @@ fn scale_encode_mortality(number: u32, mut period: u32) -> [u8; 2] {
     [(encoded & 0xff) as u8, (encoded >> 8) as u8]
 }
 
+fn append_extra(buf: &mut Vec<u8>, asset_tx_payment: bool) {
+    if asset_tx_payment {
+        buf.extend_from_slice(&[
+            0x00, /* tip */
+            0x00, /* asset_id */
+            0x00, /* mode (disable metadata hash verification) */
+        ]);
+    } else {
+        buf.extend_from_slice(&[
+            0x00, /* tip */
+            0x00, /* mode (disable metadata hash verification) */
+        ]);
+    }
+}
+
 /// The definition for the extrinsic signature can be found here:
 /// https://spec.polkadot.network/id-extrinsics#defn-extrinsic-signature
 ///
@@ -362,10 +377,7 @@ fn generate_extrinsic_signature_payload(
     // Write extra data, E.
     buf.extend_from_slice(&scale_encode_mortality(block_number, PERIOD));
     Compact(sender_nonce).encode_to(&mut buf);
-    buf.extend_from_slice(&[
-        0x00, /* tip */
-        0x00, /* mode (disable metadata hash verification) */
-    ]);
+    append_extra(&mut buf, chain_metadata.asset_tx_payment);
 
     // Write R_v.
     buf.extend_from_slice(&spec_version.to_le_bytes());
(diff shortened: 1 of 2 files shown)


components/brave_wallet/browser/internal/polkadot_extrinsic.rs
```diff
@@ -362,10 +377,7 @@ fn generate_extrinsic_signature_payload(
     // Write extra data, E.
     buf.extend_from_slice(&scale_encode_mortality(block_number, PERIOD));
     Compact(sender_nonce).encode_to(&mut buf);
-    buf.extend_from_slice(&[
-        0x00, /* tip */
-        0x00, /* mode (disable metadata hash verification) */
-    ]);
+    append_extra(&mut buf, chain_metadata.asset_tx_payment);
```
netzenbot, line 383: nit: The doc comment above this function describes `E = extra data (mortality, nonce of sender, tip, mode)` but this change conditionally inserts `asset_id` between tip and mode. Consider updating that line to reflect the optional field, e.g. `E = extra data (mortality, nonce of sender, tip, [asset_id], mode)`. ([Update Documentation Alongside Code Changes](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-001))

supermassive (approved):

components/brave_wallet/browser/internal/polkadot_extrinsic.rs
```diff
@@ -295,6 +295,21 @@ fn scale_encode_mortality(number: u32, mut period: u32) -> [u8; 2] {
     [(encoded & 0xff) as u8, (encoded >> 8) as u8]
 }
 
+fn append_extra(buf: &mut Vec<u8>, asset_tx_payment: bool) {
+    if asset_tx_payment {
+        buf.extend_from_slice(&[
+            0x00, /* tip */
+            0x00, /* asset_id */
```
cmazakas, line 305: Add a comment here explaining why the `asset_id` is hard-coded to `0x00`.

cypt4 (approved):

antonok-edm (approved):

netzenbot (approved):

PR #36284, opened 2026-05-08
[AI Chat] Don't show Task UI when executing tools in a non-agent conversation

### LINKED ISSUE #55382: [AI Chat] Task UI shouldn't show for every tool use
When executing any tool, the big Pause and Stop buttons appear.

Image

They should only appear for tasks that tend to take external actions and do so multiple times within a conversation round, such as browser agent tasks.

This is especially annoying when it's waiting for user input, such as the user choice tool or a permission to use a tool.

Steps to reproduce

With an automatically-run tool
1. Start a new conversation in chrome://leo-ai
2. Ask it to calculate something using the code execution tool
3. Observe that the code execution tool is run
4. Observe that the Pause and Stop buttons don't appear

With a tool that requires user input
1. Enable the User Choice Tool via brave://flags/#brave-ai-chat-user-choice-tool
2. Start a new conversation in chrome://leo-ai
3. Ask it to ask you some questions, e.g. "I'm testing the user choice tool, give me a question"
4. Observe that it presents you with some options
5. Observe that the Pause and Stop buttons don't appear

Note that the functionality is completely shared and cross-platform testing not required.

petemill:
It's a pretty cumbersome UI that is more suitable when the conversation is performing potentially disruptive actions, like using the browser.
- Send the current conversation capabilities to the UI (currently a static list, so does not need events to update)
- Test the UI is not shown for chat conversation (we already test that it does show for agent conversations)
- Don't set the ToolUseTaskState to RUNNING when we're waiting on user input (in the future we'll probably add a WAITING state)

- Resolves https://github.com/brave/brave-browser/issues/55382

### DIFF
--- a/browser/ai_chat/ai_chat_conversation_ui_browsertest.cc
+++ b/browser/ai_chat/ai_chat_conversation_ui_browsertest.cc
@@ -28,9 +28,11 @@ class AIChatConversationUIBrowserTest
 };
 
 // Test that when executing a tool, the task UI is not shown for chat
-// conversations (should only be shown for content agent conversations, verified
-// in the task browser tests.
-IN_PROC_BROWSER_TEST_F(AIChatConversationUIBrowserTest, NoTaskStateActions) {
+// conversations (it should only be shown for content agent conversations,
+// which is verified in the task browser tests).
+IN_PROC_BROWSER_TEST_F(
+    AIChatConversationUIBrowserTest,
+    NonAgentConversationDoesNotShowTaskStateActionsWhileToolExecutes) {
   CreateConversationWithMockEngine();
   std::string uuid = conversation_handler_->get_conversation_uuid();
 
@@ -80,7 +82,7 @@ IN_PROC_BROWSER_TEST_F(AIChatConversationUIBrowserTest, NoTaskStateActions) {
   }
 
   // Wait for running state
-  EXPECT_TRUE(base::test::RunUntil([this]() {
+  ASSERT_TRUE(base::test::RunUntil([this]() {
     return GetConversationState()->tool_use_task_state ==
            mojom::TaskState::kRunning;
   }));
--- a/browser/ai_chat/ai_chat_conversation_ui_browsertest.cc
+++ b/browser/ai_chat/ai_chat_conversation_ui_browsertest.cc
@@ -1,0 +1,92 @@
+// Copyright (c) 2026 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#include <memory>
+#include <string>
+
+#include "base/test/run_until.h"
+#include "base/types/expected.h"
+#include "brave/browser/ai_chat/ai_chat_conversation_ui_browsertest_base.h"
+#include "brave/components/ai_chat/core/browser/engine/engine_consumer.h"
+#include "brave/components/ai_chat/core/browser/tools/mock_tool.h"
+#include "brave/components/ai_chat/core/browser/tools/mock_tool_provider.h"
+#include "brave/components/ai_chat/core/browser/tools/tool_utils.h"
+#include "brave/components/ai_chat/core/common/mojom/ai_chat.mojom.h"
+#include "content/public/test/browser_test.h"
+#include "testing/gmock/include/gmock/gmock.h"
+
+namespace ai_chat {
+
+// Tests for general conversation UI rendering and interactions
+class AIChatConversationUIBrowserTest
+    : public AIChatConversationUIBrowserTestBase {
+ public:
+  AIChatConversationUIBrowserTest() = default;
+  ~AIChatConversationUIBrowserTest() override = default;
+};
+
+// Test that when executing a tool, the task UI is not shown for chat
+// conversations (should only be shown for content agent conversations, verified
+// in the task browser tests.
+IN_PROC_BROWSER_TEST_F(AIChatConversationUIBrowserTest, NoTaskStateActions) {
+  CreateConversationWithMockEngine();
+  std::string uuid = conversation_handler_->get_conversation_uuid();
+
+  NavigateToConversationUI(uuid);
+
+  // Inject a mock tool provider with a mock tool so that we can control when
+  // the tool execution completes and inspect the UI.
+  auto* mock_tool_provider = conversation_handler_->AddToolProviderForTesting(
+      std::make_unique<testing::NiceMock<MockToolProvider>>());
+  auto* mock_tool = mock_tool_provider->AddToolForTesting(
+      std::make_unique<testing::NiceMock<MockTool>>("mock_tool", "Mock tool"));
+
+  testing::Sequence tool_call_seq;
+  base::OnceClosure tool_execute;
+
+  // Submit first message
+  {
+    auto generate_future = SetupMockGenerateAssistantResponse(&tool_call_seq);
+    conversation_handler_->SubmitHumanConversationEntry("Do something",
+                                                        std::nullopt);
+    auto callbacks = generate_future->Take();
+
+    // Set up the mock tool to capture its callback so we control execution
+    // timing.
+    EXPECT_CALL(*mock_tool, UseTool)
+        .InSequence(tool_call_seq)
+        .WillOnce(testing::WithArg<1>([&](Tool::UseToolCallback callback) {
+          // Store the callback but don't call it yet so we can inspect the
+          // state whilst waiting for the tool execution to complete.
+          tool_execute = base::BindOnce(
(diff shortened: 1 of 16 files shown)


browser/ai_chat/ai_chat_conversation_ui_browsertest.cc
```diff
@@ -0,0 +1,95 @@
...
+    // Complete first message response
+    std::move(callbacks.completed_callback)
+        .Run(base::ok(
+            EngineConsumer::GenerationResultData(nullptr, std::nullopt)));
+  }
+
+  // Wait for running state
+  EXPECT_TRUE(base::test::RunUntil([this]() {
+    return GetConversationState()->tool_use_task_state ==
+           mojom::TaskState::kRunning;
```
netzenbot, line 87: Use `ASSERT_TRUE` here instead of `EXPECT_TRUE`. If `RunUntil` times out and the task state never reaches `kRunning`, execution continues and the `VerifyElementState` check on the next line produces a meaningless result rather than a clear failure. [TI-034](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-034)

components/ai_chat/core/browser/conversation_handler.cc
```diff
@@ -391,14 +391,18 @@ void ConversationHandler::GetState(GetStateCallback callback) {
   std::ranges::transform(suggestions_, std::back_inserter(suggestions),
                          [](const auto& s) { return s.title; });
 
+  // Build array of capabilities
+  std::vector<mojom::ConversationCapability> capabilities(
+      conversation_capabilities_.begin(), conversation_capabilities_.end());
```
netzenbot: nit: `base::ToVector(conversation_capabilities_)` is preferred over the manual range-constructor form per [CSA-020](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-020). The same issue exists nearby at the `entries_state->conversation_capabilities` assignment, but since that code predates this PR, the fix can be done here for the new addition.

browser/ai_chat/ai_chat_conversation_ui_browsertest.cc
```diff
@@ -0,0 +1,95 @@
...
+        mojom::ConversationEntryEvent::NewToolUseEvent(
+            CreateToolUseEvent("mock_tool", "tool_id_1")),
+        std::nullopt));
+    // Complete first message response
+    std::move(callbacks.completed_callback)
+        .Run(base::ok(
+            EngineConsumer::GenerationResultData(nullptr, std::nullopt)));
+  }
+
+  // Wait for running state
```
netzenbot, line 84: nit: This comment describes a different test entirely — it mentions reloading conversations, edit buttons, and `canSubmitUserEntries`, none of which are tested here. The test is actually verifying that the task state actions UI is hidden for non-agent conversations. Please update the comment to describe what this test actually does. See [CS-019](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-019).

components/ai_chat/core/browser/conversation_handler.cc
```diff
@@ -2252,9 +2249,16 @@ bool ConversationHandler::MaybeRespondToNextToolUseRequest() {
 
       // No user interaction needed - execute tool
 
-      // At this point task state should be kRunning (not paused, stopped, or
-      // none) since we initialized it earlier and checked for pause/stop above.
-      CHECK_EQ(tool_use_task_state_, mojom::TaskState::kRunning);
+      // Initialize the task state for this tool loop if not already set. We do
+      // this after checking for user interaction so that a tool requiring
```
netzenbot, line 2251: nit: The comment says "set it above before we `break`" — avoid positional terms like "above" since code insertions can break the meaning. Reference the specific condition or break site by name or context instead (e.g., "before the `break` in the permission/user-output branches"). See [CS-024](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-024).

components/ai_chat/core/browser/tools/mock_tool_provider.cc
```diff
@@ -0,0 +1,40 @@
...
+  }
+  return result;
+}
+
+void MockToolProvider::StartContentTask(int32_t tab_id) {
+  for (auto& observer : observers_) {
+    observer.OnContentTaskStarted(tab_id);
+  }
+}
+
```
netzenbot, line 28: nit: Use `observers_.Notify(&Observer::OnContentTaskStarted, tab_id)` instead of the manual `for` loop. See [CS-056](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-056).

browser/ai_chat/ai_chat_conversation_ui_browsertest.cc
```diff
@@ -0,0 +1,92 @@
...
+  EXPECT_TRUE(base::test::RunUntil([this]() {
+    return GetConversationState()->tool_use_task_state ==
+           mojom::TaskState::kRunning;
+  }));
```
netzenbot, line 88: nit: `NoTaskStateActions` doesn't convey what type of conversation or under what conditions. A name like `NonAgentConversationDoesNotShowTaskStateActionsWhileToolExecutes` would make the scenario self-documenting. [TI-038](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-038)

components/ai_chat/resources/page/components/input_box/index.tsx
```diff
@@ -256,7 +264,10 @@ function InputBox(props: InputBoxProps) {
     (c) => !c.conversationTurnUuid,
   )
   const showTaskStateActions =
-    props.context.toolUseTaskState !== Mojom.TaskState.kNone
+    conversationState.capabilitiesEnabled.includes(
+      Mojom.ConversationCapability.CONTENT_AGENT,
+    )
+    && props.context.toolUseTaskState !== Mojom.TaskState.kNone
     && props.context.toolUseTaskState !== Mojom.TaskState.kStopped
```
petemill, line 271: This is the change to solve this PR. Everything else is testing or getting the data here.

components/ai_chat/resources/page/stories/components_panel.tsx
```diff
@@ -567,6 +567,7 @@ function StoryContext(
               temporary: argsRef.current.isTemporaryChat,
               toolUseTaskState:
                 Mojom.TaskState[argsRef.current.toolUseTaskState],
+              capabilitiesEnabled: [Mojom.ConversationCapability.CHAT],
```
fallaciousreasoning: do we need this if its in the default state? Maybe we should be using that here?

components/ai_chat/resources/page/api/conversation_api.ts
```diff
@@ -64,6 +64,7 @@ export default function createConversationApi(
             error: Mojom.APIError.None,
             temporary: false,
             toolUseTaskState: Mojom.TaskState.kNone,
+            capabilitiesEnabled: [Mojom.ConversationCapability.CHAT],
```
fallaciousreasoning, line 67: optional, but should this be a set?

fallaciousreasoning: I guess it'd make it harder to auto wrap 🤷

components/ai_chat/core/browser/tools/mock_tool_provider.cc
fallaciousreasoning: Wow, didn't know that was a thing!

fallaciousreasoning (approved):

components/ai_chat/resources/page/api/conversation_api.ts
petemill: I don't think [mojom](https://chromium.googlesource.com/chromium/src/+/HEAD/mojo/public/tools/bindings/README.md#primitive-types) has set

components/ai_chat/resources/page/stories/components_panel.tsx
petemill: This is in the `getState` function, which replaces the placeholder state with whatever getState returns - it needs to return the full `ConversationState`. As this property evolves (deep research or agentic mode become toggleable), we will probably make this customizable in the story args. So I'm not sure we need to make a shared `defaultCapabilitiesEnabled` export in `page/api/conversation_api.ts` just yet, because we probably won't use it?

petemill: Actually I added this control to storybook now because I noticed I'm already passing the conversation capability to the conversation "frame", so I joined them up to a common control.

This shows the Task UI because it's in "CONTENT_AGENT" mode `/?path=/story/ai-chat-conversation--with-multiple-tools-multiple-turns-in-progress&args=capabilitiesEnabled[1]:CONTENT_AGENT;toolUseTaskState:kRunning;isAIChatAgentProfileFeatureEnabled:!true`

PR #36285, opened 2026-05-08
[Ads] Defer bat_ads launch until PolicyService is initialized

bbondy:
## Summary

Make `AdsServiceImpl` observe `policy::PolicyService` and wait for `IsInitializationComplete(POLICY_DOMAIN_CHROME)` before launching the `bat_ads` utility process. Previously the constructor called `MaybeStartBatAdsService()` synchronously, which raced against the policy framework's managed-pref propagation at profile init — `bat_ads` would launch before policies that gate eligibility (e.g. `BraveRewardsDisabled` set by enterprise policy or by BraveOrigin) had landed in the pref store, so ads started for users where Rewards was policy-disabled.

Fixes both the desktop BraveOrigin race and the Android race (where Origin policies arrive even later, after JNI has requested the AdsService) with no BraveOrigin-specific knowledge in the ads code.

## Pattern

Mirrors [`chrome/browser/extensions/forced_extensions/force_installed_tracker.cc`](https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/extensions/forced_extensions/force_installed_tracker.cc): if the PolicyService is already initialized, run `MaybeStartBatAdsService` immediately; otherwise register an observer and proceed from `OnPolicyServiceInitialized`. Tests pass `nullptr` for `PolicyService`, falling back to immediate launch (matches prior behavior).

## Replaces

Supersedes the temporary workaround in #36276 (which read BraveOrigin local state directly to bypass the race). With this PR, the workaround is no longer needed because the gate now waits for the policy framework itself.

## Test plan

- [ ] Brave Origin purchased on desktop, default: ads service does not start.
- [ ] Brave Origin purchased, user toggles Rewards back on: ads service starts.
- [ ] Non-Origin user, no Rewards opt-in: ads service starts (NTP/SearchResult branch).
- [ ] Non-Origin user, opted into Rewards: ads service starts (joined-Rewards branch).
- [ ] Brave Origin purchased on Android: ads service does not start.
- [ ] Enterprise policy `BraveRewardsDisabled=true` on desktop: ads service does not start.

### DIFF
--- a/browser/brave_ads/ads_service_factory.cc
+++ b/browser/brave_ads/ads_service_factory.cc
@@ -28,6 +28,7 @@
 #include "chrome/browser/history/history_service_factory.h"
 #include "chrome/browser/notifications/notification_display_service.h"
 #include "chrome/browser/notifications/notification_display_service_factory.h"
+#include "chrome/browser/policy/profile_policy_connector.h"
 #include "chrome/browser/profiles/profile.h"
 #include "chrome/browser/profiles/profile_manager.h"
 #include "components/keyed_service/content/browser_context_dependency_manager.h"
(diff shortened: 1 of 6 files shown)


bbondy:
Serg basically did this same approach here: https://github.com/brave/brave-core/pull/36151 
so I'm closing this.

PR #36288, opened 2026-05-08
[Docs] Add best practice: don't read policy-backed state in KeyedService ctors

bbondy:
## Summary

- Adds **ARCH-073** to `docs/best-practices/architecture.md` capturing the lesson from #36151.
- A `KeyedService` constructor can run before `PolicyService` has finished merging policy bundles into the managed pref store, so constructor-time checks like `prefs->IsManagedPreference(...)` or reads of policy-mapped prefs evaluate as if no policy is set — even when an admin has applied one. The service then starts (or skips startup) under the wrong assumption for the rest of the session.
- Documents the fix pattern: take `policy::PolicyService*` (nullable in tests) through the constructor, start synchronously when `IsInitializationComplete(POLICY_DOMAIN_CHROME)` is true, otherwise observe and defer to `OnPolicyServiceInitialized`. Notes that `DependsOn(...)` only orders other `KeyedService`s and gives no factory-level handle on policy initialization.

## Test plan

- [x] `npm run format`
- [x] `npm run presubmit`
- [ ] Render `docs/best-practices/architecture.md` and confirm the new ARCH-073 section reads correctly.

### DIFF
--- a/docs/best-practices/architecture.md
+++ b/docs/best-practices/architecture.md
@@ -1501,3 +1501,41 @@ BraveWalletServiceFactory::BuildServiceInstanceForBrowserContext(
 - Any state that can change while the browser is running
 
 If a feature must be fully disable-able at runtime, the service should remain instantiated and expose an `IsEnabled()` accessor, or callers should observe the pref directly.
+
+---
+
+<a id="ARCH-073"></a>
+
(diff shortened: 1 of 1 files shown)


netzenbot (approved):

bbondy:
not sure what is the best practice here so I'll just close this for now.

PR #36290, opened 2026-05-09
[plaster][android] Migrate `AUTOPLAY` patches

### LINKED ISSUE #55385: Migrate AUTOPLAY java patches
There are couple of patches for `AUTOPLAY` on android that tend to break from time to time in a very predictable way, and we should migrate them to plasters.

cdesouza-chromium:
This PR migrates several java patches for `AUTOPLAY` to plaster. It uses
the same previous techniques, where we match the enum block and add at
the end.

The only innovation in this PR as an approach for plaster is the
replacement of the value `NUM_ENTRIES` with `+1` for whatever keys we
are inserting, while at the same time using the previous value of
`NUM_ENTRIES` as the starting value for the first key.

Resolved https://github.com/brave/brave-browser/issues/55385

### DIFF
--- a/rewrite/chrome/android/java/src/org/chromium/chrome/browser/contextmenu/ChromeContextMenuItem.java.toml
+++ b/rewrite/chrome/android/java/src/org/chromium/chrome/browser/contextmenu/ChromeContextMenuItem.java.toml
@@ -1,0 +1,45 @@
+# Copyright (c) 2026 The Brave Authors. All rights reserved.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+[[substitution]]
+description = 'Register COPY_CLEAN_LINK at the end of the @IntDef list'
+re_pattern = '(@IntDef\(\{.+?),?(\n\s*\}\))'
+re_flags = ['DOTALL']
+replace = '''\1,
+        Item.COPY_CLEAN_LINK,\2'''
+
+[[substitution]]
+description = '''Define COPY_CLEAN_LINK at the end of @interface Item
+
+A `// ALWAYS UPDATE!` comment sits between the last upstream entry and the
+sentinel, so we capture it separately (`\2`) and keep it adjacent to
+`NUM_ENTRIES`. The indent before `NUM_ENTRIES` is captured into `\3` and
+reused for the inserted line, instead of hardcoding it.
+
+```
+int COPY_CLEAN_LINK = 43;
+// ALWAYS UPDATE!
+int NUM_ENTRIES = COPY_CLEAN_LINK + 1;
+```
+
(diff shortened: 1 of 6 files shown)


netzenbot (approved):

rewrite/chrome/android/java/src/org/chromium/chrome/browser/contextmenu/ChromeContextMenuItem.java.toml
```diff
@@ -0,0 +1,43 @@
...
+
+[[substitution]]
+description = '''Define COPY_CLEAN_LINK at the end of @interface Item
+
+We have to make the match a bit more lenient for this case, becuase the outcome
+has to ignore the comment line in between so we get something like this:
+
+```
+int COPY_CLEAN_LINK = 43;
+// ALWAYS UPDATE!
```
AlexeyBarabash, line 23: what does this comment mean?

AlexeyBarabash (approved):
LGTM, left the question about `// ALWAYS UPDATE!` which I didn't understand

rewrite/chrome/android/java/src/org/chromium/chrome/browser/contextmenu/ChromeContextMenuItem.java.toml
cdesouza-chromium: The original looks like this:

```java
        int INSPECT_ELEMENT = 42;
        // ALWAYS UPDATE!
        int NUM_ENTRIES = 43;
```

I'm just mentioning in this comment that this is kept the same. We don't touch any line in-between.

AlexeyBarabash: ah, missed that is was the original comment, sorry, thanks

rewrite/chrome/android/java/src/org/chromium/chrome/browser/contextmenu/ChromeContextMenuPopulator.java.toml
```diff
@@ -0,0 +1,28 @@
...
+
+[[substitution]]
+description = '''Define COPY_CLEAN_LINK at the end of @interface Action
+
+Notice that the value for `COPY_CLEAN_LINK` is being whatever was the value for
+`NUM_ENTRIES`, which allows us to take over the end of the list.
+'''
+re_pattern = '(@interface Action \{.+?;\n)\s+int NUM_ENTRIES = (\d+);'
+re_flags = ['DOTALL']
+replace = '''\1            int COPY_CLEAN_LINK = \2;
```
bridiver: it would be better to capture the whitespace before `int NUM_ENTRIES` and insert it back here again

cdesouza-chromium: Fixed.

bridiver (approved):

PR #36291, opened 2026-05-09
Add `ObliviousHttpChunkProcessor` to enable chunked OHTTP

DJAndries:

### DIFF
--- a/components/oblivious_http/BUILD.gn
+++ b/components/oblivious_http/BUILD.gn
@@ -15,10 +15,10 @@ source_set("oblivious_http") {
   ]
 
   public_deps = [
+    "//brave/services/network/public/mojom",
     "//mojo/public/cpp/bindings",
     "//net",
     "//services/network/public/cpp",
-    "//services/network/public/mojom",
   ]
 
   deps = [ "//base" ]
@@ -36,7 +36,6 @@ source_set("unit_tests") {
     "//mojo/public/cpp/bindings",
     "//net",
     "//net/third_party/quiche",
-    "//services/network/public/mojom",
     "//testing/gtest",
   ]
 }
--- a/components/oblivious_http/BUILD.gn
+++ b/components/oblivious_http/BUILD.gn
@@ -1,0 +1,42 @@
+# Copyright (c) 2026 The Brave Authors. All rights reserved.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+source_set("oblivious_http") {
+  visibility = [
+    ":unit_tests",
+    "//services/network:network_service",
+  ]
+
+  sources = [
+    "oblivious_http_chunk_processor.cc",
+    "oblivious_http_chunk_processor.h",
+  ]
+
+  public_deps = [
+    "//mojo/public/cpp/bindings",
+    "//net",
+    "//services/network/public/cpp",
+    "//services/network/public/mojom",
+  ]
+
+  deps = [ "//base" ]
+}
+
+source_set("unit_tests") {
+  testonly = true
+
+  sources = [ "oblivious_http_chunk_processor_unittest.cc" ]
+
+  deps = [
+    ":oblivious_http",
+    "//base",
+    "//base/test:test_support",
+    "//mojo/public/cpp/bindings",
+    "//net",
+    "//net/third_party/quiche",
+    "//services/network/public/mojom",
+    "//testing/gtest",
+  ]
+}
--- a/components/oblivious_http/oblivious_http_chunk_processor.cc
+++ b/components/oblivious_http/oblivious_http_chunk_processor.cc
@@ -5,6 +5,7 @@
 
 #include "brave/components/oblivious_http/oblivious_http_chunk_processor.h"
 
+#include "base/check.h"
 #include "base/memory/ptr_util.h"
 #include "base/strings/string_number_conversions.h"
 
@@ -38,23 +39,22 @@ ObliviousHttpChunkProcessor::Create(
     return nullptr;
   }
 
-  // Construct the decoder first so we have a stable address to pass to
-  // ChunkedObliviousHttpClient::Create, which holds a non-owning reference.
-  auto decoder = base::WrapUnique(new ObliviousHttpChunkProcessor(
+  auto processor = base::WrapUnique(new ObliviousHttpChunkProcessor(
       std::move(chunk_client_remote), std::move(on_request_complete)));
 
   auto client_result = quiche::ChunkedObliviousHttpClient::Create(
-      *pub_key, key_config, decoder.get());
+      *pub_key, key_config, processor.get());
   if (!client_result.ok()) {
     return nullptr;
   }
 
-  decoder->ohttp_client_.emplace(std::move(*client_result));
-  return decoder;
+  processor->ohttp_client_.emplace(std::move(*client_result));
+  return processor;
 }
 
 std::optional<std::string> ObliviousHttpChunkProcessor::EncryptRequest(
     std::string_view plaintext) {
+  CHECK(ohttp_client_);
   auto result =
       ohttp_client_->EncryptRequestChunk(plaintext, /*is_final_chunk=*/true);
   if (!result.ok()) {
@@ -65,6 +65,7 @@ std::optional<std::string> ObliviousHttpChunkProcessor::EncryptRequest(
 
 void ObliviousHttpChunkProcessor::OnDataReceived(std::string_view data,
                                                  base::OnceClosure resume) {
+  CHECK(ohttp_client_);
   auto status = ohttp_client_->DecryptResponse(data, /*end_stream=*/false);
   if (!status.ok()) {
     has_error_ = true;
@@ -73,6 +74,7 @@ void ObliviousHttpChunkProcessor::OnDataReceived(std::string_view data,
 }
 
 void ObliviousHttpChunkProcessor::OnComplete(bool success) {
+  CHECK(ohttp_client_);
   if (!success) {
     NotifyURLLoaderComplete(/*success=*/false);
     return;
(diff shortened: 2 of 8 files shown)


components/oblivious_http/BUILD.gn
```diff
@@ -0,0 +1,42 @@
...
+    ":oblivious_http",
+    "//base",
+    "//base/test:test_support",
+    "//mojo/public/cpp/bindings",
+    "//net",
+    "//net/third_party/quiche",
+    "//services/network/public/mojom",
+    "//testing/gtest",
+  ]
+}
```
netzenbot, line 41: `//base` needs to be in `public_deps`, not `deps`. The header exposes `base::OnceCallback` and `base::OnceClosure` in its public constructor, `Create()`, `OnDataReceived()`, and `OnRetry()` signatures, so any consumer of this target will need the transitive include path. ([Use `public_deps` for Header-File Includes in BUILD.gn](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-024))

components/oblivious_http/oblivious_http_chunk_processor.cc
```diff
@@ -0,0 +1,200 @@
...
+}
+
+absl::Status ObliviousHttpChunkProcessor::OnFinalResponseHeader(
+    absl::string_view name,
+    absl::string_view value) {
+  pending_headers_.emplace_back(std::string(name), std::string(value));
+  return absl::OkStatus();
+}
+
+absl::Status ObliviousHttpChunkProcessor::OnFinalResponseHeadersDone() {
```
netzenbot, line 140: nit: `DCHECK(ohttp_client_)` (also on lines 151 and 160) guards a condition that should be impossible in any build — calling these methods without a successful `Create()` is a programming error, not a debug-only concern. Per [CS-025](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-025), use `CHECK` for conditions that should never happen in any build.

components/oblivious_http/oblivious_http_chunk_processor_unittest.cc
```diff
@@ -0,0 +1,373 @@
...
+  // single network delivery.
+  RunRoundTrip(kLargeJsonResponse, net::HTTP_OK,
+               {{"content-type", "application/json"}},
+               /*bhttp_chunk_count=*/10, /*network_chunk_count=*/2);
+  ASSERT_TRUE(complete_result_.has_value());
+  EXPECT_TRUE(*complete_result_);
+  EXPECT_EQ(kLargeJsonResponse, received_server_response_body_);
+}
+
+}  // namespace oblivious_http
```
netzenbot, line 350: The `on_request_complete` callback uses the verbose `base::RunLoop` + `BindLambdaForTesting` + `Quit()` pattern, which is explicitly flagged in our async testing guidelines. Since the callback has signature `void(std::optional<std::string>)`, this can be replaced with `base::test::TestFuture<std::optional<std::string>>` — call `future.GetCallback()` when constructing the processor and `future.Get()` to wait for the result. See [TA-009](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-async.md#TA-009).

components/oblivious_http/oblivious_http_chunk_processor.cc
```diff
@@ -0,0 +1,200 @@
...
+
+  decoder->ohttp_client_.emplace(std::move(*client_result));
+  return decoder;
+}
+
+std::optional<std::string> ObliviousHttpChunkProcessor::EncryptRequest(
+    std::string_view plaintext) {
+  DCHECK(ohttp_client_);
+  auto result =
+      ohttp_client_->EncryptRequestChunk(plaintext, /*is_final_chunk=*/true);
```
aseren, line 59: I suppose this is not in UI thread?

components/oblivious_http/oblivious_http_chunk_processor.cc
```diff
@@ -0,0 +1,200 @@
...
+  if (!result.ok()) {
+    return std::nullopt;
+  }
+  return std::move(*result);
+}
+
+void ObliviousHttpChunkProcessor::OnDataReceived(std::string_view data,
+                                                 base::OnceClosure resume) {
+  DCHECK(ohttp_client_);
```
aseren: just CHECK ?

aseren: when ohttp_client_ can be null?

components/oblivious_http/oblivious_http_chunk_processor.cc
DJAndries: Correct, it executes in the network service, using the same threads as the existing encrypt calls in the upstream OHTTP client

components/oblivious_http/oblivious_http_chunk_processor.cc
DJAndries: Agreed, this should be `CHECK`. `ohttp_client_` can be null if it wasn't constructed via `Create`

components/oblivious_http/oblivious_http_chunk_processor.cc
```diff
@@ -0,0 +1,200 @@
...
+  for (const auto& [name, value] : pending_headers_) {
+    builder.AddHeader(name, value);
+  }
+  headers_ = builder.Build();
+  return absl::OkStatus();
+}
+
+absl::Status ObliviousHttpChunkProcessor::OnBodyChunk(
+    absl::string_view body_chunk) {
+  chunk_client_->OnBodyChunk(std::string(body_chunk));
```
aseren, line 152: We do a copy of body_chunk here, is it really needed?
maybe OnBodyChunk can work with std::string_view, so we can pass body_chunk to it directly without wrapping to std::string.

components/oblivious_http/oblivious_http_chunk_processor.cc
```diff
@@ -0,0 +1,200 @@
...
+  if (!status.ok()) {
+    has_error_ = true;
+  }
+  std::move(resume).Run();
+}
+
+void ObliviousHttpChunkProcessor::OnComplete(bool success) {
+  DCHECK(ohttp_client_);
+  if (!success) {
+    NotifyURLLoaderComplete(false);
```
aseren: nit: using false without comment is not consistent, as we add comment to magic constants in file: /*end_stream=*/false

components/oblivious_http/oblivious_http_chunk_processor.h
```diff
@@ -0,0 +1,117 @@
...
+namespace oblivious_http {
+
+// Implements quiche::ObliviousHttpChunkHandler,
+// quiche::BinaryHttpResponse::IndeterminateLengthDecoder::MessageSectionHandler,
+// and network::SimpleURLLoaderStreamConsumer to support chunked OHTTP
+// responses. Owned by RequestState and must not outlive it.
+//
+// Use Create() to construct — it handles key parsing and client creation.
+// Call EncryptRequest() once after Create() to encrypt the request payload.
+class ObliviousHttpChunkProcessor
```
aseren, line 33: would be helpful for reviewers if add a comment of why we add `ObliviousHttpChunkProcessor`, why it is needed

DJAndries: updated the comment

components/oblivious_http/oblivious_http_chunk_processor.cc
DJAndries: done

components/oblivious_http/oblivious_http_chunk_processor.cc
DJAndries: yes, we need a `std::string` in order to convert to `mojom::ByteString`

components/oblivious_http/oblivious_http_chunk_processor.cc
DJAndries: updated to make the constructor private, so `Create` is the only way to instantiate the class. Removed `DCHECK`s since `ohttp_client_` will never be null.

netzenbot (approved):

components/oblivious_http/oblivious_http_chunk_processor.cc
aseren: hmm, so it is double copy then

aseren: is there any way of converting std::string_view to mojom::ByteString without converting it to std::string?

components/oblivious_http/oblivious_http_chunk_processor.cc
aseren: ohttp_client_ is still optional, can it be changed to non-optional?

components/oblivious_http/oblivious_http_chunk_processor.cc
DJAndries: i don't think so. at least, i couldn't find a way.

components/oblivious_http/oblivious_http_chunk_processor.cc
DJAndries: this is tricky... in `Create`, we need to construct the processor first, so we have a pointer to pass to `ChunkedObliviousHttpClient::Create`. in that short period of time, the member has to be nullopt. I'm not sure if there's a way to get around this unfortunately

aseren: so ohttp_client_ can be nullopt then, right? if so I think we need to put CHECK here.

aseren:
approved with one comment about CHECK

components/oblivious_http/oblivious_http_chunk_processor.cc
DJAndries: `ObliviousHttpChunkProcessor::Create` is the only way to construct the processor. If ohttp client construction does not succeed, then we will not return the processor. So all methods can expect `ohttp_client_` to be present.

aseren (approved):

components/oblivious_http/oblivious_http_chunk_processor.cc
DJAndries: added CHECKs in all methods where ohttp_client is used

components/oblivious_http/oblivious_http_chunk_processor_unittest.cc
DJAndries: updated

PR #36293, opened 2026-05-09
[AI Chat] Add manager/loader for OHTTP config

DJAndries:


netzenbot (approved):

components/ai_chat/core/browser/engine/oblivious_http_config_manager.cc
```diff
@@ -0,0 +1,209 @@
...
+    return;
+  }
+
+  ScopedDictPrefUpdate update(profile_prefs_,
+                              prefs::kAIChatObliviousHttpKeyConfigs);
+  base::DictValue entry;
+  entry.Set(kKeyConfigKey, *key_config_b64);
+  entry.Set(kEndpointUrlKey, endpoint_url.spec());
+  entry.Set(kPrefExpiresAtField,
+            base::TimeToValue(base::Time::Now() + kKeyConfigExpiresAfter));
```
fallaciousreasoning, line 202: Hey, does the server guarantee that these are valid at least 3 days? Is there a reason this doesn't come back from the server?

fallaciousreasoning (approved):
this LGTM, nice work @DJAndries - thanks for taking the time to walk me through how this will be used.

components/ai_chat/core/browser/engine/oblivious_http_config_manager.cc
DJAndries: NEAR has communicated to us that key configurations will last forever, but I decided to add a TTL here just in case.

PR #36294, opened 2026-05-09
[AI Chat] Prepare `EngineConsumerOAIRemote`/`OAIAPIClient` for OHTTP support

DJAndries:

Change `EngineConsumerOAIRemote` to take `ModelOptions` instead of `CustomModelOptions`.
This will allow us to pass `LeoModelOptions`, for models that support OHTTP.
E2EE has a lot of overlap with the BYOM functionality, so it makes sense to utilize `EngineConsumerOAIRemote` for this purpose


netzenbot (approved):

components/ai_chat/core/browser/engine/engine_consumer_oai.cc
```diff
@@ -39,23 +39,28 @@ namespace ai_chat {
...
+  return opts.model_system_prompt && !opts.model_system_prompt->empty();
 }
 
 }  // namespace
 
 EngineConsumerOAIRemote::EngineConsumerOAIRemote(
-    const mojom::CustomModelOptions& model_options,
+    const mojom::ModelOptions& model_options,
     scoped_refptr<network::SharedURLLoaderFactory> url_loader_factory,
+    network::NetworkContextGetter network_context_getter,
```
fallaciousreasoning, line 55: this param is unused?

fallaciousreasoning: is it necessary in a followup?

fallaciousreasoning (approved):

components/ai_chat/core/browser/engine/engine_consumer_oai.cc
DJAndries: good catch, it is unused, but will be used in https://github.com/brave/brave-core/pull/36295. I just placed it here to reduce the diff size of the subsequent PR

mkarolin (approved):
`strings`++

components/ai_chat/core/browser/engine/engine_consumer_oai.cc
```diff
@@ -75,11 +80,13 @@ bool EngineConsumerOAIRemote::RequiresClientSideTitleGeneration() const {
 
 void EngineConsumerOAIRemote::UpdateModelOptions(
     const mojom::ModelOptions& options) {
-  if (options.is_custom_model_options()) {
-    model_options_ = *options.get_custom_model_options();
-    max_associated_content_length_ =
-        model_options_.max_associated_content_length;
+  if (!options.is_custom_model_options() ||
+      !model_options_->is_custom_model_options()) {
+    return;
```
cdesouza-chromium: Here we are early returning, but above we are CHECKing for `is_custom_model_options()`. Is this discrepancy intentional?

components/ai_chat/core/browser/engine/engine_consumer_oai.cc
```diff
@@ -39,23 +39,28 @@ namespace ai_chat {
...
+    network::NetworkContextGetter network_context_getter,
+    AIChatCredentialManager* credential_manager,
     ModelService* model_service,
     PrefService* prefs)
     : EngineConsumer(model_service, prefs) {
-  model_options_ = model_options;
-  max_associated_content_length_ = model_options.max_associated_content_length;
-
-  // Initialize the API client
+  model_options_ = model_options.Clone();
```
cdesouza-chromium: if we are gonna call `.Clone()` at this point, wouldn't it be better to take `mojom::ModelOptions model_options` by value, and let the caller decide if they are moving with `std::move()` or using `Clone()`? Same for below...

components/ai_chat/core/browser/engine/engine_consumer_oai.cc
```diff
@@ -39,23 +39,28 @@ namespace ai_chat {
...
+  CHECK(model_options_->is_custom_model_options());
+  max_associated_content_length_ =
+      model_options_->get_custom_model_options()->max_associated_content_length;
```
cdesouza-chromium, line 63: It is not clear to me why we have to duplicate this information in a local field called `max_associated_content_length_`, rather than just accessing it as `model_options_->get_custom_model_options()->max_associated_content_length` when needed.

cdesouza-chromium (approved):
LGTM % comments I've left. Also, the PR/commit message could improve a bit by having a better description describing "what" and "why".

components/ai_chat/core/browser/engine/engine_consumer_oai.cc
DJAndries: The `ModelService` retains ownership of the `Model`/`ModelOptions` after constructing this class, so we have to clone.

components/ai_chat/core/browser/engine/engine_consumer_oai.cc
DJAndries: yeah i agree. I hope it's acceptable that we leave this as a follow up in order to keep this PR focused, since this field already exists.

components/ai_chat/core/browser/engine/engine_consumer_oai.cc
DJAndries: it is not intentional. removed this additional check.

DJAndries:
> Also, the PR/commit message could improve a bit by having a better description describing "what" and "why".

Amended the PR/commit description, thanks.

components/ai_chat/core/browser/engine/engine_consumer_oai.cc
cdesouza-chromium: To give an example with a smaller function. Right now you have this:

```cxx
void EngineConsumerOAIRemote::UpdateModelOptions(
    const mojom::ModelOptions& options) {
  if (!options.is_custom_model_options()) {
    return;
  }
  model_options_ = options.Clone();
```

So anyone calling the function above passes an object as a const ref, and then you clone.

But it could be just this:

```cxx
void EngineConsumerOAIRemote::UpdateModelOptions(
    mojom::ModelOptions options) {
  if (!options.is_custom_model_options()) {
    return;
  }
  model_options_ = std::move(options);
```

Now the caller can choose if they are gonna `std::move(options)` into this function, or `options.Clone()` it, because they need to keep a copy for themselves. If you are taking ownership, prefer to pass things by move, does that make sense?

components/ai_chat/core/browser/engine/engine_consumer_oai.cc
cdesouza-chromium: Sure thing.

components/ai_chat/core/browser/engine/engine_consumer_oai.cc
DJAndries: yes, updated

PR #36300, opened 2026-05-09
Fix VisibilityTimer visibilitychange handling

### LINKED ISSUE #55394: [NTP][Brave News] VisibilityTimer ignores visibilitychange due to typo
## Summary
`VisibilityTimer` registers `visiblitychange` instead of `visibilitychange`, so the visibility event handler never runs.

As a result, timers can continue counting while the document is hidden and only react to other state transitions.

## Impact
- Incorrect viewed-timer behavior while tab/page is backgrounded.
- Potentially inaccurate metrics and unnecessary background work for visibility-gated callbacks.

## Scope
- Fix listener name typo in VisibilityTimer add/remove paths.
- Add regression unit test covering hidden->visible transition behavior with fake timers.

## Proposed fix
- Rename listener key to `visibilitychange` in start/stop tracking.
- Add test: hidden state prevents expiry and visible state resumes timer.

## Tracking PR
- brave/brave-core#36300

crbender:
## Summary
- fix typo in `VisibilityTimer` event listener from `visiblitychange` to `visibilitychange`
- add focused regression test to verify hidden state pauses timer and visible state resumes it

## Why
The typo prevented document visibility events from reaching `VisibilityTimer`, allowing view timers to continue while hidden.

## Linked issue
- Fixes brave/brave-browser#55394

## Scope fence (anti-slop)
- only listener event string and targeted unit test were changed
- no behavior refactor outside existing timer lifecycle
- no consumer-side changes in ad/article components

## Tests
- added unit test: `components/common/visibilityTimer.test.ts`
  - starts timer while visible
  - dispatches `visibilitychange` to hidden and verifies callback does not fire past threshold
  - dispatches `visibilitychange` back to visible and verifies timer resumes and callback fires

## Manual verification status
- Local automated test execution is currently blocked in this checkout (`jest: command not found`).
- VS Code diagnostics on touched files show no errors.

### DIFF
--- a/components/common/visibilityTimer.test.ts
+++ b/components/common/visibilityTimer.test.ts
@@ -1,0 +1,46 @@
+// Copyright (c) 2026 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import VisibilityTimer from './visibilityTimer'
+
+describe('VisibilityTimer', () => {
+  function setVisibilityState(state: DocumentVisibilityState) {
+    Object.defineProperty(document, 'visibilityState', {
+      configurable: true,
+      value: state,
+    })
+  }
+
+  beforeEach(() => {
+    jest.useFakeTimers()
+    setVisibilityState('visible')
+  })
+
+  afterEach(() => {
+    jest.useRealTimers()
+  })
+
+  it('does not fire while hidden and resumes after visible', () => {
+    const onTimerExpired = jest.fn()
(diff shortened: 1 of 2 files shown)


zenparsing (approved):
It looks like this module is only used on the old NTP and on the Brave News "Ads.tsx" component (which is no longer active).

tmancey:
> It looks like this module is only used on the old NTP and on the Brave News "Ads.tsx" component (which is no longer active).

@aseren can you PTAL and remove any dead ads code. Thanks

aseren (approved):
`components/brave_news/browser/resources/feed/Ad.tsx` is a dead code now, because Brave News support was removed here: https://github.com/brave/brave-core/pull/32359. 
Filed an issue for `components/brave_news/browser/resources/feed/Ad.tsx` removal: https://github.com/brave/brave-browser/issues/55419

PR #36301, opened 2026-05-09
Respect vertical tabs on right for toggle placement

### LINKED ISSUE #55395: [Vertical tabs] Place expand/collapse button on the right side when tabs are set to be in the right side
### Description

Follow up to this issue. 
https://github.com/brave/brave-browser/issues/52873

When vertical tabs are set to the right, the button to expand/collapse ends up being on the opposite edge of the browser. To fix, let's position it to the left side of the hamburger menu when that is the setting.

### Steps to reproduce
1. Enable vertical tabs
2. Go to settings > appearance
3. Set vertical tabs to the right.
4. See how the button to expand/collapse ends up in the opposite side

### Actual result

Image

### Expected result

Button should be on the right side of the toolbar

Image

### Reproduces how often

Easily reproduced

1.92.34 Chromium: 148.0.7778.96 (Official Build) nightly (arm64) 

### Channel information

- [ ] release (stable)
- [ ] beta
- [ ] nightly

### Reproducibility

### Miscellaneous information

_No response_

aguscruiz:
Resolves https://github.com/brave/brave-browser/issues/55395

# Reposition vertical-tab expand/collapse toggle when tabs are shown on the right

## Summary

When Brave vertical tabs use **“Show on the right”** (`brave.tabs.vertical_tabs_on_right`), the toolbar control that expands/collapses the vertical tab strip is moved from the **leading edge** of the toolbar (pinned to **child index `0`** in the toolbar container) to the **trailing** cluster, **immediately before the app menu**, so it stays next to the strip and the main menu. In **RTL** UI (`--force-ui-direction=rtl`), the same physical-side rule applies — the placement choice is inverted so the toggle visually lines up with the strip, since the toolbar container's horizontal `views::FlexLayout` mirrors child indices but the strip is positioned by physical coordinates.

## Motivation

With the strip on the right edge of the window, keeping the toggle on the left felt disconnected from the feature. Placing it beside the menu aligns the control with the strip and matches how other trailing toolbar actions are grouped.

## Implementation

- **`BraveToolbarView`** (`browser/ui/views/toolbar/brave_toolbar_view.h`, `browser/ui/views/toolbar/brave_toolbar_view.cc`):
  - Observe **`kVerticalTabsOnRight`** via `BooleanPrefMember` and call **`UpdateVerticalTabTogglePlacement()`** when it changes (inside the existing **`SupportsBraveVerticalTabs(browser_)`** block).
  - **`show_vertical_tabs_`** callback also refreshes placement when vertical tabs mode is toggled (same guard).
  - **`UpdateVerticalTabTogglePlacement()`** uses `views::View::ReorderChildView` on the same `container_view` as the rest of the toolbar:
    - Decision: `place_near_app_menu = tabs::utils::IsVerticalTabOnRight(browser_) != base::i18n::IsRTL()`. The vertical tab strip is placed by physical position and does not mirror in RTL, while the toolbar's `FlexLayout` mirrors logical child indices — XOR'ing with RTL keeps the toggle on the same physical side as the strip across all four (LTR/RTL × left/right) combinations.
    - **Near app menu:** target index is **`GetIndexOf(GetAppMenuButton()) - 1`** (same slot strategy as other trailing controls and the avatar/menu ordering).
    - **Leading edge:** target index is **`0`**. This avoids depending on **`GetIndexOf(back_)`**: forward is already pinnable upstream and back could become pinnable later; tying placement to back's index risks **`CHECK`** failures or odd ordering when navigation-button pinning changes.
  - Run placement at the **end of `Init()`**, **after** the avatar is reordered before the app menu, **only when** `tabs::utils::SupportsBraveVerticalTabs(browser_)` (same condition under which the toggle is created), then set `brave_initialized_` so startup order matches runtime pref updates.
  - **`UpdateVerticalTabTogglePlacement()`** is assumed to run only when vertical tabs are supported and the toggle exists as a child of `container_view`; unexpected layout state is treated as a programming error via **`CHECK`** on the toggle, location bar parent, **`GetIndexOf`** for the toggle, and (when placing near the app menu) **`GetIndexOf`** for the app menu plus **`CHECK_GT`** on that index so `menu_idx - 1` cannot underflow.

- **Tests:** `browser/ui/views/toolbar/brave_toolbar_view_browsertest.cc`
  - **`VerticalTabTogglePlacementRespectsTabsOnRight`** (LTR) — asserts **`child index 0`** when **“show on the left”**, **`menu_index - 1`** when **“show on the right”**, that moving from left to right increases the toggle's index, and uses **`<<`** diagnostic context on assertions (TI-014).
  - **`BraveToolbarViewRTLTest.RtlInvertsVerticalTabTogglePlacement`** — adds `--force-ui-direction=rtl` via `SetUpCommandLine` and asserts the *inverted* indices: **`menu_index - 1`** when “show on the left”, **`0`** when “show on the right”, and that flipping the pref *decreases* the toggle's logical index (mirror of the LTR case).

## How to test manually

1. Enable **vertical tabs** and keep **“Show on the left”** — confirm the toggle stays at the **leading edge** of the toolbar (first control in the container in typical LTR layouts).
2. Switch to **“Show on the right”** — confirm the toggle appears **on the right**, **just left of the ≡ menu**.
3. Toggle vertical tabs off/on and flip left/right again — confirm placement updates without a restart.
4. Launch with `--force-ui-direction=rtl` and repeat 1–2: the toggle should follow the strip's *physical* side (visually left/right matches the strip), even though the toolbar's child order is mirrored.

## Related

- Preference: `brave.tabs.vertical_tabs_on_right` (`browser/ui/tabs/brave_tab_prefs.h`).
- Settings copy: “Show on the right” in the tabs appearance page.

### Actual result

Image

### Expected result

Button should be on the right side of the toolbar

Image

### DIFF
--- a/browser/ui/views/toolbar/brave_toolbar_view.cc
+++ b/browser/ui/views/toolbar/brave_toolbar_view.cc
@@ -674,9 +674,13 @@ void BraveToolbarView::UpdateVerticalTabTogglePlacement() {
     return;
   }
 
-  DCHECK(location_bar_view_ && location_bar_view_->parent());
+  CHECK(location_bar_view_);
+  CHECK(location_bar_view_->parent());
   views::View* container_view = location_bar_view_->parent();
 
+  // The toggle may not yet be parented to `container_view` during early
+  // initialization or if it was removed by another flow; bail out instead of
+  // reordering an unrelated view.
   const std::optional<size_t> toggle_idx =
       container_view->GetIndexOf(vertical_tab_toggle_);
   if (!toggle_idx.has_value()) {
@@ -685,12 +689,17 @@ void BraveToolbarView::UpdateVerticalTabTogglePlacement() {
 
   size_t target_idx = 0;
   if (tabs::utils::IsVerticalTabOnRight(browser_)) {
+    // Place the toggle just before the app menu. If the menu isn't in the
+    // container yet, or is already the first child, there is no valid slot
+    // to the left of it.
     const auto menu_idx = container_view->GetIndexOf(GetAppMenuButton());
     if (!menu_idx.has_value() || *menu_idx == 0) {
       return;
     }
     target_idx = *menu_idx - 1;
   } else {
+    // Place the toggle just before the back button. If back isn't a direct
+    // child of the container yet, defer placement until it is.
     const auto back_idx = container_view->GetIndexOf(back_);
     if (!back_idx.has_value()) {
       return;
--- a/browser/ui/views/toolbar/brave_toolbar_view.cc
+++ b/browser/ui/views/toolbar/brave_toolbar_view.cc
@@ -270,6 +270,7 @@ void BraveToolbarView::Init() {
                                  [](BraveToolbarView* self) {
                                    self->UpdateHorizontalPadding();
                                    self->UpdateVerticalTabToggleVisibility();
+                                   self->UpdateVerticalTabTogglePlacement();
                                  },
                                  base::Unretained(this)));
     show_title_bar_on_vertical_tabs_.Init(
@@ -280,6 +281,10 @@ void BraveToolbarView::Init() {
     vertical_tabs_collapsed_.Init(
         brave_tabs::kVerticalTabsCollapsed, profile->GetPrefs(),
         base::BindRepeating(&BraveToolbarView::UpdateVerticalTabToggleState,
+                            base::Unretained(this)));
+    vertical_tabs_on_right_.Init(
+        brave_tabs::kVerticalTabsOnRight, profile->GetPrefs(),
+        base::BindRepeating(&BraveToolbarView::UpdateVerticalTabTogglePlacement,
                             base::Unretained(this)));
 #if BUILDFLAG(IS_LINUX)
     use_custom_chrome_frame_.Init(
@@ -381,6 +386,8 @@ void BraveToolbarView::Init() {
     container_view->ReorderChildView(
         avatar, *container_view->GetIndexOf(GetAppMenuButton()) - 1);
   }
+
+  UpdateVerticalTabTogglePlacement();
 
   brave_initialized_ = true;
   UpdateHorizontalPadding();
@@ -660,6 +667,42 @@ void BraveToolbarView::UpdateVerticalTabToggleVisibility() {
 
   vertical_tab_toggle_->SetVisible(
       tabs::utils::ShouldShowBraveVerticalTabs(browser_));
+}
+
+void BraveToolbarView::UpdateVerticalTabTogglePlacement() {
+  if (!vertical_tab_toggle_ || display_mode_ != DisplayMode::kNormal) {
+    return;
+  }
+
+  DCHECK(location_bar_view_ && location_bar_view_->parent());
+  views::View* container_view = location_bar_view_->parent();
+
+  const std::optional<size_t> toggle_idx =
+      container_view->GetIndexOf(vertical_tab_toggle_);
+  if (!toggle_idx.has_value()) {
+    return;
+  }
+
+  size_t target_idx = 0;
+  if (tabs::utils::IsVerticalTabOnRight(browser_)) {
+    const auto menu_idx = container_view->GetIndexOf(GetAppMenuButton());
+    if (!menu_idx.has_value() || *menu_idx == 0) {
+      return;
+    }
+    target_idx = *menu_idx - 1;
+  } else {
+    const auto back_idx = container_view->GetIndexOf(back_);
+    if (!back_idx.has_value()) {
+      return;
+    }
+    target_idx = *back_idx > 0 ? *back_idx - 1 : 0;
+  }
+
+  if (*toggle_idx == target_idx) {
+    return;
+  }
+
+  container_view->ReorderChildView(vertical_tab_toggle_, target_idx);
 }
 
 void BraveToolbarView::UpdateVerticalTabToggleState() {
--- a/browser/ui/views/toolbar/brave_toolbar_view.cc
+++ b/browser/ui/views/toolbar/brave_toolbar_view.cc
@@ -692,9 +692,11 @@ void BraveToolbarView::UpdateVerticalTabTogglePlacement() {
     CHECK_GT(*menu_idx, 0u);
     target_idx = *menu_idx - 1;
   } else {
-    const auto back_idx = container_view->GetIndexOf(back_);
-    CHECK(back_idx.has_value());
-    target_idx = *back_idx > 0 ? *back_idx - 1 : 0;
+    // Pin index 0 for the leading edge so placement does not depend on the
+    // back button's child index (forward is pinnable today; back could become
+    // pinnable upstream). Computing `back_idx - 1` would risk CHECK failures or
+    // surprising order if navigation-button pinning changes.
+    target_idx = 0;
   }
 
   if (*toggle_idx == target_idx) {
--- a/browser/ui/views/toolbar/brave_toolbar_view.cc
+++ b/browser/ui/views/toolbar/brave_toolbar_view.cc
@@ -387,7 +387,9 @@ void BraveToolbarView::Init() {
         avatar, *container_view->GetIndexOf(GetAppMenuButton()) - 1);
   }
 
-  UpdateVerticalTabTogglePlacement();
+  if (tabs::utils::SupportsBraveVerticalTabs(browser_)) {
+    UpdateVerticalTabTogglePlacement();
+  }
 
   brave_initialized_ = true;
   UpdateHorizontalPadding();
@@ -670,40 +672,28 @@ void BraveToolbarView::UpdateVerticalTabToggleVisibility() {
 }
 
 void BraveToolbarView::UpdateVerticalTabTogglePlacement() {
-  if (!vertical_tab_toggle_ || display_mode_ != DisplayMode::kNormal) {
-    return;
-  }
-
+  // Callers must gate on `SupportsBraveVerticalTabs(browser_)`, which both
+  // ensures we're in `DisplayMode::kNormal` (other modes early-return from
+  // `Init()` before the toggle is created) and guarantees the toggle has
+  // been added as a child of `container_view`.
+  CHECK(vertical_tab_toggle_);
   CHECK(location_bar_view_);
   CHECK(location_bar_view_->parent());
   views::View* container_view = location_bar_view_->parent();
 
-  // The toggle may not yet be parented to `container_view` during early
-  // initialization or if it was removed by another flow; bail out instead of
-  // reordering an unrelated view.
   const std::optional<size_t> toggle_idx =
       container_view->GetIndexOf(vertical_tab_toggle_);
-  if (!toggle_idx.has_value()) {
-    return;
-  }
+  CHECK(toggle_idx.has_value());
 
   size_t target_idx = 0;
   if (tabs::utils::IsVerticalTabOnRight(browser_)) {
-    // Place the toggle just before the app menu. If the menu isn't in the
-    // container yet, or is already the first child, there is no valid slot
-    // to the left of it.
     const auto menu_idx = container_view->GetIndexOf(GetAppMenuButton());
-    if (!menu_idx.has_value() || *menu_idx == 0) {
-      return;
-    }
+    CHECK(menu_idx.has_value());
+    CHECK_GT(*menu_idx, 0u);
     target_idx = *menu_idx - 1;
   } else {
-    // Place the toggle just before the back button. If back isn't a direct
-    // child of the container yet, defer placement until it is.
     const auto back_idx = container_view->GetIndexOf(back_);
-    if (!back_idx.has_value()) {
-      return;
-    }
+    CHECK(back_idx.has_value());
     target_idx = *back_idx > 0 ? *back_idx - 1 : 0;
   }
 
--- a/browser/ui/views/toolbar/brave_toolbar_view.cc
+++ b/browser/ui/views/toolbar/brave_toolbar_view.cc
@@ -12,6 +12,7 @@
 #include "base/check.h"
 #include "base/check_op.h"
 #include "base/functional/bind.h"
+#include "base/i18n/rtl.h"
 #include "brave/app/brave_command_ids.h"
 #include "brave/app/vector_icons/vector_icons.h"
 #include "brave/browser/ui/tabs/brave_tab_prefs.h"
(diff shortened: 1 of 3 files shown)


browser/ui/views/toolbar/brave_toolbar_view.cc
```diff
@@ -662,6 +669,42 @@ void BraveToolbarView::UpdateVerticalTabToggleVisibility() {
...
+    const auto menu_idx = container_view->GetIndexOf(GetAppMenuButton());
+    if (!menu_idx.has_value() || *menu_idx == 0) {
+      return;
+    }
+    target_idx = *menu_idx - 1;
+  } else {
+    const auto back_idx = container_view->GetIndexOf(back_);
+    if (!back_idx.has_value()) {
+      return;
+    }
```
netzenbot: nit: The `!toggle_idx.has_value()`, `!menu_idx.has_value() || *menu_idx == 0`, and `!back_idx.has_value()` branches are non-obvious — `GetIndexOf` can return nullopt if the view isn't a direct child of the container at that moment. A brief comment on each branch would clarify the defensive rationale. See [CS-021](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-021).

browser/ui/views/toolbar/brave_toolbar_view.cc
```diff
@@ -662,6 +669,42 @@ void BraveToolbarView::UpdateVerticalTabToggleVisibility() {
       tabs::utils::ShouldShowBraveVerticalTabs(browser_));
 }
 
+void BraveToolbarView::UpdateVerticalTabTogglePlacement() {
+  if (!vertical_tab_toggle_ || display_mode_ != DisplayMode::kNormal) {
+    return;
```
netzenbot: nit: Per [CS-025](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-025), use `CHECK` instead of `DCHECK` for conditions that should never happen in any build. If `location_bar_view_` or its parent being null here represents a programming error that can't occur in a correct program, `CHECK` is more appropriate than `DCHECK` (which is stripped in release builds).

browser/ui/views/toolbar/brave_toolbar_view_browsertest.cc
```diff
@@ -607,6 +607,34 @@ IN_PROC_BROWSER_TEST_F(BraveToolbarViewTest,
...
+  const auto ix_on_right = container->GetIndexOf(toggle);
+  ASSERT_TRUE(ix_on_right.has_value());
+
+  views::View* menu = toolbar_button_provider_->GetAppMenuButton();
+  ASSERT_TRUE(menu);
+  const auto menu_ix = container->GetIndexOf(menu);
+  ASSERT_TRUE(menu_ix.has_value() && *menu_ix > 0);
+
+  EXPECT_EQ(*ix_on_right, *menu_ix - 1);
+  EXPECT_LT(*ix_left_side, *ix_on_right);
```
netzenbot: nit: The `EXPECT_EQ(*ix_on_right, *menu_ix - 1)` and `EXPECT_LT(*ix_left_side, *ix_on_right)` assertions have no `<<` diagnostic context. On failure you'll only see raw index values with no indication of which pref state is active. See nearby `ToolbarPaddingMatchesFrameLayoutParams` for the pattern: `<< "after enabling vertical tabs, hiding title bar"`. https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-014

browser/ui/views/toolbar/brave_toolbar_view.cc
aguscruiz: Done in the latest revision — replaced the single `DCHECK(location_bar_view_ && location_bar_view_->parent())` with two `CHECK`s (one per invariant), since these are real preconditions in any build, not debug-only assumptions.

browser/ui/views/toolbar/brave_toolbar_view.cc
aguscruiz: Done — added a brief comment before each defensive early-return explaining what each branch guards against:

- toggle not yet parented to `container_view` (early init / removed by another flow)
- app menu not in the container yet, or already the first child (no valid slot to its left)
- back button not yet a direct child (defer placement until it is)

browser/ui/views/toolbar/brave_toolbar_view_browsertest.cc
aguscruiz: Done — every `ASSERT_TRUE` / `EXPECT_EQ` / `EXPECT_LT` in the new test now has `<<` diagnostic context. The two index comparisons also stream the actual values (toggle/menu/left/right indexes) so a failure surfaces the offending state instead of bare numeric mismatches, matching the pattern in `ToolbarPaddingMatchesFrameLayoutParams`.

browser/ui/views/toolbar/brave_toolbar_view.cc
```diff
@@ -382,6 +387,8 @@ void BraveToolbarView::Init() {
         avatar, *container_view->GetIndexOf(GetAppMenuButton()) - 1);
   }
 
+  UpdateVerticalTabTogglePlacement();
```
sangwoo108: I think this should be guarded with `  if (tabs::utils::SupportsBraveVerticalTabs(browser_)) {` as the button is created only under this condition

browser/ui/views/toolbar/brave_toolbar_view.cc
```diff
@@ -662,6 +669,51 @@ void BraveToolbarView::UpdateVerticalTabToggleVisibility() {
       tabs::utils::ShouldShowBraveVerticalTabs(browser_));
 }
 
+void BraveToolbarView::UpdateVerticalTabTogglePlacement() {
+  if (!vertical_tab_toggle_ || display_mode_ != DisplayMode::kNormal) {
+    return;
```
sangwoo108: If we call this only when   `if (tabs::utils::SupportsBraveVerticalTabs(browser_)) { `, we don't need to filter display_mode_ and vertical tab should exist. So CHECK(vertical_tab_tggoel_) makes more sense

browser/ui/views/toolbar/brave_toolbar_view.cc
```diff
@@ -662,6 +669,51 @@ void BraveToolbarView::UpdateVerticalTabToggleVisibility() {
...
+  }
+
+  CHECK(location_bar_view_);
+  CHECK(location_bar_view_->parent());
+  views::View* container_view = location_bar_view_->parent();
+
+  // The toggle may not yet be parented to `container_view` during early
+  // initialization or if it was removed by another flow; bail out instead of
+  // reordering an unrelated view.
```
sangwoo108: I'm not sure if this is true. For me vertical tab toggle must exist at this point. maybe CHECK(toggle_idx) is better.

browser/ui/views/toolbar/brave_toolbar_view.cc
```diff
@@ -662,6 +669,51 @@ void BraveToolbarView::UpdateVerticalTabToggleVisibility() {
...
+
+  size_t target_idx = 0;
+  if (tabs::utils::IsVerticalTabOnRight(browser_)) {
+    // Place the toggle just before the app menu. If the menu isn't in the
+    // container yet, or is already the first child, there is no valid slot
+    // to the left of it.
+    const auto menu_idx = container_view->GetIndexOf(GetAppMenuButton());
+    if (!menu_idx.has_value() || *menu_idx == 0) {
+      return;
+    }
```
sangwoo108: Also, I can't come up with when this can be true. CHECK(menu_idx) could be better?

browser/ui/views/toolbar/brave_toolbar_view.cc
```diff
@@ -662,6 +669,51 @@ void BraveToolbarView::UpdateVerticalTabToggleVisibility() {
...
+    target_idx = *menu_idx - 1;
+  } else {
+    // Place the toggle just before the back button. If back isn't a direct
+    // child of the container yet, defer placement until it is.
+    const auto back_idx = container_view->GetIndexOf(back_);
+    if (!back_idx.has_value()) {
+      return;
+    }
```
sangwoo108: ditto. I can't come up with cases where back button isn't created yet. CHECK(back_idx) could be better?

browser/ui/views/toolbar/brave_toolbar_view.cc
aguscruiz: Done — wrapped the call site in `Init()` with `if (tabs::utils::SupportsBraveVerticalTabs(browser_))` so we never reorder anything when the toggle wasn't created. The two pref-change callsites at lines 273/287 are already inside that same guard, so all entrypoints to `UpdateVerticalTabTogglePlacement()` now share the same precondition.

browser/ui/views/toolbar/brave_toolbar_view.cc
aguscruiz: Done — dropped the early-return guard entirely. Now that all callers are gated on `SupportsBraveVerticalTabs(browser_)`, the toggle is guaranteed to exist and we're guaranteed to be in `DisplayMode::kNormal` (other modes early-return from `Init()` before the toggle is ever created). Replaced with `CHECK(vertical_tab_toggle_)` and added a comment documenting the caller contract.

browser/ui/views/toolbar/brave_toolbar_view.cc
aguscruiz: Agreed — replaced the early-return with `CHECK(toggle_idx.has_value())`. With the new caller contract (gated on `SupportsBraveVerticalTabs`), the toggle is added to `container_view` in `Init()` and never reparented elsewhere, so a missing index here is a real programming error worth crashing on.

browser/ui/views/toolbar/brave_toolbar_view.cc
aguscruiz: Done — split into `CHECK(menu_idx.has_value())` and `CHECK_GT(*menu_idx, 0u)`. The first asserts the app menu is a direct child (always true for a normal-mode toolbar after `Init()`); the second guards the `*menu_idx - 1` underflow and crashes loudly if the app menu somehow ends up as the very first child, instead of silently no-op'ing.

browser/ui/views/toolbar/brave_toolbar_view.cc
aguscruiz: Done — replaced with `CHECK(back_idx.has_value())`. Same reasoning: the back button is a standard upstream child of the toolbar container, so it's always present in any normal-mode toolbar by the time we reach this code path.

browser/ui/views/toolbar/brave_toolbar_view.cc
```diff
@@ -662,6 +671,39 @@ void BraveToolbarView::UpdateVerticalTabToggleVisibility() {
...
+  size_t target_idx = 0;
+  if (tabs::utils::IsVerticalTabOnRight(browser_)) {
+    const auto menu_idx = container_view->GetIndexOf(GetAppMenuButton());
+    CHECK(menu_idx.has_value());
+    CHECK_GT(*menu_idx, 0u);
+    target_idx = *menu_idx - 1;
+  } else {
+    const auto back_idx = container_view->GetIndexOf(back_);
+    CHECK(back_idx.has_value());
+    target_idx = *back_idx > 0 ? *back_idx - 1 : 0;
```
simonhong: How about simply use `0` index always here?
For now, back button's pinned statue is not configurable. fwd button is pinnable.
Not sure but if upstream makes back button's pinned stabe configurable, this code could make crash. I think having `0` always is safe always.

simonhong (commented):
Please double check that vertical tabs toolbar button is not hidden when browser window is in minimum width.
It would be great to have test case for this. <= it's optional as it's not directly related with this PR.

One more thing I want to mention is `RTL` mode. `--force-ui-direction=rtl`
Button position also should be synced with veritcal tab position. it's another issue though.
Maybe you want to fix this issue next :) 
image

browser/ui/views/toolbar/brave_toolbar_view.cc
aguscruiz: Done — left-side placement now always uses child index `0` (leading toolbar edge) instead of `back_idx - 1`, so we don’t depend on the back button’s index if pinning/upstream ordering changes. Dropped the `GetIndexOf(back_)` / `CHECK` on that path. Added a short comment in code and an `EXPECT_EQ(ix, 0)` in `VerticalTabTogglePlacementRespectsTabsOnRight` to lock in the behavior.

aguscruiz:
@simonhong — addressed the RTL note in `92c260cfea2` instead of deferring to a follow-up, so the toggle now stays on the same physical side as the strip with `--force-ui-direction=rtl`.

**Diagnosis**
- `tabs::utils::IsVerticalTabOnRight()` returns the pref verbatim, and `VerticalTabStripWidgetDelegateView::UpdateVerticalTabBounds()` places the strip by physical coordinates (`host_bounds.right() - width`) — i.e. `kVerticalTabsOnRight` means *physically* right in any UI direction.
- The toolbar container, on the other hand, lays out children with a horizontal `views::FlexLayout`, which mirrors logical child indices in RTL: logical index 0 renders on the visual right, and the app-menu's logical position appears on the visual left.
- So the previous code placed the toggle adjacent to the menu's *logical* slot, which was on the wrong *physical* side in RTL.

**Fix (`BraveToolbarView::UpdateVerticalTabTogglePlacement`)**

```cpp
const bool place_near_app_menu =
    tabs::utils::IsVerticalTabOnRight(browser_) != base::i18n::IsRTL();
```

- LTR + on right → trailing slot (`menu_idx - 1`) → visually right → matches strip.
- LTR + on left  → leading slot (`0`) → visually left → matches strip.
- RTL + on right → leading slot (`0`) → visually right (mirrored) → matches strip.
- RTL + on left  → trailing slot (`menu_idx - 1`) → visually left (mirrored) → matches strip.

**Coverage**
- New browser test `BraveToolbarViewRTLTest.RtlInvertsVerticalTabTogglePlacement` runs under `--force-ui-direction=rtl` and asserts:
  - `kVerticalTabsOnRight=false` → toggle at `menu_idx - 1` (was leading edge in LTR).
  - `kVerticalTabsOnRight=true`  → toggle at child index `0` (was `menu_idx - 1` in LTR).
  - Flipping the pref *decreases* the toggle's logical index (mirror of the LTR test).
- Existing LTR test (`VerticalTabTogglePlacementRespectsTabsOnRight`) still passes unchanged.

Note: minimum-width visibility behavior is unchanged by this PR — the toggle still uses the same flex behavior as the other toolbar buttons — but I tested it and with minimum width the button still shows

simonhong (approved):
++

rafisics:
Waiting to get it fixed in the stable release for Linux as well.

PR #36310, opened 2026-05-11
[plaster] Remove `count = 1` lines

cdesouza-chromium:
Removing these lines so we can introduce `count` value checks to steer
users towards fixed counts only when needed, and away from `count=0`. At
the moment removing these as they are redundant.

### DIFF
--- a/rewrite/chrome/android/java/src/org/chromium/chrome/browser/toolbar/ToolbarManager.java.toml
+++ b/rewrite/chrome/android/java/src/org/chromium/chrome/browser/toolbar/ToolbarManager.java.toml
@@ -9,4 +9,3 @@
 description = 'Remove NTP check from toolbar longpress condition'
 re_pattern = '(private boolean shouldSuppressToolbarLongPress\(\)[\s\S]+?)\|\| \(getNewTabPageForCurrentTab\(\) != null\);'
 replace = '\1;'
-count = 1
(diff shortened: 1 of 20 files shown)


netzenbot (approved):

goodov:
> to steer users towards fixed counts only when needed

I don't quite understand what's the final solution. I'd expect the plaster patches to work this way:
1. no `count` implies `count=1`
2. match the pattern through all file and expect exact `count` matches
3. fail if `matches != count`
4. allow `count=0/-1` for "I don't care, replace everything" scenario, but discourage and require a closer look at those patches.

Is this how it'll work? Because accidental replaces where you don't expect them, but which are happily applied, might be a nasty bug generator.

cdesouza-chromium:
> > to steer users towards fixed counts only when needed
> 
> I don't quite understand what's the final solution. I'd expect the plaster patches to work this way:
> 
> 1. no `count` implies `count=1`
> 2. match the pattern through all file and expect exact `count` matches
> 3. fail if `matches != count`
> 4. allow `count=0/-1` for "I don't care, replace everything" scenario, but discourage and require a closer look at those patches.
> 
> Is this how it'll work? Because accidental replaces where you don't expect them, but which are happily applied, might be a nasty bug generator.

That's exactly what I'm laying the work to roll out, and in fact right now it is how it works. Count is extremly important, and we should see when it slides. That's a strength plaster has over patches/chromium_src.

no count = count 1
count = N heavily encouraged for anything not default 1 to catch slides.
count = 0 mean we don't care about count - heavily discouraged

I have a warning that I will roll out in plaster that will force anyone using `count  = 0` to add a reason for it in the comments to, but I'm trying to clean up these noisy and redundant count from the codebase, so I can write a dos and donts on this as well.

goodov (approved):

PR #36273, opened 2026-05-08
Removes importing payment info from Chrome profile importer

mkarolin:
Resolves https://github.com/brave/brave-browser/issues/55377

PR #36311, opened 2026-05-11
Hide payment checkbox from import dialog

simonhong:
Resolves: f/u PR to https://github.com/brave/brave-core/pull/36273

Manual test:
1. Add any payments info to Chrome and close
2. Go to brave://settings/importData and check no `Payment methods` checkbox


simonhong:
@mkarolin I think we can uplift this PR together with yours.

netzenbot (approved):

mkarolin (approved):
Ah, I was wondering if there was something like this left, but didn't find at the time. Thanks for catching this!

PR #36316, opened 2026-05-11
[query-filter] Add functionality to extract blocklisted params from rules for spec

### LINKED ISSUE #55412: Create a new functionality to extract the params for a spec
We should have a helper class / API to extract all the blocklisted params from a given set of |rules| and spec.  See this [change](https://github.com/brave/brave-core/pull/36222/changes#diff-67555d88e9b6cea4c29482c81035edb152f3a715e34c29a3a1a0e90dff48b1b3R17) for example.

r0hit0303:
Adds a new helper class to facilitate extractions of all blocked listed params from a given rule set for a spec. This change is a smaller change of https://github.com/brave/brave-core/pull/36222 where it specifies more details on how the helper method would be used one the client side. 

This change also adds units tests for the same. 

Resolves https://github.com/brave/brave-browser/issues/55412

### DIFF
--- a/components/query_filter/common/internal/schema_utils.cc
+++ b/components/query_filter/common/internal/schema_utils.cc
@@ -30,17 +30,13 @@ base::flat_set<std::string> GetBlocklistedParamsForSpec(
   // would need to be rewritten as ".*://.*\\.instagram\\.com/.*" first.
   const url_pattern_index::UrlPattern::UrlInfo url_info(url);
 
-  // A helper method to detect blank empty strings.
-  constexpr auto is_blank = [](std::string_view s) {
-    return base::TrimWhitespaceASCII(s, base::TrimPositions::TRIM_ALL).empty();
-  };
-
   // Go over rules that matches on the |url|.
   for (const auto& rule : rules) {
     // Check if the rule explicitly "excludes" the |url| from consideration.
-    const auto& exclude_itr = std::ranges::find_if(
-        rule.exclude, [&url_info, &is_blank](const std::string& str) {
-          return !is_blank(str) &&
+    const auto& exclude_itr =
+        std::ranges::find_if(rule.exclude, [&url_info](const std::string& str) {
+          return !base::TrimWhitespaceASCII(str, base::TrimPositions::TRIM_ALL)
+                      .empty() &&
                  url_pattern_index::UrlPattern(str).MatchesUrl(url_info);
         });
     // |url| excluded from consideration for the current |rule|
@@ -49,9 +45,10 @@ base::flat_set<std::string> GetBlocklistedParamsForSpec(
     }
 
     // Check if the rule explicitly "includes" the |url| for consideration.
-    const auto& include_itr = std::ranges::find_if(
-        rule.include, [&url_info, &is_blank](const std::string& str) {
-          return !is_blank(str) &&
+    const auto& include_itr =
+        std::ranges::find_if(rule.include, [&url_info](const std::string& str) {
+          return !base::TrimWhitespaceASCII(str, base::TrimPositions::TRIM_ALL)
+                      .empty() &&
                  url_pattern_index::UrlPattern(str).MatchesUrl(url_info);
         });
     if (include_itr != rule.include.cend()) {
--- a/components/query_filter/common/internal/schema_utils.cc
+++ b/components/query_filter/common/internal/schema_utils.cc
@@ -1,0 +1,70 @@
+// Copyright (c) 2026 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#include "brave/components/query_filter/common/internal/schema_utils.h"
+
+#include <algorithm>
+#include <string_view>
+
+#include "components/url_pattern_index/url_pattern.h"
+#include "third_party/abseil-cpp/absl/strings/ascii.h"
+#include "url/gurl.h"
+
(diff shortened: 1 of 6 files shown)


components/query_filter/common/internal/schema_utils.cc
```diff
@@ -0,0 +1,70 @@
...
+  // because RE2 requires explicit ".*" for wildcards, so the pattern above
+  // would need to be rewritten as ".*://.*\\.instagram\\.com/.*" first.
+  const url_pattern_index::UrlPattern::UrlInfo url_info(url);
+
+  // A helper method to detect blank empty strings.
+  constexpr auto is_blank = [](std::string_view s) {
+    return s.empty() ||
+           std::ranges::all_of(s.begin(), s.end(), [](unsigned char c) {
+             return absl::ascii_isspace(c);
+           });
```
cdesouza-chromium: This can be simplified to:

```cxx
std::ranges::all_of(s, absl::ascii_isspace, [](unsigned char c) { return c; });
```

However I wonder if this can be safely simplified with trimming: https://source.chromium.org/chromium/chromium/src/+/main:base/strings/string_util.h;l=274

```cxx
if (base::TrimString(input, base::kWhitespaceASCII, base::TRIM_ALL).empty())
```

components/query_filter/common/internal/schema_utils.h
```diff
@@ -0,0 +1,35 @@
...
+// contributes its params only when |spec| matches at least one include pattern
+// and no exclude pattern. Blank/whitespace-only patterns are treated as
+// non-matching. Returns an empty set if |spec| is not a valid URL or no rule
+// matches.
+// TODO(https://github.com/brave/brave-browser/issues/55305): This is not used
+// at the moment. This would be plugged in to the
+// components/query_filter/browser/utils.cc when we support the query filter
+// component.
+base::flat_set<std::string> GetBlocklistedParamsForSpec(
+    const std::vector<schema::Rule>& rules,
```
cdesouza-chromium: it's better to take `base::span<schema::Rule>`, so we don't make this specific to `vector`.

components/query_filter/common/internal/schema_utils.cc
```diff
@@ -0,0 +1,70 @@
...
+  };
+
+  // Go over rules that matches on the |url|.
+  for (const auto& rule : rules) {
+    // Check if the rule explicitly "excludes" the |url| from consideration.
+    const auto& exclude_itr = std::ranges::find_if(
+        rule.exclude.cbegin(), rule.exclude.cend(),
```
cdesouza-chromium:
```cxx
    const auto& exclude_itr = std::ranges::find_if(rule.exclude,
```

This is a ranges call, you can pass the whole container rather than a pair of iterators.

components/query_filter/common/internal/schema_utils.cc
```diff
@@ -0,0 +1,70 @@
...
+                 url_pattern_index::UrlPattern(str).MatchesUrl(url_info);
+        });
+    // |url| excluded from consideration for the current |rule|
+    if (exclude_itr != rule.exclude.cend()) {
+      continue;
+    }
+
+    // Check if the rule explicitly "includes" the |url| for consideration.
+    const auto& include_itr = std::ranges::find_if(
+        rule.include.cbegin(), rule.include.cend(),
```
cdesouza-chromium: Same

cdesouza-chromium (commented):
Looks good... a few questions.

components/query_filter/common/internal/schema_utils.cc
r0hit0303: Good suggestion. It seems we have [TrimWhitespaceASCII](https://source.chromium.org/chromium/chromium/src/+/main:base/strings/string_util.h;bpv=1;bpt=1;l=303) that does exactly that with fewer arguments.

components/query_filter/common/internal/schema_utils.cc
r0hit0303: Done

components/query_filter/common/internal/schema_utils.cc
r0hit0303: Done

components/query_filter/common/internal/schema_utils.h
r0hit0303: Done

r0hit0303:
Thanks @cdesouza-chromium for your review. Please find the asked changes in the latest ps.

components/query_filter/common/internal/schema_utils.cc
```diff
@@ -0,0 +1,66 @@
...
+  // UrlPattern handles glob-style wildcard patterns used in rule.include and
+  // rule.exclude (e.g "*://*.instagram.com/*"). We can't use RE2 directly
+  // because RE2 requires explicit ".*" for wildcards, so the pattern above
+  // would need to be rewritten as ".*://.*\\.instagram\\.com/.*" first.
+  const url_pattern_index::UrlPattern::UrlInfo url_info(url);
+
+  // A helper method to detect blank empty strings.
+  constexpr auto is_blank = [](std::string_view s) {
+    return base::TrimWhitespaceASCII(s, base::TrimPositions::TRIM_ALL).empty();
+  };
```
cdesouza-chromium: nit: Don't you think calling `base::TrimWhitespaceASCII` is easier for the reader, rather than using `is_blank` as lambda and an indirection?

cdesouza-chromium (approved):
LGTM. One nit.

components/query_filter/common/internal/schema_utils.cc
r0hit0303: You are right. I missed to simplify it before uploading the last patch-set. Done

PR #36317, opened 2026-05-11
[git-cr] Make `commit` its own module

cdesouza-chromium:
This change makes the code for `git-cr` tools more modular. This will
help with test correctness ultimately, as now we are depending fully on
`Repository` for paths, specially install paths, which eliminates a
recurring issue where tests do seem to interfere with the current
install path `brave-core`.


AlexeyBarabash (approved):
++

PR #36318, opened 2026-05-11
[query-filter] Add query filter test helper

### LINKED ISSUE #55418: Create a query filter test helper
We currently have multiple clients like [brave_site_hacks_network_delegate_helper_unittest](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/browser/net/brave_site_hacks_network_delegate_helper_unittest.cc) and [brave_site_hacks_network_delegate_helper_browsertest](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/browser/net/brave_site_hacks_network_delegate_helper_browsertest.cc) outside of the `components/query_filter` target that relies on checking the query stripping functionality. 

Their tests relies on the fact that we have the [hardcoded lists](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/components/query_filter/browser/utils.cc?L28) baked in the code. However, with [QueryFilterComponent](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/components/query_filter/common/features.h?L14) we would remove that list and instead rely on [QueryFilterData#PopulateDataFromComponent](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/components/query_filter/browser/query_filter_data.h?L44) similar to what we do [here](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/components/query_filter/browser/query_filter_data_unittest.cc?L79-L93). 

Therefore, we need a public test helper for these external clients that can reliably setup the query_filter environment for their test.

r0hit0303:
This changes adds a new test helper to allow populating the query filter rules when the query filter component is enabled. This helps outside test clients that relied on hardcoded rules previously to test query stripping functionality. The hardcoded rules are going away once the `QueryFilterComponent` flag is enabled. 

This change updates the two existing test suites to support the change when we migrate to the new query filter component. Please see the larger [change](https://github.com/brave/brave-core/pull/36222) to see the full integration. 

Note that for now the test suites would *continue* to test the hardcoded rules despite this change as the actual query filtering is not yet supported. However, the plan is that once we do, these test suites would not need to be modified and they will automatically tests the new behaviour.

Resolves https://github.com/brave/brave-browser/issues/55418


components/query_filter/browser/test_support/query_filter_test_helper.h
```diff
@@ -0,0 +1,25 @@
...
+namespace test {
+// Populates the query filter component with default set of rules.
+// Must only be called with kQueryFilterComponent enabled; calling otherwise
+// will crash.
+void SetupWithDefaultQueryFilterRules();
+
+// Clears all rules from the query filter component.
+// Must only be called with kQueryFilterComponent enabled; calling otherwise
+// will crash.
+void RemoveDefaultRules();
```
cdesouza-chromium: The best practice here would be something like:

```cxx
class ScopedTestingQueryFilterRules {
 public:
  ScopedTestingQueryFilterRules() {  // in ..cc
    ASSERT_TRUE(QueryFilterData::GetInstance()->PopulateDataFromComponent(
         kDefaultQueryFilterRules));
  }
  ~ScopedTestingQueryFilterRules() {  // in ..cc
    QueryFilterData::GetInstance()->ResetRulesForTesting()
  }

  ScopedTestingQueryFilterRules(const ScopedQueryFilterRules&) = delete;
  ScopedTestingQueryFilterRules& operator=(const ScopedQueryFilterRules&) = delete;
};
```

Then have something like

```cxx
ScopedTestingQueryFilterRules testing_filter_rules_;
```

If you absolutely need to control when the rules are created/removed, then use an `optional<ScopedTestingQueryFilterRules>`.

Then you can initialise with `SetUpOnMainThread()` with `testing_filter_rules_.emplace();`, and you can trigger destruction in `TearDownInProcessBrowserTestFixture` with `testing_filter_rules_ = std::nullopt;`

cdesouza-chromium (approved):
LGTM % one comment.

components/query_filter/browser/test_support/query_filter_test_helper.h
r0hit0303: Thanks for the suggestion. It's a really cool trick to handle setup and cleanup implicitly via ctor and dtor. I went ahead with using std::optional and hooking into SetUp and TearDown for emplace and reset calls. Done

r0hit0303:
Thanks @cdesouza-chromium for the review.

PR #36324, opened 2026-05-11
[CodeHealth] Clean up brave/browser/ui/views/frame target

### LINKED ISSUE #55423: [CodeHealth] Clean up browser/ui//views/frame target
### Description

There are currently several sources in the `frame` folder that are still included in the `//brave/browser/ui` target. These sources should be moved to the `frame` target.

zenparsing:
Resolves https://github.com/brave/brave-browser/issues/55423

The following targets have been converted to the header/impl pattern in order to avoid circular dependencies on the frame target:

* `//brave/browser/ui/views/frame/split_view`
* `//brave/browser/ui/views/frame/vertical_tabs`

### DIFF
--- a/browser/ui/config.gni
+++ b/browser/ui/config.gni
@@ -52,7 +52,9 @@ if (toolkit_views) {
     "//brave/browser/ui/views:layout_provider",
     "//brave/browser/ui/views/frame",
     "//brave/browser/ui/views/frame/split_view",
+    "//brave/browser/ui/views/frame/split_view:split_view_impl",
     "//brave/browser/ui/views/frame/vertical_tabs",
+    "//brave/browser/ui/views/frame/vertical_tabs:vertical_tabs_impl",
     "//brave/browser/ui/views/location_bar",
     "//brave/browser/ui/views/page_info",
   ]
--- a/browser/sources.gni
+++ b/browser/sources.gni
@@ -813,11 +813,13 @@ if (enable_brave_ads) {
 if (toolkit_views) {
   brave_chrome_browser_deps += [
     "//brave/browser/ui/sidebar:sidebar_impl",
+    "//brave/browser/ui/views/frame",
     "//brave/browser/ui/views/location_bar",
     "//brave/browser/ui/views/page_info",
   ]
   brave_chrome_browser_allow_circular_includes_from += [
     "//brave/browser/ui/sidebar:sidebar_impl",
+    "//brave/browser/ui/views/frame",
     "//brave/browser/ui/views/location_bar",
   ]
 }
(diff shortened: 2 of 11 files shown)


netzenbot (approved):

browser/ui/config.gni
```diff
@@ -52,7 +52,9 @@ if (toolkit_views) {
     "//brave/browser/ui/views:layout_provider",
     "//brave/browser/ui/views/frame",
     "//brave/browser/ui/views/frame/split_view",
+    "//brave/browser/ui/views/frame/split_view:split_view_impl",
     "//brave/browser/ui/views/frame/vertical_tabs",
+    "//brave/browser/ui/views/frame/vertical_tabs:vertical_tabs_impl",
```
simonhong, line 57: Can we remove public targets (`frame/split_view`, `frame/vertical_tab`) from `brave_ui_allow_circular_includes_from` ?

zenparsing: No, because even the public targets (the headers) include from `//chrome/browser/ui`. For example:

`brave_multi_contents_view.h` has

```
#include "chrome/browser/ui/views/frame/multi_contents_view.h"
```

which is in the target:

```
//chrome/browser/ui/BUILD.gn
```

simonhong: IIUC, we separate targets into `public` and `impl` to make `public` as not-circular with `ui` target.
If both(public and impl) still have circular deps with `ui`, there is no reason to have `public`.
IMO, we don't need to make all headers as a public if it adds circular now.
I think we could remove `brave_multi_contents_view.h` from public. WDYT? cc @bridiver

zenparsing: The public/impl pattern is required in this case to avoid circular dependencies between, e.g.

`//brave/browser/ui/views/frame/split_view` and `//brave/browser/ui/views/frame`

or

`//brave/browser/ui/views/frame/vertical_tabs` and `//brave/browser/ui/views/frame`

If we want to have `tab_strip_placement_coordinator` (see https://github.com/brave/brave-core/pull/36255) in the `frame` target, we'll need to break at least this dep cycle somehow.

simonhong: IMO, this PR does `frame` target cleanup quite aggressively.
Still upstream's `ui` target includes `browser_view.h` and etc.
Instead of changing a lot now, how about making public `frame` target and and it only includes `tab_strip_placement_coordinator.h`. And adding `tab_strip_placement_coordinator.cc` to renamed `frame_impl` target from current `frame`.
As `tab_strip_placement_coordinator.h` doesn't have any deps with `ui` target, maybe we could remove circular deps. This could work?

zenparsing: i'll give it a try!

PR #36334, opened 2026-05-11
[Containers] Make container icon on url bar bigger

sangwoo108:
Previously, the icon used for PageActionView was too small, because the size includes the background size. Introduces a new generator function that generates the icon without the background.

Part of https://github.com/brave/brave-browser/issues/55407

### DIFF
--- a/browser/ui/containers/containers_icon_generator.cc
+++ b/browser/ui/containers/containers_icon_generator.cc
@@ -13,7 +13,6 @@
 #include "brave/components/containers/core/browser/temporary_container.h"
 #include "brave/components/vector_icons/vector_icons.h"
 #include "ui/gfx/canvas.h"
-#include "ui/gfx/color_palette.h"
 #include "ui/gfx/geometry/rect_f.h"
 #include "ui/gfx/image/canvas_image_source.h"
 #include "ui/gfx/image/image_skia.h"
@@ -34,16 +33,8 @@ class ContainersIconImageSource : public gfx::CanvasImageSource {
         background_(background),
         dip_icon_size_(dip_icon_size) {}
 
-  ContainersIconImageSource(const gfx::ImageSkia& icon_image, int dip_size)
-      : ContainersIconImageSource(icon_image,
-                                  gfx::kPlaceholderColor,
-                                  dip_size,
-                                  dip_size) {}
-
   void Draw(gfx::Canvas* canvas) override {
-    if (background_ != gfx::kPlaceholderColor) {
-      DrawBackground(canvas);
-    }
+    DrawBackground(canvas);
     DrawIcon(canvas);
   }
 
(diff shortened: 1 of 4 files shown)


browser/ui/containers/containers_icon_generator.cc
```diff
@@ -166,19 +187,24 @@ gfx::ImageSkia GenerateContainerIcon(std::string_view container_id,
...
-                                       dip_icon_size, SK_ColorWHITE);
-  }
+  gfx::ImageSkia icon_image =
+      GetIconImageForContainer(container_id, icon, dip_icon_size, scale_factor);
   auto image_source = std::make_unique<ContainersIconImageSource>(
       std::move(icon_image), background, dip_size, dip_icon_size);
   return gfx::ImageSkia(std::move(image_source), scale_factor);
 }
 
+gfx::ImageSkia GenerateContainerIconWithoutBackground(
```
goodov: `GenerateContainerForegroundIcon` ?

browser/ui/containers/containers_icon_generator.cc
```diff
@@ -33,8 +34,16 @@ class ContainersIconImageSource : public gfx::CanvasImageSource {
         background_(background),
         dip_icon_size_(dip_icon_size) {}
 
+  ContainersIconImageSource(const gfx::ImageSkia& icon_image, int dip_size)
+      : ContainersIconImageSource(icon_image,
+                                  gfx::kPlaceholderColor,
```
goodov: that's an actual color (red). maybe go with `optional<SkColor>` + nullopt?

browser/ui/containers/containers_icon_generator.cc
```diff
@@ -166,19 +187,24 @@ gfx::ImageSkia GenerateContainerIcon(std::string_view container_id,
...
 }
 
+gfx::ImageSkia GenerateContainerIconWithoutBackground(
+    std::string_view container_id,
+    mojom::Icon icon,
+    int dip_size,
+    float scale_factor) {
+  auto image_source = std::make_unique<ContainersIconImageSource>(
+      GetIconImageForContainer(container_id, icon, dip_size, scale_factor),
+      dip_size);
```
goodov: if `dip_size` is equal to `dip_icon_size`, and there's no background, why do we need to use `ContainersIconImageSource`? Why can't we just return the result of `GetIconImageForContainer()`?

browser/ui/containers/containers_icon_generator.cc
sangwoo108: Much better

browser/ui/containers/containers_icon_generator.cc
sangwoo108: You got the right point. We can just return it.

browser/ui/containers/containers_icon_generator.cc
sangwoo108: Ah yes, that'd be more semantic

browser/ui/containers/containers_icon_generator.h
```diff
@@ -37,6 +37,14 @@ gfx::ImageSkia GenerateContainerIcon(std::string_view container_id,
                                      float scale_factor,
                                      const ui::ColorProvider* color_provider);
 
+// Returns a foreground icon for the container. In case the container_id is a
+// temporary container, it will return a temporary container foreground icon.
+// Otherwise, it will return a vector icon for the container.
+gfx::ImageSkia GenerateContainerForegroundIcon(std::string_view container_id,
```
goodov: how about we accept a color here and:
1. call this from `BraveTabStrip::GetTabAccentIcon`? instead of
```
  if (containers::IsTemporaryContainerId(container->id)) {
    return ui::ImageModel::FromImageSkia(
        containers::GenerateTemporaryContainerForegroundIcon(
            container->id, accent_colors->icon_color, 16, scale_factor));
  }

  const auto& icon = containers::GetVectorIconFromIconType(container->icon);
  return ui::ImageModel::FromVectorIcon(icon, accent_colors->icon_color, 16);
```
2. make `GenerateTemporaryContainerForegroundIcon` private?

netzenbot (approved):

browser/ui/containers/containers_icon_generator.h
sangwoo108: Such a great idea!

sangwoo108: Done.

goodov (approved):

PR #36340, opened 2026-05-12
[Wallet] Delay unlock with dev-wallet-password

supermassive:
Fix crash when `dev-wallet-password` cmd line param is used.

### DIFF
--- a/components/brave_wallet/browser/keyring_service_unittest.cc
+++ b/components/brave_wallet/browser/keyring_service_unittest.cc
@@ -121,6 +121,13 @@ std::string GenerateBtcImportPayload(std::string_view private_key_hex) {
   return Base58EncodeWithCheck(buf);
 }
 
+void WaitForPostedTask() {
+  base::RunLoop run_loop;
+  base::SingleThreadTaskRunner::GetCurrentDefault()->PostTask(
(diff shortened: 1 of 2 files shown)


cypt4 (approved):

components/brave_wallet/browser/keyring_service_unittest.cc
```diff
@@ -121,6 +121,13 @@ std::string GenerateBtcImportPayload(std::string_view private_key_hex) {
   return Base58EncodeWithCheck(buf);
 }
 
+void WaitForPostedTask() {
```
bridiver, line 124: This is not reliable and will break if there is  a chain of tasks instead of a single post task. It is better to use `ASSERT_TRUE(base::test::RunUntil([&]() { service.IsLockedSync(); }))` here

supermassive: What am I going to wait for to expect it stays unlocked?

supermassive: And it is reliable as there is no chain of tasks now. Just search for this through tests `FROM_HERE.*QuitClosure\(\)\);`

PR #36341, opened 2026-05-12
[AI Chat] Do not accept duplicate tool use event IDs

### LINKED ISSUE #55438: [AI Chat] infinite loop in `MaybeRespondToNextToolUseRequest`
Seems like there is an occasional infinite loop being logged in the crash reports on all platforms between `MaybeRespondToNextToolUseRequest` and `RespondToToolUseRequest`. 

The guess is that if duplicate tool use requests with the same Tool Use ID are received by the API response, then we only assign any UseTool output to the first matching event. Then the ConversationHandler scans the pending events to see if there are any tools with no output that need execution. If there are, then we will call UseTool on those tools, which will in turn call `ResponseToToolUseRequest`. In the case of duplicate IDs, there will be a tool to execute because we only assign the output to the first matching request by ID, and we would leave the duplicate tool use request with no output.

This can be fixed by preventing adding tool use requests with duplicate IDs in the first place. When receiving a new tool use event from an API response, we should drop it if it 

We should also try to check if there is a certain model / tool combination that causes duplicate IDs, since the model engines should have some protection against this. Instead of crashing in this scenario we should dump the stack along with the tool name (and model name if not a user-custom model).

Marking as `QA/No` because I don't have concrete steps to reproduce. Fixing an identified problem that I believe is the cause of a crash seen logged in the wild.

petemill:
- Fix https://github.com/brave/brave-browser/issues/55438

~Also adds dump data for which tool / model combination are causing this duplication. Only dumps model name if not a custom model. We should figure out which piece of the engine is causing this:~
- https://github.com/brave/brave-browser/issues/55439

### DIFF
--- a/components/ai_chat/core/browser/conversation_handler.cc
+++ b/components/ai_chat/core/browser/conversation_handler.cc
@@ -28,7 +28,6 @@
 #include "base/functional/callback.h"
 #include "base/logging.h"
 #include "base/memory/weak_ptr.h"
-#include "base/notreached.h"
 #include "base/numerics/safe_math.h"
 #include "base/rand_util.h"
 #include "base/strings/strcat.h"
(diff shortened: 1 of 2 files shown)


components/ai_chat/core/browser/conversation_handler.cc
```diff
@@ -1376,6 +1376,39 @@ void ConversationHandler::UpdateOrCreateLastAssistantEntry(
...
+            SCOPED_CRASH_KEY_STRING1024("BraveAIChatToolId", "id",
+                                        tool_use_event->id);
+            if (GetCurrentModel().options->is_leo_model_options()) {
+              SCOPED_CRASH_KEY_STRING1024("BraveAIChatModel", "key",
+                                          model_key_);
+            }
+            DUMP_WILL_BE_NOTREACHED()
+                << "Dropping tool_use event with duplicate id: "
+                << tool_use_event->id << " (tool: " << tool_use_event->tool_name
+                << ")";
```
petemill: Obviously adding this has broken the test as it now hits `NOTREACHED()`, so I need to come up with a solution

goodov:
```
#if defined(OFFICIAL_BUILD) && !DCHECK_IS_ON()
TEST_F(ConversationHandlerUnitTest, ToolUseEvents_DuplicateToolIds) {
```

SergeyZhukovsky (approved):
++

netzenbot (approved):

components/ai_chat/core/browser/conversation_handler.cc
petemill: So as long as we run tests for official builds (and without dcheck), they will at least get run before a release. But I can't help feel that it might be better to instead, in the implementation, wrap `base::debug::DumpWithoutCrashing` with this condition, so that the test can be run in developer and PR environments. The `NOTREACHED` part isn't even that useful here because this is external data we are guarding against, unlikely to happen due to a new code path, and if it does we have a unit test for it.

Happy to merge like this (only running the test on official builds), but wanted to get your feedback in case we can change in a follow-up to run the test in all environments.

goodov: Do we really need to use Dump if it's external data and is generated by us? Can't we just log such things on the server?

petemill: I don't think we have any logging on the server, but I'm putting the question out. For now, I've removed the dump to land this, and can land separately.

PR #36347, opened 2026-05-12
Fix a cause of Omaha 4 update failures on macOS

mherrmann:
Chrome and Edge both have a fix for potential update failures on macOS that we lack. It is hard to reproduce (cf. https://issues.chromium.org/issues/41492879), but I feel we should have the fix.

# Test plan

Please test that updates with Omaha 4 on macOS still work in both per-user (DMG) and system-wide (PKG) installations.


updater/install_sh_mac_test.py
```diff
@@ -0,0 +1,163 @@
+#!/usr/bin/env python3
```
goodov, line 1: let's name this and .cc files to contain "mac" or move them to `mac` subdirectory.

goodov (approved):
lgtm

updater/install_sh_mac_test.py
mherrmann: Implemented in 5d566664df8e1466d2a8fc93cf535ac90719e638.

PR #36354, opened 2026-05-12
[webgl] Add fingerprinting protections against webgl_debug_renderer_info

### LINKED ISSUE #55443: Add fingerprinting protections for UNMASKED_VENDOR and UNMASKED_RENDERER
For the given script UNMASKED_VENDOR_WEBGL and UNMASKED_RENDERER_WEBGL returns: 

- For Vendor: **"Google Inc. (Apple)"** 
- For Renderer **"ANGLE (Apple, ANGLE Metal Renderer: Apple M5 Max, Unspecified Version)"**

**Script**
```
const canvas = document.createElement("canvas"); 
const gl = canvas.getContext("webgl");
const debugInfo = gl.getExtension("WEBGL_debug_renderer_info");
const vendor = gl.getParameter(debugInfo.UNMASKED_VENDOR_WEBGL);
const renderer = gl.getParameter(debugInfo.UNMASKED_RENDERER_WEBGL);
console.log(vendor);
console.log(renderer);
```

They should be updated to return "Brave" in both cases.

r0hit0303:
This change adds fingerprinting protection for UNMASKED_VENDOR_WEBGL and UNMASKED_RENDERER_WEBGL for balanced farbling level behind a new feature flag |kWebGLBalancedFingerprintingProtection|.

Manual testing with |kWebGLBalancedFingerprintingProtection| flag enabled:
https://github.com/user-attachments/assets/3dc32625-5a85-491e-9a05-919e932313dd

Resolves https://github.com/brave/brave-browser/issues/55443


r0hit0303:
Hello @goodov Could you PTAL at this change? Thanks

browser/farbling/brave_webgl_farbling_browsertest.cc
```diff
@@ -16,6 +18,7 @@
 #include "brave/components/constants/brave_paths.h"
 #include "brave/components/constants/pref_names.h"
 #include "brave/components/webcompat/core/common/features.h"
+#include "brave/third_party/blink/renderer/brave_farbling_constants.h"
```
goodov: you shouldn't include this. Pls look inside and use mojo types from the component level instead.

browser/farbling/DEPS
```diff
@@ -9,6 +9,9 @@ specific_include_rules = {
     "+content/public/renderer/render_frame.h",
     "+third_party/blink/public",
   ],
+  "brave_webgl_farbling_browsertest\.cc": [
+    "+brave/third_party/blink/renderer/brave_farbling_constants.h",
```
goodov: del pls

chromium_src/third_party/blink/renderer/modules/webgl/webgl_rendering_context_base.cc
```diff
@@ -3,25 +3,70 @@
...
+      return original_script_value;
+    case BraveFarblingLevel::BALANCED:
+      return base::FeatureList::IsEnabled(
+                 blink::features::kWebGLBalancedFingerprintingProtection)
+                 ? blink::WebGLAny(script_state, blink::String("Brave"))
+                 : original_script_value;
+    case BraveFarblingLevel::MAXIMUM:
+      return blink::WebGLAny(
+          script_state,
+          blink::String(
```
goodov, line 58: why not `String`? I haven't seen any recommendation to follow `blink::String` in new blink code, so I'd prefer to keep `String`.

browser/farbling/brave_webgl_farbling_browsertest.cc
```diff
@@ -247,3 +217,89 @@ IN_PROC_BROWSER_TEST_F(BraveWebGLFarblingBrowserTest, GetAttachedShaders) {
...
+  std::string GetExpectedString(
+      BraveFarblingLevel level,
+      std::optional<std::string> expected_override = std::nullopt) {
+    if (expected_override.has_value()) {
+      return expected_override.value();
+    }
+    if (level == BraveFarblingLevel::MAXIMUM) {
+      return "uAfPPuXL,aseXyZzC";
+    } else if (level == BraveFarblingLevel::BALANCED) {
+      return GetParam() ? "Brave,Brave" : expected_override.value();
```
netzenbot: Bug: the `else if (level == BraveFarblingLevel::BALANCED)` branch with `!GetParam()` calls `expected_override.value()` after the `has_value()` guard already confirmed it's `std::nullopt` — this throws `std::bad_optional_access` at runtime. Same issue in the `else` branch on the next line. You likely want to return the actual live hardware value here; consider restructuring so the caller always provides a fallback, or use `expected_override.value_or(...)` instead.

browser/farbling/brave_webgl_farbling_browsertest.cc
```diff
@@ -247,3 +217,89 @@ IN_PROC_BROWSER_TEST_F(BraveWebGLFarblingBrowserTest, GetAttachedShaders) {
...
+          {}, {blink::features::kWebGLBalancedFingerprintingProtection});
+    }
+  }
+
+  std::string GetExpectedString(
+      BraveFarblingLevel level,
+      std::optional<std::string> expected_override = std::nullopt) {
+    if (expected_override.has_value()) {
+      return expected_override.value();
+    }
```
netzenbot, line 247: Bug: `expected_override.value()` is called on branches that are only reachable when `expected_override` is `std::nullopt` — the early `if (expected_override.has_value()) { return ...; }` exit means `.value()` here always throws `std::bad_optional_access`. The `else` branch is dead code and both unsafe callsites should be replaced with a concrete return value or the function logic restructured.

browser/farbling/brave_webgl_farbling_browsertest.cc
```diff
@@ -247,3 +217,89 @@ IN_PROC_BROWSER_TEST_F(BraveWebGLFarblingBrowserTest, GetAttachedShaders) {
...
+    if (level == BraveFarblingLevel::MAXIMUM) {
```
netzenbot: `GetExpectedString` returns early on lines 234-236 only when `expected_override.has_value()` is true, so after that early return the value is guaranteed to be `std::nullopt`. The subsequent calls to `expected_override.value()` on the BALANCED path (`GetParam() ? "Brave,Brave" : expected_override.value()`) and in the `else` branch will throw `std::bad_optional_access` at runtime. These branches need to handle the nullopt case explicitly.

browser/farbling/brave_webgl_farbling_browsertest.cc
```diff
@@ -247,3 +217,89 @@ IN_PROC_BROWSER_TEST_F(BraveWebGLFarblingBrowserTest, GetAttachedShaders) {
...
+          {blink::features::kWebGLBalancedFingerprintingProtection}, {});
+    } else {
+      get_parameter_feature_list_.InitWithFeatures(
+          {}, {blink::features::kWebGLBalancedFingerprintingProtection});
+    }
+  }
+
+  std::string GetExpectedString(
+      BraveFarblingLevel level,
+      std::optional<std::string> expected_override = std::nullopt) {
```
netzenbot, line 244: Bug: after the `if (expected_override.has_value()) return ...` guard, `expected_override` is guaranteed to be `std::nullopt` in all remaining branches. Calling `expected_override.value()` in the `BALANCED` (when `!GetParam()`) and `else` branches will throw `std::bad_optional_access` (or UB with exceptions disabled). Those branches should not be reachable with a valid override, so either add a `NOTREACHED()` / `CHECK` there, or restructure the logic so callers always pass an override when they need a non-hardcoded result.

browser/farbling/brave_webgl_farbling_browsertest.cc
```diff
@@ -247,3 +217,89 @@ IN_PROC_BROWSER_TEST_F(BraveWebGLFarblingBrowserTest, GetAttachedShaders) {
...
+      get_parameter_feature_list_.InitWithFeatures(
+          {blink::features::kWebGLBalancedFingerprintingProtection}, {});
+    } else {
+      get_parameter_feature_list_.InitWithFeatures(
+          {}, {blink::features::kWebGLBalancedFingerprintingProtection});
+    }
+  }
+
+  std::string GetExpectedString(
+      BraveFarblingLevel level,
```
netzenbot: Bug in `GetExpectedString`: the first `if (expected_override.has_value())` returns early, so when execution reaches `return GetParam() ? "Brave,Brave" : expected_override.value();` or the final `else { return expected_override.value(); }`, `expected_override` is guaranteed to be `nullopt`. Calling `.value()` on a `nullopt` optional is UB (throws `std::bad_optional_access` in non-NDEBUG builds). The BALANCED/false and default branches need a real fallback value instead of dereferencing the empty optional.

browser/farbling/brave_webgl_farbling_browsertest.cc
r0hit0303: Thanks for the suggestion. I've removed the dependency in the latest ps.

browser/farbling/DEPS
r0hit0303: Done

chromium_src/third_party/blink/renderer/modules/webgl/webgl_rendering_context_base.cc
r0hit0303: It's because the context here is an anonymous namespace and therefore required the blink scope specifier.

browser/farbling/brave_webgl_farbling_browsertest.cc
r0hit0303: This and in other comments the crash was intentional. I've made it more explicit in the latest change.

r0hit0303:
Thanks Aleksei for your thorough review

chromium_src/third_party/blink/renderer/modules/webgl/webgl_rendering_context_base.cc
goodov: hmm. yeah, I see now. I remember `blink::String` was re-exported as `String` into global namespace, but looks like that's no longer the case.

https://source.chromium.org/chromium/chromium/src/+/36c632009fcd90457e60c49da342712d99a341a8:third_party/blink/renderer/platform/wtf/text/wtf_string.h;dlc=7fc55ddc42aba9d2d672391053b23cd60ecbe916

goodov (approved):

r0hit0303:
Thanks Aleksei for the approval.

PR #36357, opened 2026-05-12
[Wallet]: Fix Token 2022 Insufficient Funds Error During Transfer

### LINKED ISSUE #55454: Unable to transfer SPL Token-2022 tokens
### Description

When you attempt to transfer a SPL `Token-2022` token, you get an insufficient funds error on the `Confirmation Panel`

### Steps to reproduce

1. Attempt to send an SPL `Token-2022` token such as `$CWIF`
2. You will get an insufficient funds error in the `Confirmation` panel

### Actual result

Insufficient funds error, so you are unable to confirm

### Expected result

Should not bet an `insufficient funds` error

### Reproduces how often

Easily reproduced

1.89.143 Chromium: 147.0.7727.117 (Official Build) (arm64) 

### Channel information

- [x] release (stable)
- [x] beta
- [x] nightly

### Reproducibility

### Miscellaneous information

_No response_

Douglashdaniel:
`GetSPLTokenProgramByMint` in json_rpc_service.cc now always resolves the mint’s program from the chain via `GetSolanaAccountInfo` (mint account owner = token program id). Prefs are still updated in `ContinueGetSPLTokenProgramByMint` when the mint is a user asset, but we no longer trust a possibly wrong cached program for balance / ATA derivation.

Resolves <https://github.com/brave/brave-browser/issues/55454>

## Test plan:
1. Attempt to send an SPL `Token-2022` token such as `$CWIF`
2. You should not get a insufficient funds error in the `Confirmation` panel

Screenshot 62

### DIFF
--- a/components/brave_wallet/browser/json_rpc_service.cc
+++ b/components/brave_wallet/browser/json_rpc_service.cc
@@ -3572,25 +3572,13 @@ void JsonRpcService::GetSPLTokenProgramByMint(
     return;
   }
 
-  mojom::BlockchainTokenPtr user_asset;
-  if ((user_asset = GetUserAsset(prefs_, mojom::CoinType::SOL, chain_id,
-                                 mint_address, "", false, false, false))) {
-    if (user_asset->spl_token_program != mojom::SPLTokenProgram::kUnknown) {
-      std::move(callback).Run(user_asset->spl_token_program,
-                              mojom::SolanaProviderError::kSuccess, "");
-      return;
-    }
-  } else if (auto token = BlockchainRegistry::GetInstance()->GetTokenByAddress(
-                 chain_id, mojom::CoinType::SOL, mint_address)) {
-    // In theory token in registry won't have unknown value appears, but we let
-    // it fall through to fetch from network as it won't hurt if that does
-    // happen somehow.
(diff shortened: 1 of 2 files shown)


components/brave_wallet/browser/json_rpc_service.cc
onyb, line 3613: **Suggested follow up:** remove this block - no point setting the token program if it is no longer read. Subsequently, the `mojom::BlockchainTokenPtr user_asset` argument from `JsonRpcService::ContinueGetSPLTokenProgramByMint` can also be removed.

onyb (approved):

netzenbot (approved):

cmazakas (approved):
I'm approving this because it gets a fix out the door.

That all being said, it's worth noting that nothing in prefs should ever go stale like this, and that we should only ever be committing something finalized or derived purely from on-chain data.

Skipping the BlockchainRegistry here is probably indisputably correct.

But I think in the future, we should aim towards having AddUserAsset get the token program for an asset instead of us marking it as kUnknown or using the BlockchainRegistry list.

That or we don't store this field at all and always fetch it from on-chain data.

@supermassive any thoughts on this?

PR #36365, opened 2026-05-12
Fix build broken due

cypt4:
https://github.com/brave/brave-core/pull/36340/changes since method is used under !is_official_build condition.

### DIFF
--- a/components/brave_wallet/browser/keyring_service_unittest.cc
+++ b/components/brave_wallet/browser/keyring_service_unittest.cc
@@ -121,12 +121,14 @@ std::string GenerateBtcImportPayload(std::string_view private_key_hex) {
   return Base58EncodeWithCheck(buf);
 }
 
+#if !defined(OFFICIAL_BUILD)
 void WaitForPostedTask() {
   base::RunLoop run_loop;
   base::SingleThreadTaskRunner::GetCurrentDefault()->PostTask(
       FROM_HERE, run_loop.QuitClosure());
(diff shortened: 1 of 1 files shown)


components/brave_wallet/browser/keyring_service_unittest.cc
```diff
@@ -121,12 +121,14 @@ std::string GenerateBtcImportPayload(std::string_view private_key_hex) {
   return Base58EncodeWithCheck(buf);
 }
 
+#if !defined(OFFICIAL_BUILD)
```
bridiver, line 124: Why would this be gated on official build? What about the behavior changes when it's an official build or not? Also see correct way to wait here https://github.com/brave/brave-core/pull/36340/changes#r3228532961

cypt4: This test was marked as !official_build 4 years ago https://github.com/brave/brave-core/commit/60d4fc513e686ed14e62dc4f15a635f9bbed5d0a 
cc @supermassive

bridiver: I see, it was hidden in the diff for this PR and missied it in the one that caused the problem. There should be a follow up though to get rid of this method entirely and call `ASSERT_TRUE(base::test::RunUntil([&]() { service.IsLockedSync(); }))` directly instead

netzenbot (approved):

PR #36368, opened 2026-05-12
Remove re-entrant call to lazy initialize adblock service

bridiver:
This fixes a problem that was surfaced by https://github.com/brave/brave-core/pull/36263, but technically it's a pre-existing issue that was just less likely to happen before it.


antonok-edm (approved):

netzenbot (approved):

cdesouza-chromium (approved):
LGTM

PR #36371, opened 2026-05-12
[iOS] Add support for randomized message handlers in JavaScriptFeatures

kylehickinson:
This change adds a helper class `RandomizedMessageHandlerName` that can be used in JavaScriptFeature's to support randomizing message handler names. The randomized message handler is a UUIDv4 which is injected with the script itself via placeholder replacement.

### DIFF
--- a/ios/web/js_messaging/BUILD.gn
+++ b/ios/web/js_messaging/BUILD.gn
@@ -14,6 +14,7 @@ source_set("js_messaging") {
   ]
   deps = [
     ":message_handler_token",
+    ":randomized_message_handler_name",
     ":safe_builtins_js",
   ]
   public_deps = [ "//ios/web/public/js_messaging" ]
@@ -30,15 +31,30 @@ source_set("message_handler_token") {
   ]
 }
 
+source_set("randomized_message_handler_name") {
+  sources = [
+    "randomized_message_handler_name.h",
+    "randomized_message_handler_name.mm",
+  ]
+  public_deps = [
+    "//base",
+    "//ios/web/public/js_messaging",
+  ]
+}
+
 source_set("unittests") {
   testonly = true
   sources = [
     "message_handler_token_inttest.mm",
+    "randomized_message_handler_name_inttest.mm",
     "safe_builtins_javascript_test.mm",
   ]
   deps = [
+    ":js_messaging",
     ":message_handler_token",
     ":message_handler_token_test_api",
+    ":randomized_message_handler_name",
+    ":randomized_message_handler_test_api",
     ":safe_builtins_js",
     "//base",
     "//base/test:test_support",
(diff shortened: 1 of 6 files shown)


ios/web/js_messaging/with_randomized_message_handler.h
```diff
@@ -0,0 +1,38 @@
...
+  std::string GetRandomizedMessageHandlerName() const;
+  PlaceholderReplacements GetPlaceholderReplacements() const;
+
+ private:
+  base::Uuid handler_name_ = base::Uuid::GenerateRandomV4();
+};
+
+}  // namespace web
+
+#endif  // BRAVE_IOS_WEB_JS_MESSAGING_WITH_RANDOMIZED_MESSAGE_HANDLER_H_
```
netzenbot: nit: `WithRandomizedMessageHandler` is a concrete class with state (`handler_name_`) designed to be inherited alongside `JavaScriptFeature`. [CS-069](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-069) discourages multiple inheritance beyond interface-style patterns — consider making it a composable helper that `JavaScriptFeature` subclasses hold as a member and delegate to instead.

ios/web/js_messaging/with_randomized_message_handler.h
```diff
@@ -0,0 +1,38 @@
...
+  WithRandomizedMessageHandler();
+
+  using PlaceholderReplacements =
+      web::JavaScriptFeature::FeatureScript::PlaceholderReplacements;
+
+  std::string GetRandomizedMessageHandlerName() const;
+  PlaceholderReplacements GetPlaceholderReplacements() const;
+
+ private:
+  base::Uuid handler_name_ = base::Uuid::GenerateRandomV4();
```
netzenbot: nit: These protected methods are part of the mixin's API surface for subclasses but have no documentation. Consider adding brief comments describing what each returns — e.g., that GetRandomizedMessageHandlerName() returns a stable per-instance UUID string, and that GetPlaceholderReplacements() returns the map needed by FeatureScript for substituting the placeholder name in JS. See [DOC-004](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-004).

netzenbot (approved):

StephenHeaps (approved):

stoletheminerals (approved):
LGTM.
One note though, from what I remember the previous implementation had hyphens removed from UUID because you can't access these message handlers by `webkit.messageHandlers.name-with-UUID.sendMessage()` as hyphnes make the syntax incorrect. It's still possible to do `webkit.messageHandlers['name-with-UUID'].sendMessage()`, so as long as we stick to this it shouldn't be a problem

kylehickinson:
@stoletheminerals Yep thats correct, we will always use either `sendWebKitMessage{WithReply}` or `gSafeBuiltins.sendWebKitMessage{WithReply}` which always access via square-bracket lookup so all good there.

PR #36384, opened 2026-05-12
[Polkadot] [Wallet] default Polkadot URLs to public RPCs

cmazakas:
Update Polkadot code to use hard-coded well-trusted RPC URLs instead of relying on proxying.

The tests also proved to be over-constrained because they weren't testing anything useful and only hindered further refactors of the code so several testing statements have been removed as they're no longer required.

### DIFF
--- a/components/brave_wallet/browser/polkadot/polkadot_test_utils.cc
+++ b/components/brave_wallet/browser/polkadot/polkadot_test_utils.cc
@@ -291,9 +291,6 @@ void PolkadotMockRpc::FinalizeSetup() {
           ->rpc_endpoints.front()
           .spec();
 
-  EXPECT_EQ(testnet_url_, "https://polkadot-westend.wallet.brave.com/");
-  EXPECT_EQ(mainnet_url_, "https://polkadot-mainnet.wallet.brave.com/");
-
   url_loader_factory_->SetInterceptor(base::BindRepeating(
       &PolkadotMockRpc::RequestInterceptor, base::Unretained(this)));
 }
--- a/components/brave_wallet/browser/network_manager.cc
+++ b/components/brave_wallet/browser/network_manager.cc
@@ -592,7 +592,7 @@ GURL PolkadotMainnetRpcUrl() {
   if (switch_url.is_valid()) {
     return switch_url;
   }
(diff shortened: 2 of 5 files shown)


components/brave_wallet/browser/polkadot/polkadot_test_utils.cc
```diff
@@ -291,9 +291,6 @@ void PolkadotMockRpc::FinalizeSetup() {
           ->rpc_endpoints.front()
```
netzenbot, line 291: These `EXPECT_EQ(testnet_url_, ...)` / `EXPECT_EQ(mainnet_url_, ...)` assertions were verifying which URL the test infrastructure was routing mock responses through. Rather than deleting them, they should be updated to the new public RPC URLs (`"https://westend-rpc.polkadot.io/"` and `"https://rpc.polkadot.io/"`). The same pattern was correctly followed in `network_manager_unittest.cc` which updated its assertions — the other test files should do the same. See [TI-017](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-017).

components/brave_wallet/browser/polkadot/polkadot_test_utils.cc
```diff
@@ -291,9 +291,6 @@ void PolkadotMockRpc::FinalizeSetup() {
...
           .spec();
 
-  EXPECT_EQ(testnet_url_, "https://polkadot-westend.wallet.brave.com/");
-  EXPECT_EQ(mainnet_url_, "https://polkadot-mainnet.wallet.brave.com/");
-
   url_loader_factory_->SetInterceptor(base::BindRepeating(
```
netzenbot, line 294: nit: The `EXPECT_EQ(testnet_url_, ...)` and `EXPECT_EQ(mainnet_url_, ...)` assertions are being deleted instead of updated to the new values (`"https://westend-rpc.polkadot.io/"` and `"https://rpc.polkadot.io/"`). The `network_manager_unittest.cc` change does cover the new URL, but these were additional sanity checks that explicitly tied the mock RPC setup to known endpoint values — updating them rather than dropping them would keep that coverage intact. ([PRs Should Include Reasonable Test Coverage](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-039))

PR #36385, opened 2026-05-12
[AI Chat] Add semantic history search tool for Leo

### LINKED ISSUE #55102: Export history semantic search to Leo
By tool use and behind the same flag brave://flags/#brave-history-embeddings

darkdh:
Resolves https://github.com/brave/brave-browser/issues/55102
Security/Privacy review https://github.com/brave/reviews/issues/2253

- Adds `HistorySearchTool`, a self-executing Tool that wraps
  `HistoryEmbeddingsService::Search()` so Leo can semantically search the
  user's local browsing history. Matches are returned as a JSON
  `TextContentBlock` (title, URL, last visit time, and either all indexed
  passages or a single best-matching snippet). Gated on
  `IsHistoryEmbeddingsEnabledForProfile`; the search itself runs entirely
  on-device.
- **Permission gate.** Before any results flow to the LLM, the tool
  returns a `PermissionChallenge` asking the user to confirm. The search
  query and full history never leave the device — only the matching
  pages' titles, URLs, and indexed text passages are sent to Brave AI as
  tool output. Permission is granted once per conversation.
- Wires the tool into `BrowserToolProvider` when the feature is enabled
  for the active profile. Exposes the tool name as
  `mojom::kSemanticHistorySearchToolName` and a UI label
  ("Searching your browser history for \"{query}\"") via
  `get_tool_label.ts`.
- Introduces a new `kVisitedLinksArtifactType` tool artifact carrying a
  JSON array of HTTPS URLs. `getGroupVisitedLinks` flattens these across
  an assistant entry group and contributes them to `allowedLinks` from
  both render paths (`AssistantResponse` directly, and `AssistantTask`
  via `useExtractTaskData`) so anchors referencing tool-returned URLs
  survive the `MarkdownRenderer` href filter. Web-search `sourcesEvent`
  links stay per-entry; only visited_links cross the entry boundary
  (required because the tool-use protocol splits a single logical
  response across two assistant entries, and `TASK_TOOL_COUNT=1` routes
  any tool use through the task path). No visual rendering, no mojom
  struct change, no Sources panel.
- `HistorySearchTool` emits this artifact populated from URLs in its
  results.

## Test plan

- Enable `brave://flags/#brave-history-embeddings`, visit several pages
  with distinct content, wait for embeddings to index, then ask Leo a
  question about page content (e.g. "what was that article I read about
  X"). Confirm the permission prompt appears; approve it. Verify the
  tool fires, results render, and "give me the history link" produces a
  clickable anchor.
- If you are using version later than [v1.93.50](https://github.com/brave/brave-browser/releases/tag/v1.93.50)
   you will have to go to the brave://history and enable the toggle
- Deny the permission prompt and confirm Leo gracefully falls back
  without searching.
- Submit another history query in the same conversation and confirm the
  permission prompt does NOT re-appear (granted per conversation).
- With the flag disabled (or on a profile where embeddings aren't
  eligible), confirm the tool is not advertised and Leo falls back to
  its normal behavior.

https://github.com/user-attachments/assets/ba2bd634-8356-4543-9481-276908e0e2f5

### DIFF
--- a/browser/ai_chat/BUILD.gn
+++ b/browser/ai_chat/BUILD.gn
@@ -405,6 +405,7 @@ source_set("browser_tests") {
       "code_execution_tool_browsertest.cc",
       "page_content_fetcher_browsertest.cc",
       "text_file_extractor_browsertest.cc",
+      "tools/history_search_tool_browsertest.cc",
     ]
 
     if (enable_pdf) {
@@ -438,6 +439,8 @@ source_set("browser_tests") {
       "//chrome/browser/ui/side_panel:side_panel_views_dependent",
       "//chrome/test:test_support",
       "//chrome/test:test_support_ui",
+      "//components/history/core/browser",
+      "//components/history_embeddings/core",
       "//printing/buildflags",
       "//services/screen_ai/buildflags",
       "//services/screen_ai/public/cpp:utilities",
--- a/browser/ai_chat/BUILD.gn
+++ b/browser/ai_chat/BUILD.gn
@@ -300,6 +300,7 @@ source_set("unit_tests") {
     "ai_chat_throttle_unittest.cc",
     "brave_open_ai_chat_permission_context_unittest.cc",
     "full_screenshotter_unittest.cc",
+    "tools/history_search_tool_unittest.cc",
   ]
 
   deps = [
@@ -312,11 +313,16 @@ source_set("unit_tests") {
     "//brave/components/constants",
     "//chrome/common",
     "//chrome/test:test_support",
+    "//components/history/core/browser",
+    "//components/history_embeddings/content",
+    "//components/history_embeddings/core",
     "//components/paint_preview/common:test_utils",
     "//components/paint_preview/common/mojom",
+    "//components/passage_embeddings/core",
     "//components/services/paint_preview_compositor/public/mojom",
     "//content/public/browser",
     "//content/test:test_support",
+    "//testing/gmock",
     "//testing/gtest",
     "//url",
   ]
--- a/browser/ai_chat/BUILD.gn
+++ b/browser/ai_chat/BUILD.gn
@@ -35,13 +35,20 @@ static_library("ai_chat") {
     "tab_tracker_service_factory.h",
     "tools/code_execution_tool.cc",
     "tools/code_execution_tool.h",
+    "tools/history_search_tool.cc",
+    "tools/history_search_tool.h",
   ]
 
   if (enable_ai_chat_tab_management_tool) {
     sources += [ "tools/tab_management_tool.h" ]
   }
 
-  public_deps = [ "//brave/components/restricted_web_contents_delegate" ]
+  public_deps = [
+    "//brave/components/restricted_web_contents_delegate",
+
+    # history_search_tool.h includes history_embeddings_search.h.
+    "//components/history_embeddings/core",
+  ]
 
   deps = [
     "//brave/brave_domains",
@@ -61,11 +68,14 @@ static_library("ai_chat") {
     "//brave/net/base:utils",
     "//chrome/browser:browser_process",
     "//chrome/browser/actor",
+    "//chrome/browser/history_embeddings",
     "//chrome/browser/profiles:profile",
     "//chrome/browser/ui/tabs:tabs_public",
     "//chrome/common",
     "//chrome/common:channel_info",
     "//components/browsing_data/core",
+    "//components/history/core/browser",
+    "//components/history_embeddings/content",
     "//components/keyed_service/content",
     "//components/prefs",
     "//components/user_prefs",
--- a/browser/ai_chat/browser_tool_provider.h
+++ b/browser/ai_chat/browser_tool_provider.h
@@ -23,6 +23,7 @@ class BrowserContext;
 namespace ai_chat {
 
 class CodeExecutionTool;
+class HistorySearchTool;
 class TabManagementTool;
 
 // Implementation of ToolProvider that provides browser-specific
@@ -41,11 +42,14 @@ class BrowserToolProvider : public ToolProvider {
   // ToolProvider implementation
   std::vector<base::WeakPtr<Tool>> GetTools() override;
 
+  HistorySearchTool* GetHistorySearchToolForTesting();
+
  private:
   void CreateTools(content::BrowserContext* browser_context);
 
   // Browser-specific tools owned by this provider
   std::unique_ptr<CodeExecutionTool> code_execution_tool_;
+  std::unique_ptr<HistorySearchTool> history_search_tool_;
 #if BUILDFLAG(ENABLE_AI_CHAT_TAB_MANAGEMENT_TOOL)
   std::unique_ptr<TabManagementTool> tab_management_tool_;
 #endif
--- a/browser/ai_chat/tools/history_search_tool.cc
+++ b/browser/ai_chat/tools/history_search_tool.cc
@@ -221,7 +221,9 @@ void HistorySearchTool::UseTool(const std::string& input_json,
 
   // Defense in depth: the tool is only published when the flag is on, but
   // re-check here in case the flag flipped between advertisement and use.
-  if (!history_embeddings::IsHistoryEmbeddingsEnabledForProfile(profile_)) {
+  // Tests that inject a fake search interface bypass this gate.
+  if (!search_for_testing_ &&
+      !history_embeddings::IsHistoryEmbeddingsEnabledForProfile(profile_)) {
     std::move(callback).Run(
         CreateContentBlocksForText(
             "Error: history embeddings is not enabled for this profile"),
@@ -229,7 +231,10 @@ void HistorySearchTool::UseTool(const std::string& input_json,
     return;
   }
 
-  auto* service = HistoryEmbeddingsServiceFactory::GetForProfile(profile_);
+  history_embeddings::HistoryEmbeddingsSearch* service =
+      search_for_testing_
+          ? search_for_testing_.get()
+          : HistoryEmbeddingsServiceFactory::GetForProfile(profile_);
   if (!service) {
     std::move(callback).Run(
         CreateContentBlocksForText(
@@ -249,6 +254,11 @@ void HistorySearchTool::UseTool(const std::string& input_json,
           &HistorySearchTool::OnSearchResult, weak_ptr_factory_.GetWeakPtr(),
           base::Owned(std::make_unique<UseToolCallback>(std::move(callback))),
           include_all_passages));
+}
+
+void HistorySearchTool::SetSearchForTesting(
+    history_embeddings::HistoryEmbeddingsSearch* search) {
+  search_for_testing_ = search;
 }
 
 void HistorySearchTool::OnSearchResult(
--- a/browser/ai_chat/tools/history_search_tool.cc
+++ b/browser/ai_chat/tools/history_search_tool.cc
@@ -19,12 +19,14 @@
 #include "base/values.h"
 #include "brave/components/ai_chat/core/browser/tools/tool_input_properties.h"
 #include "brave/components/ai_chat/core/browser/tools/tool_utils.h"
+#include "brave/components/ai_chat/core/common/mojom/ai_chat.mojom.h"
 #include "chrome/browser/history_embeddings/history_embeddings_service_factory.h"
 #include "chrome/browser/history_embeddings/history_embeddings_utils.h"
 #include "chrome/browser/profiles/profile.h"
 #include "components/history/core/browser/url_row.h"
 #include "components/history_embeddings/content/history_embeddings_service.h"
 #include "components/history_embeddings/core/history_embeddings_features.h"
+#include "url/url_constants.h"
 
 namespace ai_chat {
 
@@ -258,7 +260,30 @@ void HistorySearchTool::OnSearchResult(
   }
   std::string json = internal::BuildHistorySearchResultJson(
       result.query, result, include_all_passages);
-  std::move(*callback).Run(CreateContentBlocksForText(json), {});
+
+  // Surface the URLs in our results as a visited-links artifact so the
+  // conversation UI permits anchors in the assistant's reply for them.
+  // The artifact has no visual rendering -- it's a sidechannel trust-list.
+  ToolArtifacts artifacts;
+  if (!result.scored_url_rows.empty()) {
+    base::ListValue links;
+    for (const auto& row : result.scored_url_rows) {
+      if (row.row.url().is_valid() &&
+          row.row.url().SchemeIs(url::kHttpsScheme)) {
+        links.Append(row.row.url().spec());
+      }
+    }
+    if (!links.empty()) {
+      std::string links_json;
+      base::JSONWriter::Write(links, &links_json);
+      artifacts.push_back(mojom::ToolArtifact::New(
+          /*id=*/std::nullopt, mojom::kVisitedLinksArtifactType,
+          std::move(links_json)));
+    }
+  }
+
+  std::move(*callback).Run(CreateContentBlocksForText(json),
+                           std::move(artifacts));
 }
 
 }  // namespace ai_chat
--- a/browser/ai_chat/tools/history_search_tool.cc
+++ b/browser/ai_chat/tools/history_search_tool.cc
@@ -1,0 +1,264 @@
+// Copyright (c) 2026 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#include "brave/browser/ai_chat/tools/history_search_tool.h"
+
+#include <algorithm>
+#include <utility>
+
+#include "base/functional/bind.h"
+#include "base/i18n/time_formatting.h"
+#include "base/json/json_reader.h"
+#include "base/json/json_writer.h"
+#include "base/strings/strcat.h"
+#include "base/strings/string_number_conversions.h"
+#include "base/strings/utf_string_conversions.h"
+#include "base/time/time.h"
+#include "base/values.h"
+#include "brave/components/ai_chat/core/browser/tools/tool_input_properties.h"
+#include "brave/components/ai_chat/core/browser/tools/tool_utils.h"
+#include "chrome/browser/history_embeddings/history_embeddings_service_factory.h"
+#include "chrome/browser/history_embeddings/history_embeddings_utils.h"
+#include "chrome/browser/profiles/profile.h"
+#include "components/history/core/browser/url_row.h"
+#include "components/history_embeddings/content/history_embeddings_service.h"
+#include "components/history_embeddings/core/history_embeddings_features.h"
+
+namespace ai_chat {
+
+namespace {
+
+constexpr char kPropertyQuery[] = "query";
+constexpr char kPropertyCount[] = "count";
+constexpr char kPropertyTimeRangeStartDaysAgo[] = "time_range_start_days_ago";
+constexpr char kPropertyIncludeAllPassages[] = "include_all_passages";
+
+constexpr char kOutputKeyQuery[] = "query";
+constexpr char kOutputKeyResults[] = "results";
+constexpr char kOutputKeyTitle[] = "title";
+constexpr char kOutputKeyUrl[] = "url";
+constexpr char kOutputKeyLastVisitTime[] = "last_visit_time";
+constexpr char kOutputKeyPassages[] = "passages";
+constexpr char kOutputKeySnippet[] = "snippet";
+
+constexpr size_t kDefaultResultCount = 5;
+constexpr size_t kMaxResultCount = 10;
+
+base::DictValue ScoredUrlRowToDict(const history_embeddings::ScoredUrlRow& row,
+                                   bool include_all_passages) {
+  base::DictValue entry;
+  entry.Set(kOutputKeyTitle, base::UTF16ToUTF8(row.row.title()));
+  entry.Set(kOutputKeyUrl, row.row.url().spec());
+  // GetBestPassage() / GetBestScoreIndices() CHECK that there is at least
+  // one passage; some rows (e.g. URL-only history entries with no embedded
+  // passages) won't have any, so guard against it.
+  const int passages_size = row.url_data.passages.passages_size();
+  if (passages_size > 0) {
+    if (include_all_passages) {
+      base::ListValue passages;
+      // Returns indices ordered by descending score.
+      auto indices = row.GetBestScoreIndices(/*min_count=*/passages_size,
+                                             /*min_word_count=*/0);
+      for (size_t i : indices) {
+        passages.Append(row.url_data.passages.passages(i));
+      }
+      entry.Set(kOutputKeyPassages, std::move(passages));
+    } else {
+      entry.Set(kOutputKeySnippet, row.GetBestPassage());
+    }
+  }
+  if (!row.row.last_visit().is_null()) {
+    entry.Set(kOutputKeyLastVisitTime,
+              base::TimeFormatAsIso8601(row.row.last_visit()));
+  }
+  return entry;
+}
+
+}  // namespace
+
+namespace internal {
+
+std::string BuildHistorySearchResultJson(
+    const std::string& query,
+    const history_embeddings::SearchResult& result,
+    bool include_all_passages) {
+  base::ListValue results;
+  for (const auto& row : result.scored_url_rows) {
+    results.Append(ScoredUrlRowToDict(row, include_all_passages));
+  }
+
+  base::DictValue root;
+  root.Set(kOutputKeyQuery, query);
+  root.Set(kOutputKeyResults, std::move(results));
+
+  std::string json;
+  base::JSONWriter::Write(root, &json);
+  return json;
+}
+
+}  // namespace internal
+
+HistorySearchTool::HistorySearchTool(content::BrowserContext* browser_context)
+    : profile_(Profile::FromBrowserContext(browser_context)) {}
+
+HistorySearchTool::~HistorySearchTool() = default;
+
+std::string_view HistorySearchTool::Name() const {
+  return mojom::kSemanticHistorySearchToolName;
+}
+
+std::string_view HistorySearchTool::Description() const {
+  return "Performs a semantic (meaning-based) search over the user's local "
+         "browsing history. Use when the user asks about something they "
+         "previously read but can't recall the URL or title, or when you "
+         "need the page's text content to answer their question. Returns "
+         "matching pages as JSON with title, URL, last visit time, and "
+         "(by default) the indexed text passages -- the only way to "
+         "access page content from history; you do not need to ask the "
+         "user to fetch the page. Do not use for exact URL/title lookup "
+         "or general web search. Runs entirely on-device.";
+}
+
+std::optional<base::DictValue> HistorySearchTool::InputProperties() const {
+  return CreateInputProperties(
+      {{kPropertyQuery,
+        StringProperty("Natural language description of the page the user "
+                       "is trying to find in their browsing history.")},
+       {kPropertyCount,
+        IntegerProperty(
+            "Maximum number of results to return. Defaults to 5; capped "
+            "at 10.")},
+       {kPropertyTimeRangeStartDaysAgo,
+        IntegerProperty(
+            "If set, restrict the search to pages visited within the last "
+            "N days. Omit to search the entire history.")},
+       {kPropertyIncludeAllPassages,
+        BooleanProperty(
+            "Include all indexed passages per result, sorted by "
+            "relevance. Defaults to true. Set false to return a single "
+            "'snippet' (best-matching passage only) when no page "
+            "content is needed.")}});
+}
+
+std::optional<std::vector<std::string>> HistorySearchTool::RequiredProperties()
+    const {
+  return std::vector<std::string>{kPropertyQuery};
+}
+
+std::variant<bool, mojom::PermissionChallengePtr>
+HistorySearchTool::RequiresUserInteractionBeforeHandling(
+    const mojom::ToolUseEvent& tool_use) const {
+  if (user_has_granted_permission_) {
+    return false;
+  }
+  // The search itself runs entirely on-device against the local embeddings
+  // index, so the query and the full history never leave the device. Only
+  // the matching pages' titles, URLs, and indexed passages are sent to the
+  // remote LLM as the tool's output. Ask the user to confirm before doing
+  // that. The user-facing wording is in
+  // `get_tool_permission_implications.tsx` so it goes through i18n; this
+  // C++ side only needs to surface a non-null challenge.
+  return mojom::PermissionChallenge::New(/*assessment=*/std::nullopt,
+                                         /*plan=*/std::nullopt);
+}
+
+void HistorySearchTool::UserPermissionGranted(const std::string& tool_use_id) {
+  user_has_granted_permission_ = true;
+}
+
+void HistorySearchTool::UseTool(const std::string& input_json,
+                                UseToolCallback callback) {
+  auto input = base::JSONReader::ReadDict(input_json,
+                                          base::JSON_PARSE_CHROMIUM_EXTENSIONS);
+  if (!input.has_value()) {
+    std::move(callback).Run(
+        CreateContentBlocksForText("Error: failed to parse input JSON"), {});
+    return;
+  }
+
+  const std::string* query = input->FindString(kPropertyQuery);
+  if (!query || query->empty()) {
+    std::move(callback).Run(
+        CreateContentBlocksForText("Error: missing or empty 'query' field"),
+        {});
+    return;
+  }
+
+  // Match the chrome://history and omnibox gating: reject queries with fewer
+  // than `search_query_minimum_word_count` words (default 2). The service
+  // API itself does not enforce this, but very short queries produce noisy
+  // semantic results.
+  const int min_words = history_embeddings::GetFeatureParameters()
+                            .search_query_minimum_word_count;
+  if (static_cast<int>(history_embeddings::CountWords(*query)) < min_words) {
+    std::move(callback).Run(
+        CreateContentBlocksForText(base::StrCat(
+            {"Error: 'query' must contain at least ",
+             base::NumberToString(min_words),
+             " words. Provide a longer natural-language description."})),
+        {});
+    return;
+  }
+
+  size_t count = kDefaultResultCount;
+  if (auto requested = input->FindInt(kPropertyCount)) {
+    count = static_cast<size_t>(
+        std::clamp(*requested, 1, static_cast<int>(kMaxResultCount)));
+  }
+
+  std::optional<base::Time> time_range_start;
+  if (auto days = input->FindInt(kPropertyTimeRangeStartDaysAgo);
+      days && *days > 0) {
+    time_range_start = base::Time::Now() - base::Days(*days);
+  }
+
+  bool include_all_passages =
+      input->FindBool(kPropertyIncludeAllPassages).value_or(true);
+
+  // Defense in depth: the tool is only published when the flag is on, but
+  // re-check here in case the flag flipped between advertisement and use.
+  if (!history_embeddings::IsHistoryEmbeddingsEnabledForProfile(profile_)) {
+    std::move(callback).Run(
+        CreateContentBlocksForText(
+            "Error: history embeddings is not enabled for this profile"),
+        {});
+    return;
+  }
+
+  auto* service = HistoryEmbeddingsServiceFactory::GetForProfile(profile_);
+  if (!service) {
+    std::move(callback).Run(
+        CreateContentBlocksForText(
+            "Error: history embeddings service is unavailable"),
+        {});
+    return;
+  }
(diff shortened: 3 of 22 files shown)


components/ai_chat/core/common/mojom/ai_chat.mojom
```diff
@@ -43,6 +43,11 @@ const string kDeepResearchToolName = "deep_research";
 // Tool artifact types
 const string kLineChartArtifactType = "line_chart";
 
+// Carries a JSON array of HTTPS URL strings that the tool's output
+// references. The conversation UI uses this to permit the assistant's reply
+// to render those URLs as anchors. Not rendered visually.
+const string kTrustedLinksArtifactType = "trusted_links";
+
```
petemill, line 55: This is a cool use of artifacts - nice thinking @darkdh and nice work @DJAndries on adding the artifacts concept.

Without this we could have considered inspecting the tool output and extracting the links, but this more generalized approach will be cleaner!

DJAndries: awesome! glad we could find another use case

browser/ai_chat/tools/history_search_tool.cc
```diff
@@ -0,0 +1,282 @@
...
+            "Whether to include the indexed text passages from each "
+            "matching page (the actual article segments) under "
+            "'passages', sorted by relevance. Defaults to true -- this "
+            "is normally what you want, since the passages are the only "
+            "way to access the page's content from history. Set to "
+            "false only to keep the response compact when the user "
+            "wants a list of pages and no content is needed; in that "
+            "case each result has a single 'snippet' field with just "
+            "the best-matching passage.")}});
+}
```
petemill, line 145: This seems like quite a few tokens. I don't have any concrete recommendations on reducing except to say there is a bit of repetition for the parameters in the description. Did you find that was necessary in your testing?

browser/ai_chat/tools/history_search_tool.h
```diff
@@ -0,0 +1,79 @@
...
+  std::string_view Name() const override;
+  std::string_view Description() const override;
+  std::optional<base::DictValue> InputProperties() const override;
+  std::optional<std::vector<std::string>> RequiredProperties() const override;
+  std::variant<bool, mojom::PermissionChallengePtr>
+  RequiresUserInteractionBeforeHandling(
+      const mojom::ToolUseEvent& tool_use) const override;
+  void UseTool(const std::string& input_json,
+               UseToolCallback callback) override;
+
```
petemill, line 63: Do you think we should have a browser test verifying that this performs the search or is that too difficult to do without sample embeddings? Can the history embeddings search service be mocked?

If it's possible to browsertest then we do have some helpers at `ai_chat_conversation_ui_browsertest_base.h` where you can create a conversation and mock a call to use the tool with lower overhead via `SetupMockGenerateAssistantResponse` and `CreateToolUseEvent`.

As it stands we don't have a "success" unit test that RunTool causes a valid query to be sent to the service.

```cpp
CreateConversationWithMockEngine();
// Submit first message
{
  auto generate_future = SetupMockGenerateAssistantResponse();
  conversation_handler_->SubmitHumanConversationEntry(
      "Search my history for things, std::nullopt);
  auto callbacks = generate_future->Take();

  // Send first message response
  // Simulate tool use event
  callbacks.data_callback.Run(EngineConsumer::GenerationResultData(
      mojom::ConversationEntryEvent::NewToolUseEvent(
          CreateSemanticSearchToolUseEvent("tool_id_1", "expected sample query")),
      std::nullopt));
  // Complete first message response
  std::move(callbacks.completed_callback)
      .Run(base::ok(
          EngineConsumer::GenerationResultData(nullptr, std::nullopt)));
}

// Verify the history search service was queried with the expected query
...
```

petemill (commented):
Looks great, just a comment about browsertests

browser/ai_chat/tools/history_search_tool.cc
darkdh: let me see if I can reduce these further more

browser/ai_chat/tools/history_search_tool_unittest.cc
```diff
@@ -0,0 +1,246 @@
...
+  AddPassage(row, "high score passage", /*score=*/0.9f, /*word_count=*/3);
+  result.scored_url_rows.push_back(std::move(row));
+
+  std::string json = internal::BuildHistorySearchResultJson(
+      "article", result, /*include_all_passages=*/false);
+
+  auto parsed =
+      base::JSONReader::ReadDict(json, base::JSON_PARSE_CHROMIUM_EXTENSIONS);
+  ASSERT_TRUE(parsed.has_value());
+  const base::DictValue& entry = (*parsed->FindList("results"))[0].GetDict();
```
netzenbot: `(*parsed->FindList("results"))[0].GetDict()` dereferences the list pointer and accesses index 0 without asserting either is safe. Add `const base::ListValue* results = parsed->FindList("results"); ASSERT_TRUE(results); ASSERT_EQ(results->size(), 1u);` before accessing the entry, matching the pattern used in the other JSON tests. ([TI-034](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-034))

browser/ai_chat/tools/history_search_tool_unittest.cc
```diff
@@ -0,0 +1,246 @@
...
+  EXPECT_TRUE(props->Find("query"));
+  EXPECT_TRUE(props->Find("count"));
+  EXPECT_TRUE(props->Find("time_range_start_days_ago"));
+}
+
+TEST(HistorySearchToolJsonTest, EmptyResultProducesEmptyResultsArray) {
+  history_embeddings::SearchResult result;
+  std::string json = internal::BuildHistorySearchResultJson(
+      "kittens", result, /*include_all_passages=*/false);
+
```
netzenbot, line 175: nit: Per [CSA-051](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-051), test code should use `base::test::ParseJsonDict()` instead of `base::JSONReader::ReadDict()` for better readability and error messages. This applies to all four `ReadDict` calls in the `HistorySearchToolJsonTest` tests (neighboring test files like `target_util_unittest.cc` follow this pattern).

browser/ai_chat/tools/history_search_tool.cc
```diff
@@ -0,0 +1,282 @@
...
+
+void HistorySearchTool::UseTool(const std::string& input_json,
+                                UseToolCallback callback) {
+  auto input = base::JSONReader::ReadDict(input_json,
+                                          base::JSON_PARSE_CHROMIUM_EXTENSIONS);
+  if (!input.has_value()) {
+    std::move(callback).Run(
+        CreateContentBlocksForText("Error: failed to parse input JSON"), {});
+    return;
```
netzenbot, line 180: nit: The output JSON keys in `ScoredUrlRowToDict` and `BuildHistorySearchResultJson` are inline string literals (`"title"`, `"url"`, `"passages"`, `"snippet"`, `"last_visit_time"`, `"query"`, `"results"`). The input property keys already follow the pattern with named constants — apply the same here. [CSA-054](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-054)

browser/ai_chat/tools/history_search_tool_unittest.cc
```diff
@@ -0,0 +1,246 @@
...
+  auto props = tool->InputProperties();
+  ASSERT_TRUE(props.has_value());
+  EXPECT_TRUE(props->Find("query"));
+  EXPECT_TRUE(props->Find("count"));
+  EXPECT_TRUE(props->Find("time_range_start_days_ago"));
+}
+
+TEST(HistorySearchToolJsonTest, EmptyResultProducesEmptyResultsArray) {
+  history_embeddings::SearchResult result;
+  std::string json = internal::BuildHistorySearchResultJson(
```
netzenbot, line 173: nit: The JSON tests use `base::JSONReader::ReadDict()` directly; prefer `base::test::ParseJsonDict()` per [TI-020](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-020). It's simpler and produces cleaner assertions — e.g., `EXPECT_EQ(actual_dict, base::test::ParseJsonDict(R"({"query":"kittens"})"))` instead of manually finding and comparing individual keys.

browser/ai_chat/tools/history_search_tool_unittest.cc
```diff
@@ -0,0 +1,246 @@
...
+  EnableFeature();
+  auto tool = CreateTool();
+  mojom::ToolUseEvent event(tool->Name().data(), "1", "{}", std::nullopt,
+                            std::nullopt, nullptr, false);
+  auto result = tool->RequiresUserInteractionBeforeHandling(event);
+  ASSERT_TRUE(std::holds_alternative<bool>(result));
+  EXPECT_FALSE(std::get<bool>(result));
+}
+
+TEST_F(HistorySearchToolTest, MalformedJson) {
```
netzenbot: nit: `MalformedJson`, `MissingQuery`, `EmptyQuery`, and `SingleWordQueryRejected` are structurally identical — each calls `EnableFeature()`, creates a tool, and checks a specific error substring. Consider collapsing them into a single table-driven (parameterized) test with `SCOPED_TRACE` to keep coverage while removing the repetition. ([TI-023](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-023))

browser/ai_chat/tools/history_search_tool.cc
darkdh: compacting it through 1c0bf5d4793014999b784a3eb8a5bca65b5fba18

browser/ai_chat/tools/history_search_tool.h
darkdh: added browser test through mock service in 432612d3602e75b85dfa81ade043ee3192a488fe

browser/ai_chat/tools/history_search_tool_unittest.cc
darkdh: fixed in 36cba4bc42c1f66d18617a3dd1e5e9c7634e96f1

browser/ai_chat/tools/history_search_tool_unittest.cc
darkdh: fixed in 6c883e9d644686408045af5157190ea3dd309ef6

browser/ai_chat/tools/history_search_tool.cc
darkdh: fixed in aca3d997d426207e9a1e42583c87026f04dbd446

browser/ai_chat/tools/history_search_tool_unittest.cc
darkdh: fixed in 2ddbf7353b684739e894598e79d7d41bbdab70d3

browser/ai_chat/tools/history_search_tool_unittest.cc
darkdh: fixed in f096b105927d212a81c522a8edbc033d6df15339

browser/ai_chat/tools/history_search_tool_browsertest.cc
```diff
@@ -0,0 +1,274 @@
...
+          !tool_use->output.has_value() || tool_use->output->empty()) {
+        continue;
+      }
+      const auto& blocks = *tool_use->output;
+      ASSERT_TRUE(blocks[0]->is_text_content_block());
+      auto parsed =
+          base::JSONReader::ReadDict(blocks[0]->get_text_content_block()->text,
+                                     base::JSON_PARSE_CHROMIUM_EXTENSIONS);
+      ASSERT_TRUE(parsed.has_value());
+      EXPECT_EQ(*parsed->FindString("query"), "kittens article");
```
netzenbot: nit: The `find_completed_tool_use` lambda parses the tool output with `base::JSONReader::ReadDict()` and manually checks individual fields. Prefer `base::test::ParseJsonDict()` for the comparison, as already done in the unit tests — e.g., `EXPECT_EQ(base::test::ParseJsonDict(text), base::test::ParseJsonDict(R"({...})"))`. ([TI-020](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-020))

darkdh: fixed in fda12094b0b2ddc81a4936db93845dadbd5f16b3

browser/ai_chat/browser_tool_provider.h
```diff
@@ -39,11 +40,16 @@ class BrowserToolProvider : public ToolProvider {
   // ToolProvider implementation
   std::vector<base::WeakPtr<Tool>> GetTools() override;
 
+  HistorySearchTool* GetHistorySearchToolForTesting() {
+    return history_search_tool_.get();
+  }
+
```
netzenbot, line 46: nit: `GetHistorySearchToolForTesting()` makes a private member accessible via a public method. Per [ARCH-020](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-020), prefer a `friend class HistorySearchToolBrowserTest;` declaration in the `private:` section instead of adding a public test accessor.

darkdh: I will add `CHECK_IS_TEST()` but public `GetHistorySearchToolForTesting()` is better than friending class for test specific functions

darkdh: added in 56fd337ec692ffadb15b2852ae2110232e58f16c

browser/ai_chat/BUILD.gn
```diff
@@ -64,11 +66,15 @@ static_library("ai_chat") {
...
     "//chrome/browser/actor",
+    "//chrome/browser/history_embeddings",
     "//chrome/browser/profiles:profile",
     "//chrome/browser/ui/tabs:tabs_public",
     "//chrome/common",
     "//chrome/common:channel_info",
     "//components/browsing_data/core",
+    "//components/history/core/browser",
+    "//components/history_embeddings/content",
+    "//components/history_embeddings/core",
```
netzenbot: `history_search_tool.h` includes `components/history_embeddings/core/history_embeddings_search.h` directly, so `//components/history_embeddings/core` must be in `public_deps`, not `deps`, to give consumers of this target the transitive include path they need. ([BS-024](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-024))

browser/ai_chat/tools/history_search_tool.cc
```diff
@@ -0,0 +1,288 @@
...
+  ToolArtifacts artifacts;
+  if (!result.scored_url_rows.empty()) {
+    base::ListValue links;
+    for (const auto& row : result.scored_url_rows) {
+      if (row.row.url().is_valid() &&
+          row.row.url().SchemeIs(url::kHttpsScheme)) {
+        links.Append(row.row.url().spec());
+      }
+    }
+    if (!links.empty()) {
```
netzenbot, line 286: nit: `SetSearchForTesting()` appears before `Name()` in the `.cc` but is declared after `UseTool()` in the `.h`. Per [CS-017](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-017), the definition order in `.cc` should match the declaration order in `.h`. Move `SetSearchForTesting()` to after `UseTool()` in the `.cc`.

browser/ai_chat/tools/history_search_tool.cc
```diff
@@ -0,0 +1,288 @@
+// Copyright (c) 2026 The Brave Authors. All rights reserved.
```
netzenbot, line 1: nit: The `browser/ai_chat/tools/README.md` only describes Content Agent Tools (owned by `ContentAgentToolProvider`). `HistorySearchTool` is a different category — a standalone browser tool owned by `BrowserToolProvider`. Consider adding a brief section to the README describing this category so future contributors know which provider to use. ([DOC-001](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-001))

browser/ai_chat/tools/history_search_tool_browsertest.cc
```diff
@@ -0,0 +1,270 @@
...
+          base::StrCat({R"({
+        "query": "kittens article",
+        "results": [{
+          "title": "Canned title",
+          "url": "https://example.com/canned",
+          "last_visit_time": ")",
+                        base::TimeFormatAsIso8601(CannedVisitTime()),
+                        R"("
+        }]
+      })"})));
```
petemill: very nit: `EXPECT_THAT(base::test::ParseJsonDict(...), base::Test::IsJson(...))`

not sure if meaningfully better though!

browser/ai_chat/tools/history_search_tool_unittest.cc
```diff
@@ -0,0 +1,246 @@
...
+  auto row = MakeRow("https://example.com/article", u"Article title");
+  AddPassage(row, "low score passage", /*score=*/0.1f, /*word_count=*/3);
+  AddPassage(row, "high score passage", /*score=*/0.9f, /*word_count=*/3);
+  result.scored_url_rows.push_back(std::move(row));
+
+  std::string json = internal::BuildHistorySearchResultJson(
+      "article", result, /*include_all_passages=*/false);
+
+  EXPECT_EQ(base::test::ParseJsonDict(json),
+            base::test::ParseJsonDict(base::StrCat({R"({
```
petemill: ditto re: JSON matcher nit

components/ai_chat/resources/untrusted_conversation_frame/components/conversation_entries/index.tsx
```diff
@@ -312,6 +312,32 @@ function ConversationEntries(props: { scrollToBottom: () => void }) {
...
+                return Array.isArray(parsed)
+                  ? parsed.filter((u): u is string => typeof u === 'string')
+                  : []
+              } catch {
+                return []
+              }
+            }) ?? [],
+      )
+    })
+
```
petemill: We also need to do this in `use_extract_task_data.ts` as these won't have any effect. This `groupTrustedLinks` will only be used for responses that have no tool uses and not ones that are considered "tasks" which wrap the response in an `AssistantTask` component. Previously (when you started this PR), it was only responses that had 2x tool uses, but now in master it's 1x (via `isAssistantGroupTask`) and so it will be every response that calls this new tool. So for now we only need it in `useExtractTaskData` but it wouldn't hurt to extract both functions to ~a function and have it in both places, or~ a single function which combines the allowed links from all group completion events and trusted-links artifacts and passes as `allowedLinks` to both `AssistantResponse` and `AssistantTask`.

- finds all web sources events in all entries in the group
- finds all trusted links artifacts in all tool use events in all entries in the group

browser/ai_chat/tools/history_search_tool.cc
```diff
@@ -0,0 +1,288 @@
...
+#include "components/history/core/browser/url_row.h"
+#include "components/history_embeddings/content/history_embeddings_service.h"
+#include "components/history_embeddings/core/history_embeddings_features.h"
+#include "url/url_constants.h"
+
+namespace ai_chat {
+
+namespace {
+
+constexpr char kToolName[] = "semantic_history_search";
```
petemill: Please add to `ai_chat.mojom` and add a label to `get_tool_label.ts`. Something present tense, e.g. `Searching your browser history for "{query}"`

browser/ai_chat/tools/history_search_tool.cc
```diff
@@ -0,0 +1,288 @@
...
+    const {
+  return std::vector<std::string>{kPropertyQuery};
+}
+
+std::variant<bool, mojom::PermissionChallengePtr>
+HistorySearchTool::RequiresUserInteractionBeforeHandling(
+    const mojom::ToolUseEvent& tool_use) const {
+  // The search runs on locally-stored history that the user has already
+  // opted into via the brave-history-embeddings flag; no remote send.
+  return false;
```
petemill: Whilst the semantic search happens locally so that it doesn't send your whole history database to the LLM, it does send history to the remote LLM in the form of the results of any search the LLM decides to do. Did we check with security & privacy review that we don't need to get user permission first?

petemill (changes requested):

browser/ai_chat/tools/history_search_tool.cc
darkdh: https://github.com/brave/reviews/issues/2253
opened one for security team to decide

browser/ai_chat/tools/history_search_tool.cc
```diff
@@ -0,0 +1,288 @@
...
+std::optional<std::vector<std::string>> HistorySearchTool::RequiredProperties()
+    const {
+  return std::vector<std::string>{kPropertyQuery};
+}
+
+std::variant<bool, mojom::PermissionChallengePtr>
+HistorySearchTool::RequiresUserInteractionBeforeHandling(
+    const mojom::ToolUseEvent& tool_use) const {
+  // The search runs on locally-stored history that the user has already
+  // opted into via the brave-history-embeddings flag; no remote send.
```
diracdeltas: these comments are a bit misleading. they should be clear that results are remotely sent.

components/ai_chat/resources/untrusted_conversation_frame/components/conversation_entries/index.tsx
```diff
@@ -312,6 +312,32 @@ function ConversationEntries(props: { scrollToBottom: () => void }) {
...
       : null
 
+    // Trusted-links artifacts are flattened across the whole group because
+    // a client-side tool call lives in a separate assistant entry from the
+    // follow-up response that references the tool's URLs. Without this, the
+    // response's anchors would be filtered out by MarkdownRenderer since the
+    // trust-list lives on the tool_use_event in the prior entry.
+    // (sourcesEvent stays per-entry below -- web search citations belong to
+    // the specific response that produced them.)
+    const groupTrustedLinks: string[] = group.flatMap((entry) => {
```
diracdeltas: IIUC, 'trusted' links can be arbitrary HTTPS links; they aren't really audited in any way. if so can we rename them to not imply that they are trustworthy?

also is there a check somewhere to validate that the link is HTTPS? (no arbitrary protocols allowed)

diracdeltas: nvm on the second point, i see https://github.com/brave/brave-core/pull/36385/changes#diff-f7d94242dc69a65f3c53702fbb672283856e1b5d5e99ef5ea5ba7b86c3c900b9R271

darkdh: They are extracted from history an user visited so they are equivalent to the links in brave://history/.
I'm not sure how to craft up a arbitrary link without visiting the real page except accessing to the database file on disk and insert it directly

diracdeltas: i think calling them visitedLinks would be more clear but not a big deal either way

browser/ai_chat/BUILD.gn
darkdh: fixed in b5b5e26fcc48fb8b1e389cb7bb14f84805e4bc27

browser/ai_chat/tools/history_search_tool.cc
darkdh: fixed in 28cd531ebfc5685cabfddf76193adb8f6fa6cfe9

browser/ai_chat/tools/history_search_tool.cc
darkdh: fixed in d6adb7e5068797090b60b815f64dcc83894ce21b

browser/ai_chat/tools/history_search_tool_unittest.cc
darkdh: fixed in e3e9359d243c63ab93b4df8eb008a978b2504e96

browser/ai_chat/tools/history_search_tool_browsertest.cc
darkdh: fixed in a55c7c502f4b41d75667e1ab3db37523843e44d0

components/ai_chat/resources/untrusted_conversation_frame/components/conversation_entries/index.tsx
darkdh: extracted to a single function in 37ba19cd7e4384426e64b597aeef63df0486f467

browser/ai_chat/tools/history_search_tool.cc
darkdh: tool name is in now in mojom and also added label in effb7a68638094b4fa3456e503dc6d4fb38c41d8

browser/ai_chat/tools/history_search_tool.cc
darkdh: it is now a per conversation permission 038954e62b9232608ff0c015028fca5197edb020

browser/ai_chat/tools/history_search_tool.cc
darkdh: changes it to ask user permission in https://github.com/brave/brave-core/commit/038954e62b9232608ff0c015028fca5197edb020

components/ai_chat/resources/untrusted_conversation_frame/components/conversation_entries/index.tsx
darkdh: sounds good and I will change it to `visitedLinks`

darkdh: renamed in e5d1b2c7ffa2a3f73b6c590ecfa33d3981865ab7

browser/ai_chat/tools/history_search_tool_browsertest.cc
```diff
@@ -0,0 +1,305 @@
...
+      }
+    }
+    return false;
+  }));
+
+  EXPECT_TRUE(fake.last_query_.empty())
+      << "Search service must not be called with unparseable args";
+}
+
+}  // namespace ai_chat
```
netzenbot, line 301: `blocks[0]` is accessed without first asserting `!blocks.empty()` (or `ASSERT_GE(blocks.size(), 1u)`). The `find_completed_tool_use` lambda checks `!tool_use->output->empty()` internally, but that guarantee isn't surfaced as a fatal assertion in the test body — if the invariant ever breaks, the test will crash rather than emit a clear failure. Add `ASSERT_FALSE(blocks.empty())` before accessing index 0. ([TI-034](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-034))

components/ai_chat/resources/untrusted_conversation_frame/components/conversation_entries/conversation_entries_utils.test.ts
```diff
@@ -475,3 +476,93 @@ describe('isAssistantGroupTask', () => {
...
+  const makeArtifact = (contentJson: string): Mojom.ToolArtifact => ({
+    id: null,
+    type: Mojom.VISITED_LINKS_ARTIFACT_TYPE,
+    contentJson,
+  })
+
+  const turnWithArtifacts = (
+    artifacts: Mojom.ToolArtifact[],
+  ): Mojom.ConversationTurn =>
+    createConversationTurnWithDefaults({
```
netzenbot, line 491: nit: `toolName: 'semantic_history_search'` duplicates the mojom constant you just added. Use `Mojom.SEMANTIC_HISTORY_SEARCH_TOOL_NAME` so this can't drift out of sync. ([CS-040](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-040))

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/use_extract_task_data.test.ts
```diff
@@ -219,5 +219,82 @@ describe('useExtractTaskData', () => {
...
+    it('should include URLs from visited_links artifacts', () => {
+      // Client-side tools (e.g. semantic history search) attach a
+      // visited_links artifact carrying a JSON array of HTTPS URLs. These
+      // belong in allowedLinks so the follow-up response's anchors render.
+      const assistantEntries: Mojom.ConversationTurn[] = [
+        createConversationTurnWithDefaults({
+          characterType: Mojom.CharacterType.ASSISTANT,
+          events: [
+            getToolUseEvent({
+              toolName: 'semantic_history_search',
```
netzenbot: nit: `toolName: 'semantic_history_search'` duplicates the mojom constant. Use `Mojom.SEMANTIC_HISTORY_SEARCH_TOOL_NAME` here instead. ([CS-040](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-040))

browser/ai_chat/tools/history_search_tool.cc
```diff
@@ -0,0 +1,302 @@
...
+    return false;
+  }
+  // The search itself runs entirely on-device against the local embeddings
+  // index -- the query and the full history never leave the device. Only
+  // the matching pages' titles, URLs, and indexed passages are sent to the
+  // remote LLM as the tool's output. Ask the user to confirm before doing
+  // that.
+  return mojom::PermissionChallenge::New(
+      std::nullopt,
+      "Brave AI wants to search your browsing history. The search runs "
```
netzenbot: nit: `-- ` should be an em dash `—` with no spaces around it per the Brave Style Guide. Change `against your history -- your full history` to `against your history—your full history`. ([SG-010](https://github.com/brave/brave-core/tree/master/docs/best-practices/style-guide.md#SG-010))

browser/ai_chat/tools/history_search_tool.cc
```diff
@@ -0,0 +1,302 @@
...
+      "locally on your device against your history -- your full history "
+      "is not sent anywhere. Only the matching pages' titles, URLs, and "
```
netzenbot: nit: Per the Brave Style Guide, use contractions in UI text. Change `is not sent anywhere` to `isn't sent anywhere`. ([SG-002](https://github.com/brave/brave-core/tree/master/docs/best-practices/style-guide.md#SG-002))

browser/ai_chat/tools/history_search_tool_browsertest.cc
darkdh: Fixed by 8f23e793d31.

components/ai_chat/resources/untrusted_conversation_frame/components/conversation_entries/conversation_entries_utils.test.ts
darkdh: Fixed by 27c30755d44.

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/use_extract_task_data.test.ts
darkdh: Fixed by 27c30755d44.

browser/ai_chat/tools/history_search_tool.cc
darkdh: Fixed by 727533d55c2 (the em dash was a symptom of an awkward clause join; rewrote the sentence with "so" instead so there's no dash at all).

browser/ai_chat/tools/history_search_tool.cc
darkdh: Fixed by 727533d55c2.

browser/ai_chat/tools/history_search_tool_browsertest.cc
```diff
@@ -0,0 +1,306 @@
...
+          std::nullopt, std::nullopt, nullptr, false)),
+      std::nullopt));
+  std::move(callbacks.completed_callback)
+      .Run(base::ok(
+          EngineConsumer::GenerationResultData(nullptr, std::nullopt)));
+
+  // The permission challenge fires before UseTool runs; approve it so the
+  // tool actually executes and surfaces its parse error.
+  ASSERT_TRUE(base::test::RunUntil([this]() {
+    for (const auto& turn : conversation_handler_->GetConversationHistory()) {
```
netzenbot: nit: The `ASSERT_TRUE(base::test::RunUntil([this]() { ... permission_challenge ... }))` block is copy-pasted identically in both `ToolUseDispatchesSearchAndReturnsResult` and `MalformedArgsProducesErrorOutput`. Per [TI-024](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-024), extract it into a helper (e.g., `WaitForAndApprovePermissionChallenge(const std::string& tool_id)`) so the two tests share a single definition.

darkdh: Fixed by a062bb33dc1.

browser/ai_chat/tools/history_search_tool_browsertest.cc
```diff
@@ -0,0 +1,297 @@
...
+    ASSERT_TRUE(provider);
+    auto* tool = provider->GetHistorySearchToolForTesting();
+    ASSERT_TRUE(tool);
+    tool->SetSearchForTesting(fake);
+  }
+
+  // Waits for the tool's permission challenge to land in conversation
+  // history, then approves it so the tool can execute.
+  void WaitForAndApprovePermissionChallenge(const std::string& tool_id) {
+    ASSERT_TRUE(base::test::RunUntil([this]() {
```
netzenbot, line 126: nit: `WaitForAndApprovePermissionChallenge` contains `ASSERT_TRUE` but doesn't accept or forward a `base::Location` parameter. Per [TI-015](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-015), add `base::Location location = base::Location::Current()` as a parameter and `SCOPED_TRACE(location.ToString())` at the top of the helper so failures report the caller's line rather than the helper's.

darkdh: Fixed by 173141560f3.

netzenbot (approved):

diracdeltas:
this will require a privacy policy update https://github.com/brave/reviews/issues/2265

maybe also some leo support docs that need to be updated https://support.brave.app/hc/en-us/sections/20991947768717-Leo-AI @brave/user-support-team

components/ai_chat/resources/untrusted_conversation_frame/components/conversation_entries/conversation_entries_utils.ts
```diff
@@ -279,3 +279,32 @@ export function getToolArtifacts(
...
+
+/**
+ * Collects HTTPS URLs from `visited_links` artifacts on tool_use events
+ * across the group. Client-side tools (e.g. semantic history search) emit
+ * these as a sidechannel trust-list so the assistant's reply can render the
+ * tool's URLs as anchors. Bad JSON or non-string array entries are skipped.
+ */
+export function getGroupVisitedLinks(
+  group: Mojom.ConversationTurn[],
+): string[] {
```
petemill, line 327: We can do as a follow-up, but I think if we're introducing this funciton, it would be cleaner to also collect the URLs from the sources events in it, so we're not doing it twice (in `conversation_entries/index.tsx` and `use_extract_task_data.ts`), and probably name this function something like `getGroupAllowedLinks`. Going further, like my original suggestion, we can also just pass the result from this function to both AssistantTask and AssistantResponse in identical `allowedLinks` props, so that `useExtractTaskData` doesn't even need to call this function.

browser/ai_chat/tools/history_search_tool.cc
```diff
@@ -0,0 +1,301 @@
...
+  // the matching pages' titles, URLs, and indexed passages are sent to the
+  // remote LLM as the tool's output. Ask the user to confirm before doing
+  // that.
+  return mojom::PermissionChallenge::New(
+      std::nullopt,
+      "Brave AI wants to search your browsing history. The search runs "
+      "locally on your device, so your full history isn't sent anywhere. "
+      "Only the matching pages' titles, URLs, and indexed text passages "
+      "will be sent to Brave AI as part of this conversation.");
+}
```
petemill: This should be a translated string. OK for follow-up before enabling by default, but should probably leave a TODO at least, or just fix here. If there is no formatting required then probably ok to do this via C++, but see `get_tool_permission_implications.tsx` for how I did this for the Tab Management Tool where there is formatting. You could also use that function if you wanted to show what the queries are, but that might come anyway from a near future re-design where we are going to put all the tool arguments in all permission prompts.

petemill (approved):
Looking great. Will defer to you about whether to do the string here or in a follow-up before enabling the flag.

components/ai_chat/resources/untrusted_conversation_frame/components/conversation_entries/conversation_entries_utils.ts
darkdh: refactored in https://github.com/brave/brave-core/pull/36385/commits/bb9447f1212dc61c1cfd156d44a2a4108c0d7ef2

browser/ai_chat/tools/history_search_tool.cc
darkdh: fixed in 418269ee4cca2a6f8f24808a73df3c98797cb982 and 829c8a0f875ed6e6b39e2bf8e99a925fc85d5859

PR #36388, opened 2026-05-12
[AI Chat] Fix empty title generation

### LINKED ISSUE #55480: [AI Chat] Fix empty title generation
We are seeing empty title generation request in content agent conversations (not always), and chart generation conversations. 
The cause of this is we unintentionally uses assistant response when uploaded files was empty vector, and when tool calling is involved, sometimes LLM won't give us initial responses so assistant response used was empty, which leads to a title generation request with empty text.
I didn't check if this only starts to be this case recently (uploaded files was empty but not std::nullopt), but it wasn't our intention to use assistant response in both cases (uploaded files is std::nullopt or empty).

yrliou:
We were seeing title generation requests with empty text from content
agent conversations (not always) and chart generation conversations.
The root cause of this is we unintentionally uses assistant response
when uploaded files is an empty vector, and when tool calling is
involved, sometimes LLM won't give us initial responses so assistant
response we used would be empty, which leads to a title generation
request with empty text.

This PR fixes the bug by:
1. Only use assistant response if there are non-empty uploaded files and
   non-empty assistant response.
2. Do not issue title generation requests when the text to be sent is
   empty.

Resolves https://github.com/brave/brave-browser/issues/55480

Test plan:
1. Starts browser with `--enable-features="AIChatCodeExecutionTool:charts/true"`
2. Create a new Leo conversation and send "chart fib n 1 to 10" using automatic model
3. Title should be generated (please make sure code execution tool was triggered)
Screenshot 2026-05-12 at 4 03 15 PM

### DIFF
--- a/components/ai_chat/core/browser/engine/oai_message_utils.cc
+++ b/components/ai_chat/core/browser/engine/oai_message_utils.cc
@@ -617,14 +617,26 @@ BuildOAIGenerateConversationTitleMessages(
         mojom::PageExcerptContentBlock::New(*first_turn->selected_text)));
   }
 
-  // Add a message for title generation.
-  // Use first assistant response as the content if files are uploaded (image,
-  // PDF), otherwise use the first human turn text.
+  // Use the first assistant response when it's non-empty as the title source
+  // when the user uploaded files (image, PDF) since it contains more useful
+  // context for title generation.
+  const bool use_assistant_text = first_turn->uploaded_files &&
+                                  !first_turn->uploaded_files->empty() &&
+                                  !assistant_turn->text.empty();
+  std::string title_text = use_assistant_text
+                               ? assistant_turn->text
+                               : EngineConsumer::GetPromptForEntry(first_turn);
+
+  // Withdraw the request entirely if we have nothing to summarize. Sending an
+  // empty title block wastes a server round-trip and cannot produce a useful
+  // title. Title errors are silently dropped in OnTitleGenerated, so the
+  // conversation simply keeps its placeholder title.
+  if (title_text.empty()) {
+    return std::nullopt;
+  }
+
   msg.content.push_back(mojom::ContentBlock::NewRequestTitleContentBlock(
-      mojom::RequestTitleContentBlock::New(
-          first_turn->uploaded_files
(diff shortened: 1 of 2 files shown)


DJAndries (approved):

components/ai_chat/core/browser/engine/oai_message_utils.cc
```diff
@@ -617,14 +617,26 @@ BuildOAIGenerateConversationTitleMessages(
...
+                               : EngineConsumer::GetPromptForEntry(first_turn);
+
+  // Withdraw the request entirely if we have nothing to summarize. Sending an
+  // empty title block wastes a server round-trip and cannot produce a useful
+  // title. Title errors are silently dropped in OnTitleGenerated, so the
+  // conversation simply keeps its placeholder title.
+  if (title_text.empty()) {
+    return std::nullopt;
+  }
+
```
petemill, line 637: Perhaps a TODO to wait for a server response to see if there's any text we can create a title from?

petemill (approved):

components/ai_chat/core/browser/engine/oai_message_utils.cc
yrliou: It might not be the best place to put a TODO comment, I would prefer creating a general issue to retry title generation later when title is not available (currently we only try once when history.size == 2) either due to first two turns has no usable text or if the operation failed for any reason, if that's fine with you.

yrliou: https://github.com/brave/brave-browser/issues/55481

netzenbot (approved):

kjozwiak:

```
Brave | 1.92.56 Chromium: 148.0.7778.167 (Official Build) nightly (64-bit)
-- | --
Revision | f66f43b0b8f4c063fd88671464a11d1abd2eaf4d
```

Using the STR/Cases outlined https://github.com/brave/brave-core/pull/36388#issue-4433303745, ensured that the correct tile is being used re: `conversation history` when launching the browser using `--enable-features="AIChatCodeExecutionTool:charts/true"` as per the following:

Screenshot 2026-05-13 231955

PR #36391, opened 2026-05-13
[AI Chat] Expose status and error code to user on server error

### LINKED ISSUE #55483: [AI Chat] Expose status and error code to user on server error
To increase effectiveness of bug reports.

DJAndries:
Resolves https://github.com/brave/brave-browser/issues/55483

[
Screenshot-2026-05-12_20-14
](url)

## Test plan

Use a MITM proxy to intercept requests to ai-chat.bsg.brave.com. When the request is intercepted, modify the request body to contain invalid JSON. Release the request to the server and ensure you see a status code and error code in the error bubble.

### DIFF
--- a/components/ai_chat/core/browser/engine/conversation_api_v2_client.cc
+++ b/components/ai_chat/core/browser/engine/conversation_api_v2_client.cc
@@ -130,6 +130,20 @@ std::string_view GetContentBlockTypeString(
   }
 }
 
+int32_t ParseErrorCode(const base::Value& body) {
+  if (body.is_dict()) {
+    if (auto* error_dict = body.GetDict().FindDict("error")) {
+      if (auto* type_str = error_dict->FindString("type")) {
+        int parsed;
+        if (base::StringToInt(*type_str, &parsed)) {
+          return static_cast<int32_t>(parsed);
+        }
+      }
+    }
+  }
+  return 0;
+}
+
 }  // namespace
 
 // static
@@ -426,20 +440,6 @@ void ConversationAPIV2Client::PerformRequestWithCredentials(
   }
 }
 
-void ConversationAPIV2Client::ParseErrorCode(const base::Value& body) {
-  if (!body.is_dict()) {
-    return;
-  }
-  if (auto* error_dict = body.GetDict().FindDict("error")) {
-    if (auto* type_str = error_dict->FindString("type")) {
-      int parsed;
-      if (base::StringToInt(*type_str, &parsed)) {
-        last_error_code_ = static_cast<int32_t>(parsed);
-      }
-    }
-  }
-}
-
 void ConversationAPIV2Client::OnQueryCompleted(
     std::optional<CredentialCacheEntry> credential,
     GenerationCompletedCallback callback,
@@ -447,7 +447,6 @@ void ConversationAPIV2Client::OnQueryCompleted(
   const bool success = result.Is2XXResponseCode();
   // Handle successful request
   if (success) {
-    last_error_code_ = std::nullopt;
     std::optional<bool> is_near_verified = std::nullopt;
     std::optional<std::string> model_key = std::nullopt;
     const auto& headers = result.headers();
@@ -491,14 +490,12 @@ void ConversationAPIV2Client::OnQueryCompleted(
     error = mojom::APIError::ConnectionIssue;
   }
 
-  ParseErrorCode(result.value_body());
-  auto details = mojom::APIErrorDetails::New(
-      static_cast<int32_t>(result.response_code()),
-      last_error_code_.value_or(0));
-  last_error_code_ = std::nullopt;
+  auto details =
+      mojom::APIErrorDetails::New(static_cast<int32_t>(result.response_code()),
+                                  ParseErrorCode(result.value_body()));
 
-  std::move(callback).Run(base::unexpected(
-      EngineConsumer::Error(error, std::move(details))));
+  std::move(callback).Run(
+      base::unexpected(EngineConsumer::Error(error, std::move(details))));
 }
 
 void ConversationAPIV2Client::OnQueryDataReceived(
@@ -507,8 +504,6 @@ void ConversationAPIV2Client::OnQueryDataReceived(
   if (!result.has_value() || !result->is_dict()) {
     return;
   }
-
-  ParseErrorCode(*result);
 
   auto& result_params = result->GetDict();
   std::optional<std::string> model_key =
--- a/components/ai_chat/core/browser/engine/conversation_api_v2_client.cc
+++ b/components/ai_chat/core/browser/engine/conversation_api_v2_client.cc
@@ -19,6 +19,7 @@
 #include "base/json/json_writer.h"
 #include "base/memory/scoped_refptr.h"
 #include "base/strings/strcat.h"
+#include "base/strings/string_number_conversions.h"
 #include "base/strings/string_util.h"
 #include "base/types/expected.h"
 #include "brave/components/ai_chat/core/browser/constants.h"
@@ -425,6 +426,20 @@ void ConversationAPIV2Client::PerformRequestWithCredentials(
   }
 }
 
+void ConversationAPIV2Client::ParseErrorCode(const base::Value& body) {
+  if (!body.is_dict()) {
+    return;
+  }
+  if (auto* error_dict = body.GetDict().FindDict("error")) {
+    if (auto* type_str = error_dict->FindString("type")) {
+      int parsed;
+      if (base::StringToInt(*type_str, &parsed)) {
+        last_error_code_ = static_cast<int32_t>(parsed);
+      }
+    }
+  }
+}
+
 void ConversationAPIV2Client::OnQueryCompleted(
     std::optional<CredentialCacheEntry> credential,
     GenerationCompletedCallback callback,
@@ -432,6 +447,7 @@ void ConversationAPIV2Client::OnQueryCompleted(
   const bool success = result.Is2XXResponseCode();
   // Handle successful request
   if (success) {
+    last_error_code_ = std::nullopt;
     std::optional<bool> is_near_verified = std::nullopt;
     std::optional<std::string> model_key = std::nullopt;
     const auto& headers = result.headers();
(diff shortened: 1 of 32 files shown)


components/ai_chat/core/common/mojom/ai_chat.mojom
```diff
@@ -364,7 +365,7 @@ interface ConversationUI {
 
   OnAPIRequestInProgress(bool is_request_in_progress);
 
-  OnAPIResponseError(APIError error);
+  OnAPIResponseError(APIError api_error, APIErrorDetails? details);
```
fallaciousreasoning, line 369: I wonder if we should just return an `APIErrorDetails` struct, and move the error property onto that?

components/ai_chat/core/common/mojom/common.mojom
```diff
@@ -38,6 +38,11 @@ enum APIError {
   ServiceOverloaded
 };
 
+struct APIErrorDetails {
+  int32 status_code;
+  int32 error_code;
+};
```
fallaciousreasoning, line 45: I wonder if we should add an `APIError type` to this struct

components/api_request_helper/api_request_helper.h
```diff
@@ -161,6 +167,8 @@ class APIRequestHelper {
     ResponseConversionCallback conversion_callback_;
 
     bool is_sse_ = false;
+    std::optional<bool> is_response_success_;
+    base::Value error_value_;
```
fallaciousreasoning: Hey, I'm slightly worried about this as we've got into a little trouble for making `APIRequestHelper` do too much already. 

With this, we'll try and parse the error response to JSON even if the user of the APIRequestHelper doesn't care about it (which none of them do at the moment).

components/ai_chat/core/browser/engine/conversation_api_v2_client.cc
```diff
@@ -129,6 +130,20 @@ std::string_view GetContentBlockTypeString(
   }
 }
 
+int32_t ParseErrorCode(const base::Value& body) {
+  if (body.is_dict()) {
+    if (auto* error_dict = body.GetDict().FindDict("error")) {
+      if (auto* type_str = error_dict->FindString("type")) {
```
fallaciousreasoning, line 136: oh weird - we respond with a number in a string?

components/ai_chat/resources/untrusted_conversation_frame/components/alerts/error_connection.tsx
```diff
@@ -6,18 +6,37 @@
...
+  let detailsText: string | undefined
+  if (errorDetails) {
+    if (errorDetails.errorCode !== 0) {
+      detailsText = formatLocale(S.CHAT_UI_ERROR_NETWORK_DETAILS, {
+        $1: String(errorDetails.statusCode),
+        $2: String(errorDetails.errorCode),
+      })
+    } else {
+      detailsText = formatLocale(S.CHAT_UI_ERROR_NETWORK_STATUS_CODE, {
+        $1: String(errorDetails.statusCode),
```
fallaciousreasoning, line 30: do we need to cast these to strings?

netzenbot (approved):

components/ai_chat/core/common/mojom/ai_chat.mojom
DJAndries: I considered that initially, but it would require a lot of changes at many call sites.

components/ai_chat/core/common/mojom/common.mojom
DJAndries: I left it out for the same reason as https://github.com/brave/brave-core/pull/36391#discussion_r3236872419.

I created `EngineConsumer::Error` with a conversation constructor from `APIError` instead to avoid excessive result instantiation call site changes.

components/ai_chat/core/browser/engine/conversation_api_v2_client.cc
DJAndries: yeah, it is a little weird, but openai also uses a string for the type field, but they use it for actual strings rather than numbers. i'm going to change the `error_code` type into a string and avoid parsing, since parsing is not really needed.

components/api_request_helper/api_request_helper.h
DJAndries: I'm hoping we can allow an exception for this, especially given that this change only affects SSE which is only used by Playlist and Leo. We already parse error bodies for non-SSE requests from what I understand. Perhaps we could add a `enable_error_body_parsing` param to the request as a compromise.

components/ai_chat/resources/untrusted_conversation_frame/components/alerts/error_connection.tsx
DJAndries: yes, it won't build without them

components/ai_chat/core/browser/engine/conversation_api_v2_client.cc
DJAndries: updated

components/api_request_helper/api_request_helper.cc
```diff
@@ -497,6 +511,28 @@ void APIRequestHelper::URLLoaderHandler::OnParseJsonResponse(
...
+
+void APIRequestHelper::URLLoaderHandler::OnParseErrorBody(
+    ValueOrError result_value) {
+  current_decoding_operation_count_--;
+  if (result_value.has_value() && result_value->is_dict()) {
+    error_value_ = std::move(*result_value);
+  }
+  MaybeSendResult();
+}
+
```
petemill: Two thoughts about this:

1. If we're not in SSE mode (which seems like is always the case if it's an error?) then perhaps we shouldn't parse the JSON here and we should leave it up to the consumer? Or because we're `APIRequestHelper` and primarily a JSON-consuming API client, we assume that even errors are always JSON?

2. Could the error body itself be chunked over multiple chunks?

DJAndries: > 1. If we're not in SSE mode (which seems like is always the case if it's an error?) then perhaps we shouldn't parse the JSON here and we should leave it up to the consumer? Or because we're `APIRequestHelper` and primarily a JSON-consuming API client, we assume that even errors are always JSON?

We only parse error bodies if `RequestSSE` was used (`OnDataReceived` is only called if `DownloadAsStream` is used, and `OnDataReceived` is the only caller of `MaybeParseErrorBody`) and if the response content type is `application/json` rather than `text/event-stream`. If these conditions are not met, we return an empty value as usual.

> Could the error body itself be chunked over multiple chunks?

Good question. I don't think it would, but I can't guarantee it. I'll modify to assume that it can be chunked.

petemill:
These numbers will be pretty meaningless to users. Is the overall aim to eventually make string representations for some of the error types that would be helpful for the user to know about?

If it's simply to make bug reports, don't we already track the counts of each type of error occurring in the server analytics?

I can see _some_ utility when responding to users' individual requests for why something errors.

components/ai_chat/core/common/mojom/common.mojom
DJAndries: Discussed over DM. we agreed to add the error type to the struct, but as a follow-up. I'll leave a TODO here and log an issue for now

components/ai_chat/resources/untrusted_conversation_frame/components/conversation/index.tsx
```diff
@@ -104,6 +104,7 @@ function Conversation(props: ConversationProps) {
         currentErrorElement = (
           <ErrorConnection
             onRetry={() => context.conversationHandler.retryAPIRequest()}
+            errorDetails={state.currentErrorDetails}
```
petemill, line 107: Please expose in storybook either as a simple boolean that loads a constant `currentErrorDetails` or a write-in for each property

components/ai_chat/core/common/mojom/common.mojom
petemill: This was my initial reaction too. I think you're saved by the ConversationHandler struct which avoids needing to keep 2 properties in sync, so ok...

DJAndries:
> These numbers will be pretty meaningless to users. Is the overall aim to eventually make string representations for some of the error types that would be helpful for the user to know about?
> 
> If it's simply to make bug reports, don't we already track the counts of each type of error occurring in the server analytics?
> 
> I can see _some_ utility when responding to users' individual requests for why something errors.

It's primarily for bug reports. We do have server analytics for each type of error, but we don't necessarily have the steps to reproduce the error. Having the STR and error codes included in a bug report could make things easier to investigate.

components/ai_chat/core/common/mojom/common.mojom
DJAndries: added TODO, and raised https://github.com/brave/brave-browser/issues/55518

components/api_request_helper/api_request_helper.cc
DJAndries: updated to parse body on completion

components/ai_chat/resources/untrusted_conversation_frame/components/conversation/index.tsx
DJAndries: updated

components/api_request_helper/api_request_helper.cc
```diff
@@ -497,6 +515,34 @@ void APIRequestHelper::URLLoaderHandler::OnParseJsonResponse(
   std::move(result_callback_).Run(std::move(result));
 }
 
+bool APIRequestHelper::URLLoaderHandler::MaybeAppendToErrorBodyBuffer(
+    std::string_view string_piece) {
+  if (!is_response_fail_and_json_.has_value() || !*is_response_fail_and_json_) {
```
petemill: ultra nit: `if (is_response_fail_and_json.value_or(false)) {`

petemill (approved):
Nice work

emerick (approved):
chromium_src++

components/api_request_helper/api_request_helper.h
DJAndries: resolving. please lmk if you have further objections and i can follow up

components/ai_chat/core/common/mojom/ai_chat.mojom
DJAndries: resolved via https://github.com/brave/brave-core/pull/36391#discussion_r3238231357

components/api_request_helper/api_request_helper.cc
DJAndries: updated

fallaciousreasoning (approved):

kjozwiak:

```
Brave | 1.93.6 Chromium: 149.0.7827.23 (Official Build) nightly (64-bit)
-- | --
Revision | fcbe4adb3a13282b4016a0b28fe7ed4c9163a7b2
```

Using the STR/Cases outlined via https://github.com/brave/brave-core/pull/36391#issue-4434418768, used `Proxyman` to create a `Breakpoint` to capture the `REQUEST` from `https://ai-chat.bsg.brave.com/v1/chat/completions` and edit the `JSON` so it's an invalid request and ensured that `Status code: 401, error code: 40104` was displayed as per the following:

Screenshot 2026-05-29 000115

https://github.com/user-attachments/assets/c7bf2d5a-3271-449e-bb2b-1dd07d9938ed

PR #36392, opened 2026-05-13
Account: cancel in-flight network requests on account state transitions

### LINKED ISSUE #55264: Account: cancel in-flight network requests on account state transitions
When `BraveAccountService` transitions between `LoggedOutState` and `LoggedInState`, any in-flight network request that's only valid for the source state should be cancelled.

Currently, responses to in-flight requests can fire after the service has transitioned to a new state. This is benign today in that users won't end up in a structurally invalid state, but they can still end up in a valid-but-unintended one. For example, if the user clicks `Cancel registration` while `VerifyComplete` is in flight, the response may still arrive and `OnRegisterVerify()` can promote them to logged-in.

Other paths are already protected today: for example, `AuthValidate` is cancelled because `OnAccountStateChanged()` stops `auth_validate_timer_`, which owns the request handle, and `OnGetServiceToken()` has an explicit `kAuthenticationSessionChanged` guard.

That said, these invariants are easy to get wrong when maintained manually. State-owned cancellation would make the protection structural and uniform.

szilardszaloki:
Resolves https://github.com/brave/brave-browser/issues/55264.

- split `BraveAccountService` into typed account-state classes (`LoggedOutState`/`LoggedInState`) over a shared `StateBase`. Each state owns its in-flight `RequestHandle`s in an `in_flight_` list, so state transitions destroy the previous state, which cancels its requests — response callbacks for the prior state can no longer fire against the new one, therefore per-callback *Did the state change?* checks become unnecessary.
- make `StateBase` the `mojom::Authentication` impl. Each state's `mojo::ReceiverSet<>` migrates across transitions via `TakeReceivers()`/`AddReceiver()`. Wrong-state calls (race between renderer and state transition) land in `StateBase`'s default response and return a new `kCalledInWrongState` client error. The `BraveAccountService` no longer implements `mojom::Authentication` — it orchestrates transitions and owns the observer set.
- tighten `AccountStatePrefs::GetAccountState()` into a pref-shape firewall: `CHECK` the same invariants the setters enforce, so downstream consumers can trust the typed variant without re-validating pref shape. Callers `CHECK` the pref invariant locally before `Decrypt()`.
- removed `GetServiceTokenClientErrorCode::{kAuthenticationSessionChanged, kUserNotLoggedIn}`, 
  `RegisterClientErrorCode::kNoRegistrationInProgress`, `ResendConfirmationEmailClientErrorCode::kNoRegistrationInProgress` — all now unreachable by construction.


components/brave_account/brave_account_service_test.h
```diff
@@ -99,8 +99,10 @@ class BraveAccountServiceTest : public testing::TestWithParam<const TestCase*> {
                     future.GetCallback());
       EXPECT_EQ(future.Take(), test_case.mojo_expected);
     } else if constexpr (std::is_same_v<TestCase, AuthValidateTestCase>) {
-      TestCase::Run(test_case, pref_service_, task_environment_,
-                    *auth_validate_timer_);
+      TestCase::Run(
+          test_case, pref_service_, task_environment_,
+          base::BindRepeating(&BraveAccountService::AuthValidateTimerForTesting,
+                              base::Unretained(brave_account_service_.get())));
```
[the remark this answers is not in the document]
szilardszaloki: Closing as outdated.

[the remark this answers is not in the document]
szilardszaloki: Closing as outdated.

components/brave_account/logged_in_state.cc
```diff
@@ -0,0 +1,214 @@
...
+                     weak_factory_.GetWeakPtr(), std::move(service_name),
+                     std::move(callback)));
+}
+
+void LoggedInState::ScheduleAuthValidate(
+    base::TimeDelta delay,
+    RequestHandle current_auth_validate_request) {
+  auth_validate_timer_.Start(
+      FROM_HERE, delay,
+      base::BindOnce(&LoggedInState::AuthValidate, base::Unretained(this),
```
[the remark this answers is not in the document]
szilardszaloki: /fp `auth_validate_timer_` is a `LoggedInState` member and its pending task is dropped on destruction, so using `base::Unretained<>()` here is safe.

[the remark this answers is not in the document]
szilardszaloki: False positive — `auth_validate_timer_` is a `LoggedInState` member and its pending task is dropped on destruction, so using `base::Unretained<>()` here is safe.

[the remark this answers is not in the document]
goodov: that's `base::*Timer`, so it's acceptable.

components/brave_account/brave_account_service.cc
```diff
@@ -204,651 +66,64 @@ void BraveAccountService::AddObserver(
...
+    auto receivers = std::visit(
+        absl::Overload{[](StateBase& state) { return state.TakeReceivers(); },
+                       [&](std::monostate) {
+                         return std::exchange(pending_receivers_, {});
+                       }},
+        state_);
+    state_.emplace<State>(account_state_prefs_, url_loader_factory_,
+                          *encryptor_,
+                          base::BindRepeating(&BraveAccountService::AddObserver,
+                                              base::Unretained(this)));
```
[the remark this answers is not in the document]
szilardszaloki: That's safe: the callback is stored on `state_`, which is a member of `this`, so it cannot outlive `this`.

netzenbot (approved):

components/brave_account/encryption.h
```diff
@@ -0,0 +1,30 @@
...
+using OSCryptCallback =
+    base::RepeatingCallback<bool(const std::string&, std::string*)>;
+
+void SetOSCryptCallbacksForTesting(OSCryptCallback encrypt_callback,
+                                   OSCryptCallback decrypt_callback);
+
+std::string Encrypt(const os_crypt_async::Encryptor& encryptor,
+                    const std::string& plain_text);
+
+std::string Decrypt(const os_crypt_async::Encryptor& encryptor,
```
DJAndries: i'm happy to see this being split into another file. i think i've commented on this before, but it really seems like we could create a class that holds a raw_ref and hosts some methods

DJAndries (commented):
To be honest, I'm a little unsure about this approach, where you have two classes that implement the mojo interface, pass the receivers between them and swap between them at run time. The class hierarchy is also a little unorthodox. There's a lot of cleverness going on here, enough to cause concern. Are there any other examples in b-c where we do something similar?

IMHO, this solution (which seems like a partial architecture overhaul) seems overkill for the issue as it's described. Couldn't we take a few simple approaches to address the issue?

i.e.

- Disable UI action elements (i.e. "cancel registration" buttons, etc.) while an operation is in progress
- Have the service know when an operation is in progress (i.e. a bool class member), and return an error if so?

That being said, it is nice to see some of the existing code being distributed across multiple files, like the encryption stuff for instance. Regardless of the path you choose to take, I hope we can keep progressing in that regard.

szilardszaloki:
Thanks for the careful read @DJAndries.

On the suggested approaches, I don't think either quite gets at the actual issue here.

Disabling UI during an operation helps with one class of user-driven races, but it doesn't do anything for races caused by background work changing state. It is also only a best-effort frontend guard, and it doesn't protect us from buggy frontend implementations.

A service-side "operation in progress" `bool` is basically the same kind of per-callback *Did the state change?* check this PR is trying to get rid of. We already had a few of those, like `kAuthenticationSessionChanged`. They are all one-offs, and a single `bool` doesn't really cover the problem space. The issue is not "only allow one operation at a time", it's that the service state and its in-flight network requests need to move together as one atomic thing. The old code treated them as separate concerns and kept them in sync by hand, which made the code hard to reason about.

On the design I landed on: yes, `StateBase` implements the Mojom interface, and the two states inherit from it. Splitting an implementation across an inheritance hierarchy where the base provides defaults is not unusual. `ReceiverSet::TakeReceivers()` also exists exactly for migration-style cases like this, it's not a custom mechanism.

For context, these are the alternatives I tried before landing here (which is also why I think the current shape is the right tradeoff):

1. Per-method `std::visit<>()` in the service, with wrong-state error branches. This works, but leaves us with ten near-identical dispatch methods on the service, and every new Mojom method needs another copy. Wrong-state handling ends up duplicated everywhere.

2. Per-method `std::get_if<>()` + `CHECK_DEREF()`, which is effectively what I had in the draft. That gives us renderer-reachable browser crashes on races, which is not great.

3. A `DispatchTo<>()` `template` helper in the service that infers the state from the member pointer and centralizes dispatch. This works, but gets pretty ugly pretty quickly because of the TMP needed to recover the owning class from the member pointer.

4. Keep receivers on the service and forward through a `current_impl_` pointer to the active state. This avoids receiver migration, but now the service has ten one-line `current_impl_->Method(...)` forwards. Same problem as before, just moved around. It also leaves us with four classes implementing the interface: `StateBase`, the two states, and the service as a forwarder. Worst of all worlds.

5. Split the Mojom interface into per-state interfaces, like `LoggedOutAuthentication` and `LoggedInAuthentication`. This is yet another type-safe option, but it turns into a renderer-side protocol change with lots of ripples: rebinding on transitions, handling the window where neither interface is bound, and rewrites across the Accounts code on all platforms.

6. The current version: `StateBase` implements the interface, states override only what's valid, and receivers migrate to the active state. Dispatch is zero-cost because Mojom binds directly to the active state. Wrong-state responses live in one place per method, in the base. The override list makes it clear which operations are valid in which state. The tradeoff is receiver migration, but that's a single small block in `EnsureState()` using `TakeReceivers()`/`AddReceiver()`.

On the "partial architecture overhaul" point: I think that's a fair characterization. But the scope is bounded. This is contained to `BraveAccountService` and its two state classes. Callers outside the component don't change, for example, Settings still calls `BindInterface()` the same way.

components/brave_account/encryption.h
szilardszaloki: Yep, happy to look into that as a follow-up. 👍🏻

szilardszaloki: Actually, I ended up taking care of it here: https://github.com/brave/brave-core/pull/36392/commits/3abb57656f9ec166672afad227da51e2b73fbed5.

DJAndries:
> Disabling UI during an operation helps with one class of user-driven races, but it doesn't do anything for races caused by background work changing state. It is also only a best-effort frontend guard, and it doesn't protect us from buggy frontend implementations.

To clarify, I was suggesting combining this with a solution that ensures the backend only performs one operation at a time. The frontend modification would only affect the UX, to let them know that they cannot perform the operation just yet.

> A service-side "operation in progress" `bool` is basically the same kind of per-callback _Did the state change?_ check this PR is trying to get rid of. We already had a few of those, like `kAuthenticationSessionChanged`. They are all one-offs, and a single `bool` doesn't really cover the problem space.

So we couldn't in theory have some sort of "in progress" method on the endpoint client that returns true if a network request is pending, and return an error on any async operation call if it did return true? If not, we couldn't take a similar approach via other means?

> On the design I landed on: yes, `StateBase` implements the Mojom interface, and the two states inherit from it. Splitting an implementation across an inheritance hierarchy where the base provides defaults is not unusual.

Could you please provide an example that utilizes Mojo to aid my understanding? Perhaps I just need to get familiar with the idea.

> `ReceiverSet::TakeReceivers()` also exists exactly for migration-style cases like this, it's not a custom mechanism.

The only examples of `TakeReceivers` calls that I could find were from `NavigationRequest`, in order to pass the existing receivers around as navigations are committed, rather than passing receivers from one derived class to another.

szilardszaloki:
> So we couldn't in theory have some sort of "in progress" method on the endpoint client that returns true if a network request is pending, and return an error on any async operation call if it did return true? If not, we couldn't take a similar approach via other means?

Probably we could do that, but it would solve a different problem.

The issue this PR is trying to guard against is a response from a prior state landing after we've moved to a new state. If we could prove that all state transitions are caused only by operations gated by that same in-progress flag, then maybe that would avoid some of these races. But I don't think that's an invariant we should rely on.

The safer invariant is: if a state goes away, callbacks owned by that state cannot run anymore — and I want that guarantee to hold by construction.

So basically, "don't issue request B while request A is in flight" is not the same as "don't process request A's response if the state changed underneath it".

It also feels like a layering smell to put this on the endpoint client. The endpoint client doesn't know what "an operation" means at the user-flow level. For example, `AuthValidate` polling and `VerifyDelete` fire-and-forget are both network requests, but neither should block other operations. Those semantic distinctions don't belong on the endpoint client.

> Could you please provide an example that utilizes Mojo to aid my understanding? Perhaps I just need to get familiar with the idea.

This is really just ordinary C++ inheritance — the fact that `mojom::Authentication` happens to be Mojom-generated doesn't materially change anything. A pure-virtual interface can have an intermediate base class that provides default implementations, with concrete subclasses overriding only the methods relevant to them. There's nothing particularly Mojom-specific about the pattern itself.

> The only examples of TakeReceivers calls that I could find were from NavigationRequest, in order to pass the existing receivers around as navigations are committed, rather than passing receivers from one derived class to another.

The `TakeReceivers()` usage you found for migrating receivers across the navigation-commit boundary is actually quite similar to how this PR uses it across state-transition boundaries. In both cases, receivers belong to a particular logical phase/state and are transferred as ownership moves forward. The fact that there aren't many examples of it doesn't make it the wrong tool here.

components/brave_account/brave_account_encryption.h
```diff
@@ -0,0 +1,42 @@
...
+
+#include <string>
+
+#include "base/functional/callback.h"
+#include "base/memory/raw_ref.h"
+#include "components/os_crypt/async/common/encryptor.h"
+
+namespace brave_account {
+
+class BraveAccountEncryption {
```
bsclifton, line 22: This could use some comments describing what it's used for

components/brave_account/logged_in_state.h
```diff
@@ -0,0 +1,73 @@
...
+#include "brave/components/brave_account/state_base.h"
+#include "components/os_crypt/async/common/encryptor.h"
+
+namespace network {
+class SharedURLLoaderFactory;
+}  // namespace network
+
+namespace brave_account {
+
+class LoggedInState : public StateBase {
```
bsclifton, line 34: Can use a description for the state

components/brave_account/logged_out_state.h
```diff
@@ -0,0 +1,94 @@
...
+#include "brave/components/brave_account/state_base.h"
+#include "components/os_crypt/async/common/encryptor.h"
+
+namespace network {
+class SharedURLLoaderFactory;
+}  // namespace network
+
+namespace brave_account {
+
+class LoggedOutState : public StateBase {
```
bsclifton, line 35: Can use a description for the state

components/brave_account/brave_account_encryption.h
szilardszaloki: Added in https://github.com/brave/brave-core/pull/36392/commits/e245dea5fca8f36c694aebc1e89740daaf3b69ab.

components/brave_account/logged_in_state.h
szilardszaloki: Added in https://github.com/brave/brave-core/pull/36392/commits/e245dea5fca8f36c694aebc1e89740daaf3b69ab.

components/brave_account/logged_out_state.h
szilardszaloki: Added in https://github.com/brave/brave-core/pull/36392/commits/e245dea5fca8f36c694aebc1e89740daaf3b69ab.

components/brave_account/state_internal.h
```diff
@@ -0,0 +1,115 @@
...
+
+namespace brave_account::internal {
+
+inline constexpr net::NetworkTrafficAnnotationTag kTrafficAnnotation =
+    net::DefineNetworkTrafficAnnotation("brave_account_endpoints",
+                                        R"(
+  semantics {
+    sender: "Brave Account client"
+    description:
+      "Implements the creation or sign-in process for a Brave Account."
```
bsclifton, line 26: We might need to update this when doing password reset/change if the same annotation is re-used.

components/brave_account/state_base.cc
```diff
@@ -0,0 +1,108 @@
...
+
+using internal::MakeCalledInWrongStateError;
+
+void StateBase::AddReceiver(
+    mojo::PendingReceiver<mojom::Authentication> receiver) {
+  receivers_.Add(this, std::move(receiver));
+}
+
+std::vector<mojo::PendingReceiver<mojom::Authentication>>
+StateBase::TakeReceivers() {
```
bsclifton, line 24: Since this is our method, we could potentially call it `UnbindAndTakeReceivers` to make it a bit more obvious. I'm OK with it as-is though, because you can look upstream and see the unbinding happens.

bsclifton: The comment in state_base.h does already mention the unbinding 👍 Missed that. Up to you 😄

components/brave_account/brave_account_service.cc
```diff
@@ -167,34 +35,28 @@ BraveAccountService::BraveAccountService(
...
-  DecryptCallbackForTesting() = std::move(decrypt_callback);  // IN-TEST
-}
-
 void BraveAccountService::BindInterface(
     mojo::PendingReceiver<mojom::Authentication> pending_receiver) {
-  if (!encryptor_) {
-    return pending_receivers_.push_back(std::move(pending_receiver));
-  }
-  authentication_receivers_.Add(this, std::move(pending_receiver));
+  std::holds_alternative<std::monostate>(state_)
```
bsclifton: This could be (at a glance) hard to read - curious if it makes sense to abstract out to a bool. Like:
```
bool no_state_set = std::holds_alternative<std::monostate>(state_);
if (no_state_set) {
```

If the naming was different in `std` library, it might be obvious 😂 Like
`if (contains<emptystate>(state_))`

But if the person reading already knows about `std::variant` it's a non-issue though.

components/brave_account/brave_account_service.cc
```diff
@@ -204,651 +66,64 @@ void BraveAccountService::AddObserver(
...
+    for (auto& receiver : receivers) {
+      ActiveState().AddReceiver(std::move(receiver));
     }
-  }
-
-  return plain_text;
+  };
+  // NOLINTEND(readability/braces)
+
+  switch (which) {
```
bsclifton, line 107: Maybe a comment here sharing that the above templated lambda is executed based on the mojom account state. In case the reader didn't realize the above is a lamda. Or maybe explicitly name it `ensure_state_lamda`.

bsclifton (approved):
Overall, LGTM. I know you're working on password reset/change states too. It would be great to have some tests covering some of the logic (if possible) for logic in StateBase.

components/brave_account/brave_account_service.cc
szilardszaloki: Yeah, that makes sense — how about this?
```cpp
const bool initializing = std::holds_alternative<std::monostate>(state_);
initializing ? pending_receivers_.push_back(std::move(pending_receiver))
             : ActiveState().AddReceiver(std::move(pending_receiver));
```

components/brave_account/state_internal.h
szilardszaloki: Good point — I'll update the fields so they cover password reset/change too when wiring them up.

components/brave_account/brave_account_service.cc
szilardszaloki: Amended in https://github.com/brave/brave-core/pull/36392/commits/a7413da6294878fa5fe1508293be686d21fd1fc5.

components/brave_account/brave_account_service.cc
szilardszaloki: Amended in https://github.com/brave/brave-core/pull/36392/commits/a7413da6294878fa5fe1508293be686d21fd1fc5.

szilardszaloki:
> Overall, LGTM. I know you're working on password reset/change states too. It would be great to have some tests covering some of the logic (if possible) for logic in StateBase.

Good idea. Most `StateBase` entry points just forward (e.g. `SendCallerOwnedRequest<>()` and `SendUnownedRequest<>()` are one-line forwards to `endpoint_client::Client<>::Send<>()`), but I added tests in `state_base_unittest.cc` covering `SendStateOwnedRequest<>()`. 👍🏻

bsclifton:
This was set for auto merge but S3 upload failed for macOS ARM64 (all prior steps passed).
Will give this an admin merge

PR #36393, opened 2026-05-13
Remove padding around container page action icon and text

### LINKED ISSUE #55407: [Containers] Polish PageActionView for container
Check that the left-right side padding of the container bubble matches design. Implementation seems to have much more space to the left-right sides.

Check that the icon size of the bubble matches design spec.

Image

sangwoo108:
We don't need theses padding on the chip for partitioned storage page action.

Resolves https://github.com/brave/brave-browser/issues/55407

### DIFF
--- a/chromium_src/chrome/browser/ui/page_actions/page_action_model.cc
+++ b/chromium_src/chrome/browser/ui/page_actions/page_action_model.cc
@@ -69,7 +69,7 @@ std::optional<int> PageActionModel::GetOverrideTriggerableEvent() const {
   return override_triggerable_event_flags_;
 }
 
-void PageActionModel::SetOverrideBorder(base::PassKey<PageActionController>,
+void PageActionModel::SetOverrideBorder(PageActionPassKey,
                                         std::optional<gfx::Insets> border) {
   if (override_border_ == border) {
     return;
--- a/chromium_src/chrome/browser/ui/page_actions/page_action_model.cc
+++ b/chromium_src/chrome/browser/ui/page_actions/page_action_model.cc
@@ -69,4 +69,18 @@ std::optional<int> PageActionModel::GetOverrideTriggerableEvent() const {
   return override_triggerable_event_flags_;
 }
 
+void PageActionModel::SetOverrideBorder(
+    base::PassKey<PageActionController>,
+    const std::optional<gfx::Insets>& border) {
+  if (override_border_ == border) {
+    return;
+  }
+  override_border_ = border;
+  NotifyChange(Property::kOverrideBorder);
+}
+
+std::optional<gfx::Insets> PageActionModel::GetOverrideBorder() const {
+  return override_border_;
+}
+
 }  // namespace page_actions
--- a/chromium_src/chrome/browser/ui/page_actions/page_action_model.cc
+++ b/chromium_src/chrome/browser/ui/page_actions/page_action_model.cc
@@ -69,9 +69,8 @@ std::optional<int> PageActionModel::GetOverrideTriggerableEvent() const {
   return override_triggerable_event_flags_;
 }
 
-void PageActionModel::SetOverrideBorder(
-    base::PassKey<PageActionController>,
-    const std::optional<gfx::Insets>& border) {
+void PageActionModel::SetOverrideBorder(base::PassKey<PageActionController>,
+                                        std::optional<gfx::Insets> border) {
   if (override_border_ == border) {
     return;
   }
--- a/chromium_src/chrome/browser/ui/views/page_action/page_action_view.cc
+++ b/chromium_src/chrome/browser/ui/views/page_action/page_action_view.cc
@@ -15,16 +15,30 @@
 #include "ui/views/style/platform_style.h"
 #include "ui/views/view_class_properties.h"
 
+namespace {
+
+void MaybeOverrideBorder(const page_actions::PageActionModelInterface* source,
+                         gfx::Insets& border_insets) {
+  if (source && source->GetOverrideBorder().has_value()) {
+    border_insets = source->GetOverrideBorder().value();
+  }
+}
+
+}  // namespace
+
 #define GetMinimumSize GetMinimumSize_Chromium
 #define OnNewActiveController OnNewActiveController_Chromium
 #define OnPageActionModelChanged OnPageActionModelChanged_Chromium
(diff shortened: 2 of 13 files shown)


netzenbot (approved):

chromium_src/chrome/browser/ui/page_actions/page_action_model.cc
```diff
@@ -70,4 +70,18 @@ std::optional<int> PageActionModel::GetOverrideTriggerableEvent() const {
   return override_triggerable_event_flags_;
 }
 
+void PageActionModel::SetOverrideBorder(
+    base::PassKey<PageActionController>,
+    const std::optional<gfx::Insets>& border) {
+  if (override_border_ == border) {
+    return;
+  }
```
cdesouza-chromium, line 76: Would it be possible to improve the semantics for this?

`override_border_` is an optional, but the optional has to be engaged for it to be assignable, so before the optional has a value assigned first, all calls to `SetOverrideBorder` are ignored. This makes the behaviour of `SetOverrideBorder` a bit confusing, as it only "sets" an overriding value, if a previous one exist. I would prefer if this could be clearer.

sangwoo108: Do you mean that it'd be preferred having PageActionModel::SetOverrideBorder(gfx::Insets) and PageActionModel::ClearOverrideBorder() for clarity?

cdesouza-chromium: Sorry my comment was wrong above. `SetOverrideBorder` is implemented correct. So forget what I said above.

I don't see the value in `ClearOverrideBorder()` if that's the same as `SetOverrideBorder(std::nullop)`. Can we do with just one function?

One last thing as well:

```cxx
void PageActionModel::SetOverrideBorder(
    base::PassKey<PageActionController>,
    std::optional<gfx::Insets> border) 
```

Avoid using `const optional<T>&`, as this is very misleading, and it is not actually doing what you think it is doing in a lot of cases.

chromium_src/chrome/browser/ui/views/page_action/page_action_view.cc
```diff
@@ -18,11 +18,13 @@
 #define GetMinimumSize GetMinimumSize_Chromium
 #define OnNewActiveController OnNewActiveController_Chromium
 #define OnPageActionModelChanged OnPageActionModelChanged_Chromium
+#define UpdateBorder UpdateBorder_Chromium
```
cdesouza-chromium: `UpdateBorder` is called in `chrome/browser/ui/views/page_action/page_action_view.cc` at least twice, so those callsites are also getting renamed. Is this intentional?

sangwoo108: Yes, at the moment calling UpdateBorder_Chromium() in the constructor isn't that bad as the border will be replaced by following call of UpdateBorder() from the controller.

cdesouza-chromium: There are actually two calls to `UpdateBorder`: one in the constructor, and another in `UpdateIconImage`, so this is a bit confusing. On top of that, I don't think it is a good idea to virtualise functions that get called in a constructor, because can confuse people entirely, even if this was a plaster. Is there some alternative way to solve this customisation?

chromium_src/chrome/browser/ui/page_actions/page_action_model.cc
sangwoo108: Yes, one function will do. 

Removed const & from the paramter.

chromium_src/chrome/browser/ui/views/page_action/page_action_view.cc
sangwoo108: Good point, let me think about it a little bit more.

chromium_src/chrome/browser/ui/page_actions/page_action_model.cc
sangwoo108:
```
void PageActionView::UpdateBorder() {
  gfx::Insets border_insets = icon_insets_;
  if (observation_.IsObserving() &&
      !observation_.GetSource()->GetImage().IsVectorIcon()) {
    border_insets = GetInsetsForNonVectorIcon();
  }
  SetBorder(views::CreateEmptyBorder(border_insets));
}
```

This is upstream implementation. How about adding BRAVE_PAGE_ACTION_VIEW_UPDATE_BORDER ?

```diff
void PageActionView::UpdateBorder() {
  gfx::Insets border_insets = icon_insets_;
  if (observation_.IsObserving() &&
      !observation_.GetSource()->GetImage().IsVectorIcon()) {
    border_insets = GetInsetsForNonVectorIcon();
  }
+ BRAVE_PAGE_ACTION_VIEW_UPDATE_BORDER  // Get override border from model
  SetBorder(views::CreateEmptyBorder(border_insets));
}
```

cdesouza-chromium: You replied here to the wrong thread, but yeah, it looks like we should do that.

sangwoo108: Oops, sorry for the wrong anchor 😭

rewrite/chrome/browser/ui/views/page_action/page_action_view.cc.toml
```diff
@@ -7,3 +7,12 @@
...
 re_pattern = 'else if \((model.ShouldShowSuggestionChip\(\))\)'
 replace = 'else if (\1 || model.GetAlwaysShowLabel())'
+
+[[substitution]]
+description = '''
+Adding BRAVE_PAGE_ACTION_VIEW_UPDATE_BORDER in UpdateBorder() right before
+SetBorder() call in it.
+'''
+re_pattern = '(void PageActionView::UpdateBorder\(\)[\s\S]*?)(SetBorder\()'
+replace = '''\1BRAVE_PAGE_ACTION_VIEW_UPDATE_BORDER
```
sangwoo108: @cdesouza-chromium , would it be better to insert MaybeOverrideBorder() directly as we already have plaster for this file?

cdesouza-chromium (approved):
LGTM

PR #36395, opened 2026-05-13
[infobar] BraveConfirmInfoBar (3/N): drop InfoBarView base class

simonhong:
BraveConfirmInfoBar used to inherit from InfoBarView, so its layout
changed with upstream's features::kInfobarRefresh — vertical centering,
label/link gaps, icon offset and multi-line height all shifted between
the flag-off and flag-on states.

As we're using BraveConfirmInfoBar for our specific info bar only,
we don't need to get suffer from upstream's view change over time.
We can go with our own custom view based on `infobars::InfoBar` interface.

Replace the base class with direct multi-inheritance from
infobars::InfoBar, views::View and views::ExternalFocusTracker, and
implement Layout / CalculatePreferredSize / OnThemeChanged /
PlatformSpecific* ourselves. The visual is now identical regardless of
the upstream flag and we no longer depend on InfoBarView's internal
structure.

Resolves - subtask of https://github.com/brave/brave-browser/issues/48822

No regressions on existing infobars that uses BraveConfirmInfoBar.
infobar with multi-line:
image

infobar with checkbox:
image

### DIFF
--- a/browser/ui/views/infobars/brave_confirm_infobar.cc
+++ b/browser/ui/views/infobars/brave_confirm_infobar.cc
@@ -11,6 +11,7 @@
 
 #include "base/check.h"
 #include "base/functional/bind.h"
+#include "base/memory/raw_ref.h"
 #include "chrome/browser/ui/color/chrome_color_id.h"
 #include "chrome/browser/ui/views/chrome_layout_provider.h"
 #include "chrome/grit/generated_resources.h"
@@ -39,6 +40,7 @@
 #include "ui/views/controls/image_view.h"
 #include "ui/views/controls/label.h"
 #include "ui/views/controls/link.h"
+#include "ui/views/focus/external_focus_tracker.h"
 #include "ui/views/style/typography.h"
 #include "ui/views/view_class_properties.h"
 #include "ui/views/widget/widget.h"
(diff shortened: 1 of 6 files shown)


browser/ui/views/infobars/brave_confirm_infobar.cc
```diff
@@ -88,52 +138,97 @@ BraveConfirmInfoBar::BraveConfirmInfoBar(
...
-  // children. However, we just add children directly instead of
-  // that container.
-  // TODO(https://github.com/brave/brave-browser/issues/48822): Add our children
-  // into content container.
-  if (close_button_) {
-    ReorderChildView(close_button_, children().size());
+  // Close button (always last for accessibility focus order).
+  if (delegate_ptr->IsCloseable()) {
+    auto close_button = views::CreateVectorImageButton(base::BindRepeating(
+        &BraveConfirmInfoBar::CloseButtonPressed, base::Unretained(this)));
```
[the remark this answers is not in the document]
simonhong: `close_button` is owned by BraveConfirmInfoBar.

browser/ui/views/infobars/brave_confirm_infobar.cc
```diff
@@ -164,84 +258,192 @@ void BraveConfirmInfoBar::Layout(PassKey) {
...
+  return label;
+}
+
+std::unique_ptr<views::Link> BraveConfirmInfoBar::CreateLink(
+    const std::u16string& text) {
+  auto link = std::make_unique<views::Link>(
+      text, views::style::CONTEXT_DIALOG_BODY_TEXT);
+  SetLabelDetails(link.get());
+  link->SetCallback(base::BindRepeating(&BraveConfirmInfoBar::LinkClicked,
+                                        base::Unretained(this)));
```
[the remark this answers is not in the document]
simonhong: link is owned by BraveConfirmInfoBar.

browser/ui/views/infobars/brave_confirm_infobar.h
```diff
@@ -7,26 +7,50 @@
...
+// checkbox, and an optional right-aligned link. Commonly used for:
+//   "Would you like to do X?  [Yes]  [No]  [_Learn More_] [x]"
+//
+// Unlike upstream's ConfirmInfoBar/InfoBarView, this class is self-contained:
+// it inherits directly from infobars::InfoBar + views::View instead of from
+// InfoBarView, so its layout doesn't change with features::kInfobarRefresh
+// and stays insulated from future upstream restructuring of InfoBarView.
+class BraveConfirmInfoBar : public infobars::InfoBar,
+                            public views::View,
+                            public views::ExternalFocusTracker {
```
sangwoo108: I think we can have another class(maybe inner class?) instead of multiple inheritance. I believe composition would be better.

browser/ui/views/infobars/brave_confirm_infobar.cc
```diff
@@ -88,52 +138,97 @@ BraveConfirmInfoBar::BraveConfirmInfoBar(
...
+    const gfx::Insets close_button_spacing = GetCloseButtonSpacing();
+    close_button_->SizeToPreferredSize();
+    close_button_->SetPosition(gfx::Point(
+        std::max(
+            reserved_x + close_button_spacing.left(),
+            width() - close_button_spacing.right() - close_button_->width()),
+        OffsetY(close_button_)));
+
+    // For accessibility reasons, the close button should come last.
+    DCHECK_EQ(close_button_, children().back());
```
sangwoo108: I think this can be guarded with DCHECK_IS_ON()

browser/ui/views/infobars/brave_confirm_infobar.cc
```diff
@@ -154,7 +249,6 @@ void BraveConfirmInfoBar::Layout(PassKey) {
   place_button(ok_button_);
   place_button(cancel_button_);
 
-  // Place checkbox after latest button
```
sangwoo108, line 157: Wonder why this line was removed

browser/ui/views/infobars/brave_confirm_infobar.cc
simonhong: Deleted this check. It came from upstream `InfoBarView` but I think it's not necessary in production code. `BraveConfirmInfoBarTest.CloseButtonOrderTest` will check.

browser/ui/views/infobars/brave_confirm_infobar.cc
simonhong: oh, accidently removed. Reverted.

browser/ui/views/infobars/brave_confirm_infobar.h
simonhong: Updated. much better.

sangwoo108 (approved):
LGTM 👍

goodov (approved):

PR #36408, opened 2026-05-13
[Android] Apply Nala baseline color palette

### LINKED ISSUE #55497: [Android] Re-map upstream baseline colors to our Nala colors
### Description

To apply our Nala colors in upstream UI elements we should re-map upstream baseline colors to our Nala colors.

### Steps to reproduce

1. Upstream UI elements have orange accent

### Actual result

Upstream UI elements have orange accent

### Expected result

Upstream UI elements have current accent from Nala (blurple as of today)

### Reproduces how often

Easily reproduced

### Brave version

n/a

### Device

- Brand/model: n/a
- Android version: n/a

### Channel information

- [ ] release (stable)
- [ ] beta
- [ ] nightly

### Reproducibility

### Miscellaneous information

_No response_

samartnik:
Resolves https://github.com/brave/brave-browser/issues/55497

### DIFF
--- a/android/java/res/values-night/brave_colors.xml
+++ b/android/java/res/values-night/brave_colors.xml
@@ -5,14 +5,8 @@
      You can obtain one at https://mozilla.org/MPL/2.0/.
 -->
 <resources>
-    <!-- Theme colors -->
-    <!-- Overriding chromium theme colors to brave theme colors -->
-    <color name="baseline_primary_30">#F0F2FF</color>
-    <color name="baseline_primary_80">#FA7250</color>
-    <color name="baseline_primary_90">#5B2A23</color>
     <color name="default_text_color_on_accent1_baseline">@color/default_text_color_on_accent1_light</color>
     <color name="dialog_bg_color_dark_baseline">@color/default_bg_color_baseline</color>
-    <color name="modern_blue_600">@color/baseline_primary_30</color>
 
     <!-- Theme colors -->
     <color name="brave_link">#A0A5EB</color>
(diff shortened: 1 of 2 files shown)


samartnik:
Examples of new colors applied:
Screenshot_20260513_095359
Screenshot_20260513_095422
Screenshot_20260513_095432
Screenshot_20260513_095445
Screenshot_20260513_095452
Screenshot_20260513_095514

aguscruiz (approved):
Looks great!

SergeyZhukovsky (approved):
++

fallaciousreasoning (commented):
This seems like a good first step, but I suspect that we'll eventually need this to be a bit more complicated if we want to pull in dynamic themes from Material UI, as we'll need the dynamic theme to flow into the Nala theme (i.e. `@color/primitive_primary_95` should be set to `baseline_primary_95`, the opposite of what we're doing here).

Then, we should generate a default `baseline_primary` based on the default Nala theme, which we use when we're not taking dynamic colors from the material theme. This could be done on the @brave/leo side, if you can tell me what you need.

For example, on Desktop, we have this `brave_ref_color_mixer` which handles baseline vs dynamic themes:
ui/color/brave_ref_color_mixer.cc

aguscruiz:
To add to Jay's comment, yeah please let's do something that doesn't hardcode icons into the system. Icons change in the design system from time to time. We can't be pushing new code that hardcodes icons and/or colors that are not integrated dynamically into Nala.

samartnik:
For dynamic colors we should use `material_*` instead of `primitive_*` colors in code (it should be reflected in Figma). In this case it will use dynamic one if option is chosen or fall back to `primitive_*` otherwise.

fallaciousreasoning:
Awesome, thanks @samartnik !

PR #36409, opened 2026-05-13
[brockit][sync] Avoiding `bool` in `DeviceInfo`

### LINKED ISSUE #55018: Migrate `DeviceInfo` to a plaster

cdesouza-chromium:
The use of `bool` for `DeviceInfo` fields has led to bugs in the past.
This has been correct with plaster, however we should also convert the
field to an enum to make sure we follow the upstream practice of using
enums in this type.

Resolves https://github.com/brave/brave-browser/issues/55018


AlexeyBarabash (approved):
Thanks, lgtm, looks much robust than before

goodov (approved):

netzenbot (approved):

PR #36413, opened 2026-05-13
Implement Android browser-wide lockdown

### LINKED ISSUE #53000: [Android] Setting: Browser lock with pin/biometrics
**Spec:** search for `Android: Browser Lock` in drive.

Olipro:
## Description
Resolves brave/brave-browser#53000

This commit implements a browser-wide lock feature that operates much like the existing one for private tabs except it applies to the whole browser. It wholesale reuses the UI of the private tabs (with the exception of changing the button text to say "Unlock browser") so there is no (new) UI to look at.

I believe this represents a reasonable approach to implementation and could be merged as-is. Equally, someone might have a far better approach so I'll keep my powder dry and consider this a PoC.

Detailed information is included within the messages of the 2 commits in this PR. **Security review:** https://github.com/brave/reviews/issues/2249

### Locking and unlocking application
1. Launch the app and open Settings
2. Under `General`, find the one for `Browser lock`. Click this
    image
3. Enable `Entire application`
    image
4. You should get prompted to authenticate. This might be biometric (face ID, fingerprint sensor) or could be pin depending on your phone's config.
5. Close settings
6. Open a few tabs (example.com, brave.com)
7. Go back to the home screen, putting Brave into the background
8. Re-open Brave so that it takes focus again
9. You should be prompted to authenticate. It's important that no tabs are visible
10. After successful auth, verify your tabs are there

### Make sure lock works (dismiss auth)
1. Do the steps 1-8 for `Locking and unlocking application`
2. When you're prompted for authentication, you should be able to dismiss the authentication with an X button to close
3. When closed, you'll just see (brave logo) `brave` and `Browser locked` with an `Unlock browser` button.
4. There should be no way to see any tabs or use the app
5. For a last step, you could click `Unlock browser` button at the bottom and make sure you can unlock / see your tabs

### Verify you get prompted for disabling the browser lock
1. Follow steps 1-10 under `Locking and unlocking application`
2. Go back to Settings > General > Browser lock and disable `Entire application`
3. You should be prompted to authenticate. If you dismiss, you can't disable the lock.

### Private tabs (quick notes)

Folks have always been able to lock private tabs. This is a Chromium feature. Before, this was found under:
`Settings` > `Brave Shields & privacy` > `Other privacy settings` > `Lock Private tabs when you leave Brave`.

Now, you can find this under `Settings` > `General` > `Browser lock` > `Private tabs`.

No functionality has changed. Only where the setting is located at.

### Private tabs (double authentication)
1. Follow steps 1-4 from `Locking and unlocking application`
2. While on the `Browser lock` screen, also enable the private tabs feature
3. Close settings
4. Open a private tab to example.com and keep this private tab as the active tab
5. Go back to the home screen, putting Brave into the background
6. Re-open Brave so that it takes focus again
7. You should be prompted to authenticate twice. Once for browser lock, second for private tab.

### Verify screen capture is disabled by default
1, Go to `Settings` > `General` > `Browser lock`
2, Enable both `Entire application` and `Private tabs`. Authenticate if prompted.
3. `Prevent screenshot/video capture of locked tabs` should be enabled by default
4. Use the application (open new tabs, etc)
5. Take screenshots inside regular app and then also on a private tab
6. Check the screenshots and verify they are not a solid black image

### Verify screen capture can be enabled
1, Go to `Settings` > `General` > `Browser lock`
2, Enable both `Entire application` and `Private tabs`. Authenticate if prompted.
3. Disable `Prevent screenshot/video capture of locked tabs` setting
3. Use the application (open new tabs, etc)
5. Take screenshots inside regular app and then also on a private tab
6. Check the screenshots and verify they are not a solid black image

image


android/java/org/chromium/chrome/browser/incognito/reauth/BraveBrowserLockCoordinator.java
```diff
@@ -0,0 +1,84 @@
...
+                incognitoReauthCallback,
+                /* seeOtherTabsRunnable= */ () -> {});
+        mActivity = activity;
+        mIncognitoReauthManager = incognitoReauthManager;
+        mIncognitoReauthCallback = incognitoReauthCallback;
+    }
+
+    @Override
+    public void show() {
+        // fullscreen=false avoids the menu-delegate assertion and sets IS_SEE_OTHER_TABS_VISIBLE to
```
Olipro: This is perhaps a bit _too_ clever. I like it, but it does imply that we should have some automated monitoring (if we don't already) should Google decide to alter the behaviour of the UI and break this neat little trick.

android/java/org/chromium/chrome/browser/privacy/settings/BraveLockSettings.java
```diff
@@ -0,0 +1,93 @@
...
+        mReauthManager.startReauthenticationFlow(
+                new IncognitoReauthManager.IncognitoReauthCallback() {
+                    @Override
+                    public void onIncognitoReauthNotPossible() {
+                        revertToggle(previousValue);
+                    }
+
+                    @Override
+                    public void onIncognitoReauthSuccess() {
+                        ChromeSharedPreferences.getInstance()
```
Olipro: Is this correct? I worked on the basis that there's no reason to dump the setting into C++ land if it's entirely an Android thing.

android/java/org/chromium/chrome/browser/tabbed_mode/BraveTabbedRootUiCoordinator.java
```diff
@@ -208,6 +211,24 @@ public void onFinishNativeInitialization() {
         }
```
netzenbot, line 216: The browser-wide lockdown is a significant new security feature but has no `base::Feature` flag for remote disabling via Griffin. If a crash or UX regression is discovered in production, there's no way to remotely kill this without a hotfix release. Consider adding a `base::Feature` guard around the `BraveBrowserLockController` instantiation. ([Guard Significant or Experimental New Functionality Behind `base::Feature`](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-030))

android/java/brave-res/xml/brave_privacy_preferences.xml
```diff
@@ -146,6 +146,12 @@
             android:title="@string/private_tab_screenshot_title"
             android:summary="@string/private_tab_screenshot_summary"
             android:defaultValue="false" />
+        <org.chromium.components.browser_ui.settings.ChromeSwitchPreference
```
netzenbot: nit: Since `BraveLockSettings.setUp()` calls `mPreference.setVisible(false)` when the feature is unavailable, add `app:isPreferenceVisible="false"` here to avoid a brief flash of the preference before the code hides it. See the pattern used in `brave_leo_preferences.xml`. ([Use `app:isPreferenceVisible="false"` for Conditionally Shown Preferences](https://github.com/brave/brave-core/tree/master/docs/best-practices/android.md#AND-008))

android/java/org/chromium/chrome/browser/privacy/BraveBrowserLockController.java
```diff
@@ -0,0 +1,171 @@
...
+        return new BraveBrowserLockCoordinator(activity, mProfile, mReauthCallback);
+    }
+
+    private void hideCoordinatorIfShowing(@DialogDismissalCause int cause) {
+        if (mCoordinator != null) {
+            mCoordinator.hide(cause);
+            mCoordinator = null;
+        }
+    }
+}
```
netzenbot: nit: Per AND-025, test accessor annotations should use `@VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)` to communicate intent. This applies to `isLockPendingForTesting()`, `getReauthCallbackForTesting()`, and `createCoordinator()`. ([Use `@VisibleForTesting` for Package-Private Test Accessors](https://github.com/brave/brave-core/tree/master/docs/best-practices/android.md#AND-025))

android/java/org/chromium/chrome/browser/incognito/reauth/BraveBrowserLockCoordinator.java
```diff
@@ -0,0 +1,84 @@
...
+    @Override
+    public void hide(@DialogDismissalCause int dismissalCause) {
+        if (mLockView != null) {
+            ViewGroup rootView = (ViewGroup) mActivity.getWindow().getDecorView();
+            rootView.removeView(mLockView);
+            mLockView = null;
+            destroy();
+        }
+    }
+}
```
netzenbot, line 74: nit: Per [AND-042](https://github.com/brave/brave-core/tree/master/docs/best-practices/android.md#AND-042), `assert x != null` is intended for cases where you are *not* locally dereferencing the object. Since `mLockView` is immediately used on the next line, prefer `assumeNonNull(mLockView)` as a precondition statement instead.

browser/ui/android/strings/android_brave_strings.grd
```diff
@@ -629,6 +629,15 @@ This file contains all "about" strings.  It is set to NOT be translated, in tran
       <message name="IDS_CLOSE_TABS_ON_EXIT_SUMMARY" desc="Summary for closing tabs on exit.">
         Don't persist tabs between browsing sessions
       </message>
+      <message name="IDS_BRAVE_BROWSER_LOCK_TITLE" desc="Title for the browser biometrics lock setting.">
+        Lockdown Brave
+      </message>
+      <message name="IDS_BRAVE_BROWSER_LOCK_SUMMARY" desc="Summary for the browser biometrics lock setting.">
+        Require your screen lock every time you launch or switch back to Brave.
```
netzenbot: nit: Per [L10N-015](https://github.com/brave/brave-core/tree/master/docs/best-practices/localization.md#L10N-015), single-sentence settings toggle descriptions should not end with a period. Remove the trailing period from `IDS_BRAVE_BROWSER_LOCK_SUMMARY`.

android/BUILD.gn
```diff
@@ -82,6 +82,7 @@ android_library("qrreader_java") {
 
 # Brave-specific resources.
 android_resources("java_resources") {
+  resource_overlay = true
```
Olipro, line 85: Are we good with this? It struck me as a much better way of doing the override but perhaps there are other concerns I haven't thought of.

android/java/brave-res/layout/fragment_brave_browser_lock_settings.xml
```diff
@@ -0,0 +1,125 @@
...
+                android:layout_height="wrap_content"
+                android:orientation="vertical">
+
+                <!-- Entire application -->
+                <LinearLayout
+                    android:layout_width="match_parent"
+                    android:layout_height="wrap_content"
+                    android:orientation="horizontal"
+                    android:gravity="center_vertical"
+                    android:paddingStart="16dp"
```
netzenbot: nit: `paddingStart="16dp"` + `paddingEnd="16dp"` can be collapsed to `android:paddingHorizontal="16dp"` (same for the `layout_marginStart`/`layout_marginEnd` pairs on the headers). See [AND-038](https://github.com/brave/brave-core/tree/master/docs/best-practices/android.md#AND-038).

android/java/brave-res/layout/incognito_reauth_view.xml
```diff
@@ -0,0 +1,127 @@
...
+    </FrameLayout>
+
+    <!-- Centred content: Brave brand + status text. -->
+    <LinearLayout
+        android:layout_width="match_parent"
+        android:layout_height="0dp"
+        android:layout_weight="1"
+        android:orientation="vertical"
+        android:gravity="center"
+        android:paddingStart="16dp"
```
netzenbot: nit: `paddingStart="16dp"` + `paddingEnd="16dp"` here (and on the action-button container below) can be collapsed to `android:paddingHorizontal="16dp"`. See [AND-038](https://github.com/brave/brave-core/tree/master/docs/best-practices/android.md#AND-038).

android/brave_java_sources.gni
```diff
@@ -229,6 +229,7 @@ brave_java_sources = [
   "../../brave/android/java/org/chromium/chrome/browser/homepage/settings/BraveRadioButtonGroupHomepagePreference.java",
   "../../brave/android/java/org/chromium/chrome/browser/homepage/settings/BraveRadioButtonGroupHomepagePreferenceDummySuper.java",
   "../../brave/android/java/org/chromium/chrome/browser/identity_disc/BraveIdentityDiscController.java",
+  "../../brave/android/java/org/chromium/chrome/browser/incognito/reauth/BraveBrowserLockCoordinator.java",
```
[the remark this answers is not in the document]
AlexeyBarabash: Hello @Olipro ,

you can ty to use
https://github.com/brave/brave-core/pull/37653/changes/f023501c31305bf80d19e61f40fc020fd8d936c7
as an example
to move `BraveBrowserLockCoordinator.java` into a separate `BUILD.gn`.

android/BUILD.gn
AlexeyBarabash: @Olipro , which resource is at  duplicate resource conflict error without `resource_overlay = true`?

Olipro: None, this makes `android/java/brave-res/layout/incognito_reauth_view.xml` in this PR act as an override of Chromium's (which is also why this PR deletes the patch file)

Or to be more precise: This line is added so that the aforementioned XML can be trivially overridden rather than having to maintain a patch file.

android/java/org/chromium/chrome/browser/privacy/settings/BraveBrowserLockSettingsFragment.java
```diff
@@ -0,0 +1,149 @@
...
+        boolean previous = ChromeSharedPreferences.getInstance().readBoolean(prefKey, false);
+        if (previous == isChecked) return;
+
+        IncognitoReauthManager reauth = new IncognitoReauthManager(requireActivity(), profile);
+        reauth.startReauthenticationFlow(
+                new IncognitoReauthManager.IncognitoReauthCallback() {
+                    @Override
+                    public void onIncognitoReauthNotPossible() {
+                        reauth.destroy();
+                        revertToggle(toggle, prefKey, previous);
```
netzenbot, line 138: nit: Per AND-002, this reauth callback touches the fragment's `toggle` view (via `revertToggle`) after an async flow, but doesn't verify the fragment is still attached. The biometric prompt can outlive the fragment; guard the UI update with an `isAdded()`/`isDetached()` check before calling `revertToggle`. ([Check Fragment Attachment Before Async UI Updates](https://github.com/brave/brave-core/tree/master/docs/best-practices/android.md#AND-002))

android/java/org/chromium/chrome/browser/privacy/settings/BraveBrowserLockSettingsFragment.java
```diff
@@ -0,0 +1,149 @@
...
+        mPageTitle.set(getString(R.string.brave_browser_lock_title));
+        return inflater.inflate(R.layout.fragment_brave_browser_lock_settings, container, false);
+    }
+
+    @Override
+    public void onViewCreated(@NonNull View view, @Nullable Bundle savedInstanceState) {
+        MaterialSwitch switchEntireApp = view.findViewById(R.id.switch_entire_application);
+        MaterialSwitch switchPrivateTabs = view.findViewById(R.id.switch_private_tabs);
+
+        assert switchEntireApp != null;
```
netzenbot: nit: Per AND-042, `assert x != null` is for cases where you are not locally dereferencing the object. Since `switchEntireApp` and `switchPrivateTabs` are dereferenced right below, prefer `assumeNonNull(...)` as a precondition instead. ([Choose the Right Assert/Assume Pattern](https://github.com/brave/brave-core/tree/master/docs/best-practices/android.md#AND-042))

browser/ui/android/strings/android_brave_strings.grd
```diff
@@ -708,6 +708,36 @@ This file contains all "about" strings.  It is set to NOT be translated, in tran
...
+        Private tabs locked
+      </message>
+      <message name="IDS_BRAVE_UNLOCK_PRIVATE_TABS" desc="Button label to unlock private tabs via biometrics.">
+        Unlock private tabs
+      </message>
+      <message name="IDS_BRAVE_OPEN_NORMAL_TABS" desc="Button label on the private tabs lock screen to navigate to regular tabs instead.">
+        Open normal tabs
+      </message>
+      <message name="IDS_BRAVE_BROWSER_LOCK_SCREEN_DESCRIPTION" desc="Description at the top of the Browser Lock settings screen.">
+        Lock private tabs or the entire Brave app with your device's screen lock.
```
netzenbot: nit: Per [L10N-015](https://github.com/brave/brave-core/tree/master/docs/best-practices/localization.md#L10N-015), single-sentence settings toggle descriptions should not end with a period. This string is used as the `brave_browser_lock` preference summary (in `brave_main_preferences.xml` and `BraveMainPreferencesBase`), so remove the trailing period from `IDS_BRAVE_BROWSER_LOCK_SCREEN_DESCRIPTION`.

android/brave_java_sources.gni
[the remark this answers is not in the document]
Olipro: I give up on this. Moving it to a BUILD.gn just causes existing stuff that was modified to reference it to fail (like `BraveTabbedRootUiCoordinator.java` - GN might as well be a black-box. I've tried myself, I've tried throwing AI at the problem with no success at all. If someone knows how to make it work, I'm all ears but this has proven to be a boondoggle.

We either should just completely get rid of these `gni` files or this warning is noise.

android/java/brave-res/xml/brave_main_preferences.xml
```diff
@@ -121,6 +121,14 @@
         android:icon="@drawable/ic_window_tab_close"
         android:title="@string/prefs_closing_all_tabs_closes_brave"
         app:isPreferenceVisible="false"/>
+    <org.chromium.components.browser_ui.settings.ChromeBasePreference
```
netzenbot, line 124: nit: This preference is removed programmatically (removePreferenceIfPresent(PREF_BROWSER_LOCK)) when the reauth feature is unavailable, so add app:isPreferenceVisible="false" here to avoid a brief flash before the code hides it. The sibling closing_all_tabs_closes_brave preference already uses this pattern. ([Use app:isPreferenceVisible="false" for Conditionally Shown Preferences](https://github.com/brave/brave-core/tree/master/docs/best-practices/android.md#AND-008))

android/java/brave-res/drawable/ic_brave_wordmark.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright (c) 2026 The Brave Authors. All rights reserved.
+     This Source Code Form is subject to the terms of the Mozilla Public
+     License, v. 2.0. If a copy of the MPL was not distributed with this
+     file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+     Brave brand wordmark — converted from brave_wordmark.svg (root of the repo).
+     Use alongside ic_brave_icon_release_color to form the full Brand / Full Brave Brand. -->
+
```
netzenbot, line 9: This new drawable is added directly as a resource. Per the Nala/Leo icon guidelines, add the icon to `android/nala/icons.gni` instead of creating a standalone drawable file (the sibling `ic_biometric_login.xml` and `ic_brave_icon_release_color.xml` in this PR do this correctly). cc @brave/nala-token-reviewers ([Don't Add Custom Icons — Use Leo/Nala Icons](https://github.com/brave/brave-core/tree/master/docs/best-practices/nala.md#NA-009))

android/java/org/chromium/chrome/browser/tabbed_mode/BraveTabbedRootUiCoordinator.java
Olipro: Given the nature of this change, it's very unlikely that it could introduce a crash. With that said, allowing such a feature to be remotely disabled by us strikes me as something that would certainly upset users.

android/java/brave-res/drawable/ic_brave_wordmark.xml
Olipro: Currently doesn't exist in Nala

bsclifton (approved):
string reviewers ++

android/java/org/chromium/chrome/browser/privacy/settings/BraveBrowserLockSettingsFragment.java
```diff
@@ -0,0 +1,202 @@
...
+        switchEntireApp.setOnCheckedChangeListener(
+                (buttonView, isChecked) ->
+                        onToggleChanged(
+                                BravePreferenceKeys.BRAVE_BROWSER_LOCK,
+                                switchEntireApp,
+                                isChecked));
+        switchPrivateTabs.setOnCheckedChangeListener(
+                (buttonView, isChecked) ->
+                        onToggleChanged(
+                                BravePreferenceKeys.BRAVE_BROWSER_LOCK_PRIVATE_TABS_ONLY,
```
netzenbot, line 114: This toggle only writes the shared pref `browser_lock_private_tabs_only`, and nothing consumes it — the only other references are this fragment's own switch state and `updatePreventCaptureVisibility()`; `BraveBrowserLockManager.isBrowserLockEnabled()` reads only `BRAVE_BROWSER_LOCK`. Chromium's private-tab lock is driven by `Pref.INCOGNITO_REAUTHENTICATION_FOR_ANDROID` (set in `IncognitoLockSettings.onIncognitoReauthPreferenceChange`), and this commit removes the only UI that wrote it via `removePreferenceIfPresent("incognito_lock")` in `BravePrivacySettings`. As written the toggle is a no-op and private-tab locking is no longer user-configurable. ([Remove Unused Interfaces and Dead Code](https://github.com/brave/brave-core/tree/master/docs/best-practices/android.md#AND-015))

android/java/org/chromium/chrome/browser/privacy/settings/BravePrivacySettings.java
```diff
@@ -490,14 +483,12 @@ public void onCreatePreferences(@Nullable Bundle savedInstanceState, @Nullable S
...
 
-        mIncognitoScreenshot = (ChromeSwitchPreference) findPreference(PREF_INCOGNITO_SCREENSHOT);
-        mIncognitoScreenshot.setOnPreferenceChangeListener(this);
-
         mWebrtcPolicy = (ChromeBasePreference) findPreference(PREF_WEBRTC_POLICY);
 
         removePreferenceIfPresent(PREF_AD_BLOCK);
         removePreferenceIfPresent(PREF_SYNC_AND_SERVICES_LINK);
         removePreferenceIfPresent(PREF_NETWORK_PREDICTIONS);
+        removePreferenceIfPresent("incognito_lock");
```
netzenbot, line 502: Removing the upstream `incognito_lock` preference leaves nothing writing `Pref.INCOGNITO_REAUTHENTICATION_FOR_ANDROID` — `IncognitoLockSettings` is its only non-test writer, and `IncognitoReauthManager.isIncognitoReauthEnabled()` gates all private-tab reauth on that pref. The replacement "Private tabs" toggle in `BraveBrowserLockSettingsFragment` only writes the Java-side `browser_lock_private_tabs_only` shared pref, which no code reads, so private-tab locking is now unreachable (and users who already had it enabled can no longer turn it off).

browser/ui/android/strings/android_brave_strings.grd
```diff
@@ -3718,12 +3751,6 @@ If you don't accept this request, VPN will not reconnect and your internet conne
     <message name="IDS_SAFEBROWSING_ERRORS_PREF" desc="A preferences to show Safe Browsing errors">
       Show Safe Browsing errors
     </message>
-    <message name="IDS_PRIVATE_TAB_SCREENSHOT_TITLE" desc="A title to show private tab screenshot">
-      Allow screenshots in private tabs
-    </message>
-    <message name="IDS_PRIVATE_TAB_SCREENSHOT_SUMMARY" desc="A summary to show private tab screenshot">
-      If allowed, private tabs will also be visible when multiple apps are open.
```
diracdeltas, line 3737: was this intentionally removed?

diracdeltas:
1. why were there any changes to screen capture as part of this PR? i don't see it in the spec

2. please see the sec hotspot findings

android/java/org/chromium/chrome/browser/tabbed_mode/BraveTabbedRootUiCoordinator.java
DerrickRice: I may be misreading, but I believe the incognito lock was deployed behind such a flag, present in Chromium 94 and removed in 95. There was an issue where private tab locking locked people out of their private tabs, and we had to temporarily disable it to unlock users.

https://github.com/brave/brave-core/pull/18101

Seems like it might be prudent to roll this out under a feature flag and then remove the feature flag a couple releases later, once we know it's working as intended.

browser/ui/android/strings/android_brave_strings.grd
Olipro: Yes, just above there is new copy that pertains to the new UI. There's a single option toggle for disallowing screenshots that applies to whatever the user has locked.

When the screenshot toggle in the new UI is enabled:

1. If only private tabs are locked, only private tabs prevent screenshots.
2. If the whole browser is locked, you cannot screenshot anything.

If the toggle is disabled, screenshots can be taken for anything, regardless of what's locked. Changing the state of the toggle requires the user to authenticate.

diracdeltas: could you share a screenshot of this with the screenshot text/setting?

Olipro: Image

diracdeltas: i think that UI is not clear enough that "prevent screenshot..." only applies to whatever is already locked. it should be a sub-toggle or something below the locked item (only visible when lock is on) instead of a third item at the same level.

Olipro: As it currently stands, the toggle for enabling/disabling screenshots isn't visible unless at least one of the lockdown options are enabled. Pinging @aguscruiz to chime in on UX.

aguscruiz: How does this look to you @diracdeltas @Olipro? Separated that toggle into its own section and tweaked the messaging a little bit. 

image

diracdeltas: @aguscruiz that's more clear IMO though i'd suggest the text be something like "Prevent screenshot/video capture of locked tabs"

aguscruiz: Sounds good! Updated the design files 

image

android/java/org/chromium/chrome/browser/settings/BraveMainPreferencesBase.java
```diff
@@ -397,6 +400,14 @@ private void rearrangePreferenceOrders() {
             setPreferenceOrder(PREF_CLOSING_ALL_TABS_CLOSES_BRAVE, ++generalOrder);
         }
 
+        if (IncognitoReauthManager.isIncognitoReauthFeatureAvailable()) {
+            setPreferenceVisibleIfPresent(PREF_BROWSER_LOCK, true);
+            setPreferenceOrder(PREF_BROWSER_LOCK, ++generalOrder);
+        } else {
+            removePreferenceIfPresent(PREF_BROWSER_LOCK);
+        }
+        setPreferenceOrder(PREF_CLOSING_ALL_TABS_CLOSES_BRAVE, ++generalOrder);
```
netzenbot, line 409: This unconditional re-order runs even when `BRAVE_ANDROID_TAB_GROUPS_SETTINGS` is enabled, and that branch just did `removePreferenceIfPresent(PREF_CLOSING_ALL_TABS_CLOSES_BRAVE)` (line 397), so `findPreference` returns null and `setPreferenceOrder` (which does `assumeNonNull(preference); preference.setOrder(order);`) throws. It also makes the `setPreferenceOrder(PREF_CLOSING_ALL_TABS_CLOSES_BRAVE, ...)` call at line 400 dead — move this line into the same `else` branch, after the browser-lock block, rather than leaving it unconditional.

android/java/org/chromium/chrome/browser/privacy/BraveBrowserLockManager.java
```diff
@@ -0,0 +1,380 @@
...
+    public static boolean isPreventCaptureEnabled() {
+        return ChromeSharedPreferences.getInstance()
+                .readBoolean(BravePreferenceKeys.BRAVE_BROWSER_LOCK_PREVENT_CAPTURE, true);
+    }
+
+    private void applySecureFlagToAllActivities() {
+        boolean secure = isBrowserLockEnabled() && isPreventCaptureEnabled();
+        for (Activity activity : ApplicationStatus.getRunningActivities()) {
+            if (secure) {
+                activity.getWindow().addFlags(WindowManager.LayoutParams.FLAG_SECURE);
```
netzenbot, line 266: FLAG_SECURE on `ChromeTabbedActivity` is also owned by upstream `IncognitoSnapshotController.updateIncognitoTabSnapshotState()`, which does an unconditional `clearFlags(FLAG_SECURE)` whenever no incognito tab is showing; `IncognitoTabbedSnapshotController` runs it on every current-tab-model change and on HUB show/hide. Since enabling this toggle also disables the `IncognitoScreenshot` feature (taking upstream down the `expectedSecureState = isShowingIncognito` path), the flag set here is cleared as soon as the user is on a regular tab, so capture protection stops applying to the main window until the app is next stopped and `applySecureFlagToAllActivities()` re-adds it.

android/java/org/chromium/chrome/browser/tabbed_mode/BraveTabbedRootUiCoordinator.java
DerrickRice: After chatting with @Olipro, I'm comfortable going forward w/o a feature flag.

android/java/org/chromium/chrome/browser/privacy/BraveBrowserLockManager.java
Olipro: This should now be fixed.

android/java/org/chromium/chrome/browser/settings/BraveMainPreferencesBase.java
Olipro: This should be resolved.

Olipro:
@AlexeyBarabash Could you give this a once-over when you have a chance? I know there's a few `.gni` complaints but I'm loathe to faff around with these when half the time it tends to be an abortive procedure, at least it is with my level of GN-fu.

Privacy is happy with it now so code review is all that remains.

PR #36419, opened 2026-05-13
[Polkadot] Parse assets pallet chain metadata.

### LINKED ISSUE #55503: [Polkadot] Extract assets pallet information metadata, like required methods for assets support
### Description

We need to extract assets pallet information for supported polkadot networks.
Some basic information about methods, required to fetch balance and submit tx.

### Steps to reproduce

-

### Actual result

-

### Expected result

-

### Reproduces how often

Easily reproduced

-

### Channel information

- [ ] release (stable)
- [ ] beta
- [ ] nightly

### Reproducibility

### Miscellaneous information

_No response_

cypt4:
Parse basic methods for assets pallet including methods for asset id resolution, asset data resolution, balance and transfer.

Resolves https://github.com/brave/brave-browser/issues/55503
Root sec review issue(non-blocking): https://github.com/brave/reviews/issues/2250


netzenbot (approved):

components/brave_wallet/browser/polkadot/polkadot_chain_metadata_unittest.cc
```diff
@@ -19,6 +20,64 @@ namespace {
...
+      for (size_t j = 0; j < replacement.size(); ++j) {
+        (*bytes)[i + j] = static_cast<uint8_t>(replacement[j]);
+      }
+      return true;
+    }
+    ++matches;
+  }
+
+  return false;
+}
```
cmazakas: This code is a little hard to read and does manual indexing arithmetic.

Maybe we try something more along these lines:
```cpp
bool ReplaceNthOccurrence(std::vector<uint8_t>& bytes,
                          std::string_view needle,
                          std::string_view replacement,
                          size_t occurrence) {
  DCHECK(!needle.empty());

  auto it = bytes.begin();
  size_t num_found = 0;

  while (it < bytes.end()) {
    auto match =
        std::ranges::search(it, bytes.end(), needle.begin(), needle.end());
    if (match.empty()) {
      return false;
    }

    if (num_found == occurrence) {
      auto pos = bytes.erase(match.begin(), match.end());
      bytes.insert(pos, replacement.begin(), replacement.end());
      return true;
    }

    ++num_found;
    it = match.end();
  }

  return false;
}
```

Also, this function is sufficiently complex that I think it actually needs its own set of tests just to prove subsequent correctness of the tests.

components/brave_wallet/browser/polkadot/polkadot_chain_metadata_unittest.cc
```diff
@@ -148,6 +236,42 @@ TEST(PolkadotChainMetadataUnitTest, FromBytesInvalid) {
...
+
+  // SCALE string length prefix 0x4c encodes 19 bytes. The second
+  // transfer_keep_alive occurrence is the Assets pallet call variant; the first
+  // belongs to Balances.
+  constexpr std::string_view kTransferKeepAlive =
+      "\x4c"
+      "transfer_keep_alive";
+  constexpr std::string_view kTransferDeadAlive =
+      "\x4c"
+      "transfer_dead_alive";
```
cmazakas: Let's update `polkadot_extrinsic.rs` to include a helper in a similar vein to:
```rs
fn compact_scale_encode_u32(x: u32) -> Vec<u8> {
    Compact(x).encode()
}
```
and then we can consume it from this test file, becase including the manual prefixes here is unsustainable long-term.

cmazakas:
I think our approach to some of this stuff isn't very sustainable.

I think we need to make a smaller PR first, before we merge this one.

We can put struct definitions here:
```rs
use ffi::Rawr;
crate::impl_result!(Rawr, RawrResult);

#[cxx::bridge(namespace = brave_wallet)]
mod ffi {
    #[derive(Clone, Copy, Default, Debug)]
    pub struct Rawr {
        pub system_pallet_index: u8,
        pub balances_pallet_index: u8,
        pub transaction_payment_pallet_index: u8,
        pub transfer_allow_death_call_index: u8,
        pub transfer_keep_alive_call_index: u8,
        pub transfer_all_call_index: u8,
        pub assets_pallet_index: u8,
        pub assets_transfer_all_call_index: u8,
        pub assets_transfer_keep_alive_call_index: u8,
        pub has_assets_pallet: bool,
        pub ss58_prefix: u16,
        pub spec_version: u32,
        pub asset_tx_payment: bool,
    }

    extern "Rust" {
        type CxxPolkadotChainMetadataFieldsResult;
        type RawrResult;

        fn make_rawr() -> Box<RawrResult>;
        fn unwrap(self: &mut RawrResult) -> Box<Rawr>;
        fn is_ok(self: &RawrResult) -> bool;
        fn error_message(self: &RawrResult) -> String;
  }
}

fn make_rawr() -> Box<RawrResult> {
    Box::new(RawrResult(Ok(ffi::Rawr { ..Default::default() })))
}
```

Then in the C++, we can directly access this struct's data members. Right now, we're dealing with way too much boilerplate with all these fields and we can eliminate a lot of these middle layers.

```cpp
  auto rawr = *make_rawr()->unwrap();
  (void)rawr.asset_tx_payment;
```

As for the FromFields helpers, maybe we can consider using a builder pattern or something but I have concerns about how this code is evolving.

But we should be able to dramatically simplify our Rust <-> C++ here by just creating a struct whose definition is compatible among both languages. We can remove a lot of code, and I'd say we merge that cleanup PR first.

components/brave_wallet/browser/polkadot/polkadot_chain_metadata_prefs.h
```diff
@@ -18,7 +18,7 @@ namespace brave_wallet {
 
 class PolkadotChainMetadataPrefs {
  public:
-  static constexpr int kVersion = 1;
+  static constexpr int kVersion = 2;
```
kdenhartog, line 21: Is this internally unique to us or something specific to the changes being made in this PR for interoperability?

components/brave_wallet/browser/polkadot/polkadot_test_utils.cc
```diff
@@ -85,6 +95,40 @@ std::vector<uint8_t> ReadMetadataFixture(std::string_view file_name) {
...
+
+  while (it != bytes.end()) {
+    auto match =
+        std::search(it, bytes.end(), needle_bytes.begin(), needle_bytes.end());
+    if (match == bytes.end()) {
+      return false;
+    }
+
+    if (num_found == occurrence) {
+      auto pos = bytes.erase(match, std::next(match, needle_bytes.size()));
```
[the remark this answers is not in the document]
cdesouza-chromium: @cypt4 please, remove all the uses of `std::next` in this PR, because they circumvent `-Wunsafe-buffers`. For reference: https://chromium.googlesource.com/chromium/src/+/main/docs/unsafe_buffers.md#Avoid-std_next_for-silencing-warnings_use-ranges

components/brave_wallet/browser/polkadot/polkadot_test_utils.cc
```diff
@@ -85,6 +95,40 @@ std::vector<uint8_t> ReadMetadataFixture(std::string_view file_name) {
...
+      bytes.insert(pos, replacement_bytes.begin(), replacement_bytes.end());
+      return true;
+    }
+
+    ++num_found;
+    it = std::next(match, needle_bytes.size());
```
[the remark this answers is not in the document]
cdesouza-chromium: @cypt4 same here.

components/brave_wallet/browser/polkadot/polkadot_test_utils.cc
```diff
@@ -57,6 +63,10 @@ const base::ListValue* FindParamsOrNull(const base::DictValue& req_body) {
   return req_body.FindList("params");
 }
 
+std::vector<uint8_t> StringToBytes(std::string_view value) {
+  return base::ToVector(base::as_byte_span(value));
+}
```
cdesouza-chromium: I don't think we need this function to begin with. This function is copying the contents of `value` into a `std::vector<uint8_t>`. However, looking at its uses, we have:

```cxx
const auto needle_bytes = StringToBytes(needle);
const auto replacement_bytes = StringToBytes(replacement);
```

However, looking at the uses we have for this function, it looks like we really don't need these vector copies/allocations, and it could just be:

```
  auto needle_bytes = base::as_byte_span(needle);
  auto replacement_bytes = base::as_byte_span(replacement);
```

We also don't need `const` there, because span is a view type anyway.

components/brave_wallet/browser/polkadot/polkadot_test_utils.cc
```diff
@@ -85,6 +95,40 @@ std::vector<uint8_t> ReadMetadataFixture(std::string_view file_name) {
...
+  }
+
+  const auto needle_bytes = StringToBytes(needle);
+  const auto replacement_bytes = StringToBytes(replacement);
+  auto it = bytes.begin();
+  size_t num_found = 0;
+
+  while (it != bytes.end()) {
+    auto match =
+        std::search(it, bytes.end(), needle_bytes.begin(), needle_bytes.end());
```
cdesouza-chromium: Please use `std::ranges::search`

components/brave_wallet/browser/polkadot/polkadot_test_utils.cc
cdesouza-chromium:
```cxx
auto pos = bytes.erase(match, std::next(match, needle_bytes.size()));
```

If you need my help, I can try to come up with something, what you want here is some ranged approach to this erasure, rather than index-based.

components/brave_wallet/browser/polkadot/polkadot_test_utils.h
```diff
@@ -24,6 +26,10 @@ namespace brave_wallet {
 base::DictValue RequestBodyToJsonDict(const network::ResourceRequest& req);
 std::string ReadMetadataFixtureJson(std::string_view file_name);
 std::vector<uint8_t> ReadMetadataFixture(std::string_view file_name);
+bool ReplaceNthOccurrence(std::vector<uint8_t>& bytes,
+                          std::string_view needle,
+                          std::string_view replacement,
+                          size_t occurrence);
```
cdesouza-chromium, line 41: Please consider adding adequate documentation to this function, so users can infer in a glance it is used, as it has quite a lot of subtleties.

components/brave_wallet/browser/polkadot/polkadot_chain_metadata.h
```diff
@@ -47,7 +47,18 @@ class PolkadotChainMetadata {
       uint8_t transfer_all_call_index,
       uint16_t ss58_prefix,
       uint32_t spec_version,
-      bool asset_tx_payment);
+      bool asset_tx_payment,
+      bool has_assets_pallet = false,
+      uint8_t assets_pallet_index = 0,
+      uint8_t assets_transfer_all_call_index = 0,
+      uint8_t assets_transfer_keep_alive_call_index = 0);
```
cmazakas: We need to remove the default values here, otherwise it defeats the purpose of having `FromFields`.

components/brave_wallet/browser/internal/polkadot_chain_metadata.rs
```diff
@@ -396,6 +397,19 @@ fn parse_chain_metadata_fields(bytes: &[u8]) -> Result<CxxPolkadotChainMetadata,
     let transfer_all_call_index =
         get_call_index(&portable_registry, balances_pallet, "transferall")?;
 
+    let mut assets_pallet_index = 0;
+    let mut assets_transfer_all_call_index = 0;
+    let mut assets_transfer_keep_alive_call_index = 0;
+    let mut has_assets_pallet = false;
```
cmazakas: nit: let's put `has_assets_pallet` as the first variable declaration in this block

Declaration order should match assignment order, and the most important assignment here is `has_assets_pallet`.

components/brave_wallet/browser/polkadot/polkadot_chain_metadata.cc
```diff
@@ -48,22 +55,92 @@ PolkadotChainMetadata PolkadotChainMetadata::FromFields(
...
+  metadata->has_assets_pallet = has_assets_pallet;
   metadata->ss58_prefix = ss58_prefix;
   metadata->spec_version = spec_version;
   metadata->asset_tx_payment = asset_tx_payment;
 
   return metadata;
 }
 
+// static
+std::optional<PolkadotChainMetadata> PolkadotChainMetadata::FromChainName(
```
cmazakas: I'm confused, didn't we just have a PR that moved all of these to polkadot_test_utils because these shouldn't be used in production?

components/brave_wallet/browser/polkadot/polkadot_chain_metadata_unittest.cc
```diff
@@ -61,9 +101,87 @@ TEST(PolkadotChainMetadataUnitTest, EqualityOperator) {
...
+      /*assets_transfer_all_call_index=*/10,
+      /*assets_transfer_keep_alive_call_index=*/9);
+
   EXPECT_EQ(metadata_a, metadata_b);
   EXPECT_NE(metadata_a, metadata_c);
   EXPECT_NE(metadata_a, metadata_d);
+  EXPECT_NE(metadata_a, metadata_e);
+}
+
+TEST(PolkadotChainMetadataUnitTest, FromChainName) {
```
cmazakas: I think this might be a rebasing artifact but if we were ever gonna test `FromChainName`, it should be done via the fixture parsing unit tests, like we're already doing:

```cpp
  auto expected = PolkadotChainMetadata::FromFields(
      /*system_pallet_index=*/0, /*balances_pallet_index=*/0x0a,
      /*transaction_payment_pallet_index=*/0x0b,
      /*transfer_allow_death_call_index=*/0,
      /*transfer_keep_alive_call_index=*/3,
      /*transfer_all_call_index=*/4,
      /*ss58_prefix=*/42, /*spec_version=*/1'022'005,
      /*asset_tx_payment=*/true,
      /*has_assets_pallet=*/true,
      /*assets_pallet_index=*/50,
      /*assets_transfer_all_call_index=*/32,
      /*assets_transfer_keep_alive_call_index=*/9);
  EXPECT_EQ(*metadata, expected);

  auto metadata2 = PolkadotMetadataFromChainName("Westend Asset Hub");
  ASSERT_TRUE(metadata2);
  PolkadotChainMetadata expected_from_name = expected;
  expected_from_name->spec_version = 0;
  EXPECT_EQ(*metadata2, expected_from_name);
```

components/brave_wallet/browser/polkadot/polkadot_chain_metadata_unittest.cc
```diff
@@ -20,6 +22,36 @@ namespace {
...
+  std::vector<uint8_t> metadata_bytes;
+  EXPECT_TRUE(PrefixedHexStringToBytes(*metadata_hex, &metadata_bytes));
+  return metadata_bytes;
+}
+
+std::string ScaleEncodeString(std::string_view value) {
+  auto encoded = scale_encode_string(::rust::Slice<const uint8_t>(
+      reinterpret_cast<const uint8_t*>(value.data()), value.size()));
+  return std::string(encoded.begin(), encoded.end());
+}
```
cmazakas, line 25: CXX's `rust::Str` doesn't support construction from `std::string_view`, which is lame.

In general, we should try to avoid `reinterpret_cast` and this seems to work fine:
```cpp
std::string ScaleEncodeString(std::string_view value) {
  auto bytes = base::as_byte_span(value);
  auto encoded = scale_encode_string(::rust::Slice(bytes));
  return std::string(encoded.begin(), encoded.end());
}
```

We also could update the Rust code to accept `&str` and then pass `const std::string&`, which works fine.

components/brave_wallet/browser/polkadot/polkadot_test_utils_unittest.cc
```diff
@@ -0,0 +1,68 @@
...
+namespace {
+
+std::vector<uint8_t> BytesForTest(std::string_view value) {
+  std::vector<uint8_t> bytes;
+  bytes.reserve(value.size());
+  for (char c : value) {
+    bytes.push_back(static_cast<uint8_t>(c));
+  }
+  return bytes;
+}
```
cmazakas, line 21: This can be more succinctly written as:
```cpp
std::vector<uint8_t> BytesForTest(std::string_view value) {
  auto bytes = base::as_byte_span(value);
  return std::vector(bytes.begin(), bytes.end());
}
```

components/brave_wallet/browser/polkadot/polkadot_chain_metadata_unittest.cc
```diff
@@ -20,6 +22,36 @@ namespace {
 
 constexpr char kResult[] = "result";
 
+std::vector<uint8_t> ReadMetadataFixture(const char* fixture_name) {
+  const auto fixture_path =
+      base::PathService::CheckedGet(base::DIR_SRC_TEST_DATA_ROOT)
+          .AppendASCII("brave")
```
kdenhartog: hmm is this a normal way of accessing a file path URL? I've never seen this pattern used before
Also, do we need to be setting this data on disk for a test or can we just Mock it in some way instead?

kdenhartog (commented):
few questions

components/brave_wallet/browser/polkadot/polkadot_chain_metadata_prefs.h
cypt4: Since prefs schema is changed we need to update schema version so data is erased and reloaded on update.

components/brave_wallet/browser/polkadot/polkadot_chain_metadata_unittest.cc
cmazakas: We store production data to the filesystem because runtime metadata for Polkadot is around ~1 MB. This gives us reliable tests and ensures our parsing code is parsing something from the real world.

components/brave_wallet/browser/polkadot/polkadot_chain_metadata.cc
cypt4: Yep this is a rebase artifact

cmazakas (approved):
Looks great!

kdenhartog (approved):
my comments were addressed. Nothing else blocking. LGTM

PR #36422, opened 2026-05-13
[ads] Center New Tab Takeover CTA button for NTP refresh

### LINKED ISSUE #55316: [ads] Center New Tab Takeover CTA button in NTP refresh
Per discussion in Slack, the new position of CTA button will be non-configurable as it would be yet another bit of configuration to get wrong.
If we need something special, we can use dNTT for that specific case.

aseren:
Centers the CTA button on New Tab Takeover sponsored image pages in the NTP refresh when the `brave://flags/#center-ntt-cta-button` flag is enabled. The flag is disabled by default.

image

## Test plan

* Start the browser
* Trigger NTT
   EXPECTATION: CTA button is not centered
* Enable the flag on brave://flags/#center-ntt-cta-button
* Restart the browser
* Trigger NTT
   EXPECTATION: CTA button is centered

Resolves https://github.com/brave/brave-browser/issues/55316

### DIFF
--- a/browser/resources/brave_new_tab_page_refresh/components/app.style.ts
+++ b/browser/resources/brave_new_tab_page_refresh/components/app.style.ts
@@ -224,6 +224,10 @@ export const style = scoped.css`
     margin-inline-end: 0;
   }
 
+  .caption-container:has(.centered-ntt-cta-button) {
+    position: static;
+  }
+
   .widget-container {
     --widget-height: 128px;
     --widget-min-width: 380px;
--- a/browser/about_flags.cc
+++ b/browser/about_flags.cc
@@ -27,6 +27,7 @@
 #include "brave/components/debounce/core/common/features.h"
 #include "brave/components/email_aliases/buildflags/buildflags.h"
 #include "brave/components/google_sign_in_permission/features.h"
+#include "brave/components/ntp_background_images/browser/features.h"
 #include "brave/components/playlist/core/common/buildflags/buildflags.h"
 #include "brave/components/psst/buildflags/buildflags.h"
 #include "brave/components/request_otr/common/buildflags/buildflags.h"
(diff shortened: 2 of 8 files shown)


browser/resources/brave_new_tab_page_refresh/components/app.style.ts
```diff
@@ -224,6 +224,19 @@ export const style = scoped.css`
     margin-inline-end: 0;
   }
 
+  .ntt-cta-button {
```
zenparsing: I *think* it will center the element without these rules (with just `static` positioning).

browser/resources/brave_new_tab_page_refresh/components/app.tsx
```diff
@@ -36,6 +37,15 @@ export function App() {
   )
 
   const threeColumnWidth = useMediaQuery(threeColumnQuery)
+  const currentBackground = useCurrentBackground()
+  const centerNttCtaButtonFeatureEnabled = useNewTabState(
+    (s) => s.centerNttCtaButtonFeatureEnabled,
+  )
+  const backgroundCaptionClassName =
```
zenparsing: Rather than pushing the background info into the app view, I have another idea to try.

What if we set an additional CSS class on the `SponsoredBackgroundLogo` component, say `.centered-ntt-cta`.

And then we add a CSS rule `app.style.ts` to cancel out the absolute positioning if the container contains an element with that class name:

```css
  .caption-container:has(.centered-ntt-cta) {
    position: static;
  }
```

Would that work?

browser/resources/brave_new_tab_page_refresh/components/app.style.ts
aseren: Yes this works! Fixed, thanks!

browser/resources/brave_new_tab_page_refresh/components/app.tsx
aseren: Yes this works! Fixed, thanks!

zenparsing (approved):

netzenbot (approved):

PR #36433, opened 2026-05-13
[Wallet]: Handle More Permit Type Warnings

Douglashdaniel:
## Description 

Will now handle more `Permit` like warning in the `Sign` panel

Primary Types Covered:

```
[
  'Permit',
  'PermitSingle',
  'PermitBatch',
  'PermitTransferFrom',
  'PermitWitnessTransferFrom',
]
```

Permit Shapes Covered:

```
    names.has('spender')
    && (names.has('value')
      || names.has('amount')
      || t.some(
        (f) =>
          f.type === 'PermitDetails'
          || f.type === 'PermitDetails[]'
          || f.type === 'TokenPermissions',
      ))
```

https://github.com/user-attachments/assets/7fdb4382-0f03-4153-a944-ff886f7858d1

### DIFF
--- a/components/brave_wallet_ui/components/extension/sign-panel/sign_panel.test.tsx
+++ b/components/brave_wallet_ui/components/extension/sign-panel/sign_panel.test.tsx
@@ -4,6 +4,7 @@
 // You can obtain one at https://mozilla.org/MPL/2.0/.
 
 import * as React from 'react'
+import '@testing-library/jest-dom'
 import { render, screen, waitFor } from '@testing-library/react'
 import { Provider } from 'react-redux'
 
@@ -15,7 +16,10 @@ import {
   mockCardanoAccount, //
 } from '../../../stories/mock-data/mock-wallet-accounts'
 import { mockOriginInfo } from '../../../stories/mock-data/mock-origin-info'
-import { mockSolanaAccount } from '../../../common/constants/mocks'
+import {
+  mockEthAccount,
+  mockSolanaAccount,
+} from '../../../common/constants/mocks'
 
 // Utils
 import BraveCoreThemeProvider from '../../../../common/BraveCoreThemeProvider'
@@ -23,6 +27,23 @@ import { createMockStore } from '../../../utils/test-utils'
 
 // Components
 import { SignPanel } from './index'
+
+function makeEthSignTypedData(
+  overrides: Partial<BraveWallet.EthSignTypedData>,
+): BraveWallet.EthSignTypedData {
+  return {
+    addressParam: '0x',
+    chainId: '1',
+    domainHash: [],
+    domainJson: '{}',
+    messageJson: '{}',
+    meta: undefined,
+    primaryHash: [],
+    primaryType: 'Person',
+    typesJson: '{}',
+    ...overrides,
+  }
+}
 
 const signCardanoMessageData: BraveWallet.SignMessageRequest = {
   id: 0,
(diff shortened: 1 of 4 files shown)


components/brave_wallet_ui/utils/sign_utils.test.ts
```diff
@@ -0,0 +1,232 @@
...
+            { name: 'value', type: 'uint256' },
+          ]),
+        }),
+      ),
+    ).toBe(true)
+  })
+
+  it('returns false when primaryType is empty and shape does not match', () => {
+    expect(
+      isPermitLikeEthSignTypedData(
```
netzenbot, line 394: nit: The `hasPermitShape` suite has 9 separate `it()` cases all invoking the same function with different field combinations. Since you already use `it.each` for `isPermitLikeEthSignTypedData`, consider the same approach here to reduce boilerplate. ([TI-023](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-023))

components/brave_wallet_ui/components/extension/sign-panel/sign_panel.test.tsx
```diff
@@ -132,4 +195,52 @@ describe('SignTypedDataPanel', () => {
...
+
+  it('Should show sign risk warning for EIP-712 Permit primary type', async () => {
+    const store = createMockStore({})
+    const { container } = render(
+      <Provider store={store}>
+        <BraveCoreThemeProvider>
+          <SignPanel
+            signMessageData={[signEthPermitTypedDataRequest]}
+            showWarning={false}
+          />
```
netzenbot, line 251: nit: The two new tests share an identical `waitFor` assertion block (network name, warning title, warning text, continue button). Extract it into a helper like `expectPermitWarningShown(container)` to remove the duplication. ([TI-024](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-024))

components/brave_wallet_ui/utils/sign_utils.ts
```diff
@@ -0,0 +1,65 @@
...
+
+import { BraveWallet } from '../constants/types'
+
+export const PERMIT_LIKE: readonly string[] = [
+  'Permit',
+  'PermitSingle',
+  'PermitBatch',
+  'PermitTransferFrom',
+  'PermitWitnessTransferFrom',
+]
```
kdenhartog, line 26:
```suggestion
export const PERMIT_LIKE: readonly string[] = [
  // ERC-2612 and DAI-style
  'Permit',
  
  // Uniswap Permit2 (AllowanceTransfer)
  'PermitSingle',
  'PermitBatch',
  
  // Uniswap Permit2 (SignatureTransfer)
  'PermitTransferFrom',
  'PermitBatchTransferFrom',
  'PermitWitnessTransferFrom',
  'PermitBatchWitnessTransferFrom',
  
  // EIP-3009 (USDC and for x402 support)
  'TransferWithAuthorization',
  'ReceiveWithAuthorization',
  'CancelAuthorization',
]
```

components/brave_wallet_ui/utils/sign_utils.ts
```diff
@@ -0,0 +1,65 @@
...
+    return undefined
+  }
+  try {
+    return JSON.parse(typesJson) as Record<string, TypedDataField[]>
+  } catch {
+    return undefined
+  }
+}
+
+export function hasPermitShape(td: BraveWallet.EthSignTypedData): boolean {
```
kdenhartog: We should check for Dai/EIP-3009 style authorizations too:

```
// DAI-style: holder + spender + allowed(bool)
const isDaiStyle =
  names.has('spender')
  && names.has('allowed')
  && t.some((f) => f.name === 'allowed' && f.type === 'bool')

// EIP-3009: from + to + value + validAfter + validBefore + nonce
const isAuthorizationStyle =
  names.has('from')
  && names.has('to')
  && names.has('value')
  && names.has('validAfter')
  && names.has('validBefore')
  && names.has('nonce')
```

components/brave_wallet_ui/utils/sign_utils.test.ts
```diff
@@ -0,0 +1,232 @@
...
+    messageJson: '{}',
+    meta: undefined,
+    primaryHash: [],
+    primaryType: 'Person',
+    typesJson: '{}',
+    ...overrides,
+  }
+}
+
+function typesJsonForPrimary(
```
kdenhartog, line 34: this seems to only check the primary type, but I think the permit could be smuggled through a nested type structure. Can we add something with like a benign primary type of "Nested" which then contains a permit2 type authorization?

We could likely catch this with a check against the most commonly used permit2 contract address using this:

```
const PERMIT2_ADDRESS = '0x000000000022d473030f116ddee9f6b43ac78ba3'

function isPermit2Domain(td: BraveWallet.EthSignTypedData): boolean {
  const domain = parseDomain(td.domainJson)
  return domain?.verifyingContract?.toLowerCase() === PERMIT2_ADDRESS
}
```

components/brave_wallet_ui/utils/sign_utils.ts
```diff
@@ -0,0 +1,65 @@
...
+  const names = new Set(t.map((f) => f.name))
+  return (
+    names.has('spender')
+    && (names.has('value')
+      || names.has('amount')
+      || t.some(
+        (f) =>
+          f.type === 'PermitDetails'
+          || f.type === 'PermitDetails[]'
+          || f.type === 'TokenPermissions',
```
kdenhartog: needed for the batch variants being checked above:

```suggestion
          || f.type === 'TokenPermissions'
          || f.type === 'TokenPermissions[]'
```

components/brave_wallet_ui/utils/sign_utils.ts
```diff
@@ -0,0 +1,65 @@
...
+  const types = parseTypesJson(td.typesJson)
+  const primaryType = td.primaryType
+  if (!types || !primaryType) {
+    return false
+  }
+  const t = types[primaryType]
+  if (!Array.isArray(t)) {
+    return false
+  }
+  const names = new Set(t.map((f) => f.name))
```
kdenhartog:
```suggestion
  const names = new Set(
    t.flatMap((f) => (f && typeof f.name === 'string' ? [f.name] : []))
  )
```

kdenhartog (changes requested):
Found some additional cases related to other similar pull style payment authorizations like Permit2 which we should account for here.

kdenhartog:
Since this is all in react, is this going to get picked up by Android as well or do we need a 3rd PR for it? Just seeing we have to do this separately in https://github.com/brave/brave-core/pull/36463 for iOS.

components/brave_wallet_ui/utils/sign_utils.ts
Douglashdaniel: Added!

components/brave_wallet_ui/utils/sign_utils.ts
Douglashdaniel: Added!

components/brave_wallet_ui/utils/sign_utils.test.ts
Douglashdaniel: Good catch, added!

components/brave_wallet_ui/utils/sign_utils.ts
Douglashdaniel: Updated!

Douglashdaniel:
> Since this is all in react, is this going to get picked up by Android as well or do we need a 3rd PR for it? Just seeing we have to do this separately in #36463 for iOS.

No, this will only cover `Desktop` UI

kdenhartog (approved):
found two additional checks here as well for EIP-3009 and Dai style authorizations.

onyb (approved):

kdenhartog:
Apologies, that was a message still sitting in github localStorage from first review. This is good to go

Douglashdaniel:
Closing this out in favor of https://github.com/brave/brave-core/pull/36619

PR #36436, opened 2026-05-13
Fix UBSAN error in BraveTabStyleViews

sangwoo108:
During startup tab could be in partially constructed state and BraveTab could be null during startup.

Resolves https://github.com/brave/internal/issues/1719

### DIFF
--- a/browser/ui/views/tabs/brave_tab_style_views.inc.cc
+++ b/browser/ui/views/tabs/brave_tab_style_views.inc.cc
@@ -14,6 +14,7 @@
 #include "chrome/browser/ui/browser_window/public/browser_window_features.h"
 #include "chrome/browser/ui/layout_constants.h"
 #include "ui/gfx/geometry/skia_conversions.h"
+#include "ui/views/view_utils.h"
 
 namespace {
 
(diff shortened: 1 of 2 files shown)


browser/ui/views/tabs/brave_tab_style_views.inc.cc
```diff
@@ -297,7 +298,12 @@ gfx::Insets BraveVerticalTabStyle::GetContentsInsets() const {
   };
 
   auto add_left_padding_for_accent_icon = [&](gfx::Insets& insets) {
-    auto* brave_tab = static_cast<const BraveTab*>(tab());
+    auto* brave_tab = views::AsViewClass<BraveTab>(tab());
+    if (!brave_tab) {
+      // During startup tab could be in partially constructed state and BraveTab
```
simonhong, line 303: Here and other places
```suggestion
      // During startup tab could be in partially constructed state and BraveTab
      // could be null during startup.
```

simonhong (approved):
++

browser/ui/views/tabs/brave_tab_style_views.inc.cc
sangwoo108: Oops, thanks.

netzenbot (approved):

PR #36438, opened 2026-05-14
Add best practice: prefer AsViewClass/IsViewClass over static_cast for views

sangwoo108:
From experience https://github.com/brave/brave-core/pull/36436


sangwoo108:
@simonhong what do you think about this? Is this too strong?

simonhong (approved):
++ I like this. I left same comment in review.

PR #36445, opened 2026-05-14
Make GN format sort leo icons files automatically.

goodov:
Auto-format leo icons sources in GN. GN format treats variables that end with "sources", "deps" as lists where order doesn't matter, so it can format them alphabetically.


netzenbot (approved):

fallaciousreasoning (approved):
nice, that's super handy!

PR #36454, opened 2026-05-14
[AI Chat] Include specific message type for extracted file text

porteron:
- Both the PDF and text-file extracted-text paths now create NewBraveFileExtractedTextContentBlock instead of NewTextContentBlock
- Applies to both pdf and other extractable file types

Resolves  https://github.com/brave/aichat/pull/1478


porteron:
https://github.com/brave/brave-core/pull/36464

PR #36463, opened 2026-05-14
[iOS][Wallet] Display warnings for more eth sign typed data

nuo-xu:
No strict check on `primaryType` but check both `primaryType` and the shape of `EthSignTypedData .typesJson`

Resolves https://github.com/brave/internal/issues/1733

Test:
1. visit PoC dApp: https://syarif07.my.id/wallet/permit.html
2. connect the dApp with wallet
3. tap each Test 1 to Test 6
4. each test will ask for sign 
5. verify the sign request screen shows warnings

https://github.com/user-attachments/assets/05acf71e-d179-4e67-babb-63ef7ce39610

### DIFF
--- a/ios/brave-ios/Sources/BraveWallet/Extensions/BraveWalletExtensions.swift
+++ b/ios/brave-ios/Sources/BraveWallet/Extensions/BraveWalletExtensions.swift
@@ -859,3 +859,38 @@ extension BraveWallet.MeldCryptoCurrency {
     }
   }
 }
+
+extension BraveWallet.SignDataUnion {
+  var isWarningsNeeded: Bool {
+    if cardanoSignData != nil { return true }
+    if let ethSignTypedData {
+      let permitLikeTypes: [String] = [
+        "Permit",
+        "PermitSingle",
+        "PermitBatch",
+        "PermitTransferFrom",
+        "PermitWitnessTransferFrom",
+      ]
+      return permitLikeTypes.contains(ethSignTypedData.primaryType)
+        || ethSignTypedData.hasPermitShape
+    }
+    return false
+  }
+}
+
+extension BraveWallet.EthSignTypedData {
+  var hasPermitShape: Bool {
+    guard
+      let data = self.typesJson.data(using: .utf8),
+      let typeMap = try? JSONDecoder().decode([String: [[String: String]]].self, from: data),
+      let fields = typeMap[primaryType]
+    else { return false }
+    let fieldNames = Set(fields.compactMap { $0["name"] })
+    return fieldNames.contains("spender")
+      && (fieldNames.contains("value")
+        || fieldNames.contains("amount")
+        || fields.contains(where: {
+          ["PermitDetails", "PermitDetails[]", "TokenPermissions"].contains($0["type"])
+        }))
+  }
+}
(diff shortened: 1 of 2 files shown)


netzenbot (approved):

ios/brave-ios/Sources/BraveWallet/Extensions/BraveWalletExtensions.swift
```diff
@@ -859,3 +859,38 @@ extension BraveWallet.MeldCryptoCurrency {
...
+      let typeMap = try? JSONDecoder().decode([String: [[String: String]]].self, from: data),
+      let fields = typeMap[primaryType]
+    else { return false }
+    let fieldNames = Set(fields.compactMap { $0["name"] })
+    return fieldNames.contains("spender")
+      && (fieldNames.contains("value")
+        || fieldNames.contains("amount")
+        || fields.contains(where: {
+          ["PermitDetails", "PermitDetails[]", "TokenPermissions"].contains($0["type"])
+        }))
```
kylehickinson: This seems kind of fragile, do we not have anything shared/defined more concretely in the shared wallet component code?

nuo-xu: unfortunately no.  desktop logic lives entirely in the web/extension UI layer. https://github.com/brave/brave-core/pull/36433/changes#diff-6185468ee520187719c7b5df989ed4aebab5b851a678e168bb58215d601d9246

kdenhartog: We should likely refactor this into https://github.com/brave/brave-core/blob/master/components/brave_wallet/common/eth_request_helper.cc#L408 instead and then have it pass a bool over the IPC via mojo and then frontend just checks for that instead.

@brave/crypto-wallets-core is this something that would be relatively easy to do rather than 2 separate front end changes?

kdenhartog:
Requested some changes in https://github.com/brave/brave-core/pull/36433#pullrequestreview-4324540343. @nuo-xu would you be able to copy the requested changes over to this PR as well for iOS?

ios/brave-ios/Sources/BraveWallet/Extensions/BraveWalletExtensions.swift
```diff
@@ -859,3 +859,75 @@ extension BraveWallet.MeldCryptoCurrency {
...
+
+    let isAuthorizationStyle =
+      names.contains("from") && names.contains("to") && names.contains("value")
+      && names.contains("validAfter") && names.contains("validBefore")
+      && names.contains("nonce")
+
+    return isPermitStyle || isDaiStyle || isAuthorizationStyle
+  }
+
+  private var hasPermitShape: Bool {
```
kdenhartog, line 931: Can we add support for this check too: https://github.com/brave/brave-core/pull/36433#discussion_r3270708074


this seems to only check the primary type, but I think the permit could be smuggled through a nested type structure. Can we add something with like a benign primary type of "Nested" which then contains a permit2 type authorization?

We could likely catch this with a check against the most commonly used permit2 contract address using this:

```
const PERMIT2_ADDRESS = '0x000000000022d473030f116ddee9f6b43ac78ba3'

function isPermit2Domain(td: BraveWallet.EthSignTypedData): boolean {
  const domain = parseDomain(td.domainJson)
  return domain?.verifyingContract?.toLowerCase() === PERMIT2_ADDRESS
}
```

but we'll also need to recursively parse through rather than only checking the top level types

nuo-xu: added in [6975f53](https://github.com/brave/brave-core/pull/36463/commits/6975f530cb72b87dbe32c89c7ededfe7aac79bc1)

nuo-xu:
Close this as we are going to show warnings for all eip-712 tx https://github.com/brave/brave-core/pull/36634

PR #36464, opened 2026-05-14
[AI Chat] Send extracted file text via dedicated content block type

### LINKED ISSUE #54543: [AI Chat] File extracted text should have dedicated types
With https://github.com/brave/aichat/pull/1344, 
we should send extracted text from files through
```
{
  "content": str | list[TextContentPart], 
  "type": "brave-file-extracted-text"
}
```
instead of pure text content type so server can better handling large file content without hitting context limit.

darkdh:
Introduce FileExtractedTextContentBlock and serialize it as the brave-file-extracted-text content type defined in
https://github.com/brave/aichat/pull/1344 so the server can handle large extracted file content without hitting context limits. The new block is emitted for both PDF and plain text uploads whenever extracted text is available. Local-LLM fallback in the OAI client passes the text through as a plain text block.

Resolves https://github.com/brave/brave-browser/issues/54543

### pdf Leo
1. Open Leo and upload a pdf which has copiable text in it
2. Ask Leo with automatic model
3. We should be able to get answer

### text Leo
1. Open Leo and upload a text file
2. Ask Leo with automatic model
3. We should be able to get answer

### pdf BYOM
1. Open Leo and upload a pdf which has copiable text in it
2. Ask Leo with ollama model 
3. We should be able to get answer

### text Leo
1. Open Leo and upload a text file
2. Ask Leo with ollama model 
3. We should be able to get answer


yrliou (approved):
LGTM

yrliou:
We should have a manual test plan too for a sanity check that things are working fine end-to-end as expected.

netzenbot (approved):

PR #36472, opened 2026-05-15
[Sidebar V2] Make brave's panel header as `SidePanel`'s header

### LINKED ISSUE #55544: [Sidebar V2] Make brave's Bookmarks/ReadingList panel header as SidePanel's header
We wrapped upstream's Bookmarks/ReadingList panel view in additional View to use our own panel header like below.

We wrapped upstream's Bookmarks/ReadingList panel view with our header view(below image) as another panel view.
In Sidebar V2, it's set as child of `ContentParentView` in `SidePanel`.
`ContentParentView` has assumption that it's child is webview. so it updates panel radius when first child is WebView subclass. But, it's not WebView actually. It's plain `View` like mentioned above. So, radius is not updated for Bookmarks/ReadingList panel in V2.

Instead of having additional parent view, we should set our panel header as SidePanel's header view.

Image

simonhong:
Sidebar v2 attaches a Brave-styled panel header at the SidePanel level
instead of embedding it inside the content view.

## Changes

  - New BraveSidePanelHeader (view) + BraveSidePanelHeaderController (delegate) under brave/browser/ui/views/side_panel/. Visual spec (title, optional launch button + separator, close button, 60px height, Nala colors, Leo icons) mirrors V1's inline headers in BraveReadLaterSidePanelView/BraveBookmarksSidePanelView.
  - BraveSidePanelCoordinator::PopulateSidePanel attaches the header via upstream SidePanel::AddHeaderView only for entries ShouldShowBraveHeader allows (today: kReadingList, kBookmarks).
  - SidePanel chromium_src override now wraps upstream AddHeaderView / RemoveHeaderView so each call also re-runs Brave's UpdateBorder().
  UpdateBorder() reserves a top inset equal to the header's preferred height when a header is attached, in both the rounded-border and no-border states.
  - brave_side_panel_header_controller is split into its own :side_panel_impl source_set (depending on :side_panel) to keep dep chains clean.

## Test plan

  - New SidebarBrowserTest.SidebarV2BraveHeaderTest test case
  - Manual: run with enable_sidebar_v2=true, open reading list and bookmarks — header should be visible

Resolves https://github.com/brave/brave-browser/issues/55544

### DIFF
--- a/browser/ui/views/side_panel/brave_side_panel_header.cc
+++ b/browser/ui/views/side_panel/brave_side_panel_header.cc
@@ -1,0 +1,76 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "brave/browser/ui/views/side_panel/brave_side_panel_header.h"
+
+#include <utility>
+
+#include "base/check.h"
+#include "brave/ui/color/nala/nala_color_id.h"
+#include "ui/base/metadata/metadata_impl_macros.h"
+#include "ui/gfx/geometry/insets.h"
+#include "ui/views/background.h"
+#include "ui/views/controls/button/image_button.h"
+#include "ui/views/controls/label.h"
+#include "ui/views/controls/separator.h"
+#include "ui/views/layout/flex_layout.h"
+#include "ui/views/layout/flex_layout_types.h"
+#include "ui/views/layout/layout_types.h"
+#include "ui/views/view_class_properties.h"
+
+namespace {
+
+constexpr int kHeaderInteriorMargin = 16;
+constexpr int kHeaderButtonSize = 20;
+constexpr int kHeaderHeight = 60;
+constexpr int kSeparatorHorizontalSpacing = 12;
+
+}  // namespace
+
+BraveSidePanelHeader::BraveSidePanelHeader(std::unique_ptr<Delegate> delegate)
+    : delegate_(std::move(delegate)) {
+  CHECK(delegate_);
+
+  SetBackground(views::CreateSolidBackground(nala::kColorPageBackground));
+  SetLayoutManager(std::make_unique<views::FlexLayout>())
+      ->SetOrientation(views::LayoutOrientation::kHorizontal)
+      .SetInteriorMargin(gfx::Insets(kHeaderInteriorMargin))
+      .SetMainAxisAlignment(views::LayoutAlignment::kStart)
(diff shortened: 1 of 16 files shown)


cdesouza-chromium (approved):
LGTM

browser/ui/views/side_panel/brave_side_panel_header.cc
```diff
@@ -0,0 +1,76 @@
...
+
+BraveSidePanelHeader::~BraveSidePanelHeader() = default;
+
+void BraveSidePanelHeader::Layout(PassKey) {
+  LayoutSuperclass<views::View>(this);
+
+  const gfx::Rect contents_bounds = parent()->GetContentsBounds();
+  SetBoundsRect(gfx::Rect(contents_bounds.x(),
+                          contents_bounds.y() - kHeaderHeight,
+                          contents_bounds.width(), kHeaderHeight));
```
sangwoo108, line 73: Q. In general, this would be a responsibility of the parent view. Wonder why this is done here. Is it for leave upstream code untouched? This would lead to another following Layout() call

chromium_src/chrome/browser/ui/views/side_panel/reading_list/reading_list_side_panel_coordinator.cc
```diff
@@ -3,11 +3,16 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this file,
 // You can obtain one at https://mozilla.org/MPL/2.0/.
 
+#include "brave/browser/ui/sidebar/buildflags/buildflags.h"
 #include "brave/browser/ui/views/side_panel/brave_read_later_side_panel_view.h"
 #include "chrome/browser/ui/views/side_panel/reading_list/read_later_side_panel_web_view.h"
 
+#if !BUILDFLAG(ENABLE_SIDEBAR_V2)
 #define ReadLaterSidePanelWebView BraveReadLaterSidePanelView
```
sangwoo108, line 11: Optional: How about put CHECK(!kSidebarV2) in BraveReadLaterSidePanelView and other classes' constructors that shouldn't be populated under new feature flag?

sangwoo108 (approved):
LGTM!

netzenbot (approved):

browser/ui/views/side_panel/brave_side_panel_header.cc
simonhong: It's what upstream sets panel header's bounds. As `SidePanel` uses `FillLayout`, header should set its bounds by itself. Maybe due to the header is pluggable. Added comment.

chromium_src/chrome/browser/ui/views/side_panel/reading_list/reading_list_side_panel_coordinator.cc
simonhong: Added.

PR #36474, opened 2026-05-15
Change farbling seed per container.

### LINKED ISSUE #55601: [Containers] Change farbling seed per container
We randomize farbling seed per website in each profile. We should also randomize it per container for increased privacy.

goodov:
Introduce additional entropy to farbling seed generator and use container ID (internally UUID) as this entropy.

Resolves https://github.com/brave/brave-browser/issues/55601

### DIFF
--- a/browser/brave_content_browser_client.cc
+++ b/browser/brave_content_browser_client.cc
@@ -868,7 +868,7 @@ BraveContentBrowserClient::WorkerGetBraveShieldSettings(
 #if BUILDFLAG(ENABLE_CONTAINERS)
   if (storage_partition_config &&
       base::FeatureList::IsEnabled(containers::features::kContainers)) {
-    additional_entropy +=
+    additional_entropy =
         std::string(containers::GetContainerIdFromStoragePartitionConfig(
             *storage_partition_config));
   }
--- a/browser/brave_content_browser_client.cc
+++ b/browser/brave_content_browser_client.cc
@@ -852,22 +852,32 @@ BraveContentBrowserClient::GetEphemeralStorageToken(
 bool BraveContentBrowserClient::AllowWorkerFingerprinting(
     const GURL& url,
     content::BrowserContext* browser_context) {
-  return WorkerGetBraveShieldSettings(url, browser_context)->farbling_level !=
-         brave_shields::mojom::FarblingLevel::MAXIMUM;
+  return WorkerGetBraveShieldSettings(url, browser_context, nullptr)
+             ->farbling_level != brave_shields::mojom::FarblingLevel::MAXIMUM;
 }
 
 brave_shields::mojom::ShieldsSettingsPtr
 BraveContentBrowserClient::WorkerGetBraveShieldSettings(
     const GURL& url,
-    content::BrowserContext* browser_context) {
+    content::BrowserContext* browser_context,
+    const content::StoragePartitionConfig* storage_partition_config) {
   const brave_shields::mojom::FarblingLevel farbling_level =
       brave_shields::GetFarblingLevel(
           HostContentSettingsMapFactory::GetForProfile(browser_context), url);
+  std::string additional_entropy;
+#if BUILDFLAG(ENABLE_CONTAINERS)
+  if (storage_partition_config &&
+      base::FeatureList::IsEnabled(containers::features::kContainers)) {
+    additional_entropy +=
+        std::string(containers::GetContainerIdFromStoragePartitionConfig(
+            *storage_partition_config));
+  }
+#endif
   const base::Token farbling_token =
       farbling_level != brave_shields::mojom::FarblingLevel::OFF
           ? brave_shields::GetFarblingToken(
                 HostContentSettingsMapFactory::GetForProfile(browser_context),
-                url)
+                url, base::as_byte_span(additional_entropy))
           : base::Token();
 
   PrefService* pref_service = user_prefs::UserPrefs::Get(browser_context);
--- a/browser/containers/containers_browsertest.cc
+++ b/browser/containers/containers_browsertest.cc
@@ -59,7 +59,14 @@
 
 namespace containers {
 
+namespace {
+
 constexpr char kTestContainerId[] = "test-container-id";
+
+constexpr char kFarblingPluginsStringScript[] =
+    "Array.from(navigator.plugins).map(p => p.name).join(',');";
+
+}  // namespace
 
 class ContainersBrowserTest : public InProcessBrowserTest {
  public:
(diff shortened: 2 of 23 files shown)


browser/brave_content_browser_client.cc
```diff
@@ -852,22 +852,32 @@ BraveContentBrowserClient::GetEphemeralStorageToken(
...
+    content::BrowserContext* browser_context,
+    const content::StoragePartitionConfig* storage_partition_config) {
   const brave_shields::mojom::FarblingLevel farbling_level =
       brave_shields::GetFarblingLevel(
           HostContentSettingsMapFactory::GetForProfile(browser_context), url);
+  std::string additional_entropy;
+#if BUILDFLAG(ENABLE_CONTAINERS)
+  if (storage_partition_config &&
+      base::FeatureList::IsEnabled(containers::features::kContainers)) {
+    additional_entropy +=
```
cdesouza-chromium: nit: I would use just `additional_entropy =` as the `+=` can make the reader have to look again to understand that nothing was there before.

mkarolin (approved):
`chromium_src`, `DEPS` ++

cdesouza-chromium (approved):
LGTM

components/brave_shields/core/browser/brave_shields_utils.cc
```diff
@@ -871,7 +873,15 @@ base::Token GetFarblingToken(HostContentSettingsMap* map, const GURL& url) {
     SetShieldsMetadata(map, url, std::move(shields_metadata));
   }
 
-  return token;
+  if (additional_entropy.empty()) {
+    return token;
+  }
+
+  const uint32_t entropy_hash = base::PersistentHash(additional_entropy);
+  const uint32_t high = token.high() ^ entropy_hash;
```
netzenbot: `token.high()` and `token.low()` both return `uint64_t`, but assigning them to `uint32_t high` / `uint32_t low` silently truncates the upper 32 bits of each. The `base::Token(high, low)` constructor then zero-extends these 32-bit values back to 64 bits, so the resulting token has its upper 32 bits of both halves set to zero — reducing effective entropy from 128 bits to 64 bits. These variables should be `uint64_t`.

components/brave_shields/core/browser/brave_shields_utils.cc
```diff
@@ -871,7 +873,15 @@ base::Token GetFarblingToken(HostContentSettingsMap* map, const GURL& url) {
     SetShieldsMetadata(map, url, std::move(shields_metadata));
   }
 
-  return token;
+  if (additional_entropy.empty()) {
+    return token;
+  }
+
```
netzenbot, line 886: `high` and `low` are declared `uint32_t`, but `token.high()` and `token.low()` return `uint64_t` — the upper 32 bits of both words are truncated and zeroed in every container-specific token. This halves the effective entropy and means any container tab's farbling token will always have the top 32 bits of each word set to zero, which is a detectable fingerprint. Use `uint64_t` for both and XOR the full 64-bit values (e.g. combine two `PersistentHash` calls or use a wider hash).

components/brave_shields/core/browser/brave_shields_utils.cc
```diff
@@ -871,7 +873,15 @@ base::Token GetFarblingToken(HostContentSettingsMap* map, const GURL& url) {
     SetShieldsMetadata(map, url, std::move(shields_metadata));
```
netzenbot, line 880: There's a truncation bug here: `token.high()` and `token.low()` return `uint64_t`, but `high` and `low` are declared as `uint32_t`. The truncation discards the upper 32 bits of each word, so the returned `base::Token` always has zero in bits 32–63 of both halves — cutting effective entropy from 128 bits to 64. Declare them as `uint64_t` instead: `const uint64_t high = token.high() ^ static_cast<uint64_t>(entropy_hash);`.

components/brave_shields/core/browser/brave_shields_utils.cc
```diff
@@ -871,7 +873,15 @@ base::Token GetFarblingToken(HostContentSettingsMap* map, const GURL& url) {
...
   }
 
-  return token;
+  if (additional_entropy.empty()) {
+    return token;
```
netzenbot, line 884: The variables `high` and `low` are declared as `uint32_t`, but `token.high()` and `token.low()` return `uint64_t`. Assigning a `uint64_t` to a `uint32_t` silently truncates the upper 32 bits of each word, so the returned `base::Token` will always have zeros in the top 32 bits of each half — reducing the effective entropy of the container-mixed token from 128 bits to 64 bits. Consider using `uint64_t high = token.high() ^ entropy_hash;` and XORing `token.low()` with a 64-bit mix of the entropy (or using a wider hash).

components/brave_shields/core/browser/brave_shields_utils.h
```diff
@@ -131,7 +132,9 @@ bool IsWebcompatEnabled(HostContentSettingsMap* map,
 
 mojom::FarblingLevel GetFarblingLevel(HostContentSettingsMap* map,
                                       const GURL& primary_url);
-base::Token GetFarblingToken(HostContentSettingsMap* map, const GURL& url);
+base::Token GetFarblingToken(HostContentSettingsMap* map,
+                             const GURL& url,
+                             base::span<const uint8_t> additional_entropy);
```
simonhong, line 137: nit: how about this makes optional?

browser/containers/containers_browsertest.cc
```diff
@@ -1806,6 +1813,45 @@ IN_PROC_BROWSER_TEST_F(ContainersBrowserTest,
...
+  params_a.disposition = WindowOpenDisposition::NEW_FOREGROUND_TAB;
+  params_a.storage_partition_config = content::StoragePartitionConfig::Create(
+      browser()->profile(), kContainersStoragePartitionDomain, "farbling-a",
+      browser()->profile()->IsOffTheRecord());
+  ui_test_utils::NavigateToURL(&params_a);
+
+  content::WebContents* web_contents_a =
+      browser()->tab_strip_model()->GetActiveWebContents();
+  ASSERT_TRUE(web_contents_a);
+  ASSERT_TRUE(content::WaitForLoadStop(web_contents_a));
```
simonhong: nit: it's redundant as `NavigateToURL()` wait till load stop?

browser/containers/containers_browsertest.cc
```diff
@@ -1806,6 +1813,45 @@ IN_PROC_BROWSER_TEST_F(ContainersBrowserTest,
...
+      browser()->tab_strip_model()->GetActiveWebContents();
+  ASSERT_TRUE(web_contents_b);
+  ASSERT_TRUE(content::WaitForLoadStop(web_contents_b));
+  const std::string plugins_b =
+      content::EvalJs(web_contents_b, kFarblingPluginsStringScript)
+          .ExtractString();
+
+  EXPECT_FALSE(plugins_a.empty());
+  EXPECT_FALSE(plugins_b.empty());
+  EXPECT_NE(plugins_a, plugins_b);
```
simonhong: q) If same container is used, both should have same farbling?

nit: not sure it's worth to check farblings are same if same container is used. it's up to you.

simonhong (approved):
++ with all nits.

components/brave_shields/core/browser/brave_shields_utils.h
goodov: don't think we need optional here, empty size should be enough to treat this as "not required".

browser/containers/containers_browsertest.cc
goodov: yep, adjusted the test to include all scenarios.

netzenbot (approved):

browser/net/brave_reduce_language_network_delegate_helper.cc
```diff
@@ -27,6 +29,13 @@
 #include "content/public/browser/browser_context.h"
 #include "net/base/net_errors.h"
 
+#if BUILDFLAG(ENABLE_CONTAINERS)
+#include "brave/components/containers/content/browser/storage_partition_utils.h"
+#include "brave/components/containers/core/common/features.h"
+#include "content/public/browser/render_frame_host.h"
```
bridiver, line 35: content/public headers don't belong behind `ENABLE_CONTAINERS`

PR #29736, opened 2025-06-24
fix(privacy): Move Brave Shields utility functions to shared location

StephenHeaps:
- Move the functions not relying on `//content/*` from `/components/brave_shields/content/browser/` to `/components/brave_shields/core/browser/` so iOS can re-use them as [iOS cannot access `//content/*`](https://source.chromium.org/chromium/chromium/src/+/main:ios/build/config.gni;l=9).
- The only code changes are resolving the `UNSAFE_TODO` in `brave_shields_p3a.cc`.
- Opened a follow up issue for addressing the `chrome/browser` dependency in the moved unit tests: https://github.com/brave/brave-browser/issues/47454

Resolves https://github.com/brave/brave-browser/issues/46966

PR #36475, opened 2026-05-15
Fix brave shields test utils location.

goodov:
Fix test utils file location. Production file was moved here https://github.com/brave/brave-core/pull/29736, but the test utils file was left behind.

The incorrect test utils file location breaks iOS build when included into `core/browser` tests.


mkarolin (approved):
++

netzenbot (approved):

PR #36476, opened 2026-05-15
Fix Remember Browsing Mode on iPad after scene restore.

### LINKED ISSUE #55515: Remember Browsing Mode setting is not respected [iPadOS]
### Description

Re-launching Brave on iPad with `Remember Browsing Mode` enabled will default to normal browsing mode. Works on iOS

Image

### Steps to reproduce

1. Switch to Private Browsing mode
2. Tap on `More` -> `Private Tab Settings`
3. Enable `Keep Private Tabs` and `Remember Browsing Mode`
4. Relaunch the browser
5. Observe

### Actual result

Opens in normal browsing mode

### Expected result

Should open in Private mode

### Reproduces how often

Easily reproduced

### Brave version

1.90 (122)

### Device/iOS version

Apple iPad (9th Gen) - iOS 18.5

### Affected browser versions

- [ ] latest AppStore
- [x] latest TestFlight
- [x] previous TestFlight

### Reproducibility

### Miscellaneous information

_No response_

EliHini:
Resolves https://github.com/brave/brave-browser/issues/55515

## Summary
On multi-scene devices, launch private state came only from `BrowserState.getNewWindowInfo`, which treats `isPrivate` as false when `scene/NSUserActivity` userInfo only has `WindowID` (we never persist isPrivate in BrowserState.userActivity(for:) / setWindowId for the session).

### DIFF
--- a/ios/brave-ios/App/iOS/Delegates/SceneDelegate.swift
+++ b/ios/brave-ios/App/iOS/Delegates/SceneDelegate.swift
@@ -756,8 +756,11 @@ extension SceneDelegate {
       {
         // Restore the scene from the User-Info WindowID
         windowId = windowUUID
-        isPrivate = windowInfo.isPrivate
-        privateBrowsingManager.isPrivateBrowsing = windowInfo.isPrivate
+        let isPrivateFromActivity = windowInfo.isPrivate
+        isPrivate =
+          isPrivateFromActivity
+          || self.shouldLaunchInPrivateMode(windowId: windowUUID)
+        privateBrowsingManager.isPrivateBrowsing = isPrivate
         urlToOpen = windowInfo.openURL
 
         // Create a new session window if it does not already exist


netzenbot (approved):

ios/brave-ios/App/iOS/Delegates/SceneDelegate.swift
```diff
@@ -756,8 +756,11 @@ extension SceneDelegate {
       {
         // Restore the scene from the User-Info WindowID
         windowId = windowUUID
-        isPrivate = windowInfo.isPrivate
-        privateBrowsingManager.isPrivateBrowsing = windowInfo.isPrivate
+        let isPrivateFromActivity = windowInfo.isPrivate
+        isPrivate =
+          isPrivateFromActivity
+          || self.shouldLaunchInPrivateMode(windowId: windowUUID)
+        privateBrowsingManager.isPrivateBrowsing = isPrivate
```
nuo-xu, line 763: should we use `shouldLaunchInPrivateMode` for both cases?

EliHini: Good question. We need both; we use the activity when it explicitly says private, otherwise use `shouldLaunchInPrivateMode`.

`shouldLaunchInPrivateMode` only looks at settings and saved tabs. It doesn’t know what the system passed in on the `NSUserActivity`.

For a new window created with “open in private” we put private mode in that activity. At scene delegate connect time there often aren’t tabs yet for that window, so `shouldLaunchInPrivateMode` alone can wrongly say “not private.”

After a normal relaunch, the activity usually only has the window id, not private mode, so I am relying on `shouldLaunchInPrivateMode` for Remember Browsing Mode.

nuo-xu (approved):

kjozwiak:
The above didn't fix the issue as per https://github.com/brave/brave-browser/issues/55515#issuecomment-4493004735. Also see https://github.com/brave/brave-core/pull/36535 & https://github.com/brave/brave-core/pull/36536 for more context.

PR #36478, opened 2026-05-15
[Desktop NTP]: Fix high-resolution custom backgrounds not displaying and loading indefinitely

### LINKED ISSUE #55551: [NTP] Uploading a high-resolution custom background images can make brave stuck and loading indefinitely
### Description

When uploading a **high-resolution** image (around 10k resolution or more) as a custom background on the New Tab Page, Brave can get stuck in a loading state.

The image picker closes and the upload flow starts, but the background is not applied and the New Tab Page does not recover from the loading state.

### Steps to reproduce

1. Open a new tab page.
2. Open the Customize New Tab Page and press upload from device
3. Select a high-resolution image, for example a 9216x5184 JPG.

### Actual result

The image is not applied as the custom background.

The New Tab Page stays in a loading state after selecting the image.

Image

### Expected result

Brave should either apply the selected image successfully, or fail cleanly and stop showing the loading state.

### Reproduces how often

Easily reproduced

Tested on both release and nightly build:
1.92.61 Chromium: 148.0.7778.167 (Official Build) nightly (64-bit) - Windows 11 Version 25H2 (Build 26200.8246)

### Channel information

- [x] release (stable)
- [ ] beta
- [x] nightly

### Reproducibility

### Miscellaneous information

- This was reproduced with a 9216x5184 JPG image and 10k wallpaper resolution from internet.
- Chrome and other chromium based browser is able to use the same image as a custom New Tab Page background.

From local debugging, Brave is able to read the selected image, but the decoded image is too large and hits the IPC maximum message size limit.

I opened a PR with a workaround so large images can still be displayed.

kisenaa:
Resolves https://github.com/brave/brave-browser/issues/55551

### Description
High-resolution custom background images can fail to save on Desktop when the decoded bitmap is larger than the Mojo IPC maximum message size.

From my local testing and debugging:
Brave read the high-res image successfully, but `DecodeImageIsolated` returned an empty decoded image because the full decoded bitmap exceeded `kChannelMaximumMessageSize`. An empty decoded image then gets passed to the function `SaveImageAsPNG` and returned early with `base::FilePath()` because it is empty.

This PR enables `shrink_to_fit` when decoding custom background images, allowing large images to be downscaled before being returned from the decoder.

### Notes
- `kChannelMaximumMessageSize` is defined in `ipc/constants.mojom`.
- `shrink_to_fit` parameter is explained on `Services\data_decoder\public\mojom\image_decoder.mojom`

I tested this with 9216x5184 and 10k-resolution images, and both displayed correctly after the change.

Please let me know if I missed anything or there are any mistakes. Thanks.

### Demonstration to show that the issue is fixed
https://github.com/user-attachments/assets/c515cd96-23c9-4673-8b47-b52b6825ab07

### DIFF
--- a/browser/ntp_background/custom_background_file_manager.cc
+++ b/browser/ntp_background/custom_background_file_manager.cc
@@ -173,7 +173,7 @@ void CustomBackgroundFileManager::DecodeImageInIsolatedProcess(
 
   data_decoder::DecodeImageIsolated(
       image_data_span, data_decoder::mojom::ImageCodec::kDefault,
-      /*shrink_to_fit=*/false,
+      /*shrink_to_fit=*/true,
       static_cast<uint64_t>(IPC::mojom::kChannelMaximumMessageSize),
       gfx::Size() /* No particular size desired. */,
       std::move(skbitmap_to_image).Then(std::move(on_decode)));
--- a/browser/ntp_background/custom_background_file_manager_browsertest.cc
+++ b/browser/ntp_background/custom_background_file_manager_browsertest.cc
@@ -21,12 +21,17 @@
 #include "content/public/test/browser_task_environment.h"
 #include "content/public/test/browser_test.h"
 #include "third_party/abseil-cpp/absl/strings/str_format.h"
+#include "third_party/skia/include/core/SkBitmap.h"
+#include "third_party/skia/include/core/SkPixmap.h"
+#include "ui/gfx/codec/png_codec.h"
 
 // CustomBackgroundFileManager requires data decoder which can't be initialized
 // in unit tests.
 class CustomBackgroundFileManagerBrowserTest : public InProcessBrowserTest {
  public:
   static constexpr char kTestImageName[] = "background.jpg";
+  static constexpr char kTestHighResolutionImageName[] =
+      "high_resolution_background.png";
 
   CustomBackgroundFileManagerBrowserTest() = default;
   ~CustomBackgroundFileManagerBrowserTest() override = default;
@@ -57,6 +62,36 @@ class CustomBackgroundFileManagerBrowserTest : public InProcessBrowserTest {
  protected:
   base::FilePath test_file() const {
     return profile()->GetPath().AppendASCII(kTestImageName);
+  }
+
+  bool CreateHighResolutionTestImage(base::FilePath* image_path) const {
+    constexpr int kWidth = 9216;
+    constexpr int kHeight = 5184;
+
+    SkBitmap bitmap;
+    if (!bitmap.tryAllocN32Pixels(kWidth, kHeight)) {
+      return false;
+    }
+
+    bitmap.eraseColor(SK_ColorRED);
+
+    SkPixmap pixmap;
+    if (!bitmap.peekPixels(&pixmap)) {
+      return false;
+    }
+
+    *image_path =
+        profile()->GetPath().AppendASCII(kTestHighResolutionImageName);
+
+    std::optional<std::vector<uint8_t>> encoded_png =
+        gfx::PNGCodec::EncodeBGRASkBitmap(bitmap,
+                                          /*discard_transparency=*/false);
+
+    if (!encoded_png) {
+      return false;
+    }
+
+    return base::WriteFile(*image_path, *encoded_png);
   }
 
   Profile* profile() { return browser()->profile(); }
@@ -110,6 +145,30 @@ IN_PROC_BROWSER_TEST_F(CustomBackgroundFileManagerBrowserTest,
       custom_file_manager().GetCustomBackgroundDirectory().AppendASCII(
           kTestImageName)));
   EXPECT_TRUE(base::PathExists(test_file()));
+}
+
+// Verifies that high-resolution images whose decoded bitmap would exceed the
+// Mojo IPC message size limit can still be saved after being scaled down by the
+// data decoder. See https://github.com/brave/brave-browser/issues/55551
+IN_PROC_BROWSER_TEST_F(CustomBackgroundFileManagerBrowserTest,
+                       SaveHighResolutionImageToCustomBackgroundDir) {
+  base::ScopedAllowBlockingForTesting allow_blocking_call;
+
+  base::FilePath high_resolution_image_path;
+  ASSERT_TRUE(CreateHighResolutionTestImage(&high_resolution_image_path));
+  ASSERT_TRUE(base::PathExists(high_resolution_image_path));
+
+  auto check_result = base::BindOnce([](const base::FilePath& path) {
+                        EXPECT_FALSE(path.empty());
+                      }).Then(run_loop_quit_closure());
+
+  custom_file_manager().SaveImage(high_resolution_image_path,
+                                  std::move(check_result));
+  Wait();
+
+  EXPECT_TRUE(base::PathExists(
+      custom_file_manager().GetCustomBackgroundDirectory().AppendASCII(
+          kTestHighResolutionImageName)));
 }
 
 IN_PROC_BROWSER_TEST_F(CustomBackgroundFileManagerBrowserTest,
--- a/browser/ntp_background/custom_background_file_manager_browsertest.cc
+++ b/browser/ntp_background/custom_background_file_manager_browsertest.cc
@@ -9,6 +9,7 @@
 
 #include "base/files/file_path.h"
 #include "base/files/file_util.h"
+#include "base/logging.h"
 #include "base/path_service.h"
 #include "base/test/bind.h"
 #include "base/test/task_environment.h"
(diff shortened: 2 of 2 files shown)


kisenaa:
I also added a browser test covering high-resolution custom background images in commit https://github.com/brave/brave-core/pull/36478/changes/8cbbad168de7c2ff476bba9910f6d3645aeb0e69. If this test is unnecessary, it can be removed or adjusted.

browser/ntp_background/custom_background_file_manager.cc
```diff
@@ -173,7 +173,7 @@ void CustomBackgroundFileManager::DecodeImageInIsolatedProcess(
 
   data_decoder::DecodeImageIsolated(
       image_data_span, data_decoder::mojom::ImageCodec::kDefault,
-      /*shrink_to_fit=*/false,
+      /*shrink_to_fit=*/true,
```
zenparsing, line 176: When `shrink_to_fit` is `true`, an oversized image will eventually get resized by [ResizeImage](https://source.chromium.org/chromium/chromium/src/+/main:services/data_decoder/image_decoder_impl.cc;l=27;drc=36976705fdc7f834f80b5448f14288980104471a;bpv=1;bpt=1), which will shrink the image by halving the height and width until it is less than the specified max size (`kChannelMaximumMessageSize` below).

sangwoo108: I think that's what the @kisenaa intended so that we don't get empty image?

browser/ntp_background/custom_background_file_manager_browsertest.cc
```diff
@@ -59,6 +64,36 @@ class CustomBackgroundFileManagerBrowserTest : public InProcessBrowserTest {
...
+  bool CreateHighResolutionTestImage(base::FilePath* image_path) const {
+    constexpr int kWidth = 9216;
+    constexpr int kHeight = 5184;
+
+    SkBitmap bitmap;
+    if (!bitmap.tryAllocN32Pixels(kWidth, kHeight)) {
+      return false;
+    }
+
+    bitmap.eraseColor(SK_ColorRED);
```
sangwoo108, line 79: Q. Are we initializing the bitmap with this color after the initial allocation?

browser/ntp_background/custom_background_file_manager_browsertest.cc
```diff
@@ -59,6 +64,36 @@ class CustomBackgroundFileManagerBrowserTest : public InProcessBrowserTest {
...
+    if (!bitmap.peekPixels(&pixmap)) {
+      return false;
+    }
+
+    *image_path =
+        profile()->GetPath().AppendASCII(kTestHighResolutionImageName);
+
+    std::optional<std::vector<uint8_t>> encoded_png =
+        gfx::PNGCodec::EncodeBGRASkBitmap(bitmap,
+                                          /*discard_transparency=*/false);
```
sangwoo108, line 88: Can you add comments about this line? Are we verifying that the decoded image is valid?

browser/ntp_background/custom_background_file_manager.cc
kisenaa: Yes, it seems the decoded bitmap size follows the formula: width × height × 4 (RGBA).

For example, an image with a resolution of 9216 × 5184 will produce a size of 191,102,976 bytes, which exceeds the maximum IPC message size limit (134,217,728). By setting `shrink_to_fit` to true, that image can still be processed and displayed.

browser/ntp_background/custom_background_file_manager_browsertest.cc
kisenaa: Yes, we set the bitmap to SK_ColorRED (red color). The specific color is not important and can be changed to any color.

It is needed to generate a valid test image on the fly instead of storing a large real image in the repository. However, I can change it to use a stored test image instead if preferred.

browser/ntp_background/custom_background_file_manager_browsertest.cc
kisenaa: > Can you add comments about this line? Are we verifying that the decoded image is valid?

Should I write the comment like this ? 
```C++
    // Encode the generated bitmap into a valid PNG image so it can be written
    // to disk and used during the browser test.
    std::optional<std::vector<uint8_t>> encoded_png = ...
```

The encoded_png result is already checked on lines 90–92:
```
if (!encoded_png) {
      return false;
    }
```
Would it also make sense to add a `VLOG(2)` before returning false for additional clarity?

kisenaa: I have added the comments in the latest commit: https://github.com/brave/brave-core/pull/36478/commits/deb31e8112fb734fe3cd360f204c280800953578

kisenaa:
Hello. I have addressed the review comments and changed `CreateHighResolutionTestImage` to return `std::optional` instead of using an out parameter in https://github.com/brave/brave-core/pull/36478/commits/deb31e8112fb734fe3cd360f204c280800953578. Hopefully this makes the code clearer. 

Can my changes be reviewed again? Thank you.

browser/ntp_background/custom_background_file_manager_browsertest.cc
sangwoo108: No problem. I was just curious if we should init the pixels after allocating memory. Looks good as is. Thanks

browser/ntp_background/custom_background_file_manager_browsertest.cc
```diff
@@ -59,6 +65,42 @@ class CustomBackgroundFileManagerBrowserTest : public InProcessBrowserTest {
     return profile()->GetPath().AppendASCII(kTestImageName);
   }
 
+  std::optional<base::FilePath> CreateHighResolutionTestImage() const {
```
sangwoo108: I think this function can fail due to various reasons. Instead of having LOG(ERROR) for each branch, how about having `base::Location location = base::Location::Current()` as parameter and `    SCOPED_TRACE(location.ToString());` ? Then we can have assertion instead of returning nullopt.

kisenaa: > I think this function can fail due to various reasons. Instead of having LOG(ERROR) for each branch, how about having `base::Location location = base::Location::Current()` as parameter and ` SCOPED_TRACE(location.ToString());` ? Then we can have assertion instead of returning nullopt.

I see. So based on my understanding the function should take a `base::Location` parameter and guard the function by adding `SCOPED_TRACE` at the beginning of the function?

Do you mean something like this?
`base::FilePath CreateHighResolutionTestImage(const base::Location& location = FROM_HERE)`
   We return the generated image path . If something fail, use assertions instead of returning an std::nullopt or empty path.

kisenaa: From https://google.github.io/googletest/reference/testing.html, `SCOPED_TRACE(message)` will causes the current file name, line number, and the given message message to be added to the failure message for each assertion failure that occurs in the scope.

So based on my undersanding, it is option 1 right? Using assertions instead  of std::nullopt

kisenaa: Ah my bad. After reading and doing some research from google gtest, `(FAIL* and ASSERT_*)` can only be used in void-returning functions. 

So i made the function to : `void CreateHighResolutionTestImage(base::FilePath* image_path_out, const base::Location& location = FROM_HERE)` . `image_path_out` is needed because we supply the `kTestHighResolutionImageName` path inside this helper func (similar style like `test_file`). 

Let me know if that's acceptable or i need to remove the image_path_out. I pushed the new one on commit https://github.com/brave/brave-core/pull/36478/commits/bf95e16467dd571a65ba5042e75e2cc5e91d56d4

kisenaa:
I have addressed the newest review on commit https://github.com/brave/brave-core/pull/36478/changes/bf95e16467dd571a65ba5042e75e2cc5e91d56d4 . Thank you

browser/ntp_background/custom_background_file_manager_browsertest.cc
sangwoo108: Thanks for digging! Looks good to me

sangwoo108:
Let me kick of CI! Thanks :)

---

Update: https://github.com/brave/brave-core/pull/36746

kisenaa:
Hello @sangwoo108 and @zenparsing, may I ask if there is any update regarding this PR when you have time? 

I believe I have addressed the latest review comments and the CI checks seems to have succeeded in https://github.com/brave/brave-core/pull/36746

Please let me know if anything else is needed from my side. Thank you for your time

kisenaa:
Hi, just wanted to follow up on this PR. I’d appreciate it if someone from the Brave team could take a look when they have a chance. Thank you.

sangwoo108:
Oh so sorry that I missed. Let me check today

PR #36480, opened 2026-05-15
Add code to debug Omaha 4 update failures on macOS

mherrmann:
Omaha 4 is currently enabled for automatic updates of Nightly and 25% of Beta. Roughly 1% of update attempts fail with exit code 12. This PR adds debug code that attempts two possible fixes and reports their effectiveness as new exit codes between 17 and 24.

# Test plan

### DIFF
--- a/patches/chrome-updater-mac-.install.sh.patch
+++ b/patches/chrome-updater-mac-.install.sh.patch
@@ -1,5 +1,5 @@
 diff --git a/chrome/updater/mac/.install.sh b/chrome/updater/mac/.install.sh
-index 3b1874d09fca0cf88e361927dfea7c4273790030..a1b6a4fb30123dfd26ff22aad6c5a208e9777f09 100755
+index 3b1874d09fca0cf88e361927dfea7c4273790030..f2d4a4c9eec54857b07cf6442b78535bc584006d 100755
 --- a/chrome/updater/mac/.install.sh
 +++ b/chrome/updater/mac/.install.sh
 @@ -322,7 +322,7 @@ main() {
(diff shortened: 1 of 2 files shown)


netzenbot (approved):

patches/chrome-updater-mac-.install.sh.patch
```diff
@@ -20,3 +20,27 @@ index 3b1874d09fca0cf88e361927dfea7c4273790030..a1b6a4fb30123dfd26ff22aad6c5a208
...
+     if ! rsync ${RSYNC_FLAGS} --delete-before "${update_versioned_dir}/" \
+                                               "${new_versioned_dir}"; then
+       err "rsync of versioned directory failed, status ${PIPESTATUS[0]}"
+@@ -463,7 +469,9 @@ main() {
+       # The incomplete version would break code signature validation.
+       note "cleaning up new_versioned_dir"
+       rm -rf "${new_versioned_dir}"
+-      exit 12
++      rsync ${RSYNC_FLAGS} --delete-before "${update_versioned_dir}" "${installed_versions_dir}" && (( debug_flags |= 4 ))
++      rm -rf "${new_versioned_dir}"
```
cdesouza-chromium, line 42:
```sh
rm -rf "${new_versioned_dir}"
rsync ${RSYNC_FLAGS} --delete-before "${update_versioned_dir}" "${installed_versions_dir}" && (( debug_flags |= 4 ))
rm -rf "${new_versioned_dir}"
```

Is this double call of `rm -fr` to the same path intentional?

rewrite/chrome/updater/mac/.install.sh.toml
```diff
@@ -26,3 +26,28 @@ replace = '${update_dmg_mount_point}/${PRODUCT_NAME}.app'
...
+      mkdir -p "${new_versioned_dir}" && (( debug_flags |= 2 ))
+      rm -rf "${new_versioned_dir}"
+    fi'''
+
+[[substitution]]
+description = 'Return debug_flags in exit code'
+pattern = 'exit 12'
+replace = '''rsync ${RSYNC_FLAGS} --delete-before "${update_versioned_dir}" "${installed_versions_dir}" && (( debug_flags |= 4 ))
+      rm -rf "${new_versioned_dir}"
+      exit $((17 + debug_flags))'''
```
cdesouza-chromium: My concern with matching `exit 12` here, is that these codes are being listed in that script based on the order they occur, so tomorrow someone comes a long and inserts a previous early exit, and shifts all the codes down, and now the substitution will match some other place. It would be better to use a regex that first matches the note `note "cleaning up new_versioned_dir"`, and then whatever is the next `exit` statement after the note, regardless of the value.

cdesouza-chromium (commented):
LGTM

patches/chrome-updater-mac-.install.sh.patch
mherrmann: Yes. The `rsync` may put contents in `new_versioned_dir`. I want a clean state after my debug code, as if it never ran. That's why I'm deleting the dir again.

cdesouza-chromium: Understood.

rewrite/chrome/updater/mac/.install.sh.toml
mherrmann: The exit code is reported in telemetry. For this reason, I expect that upstream keeps the codes constant. Does this alleviate your concern? I can also match on something else. Thank you.

cdesouza-chromium (approved):
LGTM

PR #36493, opened 2026-05-15
[iOS] Add Skus JavaScriptFeature

### LINKED ISSUE #55561: Add support for account linking on Brave Account pages when using single profile web view configuration
Add the JS that powers linking premium features using the skus service on Brave Account pages when the `UseProfileWebViewConfiguration` feature flag is enabled

kylehickinson:
Adds a JavaScriptFeature to handle the Brave Account premium shim layer with the SKUs service when the `UseProfileWebViewConfiguration` feature flag is enabled

Resolves https://github.com/brave/brave-browser/issues/55561

### DIFF
--- a/ios/browser/skus/skus_javascript_feature.mm
+++ b/ios/browser/skus/skus_javascript_feature.mm
@@ -1,0 +1,234 @@
+// Copyright (c) 2026 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#include "brave/ios/browser/skus/skus_javascript_feature.h"
+
+#include <optional>
+#include <string>
+#include <string_view>
+#include <utility>
+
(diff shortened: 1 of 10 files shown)


netzenbot (approved):

ios/browser/skus/skus_javascript_feature.mm
```diff
@@ -0,0 +1,235 @@
...
+
+void SkusJavaScriptFeature::OnMojoConnectionError() {
+  skus_service_.reset();
+}
+
+void SkusJavaScriptFeature::OnCredentialSummary(
+    const std::string domain,
+    ScriptMessageReplyCallback callback,
+    mojom::SkusResultPtr response) {
+  credential_summary_fetched_.Run(domain, response->message);
```
nuo-xu: i think we need a null check for `credential_summary_fetched_` here even though we probably call `SetCredentialSummaryFetched` before this.

kylehickinson: Good catch, fixed in amended push

nuo-xu (approved):

stoletheminerals (approved):

PR #36510, opened 2026-05-18
[brockit] Fix reassignment bug in rebase v2

cdesouza-chromium:
The regex used to parse the subcommand was not matching properly, and
therefore `reassign!` commits were being dropped.

Bug: https://github.com/brave/brave-browser/issues/55466


AlexeyBarabash (approved):
++

PR #36515, opened 2026-05-18
[include-what-you-use] `//components/brave_wallet/common/` fixes

cdesouza-chromium:
This is an experimental pass of `include-what-you-use`.

Bug: https://github.com/brave/brave-browser/issues/42212


components/brave_wallet/common/btc_like_serializer_stream.h
```diff
@@ -6,11 +6,22 @@
...
 #include "base/numerics/safe_conversions.h"
 
+namespace base {
+namespace numerics_internal {
+template <typename T>
+  requires std::is_arithmetic_v<T>
+class StrictNumeric;
+}  // namespace numerics_internal
+}  // namespace base
+
```
cmazakas: I think I agree with the AI summary here, I'm not a fan of introducing a forward declaration that contains an `internal` and what's more, duplicates the template constraint. I think the ideal here would be for upstream to provide a `fwd.h`.

cdesouza-chromium: Fixed.

components/brave_wallet/common/brave_wallet_response_helpers.h
```diff
@@ -6,13 +6,19 @@
...
 #include <string>
 
 #include "base/values.h"
 #include "brave/components/brave_wallet/common/brave_wallet.mojom.h"
 
 namespace brave_wallet {
+namespace mojom {
+enum class ProviderError : int32_t;
+enum class SolanaProviderError : int32_t;
+}  // namespace mojom
```
cmazakas, line 18: Hmm, I'm curious how to feel about forward declarations for mojom-generated entities. On one hand, it could be incredibly useful. On the other, it seems trivial that it'd potentially fall out of sync.

What's interesting is, I actually see the `brave_wallet.mojom.h` `#include` above, which makes me wonder if the tool just messed up here.

cdesouza-chromium: Fixed. `include-what-you-use` is doing the wrong thing here, and we are supposed to be using the forward inclusion header generated by mojom.

components/brave_wallet/common/DEPS
```diff
@@ -19,6 +19,7 @@ specific_include_rules = {
   "hash_utils\.cc": [
     "+brave/third_party/argon2",
     "+brave/third_party/bitcoin-core/src/src/crypto/ripemd160.h",
+    "!brave/third_party/ethash/src/include/ethash/hash_types.h",
```
netzenbot: The DEPS entry `"!brave/third_party/ethash/src/include/ethash/hash_types.h"` uses the `!` prefix which marks the include as forbidden/being-removed, but `hash_utils.cc` adds a new `#include` of that file. The presubmit bot flags this as an illegal include. Change `!` to `+` in the DEPS entry to actually allow the include. ([DEPS Allowlist Paths Must Exactly Match Include Paths](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-036))

components/brave_wallet/common/value_conversion_utils.h
```diff
@@ -6,18 +6,27 @@
...
 #include "base/values.h"
 #include "brave/components/brave_wallet/common/brave_wallet.mojom.h"
+#include "url/gurl.h"
 #include "url/origin.h"
 
+namespace url {
+class Origin;
+}  // namespace url
+
 namespace brave_wallet {
```
netzenbot, line 30: nit: The forward declaration `namespace url { class Origin; }` is redundant here — `url/origin.h` is already included above, so the type is fully defined. Per [CS-001](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-001), remove declarations you don't need. Drop the forward declare and keep the full include.

cdesouza-chromium: Fixed.

components/brave_wallet/common/DEPS
cdesouza-chromium: Fixed.

components/brave_wallet/common/eth_address.h
```diff
@@ -6,7 +6,13 @@
 #ifndef BRAVE_COMPONENTS_BRAVE_WALLET_COMMON_ETH_ADDRESS_H_
 #define BRAVE_COMPONENTS_BRAVE_WALLET_COMMON_ETH_ADDRESS_H_
 
+#include <stddef.h>
+#include <stdint.h>
```
supermassive, line 10: I'm curious what these provide.

supermassive (approved):
lgtm, Thanks!

Is it somehting that eventually would be enforced by CI and/or presubmit?

components/brave_wallet/common/eth_address.h
cdesouza-chromium: `<stddef.h>` is because of `size_t`, and `<stdint.h>` provides `uint8_t`.

cdesouza-chromium:
> lgtm, Thanks!
> 
> Is it somehting that eventually would be enforced by CI and/or presubmit?

I'm not sure if this thing works reliably enough for that, but I think in cases like your team, we can have the tooling checked out in the tree, so each team runs it on the regular to keep their trees in a good shape.

components/brave_wallet/common/hash_utils.cc
```diff
@@ -10,12 +10,14 @@
 
 #include "base/check_op.h"
 #include "base/containers/adapters.h"
+#include "base/containers/adapters_internal.h"
```
cmazakas: Hmm, I think the concerns raised by the AI are valid here. I double-checked the `adapters.h` header and it's basically the `_internal.h` include with one extra template in it, so I think we can just `#include "base/containers/adapters.h"` here.

components/brave_wallet/common/cardano_address.h
```diff
@@ -7,7 +7,6 @@
 #define BRAVE_COMPONENTS_BRAVE_WALLET_COMMON_CARDANO_ADDRESS_H_
 
 #include <stdint.h>
-
 #include <cstdint>
```
cmazakas: Ah, the AI is right here too. Technically, this PR didn't touch anything here but it does seem like we have redundant #includes here. If we wanna fix this in this PR that's fine, otherwise I think we can let it slide.

components/brave_wallet/common/cardano_address.cc
```diff
@@ -16,7 +17,8 @@
 #include "base/containers/span_rust.h"
 #include "base/containers/span_writer.h"
 #include "base/containers/to_vector.h"
-#include "base/notreached.h"
+#include "base/memory/raw_ptr.h"
```
cmazakas: Hmm, I think the AI is right. This is a dead `#include`. None of the code in this file seems to use a `raw_ptr`.

components/brave_wallet/common/encoding_utils.cc
```diff
@@ -5,14 +5,14 @@
...
 
-#include <utility>
-
 #include "base/check.h"
 #include "base/check_op.h"
 #include "base/containers/extend.h"
 #include "base/containers/span.h"
 #include "base/containers/span_writer.h"
 #include "base/containers/to_vector.h"
+#include "base/memory/raw_ptr.h"
```
cmazakas: This looks like a dead #include to me nothing in the source file seems to use `raw_ptr`. IWYU must be getting confused by something.

components/brave_wallet/common/hash_utils.cc
cdesouza-chromium: Done. Submitted a fix https://chromium-review.googlesource.com/c/chromium/src/+/7860407.

components/brave_wallet/common/cardano_address.h
cdesouza-chromium: Done.

components/brave_wallet/common/cardano_address.cc
cdesouza-chromium: The issue here is that `base::SpanWriter` is a templated type, and this leads to a whole thread being pulled back to this header. Removing for now, but I need to think of something better to deal with this.

components/brave_wallet/common/encoding_utils.cc
cdesouza-chromium: Same.

components/brave_wallet/common/cardano_address.cc
cmazakas: Oh, interesting.

cmazakas (approved):

netzenbot (approved):

goodov (approved):

PR #36516, opened 2026-05-18
Fixes search engines icons in the Settings.

### LINKED ISSUE #55619: Fix search engine icons in the Settings
Search engine icons are shown as generic globes in the Settings -> Search Engines -> Normal | Private Window -> Change.

This is because the component that's used to show the icons (`site-favicon`) is using `chrome://favicon2` to retrieve the icons and upstream switched the handler for that to only get the icons from the history service. This means the icons only become available after the site has been visited.

Since then upstream switched to using a different component (`settings-search-engine-icon`) which is using `chrome://image` and retrieves the icons on the fly. We should also switch to using that component.

We do have the icons available in Nala so at some point we may want to consider getting the icons from there instead.

STR/ QA Steps:

1. Start Brave with a fresh profile
2. Navigate to brave://settings/search
3. Click on Change in the Normal Window section

Observe: the icons for search engines are all generic globes
Expected: the actual search engines favicons are loaded

4. Close the dialog
5. Repeat the step 3 with the Private Window section.

6. Click on "Manage search engines and site search"
7. Add a new search engine (e.g Yahoo) and set it as default

Expected: the new search engine favicon is shown.

**Note**: ecosia's favicon URL is stale upstream, so the icon can't be loaded for it.

mkarolin:
Replaces site-favicon component - which only retrieves favicons from the history service, which means the user would first have to visit the site for the favicon to be available - with settings-search-engine-icon which retrieves the favicon right away.

Resolves https://github.com/brave/brave-browser/issues/55619

### DIFF
--- a/browser/resources/settings/brave_search_engines_page/brave_search_engines_page.html
+++ b/browser/resources/settings/brave_search_engines_page/brave_search_engines_page.html
@@ -13,7 +13,7 @@
     margin-inline-end: auto;
   }
 
-  site-favicon {
+  settings-search-engine-icon {
     --site-favicon-border-radius: 4px;
     --site-favicon-height: var(--favicon-size);
     --site-favicon-width: var(--favicon-size);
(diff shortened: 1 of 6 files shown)


fallaciousreasoning (approved):
lgtm assuming the sec/privacy team is okay with us fetching these icons from the search engines (or if we're reusing an icon distributed with the browser)

mkarolin:
> lgtm assuming the sec/privacy team is okay with us fetching these icons from the search engines (or if we're reusing an icon distributed with the browser)

Thanks, @fallaciousreasoning! I'll add sec/privacy team to the reviewers.

diracdeltas (approved):
IIUC this has some privacy impact (leaks to a network observer what search engines are loaded in the settings page) but i think they're the same for basically all users? so the impact seems fine

emerick (approved):

PR #36518, opened 2026-05-18
[Polkadot] [Wallet] simplify handling of chain metadata

cmazakas:
Update how we handle chain metadata by removing the FromFields factory function and instead opt into permitting direct accesses to the CxxPolkadotChainMetadata structure. We also removed the CxxPolkadotChainMetadataFields structure and instead return it directly from the Rust functions responsible for parsing the encoded chain metadata.

This is motivated by the number of fields attached the metadata growing substantially over time.

Intended as a cleanup/CodeHealth PR that https://github.com/brave/brave-core/pull/36419 can be rebased against.

### DIFF
--- a/components/brave_wallet/browser/internal/polkadot_chain_metadata.rs
+++ b/components/brave_wallet/browser/internal/polkadot_chain_metadata.rs
@@ -3,90 +3,9 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this file,
 // You can obtain one at https://mozilla.org/MPL/2.0/.
 
-use crate::CxxPolkadotChainMetadata;
-use crate::Error;
+use crate::{CxxPolkadotChainMetadata, CxxPolkadotChainMetadataResult, Error};
 use parity_scale_codec::{Compact, Decode};
 use std::collections::HashMap;
-
-#[derive(Clone, Copy)]
-struct CxxPolkadotChainMetadataFields {
-    system_pallet_index: u8,
-    balances_pallet_index: u8,
-    transaction_payment_pallet_index: u8,
-    transfer_allow_death_call_index: u8,
-    transfer_keep_alive_call_index: u8,
-    transfer_all_call_index: u8,
-    ss58_prefix: u16,
-    spec_version: u32,
-    asset_tx_payment: bool,
-}
-
-crate::impl_result!(CxxPolkadotChainMetadataFields, CxxPolkadotChainMetadataFieldsResult);
-
-#[cxx::bridge(namespace = brave_wallet)]
-mod ffi {
-    extern "Rust" {
-        type CxxPolkadotChainMetadataFieldsResult;
-
-        fn parse_chain_metadata_from_scale(
-            metadata_bytes: &[u8],
-        ) -> Box<CxxPolkadotChainMetadataFieldsResult>;
-        fn is_ok(self: &CxxPolkadotChainMetadataFieldsResult) -> bool;
-        fn error_message(self: &CxxPolkadotChainMetadataFieldsResult) -> String;
-        fn unwrap(
-            self: &mut CxxPolkadotChainMetadataFieldsResult,
-        ) -> Box<CxxPolkadotChainMetadataFields>;
-
-        type CxxPolkadotChainMetadataFields;
-        fn system_pallet_index(self: &CxxPolkadotChainMetadataFields) -> u8;
-        fn balances_pallet_index(self: &CxxPolkadotChainMetadataFields) -> u8;
-        fn transaction_payment_pallet_index(self: &CxxPolkadotChainMetadataFields) -> u8;
-        fn transfer_allow_death_call_index(self: &CxxPolkadotChainMetadataFields) -> u8;
-        fn transfer_keep_alive_call_index(self: &CxxPolkadotChainMetadataFields) -> u8;
-        fn transfer_all_call_index(self: &CxxPolkadotChainMetadataFields) -> u8;
-        fn ss58_prefix(self: &CxxPolkadotChainMetadataFields) -> u16;
-        fn spec_version(self: &CxxPolkadotChainMetadataFields) -> u32;
-        fn asset_tx_payment(self: &CxxPolkadotChainMetadataFields) -> bool;
-    }
-}
-
-impl CxxPolkadotChainMetadataFields {
-    fn system_pallet_index(self: &CxxPolkadotChainMetadataFields) -> u8 {
-        self.system_pallet_index
-    }
-
-    fn balances_pallet_index(self: &CxxPolkadotChainMetadataFields) -> u8 {
-        self.balances_pallet_index
-    }
-
-    fn transaction_payment_pallet_index(self: &CxxPolkadotChainMetadataFields) -> u8 {
-        self.transaction_payment_pallet_index
-    }
-
-    fn transfer_allow_death_call_index(self: &CxxPolkadotChainMetadataFields) -> u8 {
-        self.transfer_allow_death_call_index
-    }
-
-    fn transfer_keep_alive_call_index(self: &CxxPolkadotChainMetadataFields) -> u8 {
-        self.transfer_keep_alive_call_index
-    }
-
-    fn transfer_all_call_index(self: &CxxPolkadotChainMetadataFields) -> u8 {
-        self.transfer_all_call_index
-    }
-
-    fn ss58_prefix(self: &CxxPolkadotChainMetadataFields) -> u16 {
-        self.ss58_prefix
-    }
-
-    fn spec_version(self: &CxxPolkadotChainMetadataFields) -> u32 {
-        self.spec_version
-    }
-
-    fn asset_tx_payment(self: &CxxPolkadotChainMetadataFields) -> bool {
-        self.asset_tx_payment
-    }
-}
 
 fn decode_scale<T: Decode>(input: &mut &[u8]) -> Result<T, Error> {
     T::decode(input).map_err(|_| Error::InvalidScale)
@@ -517,21 +436,8 @@ fn parse_chain_metadata_fields(bytes: &[u8]) -> Result<CxxPolkadotChainMetadata,
     })
 }
 
-fn parse_chain_metadata_from_scale(
+pub(super) fn parse_chain_metadata_from_scale(
     metadata_bytes: &[u8],
-) -> Box<CxxPolkadotChainMetadataFieldsResult> {
-    let fields = parse_chain_metadata_fields(metadata_bytes).map(|metadata| {
-        CxxPolkadotChainMetadataFields {
-            system_pallet_index: metadata.system_pallet_index,
-            balances_pallet_index: metadata.balances_pallet_index,
-            transaction_payment_pallet_index: metadata.transaction_payment_pallet_index,
-            transfer_allow_death_call_index: metadata.transfer_allow_death_call_index,
-            transfer_keep_alive_call_index: metadata.transfer_keep_alive_call_index,
-            transfer_all_call_index: metadata.transfer_all_call_index,
-            ss58_prefix: metadata.ss58_prefix,
-            spec_version: metadata.spec_version,
-            asset_tx_payment: metadata.asset_tx_payment,
-        }
-    });
-    Box::new(CxxPolkadotChainMetadataFieldsResult(fields))
+) -> Box<CxxPolkadotChainMetadataResult> {
+    Box::new(CxxPolkadotChainMetadataResult(parse_chain_metadata_fields(metadata_bytes)))
 }
--- a/components/brave_wallet/browser/polkadot/polkadot_chain_metadata.cc
+++ b/components/brave_wallet/browser/polkadot/polkadot_chain_metadata.cc
@@ -5,12 +5,12 @@
 
 #include "brave/components/brave_wallet/browser/polkadot/polkadot_chain_metadata.h"
 
-#include "brave/components/brave_wallet/browser/internal/polkadot_chain_metadata.rs.h"
-
 namespace brave_wallet {
 
-PolkadotChainMetadata::PolkadotChainMetadata(const PolkadotChainMetadata& other)
-    : chain_metadata_(other.chain_metadata_) {}
+PolkadotChainMetadata::PolkadotChainMetadata() = default;
+
+PolkadotChainMetadata::PolkadotChainMetadata(
+    const PolkadotChainMetadata& other) = default;
 
 PolkadotChainMetadata::PolkadotChainMetadata(PolkadotChainMetadata&&) noexcept =
     default;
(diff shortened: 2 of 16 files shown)


components/brave_wallet/browser/polkadot/polkadot_chain_metadata.h
```diff
@@ -60,21 +48,8 @@ class PolkadotChainMetadata {
...
-  // Returns the ss58 prefix used by the chain when encoding/working with ss58
-  // addresses.
-  uint16_t GetSs58Prefix() const;
-  uint32_t GetSpecVersion() const;
-
-  // Returns true if the runtime uses ChargeAssetTxPayment as a signed
-  // extension, meaning the signature payload must include an asset_id field.
-  bool UsesAssetTxPayment() const;
+  CxxPolkadotChainMetadata* operator->();
+  const CxxPolkadotChainMetadata* operator->() const;
```
cdesouza-chromium, line 63: https://google.github.io/styleguide/cppguide.html#trailing_return

Would you mind changing the trailing return types to follow the guideline?

cdesouza-chromium (approved):
LGTM % style guide suggestion.

netzenbot (approved):

components/brave_wallet/browser/polkadot/polkadot_chain_metadata.h
cypt4, line 15: The main issue here is that we are sharing internal header in non-internal header file. And AFAIR there were some issues with such rust-related header files. @bridiver could give more information.

components/brave_wallet/browser/internal/polkadot_chain_metadata.rs
```diff
@@ -3,91 +3,10 @@
...
 // You can obtain one at https://mozilla.org/MPL/2.0/.
 
-use crate::CxxPolkadotChainMetadata;
-use crate::Error;
+use crate::{CxxPolkadotChainMetadata, CxxPolkadotChainMetadataResult, Error};
 use parity_scale_codec::{Compact, Decode};
 use std::collections::HashMap;
 
-#[derive(Clone, Copy)]
-struct CxxPolkadotChainMetadataFields {
```
cypt4, line 12: I'd say if we are doing a refactoring here we should make this structure more layered. Like presenting PalletInfo and CallInfo structs.

components/brave_wallet/browser/polkadot/polkadot_chain_metadata.h
cmazakas: Hmm, this `#include` isn't new, actually, and we pass `npm run gn_check`.

I guess technically we are leaking the definition of the CxxPolkadotChainMetadata struct via `operator->()` but in this case, I'd argue it's a feature and if we really needed to, we can always hide it somehow later.

components/brave_wallet/browser/internal/polkadot_chain_metadata.rs
cmazakas: That's not a bad idea. Currently, I like this PR as it is because it keeps the diffs readable and somewhat contained, so we can trivially audit the code for any breaking changes.

If we're going this way, I'd like to do it as a follow-up PR.

antonok-edm (approved):
rust ++

components/brave_wallet/browser/polkadot/polkadot_chain_metadata.h
bridiver: After discussing this with @cmazakas I think it's reasonable to make an exception to the "no public rust headers" for things like structs that only contain primitive types. No heap allocation and nothing in the built-in bindings list https://cxx.rs/bindings.html#built-in-bindings-reference

bridiver: std::array of primitive types is probably also ok so that would be the only exception I can think of from the built in bindings

components/brave_wallet/browser/polkadot/polkadot_chain_metadata.h
cmazakas: I've added a clarifying comment here. This isn't trailing return syntax, it's an overload for the arrow operator so that we can ergonomically access the data members of the underlying struct. Normally, I would've just done some public inheritance of the POD type but the cxx crate marks all generated struct as `final`.

components/brave_wallet/browser/polkadot/polkadot_chain_metadata.cc
```diff
@@ -77,52 +47,47 @@ constexpr uint32_t kUnknownSpecVersion = 0;
 // static
 std::optional<PolkadotChainMetadata> PolkadotChainMetadata::FromChainName(
```
cypt4: This method is used only in tests, consider moving it to the polkadot_test_utils

components/brave_wallet/browser/polkadot/polkadot_metadata_provider_unittest.cc
```diff
@@ -117,27 +117,31 @@ TEST_F(PolkadotMetadataProviderUnitTest, InitInitialization) {
...
   ASSERT_TRUE(mainnet_metadata.has_value());
-  auto expected_mainnet = PolkadotChainMetadata::FromFields(
-      /*system_pallet_index=*/0, /*balances_pallet_index=*/5,
-      /*transaction_payment_pallet_index=*/0x20,
-      /*transfer_allow_death_call_index=*/0,
-      /*transfer_keep_alive_call_index=*/3,
-      /*transfer_all_call_index=*/4,
-      /*ss58_prefix=*/0, /*spec_version=*/2'000'007,
-      /*asset_tx_payment=*/false);
+  PolkadotChainMetadata expected_mainnet;
```
cypt4: Can we keep FromFields method but move it to polkadot_test_utils?
Otherwise something could be missed in tests if new args are added,

cmazakas: I asked Claude and Claude had an amazing suggestion here. Structured bindings!

If we use structured bindings we enforce handling of every single data member because the underlying struct is an aggregate type. I then had Claude update the unit tests to do a member-wise comparison, and then I hand-updated it to keep the broader `operator==` comparisons so that way our `operator==` stays in sync as well.

I think this is the best approach here.

netzenbot (approved):

components/brave_wallet/browser/polkadot/polkadot_metadata_provider_unittest.cc
cypt4: There a are still some [places](https://github.com/brave/brave-core/pull/36518/changes#diff-9f9849ea675ccb206b8181c61f826b64832b6531df04a204f9a111d43d3f01fdR233) in tests where PolkadotChainMetadata initialized with default value and with field initialization list. Can we use this approach so every PolkadotChainMetadata initialization in tests uses this approach or FromFields method to validate all fields are initialized?

components/brave_wallet/browser/polkadot/polkadot_chain_metadata.cc
cmazakas: Done!

components/brave_wallet/browser/polkadot/polkadot_metadata_provider_unittest.cc
cmazakas: I only updated the 4 fixture parsing tests, as those are the most important ones. I think the rest of the tests will be fine so long as we're diligent in code review and using LLM tooling, which is really good about actually updating all the tests.

The tests we absolutely 112% must get correct are the ones that parse from Prefs and from the metadata, and I think we've covered both of those cases. We've covered prefs, the metadata provider and the bytes-parsing.

cypt4: LLM tooling depends on context resolution that depends on agent environment(cursor, claude code), free context size and model complexity and isn't perfect. I'd prefer to see compilation errors if we forget to add some fields to some metadata in tests. Otherwise we can have false-positive tests that use default-initialized fields. For example 
[here](https://github.com/brave/brave-core/pull/36518/changes#diff-fe8c42d0947c5942a8dc5082f5a92f663e02d64f9714539cbd1398048e7b412bR38)  westend metadata could have missing pallet info and default one will be used. This should not take much effort to ask LLM to update all PolkadotMetadata initialisation in tests to use the same manner as https://github.com/brave/brave-core/pull/36518/changes#diff-31865f84436bc104d9606f8f1e0e74d68db34583b20afb0289e132c108d36322R271.

components/brave_wallet/browser/polkadot/polkadot_metadata_provider_unittest.cc
```diff
@@ -114,31 +113,84 @@ TEST_F(PolkadotMetadataProviderUnitTest, InitInitialization) {
...
-  EXPECT_EQ(*mainnet_metadata, expected_mainnet);
+  {
+    const auto& [system_pallet_index, balances_pallet_index,
+                 transaction_payment_pallet_index,
+                 transfer_allow_death_call_index,
+                 transfer_keep_alive_call_index, transfer_all_call_index,
+                 ss58_prefix, spec_version, asset_tx_payment] =
+        **mainnet_metadata;
+
+    EXPECT_EQ(system_pallet_index, expected_mainnet->system_pallet_index);
```
cypt4: Why we are just not using   EXPECT_EQ(*mainnet_metadata, expected_mainnet); here and below?

components/brave_wallet/browser/polkadot/polkadot_metadata_provider_unittest.cc
```diff
@@ -267,20 +426,67 @@ TEST_F(PolkadotMetadataProviderUnitTest, DataIsSavedToCache) {
   provider_->GetChainMetadata(mojom::kPolkadotMainnet, future2.GetCallback());
   auto result2 = future2.Take();
   ASSERT_TRUE(result2.has_value());
-  EXPECT_EQ(*result2, *result1);
+  {
```
cypt4: The same, why we can't use equals here?

components/brave_wallet/browser/polkadot/polkadot_chain_metadata_unittest.cc
```diff
@@ -21,120 +22,90 @@ constexpr char kResult[] = "result";
...
-  auto expected_westend = PolkadotChainMetadata::FromFields(
-      /*system_pallet_index=*/0, /*balances_pallet_index=*/4,
-      /*transaction_payment_pallet_index=*/0x1a,
-      /*transfer_allow_death_call_index=*/0,
-      /*transfer_keep_alive_call_index=*/3,
-      /*transfer_all_call_index=*/4,
-      /*ss58_prefix=*/42, /*spec_version=*/0,
-      /*asset_tx_payment=*/false);
-  auto westend = PolkadotChainMetadata::FromChainName("Westend");
+  PolkadotChainMetadata expected_westend;
```
cypt4: I think we do not need this test at all since FromChainName is now a test_utils method.

components/brave_wallet/browser/polkadot/polkadot_extrinsic_unittest.cc
```diff
@@ -34,47 +35,59 @@ constexpr uint8_t kSchnorrkelSeed[] = {
...
 PolkadotChainMetadata MakeWestendMetadata() {
-  return PolkadotChainMetadata::FromFields(
-      /*system_pallet_index=*/0, /*balances_pallet_index=*/4,
-      /*transaction_payment_pallet_index=*/0x1a,
-      /*transfer_allow_death_call_index=*/0,
-      /*transfer_keep_alive_call_index=*/3,
-      /*transfer_all_call_index=*/4,
-      /*ss58_prefix=*/42, kSpecVersion,
-      /*asset_tx_payment=*/false);
+  PolkadotChainMetadata metadata;
```
cypt4: Do we still need this methods? We have PolkadotMetadataFromChainName in polkadot_test_utils.

components/brave_wallet/browser/polkadot/polkadot_metadata_provider_unittest.cc
```diff
@@ -114,31 +113,84 @@ TEST_F(PolkadotMetadataProviderUnitTest, InitInitialization) {
...
+    EXPECT_EQ(transfer_keep_alive_call_index,
+              expected_mainnet->transfer_keep_alive_call_index);
+    EXPECT_EQ(transfer_all_call_index,
+              expected_mainnet->transfer_all_call_index);
+    EXPECT_EQ(ss58_prefix, expected_mainnet->ss58_prefix);
+    EXPECT_EQ(spec_version, expected_mainnet->spec_version);
+    EXPECT_EQ(asset_tx_payment, expected_mainnet->asset_tx_payment);
+
+    EXPECT_EQ(mainnet_metadata, expected_mainnet);
+  }
```
supermassive: It is effectively `operator ==` and we should not test it everywhere. We should either rely on default implementation and don't test at all, or test it once if there is no default for some reason

components/brave_wallet/browser/polkadot/polkadot_metadata_provider_unittest.cc
```diff
@@ -114,31 +113,84 @@ TEST_F(PolkadotMetadataProviderUnitTest, InitInitialization) {
...
+  expected_mainnet->system_pallet_index = 0;
+  expected_mainnet->balances_pallet_index = 5;
+  expected_mainnet->transaction_payment_pallet_index = 0x20;
+  expected_mainnet->transfer_allow_death_call_index = 0;
+  expected_mainnet->transfer_keep_alive_call_index = 3;
+  expected_mainnet->transfer_all_call_index = 4;
+  expected_mainnet->ss58_prefix = 0;
+  expected_mainnet->spec_version = 2'000'007;
+  expected_mainnet->asset_tx_payment = false;
+
```
supermassive, line 118: - This(and similar ones) should be moved moved to polkadot_test_utils as `PolkadotMainnetChainMetadata`(or so). Should be close to production values as much as possible.
- All other tests which need 'some' metadata should pick a predefined one from test utils. Again, tests should use production metadata.
- If a test needs to expect somtheing from metadata with some non-default value of, for example, `transfer_all_call_index = 777;` let's clone `PolkadotMainnetChainMetadata()` and override `transfer_all_call_index`.

components/brave_wallet/browser/polkadot/polkadot_extrinsic_unittest.cc
cmazakas: To minimize the diff, I updated these functions to just be thin wrappers around the test_utils methods.

components/brave_wallet/browser/polkadot/polkadot_chain_metadata_unittest.cc
cmazakas: I've updated the code to instead test that our factory functions generate output matching what we actually parse from the fixture metadata bytes.

components/brave_wallet/browser/polkadot/polkadot_metadata_provider_unittest.cc
cmazakas: Yeah, good points. I think I overthought this. I've cleaned up the code dramatically.

components/brave_wallet/browser/polkadot/polkadot_metadata_provider_unittest.cc
cmazakas: Overthought it, wrote verbose code for no reason. This is cleaned up.

components/brave_wallet/browser/polkadot/polkadot_metadata_provider_unittest.cc
cmazakas: I updated the code to double-check the struct was `#[derive(PartialEq)]` on the Rust-side and then we use `bool operator==() = default;` on the C++ side, so this is done.

Good point about this.

components/brave_wallet/browser/polkadot/polkadot_metadata_provider_unittest.cc
cmazakas: We can; this has been subsequently cleaned up.

components/brave_wallet/browser/polkadot/polkadot_metadata_provider_unittest.cc
cmazakas: I re-added FromFields back. These structured bindings are  terrible to read and more importantly, they rely on impl details of cxx that aren't present in crubit, which is something upstream uses that we're likely to migrate to as well.

netzenbot (approved):

cypt4 (approved):

supermassive (approved):

PR #36519, opened 2026-05-18
[plaster] Adding `token` to `[[substittution]]`

cdesouza-chromium:
This PR introduces `token` as an easy shortcut for cases where the match
is supposed to be for a specific token. This is to avoid the trap of
using `pattern` in cases where it can cause mismatches, and also to
provide guidance on how to approach such cases.


docs/plaster.md
```diff
@@ -129,6 +139,16 @@ index 00000000000..63703e6940b
 
 The produced patch is supposed to be committed in the repository as usual.
 
+You can run `apply` for specific plaster files or patches, as below:
+
+```sh
+# With the plaster file
+tools/cr/plaster.py apply rewrite/net/cookies/cookie_monster.cc.toml
+
+# With the patch
```
netzenbot, line 148: nit: The two `apply` examples are identical — both show `rewrite/net/cookies/cookie_monster.cc.toml`. The second example is labeled `# With the patch` but should show the patch file path (e.g., `patches/net-cookies-cookie_monster.cc.patch`) to actually demonstrate the difference. ([Update Documentation Alongside Code Changes](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-001))

PR #36520, opened 2026-05-18
[AI Chat]: Infer `isGlobalPanel` from `isStandalone`

### LINKED ISSUE #55536: [AI Chat]: Remove `isGlobalPanel` and `/tab` from the sidebar - this should be inferred from `isStandalone === false`
This is made somewhat more difficult given `isStandalone` is async on `SetChatUI`

fallaciousreasoning:
Series https://github.com/brave/brave-browser/issues/53416
Resolves https://github.com/brave/brave-browser/issues/55536

### DIFF
--- a/components/ai_chat/resources/page/state/ai_chat_context.tsx
+++ b/components/ai_chat/resources/page/state/ai_chat_context.tsx
@@ -9,6 +9,7 @@ import { loadTimeData } from '$web-common/loadTimeData'
 import useMediaQuery from '$web-common/useMediaQuery'
 import * as Mojom from '../../common/mojom'
 import { AIChatAPI } from '../api/ai_chat_api'
+import { tabAssociatedChatId } from './active_chat_context'
 
 export interface ConversationEntriesProps {
   /* Class to put on the element in order to customize scroll */
@@ -30,6 +31,9 @@ type AIChatContextProps = {
 export function useIsSmall() {
   return useMediaQuery('(max-width: 1024px)')
 }
+
+const initiallyTabAssociated =
+  window.location.pathname.substring(1) === tabAssociatedChatId
 
 export default function useProvideAIChatContext(props: AIChatContextProps) {
   // This hook should only have any state or anything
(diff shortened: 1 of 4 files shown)


components/ai_chat/resources/page/state/ai_chat_context.tsx
```diff
@@ -31,6 +32,8 @@ export function useIsSmall() {
   return useMediaQuery('(max-width: 1024px)')
 }
 
+const initiallyTabAssociated = window.location.pathname === tabAssociatedChatId
```
netzenbot: `window.location.pathname` returns `'/tab'` but `tabAssociatedChatId` is `'tab'` (no leading slash), so this comparison is always `false` and `initiallyTabAssociated` is never `true`. Either use `'/' + tabAssociatedChatId` or `window.location.pathname.slice(1) === tabAssociatedChatId`.

netzenbot (approved):

components/ai_chat/resources/page/state/ai_chat_context.tsx
```diff
@@ -31,6 +32,9 @@ export function useIsSmall() {
   return useMediaQuery('(max-width: 1024px)')
 }
 
+const initiallyTabAssociated =
+  window.location.pathname.substring(1) === tabAssociatedChatId
+
```
petemill: I don't love this because nothing outside `ActiveChatProviderFromUrl` knows about the url schema. And that's good for portability.

Do you have an idea for another way? Perhaps it can come from the `ActiveChatContext`?

fallaciousreasoning: Move that to `ai_chat_context` and import it here now

petemill: Can it not come from the context itself? Otherwise it's difficult to mock or give a different value in a portable version

petemill (approved):

petemill:
needs format fix

components/ai_chat/resources/page/state/ai_chat_context.tsx
fallaciousreasoning: I think having it here is fine - its only used to calculate `isGlobalPanel`, which you can just mock directly on the context.

fallaciousreasoning: I might merge this as-is, as it unblocks iOS but if you want it in the context we can do it as a followup

PR #36523, opened 2026-05-19
Fix adblock subscriptions for secondary profile startup

### LINKED ISSUE #55623: Custom adblock filter lists do not load when Brave starts with only a secondary profile
## Description

Custom filter list subscriptions added from `brave://settings/shields/filters` can fail to load when Brave is launched directly into a secondary regular profile, without opening the initial/default profile.

In that startup order, saved custom subscriptions may not be shown in settings, registered as filter providers, updated, or applied to page requests.

This appears to affect custom filter list subscriptions added from:

`brave://settings/shields/filters` → `Content Filters` → `Add custom filter lists`

Related issues:
- https://github.com/brave/brave-browser/issues/27594
- https://github.com/brave/brave-browser/issues/30323

Related community reports:
- https://community.brave.app/t/cant-add-a-filter-list-to-brave-shields/652264
- https://community.brave.app/t/having-multiple-profile-cause-custom-filter-lists-to-break/636232

## Minimal local repro fixture

The following creates a small local custom filter list and a test page. The custom filter list blocks `blocked.svg`. If the red SVG is visible on the page, the custom subscription rule is not being applied.

Create the test files:

```bash
mkdir -p /tmp/brave-filter-repo
cd /tmp/brave-filter-repo

cat > list.txt <<'EOF'
[Adblock Plus 2.0]
||asset.brave-repro.test^$image
EOF

cat > page.html <<'EOF'
<!doctype html>
<meta charset="utf-8">
<title>Brave custom subscription repro</title>
<h1>Custom filter list test</h1>
<p>If the red image below loads, the custom subscription is not applied.</p>
<img src="http://asset.brave-repro.test:8765/blocked.svg" width="240" height="120">
EOF

cat > blocked.svg <<'EOF'
<svg xmlns="http://www.w3.org/2000/svg" width="240" height="120">
  <rect width="240" height="120" fill="red"/>
  <text x="20" y="65" font-size="20" fill="white">SHOULD BE BLOCKED</text>
</svg>
EOF
```

Start a local HTTP server:

```bash
cd /tmp/brave-filter-repo
python3 -m http.server 8765 --bind 127.0.0.1
```

Verify the server is reachable:

```bash
curl -I http://127.0.0.1:8765/list.txt
curl -I http://127.0.0.1:8765/page.html
curl -I http://127.0.0.1:8765/blocked.svg
```

## Steps to reproduce

1. Start Brave with a clean or dedicated user data dir and host resolver rules:

   ```bash
   brave \
     --user-data-dir=/tmp/brave-filter-repro-user-data \
     --host-resolver-rules="MAP *.brave-repro.test 127.0.0.1" \
     --no-first-run
   ```

   For a local Brave development build, this may look like:

   ```bash
   ../out/Component/brave \
     --user-data-dir=/tmp/brave-filter-repro-user-data \
     --host-resolver-rules="MAP *.brave-repro.test 127.0.0.1" \
     --no-first-run
   ```

2. In the initial/default profile, open:

   ```text
   brave://settings/shields/filters
   ```

3. Under `Content Filters` → `Add custom filter lists`, add:

   ```text
   http://list.brave-repro.test:8765/list.txt
   ```

4. Confirm the custom filter list appears in settings.

5. Open the test page:

   ```text
   http://page.brave-repro.test:8765/page.html
   ```

6. Confirm that `blocked.svg` is blocked. The red SVG should not be visible.

7. Create a second regular profile.

   In my test setup, the profiles were:

   ```text
   Default profile path:   /tmp/brave-filter-repro-user-data/Default
   Secondary profile path: /tmp/brave-filter-repro-user-data/Profile 2
   ```

   If your secondary profile uses a different profile directory, replace `Profile 2` below with the value shown in `brave://version` → `Profile Path`.

8. Quit all Brave processes.

9. Relaunch Brave directly into the secondary profile, without opening the initial/default profile:

   ```bash
   brave \
     --user-data-dir=/tmp/brave-filter-repro-user-data \
     --profile-directory="Profile 2" \
     --host-resolver-rules="MAP *.brave-repro.test 127.0.0.1" \
     --no-first-run
   ```

   For a local Brave development build:

   ```bash
   ../out/Component/brave \
     --user-data-dir=/tmp/brave-filter-repro-user-data \
     --profile-directory="Profile 2" \
     --host-resolver-rules="MAP *.brave-repro.test 127.0.0.1" \
     --no-first-run
   ```

10. In the secondary profile, open:

    ```text
    brave://settings/shields/filters
    ```

11. Open the test page again:

    ```text
    http://page.brave-repro.test:8765/page.html
    ```

12. Optionally, try `Update now` for the custom filter list from the secondary profile.

## Actual result

When Brave is launched directly into the secondary regular profile, without loading the initial/default profile:

- the saved custom filter list subscription may be absent from `brave://settings/shields/filters`
- the saved custom subscription may not be loaded into the in-memory subscription manager
- the subscription’s cached rules may not be registered as filter providers
- `blocked.svg` may load, showing that the custom subscription rule is not applied
- refreshing/updating the subscription may no-op because the subscription download manager was not obtained

Opening the initial/default profile can cause the saved custom subscription to appear or start working again.

## Expected result

Saved custom filter list subscriptions should load when any regular profile is available.

Launching Brave directly into a secondary regular profile should:

- show the saved custom subscription in `brave://settings/shields/filters`
- register the custom subscription’s filter provider
- apply cached custom subscription rules
- allow updating/refreshing the subscription
- not require the initial/default profile to be opened first

This expected behavior assumes the current global custom subscription model. This issue is not requesting a full per-profile custom subscription implementation.

## Reproduces how often

Easily reproduced in the described startup order.

## Brave version (brave://version info)

Reproduced on official build:

```text
Brave: 1.90.122
Chromium: 148.0.7778.167
Revision: a4ba963d1c3c7d55b091ca88c355ed12d5224b90
OS: Linux x64
````

Also investigated with a local Brave development build:

```text
Brave development build from brave-core revision: fa5cf216e44252e1cb52f61bde8c9c36cc4b2545
Chromium: 148.0.7778.167
OS: Linux x64
User data dir: /tmp/brave-filter-repro-user-data
Default profile: /tmp/brave-filter-repro-user-data/Default
Secondary profile: /tmp/brave-filter-repro-user-data/Profile 2
```

## Channel information

- [x] release / stable
- [ ] beta
- [ ] nightly
- [x] local development build

Notes:

- I reproduced/investigated this with a local Brave development build.
- The linked community reports indicate similar symptoms in released desktop builds, but this issue should be verified by Brave QA on current Release, Beta, and Nightly.

## Reproducibility

- Brave Shields disabled: N/A. This issue concerns Shields custom content filter subscription loading. Disabling Shields is not a meaningful control for verifying whether Shields custom subscriptions load and apply.
- Brave Rewards disabled: N/A. Rewards is not involved in this repro.
- Latest Chrome: N/A. `brave://settings/shields/filters`, Brave Shields, and Brave custom adblock subscriptions are Brave-specific.

## Miscellaneous information

Source-level investigation suggests the issue is caused by custom subscription loading being gated on obtaining an `AdBlockSubscriptionDownloadManager`.

In the affected code path, the download manager getter only checked the initial/default profile path. If Brave was launched with only a secondary regular profile loaded, the getter callback was not satisfied. As a result, saved subscriptions from Local State were not loaded into memory, filter providers were not registered, and the custom subscription rules were not applied.

The intended fix should preserve the current global custom subscription behavior and only fix the startup-order bug: existing global custom subscriptions should not depend on the initial/default profile being opened.

rsuwa:
Resolves https://github.com/brave/brave-browser/issues/55623

## Summary

Fix custom adblock subscription loading when Brave starts directly with a secondary regular profile, without loading the initial/default profile.

Custom filter list subscriptions are stored globally in Local State, but loading them was gated on obtaining an `AdBlockSubscriptionDownloadManager`. The getter only looked up that manager from the initial profile directory. When Brave was launched directly with another regular profile directory, the initial profile was not loaded, so the service manager never loaded saved custom subscriptions.

This change keeps the existing global subscription semantics and cache location. It still prefers the initial/default profile when loaded, but falls back to an already-loaded regular profile and satisfies pending manager waits from a newly-added regular profile.

This PR does not change whether custom subscriptions are global or per-profile; it only fixes the startup-order bug where the global subscription service waited for the initial/default profile’s download manager.

## Test plan

- `git diff --check origin/master...HEAD` passed.
- `PATH=/home/suwa/oss/src/third_party/rust-toolchain/bin:/home/suwa/oss/src/third_party/depot_tools:$PATH PYTHONPATH=/home/suwa/oss/src/brave/script npm run build -- -C Component --target=brave,brave_browser_tests` passed.
- `../out/Component/brave_browser_tests --gtest_filter='AdBlockServiceSecondaryProfileStartupTest.*:AdBlockServiceTest.SubscribeTo404List:AdBlockServiceTest.SubscribeToListUrlTwice' --no-sandbox` passed.
- `../out/Component/brave_browser_tests --gtest_filter='AdBlockServiceTest.*:AdBlockServiceSecondaryProfileStartupTest.*' --no-sandbox` passed: 84/84 tests passed.

Manual verification performed with a local developer build:
1. Started Brave with a user data dir containing the initial/default profile and at least one other regular profile.
2. In the initial/default profile, added a custom list from `brave://settings/shields/filters`.
3. Confirmed the list blocked the expected test resource.
4. Quit Brave.
5. Relaunched Brave with only the secondary regular profile using `--profile-directory="<secondary profile directory>"`.
6. Confirmed the custom list still appeared in settings without opening the initial/default profile.
7. Confirmed the custom list still blocked the expected test resource.

Note: `<secondary profile directory>` is the profile directory name for the non-initial regular profile being tested. The exact value depends on the profile creation/deletion history.

## AI assistance disclosure

AI assistance was used to investigate the startup-order bug and draft parts of the localized fix and regression test. I understand the changes in this PR, reviewed the implementation myself, and manually verified the user-facing behavior locally as described above.

### DIFF
--- a/browser/brave_shields/ad_block_service_browsertest.cc
+++ b/browser/brave_shields/ad_block_service_browsertest.cc
@@ -16,6 +16,7 @@
 #include "base/check.h"
 #include "base/files/file_util.h"
 #include "base/functional/callback_helpers.h"
+#include "base/json/values_util.h"
 #include "base/memory/raw_ptr.h"
 #include "base/memory/scoped_refptr.h"
 #include "base/path_service.h"
@@ -54,13 +55,18 @@
 #include "chrome/browser/browser_process.h"
 #include "chrome/browser/content_settings/host_content_settings_map_factory.h"
 #include "chrome/browser/profiles/profile.h"
+#include "chrome/browser/profiles/profile_manager.h"
+#include "chrome/browser/profiles/profile_test_util.h"
 #include "chrome/common/chrome_isolated_world_ids.h"
+#include "chrome/common/chrome_switches.h"
 #include "chrome/test/base/chrome_test_utils.h"
 #include "chrome/test/base/platform_browser_test.h"
 #include "components/prefs/pref_change_registrar.h"
 #include "components/prefs/pref_service.h"
+#include "components/prefs/scoped_user_pref_update.h"
 #include "content/public/test/browser_test.h"
 #include "content/public/test/browser_test_utils.h"
+#include "net/base/filename_util.h"
 #include "net/dns/mock_host_resolver.h"
 #include "net/test/embedded_test_server/embedded_test_server.h"
 #include "net/test/embedded_test_server/install_default_websocket_handlers.h"
@@ -123,6 +129,15 @@ using brave_shields::features::kBraveAdblockScriptletDebugLogs;
 using brave_shields::features::kCosmeticFilteringJsPerformance;
 
 namespace {
+
+#if !BUILDFLAG(IS_ANDROID)
+constexpr char kSecondaryProfileDir[] = "Profile 2";
+constexpr char kSubscriptionListForSecondaryProfileStartup[] =
+    "! Title: Test list\n"
+    "! Homepage: https://example.com/list.txt\n"
+    "! Expires: 3 days\n"
+    "||b.com^*logo.png^\n";
+#endif
 
 void WaitForSelectorBlocked(const content::ToRenderFrameHost& target,
                             const std::string& selector) {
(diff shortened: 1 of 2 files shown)


rsuwa:
Hi @bridiver, sorry for the direct ping. I saw your recent work around the adblock startup/cache paths, so I’m hoping you may know the right reviewer for this area.

Could you take a look or help route this PR when you have a chance?

This fixes a startup-order issue where existing custom filter list subscriptions are not loaded when Brave starts with only a non-initial regular profile. The change keeps the existing global custom subscription model and includes a regression browser test plus local manual verification.

Thanks.

bridiver:
> Hi @bridiver, sorry for the direct ping. I saw your recent work around the adblock startup/cache paths, so I’m hoping you may know the right reviewer for this area.
> 
> Could you take a look or help route this PR when you have a chance?
> 
> This fixes a startup-order issue where existing custom filter list subscriptions are not loaded when Brave starts with only a non-initial regular profile. The change keeps the existing global custom subscription model and includes a regression browser test plus local manual verification.
> 
> Thanks.

I need to review with @antonok-edm because this shouldn't be tied to a profile since it's local state. Maybe we did this because we planned to move it to profile prefs or maybe there was some other issue, but seems like system profile would be more appropriate here

rsuwa:
Thanks, that makes sense. I agree that, since the subscription metadata is Local State-backed, the cleaner model would be for this path not to depend on whichever regular user profile happens to be loaded. A system-profile-backed or otherwise profile-independent download path sounds more consistent with the current global subscription model.

I kept this PR intentionally scoped to the existing profile-backed download manager behavior so it can address the user-visible startup-order bug with minimal semantic change. In the current implementation, users can end up with existing custom subscriptions not shown or applied when Brave starts directly with a non-initial regular profile, and this fixes that without changing the broader global-vs-per-profile model.

I’m happy to adjust the approach if you and @antonok-edm think this should move to a system-profile-backed manager instead. Thanks for taking a look and for checking the design intent.

PR #1399, opened 2019-01-21
Adds support for min visits

NejcZdovc:
Resolves https://github.com/brave/brave-browser/issues/2110

## Submitter Checklist:

- [x] Submitted a [ticket](https://github.com/brave/brave-browser/issues) for my issue if one did not already exist.
- [x] Used Github [auto-closing keywords](https://help.github.com/articles/closing-issues-via-commit-messages/) in the commit message.
- [x] Added/updated tests for this change (for new code or code which already has tests).
- Verified that these changes build without errors on
  - [ ] Windows
  - [x] macOS
  - [x] Linux
- Verified that these changes pass automated tests (`npm test brave_unit_tests && npm test brave_browser_tests`) on
  - [ ] Windows
  - [x] macOS
  - [ ] Linux
- [ ] Ran `git rebase master` (if needed).
- [ ] Ran `git rebase -i` to squash commits (if needed).
- [x] Tagged reviewers and labelled the pull request as needed.
- [ ] Request a security/privacy review as needed.
- [x] Add appropriate QA labels (QA/Yes or QA/No) to include the closed issue in milestone

## Test Plan:
plan 1:
- enable rewards
- visit couple of sites multiple times
- go to publisher db and make sure that number of visits in activity_info matches what you did

plan 2:
- enable rewards
- visit a site
- make sure that you see it in ac table
- go to ac settings and check 5 visits as min visits
- make sure that ac table is now empty
- visit this site 4 more times
- make sure that site is now visible in ac table

## Reviewer Checklist:

- [ ] New files have MPL-2.0 license header.
- [ ] Request a security/privacy review as needed.
- [x] Adequate test coverage exists to prevent regressions 
- [x] Verify test plan is specified in PR before merging to source

PR #1389, opened 2019-01-18
Adds error handling to the panel

NejcZdovc:
Resolves https://github.com/brave/brave-browser/issues/3016

## Submitter Checklist:

- [x] Submitted a [ticket](https://github.com/brave/brave-browser/issues) for my issue if one did not already exist.
- [x] Used Github [auto-closing keywords](https://help.github.com/articles/closing-issues-via-commit-messages/) in the commit message.
- [ ] Added/updated tests for this change (for new code or code which already has tests).
- Verified that these changes build without errors on
  - [ ] Windows
  - [x] macOS
  - [ ] Linux
- Verified that these changes pass automated tests (`npm test brave_unit_tests && npm test brave_browser_tests`) on
  - [ ] Windows
  - [x] macOS
  - [ ] Linux
- [ ] Ran `git rebase master` (if needed).
- [ ] Ran `git rebase -i` to squash commits (if needed).
- [x] Tagged reviewers and labelled the pull request as needed.
- [ ] Request a security/privacy review as needed.
- [x] Add appropriate QA labels (QA/Yes or QA/No) to include the closed issue in milestone

## Test Plan:
plan 1:
- disable internet
- start browser
- click on panel and click create wallet
- make sure that you see try again message
- enable internet
- click on try again
- make sure that wallet was crated

plan 2:
- click on panel
- click create wallet
- close panel before wallet is created
- click on panel again
- make sure that you see creating wallet message before wallet is created

## Reviewer Checklist:

- [ ] New files have MPL-2.0 license header.
- [ ] Request a security/privacy review as needed.
- [ ] Adequate test coverage exists to prevent regressions 
- [x] Verify test plan is specified in PR before merging to source

PR #1385, opened 2019-01-18
use `brave://` for the virtual url and `chrome://` for the internal u…

bridiver:
…rl to load

fix https://github.com/brave/brave-browser/issues/1616

## Submitter Checklist:

- [x] Submitted a [ticket](https://github.com/brave/brave-browser/issues) for my issue if one did not already exist.
- [x] Used Github [auto-closing keywords](https://help.github.com/articles/closing-issues-via-commit-messages/) in the commit message.
- [x] Added/updated tests for this change (for new code or code which already has tests).
- Verified that these changes build without errors on
  - [ ] Windows
  - [x] macOS
  - [ ] Linux
- Verified that these changes pass automated tests (`npm test brave_unit_tests && npm test brave_browser_tests`) on
  - [ ] Windows
  - [x] macOS
  - [ ] Linux
- [x] Ran `git rebase master` (if needed).
- [x] Ran `git rebase -i` to squash commits (if needed).
- [x] Tagged reviewers and labelled the pull request as needed.
- [x] Request a security/privacy review as needed.
- [x] Add appropriate QA labels (QA/Yes or QA/No) to include the closed issue in milestone

## Reviewer Checklist:

- [ ] New files have MPL-2.0 license header.
- [ ] Request a security/privacy review as needed.
- [ ] Adequate test coverage exists to prevent regressions 
- [ ] Verify test plan is specified in PR before merging to source

PR #36524, opened 2026-05-19
[Nala]: Update to Svelte5 version of Nala

fallaciousreasoning:
Changes https://github.com/brave/leo/compare/075a3f9ed76e03bdc3f63d7dc4a7269e6dbdec29...e4af287f27ef5bb412a57a057d0edbbc81326168
    - **BREAKING** [Icons]: Update icons from Figma ([#1378](https://github.com/brave/leo/pull/1378))
    - [svelte5]: Use unmount not destroy ([#1399](https://github.com/brave/leo/pull/1399))
    - [Svelte]: Minor bump ([#1398](https://github.com/brave/leo/pull/1398))
    - **BREAKING** [Svelte]: Upgrade to Svelte 5 ([#1078](https://github.com/brave/leo/pull/1078))
    - chore(deps): updates deps and adds .npmrc with min-release-age ([#1391](https://github.com/brave/leo/pull/1391))
    - [Toggle]: Remove :hover effect on mobile ([#1390](https://github.com/brave/leo/pull/1390))
    - [Radio]: Remove :hover effect on mobile ([#1389](https://github.com/brave/leo/pull/1389))
    - [Checkbox]: Disable `:hover` effect on mobile ([#1388](https://github.com/brave/leo/pull/1388))
    - [Button]: Disable :hover state on mobile ([#1387](https://github.com/brave/leo/pull/1387))
    - Add transform transition and press scale to button ([#1385](https://github.com/brave/leo/pull/1385))
    - Navigation style updates ([#1386](https://github.com/brave/leo/pull/1386))
    - Bump postcss from 8.5.6 to 8.5.14 in /examples/example-ui-react ([#1383](https://github.com/brave/leo/pull/1383))
    - Bump postcss from 8.5.6 to 8.5.14 ([#1384](https://github.com/brave/leo/pull/1384))

The main issue arising from the upgrade is that [Svelte now uses a `TrustedTypes` policy to create HTML](https://github.com/sveltejs/svelte/pull/8135). Unfortunately this broke things where we include multiple copies of the Nala components on a page (which we do in settings sometimes), as we'd try and register two identical policies. The ideal solution here would be to either deduplicate or iframe the subpage, but both of those come with their own issues, so for now, I'm just disabling creation of a second policy, which works because the WebComponents are registered in the same context, so the inner page can use the versions of the components which are registered with the correct policy. 

I've also needed to update all of our TrustedTypes overrides to include `svelte-trusted-html`

The other breaking changes are related to stricter type checking (i.e. no `color` prop on `Button`).


diracdeltas:
this seems reasonable but may have introduced new npm audit issues? (based on the medium cve's flagged by socket)

browser/ui/webui/email_aliases/email_aliases_panel_ui.cc
```diff
@@ -41,9 +41,11 @@ EmailAliasesPanelUI::EmailAliasesPanelUI(content::WebUI* web_ui)
   source->OverrideContentSecurityPolicy(
       network::mojom::CSPDirectiveName::StyleSrc,
       "style-src 'self' 'unsafe-inline' chrome://resources chrome://theme;");
+  content::URLDataSource::Add(profile, std::make_unique<ThemeSource>(profile));
   webui::SetupWebUIDataSource(source, kEmailAliasesPanelResources,
                               IDR_EMAIL_ALIASES_PANEL_HTML);
-  content::URLDataSource::Add(profile, std::make_unique<ThemeSource>(profile));
+
+  LOG(ERROR) << "Initialized panel";
```
petemill: leftover log

components/brave_wallet_ui/page/screens/backup-wallet/backup-recovery-phrase/backup-recovery-phrase.tsx
```diff
@@ -148,7 +147,6 @@ export const BackupRecoveryPhrase = () => {
         </NextButtonRow>
         <Button
           kind='plain-faint'
-          color={leo.color.text.secondary}
```
petemill, line 151: does it now complain about these extra props?

components/common/disableDuplicateSvelteTrustedPolicies.ts
```diff
@@ -0,0 +1,28 @@
...
+// registering duplicate trusted type policies. We don't want to
+// allow-duplicate policies in general, as that would allow any script
+// to register their own policy if they know the name of one of ours.
+// Similarly, even allowing mulitple registrations of
+// `svelte-trusted-html` is not a solution, as then an attacker could
+// gain access to all the permissions `svelte-trusted-html` has (creating
+// arbitrary HTML). Fortunately, Svelte doesn't attempt to register its
+// policy if `window.trustedTypes` don't exist, and we don't actually
+// need the policy in any of our bundles where we include a copy of
+// Svelte.
```
petemill: Perhaps indicate what error is received without this, so people have a hint of knowing in what scenario to add this import. Also let's maybe clear this up because we _do_ need the policy right? We just don't need them to be repeated because it will use the one that's already defined. I think? Would there be a conceivable case where we do need the duplicate? Actually how does the original get registered at all if we're overriding it here? Are we somehow only overriding after the first one is registered?
Either way, this is a bit difficult to hack around, or at least unergonomic, so I'd like a TODO here about avoiding the need for it. Perhaps it's vendored svelte served from chrome://resources, for example.

components/email_aliases/resources/email_aliases.tsx
```diff
@@ -3,6 +3,11 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this file,
 // You can obtain one at https://mozilla.org/MPL/2.0/.
 
+// This page contains a copy of Svelte which will attempt to register its
```
petemill:
```suggestion
// This page contains a copy of Svelte via @brave/leo which will attempt to register its
```

components/brave_wallet_ui/page/screens/backup-wallet/backup-recovery-phrase/backup-recovery-phrase.tsx
Douglashdaniel, line 151: With this update, could we replaces these buttons with a `StyledComponent` and correct `--leo-button-color` variable?

browser/ui/webui/email_aliases/email_aliases_panel_ui.cc
```diff
@@ -41,9 +41,11 @@ EmailAliasesPanelUI::EmailAliasesPanelUI(content::WebUI* web_ui)
...
 }
```
netzenbot, line 47: `LOG(ERROR) << "Initialized panel"` looks like a debug leftover — this will spam the error log in production builds and should be removed before merging.

browser/ui/webui/email_aliases/email_aliases_panel_ui.cc
```diff
@@ -41,9 +41,11 @@ EmailAliasesPanelUI::EmailAliasesPanelUI(content::WebUI* web_ui)
   source->OverrideContentSecurityPolicy(
       network::mojom::CSPDirectiveName::StyleSrc,
       "style-src 'self' 'unsafe-inline' chrome://resources chrome://theme;");
+  content::URLDataSource::Add(profile, std::make_unique<ThemeSource>(profile));
   webui::SetupWebUIDataSource(source, kEmailAliasesPanelResources,
                               IDR_EMAIL_ALIASES_PANEL_HTML);
-  content::URLDataSource::Add(profile, std::make_unique<ThemeSource>(profile));
+
```
netzenbot: This `LOG(ERROR) << "Initialized panel";` looks like a debug statement that was accidentally left in. It will spam the error log in production builds and should be removed.

browser/ui/webui/email_aliases/email_aliases_panel_ui.cc
```diff
@@ -41,9 +41,11 @@ EmailAliasesPanelUI::EmailAliasesPanelUI(content::WebUI* web_ui)
...
+  LOG(ERROR) << "Initialized panel";
 }
 
 EmailAliasesPanelUI::~EmailAliasesPanelUI() = default;
```
netzenbot, line 49: Debug log accidentally left in: `LOG(ERROR) << "Initialized panel";`. This will emit error-level log noise on every panel initialization in production builds. Should be removed before landing.

browser/ui/webui/email_aliases/email_aliases_panel_ui.cc
netzenbot: `LOG(ERROR) << "Initialized panel"` looks like leftover debug logging — initializing a panel is not an error. Per [CS-044](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-044), `LOG(ERROR)` should be reserved for truly unexpected, serious failures. This should either be removed or replaced with `VLOG(1)`.

browser/ui/webui/email_aliases/email_aliases_panel_ui.cc
```diff
@@ -41,9 +41,11 @@ EmailAliasesPanelUI::EmailAliasesPanelUI(content::WebUI* web_ui)
   source->OverrideContentSecurityPolicy(
       network::mojom::CSPDirectiveName::StyleSrc,
       "style-src 'self' 'unsafe-inline' chrome://resources chrome://theme;");
+  content::URLDataSource::Add(profile, std::make_unique<ThemeSource>(profile));
   webui::SetupWebUIDataSource(source, kEmailAliasesPanelResources,
                               IDR_EMAIL_ALIASES_PANEL_HTML);
-  content::URLDataSource::Add(profile, std::make_unique<ThemeSource>(profile));
+
+  LOG(ERROR) << "Initialized panel";
 }
```
netzenbot, line 48: nit: `LOG(ERROR) << "Initialized panel";` looks like a debug statement that shouldn't be committed. Per the Chromium style guide, remove logging before check-in; if retained logging is needed, use `DVLOG(1)` instead. ([Use `DVLOG(1)` for Retained Logging, `ScopedCrashKeyString` for Crash Diagnostics](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-067))

components/brave_wallet_ui/page/screens/backup-wallet/backup-recovery-phrase/backup-recovery-phrase.tsx
petemill: we don't need to if these props weren't doing anything previously and everything looked fine?

chromium_src/ui/webui/webui_util.cc
```diff
@@ -42,6 +42,10 @@ void SetupWebUIDataSource(content::WebUIDataSource* source,
   source->OverrideContentSecurityPolicy(
       network::mojom::CSPDirectiveName::ScriptSrc,
       IsChromeUntrustedDataSource(source) ? kBraveUntrustedCSP : kBraveCSP);
+  source->OverrideContentSecurityPolicy(
```
diracdeltas: can we only do this on pages that actually use svelte? i'm assuming upstream ones do not, for instance?

components/brave_wallet_ui/page/screens/backup-wallet/backup-recovery-phrase/backup-recovery-phrase.tsx
fallaciousreasoning: yup, fails to compile with them now

chromium_src/ui/webui/webui_util.cc
fallaciousreasoning: They do because we replace the upstream buttons/toggles with our ones from Nala. Additionally, we load other Nala components (like dialogs, segmented controls) in upstream pages.

package-lock.json
```diff
@@ -21039,40 +20913,117 @@
...
-        "@jridgewell/trace-mapping": "^0.3.18",
-        "@types/estree": "^1.0.1",
-        "acorn": "^8.9.0",
-        "aria-query": "^5.3.0",
-        "axobject-query": "^4.0.0",
-        "code-red": "^1.0.3",
-        "css-tree": "^2.3.1",
-        "estree-walker": "^3.0.3",
-        "is-reference": "^3.0.1",
+      "version": "5.55.5",
```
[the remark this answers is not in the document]
fallaciousreasoning: I've got a bump for this https://github.com/brave/leo/pull/1398

netzenbot (approved):

Douglashdaniel (approved):
++ Wallet Front-end

chromium_src/ui/webui/webui_util.cc
petemill: I assume the svelte trusted types policy's html creator function sanitizes, so this is not extra surface area a dependency could compromise compared to our own html policy? Although I just looked it up and found @fallaciousreasoning themself contributed the `svelte-trusted-types` policy to svelte! Although it just seems to return the html string un-sanitized?

https://github.com/sveltejs/svelte/pull/16271/changes#diff-e00f3f1b3b560eab14db75189695a696f396118ec80555b4987e7500b8ca4877

petemill: @fallaciousreasoning did let me know that's standard practice - we trust `svelte` itself, and malicious packages won't have a reference to it's actual policy implementation. Same as polymer already in the chromium pages

```
const policy = window.trustedTypes &&
    trustedTypes.createPolicy('polymer-html-literal', {createHTML: (s) => s});
```

petemill (approved):
LGTM and the workaround for duplicate policies seems fine as long as we're working on https://github.com/brave/brave-browser/issues/55656 to remove that complexity (which we are)

cdesouza-chromium (approved):
LGTM

zenparsing (approved):
Rewards 👍

cdesouza-chromium (approved):
LGTM

cdesouza-chromium (approved):
LGTM

browser/ui/webui/brave_webui_source.cc
```diff
@@ -63,11 +63,11 @@ content::WebUIDataSource* CreateWebUIDataSource(
   if (disable_trusted_types_csp) {
```
diracdeltas: do we still need this conditional @petemill ?

diracdeltas (approved):

package.json
```diff
@@ -69,7 +69,7 @@
   },
   "engines": {
     "node": ">=24.11.1 <25.0.0",
-    "npm": ">=11.6.2"
+    "npm": ">=11.13.0"
```
fallaciousreasoning, line 72: cc @wknapik

browser/ui/webui/brave_webui_source.cc
fallaciousreasoning: Looks like browser/ui/webui/welcome_page/brave_welcome_ui.cc is still using it - I'll see if we can remove it from there

fallaciousreasoning: https://github.com/brave/brave-browser/issues/55884

fallaciousreasoning: https://github.com/brave/brave-core/pull/36805

package.json
```diff
@@ -69,7 +69,7 @@
   },
   "engines": {
     "node": ">=24.11.1 <25.0.0",
```
mihaiplesa: I'd put `">=24.16.0 <25.0.0"` here, also to get some security bumps too.

mihaiplesa (changes requested):

mihaiplesa (approved):

mihaiplesa (approved):

PR #36525, opened 2026-05-19
[docs][🩹] Dos and Don'ts

cdesouza-chromium:
This PR introduces Dos and Don'ts for 🩹 Plaster use. This document is
being written as a general guidance of what should belong in a plaster,
but also to keep track of what are the common operations done in several
places that deserve general guidelines.

### DIFF
--- a/docs/plaster_dos_and_donts.md
+++ b/docs/plaster_dos_and_donts.md
@@ -1,0 +1,613 @@
+# 🩹 Plaster Dos and Don'ts
+
+This document collects guidance for authoring plasters. The recommendations that
+follow are not hard rules. These are observations gathered from practical use,
+and offered as context for how 🩹 Plaster is best used.
+
+## Write the match to convey the right intent
+
+Where a regex is the right tool, the regex itself is the vehicle through which
+the intent of a change is expressed. A robust match describes _what_ is being
+changed and, where it can, offers some indication of _why_, rather than
+concerning itself merely with the mechanics of _how_ the substitution happens to
+take effect today.
+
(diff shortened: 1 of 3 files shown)


netzenbot (approved):

docs/plaster_dos_and_donts.md
```diff
@@ -0,0 +1,627 @@
...
+arrays, and so must be extended before any such use occurs. Even where no such
+use is present today, nothing prevents an upstream change from introducing one,
+and it is wise to guard against that possibility.
+
+Furthermore, avoid bundling extensions together. Each extension should sit close
+to the declaration it extends. Bundling forces the insertion to a single point
+at which every array involved is both valid to extend, and yet not read yet.
+That constraint itself could be a source of future breakage.
+
+Finally, constrain the match to a particular target.
```
mkarolin: I think typically you would also need to import the .gni file where `brave_foo_sources` is defined. It would be good to have that included into the example.

mkarolin (approved):
++ % suggestion

goodov:
let's merge this. I wanted to reference this in a PR, but see the doc is not merged yet.

cdesouza-chromium:
> let's merge this. I wanted to reference this in a PR, but see the doc is not merged yet.

I will merge today. I just need to update the enum section because we have an enum rewriter now.

docs/plaster_dos_and_donts.md
cdesouza-chromium: I have removed this section because Brian has some worries about how to proceed with this advice.

PR #36526, opened 2026-05-19
Serialize TreeTabNode data to session service

sangwoo108:
When a TreeTabNode has been created, updated or destroyed, we serialize the TreeTabNode data to the session service. This data is used to restore the TreeTabNode hierarchy when the browser is restarted, or when a tab is restored from recently closed tabs.

Part of https://github.com/brave/brave-browser/issues/49792


browser/ui/brave_browser.cc
```diff
@@ -393,3 +438,66 @@ bool BraveBrowser::ShouldSuppressDialogs(content::WebContents* source) {
...
+  const std::string node_id_str = node.id().ToString();
+
+  for (const tabs::TabInterface* tab_iface : node.GetTabs()) {
+    auto* wc = tab_iface->GetContents();
+    CHECK(wc);
+    auto* session_helper = sessions::SessionTabHelper::FromWebContents(wc);
+    CHECK(session_helper);
+
+    const SessionID tab_id = session_helper->session_id();
+    const auto* tree_coll = GetTreeTabNodeCollection(tab_iface);
```
sangwoo108: Instead of GetTreeTabNodeCollection(), make a collection getter in TreeTabNode and get the collection from `node`

browser/ui/brave_browser.cc
```diff
@@ -393,3 +438,66 @@ bool BraveBrowser::ShouldSuppressDialogs(content::WebContents* source) {
...
+void BraveBrowser::UpdateTreeTabCollapsedState(const tabs::TreeTabNode& node) {
+  SessionService* const session_service =
+      SessionServiceFactory::GetForProfileIfExisting(profile());
+  if (!session_service) {
+    return;
+  }
+
+  // Only the primary (kTab) tab of a node carries the collapsed state key.
+  for (const tabs::TabInterface* tab_iface : node.GetTabs()) {
+    const auto* tree_coll = GetTreeTabNodeCollection(tab_iface);
```
sangwoo108: ditto, Instead of GetTreeTabNodeCollection(), make a collection getter in TreeTabNode and get the collection from `node`

browser/ui/tabs/brave_tab_strip_model.cc
```diff
@@ -173,6 +173,12 @@ void BraveTabStripModel::BuildTreeTabs() {
                   &BraveTabStripModel::NotifyTreeTabNodeWillBeDestroyed,
                   base::Unretained(this))));
 
+  tree_tab_node_moved_subscription_ =
+      std::make_unique<base::CallbackListSubscription>(
+          tree_tab_model_->RegisterMovedTreeTabNodeCallback(base::BindRepeating(
+              &BraveTabStripModel::NotifyTreeTabNodeReparented,
+              base::Unretained(this))));
```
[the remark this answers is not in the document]
sangwoo108: Safe to have this `Unretained` as it's part of RAII style subscription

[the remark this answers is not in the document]
sangwoo108: /fp This is RAII style subscription, which is safe to use Unretained

browser/ui/tabs/tree_tab_model.h
```diff
@@ -69,6 +69,12 @@ class TreeTabModel {
       base::RepeatingCallback<void(const tabs::TreeTabNode&)> callback);
   base::CallbackListSubscription RegisterWillRemoveTreeTabNodeCallback(
       base::RepeatingCallback<void(const tree_tab::TreeTabNodeId&)> callback);
+  base::CallbackListSubscription RegisterMovedTreeTabNodeCallback(
+      base::RepeatingCallback<void(const tree_tab::TreeTabNodeId&)> callback);
+
```
netzenbot, line 74: nit: `has_pending_add_tree_tab_node_notification()` is only called in the browser test to decide whether to wait for the deferred PostTask. Consider naming it `has_pending_add_tree_tab_node_notification_for_testing()` to signal its intent, per [CSA-035](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-035).

browser/sessions/brave_session_restore_browsertest.cc
```diff
@@ -165,3 +178,183 @@ IN_PROC_BROWSER_TEST_P(SessionCookiesCleanupOnSessionRestoreBrowserTest,
...
+  auto* tree_model = brave_tab_strip_model()->tree_model();
+  if (tree_model->has_pending_add_tree_tab_node_notification()) {
+    // Let the deferred kNodeCreated notification task run.
+    base::RunLoop loop;
+    auto subscription =
+        tree_model->RegisterAddTreeTabNodeCallback(base::BindLambdaForTesting(
+            [&](const tabs::TreeTabNode& node) { loop.Quit(); }));
+    loop.Run();
+  }
+
```
netzenbot, line 295: The `if (windows[0]->tabs.empty()) { return; }` silently passes the test if no tabs were captured in the session. Since this test explicitly adds a second tab, the test should assert the expected count with `ASSERT_GE(windows[0]->tabs.size(), 2u)` so a setup failure doesn't go undetected. ([Verify Tests Actually Test What They Claim](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-012))

browser/sessions/brave_session_restore_browsertest.cc
```diff
@@ -165,3 +178,183 @@ IN_PROC_BROWSER_TEST_P(SessionCookiesCleanupOnSessionRestoreBrowserTest,
...
+  BraveTabStripModel* brave_tab_strip_model() {
+    return static_cast<BraveTabStripModel*>(browser()->tab_strip_model());
+  }
+
+  SessionService* session_service() {
+    return SessionServiceFactory::GetForProfile(browser()->profile());
+  }
+
```
netzenbot, line 211: The tests verify `kBraveTreeNodeIdKey` and `kBraveTreeParentNodeIdKey` but never check `kBraveTreeNodeCollapsedKey`, which is also written by `UpdateTreeTabSessionDataForNode`. Per [TI-022](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-022), all fields in serialization tests should be verified. Add a check that `kBraveTreeNodeCollapsedKey` is present and has value `"0"` or `"1"` for nodes that carry the collapsed key.

browser/sessions/BUILD.gn
```diff
@@ -3,14 +3,22 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
```
netzenbot, line 3: nit: `source_set("sessions")` is consumed by `//brave/browser/ui` (a different BUILD.gn), making it a public target. Per [BS-028](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-028), `source_set` should only be used for internal targets with `visibility = [ ":*" ]`; a public target should use `static_library` or `component`. For a header-only constants target this is minor, but adding `visibility` or using a different target type would be consistent with the rule.

browser/sessions/brave_session_restore_browsertest.cc
```diff
@@ -165,3 +178,189 @@ IN_PROC_BROWSER_TEST_P(SessionCookiesCleanupOnSessionRestoreBrowserTest,
...
+    // return non-null.
+    browser()->profile()->GetPrefs()->SetBoolean(
+        brave_tabs::kVerticalTabsEnabled, true);
+    browser()->profile()->GetPrefs()->SetBoolean(brave_tabs::kTreeTabsEnabled,
+                                                 true);
+  }
+
+  BraveTabStripModel* brave_tab_strip_model() {
+    return static_cast<BraveTabStripModel*>(browser()->tab_strip_model());
+  }
```
netzenbot, line 206: nit: Several loops in these tests use the manual `find()` + `!= end()` pattern (e.g., `auto it = session_tab->extra_data.find(kBraveTreeNodeIdKey); if (it != session_tab->extra_data.end() && !it->second.empty())`). Per [CSA-049](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-049), prefer `base::FindOrNull()` for map lookups: `auto* val = base::FindOrNull(session_tab->extra_data, kBraveTreeNodeIdKey); if (val && !val->empty())`. The same pattern appears in `ParentNodeIdStoredForChildTab` as well.

browser/ui/brave_browser.cc
```diff
@@ -15,7 +15,9 @@
 #include "base/functional/bind.h"
 #include "base/functional/callback_helpers.h"
 #include "base/memory/scoped_refptr.h"
```
netzenbot, line 17: nit: `base/notimplemented.h` is added here but `NOTIMPLEMENTED()` is never called in the new functions (`OnTreeTabChanged`, `UpdateTreeTabSessionDataForNode`, `UpdateTreeTabCollapsedState`). Per [CS-001](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-001), remove includes you don't actually use.

chromium_src/chrome/browser/sessions/session_service_base.cc
```diff
@@ -0,0 +1,62 @@
+/* Copyright (c) 2024 The Brave Authors. All rights reserved.
```
netzenbot: nit: This new file has `Copyright (c) 2024` but the current year is 2026. Per [CS-012](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-012), new files must use the current year in the copyright header.

browser/ui/brave_browser.cc
```diff
@@ -15,7 +15,9 @@
...
+#include "base/notimplemented.h"
```
netzenbot: nit: `#include "base/notimplemented.h"` is added here but `NOTIMPLEMENTED()` is never called in this file. Per [CS-066](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-066), dead `#include` lines should be removed before merging.

chromium_src/chrome/browser/sessions/DEPS
```diff
@@ -0,0 +1,3 @@
+include_rules = [
```
simonhong, line 1: how `//chrome/browser/sessions/DEPS` is handled with this src overridng?

chromium_src/chrome/browser/sessions/session_service_base.cc
```diff
@@ -0,0 +1,62 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "brave/browser/sessions/brave_session_keys.h"
+#include "brave/components/tabs/public/tree_tab_node.h"
+#include "brave/components/tabs/public/tree_tab_node_tab_collection.h"
```
simonhong, line 8: FYI, need to check this includings deps as `cr149` introduced `//chrome/browser/sessions` sub target.

browser/sessions/BUILD.gn
```diff
@@ -3,14 +3,22 @@
...
 # You can obtain one at https://mozilla.org/MPL/2.0/.
```
simonhong, line 5: maybe need `assert(enable_session_service)`?

chromium_src/chrome/browser/sessions/session_service_base.cc
```diff
@@ -0,0 +1,62 @@
...
+          session_id, kBraveTreeParentNodeIdKey, parent_id));
+  command_storage_manager->AppendRebuildCommand(
+      sessions::CreateAddTabExtraDataCommand(session_id,
+                                             kBraveTreeNodeCollapsedKey,
+                                             node.collapsed() ? "1" : "0"));
+}
+
+// This will be called in SessionTabBase::BuildCommandsForBrowser(). Plaster
+// will insert this function
+// there.
```
simonhong:
```suggestion
// will insert this function there.
```

browser/ui/tabs/tree_tab_model.h
```diff
@@ -83,10 +89,14 @@ class TreeTabModel {
   std::map<tree_tab::TreeTabNodeId, std::set<tree_tab::TreeTabNodeId>>
       descendant_ids_by_collapsed_ancestor_;
 
+  int pending_add_tree_tab_node_notification_count_ = 0;
```
simonhong:
```suggestion
  int pending_add_tree_tab_node_notification_count_for_testing_ = 0;
```

browser/ui/brave_browser.cc
```diff
@@ -257,6 +273,33 @@ void BraveBrowser::OnTabStripModelChanged(
   }
 }
 
+void BraveBrowser::OnTreeTabChanged(const TreeTabChange& change) {
+  if (!base::FeatureList::IsEnabled(tabs::kBraveTreeTab)) {
+    return;
+  }
```
simonhong: It's possible to call this method when feature is disabled?

chromium_src/chrome/browser/sessions/DEPS
sangwoo108: npm run gn_check will complain without this

```
ERROR in /Users/sangwoo/brave/src/brave/chromium_src/chrome/browser/sessions/session_service_base.cc
  Illegal include: "brave/components/tabs/public/tree_tab_node.h"
    Because of no rule applying.
  Illegal include: "brave/components/tabs/public/tree_tab_node_tab_collection.h"
    Because of no rule applying.
```

I think check include script just checks deps walking down the directories, so it doesn't matter if it's under src overriding or not.

sangwoo108: If you mean that we need to have deps in a gn, then I think we'd need to patch the upstream's gn file.

chromium_src/chrome/browser/sessions/session_service_base.cc
sangwoo108: With https://github.com/brave/brave-core/pull/36526#discussion_r3286662667, it might be better to wait for cr149 to arrive.

browser/ui/brave_browser.cc
sangwoo108: Um, actually I think it should be CHECK(). This callback should never be called when the feature is disabled.

browser/ui/tabs/tree_tab_model.h
sangwoo108: done

chromium_src/chrome/browser/sessions/session_service_base.cc
sangwoo108: oops, thanks.

browser/sessions/BUILD.gn
sangwoo108: done

package-lock.json
sangwoo108, line 1: this shouldn't be included

patches/chrome-browser-sessions-session_service_base.cc.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/chrome/browser/sessions/session_service_base.cc b/chrome/browser/sessions/session_service_base.cc
+index 05c5d9430070f6fd4a2b209afa4b2a61bf647853..5d9c8a179495859e9c01b727637b4381f157c061 100644
+--- a/chrome/browser/sessions/session_service_base.cc
++++ b/chrome/browser/sessions/session_service_base.cc
+@@ -673,6 +673,8 @@ void SessionServiceBase::BuildCommandsForBrowser(
+     Browser* browser,
+     IdToRange* tab_to_available_range,
+     std::set<SessionID>* windows_to_track) {
++  BuildCommandsForTreeTab(browser->tab_strip_model(), command_storage_manager());
++
```
netzenbot: nit: The patch adds an unnecessary empty line (confirmed by the presubmit warning at line 10). Per [PATCH-001](https://github.com/brave/brave-core/tree/master/docs/best-practices/patches.md#PATCH-001), patches should contain only minimal functional changes — no empty lines or whitespace modifications. Removing the blank line also keeps the patch to one line, satisfying PATCH-007.

browser/sessions/brave_session_restore_browsertest.cc
```diff
@@ -165,3 +179,202 @@ IN_PROC_BROWSER_TEST_P(SessionCookiesCleanupOnSessionRestoreBrowserTest,
...
+
+  SessionService* session_service() {
+    return SessionServiceFactory::GetForProfile(browser()->profile());
+  }
+
+  // Saves the last session using only incremental commands (no rebuild).
+  // Tests the AddTabExtraData path triggered by kNodeCreated events.
+  // Flushes pending tasks first so that deferred kNodeCreated notifications
+  // (posted via PostTask in TreeTabModel::AddTreeTabNode) have fired and
```
netzenbot, line 215: nit: `EXPECT_TRUE(base::FindOrNull(...))` without a `<<` diagnostic message is hard to debug on failure — you won't know which tab index or what keys were actually present. Per [TI-014](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-014), add context: `<< "tab extra_data missing " << kBraveTreeParentNodeIdKey` (and similarly for the collapsed key). Same applies to the identical checks in `TreeNodeExtraDataSurvivesFullRebuild` around line 248.

browser/sessions/brave_session_restore_browsertest.cc
```diff
@@ -165,3 +179,202 @@ IN_PROC_BROWSER_TEST_P(SessionCookiesCleanupOnSessionRestoreBrowserTest,
...
+          ASSERT_FALSE(error_reading);
+          std::move(predicate).Run(std::move(windows));
+          loop.Quit();
+        }));
+    loop.Run();
+  }
+
+  // Forces a full session rebuild (mirroring what happens on browser restart)
+  // then saves and reads back the last session. Tests the
+  // BRAVE_BUILD_COMMANDS_FOR_TREE_TAB macro in BuildCommandsForBrowser.
```
netzenbot, line 236: nit: The assertion lambdas in `TreeNodeExtraDataWrittenToSession` and `TreeNodeExtraDataSurvivesFullRebuild` are nearly identical (same loop, same three key checks, same `found_tree_id` logic). Per [TI-024](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-024), extract this into a shared helper (e.g., `VerifyTreeExtraDataKeys`) and pass the failure message as a parameter so both tests can call it without duplicating ~20 lines.

chromium_src/chrome/browser/sessions/DEPS
simonhong: Ah, I was confused. I was curious this DEPS src overriding file could hid upstream's corresponding DEPS file.
Both file works independently. NVM.

patches/chrome-browser-sessions-session_service_base.cc.patch
sangwoo108: done.

browser/sessions/brave_session_restore_browsertest.cc
sangwoo108: done.

browser/sessions/brave_session_restore_browsertest.cc
sangwoo108: done.

chromium_src/chrome/browser/sessions/session_service_base.cc
sangwoo108: @simonhong Does that mean that it should be included in deps via patching?

simonhong: yeah, we're doing patch. (ex, `patches/chrome-browser-ui-views-side_panel-BUILD.gn.patch`)

browser/ui/tabs/brave_tab_strip_model.cc
```diff
@@ -209,6 +216,18 @@ void BraveTabStripModel::NotifyTreeTabNodeWillBeDestroyed(
...
+  }
+  auto change = TreeTabChange(id, TreeTabChange::ReparentedChange(*node));
+  for (auto& observer : observers_) {
+    observer.OnTreeTabChanged(change);
+  }
+}
+
 const tree_tab::TreeTabNodeId* BraveTabStripModel::GetTreeTabNodeIdForGroup(
     tab_groups::TabGroupId group_id) const {
   return static_cast<const tabs::BraveTabStripCollection*>(contents_data_.get())
```
netzenbot, line 231: nit: `base::ObserverList` supports `observers_.Notify(&TabStripModelObserver::OnTreeTabChanged, change)` — prefer that over a manual `for` loop, per [CS-056](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-056). The other Notify* functions in this file use the same manual pattern, but new additions should follow the preferred form.

browser/sessions/brave_session_restore_browsertest.cc
```diff
@@ -165,3 +179,184 @@ IN_PROC_BROWSER_TEST_P(SessionCookiesCleanupOnSessionRestoreBrowserTest,
...
+                                     kBraveTreeNodeCollapsedKey))
+            << "tab extra data should contain kBraveTreeNodeCollapsedKey";
+        break;
+      }
+    }
+    EXPECT_TRUE(found_tree_id) << "Expected at least one tab with " +
+                                      kBraveTreeNodeIdKey + " in extra_data";
+  }
+
+  base::test::ScopedFeatureList feature_list_;
```
netzenbot, line 273: nit: The previous TI-022 comment was marked done, but this test's inline lambda still doesn't check `kBraveTreeNodeCollapsedKey`. Per [TI-022](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-022), all serialized fields should be verified — add a check that at least one tab carries a valid `kBraveTreeNodeCollapsedKey` value ("0" or "1") to complete the coverage.

chromium_src/chrome/browser/sessions/session_service_base.cc
```diff
@@ -0,0 +1,61 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
```
netzenbot, line 1: nit: Per [TI-040](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-040), when overriding upstream code via `chromium_src`, consider adding a `chromium_src/chrome/browser/sessions/session_service_unittest.cc` that includes the upstream test file (`#include <chrome/browser/sessions/session_service_unittest.cc>`) to catch regressions caused by the `BuildCommandsForBrowser` injection. The new behavior is covered by the browser tests, but upstream unit test coverage would guard against future breakage.

browser/sessions/brave_session_restore_browsertest.cc
sangwoo108: Added EXPECT_* for testing value of collapsed key.

browser/ui/tabs/brave_tab_strip_model.cc
sangwoo108: done.

chromium_src/chrome/browser/sessions/session_service_base.cc
sangwoo108: I think this isn't needed as we're covering our addition in the browser tests and we don't change anything from upstream.

chromium_src/chrome/browser/sessions/session_service_base.cc
sangwoo108: I'm wondering if the sources.gni should be put under chromium_src, as the deps is for the chromium_src directory for clarity. I see the similar pattern from serveral chromium_src/*/source.gni

sangwoo108: Tried adding the source.gni to chromium_src. Please let me know how it looks :)

browser/sessions/brave_session_restore_browsertest.cc
```diff
@@ -165,3 +179,189 @@ IN_PROC_BROWSER_TEST_P(SessionCookiesCleanupOnSessionRestoreBrowserTest,
...
+  }
+
+ protected:
+  void SetUpOnMainThread() override {
+    InProcessBrowserTest::SetUpOnMainThread();
+
+    // Both vertical tabs and tree tabs prefs must be true for
+    // BraveTabStripModel::BuildTreeTabs() to be called and tree_model() to
+    // return non-null.
+    browser()->profile()->GetPrefs()->SetBoolean(
```
netzenbot, line 198: nit: The collapsed-key value check calls `session_tab->extra_data.at(kBraveTreeNodeCollapsedKey)` twice after a separate `base::FindOrNull` presence check. Per [CSA-049](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-049), use the pointer returned by `FindOrNull` directly: `auto* collapsed = base::FindOrNull(session_tab->extra_data, kBraveTreeNodeCollapsedKey); EXPECT_TRUE(collapsed && (*collapsed == "0" || *collapsed == "1"))`. This avoids the redundant double lookup and the potential for a thrown exception if the key is absent.

sangwoo108: done

browser/sessions/BUILD.gn
sangwoo108: This was false positive, because the sources = [] consists of a header, we can't make it static lib.

browser/sessions/brave_session_restore_browsertest.cc
```diff
@@ -165,3 +179,186 @@ IN_PROC_BROWSER_TEST_P(SessionCookiesCleanupOnSessionRestoreBrowserTest,
...
+
+  // Saves the last session using only incremental commands (no rebuild).
+  // Tests the AddTabExtraData path triggered by kNodeCreated events.
+  // Flushes pending tasks first so that deferred kNodeCreated notifications
+  // (posted via PostTask in TreeTabModel::AddTreeTabNode) have fired and
+  // written their AddTabExtraData commands before the session is saved.
+  void VerifyExtraData(
+      base::OnceCallback<void(
+          std::vector<std::unique_ptr<sessions::SessionWindow>>)> predicate,
+      base::Location location = base::Location::Current()) {
```
simonhong, line 220: I think `location` should not have default args like this.
If caller doesn't pass it, `location` will be same always.
Same for `VerifyExtraDataAfterFullRebuild`

browser/sessions/brave_session_restore_browsertest.cc
```diff
@@ -165,3 +179,186 @@ IN_PROC_BROWSER_TEST_P(SessionCookiesCleanupOnSessionRestoreBrowserTest,
...
+IN_PROC_BROWSER_TEST_F(BraveTreeTabSessionRestoreBrowserTest,
+                       TreeNodeExtraDataWrittenToSession) {
+  // The browser starts with one tab; tree tabs should be enabled.
+  ASSERT_TRUE(brave_tab_strip_model()->tree_model());
+
+  int tab_count = browser()->tab_strip_model()->count();
+  ASSERT_EQ(tab_count, 1);
+
+  auto* tree_model = brave_tab_strip_model()->tree_model();
+  if (tree_model->has_pending_add_tree_tab_node_notification_for_testing()) {
```
simonhong, line 287: Do we need this `if` clause? Or we can't know whether the task is pending or not here?

browser/sessions/brave_session_restore_browsertest.cc
```diff
@@ -165,3 +179,186 @@ IN_PROC_BROWSER_TEST_P(SessionCookiesCleanupOnSessionRestoreBrowserTest,
...
+              base::FindOrNull(tab->extra_data, kBraveTreeParentNodeIdKey);
+          if (!parent_id || parent_id->empty()) {
+            continue;
+          }
+          EXPECT_TRUE(all_node_ids.count(*parent_id) > 0)
+              << "brave_tree_parent_node_id '" << *parent_id
+              << "' has no matching brave_tree_node_id in the session";
+        }
+      }));
+}
```
simonhong, line 365: how about testing some more scenarios after tree tab state changes such as changing parent node or delete. Is it redundant?

browser/sessions/brave_session_restore_browsertest.cc
simonhong: If we register notification callback initially, maybe we don't need to have another `has_pending_add_tree_tab_node_notification_for_testing_` bit.

browser/sessions/brave_session_restore_browsertest.cc
sangwoo108: Hum, I guess the argument expression is evaulated in the context of caller?

```
int a = 1;
int f(int);

int g(int x = f(a)); // lookup here

void h() {
    a = 2;
    g(); // evaluates f(a) now -> uses a == 2
}
```

browser/sessions/brave_session_restore_browsertest.cc
sangwoo108: I feel like it could be flaky if we don't check pending notification explicitly. And for clarity, I'd prefer having the condition so that we can know that we have  pending notification.

browser/sessions/brave_session_restore_browsertest.cc
sangwoo108: That's good call but I think we should cover it from TreeTabsBrowserTest's observer api verification.

sangwoo108: Actually, I think you're right, we might need both.

browser/sessions/brave_session_restore_browsertest.cc
simonhong: aha, ok. you're right!

browser/sessions/brave_session_restore_browsertest.cc
simonhong: yeah, it's optional. go ahead with yours.

netzenbot (approved):

browser/sessions/brave_session_restore_browsertest.cc
sangwoo108: Done!

simonhong (approved):
++

browser/ui/brave_browser.cc
```diff
@@ -268,6 +284,31 @@ void BraveBrowser::OnTabStripModelChanged(
   }
 }
 
+void BraveBrowser::OnTreeTabChanged(const TreeTabChange& change) {
```
goodov: I don't think we should continue adding more stuff into `Browser*`. It feels like the whole Tree Tab logic can be a separate `BrowserWindowFeature`. Is it doable? can we do this in this PR or better to follow up?

rewrite/chrome/browser/sessions/session_service_base.cc.toml
```diff
@@ -0,0 +1,12 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+[[substitution]]
+description = '''
+Call BuildCommandsForTreeTab at in SessionTabBase::BuildCommandsForBrowser().
+'''
+re_pattern = '(void SessionServiceBase::BuildCommandsForBrowser\([\s\S]*?\) {)'
+replace = '''\1
+  BuildCommandsForTreeTab(browser->tab_strip_model(), command_storage_manager());'''
```
goodov: I think we should use `yaml` for new rewrites

browser/ui/brave_browser.cc
sangwoo108: It should be definitely doable. The only reason I had it in BraveBrowser is that TabGroup operation is serialized from Browser by upstream implementation. Let me try to move it.

sangwoo108: Done!

rewrite/chrome/browser/sessions/session_service_base.cc.toml
sangwoo108: Nice. Done.

browser/ui/tabs/tree_tab_model.cc
```diff
@@ -125,13 +125,17 @@ void TreeTabModel::AddTreeTabNode(tabs::TreeTabNode& node) {
       return;
     }
 
+    model->pending_add_tree_tab_node_notification_count_for_testing_--;
+    CHECK_GE(model->pending_add_tree_tab_node_notification_count_for_testing_,
```
goodov: a bit unexpected to CHECK a "for_testing" variable.

goodov (approved):

goodov:
lgtm

browser/ui/tabs/tree_tab_model.cc
sangwoo108: Thanks, removed.

PR #36528, opened 2026-05-19
[Wallet] Remove SolanaAddress default constructor

supermassive:
- Remove default constructor for `SolanaAddress`. We no longer need `IsValid` method.
- A bit of related refactoring for SNS resolver code.

Need for https://github.com/brave/brave-browser/issues/55334

### DIFF
--- a/components/brave_wallet/browser/sns_resolver_task.cc
+++ b/components/brave_wallet/browser/sns_resolver_task.cc
@@ -6,27 +6,30 @@
 #include "brave/components/brave_wallet/browser/sns_resolver_task.h"
 
 #include <algorithm>
-#include <memory>
 #include <optional>
+#include <string_view>
 #include <utility>
 #include <vector>
 
 #include "base/base64.h"
 #include "base/check.h"
 #include "base/check_op.h"
+#include "base/containers/extend.h"
 #include "base/containers/span.h"
+#include "base/containers/span_reader.h"
+#include "base/containers/to_vector.h"
 #include "base/memory/raw_span.h"
-#include "base/no_destructor.h"
 #include "base/notreached.h"
 #include "base/numerics/byte_conversions.h"
 #include "base/strings/string_number_conversions.h"
-#include "base/strings/string_split.h"
-#include "base/strings/string_util.h"
 #include "base/strings/string_view_util.h"
+#include "base/types/expected.h"
 #include "base/values.h"
 #include "brave/components/api_request_helper/api_request_helper.h"
+#include "brave/components/brave_wallet/browser/brave_wallet_utils.h"
 #include "brave/components/brave_wallet/browser/json_rpc_requests_helper.h"
 #include "brave/components/brave_wallet/browser/json_rpc_response_parser.h"
+#include "brave/components/brave_wallet/browser/json_rpc_responses.h"
 #include "brave/components/brave_wallet/browser/solana_keyring.h"
 #include "brave/components/brave_wallet/browser/solana_requests.h"
 #include "brave/components/brave_wallet/browser/solana_response_parser.h"
@@ -44,7 +47,7 @@ namespace brave_wallet {
 struct SnsFetchRecordQueueItem {
   std::string record;
   SnsRecordsVersion version;
-  SolanaAddress record_address;
+  std::optional<SolanaAddress> record_address;
 };
 
 namespace {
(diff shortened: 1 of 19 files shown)


components/brave_wallet_ui/common/slices/endpoints/address.endpoints.ts
```diff
@@ -176,16 +176,11 @@ export const addressEndpoints = ({
...
-            if (errorMessage) {
-              throw new Error(errorMessage)
-            }
+            const address = await api.jsonRpcService.snsGetSolAddr(lowercaseURL)
 
             return {
               data: {
-                address,
+                address: address.result.success!,
                 requireOffchainConsent: false,
```
netzenbot, line 189: The old code checked `errorMessage` and threw on failure; the new code drops that entirely. `address.result.success` is `undefined` when the SNS lookup fails (the mojo `result<>` union sets `failure` instead of `success`), so `address.result.success!` silently passes `undefined` as the resolved address on error. Check `address.result.failure` and throw in that branch, similar to how the previous code handled `errorMessage`.

components/brave_wallet_ui/common/slices/endpoints/address.endpoints.ts
```diff
@@ -176,16 +176,11 @@ export const addressEndpoints = ({
...
               },
             }
```
netzenbot, line 191: The old code checked `errorMessage` and threw before using the address. The new code uses `address.result.success!` with a non-null assertion but never checks whether the result is an error — if SNS resolution fails, `success` is null/undefined and `undefined` gets returned as the resolved address instead of surfacing the error to the caller.

components/brave_wallet_ui/common/slices/endpoints/address.endpoints.ts
```diff
@@ -176,16 +176,11 @@ export const addressEndpoints = ({
...
-
-            if (errorMessage) {
-              throw new Error(errorMessage)
-            }
+            const address = await api.jsonRpcService.snsGetSolAddr(lowercaseURL)
 
             return {
               data: {
-                address,
+                address: address.result.success!,
```
netzenbot: The error handling was removed here. The old code threw on `errorMessage`; the new code uses `address.result.success!` with no guard. If the SNS resolution fails, `success` will be undefined and the error is silently swallowed. Add a check like `if (!address.result.success) { throw new Error(address.result.error ?? 'SNS resolution failed') }` before accessing `address.result.success`.

components/brave_wallet_ui/common/slices/endpoints/address.endpoints.ts
```diff
@@ -176,16 +176,11 @@ export const addressEndpoints = ({
...
-            const { address, errorMessage } =
-              await api.jsonRpcService.snsGetSolAddr(lowercaseURL)
-
-            if (errorMessage) {
-              throw new Error(errorMessage)
-            }
+            const address = await api.jsonRpcService.snsGetSolAddr(lowercaseURL)
 
             return {
               data: {
```
netzenbot, line 187: This removes error handling for the SNS case. When `snsGetSolAddr` returns an error, `address.result.success` is `undefined`, but `!` suppresses TypeScript's null check and the code returns `address: undefined` silently instead of throwing. The ENS and Unstoppable Domains paths below still check for errors and throw — this path should too. Consider: `if (!address.result.success) { throw new Error(address.result.error ?? 'SNS resolution failed') }`.

components/brave_wallet/browser/sns_resolver_task.h
```diff
@@ -168,8 +147,9 @@ class SnsResolverTask {
```
netzenbot, line 147: The new `wallet_address_result_` and `url_result_` fields are typed as `std::optional<base::expected<T, std::string>>`, which stores the error variant in the field rather than handling it immediately. Per [CS-029](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-029), prefer `std::optional<T>` and propagate errors at the call site instead of deferring them through a stored `expected`.

goodov (approved):
browser/net lgtm

Douglashdaniel (approved):
++ Wallet Front-end

components/brave_wallet/browser/sns_resolver_task.cc
```diff
@@ -301,8 +285,9 @@ std::optional<SolanaAddress> ParseAndVerifySolRecordData(
     base::span<const uint8_t> sol_record_payload,
     const SolanaAddress& domain_owner) {
   if (record_item.version == SnsRecordsVersion::kRecordsV1) {
+    CHECK(record_item.record_address);
```
cmazakas, line 288: Because `record_item` is caller-supplied do we need a CHECK() here over a `if(!record_item.record_address) { return std::nullopt; }`? I guess I'm wondering if this is such a severe precondition violation that it's worth crashing over.

cmazakas (approved):
The SnsResolveTask seems to look dramatically cleaner now, even though I had some trouble following the diffs. But all the existing tests seem to pass so I'm approving.

netzenbot (approved):

components/brave_wallet/browser/sns_resolver_task.cc
supermassive: All record_item.record_address are supposed to be filled in `FillWorkData`

PR #36539, opened 2026-05-19
Remove Brave News inline content ads dead code on Android

### LINKED ISSUE #55479: [ads] Remove Brave News inline content ads code on Android
Remove Brave News inline content ads code, as it will no longer be needed once [brave/brave-browser#39474](https://github.com/brave/brave-browser/issues/39474) is merged.

Desktop Brave News inline content ads code was removed in https://github.com/brave/brave-browser/issues/50312.

aseren:
The Brave News inline content ads pipeline is already disabled via https://github.com/brave/brave-core/pull/32359. This PR removes the remaining inline content ads Android code and mojom code that no longer has any effect.

### Test case 1

* Fresh install
* Verify that Brave News functionality on Android works as expected:
    - turning on Brave News shows Brave News widget.
    - turning off Brave News hides Brave News widget.
    - the feed loads and articles appear when scrolling to the Brave News section.

### Test case 2

* Install the older version of Brave (for example 1.91)
* Start the browser
* Turn on Brave News and scroll to get some news articles
* Install Brave version with the current fix
* Start the browser
* Verify that Brave News functionality on Android works as expected:
    - turning off Brave News hides Brave News widget.
    - turning on Brave News shows Brave News widget.
    - the feed loads and articles appear when scrolling to the Brave News section.

Resolves https://github.com/brave/brave-browser/issues/55479


android/java/org/chromium/chrome/browser/ntp/BraveNewTabPageLayout.java
```diff
@@ -639,80 +631,13 @@ public void onScrollStateChanged(RecyclerView recyclerView, int newState) {
                                                 mUuid = mVisibleCard.getUuid();
                                                 mCardType = "promo";
                                             }
-
-                                            // get params for view DISPLAY_AD
-                                            if (mVisibleCard.getCardType() == CardType.DISPLAY_AD) {
-                                                mItemPosition = newsFeedViewPosition;
```
SergeyZhukovsky, line 645: `private int mItemPosition;` is dead also now, could be removed

fallaciousreasoning (commented):
C++ changes LGTM - do we have a follow up issue for removing `PromotedItems`?

I think we should get rid of the brave://flags/#brave-ntp-refresh-enabled flag on desktop, and just branch on Android/Desktop instead

fallaciousreasoning (approved):

android/java/org/chromium/chrome/browser/ntp/BraveNewTabPageLayout.java
aseren: it is still used for Promoted Articles, which will be removed separately:
```
                                            // get params for view PROMOTED_ARTICLE
                                            if (mVisibleCard.getCardType()
                                                    == CardType.PROMOTED_ARTICLE) {
                                                mItemPosition = newsFeedViewPosition;
                                                mCreativeInstanceId =
                                                        BraveNewsUtils.getPromotionIdItem(
                                                                mVisibleCard);
                                                mUuid = mVisibleCard.getUuid();
                                                mCardType = "promo";
                                            }

```

aseren:
> C++ changes LGTM - do we have a follow up issue for removing `PromotedItems`?
> 
> I think we should get rid of the brave://flags/#brave-ntp-refresh-enabled flag on desktop, and just branch on Android/Desktop instead

Filed an issue to remove promoted articles: https://github.com/brave/brave-browser/issues/55659

netzenbot (approved):

bridiver (approved):

android/java/org/chromium/chrome/browser/ntp/BraveNewTabPageLayout.java
SergeyZhukovsky: we only assign it currently and don't use anywhere. But fine if you are going to remove it separately soon.

SergeyZhukovsky (approved):
++

aseren:
> [[puLL-Merge](https://github.com/brave/pull-merge)] - [brave/brave-core@36539](https://github.com/brave/brave-core/pull/36539)
> 
> ### Description
> Removes all Display Ad (inline content ad) support from Brave News across Android UI, backend controller, Mojom interface, feed building, and local database. The `DisplayAd` struct, `DISPLAY_AD` card type, related Mojom methods (`GetDisplayAd`, `OnDisplayAdVisit`, `OnDisplayAdView`), the `DisplayAdsTable` SQLite table, ad rendering/caching logic, and associated resources (drawable, styles, colors, dimens, strings) are deleted. Database version bumped from 3→4 with migration to drop the old table.
> 
> ### Possible Issues
> * **Database migration ordering**: `onUpgrade` calls `maybeMigrate()` then `onCreate()`. `onCreate` uses `CREATE TABLE IF NOT EXISTS` for remaining tables, so this works. However, if a future version 5 is added, `maybeMigrate` needs extension — the pattern of calling `onCreate` after migrate is unusual and could cause confusion. Typically `onCreate` isn't re-invoked in `onUpgrade`.
> * **Empty `onDestroyed` callback**: `BraveToolbarLayoutImpl.java` `onDestroyed` now has an empty body for the ad cleanup block removal but the surrounding callback structure remains. Verify the tab observer still functions correctly with no-op in that path.
> * **`g_page_content_order` has 14 entries but comment says 15→14**: The test update from 17→16 (15+2→14+2) is consistent only if exactly one `DISPLAY_AD` entry was removed. Verify the array in `feed_building.cc` matches.
> * **Dead condition**: `BraveNewTabPageLayout.java` line ~490 — `mCardType.equals("promo")` check remains but `"displayad"` was the only other value previously handled there. The `"promo"` string is never set before this check runs on the first iteration (it's set later in the scroll logic), so the ordering/logic should be audited for correctness.
> 
> Changes

These are not real issues because:
1. Calling `onCreate()` from `onUpgrade` is unusual for Android, but this is a legacy code and should be refactored separately. There is no bug with current code and database tables are migrated and created without any issues. The order of calling `maybeMigrate()` and then `onCreate()` is correct for this code and is logical because it first "cleaning up old schema" and then "ensuring current schema exists".
2. Empty `onDestroyed` is not real issue. The method has working cleanup calls. Nothing is empty.
3. `g_page_content_order` count concern is not real issue. The array has 14 entries, the test comment at line 388 correctly says 14 + 2 = 16. Everything is consistent.
4. mCardType.equals("promo") dead condition is not a bug. This is legacy code and will be removed when promoted articles support is removed.

mkarolin (approved):
`strings`++

PR #36545, opened 2026-05-19
Plumbing for BravePlaylistExclusions components

EliHini:
Contributes to https://github.com/brave/brave-browser/issues/55010

Component for Playlist Exclusion list
Installer followup here: https://github.com/brave/brave-core/pull/36790


ios/browser/playlist/playlist_exclusions_bridge.h
kylehickinson, line 1: Move to `ios/browser/playlist`, we dont add new things to api anymore

ios/browser/api/brave_playlist/playlist_exclusion_ios.h
```diff
@@ -0,0 +1,30 @@
...
+
+#ifndef BRAVE_IOS_BROWSER_API_BRAVE_PLAYLIST_PLAYLIST_EXCLUSION_IOS_H_
+#define BRAVE_IOS_BROWSER_API_BRAVE_PLAYLIST_PLAYLIST_EXCLUSION_IOS_H_
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+OBJC_EXPORT
+@interface BravePlaylistExclusionsIOS : NSObject
```
kylehickinson: Use the new pattern: expose a `@protocol` instead, create private impl's of it and return that instead. This will let you mock things on the Swift side easier

ios/app/brave_core_main.mm
```diff
@@ -288,6 +291,15 @@ - (HTTPSUpgradeExceptionsService*)httpsUpgradeExceptionsService {
   return _httpsUpgradeExceptionsService;
 }
 
+- (BravePlaylistExclusionsIOS*)bravePlaylistExclusions {
+  if (!_bravePlaylistExclusions) {
+    _bravePlaylistExclusions = [[BravePlaylistExclusionsIOS alloc]
+        initWithPlaylistExclusions:brave_playlist::PlaylistExclusions::
+                                       GetInstance()];
+  }
+  return _bravePlaylistExclusions;
```
kylehickinson: I realize this is duplicating `BraveUserAgentExceptionsIOS`, but if this is a singleton why not just expose an Obj-C singleton that returns a wrapper around the C++ instance? Doesn't seem to be required that it is exposed on `BraveCoreMain` nor do we have to maintain an instance of the wrapper

ios/browser/api/brave_playlist/headers.gni
```diff
@@ -0,0 +1,7 @@
+# Copyright (c) 2026 The Brave Authors. All rights reserved.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+browser_api_brave_playlist_exclusions_ios_public_headers =
+    [ "//brave/ios/browser/api/brave_playlist/playlist_exclusion_ios.h" ]
```
kylehickinson: Follow new patterns: Just add the header directly to the `ios/BUILD.gn`

ios/browser/web/BUILD.gn
```diff
@@ -17,6 +17,7 @@ source_set("web") {
     "//base",
     "//brave/components/ai_chat/core/browser",
     "//brave/components/brave_component_updater/browser:public",
+    "//brave/components/brave_playlist/browser",
```
kylehickinson: Don't see this needed anywhere

browser/ui/browser_window/internal/BUILD.gn
```diff
@@ -27,6 +27,7 @@ source_set("internal") {
 
   if (enable_playlist) {
     deps += [
+      "//brave/components/brave_playlist/browser",
```
kylehickinson: Why is this needed?

components/brave_playlist/browser/BUILD.gn
kylehickinson, line 1: We already have `components/playlist`, not sure why you made a new folder. This should likely go in `components/playlist/core/browser`

kylehickinson (changes requested):

components/brave_playlist/browser/BUILD.gn
EliHini: I wasn't sure what that was for and if we were using it for iOS so I steered clear of it in case it was just being used by browser

kylehickinson: We're not right now aside from accessing some shared pref names, but I assume whatever list we create here will eventually want to be used by other platforms too

components/playlist/core/browser/BUILD.gn
```diff
@@ -9,22 +9,30 @@ assert(enable_playlist)
...
+    "//brave/components/brave_component_updater/browser:public",
     "//brave/components/playlist/core/common",
     "//brave/components/playlist/core/common/buildflags",
     "//components/prefs",
+    "//net",
+    "//url",
   ]
 }
 
 source_set("unit_tests") {
```
netzenbot, line 29: `playlist_exclusion.h` exposes `GURL` in its public API (`CanResolvePageSrcLater`), so `//url` must be in `public_deps`, not `deps`. Consumers that include this header will fail to find `url/gurl.h` without the transitive include path. See [BS-024](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-024).

components/playlist/core/browser/playlist_exclusion.cc
```diff
@@ -0,0 +1,163 @@
...
+    if (rule.registrable_domain != domain) {
+      continue;
+    }
+    if (rule.deny_root_path && (path.empty() || path == "/")) {
+      return false;
+    }
+    for (const std::string& prefix : rule.path_prefixes) {
+      if (prefix.empty()) {
+        continue;
+      }
```
netzenbot, line 157: The manual key-by-key parsing in `OnPlaylistExclusionsLoaded` should use `base::JSONValueConverter` instead. Comparable component rule parsers in this codebase (e.g., `DebounceRule`, `RequestOTRRule`) use `RegisterJSONConverter` + `base::JSONValueConverter<T>` for this pattern. `base::JSONValueConverter` supports `RegisterRepeatedString` for the `path_prefixes` list field. See [CSA-004](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-004).

components/playlist/core/browser/playlist_exclusion.h
```diff
@@ -0,0 +1,89 @@
...
+  // Returns true when `pageSrc` may be used for LivePlaylist-style reload /
+  // re-resolution. When the component list is not loaded yet, returns true
+  bool CanResolvePageSrcLater(const GURL& url) const;
+
+  // Debugging / UI: human-readable "domain<TAB>condition" rows.
+  std::vector<std::string> ListPlaylistExclusions() const;
+
+  void ResetForTesting();
+
+ private:
```
netzenbot, line 71: `ResetForTesting()` is a production method added solely for test isolation of the singleton. The `_for_testing` suffix is a clear signal this violates [CSA-035](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-035). Consider resetting state in the test fixture via a test-only subclass or by injecting a factory, rather than adding a state-mutating method to the production class.

components/playlist/core/browser/playlist_exclusion.cc
```diff
@@ -0,0 +1,163 @@
...
+    const auto* prefixes = rule_dict->FindList("path_prefixes");
+    if (prefixes) {
+      for (const base::Value& p : *prefixes) {
+        const std::string* ps = p.GetIfString();
+        if (ps) {
+          rule.path_prefixes.push_back(*ps);
+        }
+      }
+    }
+    rules_.push_back(std::move(rule));
```
netzenbot, line 128: `ResetForTesting()` should call `weak_factory_.InvalidateWeakPtrs()` before clearing `rules_` and `is_ready_`. Otherwise an in-flight `OnPlaylistExclusionsLoaded` reply posted from a previous `OnComponentReady()` call can fire after the reset and overwrite the cleared state, causing flaky tests. See [CSM-006](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-memory.md#CSM-006).

components/playlist/core/browser/playlist_exclusion_unittest.cc
```diff
@@ -0,0 +1,70 @@
...
+
+  exclusions->OnComponentReady(temp_dir.GetPath());
+  ASSERT_TRUE(base::test::RunUntil([&]() { return exclusions->is_ready_; }));
+
+  EXPECT_FALSE(
+      exclusions->CanResolvePageSrcLater(GURL("https://youtube.com/")));
+  EXPECT_FALSE(
+      exclusions->CanResolvePageSrcLater(GURL("https://www.youtube.com")));
+  EXPECT_FALSE(exclusions->CanResolvePageSrcLater(
+      GURL("https://www.youtube.com/results?foo=1")));
```
netzenbot, line 84: The `RulesBlockListedPaths` test only exercises the happy path of JSON deserialization. `OnPlaylistExclusionsLoaded` has three early-return guards — empty contents, failed JSON parse, and missing `rules` list — none of which are tested. Per [TI-022](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-022), deserialization tests should always include invalid/malformed input cases.

components/playlist/core/browser/playlist_exclusion.h
```diff
@@ -0,0 +1,89 @@
...
+
+  static PlaylistExclusions* GetInstance();
+
+  void OnComponentReady(const base::FilePath& component_dir);
+
+  // Returns true when `pageSrc` may be used for LivePlaylist-style reload /
+  // re-resolution. When the component list is not loaded yet, returns true
+  bool CanResolvePageSrcLater(const GURL& url) const;
+
+  // Debugging / UI: human-readable "domain<TAB>condition" rows.
```
StephenHeaps: What is `domain<TAB>condition`?

components/playlist/core/browser/playlist_exclusion.h
```diff
@@ -0,0 +1,89 @@
...
+  std::vector<std::string> ListPlaylistExclusions() const;
```
StephenHeaps: I don't think this needs to be exposed, it's internal to the component and you can always inspect the file itself when debugging or add logs if needed.

components/playlist/core/browser/playlist_exclusion.cc
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+  is_ready_ = true;
+}
+
+bool PlaylistExclusions::CanResolvePageSrcLater(const GURL& url) const {
+  if (!is_ready_) {
+    return true;
```
StephenHeaps: Should add a small comment explaining why this returns true:
https://github.com/brave/brave-core/blob/60356f81c3a420a9470f60e3fa4eb7606d818493/components/brave_user_agent/browser/brave_user_agent_exceptions.cc#L70-L71

components/playlist/core/browser/playlist_exclusion.cc
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+  if (!url.SchemeIsHTTPOrHTTPS()) {
+    return true;
```
StephenHeaps, line 138: Hmm something feels off here. `CanResolvePageSrcLater` returns true if the url isn't http/https? Do we even show playlist on non-http/https pages?

components/playlist/core/browser/playlist_exclusion.h
EliHini: That format was for the `ListPlaylistExclusions()` debug helper — each row was` registrable_domain + "\t"` + path prefix. It is no longer needed.

components/playlist/core/browser/playlist_exclusion.h
EliHini: Removed

components/playlist/core/browser/playlist_exclusion.cc
EliHini: It is redundant; it was to force an early return, but it is unnecessary. Non-web URLs won't match any exclusion rules anyway. I've removed it

components/playlist/core/browser/playlist_exclusion_unittest.cc
```diff
@@ -0,0 +1,90 @@
...
+  EXPECT_FALSE(exclusions->CanResolvePageSrcLater(
+      GURL("https://www.youtube.com/@Example")));
+  EXPECT_TRUE(exclusions->CanResolvePageSrcLater(
+      GURL("https://www.youtube.com/watch?v=1")));
```
StephenHeaps, line 88: I think would be good to include a test for a `registrable_domain` not on the list too

components/playlist/core/browser/playlist_exclusion_unittest.cc
StephenHeaps, line 1: Are these unit tests being run?

kylehickinson: They wont be for iOS because `//brave/components/playlist/core/browser:unit_tests` is not part of `component_unit_tests`

StephenHeaps: Yeah, confirmed they're running for desktop platforms under `brave_unit_tests`. But I think these could be added to `component_unit_tests` so iOS can run them too.

components/playlist/core/browser/playlist_exclusion.cc
```diff
@@ -0,0 +1,160 @@
...
+      net::registry_controlled_domains::GetDomainAndRegistry(
+          url, net::registry_controlled_domains::INCLUDE_PRIVATE_REGISTRIES);
+
+  const std::string_view path = url.path();
+
+  for (const PlaylistResolveRule& rule : rules_) {
+    if (rule.registrable_domain != domain) {
+      continue;
+    }
+    if (rule.deny_root_path && (path.empty() || path == "/")) {
```
netzenbot, line 151: nit: `PostTaskAndReplyWithResult` here only specifies `base::MayBlock()` but omits `base::TaskPriority` and shutdown behavior. Per [CSM-022](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-memory.md#CSM-022), prefer e.g. `{base::MayBlock(), base::TaskPriority::BEST_EFFORT, base::TaskShutdownBehavior::SKIP_ON_SHUTDOWN}` so the scheduler and shutdown logic have explicit guidance.

components/playlist/core/browser/playlist_exclusion.h
```diff
@@ -0,0 +1,92 @@
...
+  bool is_ready_ = false;
+  base::WeakPtrFactory<PlaylistExclusions> weak_factory_{this};
+
+  friend struct base::DefaultSingletonTraits<PlaylistExclusions>;
+  friend class PlaylistExclusionsUnitTest;
+};
+
+}  // namespace playlist
+
+#endif  // BRAVE_COMPONENTS_PLAYLIST_CORE_BROWSER_PLAYLIST_EXCLUSION_H_
```
netzenbot, line 94: nit: `friend struct base::DefaultSingletonTraits<PlaylistExclusions>` and `friend class PlaylistExclusionsUnitTest` should be placed immediately after `private:`, before the constructor, methods, and member variables. See [CS-015](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-015).

components/playlist/core/browser/playlist_exclusion.h
```diff
@@ -0,0 +1,92 @@
...
+namespace playlist {
+
+class PlaylistResolveRule {
+ public:
+  PlaylistResolveRule();
+  ~PlaylistResolveRule();
+
+  PlaylistResolveRule(const PlaylistResolveRule&);
+  PlaylistResolveRule& operator=(const PlaylistResolveRule&);
+  PlaylistResolveRule(PlaylistResolveRule&&) noexcept;
```
netzenbot, line 45: nit: `GURL` is only used as a `const GURL&` parameter in `CanResolvePageSrcLater`, so the header can use `class GURL;` forward declaration instead of `#include "url/gurl.h"`. Move the full include to the `.cc` file. See [CS-014](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-014).

components/playlist/core/browser/playlist_exclusion.h
netzenbot, line 94: nit: The header doc only says "may be used for LivePlaylist-style reload / re-resolution" but omits the notable fallback: when the exclusions list isn't loaded yet, the method always returns `true`. Per [DOC-004](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-004), the header should describe the full contract so callers don't have to read the .cc to understand behavior.

ios/browser/playlist/playlist_exclusion_ios.h
```diff
@@ -0,0 +1,32 @@
...
+#ifndef BRAVE_IOS_BROWSER_PLAYLIST_PLAYLIST_EXCLUSION_IOS_H_
+#define BRAVE_IOS_BROWSER_PLAYLIST_PLAYLIST_EXCLUSION_IOS_H_
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+OBJC_EXPORT
+@protocol PlaylistExclusionsAPI
+@required
```
netzenbot: nit: `+ (id<PlaylistExclusionsAPI>)sharedInstance` is exposed to Swift but lacks `NS_SWIFT_NAME`. Consider `+ (id<PlaylistExclusionsAPI>)sharedInstance NS_SWIFT_NAME(shared);` so Swift callers get the idiomatic `.shared` accessor. See [IOS-033](https://github.com/brave/brave-core/tree/master/docs/best-practices/ios.md#IOS-033).

components/playlist/core/browser/playlist_exclusion_unittest.cc
EliHini: great, adding them now

components/playlist/core/browser/playlist_exclusion_unittest.cc
```diff
@@ -0,0 +1,93 @@
...
+  EXPECT_TRUE(exclusions->CanResolvePageSrcLater(GURL("https://youtube.com/")));
+}
+
+TEST_F(PlaylistExclusionsUnitTest, EmptyContentsNotReady) {
+  PlaylistExclusions* exclusions = PlaylistExclusions::GetInstance();
+  exclusions->OnPlaylistExclusionsLoaded("");
+  EXPECT_FALSE(exclusions->is_ready_);
+}
+
+TEST_F(PlaylistExclusionsUnitTest, InvalidJsonNotReady) {
```
netzenbot, line 47: nit: These three tests assert `exclusions->is_ready_` (private state) rather than observable public behavior. Per [TI-002](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-002), prefer testing the public API — e.g. `EXPECT_TRUE(exclusions->CanResolvePageSrcLater(GURL("https://example.com/")))` to verify the permissive fallback. The `NotReadyIsPermissive` test already demonstrates this pattern.

StephenHeaps:
Looking good. Is the server-side component piece deployed for testing on dev?

ios/browser/playlist/playlist_exclusion_ios+private.h
```diff
@@ -0,0 +1,18 @@
...
+
+#ifndef BRAVE_IOS_BROWSER_PLAYLIST_PLAYLIST_EXCLUSION_IOS_PRIVATE_H_
+#define BRAVE_IOS_BROWSER_PLAYLIST_PLAYLIST_EXCLUSION_IOS_PRIVATE_H_
+
+#include "brave/components/playlist/core/browser/playlist_exclusion.h"
+#include "brave/ios/browser/playlist/playlist_exclusion_ios.h"
+
+@interface PlaylistExclusionsAPIImpl : NSObject <PlaylistExclusionsAPI>
+- (instancetype)initWithPlaylistExclusions:
+    (playlist::PlaylistExclusions*)playlistExclusions;
```
netzenbot: nit: This private header is missing `NS_ASSUME_NONNULL_BEGIN`/`NS_ASSUME_NONNULL_END`. Per [IOS-030](https://github.com/brave/brave-core/tree/master/docs/best-practices/ios.md#IOS-030), all Objective-C headers should wrap their declarations with these macros.

EliHini:
> Looking good. Is the server-side component piece deployed for testing on dev?

Not yet. I will update you once it is done

ios/browser/playlist/playlist_exclusion_ios+private.h
nuo-xu: valid comment. please address. 
also do we still do `+private.h`? or `_internal.h`? please confirm with @kylehickinson

kylehickinson: Naming overall is wrong based on new patterns, should be broken up into two things: 
- The Obj-C bridge (public protocol `playlist_exclusion_bridge.h`, private impl (`playlist_exclusion_bridge_impl.h/mm`) 
- The factory which returns a concrete implementation of the bridge

protocol should be named `XBridge` (with NS_SWIFT_NAME replacing it back to `X`)

EliHini: updated to match the newer patterns.

ios/browser/playlist/playlist_exclusion_bridge_impl.mm
```diff
@@ -0,0 +1,33 @@
...
+
+#include "brave/ios/browser/playlist/playlist_exclusion_bridge_impl.h"
+
+#include "base/memory/raw_ptr.h"
+#include "brave/components/playlist/core/browser/playlist_exclusion.h"
+#include "net/base/apple/url_conversions.h"
+#include "url/gurl.h"
+
+@implementation PlaylistExclusionsBridgeImpl {
+  raw_ptr<playlist::PlaylistExclusions> _playlistExclusions;  // NOT OWNED
```
netzenbot: nit: `_playlistExclusions` is always non-null (it holds the singleton) and is set once in `initWithPlaylistExclusions:` — consider `raw_ref<playlist::PlaylistExclusions>` instead of `raw_ptr<>` to express that this field must never be null. See [CSM-036](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-memory.md#CSM-036).

components/playlist/core/browser/playlist_exclusion.cc
```diff
@@ -0,0 +1,163 @@
...
+      }
+      if (base::StartsWith(path, prefix)) {
+        return false;
+      }
+    }
+  }
+  return true;
+}
+
+}  // namespace playlist
```
netzenbot, line 166: nit: `OnPlaylistExclusionsLoaded` has three uncommented early-return failure branches (`contents.empty()`, JSON parse failure, missing `rules` list). Per [CS-021](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-021), add a brief comment before each summarizing what it handles — e.g. `// Bail out if the component file was not loaded.` and `// Skip malformed JSON payloads.`

ios/browser/playlist/playlist_exclusion_bridge_factory.h
```diff
@@ -0,0 +1,27 @@
...
+
+#import <Foundation/Foundation.h>
+
+@protocol PlaylistExclusionsBridge;
+
+NS_ASSUME_NONNULL_BEGIN
+
+OBJC_EXPORT
+NS_SWIFT_NAME(PlaylistExclusionsFactory)
+@interface PlaylistExclusionsBridgeFactory : NSObject
```
StephenHeaps: nit: I would probably drop `Bridge` from the name here

StephenHeaps (approved):

ios/browser/playlist/playlist_exclusion_bridge_factory.h
kylehickinson: +1 as we're not bridging a C++ factory

ios/browser/playlist/playlist_exclusion_bridge_factory.h
```diff
@@ -0,0 +1,27 @@
...
+
++ (id<PlaylistExclusionsBridge>)sharedBridge;
```
kylehickinson: `sharedInstance` or `sharedPlaylistExclusions` (the "bridge" aspect is transparent here)

ios/browser/playlist/playlist_exclusion_bridge_factory.mm
```diff
@@ -0,0 +1,24 @@
...
+
++ (id<PlaylistExclusionsBridge>)sharedBridge {
+  static PlaylistExclusionsBridgeImpl* instance = nil;
+  static dispatch_once_t once;
+  dispatch_once(&once, ^{
+    instance = [[PlaylistExclusionsBridgeImpl alloc]
+        initWithPlaylistExclusions:*playlist::PlaylistExclusions::
+                                       GetInstance()];
+  });
+  return instance;
```
kylehickinson: Just return a new instance every time, no reason to hold onto a bridge wrapper indefinitely as it holds no state

ios/browser/playlist/playlist_exclusion_bridge.h
```diff
@@ -0,0 +1,23 @@
...
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+/// Returns whether `pageSrc` may be re-resolved later (e.g. via
+/// LivePlaylistWebLoader). Mirrors
+/// `PlaylistExclusions::CanResolvePageSrcLater`.
+NS_SWIFT_NAME(PlaylistExclusions)
+@protocol PlaylistExclusionsBridge
+- (bool)canResolvePageSrcLater:(NSURL*)url;
```
kylehickinson: Use `BOOL` over `bool` for public headers (gets some special treatment during Obj-C > Swift import)

ios/browser/playlist/playlist_exclusions.h
```diff
@@ -0,0 +1,22 @@
...
+#define BRAVE_IOS_BROWSER_PLAYLIST_PLAYLIST_EXCLUSIONS_H_
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+/// Returns whether `pageSrc` may be re-resolved later (e.g. via
+/// LivePlaylistWebLoader). Mirrors
+/// `PlaylistExclusions::CanResolvePageSrcLater`.
+@protocol PlaylistExclusions
```
kylehickinson: `PlaylistExclusionsBridge` w/ `NS_SWIFT_NAME`?

ios/browser/playlist/playlist_exclusions_factory.h
```diff
@@ -0,0 +1,27 @@
...
+#define BRAVE_IOS_BROWSER_PLAYLIST_PLAYLIST_EXCLUSIONS_FACTORY_H_
+
+#import <Foundation/Foundation.h>
+
+@protocol PlaylistExclusions;
+
+NS_ASSUME_NONNULL_BEGIN
+
+OBJC_EXPORT
+NS_SWIFT_NAME(PlaylistExclusionsFactory)
```
kylehickinson: redundant, not changing anything

ios/browser/playlist/playlist_exclusions_impl.h
```diff
@@ -0,0 +1,28 @@
...
+
+#include "brave/ios/browser/playlist/playlist_exclusions.h"
+
+namespace playlist {
+class PlaylistExclusions;
+}
+
+NS_ASSUME_NONNULL_BEGIN
+
+OBJC_EXPORT
```
kylehickinson: not exporting this, its only an implementation

ios/browser/playlist/playlist_exclusions_bridge_impl.h
```diff
@@ -0,0 +1,28 @@
...
+}
+
+NS_ASSUME_NONNULL_BEGIN
+
+OBJC_EXPORT
+@interface PlaylistExclusionsImpl : NSObject <PlaylistExclusions>
+
+- (instancetype)init NS_UNAVAILABLE;
+- (instancetype)initWithPlaylistExclusions:
+    (playlist::PlaylistExclusions&)playlistExclusions NS_DESIGNATED_INITIALIZER;
```
kylehickinson, line 21: any reason to use a ref (&) instead of a pointer here since you're using raw_ptr?

EliHini: Yes, the & is intentional and separate from why the ivar is `raw_ptr`.

`init` takes a `ref` because the dependency is mandatory and always valid (`*GetInstance()` from a process singleton). That matches [CSM-036](https://github.com/brave/brave-core/blob/master/docs/best-practices/coding-standards-memory.md#CSM-036) and the same pattern as `serp_metrics_bridge_impl`.

The ivar stays `raw_ptr` because ObjC ivars are zero-initialized before `init`, so `raw_ref` isn’t usable there.  `&playlistExclusions` is stored once in init.

See netzenbot’s [CSM-036 comment](https://github.com/brave/brave-core/pull/36545#discussion_r3314613782) to express the non-null singleton dependency more clearly.

ios/browser/playlist/playlist_exclusions.h
EliHini: Done. matches `serp_metrics` / `brave_origin` convention.

ios/browser/playlist/playlist_exclusions_factory.h
EliHini: Removed redundant `NS_SWIFT_NAME`. Kept `OBJC_EXPORT` on the public factory

ios/browser/playlist/playlist_exclusions_impl.h
EliHini: Removed.

ios/browser/playlist/playlist_exclusions_bridge.h
kylehickinson, line 1: file name should also be `playlist_exclusions_bridge.h/mm`

ios/browser/playlist/playlist_exclusions_impl.h
```diff
@@ -14,8 +14,7 @@ class PlaylistExclusions;
 
 NS_ASSUME_NONNULL_BEGIN
 
-OBJC_EXPORT
-@interface PlaylistExclusionsImpl : NSObject <PlaylistExclusions>
+@interface PlaylistExclusionsImpl : NSObject <PlaylistExclusionsBridge>
```
kylehickinson: `BridgeImpl`

ios/browser/playlist/playlist_exclusions_bridge_impl.h
kylehickinson, line 1: `bridge_impl.h/mm`

kylehickinson (approved):

components/playlist/core/browser/playlist_exclusion.cc
```diff
@@ -0,0 +1,166 @@
...
+      base::BindOnce(&brave_component_updater::GetDATFileAsString,
+                     component_path_.AppendASCII(kPlaylistExclusionsFile)),
+      base::BindOnce(&PlaylistExclusions::OnPlaylistExclusionsLoaded,
+                     weak_factory_.GetWeakPtr()));
+}
+
+void PlaylistExclusions::OnPlaylistExclusionsLoaded(
+    const std::string& contents) {
+  rules_.clear();
+  is_ready_ = false;
```
bridiver, line 99: what is the purpose of this `is_ready_` flag? It gets set to false here and then true at the end of the method, but nothing can happen in the intervening time unless this class is called from more than one thread, but this is not a safe way to handle that case

components/playlist/core/browser/playlist_exclusion.cc
```diff
@@ -0,0 +1,166 @@
...
+// static
+PlaylistExclusions* PlaylistExclusions::GetInstance() {
+  return base::Singleton<PlaylistExclusions>::get();
+}
+
+PlaylistExclusions::PlaylistExclusions() = default;
+
+PlaylistExclusions::~PlaylistExclusions() = default;
+
+void PlaylistExclusions::OnComponentReady(const base::FilePath& component_dir) {
```
bridiver, line 83: this should probably be `LoadPlaylistExclusions` and the installer should pass the full path to the file

components/playlist/core/browser/playlist_exclusion.cc
```diff
@@ -0,0 +1,166 @@
...
+  rules_.clear();
+  is_ready_ = false;
+
+  // Bail out if the component file was not loaded.
+  if (contents.empty()) {
+    return;
+  }
+
+  std::optional<base::DictValue> root =
+      base::JSONReader::ReadDict(contents, base::JSON_PARSE_RFC);
```
bridiver, line 107: parsing of the json data should also happen on the task runner

components/playlist/core/browser/playlist_exclusion.cc
EliHini: `is_ready_` means we have successfully loaded and parsed the exclusions JSON, not a cross-thread lock. `CanResolvePageSrcLater` stays permissive (returns `true` for all URLs) until then so playlist behaviour is not blocked while the component is downloading or parsing.

The `is_ready_ = false` at the start of the load callback is redundant. However, the reply always runs on the main thread after the background work finishes, so nothing can observe an intermediate state. I'm removing that and using `clear is_ready_ (and rules_)` instead when `LoadPlaylistExclusions` starts, so a component update goes back to permissive mode while a new file is loading.

components/playlist/core/browser/playlist_exclusion.cc
EliHini: Done, updated in https://github.com/brave/brave-core/pull/36790/changes/c47a25449b28b177ea1495b4d2a96b9a12089166

components/playlist/core/browser/playlist_exclusion.cc
EliHini: Done, updated in https://github.com/brave/brave-core/pull/36790/changes/c47a25449b28b177ea1495b4d2a96b9a12089166

components/playlist/core/browser/playlist_exclusion.cc
bridiver: but if this class only gets called on a single thread (which is neceesary since it's not thread safe) then nothing can happen between setting it to false and true because they both happen in the same method and I don't see any re-entrant calls

bridiver: oh sorry, because of the early returns? But still what purpose does it serve? In fact actually it seems like a problem because a new list that was malformed would prevent an existing correctly formed list from continuing to work

bridiver: I guess we could do one of two things here, but neither of them requires a separate flag:
1. continue to use the last valid list for the remainder of the session.
or
2. just let `rules_.clear();` handle things because `is_ready_ == false` is equivalent to `rules_.empty()`, right?

EliHini: For component updates, should a failed reload keep the last good list (path 1), or go permissive (current path 2 + clear)?

EliHini: Went with path 1 in commit https://github.com/brave/brave-core/pull/36790/changes/7a9372d40c2fb01159027454edf8cc1acf73b0f2.

PR #36517, opened 2026-05-18
[AI Chat] Move HandoffButton lower and make it smaller

petemill:
Brave's URL bar is centered, so it makes the URL unreadable in more cases than upstream. And Brave's AI Chat conversation UI has large buttons for pausing the tasks, which perform the same function.

- Fix https://github.com/brave/brave-browser/issues/51384

PR #36553, opened 2026-05-19
[AI Chat] Move HandoffButton lower and make it smaller (plaster version)

### LINKED ISSUE #51384: [AI browsing] `Take over task` modal might hide omnibox and URL
### Description

In some cases, the URL might be hidden during AI browsing mode by the `Take over task` modal. 

### Steps to reproduce

1. Start an AI browsing task in AI browsing profile (example: `buy me potatoes online`).
2. Switch to tab where agent is acting

### Actual result

`Take over task` modal (or any AI browsing-related modal) might overlap with omnibox, depending on screen size.

Image

### Expected result

The overlay should never hide the URL.

### Reproduces how often

Easily reproduced

1.87.65 Chromium: 143.0.7499.40 (Official Build) nightly (arm64) 

### Channel information

- [ ] release (stable)
- [ ] beta
- [x] nightly

### Reproducibility

### Miscellaneous information

cc @petemill @mattmcalister @aguscruiz

petemill:
Brave's URL bar is centered, so it makes the URL unreadable in more cases than upstream. And Brave's AI Chat conversation UI has large buttons for pausing the tasks, which perform the same function.

- Fix https://github.com/brave/brave-browser/issues/51384

Alternative to:
- https://github.com/brave/brave-core/pull/36517

Plaster seems good for this because we're largely swapping some values for another. The alternative is either direct patching, or subclassing and fragile tree walking to adjust already-addded borders (also with slight performance hit). The bounds nudging could be done by subclassing but it would need to access private variables so there is more complication there where this plaster will do. If it doesn't in the future then we can consider subclassing just for that one.


rewrite/chrome/browser/actor/ui/handoff_button_controller.cc.toml
```diff
@@ -0,0 +1,38 @@
...
+      const SkScalar pos[] = {0.0f, 0.5f, 1.0f};'''
+
+[[substitution]]
+description = '''Tighten handoff button content padding.
+
+Brave's location bar centers its URL text, so the centered handoff button
+covers it more than upstream; a smaller button is less obtrusive.
+'''
+pattern = 'gfx::Insets::TLBR(10, 10, 10, 14)'
+replace = 'gfx::Insets::TLBR(4, 6, 4, 10)'
```
netzenbot: nit: `pattern = 'gfx::Insets::TLBR(10, 10, 10, 14)'` includes function arguments rather than a simple symbol name. Per [PLSTR-002](https://github.com/brave/brave-core/tree/master/docs/best-practices/plaster.md#PLSTR-002), `pattern` is intended for simple symbol names (e.g., a single constant or method name); for anything with more context, `pattern_re` is more appropriate.

rewrite/chrome/browser/actor/ui/handoff_button_controller.cc.toml
```diff
@@ -0,0 +1,45 @@
...
+Anchors on the `HandoffButtonController::GetHandoffButtonBounds()` function
+signature and its return statement. The body between them is captured and
+emitted verbatim via `\1`; only the return is rewritten to wrap `y` with the
+Brave offset, conditioned on the function-local `is_tab_strip_visible`. This
+keeps the substitution scoped to one specific function and one specific
+statement, so unrelated returns of the same shape elsewhere in the file
+can't be mis-targeted.
+'''
+re_pattern = '(gfx::Rect HandoffButtonController::GetHandoffButtonBounds\(\)[\s\S]+?)return gfx::Rect\(\{x, y\}, preferred_size\);'
+replace = '\1return gfx::Rect({x, y + (is_tab_strip_visible ? kHandoffButtonPreferredHeight / 3 : 0)}, preferred_size);'
```
petemill: I believe this is better than subclassing `HandoffButtonController` and overriding `GetHandoffButtonBounds` because is_tab_strip_visible` comes from `tab_interface_` which is `private` and managed in a couple lifecycle places. I wouldn't want to duplicate those lifecycles in the subclass because we could miss a new lifecycle method in the future. If we still want to subclass instead, we'd need to inject the subclassing, add virtual to GetHandoffButtonBounds and make `tab_interface_`  _protected_ instead of _private_.

This seems cleaner to me for now and does not target the body of the function but the return of this specific function.

rewrite/chrome/browser/actor/ui/handoff_button_controller.cc.yaml
```diff
@@ -0,0 +1,51 @@
...
+substitutions:
+  - description: |-
+      Use Brave's gradient colors and stops for the handoff button shadow.
+
+      Replaces both branches of upstream's `kActorUiThemed` feature flag with an
+      unconditional Brave sweep gradient. Anchors on the `colors` declaration
+      and the `pos` array name (not their values) so upstream tweaks to the
+      stop positions don't break the match.
+    re_pattern:
+      'std::vector<SkColor> colors;.+?const SkScalar pos\[\] = \{[^}]+\};'
```
goodov, line 15: my gut feeling is to always use yaml's `|-` syntax to avoid quoting patterns and replaces. This is the noise that we don't need in those rewrite files.

maybe even enforce this via presubmit. cc @cdesouza-chromium

goodov (approved):

rewrite/chrome/browser/actor/ui/handoff_button_controller.cc.yaml
petemill: we can address with formatting afterwards if that's desired and a rule is introduced

rewrite/chrome/browser/actor/ui/handoff_button_controller.cc.toml
petemill: I would like this to break if it changes

PR #36557, opened 2026-05-19
[Brave News]: Enable image cache on BraveSanitizedImageSource

### LINKED ISSUE #48967: [perf] Cache reencoded images for Brave News
Currently, every time we open NTP we re-encode the Brave News images to show via a sandboxed process.
It takes CPU time & battery and slows down showing the images.
It makes sense to cache the images on a disk. The best option is to use the same HTTP cache as for usually resources.

Image

fallaciousreasoning:
Resolves https://github.com/brave/brave-browser/issues/48967

@atuchin-m I'm a bit unsure about this as the comment on the class makes it sound as if this would be stored in memory, rather than the HTTP cache?

### DIFF
--- a/browser/ui/webui/brave_sanitized_image_source.cc
+++ b/browser/ui/webui/brave_sanitized_image_source.cc
@@ -201,7 +201,7 @@ BraveSanitizedImageSource::BraveSanitizedImageSource(
 }
 
 bool BraveSanitizedImageSource::AllowCaching() {
-  return false;
+  return true;
 }
 std::string BraveSanitizedImageSource::GetSource() {
   return kChromeUIBraveImageHost;
--- a/browser/ui/webui/brave_sanitized_image_source.h
+++ b/browser/ui/webui/brave_sanitized_image_source.h
@@ -35,7 +35,6 @@ class SimpleURLLoader;
 // * supporting target_size. Resized images take less CPU to encode and
 //   less memory to store;
 // * using different encoding method: PNG FastEncodeBGRASkBitmap.
-// * disabling caching (saving renderer memory);
 // * USER_VISIBLE_PRIORITY.
 class BraveSanitizedImageSource : public content::URLDataSource {
  public:


browser/ui/webui/brave_sanitized_image_source.cc
```diff
@@ -201,7 +201,7 @@ BraveSanitizedImageSource::BraveSanitizedImageSource(
 }
 
 bool BraveSanitizedImageSource::AllowCaching() {
-  return false;
+  return true;
```
fallaciousreasoning, line 204: this seems to control the `Cache-Control` header:
https://source.chromium.org/chromium/chromium/src/+/main:content/browser/webui/url_data_manager_backend.cc;l=220

atuchin-m: @fallaciousreasoning It also controls in-memory cache. Enabling this double the memory usage.
Currently, HTTP cache is used to store the original images.

I'm not agains enabling caching here, but let's exclude in memory cache somehow.

netzenbot (approved):

browser/ui/webui/brave_sanitized_image_source.cc
atuchin-m: BTW, does this change actually cache the decoded images in HTTP cache?

fallaciousreasoning: It definitely seems to be using the HTTP memory cache:
image

fallaciousreasoning: Hey so I just did a quick test - it doesn't seem to have much effect on the memory usage. For example, with the flag as true my NTP settled at 128mb after scrolling down ~4 pages of results.

Doing the same with the flag as false resulted in 126mb after about ~4 pages of results. This is pretty unscientific but its definitely not double.

Additionally, I think the browser will evict things from its memory cache if things get under pressure, so are you sure this is a big deal?

fallaciousreasoning: I had a super quick play around trying to get it to store things in the disk cache but it doesn't seem obvious

atuchin-m: I believe in-memory caching doesn't make a lot of sense.
The thing what we really need is disk caching.

PR #36561, opened 2026-05-20
[FocusMode] Add edge reveal supporting classes

zenparsing:

See https://github.com/brave/brave-core/pull/36461 for the Focus Mode prototype and implementation overview.

### DIFF
--- a/browser/ui/views/frame/edge_reveal/BUILD.gn
+++ b/browser/ui/views/frame/edge_reveal/BUILD.gn
@@ -1,0 +1,40 @@
+# Copyright (c) 2026 The Brave Authors. All rights reserved.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+source_set("edge_reveal") {
+  sources = [
+    "edge_hover_detector.cc",
+    "edge_hover_detector.h",
+    "edge_reveal_watcher.cc",
+    "edge_reveal_watcher.h",
+  ]
+
+  public_deps = [
+    "//base",
+    "//ui/views",
+  ]
+
+  deps = []
+}
+
+source_set("unit_tests") {
+  testonly = true
+
+  sources = [
+    "edge_hover_detector_unittest.cc",
+    "edge_reveal_watcher_unittest.cc",
+  ]
+
+  deps = [
+    ":edge_reveal",
+    "//base/test:test_support",
+    "//testing/gmock",
+    "//testing/gtest",
+    "//ui/events",
+    "//ui/gfx/geometry",
+    "//ui/views",
+    "//ui/views:test_support",
+  ]
+}
--- a/browser/ui/views/frame/edge_reveal/edge_hover_detector.cc
+++ b/browser/ui/views/frame/edge_reveal/edge_hover_detector.cc
@@ -1,0 +1,86 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "brave/browser/ui/views/frame/edge_reveal/edge_hover_detector.h"
+
+#include <utility>
+
+#include "base/check.h"
+#include "base/functional/bind.h"
+#include "ui/display/screen.h"
+#include "ui/events/event.h"
+#include "ui/events/types/event_type.h"
+#include "ui/views/event_monitor.h"
+
+EdgeHoverDetector::EdgeHoverDetector(
+    views::Widget* host,
+    IsHoverPointCallback is_hover_point,
+    HoverStateChangedCallback hover_state_changed)
+    : is_hover_point_(std::move(is_hover_point)),
+      hover_state_changed_(std::move(hover_state_changed)) {
+  CHECK(host);
+  CHECK(is_hover_point_);
+  CHECK(hover_state_changed_);
+
+  host_observation_.Observe(host);
+
+  event_monitor_ = views::EventMonitor::CreateWindowMonitor(
+      this, host->GetNativeWindow(),
+      {ui::EventType::kMouseMoved, ui::EventType::kMouseExited});
+}
+
+EdgeHoverDetector::~EdgeHoverDetector() = default;
+
+void EdgeHoverDetector::DetectHoverState() {
+  if (!host_observation_.IsObserving()) {
+    return;
+  }
+  ApplyWithDelay(IsCursorHovering());
+}
+
+void EdgeHoverDetector::HandleEventForTesting(const ui::Event& event) {
+  OnEvent(event);
+}
+
+void EdgeHoverDetector::OnWidgetDestroying(views::Widget* widget) {
+  event_monitor_.reset();
+  timer_.Stop();
+  host_observation_.Reset();
+}
+
+void EdgeHoverDetector::OnEvent(const ui::Event& event) {
+  if (event.type() == ui::EventType::kMouseExited) {
+    ApplyWithDelay(false);
+  } else if (event.type() == ui::EventType::kMouseMoved) {
+    ApplyWithDelay(IsCursorHovering());
+  }
+}
+
+bool EdgeHoverDetector::IsCursorHovering() {
+  return is_hover_point_.Run(display::Screen::Get()->GetCursorScreenPoint());
+}
+
+void EdgeHoverDetector::ApplyWithDelay(bool hovering) {
+  // If we are transitioning back to the current hover state (e.g. the mouse
+  // exits the edge very quickly after entering it), then cancel the toggle
+  // timer, if active.
+  if (hovering == hovering_) {
+    timer_.Stop();
+    return;
+  }
+  // If the toggle timer is already active, then let it proceed.
+  if (timer_.IsRunning()) {
+    return;
+  }
+  // Start the toggle timer. When it fires, the hover state will be toggled.
+  timer_.Start(FROM_HERE, hovering ? kMouseEnterDelay : kMouseExitDelay,
+               base::BindOnce(&EdgeHoverDetector::ToggleHoverState,
+                              base::Unretained(this)));
+}
+
+void EdgeHoverDetector::ToggleHoverState() {
+  hovering_ = !hovering_;
+  hover_state_changed_.Run(hovering_);
+}
--- a/browser/ui/views/frame/edge_reveal/edge_hover_detector.h
+++ b/browser/ui/views/frame/edge_reveal/edge_hover_detector.h
@@ -1,0 +1,84 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#ifndef BRAVE_BROWSER_UI_VIEWS_FRAME_EDGE_REVEAL_EDGE_HOVER_DETECTOR_H_
+#define BRAVE_BROWSER_UI_VIEWS_FRAME_EDGE_REVEAL_EDGE_HOVER_DETECTOR_H_
+
+#include <memory>
+
+#include "base/functional/callback.h"
+#include "base/scoped_observation.h"
+#include "base/time/time.h"
+#include "base/timer/timer.h"
+#include "ui/events/event_observer.h"
+#include "ui/gfx/geometry/point.h"
+#include "ui/views/widget/widget.h"
+#include "ui/views/widget/widget_observer.h"
+
+namespace views {
+class EventMonitor;
+}  // namespace views
+
+// Detects when the mouse is "hovering" in a consumer-defined region of a host
+// widget's window (typically one of the edges of the browser window) and emits
+// a debounced signal suitable for driving auto-hide UI reveal.
+class EdgeHoverDetector : public views::WidgetObserver,
+                          public ui::EventObserver {
+ public:
+  using IsHoverPointCallback =
+      base::RepeatingCallback<bool(const gfx::Point& screen_point)>;
+
+  using HoverStateChangedCallback =
+      base::RepeatingCallback<void(bool hovering)>;
+
+  // Debounce delays applied before committing a hover state change.
+  static constexpr base::TimeDelta kMouseEnterDelay = base::Milliseconds(60);
+  static constexpr base::TimeDelta kMouseExitDelay = base::Milliseconds(500);
+
+  // Constructs a new detector for the specified host widget window. The caller
+  // provides the following callbacks:
+  //   - `is_hover_point`: Returns true if the specified screen coordinate is
+  //     considered within the edge's active region.
+  //   - `hover_state_changed`: Called when the hover state has changed.
+  EdgeHoverDetector(views::Widget* host,
+                    IsHoverPointCallback is_hover_point,
+                    HoverStateChangedCallback hover_state_changed);
+
+  EdgeHoverDetector(const EdgeHoverDetector&) = delete;
+  EdgeHoverDetector& operator=(const EdgeHoverDetector&) = delete;
+  ~EdgeHoverDetector() override;
+
+  // Current committed hover state (post-debounce).
+  bool hovering() const { return hovering_; }
+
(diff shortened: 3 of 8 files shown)


browser/ui/views/frame/edge_reveal/edge_reveal_watcher.cc
```diff
@@ -0,0 +1,135 @@
...
+  return view && edge_contains_view_.Run(view);
+}
+
+bool EdgeRevealWatcher::IsAnchoredToEdge(views::Widget* widget) const {
+  auto* delegate = widget->widget_delegate();
+  if (!delegate) {
+    return false;
+  }
+  auto* bubble = delegate->AsBubbleDialogDelegate();
+  if (!bubble) {
```
netzenbot, line 95: `GetFocusManager()` can return `nullptr` (it does so when `GetTopLevelWidget()` returns null or when the focus manager hasn't been set up), and `ScopedObservation::Observe()` will crash if passed null. `Recompute()` correctly null-checks `GetFocusManager()`, but the constructor does not. Add a null-check here or assert that the widget always has a focus manager at construction time. ([✅ Null-Check Return Values Before Dereferencing](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-memory.md#CSM-029))

browser/ui/views/frame/edge_reveal/BUILD.gn
```diff
@@ -0,0 +1,77 @@
...
+    "edge_reveal_watcher_unittest.cc",
+  ]
+
+  deps = [
+    ":edge_reveal",
+    "//base",
+    "//base/test:test_support",
+    "//testing/gmock",
+    "//testing/gtest",
+    "//ui/events",
```
netzenbot, line 35: nit: `child_widget_tracker_test_support.h` defines a reusable fake (`FakeChildWidgetTracker`) and is already used by two different unittest files. Per [BS-022](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-022), it should live in a dedicated `test_support` source_set so other test targets (e.g., future tests in other directories) can depend on it without pulling in all of `unit_tests`.

browser/ui/views/frame/edge_reveal/edge_hover_detector.h
```diff
@@ -0,0 +1,79 @@
...
+  void OnWidgetDestroying(views::Widget* widget) override;
+
+  // ui::EventObserver:
+  void OnEvent(const ui::Event& event) override;
+
+  bool IsCursorHovering();
+  void ApplyWithDelay(bool hovering);
+  void OnTimerFired(bool hovering);
+
+  raw_ptr<views::Widget> host_;
```
netzenbot: nit: `raw_ptr<views::Widget> host_` should come after owning members like `std::unique_ptr<views::EventMonitor> event_monitor_`. [CSM-013](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-memory.md#CSM-013)

browser/ui/views/frame/edge_reveal/edge_reveal_watcher.h
```diff
@@ -0,0 +1,107 @@
...
+  void OnWidgetVisibilityChanged(views::Widget* widget, bool visible) override;
+  void OnWidgetDestroying(views::Widget* widget) override;
+
+  bool EdgeContainsView(const views::View* view) const;
+  bool IsAnchoredToEdge(views::Widget* widget) const;
+
+  void Recompute();
+
+  raw_ptr<views::Widget> host_widget_;
+  EdgeContainsViewCallback edge_contains_view_;
```
netzenbot, line 88: nit: `raw_ptr<views::Widget> host_widget_` should be declared after owning members like `std::unique_ptr<ChildWidgetTracker> tracker_`. [CSM-013](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-memory.md#CSM-013)

browser/ui/views/frame/edge_reveal/child_widget_tracker.h
```diff
@@ -0,0 +1,72 @@
...
+
+  raw_ptr<views::Widget> host_;
+  base::ScopedObservation<views::Widget, views::WidgetObserver>
+      host_observation_{this};
+  base::ScopedMultiSourceObservation<views::Widget, views::WidgetObserver>
+      tracked_widget_observations_{this};
+  base::ObserverList<Observer> observers_;
+};
+
+#endif  // BRAVE_BROWSER_UI_VIEWS_FRAME_EDGE_REVEAL_CHILD_WIDGET_TRACKER_H_
```
netzenbot: nit: Per [CS-013](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-013), method definitions should go in `.cc` files, not headers — simple inline getters are the accepted exception. `OnHostDestroying()` is a virtual callback, not a getter; move its empty body to `child_widget_tracker.cc`.

browser/ui/views/frame/edge_reveal/child_widget_tracker.h
```diff
@@ -0,0 +1,72 @@
...
+// events to observers.
+//
+// "Child" here is the platform-native parent/child window relationship, not the
+// views view-tree parent relationship:
+//   - aura: top-level views::Widgets whose `parent()` is the host.
+//   - macOS: child NSWindows of the host's NSWindow.
+//
+// Widgets reported here include anchored bubbles, menus, the omnibox
+// autocomplete popup, drag helpers, etc.
+class ChildWidgetTracker : public views::WidgetObserver {
```
simonhong: Curious why we need platform specific subclass(aura, mac) for tracking child widget add/remove.
We can get it from WidgetObserver. Isn't it sufficient for focus mode?

zenparsing: I think it works! Thanks so much ❤️

browser/ui/views/frame/edge_reveal/edge_hover_detector.h
```diff
@@ -0,0 +1,79 @@
...
+  IsHoverPointCallback is_hover_point_;
+  HoverStateChangedCallback hover_state_changed_;
+  std::unique_ptr<views::EventMonitor> event_monitor_;
+  base::OneShotTimer timer_;
+  bool hovering_ = false;
+  base::ScopedObservation<views::Widget, views::WidgetObserver>
+      host_observation_{this};
+};
+
+#endif  // BRAVE_BROWSER_UI_VIEWS_FRAME_EDGE_REVEAL_EDGE_HOVER_DETECTOR_H_
```
netzenbot, line 84: nit: The constructor has no doc comment explaining what `is_hover_point` and `hover_state_changed` represent or any ownership/lifetime expectations. `EdgeRevealWatcher`'s constructor has this; it would be consistent to add one here too. ([DOC-004](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-004))

browser/ui/views/frame/edge_reveal/edge_hover_detector.cc
```diff
@@ -0,0 +1,86 @@
...
+      this, host_->GetNativeWindow(),
+      {ui::EventType::kMouseMoved, ui::EventType::kMouseExited});
+}
+
+EdgeHoverDetector::~EdgeHoverDetector() = default;
+
+void EdgeHoverDetector::Update() {
+  if (!host_) {
+    return;
+  }
```
simonhong: It seems `host_` is always not null.
```suggestion
  CHECK(host);
```

browser/ui/views/frame/edge_reveal/edge_hover_detector.h
simonhong: nit: it's recommended to use `raw_ref` if it's always non-null.

browser/ui/views/frame/edge_reveal/edge_hover_detector.cc
```diff
@@ -0,0 +1,86 @@
...
+  timer_.Stop();
+  host_observation_.Reset();
+  host_ = nullptr;
+}
+
+void EdgeHoverDetector::OnEvent(const ui::Event& event) {
+  if (event.type() == ui::EventType::kMouseExited) {
+    ApplyWithDelay(false);
+  } else if (event.type() == ui::EventType::kMouseMoved) {
+    Update();
```
simonhong: why not just calling `ApplyWithDelay(IsCursorHovering());` here?

browser/ui/views/frame/edge_reveal/edge_hover_detector.cc
```diff
@@ -0,0 +1,86 @@
...
+  }
```
simonhong, line 58: IMO, I think we don't need to make EdgeHoverDetector/EdgeRevealWatcher too general because focus mode will only use them.
If these classes has more descriptive method name/comment regarding focus mode UX, it would be really helpful to understand how it's used.
For example, I'm curious what is "host" and why we need to handle its kMouseExited event in focus mode scenario.

browser/ui/views/frame/edge_reveal/edge_hover_detector.cc
zenparsing: We could do that - that's exactly what `Update` does. It seemed better to not repeat however.

browser/ui/views/frame/edge_reveal/edge_hover_detector.cc
zenparsing: This pointer is set to null if the widget is destroyed before the detector.

browser/ui/views/frame/edge_reveal/edge_hover_detector_unittest.cc
```diff
@@ -0,0 +1,213 @@
...
+
+class EdgeHoverDetectorTest : public views::test::WidgetTest {
+ public:
+  EdgeHoverDetectorTest()
+      : views::test::WidgetTest(
+            base::test::TaskEnvironment::TimeSource::MOCK_TIME) {}
+
+  void SetUp() override {
+    WidgetTest::SetUp();
+    widget_ = CreateTopLevelNativeWidget();
```
simonhong: how about pass `CLIENT_OWNS_WIDGET` explicitely like you did for EdgeRevealWatcherTest?
All other types are deprecated except `CLIENT_OWNS_WIDGET`.

browser/ui/views/frame/edge_reveal/edge_hover_detector.cc
simonhong: I think  we can't call `Update()` when `host_` is null because `event_monitor_` is destroyed before setting null to `host_`.

browser/ui/views/frame/edge_reveal/edge_hover_detector.cc
simonhong: I mentioned this because `Update()` is only called from here in production.
If you have it for calling from test, ~~I think calling `OnEvent()` from test seems sufficient?~~
We already have `HandleEventForTesting()`.
Also `Update()` method name is ambiguous.

browser/ui/views/frame/edge_reveal/edge_hover_detector.cc
```diff
@@ -0,0 +1,86 @@
...
+}
+
+bool EdgeHoverDetector::IsCursorHovering() {
+  return is_hover_point_.Run(display::Screen::Get()->GetCursorScreenPoint());
+}
+
+void EdgeHoverDetector::ApplyWithDelay(bool hovering) {
+  if (hovering == hovering_) {
+    timer_.Stop();
```
simonhong, line 70: Is it possible that `timer_` is running when `hovering == hovering_` is true?
If so, we could just return here?

browser/ui/views/frame/edge_reveal/edge_hover_detector.cc
```diff
@@ -0,0 +1,86 @@
...
+  }
+  timer_.Start(FROM_HERE, hovering ? kMouseEnterDelay : kMouseExitDelay,
+               base::BindOnce(&EdgeHoverDetector::OnTimerFired,
+                              base::Unretained(this), hovering));
+}
+
+void EdgeHoverDetector::OnTimerFired(bool hovering) {
+  if (hovering != IsCursorHovering() || hovering == hovering_) {
+    return;
+  }
```
simonhong: How about having two timers? one for hover and the other for exit
If we have different timer for hover and exit, I think logic could be more simpler.
Then, we could simply update tate and run callback w/o this complicated state checking when fired.

we can simply reset hovering timer when exit event arrived. and vice versa. WDYT?

browser/ui/views/frame/edge_reveal/edge_hover_detector.cc
```diff
@@ -0,0 +1,86 @@
...
+}
+
+void EdgeHoverDetector::ApplyWithDelay(bool hovering) {
+  if (hovering == hovering_) {
+    timer_.Stop();
+    return;
+  }
+  if (timer_.IsRunning()) {
+    return;
+  }
```
simonhong, line 76: At this point, we could cancel easily if we know `timer_` is for hovering or exit.

browser/ui/views/frame/edge_reveal/edge_reveal_watcher.cc
```diff
@@ -0,0 +1,124 @@
...
+    return false;
+  }
+  views::View* anchor = bubble->GetAnchorView();
+  return anchor && EdgeContainsView(anchor);
```
simonhong:
```suggestion
  return EdgeContainsView(anchor);
```

browser/ui/views/frame/edge_reveal/edge_reveal_watcher.h
```diff
@@ -0,0 +1,99 @@
...
+#include "ui/views/widget/widget_observer.h"
+
+namespace views {
+class View;
+}  // namespace views
+
+// Watches a host widget and reports, via a callback, the set of reasons that an
+// edge may need to be revealed. Intended for use by edge-mounted, auto-hiding
+// browser "chrome".
+class EdgeRevealWatcher : public views::FocusChangeListener,
```
simonhong, line 26: Can you add comment about what edge means? it's View, Widget, host's specific region or anything? (detector class also.) If it's decided by client, just its concept would be fine.

browser/ui/views/frame/edge_reveal/edge_reveal_watcher.cc
```diff
@@ -0,0 +1,124 @@
...
+    return;
+  }
+  if (!child_widget_observations_.IsObservingSource(child)) {
+    child_widget_observations_.AddObservation(child);
+    Recompute();
+  }
+}
+
+void EdgeRevealWatcher::OnWidgetChildRemoved(views::Widget* widget,
+                                             views::Widget* child) {
```
simonhong, line 75:
```
  if (widget != host_widget_) {
    return;
  }
```

browser/ui/views/frame/edge_reveal/edge_reveal_watcher.h
zenparsing: I've expanded the comment.

browser/ui/views/frame/edge_reveal/edge_hover_detector.cc
zenparsing: Yes, that is possible because of debouncing:

* `hovering_` is false.
* Mouse enters "hot zone" at top of window; `ApplyWithDelay(true)`.
* Timer is activated to potentially set `hovering_` to true.
* Mouse exits "hot zone"; `ApplyWithDelay(false)`.
* `hovering` (false) == `hovering_` (false).
* Timer is cancelled.

I'll take a closer look into how we might make this more obvious. I'm not sure that having two timers will help, but I will consider that as an option.

browser/ui/views/frame/edge_reveal/edge_hover_detector.cc
zenparsing: Renamed to `DetectHoverState`, and I've expanded the comment to indicate why it's a public method.

browser/ui/views/frame/edge_reveal/edge_hover_detector.h
zenparsing: The raw_ptr has been removed in favor of simply using `host_observation_.IsObserving()`.

browser/ui/views/frame/edge_reveal/edge_hover_detector.cc
zenparsing: `raw_ptr` removed.

browser/ui/views/frame/edge_reveal/edge_hover_detector.cc
zenparsing: I've added some comments and simplified things just a bit. Does that help?

netzenbot (approved):

simonhong (approved):
++

PR #36563, opened 2026-05-20
[plaster] Apply on change

### LINKED ISSUE #45057: 🩹 Support applying only overlay files that changed
At the moment, we process all overlay files with plaster irrespective of them being changed or not. We should make the default behaviour to only process files that need to be updated, either because the overlay file itself changed, or the source changed.

cdesouza-chromium:
This PR introduces a check mechanism to `plaster` that will allow it by
default to just reapply patches for the plasters who are not up-to-date.
This is being implemented with a simple approach:

 * Check if the `.patchinfo` timestamp older than the plaster file,
   patch, or upstream source.
 * If `.patchinfo` is newer than any of these, we go on to check if the
   checksums are matching with what is in the `.patchinfo`.
 * If any of the checksum chcks fail, we reapply the patch.

This is very lightweight, and will make most of the uses of `apply`
reliable for reruns while working on a feature.

Resolves https://github.com/brave/brave-browser/issues/45057

### DIFF
--- a/tools/cr/plaster.py
+++ b/tools/cr/plaster.py
@@ -84,13 +84,111 @@ class PathChecksumPair:
         return True
 
 
-@dataclass
-class PatchInfo:
-    """ Manages a .patchinfo file contents.
+@dataclass(frozen=True)
+class Patchinfo:
+    """In-memory representation of a parsed .patchinfo file.
+    """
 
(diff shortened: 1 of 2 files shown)


mkarolin (approved):
LGTM

PR #36565, opened 2026-05-20
Fix intermittent iPad Remember Browsing Mode

### LINKED ISSUE #55515: Remember Browsing Mode setting is not respected [iPadOS]

EliHini:
## Summary

There was a data race condition that occurred on consecutive iPad relaunches, private mode was inferred from Core Data tab state and scene restoration data that often hadn’t been flushed yet, and a failed launch could overwrite the saved private-mode flag before the next relaunch — the setting stayed enabled, but the app sometimes reopened in regular mode instead of private.

- Persist browsing mode reliably via a new `lastPrivateBrowsingMode` preference, scene session state, and synchronous tab/selection saves on background
- Improve launch resolution with `resolveLaunchIsPrivate()` and better `windowId` fallback when session state is stale
- Prevent launch-time state from overwriting saved private mode, and preserve browsing mode during tab restore when "Remember Browsing Mode" is enabled

## Root cause

On consecutive relaunches, private mode was inferred from Core Data tab selection and scene restoration state that often wasn't flushed yet. A failed launch could also write `isPrivate: false` back into scene state, breaking the next relaunch.

https://github.com/user-attachments/assets/d22935c1-5525-4468-9053-17a7074cc307

### Setup

- [ ] Use an **iPad** (simulator or device)
- [ ] Go to **Settings → Tabs → Private Tabs**
- [ ] Enable **Keep Private Tabs**
- [ ] Enable **Remember Browsing Mode**
- [ ] Open one or more **private tabs** and browse to a real page (not just NTP)

### Consecutive relaunch — private mode (primary repro)

- [ ] Confirm you are in **private browsing mode**
- [ ] Quit Brave (Home gesture / app switcher)
- [ ] Relaunch Brave immediately
- [ ] Confirm Brave reopens in **private mode** with private tabs restored
- [ ] Repeat background → relaunch **5–10 times in a row**, as quickly as possible
- [ ] Each relaunch should consistently reopen in **private mode**

### Consecutive relaunch — after mode switch

- [ ] Start in **private mode** with private tabs open
- [ ] Open tab grid and switch to **regular mode**
- [ ] Quit Brave and relaunch
- [ ] Confirm Brave reopens in **regular mode**
- [ ] Switch back to **private mode**, background, and relaunch 3–5 times
- [ ] Confirm Brave reopens in **private mode** each time

### Mixed tabs

- [ ] With both **regular and private tabs** open, select a **private tab**
- [ ] Quit and relaunch 3–5 times consecutively
- [ ] Confirm Brave reopens in **private mode** with the private tab(s) available

### Settings unchanged

- [ ] After several consecutive relaunches, open **Settings → Tabs → Private Tabs**
- [ ] Confirm **Keep Private Tabs** and **Remember Browsing Mode** are still **enabled**

### Regression — Remember Browsing Mode off

- [ ] Disable **Remember Browsing Mode** (keep **Keep Private Tabs** on)
- [ ] Open private tabs, quit, and relaunch
- [ ] Confirm app reopens in **regular mode**, but private tabs are still restorable via tab switcher

### Private Browsing Lock

- [ ] Enable **Private Browsing Lock**
- [ ] Relaunch from private mode and complete/cancel auth as appropriate
- [ ] Confirm behaviour matches lock setting (auth failure will fall back to regular mode by design)

### iPhone Regression Test
- [ ] Repeat the tests for iPhone
- [ ] Confirm behaviour is unchanged. Both iPad and iPhone now match.

Resolves https://github.com/brave/brave-browser/issues/55515

### DIFF
--- a/ios/brave-ios/Sources/Brave/Frontend/Browser/TabManager.swift
+++ b/ios/brave-ios/Sources/Brave/Frontend/Browser/TabManager.swift
@@ -614,7 +614,7 @@ class TabManager: NSObject {
     saveTabOrder()
   }
 
-  private func saveTabOrder(synchronously: Bool = false) {
+  private func saveTabOrder() {
     if Preferences.Privacy.privateBrowsingOnly.value
       || (privateBrowsingManager.isPrivateBrowsing
         && !Preferences.Privacy.persistentPrivateBrowsing.value)
@@ -622,7 +622,7 @@ class TabManager: NSObject {
       return
     }
     let allTabIds = allTabs.compactMap { $0.id }
-    SessionTab.saveTabOrder(tabIds: allTabIds, synchronously: synchronously)
+    SessionTab.saveTabOrder(tabIds: allTabIds)
   }
 
   @MainActor func configureTab(
@@ -701,7 +701,7 @@ class TabManager: NSObject {
     }
   }
 
-  func saveAllTabs(synchronously: Bool = false) {
+  func saveAllTabs() {
     if Preferences.Privacy.privateBrowsingOnly.value
       || (privateBrowsingManager.isPrivateBrowsing
         && !Preferences.Privacy.persistentPrivateBrowsing.value)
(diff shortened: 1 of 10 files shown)


ios/brave-ios/Sources/Brave/Frontend/Browser/TabManager.swift
```diff
@@ -703,15 +711,40 @@ class TabManager: NSObject {
...
-    )
+      )
+    }
+  }
+
+  /// Persists tab session data and tab order when the app backgrounds.
+  func persistSessionOnBackground() {
+    SessionWindow.createWindow(isSelected: true, uuid: windowId)
+    saveAllTabs(synchronously: true)
+    saveTabOrder(synchronously: true)
```
kylehickinson: We cant do this synchronously on background as it causes performance issues, we used to do this and it was changed to fix the app hanging when the user foregrounded closely after backgrounding

EliHini: Good to know. Switched background persistence back to async per review. Sync is only used on willTerminate. Tab upsert via persistTabs is unchanged so private tabs still persist without blocking foreground

ios/brave-ios/Sources/Brave/States/BrowserState.swift
```diff
@@ -90,6 +112,72 @@ public class BrowserState {
     }
   }
 
+  public static func getSessionState(from session: UISceneSession) -> SessionState {
+    guard let userInfo = session.userInfo else {
+      return SessionState(windowId: nil, isPrivate: false, openURL: nil)
+    }
+
+    return SessionState(
```
netzenbot, line 120: nit: `persistBrowsingMode` is a public API but has no doc comment. Given `persistRememberedBrowsingMode` is the guarded wrapper that most callers should use, a brief note on when to call this lower-level method directly (and its threading expectations) would help. See [DOC-004](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-004).

netzenbot (approved):

PR #36468, opened 2026-05-15
Register vertical tab region view as draggable area

sangwoo108:
Resolves https://github.com/brave/brave-browser/issues/30011

PR #36567, opened 2026-05-20
Revert "Register vertical tab region view as draggable area (#36468)"

sangwoo108:
This reverts commit 23084d45c9acb14f91c4187d28c72f1f6d5bc4f8.

As per comment in #36468 , revert this feature for now.


simonhong (approved):
++ thanks.

PR #36568, opened 2026-05-20
Fix widget promo panel layout (uplift to 1.90.x)

simoarpe:
## Resolves https://github.com/brave/brave-browser/issues/55646

Fixes widget promo panel layout margins and calculates correct position when shown for the first time.

**This issue is already fixed in Beta and Nightly.**


kjozwiak (approved):
Uplift into `1.90.x` approved 👍The fix will need to be checked directly against `1.90.x` as both `Nightly` & `BETA` have already been fixed/don't have the above issue.

PR #36575, opened 2026-05-20
Updates Android best practices to prefer using Material design components versus custom

SergeyZhukovsky:
Updates Android best practices to prefer using Material design components versus custom

### DIFF
--- a/docs/best-practices/android.md
+++ b/docs/best-practices/android.md
@@ -936,3 +936,15 @@ chromium_code = false
 # Do not let chromium_code = false disable Error Prone.
 enable_errorprone = true
 ```
+
+---
+
+<a id="AND-055"></a>
+
+## ✅ Prefer Material Components Over Custom Reimplementations
+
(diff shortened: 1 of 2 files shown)


docs/best-practices/android.md
```diff
@@ -936,3 +936,15 @@ chromium_code = false
...
 enable_errorprone = true
 ```
+
+---
+
+<a id="AND-055"></a>
+
+## ✅ Prefer Material Components Over Custom Reimplementations
+
+**When a Material component exists for the UI element, use it instead of building a custom version.** Material components carry built-in motion, accessibility, and theming that hand-rolled `View` subclasses almost always miss. Common cases: `MaterialSwitch` over custom toggles, `MaterialButton` over styled `Button`, `BottomSheetDialogFragment` over hand-rolled sheets, `MaterialAlertDialogBuilder` over custom modal views, `Snackbar`, `Chip`, `TextInputLayout`, `MaterialCardView`.
```
aguscruiz, line 946: @SergeyZhukovsky Would it be worth listing out ALL of the material component names instead of just some examples? Or do you think it's redundant and the bot will be able to figure it out?

SergeyZhukovsky: Redundant I think — the verification step pins it to the `com.google.android.material.*` package, which is the authoritative list. Hand-enumerating every component would duplicate that and drift over time as Material releases new ones. The examples in the rule are just the high-traffic cases; the package check catches everything else.

aguscruiz (approved):
Hope this works :D

bbondy (approved):

fallaciousreasoning (commented):
lgtm

PR #36576, opened 2026-05-20
[plaster] Experimental integration with prototype engines

cdesouza-chromium:
This is an experimental implementation for the use of prototype engines.

### DIFF
--- a/tools/cr/comby/build_comby.py
+++ b/tools/cr/comby/build_comby.py
@@ -1,0 +1,411 @@
+#!/usr/bin/env vpython3
+# Copyright (c) 2026 The Brave Authors. All rights reserved.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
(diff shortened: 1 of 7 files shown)


tools/cr/comby/build_comby.py
```diff
@@ -0,0 +1,411 @@
...
+_CR_DIR = str(Path(__file__).resolve().parent.parent)
+if _CR_DIR not in sys.path:
+    sys.path.insert(0, _CR_DIR)
+
+from repository import brave  # noqa: E402
+from terminal import terminal  # noqa: E402
+
+# Pinned opam release. The download URL is constructed from this and the
+# extracted directory name is `opam-<version>` (GitHub's archive tarballs
+# strip the leading `v` from the tag name).
```
netzenbot, line 58: nit: The path `third_party/comby/README.md` referenced here (and again in the Windows error message in `main()`) doesn't exist — the README this PR adds lives at `tools/cr/comby/README.md`. Update both references so they point to the actual file. ([Delete Dead Documentation](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-002))

PR #30429, opened 2025-08-02
[AIChat]: Enable pdf file upload

darkdh:
Resolves https://github.com/brave/brave-browser/issues/48151

1. UploadFileHelper now accepts .pdf extension and will do basic PDF magic
   bytes check.
2. Added new conversation API event to upload pdf content through pdf MIME
   data url. (Corresponding server PR https://github.com/brave/aichat/pull/706)
3. Enforce new restrictions for upload items: 20 images per conversation
   and 5 documents per conversation and each document cannot exceed
   4.5 MB.

## Test plan
1. Use clip icon->"Upload files..."
2. Pick a pdf file
Screenshot 2025-08-04 at 12 17 05

3. Ask what is in the pdf file and should be able to get answer
Screenshot 2025-08-04 at 12 17 43

PR #36583, opened 2026-05-20
[AI Chat] Fix attachment progress not shown

### LINKED ISSUE #55694: [AI Chat] Attaching files does not show progress whilst they are being processed
No spinner is shown in the input

I believe this is caused by a combination of:
- https://github.com/brave/brave-core/pull/30429
- https://github.com/brave/brave-core/pull/35816

## Steps to Reproduce
- Open Leo
- Drag and drop a large file to Leo, e.g. a 7mb movie, pdf, or image

### Expected
Progress is shown

Image

### Actual
Nothing happens for many seconds until the file is processed

Image

petemill:
- Fix https://github.com/brave/brave-browser/issues/55694

When attaching any files, no progress is being shown. This is a regression caused by the following, either one of which alone stops progress being shown:
- https://github.com/brave/brave-core/pull/35816
- https://github.com/brave/brave-core/pull/30429

Also added tests to catch the regression.

image

image


components/ai_chat/resources/page/components/input_box/index.test.tsx
```diff
@@ -321,6 +321,47 @@ describe('input box', () => {
...
+    ).toBeInTheDocument()
+  })
+
+  it('does not show the upload spinner when isUploadingFiles is false and there are no pending files', async () => {
+    const { container } = await renderInputBox(
+      <MockContext>
+        <InputBox
+          context={{
+            ...testContext,
+            pendingMessageFiles: [],
```
darkdh: shouldn't we also show the spinner when there are pending files and isUploadingFiles is true?
ex. an user can do two steps upload, first attempt with one image and second attempt with one pdf
and the spinner should be displayed for both tries

components/ai_chat/resources/page/state/conversation_context.tsx
```diff
@@ -480,18 +484,18 @@ export function useProvideConversationContext(props: ConversationContextProps) {
...
         // TODO(https://github.com/brave/brave-browser/issues/54918): We should just expose a `ProcessFile` function on the mojo API and do the file type detection on the C++ side.
         if (type === Mojom.UploadedFileType.kText) {
-          promise = aiChat.api.uiHandler.processTextFile(data, file.name)
+          promise = processTextFileMutation.mutateAsync([data, file.name])
         } else if (type === Mojom.UploadedFileType.kImage) {
-          promise = aiChat.api.uiHandler.processImageFile(data, file.name)
+          promise = processImageFileMutation.mutateAsync([data, file.name])
         } else {
-          promise = aiChat.api.uiHandler.processPdfFile(data, file.name)
+          promise = processPdfFileMutation.mutateAsync([data, file.name])
```
darkdh, line 495: neat usage on `mutateAsync`

components/ai_chat/resources/page/components/input_box/index.test.tsx
petemill: yes, and that scenario does work with this PR. I suppose the test is poorly titled and that case isn't explicitly tested.

components/resources/ai_chat_ui_strings.grdp
```diff
@@ -757,10 +757,10 @@
     Attachment
   </message>
   <message name="IDS_AI_CHAT_UPLOAD_FILE_BUTTON_LABEL" desc="The button label for upload file" formatter_data="webui=AiChat">
-    Upload file...
+    Attach file...
```
netzenbot, line 760: nit: The `desc` attributes for `IDS_AI_CHAT_UPLOAD_FILE_BUTTON_LABEL` ("The button label for upload file") and `IDS_AI_CHAT_UPLOADING_FILE_LABEL` ("The label for uploading a file") still say "upload" even though the strings now say "attach". Per [L10N-003](https://github.com/brave/brave-core/tree/master/docs/best-practices/localization.md#L10N-003), descriptions must be accurate so translators aren't misled. Please update the `desc` values to match the new terminology.

petemill: fixed

components/ai_chat/resources/page/components/input_box/index.test.tsx
petemill: added another test

bsclifton (approved):

mkarolin (approved):
`strings`++

components/ai_chat/resources/page/components/input_box/index.test.tsx
darkdh: can we do these test in the same way like you do 
```
  it.each([
    ['text', 'processTextFile' as const, 'text/plain', 'foo.txt'],
    ['image', 'processImageFile' as const, 'image/png', 'foo.png'],
    ['pdf', 'processPdfFile' as const, 'application/pdf', 'foo.pdf'],
  ])(
  ...
  ```
  and you missed one case "not show spinner when isUploadingFiles is false alongside pending files"

petemill: done!

darkdh (approved):

PR #36495, opened 2026-05-15
[Origin] Defer policy init signal until BraveProfilePolicyProvider refreshes

SergeyZhukovsky:
`BraveProvider`'s first `RefreshPolicies` is triggered by `AdBlockOnlyModePolicyManager` (initialised at process start), well before `BraveOriginPolicyManager` (lazily initialised by `BraveOriginServiceFactory`) is ready. The resulting bundle lacks `BraveRewardsDisabled`, but the default `IsInitializationComplete` lets `PolicyService` report ready anyway. `AdsServiceImpl::MaybeStartBatAdsService` then evaluates a stale managed pref store and bat-ads starts despite the policy.

Override `IsInitializationComplete` to additionally gate on `BraveOriginPolicyManager::IsInitialized()`. Upstream unit tests
(`ProfilePolicyConnectorTest`) bypass our factory bootstrap, so add a `SetExpectedToBeInitialized` flag set from `BraveOriginServiceFactory`'s ctor; the override short-circuits when unset (desktop/Android only -- iOS doesn't run those tests).

Resolves: https://github.com/brave/brave-browser/issues/54962

PR #36585, opened 2026-05-20
[iOS] [Origin] Don't defer profile init for policy initialization

kylehickinson:
This causes a deadlock in which the Profile never completes its initialization so the app never shows the browser window

refs https://github.com/brave/brave-core/pull/36495


SergeyZhukovsky (approved):
++

kylehickinson:
Superseded by https://github.com/brave/brave-core/pull/36586

PR #36586, opened 2026-05-20
[iOS] [Origin] Initialize policy manager before profile creation

kylehickinson:
This fixes a deadlock in which the Profile never completes its initialization so the app never shows the browser window


SergeyZhukovsky (approved):
++

components/brave_policy/brave_profile_policy_provider.cc
```diff
@@ -69,7 +68,6 @@ bool BraveProfilePolicyProvider::IsInitializationComplete(
   if (!manager->IsExpectedToBeInitialized()) {
```
cdesouza-chromium, line 62: Would you mind updating the comment above ^

cdesouza-chromium (approved):
LGTM % comment

components/brave_policy/brave_profile_policy_provider.cc
kylehickinson: Fixed!

netzenbot (approved):

PR #36595, opened 2026-05-21
[ads] Add BraveAdsEnabled policy for Desktop and Android

aseren:
Adds the `BraveAdsEnabled` policy for Desktop and Android, allowing administrators to enable or disable Brave Ads. When disabled, all Brave ad types are turned off and cannot be overridden by users.

Part of https://github.com/brave/brave-browser/issues/54917.

### DIFF
--- a/browser/brave_ads/ads_policy_browsertest.cc
+++ b/browser/brave_ads/ads_policy_browsertest.cc
@@ -1,0 +1,76 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "base/check_deref.h"
+#include "base/values.h"
+#include "brave/browser/brave_ads/ads_service_factory.h"
+#include "brave/browser/brave_ads/ads_service_waiter.h"
+#include "brave/components/brave_ads/core/browser/service/ads_service.h"
+#include "chrome/browser/ui/browser.h"
+#include "chrome/test/base/platform_browser_test.h"
+#include "components/policy/core/browser/browser_policy_connector.h"
+#include "components/policy/core/common/mock_configuration_policy_provider.h"
+#include "components/policy/core/common/policy_map.h"
+#include "components/policy/core/common/policy_types.h"
+#include "components/policy/policy_constants.h"
+#include "content/public/test/browser_test.h"
+#include "testing/gmock/include/gmock/gmock.h"
+#include "third_party/abseil-cpp/absl/strings/str_format.h"
+
+namespace brave_ads {
+
+class BraveAdsPolicyBrowserTest : public PlatformBrowserTest,
+                                  public ::testing::WithParamInterface<bool> {
+ public:
+  BraveAdsPolicyBrowserTest() = default;
+  ~BraveAdsPolicyBrowserTest() override = default;
+
+  void SetUpInProcessBrowserTestFixture() override {
+    EXPECT_CALL(configuration_policy_provider_,
+                IsInitializationComplete(::testing::_))
+        .WillRepeatedly(::testing::Return(true));
+    policy::BrowserPolicyConnector::SetPolicyProviderForTesting(
+        &configuration_policy_provider_);
+    policy::PolicyMap policies;
+    policies.Set(policy::key::kBraveAdsEnabled, policy::POLICY_LEVEL_MANDATORY,
+                 policy::POLICY_SCOPE_USER, policy::POLICY_SOURCE_PLATFORM,
+                 base::Value(IsAdsEnabledByPolicy()),
+                 /*external_data_fetcher=*/nullptr);
+    configuration_policy_provider_.UpdateChromePolicy(policies);
+  }
+
+  bool IsAdsEnabledByPolicy() { return GetParam(); }
+
+  AdsService& GetAdsService() {
+    return CHECK_DEREF(AdsServiceFactory::GetForProfile(browser()->profile()));
+  }
+
+ private:
+  policy::MockConfigurationPolicyProvider configuration_policy_provider_;
+};
+
+IN_PROC_BROWSER_TEST_P(BraveAdsPolicyBrowserTest,
+                       ServiceInitializationRespectsBraveAdsEnabledPolicy) {
+  test::AdsServiceWaiter waiter(GetAdsService());
+  if (IsAdsEnabledByPolicy()) {
+    waiter.WaitForOnDidInitializeAdsService();
+    EXPECT_TRUE(GetAdsService().IsInitialized());
+  } else {
+    waiter.WaitForOnAdsServiceIneligibleToStart();
+    EXPECT_FALSE(GetAdsService().IsInitialized());
+  }
+}
+
+INSTANTIATE_TEST_SUITE_P(
+    BraveAdsPolicyBrowserTest,
+    BraveAdsPolicyBrowserTest,
+    ::testing::Bool(),
+    [](const testing::TestParamInfo<BraveAdsPolicyBrowserTest::ParamType>&
+           info) {
+      return absl::StrFormat("BraveAds%sByPolicy",
+                             info.param ? "Enabled" : "Disabled");
+    });
+
+}  // namespace brave_ads
--- a/components/policy/handlers/BUILD.gn
+++ b/components/policy/handlers/BUILD.gn
@@ -1,0 +1,41 @@
+# Copyright (c) 2026 The Brave Authors. All rights reserved.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import("//brave/components/brave_ads/buildflags/buildflags.gni")
+
+if (enable_brave_ads) {
+  static_library("handlers") {
+    public = [ "brave_ads_enabled_policy_handler.h" ]
+
+    sources = [ "brave_ads_enabled_policy_handler.cc" ]
+
+    public_deps = [ "//components/policy/core/browser" ]
+
+    deps = [
+      "//base",
+      "//brave/components/brave_ads/core/public:headers",
+      "//brave/components/ntp_background_images/common",
+      "//components/policy/core/common:common_constants",
+      "//components/prefs",
+    ]
+  }
+
+  source_set("unit_tests") {
+    testonly = true
+
+    sources = [ "brave_ads_enabled_policy_handler_unittest.cc" ]
+
+    deps = [
+      ":handlers",
+      "//base",
+      "//brave/components/brave_ads/core/public:headers",
+      "//brave/components/ntp_background_images/common",
+      "//components/policy/core/browser",
+      "//components/policy/core/common",
+      "//components/prefs",
+      "//testing/gtest",
+    ]
+  }
+}
--- a/components/policy/handlers/brave_ads_enabled_policy_handler.cc
+++ b/components/policy/handlers/brave_ads_enabled_policy_handler.cc
@@ -1,0 +1,42 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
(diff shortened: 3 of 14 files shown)


components/policy/handlers/BUILD.gn
```diff
@@ -0,0 +1,41 @@
...
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import("//brave/components/brave_ads/buildflags/buildflags.gni")
+
+if (enable_brave_ads) {
+  source_set("handlers") {
+    sources = [
+      "brave_ads_enabled_policy_handler.cc",
+      "brave_ads_enabled_policy_handler.h",
+    ]
```
netzenbot: `brave_ads_enabled_policy_handler.h` inherits from `TypeCheckingPolicyHandler` defined in `components/policy/core/browser/configuration_policy_handler.h`, so `//components/policy/core/browser` must be in `public_deps` rather than `deps`. Consumers of this target need the transitive include path. See [BS-024](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-024).

components/policy/handlers/BUILD.gn
```diff
@@ -0,0 +1,41 @@
+# Copyright (c) 2026 The Brave Authors. All rights reserved.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
```
netzenbot, line 3: nit: Since the entire file belongs to the `enable_brave_ads` feature, use `assert(enable_brave_ads)` at the top instead of wrapping both targets in `if (enable_brave_ads)`. See [BS-037](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-037) — `browser/brave_ads/BUILD.gn` uses this pattern.

components/policy/handlers/brave_ads_enabled_policy_handler.h
```diff
@@ -0,0 +1,34 @@
...
+
+class PrefValueMap;
+
+namespace policy {
+
+class PolicyMap;
+
+// Handles the `policy::key::kBraveAdsEnabled` policy. Setting the policy to
+// `false` forces all brave ads types prefs to false.
+class BraveAdsEnabledPolicyHandler final : public TypeCheckingPolicyHandler {
```
netzenbot, line 22: nit: Per [BS-049](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-049), public headers for components guarded by a build flag should include `static_assert(BUILDFLAG(ENABLE_BRAVE_ADS));` to catch accidental inclusion when the feature is disabled. Other brave_ads public headers (e.g. `core/public/prefs/pref_names.h`) follow this pattern.

browser/brave_ads/ads_policy_browsertest.cc
```diff
@@ -0,0 +1,76 @@
...
+    BraveAdsPolicyBrowserTest,
+    BraveAdsPolicyBrowserTest,
+    ::testing::Bool(),
+    [](const testing::TestParamInfo<BraveAdsPolicyBrowserTest::ParamType>&
+           info) {
+      return absl::StrFormat("BraveAds%sByPolicy",
+                             info.param ? "Enabled" : "Disabled");
+    });
+
+}  // namespace brave_ads
```
netzenbot, line 76: nit: The name `ServiceInitializationRespectsPolicyWhenRewardsDisabled` is misleading — the suffix `WhenRewardsDisabled` doesn't describe the scenario (this tests the `BraveAdsEnabled` admin policy, not a Rewards-disabled state). Since the test is parameterized over both `true` and `false`, consider a name like `ServiceInitializationRespectsBraveAdsEnabledPolicy` that describes both instantiations. [TI-038](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-038)

components/policy/handlers/BUILD.gn
aseren: Fixed

components/policy/handlers/brave_ads_enabled_policy_handler.h
aseren: Fixed

browser/brave_ads/ads_policy_browsertest.cc
aseren: Fixed

components/policy/handlers/BUILD.gn
aseren: @netzenbot, adding `assert(enable_brave_ads)` here is wrong because in future there can be other handlers added which doesn't require this flag to be enabled.

browser/brave_ads/ads_policy_browsertest.cc
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+ private:
+  policy::MockConfigurationPolicyProvider configuration_policy_provider_;
+};
+
+IN_PROC_BROWSER_TEST_P(BraveAdsPolicyBrowserTest,
+                       ServiceInitializationRespectsBraveAdsEnabledPolicy) {
+  test::AdsServiceWaiter waiter(GetAdsService());
+  if (IsAdsEnabledByPolicy()) {
```
bridiver, line 57: this feels like a change detector test, it doesn't seem like it's necessary?

components/policy/handlers/brave_ads_enabled_policy_handler.cc
```diff
@@ -0,0 +1,42 @@
...
+  const base::Value* value =
+      policies.GetValue(policy_name(), base::Value::Type::BOOLEAN);
+  if (!value || value->GetBool()) {
+    // When the policy is `true` or unset, users control each brave ads type is
+    // controlled independently.
+    return;
+  }
+
+  CHECK(prefs);
+  prefs->SetBoolean(brave_ads::prefs::kOptedInToNotificationAds, false);
```
bridiver, line 35: I think there was a misunderstanding, the idea was that the brave origin UI could combine these into one toggle. They should be set independently otherwise.

components/policy/handlers/BUILD.gn
```diff
@@ -0,0 +1,42 @@
...
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import("//brave/components/brave_ads/buildflags/buildflags.gni")
+
+if (enable_brave_ads) {
+  source_set("handlers") {
+    sources = [
+      "brave_ads_enabled_policy_handler.cc",
```
netzenbot: nit: Per [BS-055](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-055), consider adding `public = ["brave_ads_enabled_policy_handler.h"]` to make the public API of this target explicit. Even with a single header, listing it in `public` clarifies the intended interface and guards against accidental exposure if internal headers are added later.

components/policy/handlers/brave_ads_enabled_policy_handler_unittest.cc
```diff
@@ -0,0 +1,84 @@
...
+};
+
+TEST_F(BraveAdsEnabledPolicyHandlerTest,
+       PolicySettingFalseDisablesAllBraveAdsTypes) {
+  SetAndApplyPolicy(/*policy_enabled=*/false);
+
+  for (const std::string_view pref : kBraveAdsTypePrefs) {
+    bool pref_value;
+    ASSERT_TRUE(prefs().GetBoolean(pref, &pref_value));
+    EXPECT_FALSE(pref_value);
```
netzenbot: nit: The `ASSERT_TRUE` and `EXPECT_FALSE` assertions inside the `for` loop don't indicate which pref failed. Add `SCOPED_TRACE(pref)` before the assertions in the loop body so failures identify the failing pref. See [TI-014](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-014).

components/policy/handlers/brave_ads_enabled_policy_handler.cc
aseren: Yeah, we will have an Origin UI toggle which disables all 3 Brave Ads preferences via BraveAdsEnabled policy, something like:

```
    <div class="cr-row title">$i18n{braveOriginSectionAdsTitle}</div>

    <origin-toggle-button id="toggleAdsButton"
        policy-key="BraveAdsEnabled"
        label="$i18n{braveOriginAdsToggleTitle}"
        icon="product-bat-outline">
    </origin-toggle-button>

```

browser/brave_ads/ads_policy_browsertest.cc
aseren: I think it is not a change detector test, because it verifies that if `AdsServiceImpl` is initialized/not initialized, depending on `BraveAdsEnabled` policy

components/policy/handlers/BUILD.gn
aseren: Fixed

components/policy/handlers/brave_ads_enabled_policy_handler_unittest.cc
aseren: Fixed this concern by adding pref_name to ASSERTs and EXPECTs:

```
EXPECT_FALSE(pref_value) << pref;
```

components/policy/handlers/BUILD.gn
```diff
@@ -0,0 +1,41 @@
+# Copyright (c) 2026 The Brave Authors. All rights reserved.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import("//brave/components/brave_ads/buildflags/buildflags.gni")
+
+if (enable_brave_ads) {
```
netzenbot, line 8: nit: `source_set("handlers")` is consumed by `//brave/browser/policy/handlers` (an external target), so it's a public target. Per [BS-028](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-028), public component targets should use `static_library` instead of `source_set`. The existing `browser/policy/handlers/BUILD.gn` uses `static_library("handlers")` for the same reason.

aseren: Fixed

components/policy/handlers/brave_ads_enabled_policy_handler.cc
bridiver: I think you're still misunderstanding, there should be three separate policies and the UI toggle will change all three.

browser/brave_ads/ads_policy_browsertest.cc
bridiver: There isn't even any new functionality here because the ads service already handles the prefs. This test is basically just verifying that the pref gets set with the policy and that is a change detector test

components/policy/handlers/brave_ads_enabled_policy_handler.cc
tmancey: @aseren please see Slack thread with @bridiver and @bbondy, where this was discussed. Thanks

test/BUILD.gn
```diff
@@ -395,6 +395,7 @@ test("brave_unit_tests") {
       "//brave/components/brave_ads/core",
       "//brave/components/brave_ads/core/browser/service:test_support",
       "//brave/components/brave_ads/core/test:brave_ads_unit_tests",
+      "//brave/components/policy/handlers:unit_tests",
```
netzenbot, line 398: nit: `//brave/components/policy/handlers:unit_tests` has no `//chrome/...` deps, so it should be wired into `brave_components_unittests` via `components/BUILD.gn` rather than `brave_unit_tests` in `test/BUILD.gn`. See [BS-021](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-021).

components/policy/handlers/brave_ads_enabled_policy_handler.cc
```diff
@@ -0,0 +1,42 @@
...
+  prefs->SetBoolean(brave_ads::prefs::kOptedInToSearchResultAds, false);
+  prefs->SetBoolean(ntp_background_images::prefs::
+                        kNewTabPageShowSponsoredImagesBackgroundImage,
+                    false);
+}
+
+}  // namespace policy
```
netzenbot, line 42: nit: The comment reads 'users control each brave ads type is controlled independently' which is garbled — it merges two sentence fragments. Something like 'users retain full control over each ad type independently' would be clearer. See [CS-019](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-019).

aseren:
putting the PR to draft because we need to implement separate policy for each ad setting

aseren:
Closing the current PR as BraveAdsEnabled policy is now split to 3 separate policies for each ad setting:
* https://github.com/brave/brave-core/pull/37955
* https://github.com/brave/brave-core/pull/37956
* https://github.com/brave/brave-core/pull/37957

PR #36596, opened 2026-05-21
[CompactMode] Refactor constants and helpers for maintainability

zenparsing:
Subtask of https://github.com/brave/brave-browser/issues/55506


simonhong (approved):
++ much clear. thanks!

browser/ui/views/tabs/brave_tab_container.cc
```diff
@@ -496,11 +495,8 @@ void BraveTabContainer::PaintBoundingBoxForSplitTab(
         tabs::kHorizontalTabInset));
   }
 
-  const float kRadius =
-      (!is_vertical_tab && tabs::ShouldUseCompactHorizontalTabsForNonTouchUI())
-          ? tabs::compact_horizontal_tabs_layout::
-                kHorizontalSplitViewTileCornerRadiusDip
-          : 12.f;  // default matches --leo-radius-l
+  const float corner_radius =
+      !is_vertical_tab && tabs::UseCompactHorizontalTabs() ? 9.0f : 12.f;
```
cdesouza-chromium, line 499: nit: would help with readability if these values were `constexpr` named constants?

chromium_src/chrome/browser/ui/layout_constants.cc
```diff
@@ -9,27 +9,21 @@
...
+  const bool compact = UseCompactHorizontalTabs();
   switch (inset) {
     case LOCATION_BAR_PAGE_INFO_ICON_PADDING:
-      return gfx::Insets::VH(
-          compact ? tabs::compact_horizontal_tabs_layout::
-                        kPageInfoIconPaddingVertical
-                  : tabs::compact_horizontal_tabs_layout::
-                        kPageInfoIconPaddingVerticalDefault,
-          tabs::compact_horizontal_tabs_layout::kPageInfoIconPaddingHorizontal);
+      return gfx::Insets::VH(compact ? 4 : 6, 6);
```
cdesouza-chromium, line 26: nit: same, it seems that it is the practice in this source to just have literals used everywhere, but do you think readability could improve if we had named constants to choose from for this and other cases in this file?

cdesouza-chromium (approved):
LGTM % some nits, I don't feel strong about them, feel free to resolve them as you see better.

chromium_src/chrome/browser/ui/layout_constants.cc
zenparsing: IMO, the literals in this file are already effectively "named" by either their position in the switch arm or by the compact/normal getter function that wraps them. Adding additional names adds a layer of mental indirection without much benefit.

PR #36597, opened 2026-05-21
[externals]: Share React for all ESModule JS bundles

### LINKED ISSUE #55709: [bundling]: We should share a React bundle across all WebUIs which require React
At the moment, we bundle React in each WebUI which needs it (which is almost all of them in Brave). By sharing an ESModule we can share a single React bundle across all WebUIs which don't specify `output_module = false`

Test Plan:
1. New tab page should load
2. chrome://settings/system/shortcuts should load

fallaciousreasoning:
Resolves https://github.com/brave/brave-browser/issues/55709

This PR will share a React bundle across all Brave WebUIs which use module output (i.e. they don't specify `output_module = false` in their `BUILD.gn`)

### DIFF
--- a/components/common/typescript.gni
+++ b/components/common/typescript.gni
@@ -62,6 +62,10 @@ brave_common_web_compile_inputs = [
 #       - intended pak name, e.g. resourcename_generated.pak
 #       - generated resource map, e.g. resourcename_generated_map.h
 #       - resource ID (e.g. `IDR_RESOURCENAME_MYFILE_JS`)
+#   no_externals
+#     Set this if you're compiling a bundle with dependencies
+#     which will otherwise be externalized (otherwise your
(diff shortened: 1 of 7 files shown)


zenparsing (approved):
❤️

components/common/typescript.gni
```diff
@@ -181,6 +181,10 @@ template("transpile_web_ui") {
       "--depfile_path=" + rebase_path(depfile),
     ]
 
+    if (defined(invoker.no_externals) && invoker.no_externals) {
+      args += [ "--no_externals" ]
+    }
+
```
petemill, line 191: we might want to document why an invoker would want to use "no_externals"

petemill (approved):
`external_js` is as good of a name as any I can think of!

netzenbot (approved):

PR #36599, opened 2026-05-21
Remove #define in c/b/ui/accelerator_table.cc

### LINKED ISSUE #55713: Remove #define macro in chromium_src/chrome/browser/ui/accelerator_table.cc

sangwoo108:
Replaces #define macro with plaster + chromium_src override. plaster will rename GetAcceleratorList() to GetAcceleratorList_ChromiumImpl() so that we can have our own implementation of GetAcceleratorList().

Resolves https://github.com/brave/brave-browser/issues/55713


patches/chrome-browser-ui-accelerator_table.cc.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/chrome/browser/ui/accelerator_table.cc b/chrome/browser/ui/accelerator_table.cc
```
netzenbot, line 1: This patch renames `GetAcceleratorList` in the upstream file, but the previous `#define GetAcceleratorList GetAcceleratorList_ChromiumImpl` in `chromium_src/` was already a working override. [CSRC-004](https://github.com/brave/brave-core/tree/master/docs/best-practices/chromium-src-overrides.md#CSRC-004) says patches are a last resort when a chromium_src override is not feasible — what makes the `#define` approach infeasible here?

chromium_src/chrome/browser/ui/accelerator_table.cc
```diff
@@ -41,10 +36,12 @@ constexpr AcceleratorMapping kBraveAcceleratorMap[] = {
 
 }  // namespace
```
netzenbot, line 37: nit: The new comment uses `(see \`kBraveAcceleratorMap\` above)` — `above` is a fragile positional reference. Since you already name the symbol explicitly, just drop `above`: `(see \`kBraveAcceleratorMap\`)`. [CS-024](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-024)

rewrite/chrome/browser/ui/accelerator_table.cc.toml
```diff
@@ -0,0 +1,12 @@
...
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+[[substitution]]
+description = '''
+Rename GetAcceleratorList() to GetAcceleratorList_ChromiumImpl() so that we can
+have our own implementation of GetAcceleratorList().
+'''
+pattern = 'std::vector<AcceleratorMapping> GetAcceleratorList() {'
```
netzenbot: nit: `pattern` should be used for simple symbol names. Here the pattern includes the full return type and opening brace to distinguish the definition from call sites — that's exactly the use case for `pattern_re`. Consider `pattern_re = 'std::vector<AcceleratorMapping> GetAcceleratorList\(\) \{'` instead. ([Use `pattern` for Simple Symbol Replacement, `pattern_re` for Context-Aware Matches](https://github.com/brave/brave-core/tree/master/docs/best-practices/plaster.md#PLSTR-002))

patches/chrome-browser-ui-accelerator_table.cc.patch
sangwoo108: This is false positive. We're trying to minimize #define macro

chromium_src/chrome/browser/ui/accelerator_table.cc
sangwoo108: done.

rewrite/chrome/browser/ui/accelerator_table.cc.toml
sangwoo108: done

netzenbot (approved):

rewrite/chrome/browser/ui/accelerator_table.cc.yaml
```diff
@@ -0,0 +1,12 @@
...
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+[[substitution]]
+description = '''
+Rename GetAcceleratorList() to GetAcceleratorList_ChromiumImpl() so that we can
+have our own implementation of GetAcceleratorList().
+'''
+re_pattern = '(std::vector<AcceleratorMapping>\ GetAcceleratorList)(\(\)\ {)'
```
cdesouza-chromium: I was looking on this file, and I noticed it has this:

```
  for (const auto& mapping : GetAcceleratorList()) {
    if (mapping.command_id == command_id) {
      *accelerator = ui::Accelerator(mapping.keycode, mapping.modifiers);
      return true;
    }
  }
```

Is this a bug we are fixing with this plaster?

sangwoo108: Oh, didn't notice it but I believe this is why we should remove the #define. I think it shouldn't be replaced with _ChromiumImpl()

cdesouza-chromium: Yeah, I get that, but I just want to be sure this changing in meaning is not gonna introduce anything new.

rewrite/chrome/browser/ui/accelerator_table.cc.yaml
```diff
@@ -0,0 +1,12 @@
+# Copyright (c) 2026 The Brave Authors. All rights reserved.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+[[substitution]]
```
cdesouza-chromium: We are writing plasters now in YAML. Please convert this plaster to the new format.

rewrite/chrome/browser/ui/accelerator_table.cc.yaml
sangwoo108: Yes, this won't do any harm.

rewrite/chrome/browser/ui/accelerator_table.cc.yaml
sangwoo108: Done.

rewrite/chrome/browser/ui/accelerator_table.cc.yaml
```diff
@@ -0,0 +1,11 @@
+# Copyright (c) 2026 The Brave Authors. All rights reserved.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+substitutions:
+  - description: >-
```
cdesouza-chromium: `- description: |-`

cdesouza-chromium (approved):
LGTM

PR #36600, opened 2026-05-21
Remove contained site data on site data removal.

### LINKED ISSUE #55714: [Containers] Support removing containers site data via standard dialog
Delete Browsing Data (Ctrl+Shift+Del) should also remove site data populated in containers using the same filters.

goodov:
Make sure contained website data is removed the same way as normal website data.

Resolves https://github.com/brave/brave-browser/issues/55714

### DIFF
--- a/browser/browsing_data/brave_browsing_data_remover_delegate.cc
+++ b/browser/browsing_data/brave_browsing_data_remover_delegate.cc
@@ -29,12 +29,12 @@
 #include "chrome/common/buildflags.h"
 #include "components/browsing_data/content/browsing_data_helper.h"
 #include "components/content_settings/core/browser/host_content_settings_map.h"
+#include "content/public/browser/browsing_data_filter_builder.h"
 #include "content/public/browser/browsing_data_remover.h"
 
 #if BUILDFLAG(ENABLE_CONTAINERS)
 #include "brave/browser/containers/used_container_storage_partitions.h"
 #include "brave/components/containers/core/common/features.h"
-#include "content/public/browser/browsing_data_filter_builder.h"
 #endif
 
 #if BUILDFLAG(ENABLE_AI_CHAT)
@@ -147,11 +147,15 @@ void BraveBrowsingDataRemoverDelegate::RemoveEmbedderData(
       std::move(callback));
 
 #if BUILDFLAG(ENABLE_CONTAINERS)
+  // Clear containers storage when no storage partition is specified in the
+  // filter.
   if (base::FeatureList::IsEnabled(containers::features::kContainers) &&
       !filter_builder->GetStoragePartitionConfig().has_value()) {
     const uint64_t container_remove_mask =
         content::BrowsingDataRemover::DATA_TYPE_ON_STORAGE_PARTITION &
         remove_mask;
+    // Only clear the storage when the remove mask covers data types that live
+    // on a storage partition.
     if (container_remove_mask) {
       for (const content::StoragePartitionConfig& config :
            containers::GetUsedContainerStoragePartitionConfigs(profile_)) {
--- a/browser/browsing_data/brave_browsing_data_remover_delegate.cc
+++ b/browser/browsing_data/brave_browsing_data_remover_delegate.cc
@@ -16,6 +16,7 @@
 #include "brave/browser/serp_metrics/serp_metrics_service_factory.h"
 #include "brave/components/ai_chat/core/common/buildflags/buildflags.h"
 #include "brave/components/brave_news/common/buildflags/buildflags.h"
+#include "brave/components/containers/buildflags/buildflags.h"
 #include "brave/components/content_settings/core/browser/brave_content_settings_pref_provider.h"
 #include "brave/components/content_settings/core/browser/brave_content_settings_utils.h"
 #include "brave/components/misc_metrics/page_metrics.h"
@@ -29,6 +30,12 @@
 #include "components/browsing_data/content/browsing_data_helper.h"
 #include "components/content_settings/core/browser/host_content_settings_map.h"
 #include "content/public/browser/browsing_data_remover.h"
+
+#if BUILDFLAG(ENABLE_CONTAINERS)
+#include "brave/browser/containers/used_container_storage_partitions.h"
+#include "brave/components/containers/core/common/features.h"
+#include "content/public/browser/browsing_data_filter_builder.h"
+#endif
 
 #if BUILDFLAG(ENABLE_AI_CHAT)
 #include "brave/browser/ai_chat/ai_chat_service_factory.h"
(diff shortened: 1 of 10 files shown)


browser/containers/containers_browsertest.cc
```diff
@@ -284,6 +293,120 @@ class ContainersBrowserTest : public InProcessBrowserTest {
...
+      return false;
+    }
+    return content::EvalJs(web_contents,
+                           GetServiceWorkerRegistrationCountJS()) == 0;
+  }
+
+  void ExpectPartitionSiteDataClearedEventually(
+      content::WebContents* web_contents,
+      const PartitionSiteData& data,
+      const std::string& scope) {
```
netzenbot, line 375: `IsPartitionSiteDataCleared` calls `content::EvalJs()` multiple times and is invoked directly as the `base::test::RunUntil()` predicate in `ExpectPartitionSiteDataClearedEventually`. This creates nested run loops and will cause DCHECK failures on macOS arm64 (`stack_.size() < nesting_level_`). Per [TA-007](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-async.md#TA-007), EvalJs/ExecJs must not be called inside RunUntil(). Consider synchronizing on the BrowsingDataRemover completion observer (already used in `RemoveSiteDataAndWait`) and then asserting JS state afterward with plain EvalJs calls.

browser/containers/used_container_storage_partitions.h
```diff
@@ -0,0 +1,23 @@
...
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#ifndef BRAVE_BROWSER_CONTAINERS_USED_CONTAINER_STORAGE_PARTITIONS_H_
+#define BRAVE_BROWSER_CONTAINERS_USED_CONTAINER_STORAGE_PARTITIONS_H_
+
+#include <vector>
+
+#include "content/public/browser/storage_partition_config.h"
+
+class Profile;
```
netzenbot, line 13: This header includes `content/public/browser/storage_partition_config.h`, which means `//content/public/browser` must be in `public_deps` (not `deps`) of the `:containers` target so consumers get the transitive include path. See [BS-024](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-024).

components/containers/core/browser/containers_service.cc
```diff
@@ -87,6 +87,14 @@ std::vector<mojom::ContainerPtr> ContainersService::GetContainers() const {
   return GetContainersFromPrefs(*prefs_);
 }
 
+std::vector<std::string> ContainersService::GetUsedContainerIds() const {
+  std::vector<std::string> ids;
+  for (const auto& container : GetLocallyUsedContainersFromPrefs(*prefs_)) {
```
netzenbot: nit: This manual loop pattern can be replaced with `base::ToVector`. See [CSA-020](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-020).
```cpp
return base::ToVector(GetLocallyUsedContainersFromPrefs(*prefs_),
                      [](const auto& c) { return c->id; });
```

browser/containers/used_container_storage_partitions.cc
```diff
@@ -0,0 +1,37 @@
...
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#include "brave/browser/containers/used_container_storage_partitions.h"
+
+#include "brave/browser/containers/containers_service_factory.h"
+#include "brave/components/containers/content/browser/storage_partition_utils.h"
+#include "brave/components/containers/core/browser/containers_service.h"
+#include "chrome/browser/profiles/profile.h"
```
netzenbot, line 16: nit: `std::string` is used directly in the range-for loop (`const std::string& id`) but `<string>` isn't included — it's only available transitively through `containers_service.h`. Per IWYU, add `#include <string>` directly. https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-001

browser/containers/used_container_storage_partitions.h
```diff
@@ -0,0 +1,23 @@
...
+
+namespace containers {
+
+// Returns StoragePartitionConfig for each locally used container.
+std::vector<content::StoragePartitionConfig>
+GetUsedContainerStoragePartitionConfigs(Profile* profile);
```
netzenbot, line 20: nit: The comment doesn't capture the contract fully — notably that this returns an empty vector when `profile` is null or `ContainersService` is unavailable. Those are non-obvious behaviors callers depend on. ([Method Documentation Should Describe the Contract](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-004))

netzenbot (approved):

browser/browsing_data/brave_browsing_data_remover_delegate.cc
```diff
@@ -30,6 +31,12 @@
 #include "components/content_settings/core/browser/host_content_settings_map.h"
 #include "content/public/browser/browsing_data_remover.h"
 
+#if BUILDFLAG(ENABLE_CONTAINERS)
+#include "brave/browser/containers/used_container_storage_partitions.h"
+#include "brave/components/containers/core/common/features.h"
+#include "content/public/browser/browsing_data_filter_builder.h"
```
simonhong: this includes should be outside of container buildflag?

browser/browsing_data/brave_browsing_data_remover_delegate.cc
```diff
@@ -139,6 +146,26 @@ void BraveBrowsingDataRemoverDelegate::RemoveEmbedderData(
       delete_begin, delete_end, remove_mask, filter_builder, origin_type_mask,
       std::move(callback));
 
+#if BUILDFLAG(ENABLE_CONTAINERS)
+  if (base::FeatureList::IsEnabled(containers::features::kContainers) &&
+      !filter_builder->GetStoragePartitionConfig().has_value()) {
+    const uint64_t container_remove_mask =
+        content::BrowsingDataRemover::DATA_TYPE_ON_STORAGE_PARTITION &
+        remove_mask;
```
simonhong, line 156: How about having comment to this condition something like `clear container storage when no storage partition in filter` with meaning of no storage in filter.

browser/browsing_data/brave_browsing_data_remover_delegate.cc
```diff
@@ -139,6 +146,26 @@ void BraveBrowsingDataRemoverDelegate::RemoveEmbedderData(
...
+    if (container_remove_mask) {
+      for (const content::StoragePartitionConfig& config :
+           containers::GetUsedContainerStoragePartitionConfigs(profile_)) {
+        std::unique_ptr<content::BrowsingDataFilterBuilder> partition_filter =
+            filter_builder->Copy();
```
simonhong, line 163: q) upstream created new filter builder for web app. We should use `copy`?
```
        auto iwa_filter_builder = content::BrowsingDataFilterBuilder::Create(
            content::BrowsingDataFilterBuilder::Mode::kPreserve);
```

simonhong (approved):
++

browser/browsing_data/brave_browsing_data_remover_delegate.cc
goodov: the web app filter is different - it works in the inverse mode where it preserves everything EXCEPT the web app to clean.

here we need to cleanup stuff similarly to a default partition, so the filter should be the same, but include a partition.

browser/browsing_data/brave_browsing_data_remover_delegate.cc
goodov: yep, added few comments

PR #36604, opened 2026-05-21
[toolchain] Allow `stderr` through in the rust builder

cdesouza-chromium:
This PR changes the function we use to run commands in the rust builder
to always allow `stderr` through. We are also adding a log whenever the
script is launched to indicate the value of `GIT_CACHE_PATH`.


AlexeyBarabash (approved):
LGTM!

PR #36605, opened 2026-05-21
[farbling] Farble getSupportedExtensions and getExtension

### LINKED ISSUE #55858: Add farbling logic for getSupportedExtensions
The core logic for this feature is derived from the algorithm presented in the parent issue https://github.com/brave/brave-browser/issues/15904. 

We now have a [list](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/third_party/blink/renderer/bindings/core/webgl/webgl_farbled_extension_handler.cc?L26) of fake extensions. When the website makes the call to [getSupportedExtensions()](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/chromium_src/third_party/blink/renderer/modules/webgl/webgl_rendering_context_base.cc?L164) we inject one of the fake values with the result and is stable for the lifetime of that site session by making use of the same underlying farbling seed. 

This simple defense helps to thwart automated scripts that rely on a hash-based approach for fingerprinting across two different sites. With this farbling approach, the hashes are very strongly likely to be different on two different sites. Mathematically, the probability of the hash to be same for two different sites (independent events) would be `1/|fake_list_size| x 1/fake_list_size|`. With the current fake list size of 21, the probability of having the same hash is 1/441.

r0hit0303:
This change is the final change from the series of changes to add support for farbling getSupportedExtensions calls. 

**Past changes:** 
1) https://github.com/brave/brave-core/pull/36762
2) https://github.com/brave/brave-core/pull/36837

**Current change**
Adds the final integration of the `WebGLFarbledExtensionHandler` with `WebGLRenderingContextBase`. The latter is the core site which handles the actual `getSupportedExtensions` and `getExtensions` call. 

### Demo (w/o containers)

https://github.com/user-attachments/assets/e0d5d57c-b28e-43fe-8665-21fd8f51b88b

### Demo (with containers)

https://github.com/user-attachments/assets/d5c6d338-e5db-4bfe-95d0-cbf721f08d83

### Motionmark score (with this PR - Debug mode)

motionmark-score-on-pr-36605

### Motionmark score (w/o this PR - Debug mode)
motionmark-score-w:o-pr-36605

Resolves https://github.com/brave/brave-browser/issues/55858


r0hit0303:
@goodov Hey Aleksei, could I please get some early high level feedback on the approach? I'm not looking for a full review at the moment. Thanks

If everything looks good to you then I can start breaking this change into smaller chunks and send them for review along with tests. Also, do you know if we need to adapt this in the context of [containers](https://github.com/brave/brave-core/pull/36474/changes)? If so, would you have any suggestions on how we might be able to add additional entropy for them in the blink side?

goodov:
looks fine. Make sure to avoid any additional computation each time a farbled method is called in the same context.

No need to handle this specifically for containers. The token you use to farble API is already container-aware since this was landed https://github.com/brave/brave-browser/issues/55601.

chromium_src/third_party/blink/renderer/modules/webgl/webgl_rendering_context_base.cc
```diff
@@ -104,8 +105,6 @@ blink::ScriptValue GetWebGLDebugInfoValue(
     precision = 0;                                          \
   }
 
-// TODO(https://github.com/brave/brave-browser/issues/55445): Add fingerprinting
-// protections for GL_RENDERER and GL_VENDOR
```
r0hit0303, line 108: No farbling support seems to be needed for them. Please see https://github.com/brave/brave-browser/issues/55445.

chromium_src/third_party/blink/renderer/modules/webgl/webgl_rendering_context_base.cc
```diff
@@ -126,35 +125,37 @@ blink::ScriptValue GetWebGLDebugInfoValue(
...
-    if (name != WebGLDebugRendererInfo::ExtensionName()) {
-      return ScriptObject::CreateNull(v8::Isolate::GetCurrent());
-    }
+  std::optional<Vector<String>> real_extensions = getSupportedExtensions();
+  if (real_extensions == std::nullopt) {
+    return ScriptObject::CreateNull(v8::Isolate::GetCurrent());
   }
-  return getExtension_ChromiumImpl(script_state, name);
+
+  // Upstream takes care of returning nullable ScriptObject for invalid names.
```
r0hit0303, line 191: I think we would return a nullptr [here](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/webgl/webgl_rendering_context_base.cc;l=3850.). I'll add a test just to be sure.

r0hit0303: Added `GetExtensionWithInvalidName` test and indeed we have null.

browser/farbling/brave_webgl_farbling_browsertest.cc
```diff
@@ -163,57 +177,6 @@ IN_PROC_BROWSER_TEST_F(BraveWebGLFarblingBrowserTest,
   }
```
netzenbot, line 203: The new test code calls `base::SplitString`, `base::TRIM_WHITESPACE`, and `base::SPLIT_WANT_ALL`, but `"base/strings/string_split.h"` is not included. Please add it to the include list. ([CS-001](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-001))

chromium_src/third_party/blink/renderer/modules/webgl/webgl_rendering_context_base.cc
```diff
@@ -126,35 +125,37 @@ blink::ScriptValue GetWebGLDebugInfoValue(
 
 namespace blink {
 
-// If fingerprinting is disallowed, claim that the only supported
-// extension is WebGLDebugRendererInfo.
 std::optional<Vector<String>>
 WebGLRenderingContextBase::getSupportedExtensions() {
   std::optional<Vector<String>> real_extensions =
       getSupportedExtensions_ChromiumImpl();
   if (real_extensions == std::nullopt) {
```
netzenbot: nit: The previous code had per-function comments explaining why each override was needed (e.g. "If fingerprinting is disallowed, claim that the only supported extension is WebGLDebugRendererInfo."). The new implementation removes them without replacement. Per [CSRC-017](https://github.com/brave/brave-core/tree/master/docs/best-practices/chromium-src-overrides.md#CSRC-017), each overridden function should have a comment explaining why the override is needed.

chromium_src/third_party/blink/renderer/modules/webgl/webgl_rendering_context_base.cc
```diff
@@ -126,35 +125,37 @@ blink::ScriptValue GetWebGLDebugInfoValue(
...
+
+  // Upstream takes care of returning nullable ScriptObject for invalid names.
+  const ScriptObject real_extension =
+      getExtension_ChromiumImpl(script_state, name);
+
+  blink::WebGLFarbledExtensionHandler* handler =
+      brave::BraveSessionCache::From(*Host()->GetTopExecutionContext())
+          .GetWebGLFarbledExtensionHandler(real_extensions.value());
+  // Handler would apply farbling on valid extension name. If not valid it would
+  // return real_extension.
```
netzenbot, line 195: nit: The comment `// Handler would apply farbling on valid extension name. If not valid it would return real_extension.` is inaccurate — for names not in the supported list, `GetExtension` returns `CreateNull(...)`, not `real_extension`. Consider: `// Handler applies farbling for farbled extension names; returns real_extension for valid unfarbled names; returns null for unknown names.` ([CS-019](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-019))

browser/farbling/brave_webgl_farbling_browsertest.cc
```diff
@@ -317,3 +281,123 @@ IN_PROC_BROWSER_TEST_P(BraveWebGLGetParameterTest,
...
+  auto actual_balanced_extensions_list = base::SplitString(
+      actual_balanced_value, ",", base::TRIM_WHITESPACE, base::SPLIT_WANT_ALL);
+  std::sort(actual_balanced_extensions_list.begin(),
+            actual_balanced_extensions_list.end());
+  std::vector<std::string> diff;
+  std::ranges::set_difference(actual_balanced_extensions_list,
+                              actual_extensions_list, std::back_inserter(diff));
+
+  if (GetParam()) {
+    // This should contain one of the farbled values.
```
netzenbot: nit: `EXPECT_TRUE(IsFakeExtensionName(diff[0]))` won't show what `diff[0]` actually contains on failure. Consider `EXPECT_TRUE(IsFakeExtensionName(diff[0])) << diff[0];` — same applies to the equivalent line in `FarbleGetExtension`. ([TI-014](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-014))

browser/farbling/brave_webgl_farbling_browsertest.cc
```diff
@@ -317,3 +281,123 @@ IN_PROC_BROWSER_TEST_P(BraveWebGLGetParameterTest,
...
+  // WebGL getExtension returns real objects
+  AllowFingerprinting(domain);
+  ASSERT_TRUE(ui_test_utils::NavigateToURL(browser(), url));
+  std::string actual_value = EvalJs(contents(), kTitleScript).ExtractString();
+  EXPECT_NE(actual_value, kExpectedExtensionListMax);
+
+  auto actual_extensions_list = base::SplitString(
+      actual_value, ",", base::TRIM_WHITESPACE, base::SPLIT_WANT_ALL);
+  // We sort this to compute a set difference later.
+  std::sort(actual_extensions_list.begin(), actual_extensions_list.end());
```
netzenbot: nit: The balanced-mode assertion block (split, sort, set_difference, check IsFakeExtensionName or compare lists) is duplicated verbatim in `FarbleGetSupportedExtensions` and `FarbleGetExtension`. Consider extracting it into a helper like `VerifyBalancedFarblingExtensions(actual_off, actual_balanced, GetParam())` to reduce duplication. https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-024

netzenbot (approved):

chromium_src/third_party/blink/renderer/modules/webgl/webgl_rendering_context_base.cc
```diff
@@ -126,35 +125,78 @@ blink::ScriptValue GetWebGLDebugInfoValue(
...
+namespace {
+
+// An opaque method to get a valid WebGL extension handler. If the handler
+// does not exist it will create a new one.
+// |get_real_extensions| is a lambda which when invoked returns the true list of
+// supported extensions. The true list is needed to create the appropriate
+// handler.
+template <typename T>
+blink::WebGLFarbledExtensionHandler* GetValidHandler(ExecutionContext* context,
+                                                     T&& get_real_extensions) {
```
goodov: not sure if this helper simplifies anything... right now it doesn't feel so.

I think it should be fine to implement all of this inside `WebGLRenderingContextBase::getSupportedExtensions()`.

and also it makes sense to rename `GetWebGLFarbledExtensionHandler` to `GetWebGLFarbledExtensionHandler`.

r0hit0303: > not sure if this helper simplifies anything... right now it doesn't feel so.

I felt it was painting a cleaner picture that way but happy to also pipe things through getSupportedExtensions if that seems more practical. 

> and also it makes sense to rename `GetWebGLFarbledExtensionHandler` to `GetWebGLFarbledExtensionHandler`. 

I re-read this couple of times but I think there's a typo somewhere.. You want me to update the name of the method `GetWebGLFarbledExtensionHandler` to what exactly?

goodov: sorry, I meant `CreateOrGetWebGLFarbledExtensionHandler` :)

r0hit0303: I didn't find a good way to rely on `getSupportedExtensions` from `getExtension` without either trading off clarity / performance / code-duplication. Maybe there's a way but I don't see it now. I'm happy to follow-up on any ideas as a separate change if possible. 

To get a valid handler we need to either have created it in the past, or we create a fresh one which requires the current set of supported extensions. So in the getExtension call, since it's not guaranteed that we would have a valid handler, we would need to GetOrCreate that handler there as well like in getSupportedExtensions. Given this two calling sites relying on the same logic it made sense to put this logic into a common method `CreateOrGetValidWebGLExtensionHandler` inside an anonymous namespace. 

I've also made the contract much stricter to access the handler in the BraveSessionCache. The method `CreateWebGLFarbledExtensionHandler` would now guarantee to create a handler and would crash if the client tries to re-create a handler. To only access the handler the clients can use `get_webgl_farbled_extension_handler()` method.

goodov (approved):

chromium_src/third_party/blink/renderer/modules/webgl/webgl_rendering_context_base.cc
goodov: thanks for looking into this. I think I missed the fact that the getter is called from two places with different upstream implementations. With this in mind it does look reasonable the way you implemented it.

r0hit0303:
Thanks @goodov for the review

PR #36607, opened 2026-05-21
Add more Omaha 4 macOS update failure logging

mherrmann:
This PR has the same goal and test plan as #36480.

I see in remote logs that the `mkdir -p "${new_versioned_dir}"` call fails. This PR makes it so I can hopefully see exactly why it fails.


cdesouza-chromium (approved):
LGTM

PR #36609, opened 2026-05-21
Update polkadot keyring to accept custom u64 rng seed for testing.

cypt4:
It was using 0 before, so allow to setup custom one.

Resolves

### DIFF
--- a/components/brave_wallet/browser/internal/hd_key_sr25519.h
+++ b/components/brave_wallet/browser/internal/hd_key_sr25519.h
@@ -7,6 +7,7 @@
 #define BRAVE_COMPONENTS_BRAVE_WALLET_BROWSER_INTERNAL_HD_KEY_SR25519_H_
 
 #include <array>
+#include <cstdint>
 #include <optional>
 
 #include "base/containers/span.h"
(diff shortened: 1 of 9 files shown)


components/brave_wallet/browser/internal/hd_key_sr25519.h
```diff
@@ -80,10 +81,10 @@ class HDKeySr25519 {
   // of a general blob of bytes which can represent anything.
   HDKeySr25519 DeriveHard(uint32_t account_index) const;
 
-  // Update the keypair to use a mock RNG when deriving Shcnorr signatures. This
+  // Update the keypair to use a mock RNG when deriving Schnorr signatures. This
   // is intended only for testing purposes and would be disastrous for
```
netzenbot, line 85: nit: The comment doesn't describe what `seed` represents or what the default value `0` means (e.g., "seed value passed to ChaCha20Rng; defaults to 0 for a fully deterministic signature"). Per DOC-004, parameter contracts should be documented. ([Method Documentation Should Describe the Contract](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-004))

cmazakas (approved):
LGTM! Thank you for this.

components/brave_wallet/browser/internal/hd_key_sr25519.h
cypt4: In a follow up

PR #36614, opened 2026-05-21
Add best practice: convert rust::Str/rust::String at the C++ boundary

bbondy:
## Summary
- Adds a new best practice to `docs/best-practices/coding-standards-apis.md` (ID `CSA-070`) covering Rust↔C++ FFI string and container types.
- Captures the guidance from https://github.com/brave/brave-browser/issues/55722: convert `rust::Str`/`rust::String` to `std::string_view`/`base::span` (or deep-copy) at the boundary; avoid `CxxString`/`CxxVector` because Rust and C++ use incompatible allocators.

## Test plan
- [x] `npm run presubmit` passes
- [x] `python3 ./script/manage-bp-ids.py --validate` passes (596 headings, all unique IDs)
- [ ] Reviewer sanity-check that the helper signatures (`RustStringToStringView`, `RustStringToUTF16`) match what we want to standardize on

### DIFF
--- a/docs/best-practices/coding-standards-apis.md
+++ b/docs/best-practices/coding-standards-apis.md
@@ -1333,7 +1333,7 @@ class Reporter {
 };
 void Reporter::Refresh() {
   rust::String r = ffi::build_report();
-  last_report_.assign(r.data(), r.size());
+  last_report_.assign(r.data(), r.length());
 }
 ```
 
--- a/docs/best-practices/coding-standards-apis.md
+++ b/docs/best-practices/coding-standards-apis.md
@@ -1261,44 +1261,166 @@ base::ReadUnicodeCharacter(
 
 <a id="CSA-070"></a>
 
-## ✅ Convert Rust FFI Types at the C++ Boundary; Prefer `rust::Str`/`rust::Slice` Over `CxxString`/`CxxVector`
+## ✅ String Usage in Rust/C++ FFI
 
-**`rust::Str`/`rust::String` should always be converted to `std::string_view`/`base::span` (or deep-copied to `std::string`) when passed into C++. Prefer `rust::Str`/`rust::Slice` over `CxxString`/`CxxVector` for borrowed data.** Rust's `String`/`Vec` and C++'s `std::string`/`std::vector` use different allocators and memory layouts, so handing raw heap ownership across the boundary is unsafe — copy instead.
+**Pick FFI string types by data flow (borrow vs. own) and payload size (standard vs. large).** Rust's `String`/`Vec` and C++'s `std::string`/`std::vector` use different allocators and memory layouts, so the wrong choice causes hidden copies, lifetime bugs, or container overhead. Sections below cover each case with examples.
 
-```cpp
-// ❌ WRONG - storing rust::String in C++ data structures or passing it
-// around as if it were std::string. Bridge types should not outlive the
-// FFI call.
-class MyService {
-  rust::String value_;  // Don't hold rust types in C++ state.
-};
+### Rust → C++ read-only text — `&str` (maps to `rust::Str`)
 
-// ✅ CORRECT - convert at the boundary
-// For rust::Str, use the existing helper in //base:
-#include "base/strings/string_view_rust.h"
-std::string_view sv = base::RustStrToStringView(rust_str);
+Zero-copy. Rust retains ownership; C++ reads via pointer/length window. Use for logging, printing, parsing.
 
-// For rust::String, view its bytes as string_view, or deep-copy:
-std::string_view RustStringToStringView(const rust::String& s) {
-  return std::string_view(s.data(), s.length());
-}
-std::u16string RustStringToUTF16(const rust::String& s) {
-  return base::UTF8ToUTF16(RustStringToStringView(s));
+```rust
+// ❌ WRONG - forces C++ container overhead for a borrow
+unsafe extern "C++" {
+    fn log_event(name: &CxxString);
 }
 
-// When ownership must cross the boundary, deep-copy:
-std::string owned(rust_str.data(), rust_str.size());  // C++ side
-// rust_string.to_string() / rust_string.to_owned()   // Rust side
+// ✅ CORRECT - zero-copy borrow
+unsafe extern "C++" {
+    fn log_event(name: &str);
+}
 ```
 
-`CxxString`/`CxxVector` can't be passed by value across the FFI and `&mut`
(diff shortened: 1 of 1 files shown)


netzenbot (approved):

docs/best-practices/coding-standards-apis.md
```diff
@@ -1256,3 +1256,41 @@ base::ReadUnicodeCharacter(
...
     base::as_string_view(data), &i, &code_point);
 ```
+
+---
+
+<a id="CSA-070"></a>
+
+## ✅ Convert Rust FFI Types at the C++ Boundary; Avoid `CxxString`/`CxxVector`
+
+**`rust::Str`/`rust::String` should always be converted to `std::string_view`/`base::span` (or deep-copied to `std::string`) when passed into C++. Avoid `CxxString` and `CxxVector` for ownership transfer between Rust and C++.** Rust's `String`/`Vec` and C++'s `std::string`/`std::vector` use different allocators and memory layouts, so handing raw heap ownership across the boundary is unsafe — copy instead.
```
bridiver: missed the bit about `If you need to pass a large amount of data without copying, using cxx::UniquePtr is generally preferred`

bridiver: C++ allocates the via std::make_unique<T>() and drops ownership into a cxx::UniquePtr<T>. When Rust drops the UniquePtr, cxx safely invokes the C++ allocator to delete it.

bbondy: See the new commit pls

docs/best-practices/coding-standards-apis.md
```diff
@@ -1256,3 +1256,171 @@ base::ReadUnicodeCharacter(
...
+reporter.set_last_report(std::move(report));
+```
+
+### Rust → C++ ownership, MASSIVE payload — `Box<Vec<u8>>` (maps to `rust::Box<rust::Vec<uint8_t>>`)
+
+Zero-copy. Heap pointer moves to C++; C++ wraps bytes in `std::string_view` or `base::span`. Rust's allocator cleans up when `Box` drops. Use for file buffers, JSON blobs, streaming payloads.
+
+```rust
+// ❌ WRONG - copies container layout
+fn take_payload(buf: String);
```
cmazakas: Shouldn't this example function be `fn produce_payload() -> String;`? Otherwise, this looks like a transfer of ownership from C++ -> Rust.

bridiver: yeah that's definitely not right

docs/best-practices/coding-standards-apis.md
```diff
@@ -1256,3 +1256,171 @@ base::ReadUnicodeCharacter(
...
+
+```cpp
+// ❌ WRONG - bridge type stored past FFI call
+class Reporter {
+  rust::String last_report_;
+};
+
+// ✅ CORRECT - deep-copy at the boundary
+std::string report(rust_report.data(), rust_report.size());
+reporter.set_last_report(std::move(report));
```
cdesouza-chromium: I'm not sure what is the correct answer here, but I think there should be something better than a pointer/length pair. We have tried to move away from those a lot since spanification, but I'm not sure what the type of `rust_report` is supposed to be in this case.

For example, as Brian had mentioned in the thread, we have `base/strings/string_view_rust.h`, which has this:

```
inline std::string_view RustStrToStringView(rust::Str str) {
  return std::string_view(str.data(), str.size());
}
```

So we potentially should encourage people to use that function if all they have is pointer/length with `rust::Str`, because there is some effort in upstream to mark the pointer length constructors of `std::string` as buffer unsafe in C++.

docs/best-practices/coding-standards-apis.md
```diff
@@ -1256,3 +1256,171 @@ base::ReadUnicodeCharacter(
...
+
+// ✅ CORRECT - pointer move, zero-copy
+fn produce_payload() -> Box<Vec<u8>>;
+```
+
+```cpp
+void Consume(rust::Box<rust::Vec<uint8_t>> buf) {
+  base::span<const uint8_t> bytes(buf->data(), buf->size());
+  // ...consume bytes. Rust deallocator runs when buf drops.
+}
```
cdesouza-chromium, line 1356: This example will not work, because the pointer length constructor for `base::span` is marked as `UNSAFE_BUFFER_USAGE`. I 99% certain you can just pass `buf` as is to the constructor of `base::span`, because I did submit a PR for that to `cxxbridge`, and that was eventually picked up by someone else from MS, to make sure that the concepts expected were satisfied.

docs/best-practices/coding-standards-apis.md
bridiver: This example is out of context and makes no sense so I'm going to fix it, but this is for places where string_view is not appropriate like returning a new string from a function call

bridiver: this is what the example should look like

```rust
// ✅ Return owned text by value
fn build_report() -> String {
    format!("report: {}", compute())
}
```

```cpp
// ❌ WRONG - bridge type held past FFI call; ties C++ class to Rust
// allocator and bridge layout.
class Reporter {
  rust::String last_report_;
};
void Reporter::Refresh() {
  last_report_ = ffi::build_report();
}

// ✅ CORRECT - deep-copy into std::string at the boundary
class Reporter {
  std::string last_report_;
};
void Reporter::Refresh() {
  rust::String r = ffi::build_report();
  last_report_.assign(r.data(), r.size());
}
```

cmazakas (approved):
LGTM

docs/best-practices/coding-standards-apis.md
```diff
@@ -1256,3 +1256,180 @@ base::ReadUnicodeCharacter(
...
+
+<a id="CSA-070"></a>
+
+## ✅ String Usage in Rust/C++ FFI
+
+**Pick FFI string types by data flow (borrow vs. own) and payload size (standard vs. large).** Rust's `String`/`Vec` and C++'s `std::string`/`std::vector` use different allocators and memory layouts, so the wrong choice causes hidden copies, lifetime bugs, or container overhead. Sections below cover each case with examples.
+
+### Rust → C++ read-only text — `&str` (maps to `rust::Str`)
+
+Zero-copy. Rust retains ownership; C++ reads via pointer/length window. Use for logging, printing, parsing.
```
cdesouza-chromium: nit: "reads via pointer/length window" should be "reads as a view"?

cdesouza-chromium (approved):
LGTM

PR #36618, opened 2026-05-21
[VPN 2.0] Add BraveVpnService stub v2 implementation and finalize split

### LINKED ISSUE #54597: Implement a stub BraveVpnService v2
- ensure local testing works.

Acceptance criteria:

alekspop26:

This change adds the second "stub" BraveVpnService implementation based on the Architecture 2.0, under `brave_vpn::v2` namespace. The instantiation of v2 implementation is preferred when both are enabled, but v2 implementation compiled out by default (based on the GN flags).

The following combinations of BraveVpnService implementations are allowed:
1. BraveVpnServiceImpl v1 only.
2. BraveVpnServiceImpl v2 only.
3. Both BraveVpnServiceImpl v1 and v2 - the factory instantiates v2 (preparation for future runtime selection).
If neither BraveVpnServiceImpl v1 or v2 is enabled, then build fails.

In addition, GN flags controlling architecture-specific unit tests are added. These are supposed to be used in local testing only. In the CI runs, both architectures should be unit-tested, regardless of the implementation build flag settings.

In the change, browser and test GN targets were updated to include v1 and v2 dependencies conditionally, to eliminate any unwanted architecture-specific dependencies.

Resolves https://github.com/brave/brave-browser/issues/54597

Resolves

### DIFF
--- a/browser/brave_vpn/BUILD.gn
+++ b/browser/brave_vpn/BUILD.gn
@@ -7,17 +7,32 @@ import("//brave/components/brave_vpn/common/buildflags/buildflags.gni")
 
 assert(enable_brave_vpn)
 
-source_set("brave_vpn") {
+source_set("shared") {
   sources = [
     "vpn_utils.cc",
     "vpn_utils.h",
   ]
 
   deps = [
-    "//brave/components/brave_vpn/browser/connection:api",
     "//brave/components/brave_vpn/common",
     "//brave/components/brave_vpn/common/buildflags",
     "//chrome/browser/profiles:profile",
+    "//components/prefs",
+    "//components/user_prefs",
+  ]
+}
+
+source_set("brave_vpn") {
+  sources = [
+    "vpn_connection_manager_utils.cc",
+    "vpn_connection_manager_utils.h",
+  ]
+
+  public_deps = [ ":shared" ]
+
+  deps = [
+    "//brave/components/brave_vpn/browser/connection:api",
+    "//brave/components/brave_vpn/common",
     "//chrome/common:channel_info",
     "//components/prefs",
     "//components/user_prefs",
--- a/browser/brave_vpn/brave_vpn_service_factory.cc
+++ b/browser/brave_vpn/brave_vpn_service_factory.cc
@@ -7,48 +7,60 @@
 
 #include <utility>
 
-#include "base/feature_list.h"
 #include "base/no_destructor.h"
-#include "brave/browser/brave_browser_process.h"
 #include "brave/browser/brave_vpn/vpn_utils.h"
-#include "brave/browser/misc_metrics/process_misc_metrics.h"
-#include "brave/browser/misc_metrics/uptime_monitor_impl.h"
 #include "brave/browser/skus/skus_service_factory.h"
 #include "brave/components/brave_vpn/browser/brave_vpn_service.h"
-#include "brave/components/brave_vpn/browser/brave_vpn_service_impl.h"
 #include "brave/components/brave_vpn/common/brave_vpn_utils.h"
-#include "brave/components/skus/common/features.h"
+#include "brave/components/brave_vpn/common/buildflags/buildflags.h"
 #include "build/build_config.h"
-#include "chrome/browser/browser_process.h"
-#include "chrome/browser/profiles/incognito_helpers.h"
 #include "chrome/browser/profiles/profile.h"
 #include "components/keyed_service/content/browser_context_dependency_manager.h"
-#include "components/user_prefs/user_prefs.h"
 #include "content/public/browser/browser_context.h"
-#include "content/public/browser/storage_partition.h"
 
+#if BUILDFLAG(ENABLE_BRAVE_VPN_V1)
+#include "base/functional/bind.h"
+#include "base/memory/weak_ptr.h"
+#include "brave/browser/brave_browser_process.h"
+#include "brave/browser/misc_metrics/process_misc_metrics.h"
+#include "brave/browser/misc_metrics/uptime_monitor_impl.h"
+#include "brave/components/brave_vpn/browser/brave_vpn_service_impl.h"
+#include "chrome/browser/browser_process.h"
+#include "components/user_prefs/user_prefs.h"
+#include "content/public/browser/storage_partition.h"
 #if BUILDFLAG(IS_WIN)
 #include "brave/browser/brave_vpn/dns/brave_vpn_dns_observer_factory_win.h"
 #include "brave/browser/brave_vpn/dns/brave_vpn_dns_observer_service_win.h"
 #include "brave/browser/brave_vpn/win/brave_vpn_service_delegate_win.h"
 #include "brave/browser/brave_vpn/win/brave_vpn_wireguard_observer_factory_win.h"
 #include "brave/browser/brave_vpn/win/brave_vpn_wireguard_observer_service_win.h"
-#endif
+#endif  // BUILDFLAG(IS_WIN)
+#endif  // BUILDFLAG(ENABLE_BRAVE_VPN_V1)
+
+#if BUILDFLAG(ENABLE_BRAVE_VPN_V2)
+#include "brave/components/brave_vpn/v2/browser/brave_vpn_service_impl.h"
+#endif  // BUILDFLAG(ENABLE_BRAVE_VPN_V2)
 
 namespace brave_vpn {
 namespace {
 
-std::unique_ptr<KeyedService> BuildVpnService(
-    content::BrowserContext* context) {
-  if (!brave_vpn::IsAllowedForContext(context)) {
-    return nullptr;
-  }
+#if BUILDFLAG(ENABLE_BRAVE_VPN_V2)
 
+std::unique_ptr<KeyedService> BuildVpnService_V2(
+    content::BrowserContext* context) {
+  // Return stub implementation.
+  return std::make_unique<v2::BraveVpnServiceImpl>();
+}
+
+#elif BUILDFLAG(ENABLE_BRAVE_VPN_V1)
+
+std::unique_ptr<KeyedService> BuildVpnService_V1(
+    content::BrowserContext* context) {
 #if !BUILDFLAG(IS_ANDROID)
   if (!g_brave_browser_process->brave_vpn_connection_manager()) {
     return nullptr;
   }
-#endif
+#endif  // !BUILDFLAG(IS_ANDROID)
 
   auto* default_storage_partition = context->GetDefaultStoragePartition();
   auto shared_url_loader_factory =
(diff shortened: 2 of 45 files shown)


components/brave_vpn/browser/BUILD.gn
```diff
@@ -7,21 +7,37 @@ import("//brave/components/brave_vpn/common/buildflags/buildflags.gni")
...
+# Browser sources shared between v1 and v2 implementations.
+source_set("shared") {
+  sources = [
+    "brave_vpn_service.cc",
+    "brave_vpn_service.h",
+    "brave_vpn_service_observer.cc",
+    "brave_vpn_service_observer.h",
+  ]
+
+  deps = [
```
netzenbot: The new `source_set("shared")` exposes `brave_vpn_service.h` publicly, which includes headers from `//brave/components/brave_vpn/common/mojom`, `//brave/components/skus/browser`, and `//components/keyed_service/core`. Per [BS-024](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-024), these must be in `public_deps` since they appear in the target's own header files — otherwise consumers won't get the transitive include paths they need.

components/brave_vpn/v2/browser/BUILD.gn
```diff
@@ -0,0 +1,44 @@
...
+assert(enable_brave_vpn)
+
+static_library("browser") {
+  sources = [
+    "brave_vpn_service_impl.cc",
+    "brave_vpn_service_impl.h",
+  ]
+
+  deps = [
+    "//base",
```
netzenbot: `brave_vpn_service_impl.h` inherits from `BraveVpnService` which is defined in `//brave/components/brave_vpn/browser:shared`, so that dep must be in `public_deps` rather than `deps`. Per [BS-024](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-024), header-visible dependencies must be in `public_deps` so consumers of this target also get the transitive include paths.

components/brave_vpn/browser/BUILD.gn
```diff
@@ -7,21 +7,37 @@ import("//brave/components/brave_vpn/common/buildflags/buildflags.gni")
 
 assert(enable_brave_vpn)
```
netzenbot, line 9: The new `source_set("shared")` is consumed by multiple external modules but has no `visibility` restriction, making it a de-facto public target. Per [BS-028](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-028), public targets should use `static_library` (or `component`), not `source_set` — the existing target in this file already uses `static_library`. Either convert this to `static_library("shared")` or add `visibility = [ ":*" ]` and fold the sources into `:browser`.

components/brave_vpn/v2/browser/brave_vpn_service_impl_unittest.cc
```diff
@@ -0,0 +1,21 @@
...
+#include "brave/components/brave_vpn/v2/browser/brave_vpn_service_impl.h"
+
+#include "testing/gtest/include/gtest/gtest.h"
+
+namespace brave_vpn {
+namespace v2 {
+
+TEST(BraveVpnServiceImplTest, Test) {
+  // This test just verifies that the test fixture compiles and links correctly
+  // with the real BraveVpnServiceImpl implementation.
```
netzenbot, line 15: nit: This is a change detector test — `IsBraveVPNEnabled()` is hardcoded to `return true`, so `EXPECT_TRUE(service.IsBraveVPNEnabled())` always passes and validates nothing. Per [ARCH-062b](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-062b), tests should validate behavior across meaningful cases. If this is purely a compile/link smoke test, it would be better to just leave a comment explaining that, or omit the assertion entirely.

components/brave_vpn/v2/DEPS
```diff
@@ -0,0 +1,9 @@
+include_rules = [
```
netzenbot, line 1: nit: This PR introduces the new `components/brave_vpn/v2/` directory which represents a parallel VPN architecture. Per [DOC-005](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-005), major directories should have a README.md that describes the purpose, key files, and relationship to the v1 architecture. A brief README here would help orient developers working on the VPN 2.0 migration.

components/brave_vpn/v2/browser/brave_vpn_service_impl_unittest.cc
```diff
@@ -0,0 +1,21 @@
...
+
+#include "brave/components/brave_vpn/v2/browser/brave_vpn_service_impl.h"
+
+#include "testing/gtest/include/gtest/gtest.h"
+
+namespace brave_vpn {
+namespace v2 {
+
+TEST(BraveVpnServiceImplTest, Test) {
+  // This test just verifies that the test fixture compiles and links correctly
```
netzenbot, line 14: nit: `Test` is not a descriptive name per [TI-038](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-038). Even for a compilation/linkage smoke test, prefer something like `IsBraveVPNEnabled_ReturnsTrue` or `CompilesAndIsBraveVPNEnabledReturnsTrue` so the intent is clear without reading the body.

components/brave_vpn/browser/BUILD.gn
alekspop26: Fixed

components/brave_vpn/v2/browser/BUILD.gn
alekspop26: Fixed

components/brave_vpn/browser/BUILD.gn
alekspop26: Fixed

components/brave_vpn/v2/browser/brave_vpn_service_impl_unittest.cc
alekspop26: Fixed

components/brave_vpn/v2/browser/brave_vpn_service_impl.h
```diff
@@ -0,0 +1,125 @@
...
+
+#ifndef BRAVE_COMPONENTS_BRAVE_VPN_V2_BROWSER_BRAVE_VPN_SERVICE_IMPL_H_
+#define BRAVE_COMPONENTS_BRAVE_VPN_V2_BROWSER_BRAVE_VPN_SERVICE_IMPL_H_
+
+#include "brave/components/brave_vpn/browser/brave_vpn_service.h"
+#include "brave/components/skus/common/skus_sdk.mojom.h"
+#include "build/build_config.h"
+#include "mojo/public/cpp/bindings/pending_remote.h"
+#include "mojo/public/cpp/bindings/remote_set.h"
+
```
netzenbot, line 13: nit: `brave/components/skus/common/skus_sdk.mojom.h` is included here but no `skus::mojom` types appear in the class declaration. Per [CS-070](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-070), the full mojom header should be moved to the `.cc` where the types are actually used (or removed if unused there too).

components/brave_vpn/v2/DEPS
alekspop26: Fixed.

components/brave_vpn/v2/browser/brave_vpn_service_impl_unittest.cc
alekspop26: Fixed.

components/brave_vpn/v2/browser/brave_vpn_service_impl.h
alekspop26: Fixed.

browser/brave_vpn/brave_vpn_service_factory.cc
```diff
@@ -7,48 +7,60 @@
...
-#include "chrome/browser/browser_process.h"
-#include "chrome/browser/profiles/incognito_helpers.h"
 #include "chrome/browser/profiles/profile.h"
 #include "components/keyed_service/content/browser_context_dependency_manager.h"
-#include "components/user_prefs/user_prefs.h"
 #include "content/public/browser/browser_context.h"
-#include "content/public/browser/storage_partition.h"
 
+#if BUILDFLAG(ENABLE_BRAVE_VPN_V1)
+#include "base/functional/bind.h"
```
bridiver, line 22: guards in gn and cpp should match so only files guarded by ENABLE_BRAVE_VPN_V1 in gn should be behind the cpp buildflag

browser/brave_vpn/BUILD.gn
```diff
@@ -7,17 +7,32 @@ import("//brave/components/brave_vpn/common/buildflags/buildflags.gni")
 
 assert(enable_brave_vpn)
 
-source_set("brave_vpn") {
+source_set("shared") {
```
bridiver, line 10: `utils`?

components/brave_vpn/browser/BUILD.gn
```diff
@@ -7,21 +7,38 @@ import("//brave/components/brave_vpn/common/buildflags/buildflags.gni")
...
 
+# Browser sources shared between v1 and v2 implementations.
+static_library("shared") {
```
bridiver, line 11: `brave_vpn_service`?

components/brave_vpn/common/buildflags/buildflags.gni
```diff
@@ -20,6 +20,21 @@ declare_args() {
...
 }
+
+assert(!enable_brave_vpn || enable_brave_vpn_v1 || enable_brave_vpn_v2,
+       "If Brave VPN is enabled, at least one VPN architecture must be enabled")
+
+# Build Brave VPN 2.0 deps only if VPN is enabled and architecture 2.0 is supported.
+enable_brave_vpn_v2_deps = enable_brave_vpn && enable_brave_vpn_v2
+
+# Enable Brave VPN 1.0 architecture specific unit tests - turn off for debugging only.
+enable_brave_vpn_v1_tests = true
```
bridiver, line 37: why would tests have their own buildflag?

alekspop26: Because tests are always built for both v1 and v2, regardless of the feature being enabled or not. That's to make sure we always build and test the new feature even if it's compiled out in production. The particular test-related flags are for local development and facilitating the migration in the future - so that we can check we can disable certain architecture in testing, too, and nothing breaks (i.e. no unwanted test dependencies).

bridiver: > That's to make sure we always build and test the new feature even if it's compiled out in production

This isn't something that we do anywhere else. If we need another build in CI then we can add another build for vpn v2

alekspop26: That was one of the VPN 2.0 tenets from the very beginning: "Continuous unit testing to ensure seamless integration." In other words, the feature is not in prod, but continuously tested in CI to catch the issues as early as possible.

netzenbot (approved):

alekspop26:
Split into multiple smaller changes:
* https://github.com/brave/brave-core/pull/36757
* https://github.com/brave/brave-core/pull/36763
* https://github.com/brave/brave-core/pull/36787
* https://github.com/brave/brave-core/pull/36851
* https://github.com/brave/brave-core/pull/36935

PR #36624, opened 2026-05-21
Add Semantic History Search toggle history page

### LINKED ISSUE #55725: Add a setting for semantic history search
We need a toggle on shown when brave://flags/#brave-history-embeddings is enabled to control the whole feature instead of solely relying on flag so we can do Griffin random rolled out without surprising users and give users option to opt-in (default off)

darkdh:
Adds a persistent on/off toggle for the history embeddings feature directly inside the `brave://history` side bar, shown only when the `brave://flags/#brave-history-embeddings` flag is enabled. The toggle is backed by a new Brave-owned boolean pref (`brave.history_embeddings_enabled`, owned by `//brave/components/local_ai/core`, default off), which gates `IsHistoryEmbeddingsEnabledForProfile()` so flipping it actually controls whether pages are indexed and searched. Flipping the toggle triggers a page reload so the rest of the history UI picks up the new state immediately.

https://github.com/user-attachments/assets/b69b4142-93a8-4657-8662-cdf57d58103f

### Toggle visibility

1. Enable `brave://flags/#brave-history-embeddings`.
2. Navigate to `brave://history`.
3. The **Semantic History Search** toggle should appear in the side bar.
4. Disable `brave://flags/#brave-history-embeddings`.
5. Navigate to `brave://history`.
6. The **Semantic History Search** toggle should not appear.

### Toggle on/off

1. Enable `brave://flags/#brave-history-embeddings`.
2. Navigate to `brave://history`.
3. The side bar toggle should be off by default; the semantic search input should not appear at the top of the history list.
4. Flip the toggle on. The page reloads.
5. The semantic search input should now appear above the history list.
6. Flip the toggle off. The page reloads.
7. The semantic search input should disappear.

Resolves https://github.com/brave/brave-browser/issues/55725

### DIFF
--- a/browser/brave_profile_prefs.cc
+++ b/browser/brave_profile_prefs.cc
@@ -43,6 +43,7 @@
 #include "brave/components/debounce/core/browser/debounce_service.h"
 #include "brave/components/global_privacy_control/pref_names.h"
 #include "brave/components/ipfs/ipfs_prefs.h"
+#include "brave/components/local_ai/core/pref_names.h"
 #include "brave/components/ntp_background_images/browser/view_counter_service.h"
 #include "brave/components/ntp_background_images/buildflags/buildflags.h"
 #include "brave/components/ntp_background_images/common/view_counter_pref_registry.h"
@@ -555,6 +556,8 @@ void RegisterProfilePrefs(user_prefs::PrefRegistrySyncable* registry) {
   ai_chat::prefs::RegisterProfilePrefs(registry);
   ai_chat::ModelService::RegisterProfilePrefs(registry);
 #endif
+
+  local_ai::prefs::RegisterProfilePrefs(registry);
 
   brave_account::prefs::RegisterPrefs(registry);
   brave_origin::prefs::RegisterProfilePrefs(registry);
--- a/browser/resources/history/brave_history_embeddings_browser_proxy.ts
+++ b/browser/resources/history/brave_history_embeddings_browser_proxy.ts
@@ -1,0 +1,31 @@
+// Copyright (c) 2026 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import {
+  PageCallbackRouter,
+  PageHandlerFactory,
+  PageHandlerRemote,
+} from './brave_history_embeddings.mojom-webui.js'
+
+export interface BraveHistoryEmbeddingsBrowserProxy {
+  pageHandler: PageHandlerRemote
+  callbackRouter: PageCallbackRouter
+}
+
+let instance: BraveHistoryEmbeddingsBrowserProxy | null = null
+
+export function getBraveHistoryEmbeddingsBrowserProxy(): BraveHistoryEmbeddingsBrowserProxy {
+  if (!instance) {
+    const factory = PageHandlerFactory.getRemote()
+    const pageHandler = new PageHandlerRemote()
+    const callbackRouter = new PageCallbackRouter()
+    factory.createPageHandler(
+      callbackRouter.$.bindNewPipeAndPassRemote(),
+      pageHandler.$.bindNewPipeAndPassReceiver(),
+    )
+    instance = { pageHandler, callbackRouter }
+  }
+  return instance
+}
--- a/browser/ui/webui/history/brave_history_embeddings.mojom
+++ b/browser/ui/webui/history/brave_history_embeddings.mojom
@@ -1,0 +1,26 @@
+// Copyright (c) 2026 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+module brave_history_embeddings.mojom;
+
+// chrome://history → browser. Lets the side bar toggle write the
+// Semantic History Search pref and listen for changes.
+interface PageHandlerFactory {
+  // Binds a `Page` for browser→renderer notifications and binds a
+  // `PageHandler` for renderer→browser calls.
+  CreatePageHandler(pending_remote<Page> page,
+                    pending_receiver<PageHandler> handler);
+};
+
+interface PageHandler {
+  // Writes the new value of the brave.history_embeddings_enabled pref.
+  SetEnabled(bool enabled);
+};
+
+// Notifications from the browser back to all chrome://history pages so any
+// UI consuming the enabled state can react without a page reload.
+interface Page {
+  OnEnabledChanged(bool enabled);
+};
--- a/chromium_src/chrome/browser/resources/history/app.ts
+++ b/chromium_src/chrome/browser/resources/history/app.ts
@@ -4,9 +4,13 @@
 // You can obtain one at https://mozilla.org/MPL/2.0/.
 
 import {injectStyle} from '//resources/brave/lit_overriding.js'
+import {loadTimeData} from '//resources/js/load_time_data.js'
 import {css} from '//resources/lit/v3_0/lit.rollup.js'
 
 import {HistoryAppElement} from './app-chromium.js'
+import {
+  getBraveHistoryEmbeddingsBrowserProxy,
+} from './brave_history_embeddings_browser_proxy.js'
 
 injectStyle(HistoryAppElement, css`
         #tabs-content, #tabs-content>* {
@@ -14,5 +18,28 @@ injectStyle(HistoryAppElement, css`
         }
     </style>
 `)
+
+// Subscribe directly to the Mojo `onEnabledChanged` notification so the
+// chrome://history UI reacts even when the side bar isn't rendered. Update
+// each history-app's reactive `enableHistoryEmbeddings_` accessor (Lit
+// re-renders the app shell), refresh loadTimeData so any code that reads
+// `enableHistoryEmbeddings` inline picks up the new value, and force a
+// re-render of the toolbar/list whose `compute*_` methods cache the value.
+getBraveHistoryEmbeddingsBrowserProxy()
+    .callbackRouter.onEnabledChanged.addListener((enabled: boolean) => {
+      loadTimeData.overrideValues({enableHistoryEmbeddings: enabled})
+      for (const app of document.querySelectorAll('history-app')) {
+        ;(app as unknown as {enableHistoryEmbeddings_: boolean})
+            .enableHistoryEmbeddings_ = enabled
+      }
+      const root = document.querySelector('history-app')?.shadowRoot
+      if (!root) {
+        return
+      }
+      for (const el of root.querySelectorAll(
+               'history-side-bar, history-toolbar, history-list')) {
+        ;(el as unknown as {requestUpdate?: () => void}).requestUpdate?.()
+      }
+    })
 
 export * from './app-chromium.js'
--- a/chromium_src/chrome/browser/resources/history/side_bar.ts
+++ b/chromium_src/chrome/browser/resources/history/side_bar.ts
@@ -3,12 +3,21 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this file,
 // You can obtain one at https://mozilla.org/MPL/2.0/.
 
+import 'chrome://resources/cr_elements/cr_toggle/cr_toggle.js'
+
 import {injectStyle} from '//resources/brave/lit_overriding.js'
+import {loadTimeData} from '//resources/js/load_time_data.js'
 import {css} from '//resources/lit/v3_0/lit.rollup.js'
+import type {PropertyValues} from '//resources/lit/v3_0/lit.rollup.js'
 
-import {HistorySideBarElement} from './side_bar-chromium.js'
+import {
+  HistorySideBarElement as HistorySideBarElementChromium,
+} from './side_bar-chromium.js'
+import {
+  getBraveHistoryEmbeddingsBrowserProxy,
+} from './brave_history_embeddings_browser_proxy.js'
 
-injectStyle(HistorySideBarElement, css`
+injectStyle(HistorySideBarElementChromium, css`
   .cr-nav-menu-item {
     min-height: 20px !important;
     border-end-end-radius: 0px !important;
@@ -29,6 +38,72 @@ injectStyle(HistorySideBarElement, css`
   .cr-nav-menu-item cr-ripple {
     display: none !important;
   }
+  #brave-history-embeddings-toggle {
+    align-items: center;
+    border-top: 1px solid var(--cr-separator-color);
+    display: flex;
+    gap: 12px;
+    margin-top: 8px;
+    padding: 12px 16px 8px;
+  }
+  #brave-history-embeddings-toggle .text {
+    flex: 1;
+  }
+  #brave-history-embeddings-toggle .label {
+    font-weight: 500;
+  }
+  #brave-history-embeddings-toggle .description {
+    color: var(--cr-secondary-text-color);
+    font-size: 12px;
+    margin-top: 2px;
+  }
 `)
 
+// Declaration-merge the Brave-only reactive properties onto the upstream
+// class type so the lit_mangler-injected template (typed with
+// `this: HistorySideBarElement` via the upstream side_bar.html.ts import)
+// type-checks.
+declare module './side_bar-chromium.js' {
+  interface HistorySideBarElement {
+    braveHistoryEmbeddingsFeatureEnabled: boolean
+    braveHistoryEmbeddingsEnabled: boolean
+    onBraveHistoryEmbeddingsToggleChange(e: CustomEvent<boolean>): void
+  }
+}
+
+class HistorySideBarElement extends HistorySideBarElementChromium {
+  static override get properties() {
+    return {
+      ...super.properties,
+      braveHistoryEmbeddingsFeatureEnabled: {type: Boolean},
+      braveHistoryEmbeddingsEnabled: {type: Boolean},
+    }
+  }
+
+  override accessor braveHistoryEmbeddingsFeatureEnabled: boolean =
+      loadTimeData.getBoolean('isHistoryEmbeddingsFeatureEnabled')
+  override accessor braveHistoryEmbeddingsEnabled: boolean =
+      loadTimeData.getBoolean('enableHistoryEmbeddings')
+
+  // Re-read the toggle state from loadTimeData on every render cycle. The
+  // single Mojo subscription in app.ts updates loadTimeData and then calls
+  // requestUpdate() on this element, the toolbar, and the list — keeping all
+  // consumers on the same refresh mechanism.
+  override willUpdate(changedProperties: PropertyValues<this>) {
+    super.willUpdate(changedProperties)
+    this.braveHistoryEmbeddingsEnabled =
+        loadTimeData.getBoolean('enableHistoryEmbeddings')
+  }
+
+  override onBraveHistoryEmbeddingsToggleChange(e: CustomEvent<boolean>) {
+    getBraveHistoryEmbeddingsBrowserProxy().pageHandler.setEnabled(e.detail)
+  }
+}
+
+export {HistorySideBarElement}
 export * from './side_bar-chromium.js'
+
+// Register the Brave subclass instead of the upstream class. The matching
+// `customElements.define` in upstream side_bar.ts is patched out (see
+// patches/chrome-browser-resources-history-side_bar.ts.patch).
+customElements.define(HistorySideBarElement.is, HistorySideBarElement)
--- a/chromium_src/chrome/browser/ui/webui/history/history_ui.cc
+++ b/chromium_src/chrome/browser/ui/webui/history/history_ui.cc
@@ -1,0 +1,51 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+// Inject two loadTimeData entries (`isHistoryEmbeddingsFeatureEnabled` plus
+// the toggle's localized strings) into the chrome://history WebUI data
+// source. Hooked via macro substitution of the one-and-only
+// ManagedUIHandler::Initialize() call in the upstream constructor — by the
+// time it runs the data source has been created and is in scope. The
+// Mojo interface for write/observer plumbing lives in BraveHistoryUI; this
+// override only augments the data source.
+
+#include "brave/components/local_ai/core/pref_names.h"
+#include "brave/grit/brave_generated_resources.h"
+#include "chrome/browser/history_embeddings/history_embeddings_utils.h"
+#include "chrome/browser/profiles/profile.h"
+#include "chrome/browser/ui/webui/managed_ui_handler.h"
+#include "components/prefs/pref_service.h"
+#include "content/public/browser/web_ui.h"
+#include "content/public/browser/web_ui_data_source.h"
+#include "ui/base/webui/web_ui_util.h"
+
+namespace {
+
+class BraveHistoryUIInitializer {
+ public:
+  static void Initialize(content::WebUI* web_ui,
+                         content::WebUIDataSource* source) {
+    ManagedUIHandler::Initialize(web_ui, source);
+
+    source->AddBoolean("isHistoryEmbeddingsFeatureEnabled",
+                       history_embeddings::IsHistoryEmbeddingsFeatureEnabled());
+
+    static constexpr webui::LocalizedString kBraveHistoryEmbeddingsStrings[] = {
+        {"braveHistoryEmbeddingsToggleLabel",
+         IDS_BRAVE_HISTORY_EMBEDDINGS_TOGGLE_LABEL},
+        {"braveHistoryEmbeddingsToggleDescription",
+         IDS_BRAVE_HISTORY_EMBEDDINGS_TOGGLE_DESCRIPTION},
+    };
+    source->AddLocalizedStrings(kBraveHistoryEmbeddingsStrings);
+  }
+};
+
+}  // namespace
+
+#define ManagedUIHandler BraveHistoryUIInitializer
+
+#include <chrome/browser/ui/webui/history/history_ui.cc>
+
+#undef ManagedUIHandler
--- a/app/brave_generated_resources.grd
+++ b/app/brave_generated_resources.grd
@@ -1508,6 +1508,14 @@ Or change later at <ph name="SETTINGS_EXTENIONS_LINK">$2<ex>brave://settings/ext
       <message name="IDS_BRAVE_HISTORY_EMBEDDINGS_DISCLAIMER" desc="Disclaimer shown on chrome://history and in the history search IPH promo, indicating that history semantic search runs entirely on-device.">
         Your searches and page contents are processed on this device. Nothing is sent to Brave or seen by human reviewers.
       </message>
+
+      <!-- Semantic History Search toggle in the chrome://history side bar. -->
+      <message name="IDS_BRAVE_HISTORY_EMBEDDINGS_TOGGLE_LABEL" desc="Label for the toggle on chrome://history that turns Semantic History Search on or off.">
+        Semantic History Search
+      </message>
+      <message name="IDS_BRAVE_HISTORY_EMBEDDINGS_TOGGLE_DESCRIPTION" desc="One-line explainer below the Semantic History Search toggle on chrome://history.">
+        Index pages you visit locally so you can search your history by meaning. All processing stays on this device.
+      </message>
       <!--Add new items to the appropriate sections above -->
     </messages>
   </release>
--- a/browser/ai_chat/BUILD.gn
+++ b/browser/ai_chat/BUILD.gn
@@ -426,6 +426,7 @@ source_set("browser_tests") {
       "//brave/components/ai_chat/core/common/mojom",
       "//brave/components/constants",
       "//brave/components/l10n/common:test_support",
+      "//brave/components/local_ai/core",
       "//brave/components/resources:strings_grit",
       "//brave/components/sidebar/browser",
       "//brave/components/text_recognition/common/buildflags",
--- a/browser/ai_chat/tools/history_search_tool_browsertest.cc
+++ b/browser/ai_chat/tools/history_search_tool_browsertest.cc
@@ -29,10 +29,14 @@
 #include "brave/components/ai_chat/core/browser/engine/mock_engine_consumer.h"
 #include "brave/components/ai_chat/core/browser/tools/tool_provider.h"
 #include "brave/components/ai_chat/core/common/mojom/ai_chat.mojom.h"
+#include "brave/components/local_ai/core/pref_names.h"
+#include "chrome/browser/profiles/profile.h"
+#include "chrome/browser/ui/browser.h"
 #include "components/history/core/browser/url_row.h"
 #include "components/history_embeddings/core/history_embeddings_features.h"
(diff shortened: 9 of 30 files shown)


app/brave_settings_strings.grdp
```diff
@@ -1632,6 +1632,12 @@
   <message name="IDS_SETTINGS_LEO_ASSISTANT_TAB_ORGANIZATION_MODEL_DESC" desc="The description for the tab organization model selection dropdown, with a link to a support page about Tab Focus Mode">
     The model used to organize your tabs in Tab Focus Mode. <ph name="BEGIN_LINK">&lt;a target="_blank" href="$1"&gt;</ph>Learn more<ph name="END_LINK">&lt;/a&gt;</ph>
   </message>
+  <message name="IDS_SETTINGS_LEO_ASSISTANT_HISTORY_EMBEDDINGS_LABEL" desc="The text for settings option to enable semantic history search">
+    Semantic History Search
+  </message>
+  <message name="IDS_SETTINGS_LEO_ASSISTANT_HISTORY_EMBEDDINGS_DESC" desc="The description for settings option to enable semantic history search">
+    Leo AI will index the pages you visit in the background so you can search your history using everyday language. Indexing and search both run locally on your device — your history data is never sent to a remote server.
```
diracdeltas: Should this be more clear that history data is actually sent to a remote server when you accept the permission prompt?

`your history data is not sent to a remote server until you give Leo permission to use it for context` or something

cc @clening for review too

darkdh: I thought Leo per conversation ask permission is sufficient. The main feature is for history search, exporting it as a [tool for Leo](https://github.com/brave/brave-core/pull/36385) is additional. But I guess it still worth mentioning here, I will clarify it.

diracdeltas: yes, i think `never sent` is confusing here

darkdh: fixed by 06f4c68924d714b546b144ea2ae0f1db64318039

diracdeltas: also maybe this should be more clear that the local search feature is in brave://history. because of where this setting is, the user might assume it is done via the leo chat.

darkdh: what about

Search your browsing history using everyday language from brave://history or the @history omnibox shortcut. Pages are indexed and searched locally on your device — no history data leaves your device for this feature. Search is also exported as a tool that Leo can use; when enabled, Leo will only access it after you grant explicit permission for a given conversation.

diracdeltas: that looks good to me

browser/brave_profile_prefs.cc
```diff
@@ -548,6 +549,8 @@ void RegisterProfilePrefs(user_prefs::PrefRegistrySyncable* registry) {
   ai_chat::ModelService::RegisterProfilePrefs(registry);
 #endif
 
+  local_ai::prefs::RegisterProfilePrefs(registry);
+
```
netzenbot, line 561: The `#include "brave/components/local_ai/core/pref_names.h"` is inside the `#if BUILDFLAG(ENABLE_AI_CHAT)` block, but `local_ai::prefs::RegisterProfilePrefs(registry)` is called outside it. If `ENABLE_AI_CHAT` is disabled, the header won't be included and this call won't compile. Either move the include outside the guard, or wrap the call inside the guard too — whichever matches the intent.

browser/brave_profile_prefs.cc
```diff
@@ -95,6 +95,7 @@
 #include "brave/components/ai_chat/core/browser/model_service.h"
 #include "brave/components/ai_chat/core/common/features.h"
 #include "brave/components/ai_chat/core/common/pref_names.h"
+#include "brave/components/local_ai/core/pref_names.h"
```
netzenbot: The `#include "brave/components/local_ai/core/pref_names.h"` is inside the `#if BUILDFLAG(ENABLE_AI_CHAT)` block, but `local_ai::prefs::RegisterProfilePrefs(registry)` is called unconditionally after the `#endif`. This will cause a compile error when `ENABLE_AI_CHAT` is disabled. Move the include to an unconditional section (local_ai is not an AI Chat-only component).

browser/brave_profile_prefs.cc
```diff
@@ -95,6 +95,7 @@
...
 #endif
```
netzenbot, line 102: The `#include "brave/components/local_ai/core/pref_names.h"` is nested inside `#if BUILDFLAG(ENABLE_AI_CHAT)`, but `local_ai::prefs::RegisterProfilePrefs(registry)` is called unconditionally below. This will fail to compile when `ENABLE_AI_CHAT` is false (e.g., on branded origin builds). Move the include outside the `ENABLE_AI_CHAT` guard.

browser/resources/settings/brave_leo_assistant_page/brave_leo_assistant_page.html
```diff
@@ -51,6 +51,14 @@
...
+    if="[[isHistoryEmbeddingsFeatureEnabled_]]">
+    <settings-toggle-button class="cr-row hr"
+      pref="{{prefs.brave.history_embeddings_enabled}}"
+      label="$i18n{braveLeoAssistantHistoryEmbeddingsLabel}"
+      sub-label="$i18n{braveLeoAssistantHistoryEmbeddingsDesc}">
+    </settings-toggle-button>
+  </template>
   <cr-link-row class="hr"
     label="$i18n{braveLeoAssistantAutocompleteLink}"
     on-click="openAutocompleteSetting_">
```
netzenbot, line 56: This PR adds a new user-visible toggle to the Leo settings page but includes no `InteractiveBrowserTest` covering the CUJ (e.g. toggling the pref and verifying the feature responds). Per [ARCH-062](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-062), every feature that adds or modifies user-visible UI needs at least one CUJ test.

browser/brave_profile_prefs.cc
darkdh: This is intended I will move the header include outside of `ENABLE_AI_CHAT`

browser/resources/settings/brave_leo_assistant_page/brave_leo_assistant_page.html
darkdh: There is currently no existing test for this page, I'm not responsible for adding that while only modifying this file.
That is out of scope of this PR.

browser/brave_profile_prefs.cc
darkdh: moved to the group without guard

browser/brave_profile_prefs.cc
darkdh: moved to the group without guard

app/brave_settings_strings.grdp
petemill: 1) It's confusing it's in the Leo section to be honest, when it's not connected. But maybe that's a branding thing and Leo is all AI and ML features? 

@rebron @aguscruiz 

2) Either way, this is buried and it would be much more discoverable exposed where users do history search, e.g. exposed as a button on `brave://history`.

3) Are we planning on enabling history search tool in Leo without semantic history search, e.g search by title or Url when the user hasn't enabled semantic search indexing?

app/brave_settings_strings.grdp
```diff
@@ -1632,6 +1632,12 @@
   <message name="IDS_SETTINGS_LEO_ASSISTANT_TAB_ORGANIZATION_MODEL_DESC" desc="The description for the tab organization model selection dropdown, with a link to a support page about Tab Focus Mode">
     The model used to organize your tabs in Tab Focus Mode. <ph name="BEGIN_LINK">&lt;a target="_blank" href="$1"&gt;</ph>Learn more<ph name="END_LINK">&lt;/a&gt;</ph>
   </message>
+  <message name="IDS_SETTINGS_LEO_ASSISTANT_HISTORY_EMBEDDINGS_LABEL" desc="The text for settings option to enable semantic history search">
+    Semantic History Search
+  </message>
```
netzenbot: nit: The toggle label "Semantic History Search" uses title case. Per [SG-008](https://github.com/brave/brave-core/tree/master/docs/best-practices/style-guide.md#SG-008), sentence case is the default for settings labels — title case is reserved for branded product names. Should be "Semantic history search".

app/brave_settings_strings.grdp
```diff
@@ -1632,6 +1632,12 @@
...
+  <message name="IDS_SETTINGS_LEO_ASSISTANT_HISTORY_EMBEDDINGS_DESC" desc="The description for settings option to enable semantic history search">
+    Search your browsing history using everyday language from brave://history or the @history omnibox shortcut. Pages are indexed and searched locally on your device — no history data leaves your device for this feature. Search is also exported as a tool that Leo can use; when enabled, Leo will only access it after you grant explicit permission for a given conversation.
+  </message>
```
netzenbot: nit: The em dash in "on your device — no history data" has spaces before and after it. Per [SG-010](https://github.com/brave/brave-core/tree/master/docs/best-practices/style-guide.md#SG-010), em dashes take no spaces: "on your device—no history data leaves your device for this feature."

app/brave_settings_strings.grdp
darkdh: @petemill moved the toggle to history page like these
Screenshot 2026-06-01 at 15 22 39
Screenshot 2026-06-01 at 15 22 50

fallaciousreasoning:
I know its a bit of a pain, but do you reckon we could get this working without the page reload? None of our other UIs reload when you change a setting and its a bit jarring. Happy for it to be a follow up, as this is still behind a flag.

chromium_src/chrome/browser/ui/webui/history/history_ui.cc
```diff
@@ -0,0 +1,94 @@
...
+
+  ~BraveHistoryEmbeddingsToggleHandler() override = default;
+
+  void RegisterMessages() override {
+    web_ui()->RegisterMessageCallback(
+        "setHistoryEmbeddingsEnabled",
+        base::BindRepeating(&BraveHistoryEmbeddingsToggleHandler::
+                                HandleSetHistoryEmbeddingsEnabled,
+                            base::Unretained(this)));
+  }
```
fallaciousreasoning: Its probably borderline here because this exposing a pref to the UI but in general we shouldn't be adding new MessageHandlers and should add a mojom interface instead. Once set up it makes it much easier to add/change things and everything is typechecked.

fallaciousreasoning: I think it probably is worth doing here.

chromium_src/chrome/browser/resources/history/side_bar.ts
```diff
@@ -29,6 +32,52 @@ injectStyle(HistorySideBarElement, css`
...
+
+globalThis.braveHistoryEmbeddingsFeatureEnabled = featureEnabled
+globalThis.braveHistoryEmbeddingsEnabled = initialEnabled
+globalThis.braveHistoryEmbeddingsOnChange = (e: CustomEvent<boolean>) => {
+  chrome.send('setHistoryEmbeddingsEnabled', [e.detail])
+  // The upstream history WebUI consumes `enableHistoryEmbeddings` once at
+  // page load via loadTimeData, so the search UI doesn't react to pref
+  // changes mid-session. Reload so the new state takes effect immediately.
+  window.location.reload()
+}
```
fallaciousreasoning: Hey, this doesn't seem ideal, and its the reason we need to do the page reload. I think it'd be better to patch these onto the `CrLitElement` rather than placing these flags/event handlers on `window` like this

chromium_src/chrome/browser/resources/history/side_bar.ts
```diff
@@ -29,6 +32,52 @@ injectStyle(HistorySideBarElement, css`
...
+    margin-top: 2px;
+  }
 `)
 
+// Exposed as module-level bindings (not class properties) so the
+// lit_mangler-injected template in side_bar.html.ts can reference them
+// without tripping @webui-eslint/lit-reactive-properties, which only fires
+// on `this.<prop>` accesses.
+const featureEnabled =
+    loadTimeData.valueExists('isHistoryEmbeddingsFeatureEnabled') &&
```
fallaciousreasoning: do we need these `valueExists` checks? This should be either true or false at loadTime.

chromium_src/chrome/browser/resources/history/side_bar.ts
```diff
@@ -29,6 +32,52 @@ injectStyle(HistorySideBarElement, css`
...
+    loadTimeData.getBoolean('isHistoryEmbeddingsFeatureEnabled')
+const initialEnabled =
+    loadTimeData.valueExists('isHistoryEmbeddingsEnabled') &&
```
fallaciousreasoning: ditto

chromium_src/chrome/browser/resources/history/side_bar.ts
fallaciousreasoning: I think the way to do this is probable something like this:
1. In this file:
```ts
declare module './side_bar-chromium.js' {
  interface HistorySideBarElement {
    braveHistoryEmbeddingsFeatureEnabled(): boolean
    onHistoryEmbeddingsEnabledChange(e: CustomEvent<boolean>): void
  }
}

HistorySideBarElement.prototype.braveHistoryEmbeddingsFeatureEnabled = () => loadTimeData.getBoolean('isHistoryEmbeddingsFeatureEnabled')
HistorySideBarElement.prototype.onHistoryEmbeddingsEnabledChange = (e: CustomEvent<boolean>) => {
  chrome.send('setHistoryEmbeddingsEnabled', [e.detail])
  // The upstream history WebUI consumes `enableHistoryEmbeddings` once at
  // page load via loadTimeData, so the search UI doesn't react to pref
  // changes mid-session. Reload so the new state takes effect immediately.
  window.location.reload()
}

// TODO: override `connectedCallback to setup a listener for the value
const oldConnectedCallback = HistorySideBarElement.prototype.connectedCallback
HistorySideBarElement.prototype.connectedCallback = function() {
    // TODO: Add listener
    oldConnectedCallback.apply(this)
}
```

2. Unfortunately with the new eslint rules it looks like we can't add Reactive properties in the `chromium_src` override anymore. We should add a plaster and patch in:

```ts
// in get properties() {
braveHistoryEmbeddingsFeatureEnabled: { type: boolean }

// as a class property:
protected accessor braveHistoryEmbeddingsEnabled: boolean = loadTimeData.getBoolean('isHistoryEmbeddingsEnabled')
```

I've filed https://github.com/brave/brave-browser/issues/55994 to investigate finding a better way of doing this.

fallaciousreasoning (changes requested):

chromium_src/chrome/browser/resources/history/side_bar.ts
fallaciousreasoning: I thought of another approach which is a bit cleaner:

1. In the chromium_src override:
```ts
import { HistorySideBarElement as BaseHistorySideBarElementChromium } from './side_bar-chromium.js'

declare module './side_bar-chromium.js' {
  interface HistorySideBarElement {
    braveHistoryEmbeddingsFeatureEnabled: boolean
    braveHistoryEmbeddingsEnabled: boolean
    onHistoryEmbeddingsEnabledChange(e: CustomEvent<boolean>): void
  }
}

class HistorySideBarElement extends BaseHistorySideBarElementChromium {
  static override get properties() {
    return {
      ...super.properties,
      braveHistoryEmbeddingsFeatureEnabled: { type: Boolean },
      braveHistoryEmbeddingsEnabled: { type: Boolean },
    }
  }

  override accessor braveHistoryEmbeddingsFeatureEnabled = loadTimeData.getBoolean('isHistoryEmbeddingsFeatureEnabled')
  override accessor braveHistoryEmbeddingsEnabled: boolean = loadTimeData.getBoolean('isHistoryEmbeddingsEnabled')

  override onHistoryEmbeddingsEnabledChange(e: CustomEvent<boolean>) {
    chrome.send('setHistoryEmbeddingsEnabled', [e.detail])
    // The upstream history WebUI consumes `enableHistoryEmbeddings` once at
    // page load via loadTimeData, so the search UI doesn't react to pref
    // changes mid-session. Reload so the new state takes effect immediately.
    window.location.reload()
  }
}

export { HistorySideBarElement }
export * from './side_bar-chromium.js'
customElements.define(HistorySideBarElement.is, HistorySideBarElement);
```

2. Add a patch for the upstream file commenting out the `customElements.define` line:
```ts
// customElements.define(HistorySideBarElement.is, HistorySideBarElement);
```

chromium_src/chrome/browser/resources/history/side_bar.ts
```diff
@@ -29,6 +32,52 @@ injectStyle(HistorySideBarElement, css`
...
+}
+
+globalThis.braveHistoryEmbeddingsFeatureEnabled = featureEnabled
+globalThis.braveHistoryEmbeddingsEnabled = initialEnabled
+globalThis.braveHistoryEmbeddingsOnChange = (e: CustomEvent<boolean>) => {
+  chrome.send('setHistoryEmbeddingsEnabled', [e.detail])
+  // The upstream history WebUI consumes `enableHistoryEmbeddings` once at
+  // page load via loadTimeData, so the search UI doesn't react to pref
+  // changes mid-session. Reload so the new state takes effect immediately.
+  window.location.reload()
```
fallaciousreasoning: I don't think we should be reloading the window - in an ideal world we'd control this through a mojom interface and any UI interested in this should listen to changes via that interface.

chromium_src/chrome/browser/ui/webui/history/history_ui.cc
```diff
@@ -0,0 +1,94 @@
...
+class BraveHistoryUIInitializer {
+ public:
+  static void Initialize(content::WebUI* web_ui,
+                         content::WebUIDataSource* source) {
+    ManagedUIHandler::Initialize(web_ui, source);
+
+    Profile* profile = Profile::FromWebUI(web_ui);
+    source->AddBoolean("isHistoryEmbeddingsFeatureEnabled",
+                       history_embeddings::IsHistoryEmbeddingsFeatureEnabled());
+    source->AddBoolean("isHistoryEmbeddingsEnabled",
```
fallaciousreasoning: We're already providing this to the WebUI via `enableHistoryEmbeddings` in `history_ui.cc` which is how the searchbox knows to change its form. 

Its not identical to this as its false when the feature flag is disabled, but I think that's probably better.

chromium_src/chrome/browser/resources/history/side_bar.ts
fallaciousreasoning: https://github.com/brave/brave-core/pull/36913

chromium_src/chrome/browser/resources/history/side_bar.ts
darkdh: you are right, removed in bfd74e6c82ca98007ed32a61545641547dcf3e74

chromium_src/chrome/browser/resources/history/side_bar.ts
darkdh: removed in bfd74e6c82ca98007ed32a61545641547dcf3e74

chromium_src/chrome/browser/ui/webui/history/history_ui.cc
darkdh: Yes, using upstream one is cleaner. done in 984e0ce92bacaabe242b0876c084f75fe5f0a573

chromium_src/chrome/browser/resources/history/side_bar.ts
darkdh: Screenshot 2026-06-02 at 10 09 47

Screenshot 2026-06-02 at 10 11 03

These are all upstream codes and they don't listen to anything and when its toggle changed, the history page doesn't reflect the updates either.  (Required manual reload to get updates)

I will try to replace chrome.send with mojo `SetEnabled/OnEnabledChanged` and then observer mojo events.
Worst case scenario would be same behavior as upstream.

chromium_src/chrome/browser/ui/webui/history/history_ui.cc
```diff
@@ -0,0 +1,91 @@
...
+  }
+};
+
+}  // namespace
+
+#define ManagedUIHandler BraveHistoryUIInitializer
+
+#include <chrome/browser/ui/webui/history/history_ui.cc>
+
+#undef ManagedUIHandler
```
netzenbot, line 51: This PR adds a new user-visible Semantic History Search toggle to chrome://history but includes no `InteractiveBrowserTest` covering the CUJ (e.g. toggling the pref and verifying `IsHistoryEmbeddingsEnabledForProfile` responds). Per [ARCH-062](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-062), every feature that adds or modifies user-visible UI needs at least one CUJ test. The toggle moved from the Leo settings page to the history sidebar in the latest push, so this is the relevant location to cover.

netzenbot, line 51: nit: `BraveHistoryEmbeddingsToggleHandler` stores `raw_ptr<Profile>` but the only use is `profile_->GetPrefs()->SetBoolean(...)`. Per [ARCH-003](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-003), pass the most specific dependency needed — store `raw_ptr<PrefService>` and accept `PrefService*` (e.g., `profile->GetPrefs()`) in the constructor instead.

chromium_src/chrome/browser/resources/history/side_bar.ts
darkdh: removing the reload mechanism in ead8e2473c064d9087fc33275f654fb9fbeb484c

chromium_src/chrome/browser/ui/webui/history/history_ui.cc
darkdh: it is all done through mojo in ead8e2473c064d9087fc33275f654fb9fbeb484c

chromium_src/chrome/browser/ui/webui/history/history_ui.cc
fallaciousreasoning: I don't think this is right.

browser/ui/webui/history/brave_history_embeddings.mojom
```diff
@@ -0,0 +1,26 @@
...
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+module brave_history_embeddings.mojom;
+
+// chrome://history → browser. Lets the side bar toggle write the
+// Semantic History Search pref and listen for changes.
+interface PageHandlerFactory {
+  // Binds a `Page` for browser→renderer notifications and returns a
```
fallaciousreasoning: mega nit: this doesn't return a `PageHandler`
```suggestion
  // Binds a `Page` for browser→renderer notifications and binds a
```

browser/resources/history/brave_history_embeddings_browser_proxy.ts
```diff
@@ -0,0 +1,34 @@
...
+    factory.createPageHandler(
+      callbackRouter.$.bindNewPipeAndPassRemote(),
+      pageHandler.$.bindNewPipeAndPassReceiver(),
+    )
+    instance = { pageHandler, callbackRouter }
+  }
+  return instance
+}
+
+export { Page, PageCallbackRouter, PageHandlerFactory, PageHandlerRemote }
```
fallaciousreasoning: do we need this export? It seems unused.

browser/extensions/api/settings_private/brave_prefs_util.cc
```diff
@@ -290,6 +291,8 @@ const PrefsUtil::TypedPrefMap& BravePrefsUtil::GetAllowlistedKeys() {
       settings_api::PrefType::kBoolean;
   (*s_brave_allowlist)[ai_chat::prefs::kBraveAIChatTabOrganizationModelKey] =
       settings_api::PrefType::kString;
+  (*s_brave_allowlist)[local_ai::prefs::kBraveHistoryEmbeddingsEnabled] =
+      settings_api::PrefType::kBoolean;
```
fallaciousreasoning: I think this probably isn't needed now we're not exposing a toggle on the settings page.

chromium_src/chrome/browser/resources/history/side_bar.ts
darkdh: updated as you suggested in 607f30a36f6b3909432cd4ba28fd163fed23159d

browser/extensions/api/settings_private/brave_prefs_util.cc
darkdh: forgot to remove it when migrating to history page

browser/ui/webui/history/brave_history_embeddings.mojom
darkdh: reworded in fd5b1ca2e26f383868b5c58da6d7998047d0048c

browser/resources/history/brave_history_embeddings_browser_proxy.ts
darkdh: removed in fd5b1ca2e26f383868b5c58da6d7998047d0048c

browser/extensions/api/settings_private/brave_prefs_util.cc
darkdh: removed in fd5b1ca2e26f383868b5c58da6d7998047d0048c

chromium_src/chrome/browser/ui/webui/history/history_ui.cc
darkdh: fixed in cdcc7870893757498a778095fb0ca56a63ba073e

chromium_src/chrome/browser/resources/history/app.ts
```diff
@@ -15,4 +15,33 @@ injectStyle(HistoryAppElement, css`
     </style>
 `)
 
+// When the Semantic History Search pref flips, propagate the new value into
+// each history-app's reactive `enableHistoryEmbeddings_` accessor so the
+// search box, promo, and history list re-render without a page reload.
+// The event is dispatched from side_bar.ts after receiving the Mojo
+// OnEnabledChanged notification.
+window.addEventListener(
+    'brave-history-embeddings-enabled-changed',
```
fallaciousreasoning: is there any reason we wouldn't just listen to the event on the mojom interface directly here rather dispatching a custom event and listening for that?

As it stands, this will only receive notifications if the sidebar is rendered (which is okay at the moment because that's the only place you can change this setting).

chromium_src/chrome/browser/resources/history/app.ts
```diff
@@ -15,4 +15,33 @@ injectStyle(HistoryAppElement, css`
...
+      for (const selector of liveSelectors) {
+        const root = document.querySelector('history-app')?.shadowRoot
+        if (!root) {
+          continue
+        }
+        for (const el of root.querySelectorAll(selector)) {
+          ;(el as unknown as {requestUpdate?: () => void})
+              .requestUpdate?.()
+        }
+      }
```
fallaciousreasoning: `querySelectorAll` will query a comma separated list of selectors so this can be:
```suggestion
      // history-toolbar / history-list keep their own `compute*` methods that
      // read loadTimeData inline. loadTimeData has already been updated; ask
      // each instance to re-render so the cached values are recomputed.
      const root = document.querySelector('history-app')?.shadowRoot
      if (!root) {
        continue
      }
      for (const el of root.querySelectorAll('history-toolbar, history-list')) {
        ;(el as unknown as {requestUpdate?: () => void})
          .requestUpdate?.()
        }
```

chromium_src/chrome/browser/resources/history/side_bar.ts
```diff
@@ -29,6 +37,74 @@ injectStyle(HistorySideBarElement, css`
...
+  override connectedCallback() {
+    super.connectedCallback()
+    getBraveHistoryEmbeddingsBrowserProxy()
+        .callbackRouter.onEnabledChanged.addListener((enabled: boolean) => {
+          this.braveHistoryEmbeddingsEnabled = enabled
+          loadTimeData.overrideValues({enableHistoryEmbeddings: enabled})
+          window.dispatchEvent(
+              new CustomEvent('brave-history-embeddings-enabled-changed',
+                              {detail: enabled}))
+        })
```
fallaciousreasoning: optional: we could add this `loadTimeData.overrideValues({enableHistoryEmbeddings: enabled})` in `app.ts` and add a `requestUpdate` method to this, so all the bits of UI which need to update use the same mechanism?

fallaciousreasoning (commented):
Looking really good - just some small stuff. Thanks for making all those changes

chromium_src/chrome/browser/ui/webui/history/history_ui.cc
bbondy: Fixed here: https://github.com/brave/brave-core/pull/36956

chromium_src/chrome/browser/resources/history/app.ts
darkdh: It is redundant since we already have the mojo interface.
removed in 4a100eddf2cae0f3f26f89b7d6ce86d83c13910a

chromium_src/chrome/browser/resources/history/app.ts
darkdh: fixed in cec1637b436c382d8a9efe0fbdabb12b890b57c3

chromium_src/chrome/browser/resources/history/side_bar.ts
darkdh: unified it in a1b2fb0370e3c5b77f66c4db9ac144f5b562eb2c

app/brave_settings_strings.grdp
darkdh: @aguscruiz can you check if the above toggle in history page is ok?

aguscruiz: OK yeah this is fine

chromium_src/chrome/browser/resources/history/app.ts
```diff
@@ -15,4 +19,27 @@ injectStyle(HistoryAppElement, css`
...
+// chrome://history UI reacts even when the side bar isn't rendered. Update
+// each history-app's reactive `enableHistoryEmbeddings_` accessor (Lit
+// re-renders the app shell), refresh loadTimeData so any code that reads
+// `enableHistoryEmbeddings` inline picks up the new value, and force a
+// re-render of the toolbar/list whose `compute*_` methods cache the value.
+getBraveHistoryEmbeddingsBrowserProxy()
+    .callbackRouter.onEnabledChanged.addListener((enabled: boolean) => {
+      loadTimeData.overrideValues({enableHistoryEmbeddings: enabled})
+      for (const app of document.querySelectorAll('history-app')) {
+        ;(app as unknown as {enableHistoryEmbeddings_: boolean})
```
fallaciousreasoning, line 32: fwiw, I'm petitioning to allow `as any` in `chromium_src` files as I think `as unknown as whatever` is completely pointless. Not anything you need to change, but hopefully this will get less faffy

fallaciousreasoning (approved):
lgtm, thanks @darkdh

netzenbot (approved):

browser/ui/BUILD.gn
```diff
@@ -418,6 +423,7 @@ source_set("ui") {
       "//brave/components/brave_wayback_machine/buildflags",
       "//brave/components/constants",
       "//brave/components/email_aliases/buildflags",
+      "//brave/components/local_ai/core",
```
bridiver: this should presumably be added as a dep of the new webui/history/BUILD.gn target per https://github.com/brave/brave-core/pull/36624/changes#r3343897324

darkdh: fixed in 8f6b83994670bbc3ea7b094dd769ac2cc9e029b5

fallaciousreasoning (approved):
patches lgtm

bridiver (approved):
code owner changes ok

petemill (approved):

PR #36647, opened 2026-05-22
[Polkadot] Add asset hub westend\polkadot networks under flag.

cypt4:
This PR  adds 2 additional asset hun networks to network_manager that are available under the feature flag along with captured tx data.

Testplan:
To enable assethub networks they should be enabled using 
npm run start -- --enable-logging=stderr --v=0 --enable-features=BraveWalletPolkadot:parachains_enabled/true --polkadot-mainnet-asset-hub-rpc-url=https://polkadot-asset-hub-rpc.polkadot.io  --polkadot-testnet-asset-hub-rpc-url=https://westend-asset-hub-rpc.polkadot.io

1) Test that  asset-hub balance is shown for westend and polkadot networks. Westend networks should be unhidden in brave://settings.
2) Test that asset hub networks are shown only if "parachains_enabled" param mentioned with "true" value.
3) Check asset-hub token transactions between accounts. Note westend asset hub transactions may require several attempts to send.

### DIFF
--- a/components/brave_wallet/browser/brave_wallet_service.cc
+++ b/components/brave_wallet/browser/brave_wallet_service.cc
@@ -1833,7 +1833,7 @@ void BraveWalletService::GenerateReceiveAddress(
   }
 
   if (account_id->coin == mojom::CoinType::DOT) {
-    // TODO(https://github.com/brave/brave-browser/issues/34405): Implement
+    // TODO(https://github.com/brave/brave-browser/issues/55797): Implement
     // custom modal for Polkadot.
     std::move(callback).Run("", WalletInternalErrorMessage());
     return;
--- a/components/brave_wallet/browser/brave_wallet_service.cc
+++ b/components/brave_wallet/browser/brave_wallet_service.cc
@@ -1833,14 +1833,9 @@ void BraveWalletService::GenerateReceiveAddress(
   }
 
   if (account_id->coin == mojom::CoinType::DOT) {
-    if (!polkadot_wallet_service_) {
-      std::move(callback).Run("", WalletInternalErrorMessage());
-      return;
-    }
-
-    const auto chain_id = GetNetworkForPolkadotKeyring(account_id->keyring_id);
-    polkadot_wallet_service_->GetAddress(std::move(account_id), chain_id,
-                                         std::move(callback));
+    // TODO(https://github.com/brave/brave-browser/issues/34405): Implement
+    // custom modal for Polkadot.
+    std::move(callback).Run("", WalletInternalErrorMessage());
     return;
   }
 
--- a/components/brave_wallet/browser/network_manager.cc
+++ b/components/brave_wallet/browser/network_manager.cc
@@ -30,6 +30,7 @@
 #include "brave/components/brave_wallet/browser/pref_names.h"
 #include "brave/components/brave_wallet/common/brave_wallet.mojom.h"
 #include "brave/components/brave_wallet/common/common_utils.h"
+#include "brave/components/brave_wallet/common/features.h"
 #include "brave/components/brave_wallet/common/switches.h"
 #include "brave/components/brave_wallet/common/value_conversion_utils.h"
 #include "components/prefs/pref_service.h"
@@ -86,8 +87,10 @@ constexpr auto kChainSubdomains =
 
             // Polkadot mainnet chains.
             {mojom::kPolkadotMainnet, "polkadot-mainnet"},
+            {mojom::kPolkadotMainnetAssetHub, "polkadot-asset-hub"},
             // Polkadot testnet chains.
             {mojom::kPolkadotTestnet, "polkadot-westend"},
+            {mojom::kPolkadotTestnetAssetHub, "westend-asset-hub"},
         },
         CaseInsensitiveCompare());
 
@@ -598,6 +601,16 @@ GURL PolkadotMainnetRpcUrl() {
   return GetURLForKnownChainId(mojom::kPolkadotMainnet).value();
 }
 
+GURL PolkadotMainnetAssetHubRpcUrl() {
+  auto switch_url =
+      GURL(base::CommandLine::ForCurrentProcess()->GetSwitchValueASCII(
+          switches::kPolkadotMainnetAssetHubRpcUrl));
+  if (switch_url.is_valid()) {
+    return switch_url;
+  }
+  return GetURLForKnownChainId(mojom::kPolkadotMainnetAssetHub).value();
+}
+
 GURL PolkadotTestnetRpcUrl() {
   auto switch_url =
       GURL(base::CommandLine::ForCurrentProcess()->GetSwitchValueASCII(
@@ -606,6 +619,16 @@ GURL PolkadotTestnetRpcUrl() {
     return switch_url;
   }
   return GetURLForKnownChainId(mojom::kPolkadotTestnet).value();
+}
+
+GURL PolkadotTestnetAssetHubRpcUrl() {
+  auto switch_url =
+      GURL(base::CommandLine::ForCurrentProcess()->GetSwitchValueASCII(
+          switches::kPolkadotTestnetAssetHubRpcUrl));
+  if (switch_url.is_valid()) {
+    return switch_url;
+  }
+  return GetURLForKnownChainId(mojom::kPolkadotTestnetAssetHub).value();
 }
 
 const mojom::NetworkInfo* GetBitcoinMainnet() {
@@ -741,6 +764,25 @@ const mojom::NetworkInfo* GetPolkadotMainnet() {
   return network_info.get();
 }
 
+const mojom::NetworkInfo* GetPolkadotMainnetAssetHub() {
+  const auto coin = mojom::CoinType::DOT;
+  const auto* chain_id = mojom::kPolkadotMainnetAssetHub;
+
+  static base::NoDestructor<mojom::NetworkInfo> network_info(
+      {chain_id,
+       "Polkadot Asset Hub",
+       {"https://assethub-polkadot.subscan.io/"},
+       {},
+       0,
+       {PolkadotMainnetAssetHubRpcUrl()},
+       "DOT",
+       "Polkadot",
+       10,
+       coin,
+       GetSupportedKeyringsForKnownNetwork(coin, chain_id)});
+  return network_info.get();
+}
+
 const mojom::NetworkInfo* GetPolkadotTestnet() {
   const auto coin = mojom::CoinType::DOT;
   const auto* chain_id = mojom::kPolkadotTestnet;
@@ -752,6 +794,25 @@ const mojom::NetworkInfo* GetPolkadotTestnet() {
        {},
        0,
        {PolkadotTestnetRpcUrl()},
+       "WND",
+       "Westend",
+       12,
+       coin,
+       GetSupportedKeyringsForKnownNetwork(coin, chain_id)});
+  return network_info.get();
+}
+
+const mojom::NetworkInfo* GetPolkadotTestnetAssetHub() {
+  const auto coin = mojom::CoinType::DOT;
+  const auto* chain_id = mojom::kPolkadotTestnetAssetHub;
+
+  static base::NoDestructor<mojom::NetworkInfo> network_info(
+      {chain_id,
+       "Westend Asset Hub",
+       {"https://assethub-westend.subscan.io/"},
+       {},
+       0,
+       {PolkadotTestnetAssetHubRpcUrl()},
        "WND",
        "Westend",
        12,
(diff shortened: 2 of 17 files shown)


components/brave_wallet/common/common_utils.cc
```diff
@@ -490,11 +490,13 @@ std::vector<mojom::KeyringId> GetSupportedKeyringsForKnownNetwork(
...
     case mojom::CoinType::DOT:
-      if (chain_id == mojom::kPolkadotMainnet) {
+      if (chain_id == mojom::kPolkadotMainnet ||
+          chain_id == mojom::kPolkadotMainnetAssetHub) {
         return {mojom::KeyringId::kPolkadotMainnet,
                 mojom::KeyringId::kPolkadotImport};
       }
-      if (chain_id == mojom::kPolkadotTestnet) {
+      if (chain_id == mojom::kPolkadotTestnet ||
+          chain_id == mojom::kPolkadotTestnetAssetHub) {
```
cmazakas, line 486: I like these changes a lot because they reflect better how other wallets like Talisman operate, where a single seed phrase will be used on N parachains.

There's a caveat here that we need to consider.

Our deposit screen is broken for this.

```cpp
    const auto chain_id = GetNetworkForPolkadotKeyring(account_id->keyring_id);
    polkadot_wallet_service_->GetAddress(std::move(account_id), chain_id,
                                         std::move(callback));
```

because:
```cpp
std::string GetNetworkForPolkadotKeyring(const mojom::KeyringId& keyring_id) {
  if (keyring_id == mojom::KeyringId::kPolkadotMainnet ||
      keyring_id == mojom::KeyringId::kPolkadotImport) {
    return mojom::kPolkadotMainnet;
  }
  if (keyring_id == mojom::KeyringId::kPolkadotTestnet ||
      keyring_id == mojom::KeyringId::kPolkadotImportTestnet) {
    return mojom::kPolkadotTestnet;
  }
  NOTREACHED();
}
```

The problem here is that we're going to pass an incorrect `chain_id` to `polkadot_wallet_service_->GetAddress`. This is because Westend Asset Hub uses an ss58-prefix of 42 but Passet Hub (the nickname for Paseo Asset Hub) uses an ss58-prefixof 0.

It seems like we have the chain_id readily available for these calls:
```ts
  const { receiveAddress, isFetchingAddress } = useReceiveAddressQuery(
    selectedAccount?.accountId,
  )
```

I'm thinking, maybe we update the mojom endpoint to take a `string? chain_id`. I think it's better to just add a nullable chain_id parameter to the main entry point here than to bifurcate the API across all call-sites.

But for this, I'm curious what @supermassive and @Douglashdaniel think about that.

components/brave_wallet/browser/network_manager.cc
```diff
@@ -741,6 +764,25 @@ const mojom::NetworkInfo* GetPolkadotMainnet() {
...
+
+  static base::NoDestructor<mojom::NetworkInfo> network_info(
+      {chain_id,
+       "Polkadot Asset Hub",
+       {"https://assethub-polkadot.subscan.io/"},
+       {},
+       0,
+       {PolkadotMainnetAssetHubRpcUrl()},
+       "DOT",
+       "Polkadot",
```
cmazakas, line 779: I'd recommend updating this to say `"Polkadot Asset Hub"` because in the UI, it's kind of confusing to see two assets named "Polkadot" that are on different chains.

components/brave_wallet/browser/network_manager.cc
```diff
@@ -760,6 +802,25 @@ const mojom::NetworkInfo* GetPolkadotTestnet() {
...
+
+  static base::NoDestructor<mojom::NetworkInfo> network_info(
+      {chain_id,
+       "Westend Asset Hub",
+       {"https://assethub-westend.subscan.io/"},
+       {},
+       0,
+       {PolkadotTestnetAssetHubRpcUrl()},
+       "WND",
+       "Westend",
```
cmazakas, line 817: Same as: https://github.com/brave/brave-core/pull/36647/changes#r3291486957

When this is `"Westend Asset Hub"`, it becomes a lot more clear which asset belongs to which parachain.

netzenbot (approved):

components/brave_wallet/browser/network_manager.cc
cypt4: This is actually fine : 
image

components/brave_wallet/common/common_utils.cc
cypt4: Ok, i have removed this methods. and logged an issue to implement another deposit screen for polkadot 
https://github.com/brave/brave-browser/issues/55797

components/brave_wallet/browser/network_manager.cc
cypt4: This field is for token name, not network name

components/brave_wallet/common/common_utils.cc
cmazakas: Hmm, removing:
```cpp
  if (account_id->coin == mojom::CoinType::DOT) {
    if (!polkadot_wallet_service_) {
      std::move(callback).Run("", WalletInternalErrorMessage());
      return;
    }

    const auto chain_id = GetNetworkForPolkadotKeyring(account_id->keyring_id);
    polkadot_wallet_service_->GetAddress(std::move(account_id), chain_id,
                                         std::move(callback));
    return;
  }
```
seems like it breaks the deposit screen entirely, which would be a blocker for 1.92's relay chain release.

We should probably just add it back and keep the TODO.

cmazakas: Fwiw, right now the code in master is "broken" but it also works really well too. The Westend testnets seem to use the prefix of 42 and the production mainnets seem to use 0, which our code handles.

What we don't handle is something like Paseo Asset Hub using a ss58 prefix of 0. But I don't think this should be a blocker if the wallet actually works for users and lets them use their Asset Hub funds.

We should able to keep this code as it currently is and then gradually migrate it to the new UI without breaking what's in master right now.

components/brave_wallet/browser/network_manager.cc
cmazakas: Ah, yeah.

I was thinking of this screen:

image

It's kinda hard to know what I'm sending here.

components/brave_wallet/browser/brave_wallet_service.cc
```diff
@@ -1831,14 +1831,9 @@ void BraveWalletService::GenerateReceiveAddress(
...
   if (account_id->coin == mojom::CoinType::DOT) {
-    if (!polkadot_wallet_service_) {
-      std::move(callback).Run("", WalletInternalErrorMessage());
-      return;
-    }
-
-    const auto chain_id = GetNetworkForPolkadotKeyring(account_id->keyring_id);
-    polkadot_wallet_service_->GetAddress(std::move(account_id), chain_id,
-                                         std::move(callback));
+    // TODO(https://github.com/brave/brave-browser/issues/34405): Implement
```
cmazakas: This TODO should be: https://github.com/brave/brave-browser/issues/55797

components/brave_wallet/browser/network_manager.cc
cypt4: It should work similar to erc 20 tokens - token name is the same but we have network badge

cmazakas:
It seems like this patch is enough to fix the deposit-funds screen.

This is actually a pretty small diff, and I tested it locally by synthetically having Westend Asset Hub use a prefix 0, instead of 42. Worked great! You should test it locally, of course. But this gives us really nice parachain support in a proper way.

```diff
diff --git a/components/brave_wallet_ui/page/screens/fund-wallet/deposit-funds.tsx b/components/brave_wallet_ui/page/screens/fund-wallet/deposit-funds.tsx
index a6ab0576336..d65ac23fd7d 100644
--- a/components/brave_wallet_ui/page/screens/fund-wallet/deposit-funds.tsx
+++ b/components/brave_wallet_ui/page/screens/fund-wallet/deposit-funds.tsx
@@ -50,6 +50,7 @@ import { AllNetworksOption } from '../../../options/network-filter-options'
 import { useCopyToClipboard } from '../../../common/hooks/use-copy-to-clipboard'
 import {
   useGetNetworkQuery,
+  useGetPolkadotAddressForNetworkQuery,
   useGetQrCodeImageQuery,
   useGetVisibleNetworksQuery,
   useGetZCashAccountInfoQuery,
@@ -538,8 +539,21 @@ function DepositAccount() {
   const [selectedAccount, setSelectedAccount] = React.useState<
     BraveWallet.AccountInfo | undefined
   >(accountsForSelectedAssetCoinType[0])
+  const isPolkadotAccount =
+    selectedAccount?.accountId.coin === BraveWallet.CoinType.DOT
   const { receiveAddress, isFetchingAddress } = useReceiveAddressQuery(
-    selectedAccount?.accountId,
+    isPolkadotAccount ? undefined : selectedAccount?.accountId,
+  )
+  const {
+    currentData: polkadotAddress,
+    isFetching: isFetchingPolkadotAddress,
+  } = useGetPolkadotAddressForNetworkQuery(
+    isPolkadotAccount && selectedAccount && selectedAssetNetwork
+      ? {
+          accountId: selectedAccount.accountId,
+          chainId: selectedAssetNetwork.chainId,
+        }
+      : skipToken,
   )
   const [selectedZCashAddressOption, setSelectedZCashAddressOption] =
     React.useState<string>('shielded')
@@ -590,6 +604,9 @@ function DepositAccount() {
   }, [selectedAsset])
 
   const address = React.useMemo(() => {
+    if (isPolkadotAccount) {
+      return polkadotAddress || ''
+    }
     if (
       isZCashShieldedTransactionsEnabled
       && selectedAccount?.accountId.coin === BraveWallet.CoinType.ZEC
@@ -606,6 +623,8 @@ function DepositAccount() {
     }
     return receiveAddress
   }, [
+    isPolkadotAccount,
+    polkadotAddress,
     isZCashShieldedTransactionsEnabled,
     selectedAccount,
     receiveAddress,
@@ -613,6 +632,10 @@ function DepositAccount() {
     selectedZCashAddressOption,
   ])
 
+  const isLoadingAddress = isPolkadotAccount
+    ? isFetchingPolkadotAddress
+    : isFetchingAddress
+
   const { data: qrCode, isFetching: isLoadingQrCode } = useGetQrCodeImageQuery(
     address || skipToken,
   )
@@ -768,7 +791,7 @@ function DepositAccount() {
           {':'}
         </AddressTextLabel>
 
-        {address && !isFetchingAddress ? (
+        {address && !isLoadingAddress ? (
           <>
             <Row gap={'12px'}>
               <AddressText
@@ -795,7 +818,7 @@ function DepositAccount() {
 
         <Row>
           <QRCodeContainer>
-            {isLoadingQrCode || !address || isFetchingAddress ? (
+            {isLoadingQrCode || !address || isLoadingAddress ? (
               <LoadingRing />
             ) : (
               <QRCodeImage src={qrCode} />
```

components/brave_wallet/browser/network_manager.cc
cmazakas: Sounds fine to me then. Marking as resolved.

components/brave_wallet/common/common_utils.cc
cmazakas: Resolving because of: https://github.com/brave/brave-core/pull/36758

components/brave_wallet/browser/polkadot/polkadot_wallet_service_unittest.cc
```diff
@@ -2090,6 +2105,425 @@ TEST_F(PolkadotWalletServiceUnitTest, SignAndSendTransaction) {
             kExpectedExtrinsic);
 }
 
+TEST_F(PolkadotWalletServiceUnitTest, SignAndSendTransaction_WestendAssetHub) {
```
cmazakas, line 2108: This test is good, but I feel like we could probably clean it up by extending the PolkadotMockRpc service.

The test lazily sets the interceptor when `FinalizeSetup()` is called, so it should in theory work. We have some plumbing already set up to handle mapping block requests. If we need to extend it, that's fine too.

components/brave_wallet/browser/polkadot/polkadot_wallet_service_unittest.cc
```diff
@@ -2090,6 +2105,425 @@ TEST_F(PolkadotWalletServiceUnitTest, SignAndSendTransaction) {
...
+            kExpectedExtrinsic);
+  EXPECT_EQ(base::HexEncodeLower(tx_hash->second.block_hash()),
+            "a3a6ef09932ad0a6086e0900431d09a9488e3e905e277cf7bb5a8ed8bfa90470");
+  EXPECT_EQ(tx_hash->second.block_num(), 0xe812ebu);
+  EXPECT_EQ(tx_hash->second.mortality_period(), 64u);
+}
+
+TEST_F(PolkadotWalletServiceUnitTest, SignAndSendTransaction_PolkadotAssetHub) {
+  // Captured transaction:
+  // https://assethub-polkadot.subscan.io/extrinsic/0x32cf17114363bc9b26256104cab96d19a1052b4407452e05709fb6876d64daff
```
cmazakas, line 2181: Btw, thank you for including a real mainnet transaction in this.

cmazakas (approved):
LGTM!

We shouldn't merge this, however, until https://github.com/brave/brave-core/pull/36758 is merged first.

cmazakas (approved):
Man, those test changes look great. Thank you for that!

components/brave_wallet/browser/polkadot/polkadot_test_utils.cc
```diff
@@ -709,12 +971,16 @@ bool PolkadotMockRpc::HandleAuthorSubmitExtrinsic(
...
-          })");
+        const std::string_view extrinsic_hash =
+            submitted_extrinsic_hash_.empty() ? kDefaultSubmittedExtrinsicHash
+                                              : submitted_extrinsic_hash_;
+        const std::string response = R"({
+          "jsonrpc": "2.0",
+          "id": 1,
+          "result": ")" + extrinsic_hash +
+                                     R"("
+        })";
```
cmazakas: For cases like this, I think it's better to just use `absl::StrFormat`.

PR #36667, opened 2026-05-22
[toolchain] Accept Xcode license regardless of toolchain download

cdesouza-chromium:
This PR corrects corrects a problem where our
`download_hermetic_xcode.py` script was diverging from the upstream one
by early returning whenever it confirmed that the toolchain had already
been downloaded and not trying to accept the license. This has been
causing issues in CI, where nodes being used for different branches,
with different SDKs, are not properly going over the necessary plist
updates that they should have done.

This PR corrects this aspect of the behaviour, but it also makes this
particular script more alinged with the upstream script itself.


emerick (approved):

cdesouza-chromium:
Force merging, as this is a Apple/Linux script only and it has finished for both platforms, and this is right now causing CI issues.

PR #36669, opened 2026-05-23
Added UAE, Nigeria flags to VPN panel

### LINKED ISSUE #55764: Add UAE, Nigeria flags to VPN panel
Set qa/no as vpn service doesn't release uae/nigeria yet.

simonhong:
Resolves https://github.com/brave/brave-browser/issues/55764

### DIFF
--- a/ui/webui/resources/BUILD.gn
+++ b/ui/webui/resources/BUILD.gn
@@ -497,6 +497,7 @@ leo_icons_sources = [
 
 if (enable_brave_vpn_panel) {
   leo_icons_sources += [
+    "country-ae.svg",
     "country-al.svg",
     "country-ar.svg",
     "country-at.svg",
@@ -532,6 +533,7 @@ if (enable_brave_vpn_panel) {
     "country-lu.svg",
     "country-mx.svg",
     "country-my.svg",
+    "country-ng.svg",
     "country-nl.svg",
     "country-no.svg",
     "country-nz.svg",
--- a/components/brave_vpn/resources/panel/components/flag/index.tsx
+++ b/components/brave_vpn/resources/panel/components/flag/index.tsx
@@ -18,6 +18,7 @@ interface Props {
 // when adding a new entry here, please also add SVG for the entry into
 // `brave/ui/webui/resources/BUILD.gn` (under `enable_brave_vpn_panel`)
 const COUNTRIES = [
+    'AE',
     'AL',
     'AR',
     'AT',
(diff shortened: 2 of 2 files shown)


ui/webui/resources/BUILD.gn
```diff
@@ -532,6 +533,7 @@ if (enable_brave_vpn_panel) {
     "country-lu.svg",
     "country-mx.svg",
     "country-my.svg",
+    "country-ne.svg",
```
netzenbot: Wrong filename: `country-ne.svg` is Niger (NE), not Nigeria (NG). The `Flag` component constructs `country-ng` from the 'NG' country code via `toLocaleLowerCase()`, so this should be `country-ng.svg`. Nigeria's flag icon will silently not render with this mismatch.

ui/webui/resources/BUILD.gn
```diff
@@ -532,6 +533,7 @@ if (enable_brave_vpn_panel) {
     "country-lu.svg",
     "country-mx.svg",
```
netzenbot, line 534: The country code added to `index.tsx` is `'NG'` (Nigeria), but the SVG registered here is `country-ne.svg` — `NE` is Niger, not Nigeria. The flag component generates the icon name as `country-ng`, so the Nigeria flag will silently fail to render. This should be `"country-ng.svg"`.

ui/webui/resources/BUILD.gn
```diff
@@ -532,6 +533,7 @@ if (enable_brave_vpn_panel) {
...
     "country-my.svg",
```
netzenbot, line 535: The SVG filename is wrong here. `country-ne.svg` is Niger (ISO code `NE`), but Nigeria's ISO code is `NG`. The `Flag` component builds the icon name as `country-${countryCode.toLowerCase()}`, so adding `'NG'` to the TypeScript array will look for `country-ng.svg`. This entry should be `"country-ng.svg"` to match.

ui/webui/resources/BUILD.gn
```diff
@@ -532,6 +533,7 @@ if (enable_brave_vpn_panel) {
...
+    "country-ne.svg",
     "country-nl.svg",
     "country-no.svg",
     "country-nz.svg",
```
netzenbot, line 539: Bug: `country-ne.svg` is the flag for Niger (ISO 3166-1: NE), not Nigeria (NG). Since `index.tsx` adds `'NG'` and the flag component resolves to `country-ng` at runtime, this entry should be `"country-ng.svg"` — otherwise Nigeria's flag icon will silently fail to load.

fallaciousreasoning (approved):

bsclifton (approved):
LGTM!

PR #36674, opened 2026-05-23
[Sidebar V2] Fix side panel's resize area is not visible

simonhong:

F/U PR for https://github.com/brave/brave-core/pull/36472.
From https://github.com/brave/brave-core/pull/36472, we set side panel header.
So far,  `BraveSidePanelResizeArea` checks panel's inset emptiness to determine its position.
If empty, it's set over the panel contents. Otherwise, its set over the panel's side border area.
After header is used, panel has non-empty insets because it uses top inset to reserve header.
So, checking panel's inset emptiness is not valid anymore. It should check inset's width instead.

Also, resize area should be top-most child view. If not, can't grp around the header because
header is added later than resize area. So, reordered whenever header is added.

TEST=SidebarBrowserTest.SidebarV2ResizeAreaPositionMatchesBorderState

### DIFF
--- a/browser/ui/sidebar/sidebar_browsertest.cc
+++ b/browser/ui/sidebar/sidebar_browsertest.cc
@@ -2194,44 +2194,61 @@ IN_PROC_BROWSER_TEST_F(SidebarBrowserTest,
   // --- Border case (rounded corners ON) ---
   prefs->SetBoolean(kWebViewRoundedCorners, true);
   RunScheduledLayouts();
-  EXPECT_FALSE(side_panel->GetInsets().IsEmpty());
+  EXPECT_GT(side_panel->GetInsets().width(), 0)
+      << "panel should have left/right insets";
 
   // In the bordered case the resize strip sits in the gap between the panel
   // edge and the content, at the left edge (panel is on the right in LTR).
-  const gfx::Rect bordered_bounds = resize_area->bounds();
-  EXPECT_EQ(bordered_bounds.top_right(),
-            side_panel->GetContentParentView()->origin())
+  // Check resize area and panel's screen bounds.origin().
+  EXPECT_EQ(resize_area->GetBoundsInScreen().origin(),
+            side_panel->GetBoundsInScreen().origin())
       << "Bordered resize area right edge should align with content origin";
-  EXPECT_EQ(bordered_bounds.width(), side_panel->GetInsets().left())
+  EXPECT_EQ(resize_area->width(), side_panel->GetInsets().left())
       << "Bordered resize area width should match the left border inset";
+  EXPECT_EQ(side_panel->GetIndexOf(resize_area),
+            side_panel->children().size() - 1)
+      << "resize area should be the top-most view";
 
   // --- No-border case (rounded corners OFF) ---
   prefs->SetBoolean(kWebViewRoundedCorners, false);
   RunScheduledLayouts();
-  EXPECT_TRUE(side_panel->GetInsets().IsEmpty());
+  EXPECT_EQ(0, side_panel->GetInsets().width());
+
+  // Shared assertions for the no-border state after a panel switch.
+  auto verify_no_border_state = [&]() {
+    EXPECT_EQ(0, side_panel->GetInsets().width())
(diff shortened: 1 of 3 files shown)


browser/ui/views/side_panel/brave_side_panel_resize_area.cc
```diff
@@ -28,7 +28,11 @@ gfx::Rect BraveSidePanelResizeArea::GetNoBorderResizeBounds(
 }
 
 void BraveSidePanelResizeArea::Layout(PassKey) {
-  if (!side_panel_->GetInsets().IsEmpty()) {
+  // Upstream puts resize area over the left or right border.
+  // If it doesn't have left or right border, we should put over the contents.
+  // Don't check inset's emptiness because it can only have top insets due to
+  // panel header.
+  if (side_panel_->GetInsets().width() != 0) {
```
netzenbot, line 36: nit: "Don't check inset's emptiness" reads as an instruction referencing the removed `IsEmpty()` approach. Reword to describe the current logic, e.g. "Check only horizontal insets — top insets may be non-zero when a panel header is present, so IsEmpty() would incorrectly treat a header-only inset as a side border." ([✅ Comments Must Make Sense to Future Readers of the Codebase](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-019))

simonhong: Done.

browser/ui/sidebar/sidebar_browsertest.cc
```diff
@@ -2194,44 +2194,61 @@ IN_PROC_BROWSER_TEST_F(SidebarBrowserTest,
...
+            side_panel->children().size() - 1)
+      << "resize area should be the top-most view";
 
   // --- No-border case (rounded corners OFF) ---
   prefs->SetBoolean(kWebViewRoundedCorners, false);
   RunScheduledLayouts();
-  EXPECT_TRUE(side_panel->GetInsets().IsEmpty());
+  EXPECT_EQ(0, side_panel->GetInsets().width());
+
+  // Shared assertions for the no-border state after a panel switch.
```
netzenbot, line 2217: nit: `verify_no_border_state` is called from two places. Add `SCOPED_TRACE(base::Location::Current())` at the top of the lambda so that assertion failures identify which call site failed, not just the lambda body. ([TI-015](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-015))

chromium_src/chrome/browser/ui/views/side_panel/side_panel.cc
```diff
@@ -69,6 +69,12 @@ void SidePanel::UpdateBorder() {
 void SidePanel::AddHeaderView(std::unique_ptr<views::View> view) {
   AddHeaderView_ChromiumImpl(std::move(view));
   UpdateBorder();
+
+  // Resize area is overlapped with other views such as contents and
+  // header view. Make it top-most view to get event for dragging.
+  if (!rounded_border_enabled_) {
```
sangwoo108, line 76: I'm wondering why we reorder the resize_area only when we don't use rounded border. Could you add some comments about it?

browser/ui/sidebar/sidebar_browsertest.cc
sangwoo108: +1

sangwoo108 (approved):

browser/ui/sidebar/sidebar_browsertest.cc
simonhong: Updated.

chromium_src/chrome/browser/ui/views/side_panel/side_panel.cc
simonhong: Comments updated.
```
  // The resize area overlaps other views (contents, header), so it must be
  // top-most to receive drag events. With rounded corners, it sits over the
  // border area instead and doesn't overlap other views, so no reorder needed.
```

goodov (approved):

PR #36683, opened 2026-05-25
Restore pinned tabs with containers.

### LINKED ISSUE #55598: [Containers] Pinned tabs lose container ID when session restore is disabled
When "Continue where you left off" is NOT selected, pinned contained tabs restore as normal tabs (without containers).

This is because pinned tabs are created via `StartupTabs` mechanism (bypassing full session restore) in this case.

This needs to be fixed.

goodov:
Add support for Containers to `StartupTab`. This allows us to restore pinned tabs with the right container and support container passing via a command line: `brave --container=Work https://brave.com`.

Resolves https://github.com/brave/brave-browser/issues/55598

### DIFF
--- a/browser/containers/container_specifier_utils.cc
+++ b/browser/containers/container_specifier_utils.cc
@@ -1,0 +1,85 @@
+// Copyright (c) 2026 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#include "brave/browser/containers/container_specifier_utils.h"
+
+#include "base/logging.h"
+#include "brave/browser/containers/containers_service_factory.h"
+#include "brave/components/containers/content/browser/storage_partition_utils.h"
+#include "brave/components/containers/core/browser/containers_service.h"
+#include "brave/components/containers/core/common/features.h"
+#include "chrome/browser/profiles/profile.h"
+#include "chrome/browser/ui/navigator/browser_navigator_params.h"
+#include "content/public/browser/browser_context.h"
+#include "content/public/browser/storage_partition_config.h"
+
+namespace containers {
+namespace {
+
+constexpr char kContainerIdKey[] = "container_id";
+
+}  // namespace
+
+void SerializeContainerSpecifier(content::WebContents* web_contents,
+                                 base::Value& value) {
+  if (!base::FeatureList::IsEnabled(features::kContainers)) {
+    return;
+  }
+
+  CHECK(value.is_dict());
+  if (auto container_id = GetContainerIdForWebContents(web_contents);
+      !container_id.empty()) {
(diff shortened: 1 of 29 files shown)


browser/containers/container_specifier_utils.cc
```diff
@@ -0,0 +1,85 @@
...
+}  // namespace
+
+void SerializeContainerSpecifier(content::WebContents* web_contents,
+                                 base::Value& value) {
+  if (!base::FeatureList::IsEnabled(features::kContainers)) {
+    return;
+  }
+
+  CHECK(value.is_dict());
+  if (auto container_id = GetContainerIdForWebContents(web_contents);
```
cdesouza-chromium: nit: I think we should follow the guidelines here and avoid the use of `auto`.

https://google.github.io/styleguide/cppguide.html#Type_deduction

rewrite/chrome/browser/ui/startup/startup_tab.h.yaml
```diff
@@ -0,0 +1,32 @@
...
+      #include "url/gurl.h"
+    replace: |-
+      #include "url/gurl.h"
+      // Containers headers begin
+      #include "brave/components/containers/buildflags/buildflags.h"
+
+      #if BUILDFLAG(ENABLE_CONTAINERS)
+      #include "brave/components/containers/core/browser/container_specifier.h"
+      #endif
+      // Containers headers end
```
cdesouza-chromium: For this one, just use the `chromium_src` header.

cdesouza-chromium (approved):
LGTM

chromium_src/chrome/browser/sessions/session_restore.cc
```diff
@@ -0,0 +1,32 @@
...
+#include "chrome/browser/ui/startup/startup_tab.h"
+
+#if BUILDFLAG(ENABLE_CONTAINERS)
+#include "brave/browser/containers/container_specifier_utils.h"
+#endif  // BUILDFLAG(ENABLE_CONTAINERS)
+
+namespace {
+
+void BraveModifyStartupTabNavigationParams(const StartupTab& tab,
+                                           Browser* browser,
```
cdesouza-chromium: Do you think we can use `BrowserWindowInterface*` here?

chromium_src/chrome/browser/ui/startup/startup_browser_creator_impl.cc
```diff
@@ -3,10 +3,36 @@
...
 #include "chrome/browser/ui/startup/startup_tab_provider.h"
 
+#if BUILDFLAG(ENABLE_CONTAINERS)
+#include "brave/browser/containers/container_specifier_utils.h"
+#endif  // BUILDFLAG(ENABLE_CONTAINERS)
+
+namespace {
+
+void BraveModifyStartupTabNavigationParams(const StartupTab& tab,
+                                           Browser* browser,
```
cdesouza-chromium: Same.

netzenbot (approved):

PR #36691, opened 2026-05-25
Fix search query concatenation with + symbols instead of spaces

### LINKED ISSUE #55802: Fix search query concatenation with + symbols instead of spaces
When using the search widget, the query is concatenated using `+` symbols instead of spaces; so `hello brave` becomes `hello+brave`.

Image

Image

X link: https://x.com/tschweren/status/2058507980895609126?s=46&t=c88ZoLNNPThnGggE75IoLQ

simoarpe:
## Resolves https://github.com/brave/brave-browser/issues/55802

Fix widget initiated Brave searches whose queries contain literal + characters (e.g. C++ tutorial) being corrupted by `maybeReplaceWidgetSearchSource` in `BraveIntentHandler`.

When a search is launched from the Search widget, `maybeReplaceWidgetSearchSource` rewrites the source query parameter from `android` to `android-widget` so the backend can attribute the request correctly. The previous implementation did this by walking every parameter through `Uri.getQueryParameters()` and emitting it again via `Uri.Builder.appendQueryParameter()`.

That decoding and encoding cycle is lossy for search URLs, because the two APIs use different encodings:

Search URLs use `application/x-www-form-urlencoded`, where ` ` is `+` and a literal `+` is `%2B`.
Android's Uri follows RFC 3986, where ` ` is `%20` and `+` is a literal.
`getQueryParameters()` decodes `%2B` to `+` but leaves a form encoded `+` (meaning space) alone. After that step a literal `+` and an encoded space are indistinguishable, and encoding the value again with `appendQueryParameter()` (which encodes `+` as `%2B`) corrupts the query.

### Concrete example

A widget search for `C++ tutorial` arrives as: `?q=C%2B%2B+tutorial&source=android`
`getQueryParameters("q")` returns the string `C+++tutorial`. 

Appending that again produces: `?q=C%2B%2B%2Btutorial&source=android-widget`
Which the server interprets as the literal string `C+++tutorial`.

### Fix
Operate on the raw encoded query via `Uri.getEncodedQuery()`, split on `&`, and rewrite only the `source=android` pair. Every other parameter, including `q`, passes through byte for byte, so whatever encoding the widget produced reaches the server intact. The substituted query is written back via `Uri.Builder.encodedQuery()` and not `query()` to prevent encoding it again.

Extensive inline comments document the encoding mismatch and explain why `Uri.getQueryParameters()` and `appendQueryParameter()` are not used.

### Core regression case (the bug being fixed)

- Type `C++ tutorial` into the widget and submit.
- Confirm the rendered results page is for `C++ tutorial`, not `C tutorial` and not `C+++tutorial`.
- Confirm source=android-widget is present.

### Plain query (no special chars)
- Type `hello world` into the widget and submit.
- Confirm the rendered results page is for `hello world`.
- Confirm source=android-widget is present.

### DIFF
--- a/android/java/org/chromium/chrome/browser/BraveIntentHandler.java
+++ b/android/java/org/chromium/chrome/browser/BraveIntentHandler.java
@@ -86,15 +86,67 @@ public class BraveIntentHandler {
             return url;
         }
 
-        final Uri.Builder builder = parsedUrl.buildUpon().clearQuery();
-        for (String name : parsedUrl.getQueryParameterNames()) {
-            for (String value : parsedUrl.getQueryParameters(name)) {
-                builder.appendQueryParameter(
-                        name,
-                        SOURCE.equals(name) && ANDROID.equals(value) ? ANDROID_WIDGET : value);
+        // Rewrite only the `source` parameter on the raw encoded query, so
+        // every other parameter's bytes pass through completely untouched.
+        //
+        // We can't safely decode and re-encode the query through
+        // Uri.getQueryParameters() / appendQueryParameter(), because the two
+        // sides use different encodings:
+        //
+        //   - Search URLs use application/x-www-form-urlencoded, where ' '
+        //     is encoded as '+' and a literal '+' as '%2B'.
+        //   - Android's Uri follows RFC 3986, where ' ' is '%20' and '+' is
+        //     a literal.
+        //
+        // getQueryParameters() percent-decodes '%2B' to '+' but leaves a
+        // form-encoded '+' (meaning space) alone. After that step, a
+        // literal '+' and an encoded space are indistinguishable, so any
+        // attempt to re-encode the value corrupts the query.
+        //
+        // For example, a widget search for "C++ tutorial" arrives as
+        // "q=C%2B%2B+tutorial". getQueryParameters() would return "C+++tutorial"
+        // (three '+' chars, origin unknown). If those were then re-encoded
+        // and the '+' chars converted back to spaces, the server would
+        // receive "C   tutorial"; if they were left as '+', the server
+        // would receive a literal "C+++tutorial". Either way the query is
+        // broken.
+        final String encodedQuery = parsedUrl.getEncodedQuery();
+        if (encodedQuery == null) {
+            return url;
+        }
+
+        final StringBuilder newQuery = new StringBuilder(encodedQuery.length());
+        // Walk each name=value pair in encoded form. We intentionally do not
+        // use Uri.getQueryParameters() here. See the comment above for why
+        // decoding and re-encoding through it is lossy.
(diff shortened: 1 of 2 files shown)


netzenbot (approved):

android/java/org/chromium/chrome/browser/BraveIntentHandler.java
```diff
@@ -86,15 +86,67 @@ private static String maybeReplaceWidgetSearchSource(final Intent intent, final
...
+            } else {
+                // Any other parameter including the search query `q`
+                // passes through untouched, byte-for-byte.
+                newQuery.append(pair);
             }
         }
-        return builder.build().toString();
+        // encodedQuery() tells Uri.Builder our string is
+        // already encoded and must not be re-encoded.
+        return parsedUrl.buildUpon().encodedQuery(newQuery.toString()).build().toString();
```
SergeyZhukovsky, line 149: could you extend unit tests inside BraveIntentHandlerUnitTest.java to cover + and special characters?

simoarpe: Yes, good idea 👉 [`9e4607c` (this PR)](https://github.com/brave/brave-core/pull/36691/commits/9e4607cdc3b5d1e6616d4c9d59b3821cab644963)

SergeyZhukovsky (approved):

kjozwiak:
Verification **PASSED** on `Pixel 8 Pro` using the following build(s):

```
Brave | 1.92.102 Chromium: 149.0.7827.23 (Official Build) nightly (64-bit) 
--- | ---
Revision | e0b03fdfccb8de142e0a697ae5a8f36ab1e5b950
OS | Android 16; Build/CP1A.260505.005; 36; REL
```

Using the STR/Cases outlined via https://github.com/brave/brave-core/pull/36691#issue-4518152107, ensured that search queries with spaces are not being concatenated with `+` as per the following:

### `Test Case #1` - Searching `C++ tutorial`

* ensured that searching for `C++ tutorial` via the widget searches for `C++ tutorial` within Brave
* ensured that `source=android-widget` is being added into the URL when doing searches from the widget

```
https://search.brave.com/search?q=C%2B%2B+tutorial&source=android-widget
```

**`Example`** | **`Example`**
---------------|----------------
Screenshot_20260526-215335 | Screenshot_20260526-215344

### `Test Case #2` - Searching `hello world`

* ensured that searching for `hello world` via the widget searches for `hello world` and NOT `hello+world` within Brave
* ensured that `source=android-widget` is being added into the URL when doing searches from the widget

```
https://search.brave.com/search?q=hello+world&source=android-widget
```

**`Example`** | **`Example`**
---------------|----------------
Screenshot_20260526-215442 | Screenshot_20260526-215447

PR #36696, opened 2026-05-25
[iOS] Refactor ReaderModeTabHelper as single source of truth for reader mode

nuo-xu:
Move all reader mode logic — readability checking, toggle/enable/disable, style updates, and P3A recording into ReaderModeTabHelper, removing the scattered flag-branching between kUseProfileWebViewConfiguration paths from BrowserViewController.
The helper now:
- Conforms to TabObserver directly, triggering readability checks on navigation events for both the JSFeature and legacy content script paths
- Conforms to ReaderModeScriptHandlerDelegate to handle legacy readability results and state change
-  Installs ReaderModeScriptHandler as a content script for the legacy path
-  Accepts ReaderModeCache as an init dependency to ensure cache writes complete before reader mode navigation is triggered
- Exposes onStateChanged, onReaderModeDisplayed and updateTranslateURLBar closures for BVC UI callbacks
- Exposes a static recordTimeBasedNumberReaderModeUsedP3A replacing the BVC instance method

Resolves

Test:
Check reader mode flow is working as before with and without `kUseProfileWebViewConfiguration` enabled.

### DIFF
--- a/ios/brave-ios/Sources/Brave/Frontend/Browser/BrowserViewController/BVC+P3A.swift
+++ b/ios/brave-ios/Sources/Brave/Frontend/Browser/BrowserViewController/BVC+P3A.swift
@@ -435,9 +435,6 @@ extension P3ATimedStorage where Value == Int {
   fileprivate static var braveVPNDaysInMonthUsedStorage: Self {
     .init(name: "vpn-days-in-month-used", lifetimeInDays: 30)
   }
-  internal static var readerModeActivated: Self {
-    .init(name: "reader-mode-activated", lifetimeInDays: 7)
-  }
   fileprivate static var navigationActionPerformedStorage: Self {
     .init(name: "navigation-action-performed", lifetimeInDays: 7)
   }
--- a/ios/brave-ios/Sources/Brave/Frontend/Browser/BrowserViewController/BVC+P3A.swift
+++ b/ios/brave-ios/Sources/Brave/Frontend/Browser/BrowserViewController/BVC+P3A.swift
@@ -360,26 +360,6 @@ extension BrowserViewController {
     UmaHistogramEnumeration("Brave.General.BottomBarLocation", sample: answer)
   }
 
-  func recordTimeBasedNumberReaderModeUsedP3A(activated: Bool) {
-    var storage = P3ATimedStorage<Int>.readerModeActivated
-    if activated {
-      storage.add(value: 1, to: Date())
-    }
-
-    // Q102- How many times did you use reader mode in the last 7 days?
-    UmaHistogramRecordValueToBucket(
-      "Brave.ReaderMode.NumberReaderModeActivated",
-      buckets: [
-        0,
-        .r(1...5),
-        .r(5...20),
-        .r(20...50),
-        .r(51...),
-      ],
-      value: storage.combinedValue
-    )
-  }
-
   func recordAdsUsageType() {
     enum Answer: Int, CaseIterable {
       case none = 0
@@ -455,7 +435,7 @@ extension P3ATimedStorage where Value == Int {
   fileprivate static var braveVPNDaysInMonthUsedStorage: Self {
     .init(name: "vpn-days-in-month-used", lifetimeInDays: 30)
   }
-  fileprivate static var readerModeActivated: Self {
+  internal static var readerModeActivated: Self {
     .init(name: "reader-mode-activated", lifetimeInDays: 7)
   }
   fileprivate static var navigationActionPerformedStorage: Self {
--- a/ios/brave-ios/Sources/Brave/Frontend/Browser/BrowserViewController/BVC+TabManagerDelegate.swift
+++ b/ios/brave-ios/Sources/Brave/Frontend/Browser/BrowserViewController/BVC+TabManagerDelegate.swift
@@ -95,9 +95,19 @@ extension BrowserViewController: TabManagerDelegate {
     tab.addPolicyDecider(braveShieldsHelper)
     tab.logins = .init(tab: tab, passwordAPI: profileController.passwordAPI)
     tab.nightMode = .init(tab: tab)
-
-    if FeatureList.kUseProfileWebViewConfiguration.enabled {
-      tab.readerMode = .init(tab: tab)
+    // reader mode
+    tab.readerMode = .init(tab: tab, readerModeCache: ReaderModeScriptHandler.cache(for: tab))
+    tab.readerMode?.onStateChanged = { [weak self] in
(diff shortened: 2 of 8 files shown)


ios/brave-ios/Sources/Brave/Frontend/Browser/Helpers/ReaderModeTabHelper.swift
```diff
@@ -43,39 +51,255 @@ class ReaderModeTabHelper: TabObserver {
...
+}
+
+// MARK: - TabObserver
+
+extension ReaderModeTabHelper: TabObserver {
+  func tabDidCreateWebView(_ tab: some TabState) {
+    // add content script for legacy reader mode only
+    guard let browserData = tab.browserData else { return }
+    let handler = ReaderModeScriptHandler()
+    browserData.addContentScript(
```
kylehickinson: needs profile web view configuration feature flag check, we dont want to add content scripts when its enabled

ios/brave-ios/Sources/Brave/Frontend/Browser/BrowserViewController/BVC+TabManagerDelegate.swift
```diff
@@ -95,9 +95,19 @@ extension BrowserViewController: TabManagerDelegate {
...
+    // reader mode
+    tab.readerMode = .init(tab: tab, readerModeCache: ReaderModeScriptHandler.cache(for: tab))
+    tab.readerMode?.onStateChanged = { [weak self] in
+      guard let self, self.tabManager.selectedTab === tab else { return }
+      self.topToolbar.updateReaderModeState(tab.readerMode?.state ?? .unavailable)
+    }
+    tab.readerMode?.onReaderModeDisplayed = { [weak self] in
+      guard let self else { return }
+      self.showReaderModeBar(animated: true)
+      tab.showContent(true)
```
netzenbot, line 107: The `onStateChanged` and `onReaderModeDisplayed` closures strongly capture `tab`, but both are stored properties on `tab.readerMode` (which is itself stored on `tab`), creating a retain cycle. Capture `tab` weakly in both closures — e.g. `{ [weak self, weak tab] in ... }`.

https://github.com/brave/brave-core/tree/master/docs/best-practices/ios.md#IOS-001

ios/brave-ios/Sources/Brave/Frontend/Browser/Helpers/ReaderModeTabHelper.swift
```diff
@@ -43,39 +51,255 @@ class ReaderModeTabHelper: TabObserver {
...
+        0,
+        .r(1...5),
+        .r(5...20),
+        .r(20...50),
+        .r(51...),
+      ],
+      value: storage.combinedValue
+    )
+  }
+}
```
netzenbot, line 268: The content script registration in `tabDidCreateWebView` needs to be guarded by `!FeatureList.kUseProfileWebViewConfiguration.enabled`. Per [IOS-044](https://github.com/brave/brave-core/tree/master/docs/best-practices/ios.md#IOS-044), feature flag checks should gate registration, not be left to runtime branching inside handlers. As @kylehickinson noted, this will add the legacy script even when the new profile web view configuration is enabled.

ios/brave-ios/Sources/Brave/Frontend/Browser/Helpers/ReaderModeTabHelper.swift
nuo-xu: fixed in [23ccae9](https://github.com/brave/brave-core/pull/36696/commits/23ccae9fb475f37929ea548bba05562afeb50b15)

ios/brave-ios/Sources/Brave/Frontend/Browser/Helpers/ReaderModeTabHelper.swift
```diff
@@ -43,39 +51,257 @@ class ReaderModeTabHelper: TabObserver {
...
+
+    let playlistItem = tab.playlistItem
+    let translationState = tab.translationState
+    if backList.count > 1 && backList.last?.url == readerModeURL {
+      tab.goToBackForwardListItem(backList.last!)
+      PlaylistScriptHandler.updatePlaylistTab(tab: tab, item: playlistItem)
+      self.updateTranslateURLBar?(tab, translationState)
+    } else if !forwardList.isEmpty && forwardList.first?.url == readerModeURL {
+      tab.goToBackForwardListItem(forwardList.first!)
+      PlaylistScriptHandler.updatePlaylistTab(tab: tab, item: playlistItem)
```
kylehickinson: Seems like this is not in the right place anymore (doesn't have anything to do with reader mode)

kylehickinson: Wonder if we should just get rid of all this extra logic around checking back/forward history

nuo-xu: good point. i tested a bit more and i think it's safe to remove since the back/foward list in enable and disable reader mode only affact how user navigates. plus since now `readabilityResult` is the source of truth. whenever user enables/disables reader mode webview should always load correctly.
also `backList.count > 1` seems like a latent bug as if the back list has exactly one item that is the reader mode URL, it skips the back navigation and falls through to the cache and load path anyways which i think is the most common case.

ios/brave-ios/Sources/Brave/Frontend/Browser/BrowserViewController/BVC+P3A.swift
```diff
@@ -455,7 +435,7 @@ extension P3ATimedStorage where Value == Int {
   fileprivate static var braveVPNDaysInMonthUsedStorage: Self {
     .init(name: "vpn-days-in-month-used", lifetimeInDays: 30)
   }
-  fileprivate static var readerModeActivated: Self {
+  internal static var readerModeActivated: Self {
```
kylehickinson: Missed this but if P3A logic now belongs to tab helper then move this to the helper file as well so it can maintain fileprivate

nuo-xu: addressed in [f1ffa33](https://github.com/brave/brave-core/pull/36696/commits/f1ffa33c7c1eb723919c327e3528ab6ebbdb5500)

kylehickinson (approved):

netzenbot (approved):

PR #36697, opened 2026-05-25
[Android] Adjust version code generation to higher build numbers

### LINKED ISSUE #36960: [Android] Update/Modify version code generation script
Current version code generation script `build/util/android_chrome_version.py` supports only 2 digits for **Build** number. 
Existing code will fail if **Build** number goes more than 99.
This issue is to understand and make changes on how our versioning will look like for beyond **Build** 99.

samartnik:
Resolves https://github.com/brave/brave-browser/issues/36960


SergeyZhukovsky (approved):
++ awesome patch!

goodov (approved):
lgtm since we already have a patch that imports the same file.

but better to use chromium_src inline to have auto-rebuild work on imported file change. right now any change to `brave_version_utils.py` won't trigger a rebuild.

PR #36711, opened 2026-05-26
[iOS] Add and edit saved passwords in Manage Passwords settings

### LINKED ISSUE #52212: Update Editable Login Settings UI to match new designs
Implement Add/Edit Password feature as per spec

1. User Taps "Add" or Plus button
2. User sees Add Mode
3. User completes the form
4. User taps on "Save " or "Confirm" button 
5. Password Edits are saved to using the autofill data manager (CWV Autofill Data Manager)
6. Managed Password can be observed on desktop to confirm that syncing

EliHini:
Resolves https://github.com/brave/brave-browser/issues/52212

## Summary

- Implements **create, read, update, and delete** for saved logins in iOS **Settings → Autofill → Manage Passwords**, wiring the previously stubbed **`Add Button`** (**+**) actions on the main list and per-domain group screens.
- Refactors the password detail screen into a **container** (`ManagePasswordDetailContainerView`) that switches between **read-only** (`ManagePasswordDetailReadOnlyView`, renamed from `ManagePasswordDetailView`) and **editable** (`ManagePasswordDetailEditableView`) based on add vs. edit mode and SwiftUI `editMode`.
- Introduces **`ManagePasswordDraft`** for in-flight form state, with validation (non-empty username/password, fully qualified `http`/`https` URL via `URL.isFullyQualifiedWebPage()`), and **`ManagePasswordsViewModel`** helpers that call `CWVAutofillDataManager` for add/update.
- **Add flow:** sheet with Cancel/Save; credentials are saved **only** when the user taps Save. Site can be pre-filled (and locked) when adding from a domain group.
- **Edit flow:** `EditButton` toggles read-only ↔ editable; changes persist **only** when the user taps **Done** (`editMode` becomes inactive). Cancel restores the original snapshot. Delete is available from the editable detail view with a confirmation dialog.
- **Discard on dismiss:** Leaving the editable UI (sheet swipe-down, back navigation, pop) resets the draft and does **not** write to storage—accidental edits are not committed just because the screen was dismissed.

## Test plan
- [ ] Go to brave://flags
      a. Look for Chromium WebView's Autofill 
      b. Select Enabled
- [ ] Open **Settings → Autofill → Manage Passwords** (authenticate if prompted).
- [ ] Tap **`Add Button`**  (**+**)  on the main list → add sheet opens; Save stays disabled until site, username, and password are valid.
- [ ] Fill in add fields, then **swipe down to dismiss the sheet** → no credential is created; list unchanged.
- [ ] Add again, tap **Save** → credential appears in the correct domain group.
- [ ] Add again, tap **Cancel** → sheet closes with no save.
- [ ] Open an existing credential → read-only detail; tap **Edit** → fields editable.
- [ ] Change fields, tap **Done** → changes persist when returning to read-only/detail.
- [ ] Change fields, tap **Cancel** (leading toolbar) → changes discarded; original values shown.
- [ ] Change fields, **navigate back** (or otherwise leave the screen) **without** tapping Done → credential unchanged in storage.
- [ ] Delete a credential from edit detail → confirm dialog → credential removed and screen dismissed.
- [ ] From a multi-credential domain group, tap **`Add Button`**  (**+**)  → site pre-filled and disabled; fill username/password, **swipe down to dismiss** → nothing saved.
- [ ] From a domain group, tap **`Add Button`**  (**+**) , tap **Save** → new credential appears under that domain.
- [ ] Privacy overlay: background app / Control Center → list redacts; detail/add sheets respect redaction.
- [ ] Single-credential row navigates directly to detail; multi-credential row opens group view.


ios/brave-ios/Sources/BraveStrings/BraveStrings.swift
```diff
@@ -9434,6 +9434,22 @@ extension Strings {
...
+        comment: "Navigation title when creating a new saved login from password management"
+      )
+    public static let managePasswordDetailDeleteConfirmMessage = NSLocalizedString(
+      "autofill.managePasswords.passwordDetail.deleteConfirmMessage",
+      tableName: "BraveShared",
+      bundle: .module,
+      value: "This password will be removed from your sync chain and your device.",
+      comment:
+        "Alert message warning the user that deleting non-synced the current password will permanently remove them"
+    )
```
netzenbot, line 9456: nit: The `comment` field says "deleting non-synced the current password" but the string value says it removes from "your sync chain and your device" — these contradict each other. The comment should be updated to accurately describe what the string communicates so translators have correct context. ([Delete Dead Documentation](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-002))

StephenHeaps: Comment needs fixed

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailEditableView.swift
```diff
@@ -0,0 +1,143 @@
...
+                .textInputAutocapitalization(.never)
+                .autocorrectionDisabled()
+                .accessibilityLabel(Strings.Login.loginInfoDetailsPasswordFieldTitle)
+              } else {
+                SecureField(
+                  Strings.Autofill.managePasswordDetailInputPasswordPlaceholder,
+                  text: $passwordValue
+                )
+                .textContentType(.password)
+                .accessibilityLabel(Strings.Login.loginInfoDetailsPasswordFieldTitle)
```
StephenHeaps: Should be on the Group

StephenHeaps (commented):
There's a weird flash when action sheet appears:

https://github.com/user-attachments/assets/f887b7f0-7cb6-449c-8e46-89d866aa0cdd

StephenHeaps:
Is there supposed to be this much padding at the top of the view? I don't see any design document link to verify against.

Image

StephenHeaps:
Selecting a row on iPad / iOS 18 hides all details:

https://github.com/user-attachments/assets/6a6e7380-ffec-48c8-993d-4bbdbb7cf45d

EliHini:
> Is there supposed to be this much padding at the top of the view? I don't see any design document link to verify against.
> 
> Image

Yeah, that's not in the design. Unfortunately that's what we get when we use searchable. @kylehickinson and I went back and forth on the placement of the search bar and we landed on using the default searchable

StephenHeaps:
> Unfortunately that's what we get when we use searchable.

This isn't true, `searchable()` api is unrelated here. There is extra padding on the view itself somewhere causing this.

searchable

EliHini:
> Selecting a row on iPad / iOS 18 hides all details:
> 
>  select.row.mov

That's a strange behaviour; it almost look as though something else comes up.Could this be your device? 
I am not seeing this on any of my physical devices (iPhone, iPad mini, iPad Pro). This part of the code wasn't touched in this PR.

https://github.com/user-attachments/assets/98ecc749-ec2f-481b-a847-4233517755a7


https://github.com/user-attachments/assets/7061f33c-fec0-42e6-80a3-5d0ed243ab23

EliHini:
> There's a weird flash when action sheet appears:
> 
>  action.sheet.mov

I wonder if the flash and the details hiding is a simulator issue. I am not seeing that either on any of the 3 physical test devices

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailEditableView.swift
EliHini: 100%

ios/brave-ios/Sources/BraveStrings/BraveStrings.swift
EliHini: Fixed

EliHini:
> > Unfortunately that's what we get when we use searchable.
> 
> This isn't true, `searchable()` api is unrelated here. There is extra padding on the view itself somewhere causing this.
> 
> searchable

I wasn't meaning the searchable itself. More about the original implementation to match the figma design before going down this route.
I've tightened up the spacing using `listSectionSpacing(.compact)`, but as expected, it does affect the spacing for the other section header.

If we want to iterate further, I will create a new issue for it so we don't pollute this PR.

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailEditableView.swift
StephenHeaps: Missed the `.textContentType(.password)`

kylehickinson:
@EliHini dont enable auto-merge on non-trivial PRs that haven't been approved/are in the midst of reviews.

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailEditableView.swift
```diff
@@ -0,0 +1,143 @@
...
+            .multilineTextAlignment(.trailing)
+            .focused($focusedField, equals: .site)
+            .foregroundStyle(
+              Color(braveSystemName: isSitePrefilled ? .textTertiary : .textSecondary)
+            )
+        } label: {
+          Text(Strings.Login.loginInfoDetailsWebsiteFieldTitle)
+            .foregroundStyle(Color(braveSystemName: .textPrimary))
+        }
+        .listRowBackground(Color(.secondaryBraveGroupedBackground))
```
kylehickinson: remove this and all legacy color usages as they will be removed anyways by the time this ships

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailEditableView.swift
```diff
@@ -0,0 +1,143 @@
...
+            .accessibilityLabel(Strings.Login.loginInfoDetailsWebsiteFieldTitle)
+            .disabled(isSitePrefilled)
+            .multilineTextAlignment(.trailing)
+            .focused($focusedField, equals: .site)
+            .foregroundStyle(
+              Color(braveSystemName: isSitePrefilled ? .textTertiary : .textSecondary)
+            )
+        } label: {
+          Text(Strings.Login.loginInfoDetailsWebsiteFieldTitle)
+            .foregroundStyle(Color(braveSystemName: .textPrimary))
```
kylehickinson: remove any `textPrimary` usages, replace any `textSecondary` with system `.secondary` style

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailContainerView.swift
```diff
@@ -0,0 +1,162 @@
...
+    /// Opens the detail screen to add a new login, optionally pre-seeded with a draft.
+    case add(ManagePasswordDraft)
+  }
+
+  @Environment(\.dismiss) private var dismiss
+  @Environment(\.editMode) private var editMode
+  @Environment(\.redactionReasons) private var redactionReasons
+
+  @State private var isPasswordRevealed = false
+  @State private var passwordDraft: ManagePasswordDraft = .init()
```
kylehickinson: Why would the password draft live in this parent container if its only usage is in the add/edit mode?

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailContainerView.swift
```diff
@@ -0,0 +1,162 @@
...
+    case edit(CWVPassword)
+    /// Opens the detail screen to add a new login, optionally pre-seeded with a draft.
+    case add(ManagePasswordDraft)
+  }
+
+  @Environment(\.dismiss) private var dismiss
+  @Environment(\.editMode) private var editMode
+  @Environment(\.redactionReasons) private var redactionReasons
+
+  @State private var isPasswordRevealed = false
```
kylehickinson, line 13: Why is this shared between both modes? During editing you'd already be revealing it?

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailContainerView.swift
```diff
@@ -0,0 +1,162 @@
...
+import SwiftUI
+
+struct ManagePasswordDetailContainerView: View {
+  /// How the password detail screen is opened: view/edit an existing login, or add a new one.
+  enum Mode {
+    /// Opens the detail screen to view or edit an existing saved login.
+    case edit(CWVPassword)
+    /// Opens the detail screen to add a new login, optionally pre-seeded with a draft.
+    case add(ManagePasswordDraft)
+  }
```
kylehickinson: This is weird, why would I ever want to present a `ManagePasswordDetailContainerView` when adding a brand new password which will supply no defaults?

kylehickinson: Honestly you may want to just split up Add & Edit entirely

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailEditableView.swift
```diff
@@ -0,0 +1,143 @@
...
+  @FocusState private var focusedField: Field?
+  @Binding var isPasswordRevealed: Bool
+  @Binding var site: String
+  @Binding var username: String
+  @Binding var passwordValue: String
+  @State private var isDeleteDialogPresented = false
+
+  var isSitePrefilled: Bool = false
+  var deleteAction: DeleteExistingPasswordAction? = nil
+  var body: some View {
```
kylehickinson, line 58: once you split out add & edit this will become much simpler,  edit will hold its own state, and can assume sites & delete action will always be supplied

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailEditableView.swift
```diff
@@ -0,0 +1,143 @@
...
+            Group {
+              if isPasswordRevealed {
+                TextField(
+                  Strings.Autofill.managePasswordDetailInputPasswordPlaceholder,
+                  text: $passwordValue
+                )
+                .textContentType(.password)
+                .textInputAutocapitalization(.never)
+                .autocorrectionDisabled()
+                .accessibilityLabel(Strings.Login.loginInfoDetailsPasswordFieldTitle)
```
kylehickinson: This is already part of the group

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailEditableView.swift
```diff
@@ -0,0 +1,143 @@
...
+          HStack(spacing: 8) {
+            Group {
+              if isPasswordRevealed {
+                TextField(
+                  Strings.Autofill.managePasswordDetailInputPasswordPlaceholder,
+                  text: $passwordValue
+                )
+                .textContentType(.password)
+                .textInputAutocapitalization(.never)
+                .autocorrectionDisabled()
```
kylehickinson: I assume all of these should be part of the group?

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailEditableView.swift
```diff
@@ -0,0 +1,143 @@
...
+      if let deleteAction {
+        Section {
+          Button {
+            isDeleteDialogPresented = true
+          } label: {
+            Label(
+              Strings.Autofill.managePasswordsDeleteCredentialButtonTitle,
+              braveSystemImage: "leo.trash"
+            )
+            .labelStyle(.titleOnly)
```
kylehickinson: Why supply an image and then use the `titleOnly` label?

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordsViewModel.swift
```diff
@@ -171,3 +191,61 @@ private class AutofillDataManagerObserver: NSObject, CWVAutofillDataManagerObser
     notify()
   }
 }
+
+/// Editable snapshot of a saved login’s fields while the user is in add or edit mode.
+/// `CWVPassword` is the persisted model; this type holds the in-flight strings bound to text fields.
+struct ManagePasswordDraft: Identifiable {
```
kylehickinson: Doesn't seem like this type is needed once you split up add & edit

ios/brave-ios/Sources/Shared/Extensions/URLExtensions.swift
```diff
@@ -215,6 +215,13 @@ extension URL {
     return scheme.map { schemes.contains($0) } ?? false
   }
 
+  /// Requires both an http/https scheme *and* a non-empty host. `data:` URLs are excluded because they have no host.
```
kylehickinson: where is this requirement set? All I see is that the `GURL` must be valid here: https://source.chromium.org/chromium/chromium/src/+/main:ios/web_view/internal/autofill/cwv_autofill_data_manager.mm;l=366

ios/brave-ios/Sources/Shared/Extensions/URLExtensions.swift
```diff
@@ -215,6 +215,13 @@ extension URL {
...
+  ///
+  /// Use this when a URL is being persisted or sent to an API that expects a real reachable web address.
+  public func isFullyQualifiedWebPage() -> Bool {
+    isWebPage(includeDataURIs: false) && !(host?.isEmpty ?? true)
+  }
```
kylehickinson: If this can't be removed, it shouldn't go in `Shared`, its a legacy package, new things should be added in `BraveShared`

netzenbot (approved):

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailEditableView.swift
EliHini: what are the direct replacements for `.secondaryBraveGroupedBackground` and `textPrimary`?

kylehickinson: None, remove and allow iOS to pick defaults

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailContainerView.swift
EliHini: That's because the container was once the detail view itself before I refactored Edit/Add specifics further out of it. I will refactor the whole thing out

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailEditableView.swift
EliHini: good catch

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailContainerView.swift
EliHini: This is in keeping with other password managers. Even when editing the password is hidden unless the user opts to reveal it. Same in Chrome

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailEditableView.swift
EliHini: yeah I missed that one

ios/brave-ios/Sources/Shared/Extensions/URLExtensions.swift
EliHini: It may not be documented but it behaves just as so.  
I guess, to be more correct. It needs a valid scheme and domain.

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailEditableView.swift
EliHini: This is a pattern we had established in previously accepted PRs which is why I kept on with it.

ios/brave-ios/Sources/Shared/Extensions/URLExtensions.swift
kylehickinson: Please point to me where the check happens. Either way if this is specific to the autofill manager (or Chromium) then we should add the check over there instead. This isn't really an extension we need

EliHini: I will remove from the extension.
I am doing the check here: https://github.com/brave/brave-core/pull/36711/changes#diff-850b937f1fcc1c3e8f7d886bdd10b57dab352b4a7717acf9dfb0dd3146f66ddfR223


Screenshot 2026-06-12 at 4 59 45 PM

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailEditableView.swift
EliHini: should I also remove these `.braveGroupedBackground`?

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailEditableView.swift
EliHini: yeah, it is meant to be on the containing group

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailEditableView.swift
kylehickinson: Yes

ios/brave-ios/Sources/Shared/Extensions/URLExtensions.swift
kylehickinson: Okay so it is the URL validity check, I guess `URL` lets you create one without a host? We should probably be picking some sensible defaults, `www.dhdhd.com` is a valid url yet still hit this check because a scheme wasn't included. If a user doesn't include a scheme we should add one ourselves (https by default), that sort of thing. Failing because the user didnt type in an exactly valid URL would feel overall bad

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailContainerView.swift
EliHini: Ended up removing the draft altogether.

ios/brave-ios/Sources/Shared/Extensions/URLExtensions.swift
EliHini: ok, defaulting to `https ` on missing scheme

ios/brave-ios/Sources/Shared/Extensions/URLExtensions.swift
EliHini: removed

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordAddView.swift
```diff
@@ -0,0 +1,88 @@
...
+
+  var body: some View {
+    Form {
+      ManagePasswordCredentialFields(
+        site: $site,
+        username: $username,
+        password: $password,
+        isPasswordRevealed: $isPasswordRevealed,
+        focusedField: $focusedField,
+        isSiteDisabled: isSitePrefilled
```
kylehickinson: I dont think we should disable the site field even if you tap the add button from a group of that site

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordCredentialFields.swift
```diff
@@ -0,0 +1,98 @@
...
+  @FocusState.Binding var focusedField: Field?
+
+  var isSiteDisabled: Bool
+
+  var body: some View {
+    Section {
+      LabeledContent {
+        Group {
+          if isSiteDisabled {
+            Text(site)
```
kylehickinson: Does it look bad if you just use the `TextField` for both states but just add `.disabled(isSiteDisabled)` on it?

EliHini: No, not the way it was after I refactored the form. That's the way I had before the refactor.
The split was a workaround for a `.constant` binding regression; 

I've reverted to a single `TextField` with `.disabled()`, edit now uses `@State` site like before.

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordCredentialFields.swift
```diff
@@ -0,0 +1,90 @@
...
+              .textContentType(.password)
+              .textInputAutocapitalization(.never)
+            } else {
+              SecureField(
+                Strings.Autofill.managePasswordDetailInputPasswordPlaceholder,
+                text: $password
+              )
+              .textContentType(.password)
+            }
+          }
```
StephenHeaps, line 64: Anything preventing `.textContentType(.password)` from being on the `Group`?

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordAddView.swift
```diff
@@ -0,0 +1,88 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import BraveStrings
+import SwiftUI
+
+/// Drives add password sheet presentation so each presentation carries its own prefilled site.
+struct ManagePasswordAddPresentation: Identifiable {
+  let id = UUID()
+  let prefilledSite: String
```
StephenHeaps: Seems like this can be optional, unless there's a reason to use empty string?

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailReadOnlyView.swift
StephenHeaps, line 1: I think this is the correct view I am looking at, but it seems like we lost the list row background colours (possibly during Nala colour migrations?)

Image of Manage Passwords view

EliHini: Yeah, we removed `.braveGroupedBackground` and `.secondaryBraveGroupedBackground`
@kylehickinson

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordCredentialFields.swift
EliHini: Nope, I forgot about these after the refactor. This is carry-over duplication from before `ManagePasswordCredentialFields` existed. In the original editable view, `.textContentType(.password)` was on each branch inside the `if/else`.

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordAddView.swift
EliHini: oh yes it should be nil, another casualty of the refactor when a password object was always passed in.

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordAddView.swift
EliHini: We can change it. It will mean that it is not an explicit add to the group. Therefore if the user changes the site there might be a slight confusion because they will return to the group not realizing that the password has been added as a standalone entry

kylehickinson: Better to treat it as a pre-filled convenience for the group, not an explicit requirement that you add it to that specific group.

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordDetailReadOnlyView.swift
kylehickinson: I would have just removed `listRowBackground`s in my PR, maybe missed an explicit background?

EliHini: yup, found the culprit: missed `.scrollContentBackground(.hidden)`. Fixed now

Screenshot 2026-06-24 at 5 44 36 PM

ios/brave-ios/Sources/BraveShared/Extensions/URLExtensions.swift
```diff
@@ -354,6 +354,30 @@ extension URL {
...
+  ///   `http` or any other scheme, it is left as-is.
+  ///
+  /// ## Examples
+  /// ```
+  /// URL(string: "example.com")!.resolvingScheme      // https://example.com
+  /// URL(string: "www.example.com")!.resolvingScheme  // https://www.example.com
+  /// URL(string: "https://example.com")!.resolvingScheme // https://example.com
+  /// URL(string: "http://example.com")!.resolvingScheme  // http://example.com
+  /// ```
+  public var resolvingScheme: URL {
```
StephenHeaps: nit: may be more clear to include `httpsIfSchemeless` or `secureIfSchemeless` (example names, might be something better 😅) in the variable name, or make this one a file private extension where it's used

ios/brave-ios/Sources/Brave/Frontend/Autofill/ManagePasswordCredentialFields.swift
```diff
@@ -0,0 +1,89 @@
...
+          .multilineTextAlignment(.trailing)
+          .focused($focusedField, equals: .password)
+          .foregroundStyle(.secondary)
+
+          Button {
+            isPasswordRevealed.toggle()
+          } label: {
+            Label(
+              Strings.Autofill.managePasswordDetailRevealPassword,
+              braveSystemImage: isPasswordRevealed ? "leo.eye.on" : "leo.eye.off"
```
StephenHeaps: cc @aguscruiz 
Should it be `eye.on` for password revealed, or `eye.off` 🤔. Currently this view differs from Wallet so we should keep consistent

Image | Image
--|--

Image |

aguscruiz: I think the standard is 

- eye.on to show ********.
- eye.off when it actually shows the password. 

This one is correct
image

StephenHeaps (approved):

PR #36713, opened 2026-05-26
[Wallet] Refactor solana tx serialization

supermassive:
Resolves

Need for https://github.com/brave/brave-browser/issues/55334


components/brave_wallet/browser/solana_message.cc
```diff
@@ -334,93 +325,94 @@ std::optional<std::vector<uint8_t>> SolanaMessage::Serialize(
...
 
   // Instructions length
-  ret = CompactU16Decode(bytes, bytes_index);
-  if (!ret) {
-    return std::nullopt;
-  }
-  bytes_index += std::get<1>(*ret);
-  size_t num_of_instructions = std::get<0>(*ret);
-
+  auto num_of_instructions = CompactU16Decode(reader);
```
netzenbot, line 432: `num_of_instructions` is not validated before `*num_of_instructions` is dereferenced on this line. The corresponding old code had an explicit `if (!ret) { return std::nullopt; }` guard that was dropped in this refactor. If `CompactU16Decode` returns `std::nullopt` (e.g. the buffer is truncated), this is undefined behavior. Add `if (!num_of_instructions) { return std::nullopt; }` between the `CompactU16Decode` call and the `for` loop, consistent with how `num_of_accounts` is guarded a few lines above.

components/brave_wallet/browser/solana_message.cc
```diff
@@ -334,93 +325,94 @@ std::optional<std::vector<uint8_t>> SolanaMessage::Serialize(
...
   std::vector<SolanaCompiledInstruction> compiled_instructions;
-  for (size_t i = 0; i < num_of_instructions; ++i) {
-    auto compiled_instruction =
-        SolanaCompiledInstruction::Deserialize(bytes, &bytes_index);
+  for (size_t i = 0; i < *num_of_instructions; ++i) {
+    auto compiled_instruction = SolanaCompiledInstruction::Deserialize(reader);
     if (!compiled_instruction) {
```
netzenbot, line 439: The null check that was present in the old code (`if (!ret) { return std::nullopt; }`) was removed in this refactor. `CompactU16Decode(reader)` returns `std::optional<uint16_t>`, but `*num_of_instructions` is dereferenced without checking first — this is undefined behavior on malformed input. Add `if (!num_of_instructions) { return std::nullopt; }` before the loop.

components/brave_wallet/browser/solana_message.cc
```diff
@@ -430,14 +422,14 @@ std::optional<SolanaMessage> SolanaMessage::Deserialize(
   std::optional<std::vector<SolanaMessageAddressTableLookup>>
```
netzenbot, line 445: The `num_of_instructions` optional is dereferenced directly in the loop condition (`*num_of_instructions`) without a prior null-check, unlike the other `CompactU16Decode` calls in the same function that all guard with `if (!num_of_accounts) { return std::nullopt; }`. If `CompactU16Decode` returns `std::nullopt` here (e.g. on malformed input) this is undefined behavior. Add the missing null-check before the loop. See [CSM-029](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-memory.md#CSM-029).

components/brave_wallet/browser/solana_message.cc
```diff
@@ -334,93 +325,94 @@ std::optional<std::vector<uint8_t>> SolanaMessage::Serialize(
...
-
+  auto num_of_instructions = CompactU16Decode(reader);
   std::vector<SolanaCompiledInstruction> compiled_instructions;
-  for (size_t i = 0; i < num_of_instructions; ++i) {
-    auto compiled_instruction =
-        SolanaCompiledInstruction::Deserialize(bytes, &bytes_index);
+  for (size_t i = 0; i < *num_of_instructions; ++i) {
+    auto compiled_instruction = SolanaCompiledInstruction::Deserialize(reader);
     if (!compiled_instruction) {
       return std::nullopt;
```
netzenbot, line 440: Missing null check: `num_of_instructions` is the result of `CompactU16Decode(reader)` which returns `std::optional<uint16_t>`, but it is dereferenced (`*num_of_instructions`) in the loop without first checking `if (!num_of_instructions) { return std::nullopt; }`. The old code had an explicit null check (`if (!ret) { return std::nullopt; }`) that was dropped in this refactor. Deserializing a malformed transaction will cause undefined behavior here.

components/brave_wallet/browser/solana_message.cc
```diff
@@ -430,14 +422,14 @@ std::optional<SolanaMessage> SolanaMessage::Deserialize(
   std::optional<std::vector<SolanaMessageAddressTableLookup>>
       addr_table_lookups = std::vector<SolanaMessageAddressTableLookup>();
   if (version == mojom::SolanaMessageVersion::kV0) {
-    addr_table_lookups = DeserializeAddressTableLookups(bytes, &bytes_index);
+    addr_table_lookups = DeserializeAddressTableLookups(reader);
     if (!addr_table_lookups) {
```
netzenbot, line 449: Missing null check: `num_of_instructions` from `CompactU16Decode(reader)` is dereferenced unconditionally as `*num_of_instructions` in the loop condition, but the null check (`if (!ret) { return std::nullopt; }`) that existed in the old code was not carried over. If the reader is truncated here, this is UB. Add `if (!num_of_instructions) { return std::nullopt; }` before the loop.

components/brave_wallet/browser/solana_message.cc
```diff
@@ -334,93 +325,94 @@ std::optional<std::vector<uint8_t>> SolanaMessage::Serialize(
...
+  if (!blockhash_bytes) {
     return std::nullopt;
   }
-  base::span<const uint8_t> blockhash_bytes =
-      base::span(bytes).subspan(bytes_index, kSolanaHashSize);
-  bytes_index += kSolanaHashSize;
-  const std::string recent_blockhash = Base58Encode(blockhash_bytes);
+  const std::string recent_blockhash = Base58Encode(*blockhash_bytes);
 
   // Instructions length
```
netzenbot, line 431: Missing null check: `num_of_instructions` is dereferenced unconditionally (`*num_of_instructions`) but `CompactU16Decode` returns `std::optional<uint16_t>` and can return `std::nullopt`. The old code had an explicit `if (!ret) { return std::nullopt; }` guard that was removed in the refactor. This is UB in a wallet deserialization path processing untrusted data — add `if (!num_of_instructions) { return std::nullopt; }` before the loop. ([Unit Tests Required for Bad/Invalid Input](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-016))

components/brave_wallet/browser/brave_wallet_service.cc
```diff
@@ -1696,7 +1697,7 @@ void BraveWalletService::NotifyDecryptRequestProcessed(
 void BraveWalletService::IsBase58EncodedSolanaPubkey(
     const std::string& key,
     IsBase58EncodedSolanaPubkeyCallback callback) {
-  std::move(callback).Run(::brave_wallet::IsBase58EncodedSolanaPubkey(key));
+  std::move(callback).Run(!!SolanaAddress::FromBase58(key));
```
cmazakas: nit: maybe `.has_value()` is more readable here

This is just personal preference here though.

components/brave_wallet/browser/solana_compiled_instruction.cc
```diff
@@ -123,36 +118,34 @@ SolanaCompiledInstruction::FromInstruction(
                                    instruction.data());
 }
 
-void SolanaCompiledInstruction::Serialize(std::vector<uint8_t>* bytes) const {
-  DCHECK(bytes);
+void SolanaCompiledInstruction::Serialize(std::vector<uint8_t>& bytes) const {
+  bytes.emplace_back(program_id_index_);
 
-  bytes->emplace_back(program_id_index_);
+  base::Extend(bytes, CompactU16Encode(account_indexes_.size()));
```
cmazakas: This code looks like it can silently truncate `account_indexes_.size()`

We should maybe consider updating `CompactU16Encode` to accept a `StrictNumeric` here and force the caller to convert and handle it failing. Unless we can somehow guarantee the size of `account_indexes_` here to be within a `uint16_t`'s range.

components/brave_wallet/common/solana_utils_unittest.cc
```diff
@@ -28,38 +45,36 @@ TEST(SolanaUtilsUnitTest, CompactU16Encode) {
       {0x7fff, {0xff, 0xff, 0x01}},
       {0xffff, {0xff, 0xff, 0x03}}};
```
cmazakas, line 48: Should we consider adding a couple of extra test cases here for the values `0x3fff` and `0x4000`?

components/brave_wallet/browser/solana_compiled_instruction.cc
supermassive: I had StrictNumeric here, then reverted as it increased PR size :)
Let me think if I can get it back.

components/brave_wallet/common/solana_utils.h
```diff
@@ -10,30 +10,28 @@
...
 // See
 // https://docs.solana.com/developing/programming-model/transactions#compact-u16-format
-void CompactU16Encode(uint16_t u16, std::vector<uint8_t>* compact_u16);
-
-std::optional<std::tuple<uint16_t, size_t>> CompactU16Decode(
-    const std::vector<uint8_t>& compact_u16,
-    size_t start_index);
-
-bool IsBase58EncodedSolanaPubkey(const std::string& key);
+absl::InlinedVector<uint8_t, 3> CompactU16Encode(uint16_t u16);
```
netzenbot: `IsBase58EncodedSolanaPubkey` is removed here, but `brave/ios/browser/api/brave_wallet/brave_wallet_solana_utils.mm` still calls `brave_wallet::IsBase58EncodedSolanaPubkey` in its `-[NSString isBase58EncodedSolanaPubkey]` method. This will break the iOS build. That file needs to be updated to use `SolanaAddress::FromBase58` instead. ([Clean Up All Dead Code When Removing Features](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-068))

components/brave_wallet/browser/solana_transaction.cc
```diff
@@ -609,28 +598,29 @@ SolanaTransaction::FromSignedTransactionBytes(
...
+  std::optional<uint16_t> num_of_signatures = CompactU16Decode(reader);
+  if (!num_of_signatures) {
     return nullptr;
   }
-  const uint16_t num_of_signatures = std::get<0>(*ret);
-  index += std::get<1>(*ret);
-  if (index + num_of_signatures * kSolanaSignatureSize > bytes.size()) {
+  auto raw_signatures = reader.Read(
+      base::CheckMul<size_t>(*num_of_signatures, kSolanaSignatureSize)
+          .ValueOrDie());
```
netzenbot, line 621: `base::CheckMul<size_t>(*num_of_signatures, kSolanaSignatureSize).ValueOrDie()` calls `ValueOrDie()` — equivalent to `CHECK` — on a value derived from untrusted external input. While the product of a `uint16_t` and 64 cannot overflow `size_t` in practice, the pattern is against [CS-027](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-027), which requires graceful error handling for external data rather than crash-on-failure. Prefer `auto byte_count = base::CheckMul<size_t>(*num_of_signatures, kSolanaSignatureSize); if (!byte_count.IsValid()) { return nullptr; }` to be consistent with the rest of this function.

components/brave_wallet/common/solana_utils.h
```diff
@@ -10,30 +10,28 @@
...
-#include "base/containers/span.h"
+#include "base/containers/span_reader.h"
 #include "brave/components/brave_wallet/common/brave_wallet.mojom-forward.h"
+#include "third_party/abseil-cpp/absl/container/inlined_vector.h"
 
 namespace base {
 class DictValue;
 }  // namespace base
 
 namespace brave_wallet {
```
netzenbot, line 24: nit: Per [CSA-067](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-067), element counts should use `size_t`. `uint8_t signatures_count` is a narrowing type for a count — callers pass `signers.size()` which is `size_t`. The UINT8_MAX guards at each call site protect correctness, but accepting `size_t` here would make the constraint explicit at the API boundary rather than relying on caller discipline.

components/brave_wallet/common/solana_utils.cc
```diff
@@ -8,43 +8,48 @@
 #include <optional>
 
 #include "base/check.h"
```
netzenbot, line 10: nit: `#include "base/numerics/safe_conversions.h"` is added but no symbol from that header (e.g. `base::saturated_cast`, `base::StrictNumeric`) appears to be used in this file. Per [CS-001](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-001), remove includes you don't actually use.

components/brave_wallet/common/solana_utils.h
```diff
@@ -10,30 +10,28 @@
...
 
+void ExtendWithEmptySignatures(std::vector<uint8_t>& bytes,
```
netzenbot, line 27: nit: `ExtendWithEmptySignatures` is a new public API but has no doc comment. The parameter `signatures_count` is non-obvious — it appends `signatures_count * 64` zero bytes, not `signatures_count` bytes. A brief comment would help callers understand the contract. ([Method Documentation Should Describe the Contract](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-004))

cmazakas (approved):
Core changes LGTM!

nuo-xu (approved):
iOS++

netzenbot (approved):

PR #36720, opened 2026-05-26
[Wallet] Fix Solana signMessage transaction guard bypass

kdenhartog:
closes https://github.com/brave/internal/issues/1750

## Summary
Prevent the submission of a solana transaction being able to be signed to match phantom behavior of rejecting

## Test Plan
See the issue for testing page

### DIFF
--- a/components/brave_wallet/browser/solana_instruction.cc
+++ b/components/brave_wallet/browser/solana_instruction.cc
@@ -13,6 +13,7 @@
 
 #include "base/base64.h"
 #include "base/containers/span.h"
+#include "base/numerics/checked_math.h"
 #include "base/values.h"
 #include "brave/components/brave_wallet/browser/solana_compiled_instruction.h"
 #include "brave/components/brave_wallet/browser/solana_instruction_data_decoder.h"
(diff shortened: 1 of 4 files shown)


components/brave_wallet/browser/solana_provider_impl.cc
```diff
@@ -650,6 +651,14 @@ void SolanaProviderImpl::SignMessage(
                             base::DictValue());
     return;
   }
+  // Reject non-UTF-8 binary payloads; signMessage is for plain text only.
```
kdenhartog, line 654: added this as a defense in depth check to match phantom behavior here too

components/brave_wallet/browser/solana_instruction.cc
supermassive, line 113: num_readonly_unsigned_accounts lgtm

btw I'd also add CheckedMath to subtractions above.

kdenhartog: ahh good call out, will add CheckedMath here

components/brave_wallet/browser/solana_instruction_unittest.cc
```diff
@@ -138,6 +138,50 @@ TEST(SolanaInstructionUnitTest, FromToCompiledInstruction) {
...
+// 8-5-4=-1 and return nullopt, bypassing the signMessage transaction guard that
+// rejects blobs which successfully deserialize as transactions.
+TEST(SolanaInstructionUnitTest,
+     FromCompiledInstructionUsesReadonlyUnsignedForWritableUnsignedCount) {
+  // 8 accounts: [0] writable signer, [1..4] readonly signers,
+  //             [5..6] writable unsigned, [7] readonly unsigned (program)
+  std::vector<SolanaAddress> accounts = {
+      *SolanaAddress::FromBase58(kAccount1),
+      *SolanaAddress::FromBase58(kAccount2),
+      *SolanaAddress::FromBase58(kAccount3),
```
netzenbot, line 149: nit: "the typo causes Brave to compute 8-5-4=-1" describes the old broken code that no longer exists. Per [CS-019](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-019), comments should read clearly without prior knowledge of the change — prefer describing what the test verifies rather than what the bug did (e.g., "A crafted header [5,4,1] with 8 accounts is valid per Solana (8-5-1=2 writable unsigned); verify FromCompiledInstruction handles it correctly.").

components/brave_wallet/browser/solana_instruction.cc
```diff
@@ -99,16 +100,22 @@ std::optional<SolanaInstruction> SolanaInstruction::FromCompiledInstruction(
...
+      message_header.num_readonly_unsigned_accounts;
+  if (!checked_writable_signed.IsValid() ||
+      !checked_writable_unsigned.IsValid() ||
+      checked_writable_signed.ValueOrDie() < 0 ||
+      checked_writable_unsigned.ValueOrDie() < 0) {  // invalid message header
     return std::nullopt;
   }
+  const int num_writable_signed_accounts = checked_writable_signed.ValueOrDie();
+  const int num_writable_unsigned_accounts =
+      checked_writable_unsigned.ValueOrDie();
```
supermassive, line 118: We should not use `int` here.
`static_accounts.size()` is capped at 256
https://solana.com/docs/core/constants-reference#transactions

So it should be `size_t` or (`uint8_t`)
```
const auto checked_writable_unsigned =
    base::CheckedNumeric<uint8_t>(static_accounts.size()) -
    message_header.num_required_signatures -
    message_header.num_readonly_unsigned_accounts;

const auto checked_writable_signed =
    base::CheckedNumeric<uint8_t>(message_header.num_required_signatures) -
    message_header.num_readonly_signed_accounts;
```
Later no need to check if it is less than 0, just `IsValid()` is enough

netzenbot (approved):

supermassive (approved):

PR #36721, opened 2026-05-26
[iOS] Add application context check in Origin factory constructor

kylehickinson:
Fixes IOSChromeUnitTest bundles crashing since the application context is not set up for those unit tests

### DIFF
--- a/ios/browser/brave_origin/brave_origin_service_factory.mm
+++ b/ios/browser/brave_origin/brave_origin_service_factory.mm
@@ -184,10 +184,12 @@ BraveOriginServiceFactory::BraveOriginServiceFactory()
   DependsOn(skus::SkusServiceFactory::GetInstance());
   auto* policy_manager = BraveOriginPolicyManager::GetInstance();
   policy_manager->SetExpectedToBeInitialized();
-  if (!policy_manager->IsInitialized()) {
+  auto* application_context = GetApplicationContext();
+  // ApplicationContext may not be set up in unit tests yet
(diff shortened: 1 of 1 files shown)


ios/browser/brave_origin/brave_origin_service_factory.mm
```diff
@@ -184,10 +184,12 @@ explicit BraveOriginDelegateIOS(SkusServiceGetter skus_service_getter)
...
   auto* policy_manager = BraveOriginPolicyManager::GetInstance();
   policy_manager->SetExpectedToBeInitialized();
-  if (!policy_manager->IsInitialized()) {
+  auto* application_context = GetApplicationContext();
+  // ApplicationContext may not be set up in unit tests yet
+  if (!policy_manager->IsInitialized() && application_context) {
     policy_manager->Init(GetBrowserPolicyDefinitions(),
                          GetProfilePolicyDefinitions(),
-                         GetApplicationContext()->GetLocalState());
+                         application_context->GetLocalState());
```
supermassive, line 192: Not sure if that's the case, but would that be OK if policy_manager->Init might never be called in tests?

kylehickinson: Yeah if any tests actually needed to test this factory we could set up an `TestApplicationContext` for that test in particular (we dont currently have anything though that needs it)

supermassive (approved):

goodov (approved):

PR #36725, opened 2026-05-26
Import passwords from Brave when importing into Brave Origin for Linux and macOS

### LINKED ISSUE #55631: Passwords are not imported from standard Brave to Brave Origin branded (macOS and Linux)
### Description

Passwords are not imported 

### Steps to reproduce

1. Install standard Brave
2. Open `https://www.w3schools.com/Tags/tryit.asp?filename=tryhtml5_input_type_password`
3. Set the password
4. Open `brave://password-manager/passwords`
5. Clean profile Brave Origin branded build
6. Run Brave Origin
7. On `brave://welcome` page, click `Skip`
8. On `brave://welcome` page, import from standard Brave
9. Open `brave://password-manager/passwords`

### Actual result

Passwords are not imported 

Image

### Expected result

Passwords are imported 

Image

### Reproduces how often

Easily reproduced

Brave Origin | 1.91.154 Chromium: 148.0.7778.167 (Official Build) beta (64-bit)
-- | --
Revision | 11b4ffabf27c9a5e4d52bafa3ebab8239472d276
OS | Linux

cc @bbondy @fmarier

bbondy:
- Adds `BravePasswordImporter` which runs in the browser process and uses `OSCryptAsync` to decrypt the source Brave  `Login Data` SQLite store; decrypted credentials are written to the destination profile's `PasswordStore`.
- Brave Origin and Standard Brave share the same OS-level encryption key

note: Windows is not included because it uses a different key

note 2: Right now you can only import Brave from Brave Origin. If you are using a Brave build you can't use this.

Fixes https://github.com/brave/brave-browser/issues/55631

## Test plan

If testing with a dev build, then use `--use_real_keychain`

1. Have a Brave profile with some passwords in it
2. Quit the application
3. Start a new install of Brave Origin
4. On `brave://welcome`, click Skip, then import from Brave.
5. Open `brave://password-manager/passwords` in Brave Origin (in one of the imported profiles) and verify the passwords are present.
6. Repeat the process with brave://settings "Import bookmarks and settings" option.  It should list passwords only on Linux and macOS and import only when selected.

### DIFF
--- a/browser/importer/brave_external_process_importer_host.cc
+++ b/browser/importer/brave_external_process_importer_host.cc
@@ -5,19 +5,44 @@
 
 #include "brave/browser/importer/brave_external_process_importer_host.h"
 
+#include <utility>
+
 #include "base/check.h"
+#include "base/logging.h"
 #include "brave/browser/importer/brave_importer_p3a.h"
 #include "brave/browser/importer/extensions_import_helpers.h"
 #include "brave/grit/brave_generated_resources.h"
+#include "build/build_config.h"
 #include "chrome/browser/importer/importer_lock_dialog.h"
+#include "components/user_data_importer/common/importer_data_types.h"
+#include "components/user_data_importer/common/importer_type.h"
 
 BraveExternalProcessImporterHost::BraveExternalProcessImporterHost()
     : weak_ptr_factory_(this) {}
 BraveExternalProcessImporterHost::~BraveExternalProcessImporterHost() = default;
 
 void BraveExternalProcessImporterHost::NotifyImportEnded() {
-  if (!cancelled_) {
+  // Guard against re-recording: this method is re-entered to continue the
+  // import lifecycle after browser-process steps (password and extension
+  // import), and P3A must be recorded only once per import.
+  if (!cancelled_ && !p3a_recorded_) {
+    p3a_recorded_ = true;
     RecordImporterP3A(source_profile_.importer_type);
+  }
+
+  // Passwords are imported in the browser process (not the utility process) so
+  // OSCryptAsync can decrypt the source's `Login Data`. `password_importer_` is
+  // created lazily here and reset in OnPasswordsImported(), which re-enters
+  // this method to continue the lifecycle.
+  if (NeedToImportPasswords() && !password_import_started_) {
+    password_import_started_ = true;
+    password_importer_ = std::make_unique<BravePasswordImporter>();
+    NotifyImportItemStarted(user_data_importer::PASSWORDS);
+    password_importer_->StartImport(
+        source_profile_.source_path, profile_,
+        base::BindOnce(&BraveExternalProcessImporterHost::OnPasswordsImported,
+                       weak_ptr_factory_.GetWeakPtr()));
+    return;
   }
 
   // If user chooses extension importing, start importing extensions.
@@ -40,6 +65,42 @@ void BraveExternalProcessImporterHost::NotifyImportEnded() {
 
   // Otherwise, notifying here and importing is finished.
   ExternalProcessImporterHost::NotifyImportEnded();
+}
+
+bool BraveExternalProcessImporterHost::NeedToImportPasswords() const {
+#if BUILDFLAG(IS_MAC) || BUILDFLAG(IS_LINUX)
+  // Only Brave-to-Brave imports share the OS keychain entry needed to decrypt
+  // the source's `Login Data`.
+  return !cancelled_ &&
+         (items_ & user_data_importer::PASSWORDS) ==
+             user_data_importer::PASSWORDS &&
+         source_profile_.importer_type == user_data_importer::TYPE_BRAVE;
+#else
+  return false;
+#endif
+}
+
+void BraveExternalProcessImporterHost::OnPasswordsImported(
+    BravePasswordImporter::Result result,
+    size_t /*submitted*/) {
+  password_importer_.reset();
+
+  if (result == BravePasswordImporter::Result::kSuccess) {
+    // Only report the item as ended on success; reporting it on failure would
+    // mark the overall import as succeeded. A successful import of zero
+    // credentials just means the source had no saved passwords.
+    NotifyImportItemEnded(user_data_importer::PASSWORDS);
+  } else {
+    // Failures are logged so they are not lost, since the import UI has no
+    // per-item failure state.
+    LOG(ERROR) << "Brave password import failed, result="
+               << static_cast<int>(result);
+  }
+
+  // Continue the remainder of the import lifecycle (extensions, if any, then
+  // the final "ended" notification) by re-entering NotifyImportEnded() now
+  // that `password_importer_` has been cleared.
+  NotifyImportEnded();
 }
 
 void BraveExternalProcessImporterHost::LaunchImportIfReady() {
--- a/browser/importer/brave_password_importer.cc
+++ b/browser/importer/brave_password_importer.cc
@@ -1,0 +1,161 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "brave/browser/importer/brave_password_importer.h"
+
+#include <cstddef>
+#include <utility>
+#include <vector>
+
+#include "base/check.h"
+#include "base/files/file_util.h"
+#include "base/functional/bind.h"
+#include "base/logging.h"
+#include "base/memory/weak_ptr.h"
+#include "base/task/thread_pool.h"
+#include "brave/common/importer/scoped_copy_file.h"
+#include "chrome/browser/browser_process.h"
+#include "chrome/browser/password_manager/factories/profile_password_store_factory.h"
+#include "components/keyed_service/core/service_access_type.h"
+#include "components/os_crypt/async/browser/os_crypt_async.h"
+#include "components/password_manager/core/browser/password_store/login_database.h"
+#include "components/password_manager/core/browser/password_store/password_store_interface.h"
+#include "components/password_manager/core/browser/password_store/stored_credential.h"
+
+namespace {
+
+struct ReadResult {
+  ReadResult() = default;
+  ReadResult(ReadResult&&) = default;
+  ReadResult& operator=(ReadResult&&) = default;
+  ~ReadResult() = default;
+
+  BravePasswordImporter::Result result =
+      BravePasswordImporter::Result::kReadFailed;
+  std::vector<password_manager::StoredCredential> credentials;
+};
+
+// Runs on a background thread. Copies the source `Login Data` to a temporary
+// location (to avoid conflicting with a possibly-running source browser),
+// opens it with the destination's `encryptor`, and returns the outcome and
+// decrypted credentials.
+ReadResult ReadCredentialsFromLoginData(
+    const base::FilePath& login_data_path,
+    scoped_refptr<os_crypt_async::Encryptor> encryptor) {
+  ReadResult read_result;
+  if (!base::PathExists(login_data_path)) {
+    read_result.result = BravePasswordImporter::Result::kLoginDataNotFound;
+    return read_result;
+  }
+
+  ScopedCopyFile copy_login_data(login_data_path);
+  if (!copy_login_data.copy_success()) {
+    LOG(ERROR) << "Failed to copy Login Data for password import";
+    read_result.result = BravePasswordImporter::Result::kCopyFailed;
+    return read_result;
+  }
+
+  password_manager::LoginDatabase database(
+      copy_login_data.copied_file_path(),
+      password_manager::IsAccountStore(false));
+  if (!database.Init(
+          /*on_undecryptable_passwords_removed=*/base::NullCallback(),
+          std::move(encryptor))) {
+    LOG(ERROR) << "LoginDatabase Init() failed for password import";
+    read_result.result = BravePasswordImporter::Result::kDatabaseInitFailed;
+    return read_result;
+  }
+
+  // GetAllLogins returns a FormRetrievalResult, letting us distinguish a
+  // decryption/key failure from an empty database. This is the same call the
+  // password store backend uses.
+  password_manager::FormRetrievalResult retrieval_result =
+      database.GetAllLogins(&read_result.credentials);
+  if (retrieval_result != password_manager::FormRetrievalResult::kSuccess &&
+      retrieval_result != password_manager::FormRetrievalResult::
+                              kEncryptionServiceFailureWithPartialData) {
+    LOG(ERROR) << "Failed to read logins for password import, result="
+               << static_cast<int>(retrieval_result);
+    read_result.credentials.clear();
+    read_result.result = BravePasswordImporter::Result::kReadFailed;
+    return read_result;
+  }
+
+  // PasswordStore::AddLogins CHECKs that blocklisted entries carry no username
+  // or password. A source `Login Data` may violate that, so normalize here.
+  for (auto& credential : read_result.credentials) {
+    if (credential.blocked_by_user) {
+      credential.username_value.clear();
+      credential.password_value.clear();
+    }
+  }
+
+  read_result.result = BravePasswordImporter::Result::kSuccess;
+  return read_result;
+}
+
+}  // namespace
+
+BravePasswordImporter::BravePasswordImporter() = default;
+BravePasswordImporter::~BravePasswordImporter() = default;
+
+void BravePasswordImporter::StartImport(const base::FilePath& source_path,
+                                        Profile* destination_profile,
+                                        CompletionCallback callback) {
+  CHECK(destination_profile);
+  CHECK(callback);
+  source_path_ = source_path;
+  callback_ = std::move(callback);
+  password_store_ = ProfilePasswordStoreFactory::GetForProfile(
+      destination_profile, ServiceAccessType::EXPLICIT_ACCESS);
+  if (!password_store_) {
+    std::move(callback_).Run(Result::kNoPasswordStore, 0);
+    return;
+  }
+
+  auto* os_crypt = g_browser_process->os_crypt_async();
+  if (!os_crypt) {
+    std::move(callback_).Run(Result::kNoEncryptor, 0);
+    return;
+  }
+  os_crypt->GetInstance(base::BindOnce(&BravePasswordImporter::OnEncryptorReady,
+                                       weak_factory_.GetWeakPtr()));
+}
+
+void BravePasswordImporter::OnEncryptorReady(
+    scoped_refptr<os_crypt_async::Encryptor> encryptor) {
+  base::FilePath login_data_path = source_path_.Append(
+      base::FilePath::StringType(FILE_PATH_LITERAL("Login Data")));
+  base::ThreadPool::PostTaskAndReplyWithResult(
+      FROM_HERE, {base::MayBlock(), base::TaskPriority::USER_VISIBLE},
+      base::BindOnce(&ReadCredentialsFromLoginData, login_data_path,
+                     std::move(encryptor)),
+      base::BindOnce(
+          [](base::WeakPtr<BravePasswordImporter> importer,
+             ReadResult read_result) {
+            if (importer) {
+              importer->OnCredentialsRead(read_result.result,
+                                          std::move(read_result.credentials));
+            }
+          },
+          weak_factory_.GetWeakPtr()));
+}
+
+void BravePasswordImporter::OnCredentialsRead(
+    Result result,
+    std::vector<password_manager::StoredCredential> credentials) {
+  if (result != Result::kSuccess || !password_store_) {
+    std::move(callback_).Run(result, 0);
+    return;
+  }
+  if (credentials.empty()) {
+    std::move(callback_).Run(Result::kSuccess, 0);
+    return;
+  }
+  const size_t count = credentials.size();
+  password_store_->AddLogins(
+      std::move(credentials),
+      base::BindOnce(std::move(callback_), Result::kSuccess, count));
+}
--- a/browser/importer/brave_password_importer.cc
+++ b/browser/importer/brave_password_importer.cc
@@ -12,8 +12,9 @@
 #include "base/check.h"
 #include "base/files/file_util.h"
 #include "base/functional/bind.h"
+#include "base/location.h"
 #include "base/logging.h"
-#include "base/memory/weak_ptr.h"
+#include "base/task/sequenced_task_runner.h"
 #include "base/task/thread_pool.h"
 #include "brave/common/importer/scoped_copy_file.h"
 #include "chrome/browser/browser_process.h"
@@ -24,25 +25,16 @@
 #include "components/password_manager/core/browser/password_store/password_store_interface.h"
 #include "components/password_manager/core/browser/password_store/stored_credential.h"
 
-namespace {
+BravePasswordImporter::ReadResult::ReadResult() = default;
+BravePasswordImporter::ReadResult::ReadResult(ReadResult&&) = default;
+BravePasswordImporter::ReadResult& BravePasswordImporter::ReadResult::operator=(
+    ReadResult&&) = default;
+BravePasswordImporter::ReadResult::~ReadResult() = default;
 
-struct ReadResult {
-  ReadResult() = default;
-  ReadResult(ReadResult&&) = default;
-  ReadResult& operator=(ReadResult&&) = default;
-  ~ReadResult() = default;
-
-  BravePasswordImporter::Result result =
-      BravePasswordImporter::Result::kReadFailed;
-  std::vector<password_manager::StoredCredential> credentials;
-};
-
-// Runs on a background thread. Copies the source `Login Data` to a temporary
-// location (to avoid conflicting with a possibly-running source browser),
-// opens it with the destination's `encryptor`, and returns the outcome and
-// decrypted credentials.
-ReadResult ReadCredentialsFromLoginData(
-    const base::FilePath& login_data_path,
+// static
+BravePasswordImporter::ReadResult
+BravePasswordImporter::ReadCredentialsFromLoginData(
+    base::FilePath login_data_path,
     scoped_refptr<os_crypt_async::Encryptor> encryptor) {
   ReadResult read_result;
   if (!base::PathExists(login_data_path)) {
@@ -96,8 +88,6 @@ ReadResult ReadCredentialsFromLoginData(
   return read_result;
 }
 
-}  // namespace
-
 BravePasswordImporter::BravePasswordImporter() = default;
 BravePasswordImporter::~BravePasswordImporter() = default;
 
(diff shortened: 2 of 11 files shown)


bsclifton:
@bbondy does this only handle release channel? 

This could be a great starting point (code-wise) for this long-standing issue we've had 😄 https://github.com/brave/brave-browser/issues/1357

bbondy:
> @bbondy does this only handle release channel?

Yes it uses the already exposed import from Brave for Brave Origin branded builds. But it adds password import since the keyring is shared. 

> This could be a great starting point (code-wise) for this long-standing issue we've had 😄 https://github.com/brave/brave-browser/issues/1357

I might take a look.

utility/importer/chrome_importer.cc
```diff
@@ -230,21 +230,15 @@ void ChromeImporter::LoadFaviconData(
...
-      auto decoded_data = importer::ReencodeFavicon(base::span(data));
-      if (!decoded_data) {
-        continue;  // Unable to decode.
+      std::optional<std::vector<uint8_t>> decoded_data;
+      // Skip favicons with invalid URLs, empty data, or that fail to decode.
+      if (usage.favicon_url.is_valid() && !data.empty() &&
+          (decoded_data = importer::ReencodeFavicon(base::span(data)))) {
+        usage.urls = entry.second;
+        usage.png_data = std::move(decoded_data).value();
+        favicons->push_back(usage);
```
darkdh, line 243: These changes don't seem relevant to this PR

browser/importer/brave_password_importer.h
```diff
@@ -0,0 +1,62 @@
...
+  BravePasswordImporter();
+  ~BravePasswordImporter();
+
+  BravePasswordImporter(const BravePasswordImporter&) = delete;
+  BravePasswordImporter& operator=(const BravePasswordImporter&) = delete;
+
+  // `source_path` is the source Brave profile directory (containing the
+  // `Login Data` file). `destination_profile` receives the imported
+  // credentials.
+  void Start(const base::FilePath& source_path,
```
darkdh: can we name it `StartImport` to be more align with `//chrome/browser/password_manager/startup_passwords_importer.h` and `components/webauthn/core/browser/import/passkey_importer.h`, ... `*importer` etc

browser/importer/brave_password_importer.cc
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  password_manager::LoginDatabase database(
+      copy_login_data.copied_file_path(),
+      password_manager::IsAccountStore(false));
+  if (!database.Init(
+          /*on_undecryptable_passwords_removed=*/base::NullCallback(),
+          std::move(encryptor))) {
+    LOG(ERROR) << "LoginDatabase Init() failed for password import";
+    return credentials;
```
darkdh: did you verify this on Windows? OSCrypt shares the same key in `Brave Safe Storage` on MacOS and `kKWalletFolder = "Brave Keys";  kKeyName = "Brave Safe Storage";  kAppName = "brave"` on Linux but `DPAPIKeyProvider(local_state())` reads `os_crypt.encrypted_key` which is DPAPI encrypted from Brave Origin's own Local State on Windows, so standard Brave has different encrypted key which is only protected by the same DPAPI. The same DPAPI key usage only exists in very old version.

If that is the case, we have to make it MacOS and Linux only

browser/ui/webui/settings/brave_import_data_handler.cc
```diff
@@ -96,22 +99,49 @@ void BraveImportDataHandler::StartImportImpl(
...
 }
 
+void BraveImportDataHandler::OnPasswordImportFinished(
+    base::FilePath source_path,
+    size_t added) {
+  password_importers_.erase(source_path);
+  if (added > 0) {
+    import_did_succeed_ = true;
+  } else {
+    LOG(WARNING) << "Brave password import finished with no credentials added";
```
darkdh: this should not fail silently that users won't notice unless they explicitly check passwords and found out it is empty

browser/importer/brave_password_importer_unittest.cc
```diff
@@ -0,0 +1,162 @@
...
+    base::FilePath login_data_path =
+        source_path_.Append(FILE_PATH_LITERAL("Login Data"));
+    password_manager::LoginDatabase database(
+        login_data_path, password_manager::IsAccountStore(false));
+    ASSERT_TRUE(database.Init(base::NullCallback(), GetEncryptorSync()));
+    for (const auto& cred : credentials) {
+      ASSERT_FALSE(
+          database.AddLogin(password_manager::CloneStoredCredential(cred))
+              .empty());
+    }
```
darkdh, line 111: we don't have test cover key mismatched and that is possibly even on MacOS because ex. a Brave Origin user can delete the keychain item and force it to regen while keeping standard Brave closed, or simply copy the backup profile from other machine

browser/importer/brave_password_importer.cc
```diff
@@ -0,0 +1,116 @@
...
+  if (!database.GetAutofillableLogins(&credentials)) {
+    LOG(ERROR) << "Failed to read autofillable logins for password import";
+    credentials.clear();
+  }
+  std::vector<password_manager::StoredCredential> blocklisted;
+  if (database.GetBlocklistLogins(&blocklisted)) {
+    credentials.insert(credentials.end(),
+                       std::make_move_iterator(blocklisted.begin()),
+                       std::make_move_iterator(blocklisted.end()));
+  }
```
darkdh: why don't we use 
```
  // Gets the complete list of all credentials.
  [[nodiscard]] FormRetrievalResult GetAllLogins(
      std::vector<StoredCredential>* credentials);
```
since we already got the database? And this can propagate the result including error so we can let users know exactly what went wrong instead of guessing.

utility/importer/chrome_importer.cc
bbondy: reverted that change

browser/importer/brave_password_importer.h
bbondy: Done

browser/importer/brave_password_importer.cc
bbondy: now wrapped with OS checks for Mac and Linux. I'll test both imports before merging btw.

bbondy: Mac already tested

browser/ui/webui/settings/brave_import_data_handler.cc
bbondy: We now log the error with ERROR

browser/importer/brave_password_importer_unittest.cc
bbondy: 👍 See `KeyMismatchImportsNothing`

browser/importer/brave_password_importer.cc
bbondy: Fixed

browser/ui/webui/settings/brave_import_data_handler.cc
```diff
@@ -96,22 +99,60 @@ void BraveImportDataHandler::StartImportImpl(
...
+          importer_host, source_profile, utility_items,
           base::BindRepeating(&BraveImportDataHandler::NotifyImportProgress,
                               weak_factory_.GetWeakPtr()));
 
   importer_host->set_parent_window(
       web_ui()->GetWebContents()->GetTopLevelNativeWindow());
   importer_host->set_parent_view(web_ui()->GetWebContents()->GetNativeView());
 
-  importer_host->StartImportSettings(source_profile, profile, imported_items,
+  importer_host->StartImportSettings(source_profile, profile, utility_items,
```
darkdh: There is an edge case that if a user only select password, the user will see import failed because `utility_items` has no items in it but later `OnPasswordImportFinished` will report success and get password imported

browser/ui/webui/settings/brave_import_data_handler.cc
```diff
@@ -96,22 +99,60 @@ void BraveImportDataHandler::StartImportImpl(
...
+  }
+#endif  // BUILDFLAG(IS_MAC) || BUILDFLAG(IS_LINUX)
+
   // Using weak pointers because it destroys itself when finshed.
   auto* importer_host = new BraveExternalProcessImporterHost();
   import_observers_[source_profile.source_path] =
       std::make_unique<BraveImporterObserver>(
-          importer_host, source_profile, imported_items,
+          importer_host, source_profile, utility_items,
           base::BindRepeating(&BraveImportDataHandler::NotifyImportProgress,
```
darkdh, line 106: we should not fire `BraveImportDataHandler::OnImportEnded` when there is password importing running which will make failed password import unnoticed.
Also `BraveImportDataHandler::StartImport` never reset `import_did_succeed_`

browser/ui/webui/settings/brave_import_data_handler.cc
```diff
@@ -96,22 +99,60 @@ void BraveImportDataHandler::StartImportImpl(
 
   FireWebUIListener("import-data-status-changed", base::Value("inProgress"));
 
+  uint16_t utility_items = imported_items;
+#if BUILDFLAG(IS_MAC) || BUILDFLAG(IS_LINUX)
```
darkdh: shouldn't we also make changes to `CanImportPasswordsForType()` otherwise users can still see password checkbox on Windows? 
As we currently let's them pick and ignore it.

browser/importer/brave_password_importer_unittest.cc
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  size_t RunImport() {
+    BravePasswordImporter importer;
+    base::test::TestFuture<BravePasswordImporter::Result, size_t> future;
+    importer.StartImport(source_path_, profile_.get(), future.GetCallback());
+    return future.Get<1>();
+  }
```
darkdh: we should verify the `Result`

browser/importer/brave_password_importer.h
```diff
@@ -0,0 +1,93 @@
...
+
+    Result result = Result::kReadFailed;
+    std::vector<password_manager::StoredCredential> credentials;
+  };
+
+  // Runs on a background thread. Copies the source `Login Data`, opens it with
+  // the destination's `encryptor`, and returns the outcome and credentials.
+  static ReadResult ReadCredentialsFromLoginData(
+      const base::FilePath& login_data_path,
+      scoped_refptr<os_crypt_async::Encryptor> encryptor);
```
darkdh, line 86: this can just be an anonymous free function

browser/importer/brave_password_importer.cc
```diff
@@ -0,0 +1,131 @@
...
+    return;
+  }
+  if (read_result.credentials.empty()) {
+    std::move(callback_).Run(Result::kSuccess, 0);
+    return;
+  }
+  const size_t count = read_result.credentials.size();
+  password_store_->AddLogins(
+      std::move(read_result.credentials),
+      base::BindOnce(std::move(callback_), Result::kSuccess, count));
```
darkdh, line 148: this doesn't really report successfully added login, it is possible that your input credentials won't be added.
So the comment of `number of credentials added` is not accurate, we can just update the comments or if we really want the correct number, we need `PasswordStoreInterface::Observer` and count `PasswordStoreChange::ADD` entries in `OnLoginsChanged(store, changes)`

browser/ui/webui/settings/brave_import_data_handler.cc
darkdh: I think we can reference how extensions import works(which also needs to live in browser)
```
  void BraveExternalProcessImporterHost::NotifyImportEnded() {
    if (!cancelled_) { RecordImporterP3A(source_profile_.importer_type); }
  #if BUILDFLAG(ENABLE_EXTENSIONS)
    if (NeedToImportExtensions() && extensions_importer_) {
      NotifyImportItemStarted(user_data_importer::EXTENSIONS);
      if (extensions_importer_->Import(base::BindRepeating(
              &BraveExternalProcessImporterHost::OnExtensionImported,
              weak_ptr_factory_.GetWeakPtr()))) {
        return;    // <-- swallow the utility import's "ended"
      }
    }
  #endif
    ExternalProcessImporterHost::NotifyImportEnded();
  }
```
so we can have `NotifyImportItemStarted/Ended(PASSWRODS)`

darkdh: We also need to add something like
```
bool BraveExternalProcessImporterHost::NeedToImportExtensions() const {
  return !cancelled_ && (items_ & user_data_importer::EXTENSIONS) ==
                            user_data_importer::EXTENSIONS;
}
``` 
to check cancellation, closing the setting page won't stop it from keeping running

browser/ui/webui/settings/brave_import_data_handler.cc
```diff
@@ -96,22 +99,60 @@ void BraveImportDataHandler::StartImportImpl(
...
+  // cannot decrypt the source's `Login Data`.
+  if ((imported_items & user_data_importer::PASSWORDS) &&
+      source_profile.importer_type == user_data_importer::TYPE_BRAVE) {
+    utility_items &= ~user_data_importer::PASSWORDS;
+    password_importers_[source_profile.source_path] =
+        std::make_unique<BravePasswordImporter>();
+    password_importers_[source_profile.source_path]->StartImport(
+        source_profile.source_path, profile,
+        base::BindOnce(&BraveImportDataHandler::OnPasswordImportFinished,
+                       weak_factory_.GetWeakPtr(), source_profile.source_path));
```
darkdh: This will have problem if a user import password while another password importing is on the fly, the second import would fail and show success for the first one.
```
  t0  user starts import of Brave profile P
      password_importers_[P] = A
      A->StartImport(P, …, cb_A)          cb_A = {handler_weak, P}
      A: OSCryptAsync::GetInstance()      → async

  t1  encryptor ready → ThreadPool read   → async

  t2  A::OnCredentialsRead
      AddLogins(creds, BindOnce(std::move(cb_A), kSuccess, N))
      ↑ cb_A now owned by the PasswordStore barrier. A holds nothing.

  t3  user clicks Import again for P (A still in flight)
      StartImportImpl → password_importers_[P] = make_unique<B>()
      ↑ unique_ptr move-assign destroys A. cb_A is untouched.
      B->StartImport(P, …, cb_B)          → async

  t4  A's AddLogins completion fires → cb_A runs
      handler_weak still valid → OnPasswordImportFinished(P, kSuccess, N)
        password_importers_.erase(P);     ← erases B
        import_did_succeed_ = true;

  t5  B's pending reply arrives → B's WeakPtr invalidated → dropped
```

darkdh: I think it would be simpler if we can just let `BraveExternalProcessImporterHost` owns `password_importer_` like it owns `extensions_importer_` because it is created per-import and self-destructs so we don't need to manage this kind of map.

browser/importer/brave_password_importer.cc
bbondy: Updated the comment only for simplicity

browser/importer/brave_password_importer.h
bbondy: Done

browser/importer/brave_password_importer_unittest.cc
bbondy: Done in each test

browser/ui/webui/settings/brave_import_data_handler.cc
bbondy: Done although tbh I don't ever see checkboxes for these things, it just import everything from my testing.  At least from brave://welcome

browser/ui/webui/settings/brave_import_data_handler.cc
bbondy: Fixed by moving the importer into `BraveExternalProcessImporterHost`, which owns `password_importer_` per-import and self-destructs

browser/ui/webui/settings/brave_import_data_handler.cc
bbondy: Thanks. Done.

browser/ui/webui/settings/brave_import_data_handler.cc
bbondy: Fixed by the lifecycle changes

bbondy:
@darkdh 
I tested import for macOS and Linux (Ubuntu) and it works. 
I also tested import for Windows and it has no visible regressions (doesn't import passwords as intended in this PR)

browser/ui/webui/settings/brave_import_data_handler.cc
darkdh: Screenshot 2026-08-07 at 08 58 10
Screenshot 2026-08-07 at 08 59 07


can you skip first time import and test import in this section?

darkdh: also from menu
Screenshot 2026-08-07 at 09 00 26
it will open the same dialog in the setting

bbondy: ok tested all platforms from that way as well (previously I did via brave://welcome which just imports everything). Windows correctly doesn’t show passwords from that screen.
Linux and macOS correctly import only when selected

common/importer/chrome_importer_utils.cc
```diff
@@ -149,6 +149,16 @@ bool CanImportPasswordsForType(user_data_importer::ImporterType type) {
   }
 #endif
 
+  // Brave-to-Brave password import relies on both installations sharing the
+  // same OS keychain entry, which only holds on macOS and Linux. On Windows
+  // each installation stores its own key in its own Local State, so the
```
darkdh: we can elaborate the key in Local State is encrypted by DPAPI so people would know DPAPI is still in use but not the way it used to be without diving into `BravePasswordImporter`

browser/ui/webui/settings/brave_import_data_handler.cc
```diff
@@ -94,8 +96,13 @@ void BraveImportDataHandler::StartImportImpl(
     import_observers_.erase(source_profile.source_path);
   }
 
+  import_did_succeed_ = false;
+
   FireWebUIListener("import-data-status-changed", base::Value("inProgress"));
 
+  // Password import (for Brave-to-Brave imports on macOS and Linux) is handled
+  // by BraveExternalProcessImporterHost in the browser process, alongside the
+  // utility-process import, so it participates in the same import lifecycle.
```
darkdh: I don't think this comment is needed. It is here because AI add it to reason about moving from owning password importer here to `BraveExternalProcessImporterHost`. 
But for people later see this PR, they don't need to know this progress because the initial attempt was wrong

browser/importer/brave_password_importer.cc
```diff
@@ -0,0 +1,152 @@
...
+  if (result != Result::kSuccess || !password_store_) {
+    std::move(callback_).Run(result, 0);
+    return;
+  }
+  if (credentials.empty()) {
+    std::move(callback_).Run(Result::kSuccess, 0);
+    return;
+  }
+  const size_t count = credentials.size();
+  password_store_->AddLogins(
```
darkdh, line 146: https://source.chromium.org/chromium/chromium/src/+/main:components/password_manager/core/browser/password_store/password_store.cc;l=126;drc=3603db4c7b3a3fe4f579e989d319427d6807d93a
because of this CHECK, I think we need to make sure blocklist item has no username and password in `ReadCredentialsFromLoginData`

We can do normalization like this
```
    for (auto& credential : read_result.credentials) {
      if (credential.blocked_by_user) {
        credential.username_value.clear();
        credential.password_value.clear();
      }
    }
```
Of course a unit test case to intentionally write a block list with either non empty username or password, after import they should be empty

browser/importer/brave_external_process_importer_host.cc
```diff
@@ -42,6 +69,42 @@ void BraveExternalProcessImporterHost::NotifyImportEnded() {
...
+    // A successful import of zero credentials is not an error; the source
+    // simply had no saved passwords. Genuine failures are logged so they are
+    // not lost, since the import UI has no per-item failure state.
+    LOG(ERROR) << "Brave password import failed, result="
+               << static_cast<int>(result);
+  }
+
+  if (observer_) {
+    NotifyImportItemEnded(user_data_importer::PASSWORDS);
+  }
```
darkdh: we should only call this upon success otherwise importing password failure would still report success
Also I don't see a reason to check `observer_` as `NotifyImportItemEnded` already does null check

browser/importer/brave_external_process_importer_host.cc
```diff
@@ -5,21 +5,48 @@
...
 BraveExternalProcessImporterHost::BraveExternalProcessImporterHost()
     : weak_ptr_factory_(this) {}
 BraveExternalProcessImporterHost::~BraveExternalProcessImporterHost() = default;
 
 void BraveExternalProcessImporterHost::NotifyImportEnded() {
-  if (!cancelled_) {
+  // Guard against re-recording: OnPasswordsImported() re-enters this method to
+  // continue the lifecycle after password import, and P3A must be recorded
+  // only once per import.
+  if (!cancelled_ && !p3a_recorded_) {
```
darkdh: I don't think this is password import specific, extension import has the same flow

browser/importer/brave_external_process_importer_host.cc
```diff
@@ -5,21 +5,48 @@
...
   }
 
+  // Passwords are imported in the browser process after the utility import has
+  // finished. Like extensions below, this swallows the utility import's
+  // "ended" notification until password import completes so a password failure
+  // is not reported as an overall success. `password_importer_` is created
+  // lazily here and reset in OnPasswordsImported(), which re-enters this
+  // method to continue the lifecycle.
+  if (NeedToImportPasswords() && !passwords_imported_) {
+    passwords_imported_ = true;
```
darkdh: The naming is kind of misleading, can we make it `password_import_started_`?

browser/importer/brave_password_importer_unittest.cc
```diff
@@ -0,0 +1,198 @@
...
+  // A fresh test Encryptor uses a random key distinct from the destination's.
+  ASSERT_NO_FATAL_FAILURE(WriteLoginData(
+      credentials, os_crypt_async::GetTestEncryptorForTesting()));
+
+  auto [result, submitted] = RunImport();
+  // The exact Result varies by platform: where undecryptable logins are
+  // dropped the read reports kSuccess with nothing decrypted, otherwise it
+  // reports kReadFailed. Either way, no credential may be submitted or leak.
+  EXPECT_TRUE(result == BravePasswordImporter::Result::kReadFailed ||
+              result == BravePasswordImporter::Result::kSuccess)
```
darkdh: can we use buildflag to know which platforms return `kSuccess` and which return  `kReadFailed`?

common/importer/chrome_importer_utils.cc
bbondy: Expanded the comment

browser/ui/webui/settings/brave_import_data_handler.cc
bbondy: Removed

browser/importer/brave_password_importer.cc
bbondy: Fixed

browser/importer/brave_external_process_importer_host.cc
bbondy: Only fires on success now

browser/importer/brave_external_process_importer_host.cc
bbondy: described the browser-process/OSCryptAsync reason

browser/importer/brave_external_process_importer_host.cc
bbondy: done

browser/importer/brave_password_importer_unittest.cc
bbondy: Done

browser/ui/webui/settings/brave_import_data_handler.cc
```diff
@@ -15,6 +15,8 @@
 #include "chrome/browser/importer/importer_list.h"
 #include "chrome/browser/importer/profile_writer.h"
 #include "chrome/browser/profiles/profile.h"
+#include "components/user_data_importer/common/importer_data_types.h"
+#include "components/user_data_importer/common/importer_type.h"
```
darkdh: are these headers still needed?

darkdh (approved):

browser/importer/brave_password_importer.cc
```diff
@@ -0,0 +1,161 @@
...
+  return read_result;
+}
+
+}  // namespace
+
+BravePasswordImporter::BravePasswordImporter() = default;
+BravePasswordImporter::~BravePasswordImporter() = default;
+
+void BravePasswordImporter::StartImport(const base::FilePath& source_path,
+                                        Profile* destination_profile,
```
netzenbot, line 95: These two early-failure paths run `callback_` synchronously while the success path is async. Since `OnPasswordsImported()` does `password_importer_.reset()` and then re-enters `NotifyImportEnded()`, a sync callback destroys this importer while `StartImport()` is still on its stack, and `ExternalProcessImporterHost::NotifyImportEnded()` (`delete this`) then runs with two host frames still live. Please post the failure callbacks via `SequencedTaskRunner::GetCurrentDefault()->PostTask()` so completion is always async ([ARCH-060](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-060)).

browser/importer/brave_password_importer.cc
```diff
@@ -0,0 +1,161 @@
...
+  auto* os_crypt = g_browser_process->os_crypt_async();
+  if (!os_crypt) {
+    std::move(callback_).Run(Result::kNoEncryptor, 0);
+    return;
+  }
+  os_crypt->GetInstance(base::BindOnce(&BravePasswordImporter::OnEncryptorReady,
+                                       weak_factory_.GetWeakPtr()));
+}
+
+void BravePasswordImporter::OnEncryptorReady(
```
netzenbot, line 124: Binding a `WeakPtr` into a lambda and null-checking it by hand is the pattern [CSM-009](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-memory.md#CSM-009) calls out; binding directly to a member function gets the cancellation for free. Moving `ReadResult` into the header (or into the class as a private struct) lets this be `base::BindOnce(&BravePasswordImporter::OnCredentialsRead, weak_factory_.GetWeakPtr())` with the member taking the whole `ReadResult`.

browser/importer/brave_external_process_importer_host.cc
```diff
@@ -42,6 +67,42 @@ void BraveExternalProcessImporterHost::NotifyImportEnded() {
   ExternalProcessImporterHost::NotifyImportEnded();
 }
 
+bool BraveExternalProcessImporterHost::NeedToImportPasswords() const {
+#if BUILDFLAG(IS_MAC) || BUILDFLAG(IS_LINUX)
+  // Only Brave-to-Brave imports share the OS keychain entry needed to decrypt
+  // the source's `Login Data`.
```
netzenbot, line 73: The new lifecycle logic here (re-entering `NotifyImportEnded()`, the `p3a_recorded_` guard, and the PASSWORDS item start/end notifications) isn't covered by any test — `brave_external_process_importer_host_unittest.cc` only exercises the extensions path. That file already has the harness for this (`DoNotLaunchImportForTesting()` + `ImportEndedObserver` + `StartImportSettings`), so a case with `user_data_importer::PASSWORDS` asserting `ImportEnded` fires exactly once and P3A is recorded once would be cheap to add. ([✅ PRs Should Include Reasonable Test Coverage](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-039))

browser/ui/webui/settings/brave_import_data_handler.cc
bbondy: no they were removed

netzenbot (approved):

PR #36727, opened 2026-05-26
[iOS] Handle all executions errors from simulator test runner summary

kylehickinson:
The `step_text` key is only ever set if an exception was raised while executing the iOS simulator test runners, we should ensure that all exceptions result in a failure to catch issues such as `AppLaunchError` which occurs when the test suite crashes and fails to run any tests

### DIFF
--- a/tools/test/chromium_to_junit_converter.py
+++ b/tools/test/chromium_to_junit_converter.py
@@ -52,8 +52,7 @@ class ChromiumToJUnitConverter:
         # Calculate totals
         total_passed = len(passed_tests)
         total_failed = len(failed_tests)
-        has_execution_error = bool(step_text
-                                   and 'TEST EXECUTE FAILED' in step_text)
+        has_execution_error = bool(step_text)
         total_tests = total_passed + total_failed + (1 if has_execution_error
                                                      else 0)
 


tools/test/chromium_to_junit_converter.py
```diff
@@ -52,8 +52,7 @@ def _convert_ios_summary(self, json_data: Dict) -> str:
         # Calculate totals
         total_passed = len(passed_tests)
         total_failed = len(failed_tests)
-        has_execution_error = bool(step_text
-                                   and 'TEST EXECUTE FAILED' in step_text)
+        has_execution_error = bool(step_text)
```
netzenbot, line 55: This changes the execution error detection logic (from a specific string check to any non-empty `step_text`), but there are no tests for this converter at all. A test covering both the original behavior and the new broader detection would prevent regressions. https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-039

bridiver (approved):

tools/test/chromium_to_junit_converter.py
kylehickinson: out of scope for this PR

PR #36733, opened 2026-05-26
[Origin] P3A reports data when disabled by policy at startup

### LINKED ISSUE #55838: [Origin] P3A submits metrics despite BraveP3AEnabled policy at startup
`P3AService::Init()` reads `kP3AEnabled` synchronously and may call `message_manager_->Start()` before `BraveOriginPolicyManager` has merged its managed prefs into local state. Result: a window where the
collector receives uploads even when `BraveP3AEnabled` is policy-disabled.

SergeyZhukovsky:
`P3AService::Init()` reads `kP3AEnabled` and may call `message_manager_->Start()` before `BraveOriginPolicyManager` has merged its managed prefs into local state, so constellation uploads can fire even when `BraveP3AEnabled` is policy-disabled. Make `P3AService` observe `BraveOriginPolicyManager`, split `Init` into `Init` + `TryFinalizeInit`, and gate the message-manager start on a `brave_policies_ready_` flag set by `OnBravePoliciesReady` (posted so the async pref-propagation chain has landed first).

Resolves: https://github.com/brave/brave-browser/issues/55838


components/p3a/p3a_service.cc
```diff
@@ -217,6 +234,30 @@ void P3AService::Init(
   }
 }
 
+void P3AService::OnBravePoliciesReady() {
```
bridiver, line 237: I don't think P3AService should need to know anything about brave origin policies and also this could be disabled by policy without brave origin. Need to dig in a little bit more to see what should happen here because nothing should be accessing prefs before policies are loaded in general. I think this is a side effect of the async issue with how brave origin policies work and we need to fix that or we're going to be playing whack-a-mole here.

SergeyZhukovsky:
Closing as after discussing in DMs we actually send anything in ~1minute after the start and the checks are respected already there.

PR #36751, opened 2026-05-27
[AI Chat] Add search support for NEAR E2EE

### LINKED ISSUE #55876: [AI Chat] Add search support for NEAR E2EE
Add support for the `web_context_search` tool which is handled by the agent in the TEE.

https://github.com/brave/aichat/issues/1493

DJAndries:
Resolves https://github.com/brave/aichat/issues/1493
Resolves https://github.com/brave/brave-browser/issues/55876

## Test plan

Prompt NEAR GLM 5.1 with "weather today london". Ensure it performs a web search, and outputs weather information.

### DIFF
--- a/components/ai_chat/core/common/mojom/ai_chat.mojom
+++ b/components/ai_chat/core/common/mojom/ai_chat.mojom
@@ -39,6 +39,8 @@ const string kBraveNewsSearchToolName = "brave_news_search";
 const string kBraveWebSearchToolName = "brave_web_search";
 const string kBraveFaqsSearchToolName = "brave_faqs_search";
 const string kDeepResearchToolName = "deep_research";
+// Currently used for NEAR E2EE only
+const string kWebContextSearchToolName = "web_context_search";
 
 // Tool artifact types
 const string kLineChartArtifactType = "line_chart";
--- a/components/ai_chat/core/browser/engine/oblivious_http_api_client.cc
+++ b/components/ai_chat/core/browser/engine/oblivious_http_api_client.cc
@@ -9,8 +9,10 @@
 
 #include "base/containers/flat_map.h"
 #include "base/functional/bind.h"
+#include "base/strings/string_util.h"
 #include "base/task/thread_pool.h"
 #include "base/types/expected.h"
+#include "brave/components/ai_chat/core/browser/engine/oai_parsing.h"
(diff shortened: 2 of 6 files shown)


netzenbot (approved):

components/ai_chat/core/common/mojom/ai_chat.mojom
```diff
@@ -39,6 +39,8 @@ const string kBraveNewsSearchToolName = "brave_news_search";
 const string kBraveWebSearchToolName = "brave_web_search";
 const string kBraveFaqsSearchToolName = "brave_faqs_search";
 const string kDeepResearchToolName = "deep_research";
+// Currently used for NEAR E2EE only
+const string kWebContextSearchToolName = "web_context_search";
```
fallaciousreasoning, line 43: Why do we need a separate string for this? Can we not use the kBraveNewsSearchToolName?

DJAndries: The name of the tool that the NEAR server provides is `web_context_search`. We could use one of the existing names here, but then we'd have to convert between the tool names in requests/responses. If you think that makes sense, I can change it. Also, what makes `brave_news_search` a better fit compared to `brave_web_search`?

fallaciousreasoning: whoops sorry, yeah I meant `kBraveWebSearchToolName` but copied the wrong name. Do we not have control over what comes back from the server (i.e. is that tool name controlled by NEAR?)?

fallaciousreasoning (approved):
lgtm + question

components/ai_chat/core/common/mojom/ai_chat.mojom
fallaciousreasoning: If they do control it, I'm not sure what the best approach is - might be nice to convert it to something we already support but I guess it isn't necessarily supported. Happy to defer to you 😆

DJAndries: They do control it. I tried converting the tool call to `brave_web_search` but the changes ended up being more involved since the argument structure is different. I'm not sure if it's worth it to do the conversion, but if you disagree please lmk and I'll be happy to follow up.

PR #36757, opened 2026-05-27
[VPN 2.0] Split VPN utils into shared and V1-only targets

### LINKED ISSUE #54597: Implement a stub BraveVpnService v2

alekspop26:

This incremental change splits "brave_vpn" target into two:
- "brave_vpn" itself, shared between V1 and V2 implementations.
- "v1_utils", a strictly V1 target, currently containing the VPN connection manager utility code for Architecture 1.0.

Most existing targets will continue depending on the shared "brave_vpn"; "v1_utils" is used only in the specific place where V1 connection manager creation is done, and will be later gated by a buildflag.

### DIFF
--- a/browser/brave_vpn/vpn_connection_manager_utils.cc
+++ b/browser/brave_vpn/vpn_connection_manager_utils.cc
@@ -1,0 +1,82 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "brave/browser/brave_vpn/vpn_connection_manager_utils.h"
+
+#include "base/functional/bind.h"
+#include "base/memory/scoped_refptr.h"
+#include "base/notreached.h"
+#include "brave/components/brave_vpn/browser/connection/brave_vpn_connection_manager.h"
+#include "brave/components/brave_vpn/common/brave_vpn_utils.h"
+#include "build/build_config.h"
+#include "chrome/common/channel_info.h"
+#include "components/prefs/pref_service.h"
+#include "components/user_prefs/user_prefs.h"
+#include "services/network/public/cpp/shared_url_loader_factory.h"
+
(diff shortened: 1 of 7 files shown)


browser/brave_vpn/vpn_connection_manager_utils.cc
```diff
@@ -0,0 +1,82 @@
...
+
+}  // namespace
+
+std::unique_ptr<BraveVPNConnectionManager> CreateBraveVPNConnectionManager(
+    scoped_refptr<network::SharedURLLoaderFactory> url_loader_factory,
+    PrefService* local_prefs) {
+#if BUILDFLAG(IS_ANDROID)
+  // Android doesn't use connection api.
+  return nullptr;
+#else
```
cdesouza-chromium, line 60: I know you are just moving code, so this line of inquiry can be pursued in a follow-up if you prefer.

I wonder if it would be better to just not build these sources on android. If they can't be used on android, I find difficult to think that allowing people to call these functions will produce any good outcome.

If the plan long term is to never support these v1 utilities in android, as would go as far as putting this on the header:
```cxx
#include "build/build_config.h"

#if BUILDFLAG(IS_ANDROID)
#error "Not supported for android."
#endif
```

The follow-up question I have if all this holds, is if v1 is ever supported for android?

alekspop26: Thanks for the feedback! Currently, `CreateBraveVPNConnectionManager` gets called on Android, which is certainly useless (and conceptually wrong). However, the tenet is that **we're not refactoring V1 code since it's going to be dead once V2 is done** - so I'd prefer to keep this as is, but do the right things for V2. Even adding an `#error` in the header would require some refactoring of the calling site (buildflag checks, etc.).

Yes, v1 is supported for Android, but partially - the connection manager lives in the Android code, BraveVpnService lives in the browser code, and gets accessed via an Android native worker.
GN flag is defined as:
```
enable_brave_vpn =
      (is_win || is_android || is_mac || is_ios) && !is_brave_origin_branded
```

cdesouza-chromium (approved):
LGTM

netzenbot (approved):

PR #36760, opened 2026-05-27
[toolchain] Disabling unused Mac toolchain downloads

cdesouza-chromium:
This PR disables the hook for downloading the Mac toolchain, that can be
only used by Google employees, as that hook own output was quit
confusing, as it preceeded our own hook being run.

We are also disabling downloading the hermetic toolchain for ios
checkouts. IOS builds do not use the hermetic tooolchain, and recently
we've run into significant issues in CI relating to that.

https://github.com/brave/brave-browser/issues/55812

### DIFF
--- a/rewrite/DEPS.yaml
+++ b/rewrite/DEPS.yaml
@@ -1,0 +1,17 @@
+# Copyright (c) 2026 The Brave Authors. All rights reserved.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+substitutions:
+  - description: |
+      Disable the mac_toolchain hook by neutralising its condition.
+
+      Brave has its own script to download the hermetic toolchain,
+      brave/build/mac/download_hermetic_xcode.py, so it is okay to disable the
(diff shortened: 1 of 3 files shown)


rewrite/DEPS.yaml
```diff
@@ -0,0 +1,15 @@
...
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+substitutions:
+  - description: |
+      Disable the mac_toolchain hook by neutralising its condition.
+
+      Brave has its own script to download the hermetic toolchain,
+      brave/build/mac/download_hermetic_xcode.py, so it is okay to disable the
+      upstream hook as it causes confusing output to be printed to the user.
+    re_pattern: "('name': 'mac_toolchain',.+?'condition':\\s*)'checkout_mac or checkout_ios'"
```
goodov: let's go with `: |-` syntax? it's a bit cleaner and easier to read (no need to escape `\`), also no need to wrap with quotes:
```
re_pattern: |-
  ('name': 'mac_toolchain',.+?'condition':\s*)'checkout_mac or checkout_ios'
replace: |-
  \1'False'
```

goodov (approved):

rewrite/DEPS.yaml
cdesouza-chromium: Done.

netzenbot (approved):

PR #36762, opened 2026-05-27
[farbling] Add WebGLExtensionHandler to handle extensions related calls

r0hit0303:
This change adds a dedicated `WebGLExtensionHandler` handler to handle farbling around WebGL's `getSupportedExtensions` and `getExtension` calls. The  handler is instantiated taking into account different levels of fingerprinting protections as follows:

#### 1) Balanced fingerprinting protections:
  - [WebGLBalancedFingerprintingProtection](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/chromium_src/third_party/blink/common/features.cc?L63) **enabled**:  calls to `getSupportedExtensions` will injects a fake extension to the list and will keep this new list consistent till the lifetime of that handler. Calls to getExtension on the fake extension would return a fake `ScriptObject` with the name of the fake extension in CamelCase.
  - [WebGLBalancedFingerprintingProtection](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/chromium_src/third_party/blink/common/features.cc?L63) **disabled**: No change in behaviour i.e we don't farble and return the true list of supported extensions. 
  
 #### 2) Maximum fingerprinting protections:
 - No change in behaviour i.e we continue to return a single entry for the getSupportedExtensions call viz. WEBGL_debug_renderer_info. 
 
 #### 3) No fingerprinting protections:
 - No change in behaviour. We return the true list of supported extensions. 

Please note that this is part of the change from the bigger [PR](https://github.com/brave/brave-core/pull/36605). The `WebGLExtensionHandler` would later be created by [BraveSessionCache](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/third_party/blink/renderer/core/farbling/brave_session_cache.h). This integration with `BraveSessionCache` and then with `WebGLRenderingContextBase` would be followed in later CLs.  

This change also adds unit_tests for the same. Browser tests would be followed up later once we have the full integration complete. 

Related https://github.com/brave/brave-browser/issues/55858

### DIFF
--- a/third_party/blink/renderer/bindings/core/webgl/webgl_extension_handler.cc
+++ b/third_party/blink/renderer/bindings/core/webgl/webgl_extension_handler.cc
@@ -1,0 +1,161 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "brave/third_party/blink/renderer/bindings/core/webgl/webgl_extension_handler.h"
+
+#include <array>
+#include <optional>
+
+#include "base/check.h"
+#include "base/check_is_test.h"
+#include "base/feature_list.h"
+#include "base/no_destructor.h"
+#include "third_party/blink/public/common/features.h"
+#include "third_party/blink/renderer/bindings/core/v8/script_value.h"
+#include "third_party/blink/renderer/platform/bindings/v8_binding.h"
+#include "third_party/blink/renderer/platform/wtf/text/ascii_ctype.h"
+#include "third_party/blink/renderer/platform/wtf/text/string_builder.h"
+#include "third_party/blink/renderer/platform/wtf/text/wtf_string.h"
+
+namespace blink {
+
+namespace {
+// The below *fake* extension names are inspired from existing names (see
+// https://registry.khronos.org/webgl/extensions/). This fake list serves to
+// defuse any automated fingerprinting scripts that relies on a hash based
+// approach. Through farbling, we will inject one of the fake values below to
+// the result of the getSupportedExtensions call which would be consistent
+// across that eTLD+1 session. A different session would pseudo-randomly yield a
+// different list resulting in a different hash. Hence the two call sites can't
+// link it's the same user via the hash based approach.
+//
+// The list below is a flat representation of the two sets {texture, expanded,
+// polygon, circle, triangle, blend, draw} and {sampler, blender, compressor}.
+const std::array<String, kFakeExtensionsSize> GetFakeSupportedExtensions() {
+  static const base::NoDestructor<std::array<String, kFakeExtensionsSize>>
+      fake_values({{
+          {"EXT_texture_sampler"},     {"EXT_texture_compressor"},
+          {"EXT_texture_blender"},     {"EXT_expanded_sampler"},
+          {"EXT_expanded_compressor"}, {"EXT_expanded_blender"},
+          {"EXT_polygon_sampler"},     {"EXT_polygon_compressor"},
+          {"EXT_polygon_blender"},     {"EXT_circle_sampler"},
+          {"EXT_circle_compressor"},   {"EXT_circle_blender"},
+          {"EXT_triangle_sampler"},    {"EXT_triangle_compressor"},
+          {"EXT_triangle_blender"},    {"EXT_blend_sampler"},
+          {"EXT_blend_compressor"},    {"EXT_blend_blender"},
+          {"EXT_draw_sampler"},        {"EXT_draw_compressor"},
+          {"EXT_draw_blender"},
+      }});
+  return *fake_values;
+}
+
+// Converts an extension name like "EXT_draw_compressor" to a CamelCase class
+// name like "ExtDrawCompressor".
+String ToExtensionClassName(const String& name) {
+  StringBuilder result;
+  bool capitalize_next = true;
+  for (unsigned i = 0; i < name.length(); ++i) {
+    const UChar c = name[i];
+    if (c == '_') {
+      capitalize_next = true;
+    } else if (capitalize_next) {
+      result.Append(static_cast<UChar>(ToAsciiUpper(c)));
+      capitalize_next = false;
+    } else {
+      result.Append(static_cast<UChar>(ToAsciiLower(c)));
+    }
+  }
+  return result.ToString();
+}
+
+}  // namespace
+
+WebGLExtensionHandler::WebGLExtensionHandler(
+    const ExtensionVector& supported_extensions,
+    std::optional<size_t> fake_index)
+    : supported_extensions_(supported_extensions), fake_index_(fake_index) {}
+
+WebGLExtensionHandler::~WebGLExtensionHandler() = default;
+
+ExtensionVector WebGLExtensionHandler::GetSupportedExtensions() const {
+  return supported_extensions_;
+}
+
+// TODO(https://github.com/brave/brave-browser/issues/55858): Cover testing this
+// in browser_tests as it's not straightforward to test it as unittest due to
+// various v8 dependencies.
+ScriptObject WebGLExtensionHandler::GetExtension(
+    ScriptState* script_state,
+    const String& name,
+    const ScriptObject* real_extension) {
+  // Create and return a fake object if the extension was farbled.
+  if (IsExtensionFarbled(name)) {
+    v8::Isolate* isolate = script_state->GetIsolate();
+    v8::Local<v8::Context> context = script_state->GetContext();
+    v8::Local<v8::FunctionTemplate> tmpl = v8::FunctionTemplate::New(isolate);
+    tmpl->SetClassName(V8String(isolate, ToExtensionClassName(name)));
+    v8::Local<v8::Object> obj = tmpl->GetFunction(context)
+                                    .ToLocalChecked()
+                                    ->NewInstance(context)
+                                    .ToLocalChecked();
+    return ScriptObject(isolate, obj);
+  }
+
+  // If extension is part of the supported list return the real extension.
+  if (supported_extensions_.Contains(name)) {
+    CHECK(real_extension);
+    return *real_extension;
+  }
+
+  return ScriptObject::CreateNull(script_state->GetIsolate());
+}
+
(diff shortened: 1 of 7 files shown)


r0hit0303:
@goodov Could you ptal at this change? Thanks! 

(Will resolve https://github.com/brave/brave-core/pull/36762#issuecomment-4556532250 in the next commit after your review.)

third_party/blink/renderer/bindings/core/webgl/webgl_extension_handler.cc
```diff
@@ -0,0 +1,161 @@
...
+// defuse any automated fingerprinting scripts that relies on a hash based
+// approach. Through farbling, we will inject one of the fake values below to
+// the result of the getSupportedExtensions call which would be consistent
+// across that eTLD+1 session. A different session would pseudo-randomly yield a
+// different list resulting in a different hash. Hence the two call sites can't
+// link it's the same user via the hash based approach.
+//
+// The list below is a flat representation of the two sets {texture, expanded,
+// polygon, circle, triangle, blend, draw} and {sampler, blender, compressor}.
+const std::array<String, kFakeExtensionsSize> GetFakeSupportedExtensions() {
```
goodov: `const std::array<String, kFakeExtensionsSize>&`

third_party/blink/renderer/bindings/core/webgl/webgl_extension_handler.cc
```diff
@@ -0,0 +1,161 @@
...
+ScriptObject WebGLExtensionHandler::GetExtension(
+    ScriptState* script_state,
+    const String& name,
+    const ScriptObject* real_extension) {
+  // Create and return a fake object if the extension was farbled.
+  if (IsExtensionFarbled(name)) {
+    v8::Isolate* isolate = script_state->GetIsolate();
+    v8::Local<v8::Context> context = script_state->GetContext();
+    v8::Local<v8::FunctionTemplate> tmpl = v8::FunctionTemplate::New(isolate);
+    tmpl->SetClassName(V8String(isolate, ToExtensionClassName(name)));
```
goodov: consider pre-calculating this into a similar array you have in `GetFakeSupportedExtensions`.

third_party/blink/renderer/bindings/core/webgl/webgl_extension_handler.cc
r0hit0303: Good catch. I thought RVO / NRVO with copy elision would kick in but forgot that it doesn't work for member static variables. Done

third_party/blink/renderer/bindings/core/webgl/webgl_extension_handler.cc
r0hit0303: Done

third_party/blink/renderer/bindings/core/webgl/webgl_extension_handler.cc
```diff
@@ -0,0 +1,151 @@
...
+    std::optional<FakeExtensionType> fake_extension)
+    : supported_extensions_(supported_extensions),
+      fake_extension_(fake_extension) {}
+
+WebGLExtensionHandler::~WebGLExtensionHandler() = default;
+
+ExtensionVector WebGLExtensionHandler::GetSupportedExtensions() const {
+  return supported_extensions_;
+}
+
```
netzenbot, line 80: nit: `ascii_ctype.h` and `string_builder.h` appear to be unused includes — no `IsASCII*` functions or `StringBuilder` are used anywhere in this file. Per [CS-001](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-001), remove includes you don't actually use.

third_party/blink/renderer/bindings/core/webgl/webgl_extension_handler.cc
```diff
@@ -0,0 +1,151 @@
...
+// TODO(https://github.com/brave/brave-browser/issues/55858): Cover testing this
+// in browser_tests as it's not straightforward to test it as unittest due to
+// various v8 dependencies.
+ScriptObject WebGLExtensionHandler::GetExtension(
+    ScriptState* script_state,
+    const String& name,
+    const ScriptObject* real_extension) {
+  // Create and return a fake object if the extension was farbled.
```
netzenbot, line 88: nit: Avoid positional terms like "below" in comments — another developer could insert code between this comment and the array, breaking the reference. Consider something like "The following fake extension names..." or "The `fake_values` array below contains..." with an explicit name instead. See [CS-024](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-024).

third_party/blink/renderer/bindings/core/webgl/webgl_extension_handler.h
```diff
@@ -0,0 +1,92 @@
...
+// |real_extensions| is the set of actual supported WebGL extensions.
+CORE_EXPORT std::unique_ptr<WebGLExtensionHandler> CreateMaximumHandler(
+    const ExtensionVector& real_extensions);
+
+// A test only method to fetch the list of fake extensions.
+CORE_EXPORT std::array<FakeExtensionType, kFakeExtensionsSize>
+GetFakeSupportedExtensionsForTesting();
+}  // namespace blink
+
+#endif  // BRAVE_THIRD_PARTY_BLINK_RENDERER_BINDINGS_CORE_WEBGL_WEBGL_EXTENSION_HANDLER_H_
```
netzenbot, line 94: nit: The doc comment for `GetExtension` doesn't mention that `real_extension` must be non-null when `name` is present in the supported extensions list — the implementation does `CHECK(real_extension)` in that branch. Per [DOC-004](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-004), contract docs should cover gotchas and restrictions. Consider adding something like `// |real_extension| must be non-null when |name| is in |supported_extensions_|.`

third_party/blink/renderer/bindings/core/webgl/webgl_extension_handler.h
```diff
@@ -0,0 +1,92 @@
...
+#include "third_party/blink/renderer/platform/wtf/vector.h"
+
+namespace blink {
+class ScriptObject;
+class ScriptState;
+
+using ExtensionVector = Vector<String>;
+
+// Internal data type of the fake WebGL extensions array.
+struct FakeExtensionType {
```
goodov, line 27:
```
using ExtensionVector = Vector<String>;

// Internal data type of the fake WebGL extensions array.
struct FakeExtensionType {
```

these type names are too generic. Can you mention in the name farbling/webgl? You're adding these types to `blink` namespace, but they are used only internally in `.cc` and in tests.

third_party/blink/renderer/bindings/core/webgl/webgl_extension_handler.h
```diff
@@ -0,0 +1,92 @@
...
+  // The real name of the fake extension returned by the getSupportedExtension
+  // call when farbled. For example: EXT_texture_sampler.
+  String name;
+  // The underlying name of the script object which is returned by the
+  // getExtension call when farbled.
+  String script_object_name;
+};
+
+inline constexpr size_t kFakeExtensionsSize = 21;
```
goodov: maybe keep this private and return `span` from public interface for tests?

goodov (approved):
lgtm, but please make the types be not that generic.

also some things can be internalized.

third_party/blink/renderer/bindings/core/webgl/webgl_extension_handler.h
r0hit0303: > Can you mention in the name farbling/webgl? You're adding these types to blink namespace [...]

I've moved them from blink to webgl namespace so I'm guessing we could still keep the same typename without adding `webgl` but please let me know if you think I still should update the type name. 

> but they are used only internally in .cc and in tests.

I've also moved these types and other constants into `...common/constants.h`. I needed to access them in tests like you said so I had to put them inside a header file.

third_party/blink/renderer/bindings/core/webgl/webgl_extension_handler.h
r0hit0303: Yeah makes sense. `kFakeExtensionsSize` seems like an implementation detail that should probably be not leaked outside.

third_party/blink/renderer/bindings/core/webgl/webgl_extension_handler.h
r0hit0303: > I've also moved these types and other constants into ...common/constants.h.

I had to revert back to putting the types inside the `webgl_extension_handler.h` as creating a separate source_set for common was failing `npm run gn_check` due to importing the "wtf" dependency which is inside "third_party//" but our target was "brave//third_party..." which failed the visibility test in the [upstream](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/platform/wtf/BUILD.gn;l=28-32). There was no issues with `webgl_extension_handler` because it was being baked in inside the "//third_party/blink/renderer" target [directly](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/third_party/blink/renderer/includes.gni?L30). Creating a separate target for constants.h blurred that boundary as it was indeed getting baked inside "//third_party/blink/renderer" but also created a separate "common" target. I'm not super knowledgeable with build files so there may be a way but I'm ok with keeping it here for the moment to keep the momentum.

r0hit0303:
Thanks @goodov for the review. All changes reflected in the latest ps.

third_party/blink/renderer/bindings/core/webgl/webgl_extension_handler.h
goodov: most of the code in `//third_party/blink` should be in `blink::`. Just put our things either in a nested namespace or made the type names be less generic.

I'd look into those:
1. Do you really need `ExtensionVector` alias? Just use `Vector<String>`, that's totally fine.
2. Name `FakeExtensionType` as `WebGLFakeExtension`.
3. Name `WebGLExtensionHandler` as `WebGLFarbledExtensionHandler`

You can have `ExtensionVector` as a part of `WebGLExtensionHandler` and put all those `Create*` as static functions of `WebGLExtensionHandler` and make the constructor private.

r0hit0303: Apologies I had enabled auto merge on the PR so it got merged before your last comment. Let me file an issue to track this and resolve it immediately.

Update: Filed the issue https://github.com/brave/brave-browser/issues/55898

PR #36763, opened 2026-05-27
[VPN 2.0] Move ClearSubscriberCredential from "browser" to "common"

### LINKED ISSUE #54597: Implement a stub BraveVpnService v2

alekspop26:

This change moves ClearSubscriberCredential from a helper located in a "browser" directory, to utils located in a "common" directory. This function is used outside of the Brave VPN targes, so an additional dependency on V1's "browser" there is unnecessary.


netzenbot (approved):

simonhong (approved):
++

PR #36766, opened 2026-05-27
Large chunked OHTTP request fix

DJAndries:

The RFC suggests a maximum chunk size of 16,384 bytes. If a request is made with a plaintext length that is greater than the limit, split it up into 16kb chunks and encrypt them individually.

### DIFF
--- a/components/oblivious_http/oblivious_http_chunk_processor.cc
+++ b/components/oblivious_http/oblivious_http_chunk_processor.cc
@@ -11,6 +11,10 @@
 
 namespace oblivious_http {
 
+namespace {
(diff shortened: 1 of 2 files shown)


components/oblivious_http/oblivious_http_chunk_processor.cc
```diff
@@ -11,6 +11,10 @@
 
 namespace oblivious_http {
 
+namespace {
+constexpr size_t kMaxEncryptChunkSize = 15000;
```
tmancey: Maybe 16384?

tmancey (approved):

PR #36777, opened 2026-05-27
[iOS][Wallet] Clear window.localStorage from swift side when wallet is reset

nuo-xu:
Override ClearWalletUIStoragePartition() in BraveWalletServiceDelegateIos  to spin up a hidden WKWebView using the profile's WKWebViewConfiguration and load wallet webui url then evaluate localStorage.clear(). 
This runs at the core level so it works regardless of whether any wallet tab is open. 

Resolves 
https://github.com/brave/brave-browser/issues/55673
https://github.com/brave/brave-browser/issues/55121

Test:
1. create or restore a wallet 
2. in the webui wallet portfolio screen, set some preference like hide balance, hide NFT tab, show graph 
3. close or keep this tab
4. reset wallet in Settings/Web3
5. verify now the wallet webui tab is now back at welcome screen
6. create a new wallet from the welcome screen and finish all the onboarding steps
7. verify in the portfolio screen, all the previous stored value are reset as the default value (balance is default unhidden, NFT tab is default unhidden, graph is default hidden)
8. verify the reset wallet's data does not reappearing(balance, accounts, etc)

### DIFF
--- a/ios/browser/brave_wallet/brave_wallet_service_factory.mm
+++ b/ios/browser/brave_wallet/brave_wallet_service_factory.mm
@@ -9,6 +9,9 @@
 
 #include <memory>
 
+#include "base/functional/bind.h"
+#include "base/memory/weak_ptr.h"
+#include "base/threading/sequence_bound.h"
 #include "brave/components/brave_wallet/browser/brave_wallet_service.h"
 #include "brave/components/brave_wallet/browser/brave_wallet_service_delegate.h"
 #include "brave/components/constants/webui_url_constants.h"
@@ -16,48 +19,39 @@
 #include "ios/chrome/browser/shared/model/application_context/application_context.h"
 #include "ios/chrome/browser/shared/model/profile/profile_ios.h"
 #include "ios/web/public/browser_state.h"
+#include "ios/web/public/thread/web_task_traits.h"
+#include "ios/web/public/thread/web_thread.h"
 #include "ios/web/web_state/ui/wk_web_view_configuration_provider.h"
 #include "services/network/public/cpp/shared_url_loader_factory.h"
 
-@interface BraveWalletWebUILocalStorageCleaner : NSObject <WKNavigationDelegate>
-+ (void)clearForProfile:(ProfileIOS*)profile;
+@interface WalletWebViewRunnerDelegate : NSObject <WKNavigationDelegate> {
+  base::OnceClosure _onDone;
+}
+- (instancetype)initWithClosure:(base::OnceClosure)closure;
 @end
 
-@implementation BraveWalletWebUILocalStorageCleaner {
-  WKWebView* _webView;
-}
+@implementation WalletWebViewRunnerDelegate
 
-static BraveWalletWebUILocalStorageCleaner* currentCleaner;
-
-+ (void)clearForProfile:(ProfileIOS*)profile {
-  dispatch_async(dispatch_get_main_queue(), ^{
-    WKWebViewConfiguration* config =
-        [web::WKWebViewConfigurationProvider::FromBrowserState(profile)
-                .GetWebViewConfiguration() copy];
-    WKWebView* webView = [[WKWebView alloc] initWithFrame:CGRectZero
-                                            configuration:config];
-    BraveWalletWebUILocalStorageCleaner* cleaner =
-        [[BraveWalletWebUILocalStorageCleaner alloc] init];
-    cleaner->_webView = webView;
-    webView.navigationDelegate = cleaner;
-    currentCleaner = cleaner;
-    [webView loadRequest:[NSURLRequest
-                             requestWithURL:
-                                 [NSURL URLWithString:@(kBraveUIWalletURL)]]];
-  });
+- (instancetype)initWithClosure:(base::OnceClosure)closure {
+  self = [super init];
+  if (self) {
+    _onDone = std::move(closure);
+  }
+  return self;
 }
 
 - (void)done {
-  _webView.navigationDelegate = nil;
-  _webView = nil;
-  currentCleaner = nil;
+  if (_onDone) {
+    std::move(_onDone).Run();
+  }
 }
 
 - (void)webView:(WKWebView*)webView
     didFinishNavigation:(WKNavigation*)navigation {
+  __weak WalletWebViewRunnerDelegate* weakSelf = self;
   [webView evaluateJavaScript:@"localStorage.clear()"
             completionHandler:^(id result, NSError* error) {
-              [self done];
+              [weakSelf done];
             }];
 }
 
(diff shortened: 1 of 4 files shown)


netzenbot (approved):

StephenHeaps:
This doesn't work if the WebUI tab is closed when you reset wallet from native settings.
I wonder if we might be able to include this change in WebUI front-end? If `brave://wallet` is visited and no wallet is setup, can we reset `window.localStorage`? cc @Douglashdaniel

EliHini (approved):

EliHini (commented):
> This doesn't work if the WebUI tab is closed when you reset wallet from native settings. I wonder if we might be able to include this change in WebUI front-end? If `brave://wallet` is visited and no wallet is setup, can we reset `window.localStorage`? cc @Douglashdaniel

This is a valid catch. Let's wait for @Douglashdaniel

ios/browser/brave_wallet/brave_wallet_service_factory.mm
```diff
@@ -5,32 +5,102 @@
...
+}
+
+static BraveWalletWebUILocalStorageCleaner* currentCleaner;
+
++ (void)clearForProfile:(ProfileIOS*)profile {
+  dispatch_async(dispatch_get_main_queue(), ^{
+    WKWebViewConfiguration* config =
+        [web::WKWebViewConfigurationProvider::FromBrowserState(profile)
+                .GetWebViewConfiguration() copy];
+    WKWebView* webView = [[WKWebView alloc] initWithFrame:CGRectZero
```
StephenHeaps: Likely want to use `BraveWebView` for this for consistency? I believe we should use our WebView everywhere cc @kylehickinson

kylehickinson: I'd say its fine to not use a whole Chromium web view for this sort of task (isolated WebKit specific job). Chromium does the same for some sandboxed tasks

ios/browser/brave_wallet/brave_wallet_service_factory.mm
```diff
@@ -5,32 +5,102 @@
...
+  WKWebView* _webView;
+}
+
+static BraveWalletWebUILocalStorageCleaner* currentCleaner;
+
++ (void)clearForProfile:(ProfileIOS*)profile {
+  dispatch_async(dispatch_get_main_queue(), ^{
+    WKWebViewConfiguration* config =
+        [web::WKWebViewConfigurationProvider::FromBrowserState(profile)
+                .GetWebViewConfiguration() copy];
```
kylehickinson: this is already copied, no need to call copy again

ios/browser/brave_wallet/brave_wallet_service_factory.mm
kylehickinson: examples: 
- https://source.chromium.org/chromium/chromium/src/+/main:ios/web/browsing_data/browsing_data_removing_util.mm;l=56
- https://source.chromium.org/chromium/chromium/src/+/main:ios/web/js_image_transcoder/java_script_image_transcoder.mm;l=24

ios/browser/brave_wallet/brave_wallet_service_factory.mm
```diff
@@ -5,32 +5,102 @@
...
+@end
+
+@implementation BraveWalletWebUILocalStorageCleaner {
+  WKWebView* _webView;
+}
+
+static BraveWalletWebUILocalStorageCleaner* currentCleaner;
+
++ (void)clearForProfile:(ProfileIOS*)profile {
+  dispatch_async(dispatch_get_main_queue(), ^{
```
kylehickinson: Use Chromium task runners instead of GCD at the C++ level, run `DCHECK_CURRENTLY_ON` in this method

ios/browser/brave_wallet/brave_wallet_service_factory.mm
```diff
@@ -5,32 +5,102 @@
...
+    currentCleaner = cleaner;
+    [webView loadRequest:[NSURLRequest
+                             requestWithURL:
+                                 [NSURL URLWithString:@(kBraveUIWalletURL)]]];
+  });
+}
+
+- (void)done {
+  _webView.navigationDelegate = nil;
+  _webView = nil;
```
kylehickinson: technically not needed if self dies on the next line setting the global to nil

ios/browser/brave_wallet/brave_wallet_service_factory.mm
```diff
@@ -5,32 +5,102 @@
...
 
+@interface BraveWalletWebUILocalStorageCleaner : NSObject <WKNavigationDelegate>
++ (void)clearForProfile:(ProfileIOS*)profile;
+@end
+
+@implementation BraveWalletWebUILocalStorageCleaner {
+  WKWebView* _webView;
+}
+
+static BraveWalletWebUILocalStorageCleaner* currentCleaner;
```
kylehickinson: Probably the wrong approach to have a static global here, `BraveWalletServiceDelegateIos` may be better at holding the lifetime of the web view and can nil delete it in the callback. `WebViewRunner` is a better example to follow here: https://source.chromium.org/chromium/chromium/src/+/main:components/user_data_importer/ios/ios_bookmark_parser.mm;l=198

kylehickinson: Also something to do as well re task runners/GCD: https://source.chromium.org/chromium/chromium/src/+/main:components/user_data_importer/ios/ios_bookmark_parser.h;l=39

StephenHeaps (approved):

ios/browser/brave_wallet/brave_wallet_service_factory.mm
```diff
@@ -5,34 +5,133 @@
...
 #include "ios/web/public/browser_state.h"
+#include "ios/web/public/thread/web_task_traits.h"
+#include "ios/web/public/thread/web_thread.h"
+#include "ios/web/web_state/ui/wk_web_view_configuration_provider.h"
 #include "services/network/public/cpp/shared_url_loader_factory.h"
 
+@interface WalletWebViewRunnerDelegate : NSObject <WKNavigationDelegate> {
+  base::OnceClosure _onDone;
+}
+- (instancetype)initWithClosure:(base::OnceClosure)closure;
```
StephenHeaps: Might want to reference what this closure is used for? `initWithCompletion`? 🤔

kylehickinson (approved):

ios/browser/brave_wallet/brave_wallet_service_factory.mm
nuo-xu: updated in [46ee8d1](https://github.com/brave/brave-core/pull/36777/commits/46ee8d1d67f9106c8f07be6bf4d9c7198ccdbcf1)

stoletheminerals (approved):

ios/browser/brave_wallet/brave_wallet_service_factory.mm
```diff
@@ -5,34 +5,133 @@
...
+    web_view_ = nil;
+    delegate_ = nil;
+  }
+
+  WKWebView* web_view_ = nil;
+  WalletWebViewRunnerDelegate* delegate_ = nil;
+  base::WeakPtrFactory<WalletWebViewRunner> weak_factory_{this};
+};
+
 class BraveWalletServiceDelegateIos : public BraveWalletServiceDelegate {
```
goodov, line 111: consider moving this class out of the `factory.mm`. Factory is factory, delegate impl should be separate.

goodov (approved):

ios/browser/brave_wallet/brave_wallet_service_factory.mm
nuo-xu: thank you! will address it separately

PR #36781, opened 2026-05-27
[Android] Respect `Show search suggestions` in search widget

### LINKED ISSUE #44489: Search `suggestions`/`websites` shown in `Search widget` if `Show browser suggestions` is OFF
### Description

Search `suggestions`/`websites` shown in the `Search widget` query results if `Show browser suggestions` is OFF while `Show search suggestions`/`Show top sites suggestions` are ON.

### Steps to reproduce

1. Launch Brave
2. Hamburger menu > Settings > Search engines > *Make sure that `Show browser suggestions` is ON*
3. Enable both `Show search suggestions` and `Show top sites suggestions`
4. Disable `Show browser suggestions`
5. Close Brave and add `Search widget` to a Home screen
6. Type a query > *Observe*

### Actual result

Search `suggestions`/`websites` shown in the `Search widget` query results if `Show browser suggestions` is OFF while `Show search suggestions`/`Show top sites suggestions` are ON.

--------------

https://github.com/user-attachments/assets/f50511cb-4859-4640-8a5b-347645d086e4

### Expected result

Search `suggestions`/`websites` must not be shown in the `Search widget` query results if `Show browser suggestions` is OFF while `Show search suggestions`/`Show top sites suggestions` are ON.

Here is the screencast of this behavior when doing the search inside of Brave.

----------

https://github.com/user-attachments/assets/93d6462a-a896-45fe-943d-ecba2089180a

### Reproduces how often

Easily reproduced

### Brave version

Brave build: 1.78.21 
Chromium: 134.0.6998.45 (Official Build) canary (64-bit) 

### Device

- Brand/model: ANY
- Android version: ANY

### Channel information

- [ ] release (stable)
- [ ] beta
- [x] nightly

### Reproducibility

### Miscellaneous information

_No response_

samartnik:
Resolves https://github.com/brave/brave-browser/issues/44489

### DIFF
--- a/android/java/org/chromium/base/BravePreferenceKeys.java
+++ b/android/java/org/chromium/base/BravePreferenceKeys.java
@@ -68,8 +68,6 @@ public final class BravePreferenceKeys {
     public static final String BRAVE_LEO_AUTOCOMPLETE =
             "org.chromium.chrome.browser.Brave_Leo_AutoComplete";
 
-    public static final String BRAVE_AUTOCOMPLETE_ENABLED = "brave.autocomplete_enabled";
-
     public static final String CUSTOMIZABLE_BRAVE_MENU_ITEM_ID_FORMAT =
             "customizable_brave_menu_item_id_%s";
 
--- a/android/java/org/chromium/base/BravePreferenceKeys.java
+++ b/android/java/org/chromium/base/BravePreferenceKeys.java
@@ -68,6 +68,8 @@ public final class BravePreferenceKeys {
     public static final String BRAVE_LEO_AUTOCOMPLETE =
             "org.chromium.chrome.browser.Brave_Leo_AutoComplete";
 
+    public static final String BRAVE_AUTOCOMPLETE_ENABLED = "brave.autocomplete_enabled";
+
     public static final String CUSTOMIZABLE_BRAVE_MENU_ITEM_ID_FORMAT =
             "customizable_brave_menu_item_id_%s";
 
--- a/browser/ui/android/omnibox/java/src/org/chromium/chrome/browser/omnibox/suggestions/BraveDropdownItemViewInfoListBuilder.java
+++ b/browser/ui/android/omnibox/java/src/org/chromium/chrome/browser/omnibox/suggestions/BraveDropdownItemViewInfoListBuilder.java
@@ -31,6 +31,7 @@ import org.chromium.components.omnibox.AutocompleteResult;
 import org.chromium.components.omnibox.GroupsProto.GroupConfig;
 import org.chromium.components.omnibox.action.OmniboxActionDelegate;
 import org.chromium.components.omnibox.suggestions.OmniboxSuggestionUiType;
+import org.chromium.components.user_prefs.UserPrefs;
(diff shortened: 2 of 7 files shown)


android/java/org/chromium/base/BravePreferenceKeys.java
```diff
@@ -68,6 +68,8 @@ public final class BravePreferenceKeys {
     public static final String BRAVE_LEO_AUTOCOMPLETE =
             "org.chromium.chrome.browser.Brave_Leo_AutoComplete";
 
+    public static final String BRAVE_AUTOCOMPLETE_ENABLED = "brave.autocomplete_enabled";
```
SergeyZhukovsky: isn't it `BravePref.AUTOCOMPLETE_ENABLED`? I think we autopopulate it from c++ enum. We should just use `BravePref.AUTOCOMPLETE_ENABLED` in both `BraveAutocompleteMediator` and `BraveDropdownItemViewInfoListBuilder`, and drop the new constant. That also removes the risk if the native key is ever renamed.

browser/ui/android/omnibox/java/src/org/chromium/chrome/browser/omnibox/suggestions/BraveDropdownItemViewInfoListBuilder.java
```diff
@@ -122,10 +130,18 @@ List<DropdownItemViewInfo> buildDropdownViewInfoList(
...
-                        : true;
+        boolean autocompleteEnabled;
+        if (tab != null) {
+            autocompleteEnabled =
+                    mLeoAutocompleteDelegate.isAutoCompleteEnabled(tab.getWebContents());
+        } else {
+            // The search widget launches SearchActivity without a tab, so use the profile
+            // pushed in via setProfile to honor the "Show browser suggestions" preference.
+            autocompleteEnabled =
+                    mProfile == null
```
SergeyZhukovsky: do you see cases when it's null? Maybe a cold start, like Force stop Brave from App info and try the widget? The thing is we show suggestions when `mProfile == null` always as `autocompleteEnabled == true` by default. Maybe from privacy prospective it's better to not show in that case and the cost (one possibly-empty frame on cold start)?

android/java/org/chromium/base/BravePreferenceKeys.java
samartnik: Changed here 38a4123999f7251775f3e04f622fdb047cbc1102

browser/ui/android/omnibox/java/src/org/chromium/chrome/browser/omnibox/suggestions/BraveDropdownItemViewInfoListBuilder.java
samartnik: That's a theoretical possibility, cold start is the possible case when it may happen. Changed the default value for such case 0120465bb5be97af43250bc8e191c4d71966857d

SergeyZhukovsky (approved):
++

netzenbot (approved):

PR #36785, opened 2026-05-27
[AI Chat] Enable global side panel by default

### LINKED ISSUE #53416: [AI Chat] Global Panel features needed
By default we will keep the same conversation open across active tab changes and page navigations. Even if the tabs are linked within the same session (Tab B opened via Tab A).

- If I select another conversation, then show the back button
  - back button goes to the conversation I was on (stored client side)
_In other words, don't use any of the tab-related GetConversationForAssociatedContent AIChatService stuff._

This version of global conversation per-panel will have the following UX:
- Every time I close and open the panel it will be a new conversation
- Each window will have a different conversation

Optional in-between (perhaps put behind an additional flag):
If I have an active conversation that has tabs attached to it, and then I go to a new tab and explicitly click the new conversation button, when I go back to a Tab that has an active conversation with it attached then I should switch to that.
Also if I go back to a tab that had a conversation without any tabs associated to it, then I should show that conversation by default.

Optional additional:
Have a concept of stale-ness for a conversation. Don't switch to the conversation when switching active tab if the conversation is stale. Staleness could be decided by:
- age of conversation
- whether it has the page attached or not
- number of turns

Future
- Maybe have the LLM decide whether the conversation is tab-associated or not

#### Initial version
- [x] Feature flag
- [X] Panel stays open across navigations and active tab, to the same conversation
  - https://github.com/brave/brave-core/issues/30720
- [x] Active tab is automatically attached, and changed whenever active tab changes or navigates
  - https://github.com/brave/brave-browser/issues/53533
- [x] "Active conversation" is initial conversation, not Service-provided conversation keyed via Page. This fixes the back arrow to the _current_ conversation.
  - https://github.com/brave/brave-browser/issues/55399
 - [ ] https://github.com/brave/brave-browser/issues/55702
- [ ] Enable by default (hard-coded, ride trains)

#### Follow-ups
- [ ] Placeholder is sent to LLM instead of full page contents, possibly only for active tab (though could be useful for all tabs as large tool output is stripped).
  - [ ] Create a tool, `get_tab_content` which returns entire content for an attachment that was sent as a placeholder
  - [ ] https://github.com/brave/brave-browser/issues/53417
      - as a follow up to the above, we should only remove automatically staged content when changing tabs/navigating.
- [ ] Tool for LLM to query all tabs and ask user permission to attach some of them

petemill:
- Resolves https://github.com/brave/brave-browser/issues/53416

### DIFF
--- a/browser/ai_chat/ai_chat_policy_browsertest.cc
+++ b/browser/ai_chat/ai_chat_policy_browsertest.cc
@@ -46,7 +46,15 @@ namespace policy {
 class AIChatPolicyTest : public InProcessBrowserTest,
                          public ::testing::WithParamInterface<bool> {
  public:
-  AIChatPolicyTest() = default;
+  AIChatPolicyTest() {
+    // Disable the global side panel feature so the AI Chat side panel entry
+    // is registered on the tab's SidePanelRegistry (the non-global path) for
+    // TabSidePanelRegistry below.
+    feature_list_.InitWithFeatures(
+        /*enabled_features=*/{ai_chat::features::kAIChat},
+        /*disabled_features=*/
+        {ai_chat::features::kAIChatGlobalSidePanelEverywhere});
+  }
   ~AIChatPolicyTest() override = default;
 
   void SetUpInProcessBrowserTestFixture() override {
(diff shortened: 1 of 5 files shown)


fallaciousreasoning (approved):
exciting! Going to be interesting to hear what people think!

netzenbot (approved):

PR #36787, opened 2026-05-27
[VPN 2.0] Add GN buildflags for VPN Architectures 1.0 and 2.0

### LINKED ISSUE #54597: Implement a stub BraveVpnService v2

alekspop26:

This change revisits the approach how Brave VPN is enabled in GN. New flags are introduced: enable_brave_vpn_v1 and enable_brave_vpn_v2. Currently, only V1 architecture is implemented, hence, the GN flag enable_brave_vpn_v1 is enabled by default, and enable_brave_vpn_v2 is disabled by default. enable_brave_vpn becomes a dependent flag: it is enabled when either enable_brave_vpn_v1, or enable_brave_vpn_v2, or both are enabled.

### DIFF
--- a/chromium_src/chrome/browser/sources.gni
+++ b/chromium_src/chrome/browser/sources.gni
@@ -26,7 +26,7 @@ if (!is_android) {
 
 if (enable_brave_vpn) {
   brave_chromium_src_chrome_browser_deps +=
-      [ "//brave/components/brave_vpn/browser" ]
+      [ "//brave/components/brave_vpn/common" ]
 }
 
 if (toolkit_views) {
--- a/components/brave_vpn/common/buildflags/BUILD.gn
+++ b/components/brave_vpn/common/buildflags/BUILD.gn
@@ -10,6 +10,8 @@ buildflag_header("buildflags") {
   header = "buildflags.h"
   flags = [
     "ENABLE_BRAVE_VPN=$enable_brave_vpn",
+    "ENABLE_BRAVE_VPN_V1=$enable_brave_vpn_v1",
+    "ENABLE_BRAVE_VPN_V2=$enable_brave_vpn_v2",
     "ENABLE_BRAVE_VPN_WIREGUARD=$enable_brave_vpn_wireguard",
   ]
 }
--- a/components/brave_vpn/common/buildflags/buildflags.gni
+++ b/components/brave_vpn/common/buildflags/buildflags.gni
@@ -7,9 +7,18 @@ import("//brave/components/brave_origin/buildflags/buildflags.gni")
 
 declare_args() {
   # On macOS, vpn is not available w/o signing.
-  enable_brave_vpn =
+  # Build with Brave VPN architecture 1.0 support.
+  enable_brave_vpn_v1 =
       (is_win || is_android || is_mac || is_ios) && !is_brave_origin_branded
+
+  # Build with Brave VPN architecture 2.0 support.
+  enable_brave_vpn_v2 = false
 }
+
+# Brave VPN is enabled, if at least one VPN architecture is enabled.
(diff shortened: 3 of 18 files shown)


chromium_src/chrome/browser/sources.gni
```diff
@@ -26,7 +26,7 @@ if (!is_android) {
 
 if (enable_brave_vpn) {
   brave_chromium_src_chrome_browser_deps +=
-      [ "//brave/components/brave_vpn/browser" ]
+      [ "//brave/components/brave_vpn/common" ]
```
bridiver, line 29: why did this change to common?

components/brave_vpn/common/buildflags/buildflags.gni
```diff
@@ -19,7 +28,5 @@ declare_args() {
   enable_brave_vpn_wireguard = (is_mac || is_win) && enable_brave_vpn
 }
 
-declare_args() {
-  # vpn 2.0 build flags
-  enable_brave_vpn2_deps = false
-}
+# Build Brave VPN 2.0 deps only if VPN architecture 2.0 is supported.
+enable_brave_vpn2_deps = enable_brave_vpn_v2
```
bridiver: do we still need this or can we remove in a follow-up?

components/brave_vpn/common/buildflags/BUILD.gn
```diff
@@ -10,6 +10,8 @@ buildflag_header("buildflags") {
   header = "buildflags.h"
   flags = [
     "ENABLE_BRAVE_VPN=$enable_brave_vpn",
+    "ENABLE_BRAVE_VPN_V1=$enable_brave_vpn_v1",
+    "ENABLE_BRAVE_VPN_V2=$enable_brave_vpn_v2",
```
netzenbot, line 14: nit: Per [BS-049](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-049), when introducing a new build flag for a component, add `static_assert(BUILDFLAG(ENABLE_BRAVE_VPN_V1));` near the top of `brave/components/brave_vpn/browser/connection/brave_vpn_connection_manager.h` to guard against accidental inclusion when the V1 architecture is disabled.

chromium_src/chrome/browser/sources.gni
alekspop26: Actually I don't understand why it was "browser" in the first place. I checked all the included headers in the `chromium_src/chrome/browser/" and it seems that "common" dependency should suffice here.

```
grep -roh '#include "brave/components/brave_vpn/[^"]*"' chromium_src/chrome/browser \
  | sort | uniq -c | sort -rn \
  | awk '{count=$1; $1=""; sub(/^ /, ""); printf "`%s` : %d times\n", $0, count}'
```
```
`#include "brave/components/brave_vpn/common/buildflags/buildflags.h"` : 6 times
`#include "brave/components/brave_vpn/common/features.h"` : 3 times
`#include "brave/components/brave_vpn/common/brave_vpn_utils.h"` : 2 times
```
Given that I'm splitting up the service dependencies - this effectively reduces the unwanted dependency on V1.

components/brave_vpn/common/buildflags/BUILD.gn
alekspop26: Will fix in a follow up CR.

components/brave_vpn/common/buildflags/buildflags.gni
alekspop26: Good point, I don't think we need that anymore. Removed the `enable_brave_vpn2_deps` buildflag, and replaced it with `enable_brave_vpn_v2`.

bridiver (approved):

PR #36790, opened 2026-05-27
Install and Register playlist exclusions component

### LINKED ISSUE #55010: Playlist: Capture allowed from non‑playable pages produces permanently broken items

EliHini:
Add component installer policy and shared constants for the playlist exclusions CRX, and register it with `ComponentUpdateService` on iOS.

Screenshot 2026-06-04 at 1 55 10 PM

Resolves https://github.com/brave/brave-browser/issues/55010

### DIFF
--- a/components/playlist/core/browser/playlist_exclusion_component_installer.cc
+++ b/components/playlist/core/browser/playlist_exclusion_component_installer.cc
@@ -1,0 +1,146 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "brave/components/playlist/core/browser/playlist_exclusion_component_installer.h"
+
+#include <stdint.h>
+
+#include <array>
+#include <memory>
+#include <string>
+#include <vector>
+
+#include "base/feature_list.h"
+#include "base/files/file_path.h"
+#include "base/files/file_util.h"
+#include "base/functional/bind.h"
+#include "brave/components/brave_component_updater/browser/brave_on_demand_updater.h"
+#include "brave/components/playlist/core/browser/playlist_exclusion.h"
+#include "brave/components/playlist/core/common/constants.h"
+#include "brave/components/playlist/core/common/features.h"
+#include "components/component_updater/component_installer.h"
+#include "components/component_updater/component_updater_service.h"
+
+using brave_component_updater::BraveOnDemandUpdater;
+
+namespace playlist {
+
+namespace {
+
+inline constexpr std::array<uint8_t, 32>
+    kPlaylistExclusionsComponentPublicKeySHA256 = {
+        0xa0, 0x92, 0x8e, 0x87, 0x35, 0xdd, 0xa3, 0x95, 0xf2, 0x88, 0x3d,
+        0xc7, 0x2a, 0x04, 0x2b, 0xc9, 0xff, 0x4b, 0xcc, 0xfb, 0x77, 0xb6,
+        0xd4, 0xc8, 0x89, 0x68, 0x28, 0xf7, 0xda, 0x08, 0x03, 0xb4};
+
+class PlaylistExclusionsComponentInstallerPolicy
+    : public component_updater::ComponentInstallerPolicy {
+ public:
+  PlaylistExclusionsComponentInstallerPolicy();
+  ~PlaylistExclusionsComponentInstallerPolicy() override;
+
+  PlaylistExclusionsComponentInstallerPolicy(
+      const PlaylistExclusionsComponentInstallerPolicy&) = delete;
+  PlaylistExclusionsComponentInstallerPolicy& operator=(
+      const PlaylistExclusionsComponentInstallerPolicy&) = delete;
+
+  // component_updater::ComponentInstallerPolicy:
+  bool SupportsGroupPolicyEnabledComponentUpdates() const override;
+  bool RequiresNetworkEncryption() const override;
+  update_client::CrxInstaller::Result OnCustomInstall(
+      const base::DictValue& manifest,
+      const base::FilePath& install_dir) override;
+  void OnCustomUninstall() override;
+  bool VerifyInstallation(const base::DictValue& manifest,
+                          const base::FilePath& install_dir) const override;
+  void ComponentReady(const base::Version& version,
+                      const base::FilePath& path,
+                      base::DictValue manifest) override;
+  base::FilePath GetRelativeInstallDir() const override;
+  void GetHash(std::vector<uint8_t>* hash) const override;
+  std::string GetName() const override;
+  update_client::InstallerAttributes GetInstallerAttributes() const override;
+  bool IsBraveComponent() const override;
+};
+
+PlaylistExclusionsComponentInstallerPolicy::
+    PlaylistExclusionsComponentInstallerPolicy() = default;
+
+PlaylistExclusionsComponentInstallerPolicy::
+    ~PlaylistExclusionsComponentInstallerPolicy() = default;
+
+bool PlaylistExclusionsComponentInstallerPolicy::
+    SupportsGroupPolicyEnabledComponentUpdates() const {
+  return true;
+}
+
+bool PlaylistExclusionsComponentInstallerPolicy::RequiresNetworkEncryption()
+    const {
+  return false;
+}
+
+update_client::CrxInstaller::Result
+PlaylistExclusionsComponentInstallerPolicy::OnCustomInstall(
+    const base::DictValue& manifest,
+    const base::FilePath& install_dir) {
+  return update_client::CrxInstaller::Result(0);
+}
+
+void PlaylistExclusionsComponentInstallerPolicy::OnCustomUninstall() {}
+
+void PlaylistExclusionsComponentInstallerPolicy::ComponentReady(
+    const base::Version& version,
+    const base::FilePath& path,
+    base::DictValue manifest) {
+  PlaylistExclusions::GetInstance()->OnComponentReady(path);
+}
+
+bool PlaylistExclusionsComponentInstallerPolicy::VerifyInstallation(
+    const base::DictValue& manifest,
+    const base::FilePath& install_dir) const {
+  return base::PathExists(install_dir.AppendASCII(kPlaylistExclusionsJsonFile));
+}
+
+base::FilePath
+PlaylistExclusionsComponentInstallerPolicy::GetRelativeInstallDir() const {
+  return base::FilePath::FromUTF8Unsafe(kPlaylistExclusionsComponentId);
+}
+
+void PlaylistExclusionsComponentInstallerPolicy::GetHash(
+    std::vector<uint8_t>* hash) const {
+  hash->assign(kPlaylistExclusionsComponentPublicKeySHA256.begin(),
+               kPlaylistExclusionsComponentPublicKeySHA256.end());
+}
+
+std::string PlaylistExclusionsComponentInstallerPolicy::GetName() const {
+  return kPlaylistExclusionsComponentName;
+}
+
+update_client::InstallerAttributes
+PlaylistExclusionsComponentInstallerPolicy::GetInstallerAttributes() const {
+  return update_client::InstallerAttributes();
+}
+
+bool PlaylistExclusionsComponentInstallerPolicy::IsBraveComponent() const {
+  return true;
+}
+
+}  // namespace
+
+void RegisterPlaylistExclusionsComponent(
+    component_updater::ComponentUpdateService* cus) {
+  if (!base::FeatureList::IsEnabled(playlist::features::kPlaylist) || !cus) {
+    return;
+  }
+
+  auto installer = base::MakeRefCounted<component_updater::ComponentInstaller>(
+      std::make_unique<PlaylistExclusionsComponentInstallerPolicy>());
+  installer->Register(cus, base::BindOnce([]() {
+                        BraveOnDemandUpdater::GetInstance()->EnsureInstalled(
+                            kPlaylistExclusionsComponentId);
+                      }));
+}
+
+}  // namespace playlist
--- a/components/playlist/core/browser/playlist_exclusion_unittest.cc
+++ b/components/playlist/core/browser/playlist_exclusion_unittest.cc
@@ -9,6 +9,7 @@
 #include "base/files/scoped_temp_dir.h"
 #include "base/test/run_until.h"
 #include "base/test/task_environment.h"
+#include "brave/components/playlist/core/common/constants.h"
 #include "testing/gtest/include/gtest/gtest.h"
 #include "url/gurl.h"
 
@@ -71,7 +72,7 @@ TEST_F(PlaylistExclusionsUnitTest, RulesBlockListedPaths) {
   base::ScopedTempDir temp_dir;
   ASSERT_TRUE(temp_dir.CreateUniqueTempDir());
   ASSERT_TRUE(base::WriteFile(
-      temp_dir.GetPath().AppendASCII("playlist_exclusions.json"), kJson));
+      temp_dir.GetPath().AppendASCII(kPlaylistExclusionsJsonFile), kJson));
 
   exclusions->OnComponentReady(temp_dir.GetPath());
   ASSERT_TRUE(base::test::RunUntil([&]() { return exclusions->is_ready_; }));
--- a/components/playlist/core/browser/playlist_exclusion_unittest.cc
+++ b/components/playlist/core/browser/playlist_exclusion_unittest.cc
@@ -5,10 +5,12 @@
 
 #include "base/files/file_util.h"
 #include "base/files/scoped_temp_dir.h"
+#include "base/test/scoped_feature_list.h"
 #include "base/test/task_environment.h"
 #include "base/test/test_future.h"
 #include "brave/components/playlist/core/browser/playlist_exclusions.h"
 #include "brave/components/playlist/core/common/constants.h"
+#include "brave/components/playlist/core/common/features.h"
 #include "testing/gtest/include/gtest/gtest.h"
 #include "url/gurl.h"
 
@@ -19,7 +21,10 @@ class PlaylistExclusionsUnitTest : public testing::Test {
   PlaylistExclusionsUnitTest() = default;
 
  protected:
-  void SetUp() override { ResetExclusions(PlaylistExclusions::GetInstance()); }
+  void SetUp() override {
+    feature_list_.InitAndEnableFeature(features::kPlaylist);
+    ResetExclusions(PlaylistExclusions::GetInstance());
+  }
 
   void ResetExclusions(PlaylistExclusions* exclusions) {
     exclusions->weak_factory_.InvalidateWeakPtrs();
@@ -35,6 +40,7 @@ class PlaylistExclusionsUnitTest : public testing::Test {
 
   base::test::TaskEnvironment task_environment_{
       base::test::TaskEnvironment::MainThreadType::UI};
+  base::test::ScopedFeatureList feature_list_;
 };
 
 TEST_F(PlaylistExclusionsUnitTest, NotReadyIsPermissive) {
--- a/components/playlist/core/browser/playlist_exclusion_unittest.cc
+++ b/components/playlist/core/browser/playlist_exclusion_unittest.cc
@@ -3,12 +3,11 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this file,
  * You can obtain one at https://mozilla.org/MPL/2.0/. */
 
-#include "brave/components/playlist/core/browser/playlist_exclusion.h"
-
 #include "base/files/file_util.h"
 #include "base/files/scoped_temp_dir.h"
-#include "base/test/run_until.h"
 #include "base/test/task_environment.h"
+#include "base/test/test_future.h"
+#include "brave/components/playlist/core/browser/playlist_exclusions.h"
 #include "brave/components/playlist/core/common/constants.h"
 #include "testing/gtest/include/gtest/gtest.h"
 #include "url/gurl.h"
@@ -25,8 +24,13 @@ class PlaylistExclusionsUnitTest : public testing::Test {
   void ResetExclusions(PlaylistExclusions* exclusions) {
     exclusions->weak_factory_.InvalidateWeakPtrs();
     exclusions->rules_.clear();
-    exclusions->is_ready_ = false;
-    exclusions->component_path_.clear();
+  }
+
+  void LoadExclusionsAndWait(PlaylistExclusions* exclusions,
+                             const base::FilePath& exclusions_file) {
+    base::test::TestFuture<void> future;
+    exclusions->LoadPlaylistExclusions(exclusions_file, future.GetCallback());
+    ASSERT_TRUE(future.Wait());
   }
 
   base::test::TaskEnvironment task_environment_{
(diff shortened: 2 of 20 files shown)


chromium_src/chrome/browser/component_updater/registration.cc
```diff
@@ -55,6 +60,9 @@ void RegisterComponentsForUpdate() {
   RegisterZxcvbnDataComponent(cus);
 #endif  // BUILDFLAG(IS_ANDROID)
   brave_user_agent::RegisterBraveUserAgentComponent(cus);
+#if BUILDFLAG(ENABLE_PLAYLIST)
+  playlist::RegisterPlaylistExclusionsComponent(cus);
```
StephenHeaps: I wouldn’t register this on desktop yet (unless they plan on using it asap too) as it’s a waste of a download/bandwidth on clients not yet using it. They can just register when it’s going to be used there 👍

netzenbot (approved):

chromium_src/chrome/browser/component_updater/registration.cc
EliHini: perfect! I will fix that

components/playlist/core/browser/playlist_exclusion_component_installer.cc
```diff
@@ -0,0 +1,148 @@
...
+}
+
+}  // namespace
+
+void RegisterPlaylistExclusionsComponent(
+    component_updater::ComponentUpdateService* cus) {
+#if !BUILDFLAG(ENABLE_PLAYLIST)
+  return;
+#endif
+  if (!cus) {
```
StephenHeaps: Since this likely will be used by other platforms, we should check against [playlist::features::kPlaylist](https://github.com/brave/brave-core/blob/4c8878f1b20c68e5ddeaca6f89330d5883dbf42c/components/playlist/core/common/features.h#L16) too.
But since this flag wasn't previously used on iOS (iOS has had playlist for a while), you likely need to update the feature flag to [enable by default for iOS only](https://github.com/brave/brave-core/blob/4c8878f1b20c68e5ddeaca6f89330d5883dbf42c/components/playlist/core/common/features.cc#L12).

EliHini: good point. Fixed.

components/playlist/core/common/features.cc
```diff
@@ -9,7 +9,12 @@
 
 namespace playlist::features {
 
-BASE_FEATURE(kPlaylist, base::FEATURE_DISABLED_BY_DEFAULT);
+BASE_FEATURE(kPlaylist,
+#if BUILDFLAG(ENABLE_PLAYLIST_BY_DEFAULT)
```
StephenHeaps: Why not just use `IS_IOS` buildflag?

EliHini: oh I see what you mean. just use the overall ios build flag rather than a feature specific one.

StephenHeaps: Yeah exactly 👍

bridiver: yes, we don't need an additional buildflag for this

bridiver: but this seems like a distinct change from adding the playlist exclusion component that should possibly be a separate PR?

components/playlist/core/browser/playlist_exclusion_component_installer.cc
```diff
@@ -0,0 +1,146 @@
...
+  return update_client::CrxInstaller::Result(0);
+}
+
+void PlaylistExclusionsComponentInstallerPolicy::OnCustomUninstall() {}
+
+void PlaylistExclusionsComponentInstallerPolicy::ComponentReady(
+    const base::Version& version,
+    const base::FilePath& path,
+    base::DictValue manifest) {
+  PlaylistExclusions::GetInstance()->OnComponentReady(path);
```
bridiver: for follow-up the filename should be playlist_exclusions to match the class name

components/playlist/core/browser/playlist_exclusion_component_installer.cc
```diff
@@ -0,0 +1,146 @@
...
+  return update_client::InstallerAttributes();
+}
+
+bool PlaylistExclusionsComponentInstallerPolicy::IsBraveComponent() const {
+  return true;
+}
+
+}  // namespace
+
+void RegisterPlaylistExclusionsComponent(
```
bridiver: `MaybeRegisterPlaylistExclusionsComponent`

EliHini: `RegisterPlaylistExclusionsComponent` for the filename I am assuming

bridiver: not sure what you're asking? I'm just saying it should be `Maybe*` because it installs conditionally based on the feature flag

components/playlist/core/browser/playlist_exclusions.h
```diff
@@ -81,14 +82,15 @@ class PlaylistExclusions {
...
   PlaylistExclusions();
 
-  void OnPlaylistExclusionsLoaded(const std::string& contents);
+  void OnPlaylistExclusionsLoaded(std::vector<PlaylistResolveRule> rules,
+                                  bool valid);
 
-  base::FilePath component_path_;
   std::vector<PlaylistResolveRule> rules_;
   bool is_ready_ = false;
+  uint64_t load_generation_ = 0;
```
bridiver: what is this for?

components/playlist/core/browser/playlist_exclusions.cc
```diff
@@ -42,6 +41,47 @@ bool GetPathPrefixesFromValue(const base::Value* value,
   return true;
 }
 
+struct PlaylistExclusionsLoadResult {
```
bridiver: use std::optional or just return empty rules. I'm still not really clear on why we need is_ready_ because doesn't it behave exactly like having empty rules?

components/playlist/core/browser/playlist_exclusions.h
EliHini: Removed. It was only there to synchronize unit tests without `RunUntilIdle`. Tests now use `SetOnLoadCompleteForTesting` and `TestFuture` (same pattern as query filter).

components/playlist/core/browser/playlist_exclusions.cc
EliHini: You're right,  `is_ready_` wasn't adding anything for `CanResolvePageSrcLater`. Whether the list wasn't loaded yet or loaded with zero rules, the behaviour was the same: stay permissive.

I've removed `is_ready_` and now store the rules in `std::optional<std::vector<PlaylistResolveRule>>`. `nullopt` means we're still loading or parsing failed; a value means parsing succeeded, even if the vector is empty. The background loader returns that optional directly, so the extra result struct is gone too.

components/playlist/core/browser/playlist_exclusions.cc
```diff
@@ -3,29 +3,33 @@
...
 namespace {
 
-constexpr char kPlaylistExclusionsFile[] = "playlist_exclusions.json";
 constexpr char kRules[] = "rules";
 constexpr char kRegistrableDomain[] = "registrable_domain";
 constexpr char kDenyRootPath[] = "deny_root_path";
 constexpr char kPathPrefixes[] = "path_prefixes";
 
+// IN-TEST
+base::OnceClosure* g_on_load_complete_for_testing = nullptr;
```
bridiver: Using globals and test only is an anti-pattern that should be avoided. This seems like it could be handled more simply with just a standard public observer?

bridiver: or you could even keep this mostly as-is, but make g_on_load_complete_for_testing an instance var instead of global

EliHini: I'm not a fan of the globals as well. I thought maybe the public observer might have been an overkill. Let me revise it

components/playlist/core/browser/playlist_exclusions.cc
```diff
@@ -80,59 +119,47 @@ PlaylistExclusions::PlaylistExclusions() = default;
...
       FROM_HERE,
       {base::MayBlock(), base::TaskPriority::BEST_EFFORT,
        base::TaskShutdownBehavior::SKIP_ON_SHUTDOWN},
-      base::BindOnce(&brave_component_updater::GetDATFileAsString,
-                     component_path_.AppendASCII(kPlaylistExclusionsFile)),
-      base::BindOnce(&PlaylistExclusions::OnPlaylistExclusionsLoaded,
-                     weak_factory_.GetWeakPtr()));
+      base::BindOnce(&LoadPlaylistExclusionsFromFile, exclusions_file_path),
+      base::BindOnce(
+          [](base::WeakPtr<PlaylistExclusions> exclusions,
```
bridiver: This is not necessary, just bind a normal callback with weak_factory_.GetWeakPtr() like it was originally

components/playlist/core/browser/playlist_exclusions.cc
bridiver: actually you can just add an optional callback to LoadPlaylistExclusions

EliHini: I'm thinking of switching it to an instance variable instead of the global retaining the same test hook pattern, no file-scope state on the singleton.

components/playlist/core/browser/playlist_exclusions.cc
```diff
@@ -144,7 +161,7 @@ bool PlaylistExclusions::CanResolvePageSrcLater(const GURL& url) const {
 
   const std::string_view path = url.path();
 
-  for (const PlaylistResolveRule& rule : rules_) {
+  for (const PlaylistResolveRule& rule : *rules_) {
     if (rule.registrable_domain != domain) {
       continue;
     }
```
netzenbot, line 148: The `absl::Cleanup cleanup_for_test_only` block runs on every production call to `OnPlaylistExclusionsLoaded`, not just in tests — it creates a lambda and checks `on_load_complete_for_testing_` on every invocation. Per CSA-035, test-specific workarounds should not affect production behavior. Per the ongoing thread, `@bridiver` suggested passing an optional callback directly to `LoadPlaylistExclusions` instead, which would keep all test-hook wiring at the call site and leave `OnPlaylistExclusionsLoaded` free of test code. https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-035

components/playlist/core/browser/playlist_exclusions.cc
```diff
@@ -80,59 +116,40 @@ PlaylistExclusions::PlaylistExclusions() = default;
...
   base::ThreadPool::PostTaskAndReplyWithResult(
       FROM_HERE,
       {base::MayBlock(), base::TaskPriority::BEST_EFFORT,
        base::TaskShutdownBehavior::SKIP_ON_SHUTDOWN},
-      base::BindOnce(&brave_component_updater::GetDATFileAsString,
-                     component_path_.AppendASCII(kPlaylistExclusionsFile)),
+      base::BindOnce(&LoadPlaylistExclusionsFromFile, exclusions_file_path),
       base::BindOnce(&PlaylistExclusions::OnPlaylistExclusionsLoaded,
                      weak_factory_.GetWeakPtr()));
 }
```
netzenbot: nit: The definition order in `.cc` doesn't match the declaration order in `.h`. `SetOnLoadCompleteForTesting` is defined before `LoadPlaylistExclusions` here, but in the header `LoadPlaylistExclusions` comes first. See [CS-017](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-017).

components/playlist/core/browser/playlist_exclusions.h
```diff
@@ -61,13 +64,17 @@ class PlaylistExclusions {
```
netzenbot, line 63: nit: `LoadPlaylistExclusions` is a new public API replacing `OnComponentReady` but has no doc comment. Given it triggers async loading on the thread pool and the caller needs to know that (e.g., results arrive later via callback), a brief comment describing the contract — especially threading and what `exclusions_file_path` should point to — would help. See [DOC-004](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-004).

components/playlist/core/browser/playlist_exclusion_unittest.cc
```diff
@@ -34,26 +43,47 @@ class PlaylistExclusionsUnitTest : public testing::Test {
```
netzenbot, line 45: nit: `EmptyContentsNotReady`, `InvalidJsonNotReady`, and `MissingRulesListNotReady` are structurally identical — each creates a temp dir, writes a different string, calls `LoadExclusionsAndWait`, then checks `!rules_.has_value()`. Per [TI-023](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-023), these should be a single parameterized test with a `struct { const char* contents; }` case list.

components/playlist/core/browser/playlist_exclusions.cc
bridiver: imo the optional callback is the cleanest and doesn't require and ForTesting methods

components/playlist/core/browser/playlist_exclusions.cc
```diff
@@ -163,4 +159,15 @@ bool PlaylistExclusions::CanResolvePageSrcLater(const GURL& url) const {
   return true;
 }
 
+PlaylistExclusions::PlaylistExclusions() = default;
+
+void PlaylistExclusions::OnPlaylistExclusionsLoaded(
+    base::OnceClosure on_complete,
+    std::optional<std::vector<PlaylistResolveRule>> rules) {
```
bridiver, line 170: the vector doesn't need to be optional, that only complicates checks since an empty vector is equivalent to nullopt in functionality

components/playlist/core/browser/playlist_exclusions.cc
```diff
@@ -42,6 +42,41 @@ bool GetPathPrefixesFromValue(const base::Value* value,
...
+  }
+
+  const base::ListValue* rules_list = root->FindList(kRules);
+  if (!rules_list) {
+    return std::nullopt;
+  }
+
+  std::vector<PlaylistResolveRule> rules;
+  base::JSONValueConverter<PlaylistResolveRule> converter;
+  for (const base::Value& rule_val : *rules_list) {
```
bridiver, line 73: nit
```
for (const base::Value& rule_val : *rules_list) {
    PlaylistResolveRule rule;
    if (converter.Convert(rule_val, &rule) &&
        !rule.registrable_domain.empty()) {
      rules.push_back(std::move(rule));
    }
  }
```

components/playlist/core/common/features.cc
bridiver: are you planning to make this change?

components/playlist/core/browser/playlist_exclusions.cc
EliHini: I guess my ObjC/Swift side is betraying me here :)

components/playlist/core/browser/playlist_exclusions.cc
```diff
@@ -42,6 +43,38 @@ bool GetPathPrefixesFromValue(const base::Value* value,
   return true;
 }
 
+std::vector<PlaylistResolveRule> LoadPlaylistExclusionsFromFile(
+    const base::FilePath& exclusions_file_path) {
+  const std::string contents =
+      brave_component_updater::GetDATFileAsString(exclusions_file_path);
+  if (contents.empty()) {
```
netzenbot, line 54: nit: `LoadPlaylistExclusionsFromFile` has three distinct early-return failure paths but none have explanatory comments. The original `OnPlaylistExclusionsLoaded` had them ("Bail out if the component file was not loaded", "Skip malformed JSON payloads", "Skip payloads without a rules list"); the extracted helper should carry them forward. See [CS-021](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-021).

components/playlist/core/common/features.cc
EliHini: Reverted. I had made that change but after running it through claude code `/review` it reported that I had violated.

Image

EliHini: Fixed now

components/playlist/core/browser/playlist_exclusions.cc
EliHini: Fixed. Removed the `absl::Cleanup` / `SetOnLoadCompleteForTesting` path. `LoadPlaylistExclusions` now takes an optional `on_complete` callback; tests pass `TestFuture::GetCallback()` at the call site and production passes the default empty closure.

components/playlist/core/common/features.cc
```diff
@@ -6,10 +6,16 @@
 #include "brave/components/playlist/core/common/features.h"
 
 #include "base/feature_list.h"
+#include "build/build_config.h"
 
 namespace playlist::features {
 
-BASE_FEATURE(kPlaylist, base::FEATURE_DISABLED_BY_DEFAULT);
+BASE_FEATURE(kPlaylist,
```
netzenbot, line 13: nit: IOS-074 says platform-specific feature defaults should be driven by GN build args, not `#if BUILDFLAG(IS_IOS)` conditionals in the `.cc` file. The default should be set via a GN `declare_args()` entry (e.g., `playlist_feature_enabled_by_default = is_ios`) and a generated buildflag header, keeping the C++ source platform-agnostic. https://github.com/brave/brave-core/tree/master/docs/best-practices/ios.md#IOS-074

components/playlist/core/browser/playlist_exclusions_unittest.cc
```diff
@@ -0,0 +1,140 @@
...
+#include "base/files/scoped_temp_dir.h"
+#include "base/test/task_environment.h"
+#include "base/test/test_future.h"
+#include "brave/components/playlist/core/common/constants.h"
+#include "testing/gtest/include/gtest/gtest.h"
+#include "url/gurl.h"
+
+namespace playlist {
+
+class PlaylistExclusionsUnitTest : public testing::Test {
```
bridiver, line 19: this is why I suggested fixing the filename in a follow-up. git doesn't detect the move so this looks like a new file instead of a change. It's better to handle rename and similar refactoring in separate PRs

components/playlist/core/browser/playlist_exclusions.cc
```diff
@@ -42,6 +43,41 @@ bool GetPathPrefixesFromValue(const base::Value* value,
   return true;
 }
 
+std::pair<bool, std::vector<PlaylistResolveRule>>
```
bridiver: how did we get to `std::pair<bool, vector>`? Why do we care whether it failed or not if the result is the same as just returning an empty vector? If we did need this information it would be better to return optional.

components/playlist/core/browser/playlist_exclusions_unittest.cc
EliHini: My apologies, I completely missed the part on the follow-up. I normally do renames in separate commits as well for the same reasons.

components/playlist/core/browser/playlist_exclusions.cc
EliHini: We do need to distinguish failure from success. On reload, a failed parse should keep the last valid rules instead of clearing them. An empty vector alone cannot tell failure apart from a successful `"rules": []` payload. Yeah, I agree that `pair` is awkward; I'm switching it to `optional<vector>` on the async loader return only. `rules_` stays a plain vector.

components/playlist/core/browser/playlist_exclusion_unittest.cc
```diff
@@ -34,26 +39,66 @@ class PlaylistExclusionsUnitTest : public testing::Test {
...
+TEST_F(PlaylistExclusionsUnitTest, FailedLoadsLeaveRulesEmpty) {
   PlaylistExclusions* exclusions = PlaylistExclusions::GetInstance();
-  exclusions->OnPlaylistExclusionsLoaded(R"({"version": 1})");
-  EXPECT_FALSE(exclusions->is_ready_);
+
+  const struct {
+    const char* trace_name;
+    const char* contents;
+  } kCases[] = {
+      {"EmptyContents", ""},
```
netzenbot, line 88: nit: `ASSERT_TRUE(exclusions->rules_.empty())` and `EXPECT_TRUE(exclusions->rules_.empty())` in the new tests check private implementation state directly. Prefer asserting via the public API — e.g., `EXPECT_TRUE(exclusions->CanResolvePageSrcLater(GURL("https://example.com/")))` — so the tests remain valid if the internal representation changes. https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-002

components/playlist/core/browser/playlist_exclusions.cc
bridiver: Oh right. I mean I guess it's possible we could intentionally return empty rules to disable it or something

EliHini: yup, 100%

cgjangid:
any idea when this will be merged and released?

components/playlist/core/browser/playlist_exclusions_component_installer.cc
```diff
@@ -0,0 +1,147 @@
...
+                      base::DictValue manifest) override;
+  base::FilePath GetRelativeInstallDir() const override;
+  void GetHash(std::vector<uint8_t>* hash) const override;
+  std::string GetName() const override;
+  update_client::InstallerAttributes GetInstallerAttributes() const override;
+  bool IsBraveComponent() const override;
+};
+
+PlaylistExclusionsComponentInstallerPolicy::
+    PlaylistExclusionsComponentInstallerPolicy() = default;
```
bridiver: this should CHECK that playlist::features::kPlaylist is enabled

components/playlist/core/browser/playlist_exclusions.cc
```diff
@@ -163,4 +158,17 @@ bool PlaylistExclusions::CanResolvePageSrcLater(const GURL& url) const {
   return true;
 }
 
+PlaylistExclusions::PlaylistExclusions() = default;
```
bridiver: CHECK that `playlist::features::kPlaylist` is enabled

components/playlist/core/browser/playlist_exclusions.cc
```diff
@@ -163,4 +158,17 @@ bool PlaylistExclusions::CanResolvePageSrcLater(const GURL& url) const {
...
+
+void PlaylistExclusions::OnPlaylistExclusionsLoaded(
+    base::OnceClosure on_complete,
+    std::optional<std::vector<PlaylistResolveRule>> rules) {
+  if (rules) {
+    rules_ = std::move(*rules);
+  }
+  if (!on_complete.is_null()) {
```
bridiver: this should never be null

components/playlist/core/common/features.cc
bridiver: This is a misinterpretation of that best practice and it probably needs to be clarified. You absolutely should not create `playlist_feature_enabled_by_default` to control this. Platforms specific code (including whether a feature is enabled or not by default) should use a platform buildflag. What we want to avoid is putting feature specific code behind a platform buildflag

bridiver: also I don't know why that best practice would be ios specific

components/playlist/core/browser/playlist_exclusion_unittest.cc
EliHini: Addressed in FailedLoadsLeaveRulesEmpty, which combines those three cases into one loop with `SCOPED_TRACE` per [TI-023.](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-023)

components/playlist/core/browser/playlist_exclusion_unittest.cc
EliHini: Resolved in https://github.com/brave/brave-core/pull/36790/changes/277536f6b15a57ebf9ff5b3ebe519e7d6e8749c5

bridiver (approved):

netzenbot (approved):

PR #36791, opened 2026-05-27
[plaster] `DeviceInfoSyncBridge` 🩹🩹🩹🩹

cdesouza-chromium:
This PR migrates all the uses of `#define` macros involving
`DeviceInfoSyncBridge` over onto plasters. These overrides have been a
reccurring cause for bugs, due to how a lot of code duplication was in
place to instantiate `DeviceInfo` from the protobuffers input.

In particular, `BraveSpecificsToModel` is gone now, and we merely insert
the instantiation for our feed in the upstream function, which will
prevent our implementation from sliding going forward.

A move ctor has been added to `DeviceInfo`, to be able to move tha value
returned by `SpecificsToModel` into a `unique_ptr<DeviceInfo>`.
Unfortunately this is necessary now, because `SpecificsToModel` return
`DeviceInfo` by value, rather than wrapped in a `unique_ptr` like
`BraveSpecificsToModel` was doing. Upstream has not added a move ctor to
this type because they emplace it inside a map, which means it has a
stable storage.

In a follow up PR, the use of `unique_ptr<DeviceInfo>` will be dropped
everywhere, now that we have a move ctor.

Bug: https://github.com/brave/brave-browser/issues/55018


netzenbot (approved):

rewrite/components/sync_device_info/device_info_sync_bridge.cc.yaml
```diff
@@ -0,0 +1,89 @@
...
+
+  - description: |
+      Disable the local-device re-upload block in `ApplyIncrementalSyncChanges`.
+
+      These macro guards disable Chromium's block which detects whether the
+      local device record should be re-uploaded to server. We disable it
+      because it breaks the ability to remove other device in sync chain for
+      Brave
+    re_pattern:
+      '(    // Reupload local device if it was deleted from the
```
AlexeyBarabash, line 46: it seems this violates the rule about spaces at the pattern
```
# ❌ WRONG - pattern requires exact whitespace match, fragile to upstream changes
pattern: '    MyMethod()'  # Breaks if upstream changes indentation
replacement: '    BraveMethod()'
```
isn't it?

AlexeyBarabash: from https://github.com/brave-experiments/brave-core-tools/blob/master/docs/best-practices/plaster.md#-use-pattern-for-simple-symbol-replacement-pattern_re-for-context-aware-matches

AlexeyBarabash (approved):
LGTM, thanks

PR #36793, opened 2026-05-27
[plaster][cr150] `NavigationRequest` 🩹🩹🩹

cdesouza-chromium:
This broke in `cr150`. This change adds a plaster that would have been
resilient to the macro breakage encountered.

Chromium changes:
https://chromium.googlesource.com/chromium/src/+/b91bc19194d8d74a1e18600c2f67537e7c3d2bcf

commit b91bc19194d8d74a1e18600c2f67537e7c3d2bcf
Author: Vlad Krot <vkrot@google.com>
Date:   Fri May 22 11:08:23 2026 -0700

    [IWA] enforce headers on navigations and in workers.

    This CL makes sure that in Isolated Web Apps CSP, COEP, COOP, CORP
    headers are injected on navigation / shared worker main script load /
    dedicated worker main script load.

    Previously, the headers were only injected via
    components/webapps/isolated_web_apps/url_loading/url_loader_factory.h,
    however, that is not enough because ServiceWorker interception bypasses
    url_loader_factory and correct headers were not injected.

    Change-Id: Iad801ac9e59f225521de44bc3bd9aa9c025c2f93
    Fixed: 500475136
    Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/7859854
    Reviewed-by: Andrew Rayskiy <greengrape@google.com>
    Commit-Queue: Vlad Krot <vkrot@google.com>
    Cr-Commit-Position: refs/heads/main@{#1635085}

Bug: https://github.com/brave/brave-browser/issues/55302

### DIFF
--- a/patches/content-browser-renderer_host-navigation_request.cc.patch
+++ b/patches/content-browser-renderer_host-navigation_request.cc.patch
@@ -1,20 +1,26 @@
 diff --git a/content/browser/renderer_host/navigation_request.cc b/content/browser/renderer_host/navigation_request.cc
-index 9ec42c6dc1c2ce18e448d905087f6ae7fe96d94c..26e464a7d941ebf56e6caac018423a632dcc731d 100644
+index 9ec42c6dc1c2ce18e448d905087f6ae7fe96d94c..7979fd99332eca0968c83c29013b49a90ad5630a 100644
 --- a/content/browser/renderer_host/navigation_request.cc
 +++ b/content/browser/renderer_host/navigation_request.cc
-@@ -3783,6 +3783,7 @@ void NavigationRequest::OnRequestRedirected(
+@@ -3783,6 +3783,10 @@ void NavigationRequest::OnRequestRedirected(
    common_params_->url = redirect_info.new_url;
    common_params_->method = redirect_info.new_method;
    common_params_->referrer->url = GURL(redirect_info.new_referrer);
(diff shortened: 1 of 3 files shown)


patches/content-browser-renderer_host-navigation_request.cc.patch
```diff
@@ -1,20 +1,25 @@
...
    headers.MergeFrom(TakeModifiedRequestHeaders());
    begin_params_->headers = headers.ToString();
++  GetContentClient()->browser()->MaybeHideReferrer(
++      frame_tree_node_->navigator().controller().GetBrowserContext(),
++      common_params_->url, GetTopDocumentGURL(frame_tree_node_),
++      &common_params_->referrer);
  
-+  BRAVE_ONSTARTCHECKSCOMPLETE_MAYBEHIDEREFERRER
    // TODO(clamy): Avoid cloning the navigation params and create the
    // ResourceRequest directly here.
```
netzenbot, line 26: The updated patch inlines the `MaybeHideReferrer` call as 3-4 lines directly in the patch, which violates the rule against multiline patches ([PATCH-004](https://github.com/brave/brave-core/tree/master/docs/best-practices/patches.md#PATCH-004)). The `rewrite/` YAML handles the substitution logic, but the patch itself should remain single-line if possible — or if the macro approach is truly unworkable for cr150, a comment explaining why the multiline patch is unavoidable here would help.

PR #36805, opened 2026-05-28
[csp]: Remove option to disable CSP

### LINKED ISSUE #55884: [csp]: We should remove the option to disable trusted types from `CreateWebUIDataSource`

fallaciousreasoning:
Resolves https://github.com/brave/brave-browser/issues/55884

The only UI still using this was the welcome page, which I converted to an esmodule (required because it uses bundle splitting). Welcome page still works nicely.

## Test Plan:

Launch a new instance of Brave and got through the welcome flow - nothing should break.


diracdeltas (approved):
thx for doing this!

PR #36808, opened 2026-05-28
Fix flicker dragging tabs into a collapsed vertical strip across windows

### LINKED ISSUE #32932: Can’t reattach tab when using always collapsed vertical tabs
## Description 
When using always collapsed vertical tabs (the ones that do not expand even if you mouse hover over them), it's possible do deattach tabs, but it's not possible to attach them back. This functionality works if tabs are not collapsed.

## Steps to Reproduce
   1. Setup Brave to have collapsed vertical tabs that do not automatically expand.
   2. Deattach one tab by dragging it out.
   3. Try to reattach the tab by dragging it in back.

## Actual result:
The tab will not be reattached, and some rendering issues happen in the detached tab (page stutters white).

## Expected result:
The tab should be reattached to the main window.

## Reproduces how often: 
Easily reproduced

## Brave version (brave://version info)

1.57.62 Chromium: 116.0.5845.180 (Official Build) (64-bit)
Revision 77ae816289bb4557b7e04b3ff2e52d091ab68d66
OS Windows 10 Version 20H2 (Build 19042.2965)

## Miscellaneous Information
https://github.com/brave/brave-browser/assets/42127323/8fdcf0cc-2dc6-4842-97e1-109a279ef162

tenmo2003:
When a tab is dragged from one window into another whose vertical tab strip is collapsed, ExpandTabStripForDragging arms a 200ms width animation (collapsed -> expanded) and tries to snap the strip via SetSize(GetPreferredSize()). But GetPreferredSize() interpolates against the in-flight animation and returns ~collapsed width on the first call, so the snap is silently a no-op. For the next ~200ms, every animation frame triggers OnBoundsChanged which force-relayouts the tab container under the active drag, recomputing the dragged-tab insertion index against a different width each frame -- the visible flicker.

When the destination is already expanded ExpandTabStripForDragging returns early, which is why the bug is invisible in that case.

Add a snap_to_expanded parameter to ExpandTabStripForDragging that, when true, resets the width animation to its end value before SetSize so it actually lands at the full expanded width. The destination call in DetachAndAttachToNewContext passes true so the strip snaps in a single frame. The source-side call keeps the default false; its existing micro-animation is immediately reversed by the resetter destruction and is invisible.

**THIS PR WAS MADE WITH THE HELP OF AI, I JUST WANT THE BUG TO GET FIXED**

Resolves https://github.com/brave/brave-browser/issues/32932

### DIFF
--- a/browser/ui/views/frame/vertical_tabs/vertical_tab_strip_region_view.cc
+++ b/browser/ui/views/frame/vertical_tabs/vertical_tab_strip_region_view.cc
@@ -568,7 +568,8 @@ void BraveVerticalTabStripRegionView::UpdateStateAfterDragAndDropFinished(
 }
 
 BraveVerticalTabStripRegionView::ScopedStateResetter
-BraveVerticalTabStripRegionView::ExpandTabStripForDragging() {
+BraveVerticalTabStripRegionView::ExpandTabStripForDragging(
+    bool snap_to_expanded) {
   if (state_ == State::kExpanded) {
     return {};
   }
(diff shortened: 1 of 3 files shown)


Teddy131:
PLEASE ACCEPT THIS PULL REQUEST FINALLY. WE ARE SUFFERING FROM THIS ISSUE FOR SO LONG. PLEASEEE

PR #36810, opened 2026-05-28
[query-filter] Refactor query parameter stripping logic to improve code readability

### LINKED ISSUE #55717: Refactor QueryStripParameter method to improve code comprehensibility
The current method [QueryStripParameter](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/components/query_filter/browser/utils.cc?L52) can be improved like this [patch](https://github.com/brave/brave-core/blob/328ef3512ad277502a7ca388e65a8d53d85ce9e7/components/query_filter/browser/utils.cc#L53). Follow-up on this change only **after** the [May 27th branch point](https://github.com/brave/brave-browser/wiki/Brave-Release-Schedule).

r0hit0303:
This change is a follow-up to refactor the [StripQueryParameter](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/components/query_filter/browser/utils.cc?L52). 

The refactor: 

a) Adds general improvements: 

- Improved variable names to reflect more concisely what they are capturing. For example `input_kv_strings` to `tokens` as it's not guaranteed that it will contain a key-value string all the time. 
- Rely on c++ auto bindings to get hold of param and value. 
- Replace the int `disallowed_count` with a bool `did_strip`. 
- Rely on an existing `base::FindOrNull` to check if an entry occurs inside a map instead of proxying via the member method [count](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/components/query_filter/browser/utils.cc?L77). 
- Adds comment for non-trivial filtering logic with examples.

b) Replaces the use of `base::SplitStringPiece` to `base::SplitStringOnce` when splitting a single token from `tokens`. 

c) Unwinds the complicated `if` [statement](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/components/query_filter/browser/utils.cc?L75-L79) into smaller chunks to capture the flow better.

Resolves https://github.com/brave/brave-browser/issues/55717


r0hit0303:
@fmarier : Could you ptal at this follow-up refactor? Thanks

components/query_filter/browser/utils.cc
```diff
@@ -51,40 +52,75 @@ static constexpr auto kExemptedHostnames =
...
-  // Set of all the params which are blocked for |spec| from our rules.
-  const auto& rules = query_filter::QueryFilterData::GetInstance()->rules();
-  const base::flat_set<std::string> blocked_params_set =
-      query_filter::GetBlocklistedParamsForSpec(base::span(rules), spec);
+    const auto [param, value] = param_value.value();
+    // For cases like "https://example.com/?fbclid=&foo=1",
+    // which would have the first |token| set as "fbclid=" which when split
+    // around '=' would have an empty value string. Similarly for param and
+    // value with a URL like "https://example.com/?=&foo=1".
+    if (param.empty() || value.empty()) {
```
netzenbot, line 88: nit: The refactor introduces three explicitly-documented edge cases (`?fbclid=&foo=1`, `?=&foo=1`, `?fbclid==`) with no test coverage. These are testable via `MaybeApplyQueryStringFilter` since GURL preserves these query strings. Consider adding a few parameterized cases to cover them. [TI-039](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-039)

r0hit0303: There are tests covered for it at https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/browser/net/brave_site_hacks_network_delegate_helper_unittest.cc?L200-220

components/query_filter/browser/utils.cc
```diff
@@ -51,40 +52,75 @@ static constexpr auto kExemptedHostnames =
...
+        value.cbegin(), value.cend(), [](const char ch) { return ch == '='; });
+    // The value is all "=" so we keep the param.
+    if (itr == value.cend()) {
+      output_tokens.emplace_back(token);
+      continue;
     }
+
+    if (blocked_params_set.contains(param)) {
+      did_strip = true;
+      continue;
```
fmarier: Let's make it explicit since this is where the actual stripping takes place.
```suggestion
      continue;  // Not adding to output_tokens
```

components/query_filter/browser/utils.cc
```diff
@@ -51,40 +52,75 @@ static constexpr auto kExemptedHostnames =
...
+    }
+
+    // Check next for conditional query parameter stripping which is
+    // completely independent of the |blocked_params_set|.
+    if (const auto* pattern =
+            base::FindOrNull(kConditionalQueryStringTrackers, param);
+        pattern && !re2::RE2::PartialMatch(spec, pattern->data())) {
+      did_strip = true;
+      continue;
```
fmarier:
```suggestion
      continue;  // Not adding to output_tokens
```

components/query_filter/browser/utils.cc
```diff
@@ -51,40 +52,75 @@ static constexpr auto kExemptedHostnames =
...
     }
+
+    if (blocked_params_set.contains(param)) {
+      did_strip = true;
+      continue;
+    }
+
+    // Check next for conditional query parameter stripping which is
+    // completely independent of the |blocked_params_set|.
```
fmarier, line 109: Good point about these being "completely independent". Let's add that invariant to our CI checks in https://github.com/brave/adblock-lists/tree/master/.github/workflows similar to the custom checks we have for the debounce list.

(This is obviously a follow-up, it's not related to this PR.)

components/query_filter/browser/utils.cc
```diff
@@ -51,40 +52,75 @@ static constexpr auto kExemptedHostnames =
...
+    if (const auto* pattern =
+            base::FindOrNull(kConditionalQueryStringTrackers, param);
+        pattern && !re2::RE2::PartialMatch(spec, pattern->data())) {
+      did_strip = true;
+      continue;
+    }
+
+    output_tokens.emplace_back(token);
```
fmarier:
```suggestion
    // param wasn't found in any of our lists
    output_tokens.emplace_back(token);
```

fmarier (approved):

components/query_filter/browser/utils.cc
fmarier: See https://github.com/brave/adblock-lists/pull/2900.

netzenbot (approved):

r0hit0303:
Thanks @fmarier for your thoughtful review. All changes made in the latest ps. 

For posterity mentioning our offline discussion around this change adding extra logic to handle explicitly cases like "example.com?fblcid==" with more than one "=". 

The [old code](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/components/query_filter/browser/utils.cc?L73) was relying on [SPLIT_WANT_NONEMPTY](https://source.chromium.org/gn/gn/+/main:src/base/strings/string_split.h;l=27-33) was returning an empty vector for pieces, which implicitly hit the else case [here](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/components/query_filter/browser/utils.cc?L82) and adding the original string in the `output_kv_strings` vector. This change made that behaviour more explicit as a lot was getting abstracted away from `SplitStringPiece`. 

That said, we think that it could make sense to update this behaviour to actually remove these tracking params. However, it may potentially introduce webcompat issues for sites that were just checking for the existence of these parameters.. therefore it makes sense to follow-up on that change in a separate CL for cleaner history.

components/query_filter/browser/utils.cc
r0hit0303: Thanks for looking into this :)

PR #36814, opened 2026-05-28
[FocusMode] Add FocusModeTopOverlay view

zenparsing:
Hosts the browser's top container and horizontal tab strip during focus mode. Animates y-offset to slide the overlay into view when required.

### DIFF
--- a/browser/ui/views/frame/BUILD.gn
+++ b/browser/ui/views/frame/BUILD.gn
@@ -8,6 +8,8 @@ source_set("frame") {
     "brave_contents_view_util.h",
     "brave_side_panel_shadow_overlay_view.h",
     "focus_mode_title_bar_view.h",
+    "focus_mode_top_overlay.cc",
+    "focus_mode_top_overlay.h",
     "layout/brave_browser_view_layout_delegate_impl.cc",
     "layout/brave_browser_view_layout_delegate_impl.h",
     "layout/brave_browser_view_tabbed_layout_impl.cc",
@@ -17,6 +19,7 @@ source_set("frame") {
 
   public_deps = [
     "//base",
+    "//brave/browser/ui/views/frame/edge_reveal",
     "//components/tabs:public",
     "//ui/views",
   ]
@@ -89,6 +92,7 @@ source_set("browser_tests") {
   sources = [
     "//brave/browser/ui/views/frame/brave_browser_view_browsertest.cc",
     "//brave/browser/ui/views/frame/brave_horizontal_tab_strip_region_view_browsertest.cc",
+    "//brave/browser/ui/views/frame/focus_mode_top_overlay_browsertest.cc",
     "//brave/browser/ui/views/frame/tab_strip_placement_coordinator_browsertest.cc",
   ]
 
@@ -98,6 +102,7 @@ source_set("browser_tests") {
     "//base",
     "//base/test:test_support",
     "//brave/browser/ui:brave_tab_prefs",
+    "//brave/browser/ui/focus_mode",
     "//brave/browser/ui/sidebar:sidebar_impl",
     "//brave/browser/ui/views/frame/vertical_tabs",
     "//brave/browser/ui/views/sidebar",
--- a/browser/ui/views/frame/brave_browser_frame_view_mac.mm
+++ b/browser/ui/views/frame/brave_browser_frame_view_mac.mm
@@ -7,6 +7,9 @@
 
 #include <memory>
 
+#include "base/check_deref.h"
+#include "base/memory/raw_ref.h"
+#include "brave/browser/ui/focus_mode/focus_mode_controller.h"
 #include "brave/browser/ui/tabs/brave_tab_prefs.h"
 #include "brave/browser/ui/views/frame/brave_non_client_hit_test_helper.h"
 #include "brave/browser/ui/views/frame/brave_window_frame_graphic.h"
@@ -26,6 +29,25 @@
 #include "ui/gfx/geometry/outsets_f.h"
 #include "ui/gfx/geometry/rect.h"
 #include "ui/gfx/scoped_canvas.h"
+
+class BraveBrowserFrameViewMac::ScopedFocusModeDisable {
+ public:
+  explicit ScopedFocusModeDisable(FocusModeController* controller)
+      : controller_(CHECK_DEREF(controller)),
+        was_enabled_(controller->IsEnabled()) {
+    controller_->SetEnabled(false);
+  }
+
+  ~ScopedFocusModeDisable() {
+    if (was_enabled_) {
+      controller_->SetEnabled(true);
+    }
+  }
+
+ private:
+  const raw_ref<FocusModeController> controller_;
+  bool was_enabled_;
+};
 
 BraveBrowserFrameViewMac::BraveBrowserFrameViewMac(
     BrowserWidget* browser_widget,
@@ -199,6 +221,27 @@ gfx::Size BraveBrowserFrameViewMac::GetMinimumSize() const {
   }
 
   return BrowserFrameViewMac::GetMinimumSize();
+}
+
+void BraveBrowserFrameViewMac::OnFullscreenStateChanged() {
+  // When entering fullscreen ensure that focus mode is disabled. Focus mode and
+  // immersive fullscreen on MacOS have conflicting presentation requirements,
+  // and both modes want to move the tabstrip into different parent views.
+  // Disabling focus mode before immersive mode is enabled ensures that the
+  // tabstrip view is returned to the expected placement before the immersive
+  // controller attempts to reparent it.
+  if (GetBrowserView()->IsFullscreen()) {
+    if (!scoped_focus_mode_disable_) {
+      auto* browser = GetBrowserView()->browser();
+      if (auto* controller = browser->GetFeatures().focus_mode_controller()) {
+        scoped_focus_mode_disable_ =
+            std::make_unique<ScopedFocusModeDisable>(controller);
+      }
+    }
+  } else {
+    scoped_focus_mode_disable_.reset();
+  }
+  BrowserFrameViewMac::OnFullscreenStateChanged();
 }
 
 bool BraveBrowserFrameViewMac::ShouldHideTopUIInFullscreen() const {
--- a/browser/ui/views/frame/brave_browser_view.cc
+++ b/browser/ui/views/frame/brave_browser_view.cc
@@ -23,6 +23,8 @@
 #include "brave/browser/ui/color/brave_color_id.h"
 #include "brave/browser/ui/commands/accelerator_service.h"
 #include "brave/browser/ui/commands/accelerator_service_factory.h"
+#include "brave/browser/ui/focus_mode/focus_mode_features.h"
+#include "brave/browser/ui/focus_mode/focus_mode_utils.h"
 #include "brave/browser/ui/page_info/features.h"
 #include "brave/browser/ui/sidebar/sidebar_controller.h"
 #include "brave/browser/ui/sidebar/sidebar_utils.h"
@@ -33,6 +35,7 @@
 #include "brave/browser/ui/views/frame/brave_contents_layout_manager.h"
 #include "brave/browser/ui/views/frame/brave_contents_view_util.h"
 #include "brave/browser/ui/views/frame/brave_side_panel_shadow_overlay_view.h"
+#include "brave/browser/ui/views/frame/focus_mode_top_overlay.h"
 #include "brave/browser/ui/views/frame/split_view/brave_contents_container_view.h"
 #include "brave/browser/ui/views/frame/split_view/brave_multi_contents_view.h"
 #include "brave/browser/ui/views/frame/tab_strip_placement_coordinator.h"
@@ -92,6 +95,7 @@
 #include "chrome/browser/ui/web_applications/app_browser_controller.h"
 #include "chrome/common/pref_names.h"
 #include "components/javascript_dialogs/tab_modal_dialog_manager.h"
+#include "components/omnibox/browser/location_bar_model.h"
 #include "components/permissions/permission_request_manager.h"
 #include "components/web_modal/web_contents_modal_dialog_manager.h"
 #include "content/public/browser/page_navigator.h"
@@ -397,8 +401,14 @@ BraveBrowserView::BraveBrowserView(Browser* browser) : BrowserView(browser) {
         views::CreateSolidBackground(kColorToolbar));
   }
 
-  if (!supports_vertical_tabs && !can_have_sidebar) {
-    return;
+  if (BrowserSupportsFocusMode(browser_)) {
+    auto* controller = browser_->GetFeatures().focus_mode_controller();
+    CHECK(controller);
+    focus_mode_observation_.Observe(controller);
+
+    focus_mode_top_overlay_ =
+        AddChildView(std::make_unique<FocusModeTopOverlay>(
+            base::PassKey<BraveBrowserView>(), this));
   }
 
   EnsureFindBarHostViewIsLastChild();
@@ -750,6 +760,10 @@ void BraveBrowserView::OnAcceleratorsChanged(
   }
 }
 
+void BraveBrowserView::OnFocusModeToggled(bool enabled) {
+  UpdateFocusModeState();
+}
+
 #if BUILDFLAG(ENABLE_BRAVE_WALLET)
 void BraveBrowserView::CreateWalletBubble() {
   DCHECK(GetWalletButton());
@@ -785,10 +799,17 @@ void BraveBrowserView::AddedToWidget() {
     vertical_tab_strip_container_view_ =
         AddChildView(std::make_unique<BraveVerticalTabStripContainerView>(
             this, vertical_tab_strip_host_view_));
-    EnsureFindBarHostViewIsLastChild();
     GetBrowserViewLayout()->set_vertical_tab_strip_host(
         vertical_tab_strip_host_view_.get());
   }
+
+  UpdateFocusModeState();
+  EnsureFindBarHostViewIsLastChild();
+}
+
+void BraveBrowserView::RemovedFromWidget() {
+  focus_mode_observation_.Reset();
+  BrowserView::RemovedFromWidget();
 }
 
 bool BraveBrowserView::ShowBraveHelpBubbleView(const std::string& text) {
@@ -1026,7 +1047,8 @@ void BraveBrowserView::HideSplitView() {
 }
 
 void BraveBrowserView::ReparentTopContainerForEndOfImmersive() {
-  if (tabs::utils::ShouldShowBraveVerticalTabs(browser())) {
+  if (tabs::utils::ShouldShowBraveVerticalTabs(browser()) ||
+      IsFocusModeEnabled(browser())) {
     return;
   }
 
@@ -1190,7 +1212,20 @@ void BraveBrowserView::OnActiveTabChanged(content::WebContents* old_contents,
                                           content::WebContents* new_contents,
                                           int index,
                                           int reason) {
+  bool tab_change_in_split_view =
+      IsTabChangeInSplitView(old_contents, new_contents);
+
   BrowserView::OnActiveTabChanged(old_contents, new_contents, index, reason);
+
+  // Switching between tabs may change state that is relevant for focus mode
+  // (e.g. when switching between an https tab and an http tab).
+  UpdateFocusModeState();
+
+  // In focus mode, when switching between tabs that aren't split-view pairs
+  // temporarily reveal the location bar.
+  if (focus_mode_top_overlay_ && !tab_change_in_split_view) {
+    focus_mode_top_overlay_->RevealTemporarily(base::Seconds(2));
+  }
 
   // Update UI after active tab changing is handled because
   // ShouldUseBraveWebViewRoundedCornersForContents() check split view UI for
@@ -1240,6 +1275,24 @@ void BraveBrowserView::OnActiveTabChanged(content::WebContents* old_contents,
     CHECK(tab_modal_dialog_manager);
     tab_modal_dialog_manager->OnTabActiveStateChanged();
   }
+}
+
+void BraveBrowserView::UpdateToolbar(content::WebContents* contents) {
+  BrowserView::UpdateToolbar(contents);
+
+  // Re-evaluate focus mode on every active-tab toolbar refresh. Same-tab
+  // navigations that only change the security level between non-secure states
+  // (e.g. an https page to a file:// page) do not produce a security-state
+  // change, so `UpdateToolbarSecurityState` alone is not sufficient.
+  UpdateFocusModeState();
+}
+
+bool BraveBrowserView::UpdateToolbarSecurityState() {
+  bool state_changed = BrowserView::UpdateToolbarSecurityState();
+  if (state_changed) {
+    UpdateFocusModeState();
+  }
+  return state_changed;
 }
 
 bool BraveBrowserView::AcceleratorPressed(const ui::Accelerator& accelerator) {
@@ -1333,6 +1386,22 @@ ClientFrameElementInfo BraveBrowserView::GetFrameElementInfo() const {
 #endif
   }
   return info;
+}
+
+void BraveBrowserView::OnImmersiveFullscreenExited() {
+  BrowserView::OnImmersiveFullscreenExited();
+  tab_strip_placement_->UpdatePlacement();
+}
+
+void BraveBrowserView::OnImmersiveModeControllerDestroyed() {
+  // When the immersive mode controller is destroyed during browser teardown,
+  // ensure that top-reveal views are returned to their original and expected
+  // placement in order to avoid violating view heirarchy assumptions in the
+  // immersive fullscreen controller.
+  if (focus_mode_top_overlay_) {
+    focus_mode_top_overlay_->Deactivate();
+  }
+  BrowserView::OnImmersiveModeControllerDestroyed();
 }
 
 bool BraveBrowserView::IsSidebarVisible() const {
@@ -1431,6 +1500,42 @@ void BraveBrowserView::UpdateWebViewRoundedCorners() {
                          contents_container_view->devtools_docked_placement(),
                          corners);
   }
+}
+
+void BraveBrowserView::UpdateFocusModeState() {
+  bool enabled = IsFocusModeEnabled(browser());
+
+  // If the location bar is showing a warning (e.g. http, cert errors, etc.),
+  // configure the browser view as if Focus Mode is not enabled.
+  if (enabled && ShouldDisableFocusModeForActiveTab()) {
+    enabled = false;
+  }
+
+  if (focus_mode_top_overlay_) {
+    if (enabled) {
+      // Ensure that the overlay is at the end of the child list for correct
+      // z-order rendering.
+      ReorderChildView(focus_mode_top_overlay_, -1);
+      focus_mode_top_overlay_->Activate();
+    } else {
+      focus_mode_top_overlay_->Deactivate();
+    }
+    EnsureFindBarHostViewIsLastChild();
+    InvalidateLayout();
+  }
+}
+
+bool BraveBrowserView::ShouldDisableFocusModeForActiveTab() const {
+  auto* location_bar = GetLocationBar();
+  if (!location_bar) {
+    return true;
+  }
+  auto* model = location_bar->GetLocationBarModel();
+  if (!model) {
+    return true;
+  }
+  auto level = model->GetSecurityLevel();
+  return level != security_state::SecurityLevel::SECURE;
 }
 
 void BraveBrowserView::Layout(PassKey) {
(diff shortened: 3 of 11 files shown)


browser/ui/views/frame/focus_mode_top_overlay.cc
```diff
@@ -0,0 +1,265 @@
...
+  SetPaintToLayer();
+  layer()->SetFillsBoundsOpaquely(true);
+
+  auto* coordinator = GetTabStripPlacementCoordinator(browser_view_.get());
+  CHECK(coordinator);
+  coordinator->SetPlacement(TabStripPlacementKind::kTopContainer,
+                            browser_view_->top_container(), 0);
+
+  animation_.SetSlideDuration(kRevealAnimationDuration);
+  animation_.Reset(1.0);
```
simonhong, line 84: need to reset 1.0 here?

browser/ui/views/frame/focus_mode_top_overlay.cc
```diff
@@ -0,0 +1,265 @@
...
+    StopRevealing();
+  }
+}
+
+void FocusModeTopOverlay::Activate() {
+  if (active_) {
+    return;
+  }
+
+  CHECK(GetWidget());
```
simonhong, line 100: q) GetWidget() is always same with browser_view_->GetWidget()?

browser/ui/views/frame/focus_mode_top_overlay.cc
```diff
@@ -0,0 +1,265 @@
...
+  active_ = true;
+
+  // Move tab strip into top container, if appropriate.
+  UpdateTabStripPlacement(browser_view_.get());
+
+  // Move top container into overlay.
+  auto* top_container = browser_view_->top_container();
+  CHECK(top_container);
+  top_container_index_ = browser_view_->GetIndexOf(top_container);
+  AddChildView(top_container);
```
simonhong, line 118: Q) curious upstream's layout works well if top container is not the brower view's direct child?
I can see `IsParentedTo(views().top_container, views().browser_view)`.

As this PR doesn't add FocusModeTopOverlay as child view to browser view, can't understand how this view's bounds is set.

browser/ui/views/frame/focus_mode_top_overlay.cc
```diff
@@ -0,0 +1,265 @@
...
+  // Restore top container placement.
+  if (auto* top_container = top_container_observation_.GetSource()) {
+    size_t max_index = browser_view_->children().size();
+    browser_view_->AddChildViewAt(
+        top_container,
+        std::min(top_container_index_.value_or(max_index), max_index));
+    top_container_observation_.Reset();
+  }
+
+  // Restore tab strip placement if necessary.
```
simonhong: This comment is correct? I think `if necessary` seems invalid because calling `UpdateTabStripPlacement()` here will change from `kTopContainer` to others always.

browser/ui/views/frame/focus_mode_top_overlay.cc
```diff
@@ -0,0 +1,265 @@
...
+void FocusModeTopOverlay::Deactivate() {
+  if (!active_) {
+    return;
+  }
+
+  active_ = false;
+  StopRevealing();
+
+  // Restore top container placement.
+  if (auto* top_container = top_container_observation_.GetSource()) {
```
simonhong, line 152: This `top_container` could null here?

browser/ui/views/frame/focus_mode_top_overlay.h
```diff
@@ -0,0 +1,125 @@
...
+  // Begins hosting the top container and hides the overlay with an animation.
+  void Activate();
+
+  // Restores the top container views to their previous placements and hides the
+  // overlay.
+  void Deactivate();
+
+  // Returns a value indicating whether the overlay is currently hosting the
+  // top container views.
+  bool IsActive() const { return active_; }
```
simonhong: nit: `active()` is chromium convention?

browser/ui/views/frame/focus_mode_top_overlay.cc
```diff
@@ -0,0 +1,265 @@
...
+  if (ShouldReveal()) {
+    animation_.Show();
+  } else {
+    animation_.Hide();
+  }
+}
+
+void FocusModeTopOverlay::UpdateBounds() {
+  auto* top_container = top_container_observation_.GetSource();
+  if (!top_container || !parent()) {
```
simonhong, line 259: Curious how `UpdateBounds()` could be called when there is no parent.

browser/ui/views/frame/focus_mode_top_overlay.cc
```diff
@@ -0,0 +1,265 @@
...
+    return;
+  }
+  int height = top_container->bounds().height();
```
simonhong, line 262: question how `top_container`'s width is calculated.
By browser view layout (even top_container is not the child of browser view)?

browser/ui/views/frame/focus_mode_top_overlay.cc
```diff
@@ -0,0 +1,265 @@
...
+  }
+
+  // Restore tab strip placement if necessary.
+  UpdateTabStripPlacement(browser_view_.get());
+
+  SetVisible(false);
+  NotifyRevealFractionChanged();
+}
+
+void FocusModeTopOverlay::RevealTemporarily(base::TimeDelta duration) {
```
simonhong, line 171: I assume that this method is used when we want to force to reveal for `duration`.
But, only UpdateRevealState() is called and scheduled.
As `UpdateRevealState()` will show or hide based on `ShouldReveal()`,
I think calling this method could not reveal.

simonhong (commented):
IMO, we could include minimal changes to browser layout for layout overlay view.
Then, we could have browser test for testing this overlay behavior.
Something like
1. set enable to focus mode controller.
2. call Activate()
3. test reveal/hide scenario and some states
4. call Deactivate
5. test some states

WDYT?

browser/ui/views/frame/focus_mode_top_overlay.cc
zenparsing: Technically no, since it is set to `1.0` in `Activate()`. Semantically though, it should be `1.0` since the the overlay starts as fully revealed as we transition to focus mode.

browser/ui/views/frame/focus_mode_top_overlay.cc
zenparsing: Yes. We could add a CHECK, but I don't believe it is an essential invariant.

browser/ui/views/frame/focus_mode_top_overlay.cc
zenparsing: [This branch](https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/ui/views/frame/layout/browser_view_layout_impl.cc;l=229) in `browser_view_layout_impl.cc` handles the case where the top container is not parented to the browser view.

browser/ui/views/frame/focus_mode_top_overlay.cc
zenparsing: Fixed.

browser/ui/views/frame/focus_mode_top_overlay.cc
zenparsing: It could be `null` if `top_container` is destroyed before this view. I suppose that won't be the case, but it seems better to me to avoid non-essential assumptions.

browser/ui/views/frame/focus_mode_top_overlay.h
zenparsing: Fixed.

browser/ui/views/frame/focus_mode_top_overlay.cc
zenparsing: It seems like `parent()` would always be non-null when this method executes. But since that depends on how this class is used (external), it seems better to me to not make that assumption.

browser/ui/views/frame/focus_mode_top_overlay.cc
zenparsing: That's correct.

browser/ui/views/frame/focus_mode_top_overlay.cc
zenparsing: `ShouldReveal` takes into account whether the temporary reveal timer is currently running (if running, then `ShouldReveal` will return true).

browser/ui/views/frame/focus_mode_top_overlay.cc
```diff
@@ -0,0 +1,265 @@
...
+  CHECK(top_container);
+  top_container_index_ = browser_view_->GetIndexOf(top_container);
+  AddChildView(top_container);
+
+  top_container_observation_.Observe(top_container);
+
+  hover_detector_ = std::make_unique<EdgeHoverDetector>(
+      GetWidget(),
+      base::BindRepeating(&FocusModeTopOverlay::IsHoverPoint,
+                          base::Unretained(this)),
```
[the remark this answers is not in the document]
zenparsing: Guarded by sub-object lifetime.

browser/ui/views/frame/focus_mode_top_overlay.cc
```diff
@@ -0,0 +1,265 @@
...
+      base::BindRepeating(&FocusModeTopOverlay::OnHoverStateChanged,
+                          base::Unretained(this)));
```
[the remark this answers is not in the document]
zenparsing: Guarded by sub-object lifetime.

browser/ui/views/frame/focus_mode_top_overlay.cc
```diff
@@ -0,0 +1,265 @@
...
+
+  reveal_watcher_ = std::make_unique<EdgeRevealWatcher>(
+      GetWidget(),
+      base::BindRepeating(&FocusModeTopOverlay::EdgeContainsView,
+                          base::Unretained(this)),
```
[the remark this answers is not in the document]
zenparsing: Guarded by sub-object lifetime.

browser/ui/views/frame/focus_mode_top_overlay.cc
```diff
@@ -0,0 +1,265 @@
...
+      base::BindRepeating(&FocusModeTopOverlay::OnRevealReasonsChanged,
+                          base::Unretained(this)));
```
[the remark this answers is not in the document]
zenparsing: Guarded by sub-object lifetime.

browser/ui/views/frame/focus_mode_top_overlay.cc
```diff
@@ -0,0 +1,265 @@
...
+}
+
+void FocusModeTopOverlay::RevealTemporarily(base::TimeDelta duration) {
+  if (!active_) {
+    return;
+  }
+  temporary_reveal_timer_.Start(
+      FROM_HERE, duration,
+      base::BindRepeating(&FocusModeTopOverlay::UpdateRevealState,
+                          base::Unretained(this)));
```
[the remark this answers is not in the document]
zenparsing: Guarded by sub-object lifetime.

browser/ui/views/frame/layout/brave_browser_view_tabbed_layout_impl.cc
```diff
@@ -389,9 +409,11 @@ void BraveBrowserViewTabbedLayoutImpl::CalculateBraveVerticalTabStripLayout(
...
     CHECK(views().top_container);
-    auto* top_container_layout = layout.GetLayoutFor(views().top_container);
-    CHECK(top_container_layout);
-    return top_container_layout->bounds.bottom() - GetContentsMargins().top();
+    if (auto* top_layout = layout.GetLayoutFor(views().top_container)) {
+      return top_layout->bounds.bottom() - GetContentsMargins().top();
+    }
+
+    return GetContentsMargins().top();
   };
```
netzenbot, line 371: The original `CHECK(top_container_layout)` has been replaced with a silent fallback to `GetContentsMargins().top()`. Since focus mode is handled by the early-return block above, this fallback path should never be reached in non-focus-mode; dropping the CHECK silently hides unexpected states. Prefer keeping the CHECK or adding a NOTREACHED() here to preserve the original crash-on-bug behavior. See [ARCH-032](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-032).

browser/ui/views/frame/BUILD.gn
```diff
@@ -20,6 +22,7 @@ source_set("frame") {
 
   deps = [
     "//brave/browser/ui:brave_tab_prefs",
+    "//brave/browser/ui/views/frame/edge_reveal",
```
netzenbot: `focus_mode_top_overlay.h` includes `edge_reveal_watcher.h` in its public header, so `//brave/browser/ui/views/frame/edge_reveal` must be in `public_deps`, not `deps`. See [BS-024](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-024).

browser/ui/views/frame/brave_browser_view.h
```diff
@@ -173,6 +182,10 @@ class BraveBrowserView : public BrowserView,
     return tab_strip_placement_.get();
   }
 
+  FocusModeTopOverlay* focus_mode_top_overlay() {
+    return focus_mode_top_overlay_;
+  }
```
netzenbot, line 188: nit: `focus_mode_top_overlay()` is used only in the browsertest but doesn't follow the `*_for_testing()` naming convention (see `top_container_separator_for_testing()` in this same class). Consider renaming to `focus_mode_top_overlay_for_testing()` or using a `friend` declaration. [ARCH-020](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-020)

browser/ui/views/frame/layout/brave_browser_view_tabbed_layout_impl.cc
```diff
@@ -515,6 +537,33 @@ void BraveBrowserViewTabbedLayoutImpl::CalculateSideBarLayout(
   layout.AddChild(views().sidebar_container, sidebar_bounds);
 }
 
+void BraveBrowserViewTabbedLayoutImpl::AdjustLayoutForFocusMode(
+    ProposedLayout& layout) const {
+  if (!delegate().IsFocusModeEnabled()) {
```
netzenbot: nit: The declaration order in the `.h` puts `InsetContentsContainerBounds` before `AdjustLayoutForFocusMode`, but the `.cc` defines `AdjustLayoutForFocusMode` first. Per [CS-017](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-017), definitions in the `.cc` should follow the same order as declarations in the `.h`.

browser/ui/views/frame/layout/brave_browser_view_tabbed_layout_impl.cc
zenparsing: The `CHECK` was not valid - crashes have been observed regardless of focus mode state.

browser/ui/views/frame/brave_browser_view.h
zenparsing: `focus_mode_top_overlay()` will be used outside of tests in a follow-up.

browser/ui/views/frame/BUILD.gn
zenparsing: Fixed.

browser/ui/views/frame/layout/brave_browser_view_tabbed_layout_impl.cc
zenparsing: Fixed.

browser/ui/views/frame/layout/brave_browser_view_tabbed_layout_impl.cc
```diff
@@ -389,9 +409,11 @@ void BraveBrowserViewTabbedLayoutImpl::CalculateBraveVerticalTabStripLayout(
...
 
     CHECK(views().top_container);
-    auto* top_container_layout = layout.GetLayoutFor(views().top_container);
-    CHECK(top_container_layout);
-    return top_container_layout->bounds.bottom() - GetContentsMargins().top();
+    if (auto* top_layout = layout.GetLayoutFor(views().top_container)) {
+      return top_layout->bounds.bottom() - GetContentsMargins().top();
+    }
+
+    return GetContentsMargins().top();
```
simonhong, line 369: Q) do you have reason about this change? When focus mode is enabled, it's not reached here.

browser/ui/views/frame/layout/brave_browser_view_tabbed_layout_impl.cc
```diff
@@ -659,6 +681,33 @@ void BraveBrowserViewTabbedLayoutImpl::UpdateMarginsForSideBar() {
   }
 }
 
+void BraveBrowserViewTabbedLayoutImpl::AdjustLayoutForFocusMode(
```
simonhong: In this method, three views are adjusted. contents container, infobar container and contents background. browser view also has many other child views that can be affected by focus mode.
I think this we need to adjust all. Also new child view could be added later.

Is it sufficient to adjust above three child views for focus mode? If not, how can we handle other views?

Image

simonhong: Also layout manager has `DoPostLayoutVisualAdjustments()`.
It's called later and maybe it could have adjustment. Is this method works well with post layout?

browser/ui/views/frame/layout/brave_browser_view_tabbed_layout_impl.cc
```diff
@@ -670,10 +719,20 @@ gfx::Insets BraveBrowserViewTabbedLayoutImpl::GetContentsMargins() const {
...
-  // If there is a visible view above the contents container, then there is no
-  // need for a top margin.
-  if (delegate().ShouldDrawTabStrip() || delegate().IsToolbarVisible() ||
-      delegate().IsBookmarkBarVisible() || delegate().IsInfobarVisible()) {
+  auto contents_at_top_edge = [&]() {
+    if (delegate().IsInfobarVisible()) {
+      return false;
+    }
+    if (IsParentedTo(views().top_container, views().browser_view)) {
+      if (delegate().ShouldDrawTabStrip() || delegate().IsToolbarVisible() ||
```
simonhong, line 610: If focus mode is disabled, tab strip and top container are sibling.

browser/ui/views/frame/brave_browser_frame_view_mac.mm
```diff
@@ -196,6 +218,27 @@
   return BrowserFrameViewMac::GetMinimumSize();
 }
 
+void BraveBrowserFrameViewMac::OnFullscreenStateChanged() {
+  // When entering fullscreen ensure that focus mode is disabled. Focus mode and
```
simonhong, line 227: How about handling this for all platform not just macOS?

browser/ui/views/frame/brave_browser_view.cc
```diff
@@ -794,6 +806,14 @@ void BraveBrowserView::AddedToWidget() {
     GetBrowserViewLayout()->set_vertical_tab_strip_host(
         vertical_tab_strip_host_view_.get());
   }
+
+  // Move focus mode overlay to end of child list to ensure correct z-order.
+  if (focus_mode_top_overlay_) {
+    ReorderChildView(focus_mode_top_overlay_, -1);
```
simonhong: This called only once.
How about reorder overlay whenever focus mode activates?

browser/ui/views/frame/brave_browser_view.cc
```diff
@@ -1415,6 +1461,20 @@ void BraveBrowserView::UpdateWebViewRoundedCorners() {
   }
 }
 
+void BraveBrowserView::UpdateFocusModeState() {
+  bool enabled = IsFocusModeEnabled(browser());
```
simonhong, line 1506: This could be moved inside below if clause.

browser/ui/views/frame/brave_browser_view.cc
simonhong: Also, we need have test for this reordering - overlay and findbar host z-orders.

browser/ui/views/frame/layout/brave_browser_view_tabbed_layout_impl.cc
simonhong: Also needs test about adjusted view's bounds such as infobar, contents background and etc.

browser/ui/views/frame/layout/brave_browser_view_tabbed_layout_impl_unittest.cc
simonhong, line 89: maybe we could use this unit test for some layout manager changes?

browser/ui/views/frame/layout/brave_browser_view_tabbed_layout_impl.cc
zenparsing: I noticed the `CHECK` was being hit occasionally while working on this feature, independent of focus mode. I also saw instances of this crash in Backtrace. I can remove the fix from the PR if preferred, but it should be fixed at some point.

browser/ui/views/frame/layout/brave_browser_view_tabbed_layout_impl.cc
zenparsing: Right. The "contents has view above it" check needed to be changed for focus mode, because the top container views and the tabstrip are in the overlay instead. Similar to upstream, `IsParentedTo` tells us whether we are in overlay mode or not.

browser/ui/views/frame/brave_browser_frame_view_mac.mm
zenparsing: For Windows and Linux platforms, focus mode and fullscreen actually work well together.

browser/ui/views/frame/brave_browser_view.cc
zenparsing: It's needed here for a follow-up (when we add the focus mode title bar).

browser/ui/views/frame/brave_browser_view.cc
zenparsing: Done - and test added.

browser/ui/views/frame/layout/brave_browser_view_tabbed_layout_impl_unittest.cc
zenparsing: Test cases added to layout unit test file. I wasn't quite sure how to test, but followed the existing pattern in that file.

browser/ui/views/frame/layout/brave_browser_view_tabbed_layout_impl.cc
zenparsing: The layout changes seem to work well as-is. Let me know if you have any suggestions.

browser/ui/views/frame/focus_mode_top_overlay_browsertest.cc
```diff
@@ -0,0 +1,105 @@
...
+#include "base/test/scoped_feature_list.h"
+#include "brave/browser/ui/focus_mode/focus_mode_controller.h"
+#include "brave/browser/ui/focus_mode/focus_mode_features.h"
+#include "brave/browser/ui/views/frame/brave_browser_view.h"
+#include "chrome/browser/ui/browser_window/public/browser_window_features.h"
+#include "chrome/browser/ui/views/frame/browser_view.h"
+#include "chrome/browser/ui/views/frame/top_container_view.h"
+#include "chrome/browser/ui/views/tabs/tab_strip.h"
+#include "chrome/browser/ui/views/toolbar/toolbar_view.h"
+#include "chrome/test/base/in_process_browser_test.h"
```
netzenbot, line 24: This test file uses `InProcessBrowserTest`, but per [ARCH-062](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-062) every feature adding user-visible UI (new views, animations) should have at least one CUJ test using `InteractiveBrowserTest`. Consider adding an `InteractiveBrowserTest`-based test that drives the reveal/hide behavior via simulated user input (e.g., mouse hover near the top edge).

browser/ui/views/frame/focus_mode_top_overlay.cc
```diff
@@ -0,0 +1,265 @@
...
+#include <algorithm>
+#include <utility>
+
+#include "base/check.h"
+#include "base/check_deref.h"
+#include "brave/browser/ui/views/frame/brave_browser_view.h"
+#include "brave/browser/ui/views/frame/edge_reveal/edge_hover_detector.h"
+#include "brave/browser/ui/views/frame/tab_strip_placement_coordinator.h"
+#include "chrome/browser/ui/views/frame/top_container_view.h"
+#include "third_party/skia/include/core/SkColor.h"
```
netzenbot, line 17: nit: `top_container_view.h` appears unused — `TopContainerView` is never referenced in this file. [`top_container()`](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-001) returns a generic `views::View*`, so this header can be removed per [CS-001](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-001).

browser/ui/views/frame/layout/brave_browser_view_tabbed_layout_impl.h
```diff
@@ -79,11 +79,14 @@ class BraveBrowserViewTabbedLayoutImpl : public BrowserViewTabbedLayoutImpl {
 
   void UpdateInsetsForVerticalTabStrip();
   void UpdateMarginsForSideBar();
+  void AdjustLayoutForFocusMode(ProposedLayout& layout) const;
 
   gfx::Insets GetContentsMargins() const;
   bool ShouldPushBookmarkBarForVerticalTabs() const;
   gfx::Insets GetInsetsConsideringVerticalTabHost() const;
```
netzenbot, line 114: nit: Per [CS-015](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-015), `friend` declarations should go right after `private:`, before any member functions or variables. Move `friend class BraveBrowserViewTabbedLayoutImplFocusModeTest` to immediately after the `private:` line.

browser/ui/views/frame/focus_mode_top_overlay.cc
zenparsing: This is incorrect. See [here](https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/ui/views/frame/browser_view.h;l=232?q=browser_view.h&ss=chromium).

browser/ui/views/frame/layout/brave_browser_view_tabbed_layout_impl.cc
simonhong: Hmm, I think this approach could make regressions in the future if we(or upstream) makes another child views. Then, we should adjust their `y`.

As this method only adjusts contents, infobar and background, I think still other views(such as ShadowOverlayView, ScrimView and etc..) also have to be adjusted?

And, we'll make upstream's `SidePanel` visible soon. Then, it also should be adjusted here?

Q) don't need to adjust SidebarContainerView's `y`?

browser/ui/views/frame/layout/brave_browser_view_tabbed_layout_impl.cc
simonhong: Ok, having a comment about focus mode here could be helpful for understand.

browser/ui/views/frame/brave_browser_frame_view_mac.mm
simonhong: 👍🏼

browser/ui/views/frame/layout/brave_browser_view_tabbed_layout_impl.cc
simonhong: I think this method seems not extensible with the reason written above.
but not sure how to handle nicely.
You also think that we should update this method whenever we found new views?

zenparsing: It turns out `AdjustLayoutForFocusMode` is not necessary. The upstream layout logic already takes care of the case where the horizontal tabstrip and the top container are not direct children of the browser view. I've removed the method.

browser/ui/views/frame/layout/brave_browser_view_tabbed_layout_impl.cc
zenparsing: Refactored slightly with a comment.

simonhong (approved):
++

netzenbot (approved):

thypon (approved):

browser/ui/views/frame/layout/brave_browser_view_tabbed_layout_impl.cc
```diff
@@ -58,6 +58,17 @@ void BraveBrowserViewTabbedLayoutImpl::ConfigureTopContainerBackground(
     background->SetCorners(corners);
   }
 #endif  // BUILDFLAG(IS_LINUX)
+
+  // In focus mode, the horizontal tab strip is reparented into the top
```
simonhong, line 62: We're inferring focus mode by checking each view's parent-child relationship.
I think it could be fragile. If we use `IsFocusModeEnabled()` instead, we don't need to have complicated parent-child relationship. WDYT?

zenparsing: Follow-up created: https://github.com/brave/brave-browser/issues/56806

PR #36815, opened 2026-05-28
[cr149][Android][Sync] Migrate sync_everything option to per-type on Android

### LINKED ISSUE #55862: [CR149] Toggles cannot be disabled after upgrade
### Description

Enabling `Sync Everything` and then upgrading will result in the new toggles being fixed in the enabled position. 

### Steps to reproduce

1. Install 1.92.91 (CR148)
2. Start a sync chain, enable `Sync everything` toggle
3. Upgrade to 1.92.95 (CR149)
4. Open sync settings, check sync toggles

### Actual result

Not able to disable the sync toggles

### Expected result

Should be able to toggle the sync categories

### Reproduces how often

Easily reproduced

### Brave version

1.92.95

### Device

- Brand/model: Google Pixel 9a
- Android version: 16

### Channel information

- [ ] release (stable)
- [ ] beta
- [x] nightly

### Reproducibility

### Miscellaneous information

_No response_

AlexeyBarabash:
Starting from `cr149` there is no `Sync Everything` UI option on Android.
Android apps which had `Sync Everything` and were updated from cr148 to cr149 cannot modify any types, because of failure `DUMP_WILL_BE_CHECK(!IsSyncEverythingEnabled());` at `SyncUserSettingsImpl::SetSelectedType`.

So this PR migrates sync_everything option to per-type on Android.

Resolves https://github.com/brave/brave-browser/issues/55862

### DIFF
--- a/components/sync/service/brave_sync_service_impl.cc
+++ b/components/sync/service/brave_sync_service_impl.cc
@@ -363,11 +363,22 @@ void BraveSyncServiceImpl::OnEngineInitialized(
     VLOG(1) << "Forced set decryption passphrase result is "
             << set_passphrase_result;
   }
+
+#if BUILDFLAG(IS_ANDROID)
+  DCHECK(GetSyncAccountStateForPrefs() !=
+         SyncPrefs::SyncAccountState::kNotSignedIn);
+  MigrateAndroidSyncEverythingIfRequired();
+#endif
 }
 
 SyncServiceCrypto* BraveSyncServiceImpl::GetCryptoForTesting() {
   CHECK_IS_TEST();
   return &crypto_;
+}
+
+DataTypeManager* BraveSyncServiceImpl::GetDataTypeManagerForTesting() {
+  CHECK_IS_TEST();
+  return data_type_manager_.get();
 }
 
 void BraveSyncServiceImpl::SetInitializingForTesting() {
@@ -561,5 +572,22 @@ void BraveSyncServiceImpl::OnAccountsInCookieUpdated(
 
 void BraveSyncServiceImpl::OnPrimaryAccountChanged(
     const signin::PrimaryAccountChangeEvent& event_details) {}
+
+#if BUILDFLAG(IS_ANDROID)
+void BraveSyncServiceImpl::MigrateAndroidSyncEverythingIfRequired() {
+  if (!GetUserSettings()->IsSyncEverythingEnabled()) {
+    return;
+  }
+
+  syncer::UserSelectableTypeSet selected_types;
+  UserSelectableTypeSet registered_types =
+      GetUserSettings()->GetRegisteredSelectableTypes();
+  for (UserSelectableType type : registered_types) {
+    selected_types.Put(type);
+  }
+  GetUserSettings()->SetSelectedTypes(false, selected_types);
+  CHECK(!GetUserSettings()->IsSyncEverythingEnabled());
+}
+#endif
 
 }  // namespace syncer
--- a/components/sync/service/brave_sync_service_impl.cc
+++ b/components/sync/service/brave_sync_service_impl.cc
@@ -584,7 +584,6 @@ void BraveSyncServiceImpl::MaybeAndroidSyncEverythingIfRequired() {
     selected_types.Put(type);
   }
   GetUserSettings()->SetSelectedTypes(false, selected_types);
-  DCHECK(!GetUserSettings()->IsSyncEverythingEnabled());
 }
 #endif
 
--- a/components/sync/service/brave_sync_service_impl.cc
+++ b/components/sync/service/brave_sync_service_impl.cc
@@ -365,9 +365,7 @@ void BraveSyncServiceImpl::OnEngineInitialized(
   }
 
 #if BUILDFLAG(IS_ANDROID)
-  DCHECK(GetSyncAccountStateForPrefs() !=
-         SyncPrefs::SyncAccountState::kNotSignedIn);
-  MigrateAndroidSyncEverythingIfRequired();
+  MaybeAndroidSyncEverythingIfRequired();
 #endif
 }
 
@@ -574,7 +572,7 @@ void BraveSyncServiceImpl::OnPrimaryAccountChanged(
     const signin::PrimaryAccountChangeEvent& event_details) {}
 
 #if BUILDFLAG(IS_ANDROID)
-void BraveSyncServiceImpl::MigrateAndroidSyncEverythingIfRequired() {
+void BraveSyncServiceImpl::MaybeAndroidSyncEverythingIfRequired() {
   if (!GetUserSettings()->IsSyncEverythingEnabled()) {
     return;
   }
@@ -586,7 +584,7 @@ void BraveSyncServiceImpl::MigrateAndroidSyncEverythingIfRequired() {
     selected_types.Put(type);
   }
   GetUserSettings()->SetSelectedTypes(false, selected_types);
-  CHECK(!GetUserSettings()->IsSyncEverythingEnabled());
+  DCHECK(!GetUserSettings()->IsSyncEverythingEnabled());
 }
 #endif
 
--- a/components/sync/service/brave_sync_service_impl.h
+++ b/components/sync/service/brave_sync_service_impl.h
@@ -91,6 +91,7 @@ class BraveSyncServiceImpl : public SyncServiceImpl {
   SyncServiceCrypto* GetCryptoForTesting();
   bool SetSeedForTesting(const std::string& seed);
   void SetInitializingForTesting();
+  DataTypeManager* GetDataTypeManagerForTesting();
 
  private:
   friend class BraveSyncServiceImplGACookiesTest;
@@ -148,6 +149,10 @@ class BraveSyncServiceImpl : public SyncServiceImpl {
       const signin::PrimaryAccountChangeEvent& event_details) override;
 
   bool SetSeed(const std::string& seed);
+
+#if BUILDFLAG(IS_ANDROID)
+  void MigrateAndroidSyncEverythingIfRequired();
+#endif
 
   brave_sync::Prefs brave_sync_prefs_;
 
--- a/components/sync/service/brave_sync_service_impl.h
+++ b/components/sync/service/brave_sync_service_impl.h
@@ -90,8 +90,8 @@ class BraveSyncServiceImpl : public SyncServiceImpl {
   void RemoveAllPrefsChangeRegistrarForTesting();
   SyncServiceCrypto* GetCryptoForTesting();
   bool SetSeedForTesting(const std::string& seed);
-  void SetInitializingForTesting();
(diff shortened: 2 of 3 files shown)


components/sync/service/brave_sync_service_impl.cc
```diff
@@ -363,13 +363,24 @@ void BraveSyncServiceImpl::OnEngineInitialized(
     VLOG(1) << "Forced set decryption passphrase result is "
             << set_passphrase_result;
   }
+
+#if BUILDFLAG(IS_ANDROID)
+  DCHECK(GetSyncAccountStateForPrefs() !=
+         SyncPrefs::SyncAccountState::kNotSignedIn);
```
darkdh: I don't see why we need this DCHECK

components/sync/service/brave_sync_service_impl.cc
```diff
@@ -562,4 +573,21 @@ void BraveSyncServiceImpl::OnAccountsInCookieUpdated(
...
+  }
+
+  syncer::UserSelectableTypeSet selected_types;
+  UserSelectableTypeSet registered_types =
+      GetUserSettings()->GetRegisteredSelectableTypes();
+  for (UserSelectableType type : registered_types) {
+    selected_types.Put(type);
+  }
+  GetUserSettings()->SetSelectedTypes(false, selected_types);
+  CHECK(!GetUserSettings()->IsSyncEverythingEnabled());
```
darkdh: Users will experience crashes if there is a bug in `SetSelectedTypes` and this is more like test scenario not a production check.

components/sync/service/brave_sync_service_impl.cc
```diff
@@ -562,4 +573,21 @@ void BraveSyncServiceImpl::OnAccountsInCookieUpdated(
 void BraveSyncServiceImpl::OnPrimaryAccountChanged(
     const signin::PrimaryAccountChangeEvent& event_details) {}
 
+#if BUILDFLAG(IS_ANDROID)
+void BraveSyncServiceImpl::MigrateAndroidSyncEverythingIfRequired() {
+  if (!GetUserSettings()->IsSyncEverythingEnabled()) {
+    return;
+  }
```
darkdh: shouldn't we also remove the user facing toggle so migration won't keep running if user keep toggling on "sync everything"?

components/sync/service/brave_sync_service_impl.cc
```diff
@@ -363,13 +363,24 @@ void BraveSyncServiceImpl::OnEngineInitialized(
     VLOG(1) << "Forced set decryption passphrase result is "
             << set_passphrase_result;
   }
+
+#if BUILDFLAG(IS_ANDROID)
+  DCHECK(GetSyncAccountStateForPrefs() !=
+         SyncPrefs::SyncAccountState::kNotSignedIn);
+  MigrateAndroidSyncEverythingIfRequired();
+#endif
```
netzenbot, line 369: nit: `DCHECK` only fires in debug builds. If `GetSyncAccountStateForPrefs() == kNotSignedIn` here is truly impossible (i.e., `OnEngineInitialized` can only be reached after sign-in), use `CHECK` so the invariant is enforced in release builds too. See [CS-025](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-025).

components/sync/service/brave_sync_service_impl.cc
```diff
@@ -363,13 +363,24 @@ void BraveSyncServiceImpl::OnEngineInitialized(
...
 }
 
 SyncServiceCrypto* BraveSyncServiceImpl::GetCryptoForTesting() {
   CHECK_IS_TEST();
   return &crypto_;
 }
 
+DataTypeManager* BraveSyncServiceImpl::GetDataTypeManagerForTesting() {
```
netzenbot, line 377: nit: The `.h` declares `GetDataTypeManagerForTesting` after `SetInitializingForTesting`, but the `.cc` defines it before. Per [CS-017](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-017), definition order in `.cc` should match declaration order in `.h`.

components/sync/service/brave_sync_service_impl.h
```diff
@@ -149,6 +150,10 @@ class BraveSyncServiceImpl : public SyncServiceImpl {
 
   bool SetSeed(const std::string& seed);
```
netzenbot, line 152: nit: This function conditionally performs an action (returns early if sync-everything is already off). Per [CS-008](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-008), prefer the `Maybe*` prefix: `MaybeMigrateAndroidSyncEverything`.

components/sync/service/brave_sync_service_impl_unittest.cc
```diff
@@ -734,6 +737,55 @@ TEST_F(BraveSyncServiceImplTest, BookmarksAndPasswordsAfterSetup) {
...
+       MigrateAndroidSyncEverythingOnEngineInitialized) {
+  CreateSyncService();
+  EXPECT_FALSE(engine());
+
+  brave_sync_service_impl()->SetSyncCode(kValidSyncCode);
+  ASSERT_TRUE(base::test::RunUntil([&] { return engine() != nullptr; }));
+
+  // Complete setup so IsInitialSyncFeatureSetupComplete() returns true.
+  // OnEngineInitialized was already called above but exited early because
+  // setup was not yet complete at that point.
```
netzenbot: nit: Avoid positional references like "called above" — if code is inserted between the reference and the call, the comment becomes misleading. Reference the specific call by name instead, e.g., "OnEngineInitialized was already called via SetSyncCode/RunUntil". See [CS-024](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-024).

components/sync/service/brave_sync_service_impl.cc
AlexeyBarabash: thanks, updated at https://github.com/brave/brave-core/pull/36815/changes/483b0a149b10d4a3f37744917b8f0cb1561551b6#diff-7ac9a9920c208381771faef286897151f591db06d29cdeb584f14e372b22e029R94 .

components/sync/service/brave_sync_service_impl.cc
AlexeyBarabash: The reason why I did put it - there is `NOTREACHED` at 
```c++
void SyncUserSettingsImpl::SetSelectedTypes(bool sync_everything,
                                            UserSelectableTypeSet types) {
  DCHECK(registered_types.HasAll(types))
      << "\n registered: " << UserSelectableTypeSetToString(registered_types)
      << "\n setting to: " << UserSelectableTypeSetToString(types);

  switch (delegate_->GetSyncAccountStateForPrefs()) {
    case SyncPrefs::SyncAccountState::kNotSignedIn:
      NOTREACHED();
...
```

so it would crash in anyway later.

Taking to attention the comment https://github.com/brave/brave-core/pull/36815#discussion_r3319513041  about make it `CHECK` - I removed it completely at https://github.com/brave/brave-core/pull/36815/changes/483b0a149b10d4a3f37744917b8f0cb1561551b6#diff-9cc578b6bc0ff8c20ea911e73c25e9aa392ad37c3a6877bc71b45bbf611ca649L368 with explanation that I don't want to crash app on start if something will go wrong.

components/sync/service/brave_sync_service_impl.cc
AlexeyBarabash: I agree, thanks, `CHECK` is too much, modified it to `DCHECK` at https://github.com/brave/brave-core/pull/36815/changes/483b0a149b10d4a3f37744917b8f0cb1561551b6#diff-9cc578b6bc0ff8c20ea911e73c25e9aa392ad37c3a6877bc71b45bbf611ca649R587

components/sync/service/brave_sync_service_impl.cc
AlexeyBarabash: Yes, on Android `Sync Everything` toggle is already removed by upstream along with `void setSelectedTypes(boolean syncEverything, Set<Integer> enabledTypes);` at https://source.chromium.org/chromium/chromium/src/+/7fcfbcad76f7c4781f352b8154c73eaed30e3123  and completely `manage_sync_preferences.xml`  at https://source.chromium.org/chromium/chromium/src/+/008e21eee2ca1b4d967a68367121423f6e411e2b .  So user is not able to turn on `sync everything` on Android through UI.

components/sync/service/brave_sync_service_impl.cc
AlexeyBarabash: thank you, I have removed this DCHECK completely then, because in the case when by whatever reasons that `CHECK` will be fired at production, the browser will crash on start. I really don't want that. Removed it completely at https://github.com/brave/brave-core/pull/36815/changes/483b0a149b10d4a3f37744917b8f0cb1561551b6#diff-9cc578b6bc0ff8c20ea911e73c25e9aa392ad37c3a6877bc71b45bbf611ca649L368

components/sync/service/brave_sync_service_impl.h
AlexeyBarabash: thanks, updated at https://github.com/brave/brave-core/pull/36815/changes/483b0a149b10d4a3f37744917b8f0cb1561551b6#diff-7ac9a9920c208381771faef286897151f591db06d29cdeb584f14e372b22e029R154

components/sync/service/brave_sync_service_impl_unittest.cc
AlexeyBarabash: updated at https://github.com/brave/brave-core/pull/36815/changes/483b0a149b10d4a3f37744917b8f0cb1561551b6#diff-e3672a42c1162ea76a62470435075b71758cc12f242b05b73efc669f2af7023bR751

components/sync/service/brave_sync_service_impl.cc
darkdh: I'm confused, why does the issue reported seeing the toggle and turned it on?

components/sync/service/brave_sync_service_impl.cc
darkdh: There should be no DCHECK/CHECK for this condition

components/sync/service/brave_sync_service_impl.cc
AlexeyBarabash: Maybe this issue is not quite exact,  updated it with screenshots/video

> the new toggles being fixed 

this is about the other toggles then `Sync Everything`

https://github.com/brave/brave-browser/issues/55862#issuecomment-4567596441

components/sync/service/brave_sync_service_impl.cc
AlexeyBarabash: thanks, removed it away at https://github.com/brave/brave-core/pull/36815/changes/6b1b15049258da90df741274b5adb64eefc4e0bf

darkdh (approved):

PR #36817, opened 2026-05-28
[fingerprinting] Handle device info leaks for WebGPU 

### LINKED ISSUE #55677: Add fingerprinting protections for WebGPU and WebGL 2.0 for leaking device info
- [x] WebGPU exposes [GPUAdapterInfo](https://developer.mozilla.org/en-US/docs/Web/API/GPUAdapterInfo) which contains device specific information like vendor and architecture. 
See [Comment#1](https://github.com/brave/brave-browser/issues/55677#issuecomment-4500291829) for more details. 

- [x] WebGL2 relies on the default WEBGL_debug_renderer_info to get hold of vendor and renderer info which we have previously handled at https://github.com/brave/brave-browser/issues/55443.

r0hit0303:
This change scrubs the device related info from [GPUAdapterInfo](https://developer.mozilla.org/en-US/docs/Web/API/GPUAdapterInfo) namely `vendor`, `architecture` and `device`.  

This change adds a patch to the upstream [gpu_adapter.cc](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/webgpu/gpu_adapter.cc;l=125) to intercept right before the moment we create the gpu adapter info, to scrub out these attributes. It was not possible to do it in [gpu_adapter_info.cc](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/webgpu/gpu_adapter_info.cc;) as it lacked any context around `ExecutionContext` which was needed to get the farbling level. 

Below are the final behaviour with different farbling levels after this change:

#### Farbling "off"
- No change in behaviour i.e no scrubbing will occur. 

#### Farbling "balanced" with [WebGLBalancedFingerprintingProtection](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/chromium_src/third_party/blink/common/features.cc?L63) **enabled**
- Scrub the device related info from GPUAdapterInfo. 

#### Farbling "balanced" with [WebGLBalancedFingerprintingProtection](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/chromium_src/third_party/blink/common/features.cc?L63) **disabled**
-  No change in behaviour i.e no scrubbing will occur. 

#### Farbling "maximum": 
- Scrub the device related info from GPUAdapterInfo. **Note** that this is a new behaviour and not to tied to any feature flag. For maximum farbling it makes sense to have the maximum protection. 

### Demo

https://github.com/user-attachments/assets/536fa971-f1e0-4c4e-9b93-cebf93347441

Resolves https://github.com/brave/brave-browser/issues/55677

### DIFF
--- a/browser/farbling/brave_webgpu_farbling_browsertest.cc
+++ b/browser/farbling/brave_webgpu_farbling_browsertest.cc
@@ -55,8 +55,8 @@ class BraveWebGPUFarblingBrowserTest : public InProcessBrowserTest {
     host_resolver()->AddRule("*", "127.0.0.1");
     content::SetupCrossSiteRedirector(embedded_test_server());
 
-    base::FilePath test_data_dir;
-    base::PathService::CheckedGet(brave::DIR_TEST_DATA, &test_data_dir);
+    base::FilePath test_data_dir =
+        base::PathService::CheckedGet(brave::DIR_TEST_DATA);
     test_data_dir = test_data_dir.AppendASCII(kEmbeddedTestServerDirectory);
     embedded_test_server()->ServeFilesFromDirectory(test_data_dir);
 
--- a/browser/farbling/brave_webgpu_farbling_browsertest.cc
+++ b/browser/farbling/brave_webgpu_farbling_browsertest.cc
@@ -1,0 +1,205 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "base/path_service.h"
+#include "base/test/scoped_feature_list.h"
+#include "brave/components/brave_shields/core/browser/brave_shields_utils.h"
+#include "brave/components/brave_shields/core/common/features.h"
+#include "brave/components/constants/brave_paths.h"
+#include "brave/components/webcompat/core/common/features.h"
+#include "chrome/browser/content_settings/host_content_settings_map_factory.h"
+#include "chrome/browser/profiles/profile.h"
+#include "chrome/browser/ui/browser.h"
+#include "chrome/test/base/in_process_browser_test.h"
+#include "chrome/test/base/ui_test_utils.h"
+#include "components/content_settings/core/browser/host_content_settings_map.h"
+#include "content/public/test/browser_test.h"
+#include "content/public/test/browser_test_utils.h"
+#include "net/dns/mock_host_resolver.h"
+#include "net/test/embedded_test_server/embedded_test_server.h"
+#include "third_party/blink/public/common/features.h"
+#include "url/gurl.h"
+
+using brave_shields::ControlType;
+
+namespace {
+
+constexpr char kEmbeddedTestServerDirectory[] = "webgpu";
+
+// Script to await the Promise set by the test HTML pages.
+constexpr char kGPUAdapterInfoScript[] =
+    "(async () => await window.gpuAdapterInfoPromise)()";
+
+// Expected result when all three adapter info fields (vendor, architecture,
+// device) are farbled to empty strings.
+constexpr char kEmptyAdapterInfo[] = ",,";
+
+}  // namespace
+
+class BraveWebGPUFarblingBrowserTest : public InProcessBrowserTest {
+ public:
+  BraveWebGPUFarblingBrowserTest() {
+    scoped_feature_list_.InitWithFeatures(
+        {
+            brave_shields::features::kBraveShowStrictFingerprintingMode,
+            webcompat::features::kBraveWebcompatExceptionsService,
+        },
+        {});
+  }
+
+  void SetUpOnMainThread() override {
+    InProcessBrowserTest::SetUpOnMainThread();
+
+    host_resolver()->AddRule("*", "127.0.0.1");
+    content::SetupCrossSiteRedirector(embedded_test_server());
+
+    base::FilePath test_data_dir;
+    base::PathService::CheckedGet(brave::DIR_TEST_DATA, &test_data_dir);
+    test_data_dir = test_data_dir.AppendASCII(kEmbeddedTestServerDirectory);
+    embedded_test_server()->ServeFilesFromDirectory(test_data_dir);
+
+    ASSERT_TRUE(embedded_test_server()->Start());
+  }
+
+  HostContentSettingsMap* content_settings() {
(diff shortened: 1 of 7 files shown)


r0hit0303:
@goodov Could you ptal at this change as well? It's not yet ready for a full review as it's missing tests. I wanted to get your thoughts before I start writing tests. Thanks

chromium_src/third_party/blink/renderer/modules/webgpu/gpu_adapter.cc
```diff
@@ -0,0 +1,36 @@
...
+                   ? String()                                                  \
+                   : s;                                                        \
+      case BraveFarblingLevel::MAXIMUM:                                        \
+        return String();                                                       \
+    }                                                                          \
+    return s;                                                                  \
+  };                                                                           \
+  base::AutoReset<String> reset_vendor(&vendor_, farble(vendor_));             \
+  base::AutoReset<String> reset_arch(&architecture_, farble(architecture_));   \
+  base::AutoReset<String> reset_device(&device_, farble(device_));
```
goodov: please make all of this be a class that you'll instantiate in a patch in a single line. There's many benefits to this approach, including easier crash reporting analysis, running code analysis tools, etc.

the class can accept references to all vars from the outer scope it needs to do the job.

r0hit0303: I've made the change in the latest ps. Hopefully, that's how you had envisioned it too. Please let me know if that wasn't the case. 

Just on a side note, I couldn't rely on the `base::AutoReset` anymore since the scope [changed](http://source.chromium.org/chromium/chromium/src/+/main:base/auto_reset.h;l=16-19). It allowed to reset these values back to the original values inside the adapter, but opaque to `gpu_adapter_info`. I think it's ok to  just scrub them completely like we are doing now.

goodov: was the scope changed because of the class introduction? I think it should be possible to have `AutoReset` as class members and set all values in the constructor, i.e. have `BraveFarblingLevel level` as a class member, initialize it first, then call `String Farble(String)` method to construct AutoReset members:
```cpp
BraveScrubWebGpuAdapterInfo(
  ExecutionContext* context,
  String& vendor,
  String& architecture,
  String& device) : level_(CalcLevel()), reset_vendor_(&vendor, Farble(vendor)), ...
```

if that's not important or won't work (maybe I missed something), it's ok to keep it as is.

put the class into anonymous ns though.

browser/farbling/brave_webgpu_farbling_browsertest.cc
```diff
@@ -0,0 +1,205 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
```
r0hit0303, line 1: This test suite was generated by LLM. There were minor issues which I fixed but all looks good to me. 

Prompt: 

> Similar to @browser/farbling/brave_webgl_farbling_browsertest.cc write a test for brave_webgpu_farbling_browsertest. You would need to create an html file here similar to https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/test/data/webgl/getSupportedExtensions.html to check: 
> a) Navigator.gpu.requestAdapter via: 
> const adapter = await navigator.gpu.requestAdapter({
>   featureLevel: "compatibility",
> });
> adapter.info should have the vendor, architecture and the device set to '' for BraveFarblingLevel::BALANCED and  kWebGLBalancedFingerprintingProtection enabled. 
> 
> b) GPUDevice
> const device = await adapter.requestDevice();
> similar to above adapter.info should have the vendor, architecture and the device set to '' for BraveFarblingLevel::BALANCED and  kWebGLBalancedFingerprintingProtection enabled. 
> 
> Add tests for BraveFarblingLevel::BALANCED with  kWebGLBalancedFingerprintingProtection feature flag enabled and disabled. and for BraveFarblingLevel::OFF and BraveFarblingLevel::MAXIMUM. 
>

chromium_src/third_party/blink/renderer/modules/webgpu/gpu_adapter.cc
r0hit0303: > I think it should be possible to have AutoReset as class members

That's a nice idea. I think it should work since the `BraveScrubWebGpuAdapterInfo ` would be in the stack until the end of the method call keeping the member `AutoReset` variables alive. Will update in the next change. 

Slightly tangential, please also note https://github.com/brave/brave-browser/issues/55927. I've added a TODO in the latest ps. 

> put the class into anonymous ns though.
Done

r0hit0303: > AutoReset as class members

Done

chromium_src/third_party/blink/renderer/modules/webgpu/gpu_adapter.cc
```diff
@@ -0,0 +1,65 @@
...
+
+#include "brave/third_party/blink/renderer/core/farbling/brave_session_cache.h"
+#include "third_party/blink/public/common/features.h"
+#include "third_party/blink/renderer/platform/wtf/text/wtf_string.h"
+
+namespace blink {
+
+namespace {
+
+String ApplyFarbling(ExecutionContext* context, const String& s) {
```
goodov: my idea is to put this as class function that you can call and set `level_` class memeber. This member then should be used to initialize auto-reset fields without calling `brave::GetBraveFarblingLevelFor()` for each field.

r0hit0303: Done

chromium_src/third_party/blink/renderer/modules/webgpu/gpu_adapter.cc
```diff
@@ -11,43 +11,44 @@ namespace blink {
...
         reset_architecture_(&architecture,
-                            ApplyFarbling(context, architecture)),
-        reset_device_(&device, ApplyFarbling(context, device)) {}
+                            ApplyFarbling(farbling_level_, architecture)),
+        reset_device_(&device, ApplyFarbling(farbling_level_, device)) {}
 
   ~BraveScrubWebGpuAdapterInfo() = default;
 
+  String ApplyFarbling(const BraveFarblingLevel level, const String& s) {
+    switch (level) {
```
goodov:
```cpp
private:
  String ApplyFarbling(const String& s) const {
    switch (farbling_level_) {
```

r0hit0303: Done

goodov (approved):

r0hit0303:
Thanks @goodov for the review

PR #36819, opened 2026-05-28
[Android] Surface Play in-app messages for subscription events

### LINKED ISSUE #55735: [Android] Implement native in-app messaging to reduce involuntary churn
### Description
Android has a native in-app notification for when there's a payment issue or an outstanding opt-in price increase.

https://developer.android.com/google/play/billing/subscriptions#in-app-messaging

It deep links to play store app and the app shows the dialog on top of Brave. So basically we just need to call showInAppMessages from the billing client library we use today and it will do everything for us.

Likely a quick win to reduce payment issue `involuntary` churn.

Image

SergeyZhukovsky:
Calls `BillingClient.showInAppMessages` on app foreground so the Play Store can render its recovery dialog (declined payment, opt-in price change) on top of Brave. Gated on the user having an active VPN or Leo subscription, throttled to one call per real foreground session via an `ApplicationStatus` listener that resets only on `HAS_STOPPED_ACTIVITIES` / `HAS_DESTROYED_ACTIVITIES` — in-app navigation does not re-trigger.

Resolves: https://github.com/brave/brave-browser/issues/55735

I didn't find any way how to test that unfortunately, that's why I put `QA/No` label on the issue. I checked the flow via logs and the thing is that if we don't any real event(price change/card decline and so on) the dialog just isn't shown. I checked that we call the function and everything is fine. So we need a real event to check that to confirm it works.

### DIFF
--- a/android/java/org/chromium/chrome/browser/app/BraveActivity.java
+++ b/android/java/org/chromium/chrome/browser/app/BraveActivity.java
@@ -446,6 +446,12 @@ public abstract class BraveActivity extends ChromeActivity
                 mMiscAndroidMetrics.recordIntentUrl(resumeIntent.getData().toString());
             }
         }
+
+        Profile profile = mTabModelProfileSupplier.get();
+        if (profile != null) {
+            InAppPurchaseWrapper.getInstance()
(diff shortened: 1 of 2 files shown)


samartnik (approved):
++

PR #36826, opened 2026-05-29
Remove uses of `$tag` for social media filter options

### LINKED ISSUE #55165: Replace `$tag` filters with a toggleable `FiltersProvider` implementation
adblock-rust's `$tag` option for network filters is a bespoke Brave-specific extension that:
- isn't very useful because it only works on plain network filters
- reduces performance by ~doubling the number of queries done in the adblock engine
- adds additional metadata to each filter, bloating the engine's memory consumption
- isn't supported by any other adblockers or lists
- doesn't add much over just recompiling filter lists with/out the relevant filters

For these reasons, it'd be beneficial to reimplement Brave's `Social media blocking` feature options to no longer rely on tagged filters. Instead, they can be conditionally added or removed from the adblock engine using brave-core's existing `FiltersProvider` interface.

antonok-edm:
Resolves https://github.com/brave/brave-browser/issues/55165

closes https://github.com/brave/brave-core/pull/36870

see also: https://github.com/brave/adblock-lists/pull/2874, https://github.com/brave/adblock-resources/pull/318

The above are not yet merged, but I've confirmed that the changes work correctly when used with different hard-coded list UUIDs.

### Deployment plan

Once merged:

#### Old `brave-core` versions

- will keep using the original `$tag` filters
- new components are hidden and will never be enabled

#### New `brave-core` versions

- original `$tag` filters will always be inactive (nothing enables the tag anymore)
- the new components will be enabled as needed by the settings toggles

### Test plan

Confirm that the `Social media blocking` toggles from `brave://settings/shields` correctly block or allow embedded social media content across several websites.

### DIFF
--- a/components/brave_shields/content/browser/ad_block_service.cc
+++ b/components/brave_shields/content/browser/ad_block_service.cc
@@ -33,7 +33,6 @@
 #include "brave/components/brave_shields/core/browser/ad_block_filters_provider_manager.h"
 #include "brave/components/brave_shields/core/browser/ad_block_resource_provider.h"
 #include "brave/components/brave_shields/core/common/adblock/rs/src/lib.rs.h"
-#include "brave/components/brave_shields/core/common/brave_shield_constants.h"
 #include "brave/components/brave_shields/core/common/features.h"
 #include "brave/components/brave_shields/core/common/pref_names.h"
 #include "components/prefs/pref_change_registrar.h"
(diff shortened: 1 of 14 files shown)


cmazakas (approved):
Rust changes LGTM!

netzenbot (approved):

components/brave_shields/content/browser/ad_block_service.cc
```diff
@@ -404,19 +383,7 @@ void AdBlockService::RemoveObserver(Observer* observer) {
...
-  bool enabled = local_state_->GetBoolean(pref_name);
-  EnableTag(tag, enabled);
-}
-
-void AdBlockService::EnableTag(const std::string& tag, bool enabled) {
-  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
-  // Tags only need to be modified for the default engine.
-  engine_wrapper_.AsyncCall(&AdBlockEngineWrapper::EnableTag)
-      .WithArgs(tag, enabled);
+  component_service_manager_->ResetProviders();
```
boocmp, line 386: Do we really need to rebuild all providers?

antonok-edm: not necessarily... this is reusing the logic of the prior `OnAdBlockOnlyModePrefChanged` method for simplicity, it's not a hot path (only triggered by user gesture in the settings page), and reloading the adblock engine does require re-reading all of the providers anyways

boocmp (approved):

antonok-edm:
prerequisites merged, merging now

StephenHeaps:
If uplifted please include and/or uplift: https://github.com/brave/brave-core/pull/37117

PR #36829, opened 2026-05-29
[Android] Fix: Sync is impossible to set up without Google Play Services

### LINKED ISSUE #11477: [Android] Handle QR code for sync when Google Play Services aren't available
## Description 
Android's built-in QR code scanning depends on Google Play services. Right now, attempting to use the QR code scanning feature brings up an error message about Brave not running without Google Play services installed. Brave is a privacy browser, so it would be nice to have better handling for for devices that do not have Google Play services available.

One option is to hide the QR code scanning altogether. This is not ideal but better than throwing an error to the user.

Another, more preferable option is to delegate scanning to the `com.google.zxing.client.android.SCAN` intent. It appears to be widely used as per [this tutorial](https://tekeye.uk/android/examples/scan-barcode-from-android-app) and [examples in the wild](https://github.com/wewewe718/QrAndBarcodeScanner/blob/master/app/src/main/AndroidManifest.xml#L71).

dark-penguin:
Resolves brave/brave-browser#11477

Fix setting up Sync by QRcode without Google Play Services

Before: the error dialog was popping up repeatedly, making it impossible to set up Sync.
Now: only show the error dialog once, making it possible to set up Sync by code words.

Tested manually on my LineageOS 19 device.

### DIFF
--- a/android/java/org/chromium/chrome/browser/qrreader/QRCodeCameraManager.java
+++ b/android/java/org/chromium/chrome/browser/qrreader/QRCodeCameraManager.java
@@ -61,6 +61,7 @@ public class QRCodeCameraManager
     private @Nullable CameraSourcePreview mCameraSourcePreview;
     private @Nullable DisplayManager mDisplayManager;
     private int mLastRotation = INITIAL_ROTATION;
+    private boolean mErrorShown;
     private final Callback mCallback;
     private final HostProvider mHostProvider;
 
@@ -212,7 +213,11 @@ public class QRCodeCameraManager
                 if (mCallback.onPlayServicesUnavailable(dlg)) {
                     return false;
                 }
+                if (mErrorShown) {
+                    return false;
+                }
                 if (dlg != null) {
+                    mErrorShown = true;
                     dlg.show();
                 }
             }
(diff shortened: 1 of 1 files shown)


android/java/org/chromium/chrome/browser/qrreader/QRCodeCameraManager.java
```diff
@@ -61,6 +61,7 @@ public interface HostProvider {
     private @Nullable CameraSourcePreview mCameraSourcePreview;
     private @Nullable DisplayManager mDisplayManager;
     private int mLastRotation = INITIAL_ROTATION;
+    private boolean mErrorShown;
```
antonok-edm: from a maintainability standpoint, it'd be better to name this something more descriptive like `mGooglePlayServicesAvailabilityDialogShown`

dark-penguin: This looks kinda unholy to me, but you know better which is better (too long or too short). Maybe `mGooglePlayServicesUnavailableDialogShown` ?

antonok-edm: yeah, that's fine too. Just anything more descriptive

dark-penguin: Done, but I don't see a good example on commit message style; should I specify the variable name or keep it short? or will it be squashed on pull anyway?

antonok-edm: that's fine, it'll be squashed. Thanks! Someone from the @brave/android team should take a look soon

simoarpe:
Thanks for tracking this down, and for the manual test on a real de-Googled device. The symptom is definitely a bug and your diagnosis of where it fires is right.

I took a safer approach in https://github.com/brave/brave-core/pull/38292 that moves the Play Services check out of `startCameraSource()` rather than guarding the dialog, and also fixes a window leak in the error dialog.

Thank you again for bringing this up to our attention!
I'm closing this now.

PR #36833, opened 2026-05-29
[query-filter] Add logic to strip blocked query parameter for "=="

### LINKED ISSUE #55924: Remove  blocked query parameters even for values like "="
Currently for cross-site requests like "www.example.com?fbclid==", we don't remove fbclid even though it's marked as a blocked parameter in the [code](https://sourcegraph.com/r/github.com/brave/adblock-lists/-/blob/brave-lists/query-filter.json). `fbclid` is one of the many blocked parameters the same is true also for other parameters listed in the code. Please see this [comment](https://github.com/brave/brave-core/pull/36810#issuecomment-4573549256) for more context around this. 

We should remove the parameter even if the value was only "==".

r0hit0303:
This change removes the additional check we did to preserve the blocked query parameter in the URL if its value was of the form "="+ (i.e one of more "=").

Resolves https://github.com/brave/brave-browser/issues/55924

### DIFF
--- a/browser/net/brave_site_hacks_network_delegate_helper_unittest.cc
+++ b/browser/net/brave_site_hacks_network_delegate_helper_unittest.cc
@@ -212,7 +212,6 @@ TYPED_TEST(BraveSiteHacksNetworkDelegateHelperTest, QueryStringUntouched) {
        "https://example.com/?+fbclid=1",
        "https://example.com/?%20fbclid=1",
        "https://example.com/#fbclid=1",
-       "https://example.com/?fbclid==",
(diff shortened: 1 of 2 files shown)


goodov (approved):

r0hit0303:
Thanks @goodov for the review

netzenbot (approved):

fmarier (approved):

r0hit0303:
Thanks @fmarier for the review

MadhaviSeelam:
@r0hit0303 Can you please attach a testplan please?

PR #36834, opened 2026-05-29
[Sidebar V2] Update panel contents border radius

simonhong:
Applied rounded corners to panel's contents view.
Introduced brave::GetPanelContentsRoundedCorners(PrefService*, bool has_header)
which derives corners from the pref and current header presence.
Route GetRoundedCorners() through it via patch, reading header_view_ state
(set before AddChildView) instead of the stale coordinator entry ID.
Call UpdateContentWrapperChildCorners() from UpdateBorder() overrides so 
corners are re-applied on panel open and pref change.

TEST=SidePanelUtilsTest.*, SidebarBrowserTest.SidebarV2ContentCornersUpdateOnStateChange

No roundness w/o rounded corners:
image
image

Roundness with rounded corners:
image

image

### DIFF
--- a/chromium_src/chrome/browser/ui/views/side_panel/side_panel.cc
+++ b/chromium_src/chrome/browser/ui/views/side_panel/side_panel.cc
@@ -30,17 +30,23 @@
 namespace {
 
 // Applies the current rounded-corner values to every direct child of the
-// content wrapper. Used when the pref changes or the panel opens with
+// content parent view. Used when the pref changes or the panel opens with
 // existing content (so OnChildViewAdded never fired with the new values).
-void UpdateContentWrapperChildCorners(views::View* wrapper,
+void UpdateContentWrapperChildCorners(views::View* content_parent_view,
                                       PrefService* prefs,
                                       bool has_header) {
+  // ContentParentView hosts multiple content view and shows at once.
+  CHECK(content_parent_view->GetUseDefaultFillLayout());
+
   auto corners = brave::GetPanelContentsRoundedCorners(prefs, has_header);
-  for (views::View* child : wrapper->children()) {
+  for (views::View* child : content_parent_view->children()) {
+    // If the child is a WebView or paints to a layer, round its corners.
     if (views::IsViewClass<views::WebView>(child)) {
       views::AsViewClass<views::WebView>(child)->holder()->SetCornerRadii(
           corners);
     }
+    // Try to detect if the child is a views::View wrapper of a WebView. If so,
+    // round its corners.
     if (child->children().size() == 1 &&
         views::IsViewClass<views::WebView>(child->children()[0])) {
       views::AsViewClass<views::WebView>(child->children()[0])
--- a/chromium_src/chrome/browser/ui/views/side_panel/side_panel.cc
+++ b/chromium_src/chrome/browser/ui/views/side_panel/side_panel.cc
@@ -7,12 +7,14 @@
 #include "build/buildflag.h"
 
 #if BUILDFLAG(ENABLE_SIDEBAR_V2)
+#include "brave/browser/ui/views/side_panel/side_panel_utils.h"
 #include "chrome/browser/ui/color/chrome_color_id.h"
 #include "chrome/browser/ui/layout_constants.h"
 #include "chrome/browser/ui/views/side_panel/side_panel.h"
+#include "ui/compositor/layer.h"
 
-// Rename the upstream Add/RemoveHeaderView implementation so we can provide
-// a thin wrapper that reapplies border state.
+// Rename upstream methods so we can provide thin wrappers that reapply
+// rounded-corner and border state.
 #define AddHeaderView AddHeaderView_ChromiumImpl
 #define RemoveHeaderView RemoveHeaderView_ChromiumImpl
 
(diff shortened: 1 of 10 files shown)


rewrite/chrome/browser/ui/views/side_panel/side_panel.cc.yaml
```diff
@@ -0,0 +1,23 @@
+# Copyright (c) 2026 The Brave Authors. All rights reserved.
```
sangwoo108, line 1: Totally optionl:

I defer the proper review to plaster reviewers, but I'm slightly concerned with that the patch leaves dead code.

We can have a plaster that renames `GetRoundedCorners_ChromiumImpl() and puts declaration of `GetRoundedCorners()`. And with that, we can make a definition of ContentsParentView::GetRoundedCorners() in the chromium_src override file.

But this is mere my thought, hope the proper reviewers would chime in.

chromium_src/chrome/browser/ui/views/side_panel/side_panel.cc
```diff
@@ -25,6 +27,35 @@
 
 #if BUILDFLAG(ENABLE_SIDEBAR_V2)
 
+namespace {
+
+// Applies the current rounded-corner values to every direct child of the
+// content wrapper. Used when the pref changes or the panel opens with
+// existing content (so OnChildViewAdded never fired with the new values).
```
sangwoo108, line 34: I think this function assumes that the `wrapper` is using `FillLayout` as it iterates all children. If so, We might want to put assertion with `GetUseDefaultFillLayout` or comment about it.

chromium_src/chrome/browser/ui/views/side_panel/side_panel.cc
```diff
@@ -25,6 +27,35 @@
...
+void UpdateContentWrapperChildCorners(views::View* wrapper,
+                                      PrefService* prefs,
+                                      bool has_header) {
+  auto corners = brave::GetPanelContentsRoundedCorners(prefs, has_header);
+  for (views::View* child : wrapper->children()) {
+    if (views::IsViewClass<views::WebView>(child)) {
+      views::AsViewClass<views::WebView>(child)->holder()->SetCornerRadii(
+          corners);
+    }
+    if (child->children().size() == 1 &&
```
sangwoo108, line 50: Can we make this `else if`?

chromium_src/chrome/browser/ui/views/side_panel/side_panel.cc
```diff
@@ -25,6 +27,35 @@
 
 #if BUILDFLAG(ENABLE_SIDEBAR_V2)
 
+namespace {
+
+// Applies the current rounded-corner values to every direct child of the
```
sangwoo108, line 32: Do we need more comments regarding web view in the direct children's hierarchy?

sangwoo108 (approved):
LGTM!

netzenbot (approved):

chromium_src/chrome/browser/ui/views/side_panel/side_panel.cc
simonhong: Added.

chromium_src/chrome/browser/ui/views/side_panel/side_panel.cc
simonhong: Updated comments below.

chromium_src/chrome/browser/ui/views/side_panel/side_panel.cc
simonhong: I think we can use `else if` here. but better to have same logic with upstream. - `ContentParentView::OnChildViewAdded()`.

rewrite/chrome/browser/ui/views/side_panel/side_panel.cc.yaml
simonhong: I think  it's better to put the code directly if it's patching.
It made upstream code dead code but new `_ChromiumImpl()` also becomes unused code.

rewrite/chrome/browser/ui/views/side_panel/side_panel.cc.yaml
```diff
@@ -0,0 +1,23 @@
...
+      Route ContentParentView::GetRoundedCorners() through Brave's corner helper.
+
+      Calls brave::GetPanelContentsRoundedCorners() with the current prefs and
+      whether the side panel has a Brave header attached.  Reading header_view_
+      state (set before AddChildView in PopulateSidePanel) avoids the timing
+      issue where coordinator->GetCurrentEntryId() still returns the previous
+      entry when OnChildViewAdded fires.
+    re_pattern:
+      '(  gfx::RoundedCornersF GetRoundedCorners\(\)
+      \{)\n    ChromeDistanceMetric corner_radius ='
```
cdesouza-chromium: Just match to `GetRoundedCorners` declaration and make the insertion the first line of the function. In this case you are matching the next line too, `\n ChromeDistanceMetric corner_radius =`, so this will break if they change anything in the next line, and in this case, it doesn't look like you care about the contents for the next line.

simonhong: Thanks. updated.

cdesouza-chromium (approved):
LGTM

PR #36835, opened 2026-05-29
Add a workaround to fix build hangs because of upstream tsc/node issues.

### LINKED ISSUE #54004: Build sometimes hangs on `ts_library` actions
We see that linux builds randomly hang without obvious reason.

after analyzing siso logs I see that these targets never finish:

```
ACTION //chrome/browser/resources/segmentation_internals:build_ts(//build/toolchain/mac:clang_x64)
```
in https://teamcity.brave.com/buildConfiguration/BraveBrowser_Component_Build/167539?buildTab=log&focusLine=225819&logView=linear&linesState=225819

and:
```
ACTION //third_party/devtools-frontend/src/front_end/entrypoints/inspector_main:devtools_entrypoint-meta-tsconfig(//build/toolchain/mac:clang_x64)
```
in https://teamcity.brave.com/buildConfiguration/BraveBrowser_Component_Build/169956?buildTab=log&linesState=223968.223996&logView=flowAware&focusLine=503089

I don't see anything interesting in siso logs.

These targets are `ts_library` targets, but they use different scripts:
`tools/typescript/ts_library.py`
`third_party/devtools-frontend/src/scripts/build/typescript/ts_library.py`

both have our patches, but they solve different problems, meaning there's no similar logic that I can see as the culprit.

goodov:
Add a workaround to restart upstream `tsc` calls on hang. Timeout is set to 90 seconds which increase x2 two times. In my tests existing `tsc` calls complete in less than 40 seconds, so 90 should be plenty to catch the actual hang.

This is a reoccurring issue that also exists in Chromium infra (see link in the issue). It becomes a bit annoying as builds randomly timeout (both PR and release ones). This is an attempt to workaround the problem.

Resolves https://github.com/brave/brave-browser/issues/54004

### DIFF
--- a/build/config/siso/brave_siso_config.star
+++ b/build/config/siso/brave_siso_config.star
@@ -341,6 +341,7 @@ def __wrap_python_with_chromium_src_inputs_handler(ctx, rule, handlers):
         handlers[new_handler_name] = __create_handler([
             "brave/script/brave_chromium_utils.py",
             "brave/script/override_utils.py",
+            "brave/tools/typescript/tsc_timeout_retry.py",
         ])
 
     rule["handler"] = new_handler_name
--- a/chromium_src/third_party/devtools-frontend/src/scripts/build/typescript/ts_library.py
+++ b/chromium_src/third_party/devtools-frontend/src/scripts/build/typescript/ts_library.py
@@ -10,6 +10,15 @@ import re
 import brave_chromium_utils
 import override_utils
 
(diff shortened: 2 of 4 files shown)


cdesouza-chromium (approved):
LGTM

mkarolin (approved):
++

fallaciousreasoning (commented):
lgtm

PR #36838, opened 2026-05-29
Unify YouTube Picture-in-Picture flow into a dedicated Java controller

### LINKED ISSUE #51859: Toolbar disappears after returning from PiP
### Description

Toolbar disappears after returning from youtube PiP view 

### Steps to reproduce

1. Play youtube media
2. Switch to PiP view
3. return to normal page
4. repeat if necessary

### Actual result

bottom toolbar disappears

https://github.com/user-attachments/assets/dc5e772e-a540-4110-8620-508aeb3e9846

### Expected result

Toolbar should be shown

### Reproduces how often

Easily reproduced

### Brave version

1.85.117

### Device

- Brand/model: Xiaomi Poco F3
- Android version: 

### Channel information

- [x] release (stable)
- [ ] beta
- [ ] nightly

### Reproducibility

### Miscellaneous information

_No response_

### LINKED ISSUE #52579: Occasionally, when the device is locked while `PiP` mode is active, both the background playback stops and the `PiP` window closes
### Description

Occasionally, when the device is locked while `PiP` mode is active, both the background playback stops and the `PiP` window closes.

### Steps to reproduce

1. Launch Brave
2. Open YT in the new tab
3. Play any video
4. Tap `PiP` button in the URL search bar
5. When `PiP` mode is active lock device > *Observe that either background play is paused or background play is paused + `PiP` is closed when the device is unlocked*

### Actual result

Occasionally, when the device is locked while `PiP` mode is active, both the background playback stops and the `PiP` window closes.

--------------------

| Galaxy A32 5G / Timestamp ~0:50 | Moto g Power 5G - 2023 Timestamp ~0:40 |
|:-:|:-:|
| **SCREENCAST** | **SCREENCAST** |
| [](https://github.com/user-attachments/assets/d33c5238-c31c-4da1-a4e7-a264efd21521) | [](https://github.com/user-attachments/assets/9fc3b36d-9a94-4d78-a167-64b60fb50727) |

### Expected result

Background play and `PiP` must continue working when locking the device.

### Reproduces how often

Easily reproduced

### Brave version

Brave build: 1.86.148
Chromium: 144.0.7559.133 (Official Build) (64-bit)

### Device

- Galaxy A32 5G / a32xsq-user 13 TP1A.220624.014 release-keys
- Moto g Power 5G - 2023 / devonn_g_sys-user 14 U1TOS34.1-157-5-21 release-keys

### Channel information

- [x] release (stable)
- [ ] beta
- [ ] nightly

### Reproducibility

### Miscellaneous information

_No response_

### LINKED ISSUE #53760: Returning to Brave from `PiP` mode is broken, and YT tab needs to be closed & reopened to fix the issue
### Description

Returning to Brave from `PiP` mode is not working correctly, and the YT tab needs to be closed & reopened in order to resolve the issue:
- YT page UI is not shown;
- No way to exit full screen;
- Video is paused when tapping `PiP` button again;

### Steps to reproduce

1. Launch Brave
2. In the new tab open YT
3. Play any video > Tap `PiP` button in the URL search bar
4. After the video starts playing in the `PiP` mode, tap the button to return to Brave
5. *Observe that Brave UI is broken*
6. Refresh the page
7. Tap the `PiP` button again > *Observe that the video is paused instead*

### Actual result

Returning to Brave from `PiP` mode is not working correctly, and the YT tab needs to be closed & reopened in order to resolve the issue.

-------------------

| Pixel 7 / 17 Beta | SM Fold 6 / 16 |
|:-:|:-:|
| <video src="https://github.com/user-attachments/assets/c64cdadc-c700-4689-bc97-ec87d2e0c70c"></video> | <video src="https://github.com/user-attachments/assets/6f54afcf-eb5b-4ce9-a9b9-74705767fe27"></video> |

### Expected result

Returning to Brave from `PiP` mode must work correctly as before.

### Reproduces how often

Easily reproduced

### Brave version

Brave build: 1.88.133 
Chromium: 146.0.7680.119 (Official Build) (64-bit) 

### Device

- Galaxy Z Fold 6 / q6quew-user 16 BP2A.250605.031.A3 release-keys
- Pixel 7 / panther_beta-user 17 CP21.260206.011.A1 release-keys

### Channel information

- [x] release (stable)
- [x] beta
- [x] nightly

### Reproducibility

### Miscellaneous information

_No response_

### LINKED ISSUE #55163: Pip/full screen broken
### Description

I'm in Samsung s25 Android 16

For a little while now full screen has been quite broken. Or maybe it's the PIP feature?

When you go to full screen and then you use the pip feature in any way, it basically never recovers when you go back to the video. There are no more controls, it shows in the middle of the screen but small, cuts of all the UI elements.

I've attached a screen video recording. 

The only way to recover is to close down the app and then restart. 

I can't seem to upload a video showing you what's going on

### Steps to reproduce

1. Play video in full screen
2. Attempt to do anything which brings it to pip
3. Try to go back to video in full screen

### Expected result

It should work

### Reproduces how often

Easily reproduced

### Brave version

Every time 

### Device

- Brand/model: Samsung Galaxy s25
- Android version: 16

### Channel information

- [x] release (stable)
- [ ] beta
- [ ] nightly

### Reproducibility

### Miscellaneous information

_No response_

simoarpe:
## Summary

Refactor and harden the Brave managed YouTube Picture-in-Picture flow on Android. The session lifecycle that previously lived inline in `BraveActivity` is extracted into a dedicated controller, the manual ("YT PiP toolbar button") and automatic ("upstream fullscreen-video → auto-PiP") entry paths are unified behind it, and a number of synchronization issues and edge case bugs are fixed along the way.

### Architectural Changes
- New `BraveYouTubePictureInPictureController` owns:
  - Session state (active / exiting / interrupted / sessionId / WebContents / owning tab id).
  - Save/restore across activity recreation via the saved-instance bundle, **validated against
    ground truth on restore**: a persisted session is only adopted when the PiP window survived
    the recreation or the session is parked behind a screen-lock interruption, so a stale
    bundle (process death, PiP dismissed in the recreation gap) can never latch a dead session
    active.
  - A **session-lifetime** screen-state `BroadcastReceiver` for `ACTION_SCREEN_OFF` /
    `ACTION_SCREEN_ON` / `ACTION_USER_PRESENT`. SCREEN_OFF is the pause trigger: it is the only
    reliable lock signal when the PiP window survives the lock (PiP mode doesn't change and
    still-playing media emits no upstream signal). Registration is deferred to
    `onResumeWithNative` because `onPostCreate` runs pre-native.
  - Play/pause state tracking via a `MediaSessionObserver` (the Java MediaSession API has no
    state getter and callbacks are async, so the state is tracked continuously while the
    session is bound).
  - Synchronous teardown for the new-tab path.
- `BraveActivity` becomes a thin set of delegating shims onto the controller.
- The manual-PiP path (`BraveYouTubeScriptInjectorNativeHelper.enterPictureInPicture`) and the
  upstream auto-PiP path (`FullscreenVideoPictureInPictureController`) both register the session
  through `onYouTubePictureInPictureRequested`, so both flows share the same exit/resume code.
  Upstream's entry is delegated via a Plaster-injected guarded early return into
  `braveAttemptPictureInPicture()` (always returns `true`); upstream's
  `enterPictureInPictureMode` try/catch is left in the source, just unreachable.
- **Fullscreen exit uses no JS injection.** The session exit asks the renderer to drop DOM
  fullscreen via `WebContents#exitFullscreen()` (the same path as the Back button, which
  YouTube handles natively), restores the browser UI off the renderer's
  `didToggleFullscreenModeForTab` signal with a 500 ms fallback, and tears down via the stock
  `FullscreenManager#exitPersistentFullscreenMode()`. The fallback logs a warning as a field
  detector for the (never observed) renderer-state desync that a page-side exit script would
  be needed for. The JS *entry* script is unchanged.
- The `FullscreenHtmlApiHandlerBase` patch is reduced to two Plaster early-return hooks that
  preserve browser + DOM fullscreen across transient hides (tab hidden, activity STOPPED)
  while a Brave-managed session is active.
- Expanding the PiP lands on the browsing layout: `BraveActivity` exits overview mode if the
  activity was recreated into the tab switcher while the PiP window was up (local
  implementation; deliberately not coupled to the actor/Glic helper API).
- Background-video-playback is checked at the chrome layer (`ChromeFeatureList` + `UserPrefs`)
  in the controller; the components-layer `BraveMediaSessionHelper` keeps the command-line
  switch, which it cannot avoid for layering reasons.
- `FullscreenHtmlApiHandlerBase.java.patch` and
  `FullscreenVideoPictureInPictureController.java.patch` are converted to Plaster files.

### Bug fixes
- **Pause on screen lock (background playback disabled)**: locking/turning off the screen now
  pauses the PiP video, driven directly by the SCREEN_OFF broadcast. After unlock the video
  stays paused — matching how YouTube behaves outside PiP — and the user resumes from the PiP
  controls. With background playback enabled, audio keeps playing through the lock as before.
- **No forced playback after unlock**: the playing state is snapshotted when the lock hits
  (before our own pause), so PiP re-entry after unlock only resumes media that the lock
  actually interrupted — never a video the user had paused — and only when background playback
  is enabled. The re-entry resume also targets the session's own WebContents instead of the
  foreground tab, which can differ after a tab switch or recreation while locked.
- **Launcher shortcut "New tab" during PiP**: intercept upstream's `MetricsEndReason.NEW_TAB`
  so it does not call `moveTaskToBack(true)`, and force the toolbar back into view via
  `BrowserControlsManager.showAndroidControls(false)` after the canonical
  `FullscreenManager.onExitFullscreen(tab)`. The latter alone leaves the toolbar at negative
  height with `View.INVISIBLE` on tablets, where upstream's `onConstraintsChanged` relayout
  doesn't always recover.
- **Media-session pause suppression**: scoped to the WebContents the controller has registered
  via a static `WeakReference` slot in `BraveMediaSessionHelper`, instead of suppressing pause
  for any YouTube tab whose hosting Activity happens to be in PiP. Eliminates phantom SystemUI
  controls for background YouTube tabs. Registry clears are guarded to a non-null WebContents
  so a session-less window being destroyed (multi-window) can never wipe another window's live
  entry.
- **Stale session restore**: a restored `mActive` with no surviving PiP window and no lock
  interruption is discarded; previously it could persist for the activity lifetime and
  suppress upstream's fullscreen teardown browser-wide.
- **Startup crash guards**: the screen-state receiver is never armed pre-native, and tab-model
  access goes through the nullable supplier (`getTabModelSelector()` throws before initialization, and the receiver path is reachable during startup).
- **Screen-lock recovery**: media-session resume is deferred until Android confirms PiP entry;
  interrupt handling is idempotent across the multiple signals upstream delivers per lock; and
  session state is reset cleanly in `onDestroy()` so in-flight `PostTask` callbacks no-op via
  `isExitingForSession()`.
- **`shouldKeepPictureInPictureAlive`**: gated up front on `BraveActivity` and
  `isYouTubePictureInPictureActive()` so non-Brave activities and inactive sessions are not
  accidentally swallowed by the dismiss hook.

### Tests
- New `BraveYouTubePictureInPictureControllerTest` (25 Robolectric tests) covering: session
  state transitions; bundle save/restore including the stale-session discard,
  restore-while-in-PiP, and the interrupted-restore receiver arming on resume; the SCREEN_OFF
  broadcast pause wiring; the lock/unlock matrix for both background-playback settings,
  including the user-paused-before-lock guard; the registry non-null guards (destroy,
  clearSession, session-less multi-window destroy); resume targeting the session WebContents;
  the event-driven fullscreen exit fast path and its timeout fallback; the new tab cleanup
  path; and the finishing-activity guard. JNI is mocked via
  `BraveYouTubeScriptInjectorNativeHelperJni.setInstanceForTesting`, with the remaining
  JNI-backed calls isolated behind `@VisibleForTesting` seams stubbed on Mockito spies.
- New `BraveFullscreenVideoPictureInPictureControllerTest` covering each `MetricsEndReason`
  short-circuit (`START`, `RESUME`, `LEFT_FULLSCREEN`, `WEB_CONTENTS_LEFT_FULLSCREEN`,
  `NEW_TAB`).
- New `BraveFullscreenHtmlApiHandlerBaseTest` covering the persistent fullscreen preservation
  predicate (non-Brave activity, inactive and active sessions).
- New `BraveMediaSessionHelperTest` covering the Brave Talk, YouTube, and YouTube-PiP
  WebContents predicates, including the untracked-WebContents case.

## Test plan (Check both mobile and tablet - no desktop mode)

1. **Manual YouTube PiP**: open `m.youtube.com/watch?v=…`, tap the YouTube PiP icon in the
   toolbar. The video should enter PiP. Tap the PiP window to expand: the tab should return to
   its normal page layout with the browser toolbar visible (not stuck in fullscreen styles).
2. **Auto-PiP from fullscreen**: enter video fullscreen via YouTube's own button, then send the
   app to the home screen. The video should auto-PiP. Dismissing the PiP with the X should stop
   the audio within about a second.
3. **Launcher "New tab" shortcut while in PiP**: long-press the Brave icon, tap "New tab". The
   new tab should appear in the foreground with the browser toolbar visible, and playback should
   pause. Verify on **both phone and tablet form factors**.
4. **Switch back to original tab after step 3**: the YouTube video should be in its normal
   in-page layout (not stuck in fullscreen styles).
5. **Screen lock during PiP with background video playback ENABLED**: start a YT-PiP session, lock
   the device, wait 10 s, unlock. The PiP window should survive and audio should keep playing
   throughout.
6. **Screen lock during PiP with background video playback DISABLED**: start a YT-PiP session,
   lock (or just turn off) the screen, audio should pause. Unlock: the PiP window should be
   back with the video paused; tapping play in the PiP should resume it. Lock again, it should
   pause again (repeat the cycle at least twice).
7. **Manual pause before lock**: with background playback enabled, pause the video in the PiP,
   lock, then unlock. The video must stay paused and it must not auto-resume.
8. **Recreation during PiP**: with the PiP window up, change a configuration that recreates the
   activity (e.g. rotate with auto-rotate, or toggle dark theme), then expand the PiP. The
   session should survive, and the expand should land on the browsing layout (not the tab
   switcher) in the normal page layout.
9. **Android 10 (API 29) emulator**: the toolbar PiP icon should be hidden; no JS injection or
   PiP entry should be attempted.

## Notes for reviewers

- The Plaster files `FullscreenHtmlApiHandlerBase.java.yaml` and `FullscreenVideoPictureInPictureController.java.yaml` have been converted from TOML.
- The `BraveMediaSessionHelper` registry uses a single static `WeakReference`, which is sufficient because Android only allows at most one PiP window per device at a time.


android/brave_java_sources.gni
```diff
@@ -241,6 +241,7 @@ brave_java_sources = [
   "../../brave/android/java/org/chromium/chrome/browser/local_database/TopSiteTable.java",
   "../../brave/android/java/org/chromium/chrome/browser/media/BraveFullscreenVideoPictureInPictureController.java",
   "../../brave/android/java/org/chromium/chrome/browser/media/BravePictureInPictureActivity.java",
+  "../../brave/android/java/org/chromium/chrome/browser/media/BraveYouTubePictureInPictureController.java",
```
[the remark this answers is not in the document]
simoarpe: Created a separate issue to investigate feasibility of this: https://github.com/brave/brave-browser/issues/55338

android/java/org/chromium/chrome/browser/media/BraveYouTubePictureInPictureController.java
```diff
@@ -0,0 +1,651 @@
...
+
+        // BraveMediaSessionHelper holds the active WebContents through a WeakReference that GC
+        // may have collected between save and restore. Recover via the persisted tab id instead:
+        // the TabModel outlives the Activity, so the id resolves back to the original tab and
+        // through it to its WebContents. We deliberately do not fall back to whatever tab is now
+        // foreground; if the user switched tabs during recreation, that would silently retarget
+        // the session at the wrong WebContents.
+        if (mTabId == Tab.INVALID_TAB_ID) {
+            return null;
+        }
```
netzenbot, line 512: nit: `maybeRestoreFullscreenUi` is called from `PostTask.postDelayedTask` callbacks but doesn't check `mActivity.isActivityFinishingOrDestroyed()` before calling `mActivity.getFullscreenManager()` and `exitPersistentFullscreenModeForPictureInPicture()`. The analogous delayed task in `onNewTabDuringPictureInPicture()` does include this guard — consider adding it here for consistency. See [AND-001](https://github.com/brave/brave-core/tree/master/docs/best-practices/android.md#AND-001).

patches/chrome-android-java-src-org-chromium-chrome-browser-fullscreen-FullscreenHtmlApiHandlerBase.java.patch
```diff
@@ -1,12 +1,74 @@
...
-@@ -320,6 +320,7 @@ public abstract class FullscreenHtmlApiHandlerBase
+@@ -320,6 +320,11 @@ public abstract class FullscreenHtmlApiHandlerBase
  
                      @Override
                      public void onHidden(Tab tab, @TabHidingType int reason) {
-+                        BraveFullscreenHtmlApiHandlerBase.class.cast(FullscreenHtmlApiHandlerBase.this).mTabHiddenByChangedTabs = reason == TabHidingType.CHANGED_TABS;
++                        if (BraveFullscreenHtmlApiHandlerBase.class.cast(
++                                        FullscreenHtmlApiHandlerBase.this)
++                                .maybeSkipExitFullscreenOnTabHidden(mActivity, reason)) {
++                            return;
```
netzenbot, line 12: This patch was previously a single-line addition but is now multiline across multiple hunks. Per the patch style guidelines, patches should be kept to one line. The same changes are covered by the new `.toml` Plaster substitution file — the patch content should stay minimal (ideally single-line) or the patch file should be removed in favor of the `.toml` approach. ([Patch Style Guidelines](https://github.com/brave/brave-core/tree/master/docs/best-practices/patches.md#PATCH-007))

patches/chrome-android-java-src-org-chromium-chrome-browser-media-FullscreenVideoPictureInPictureController.java.patch
```diff
@@ -0,0 +1,47 @@
+diff --git a/chrome/android/java/src/org/chromium/chrome/browser/media/FullscreenVideoPictureInPictureController.java b/chrome/android/java/src/org/chromium/chrome/browser/media/FullscreenVideoPictureInPictureController.java
```
netzenbot, line 1: This new patch file contains multiline additions (4 hunks with multiple added lines each). Patches should be kept to one line per the style guidelines. The identical changes are already expressed in the new `.toml` Plaster substitution file — consider whether this `.patch` file is needed at all, and if so, keep each hunk to a single added line. ([Patch Style Guidelines](https://github.com/brave/brave-core/tree/master/docs/best-practices/patches.md#PATCH-007))

AlexeyBarabash: this patch is generated by rewrite/chrome/android/java/src/org/chromium/chrome/browser/media/FullscreenVideoPictureInPictureController.java.toml, so the diagnostics is false-positive

patches/chrome-android-java-src-org-chromium-chrome-browser-fullscreen-FullscreenHtmlApiHandlerBase.java.patch
AlexeyBarabash: this patch is for `rewrite/chrome/android/java/src/org/chromium/chrome/browser/fullscreen/FullscreenHtmlApiHandlerBase.java.toml`, so it should be ok with multilines

android/java/org/chromium/chrome/browser/youtube_script_injector/BraveYouTubeScriptInjectorNativeHelper.java
```diff
@@ -30,6 +31,11 @@ public static void setFullscreen(WebContents webContents) {
         BraveYouTubeScriptInjectorNativeHelperJni.get().setFullscreen(webContents);
     }
 
+    // TODO - Uncomment once implemented in core.
```
AlexeyBarabash: if there is an issue or draft PR, it worth to specify in here along with TODO

android/java/org/chromium/chrome/browser/youtube_script_injector/BraveYouTubeScriptInjectorNativeHelper.java
```diff
@@ -71,9 +87,12 @@ public static void enterPictureInPicture(WebContents webContents) {
      * @noinspection unused
      */
     @NativeMethods
-    interface Natives {
+    public interface Natives {
         void setFullscreen(WebContents webContents);
 
+        // TODO - Uncomment once implemented in core.
```
AlexeyBarabash: same for this TODO, please put the issue or the draft PR to keep tracking

android/java/org/chromium/chrome/browser/media/BraveYouTubePictureInPictureController.java
simoarpe: This observation is reasonable, I've introduced a check to early return when the activity is finishing or destroyed 👉 [`5bde52d` (this PR)](https://github.com/brave/brave-core/pull/36838/commits/5bde52ddc32f03d9ec3c193489eddbbf36f53b82)

android/java/org/chromium/chrome/browser/youtube_script_injector/BraveYouTubeScriptInjectorNativeHelper.java
simoarpe: Added a link to an open PR where the core implementation is currently being implemented 👉 [`9cea421` (this PR)](https://github.com/brave/brave-core/pull/36838/commits/9cea42132236094b387fc78b0923753151dc132d)

android/java/org/chromium/chrome/browser/youtube_script_injector/BraveYouTubeScriptInjectorNativeHelper.java
simoarpe: Added a link to an open PR where the core implementation is currently being implemented 👉 [`9cea421` (this PR)](https://github.com/brave/brave-core/pull/36838/commits/9cea42132236094b387fc78b0923753151dc132d)

AlexeyBarabash (approved):
LGTM

rewrite/chrome/android/java/src/org/chromium/chrome/browser/fullscreen/FullscreenHtmlApiHandlerBase.java.yaml
```diff
@@ -0,0 +1,79 @@
...
+substitutions:
+  - description:
+      Preserve fullscreen state while Brave-managed YouTube PiP is active when a
+      tab is hidden
+    re_pattern:
+      '(public void onHidden\(Tab tab, @TabHidingType int
+      reason\)\s+\{\n)([\s\S]*?\n)
+      +mFullscreenManagerDelegate\.onExitFullscreen\(null\);'
+    replace: |-
+      \1                        if (BraveFullscreenHtmlApiHandlerBase.class.cast(
```
cdesouza-chromium: So, fair disclaimer: My `java` knowledge is atrocious, and I don't have much domain on the language or the codebase as a whole. So here's what I'm wondering: We are turning this:

```java
@Override
public void onHidden(Tab tab, @TabHidingType int reason) {
    // Clean up any fullscreen state that might impact other tabs.
    mFullscreenManagerDelegate.onExitFullscreen(null);
}
```

Into this:

```java
@Override
public void onHidden(Tab tab, @TabHidingType int reason) {
    if (BraveFullscreenHtmlApiHandlerBase.class.cast(
                    FullscreenHtmlApiHandlerBase.this)
            .maybeSkipExitFullscreenOnTabHidden(mActivity, reason)) {
        return;
    }
    // Clean up any fullscreen state that might impact other tabs.
    mFullscreenManagerDelegate.onExitFullscreen(null);
}
```

Basically an insert at the top, but seeing this obviously I get to wonder if the solution here is to use an override, where we our class does:

```java
@Override
public void onHidden(Tab tab, @TabHidingType int reason) {
    if (BraveFullscreenHtmlApiHandlerBase.class.cast(
                    FullscreenHtmlApiHandlerBase.this)
            .maybeSkipExitFullscreenOnTabHidden(mActivity, reason)) {
        return;
    }
    SuperClass.onHidden(tab, reason);
}
```

Of course this means inserting our derived class where it is needed, and I'm not sure if that's worse, but it would be good to know if that direction would result in something that is clearer.

rewrite/chrome/android/java/src/org/chromium/chrome/browser/fullscreen/FullscreenHtmlApiHandlerBase.java.yaml
```diff
@@ -0,0 +1,79 @@
...
+                                              FullscreenHtmlApiHandlerBase.this)
+                                      .maybeSkipExitFullscreenOnTabHidden(mActivity, reason)) {
+                                  return;
+                              }
+      \2                        mFullscreenManagerDelegate.onExitFullscreen(null);
+  - description: Add a PiP-specific persistent fullscreen exit path
```
cdesouza-chromium: This description field and others can receive a bit more of love. We are actually inserting overloads to `exitPersistentFullscreenMode`, so it would be nice to have a What and Why at this field.

rewrite/chrome/android/java/src/org/chromium/chrome/browser/fullscreen/FullscreenHtmlApiHandlerBase.java.yaml
```diff
@@ -0,0 +1,79 @@
...
+    re_pattern:
+      '(@Override\n\s+public void
+      exitPersistentFullscreenMode\(\)\s+\{\n)([\s\S]*?\n)
+      +returnFromTargetScreenIfNeeded\(\);'
```
cdesouza-chromium: Is there a reason to be matching to `returnFromTargetScreenIfNeeded` here?

rewrite/chrome/android/java/src/org/chromium/chrome/browser/fullscreen/FullscreenHtmlApiHandlerBase.java.yaml
```diff
@@ -0,0 +1,79 @@
...
+  - description:
+      Thread the WebContents fullscreen exit flag into fullscreen teardown
+    re_pattern:
+      '(private void exitPersistentFullscreenMode\(boolean
+      exitWebContentsFullscreen\)[\s\S]+?assumeNonNull\(mContentViewInFullscreen\);\n\s+)exitFullscreen\(\s*mWebContentsInFullscreen,\s*mContentViewInFullscreen\);'
+    replace: |-
+      \1exitFullscreen(
+                              mWebContentsInFullscreen,
+                              mContentViewInFullscreen,
+                              exitWebContentsFullscreen);
```
cdesouza-chromium: So the previous substitution changes the signature of `exitPersistentFullscreenMode`, but this one has to plumb the argument into the `exitFullscreen` call, but then down below we also change the signature of `exitFullscreen`, and then use an extra substitution to inject handling.

Again, I don't have much knowledge about how this code should work, but this seems to be getting quite invasive, and looking at this plaster file, it is very complicated to me to know what to do when this plaster break during rebase, so I think it would be good to know what are the other options that could be potential different path ways to do this, or if we absolutely have no other option.

rewrite/chrome/android/java/src/org/chromium/chrome/browser/fullscreen/FullscreenHtmlApiHandlerBase.java.yaml
```diff
@@ -0,0 +1,79 @@
...
+              exitPersistentFullscreenMode(false);
+          }
+
+          private void exitPersistentFullscreenMode(boolean exitWebContentsFullscreen) {
+      \2        returnFromTargetScreenIfNeeded();
+  - description:
+      Thread the WebContents fullscreen exit flag into fullscreen teardown
+    re_pattern:
+      '(private void exitPersistentFullscreenMode\(boolean
+      exitWebContentsFullscreen\)[\s\S]+?assumeNonNull\(mContentViewInFullscreen\);\n\s+)exitFullscreen\(\s*mWebContentsInFullscreen,\s*mContentViewInFullscreen\);'
```
cdesouza-chromium: There are other examples in the codebase of plasters doing this, but basically you the original content is this:

```java
exitFullscreen(mWebContentsInFullscreen, mContentViewInFullscreen);
```

And you want this:
```java
exitFullscreen(mWebContentsInFullscreen, mContentViewInFullscreen, exitWebContentsFullscreen);
```

So all you want is to insert an extra arg at the end, and this is what your plaster should be doing here: find this function call and insert an argument at the end of the arg list, this way you don't have to touch any of the arguments already passed, and this doesn't break because things changed, and a new argument is being passed into this function.

rewrite/chrome/android/java/src/org/chromium/chrome/browser/fullscreen/FullscreenHtmlApiHandlerBase.java.yaml
```diff
@@ -0,0 +1,79 @@
...
+  - description: Add a PiP-specific persistent fullscreen exit path
+    re_pattern:
+      '(@Override\n\s+public void
+      exitPersistentFullscreenMode\(\)\s+\{\n)([\s\S]*?\n)
+      +returnFromTargetScreenIfNeeded\(\);'
+    replace: |-
+      \1        exitPersistentFullscreenMode(true);
+          }
+
+          public void exitPersistentFullscreenModeForPictureInPicture() {
```
cdesouza-chromium: Rather than having:

```java
    @Override
    public void exitPersistentFullscreenMode() {
        exitPersistentFullscreenMode(true);
    }

// ...

  private void exitPersistentFullscreenMode(boolean exitWebContentsFullscreen) {
```

Have you considered just making:

```java
private void exitPersistentFullscreenMode(boolean exitWebContentsFullscreen = true) {
```

I'm not sure if this is valid in Java, and this function also has this `@Override` thing, so maybe that can't be done, but in C++ this can be doable in certain cases.

rewrite/chrome/android/java/src/org/chromium/chrome/browser/fullscreen/FullscreenHtmlApiHandlerBase.java.yaml
```diff
@@ -0,0 +1,79 @@
...
+  - description:
+      Allow fullscreen teardown to decide whether WebContents fullscreen should
+      exit
+    re_pattern:
+      'private void exitFullscreen\(WebContents webContents, View
+      contentView\)\s+\{'
+    replace: |-
+      private void exitFullscreen(
+                  WebContents webContents, View contentView, boolean exitWebContentsFullscreen) {
+  - description: Skip WebContents fullscreen exit for the Brave-managed PiP path
```
cdesouza-chromium: Please keep an empty line between each substitution

rewrite/chrome/android/java/src/org/chromium/chrome/browser/fullscreen/FullscreenHtmlApiHandlerBase.java.yaml
```diff
@@ -0,0 +1,79 @@
...
+    re_pattern:
+      '(private void exitFullscreen\(\s*WebContents webContents,\s*View
+      contentView,\s*boolean
+      exitWebContentsFullscreen\)\s+\{[\s\S]+?resetExitFullscreenLayoutChangeListener\(contentView\);\n\s+)if\s+\(webContents
+      != null && !webContents\.isDestroyed\(\)\)
+      webContents\.exitFullscreen\(\);'
+    replace: |-
+      \1if (exitWebContentsFullscreen && webContents != null && !webContents.isDestroyed()) {
```
cdesouza-chromium: This goes back to the advice I gave elsewhere:

We have this:

```java
if (webContents != null && !webContents.isDestroyed()) webContents.exitFullscreen();
```

All you want is to do this:

```java
if (exitWebContentsFullscreen && webContents != null && !webContents.isDestroyed()) webContents.exitFullscreen();
```

So this is a matter of inserting in the beginning of the statement, regardless of what is in the statement, so it would be better to: match to exitFullscreen, then match to the if(*) that has `webContents.exitFullscreen()`, and insert at the beginning.

rewrite/chrome/android/java/src/org/chromium/chrome/browser/fullscreen/FullscreenHtmlApiHandlerBase.java.yaml
```diff
@@ -0,0 +1,79 @@
...
+                  webContents.exitFullscreen();
+              }
+  - description:
+      Preserve fullscreen state while Brave-managed YouTube PiP is active when
+      the activity stops
+    re_pattern:
+      '(public void onActivityStateChange\(Activity activity, int
+      newState\)\s+\{\n\s+if\s+\(newState == ActivityState\.STOPPED &&
+      mExitFullscreenOnStop\)\s+\{\n)([\s\S]*?\n)
+      +mFullscreenManagerDelegate\.onExitFullscreen\(null\);(\n\s+\} else if
```
cdesouza-chromium: This pattern is matching onto too many things... This is very fragile, and will break easily for unrelated reason. I'm not sure what your intent was exactly, but looking on the java source, it seems to me that the intent here is to prevent `mFullscreenManagerDelegate.onExitFullscreen(null)` from being called if a certain condition is met, so I wonder if it would be cleaner to match `onActivityStateChange` and then `mFullscreenManagerDelegate.onExitFullscreen(null);`, and insert prior to that.

cdesouza-chromium (commented):
Leaving some comments. My knowledge of Java is not great, so if you want to use Plaster here, we will need a fair amount of communication to determine what is the best way use it in these cases.

Also, please take a look at https://github.com/brave/brave-core/pull/36525

rewrite/chrome/android/java/src/org/chromium/chrome/browser/fullscreen/FullscreenHtmlApiHandlerBase.java.yaml
simoarpe: This was a solution that I've considered in the past, and I've invested quite some time trying to achieve it. Unfortunately there are some limitations with our current bytecode manipulation algorithm.
More specifically the method above is part of an inner class inside `FullscreenHtmlApiHandlerBase`:

```java
mTabFullscreenObserver =
                new TabModelSelectorTabObserver(modelSelector) {
                    @Override
                    public void onContentChanged(Tab tab) {
                        setContentView(tab.getContentView());
                    }

                    @Override
                    public void onHidden(Tab tab, @TabHidingType int reason) {
                   // ...
```

The class is instantiated as an anonymous class and is not supported by our `redirectContructor` method, so at the moment we wouldn't have any way to re-insert the derived class where it's needed.
I've already filed a GitHub issue some time ago to keep track of this limitation: https://github.com/brave/brave-browser/issues/44473.

rewrite/chrome/android/java/src/org/chromium/chrome/browser/fullscreen/FullscreenHtmlApiHandlerBase.java.yaml
simoarpe: Done by [`50701d2` (this PR)](https://github.com/brave/brave-core/pull/36838/commits/50701d2ad3dd749b978332f4ed594fb41ad319c2)

rewrite/chrome/android/java/src/org/chromium/chrome/browser/fullscreen/FullscreenHtmlApiHandlerBase.java.yaml
simoarpe: Oh, right! `returnFromTargetScreenIfNeeded` can be safely removed as we just need to replace the beginning of the method. I've improved this substitution in commit 👉 [`fde8a76` (this PR)](https://github.com/brave/brave-core/pull/36838/commits/fde8a7687060cb619153f42f3d054aa493d2e1a7)

rewrite/chrome/android/java/src/org/chromium/chrome/browser/fullscreen/FullscreenHtmlApiHandlerBase.java.yaml
simoarpe: Okay, there's another solution that would simplify the modifications:
we can save on a boolean member the state in the super class (i.e. `mSkipWebContentsExitFullscreenForPip`) and then perform the check only in the `exitFullscreen` method where we interrogate the value.
This solution would remove all the plumbings. The only part where we should take extra care is resetting the value to its original state once done; we can lock it using a `try`/`finally` clause that will guarantee it will be always reset.

Pushed the changes in [`a2013ad` (this PR)](https://github.com/brave/brave-core/pull/36838/commits/a2013ad021927d9e8076f2c5dbbe6e5922d4172c)
Let me know your thoughts.

rewrite/chrome/android/java/src/org/chromium/chrome/browser/fullscreen/FullscreenHtmlApiHandlerBase.java.yaml
simoarpe: I've improved this in commit 👉 [`96deeda` (this PR)](https://github.com/brave/brave-core/pull/36838/commits/96deedacdaf054d3f66895f1b87e229f571788c6)

rewrite/chrome/android/java/src/org/chromium/chrome/browser/fullscreen/FullscreenHtmlApiHandlerBase.java.yaml
simoarpe: I've improved all the descriptions for FullscreenHtmlApiHandlerBase.java.yaml. Let me know if anything is unclear now.

rewrite/chrome/android/java/src/org/chromium/chrome/browser/fullscreen/FullscreenHtmlApiHandlerBase.java.yaml
simoarpe: With the new strategy adopted [here](https://github.com/brave/brave-core/pull/36838#discussion_r3342921614), this substitution is not needed anymore.

rewrite/chrome/android/java/src/org/chromium/chrome/browser/fullscreen/FullscreenHtmlApiHandlerBase.java.yaml
simoarpe: With the new strategy adopted [here](https://github.com/brave/brave-core/pull/36838#discussion_r3342921614), this substitution is not needed anymore.

rewrite/chrome/android/java/src/org/chromium/chrome/browser/fullscreen/FullscreenHtmlApiHandlerBase.java.yaml
simoarpe: Improve substitution in commit 👉 [`96deeda` (this PR)](https://github.com/brave/brave-core/pull/36838/commits/96deedacdaf054d3f66895f1b87e229f571788c6)

rewrite/chrome/android/java/src/org/chromium/chrome/browser/fullscreen/FullscreenHtmlApiHandlerBase.java.yaml
```diff
@@ -0,0 +1,75 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+substitutions:
+  - description:
+      Check at the top of the method if exiting fullscreen should be skipped to
+      preserve the state while Brave-managed YouTube PiP is active when a tab is
+      hidden
+    re_pattern:
+      '(public void onHidden\(Tab tab, @TabHidingType int reason\)\s+\{)'
```
cdesouza-chromium: This is better, but I think we can make intent a bit clearer here: let's match to `public void onHidden(***)` not matching the specifics in the arglist.

rewrite/chrome/android/java/src/org/chromium/chrome/browser/fullscreen/FullscreenHtmlApiHandlerBase.java.yaml
```diff
@@ -0,0 +1,75 @@
...
+    replace: |-
+      \1
+                              if (BraveFullscreenHtmlApiHandlerBase.class.cast(
+                                              FullscreenHtmlApiHandlerBase.this)
+                                      .maybeSkipExitFullscreenOnTabHidden(mActivity, reason)) {
+                                  return;
+                              }
+
+  - description:
+      Add a PiP-specific persistent fullscreen exit entry point,
```
cdesouza-chromium: So, this should have been better documented on my end, and I will do that, but the practice for description fields is to have one Subject line, an empty line, and then the long description, similar to how docstrings are.

rewrite/chrome/android/java/src/org/chromium/chrome/browser/fullscreen/FullscreenHtmlApiHandlerBase.java.yaml
```diff
@@ -0,0 +1,75 @@
...
+              self.mSkipWebContentsExitFullscreenForPip = true;
+              try {
+                  exitPersistentFullscreenMode();
+              } finally {
+                  self.mSkipWebContentsExitFullscreenForPip = false;
+              }
+          }
+          \1
+
+  - description:
```
cdesouza-chromium: If we are starting the description let's use `- description: |-` on this and all the other fields, so it makes sense when we start harvesting these fields.

rewrite/chrome/android/java/src/org/chromium/chrome/browser/media/FullscreenVideoPictureInPictureController.java.yaml
```diff
@@ -0,0 +1,48 @@
+# Copyright (c) 2026 The Brave Authors. All rights reserved.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+substitutions:
+  - description:
+      Let Brave handle YouTube PiP dismiss signals using upstream enum names
+    re_pattern:
+      '(\s+private void dismissActivityIfNeeded\(Activity activity,
```
cdesouza-chromium: Same here, let's not match the arguments in the arglist.

rewrite/chrome/android/java/src/org/chromium/chrome/browser/media/FullscreenVideoPictureInPictureController.java.yaml
```diff
@@ -0,0 +1,48 @@
+# Copyright (c) 2026 The Brave Authors. All rights reserved.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+substitutions:
```
cdesouza-chromium, line 6: Let's try to list the substitutions in the order they are affecting the file, as they seem to be ordered in different order than they occur.

rewrite/chrome/android/java/src/org/chromium/chrome/browser/media/FullscreenVideoPictureInPictureController.java.yaml
```diff
@@ -0,0 +1,48 @@
...
+                      .onYouTubePictureInPictureAttempt(mActivity, webContents);
+
+  - description: Notify Brave when manual PiP entry returns false
+    re_pattern:
+      'if \(!mActivity\.enterPictureInPictureMode\(builder\.build\(\)\)\)
+      return;'
+    replace: |-
+      if (!mActivity.enterPictureInPictureMode(builder.build())) {
+                      BraveFullscreenVideoPictureInPictureController.class.cast(this)
+                              .onYouTubePictureInPictureEnterFailed();
```
cdesouza-chromium: So, in this case, `attemptPictureInPicture` has three insertions in its body, one for `onYouTubePictureInPictureAttempt`, and then two others for the try/catch block, so I'm wondering if it would be easier to rename `attemptPictureInPicture` to `attemptPictureInPictureInternal` and then we add our own `attemptPictureInPicture` that verifies the outside state afterwards and decides what to do. If that's possible we would not need as many separate in-body insertions.

cdesouza-chromium (commented):
Hey @simoarpe this is better... thanks for the PR... we are nearly there. Just a few more things.

rewrite/chrome/android/java/src/org/chromium/chrome/browser/media/FullscreenVideoPictureInPictureController.java.yaml
simoarpe: Yes, it makes perfect sense. I've adjusted the order of the substitutions and they now match the git patch.
Done by 👉 [`08df42b` (this PR)](https://github.com/brave/brave-core/pull/36838/commits/08df42bc4130f3002a8e8b3ad88bdb6a50ce98ef).

rewrite/chrome/android/java/src/org/chromium/chrome/browser/fullscreen/FullscreenHtmlApiHandlerBase.java.yaml
simoarpe: Improved regex with `(public void onHidden\(.*\)\s+\{)` by commit [`767d35a` (this PR)](https://github.com/brave/brave-core/pull/36838/commits/767d35af858ed9705f7ae902f7597ea19ed3cdd0)

Note: I've opted for a greedy greedy quantifier because _potentially_ we can have argument annotations with parenthesis, e.g:

```java
public @Nullable Region getRegionAt(@IntRange(from = 0) int index) {
```

It's really unlikely it will be introduced, but using a greedy quantifier we are safe here.

rewrite/chrome/android/java/src/org/chromium/chrome/browser/fullscreen/FullscreenHtmlApiHandlerBase.java.yaml
simoarpe: Fixed by 👉 [`bfe416d` (this PR)](https://github.com/brave/brave-core/pull/36838/commits/bfe416d59b41d253e823174affa5328123084069)

rewrite/chrome/android/java/src/org/chromium/chrome/browser/media/FullscreenVideoPictureInPictureController.java.yaml
simoarpe: Done by 👉 [`858923b` (this PR)](https://github.com/brave/brave-core/pull/36838/commits/858923b89193d2bba1f944c645c4752a4b5a4679)

I've opted for a greedy quantifier here as well. Explained [in this comment](https://github.com/brave/brave-core/pull/36838#discussion_r3388782845) why.

rewrite/chrome/android/java/src/org/chromium/chrome/browser/fullscreen/FullscreenHtmlApiHandlerBase.java.yaml
simoarpe: Descriptions improved by commit 👉 [`6704f7b` (this PR)](https://github.com/brave/brave-core/pull/36838/commits/6704f7bb00022bcb3bc4fee7f1022495a7734088)

rewrite/chrome/android/java/src/org/chromium/chrome/browser/media/FullscreenVideoPictureInPictureController.java.yaml
simoarpe: I've improved the replace strategy delegating all the three substitutions to one single method.
Applied by commit 👉 [`ac06c7f` (this PR)](https://github.com/brave/brave-core/pull/36838/commits/ac06c7fa5d1711902a346a22ae6845052bf2e741)

Let me know if it makes it easier to maintain. The `re_pattern` is now more complex but still robust: it matches the try/catch block inside `attemptPictureInPicture` method. That would guarantee us a safe substitution even in case of small changes (unlikely, but we would be protected from conflicts).

netzenbot (approved):

rewrite/chrome/android/java/src/org/chromium/chrome/browser/media/FullscreenVideoPictureInPictureController.java.yaml
```diff
@@ -0,0 +1,47 @@
...
+      session state back if entry is refused or throws. Keeping all of that in the
+      Brave superclass means upstream just hands off the built params. The pattern
+      is anchored inside attemptPictureInPicture() and matches the try/catch by
+      balanced braces (tolerating up to two levels of nested {} in either body,
+      independent of line breaks or indentation), with a content-agnostic
+      catch(...), so it stays scoped to this method and does not depend on the
+      refusal check, the caught exception types, or the log statement inside.
+    re_pattern:
+      '(public void attemptPictureInPicture\(\)
+      \{[\s\S]*?)\btry\s*\{(?:[^{}]|\{(?:[^{}]|\{[^{}]*\})*\})*\}\s*catch\s*\([^)]*\)\s*\{(?:[^{}]|\{(?:[^{}]|\{[^{}]*\})*\})*\}'
```
cdesouza-chromium: I think we can simplify this a bit by find this scope and then doing our insertion just before the `try...catch` block, and doing an early return. This means we don't delete the `try...catch` block itself from the source. We just make it unreachable.

cdesouza-chromium (approved):
LGTM % one comment

rewrite/chrome/android/java/src/org/chromium/chrome/browser/media/FullscreenVideoPictureInPictureController.java.yaml
simoarpe: Oh, right! This would be even better; I'm going to try it and if lint won't complaint it would make things easier to read and apply.

android/java/org/chromium/chrome/browser/media/BraveFullscreenVideoPictureInPictureController.java
```diff
@@ -6,28 +6,160 @@
...
+            @Nullable final Rect bounds) {
+        onYouTubePictureInPictureAttempt(activity, webContents);
+        try {
+            if (!activity.enterPictureInPictureMode(builder.build())) {
+                onYouTubePictureInPictureEnterFailed();
+            }
+        } catch (IllegalStateException | IllegalArgumentException e) {
+            Log.e(TAG, "Error entering PiP with bounds %s", bounds, e);
+            onYouTubePictureInPictureEnterFailed();
+        }
```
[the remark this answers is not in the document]
simoarpe: `mPendingBraveActivityForPiP` is already cleared in a `finally` block: safer. This comment is wrong.

android/java/org/chromium/chrome/browser/media/BraveYouTubePictureInPictureController.java
```diff
@@ -0,0 +1,892 @@
...
+    public void onDestroy() {
+        unregisterScreenStateReceiver();
+        // On a configuration change the recreated activity restores the session from the
+        // saved instance bundle, so leave the registry intact for it to pick up. Only clear
+        // when this controller actually owns a session: every BraveActivity lazily creates a
+        // controller on its destroy path, and clear(null) from a session-less window would wipe
+        // another window's live registry entry (the slot is process-global).
+        if (!mActivity.isChangingConfigurations() && (mActive || mWebContents != null)) {
+            BraveMediaSessionHelper.clearYouTubePictureInPictureWebContents(mWebContents);
+        }
```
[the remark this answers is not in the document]
simoarpe: Legit suggestion. Applied!

android/java/org/chromium/chrome/browser/media/BraveYouTubePictureInPictureController.java
```diff
@@ -0,0 +1,892 @@
...
+    }
+
+    private void clearSession(final int sessionId, @Nullable final WebContents webContents) {
+        if (sessionId != mSessionId || webContents != mWebContents) {
+            return;
+        }
+
+        unregisterScreenStateReceiver();
+        BraveMediaSessionHelper.clearYouTubePictureInPictureWebContents(webContents);
+        resetSessionState();
```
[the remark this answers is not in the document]
simoarpe: Legit suggestion. Applied!

rewrite/chrome/android/java/src/org/chromium/chrome/browser/media/FullscreenVideoPictureInPictureController.java.yaml
simoarpe: Done by [`e3b5c3e` (this PR)](https://github.com/brave/brave-core/pull/36838/commits/e3b5c3efae6ec1759d7c8364400ea21c7cc13a67)

PR #36845, opened 2026-05-29
Uplift upstream CL: add URL-id filter to vector search

darkdh:
Resolves https://github.com/brave/brave-browser/issues/55935

Mirrors https://chromium-review.googlesource.com/c/chromium/src/+/7902082
(merged as 95bcdbb98451). Threads a vector<URLID> url_id_filter from
HistoryEmbeddingsSearch::Search through SearchParams into
VectorDatabase::MakeUrlDataIterator. When non-empty, SqlDatabase fetches
only matching rows via 'WHERE passages.url_id IN (?,...)' in a single
statement; VectorDatabaseInMemory honors it inline.

Existing upstream callers (omnibox provider, WebUI handler, AI data
service) pass {} to preserve current behavior. 

Patches drop once the
local chromium baseline includes the merged CL.

PR #36846, opened 2026-05-29
Augment tab_search popup with on-device semantic tab ranking

### LINKED ISSUE #55936: Integrate semantic tab search into tab_search UI
Image

There should be no UI changes, we just also provide semantic tab search result

darkdh:
- Resolves https://github.com/brave/brave-browser/issues/55936

 - Depends on #36845
 - Depends on https://github.com/brave/brave-core/pull/37287

## Summary
Adds `SearchTabsByContent(query)` to the upstream `tab_search.mojom`
PageHandler interface (via chromium_src `[BraveExtend]`) and implements
it in the Brave-side `TabSearchPageHandler` override.

The handler is gated on `IsHistoryEmbeddingsEnabledForProfile(profile)`
so the feature is unavailable when the Leo "Semantic History Search"
toggle is off. When enabled, it enumerates open tabs across this
profile's windows via `GlobalBrowserCollection`/`TabStripModel`,
resolves each tab URL to its `history::URLID` in a single
`HistoryService::QueryUrlIds` call, then runs
`HistoryEmbeddingsService::Search` with that URL-id filter so the SQL
layer restricts the vector search to the open-tab set. Scored results
are mapped back to `TabHandle`s for the frontend.

A chromium_src TS override of `tab_search_page.ts` subclasses
`TabSearchPageElement` (registered under a new `<brave-tab-search-page>`
tag, so upstream's `customElements.define` doesn't need patching) and
overrides `onSearchTermInput()` to fire `searchTabsByContent` async. A
companion chromium_src override of `search.ts` wraps upstream's
synchronous matcher and, on the next render pass, merges in any
semantic results the subclass cached out-of-band — so no patching of
upstream's `updateFilteredTabs_` is required. A monotonic token
discards stale responses when the user types faster than the on-device
ranker can complete.

## Test plan
1. Enable brave://flags/#brave-history-embeddings
2. Enable "Semantic History Search" toggle in brave://history
3. Navigate to some sites and make sure semantic search in history page has the result
 (ex. \`zkp and age verification\` for https://www.eff.org/deeplinks/2025/07/zero-knowledge-proofs-alone-are-not-digital-id-solution-protecting-user-privacy
4. And do the same query on tab search
Screenshot 2026-06-26 at 13 41 30


chromium_src/chrome/browser/ui/webui/tab_search/tab_search_page_handler.cc
```diff
@@ -5,15 +5,32 @@
...
 
+#include <memory>
+#include <numeric>
+
 #include "base/check.h"
+#include "base/containers/flat_map.h"
+#include "base/containers/map_util.h"
+#include "base/functional/bind.h"
 #include "base/strings/string_number_conversions.h"
 #include "brave/components/ai_chat/core/common/buildflags/buildflags.h"
```
netzenbot, line 17: These new includes (history_service_factory.h, history_embeddings_service_factory.h, history_service.h, history_types.h, tab_interface.h, web_contents.h, etc.) are only used by `SearchTabsByContent`, which is under `#if BUILDFLAG(ENABLE_AI_CHAT)`. Move them inside that guard so they aren't pulled in when AI Chat is disabled. See [BS-034](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-034).

components/history_embeddings/BUILD.gn
```diff
@@ -11,3 +11,17 @@ source_set("history_embeddings") {
...
+source_set("test_support") {
+  testonly = true
+  sources = [
+    "test/fake_history_embeddings_search.cc",
+    "test/fake_history_embeddings_search.h",
+  ]
+  deps = [
+    "//base",
+    "//components/history/core/browser",
+    "//components/history_embeddings/core",
```
netzenbot, line 23: nit: the fake's header includes `history_embeddings_search.h` (the base class) and `history_types.h` (`history::URLID` appears in the public API), so `//components/history_embeddings/core` and `//components/history/core/browser` should be `public_deps`, not `deps`. See [BS-024](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-024).

chromium_src/chrome/browser/ui/webui/tab_search/tab_search_page_handler.cc
```diff
@@ -320,4 +458,10 @@ void TabSearchPageHandler::GetTabFocusShowFRE(
     GetTabFocusShowFRECallback callback) {
   std::move(callback).Run(false);
 }
+
+void TabSearchPageHandler::SearchTabsByContent(
+    const std::string& query,
+    SearchTabsByContentCallback callback) {
+  std::move(callback).Run({});
+}
 #endif  // BUILDFLAG(ENABLE_AI_CHAT)
```
netzenbot: nit: this manual reserve + push_back loop can be replaced with `base::ToVector(tabs, &PendingTab::url)`, which handles the reserve and projection for you. ([Use `base::ToVector` for Range-to-Vector Conversions](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-020))

chromium_src/chrome/browser/ui/webui/tab_search/tab_search_page_handler.cc
```diff
@@ -294,6 +311,127 @@ void TabSearchPageHandler::GetTabFocusShowFRE(
...
+  }
+  history_service->QueryUrlIds(
+      urls,
+      base::BindOnce(&OnUrlIdsResolved, std::move(tabs), query,
+                     embeddings_search, std::move(callback)),
+      &query_url_task_tracker_);
+}
 #else   // !BUILDFLAG(ENABLE_AI_CHAT)
 // Stub implementations when AI Chat is disabled
 void TabSearchPageHandler::GetSuggestedTopics(
```
netzenbot: nit: SearchTabsByContent has multiple early-return branches that all `Run({})` (embeddings disabled, empty query / null services, no eligible tabs). A one-line comment before each summarizing what it guards would make the failure paths easier to follow. ([Document Non-Obvious Failure Branches](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-021))

chromium_src/chrome/browser/ui/webui/tab_search/tab_search_page_handler.cc
netzenbot: nit: This BindRepeating callback is complex enough (a loop mapping scored rows back to tab ids) that extracting it into a named free function in the anonymous namespace would make the Search() call site easier to read. See [CS-052](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-052).

chromium_src/chrome/browser/resources/tab_search/tab_search_page.ts
```diff
@@ -0,0 +1,57 @@
...
+// Augments the substring-filtered open-tab list with on-device semantic
+// matches from `searchTabsByContent`. A monotonic token discards stale
+// responses when the user types faster than the on-device ranker completes.
+class BraveTabSearchPageElement extends TabSearchPageElement {
+  private semanticSearchToken_: number = 0
+
+  protected override async processFilteredOpenTabs_(
+      tabs: Array<TabData|SplitViewData>):
+      Promise<Array<TabData|SplitViewData>> {
+    const searchText = (this as unknown as {searchText_: string}).searchText_
```
fallaciousreasoning, line 41: just confirming, but we need this because the field is private?

chromium_src/chrome/browser/ui/webui/tab_search/tab_search_page_handler.cc
```diff
@@ -294,6 +311,127 @@ void TabSearchPageHandler::GetTabFocusShowFRE(
   std::move(callback).Run(!Profile::FromWebUI(web_ui_)->GetPrefs()->HasPrefPath(
       ai_chat::prefs::kBraveAIChatTabOrganizationEnabled));
 }
+
+namespace {
```
fallaciousreasoning: nit: should this anonymous namespace be at the top? I haven't seen one added in the middle before

patches/chrome-browser-resources-tab_search-tab_search_page.ts.patch
```diff
@@ -0,0 +1,29 @@
...
++++ b/chrome/browser/resources/tab_search/tab_search_page.ts
+@@ -158,6 +158,11 @@ export class TabSearchPageElement extends TabSearchSearchFieldBase {
+   private filteredOpenHeaderIndices_: number[] = [];
+   private initiallySelectedIndex_: number = NO_SELECTION;
+   private documentVisibilityChangedListener_: () => void;
++  // Override hook: post-process the substring-filtered open-tab list.
++  protected async processFilteredOpenTabs_(
++      tabs: Array<TabData|SplitViewData>): Promise<Array<TabData|SplitViewData>> {
++    return tabs;
++  }
```
fallaciousreasoning: I think this function can be injected via a `chromium_src` override. If it turns out it can't, this injection should be done as a single line without a comment.

patches/chrome-browser-resources-tab_search-tab_search_page.ts.patch
```diff
@@ -0,0 +1,29 @@
...
+ 
+   constructor() {
+     super();
+@@ -856,6 +861,7 @@ export class TabSearchPageElement extends TabSearchSearchFieldBase {
+ 
+     let filteredOpenTabs = search<TabData|SplitViewData>(
+         this.searchText_, this.openTabs_, this.searchOptions_);
++    filteredOpenTabs = await this.processFilteredOpenTabs_(filteredOpenTabs);
```
fallaciousreasoning: I think we could completely remove the patching here by overriding the `search` function in a `chromium_src` override - it gets passed all the things we need (this.searchText_, this.openTabs_).

fallaciousreasoning (commented):
mostly lgtm, but I think we should override the search function with a chromium_src override and remove the patch.

chromium_src/chrome/browser/resources/tab_search/tab_search_page.ts
darkdh: yes, both `searchText_` and `openTabs_` are declared private upstream

chromium_src/chrome/browser/ui/webui/tab_search/tab_search_page_handler.cc
darkdh: This is wrong, we should only put header include under guard when those headers are guarded in gn

components/history_embeddings/BUILD.gn
darkdh: fixed in 5e884f249295458a24011763ab92839bcbc274c8

chromium_src/chrome/browser/ui/webui/tab_search/tab_search_page_handler.cc
darkdh: fixed in ab46c63303de048b5d1b59f75b5e5c5296cbe2ed

chromium_src/chrome/browser/ui/webui/tab_search/tab_search_page_handler.cc
darkdh: comments added in b503d1b59b73e3f9083824312ea415186c0fb04a

chromium_src/chrome/browser/ui/webui/tab_search/tab_search_page_handler.cc
darkdh: fixed in 79733ed171dba7a7856c9a925b1a2aa6664c51a2

chromium_src/chrome/browser/ui/webui/tab_search/tab_search_page_handler.cc
darkdh: moved in 06bc564e86f0db107823fcedf29c06b3b8c8873d

patches/chrome-browser-resources-tab_search-tab_search_page.ts.patch
darkdh: patch removed in 1705abff73ae139c5ee25293a8b5eb197f2ea3b0

patches/chrome-browser-resources-tab_search-tab_search_page.ts.patch
darkdh: patch removed in 1705abff73ae139c5ee25293a8b5eb197f2ea3b0

chromium_src/chrome/browser/ui/webui/tab_search/tab_search_page_handler.cc
```diff
@@ -321,3 +406,83 @@ void TabSearchPageHandler::GetTabFocusShowFRE(
...
+                     embeddings_search, std::move(callback)),
+      &query_url_task_tracker_);
+}
+#else   // !BUILDFLAG(ENABLE_LOCAL_AI)
+void TabSearchPageHandler::SearchTabsByContent(
+    const std::string& query,
+    SearchTabsByContentCallback callback) {
+  std::move(callback).Run({});
+}
+#endif  // BUILDFLAG(ENABLE_LOCAL_AI)
```
netzenbot, line 489: nit: this comment references the filtering "above"; positional references break if code is inserted between. Reference the filtering step by what it does instead (e.g. "the ForEach loop only keeps tracked-browser HTTP/HTTPS tabs"). See [CS-024](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-024).

chromium_src/chrome/browser/resources/tab_search/tab_search_page.ts
```diff
@@ -0,0 +1,65 @@
...
+  override onSearchTermInput() {
+    super.onSearchTermInput()
+    const searchText = (this as unknown as {searchText_: string}).searchText_
+    if (!searchText) {
+      return
+    }
+    const token = ++this.semanticSearchToken_
+    const proxy =
+        BraveTabSearchApiProxyImpl.getInstance() as BraveTabSearchApiProxy
+    void proxy.searchTabsByContent(searchText).then(({tabIds}) => {
```
fallaciousreasoning, line 47: why do we need this `void` keyword? Is it because the previous line is a cast? In that case the convention is to start this line with a semicolon, rather than adding the void here

chromium_src/chrome/browser/resources/tab_search/tab_search_api_proxy.ts
fallaciousreasoning: I'm not 100% sure but I suspect we could just shadow this declaration on the Brave implementation and not need the casts where we use it.

fallaciousreasoning (approved):

chromium_src/chrome/browser/resources/tab_search/tab_search_page.ts
darkdh: It's for for explicitly discarding a floating Promise. `searchTabsByContent(...).then(...)` returns a `Promise<void>` we deliberately don't `await` (the `.then` handler does the work via side effects), and `void` marks the discard as intentional so `@typescript-eslint/no-floating-promises` (and readers) see it's on purpose rather than a missing `await`.

But I can make it like this to explicitly swallow the error if you prefer
```
  proxy.searchTabsByContent(searchText)
      .then(({tabIds}) => { ... })
      .catch(() => {}) 
```

chromium_src/chrome/browser/resources/tab_search/tab_search_api_proxy.ts
darkdh: But that is already there before my changes, is it necessary to do that in this PR?

chromium_src/chrome/browser/ui/webui/tab_search/tab_search_page_handler.cc
darkdh: dismissed per https://github.com/brave/brave-core/pull/37618

chromium_src/chrome/browser/ui/webui/tab_search/tab_search_page_handler.cc
darkdh: rephrased in 642246b3b4ceb07623bea90792bf334bb6610fec

chromium_src/chrome/browser/resources/tab_search/search.ts
```diff
@@ -0,0 +1,61 @@
...
+    return result
+  }
+  const tabIds = context.getSemanticTabIds(input)
+  if (!tabIds || tabIds.size === 0) {
+    return result
+  }
+  const present = new Set<number>()
+  for (const item of result) {
+    if (item instanceof TabDataClass) {
+      present.add(((item as TabData).tab as Tab).tabId)
```
emerick: It seems weird that we have to cast `tab` to `Tab`, that was definitely necessary?

emerick (approved):

chromium_src/chrome/browser/resources/tab_search/search.ts
darkdh: yes both casts `item as TabData` and `tab as Tab` are redundant
removed in https://github.com/brave/brave-core/pull/36846/commits/fe35843635a8aeab0e7b5ea0eab4e73c1fb1610c

chromium_src/chrome/browser/resources/tab_search/tab_search_api_proxy.ts
darkdh: shadow it as a separate commit
6b5d825418cf36fca986e05cebb26817ad75f9c5

chromium_src/chrome/browser/resources/tab_search/tab_search_page.ts
fallaciousreasoning: Happy to leave this up to you - I've never seen this before so was just a little lost :smile:

chromium_src/chrome/browser/resources/tab_search/tab_search_api_proxy.ts
fallaciousreasoning: sorry, should've said optional - thanks for making the change, I think its a bit neater now :smile:

fallaciousreasoning (approved):
still lgtm

chromium_src/chrome/browser/resources/tab_search/tab_search_page.ts
darkdh: it is from https://typescript-eslint.io/rules/no-floating-promises/#ignorevoid

PR #36847, opened 2026-05-29
[AI Chat] Search open tabs with a `tabSearch` markdown directive

### LINKED ISSUE #55937: Semantic tab search inline markdown in Leo
Allow Leo to do semantic tab search and WITHOUT including passages info in the result and append a clickable card for users to switch to the tab.

darkdh:
Resolves https://github.com/brave/brave-browser/issues/55937

- Depends on #36846.
- Depends on #37829.
- Depends on https://github.com/brave/aichat/pull/1741.

Lets Leo surface the user's own open tabs, matched semantically against
indexed page content, fully on device.

The model emits `::tabSearch[query]` in its answer; the browser searches the
open tabs via `UntrustedUIHandler::SearchForTabs` and renders the matches in
that position as clickable cards. The results never reach the model, so it
can't repeat or invent them.

Four commits:

1. `SearchOpenTabsByContent` returns ranked `OpenTabInfo` (with a `title`
   that falls back to `url.host()`) instead of bare tab_ids; tab_search's
   page handler maps them back to ids for its mojom callback.
2. Emit every matching tab per URL, so tabs sharing a URL each get an entry.
3. Add `SearchForTabs` and `SwitchToTab` to `UntrustedUIHandler`, with the
   desktop and iOS implementations.
4. Render the matches for the `tabSearch` directive in the untrusted frame.

Clicking a card calls `UntrustedUIHandler::SwitchToTab`, which resolves the id
via `tabs::TabHandle` and activates it through `TabStripModel`. Tab activation
carries no conversation state, so it lives on the UI handler rather than
routing through `ConversationHandler`.

`SearchForTabs` returns null where on-device tab search is unavailable
(non-desktop, or the history embeddings setting is off) and the directive
renders nothing — distinct from an empty list, which reports that no open
tab matched.

## Test plan
1. Enable brave://flags/#brave-history-embeddings
2. Enable "Semantic History Search" toggle in brave://history
3. Navigate to some sites and make sure semantic search in history page has the result
(ex. `zkp and age verification` for https://www.eff.org/deeplinks/2025/07/zero-knowledge-proofs-alone-are-not-digital-id-solution-protecting-user-privacy
4. Ask Leo to find open tabs related to `zkp and age verification` for example
Screenshot 2026-08-11 at 15 06 51

5. It should render the clickable card that switches to the tab

### DIFF
--- a/browser/ai_chat/tools/tab_semantic_search_tool.cc
+++ b/browser/ai_chat/tools/tab_semantic_search_tool.cc
@@ -46,7 +46,16 @@ void OnRanked(Tool::UseToolCallback callback,
     return;
   }
   std::string serialized = internal::BuildSemanticSearchResultsJson(tabs);
-  std::move(callback).Run(CreateContentBlocksForText(serialized), {});
+  std::vector<mojom::ToolArtifactPtr> artifacts;
+  std::string sources_json = internal::BuildSemanticSearchTabSourcesJson(tabs);
+  if (!sources_json.empty()) {
+    auto artifact = mojom::ToolArtifact::New();
+    artifact->type = mojom::kTabSourcesArtifactType;
+    artifact->content_json = std::move(sources_json);
+    artifacts.push_back(std::move(artifact));
+  }
+  std::move(callback).Run(CreateContentBlocksForText(serialized),
+                          std::move(artifacts));
 }
 
 }  // namespace
@@ -65,6 +74,26 @@ std::string BuildSemanticSearchResultsJson(
   }
   base::DictValue root;
   root.Set("results", std::move(results_list));
+  std::string serialized;
+  base::JSONWriter::Write(root, &serialized);
+  return serialized;
+}
+
+std::string BuildSemanticSearchTabSourcesJson(
+    const std::vector<history_embeddings::OpenTabInfo>& tabs) {
+  base::ListValue sources;
+  for (const auto& tab : tabs) {
+    base::DictValue entry;
+    entry.Set("tab_id", tab.tab_id);
+    entry.Set("title", tab.title);
+    entry.Set("url", tab.url.spec());
+    sources.Append(std::move(entry));
+  }
+  if (sources.empty()) {
+    return std::string();
+  }
+  base::DictValue root;
+  root.Set("sources", std::move(sources));
   std::string serialized;
   base::JSONWriter::Write(root, &serialized);
   return serialized;
--- a/browser/ai_chat/tools/tab_semantic_search_tool.cc
+++ b/browser/ai_chat/tools/tab_semantic_search_tool.cc
@@ -1,0 +1,166 @@
+// Copyright (c) 2026 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#include "brave/browser/ai_chat/tools/tab_semantic_search_tool.h"
+
+#include <optional>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include "base/functional/bind.h"
+#include "base/json/json_reader.h"
+#include "base/json/json_writer.h"
+#include "base/values.h"
+#include "brave/browser/history_embeddings/open_tab_search.h"
+#include "brave/components/ai_chat/core/browser/tools/tool_input_properties.h"
+#include "brave/components/ai_chat/core/browser/tools/tool_utils.h"
+#include "brave/components/ai_chat/core/common/mojom/ai_chat.mojom.h"
+#include "brave/components/ai_chat/core/common/mojom/common.mojom.h"
+#include "chrome/browser/history/history_service_factory.h"
+#include "chrome/browser/history_embeddings/history_embeddings_service_factory.h"
+#include "chrome/browser/profiles/profile.h"
+#include "components/history/core/browser/history_service.h"
+#include "components/history_embeddings/content/history_embeddings_service.h"
+#include "components/history_embeddings/core/history_embeddings_search.h"
+#include "components/keyed_service/core/service_access_type.h"
+
+namespace ai_chat {
+
+namespace {
+
+std::string EmptyResultsJson() {
+  base::DictValue root;
+  root.Set("results", base::ListValue());
+  std::string serialized;
+  base::JSONWriter::Write(root, &serialized);
+  return serialized;
+}
+
+void OnRanked(Tool::UseToolCallback callback,
+              std::vector<history_embeddings::OpenTabInfo> tabs) {
+  if (tabs.empty()) {
+    std::move(callback).Run(CreateContentBlocksForText(EmptyResultsJson()), {});
+    return;
+  }
+  std::string serialized = internal::BuildSemanticSearchResultsJson(tabs);
+  std::move(callback).Run(CreateContentBlocksForText(serialized), {});
+}
+
+}  // namespace
+
+namespace internal {
+
+std::string BuildSemanticSearchResultsJson(
+    const std::vector<history_embeddings::OpenTabInfo>& tabs) {
+  base::ListValue results_list;
+  for (const auto& tab : tabs) {
+    base::DictValue entry;
+    entry.Set("tab_id", tab.tab_id);
+    entry.Set("title", tab.title);
+    entry.Set("url", tab.url.spec());
+    results_list.Append(std::move(entry));
+  }
+  base::DictValue root;
+  root.Set("results", std::move(results_list));
+  std::string serialized;
+  base::JSONWriter::Write(root, &serialized);
+  return serialized;
+}
+
+}  // namespace internal
+
+TabSemanticSearchTool::TabSemanticSearchTool(Profile* profile)
+    : profile_(profile) {}
+
+TabSemanticSearchTool::~TabSemanticSearchTool() = default;
+
+std::string_view TabSemanticSearchTool::Name() const {
+  return mojom::kSemanticTabSearchToolName;
+}
+
+std::string_view TabSemanticSearchTool::Description() const {
+  return "Semantically search the user's currently-open browser tabs by page "
+         "content. Use this when the user asks to find one of their open tabs "
+         "by what it contains (e.g. \"the tab about react hooks\"), not just "
+         "by title or URL. The query and full page content stay on device; "
+         "only matched tabs' titles and URLs are returned. Provide a "
+         "natural-language query.";
+}
+
+std::optional<base::DictValue> TabSemanticSearchTool::InputProperties() const {
+  return CreateInputProperties(
+      {{"query",
+        StringProperty(
+            "Natural-language query describing the tab to find by content.")}});
+}
+
+std::optional<std::vector<std::string>>
+TabSemanticSearchTool::RequiredProperties() const {
+  return std::vector<std::string>{"query"};
+}
+
+std::variant<bool, mojom::PermissionChallengePtr>
+TabSemanticSearchTool::RequiresUserInteractionBeforeHandling(
+    const mojom::ToolUseEvent& tool_use) const {
+  if (user_has_granted_permission_) {
+    return false;
+  }
+  // The search runs on-device against the local embeddings index over the
+  // user's open tabs; only matched tabs' title and URL are sent to the remote
+  // LLM as this tool's output. User-facing wording lives in
+  // `get_tool_permission_implications.tsx` so it goes through i18n; this
+  // C++ side only needs to surface a non-null challenge.
+  return mojom::PermissionChallenge::New(/*assessment=*/std::nullopt,
+                                         /*plan=*/std::nullopt);
+}
+
+void TabSemanticSearchTool::UserPermissionGranted(
+    const std::string& tool_use_id) {
+  user_has_granted_permission_ = true;
+}
+
+void TabSemanticSearchTool::UseTool(const std::string& input_json,
+                                    UseToolCallback callback) {
+  if (!user_has_granted_permission_) {
+    std::move(callback).Run(
+        CreateContentBlocksForText(
+            "Permission to search open tabs has not been granted."),
+        {});
+    return;
+  }
+  auto input = base::JSONReader::ReadDict(input_json,
+                                          base::JSON_PARSE_CHROMIUM_EXTENSIONS);
+  if (!input.has_value()) {
+    std::move(callback).Run(
+        CreateContentBlocksForText(
+            "Failed to parse input JSON. Provide a 'query' field."),
+        {});
+    return;
+  }
+  const auto* query = input->FindString("query");
+  if (!query || query->empty()) {
+    std::move(callback).Run(
+        CreateContentBlocksForText("Missing required 'query' field."), {});
+    return;
+  }
+
+  auto* embeddings_search =
+      embeddings_search_for_testing_
+          ? embeddings_search_for_testing_.get()
+          : HistoryEmbeddingsServiceFactory::GetForProfile(profile_);
+  auto* history_service = HistoryServiceFactory::GetForProfile(
+      profile_, ServiceAccessType::EXPLICIT_ACCESS);
+  if (!embeddings_search || !history_service) {
+    std::move(callback).Run(CreateContentBlocksForText(EmptyResultsJson()), {});
+    return;
+  }
+
+  history_embeddings::SearchOpenTabsByContent(
+      profile_, history_service, embeddings_search, *query,
+      base::BindOnce(&OnRanked, std::move(callback)), &query_url_task_tracker_);
+}
+
+}  // namespace ai_chat
--- a/browser/ai_chat/tools/tab_semantic_search_tool.cc
+++ b/browser/ai_chat/tools/tab_semantic_search_tool.cc
@@ -168,10 +168,9 @@ void TabSemanticSearchTool::UseTool(const std::string& input_json,
   auto* history_service = HistoryServiceFactory::GetForProfile(
       profile_, ServiceAccessType::EXPLICIT_ACCESS);
   if (!embeddings_search || !history_service) {
-    std::move(callback).Run(
-        CreateContentBlocksForText(
-            internal::BuildSemanticSearchResultsJson({})),
-        {});
+    std::move(callback).Run(CreateContentBlocksForText(
+                                internal::BuildSemanticSearchResultsJson({})),
+                            {});
     return;
   }
 
--- a/browser/ai_chat/tools/tab_semantic_search_tool.cc
+++ b/browser/ai_chat/tools/tab_semantic_search_tool.cc
@@ -7,6 +7,7 @@
 
 #include <optional>
 #include <string>
+#include <string_view>
 #include <utility>
 #include <vector>
 
@@ -31,6 +32,25 @@ namespace ai_chat {
 
 namespace {
 
+// Serializes `tabs` as a list of {tab_id, title, url} dicts under `key`.
+std::string SerializeTabsUnderKey(
+    std::string_view key,
+    const std::vector<history_embeddings::OpenTabInfo>& tabs) {
+  base::ListValue list;
+  for (const auto& tab : tabs) {
+    base::DictValue entry;
+    entry.Set("tab_id", tab.tab_id);
+    entry.Set("title", tab.title);
+    entry.Set("url", tab.url.spec());
+    list.Append(std::move(entry));
+  }
+  base::DictValue root;
+  root.Set(key, std::move(list));
+  std::string serialized;
+  base::JSONWriter::Write(root, &serialized);
+  return serialized;
+}
+
 void OnRanked(Tool::UseToolCallback callback,
               std::vector<history_embeddings::OpenTabInfo> tabs) {
   std::string serialized = internal::BuildSemanticSearchResultsJson(tabs);
@@ -52,19 +72,7 @@ namespace internal {
 
 std::string BuildSemanticSearchResultsJson(
     const std::vector<history_embeddings::OpenTabInfo>& tabs) {
-  base::ListValue results_list;
-  for (const auto& tab : tabs) {
-    base::DictValue entry;
-    entry.Set("tab_id", tab.tab_id);
-    entry.Set("title", tab.title);
-    entry.Set("url", tab.url.spec());
-    results_list.Append(std::move(entry));
-  }
-  base::DictValue root;
-  root.Set("results", std::move(results_list));
-  std::string serialized;
-  base::JSONWriter::Write(root, &serialized);
-  return serialized;
+  return SerializeTabsUnderKey("results", tabs);
 }
 
 std::string BuildSemanticSearchTabSourcesJson(
--- a/browser/ai_chat/tools/tab_semantic_search_tool.cc
+++ b/browser/ai_chat/tools/tab_semantic_search_tool.cc
@@ -106,9 +106,8 @@ std::string_view TabSemanticSearchTool::Description() const {
 
 std::optional<base::DictValue> TabSemanticSearchTool::InputProperties() const {
   return CreateInputProperties(
-      {{"query",
-        StringProperty(
-            "Natural-language query describing the tab to find by content.")}});
+      {{"query", StringProperty(
+                     "Natural-language query for finding a tab by content.")}});
 }
 
 std::optional<std::vector<std::string>>
--- a/browser/ai_chat/tools/tab_semantic_search_tool.cc
+++ b/browser/ai_chat/tools/tab_semantic_search_tool.cc
@@ -72,7 +72,7 @@ namespace internal {
 
 std::string BuildSemanticSearchResultsJson(
     const std::vector<history_embeddings::OpenTabInfo>& tabs) {
-  return SerializeTabsUnderKey("results", tabs);
+  return SerializeTabsUnderKey("tabs", tabs);
 }
 
 std::string BuildSemanticSearchTabSourcesJson(
--- a/browser/ai_chat/tools/tab_semantic_search_tool.cc
+++ b/browser/ai_chat/tools/tab_semantic_search_tool.cc
@@ -77,22 +77,11 @@ std::string BuildSemanticSearchResultsJson(
 
 std::string BuildSemanticSearchTabSourcesJson(
     const std::vector<history_embeddings::OpenTabInfo>& tabs) {
-  base::ListValue sources;
-  for (const auto& tab : tabs) {
-    base::DictValue entry;
-    entry.Set("tab_id", tab.tab_id);
-    entry.Set("title", tab.title);
-    entry.Set("url", tab.url.spec());
-    sources.Append(std::move(entry));
-  }
-  if (sources.empty()) {
+  // An empty payload signals `OnRanked` to skip the tab_sources artifact.
+  if (tabs.empty()) {
     return std::string();
   }
-  base::DictValue root;
-  root.Set("sources", std::move(sources));
-  std::string serialized;
-  base::JSONWriter::Write(root, &serialized);
-  return serialized;
+  return SerializeTabsUnderKey("sources", tabs);
 }
 
 }  // namespace internal
--- a/browser/ai_chat/tools/tab_semantic_search_tool.cc
+++ b/browser/ai_chat/tools/tab_semantic_search_tool.cc
@@ -31,20 +31,8 @@ namespace ai_chat {
 
 namespace {
 
-std::string EmptyResultsJson() {
-  base::DictValue root;
-  root.Set("results", base::ListValue());
-  std::string serialized;
-  base::JSONWriter::Write(root, &serialized);
-  return serialized;
-}
-
 void OnRanked(Tool::UseToolCallback callback,
               std::vector<history_embeddings::OpenTabInfo> tabs) {
-  if (tabs.empty()) {
-    std::move(callback).Run(CreateContentBlocksForText(EmptyResultsJson()), {});
-    return;
-  }
   std::string serialized = internal::BuildSemanticSearchResultsJson(tabs);
   std::vector<mojom::ToolArtifactPtr> artifacts;
   std::string sources_json = internal::BuildSemanticSearchTabSourcesJson(tabs);
@@ -183,7 +171,10 @@ void TabSemanticSearchTool::UseTool(const std::string& input_json,
   auto* history_service = HistoryServiceFactory::GetForProfile(
       profile_, ServiceAccessType::EXPLICIT_ACCESS);
   if (!embeddings_search || !history_service) {
-    std::move(callback).Run(CreateContentBlocksForText(EmptyResultsJson()), {});
+    std::move(callback).Run(
+        CreateContentBlocksForText(
+            internal::BuildSemanticSearchResultsJson({})),
+        {});
     return;
   }
 
--- a/browser/ai_chat/tools/tab_semantic_search_tool_unittest.cc
+++ b/browser/ai_chat/tools/tab_semantic_search_tool_unittest.cc
@@ -149,4 +149,29 @@ TEST(TabSemanticSearchToolBuildResultsJsonTest, EmptyTabsEmitsEmptyResults) {
       base::test::ParseJsonDict(R"({"results":[]})"));
 }
 
+TEST(TabSemanticSearchToolBuildTabSourcesJsonTest, EmitsRankedTabSources) {
+  const std::vector<history_embeddings::OpenTabInfo> tabs = {
+      {/*tab_id=*/22, "Tab Two", GURL("https://two.example/")},
+      {/*tab_id=*/11, "Tab One", GURL("https://one.example/")},
+  };
+
+  // tab_id is emitted as an int32 so the frontend can hand it straight to
+  // `SwitchToTab(int32)` without re-parsing.
+  EXPECT_EQ(base::test::ParseJsonDict(
+                internal::BuildSemanticSearchTabSourcesJson(tabs)),
+            base::test::ParseJsonDict(R"({
+              "sources": [
+                {"tab_id": 22, "title": "Tab Two",
+                 "url": "https://two.example/"},
+                {"tab_id": 11, "title": "Tab One",
+                 "url": "https://one.example/"}
+              ]
+            })"));
+}
+
+TEST(TabSemanticSearchToolBuildTabSourcesJsonTest, EmptyTabsEmitsEmptyString) {
+  // Empty payload signals `OnRanked` to skip emitting a tab_sources artifact.
+  EXPECT_TRUE(internal::BuildSemanticSearchTabSourcesJson({}).empty());
+}
+
 }  // namespace ai_chat
--- a/browser/ai_chat/tools/tab_semantic_search_tool_unittest.cc
+++ b/browser/ai_chat/tools/tab_semantic_search_tool_unittest.cc
@@ -124,7 +124,7 @@ TEST_F(TabSemanticSearchToolTest, NoOpenTabsReturnsEmptyResults) {
   // no browser windows, exercising the early-return JSON path.
   std::string output = RunTool(tool_.get(), kValidInput);
   EXPECT_EQ(base::test::ParseJsonDict(output),
-            base::test::ParseJsonDict(R"({"results":[]})"));
+            base::test::ParseJsonDict(R"({"tabs":[]})"));
 }
 
 TEST(TabSemanticSearchToolBuildResultsJsonTest, EmitsRankedTabs) {
@@ -136,7 +136,7 @@ TEST(TabSemanticSearchToolBuildResultsJsonTest, EmitsRankedTabs) {
   EXPECT_EQ(
       base::test::ParseJsonDict(internal::BuildSemanticSearchResultsJson(tabs)),
       base::test::ParseJsonDict(R"({
-        "results": [
+        "tabs": [
           {"tab_id": 22, "title": "Tab Two", "url": "https://two.example/"},
           {"tab_id": 11, "title": "Tab One", "url": "https://one.example/"}
         ]
@@ -146,7 +146,7 @@ TEST(TabSemanticSearchToolBuildResultsJsonTest, EmitsRankedTabs) {
 TEST(TabSemanticSearchToolBuildResultsJsonTest, EmptyTabsEmitsEmptyResults) {
   EXPECT_EQ(
       base::test::ParseJsonDict(internal::BuildSemanticSearchResultsJson({})),
-      base::test::ParseJsonDict(R"({"results":[]})"));
+      base::test::ParseJsonDict(R"({"tabs":[]})"));
 }
 
 TEST(TabSemanticSearchToolBuildTabSourcesJsonTest, EmitsRankedTabSources) {
--- a/browser/ai_chat/tools/tab_semantic_search_tool_unittest.cc
+++ b/browser/ai_chat/tools/tab_semantic_search_tool_unittest.cc
@@ -1,0 +1,152 @@
+// Copyright (c) 2026 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#include "brave/browser/ai_chat/tools/tab_semantic_search_tool.h"
+
+#include <memory>
+#include <optional>
+#include <string>
+#include <variant>
+#include <vector>
+
+#include "base/test/test_future.h"
+#include "base/test/values_test_util.h"
+#include "brave/browser/history_embeddings/open_tab_search.h"
+#include "brave/components/ai_chat/core/common/mojom/ai_chat.mojom.h"
+#include "brave/components/ai_chat/core/common/mojom/common.mojom.h"
+#include "chrome/test/base/testing_profile.h"
+#include "content/public/test/browser_task_environment.h"
+#include "testing/gmock/include/gmock/gmock.h"
+#include "testing/gtest/include/gtest/gtest.h"
+#include "url/gurl.h"
+
+namespace ai_chat {
+
+namespace {
+
+constexpr char kValidInput[] = R"({"query":"react hooks"})";
+
+std::string ExtractText(const std::vector<mojom::ContentBlockPtr>& blocks) {
+  if (blocks.empty() || !blocks[0]->is_text_content_block()) {
+    return std::string();
+  }
+  return blocks[0]->get_text_content_block()->text;
+}
+
+std::string RunTool(TabSemanticSearchTool* tool, const std::string& json) {
+  base::test::TestFuture<std::vector<mojom::ContentBlockPtr>,
+                         std::vector<mojom::ToolArtifactPtr>>
+      future;
+  tool->UseTool(json, future.GetCallback());
+  return ExtractText(future.Get<std::vector<mojom::ContentBlockPtr>>());
+}
+
+mojom::ToolUseEventPtr CreateToolUseEvent(const std::string& json) {
+  return mojom::ToolUseEvent::New(mojom::kSemanticTabSearchToolName, "1", json,
+                                  std::nullopt, std::nullopt, nullptr, false);
+}
+
+}  // namespace
+
+class TabSemanticSearchToolTest : public ::testing::Test {
+ public:
+  TabSemanticSearchToolTest() = default;
+  ~TabSemanticSearchToolTest() override = default;
+
+  void SetUp() override {
+    profile_ = std::make_unique<TestingProfile>();
+    tool_ = std::make_unique<TabSemanticSearchTool>(profile_.get());
+  }
+
+  void TearDown() override {
+    tool_.reset();
+    profile_.reset();
+  }
+
+ protected:
+  content::BrowserTaskEnvironment task_environment_;
+  std::unique_ptr<TestingProfile> profile_;
+  std::unique_ptr<TabSemanticSearchTool> tool_;
+};
+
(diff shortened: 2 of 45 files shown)


components/ai_chat/core/browser/tab_tracker_service.h
```diff
@@ -29,6 +30,19 @@ class TabTrackerService : public KeyedService, public mojom::TabTrackerService {
...
   void Bind(mojo::PendingReceiver<mojom::TabTrackerService> receiver);
 
+  // Callback that activates the tab matching `tab_id`. Implementations live
+  // in code with chrome/browser/ui access (e.g. iterating BrowserList).
+  // Returns true if a tab was activated.
+  using ActivateTabCallback = base::RepeatingCallback<bool(int32_t tab_id)>;
+
+  // Sets the bridge used by ActivateTab(). Called once at startup by the code
+  // owning the UI-side implementation.
+  void SetActivator(ActivateTabCallback activator);
```
netzenbot: TabTrackerService is a component-level service and this wires platform-specific tab activation in through a raw `base::RepeatingCallback` (`SetActivator`/`activator_`). Per CSM-024 a small delegate interface implemented by the UI layer would give cleaner lifetime and testability semantics than the raw-callback bridge. Not blocking, but worth considering. ([Use Delegates Instead of Raw Callbacks for Cross-Layer Dependencies](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-memory.md#CSM-024))

browser/ui/ai_chat/tab_activator_factory.h
```diff
@@ -0,0 +1,44 @@
...
+
+namespace content {
+class BrowserContext;
+}
+
+namespace ai_chat {
+
+class TabActivator;
+
+class TabActivatorFactory : public BrowserContextKeyedServiceFactory {
```
netzenbot: nit: New desktop factories should inherit from `ProfileKeyedServiceFactory` (with `ProfileSelections`) rather than `BrowserContextKeyedServiceFactory`; that also removes the need to hand-roll the regular-profile check in `GetBrowserContextToUse`. (Sibling ai_chat factories predate this guidance.) ([Use `ProfileKeyedServiceFactory` for New Desktop Factories](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-029))

browser/history_embeddings/open_tab_search.cc
```diff
@@ -75,33 +70,40 @@ std::vector<OpenTabInfo> SnapshotOpenTabs(Profile* profile) {
...
     SearchResult result) {
-  std::vector<int32_t> tab_ids;
+  std::vector<OpenTabInfo> ranked;
   for (const auto& row : result.scored_url_rows) {
-    if (auto* tab_id =
-            base::FindOrNull(tab_id_by_url_id, row.scored_url.url_id)) {
-      tab_ids.push_back(*tab_id);
+    if (auto* matched =
+            base::FindOrNull(tabs_by_url_id, row.scored_url.url_id)) {
+      ranked.insert(ranked.end(), matched->begin(), matched->end());
```
netzenbot: nit: `ranked.insert(ranked.end(), matched->begin(), matched->end())` can be `base::Extend(ranked, *matched)` (base/containers/extend.h) per the container guidelines. ([Use `base::Extend` for Appending Ranges to Vectors](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-050))

browser/ai_chat/tools/tab_semantic_search_tool_unittest.cc
```diff
@@ -0,0 +1,185 @@
...
+
+TEST(TabSemanticSearchToolBuildResultsJsonTest, EmitsRankedTabs) {
+  const std::vector<history_embeddings::OpenTabInfo> tabs = {
+      {/*tab_id=*/22, "Tab Two", GURL("https://two.example/")},
+      {/*tab_id=*/11, "Tab One", GURL("https://one.example/")},
+  };
+
+  std::string json = internal::BuildSemanticSearchResultsJson(tabs);
+  auto parsed = base::JSONReader::ReadDict(json, base::JSON_PARSE_RFC);
+  ASSERT_TRUE(parsed.has_value());
```
netzenbot: nit: consider `base::test::ParseJsonDict()` plus the `base::test::*` matchers from base/test/values_test_util.h instead of manual `ReadDict` + `FindList`/`FindInt` — clearer assertions and better failure messages. ([Use `base::test::ParseJson` and `base::ExpectDict*` in Tests](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-051))

browser/ai_chat/tools/tab_semantic_search_tool_unittest.cc
```diff
@@ -0,0 +1,185 @@
...
+  EXPECT_THAT(RunTool(tool_.get(), R"({"query":""})"),
+              testing::HasSubstr("Missing required 'query' field"));
+}
+
+TEST_F(TabSemanticSearchToolTest, NoOpenTabsReturnsEmptyResults) {
+  tool_->UserPermissionGranted("1");
+  // `SearchOpenTabsByContent()` returns no tabs for a `TestingProfile` with
+  // no browser windows, exercising the early-return JSON path.
+  std::string output = RunTool(tool_.get(), kValidInput);
+  EXPECT_THAT(output, testing::HasSubstr("\"results\":[]"));
```
netzenbot: nit: this validates JSON output via substring matching. Per TI-020, prefer parsing (e.g. base::test::ParseJsonDict(output) compared against the expected dict) so the structure is checked, matching how the Build*JsonTest cases in this file already use JSONReader::ReadDict. ([Don't use testing::HasSubstr to validate JSON](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-020))

browser/ui/ai_chat/tab_activator.h
```diff
@@ -0,0 +1,43 @@
...
+#include "components/keyed_service/core/keyed_service.h"
+
+class Profile;
+
+namespace ai_chat {
+
+class TabTrackerService;
+
+// Installs an activator callback into the profile's TabTrackerService so that
+// Leo can switch the user to a tab by id. The activator iterates the browsers
```
netzenbot: nit: This comment says the activator 'iterates the browsers belonging to `profile`', but `ActivateTab()` in the .cc resolves the tab via `tabs::TabHandle(tab_id).Get()` and explicitly notes it does so 'rather than iterating BrowserList'. Update the comment to match the actual global-handle lookup. ([Delete Dead Documentation](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-002))

components/ai_chat/core/browser/tab_tracker_service.h
darkdh: fixed in 7c3ba26e8a0

browser/ui/ai_chat/tab_activator_factory.h
darkdh: fixed in a80fca73835

browser/history_embeddings/open_tab_search.cc
darkdh: fixed in ca6c2569e30

browser/ai_chat/tools/tab_semantic_search_tool_unittest.cc
darkdh: fixed in deff65b2973

browser/ai_chat/tools/tab_semantic_search_tool_unittest.cc
darkdh: fixed in 092d9cb2e57

browser/ui/ai_chat/tab_activator.h
darkdh: fixed in 65735a63f1e

browser/ui/ai_chat/tab_activator.cc
```diff
@@ -0,0 +1,59 @@
...
+  }
+}
+
+TabActivator::~TabActivator() {
+  // KeyedService destruction ordering guarantees `tab_tracker_` outlives us
+  // (we depend on it), so clear ourselves as its delegate.
+  if (tab_tracker_) {
+    tab_tracker_->SetDelegate(nullptr);
+  }
+}
```
fallaciousreasoning: I wonder if we should have a virtual implementation of this in TabTrackerServer which handles calling `SetDelegate`.

optional, as this is probably going to be the only implementation.

browser/ui/ai_chat/tab_activator.cc
```diff
@@ -0,0 +1,59 @@
...
+
+bool TabActivator::ActivateTab(int32_t tab_id) {
+  // The ids stored in TabTrackerService come from `tabs::TabInterface::Handle`
+  // (see TabDataWebContentsObserver), so we resolve them via the handle's
+  // global lookup rather than iterating BrowserList by SessionID.
+  tabs::TabInterface* tab = tabs::TabHandle(tab_id).Get();
```
fallaciousreasoning: completely separate, as no super relevant to this PR, but TabTrackerService came about because upstream hadn't got BrowserWindowInterface working at the time we needed it. If we were to implement now, we'd probably use that to iterate the lsit of tabs.

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/inline_tab_search.module.scss
```diff
@@ -0,0 +1,62 @@
...
+    display: flex;
+    align-items: center;
+    gap: var(--leo-spacing-m);
+    width: 100%;
+    padding: var(--leo-spacing-m);
+    border: 1px solid var(--leo-color-divider-subtle);
+    border-radius: var(--leo-radius-m);
+    background: transparent;
+    cursor: pointer;
+    text-align: left;
```
fallaciousreasoning, line 28: do we need to specify text-align here?

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/tab_sources_event.module.scss
```diff
@@ -0,0 +1,62 @@
...
+    align-items: center;
+    gap: var(--leo-spacing-m);
+    width: 100%;
+    padding: var(--leo-spacing-m);
+    border: 1px solid var(--leo-color-divider-subtle);
+    border-radius: var(--leo-radius-m);
+    background: transparent;
+    cursor: pointer;
+    text-align: left;
+    color: inherit;
```
fallaciousreasoning: ditto, isn't this implied?

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/inline_tab_search.module.scss
```diff
@@ -0,0 +1,62 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+.tabSources {
+  margin-top: var(--leo-spacing-m);
+
+  ul {
+    list-style: none;
+    margin: 0;
+    padding: 0;
```
fallaciousreasoning, line 12: nit: unsure, but is this the default (it might not be)

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/tab_sources_event.tsx
```diff
@@ -0,0 +1,105 @@
...
+    try {
+      return new URL(source.url).hostname
+    } catch {
+      return source.url
+    }
+  })()
+
+  const faviconSrc =
+    `chrome-untrusted://favicon2?size=32&pageUrl=`
+    + encodeURIComponent(source.url)
```
fallaciousreasoning: man, we should really have a component for this - I think we have four or five places we do this in AI Chat.

not for this PR though :smile:

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/tab_sources_event.tsx
```diff
@@ -0,0 +1,105 @@
...
+  const result: TabSource[] = []
+  for (const artifact of artifacts) {
+    if (artifact.type !== Mojom.TAB_SOURCES_ARTIFACT_TYPE) {
+      continue
+    }
+    try {
+      const parsed = JSON.parse(artifact.contentJson) as {
+        sources?: TabSource[]
+      }
+      if (Array.isArray(parsed.sources)) {
```
fallaciousreasoning: I think there's a chance JSON.parse returns something falsey so might be worth a 
```suggestion
      const parsed = JSON.parse(artifact.contentJson) as {
        sources?: TabSource[]
      } | undefined
      if (Array.isArray(parsed?.sources)) {
```

browser/ai_chat/tools/tab_semantic_search_tool.cc
```diff
@@ -0,0 +1,195 @@
...
+
+namespace {
+
+std::string EmptyResultsJson() {
+  base::DictValue root;
+  root.Set("results", base::ListValue());
+  std::string serialized;
+  base::JSONWriter::Write(root, &serialized);
+  return serialized;
+}
```
fallaciousreasoning: could we just hardcode a string for this?

```
constexpr kEmptyResultsJson = "{ \"results\": [] }";

File: browser/ai_chat/tools/tab_semantic_search_tool.cc [hunk 15]
```diff
@@ -0,0 +1,195 @@
...
+std::string BuildSemanticSearchTabSourcesJson(
+    const std::vector<history_embeddings::OpenTabInfo>& tabs) {
+  base::ListValue sources;
+  for (const auto& tab : tabs) {
+    base::DictValue entry;
+    entry.Set("tab_id", tab.tab_id);
+    entry.Set("title", tab.title);
+    entry.Set("url", tab.url.spec());
+    sources.Append(std::move(entry));
+  }
```
fallaciousreasoning: maybe worth sharing the code for building a ListValue of tabs, rather than duplicating it in case we add some new properties

browser/ai_chat/tools/tab_semantic_search_tool.cc
```diff
@@ -0,0 +1,195 @@
...
+    results_list.Append(std::move(entry));
+  }
+  base::DictValue root;
+  root.Set("results", std::move(results_list));
+  std::string serialized;
+  base::JSONWriter::Write(root, &serialized);
+  return serialized;
+}
+
+std::string BuildSemanticSearchTabSourcesJson(
```
fallaciousreasoning: is there a reason we need these two separate shapes? Could we just have a 

`{ "tabs": [ ... ] }` 

for both cases?

browser/ai_chat/tools/tab_semantic_search_tool.cc
```diff
@@ -0,0 +1,195 @@
...
+         "by title or URL. The query and full page content stay on device; "
+         "only matched tabs' titles and URLs are returned. Provide a "
+         "natural-language query.";
+}
+
+std::optional<base::DictValue> TabSemanticSearchTool::InputProperties() const {
+  return CreateInputProperties(
+      {{"query",
+        StringProperty(
+            "Natural-language query describing the tab to find by content.")}});
```
fallaciousreasoning:
```suggestion
            "Natural-language query for finding a tab by content.")}});
```

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/tab_sources_event.tsx
fallaciousreasoning, line 1: probably worth adding some jest unit tests for this, and potentially updating a Storybook so there's a way to see this component in action.

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/assistant_task.tsx
fallaciousreasoning, line 1: could we add a simple jest test for this too, just that the TabSourcesEvent shows up when present.

browser/ui/ai_chat/tab_activator.cc
darkdh: Current behavior:

```cpp
TabActivator::TabActivator(Profile* profile, TabTrackerService* tab_tracker)
    : profile_(profile), tab_tracker_(tab_tracker) {
  if (tab_tracker_) {
    tab_tracker_->SetDelegate(this);
  }
}

TabActivator::~TabActivator() {
  if (tab_tracker_) {
    tab_tracker_->SetDelegate(nullptr);
  }
}
```

With registration owned by `TabTrackerService`:

```cpp
TabActivator::TabActivator(Profile* profile, TabTrackerService* tab_tracker)
    : profile_(profile), scoped_delegate_(tab_tracker, this) {}
```

No destructor, and `tab_tracker_` goes away — it was only used for registration. Members become:

```cpp
  raw_ptr<Profile> profile_;
  // Declared last: registration happens after `profile_` is initialized.
  TabTrackerService::ScopedDelegate scoped_delegate_;
```

This needs a new helper on `TabTrackerService`:

```cpp
  class ScopedDelegate {
   public:
    ScopedDelegate(TabTrackerService* service, Delegate* delegate)
        : service_(service) {
      if (service_) {
        service_->SetDelegate(delegate);
      }
    }
    ~ScopedDelegate() {
      if (service_) {
        service_->SetDelegate(nullptr);
      }
    }

   private:
    raw_ptr<TabTrackerService> service_;
  };
```

A scoped member rather than `Delegate` registering itself in its own constructor: a base-class constructor would hand `this` to the service before the derived vptr and members are set up. Same shape as `base::ScopedObservation`.

Let me know if this is what you have in mind.

browser/ui/ai_chat/tab_activator.cc
darkdh: We already use `BrowserWindowInterface` for tabs iteration in `//brave/browser/history_embeddings/open_tab_search.cc`
```cpp
  std::vector<OpenTabInfo> SnapshotOpenTabs(Profile* profile) {
    std::vector<OpenTabInfo> tabs;
    GlobalBrowserCollection::GetInstance()->ForEach(
        [profile, &tabs](BrowserWindowInterface* browser_window) {
...
}
```

This seems more like migration `TabTrackerService` to `BrowserWindowInterface` which should not be in the scope of this PR.

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/tab_sources_event.module.scss
darkdh: yes, `reset.css` already set it on `button`. Will remove it.

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/inline_tab_search.module.scss
darkdh: Not the default — the UA stylesheet gives `ul` `padding-inline-start: 40px`, and nothing in the frame's cascade resets it.

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/inline_tab_search.module.scss
darkdh: Yes — the UA sets `text-align: center` on `button`, so without this the host line renders centered.

browser/ui/ai_chat/tab_activator.cc
fallaciousreasoning: Wasn't saying for this PR, just noting that we should get rid of it at some point :smile:

browser/ui/ai_chat/tab_activator.cc
fallaciousreasoning: I was thinking we could make the constructor/destructor virtual on TabTrackerService::Delegate and make it take a `TabTrackerService*` and then call `SetDelegate` in its ctor/dtor so any future implementations would handle setting/clearing the delegate.

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/tab_sources_event.module.scss
darkdh: fixed with 2595b0a6a7f

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/tab_sources_event.tsx
darkdh: fixed with a4f5163ca55

browser/ai_chat/tools/tab_semantic_search_tool.cc
darkdh: fixed with d2b86d99149 and 03cffec5931

browser/ai_chat/tools/tab_semantic_search_tool.cc
darkdh: Dropped the helper instead — `BuildSemanticSearchResultsJson({})` already emits the empty payload, so the `tabs.empty()` early return isn't needed either. Fixed with 2bffad6daf8.

browser/ai_chat/tools/tab_semantic_search_tool.cc
darkdh: Two payloads rather than two shapes. The tool-result JSON is what the model reads; the other is a `ToolArtifact` the frontend parses to call `switchToTab`, and it's persisted into stored conversations. They also differ when empty — the artifact is omitted rather than emitted empty.

Renamed the result key to `tabs` in 9d0021ad355. Left the artifact as `sources` since it's persisted.

browser/ui/ai_chat/tab_activator.cc
darkdh: Registering from the base ctor publishes `this` before the object is complete. Construction runs base-first:

1. `Delegate(service)` → `SetDelegate(this)`
2. `TabActivator`'s members initialized (`profile_`, …)
3. `TabActivator`'s ctor body

After step 1 the service can already reach `delegate_->ActivateTab()`, but the vtable still points at `Delegate` (where it's pure virtual) and `profile_` is uninitialized. Destruction mirrors it: `~TabActivator` runs first, `~Delegate` unregisters last, leaving the service pointing at an object whose `TabActivator` half is gone.

Both are lifetime landmines — the service holds a pointer to something that isn't, or is no longer, a `TabActivator`. Whether a caller happens to hit those windows today isn't something to depend on, and the current dtor deliberately clears the delegate as its first statement, which the base-class version gives up.

(Also ctors can't be virtual — only the dtor, which it already is.)

A member achieves the same goal — a future implementation still can't forget to set/clear it — and is initialized after the object is fully itself:

```cpp
// TabActivator, declared last
TabTrackerService::ScopedDelegate scoped_delegate_{tab_tracker, this};
```

Want me to do that instead?

fallaciousreasoning: nah, lets leave it as is.

netzenbot (approved):

browser/ai_chat/tools/tab_semantic_search_tool.cc
darkdh: fixed with 31eaea6a234

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/tab_sources_event.tsx
darkdh: fixed with 984600feeee — jest tests plus a Storybook story

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_task/assistant_task.tsx
darkdh: fixed with da3b6762cb5

fallaciousreasoning (approved):

components/resources/ai_chat_ui_strings.grdp
```diff
@@ -936,6 +936,9 @@
   <message name="IDS_CHAT_UI_TOOL_SEMANTIC_HISTORY_SEARCH_PERMISSION_IMPLICATIONS" desc="Explanation shown in the permission prompt when the assistant wants to search the user's local browsing history. Calls out that the search itself is on-device and only the matched pages flow to Brave AI." formatter_data="webui=AiChat">
     Brave AI wants to search your browsing history. The search runs locally on your device, so your full history isn't sent anywhere. Only the matching pages' titles, URLs, and indexed text passages will be sent to Brave AI as part of this conversation.
   </message>
+  <message name="IDS_CHAT_UI_TOOL_SEMANTIC_TAB_SEARCH_PERMISSION_IMPLICATIONS" desc="Explanation shown in the permission prompt when the assistant wants to semantically search the user's currently-open tabs. Calls out that the search itself is on-device and only the matched tabs flow to Brave AI." formatter_data="webui=AiChat">
+    Brave AI wants to semantically search your currently-open tabs. The search query and the full content of your tabs stay on this device. Only the matched tabs' titles and URLs will be sent to Brave AI as part of this conversation.
+  </message>
```
petemill: Do we need the word "semantically" here? I think it's a detail that can confuse and this has the same meaning if it's removed.

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/tab_sources_event.tsx
```diff
@@ -0,0 +1,107 @@
...
+import * as React from 'react'
+import * as Mojom from '../../../common/mojom'
+import { useUntrustedConversationContext } from '../../untrusted_conversation_context'
+import styles from './tab_sources_event.module.scss'
+
+interface TabSource {
+  tab_id: number
+  title: string
+  url: string
+}
```
petemill: nit: consider making this a mojom type so that the C++ and the TS will have compile errors if it changes. We could also consider `ToolArtifact` has a union `content` property between the existing chart json string and `TabSource`.

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/tab_sources_event.tsx
```diff
@@ -0,0 +1,107 @@
...
+          alt=''
+        />
+        <span className={styles.text}>
+          <span className={styles.title}>{source.title || host}</span>
+          <span className={styles.host}>{host}</span>
+        </span>
+      </button>
+    </li>
+  )
+}
```
petemill: We need to support what to do for this result on the shared conversation website. We won't have favicon, though we can inline as data uri when the shared payload is created), or ability to click to the Tab, of course. At the moment this is going to cause a call to `chrome://favicon2` on an `https` URL which will fail. 
Maybe `switchToTab` could pass the whole `source` object and the remote could decide to switch to the Tab _or_ open a new Tab if the tab doesn't exist or if the platform doesn't support tab switching?

This is another reason to have TabSource be typed at the mojom level so that `conversationHandler` can do more with the object, and so [`createSharedConversationPayload`](https://github.com/brave/brave-core/blob/11f61f31e14a19db3591511dc287baa59afe516f/components/ai_chat/resources/page/components/share_conversation_modal/create_shared_conversation_payload.ts#L59-L89) can add favicon data to it like it currently does for attached Tabs.

Alternatively for the short term we can just render `null` in `TabSourcesEvent` if `conversationContext.isReadOnly` and address the above as follow-up.

petemill (commented):
Did you find it's useful to have this as a Tool where we don't send much information to the LLM? It seems like the main purpose is to show the user the results and not the LLM? In your screenshot example in the description it looks like the LLM is just repeating the information. It's also asking if the user wants to summarize the article, which it won't be able to do because we didn't provide the context, right? It might make things up.
Would this be better as a markdown inline tool? Like we offer with inline web search? So the LLM decides to put tab search results somewhere in it's answer, and we immediately render the tab results in that place, not sending them to the LLM as a tool result.

If we're going to keep it as a response Tool then I would suggest we at least need to not render it on the http website, and then the follow-up if we want to render the result there.

But I would also suggest we need to change the Tool description to let the LLM know that we will display the results for the user so it doesn't need to repeat. If that's the only purpose of the Tool though, then we can save the additional tokens and back-and-forth by making it an inline markdown thing.

darkdh:
> Did you find it's useful to have this as a Tool where we don't send much information to the LLM? It seems like the main purpose is to show the user the results and not the LLM? In your screenshot example in the description it looks like the LLM is just repeating the information. It's also asking if the user wants to summarize the article, which it won't be able to do because we didn't provide the context, right? It might make things up. Would this be better as a markdown inline tool? Like we offer with inline web search? So the LLM decides to put tab search results somewhere in it's answer, and we immediately render the tab results in that place, not sending them to the LLM as a tool result.
> 
> If we're going to keep it as a response Tool then I would suggest we at least need to not render it on the http website, and then the follow-up if we want to render the result there.
> 
> But I would also suggest we need to change the Tool description to let the LLM know that we will display the results for the user so it doesn't need to repeat. If that's the only purpose of the Tool though, then we can save the additional tokens and back-and-forth by making it an inline markdown thing.

Yes, the purpose of the current tool is to provide a different entry of tab semantic search in Leo like tab search UI here. Clicking on the result will lead users to the tab
Screenshot 2026-08-11 at 10 42 50

And you are right, in this tool implementation LLM job is only to call the `SearchOpenTabsByContent()` and then relay the info to users. Markdown directive makes sense but it is only for flow that involves server like inline web search. 
But it only support server flow, for client only flow we probably need 
1. Makes LLM writes `:tabs[...]` in its answer to trigger the markdown directive
2. a new client-side directive parsing for detecting ::tabs[...] in the completion stream
3. consent? not sure if we really need it or not since it is performing tab search and we don't need users consent for performing web search
4. new event and similar card rendering as search

I currently have no idea how to make it possible without a separate tool calling.

petemill:
> And you are right, in this tool implementation LLM job is only to call the `SearchOpenTabsByContent()` and then relay the info to users. Markdown directive makes sense but it is only for flow that involves server like inline web search. But it only support server flow, for client only flow we probably need
> 
> 1. Makes LLM writes `:tabs[...]` in its answer to trigger the markdown directive
> 2. a new client-side directive parsing for detecting ::tabs[...] in the completion stream
> 3. consent? not sure if we really need it or not since it is performing tab search and we don't need users consent for performing web search
> 4. new event and similar card rendering as search
> 
> I currently have no idea how to make it possible without a separate tool calling.

I don't think we would need a new event or to involve the server at all. The inline-search markdown directive only uses an `inlineSearchEvent` to get the data from the Search API because we can't call it from the client. 
For this you can perform the search and get the results from the browser via mojom. I think @fallaciousreasoning has made this pretty simple to implement:

- Create a React component to perform tab search and display results, e.g. 
```tsx
function InlineTabSearch(props: { query: string }) {
  const { isPending: isSearching, data: tabResults } = untrustedConversationContext.api.useSearchForTabs(props.query)
  
  if (isSearching) {
    return (
      <p>Searching Tabs for "{props.query}"</p>
    }
  }
  
  if (!tabResults?.length) {
     return (
       <p>No Tabs found for "{props.query}"</p>
     )
   }

  return (
    <div
        className={styles.tabSources}
        data-testid='tab-sources-event'
      >
        <ul>
          {tabResults.map((source) => (
            <TabSourceCard
              key={source.tab_id}
              source={source}
            />
          ))}
        </ul>
      </div>    
  )
}
```

and in `components/ai_chat/resources/untrusted_conversation_frame/components/markdown_renderer/remark_directives.tsx` we just need to add a new entry to pass the props to that component, e.g.

```tsx
export const ALLOWED_DIRECTIVES = ['search', 'tabSearch']
[...]

export const directiveComponents = {
  search: ...,
  tabSearch: function (props: any) {
    // Maybe validate props.query exists first
    return <InlineTabSearch query={props.query} />
  }
```

And then for the tab search in the browser, implement a `UntrustedUIHandler::SearchForTabs` method or similar.

We would need to change the system prompt to tell the LLM to use the `tabSearch` markdown directive when wishing to show the users some search results for those Tabs, like we do for inline web search.

So if you don't want or need the LLM to see these results, then this is what I'd suggest to try.

darkdh:
> > And you are right, in this tool implementation LLM job is only to call the `SearchOpenTabsByContent()` and then relay the info to users. Markdown directive makes sense but it is only for flow that involves server like inline web search. But it only support server flow, for client only flow we probably need
> > 
> > 1. Makes LLM writes `:tabs[...]` in its answer to trigger the markdown directive
> > 2. a new client-side directive parsing for detecting ::tabs[...] in the completion stream
> > 3. consent? not sure if we really need it or not since it is performing tab search and we don't need users consent for performing web search
> > 4. new event and similar card rendering as search
> > 
> > I currently have no idea how to make it possible without a separate tool calling.
> 
> I don't think we would need a new event or to involve the server at all. The inline-search markdown directive only uses an `inlineSearchEvent` to get the data from the Search API because we can't call it from the client. For this you can perform the search and get the results from the browser via mojom. I think @fallaciousreasoning has made this pretty simple to implement:
> 
> * Create a React component to perform tab search and display results, e.g.
> 
> ```tsx
> function InlineTabSearch(props: { query: string }) {
>   const { isPending: isSearching, data: tabResults } = untrustedConversationContext.api.useSearchForTabs(props.query)
>   
>   if (isSearching) {
>     return (
>       <p>Searching Tabs for "{props.query}"</p>
>     }
>   }
>   
>   if (!tabResults?.length) {
>      return (
>        <p>No Tabs found for "{props.query}"</p>
>      )
>    }
> 
>   return (
>     <div
>         className={styles.tabSources}
>         data-testid='tab-sources-event'
>       >
>         <ul>
>           {tabResults.map((source) => (
>             <TabSourceCard
>               key={source.tab_id}
>               source={source}
>             />
>           ))}
>         </ul>
>       </div>    
>   )
> }
> ```
> 
> and in `components/ai_chat/resources/untrusted_conversation_frame/components/markdown_renderer/remark_directives.tsx` we just need to add a new entry to pass the props to that component, e.g.
> 
> ```tsx
> export const ALLOWED_DIRECTIVES = ['search', 'tabSearch']
> [...]
> 
> export const directiveComponents = {
>   search: ...,
>   tabSearch: function (props: any) {
>     // Maybe validate props.query exists first
>     return <InlineTabSearch query={props.query} />
>   }
> ```
> 
> And then for the tab search in the browser, implement a `UntrustedUIHandler::SearchForTabs` method or similar.
> 
> We would need to change the system prompt to tell the LLM to use the `tabSearch` markdown directive when wishing to show the users some search results for those Tabs, like we do for inline web search.
> 
> So if you don't want or need the LLM to see these results, then this is what I'd suggest to try.

It is neater than tool for this purpose and that works well with the [system prompt change](https://github.com/brave/aichat/pull/1741)
Screenshot 2026-08-11 at 15 06 51

I will push changes after cleaning up

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/tab_sources_event.tsx
darkdh: This is gone with markdown directive you suggested.

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/tab_sources_event.tsx
darkdh: This is gone with markdown directive you suggested

components/resources/ai_chat_ui_strings.grdp
darkdh: This is gone with markdown directive you suggested

components/ai_chat/core/common/mojom/untrusted_frame.mojom
```diff
@@ -11,6 +11,17 @@ import "url/mojom/url.mojom";
 // Interfaces for communication between the untrusted content frame and both
 // the Browser and the parent trusted UI frame.
 
+// An open tab matching a `tabSearch` directive's query.
+struct TabSearchResult {
```
fallaciousreasoning: rather than adding this, could we reuse `TabData` from `tab_tracker.mojom`?

components/ai_chat/core/browser/tab_tracker_service.cc
```diff
@@ -47,6 +47,14 @@ void TabTrackerService::UpdateTab(int32_t tab_id, mojom::TabDataPtr tab) {
...
 }
 
+void TabTrackerService::SetDelegate(Delegate* delegate) {
+  delegate_ = delegate;
+}
+
+bool TabTrackerService::ActivateTab(int32_t tab_id) {
+  return delegate_ && delegate_->ActivateTab(tab_id);
+}
+
```
fallaciousreasoning: Just looking at this again, is there any point this being on the TabTrackerService? This could just live in the browser/ui/webui/ai_chat and we could get rid of the delegate (this method isn't part of the mojo interface and it doesn't need anything from TabTrackerService).

components/ai_chat/core/common/mojom/untrusted_frame.mojom
```diff
@@ -87,6 +98,10 @@ interface UntrustedConversationHandler {
 
   // Retry the last failed API request.
   RetryAPIRequest();
+
+  // Activate the open tab with the given handle. Used by the cards rendered
+  // for the `tabSearch` markdown directive.
+  SwitchToTab(int32 tab_id);
```
fallaciousreasoning: Can we move this method onto the UIHandler? That way we could get rid of a bunch of plumbing and delete

- tab_activator
- tab_activator_factory

and remove the changes to `ai_chat_service.h`, `conversation_handler` & `tab_tracker_service`.

Is there some reason you want to go through the conversation handler for this?

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/inline_tab_search.tsx
```diff
@@ -0,0 +1,97 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import * as React from 'react'
+import { formatLocale } from '$web-common/locale'
+import * as Mojom from '../../../common/mojom'
+import { useUntrustedConversationContext } from '../../untrusted_conversation_context'
+import styles from './inline_tab_search.module.scss'
+
+function TabSourceCard(props: { source: Mojom.TabSearchResult }) {
```
fallaciousreasoning: Hey, this TabSourceCard looks extremely similar to the `AttachmentPageItem` - I wonder if it'd be worth sharing them (potentially under a different name). It has a few advantages like handling data: uris. 

I think its worth looking at (though might not be worth doing), as these are both conceptually rendering a tab and it'd be nice to ensure everything is consistent.

fallaciousreasoning (changes requested):

components/ai_chat/core/browser/tab_tracker_service.cc
darkdh: fixed with 0635ec92db9

components/ai_chat/core/common/mojom/untrusted_frame.mojom
darkdh: fixed with 0635ec92db9

components/ai_chat/core/common/mojom/untrusted_frame.mojom
darkdh: `TabData` carries a `content_id` (the SiteInfo uuid of the loaded content) that `SearchForTabs` has no meaningful value for, so reusing it means filling a documented field with a placeholder. It'd also mean importing `tab_tracker.mojom` here just to borrow the struct.

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/inline_tab_search.tsx
darkdh: Agreed they're conceptually the same. Sharing isn't just a move though: `AttachmentPageItem` is a static chip with an optional remove button, while this card's whole surface activates the tab, and the secondary line differs (host vs scheme-stripped URL with a tooltip). Unifying means adding a click affordance and picking one subtitle, which changes how one of the two looks — can we do this as a follow-up?

Took the `allowGoogleServerFallback=0` from your version in 547d74a6742.

emerick (approved):
++chromium_src

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/inline_tab_search.tsx
fallaciousreasoning: Scheme stripped URL is probably preferable here, and I suspect the security team would probably like having the tooltip which shows the full URL.

Happy for this to happen as a followup, but I think your `allowGoogleServerFallback=0` is a good example of why unifying is helpful :smile:

browser/history_embeddings/open_tab_search.h
```diff
@@ -26,23 +27,31 @@ namespace history_embeddings {
...
-    base::OnceCallback<void(const std::vector<int32_t>&)>;
+// Descriptor of a matched open tab returned by `SearchOpenTabsByContent`.
+struct OpenTabInfo {
+  int32_t tab_id = 0;
+  // Tab title. Falls back to `url.host()` when the tab has no title yet
+  // (still loading or a page that never set one) so downstream consumers
+  // that display the tab always have something readable.
+  std::string title;
+  GURL url;
+};
```
fallaciousreasoning, line 42: is there a reason we can't use the mojo struct here? These have the same shape.

components/ai_chat/core/common/mojom/untrusted_frame.mojom
fallaciousreasoning: Not a hill I want to die on but it feels weird having two different representations of Tabs in AIChat. Additionally, if we want to have the tabs you surface here interact with the conversation (i.e. let the user add them to the conversation, or show an indicator that they're part of the conversation) we'll need the content_id.

components/ai_chat/core/common/mojom/untrusted_frame.mojom
```diff
@@ -139,6 +150,15 @@ interface UntrustedUIHandler {
...
   OpenAIChatCustomizationSettings();
 
+  // Semantically searches the user's open tabs by page content, on device.
+  // Null where on-device tab search is unavailable (non-desktop); empty when
+  // no open tab matched.
+  SearchForTabs(string query) => (array<TabSearchResult>? tabs);
+
+  // Activate the open tab with the given handle. Used by the cards rendered
+  // for the `tabSearch` markdown directive.
+  SwitchToTab(int32 tab_id);
```
fallaciousreasoning, line 151: optional, but should these be `[EnableIfNot=is_ios]`, seeing as we don't support these features there?

browser/history_embeddings/open_tab_search.h
darkdh: yes, `SearchOpenTabsByContent` is also used by the tab_search page handler, so returning `mojom::TabData`(previously `mojom:: TabSearchResult`) here would make `browser/history_embeddings` depend on AI Chat's mojom.

components/ai_chat/core/common/mojom/untrusted_frame.mojom
darkdh: fixed with e6bedc2555e, 974ccc843f7 and 051f755afe8

components/ai_chat/core/common/mojom/untrusted_frame.mojom
darkdh: iOS serves the same conversation frame, so compiling these out means guarding the TS that references them: the `searchForTabs` endpoint, `switchToTab` in `UIHandlerActions`, and the `tabSearch` directive entries. That's four `<if expr="!is_ios">` blocks to remove two stubs, and the directive renders nothing on iOS either way.

browser/history_embeddings/open_tab_search.h
fallaciousreasoning: Is there a reason it can't? Is it possible to have history_embeddings enabled and AI_CHAT off?

fallaciousreasoning (approved):

components/ai_chat/resources/untrusted_conversation_frame/components/assistant_response/inline_tab_search.tsx
darkdh: [follow-up](https://github.com/brave/brave-browser/issues/58107) created

browser/history_embeddings/open_tab_search.h
darkdh: yes, history_embeddings should not introduce ai_chat dependency since it is not Leo related. It is an independent on device ml feature that is used by `about:history`, `tab search UI` and now extend to access for ai_chat, not the other way around.

netzenbot (approved):

components/resources/ai_chat_ui_strings.grdp
```diff
@@ -498,6 +498,12 @@
   <message name="IDS_CHAT_UI_SEARCH_QUERIES" desc="AI Response status to indicate that search queries were used to enhance the response" formatter_data="webui=AiChat">
     Searching the web for <ph name="SEARCH_TERMS">$1<ex>"my first search engine query", "my second search engine query"</ex></ph>.
   </message>
+  <message name="IDS_CHAT_UI_TAB_SEARCH_IN_PROGRESS" desc="Status shown while the browser searches the user's open tabs for a query the AI assistant asked about" formatter_data="webui=AiChat">
+    Searching your open tabs for <ph name="QUERY">$1<ex>react hooks</ex></ph>…
+  </message>
+  <message name="IDS_CHAT_UI_TAB_SEARCH_NO_RESULTS" desc="Message shown when none of the user's open tabs match what the AI assistant searched for" formatter_data="webui=AiChat">
+    No open tabs match <ph name="QUERY">$1<ex>react hooks</ex></ph>.
+  </message>
```
petemill, line 524: Do we want to put quote marks around the query?

petemill (approved):
++

components/resources/ai_chat_ui_strings.grdp
darkdh: quote added!

PR #36850, opened 2026-05-29
[farbling] Fix webgpu tests and scrubbing issue

r0hit0303:
Currently the [BraveScrubWebGpuAdapterInfo](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/chromium_src/third_party/blink/renderer/modules/webgpu/gpu_adapter.cc?L60) instance in [gpu_adapter.cc](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/chromium_src/third_party/blink/renderer/modules/webgpu/gpu_adapter.cc?L60) is a temporary object that gets destroyed immediately after the semi-colon. This object was originally intended to be preserved until the end of the [method scope](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/webgpu/gpu_adapter.cc;l=125?q=gpu_adapter.cc) where it's defined, and, because of the early destruction the scrubbing didn't occur. 

This bug was also not also flagged by the corresponding [brave_webgpu_farbling_browsertests](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/browser/farbling/brave_webgpu_farbling_browsertest.cc) either. The tests were shown as passed but they were actually [skipped](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/browser/farbling/brave_webgpu_farbling_browsertest.cc?L179-181) because of not having a valid instance of `navigator.gpu`. The GTEST_SKIP() should have ideally printed out the SKIPPED with the skipped message, but due to the fact that our tests are piped through `npm run test` it added extra constraints on the output shown to the user. So, both "SKIPPED" and the message were omitted. 

This change also fixes the webgpu browser tests which skipped the tests. The fix was to serving the test page from a secured context, as it was required to access the `navigator.gpu` instance. This constraint is set in [navigator_gpu.idl](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/webgpu/navigator_gpu.idl;l=10?q=navigator_gpu.idl) SECURE_CONTEXT constraint. Along with the secure context, the tests also had to add the command line switch [kEnableUnsafeWebGPU](https://source.chromium.org/chromium/chromium/src/+/main:gpu/config/gpu_switches.cc;l=52). 

This change also adds another test to ensure we don't scrub when the [kEnableWebGPUDeveloperFeatures](https://source.chromium.org/chromium/chromium/src/+/main:gpu/config/gpu_switches.cc;l=56-59) switch was enabled. 

Related https://github.com/brave/brave-browser/issues/55677

### DIFF
--- a/browser/farbling/brave_webgpu_farbling_browsertest.cc
+++ b/browser/farbling/brave_webgpu_farbling_browsertest.cc
@@ -9,6 +9,7 @@
 #include "brave/components/brave_shields/core/common/features.h"
 #include "brave/components/constants/brave_paths.h"
 #include "brave/components/webcompat/core/common/features.h"
+#include "build/build_config.h"
 #include "chrome/browser/content_settings/host_content_settings_map_factory.h"
 #include "chrome/browser/profiles/profile.h"
 #include "chrome/browser/ui/browser.h"
@@ -162,10 +163,18 @@ INSTANTIATE_TEST_SUITE_P(
                         : "WebGLBalancedFingerprintingProtection_Disabled";
     });
 
+// Windows x64 doesn't have the WebGPU adapter support.
+#if BUILDFLAG(IS_WIN) && defined(ARCH_CPU_X86_64)
+#define NON_WIN_X64_TEST(test) DISABLED_##test
+#else
+#define NON_WIN_X64_TEST(test) test
+#endif
+
 // Tests that navigator.gpu.requestAdapter() → adapter.info fields (vendor,
 // architecture, device) are emptied under BALANCED (feature on) and MAXIMUM
 // farbling, and are returned as-is under OFF and BALANCED (feature off).
-IN_PROC_BROWSER_TEST_P(BraveWebGPUAdapterInfoTest, FarbleAdapterInfo) {
+IN_PROC_BROWSER_TEST_P(BraveWebGPUAdapterInfoTest,
+                       NON_WIN_X64_TEST(FarbleAdapterInfo)) {
   const std::string domain = "a.com";
   const GURL url = https_server_.GetURL(domain, "/adapter-info.html");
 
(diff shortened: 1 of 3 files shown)


r0hit0303:
Thanks @goodov for catching this issue

browser/farbling/brave_webgpu_farbling_browsertest.cc
```diff
@@ -203,3 +243,60 @@ IN_PROC_BROWSER_TEST_P(BraveWebGPUAdapterInfoTest, FarbleDeviceAdapterInfo) {
...
+
+  // Farbling level: off — collect the real adapter info as a baseline.
+  AllowFingerprinting(domain);
+  ASSERT_TRUE(ui_test_utils::NavigateToURL(browser(), url));
+  const std::string actual_off =
+      EvalJs(contents(), kGPUAdapterInfoScript).ExtractString();
+
+  if (actual_off == "no-gpu" || actual_off == "no-adapter") {
+    GTEST_SKIP() << "WebGPU adapter not available in this environment";
+  }
```
r0hit0303: @goodov would you know how can we reliably print this string in the bots where we didn't find webgpu support? I would love to have this explicitly confirmed during the tests w/o needing to fail it in such cases.

It would be great if we could print un-conditionally the WebGPU state before running the tests or at-least get some indicators about the availability of the webgpu in the [bots](https://ci.brave.com/job/brave-core-build-pr-windows-x64/job/PR-36850/1/testReport/(root)/BraveWebGPUAdapterInfoTest/). Would you know how we could do it?  

There seems to be [--gtest_output=xml](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/build/commands/lib/test.js?L213) support but I didn't see any corresponding artifacts maybe I missed it. In the upstream there seems to be support for [test-launcher-print-test-stdio=always](https://source.chromium.org/search?q=test-launcher-print-test-stdio%3Dalways) I was not too sure of that since printing to I/O is usually expensive and I didn't want to increase the overall time to run the whole brave_browser_tests. Right now, I can only verify locally in my macOS that these tests are getting a valid webgpu, but would be cool to have evidence also for other OSs where they will run on.

goodov: enable the test everywhere and then disable it per-platform via `MAYBE_` pattern. I think that's the only way to reliably detect if we lost webgpu on CI.

r0hit0303: Yeah, that seems like a good way of tackling this. Let me remove the `GTEST_SKIP` and see which CI's would be failing.

r0hit0303: Circling back on this with update. I think we might not have WebGPU for `windows-x64` as per [1](https://ci.brave.com/job/brave-core-build-pr-origin-windows-x64/job/PR-36850/2/) and [2](https://ci.brave.com/job/brave-core-build-pr-windows-x64/job/PR-36850/2/). So will disable the tests for windows 64 platform.

netzenbot (approved):

goodov (approved):

r0hit0303:
Thanks @goodov for the review

PR #36853, opened 2026-05-29
[iOS] Add support for token validation in JavaScriptFeatures

kylehickinson:
This change adds a helper class `MessageHandlerToken` that can be used in JavaScriptFeature's to support tokenized communication validation. The token is a UUIDv4 which is injected with the script itself via placeholder

### DIFF
--- a/ios/web/js_messaging/BUILD.gn
+++ b/ios/web/js_messaging/BUILD.gn
@@ -5,28 +5,56 @@
 
 import("//ios/web/public/js_messaging/compile_ts.gni")
 import("//ios/web/public/js_messaging/optimize_js.gni")
+import("//ios/web/public/js_messaging/optimize_ts.gni")
 
 source_set("js_messaging") {
   sources = [
     "safe_builtins_javascript_feature.h",
     "safe_builtins_javascript_feature.mm",
   ]
-  deps = [ ":safe_builtins_js" ]
+  deps = [
+    ":message_handler_token",
+    ":safe_builtins_js",
+  ]
   public_deps = [ "//ios/web/public/js_messaging" ]
+}
+
+source_set("message_handler_token") {
+  sources = [
+    "message_handler_token.h",
+    "message_handler_token.mm",
+  ]
+  public_deps = [
+    "//base",
+    "//ios/web/public/js_messaging",
+  ]
 }
 
 source_set("unittests") {
   testonly = true
-  sources = [ "safe_builtins_javascript_test.mm" ]
+  sources = [
+    "message_handler_token_inttest.mm",
+    "safe_builtins_javascript_test.mm",
+  ]
   deps = [
+    ":message_handler_token",
+    ":message_handler_token_test_api",
     ":safe_builtins_js",
     "//base",
     "//base/test:test_support",
+    "//ios/web/public/js_messaging",
     "//ios/web/public/test:javascript_test",
+    "//ios/web/public/test:test_fixture",
     "//ios/web/public/test:util",
+    "//ios/web/public/test/fakes",
     "//testing/gtest",
     "//url",
   ]
+}
+
+compile_ts("util_scripts") {
+  sources = [ "resources/utils.ts" ]
+  deps = [ ":safe_builtins" ]
 }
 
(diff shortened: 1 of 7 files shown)


ios/web/js_messaging/BUILD.gn
```diff
@@ -5,30 +5,58 @@
...
+  sources = [
+    "message_handler_token.h",
+    "message_handler_token.mm",
+  ]
+  public_deps = [
+    "//base",
+    "//ios/web/public/js_messaging",
+  ]
+}
+
```
netzenbot, line 32: nit: Per [BS-028](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-028), internal `source_set` targets should have a restricted visibility (e.g. `visibility = [ ":*" ]`) to prevent external use. Consider adding one here since this target is a composable helper rather than a top-level public API.

ios/web/js_messaging/message_handler_token.h
```diff
@@ -0,0 +1,48 @@
...
+  // the JavaScript-side placeholder (`gCrWebPlaceholderMessageHandlerToken`)
+  PlaceholderReplacements GetPlaceholderReplacements() const;
+
+ private:
+  base::Uuid token_ = base::Uuid::GenerateRandomV4();
+};
+
+}  // namespace web
+
+#endif  // BRAVE_IOS_WEB_JS_MESSAGING_MESSAGE_HANDLER_TOKEN_H_
```
netzenbot, line 50: nit: `std::optional` is used in the return type but `<optional>` isn't included — it arrives transitively via `java_script_feature.h`. Per IWYU, add `#include <optional>` directly. Similarly, `base::Value` is only used as a pointer here; prefer a forward declaration (`namespace base { class Value; }`) over relying on the transitive `base/values.h` from `java_script_feature.h`. [CS-001](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-001)

ios/web/js_messaging/message_handler_token.h
```diff
@@ -0,0 +1,48 @@
...
+// A composable helper that generates a stable, per-instance token for use with
+// JavaScriptFeature subclasses.
+//
+// Hold an instance as a member of a JavaScriptFeature and provide
+// FeatureScripts replacements callback with the GetPlaceholderReplacements call
+// and use `GetValidatedScriptMessageBody` in your JavaScriptMessage's message
+// received override.
+//
+// To send tokenized messages to the JavaScriptFeature from TypeScript, use
+// `sendTokenizedWebKitMessage` or `sendTokenizedWebKitMessageWithReply` from
```
netzenbot, line 28: nit: The doc comment only says "Returns the message body if the token matches" but doesn't describe the `std::nullopt` return cases (mismatched token, non-dict body, missing `message` key). Per DOC-004, public method docs should describe the full contract including all return values. See https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-004

stoletheminerals (approved):

ios/web/js_messaging/BUILD.gn
kylehickinson: Going to leave this as is, a future PR will move them into the `js_messaging` target

netzenbot (approved):

ios/web/js_messaging/resources/utils.ts
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) 2026 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import { gSafeBuiltins }
+  from "//brave/ios/web/js_messaging/resources/safe_builtins.js";
+
+// A token to be used when communication with the browser
```
StephenHeaps:
```suggestion
// A token to be used for validating communication with the browser
```

StephenHeaps (approved):

nuo-xu (approved):

PR #36857, opened 2026-05-29
Upgrade from Chromium 149.0.7827.23 to Chromium 149.0.7827.54 (1.92.x).

mkarolin:
Resolves https://github.com/brave/brave-browser/issues/55943


kjozwiak (approved):
Uplift into `1.92.x` approved 👍 `origin-macos-arm64` failed due to `WASM/Xcode license` issues which is known/being worked through via https://bravesoftware.slack.com/archives/C7VLGSR55/p1779466219679559 & https://bravesoftware.slack.com/archives/C7VLGSR55/p1780311110558229. As discussed in those thread, going to go ahead and not block uplifts to keep things moving. Other platforms passed without any issues/failures.

PR #36861, opened 2026-05-29
Enable Bosnian to be used as a UI language.

### LINKED ISSUE #37196: [l10n] add bs language
Hello,

We would like to begin supporting the Bosnian language which is indicated by the locale code `bs`. 

We will not be paying for localization services in Bosnian; however we'll be leveraging a combination of resources from Chromium and previously existing translations. 

For context, Bosnian, Croatian, Montenegrin and Serbian languages are all very similar languages. In fact, they are fundamentally the same language with some political differences due to their history. 

Currently we have a substantial user base in Bosnian but as we do not support the locale `bs`, all of those users are viewing the app and content in English. 

Although it isn't a perfect solution, we would like to copy over the Croatian localizations since it is the closest to Bosnian and the least political. 

Therefore, moving forward, can we please use a one-to-many mapping for the `hr` language:

``hr`: bs, hr`

mkarolin:
Resolves https://github.com/brave/brave-browser/issues/37196


netzenbot (approved):

bsclifton (commented):
string reviewer ++

bsclifton (approved):

PR #36864, opened 2026-05-31
[toolchain] `install_extra_deps.py` further fixes

cdesouza-chromium:
This PR does some additional fixes to `install_extra_deps.py`, making it
more stable for use. The main things fixed in this change are:

 * `DEPS` is now parsed as read-only, to avoid any unwanted
   side-effects.
 * `.vpython3` is being introduced `install_extra_deps.py`, as this
   script is expected to be run in place.
 * The script is being renamed to `install_extra_deps.py`, as this is
   more self-descriptive.

The values committed now in the `EXTRA_DEPS` set are valid data.


tools/cr/toolchains/.vpython3
```diff
@@ -0,0 +1,15 @@
+# Copyright (c) 2025 The Brave Authors. All rights reserved.
```
netzenbot: nit: New file copyright says 2025, but the current year is 2026. Per [CS-012](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-012), new files must use the current year.

tools/cr/toolchains/install_extra_deps.py
```diff
@@ -102,44 +105,48 @@ def _select_object(objects: list[dict],
 def _extract(archive_path: Path, dest_dir: Path) -> None:
     """Overlay the archive onto the destination directory.
 
-    The archive is expected to lay its members out relative to the destination
-    root, so it is extracted straight into `dest_dir`, overwriting any files it
-    matches and leaving everything else in the directory untouched.
+    For the first case that this script is being used, the Rust/WASM toolchain,
```
netzenbot: nit: The new `_extract` docstring replaces the clear behavioral contract with background history that includes temporal qualifiers like "at the moment" and "may decide to change at some point" — the kind of prose that goes stale and doesn't get updated. Also a typo: "determinatio". The old description (archive laid out relative to destination root, extracted straight in, overwrites matches, leaves rest untouched) was more useful. Consider keeping it and trimming the history. [DOC-011](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-011)

tools/cr/toolchains/install_extra_deps.py
```diff
@@ -0,0 +1,332 @@
...
+                'object_name': 'mac-rust-toolchain-4c4205163abcbd08948b3efab796c543ba1ea687-2-llvmorg-23-init-10931-g20b6ec66-1.tar.xz',
+                'sha256sum': 'f0ff955a916e094bf31e61606e50efdab49205576358b42510c20002f9dd0633',
+                'overlayed_on': 'Mac/rust-toolchain-4c4205163abcbd08948b3efab796c543ba1ea687-2-llvmorg-23-init-10931-g20b6ec66.tar.xz',
+                'condition': 'host_os == "mac" and host_cpu == "x64"',
+            },
+            {
+                'object_name': 'win-rust-toolchain-4c4205163abcbd08948b3efab796c543ba1ea687-2-llvmorg-23-init-10931-g20b6ec66-1.tar.xz',
+                'sha256sum': '0fbafcdaab3a76335e8d6c14398cbb7c0eb46f240e831f7607822ae1c6ee4162',
+                'overlayed_on': 'Win/rust-toolchain-4c4205163abcbd08948b3efab796c543ba1ea687-2-llvmorg-23-init-10931-g20b6ec66.tar.xz',
+                'condition': 'host_os == "win"',
```
goodov, line 85: so we're adding a bunch of depot_tools imports and try to use them in a reasonable manner. I guess it works, but do you really want to support this long term? Parsing gclient/DEPS manually, evaluating conditions manually - all of that requires support, and without clear understanding of all edge cases this might be problematic, there's a lot of going on in this script.

The existing hooks approach does the job, but also clear and doesn't require us track any depot_tools changes. We can even implement it to look kinda the way you want:

```python
vars = {
  'rust_toolchain_linux_condition': 'not rust_force_head_revision and host_os == "linux"',
  'rust_toolchain_mac_arm64_condition': 'not rust_force_head_revision and host_os == "mac" and host_cpu == "arm64"',
  'rust_toolchain_mac_x64_condition': 'not rust_force_head_revision and host_os == "mac" and host_cpu == "x64"',
  'rust_toolchain_win_condition': 'not rust_force_head_revision and host_os == "win"',
}

{
  'name':   'download_rust_toolchain_aux',
  'pattern':   '.',
  'action': [
    'vpython3', 'tools/cr/toolchains/download_published_dep.py',
    '--path', 'src/third_party/rust-toolchain'
    '--bucket', 'https://brave-build-deps-public.s3.brave.com/rust-toolchain-aux/',
    '--object',
    'linux-x64-rust-toolchain-4c4205163abcbd08948b3efab796c543ba1ea687-2-llvmorg-23-init-10931-g20b6ec66-1.tar.xz',
    'bcfc7b7b273bcce1570f835727f53c53be0e319a87cd7da9ac7fe2e03752bee4',
    '{rust_toolchain_linux_condition}',
    '--object',
    'mac-arm64-rust-toolchain-4c4205163abcbd08948b3efab796c543ba1ea687-2-llvmorg-23-init-10931-g20b6ec66-1.tar.xz',
    '2fd9e14a5eb879c893fdfca25fb752f3599f1753cbca7e6da404c60721bb9060',
    '{rust_toolchain_mac_arm64_condition}',
    '--object',
    'mac-rust-toolchain-4c4205163abcbd08948b3efab796c543ba1ea687-2-llvmorg-23-init-10931-g20b6ec66-1.tar.xz',
    'fe14b9a46c3cd16d9da2b0a8dab0a9a23f791af9b3a8be617d83915bbf961bc1',
    '{rust_toolchain_mac_x64_condition}',
    '--object',
    'win-rust-toolchain-4c4205163abcbd08948b3efab796c543ba1ea687-2-llvmorg-23-init-10931-g20b6ec66-1.tar.xz',
    '1529acb4b61ac48b8ec3c4fe6178d67261b2b5ba64ceca42c2f5abcbe3f60cf1',
    '{rust_toolchain_win_condition}',
  ],
},
```

The overlay check feature most likely should be opt-in per-dependency. When enabled, then we can parse the upstream `DEPS` and check for a dep match (should be enough to get a single match, i.e. we shouldn't care about conditions).

This way the solution should look much simpler, unless you have some other expectations to further improve things here.

cdesouza-chromium: I have been trying to evolve this design. At the moment, the only dep I'm testing with is an overlay dep, so I have left as is for now, but I have question whether or not we even should have overlay deps to begin with, or if we should just package the whole rust toolchain ourselves, as that is safer when it comes to staleness.

> The overlay check feature most likely should be opt-in per-dependency.

I intend to go on that direction, once there is a non-overlay DEP added, but I'm trying to prototype this idea in a basic way to experiment with things.

I know you may have qualms about the gclient imports, but this hits on the rebase team first, so I want to run experiments with it to see what works best. I'm not opposed to trying the way you are suggesting, but I would prefer to give this approach a go to see how well it can work.

netzenbot (approved):

emerick (approved):

PR #36869, opened 2026-05-31
[farbling] Fix Accept-Language header position

### LINKED ISSUE #55271: Language fingerprinting issues on Albertsons websites (and potentially more)
When accessing sites under the Albertsons banner (albertsons.com, safeway.com, shaws.com, etc), the search functionality is broken when language farbling is enabled. When we disable the language farbling, the search functions as expected.

After investigating, the likely source of these issues is a bot detection tool called Imperva, which is loaded on these sites. When the search request initiates, it waits for a bot detection check from Imperva to pass before returning results. When we activate the language farbling, Imperva determines that we are a bot and the search hangs because it is still waiting for Imperva to approve.

After removing the q-value farbling and navigator.language farbling in brave-core and rebuilding, these possible causes have been ruled out.

Right now, these sites have language fingerprinting disabled by default through a webcompat exceptions list, allowing users full access to these sites by default.

Note, this may not be limited to Albertsons sites, and may be an issue for other sites that use Imperva.

### Steps to reproduce
1. Go to https://www.safeway.com/shop/search-results.html?q=chocolate&tab=products
2. Check that it works.
3. Check that it works if Shields > Advanced options > `Block fingerprinting` > press the caret > `Language` toggle is on.
4. Try again.

#### Broken page

Image

#### Unbroken page (with language fingerprinting protection turned off)

Image

r0hit0303:
Brave farbles the Accept-Language header when the shields are up. Farbling is done such that sites don't get to know whether they received a farble value. However, there are cases where sites are looking into the position of the Accept-Language header to deterministically tell the user is farbling that header. This leads to site flagging the user unfortunately.

This change tries to mitigate it by readjusting the position of the Accept-Language header such that there's no parity in the position when the shields are "on" vs "off". 

**Demo (before fix):** 

https://github.com/user-attachments/assets/128afc25-29bd-49f7-9dd4-ea625e754419

**Demo (after fix):** 

https://github.com/user-attachments/assets/0ce1d400-4388-4346-87d1-2b6be74f29ca

Resolves: https://github.com/brave/brave-browser/issues/55271

### DIFF
--- a/browser/farbling/brave_navigator_languages_farbling_browsertest.cc
+++ b/browser/farbling/brave_navigator_languages_farbling_browsertest.cc
@@ -4,7 +4,6 @@
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 #include <map>
-#include <memory>
 #include <optional>
 #include <string>
 #include <string_view>
@@ -13,15 +12,9 @@
 #include "base/path_service.h"
 #include "base/strings/string_split.h"
 #include "base/strings/string_util.h"
-#include "base/test/bind.h"
-#include "base/test/thread_test_helper.h"
-#include "brave/browser/brave_browser_process.h"
-#include "brave/browser/extensions/brave_base_local_data_files_browsertest.h"
-#include "brave/components/brave_component_updater/browser/local_data_files_service.h"
 #include "brave/components/brave_shields/core/browser/brave_shields_utils.h"
 #include "brave/components/brave_shields/core/common/features.h"
 #include "brave/components/constants/brave_paths.h"
-#include "brave/components/constants/pref_names.h"
 #include "chrome/browser/content_settings/host_content_settings_map_factory.h"
 #include "chrome/browser/extensions/extension_browsertest.h"
 #include "chrome/browser/profiles/profile.h"
@@ -30,7 +23,6 @@
 #include "chrome/test/base/ui_test_utils.h"
 #include "components/language/core/browser/pref_names.h"
 #include "components/network_session_configurator/common/network_switches.h"
-#include "components/permissions/permission_request.h"
 #include "components/prefs/pref_service.h"
 #include "components/user_prefs/user_prefs.h"
 #include "content/public/browser/render_frame_host.h"
@@ -163,8 +155,6 @@ class BraveNavigatorLanguagesFarblingBrowserTest : public InProcessBrowserTest {
     prefs->Set(language::prefs::kSelectedLanguages,
                base::Value(accept_languages));
   }
-
-  void HandleHTTPRequest() {}
 
   void MonitorHTTPRequest(const net::test_server::HttpRequest& request) {
     if (request.relative_url.find("/reduce-language/") == std::string::npos) {
--- a/browser/farbling/brave_navigator_languages_farbling_browsertest.cc
+++ b/browser/farbling/brave_navigator_languages_farbling_browsertest.cc
@@ -5,9 +5,15 @@
 
 #include <map>
 #include <memory>
+#include <optional>
 #include <string>
+#include <string_view>
+#include <vector>
 
 #include "base/path_service.h"
+#include "base/strings/string_split.h"
+#include "base/strings/string_util.h"
+#include "base/test/bind.h"
 #include "base/test/thread_test_helper.h"
 #include "brave/browser/brave_browser_process.h"
 #include "brave/browser/extensions/brave_base_local_data_files_browsertest.h"
@@ -40,6 +46,30 @@ using content::TitleWatcher;
 
 namespace {
 constexpr char kNavigatorLanguagesScript[] = "navigator.languages.toString()";
+
+// Returns the zero-based index of `header_name` among the request headers in a
+// raw header block (as exposed via
+// `net::test_server::HttpRequest::all_headers`). The block starts with the
+// request line (e.g. "GET / HTTP/1.1"), which is ignored, followed by one
+// header per line. Returns std::nullopt if the header is not present.
+std::optional<size_t> GetHeaderPosition(const std::string& all_headers,
+                                        std::string_view header_name) {
+  std::vector<std::string> lines = base::SplitString(
+      all_headers, "\r\n", base::TRIM_WHITESPACE, base::SPLIT_WANT_NONEMPTY);
+  // Skip the request line at index 0; headers start at index 1.
+  for (size_t i = 1; i < lines.size(); ++i) {
+    size_t colon_pos = lines[i].find(':');
+    if (colon_pos == std::string::npos) {
+      continue;
+    }
+    std::string_view name = base::TrimWhitespaceASCII(
+        std::string_view(lines[i]).substr(0, colon_pos), base::TRIM_ALL);
+    if (base::EqualsCaseInsensitiveASCII(name, header_name)) {
+      return i - 1;
+    }
+  }
+  return std::nullopt;
+}
 }  // namespace
 
 class BraveNavigatorLanguagesFarblingBrowserTest : public InProcessBrowserTest {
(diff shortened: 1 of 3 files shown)


browser/farbling/brave_navigator_languages_farbling_browsertest.cc
```diff
@@ -40,6 +38,30 @@ using content::TitleWatcher;
 
 namespace {
 constexpr char kNavigatorLanguagesScript[] = "navigator.languages.toString()";
+
+// Returns the zero-based index of `header_name` among the request headers in a
```
r0hit0303, line 42: The test was written with the help of LLM. I have verified the test locally. 

We needed this ugly `GetHeaderPosition` because the `net::test_server::HttpRequest` exposed each headers only through a Map which unfortunately loses the actual header ordering. Only way was to tap into the HttpRequest::all_headers member variable which records the whole request exactly as it was sent as a std::string. Therefore we had to parse it with `GetHeaderPosition`.  Overall the test is to see that the position of the Accept-Language in the header when farbling was "off" is exactly the same as when farbling was "on".

r0hit0303:
@goodov could you ptal? Thanks

chromium_src/net/url_request/url_request_http_job.cc
```diff
@@ -7,6 +7,35 @@
...
+    auto modified_language_value =
+        headers.GetHeader(HttpRequestHeaders::kAcceptLanguage);
+    if (modified_language_value.has_value()) {
+      // We remove the header first from the underlying vector.
+      headers.RemoveHeader(HttpRequestHeaders::kAcceptLanguage);
+      // This will put the header at the end.
+      headers.SetHeader(HttpRequestHeaders::kAcceptLanguage,
+                        modified_language_value.value());
+    }
+  }
```
goodov, line 31: this looks like a workaround, rather than a long term solution, and I'm not sure if the workaround is right.

What if some other header is expected after `accept-language`? I don't think this preserves the original order in which Chromium sends headers, it just happens to be the case right now. What if a websites sends additional headers via `fetch`? Will the order be right?

Upstream has `content/public/browser/reduce_accept_language_utils.cc` that seems like does the same thing that we do. Did we look into it? Maybe just switch to it and remove our impl?

r0hit0303: It's indeed a workaround. I couldn't find a better way to resolve the issue. I didn't know `content/public/browser/reduce_accept_language_utils.cc` existed in the upstream. Let me look at that to explore other potential solutions. 

> What if some other header is expected after accept-language?  What if a websites sends additional headers via fetch? Will the order be right?

All good points. FWIW, this change doesn't need to care about where the Accept-Language would be placed **precisely**, but rather placing it in the same position where by default the browser would have put the `Accept-Language` header had it not be farbled. Please see https://source.chromium.org/chromium/chromium/src/+/main:net/url_request/url_request_http_job.cc;l=810 when the browser is putting this header by default.

chromium_src/net/url_request/url_request_http_job.cc
```diff
@@ -7,6 +7,35 @@
 
 #include "net/http/transport_security_state.h"
 
+namespace net {
+
```
netzenbot, line 11: nit: Per [CS-007](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-007), the `Brave*` prefix should only be used when overriding or subclassing Chromium classes. Since `BraveAcceptLanguageHttpHeaderReposition` is a new Brave-only helper and already lives in `namespace net::brave`, consider naming it `AcceptLanguageHttpHeaderReposition` instead.

browser/farbling/brave_navigator_languages_farbling_browsertest.cc
```diff
@@ -338,6 +377,46 @@ IN_PROC_BROWSER_TEST_F(BraveNavigatorLanguagesFarblingBrowserTest,
...
+      domain, "/reduce-language/page-with-subresources.html?off");
+  GURL url_default = https_server_.GetURL(
+      domain, "/reduce-language/page-with-subresources.html?default");
+  GURL url_max = https_server_.GetURL(
+      domain, "/reduce-language/page-with-subresources.html?max");
+  SetAcceptLanguages("la,es,en");
+
+  // Farbling level: off. This is the baseline position to compare against.
+  AllowFingerprinting(domain);
+  ASSERT_TRUE(ui_test_utils::NavigateToURL(browser(), url_off));
```
netzenbot, line 403: nit: The two `EXPECT_EQ(accept_language_position_, position_when_off)` assertions have no `<<` context, so a failure won't tell you which farbling level triggered it or what the actual vs expected position was. Consider adding e.g. `<< "farbling=default"` and `<< "farbling=max"` respectively. See [TI-014](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-014).

patches/net-url_request-url_request_http_job.cc.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/net/url_request/url_request_http_job.cc b/net/url_request/url_request_http_job.cc
+index 15b796f4c61b325305d4f9115018483fc95f6f70..75cca1f43ffdb920ee07a36c1a059cb1596cc185 100644
+--- a/net/url_request/url_request_http_job.cc
++++ b/net/url_request/url_request_http_job.cc
+@@ -787,6 +787,7 @@ void URLRequestHttpJob::AddExtraHeaders() {
+       request()->context()->enable_brotli(),
+       request()->context()->enable_zstd());
+ 
++  BRAVE_ACCEPT_LANGUAGE_HTTP_HEADER_REPOSITION
```
goodov: `BRAVE_URL_REQUEST_HTTP_JOB_ADD_EXTRA_HEADERS`

chromium_src/net/url_request/url_request_http_job.cc
```diff
@@ -7,6 +7,35 @@
...
+namespace brave {
+
+// The Accept-Language header position is different when farbling is "off" vs
+// when farbling is "on". This class helps to keep the position same to remove
+// farbling detection by sites.
+// See https://github.com/brave/brave-browser/issues/55271 for more details.
+class BraveAcceptLanguageHttpHeaderReposition {
```
goodov, line 17: since this is used only internally, pls put it into anonymous namespace.

chromium_src/net/url_request/url_request_http_job.cc
goodov: okay, that's fine then.

goodov (approved):
lgtm with few things to improve.

overall maybe we should ditch this in favor of upstream impl down the line.

chromium_src/net/url_request/url_request_http_job.cc
r0hit0303: I've created an issue [here](https://github.com/brave/brave-browser/issues/55980) to track the migration to the upstream code. I'll follow up on this separately.

patches/net-url_request-url_request_http_job.cc.patch
r0hit0303: Done

chromium_src/net/url_request/url_request_http_job.cc
r0hit0303: Done

r0hit0303:
Thanks @goodov for the review. Filed a separate [bug](https://github.com/brave/brave-browser/issues/55980) to explore feasibility for migrating to the upstream's fingerprinting protections.

chromium_src/net/url_request/url_request_http_job.cc
goodov: I don't see anonymous namespace

r0hit0303: Apologies, I'm not sure what happened there. It should be good now.

PR #36871, opened 2026-05-31
[toolchain] Xcode toolchain builder to handle Xcode installs

cdesouza-chromium:
This PR further improves the Xcode toolchain builder to be capable to
deploy an Xcode toolchain based on what the toolchain used by Chromium
upstream is using. This makes the toolchain we produce deterministic,
rather than the builder packing whatever is installed in the node.

The archives are confirmed using `xcodereleases.com`, and are pulled
from a local cache. The builder carries out the necessary sanity checks
to make sure we are using an Xcode version that is exactly what is
published in `xcodereleases.com`.

With this system, the builder deploys an Xcode install under
`/Applications`, keyed on the official build number,
`xcode_<build>.app`. This means that we can have multiple Xcode versions
deployed in a single node, if necessary. It is also the case that no
installation occurs if a version is already found in place.

Additionally, an index is also being introduced, to preserve the
relevant reproducibility details regarding any given SDK being packaged.

### DIFF
--- a/tools/cr/toolchains/README.md
+++ b/tools/cr/toolchains/README.md
@@ -111,7 +111,7 @@ be extracted directly over it. Invoke with the target path, e.g.
 
 ### `build_xcode_toolchain.py`
 
-macOS-only. Builds a hermetic, reproducible Xcode toolchain archive from
+Builds a hermetic, reproducible Xcode toolchain archive from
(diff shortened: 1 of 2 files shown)


AlexeyBarabash (approved):
LGTM

PR #36879, opened 2026-06-01
[toolchain] Use `unxip` in the Xcode toolchain builder

cdesouza-chromium:
This PR adds the use of `unxip` when possible in the builder, and falls
back to `xip` with a warning, when `unxip` is not found in the system


AlexeyBarabash (approved):
++

PR #36884, opened 2026-06-01
[toolchain] Downloading the Xcode toolchain with a new scheme

cdesouza-chromium:
This PR updates the script that downloads the hermetic toolchain to rely
on a new URL scheme, which is now used by the new toolchain builder.
This scheme is based on the two deciding factors when generating the
toolchain: SDK version, and build number for it. This should be enough
to go by to determine which SDK to use.

This script still needs the Xcode number for validation, and that can be
provided by the index generated for the given SDK, in this case,
`xcode-hermetic-toolchain-26.4-25E251.yaml`.

This tarball being pulled with this change is not the same that was
being pulled before. This tarball was generated today, with the builder
pulling in and installing Xcode. For more details, see:

https://github.com/brave/brave-core/pull/36871

### DIFF
--- a/build/mac/download_hermetic_xcode.py
+++ b/build/mac/download_hermetic_xcode.py
@@ -35,15 +35,16 @@ sys.path.append(str(Path(__file__).resolve().parents[2] / 'script'))
 import deps  # pylint: disable=wrong-import-position
 
 # The hash sum for the archive expected to be downloaded.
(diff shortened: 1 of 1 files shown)


netzenbot (approved):

AlexeyBarabash (approved):
LGTM

PR #36891, opened 2026-06-01
Account: parameterize verification email resend/cancel by intent

### LINKED ISSUE #55976: Account: parameterize verification email resend/cancel by intent
This will lay the groundwork for supporting verification email resends across different verification intents (registration, password reset, and password change), as well as canceling those verifications.

szilardszaloki:
Resolves https://github.com/brave/brave-browser/issues/55976.

Follow-ups:
- https://github.com/brave/brave-browser/issues/55978

### DIFF
--- a/browser/resources/settings/getting_started_page/brave_account_row.ts
+++ b/browser/resources/settings/getting_started_page/brave_account_row.ts
@@ -101,7 +101,8 @@ export class SettingsBraveAccountRowElement extends I18nMixinLit(CrLitElement) {
     let error: ResendConfirmationEmailError | undefined
 
     try {
-      await this.browserProxy.authentication.resendConfirmationEmail()
+      await this.browserProxy.authentication.resendVerificationEmail(
+        { loggedOutIntent: this.state!.loggedOut!.verification!.intent })
     } catch (e) {
       if (e && typeof e === 'object') {
         error = e as ResendConfirmationEmailError
@@ -133,7 +134,8 @@ export class SettingsBraveAccountRowElement extends I18nMixinLit(CrLitElement) {
   }
 
   protected onCancelRegistrationButtonClicked() {
-    this.browserProxy.authentication.cancelRegistration()
+    this.browserProxy.authentication.cancelVerification(
+      { loggedOutIntent: this.state!.loggedOut!.verification!.intent })
   }
 
   protected openBraveAccountDialog() {
--- a/browser/resources/settings/getting_started_page/brave_account_row.ts
+++ b/browser/resources/settings/getting_started_page/brave_account_row.ts
@@ -100,9 +100,10 @@ export class SettingsBraveAccountRowElement extends I18nMixinLit(CrLitElement) {
 
     let error: ResendConfirmationEmailError | undefined
 
+    assert(this.state?.loggedOut?.verification)
     try {
       await this.browserProxy.authentication.resendVerificationEmail(
-        { loggedOutIntent: this.state!.loggedOut!.verification!.intent })
+        { loggedOutIntent: this.state.loggedOut.verification.intent })
     } catch (e) {
       if (e && typeof e === 'object') {
         error = e as ResendConfirmationEmailError
@@ -134,8 +135,9 @@ export class SettingsBraveAccountRowElement extends I18nMixinLit(CrLitElement) {
   }
 
   protected onCancelRegistrationButtonClicked() {
+    assert(this.state?.loggedOut?.verification)
     this.browserProxy.authentication.cancelVerification(
-      { loggedOutIntent: this.state!.loggedOut!.verification!.intent })
+      { loggedOutIntent: this.state.loggedOut.verification.intent })
   }
 
   protected openBraveAccountDialog() {
--- a/components/brave_account/brave_account_state_prefs.cc
+++ b/components/brave_account/brave_account_state_prefs.cc
@@ -6,12 +6,15 @@
 #include "brave/components/brave_account/brave_account_state_prefs.h"
 
 #include <optional>
+#include <utility>
 
 #include "base/check.h"
+#include "base/check_deref.h"
 #include "base/json/values_util.h"
 #include "base/time/time.h"
 #include "base/values.h"
 #include "brave/components/brave_account/brave_account_service_constants.h"
+#include "brave/components/brave_account/pref_names.h"
 #include "components/prefs/scoped_user_pref_update.h"
 
 namespace brave_account {
@@ -78,6 +81,11 @@ void AccountStatePrefs::SetLoggedIn(
                encrypted_authentication_token));
 }
 
+void AccountStatePrefs::ClearVerification() {
+  ScopedDictPrefUpdate(&*pref_service_, prefs::kBraveAccountState)
+      ->Remove(prefs::keys::kVerification);
+}
+
 // Firewall against tampered prefs: enforce on the way out the same
 // invariants the setters enforce on the way in, so downstream code can
 // trust the returned variant matches the underlying prefs without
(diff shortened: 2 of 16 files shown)


ios/brave-ios/Sources/Brave/Frontend/Settings/SettingsViewController.swift
```diff
@@ -487,7 +487,11 @@ class SettingsViewController: TableViewController, BraveAccountAuthenticationObs
                 rowUUID: braveAccountResendConfirmationEmailRowUUID,
                 sectionUUID: braveAccountSectionUUID
               )
-              braveAccountAuthentication.resendConfirmationEmail { [weak self] _, failure in
+              braveAccountAuthentication.resendVerificationEmail(
+                intent: BraveAccount.VerificationIntent(
+                  loggedOutIntent: braveAccountState.loggedOut!.verification!.intent
```
kylehickinson: Is both `loggedOut` and `verification` guaranteed to be non-nil in these instances? Otherwise may want to safely unwrap before calling `resendVerificationEmail` or provide a fallback

SergeyZhukovsky (approved):
android ++

components/brave_account/state_base.cc
```diff
@@ -71,13 +85,61 @@ void StateBase::RegisterVerify(const std::string&,
...
 
-void StateBase::ResendConfirmationEmail(
-    ResendConfirmationEmailCallback callback) {
-  std::move(callback).Run(
-      MakeCalledInWrongStateError<mojom::ResendConfirmationEmailError>());
+void StateBase::ResendVerificationEmail(
+    mojom::VerificationIntentPtr intent,
+    ResendVerificationEmailCallback callback) {
+  const auto encrypted_verification_token =
+      account_state_prefs_->GetVerificationToken(std::move(intent));
```
DJAndries, line 92: The "get and decrypt" stuff could be extracted into a helper method/function

components/brave_account/brave_account_state_prefs.cc
```diff
@@ -124,6 +132,38 @@ std::string AccountStatePrefs::GetAuthenticationToken() const {
...
 
+std::string AccountStatePrefs::GetVerificationToken(
+    mojom::VerificationIntentPtr intent) const {
+  CHECK(intent);
+
+  const bool has_matching_verification = [&] {
+    const auto account_state = GetAccountState();
+
+    switch (intent->which()) {
+      case mojom::VerificationIntent::Tag::kLoggedOutIntent:
```
DJAndries, line 143: is this the only use of the intent param added to the callers of this method? is it really necessary to validate the intent, or could we afford to assume that it is correct? Given that the server can handle tokens with invalid intents, it seems that we could forego this validation to simplify the interface and reduce overall complexity everywhere.

ios/brave-ios/Sources/Brave/Frontend/Settings/SettingsViewController.swift
szilardszaloki: Yes, the `where` clause above (`where braveAccountState.loggedOut!.verification != nil`) guarantees it.

browser/resources/settings/getting_started_page/brave_account_row.ts
```diff
@@ -101,7 +101,8 @@ export class SettingsBraveAccountRowElement extends I18nMixinLit(CrLitElement) {
     let error: ResendConfirmationEmailError | undefined
 
     try {
-      await this.browserProxy.authentication.resendConfirmationEmail()
+      await this.browserProxy.authentication.resendVerificationEmail(
+        { loggedOutIntent: this.state!.loggedOut!.verification!.intent })
```
fallaciousreasoning: are these guaranteeed to be defined? Possibly we should update the types to indicate that?

fallaciousreasoning: Looking at the type definition and this code, I don't really see why we can assert these will exist?

browser/resources/settings/getting_started_page/brave_account_row.ts
```diff
@@ -133,7 +134,8 @@ export class SettingsBraveAccountRowElement extends I18nMixinLit(CrLitElement) {
   }
 
   protected onCancelRegistrationButtonClicked() {
-    this.browserProxy.authentication.cancelRegistration()
+    this.browserProxy.authentication.cancelVerification(
+      { loggedOutIntent: this.state!.loggedOut!.verification!.intent })
```
fallaciousreasoning: ditto

components/brave_account/resources/brave_account_otp_dialog.ts
```diff
@@ -94,6 +98,7 @@ export class BraveAccountOtpDialogElement extends CrLitElement {
   private browserProxy: BraveAccountBrowserProxy =
     BraveAccountBrowserProxyImpl.getInstance()
 
+  protected accessor intent!: VerificationIntent
```
fallaciousreasoning, line 101: why do we need the `!` here? Isn't that implied?

components/brave_account/resources/brave_account_dialogs.html.ts
```diff
@@ -13,15 +13,15 @@ import './brave_account_sign_in_dialog.js'
 import { BraveAccountDialogsElement } from './brave_account_dialogs.js'
 
 export function getHtml(this: BraveAccountDialogsElement) {
-  switch (this.dialog) {
+  switch (this.dialog?.type) {
     case undefined:
       return nothing
```
fallaciousreasoning, line 18: Looking at the docs, returning `null` or `undefined` seems like it should work find for child elements:
https://lit.dev/docs/templates/conditionals/#conditionally-rendering-nothing

and its slightly more in line with what we do in other places. Sounds like the only place `nothing` is strictly needed is for attributes

fallaciousreasoning: I guess not changed in this PR, but worth considering for future

kylehickinson (approved):
ios++

components/brave_account/state_base.cc
szilardszaloki: Yeah, although it couldn't be a plain function (it would need to be a `template`), as the empty-token branch returns `kCalledInWrongState` as a different type per caller. Also, we couldn't shared the same logic with auth tokens, as there the empty-token branch is a `CHECK` (since it's guaranteed to be present in the `LoggedInState`).

components/brave_account/brave_account_state_prefs.cc
szilardszaloki: That's actually necessary, as the `intent` parameter validates the caller's assumption about the current account state. It's there to prevent scenarios where, for example, a stale renderer still assumes a logged-out state with a `kRegistration` intent after the account has transitioned to a logged-in state with a pending `kChangePassword` verification. Without this check, it could read the verification slot and mistakenly treat the change-password token as the registration token. The `intent` match is what eventually turns that into a `kCalledInWrongState` instead of silently using the wrong token.

ios/brave-ios/Sources/Brave/Frontend/Settings/SettingsViewController.swift
szilardszaloki: I went ahead and addressed it anyway 🙂: https://github.com/brave/brave-core/pull/36891/commits/80352ffad250180426d6f672ee8fbab29cbd6bea.

netzenbot (approved):

browser/resources/settings/getting_started_page/brave_account_row.ts
szilardszaloki: Both handlers are rendered only in the `this.state.loggedOut!.verification` branch (nested under `state !== undefined` and not-`LOGGED_IN`), so `state`/`loggedOut`/`verification` are all guaranteed when they fire. If you prefer, we could make them an explicit `assert` so a future miswiring fails loudly.

browser/resources/settings/getting_started_page/brave_account_row.ts
szilardszaloki: https://github.com/brave/brave-core/pull/36891#discussion_r3344623087

components/brave_account/resources/brave_account_otp_dialog.ts
szilardszaloki: `intent` has no initializer and its type is `VerificationIntent` (non-optional), so without the `!` you get
```
error TS2564: Property 'intent' has no initializer and is not definitely assigned in the constructor.
```
An alternative would be to make it `intent?: VerificationIntent`, but that would incorrectly imply that it's optional, as it's always provided.

components/brave_account/resources/brave_account_dialogs.html.ts
szilardszaloki: Yep, that's right, but Lit's guidance is to prefer `nothing` over other falsy values (https://lit.dev/docs/api/templates/#nothing):
> Prefer using `nothing` over other falsy values as it provides a consistent behavior between various expression binding contexts.

components/brave_account/resources/brave_account_otp_dialog.ts
fallaciousreasoning: Ah right - makes sense. Let's leave it as is

browser/resources/settings/getting_started_page/brave_account_row.ts
fallaciousreasoning: Let's add an assert :smile:

components/brave_account/resources/brave_account_dialogs.html.ts
fallaciousreasoning: Things like this are why I reckon React landed on a better abstraction :laughing:

fallaciousreasoning (approved):

browser/resources/settings/getting_started_page/brave_account_row.ts
szilardszaloki: `assert` it is: https://github.com/brave/brave-core/pull/36891/commits/11aa67aa2646dfc2a666f3c6f19786fdaac09b08

PR #36906, opened 2026-06-01
Fix iPhone Playlist full screen force not taking effect

### LINKED ISSUE #55984: Full screen not forced in landscape for Playlist
### Description

Playlist full screen issue in landscape

Slack: https://bravesoftware.slack.com/archives/C07LWRLAYH5/p1780302514126049

### Steps to reproduce

1. Play YT video
2. Add to playlist, open in playlist
3. switch to landscape

### Actual result

Image

### Expected result

Should force to full in landscape

### Reproduces how often

Easily reproduced

### Brave version

1.90.x

### Device/iOS version

iPhone 15 / iOS 18.6.2

### Affected browser versions

- [x] latest AppStore
- [x] latest TestFlight
- [ ] previous TestFlight

### Reproducibility

### Miscellaneous information

_No response_

EliHini:
Resolves https://github.com/brave/brave-browser/issues/55984

## Summary
On iPhone, rotating to landscape should force fullscreen mode. This was broken because `handleFullScreenOrientationChanges` derived whether `fullscreen` or `orientation` had changed by comparing an expected value against the environment. However, SwiftUI updates environment values before `.onChange` fires, so the comparison was always `false` and `toggleFullScreen` was never called.

The fix uses handler identity instead of value comparison: the `isFullScreen` handler passes `isFullScreenModeChanging: true`; the `interfaceOrientation` handler passes `isOrientationChanging: true`. The switch then reads the already-updated environment values directly.

This also affected Max-style iPhones (Plus/Max), which report `.regular` horizontal size class in landscape. Because `fullscreen` was never triggered, `PlaylistSplitView.sidebarLayoutMode` never hit the `if isFullScreen { return .none } ` early exit and fell through to the sidebar layout path. Once `fullscreen` is triggered correctly, this path is unreachable for iPhones.

iPad behaviour remains unchanged: landscape + regular width still uses the sidebar layout.

## Root cause

The broken comparison logic seems to have been in the to the original playlist UI ([#23425](https://github.com/brave/brave-core/pull/23425)). 

The `onChange` API migration ([#35164](https://github.com/brave/brave-core/pull/35164)) made the timing issue more consistent in 1.90.x.

## Test plan

| Device | Test Case | Expected result |
|--------|----------|-----------------|
| iPhone | Play a YouTube video, add to playlist, open from playlist, rotate to landscape while playing | Player forces fullscreen (no sidebar, toolbar, or windowed controls) |
| iPhone | Rotate back to portrait while in fullscreen | Exits fullscreen and returns to portrait layout |
| iPhone | Tap fullscreen button manually (non-portrait video) | Enters fullscreen and rotates to landscape |
| iPhone | Portrait-only video | Fullscreen toggle works without requiring landscape rotation |
| iPad | Same playlist flow in landscape | Windowed/sidebar layout still shown (unchanged) |
| iPad | Portrait | Bottom sheet sidebar still works |

### DIFF
--- a/ios/brave-ios/Sources/PlaylistUI/MediaContentView.swift
+++ b/ios/brave-ios/Sources/PlaylistUI/MediaContentView.swift
@@ -58,20 +58,16 @@ struct MediaContentView: View {
         .ignoresSafeArea()
         .allowsHitTesting(false)
     }
-    .onChange(of: isFullScreen) { oldValue, newValue in
+    .onChange(of: isFullScreen) {
       handleFullScreenOrientationChanges(
-        expectedFullScreen: newValue,
-        expectedOrientation: interfaceOrientation,
-        didFullScreenChange: oldValue != newValue,
-        didOrientationChange: false
+        isFullScreenModeChanging: true,
+        isOrientationChanging: false
       )
     }
-    .onChange(of: interfaceOrientation) { oldValue, newValue in
+    .onChange(of: interfaceOrientation) {
       handleFullScreenOrientationChanges(
-        expectedFullScreen: isFullScreen,
-        expectedOrientation: newValue,
-        didFullScreenChange: false,
-        didOrientationChange: oldValue != newValue
+        isFullScreenModeChanging: false,
+        isOrientationChanging: true
       )
     }
     .persistentSystemOverlays(isFullScreen ? .hidden : .automatic)
@@ -80,10 +76,8 @@ struct MediaContentView: View {
   }
 
   private func handleFullScreenOrientationChanges(
-    expectedFullScreen: Bool,
-    expectedOrientation: UIInterfaceOrientation,
-    didFullScreenChange: Bool,
-    didOrientationChange: Bool
+    isFullScreenModeChanging: Bool,
+    isOrientationChanging: Bool
   ) {
     if UIDevice.current.userInterfaceIdiom != .phone {
       // Full screen and orientations don't affect iPads
@@ -100,12 +94,12 @@ struct MediaContentView: View {
 
     let isPortraitVideo = model.isPortraitVideo
 
-    if !didFullScreenChange && !didOrientationChange {
+    if !isFullScreenModeChanging && !isOrientationChanging {
       // Nothing to do
       return
     }
 
-    switch (expectedFullScreen, expectedOrientation.isLandscape) {
+    switch (isFullScreen, interfaceOrientation.isLandscape) {
     case (true, false):
       // When a user taps full screen mode we specifically want to shift into landscape orientation
       // Likewise if the user was in full screen mode already and is no longer in landscape
@@ -113,9 +107,9 @@ struct MediaContentView: View {
       // since potrait video can display fullscreen without being in landscape
       if !isPortraitVideo {
         ignoreFollowingFullScreenOrientationChange = true
-        if didFullScreenChange {
+        if isFullScreenModeChanging {
           requestGeometryUpdate(orientation: .landscapeLeft)
-        } else if didOrientationChange {
+        } else if isOrientationChanging {
           toggleFullScreen(explicitFullScreenMode: false)
         }
       }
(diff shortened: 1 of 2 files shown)


ios/brave-ios/Sources/PlaylistUI/PlaylistSplitView.swift
```diff
@@ -58,7 +58,12 @@ struct PlaylistSplitView<Sidebar: View, SidebarHeader: View, Content: View, Tool
     //
     // This is not actually an inverse of whether or not the bottom drawer is visible, since
     // on iPhone in landscape, we show neither sidebar or drawer.
-    if horizontalSizeClass == .regular && interfaceOrientation.isLandscape {
+    // iPhone landscape uses a regular horizontal size class but should force fullscreen
+    // (see MediaContentView), not the iPad sidebar layout.
+    if UIDevice.current.userInterfaceIdiom == .pad,
+      horizontalSizeClass == .regular,
+      interfaceOrientation.isLandscape
+    {
```
kylehickinson: This logic is not needed, we explicitly trigger fullscreen on Max-style phones that can return regular horizontal size classes

ios/brave-ios/Sources/PlaylistUI/MediaContentView.swift
```diff
@@ -58,16 +58,20 @@ struct MediaContentView: View {
...
     }
-    .onChange(of: isFullScreen) { _, newValue in
+    .onChange(of: isFullScreen) { oldValue, newValue in
       handleFullScreenOrientationChanges(
         expectedFullScreen: newValue,
-        expectedOrientation: interfaceOrientation
+        expectedOrientation: interfaceOrientation,
+        didFullScreenChange: oldValue != newValue,
+        didOrientationChange: false
       )
```
kylehickinson, line 65: Lets clean this up, given the new logic:

1. Drop the 2-argument version of `onChange`, if the state in the closure is run in the new environment then we can simply just read the value and use the 0-arg version
2. `oldValue != newValue` is redundant if we're hitting the `onChange` it means the value has changed, so this can just be simplified to `true`
3. Lets use the original naming in `handleFullScreenOrientationChanges` for the new args so they can replace them entirely: `isFullScreenModeChanging`/`isOrientationChanging`

Example of `isFullScreen` change:

```swift
.onChange(of: isFullScreen) {
  handleFullScreenOrientationChanges(
    expectedFullScreen: isFullScreen,
    expectedOrientation: interfaceOrientation
    isFullScreenModeChanging: true,
    isOrientationChanging: false
}
```

ios/brave-ios/Sources/PlaylistUI/MediaContentView.swift
```diff
@@ -77,7 +81,9 @@ struct MediaContentView: View {
 
   private func handleFullScreenOrientationChanges(
     expectedFullScreen: Bool,
-    expectedOrientation: UIInterfaceOrientation
+    expectedOrientation: UIInterfaceOrientation,
+    didFullScreenChange: Bool,
+    didOrientationChange: Bool
```
kylehickinson: As mentioned above, rename these to `isFullScreenModeChanging` and `isOrientationChanging`, then remove all the changes below (except for deletion of local vars)

kylehickinson (changes requested):
New logic is sound, just needs a bit of cleanup and removal of change in `MediaContentView`

ios/brave-ios/Sources/PlaylistUI/MediaContentView.swift
kylehickinson: Actually after looking again, we can simplify this even further by removing `expectedFullScreen`/`expectedOrientation` and instead read those directly in `handleFullScreenOrientationChanges`.

EliHini: +1

netzenbot (approved):

kylehickinson (approved):

kylehickinson:
@EliHini Given this is a regression in [1.91](https://github.com/brave/brave-core/pull/35164) please uplift to release after merging

kjozwiak:
Verification **PASSED** on `iPhone 11` running `iOS 26.5` using the following build(s):

```
Brave | 1.93.24 Chromium: 149.0.7827.54 (Official Build)nightly (64-bit) 
--- | ---
Revision | f8cdea6c02b2...
OS | iOS
```

Using the STR/Cases outlined via https://github.com/brave/brave-browser/issues/55984#issue-4565350335, ensured that full screen is being forced when rotating the device/iPhone via `landscape` mode when watching a saved video via `Playlist` as per the following:

**`1.93.19 Chromium: 149.0.7827.54 (Broken)`** | **`1.93.24 Chromium: 149.0.7827.54 (Fixed)`**
-----------------------------------------------------|-----------------------------------------------------
IMG_1644 | <video src="https://github.com/user-attachments/assets/763ae94b-29aa-4c7e-9cfa-8a9b15a43052" />

PR #36907, opened 2026-06-02
[toolchain] Use hash checks to install the Xcode toolchain

cdesouza-chromium:
This PR adds a hash constant to the Xcode toolchain script, similar to
how upstream `MAC_BINARIES_TAG` on their counterpart of this source.
With this change we check the hash sum when downloading the file to make
sure it matches the expected hash.

Prior to this change, the decision to download/extract a new SDK was
entirely based on the version of the SDK found in the destination path.
This has served us well, however is a weaker scheme for reproducibility
checks. With this change, it is being introduced a scheme, similar to
how this type of check is done in other places, where a sidecar file is
created based on the name of the toolchain to be downloaded, and that
sidecar file has a hashsum in it.

With this new verification scheme we can now detect accidental
overwrites of the toolchain, but we can also leave the door open to use
issue a new toolchain under a subversion in case of some emergency fix
for a toolchain file

Finally, this change also corrects a few things across this source that
are outdated, and do not conform to the python style guide currently
enforced in Chromium, in particular the naming convention that was being
used.

### DIFF
--- a/build/mac/download_hermetic_xcode.py
+++ b/build/mac/download_hermetic_xcode.py
@@ -34,6 +34,9 @@ sys.path.append(str(Path(__file__).resolve().parents[2] / 'script'))
 
 import deps  # pylint: disable=wrong-import-position
 
+# The hash sum for the archive expected to be downloaded.
+MAC_BINARIES_HASH = '79993cddfb0405fd39c582c88f3982f91f92c1b14c5881649bfa07aea873d1e4'
+
 # This contains binaries from Xcode 26.4 (17E202) along with the macOS 26.4 SDK
 # (25E251) and the Metal toolchain (17E188).
 XCODE_VERSION = '26.4.1'
@@ -49,16 +52,23 @@ XCODE_TOOLCHAIN_DOWNLOAD_URL = (
 # work.
 MAC_MINIMUM_OS_VERSION = [19, 4]
 
-MAC_TOOLCHAIN_ROOT = Path(
-    __file__).resolve().parents[3] / 'build' / 'mac_files'
+# Destination for the hermetic Xcode binaries to be extracted at.
+MAC_BINARIES_ROOT = Path(
+    __file__).resolve().parents[3] / 'build' / 'mac_files' / 'xcode_binaries'
+
+# The name for the hash file use to track the installation of the toolchain.
+DOWNLOADED_TOOLCHAIN_HASH = (
+    MAC_BINARIES_ROOT /
+    f'.{os.path.basename(XCODE_TOOLCHAIN_DOWNLOAD_URL).replace(".", "_")}_hash'
+)
 
 
-def LoadPList(path: Path) -> dict:
-    """Loads Plist at |path| and returns it as a dictionary."""
+def _load_plist(path: Path) -> dict:
+    """Loads the plist at path and returns it as a dictionary."""
     return plistlib.loads(path.read_bytes())
 
 
-def PlatformMeetsHermeticXcodeRequirements() -> bool:
+def _platform_meets_hermetic_xcode_requirements() -> bool:
     if sys.platform == 'darwin':
         needed = MAC_MINIMUM_OS_VERSION
         major_version = [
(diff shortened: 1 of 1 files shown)


build/mac/download_hermetic_xcode.py
```diff
@@ -68,78 +73,122 @@ def PlatformMeetsHermeticXcodeRequirements() -> bool:
...
+def _install_xcode_binaries() -> int:
     """Installs the Xcode binaries needed to build Brave and accepts the
     license."""
-    binaries_root = MAC_TOOLCHAIN_ROOT / 'xcode_binaries'
-
-    # Tarball extraction or not, if we have a hermetic toolchain,we still want
-    # to process the license if the version is newer than the currently
-    # accepted one.
-    if (XCODE_VERSION != GetHermeticXcodeVersion(binaries_root)
-            or binaries_root.is_symlink()):
```
cdesouza-chromium, line 91: I've removed this `is_symlink` because there was no indication it was being used, and I was hoping to find out more about it during review.

build/mac/download_hermetic_xcode.py
```diff
@@ -68,78 +73,122 @@ def PlatformMeetsHermeticXcodeRequirements() -> bool:
...
+    if not sdk_installer.is_installed():
         try:
-            deps.DownloadAndUnpack(url, binaries_root)
+            sdk_installer.install()
         except URLError:
-            print(f"Failed to download hermetic Xcode: {url}")
-            print("Exiting.")
+            print(f'Failed to download hermetic Xcode: '
+                  f'{XCODE_TOOLCHAIN_DOWNLOAD_URL}')
+            print('Exiting.')
```
cdesouza-chromium: I suspect we could just move this too inside `install`, print the failure message, then re-raise. Raising this here for input.

build/mac/download_hermetic_xcode.py
```diff
@@ -68,78 +73,122 @@ def PlatformMeetsHermeticXcodeRequirements() -> bool:
...
+    @classmethod
+    def create(cls) -> SdkInstaller:
+        """Builds an SdkInstaller instance for the current SDK.
+
+        The sidecar file is named based on the URL name for it, with a few
+        transformations to make it less conspicuous.
+        """
+        tarball_name = XCODE_TOOLCHAIN_DOWNLOAD_URL.rsplit('/', 1)[-1]
+        file_prefix = tarball_name.replace('/', '_').replace('.', '_')
+        return cls(MAC_BINARIES_ROOT / f'.{file_prefix}_hash')
```
goodov: is `_hash_file` set here implicitly? that's not something I used to see. can you make it explicit?

why go with `create()` at all? this seems to complicate things for no reason. that's not a pure dataclass, I think you should go with a normal class and just initialize it in `__init__`.

build/mac/download_hermetic_xcode.py
goodov: `is_symlink` means the toolchain is symlinked into `/Applications/Xcode.app`, so it's not hermetic by nature and can be updated at any point. In case of a symlink we want the hermetic toolchain to be installed explicitly even if the symlinked toolchain is acceptable.

build/mac/download_hermetic_xcode.py
goodov: yeah, reraising should be better here.

build/mac/download_hermetic_xcode.py
cdesouza-chromium: The reason I used a frozen data class is because that easily tells you that the member data never changes throughout the objects life, but I’ll change to your suggestion as maybe there isn’t that much of gain in this case.

netzenbot (approved):

build/mac/download_hermetic_xcode.py
cdesouza-chromium: Fixed. When a symlink is detected we return `False` for `is_installed`.

build/mac/download_hermetic_xcode.py
cdesouza-chromium: Fixed. That's much cleaner.

build/mac/download_hermetic_xcode.py
cdesouza-chromium: Fixed.

build/mac/download_hermetic_xcode.py
```diff
@@ -68,78 +72,123 @@ def PlatformMeetsHermeticXcodeRequirements() -> bool:
...
+            return False
+        url = XCODE_TOOLCHAIN_DOWNLOAD_URL
+        print(f'Downloading hermetic Xcode: {url}')
+        try:
+            deps.DownloadAndUnpack(url,
+                                   MAC_BINARIES_ROOT,
+                                   sha256=MAC_BINARIES_HASH)
+        except URLError:
+            print(f'Failed to download hermetic Xcode: {url}')
+            print('Exiting.')
```
goodov: nit: no need for `Exiting.` line imo.

build/mac/download_hermetic_xcode.py
```diff
@@ -68,78 +72,123 @@ def PlatformMeetsHermeticXcodeRequirements() -> bool:
...
+            raise
+        self._hash_file.write_text(MAC_BINARIES_HASH + '\n',
```
goodov: `self._hash_file.write_text(MAC_BINARIES_HASH)`

build/mac/download_hermetic_xcode.py
```diff
@@ -68,78 +72,123 @@ def PlatformMeetsHermeticXcodeRequirements() -> bool:
...
+        """Returns True if the sidecar exists and has the expected hash.
+
+        A symlinked destination is never treated as installed, so the caller
+        always redeploys a real directory on top of it.
+        """
+        if MAC_BINARIES_ROOT.is_symlink():
+            return False
+        if not self._hash_file.is_file():
+            return False
+        recorded = self._hash_file.read_bytes().decode('utf-8').rstrip()
```
goodov: `installed_hash = self._hash_file.read_text()`

build/mac/download_hermetic_xcode.py
```diff
@@ -68,78 +72,123 @@ def PlatformMeetsHermeticXcodeRequirements() -> bool:
...
+
+        Returns True if the archive was downloaded and extracted, or False if
+        the expected hash was already in place and nothing needed to be done.
+        Also writes the expected hash to the sidecar file, which is used to
+        detect the current install. On a download failure it prints a
+        diagnostic and re-raises.
+        """
+        if self.is_installed():
+            return False
+        url = XCODE_TOOLCHAIN_DOWNLOAD_URL
```
goodov: nit: use `XCODE_TOOLCHAIN_DOWNLOAD_URL` directly, no need for the var.

build/mac/download_hermetic_xcode.py
```diff
@@ -68,78 +72,123 @@ def PlatformMeetsHermeticXcodeRequirements() -> bool:
...
+    present in the destination.
+    """
+
+    def __init__(self) -> None:
+        # The sidecar file is named based on the URL name for it, with a few
+        # transformations to make it less conspicuous.
+        tarball_name = XCODE_TOOLCHAIN_DOWNLOAD_URL.rsplit('/', 1)[-1]
+        file_prefix = tarball_name.replace('/', '_').replace('.', '_')
+        # The path to the sidecar file with the hash for the archive used.
+        self._hash_file = MAC_BINARIES_ROOT / f'.{file_prefix}_hash'
```
goodov: just put this at the top as
```python
HASH_FILE = MAC_BINARIES_ROOT / f'.{os.path.basename(XCODE_TOOLCHAIN_DOWNLOAD_URL)}.sha256'
```
?
not sure why all these replacements are needed..

goodov (approved):

build/mac/download_hermetic_xcode.py
cdesouza-chromium: I want to keep these replacements because I don't want someone naively confusing these stamps when using some script to read file from a folder, and I also think it gives the wrong impression to see `x.tar.gz.sha256` as that feels to me like another aspect of the compression itself, so I would rather keep the `_` transformation to make it very easy to conclude that this is not a derivative.

I have created the constant though as you requested.

For the record, a picture is slowly starting to emerge of which direction some of this stuff is gonna take, so long term I've pondering to move all these hashing checks, etc, into a single place, and that's potentially `build/download_dep.py`. I'm just not rushing to unify everything I'm doing at once, to not get myself bogged down with making everything consistent.

build/mac/download_hermetic_xcode.py
cdesouza-chromium: Done

build/mac/download_hermetic_xcode.py
cdesouza-chromium: Done

build/mac/download_hermetic_xcode.py
cdesouza-chromium: Done

build/mac/download_hermetic_xcode.py
cdesouza-chromium: Done

build/mac/download_hermetic_xcode.py
cdesouza-chromium: @goodov I'm merging this now, as you are on PTO. Feel free to reply on this thread and I will absolutely follow up on any requests you put here, if you have any thoughts of your own about my explanation above.

goodov: Ah, I see where you're going. I think if we want to generalize this, then it makes sense to have a metadata file that will keep the URL, the timestamp when it was created (downloaded) and the hash.

The metadata file should have a fixed name that will be used by the shared deps downloader, something like `.dep_metadata.json`.

In this exact PR I wanted to rename the hash file into a generic name, but I decided to not ask for this, because seeing what the hash is for is valuable bit of information. With the metadata file we can go for a generalized name, because the name/url will be inside and it will be even better to connect the dots in case of issues.

PR #36912, opened 2026-06-02
Remove #define macros from browser_commands.cc

### LINKED ISSUE #55997: Remove #define macro from chromium_src/chrome/browser/ui/browser_commands.cc

sangwoo108:
As #define macros are fragile and error-prone, this commit removes them from browser_commands.cc and replaces them plaster and patching.

Resolves https://github.com/brave/brave-browser/issues/55997

### DIFF
--- a/chromium_src/chrome/browser/ui/browser_commands.cc
+++ b/chromium_src/chrome/browser/ui/browser_commands.cc
@@ -12,23 +12,18 @@
 #include "chrome/browser/ui/browser_commands.h"
 #include "chrome/common/webui_url_constants.h"
 
-#define ReloadBypassingCache ReloadBypassingCache_ChromiumImpl
-#define GetReadingListModel GetReadingListModel_ChromiumImpl
-#define kChromeUISplitViewNewTabPageURL kChromeUINewTabURL
-
-// Don't show toast when user tries to close a pinned tab via the keyboard
-// accelerator.
-#define BRAVE_CLOSE_TAB toast_controller = nullptr;
-
-#include <chrome/browser/ui/browser_commands.cc>
-
-#undef BRAVE_CLOSE_TAB
-#undef kChromeUISplitViewNewTabPageURL
-#undef ReloadBypassingCache
-#undef GetReadingListModel
+class ReadingListModel;
 
 namespace chrome {
 
+// Upstream function, renamed by rewrite rule.
+void ReloadBypassingCache_ChromiumImpl(BrowserWindowInterface* browser,
+                                       WindowOpenDisposition disposition);
+
+// We override upstream's ReloadBypassingCache_ChromiumImpl with this function.
+// This is needed to handle the case where the profile is a Tor profile,
+// in which case we want to trigger a new Tor connection instead of a regular
+// hard reload.
 void ReloadBypassingCache(BrowserWindowInterface* browser,
                           WindowOpenDisposition disposition) {
 #if BUILDFLAG(ENABLE_TOR)
@@ -43,8 +38,6 @@ void ReloadBypassingCache(BrowserWindowInterface* browser,
   ReloadBypassingCache_ChromiumImpl(browser, disposition);
 }
 
-ReadingListModel* GetReadingListModel(Browser* browser) {
-  return nullptr;
-}
-
 }  // namespace chrome
+
+#include <chrome/browser/ui/browser_commands.cc>
--- a/rewrite/chrome/browser/ui/browser_commands.cc.yaml
+++ b/rewrite/chrome/browser/ui/browser_commands.cc.yaml
@@ -1,0 +1,33 @@
+# Copyright (c) 2026 The Brave Authors. All rights reserved.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+substitutions:
+  - description:
+      Rename ReloadBypassingCache to ReloadBypassingCache_ChromiumImpl to avoid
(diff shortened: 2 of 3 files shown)


chromium_src/chrome/browser/ui/browser_commands.cc
```diff
@@ -6,29 +6,23 @@
 #include "brave/browser/ui/browser_commands.h"
 
 #include "base/check.h"
+#include "base/logging.h"
```
netzenbot: nit: `base/logging.h` doesn't appear to be used — neither the Brave code here nor the upstream `browser_commands.cc` uses `LOG()`, `DLOG()`, or any other logging macro from that header. Per [IWYU](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-001), remove includes you don't actually use.

chromium_src/chrome/browser/ui/browser_commands.cc
```diff
@@ -43,8 +37,6 @@ void ReloadBypassingCache(BrowserWindowInterface* browser,
   ReloadBypassingCache_ChromiumImpl(browser, disposition);
 }
 
-ReadingListModel* GetReadingListModel(Browser* browser) {
-  return nullptr;
-}
-
 }  // namespace chrome
+
```
netzenbot, line 42: nit: The comment says "We override upstream's `ReloadBypassingCache_ChromiumImpl` with this function", but it's placed above the declaration of `ReloadBypassingCache_ChromiumImpl` — making "this function" refer to the wrong thing. This comment would read more clearly if it were moved above `ReloadBypassingCache` (the actual Brave override), and the declaration of `ReloadBypassingCache_ChromiumImpl` had a simpler note like "// Upstream function, renamed by rewrite rule.". Per [CS-019](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-019), comments should be unambiguous to future readers.

sangwoo108: done

chromium_src/chrome/browser/ui/browser_commands.cc
sangwoo108: done

netzenbot (approved):

rewrite/chrome/browser/ui/browser_commands.cc.yaml
```diff
@@ -0,0 +1,33 @@
...
+      conflict with Brave implementation
+    re_pattern: (void ReloadBypassingCache)\(
+    replace: \1_ChromiumImpl(
+
+  - description:
+      Return nullptr in GetReadingListModel(). We don't want ReadingList panel
+      in the side panel.
+    re_pattern: (ReadingListModel\* GetReadingListModel\(.*?)(return model;)
+    re_flags: [DOTALL]
+    replace: \1model = nullptr;\n  \2
```
cdesouza-chromium: if the intent here is to always return `nullptr` for `GetReadingListModel`, just add `if ((true)) return nullptr;` as the first line of `GetReadingListModel`.

rewrite/chrome/browser/ui/browser_commands.cc.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+  - description:
+      Replace chrome::kChromeUISplitViewNewTabPageURL to kChromeUINewTabURL in
+      NewSplitTab(). We don't want to have a separate NTP URL for split view.
+    re_pattern: |-
+      (void NewSplitTab\(.*?const GURL new_tab_url =.*?)GURL\(chrome::kChromeUISplitViewNewTabPageURL\);
```
cdesouza-chromium: This match should be more generic. It is okay to constrain the match to scope of `NewSplitTab`, but let's just replace any occurrence `kChromeUINewTabURL` inside `NewSplitTab`, and not match with anything else inside that function.

rewrite/chrome/browser/ui/browser_commands.cc.yaml
sangwoo108: Done.

rewrite/chrome/browser/ui/browser_commands.cc.yaml
sangwoo108: Thanks, done. Used MULTILINE flag to match the closing bracket

cdesouza-chromium (approved):
LGTM

rewrite/chrome/browser/ui/browser_commands.cc.yaml
```diff
@@ -0,0 +1,37 @@
...
+      NewSplitTab(). We don't want to have a separate NTP URL for split view.
+    re_pattern: |-
+      (void NewSplitTab\(.*?)(chrome::kChromeUISplitViewNewTabPageURL)(.*?^})
+    re_flags: [DOTALL, MULTILINE]
+    replace: \1chrome::kChromeUINewTabURL\3
+
+  - description:
+      Nullify toast controller in CloseTab() not to show toast when closing a
+      pinned tab
+    re_pattern: (void CloseTab\(.*?toast_controller\(\);)
```
cdesouza-chromium, line 35: This last one, I'm just wondering if we should just replace `browser- >GetFeatures().toast_controller()` indiscriminately with `nullptr`?

sangwoo108: IIUC, I think we shouldn't as we only suppress a few of toasts

```
bool ToastController::MaybeShowToast(ToastParams params) {
  if (params.toast_id == ToastId::kLinkCopied ||
      params.toast_id == ToastId::kImageCopied ||
      params.toast_id == ToastId::kAddedToReadingList) {
    return false;
```

I believe we're still using toast for other cases.

PR #36913, opened 2026-06-02
[Lit]: Update docs with example of adding a reactive property

### LINKED ISSUE #55994: [lit]: No way to add a reactive property to an upstream class via chromium_src override (requires patching)

fallaciousreasoning:
Resolves https://github.com/brave/brave-browser/issues/55994

### DIFF
--- a/docs/webui_overriding.md
+++ b/docs/webui_overriding.md
@@ -81,6 +81,48 @@ export * from './fancy-chromium.js'
 See [this PR](https://github.com/brave/brave-core/pull/29598/files) for a real
 example.
 
+## Lit Properties
+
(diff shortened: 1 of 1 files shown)


netzenbot (approved):

petemill (approved):
Wouldn't this subclassing replacement _and_ declaration merging also be preferable in the previous functions example because there's at least _some_ chance that if we wire up typescript correctly then we'll have first class type support for the base class? Or would prototype extensions get that too?

fallaciousreasoning:
The prototype extensions will get that too, via the declaration merging - I think that's preferable than the subclassing approach for now because you don't need to do any patching.

PR #36914, opened 2026-06-02
Gracefully handle Omaha 4 update install timeout on macOS

mherrmann:
Browser updates on macOS with Omaha 4 sometimes run into timeouts. @LaurenWags reported a case where this bricked the browser. This PR makes it so the timeout gets handled more gracefully, and thus hopefully avoids bricking the browser.

# Test plan


cdesouza-chromium (approved):
LGTM

LaurenWags:
FYI - @mherrmann looks like you've got a test plan but no associated issue. If there's not an issue associated this will not get verified by QA (current process is that QA verifies issues, not PRs). I only noticed because I was tagged.

cc @kjozwiak @rebron

netzenbot (approved):

PR #36917, opened 2026-06-02
Add workspace button and UI for saving/loading/deleting workspaces

### LINKED ISSUE #55108: Workspaces - Basic save and load dialogs (phase 1.5)
## Description
Direct follow up to https://github.com/brave/brave-browser/issues/11096

Having a UI on top of [phase 1 work](https://github.com/brave/brave-browser/issues/11096) where a person can specify the name of the workspace and can save multiple workspaces. There will be a load workspace interface and folks can delete workspaces too.

### LINKED ISSUE #55869: Provide a way to save/load/delete workspaces
## Description
With workspaces phase 1 landed, it would be great to extend the functionality.
Currently, the workspace is hardcoded as `example-workspace`. This change would let folks specify their own name and then also load other workspaces.

While sessions might not be the long-term way to handle multiple workspaces (can't switch quickly between them), we do want to provide a way for folks to backup a given workspace.

## Example screens
These are some crude mock-ups. But solving this issue should offer a way to handle the following situations
### Saving a workspace
Image

### Loading a workspace
Image

### Deleting a workspace
Image

bsclifton:
Thanks to the automated reviews by Copilot and @netzenbot!

I think this is properly ready for review now

Fixes https://github.com/brave/brave-browser/issues/55108
Fixes https://github.com/brave/brave-browser/issues/55869

When brave://flags/#brave-workspace is enabled, this will add a Spaces button to the UI.

Regular mode
image

Not implemented w/ Vertical tabs yet. Will be added as follow-up.
See https://github.com/brave/brave-browser/issues/56728

When clicked, you can save your current tabs as a workspace OR load an existing workspace
image

You can click the kabob (...) menu to delete the workspace
image

Delete has a confirmation
image

If you don't have a workspace yet and you hit the Spaces button, you'll be prompted to make one.
Note that the `Settings` button is intentionally disabled. This will link into settings for workspace (which doesn't exist yet).
image


browser/ui/views/workspaces/workspace_row_view.h
```diff
@@ -0,0 +1,56 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#ifndef BRAVE_BROWSER_UI_VIEWS_WORKSPACES_WORKSPACE_ROW_VIEW_H_
+#define BRAVE_BROWSER_UI_VIEWS_WORKSPACES_WORKSPACE_ROW_VIEW_H_
+
+#include <memory>
+
```
netzenbot, line 10: nit: `WorkspaceMetadata` is only used by const-reference in the declarations here, so you could forward-declare it (`struct WorkspaceMetadata;`) and move `#include "brave/browser/workspaces/workspace_metadata.h"` to the .cc. ([Use Forward Declarations in Headers, Include in .cc](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-014))

browser/ui/views/frame/vertical_tabs/vertical_tab_strip_region_view.cc
```diff
@@ -312,6 +403,13 @@ BraveVerticalTabStripRegionView::BraveVerticalTabStripRegionView(
...
 
+  if (base::FeatureList::IsEnabled(features::kWorkspaces)) {
+    workspaces_button_ = AddChildView(
+        std::make_unique<VerticalTabWorkspacesButton>(base::BindRepeating(
+            &BraveVerticalTabStripRegionView::OnWorkspacesButtonPressed,
+            base::Unretained(this))));
+  }
+
   resize_area_ = AddChildView(std::make_unique<ResettableResizeArea>(this));
   SetBackground(views::CreateSolidBackground(kColorToolbar));
```
netzenbot, line 317: nit: `icon_` and `label_` are always set in the constructor (via `AddChildView`), so the `= nullptr` initializers are misleading per CS-054. `WorkspaceInfoButton` in this PR already omits them on its label members for the same reason. See https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-054

app/brave_generated_resources.grd
```diff
@@ -256,21 +256,105 @@
...
+          Cancel
+        </message>
+        <message name="IDS_WORKSPACE_ROW_MORE_BUTTON_TOOLTIP" desc="Tooltip for the per-row button that opens the workspace's options menu">
+          Workspace options
+        </message>
+        <message name="IDS_WORKSPACE_ROW_MENU_DELETE_WORKSPACE" desc="Context menu item that deletes the workspace shown in this row">
+          Delete Space
+        </message>
+        <message name="IDS_WORKSPACE_DELETE_CONFIRM_TITLE" desc="Title of the confirmation dialog shown before deleting a workspace">
+          Delete space
```
netzenbot, line 337: nit: The feature term "Space" is capitalized inconsistently within the sentence-case block: "Delete space" (here) vs "Delete Space" (IDS_WORKSPACE_ROW_MENU_DELETE_WORKSPACE) for the same delete action, plus "No spaces created" lowercase but "Enter your Space name" / "delete this Space" capitalized mid-sentence. Pick one convention for "space(s)" across these strings. ([Capitalization Consistency Across Feature Strings](https://github.com/brave/brave-core/tree/master/docs/best-practices/localization.md#L10N-009))

netzenbot (approved):

app/brave_generated_resources.grd
mkarolin, line 1: We seem to be using the terms `workspaces` and `spaces` interchangeably in the UI. Should we just pick one and stick to it everywhere?

goodov (approved):
chromium_src lgtm

app/brave_generated_resources.grd
bsclifton: Maybe @aguscruiz can help with this one! 😄 I think the full feature name is workspaces, but you're right - the messaging seems to be a bit mixed. I think we want to use `spaces` for everywhere that a customer sees?

@aguscruiz I'll do some edits and make sure `spaces` is preferred in the UI

aguscruiz: Yeah I'd go with `Spaces`. I think the `Work` part of `Workspaces` is pretty loaded and I'd prefer to go simpler with just spaces.

chromium_src/ui/views/bubble/bubble_dialog_delegate_view.h
```diff
@@ -29,6 +30,7 @@ class BraveBubbleDialogDelegateView;
   friend class ::BraveNewsBubbleView;                  \
   friend class ::BraveHelpBubbleDelegateView;          \
   friend class ::WaybackMachineBubbleView;             \
+  friend class ::WorkspacesBubbleView;                 \
```
simonhong: We should not add new class here
`BubbleDialogDelegateView` is deprecated.

chromium_src/ui/views/window/dialog_delegate.h
```diff
@@ -36,6 +37,7 @@ class BraveVpnDnsSettingsNotificiationDialogView;
 #define CreatePassKey                                                 \
   CreatePassKey_Unused();                                             \
   friend class ::CrashReportPermissionAskDialogView;                  \
+  friend class ::SaveWorkspaceDialog;                                 \
```
simonhong: Ditto - `DialogDelegateView` is deprecated.

browser/ui/views/workspaces/workspaces_bubble_view.cc
```diff
@@ -0,0 +1,169 @@
...
+
+  SetLayoutManager(std::make_unique<views::BoxLayout>(
+      views::BoxLayout::Orientation::kVertical,
+      /*inside_border_insets=*/gfx::Insets(),
+      /*between_child_spacing=*/0));
+
+  auto* service = WorkspaceServiceFactory::GetForProfile(browser->profile());
+  if (!service) {
+    return;
+  }
```
sangwoo108: I think this bubble shouldn't be created without service. So `CHECK(service)` would be better here.

browser/ui/views/workspaces/workspaces_bubble_view.cc
```diff
@@ -0,0 +1,169 @@
...
+  auto* workspaces_container = AddChildView(std::make_unique<views::View>());
+  workspaces_container->SetLayoutManager(std::make_unique<views::BoxLayout>(
+      views::BoxLayout::Orientation::kVertical));
+  workspaces_container->SetProperty(views::kMarginsKey, gfx::Insets(2));
+  if (workspaces.empty()) {
+    auto* empty_title =
+        workspaces_container->AddChildView(std::make_unique<views::Label>(
+            l10n_util::GetStringUTF16(IDS_WORKSPACES_BUBBLE_EMPTY_TITLE)));
+    empty_title->SetFontList(
+        empty_title->font_list().DeriveWithSizeDelta(2).DeriveWithWeight(
```
sangwoo108: Passing relative value(2) could be fragile, when the default font list set to the Label is changed. So referencing the current font size would be more robust.

```
auto current_fontlist = empty_title->font_list();
empty_title->SetFontList(current_fontlist.DeriveWithSizeDelta({target font size} - current_fontlist.GetFontSize())...
```

browser/ui/views/workspaces/workspaces_bubble_view.cc
```diff
@@ -0,0 +1,169 @@
...
+  button_row_layout->SetFlexForView(new_space_button, 1);
+}
+
+void WorkspacesBubbleView::OnSaveClicked() {
+  SaveWorkspaceDialog::Show(browser_);
+  GetWidget()->Close();
+}
+
+void WorkspacesBubbleView::OnWorkspaceSelected(const std::string& name) {
+  if (auto* service =
```
sangwoo108: Ditto: this bubble shouldn't be visible without the service. If we `CHECK(service)` from the constructor, we don't need this condition.

browser/ui/views/workspaces/workspaces_bubble_view.cc
```diff
@@ -0,0 +1,169 @@
...
+          WorkspaceServiceFactory::GetForProfile(browser_->profile())) {
+    service->RestoreWorkspace(name);
+  }
+  GetWidget()->Close();
+}
+
+void WorkspacesBubbleView::OnDeleteClicked(const std::string& name) {
+  auto* service = WorkspaceServiceFactory::GetForProfile(browser_->profile());
+  if (!service) {
```
sangwoo108: ditto

browser/ui/views/workspaces/workspaces_bubble_view.h
```diff
@@ -0,0 +1,44 @@
...
+
+  WorkspacesBubbleView(views::View* anchor_view, Browser* browser);
+  ~WorkspacesBubbleView() override;
+
+ private:
+  void OnSaveClicked();
+  void OnWorkspaceSelected(const std::string& name);
+  void OnDeleteClicked(const std::string& name);
+
+  raw_ptr<Browser> browser_;
```
sangwoo108: optional: I would prefer raw_ref<Browser> to raw_ptr as Browser should not be null.

browser/ui/views/workspaces/workspace_row_view.h
```diff
@@ -0,0 +1,57 @@
...
+                         public ui::SimpleMenuModel::Delegate {
+  METADATA_HEADER(WorkspaceRowView, views::View)
+ public:
+  WorkspaceRowView(const WorkspaceMetadata& info,
+                   WorkspaceRowClickedCallback on_workspace_selected,
+                   WorkspaceRowClickedCallback on_delete_clicked);
+  ~WorkspaceRowView() override;
+
+  void SetSelected(bool selected);
+  void OnMouseEntered(const ui::MouseEvent& event) override;
```
sangwoo108, line 60: nit: for consistency
```suggestion
  void SetSelected(bool selected);

  // views::View:
  void OnMouseEntered(const ui::MouseEvent& event) override;
```

browser/ui/views/workspaces/workspace_row_view.cc
```diff
@@ -0,0 +1,205 @@
...
+        views::BoxLayout::Orientation::kVertical,
+        /*inside_border_insets=*/gfx::Insets::VH(4, 8),
+        /*between_child_spacing=*/2));
+    layout->set_cross_axis_alignment(
+        views::BoxLayout::CrossAxisAlignment::kStart);
+
+    name_label_ = AddChildView(
+        std::make_unique<views::Label>(base::UTF8ToUTF16(info.name)));
+    name_label_->SetFontList(
+        name_label_->font_list().DeriveWithSizeDelta(2).DeriveWithWeight(
```
sangwoo108: ditto - it'd be better to reference the current font size.

browser/ui/views/workspaces/workspace_row_view.cc
```diff
@@ -0,0 +1,205 @@
...
+
+WorkspaceRowView::WorkspaceRowView(
+    const WorkspaceMetadata& info,
+    WorkspaceRowClickedCallback on_workspace_selected,
+    WorkspaceRowClickedCallback on_delete_clicked)
+    : on_delete_(std::move(on_delete_clicked)) {
+  SetNotifyEnterExitOnChild(true);
+
+  auto row_box = std::make_unique<views::BoxLayout>(
+      views::BoxLayout::Orientation::kHorizontal, gfx::Insets(), 0);
```
sangwoo108: I think gfx::Insts() and 0 already are default argument, so we can omit them for simplicity
```suggestion
  auto row_box = std::make_unique<views::BoxLayout>(
      views::BoxLayout::Orientation::kHorizontal);
```

browser/ui/views/workspaces/workspace_row_view.cc
```diff
@@ -0,0 +1,205 @@
...
+  row_box->set_cross_axis_alignment(
+      views::BoxLayout::CrossAxisAlignment::kCenter);
+  auto* row_layout = row_box.get();
+
+  SetLayoutManager(std::move(row_box));
```
sangwoo108: We can use chaining here so that we can do without a pair of unique and raw ptr variables.

```
auto* row_layout = SetLayoutManager(std::make_unique<BoxLayout>(...));
row_layout->set_cross_axis_alignmenet(...)
```

browser/ui/views/workspaces/workspace_row_view.cc
```diff
@@ -0,0 +1,205 @@
...
+
+  auto info_btn = std::make_unique<WorkspaceInfoButton>(
+      std::move(on_workspace_selected), info);
+  auto* info_btn_raw = AddChildView(std::move(info_btn));
```
sangwoo108: ditto

browser/ui/views/workspaces/workspace_row_view.h
```diff
@@ -0,0 +1,57 @@
...
+#include "ui/views/view.h"
+
+struct WorkspaceMetadata;
+
+namespace views {
+class ImageButton;
+class MenuRunner;
+}  // namespace views
+
+using WorkspaceRowClickedCallback = base::RepeatingCallback<void()>;
```
sangwoo108: I think this alias should be inside the class, and we can simplify the name, like `WorkspaceRowView::RowClickedCallack`

browser/ui/views/workspaces/save_workspace_dialog.h
```diff
@@ -0,0 +1,52 @@
...
+  // views::DialogDelegateView:
+  ui::mojom::ModalType GetModalType() const override;
+  std::u16string GetWindowTitle() const override;
+  bool IsDialogButtonEnabled(ui::mojom::DialogButton button) const override;
+
+  // views::TextfieldController:
+  void ContentsChanged(views::Textfield* sender,
+                       const std::u16string& new_contents) override;
+
+  void OnAccept();
```
sangwoo108: Generally, non-overridden methods should be put above overridden methods

browser/ui/views/workspaces/save_workspace_dialog.cc
```diff
@@ -0,0 +1,110 @@
...
+
+void SaveWorkspaceDialog::ContentsChanged(views::Textfield* sender,
+                                          const std::u16string& new_contents) {
+  DialogModelChanged();
+}
+
+void SaveWorkspaceDialog::OnAccept() {
+  if (!name_field_) {
+    return;
+  }
```
sangwoo108: I think this shouldn't happen.

browser/ui/views/workspaces/save_workspace_dialog.cc
```diff
@@ -0,0 +1,110 @@
...
+  std::string name = base::UTF16ToUTF8(name_field_->GetText());
+  if (name.empty()) {
+    return;
+  }
+
+  auto* service = WorkspaceServiceFactory::GetForProfile(browser_->profile());
+  if (!service) {
+    return;
+  }
```
sangwoo108: ditto - service should exist when this dialog shows up

browser/ui/views/frame/brave_tab_strip_region_view.cc
sangwoo108: Should add workspace button here too for compact mode?

browser/ui/views/frame/vertical_tabs/vertical_tab_strip_region_view.cc
```diff
@@ -271,6 +273,95 @@ class ResettableResizeArea : public views::ResizeArea {
 BEGIN_METADATA(ResettableResizeArea)
 END_METADATA
 
+class VerticalTabWorkspacesButton : public views::Button {
```
sangwoo108: If this button only contains icon and label, views::LabelButton could be better. But not 100% sure if it lacks some customizability for this case.

browser/ui/views/workspaces/workspaces_bubble_view.cc
bsclifton: I think it's OK to keep as 2 because it's already relative to the current font size.
https://source.chromium.org/chromium/chromium/src/+/main:ui/gfx/font_list.h;l=102-104;drc=4c83a1b71f87ad8cbf84f1adf17b51a1bc36c22e

Chromium does this too - but they do capture it with a constant name, for example:
https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/ui/ash/input_method/candidate_window_constants.h;l=11-13;drc=5cf2304bbe9c598ab4b41890a7ee20138df5d765

I could do this too! Like `kSpacesHeaderDelta`

browser/ui/views/workspaces/workspaces_bubble_view.cc
bsclifton: Fixed

browser/ui/views/workspaces/workspaces_bubble_view.cc
bsclifton: Fixed

browser/ui/views/workspaces/workspace_row_view.h
bsclifton: Fixed

browser/ui/views/workspaces/workspace_row_view.cc
bsclifton: Fixed - used a constant. Let me know what you think. It's relative to current size already

browser/ui/views/workspaces/workspace_row_view.cc
bsclifton: I need to double check this - I remember I needed to explicitly put this because it was defaulting to a larger inset (not 0).

browser/ui/views/workspaces/workspace_row_view.cc
bsclifton: Fixed!

browser/ui/views/workspaces/workspace_row_view.h
bsclifton: Fixed

browser/ui/views/workspaces/save_workspace_dialog.cc
bsclifton: Fixed

browser/ui/views/workspaces/save_workspace_dialog.cc
bsclifton: Fixed

browser/ui/views/workspaces/workspaces_bubble_view.cc
bsclifton: Fixed - used a constant. Let me know what you think. It's relative to current size already

browser/ui/views/workspaces/workspaces_bubble_view.cc
bsclifton: Fixed

browser/ui/views/frame/brave_tab_strip_region_view.cc
bsclifton: Will check this out...

browser/ui/views/workspaces/workspaces_bubble_view.h
bsclifton: Fixed

browser/ui/views/workspaces/workspace_row_view.cc
bsclifton: Fixed

browser/ui/views/workspaces/save_workspace_dialog.h
bsclifton: Fixed

app/brave_generated_resources.grd
bsclifton: Fixed!

browser/ui/views/workspaces/workspaces_bubble_view.cc
sangwoo108: > I think it's OK to keep as 2 because it's already relative to the current font size.

Yeah, I believe that makes it fragile if you meant to set a destination size.

Let's say we wanted 16px and current font size of the label is 14 by default.
Then the delta should be 2. Then if upstream changes the default size of label to 15, then the resulting size of this code would be 17px. 

Chromium uses kind of centralized font size system and uses relative value here and there but I don't think we're doing that - we usually use destination font size everywhere.

If you intended to use relative font size, that's okay.

chromium_src/ui/views/window/dialog_delegate.h
bsclifton: Fixed!

chromium_src/ui/views/bubble/bubble_dialog_delegate_view.h
bsclifton: Fixed!

browser/ui/views/workspaces/workspace_row_view.cc
bsclifton: Doesn't seem to have a difference on macOS. But I swear it does on Windows.
Screenshot 2026-06-22 at 10 07 24 AM

On Windows, it helped to get rid of the borders around the selection that you can see here. I'll need to check that again when I have access to Windows again (next week)

bsclifton:
Identified some problems w/ presubmit; fixing those...

browser/ui/views/workspaces/workspaces_bubble_view.cc
bsclifton: Fixed!

browser/ui/views/workspaces/workspace_row_view.cc
bsclifton: Fixed with absolute size

bsclifton:
@simonhong @sangwoo108 ready for re-review please 😄

app/brave_generated_resources.grd
```diff
@@ -256,21 +256,129 @@
...
+        </message>
+        <message name="IDS_WORKSPACE_ROW_STATS_MULTI_TAB" desc="Stats label shown on a workspace row when the workspace contains one window with multiple tabs">
+          <ph name="TAB_COUNT">$1<ex>2</ex></ph> Tabs
+        </message>
+        <message name="IDS_WORKSPACE_ROW_STATS_MULTI_WINDOW_SINGLE_TAB" desc="Stats label shown on a workspace row when the workspace contains multiple windows and one tab">
+          <ph name="WINDOW_COUNT">$1<ex>2</ex></ph> Windows – 1 Tab
+        </message>
+        <message name="IDS_WORKSPACE_ROW_STATS_MULTI_WINDOW_MULTI_TAB" desc="Stats label shown on a workspace row when the workspace contains multiple windows and multiple tabs">
+          <ph name="WINDOW_COUNT">$1<ex>2</ex></ph> Windows – <ph name="TAB_COUNT">$2<ex>2</ex></ph> Tabs
         </message>
```
simonhong: We can use `GetPluralStringFUTF16()` instead having above two strings. 
Ex, `IDS_APP_LIST_FOLDER_NUMBER_OF_APPS_ACCESSIBILE_DESCRIPTION`

browser/ui/views/frame/vertical_tabs/vertical_tab_strip_region_view.cc
```diff
@@ -312,6 +403,13 @@ BraveVerticalTabStripRegionView::BraveVerticalTabStripRegionView(
                           base::Unretained(original_region_view_->tab_strip_)),
       GetShortcutTextForNewTabButton(browser_view), browser_));
 
+  if (base::FeatureList::IsEnabled(features::kWorkspaces)) {
```
simonhong: Curious why this PR added workspace button below the vertical tab's new tab button.
In our design, it puts right after the vetical tab toggle button in toolbar.

I would recommend like this step. WDYT?
1. In this PR, only add button in horizontal tab mode.
2. In the f/u PR, add workspace button in toolbar in vertical tab.

browser/ui/views/frame/brave_tab_strip_region_view.cc
```diff
@@ -325,16 +329,27 @@ bool BraveHorizontalTabStripRegionView::ShouldShowHorizontalScrollButton()
...
 }
 
+void BraveHorizontalTabStripRegionView::OnWorkspacesButtonPressed() {
+  if (!workspaces_bubble_controller_) {
+    workspaces_bubble_controller_ =
+        std::make_unique<WorkspacesBubbleController>();
+  }
+  workspaces_bubble_controller_->ShowBubble(
+      workspaces_button_,
+      tab_strip_->GetBrowserWindowInterface()->GetBrowserForMigrationOnly());
```
simonhong: We should not use `GetBrowserForMigrationOnly()` in new code.

browser/ui/views/frame/brave_tab_strip_region_view.cc
```diff
@@ -405,6 +423,9 @@ void BraveHorizontalTabStripRegionView::Layout(PassKey) {
 
   // in vertical tabs mode, we make tab strip's height is the same with this
   // view's height to avoid extra gaps.
+  if (workspaces_button_) {
+    workspaces_button_->SetVisible(false);
+  }
```
simonhong, line 492: Let's move before the above comment. Above comment is not related with workspace button.

browser/ui/views/frame/brave_tab_strip_region_view.cc
```diff
@@ -545,6 +566,32 @@ void BraveHorizontalTabStripRegionView::Initialize() {
     tab_search_button_->SetVisible(false);
   }
 
+  if (base::FeatureList::IsEnabled(features::kWorkspaces)) {
```
simonhong: let's add one more condition - `browser->GetType() == BrowserWindowInterface::Type::TYPE_NORMAL`.

browser/ui/views/workspaces/workspaces_bubble_controller.h
```diff
@@ -0,0 +1,56 @@
...
+namespace views {
+class View;
+}  // namespace views
+
+// Owns the workspaces bubble and the "save workspace" dialog widgets. Both use
+// the CLIENT_OWNS_WIDGET ownership model, so the controller holds each
+// Widget's unique_ptr and tears it down when the widget closes. The controller
+// is owned by the tab strip region view that hosts the workspaces button, so
+// it reliably outlives the widgets it manages.
+class WorkspacesBubbleController {
```
simonhong, line 31: I think we can have one `WorkspacesBubbleController` per browser window.
How about having this in `BrowserWindowFeatures`?

browser/ui/views/workspaces/save_workspace_dialog.h
```diff
@@ -0,0 +1,65 @@
...
+
+// A modal dialog that prompts the user to name a new workspace. On accept, the
+// current state of all open windows for the profile is saved to disk.
+//
+// This combines the dialog delegate and contents view by inheriting from both
+// views::DialogDelegate and views::View directly, rather than the deprecated
+// views::DialogDelegateView. It also serves as the name field's controller so
+// the OK button can be enabled only while a name has been entered.
+class SaveWorkspaceDialog : public views::DialogDelegate,
+                            public views::View,
```
simonhong: This `multiple inheritance` is why upstream deprecated `DialogDelegateView`.
It seems this multiple inheritance caused many lifetime & ownership issues in upstream.
Also, it forces the class to disambiguate GetWidget()(inherited from both) and return this from GetContentsView(). Doing it by hand is unnecessary now that WidgetDelegate::SetContentsView() (inherited by DialogDelegate) can accept an externally-constructed view and take ownership of it.

```
SaveWorkspaceDialog::SaveWorkspaceDialog(Browser* browser) : browser_(browser) {
    SetOwnershipOfNewWidget(views::Widget::InitParams::CLIENT_OWNS_WIDGET);
    SetModalType(ui::mojom::ModalType::kWindow);
    SetTitle(...);
    SetButtonLabel(...);
    SetAcceptCallback(...);

    auto contents = std::make_unique<views::View>();
    contents->SetLayoutManager(std::make_unique<views::BoxLayout>(
        views::BoxLayout::Orientation::kVertical, gfx::Insets(kPadding), kSpacing));
    contents->AddChildView(std::make_unique<views::Label>(...));
    auto textfield = std::make_unique<views::Textfield>(); 
    // ... configure textfield, SetController(this) ...
    name_field_ = contents->AddChildView(std::move(textfield));
    SetButtonEnabled(ui::mojom::DialogButton::kOk, !name_field_->GetText().empty());
    SetContentsView(std::move(contents));
}
```

browser/ui/views/workspaces/workspaces_bubble_view.h
```diff
@@ -0,0 +1,64 @@
...
+// exposes actions to restore, delete, or save a new one. The bubble is the
+// primary entry point for the workspaces UI and is anchored to the workspaces
+// button in the tab strip.
+//
+// This combines the bubble delegate and contents view by inheriting from both
+// views::BubbleDialogDelegate and views::View directly, rather than the
+// deprecated views::BubbleDialogDelegateView. Its widget uses the
+// CLIENT_OWNS_WIDGET model and is owned by WorkspacesBubbleController.
+class WorkspacesBubbleView : public views::BubbleDialogDelegate,
+                             public views::View {
```
simonhong: ditto

app/brave_generated_resources.grd
bsclifton: Glad you caught this! I know better, being part of string reviewers 😓 Will work on a fix. I wasn't sure if you can nest plural calls and then forgot to revisit.

bsclifton: Fixed

browser/ui/views/frame/vertical_tabs/vertical_tab_strip_region_view.cc
sangwoo108: +1 for follow-up PR. I think having workspace button below new tab button require transient changes(for size of viewport and etc), which will be discarded soon, so it would be better to have a follow-up PR implementing the workspace button properly laid out horizontally.

browser/ui/views/frame/brave_tab_strip_region_view.cc
sangwoo108: One more request, I think having a separate method for workspace creation would be good.

CreateWorkspaceButtonIfNeeded() would be good for consistency with CreateScrollButtonIfNeeded().

browser/ui/views/workspaces/workspace_row_view.cc
```diff
@@ -0,0 +1,196 @@
...
+  } else {
+    SetBackground(nullptr);
+  }
+}
+
+void WorkspaceRowView::UpdateChildSelectionStyling() {
+  for (views::View* child : children()) {
+    if (auto* info_btn = views::AsViewClass<WorkspaceInfoButton>(child)) {
+      info_btn->SetSelected(selected_);
+    } else if (auto* img_btn = views::AsViewClass<views::ImageButton>(child)) {
```
sangwoo108: optional: I think storing info_btn and img_btn as member variables in the constructor would be good so that we can avoid iterating through all children.

browser/ui/views/workspaces/workspace_row_view.cc
```diff
@@ -0,0 +1,196 @@
...
+          img_btn, kLeoMoreVerticalIcon,
+          selected_
+              ? views::IconColors(SK_ColorWHITE, SK_ColorWHITE)
+              : views::IconColors(ui::kColorIcon, ui::kColorIconDisabled));
+    }
+  }
+}
+
+void WorkspaceRowView::ShowMoreMenu() {
+  menu_model_ = std::make_unique<ui::SimpleMenuModel>(this);
```
sangwoo108, line 162: Not a big deal but, maybe we should override SimpleMenuModelDelegate::MenuClosed() and clear menu_model_.

```
void WorkspaceRowView::MenuClosed(SimpleMenuModel* model) {
  base::SingleThreadTaskRunner::GetCurrentDefault()->DeleteSoon(
      FROM_HERE, menu_model.release());
}

```

browser/ui/views/workspaces/workspaces_bubble_view_browsertest.cc
```diff
@@ -0,0 +1,193 @@
...
+}
+
+}  // namespace
+
+class WorkspacesBubbleBrowserTest : public InProcessBrowserTest {
+ public:
+  WorkspacesBubbleBrowserTest() {
+    feature_list_.InitAndEnableFeature(features::kWorkspaces);
+  }
+
```
sangwoo108, line 90: nit: missing destructor?

browser/ui/views/workspaces/workspaces_bubble_view_browsertest.cc
```diff
@@ -0,0 +1,193 @@
...
+            return;
+          }
+          if (auto* b = views::AsViewClass<WorkspacesBubbleView>(
+                  delegate->GetContentsView())) {
+            bubble_future.SetValue(b);
+          }
+        }));
+
+    auto* button = GetWorkspacesButton();
+    EXPECT_TRUE(button) << "Workspaces button not found in the tab strip";
```
sangwoo108, line 113: I feel like this line is precondition of subsequent tests, so ASSERT_TRUE() and SCOPED_TRACE() could be better?

browser/workspaces/workspace_service.cc
```diff
@@ -58,6 +59,17 @@ void WorkspaceService::RemoveWorkspaceMetadata(const std::string& name) {
   updated->Remove(ComputeKey(name));
 }
 
+void WorkspaceService::DeleteWorkspace(const std::string& name) {
+  RemoveWorkspaceMetadata(name);
+  io_task_runner_->PostTask(FROM_HERE, base::BindOnce(
```
sangwoo108, line 64: Thinking out loud, definitely no need to addressed in this PR.

1. io_task_runner_ is created with USER_VISIBLE trait, but I feel like the disk operation doesn't have to be user visible operation, as we depend on the metadata in prefs.

2. We might have orphaned workspace data on the disk, so we periodically try finding orphaned dir and clear things up. Maybe it'd be good to schedule the job on Workspace service creation.

browser/ui/views/frame/vertical_tabs/vertical_tab_strip_region_view.cc
bsclifton: I like the idea of only doing regular (horizontal) tabs first. That helps remove some complexity - as I just made something working and didn't revisit after it was clickable. It's not matching design as-is. I can create a new issue for this specifically as a follow up when this PR is in good shape. Good call 😄

browser/ui/views/frame/vertical_tabs/vertical_tab_strip_region_view.cc
bsclifton: Removed - we can address w/ https://github.com/brave/brave-browser/issues/56728

browser/ui/views/frame/vertical_tabs/vertical_tab_strip_region_view.cc
bsclifton: Removed all the vertical tabs code - we can track that with https://github.com/brave/brave-browser/issues/56728 and address in a follow up 😄

browser/ui/views/frame/brave_tab_strip_region_view.cc
bsclifton: Fixed

browser/ui/views/frame/brave_tab_strip_region_view.cc
bsclifton: Fixed! Removed several browser* references too

browser/ui/views/frame/brave_tab_strip_region_view.cc
bsclifton: Good feedback! Fixed!

browser/ui/views/workspaces/workspaces_bubble_controller.h
bsclifton: Good one - will look at this next...

browser/ui/views/workspaces/save_workspace_dialog.h
bsclifton: Will look at this next, thanks!

browser/ui/views/workspaces/workspace_row_view.cc
bsclifton: Fixed

browser/ui/views/workspaces/workspace_row_view.cc
bsclifton: Fixed

browser/ui/views/workspaces/workspaces_bubble_view_browsertest.cc
bsclifton: Fixed

browser/ui/views/workspaces/workspaces_bubble_view_browsertest.cc
bsclifton: Fixed / addressed the best I could (can't use assert because it's a macro for return)

sangwoo108 (approved):
LGTM. I think other stuff could be addressed as follow-up except the multiple inheritance problem. I'll defer it to @simonhong .

browser/ui/views/frame/brave_tab_strip_region_view.cc
bsclifton: Looks like this diff doesn't resolve anymore - but compact mode looks OK when enabled! Button shows
Screenshot 2026-06-30 at 9 58 32 AM

browser/ui/views/workspaces/workspaces_bubble_view.h
bsclifton: Fixed!

browser/ui/views/workspaces/save_workspace_dialog.h
bsclifton: Fixed! Thanks for the super helpful feedback - this was pretty easy to do!

browser/ui/views/workspaces/workspaces_bubble_controller.h
bsclifton: Fixed! Great recommendation

browser/ui/browser_window/internal/browser_window_features.cc
```diff
@@ -189,6 +190,9 @@ void BrowserWindowFeatures::InitPostBrowserViewConstruction(
       browser_view->browser()->GetType(),
       browser_view->GetProfile()->GetPrefs());
 
+  workspaces_bubble_controller_ =
+      std::make_unique<WorkspacesBubbleController>();
```
simonhong: Maybe we want to use this guard?
```
  if (base::FeatureList::IsEnabled(features::kWorkspaces) &&
      browser_->GetType() != BrowserWindowInterface::Type::TYPE_NORMAL) {
  }
```

browser/ui/views/workspaces/workspaces_bubble_controller.cc
```diff
@@ -0,0 +1,81 @@
...
+#include "ui/views/view.h"
+#include "ui/views/widget/widget.h"
+
+WorkspacesBubbleController::WorkspacesBubbleController() = default;
+
+WorkspacesBubbleController::~WorkspacesBubbleController() = default;
+
+void WorkspacesBubbleController::ShowBubble(views::View* anchor_view,
+                                            Profile* profile) {
+  if (bubble_widget_) {
```
simonhong, line 28: is it worth to check `save_dialog_widget_` together here?

browser/ui/views/workspaces/workspace_row_view.h
```diff
@@ -0,0 +1,82 @@
...
+  // ui::SimpleMenuModel::Delegate:
+  void ExecuteCommand(int command_id, int event_flags) override;
+  void MenuClosed(ui::SimpleMenuModel* source) override;
+
+ private:
+  void UpdateBackground();
+  void UpdateChildSelectionStyling();
+  void ShowMoreMenu();
+
+  bool hovered_ = false;
```
simonhong: We have `View::IsMouseHovered()`.
Can you check we can use `View::IsMouseHovered()` instead of caching `hovered_`?

browser/ui/views/workspaces/workspaces_bubble_controller.cc
bsclifton: Not needed for this case, because Save dialog is modal. Which means you can't show a bubble while save dialog is happening. Picture for reference:
<kbd></kbd>

It is possible to show on another window, but that one is owned separately and is a different instance
<kbd></kbd>

browser/ui/browser_window/internal/browser_window_features.cc
bsclifton: Fixed! Great one - definitely needed this

browser/ui/views/workspaces/workspace_row_view.h
bsclifton: Fixed!

simonhong (approved):
++ 👍🏼

mkarolin (approved):
`strings`++

bsclifton:
Looks like some browser test errors to sort out (likely regression after I changed inheritance). I'll check those out next. Thanks all! 😄

bsclifton:
Failing upstream test (`ShareOperationUnitTest.BasicFields`) is not related and passed before (seems intermittent).

Everything else is good. Merging!

PR #36918, opened 2026-06-02
Set extra data for tree tab with BrowserLiveTabContext

sangwoo108:
BrowserLiveTabContext is responsible for providing extra data for live tabs, which are used for tab restore and session restore. This change adds support for tree tab extra data, allowing the tree structure of tabs to be preserved during restore.

Part of https://github.com/brave/brave-browser/issues/49792

### DIFF
--- a/browser/sessions/brave_session_restore_browsertest.cc
+++ b/browser/sessions/brave_session_restore_browsertest.cc
@@ -16,9 +16,11 @@
 #include "brave/browser/ui/tabs/brave_tab_prefs.h"
 #include "brave/browser/ui/tabs/brave_tab_strip_model.h"
 #include "brave/browser/ui/tabs/tree_tab_model.h"
+#include "brave/components/tabs/public/tree_tab_node.h"
 #include "brave/components/tabs/public/tree_tab_node_tab_collection.h"
 #include "chrome/browser/sessions/session_service.h"
 #include "chrome/browser/sessions/session_service_factory.h"
+#include "chrome/browser/sessions/tab_restore_service_factory.h"
 #include "chrome/browser/ui/browser.h"
 #include "chrome/browser/ui/tabs/features.h"
 #include "chrome/browser/ui/tabs/tab_enums.h"
@@ -29,6 +31,11 @@
 #include "components/prefs/pref_service.h"
 #include "components/sessions/core/serialized_navigation_entry.h"
 #include "components/sessions/core/session_types.h"
+#include "components/sessions/core/tab_restore_service.h"
+#include "components/sessions/core/tab_restore_types.h"
+#include "components/tabs/public/tab_collection.h"
+#include "components/tabs/public/tab_interface.h"
+#include "content/public/browser/web_contents.h"
 #include "content/public/test/browser_test.h"
 #include "content/public/test/browser_test_utils.h"
 #include "net/dns/mock_host_resolver.h"
@@ -203,6 +210,22 @@ class BraveTreeTabSessionRestoreBrowserTest : public InProcessBrowserTest {
 
   BraveTabStripModel* brave_tab_strip_model() {
     return static_cast<BraveTabStripModel*>(browser()->tab_strip_model());
+  }
+
+  // Returns the tree node id string for the tab at |index|, or an empty string
+  // if the tab is not inside a tree node. Mirrors the GetTreeTabNodeCollection
+  // helper used by TreeTabSessionObserver.
+  std::string GetTreeNodeIdForTab(int index) {
+    const tabs::TabInterface* tab =
+        brave_tab_strip_model()->GetTabAtIndex(index);
+    const tabs::TabCollection* parent = tab->GetParentCollection();
+    if (parent && parent->type() == tabs::TabCollection::Type::TREE_NODE) {
+      return static_cast<const tabs::TreeTabNodeTabCollection*>(parent)
+          ->node()
+          .id()
+          .ToString();
+    }
+    return std::string();
   }
 
   SessionService* session_service() {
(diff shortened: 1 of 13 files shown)


chromium_src/chrome/browser/ui/browser_live_tab_context.cc
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) 2026 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#include <map>
+#include <string>
+
+#include "base/feature_list.h"
+#include "brave/browser/ui/tabs/tree_tab_session_observer.h"
```
netzenbot: The new chromium_src override includes `brave/browser/ui/tabs/tree_tab_session_observer.h`, but `chromium_src/chrome/browser/ui/BUILD.gn` is not updated to add `//brave/browser/ui/tabs:tree_tab_session_observer` as a dep. The upstream target compiling `browser_live_tab_context.cc` doesn't transitively have this dep, so `gn check` will likely fail. You need to add the dep to the `group("ui")` block in `chromium_src/chrome/browser/ui/BUILD.gn` (under `!is_android`). ([Verify chromium_src Header GN Dependencies](https://github.com/brave/brave-core/tree/master/docs/best-practices/chromium-src-overrides.md#CSRC-018))

browser/ui/tabs/tree_tab_session_manager.h
```diff
@@ -33,6 +33,14 @@ class TreeTabSessionObserver : public TabStripModelObserver {
   TreeTabSessionObserver& operator=(const TreeTabSessionObserver&) = delete;
   ~TreeTabSessionObserver() override;
 
+  // Called from BrowserLiveTabContext::GetExtraDataForTab to record the
+  // tree-tab position of the tab being closed. Writes kBraveTreeNodeIdKey and
+  // kBraveTreeParentNodeIdKey, and kBraveTreeNodeCollapsedKey into
+  // |extra_data|.
+  void MaybePopulateTreeTabExtraData(
```
netzenbot, line 43: nit: The declaration uses `std::map<std::string, std::string>` but neither `<map>` nor `<string>` are included in this header. Per [CS-001](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-001), include what you use directly.

rewrite/chrome/browser/ui/browser_live_tab_context.cc.yaml
```diff
@@ -0,0 +1,14 @@
+# Copyright (c) 2026 The Brave Authors. All rights reserved.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+substitutions:
+  - description:
+      Call MaybePopulateTreeTabExtraData() from
```
netzenbot, line 8: nit: The `description` field says `CreateLiveTabContext()` but the substitution targets `GetExtraDataForTab()`. Per [CS-019](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-019), comments should be accurate for future readers — update the description to reference `GetExtraDataForTab()`.

browser/sessions/brave_session_restore_browsertest.cc
```diff
@@ -457,3 +480,114 @@ IN_PROC_BROWSER_TEST_F(BraveTreeTabSessionRestoreBrowserTest,
...
+  const auto* node_id =
+      base::FindOrNull(restored_tab->extra_data, kBraveTreeNodeIdKey);
+  ASSERT_TRUE(node_id);
+  EXPECT_EQ(*node_id, node_b_id);
+
+  // ...its parent's node id (A)...
+  const auto* parent_id =
+      base::FindOrNull(restored_tab->extra_data, kBraveTreeParentNodeIdKey);
+  ASSERT_TRUE(parent_id);
+  EXPECT_EQ(*parent_id, node_a_id);
```
netzenbot, line 542: nit: `EXPECT_TRUE(*collapsed == "0" || *collapsed == "1")` won't print the actual value of `*collapsed` on failure. The identical assertion in `VerifyTreeNodeExtraDataKeys` adds `<< "tab extra data should contain kBraveTreeNodeCollapsedKey with value 0 or 1"` — prefer the same pattern here, or use `EXPECT_THAT(*collapsed, testing::AnyOf("0", "1")) << "actual: " << *collapsed;`. ([Include Diagnostic State in Test Assertions](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-014))

chromium_src/chrome/browser/ui/browser_live_tab_context.cc
```diff
@@ -0,0 +1,40 @@
...
+#include "chrome/browser/ui/browser_window/public/browser_window_features.h"
+#include "chrome/browser/ui/browser_window/public/browser_window_interface.h"
+#include "chrome/browser/ui/tabs/features.h"
+
+namespace {
+
+// Called from BrowserLiveTabContext::GetExtraDataForTab to record the tree-tab
+// position of the tab being closed. Writes kBraveTreeNodeIdKey and
+// kBraveTreeParentNodeIdKey, and kBraveTreeNodeCollapsedKey into |extra_data|.
+void MaybePopulateTreeTabExtraData(
```
simonhong, line 20: nit: I think this method could standalone util method w/o depending on observer.

simonhong (approved):
++

chromium_src/chrome/browser/ui/browser_live_tab_context.cc
sangwoo108: Yes, it's possible but I'd prefer keep this in this class so that we can put responsibilities of serialize and deserialize of the data to this class. What do you think?

simonhong: It's ok. The reason why I thought was the class name `TreeTabSessionObserver`.
It's observer but this PR added more than `observer` like this method.
If this class can have more role, it's fine.

sangwoo108: Thanks, let me think about better name.

browser/sessions/brave_session_restore_browsertest.cc
sangwoo108: done

browser/ui/tabs/tree_tab_session_manager.h
sangwoo108: done

chromium_src/chrome/browser/ui/browser_live_tab_context.cc
sangwoo108: done

rewrite/chrome/browser/ui/browser_live_tab_context.cc.yaml
sangwoo108: done

netzenbot (approved):

goodov (approved):

PR #36925, opened 2026-06-02
[Wallet] Don't require user activation for hid api on ledger subframe

### LINKED ISSUE #56111: [Wallet] Don't require user activation for hid api on ledger subframe
Ledger js lib uses `navigator.hid.requestDevices()` to get access to connected Ledger wallet device. This method requires [transient user activation](https://developer.mozilla.org/en-US/docs/Web/API/HID/requestDevice#security). 
We host Ledger js lib in separate frame and it doesn't share activation flags with mainframe. To workaround that we prompt user to click 'Authorize' button in ledger subframe on start of interaction with device and after some time period when flag gets reset. 
This is just bad UX and complicates implementation.

Proposed fix: don't require transient activation for this api method when it's called from `chrome-untrusted://ledger-bridge` subframe. 

Button will be removed from UI in the next PR

supermassive:
Sec review: https://github.com/brave/reviews/issues/2269

Resolves https://github.com/brave/brave-browser/issues/56111

### DIFF
--- a/chromium_src/third_party/blink/public/platform/web_content_settings_client.h
+++ b/chromium_src/third_party/blink/public/platform/web_content_settings_client.h
@@ -40,8 +40,8 @@
                                      const blink::WebURL& script_url) {     \
     return enabled_per_settings;                                            \
   }                                                                         \
-  virtual bool RequireTransientActivationForHidRequestDevice() {            \
-    return true;                                                            \
+  virtual bool AllowHidRequestDeviceWithoutTransientActivation() {          \
+    return false;                                                           \
   }                                                                         \
   virtual bool AllowStorageAccessSync
 
--- a/chromium_src/third_party/blink/public/platform/web_content_settings_client.h
+++ b/chromium_src/third_party/blink/public/platform/web_content_settings_client.h
@@ -40,6 +40,9 @@
                                      const blink::WebURL& script_url) {     \
     return enabled_per_settings;                                            \
   }                                                                         \
+  virtual bool RequireTransientActivationForHidRequestDevice() {            \
+    return true;                                                            \
+  }                                                                         \
   virtual bool AllowStorageAccessSync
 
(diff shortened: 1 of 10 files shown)


chromium_src/third_party/blink/public/platform/web_content_settings_client.h
```diff
@@ -42,6 +42,9 @@ class GURL;
                                      const blink::WebURL& script_url) {     \
     return enabled_per_settings;                                            \
   }                                                                         \
+  virtual bool RequireTransientActivationForHidRequestDevice() {            \
```
goodov: let's make the default be `false` and name it something like `AllowHidRequestDeviceWithoutTransientActivation`.

cmazakas (approved):
I didn't test the change locally but wallet-core LGTM!

netzenbot (approved):

goodov (approved):

kdenhartog (commented):
Code looks fine. I was trying to build locally to test it and getting some build failures. Will approve once I get a chance to test it with a dev build. I think your most recent commit may have fixed the issue I was hitting. I want to make sure UX is behaving like I expect and no weird side effects get picked up by this.

supermassive:
> Code looks fine. I was trying to build locally to test it and getting some build failures. Will approve once I get a chance to test it with a dev build. I think your most recent commit may have fixed the issue I was hitting. I want to make sure UX is behaving like I expect and no weird side effects get picked up by this.

@kdenhartog 
You'd better try with the follup PR which actually removes the button
https://github.com/brave/brave-core/pull/37148
I'm going to get approves on both and merge them together

kdenhartog (approved):
From a security perspective this seems fine. When testing with https://github.com/brave/brave-core/pull/37148 there might be a regression issue that I spotted while testing, but @supermassive is aware and looking for a fix so not going to block this longer.

PR #36874, opened 2026-06-01
Fixed download toolbar icon regressions with insecure file

simonhong:
Resolves https://github.com/brave/brave-browser/issues/55816

This ensures UpdateIcon_BraveImpl is called at the point where the upstream has already updated the
image, rather than at the end of the function where it would be skipped by early returns.

Regressions from https://github.com/brave/brave-core/pull/35750
TEST=DownloadToolbarInsecureIconTest.InsecureDownloadIcon

Screenshot from 2026-06-01 17-41-04
Screenshot from 2026-06-01 17-41-13

PR #36929, opened 2026-06-02
Revert "Fixed download toolbar icon regressions with insecure file"

simonhong:
Reverts brave/brave-core#36874

Original PR had one more commit but it's missed.


cdesouza-chromium (approved):
LGTM

PR #36935, opened 2026-06-02
[VPN 2.0] Add BraveVpnServiceImpl V2 stubs

### LINKED ISSUE #54597: Implement a stub BraveVpnService v2

alekspop26:

This change adds the second service implementation, BraveVpnServiceImpl, for Architecture 2.0. It will live in a "components/brave_vpn/v2" subdirectory, but will eventually move into "components/brave_vpn" once V1 implementation has been removed.

Notable changes:
- fully testable BraveVpnServiceImpl v2 stub implementation, in a dedicated "v2" namespace;
- BraveVpnServiceFactory can now create a proper implementation based on the compile-time GN flags;
- v1/v2 include isolation using DEPS include rules (so that the implementations can't depend on each other);
- stub unit test for BraveVpnServiceImpl v2.

Resolves https://github.com/brave/brave-browser/issues/54597

### DIFF
--- a/browser/sources.gni
+++ b/browser/sources.gni
@@ -554,7 +554,10 @@ if (enable_brave_vpn) {
 }
 
 if (enable_brave_vpn_v2) {
-  brave_chrome_browser_deps += [ "//brave/third_party/boringtun" ]
+  brave_chrome_browser_deps += [
+    "//brave/components/brave_vpn/browser/v2:browser",
+    "//brave/third_party/boringtun",
+  ]
 }
 
 if (is_android) {
--- a/components/brave_vpn/browser/v2/BUILD.gn
+++ b/components/brave_vpn/browser/v2/BUILD.gn
@@ -1,0 +1,32 @@
+# Copyright (c) 2026 The Brave Authors. All rights reserved.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import("//brave/components/brave_vpn/common/buildflags/buildflags.gni")
+
+assert(enable_brave_vpn_v2)
+
+static_library("browser") {
+  sources = [
+    "brave_vpn_service_impl.cc",
+    "brave_vpn_service_impl.h",
+  ]
+
+  if (is_android) {
+    sources += [ "brave_vpn_service_impl_android.cc" ]
+  } else if (is_linux || is_mac || is_win) {
+    sources += [ "brave_vpn_service_impl_desktop.cc" ]
+  }
+
+  public_deps = [
+    "//brave/components/brave_vpn/browser",
+    "//brave/components/skus/browser",
+    "//brave/components/skus/common:mojom",
+  ]
+
+  deps = [
+    "//base",
+    "//brave/components/brave_vpn/common",
+  ]
+}
--- a/components/brave_vpn/browser/v2/brave_vpn_service_impl.cc
+++ b/components/brave_vpn/browser/v2/brave_vpn_service_impl.cc
@@ -1,0 +1,60 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "brave/components/brave_vpn/browser/v2/brave_vpn_service_impl.h"
+
+#include "base/notimplemented.h"
+#include "brave/components/skus/browser/skus_utils.h"
+
+namespace brave_vpn {
+namespace v2 {
+
+BraveVpnServiceImpl::BraveVpnServiceImpl()
+    : connection_state_(mojom::ConnectionState::DISCONNECTED),
+      purchased_state_(mojom::PurchasedState::NOT_PURCHASED) {}
+
+BraveVpnServiceImpl::~BraveVpnServiceImpl() = default;
+
+bool BraveVpnServiceImpl::IsBraveVPNEnabled() const {
+  NOTIMPLEMENTED();
+  return false;
+}
+
+bool BraveVpnServiceImpl::IsPurchased() const {
+  NOTIMPLEMENTED();
+  return purchased_state_ == mojom::PurchasedState::PURCHASED;
+}
+
+void BraveVpnServiceImpl::ReloadPurchasedState() {
+  NOTIMPLEMENTED();
+}
+
+std::string BraveVpnServiceImpl::GetCurrentEnvironment() const {
+  NOTIMPLEMENTED();
+  return skus::kEnvProduction;
+}
+
+void BraveVpnServiceImpl::GetPurchasedState(
+    GetPurchasedStateCallback callback) {
+  NOTIMPLEMENTED();
+  std::move(callback).Run(
+      mojom::PurchasedInfo::New(purchased_state_, std::nullopt));
+}
+
+void BraveVpnServiceImpl::LoadPurchasedState(const std::string& domain) {
+  NOTIMPLEMENTED();
+}
+
+void BraveVpnServiceImpl::GetAllRegions(GetAllRegionsCallback callback) {
+  NOTIMPLEMENTED();
+  std::move(callback).Run({});
+}
+
+void BraveVpnServiceImpl::Shutdown() {
+  BraveVpnService::Shutdown();
+}
+
+}  // namespace v2
+}  // namespace brave_vpn
--- a/components/brave_vpn/browser/v2/brave_vpn_service_impl.h
+++ b/components/brave_vpn/browser/v2/brave_vpn_service_impl.h
@@ -1,0 +1,123 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#ifndef BRAVE_COMPONENTS_BRAVE_VPN_BROWSER_V2_BRAVE_VPN_SERVICE_IMPL_H_
+#define BRAVE_COMPONENTS_BRAVE_VPN_BROWSER_V2_BRAVE_VPN_SERVICE_IMPL_H_
+
+#include "brave/components/brave_vpn/browser/brave_vpn_service.h"
+#include "build/build_config.h"
+
+namespace brave_vpn {
+namespace v2 {
+
+class BraveVpnServiceImpl : public BraveVpnService {
+ public:
+  BraveVpnServiceImpl();
+  ~BraveVpnServiceImpl() override;
+
+  BraveVpnServiceImpl(const BraveVpnServiceImpl&) = delete;
+  BraveVpnServiceImpl& operator=(const BraveVpnServiceImpl&) = delete;
+
+  // BraveVpnService overrides:
+  // Implementation of public interface exposed to other components.
+  bool IsBraveVPNEnabled() const override;
+  bool IsPurchased() const override;
+  void ReloadPurchasedState() override;
+  std::string GetCurrentEnvironment() const override;
+
+  // mojom::ServiceHandler overrides:
+  void GetPurchasedState(GetPurchasedStateCallback callback) override;
+  void LoadPurchasedState(const std::string& domain) override;
+  void GetAllRegions(GetAllRegionsCallback callback) override;
+
+#if !BUILDFLAG(IS_ANDROID)
+  // BraveVpnService overrides:
+  bool IsConnected() const override;
+  void ToggleConnection() override;
+  mojom::ConnectionState GetConnectionState() const override;
+  void RecordWidgetUsageMetrics(bool new_usage) override;
+
+  // mojom::ServiceHandler overrides:
+  void GetConnectionState(GetConnectionStateCallback callback) override;
+  void Connect() override;
+  void Disconnect() override;
+  void GetSelectedRegion(GetSelectedRegionCallback callback) override;
+  void SetSelectedRegion(mojom::RegionPtr region) override;
+  void ClearSelectedRegion() override;
+  void GetProductUrls(GetProductUrlsCallback callback) override;
+  void CreateSupportTicket(const std::string& email,
+                           const std::string& subject,
+                           const std::string& body,
+                           CreateSupportTicketCallback callback) override;
+  void GetSupportData(GetSupportDataCallback callback) override;
+  void ResetConnectionState() override;
+  void EnableOnDemand(bool enable) override;
+  void GetOnDemandState(GetOnDemandStateCallback callback) override;
+  void EnableSmartProxyRouting(bool enable) override;
+  void GetSmartProxyRoutingState(
+      GetSmartProxyRoutingStateCallback callback) override;
+#else   // !BUILDFLAG(IS_ANDROID)
+  // mojom::ServiceHandler overrides:
+  void GetPurchaseToken(GetPurchaseTokenCallback callback) override;
+
+  // Implementation of public interface for Android native worker.
+  void GetTimezonesForRegions(ResponseCallback callback) override;
+  void GetHostnamesForRegion(ResponseCallback callback,
+                             const std::string& region,
+                             const std::string& region_precision) override;
+  void GetProfileCredentials(ResponseCallback callback,
+                             const std::string& subscriber_credential,
+                             const std::string& hostname) override;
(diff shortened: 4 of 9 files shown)


components/brave_vpn/browser/v2/BUILD.gn
```diff
@@ -0,0 +1,46 @@
+# Copyright (c) 2026 The Brave Authors. All rights reserved.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import("//brave/components/brave_vpn/common/buildflags/buildflags.gni")
+
```
netzenbot, line 7: The new `browser` target implements a Mojo service but uses `static_library` instead of `component()`. Per [BS-032](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-032), Mojo service targets should use `component()` for correct dynamic linking support.

components/brave_vpn/v2/browser/BUILD.gn
```diff
@@ -0,0 +1,46 @@
...
+
+  sources = [ "brave_vpn_service_impl_unittest.cc" ]
+
+  deps = [
+    ":browser",
+    "//base",
+    "//brave/components/brave_vpn/browser",
+    "//brave/components/brave_vpn/common",
+    "//brave/components/brave_vpn/common/mojom",
+    "//brave/components/constants",
```
netzenbot: The new `unit_tests` target adds `"//brave/components/constants"` as a dep, but `brave_vpn_service_impl_unittest.cc` doesn't include or use anything from it. Per [BS-050](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-050), new deps on `//brave/components/constants` should be avoided — it's a deprecated legacy target. Remove this dep.

components/brave_vpn/browser/v2/brave_vpn_service_impl_unittest.cc
```diff
@@ -0,0 +1,23 @@
...
+  // This test just verifies that the test fixture compiles and links correctly
+  // with the real BraveVpnServiceImpl implementation.
+  BraveVpnServiceImpl service;
+
+  // Service is not purchased by default.
+  EXPECT_FALSE(service.IsPurchased());
+}
+
+}  // namespace v2
+}  // namespace brave_vpn
```
netzenbot: The test only verifies the default not-purchased state. Per [TI-026](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-026), guard conditions should be tested on both sides — add a second test that calls `SetPurchasedStateForTesting` with `mojom::PurchasedState::PURCHASED` and asserts `IsPurchased()` returns true.

alekspop26: This unit test is in fact a stub, correct unit tests will be added later.

components/brave_vpn/v2/browser/BUILD.gn
alekspop26: Fixed.

components/brave_vpn/browser/v2/BUILD.gn
alekspop26: Confirmed with @bridiver that it's ok to proceed with the `static_library` for now, but eventually we'll convert to a component. Tracking ticket: https://github.com/brave/brave-browser/issues/56020

browser/sources.gni
```diff
@@ -554,7 +554,10 @@ if (enable_brave_vpn) {
 }
 
 if (enable_brave_vpn_v2) {
-  brave_chrome_browser_deps += [ "//brave/third_party/boringtun" ]
+  brave_chrome_browser_deps += [
+    "//brave/components/brave_vpn/browser/v2:browser",
+    "//brave/third_party/boringtun",
```
bridiver, line 559: why is boringtun added directly to browser sources? I thought this specifically was not going to be linked in the browser process?

bridiver: also this should be a data dep since it creates a shared library loaded at runtime

bridiver: The right place for this is probably `dist_resources` as a data_dep, but seems better to encapsulate it within some brave_vpn/app/v2 target instead of adding it directly since it's really a data dep for a separate exe

bridiver: I guess this dep was already here so maybe this change should be a follow-up PR

alekspop26: There was an agreement to keep it in browser deps for now, until we create a helper app that in fact depends on Boringtun. So yeah, it's going to be a data_dep of an `app/v2/helper` target (once we add it).

bridiver: but even if it's here, it should still be a data dep now because otherwise it blocks targets that depend on`//chrome/browser`. As a data_dep gn just ensures it eventually gets built so you can have more parallelization in the build

alekspop26: It will be changed here: https://github.com/brave/brave-browser/issues/54626 (one of the next tasks I'm going to pick).

components/brave_vpn/browser/v2/brave_vpn_service_impl.cc
```diff
@@ -0,0 +1,204 @@
...
+      mojom::PurchasedInfo::New(purchased_state_, std::nullopt));
+}
+
+void BraveVpnServiceImpl::LoadPurchasedState(const std::string& domain) {}
+
+void BraveVpnServiceImpl::GetAllRegions(GetAllRegionsCallback callback) {
+  std::move(callback).Run({});
+}
+
+#if !BUILDFLAG(IS_ANDROID)
```
bridiver: Rather than having these large blocks of IS_ANDROID/!IS_ANDROID I think it would be better to have brave_vpn_service_impl_android.cc and brave_vpn_service_impl_desktop.cc to handle the android and desktop specific implementations.

components/brave_vpn/browser/v2/brave_vpn_service_impl.h
```diff
@@ -0,0 +1,124 @@
...
+  void ToggleConnection() override;
+  mojom::ConnectionState GetConnectionState() const override;
+  void RecordWidgetUsageMetrics(bool new_usage) override;
+#endif  // !BUILDFLAG(IS_ANDROID)
+
+  // mojom::ServiceHandler overrides:
+  void GetPurchasedState(GetPurchasedStateCallback callback) override;
+  void LoadPurchasedState(const std::string& domain) override;
+  void GetAllRegions(GetAllRegionsCallback callback) override;
+#if !BUILDFLAG(IS_ANDROID)
```
bridiver: please group these together so there's only a single !IS_ANDROID block

components/brave_vpn/browser/v2/BUILD.gn
```diff
@@ -0,0 +1,45 @@
...
+assert(enable_brave_vpn)
```
bridiver: shouldn't this assert enable_brave_vpn_v2?

components/brave_vpn/browser/v2/brave_vpn_service_impl_unittest.cc
```diff
@@ -0,0 +1,23 @@
...
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "brave/components/brave_vpn/browser/v2/brave_vpn_service_impl.h"
+
+#include "testing/gtest/include/gtest/gtest.h"
+
+namespace brave_vpn {
+namespace v2 {
+
+TEST(BraveVpnServiceImplTest, CompilesAndIsPurchasedReturnsFalse) {
```
bridiver: This isn't necessary and should be removed until there are actual unit tests

components/brave_vpn/browser/v2/BUILD.gn
```diff
@@ -0,0 +1,45 @@
...
+    "//brave/components/brave_vpn/common",
+  ]
+}
+
+source_set("unit_tests") {
+  testonly = true
+
+  sources = [ "brave_vpn_service_impl_unittest.cc" ]
+
+  deps = [
```
bridiver: per above I think this unit test should be removed, but where does this list of deps come from? There are only two headers in brave_vpn_service_impl_unittest.cc so the rest of these deps are not correct

components/brave_vpn/browser/v2/BUILD.gn
```diff
@@ -0,0 +1,45 @@
...
+  ]
+
+  public_deps = [
+    "//brave/components/brave_vpn/browser",
+    "//brave/components/skus/browser",
+    "//brave/components/skus/common:mojom",
+  ]
+
+  deps = [
+    "//base",
```
bridiver, line 29: there are no `//base` headers so this dep is not needed

test/BUILD.gn
```diff
@@ -430,6 +430,9 @@ test("brave_unit_tests") {
       ]
     }
   }
+  if (enable_brave_vpn_v2) {
```
bridiver: unit tests for components belong in components/BUILD.gn test target

components/brave_vpn/browser/v2/brave_vpn_service_impl.cc
```diff
@@ -0,0 +1,204 @@
...
+
+#endif  // BUILDFLAG(IS_ANDROID)
+
+void BraveVpnServiceImpl::Shutdown() {
+  BraveVpnService::Shutdown();
+}
+
+#if !BUILDFLAG(IS_ANDROID)
+
+void BraveVpnServiceImpl::SetConnectionStateForTesting(  // IN-TEST
```
bridiver: Please do not add `ForTesting` methods until/unless they are actually needed. They should generally be avoided per chromium guidelines https://chromium.googlesource.com/chromium/src/+/HEAD/docs/patterns/fortesting-methods.md

components/brave_vpn/browser/v2/brave_vpn_service_impl.cc
```diff
@@ -0,0 +1,204 @@
...
+namespace brave_vpn {
+namespace v2 {
+
+BraveVpnServiceImpl::BraveVpnServiceImpl()
+    : connection_state_(mojom::ConnectionState::DISCONNECTED),
+      purchased_state_(mojom::PurchasedState::NOT_PURCHASED) {}
+
+BraveVpnServiceImpl::~BraveVpnServiceImpl() = default;
+
+bool BraveVpnServiceImpl::IsBraveVPNEnabled() const {
```
bridiver, line 20: all of these methods should be `NOTIMPLEMENTED();` so you don't accidentally leave them with a non-functional implementation

bridiver (changes requested):

components/brave_vpn/browser/v2/brave_vpn_service_impl.cc
alekspop26: I know that `ForTesting` methods should be avoided, but these particular ones are actually helpful. They greatly simplify all the existing tests and make them implementation-agnostic. Without those, the tests of various parts of VPN (not the service itself) must rely on the internal knowledge of the service - which would be different between V1 and V2.

components/brave_vpn/browser/v2/brave_vpn_service_impl.cc
alekspop26: Sure, that's the plan, but that would increase the size of the PR, right? And you'll ask to split again?
If that's fine, I can do that in this PR.

components/brave_vpn/browser/v2/brave_vpn_service_impl.h
```diff
@@ -0,0 +1,124 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#ifndef BRAVE_COMPONENTS_BRAVE_VPN_BROWSER_V2_BRAVE_VPN_SERVICE_IMPL_H_
+#define BRAVE_COMPONENTS_BRAVE_VPN_BROWSER_V2_BRAVE_VPN_SERVICE_IMPL_H_
+
+#include "brave/components/brave_vpn/browser/brave_vpn_service.h"
```
netzenbot, line 9: nit: `pending_remote.h` and `remote_set.h` are not used directly in this class — `mojo::PendingRemote` and `mojo::RemoteSet` don't appear in any member or method signature here. They come in transitively through `brave_vpn_service.h`. Per [CS-001](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-001), remove includes you don't actually use.

components/brave_vpn/browser/v2/brave_vpn_service_impl.cc
```diff
@@ -0,0 +1,204 @@
...
+    const std::string& validation_method,
+    const std::string& purchase_token,
+    const std::string& bundle_id) {}
+
+void BraveVpnServiceImpl::GetSubscriberCredentialV12(
+    ResponseCallback callback) {}
+
+void BraveVpnServiceImpl::RecordAllMetrics() {}
+
+void BraveVpnServiceImpl::RecordAndroidBackgroundP3A(
```
netzenbot: nit: `skus::kEnvProduction` is used here but `"brave/components/skus/browser/skus_utils.h"` is not directly included — it only arrives transitively via `brave_vpn_service.h`. Per [CS-001](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-001), include what you use directly.

components/brave_vpn/browser/v2/brave_vpn_service_impl.cc
bridiver: I missed that these are existing methods from the superclass

components/brave_vpn/browser/v2/brave_vpn_service_impl.cc
bridiver: I don't think it increases the size, it's still the same code split across more files?

test/BUILD.gn
alekspop26: I'll remove the tests for now since they're placeholders, but just wanted to mention that existing Brave VPN unit tests are not in `components/BUILD.gn`, but rather `test/BUILD.gn`, I just followed the existing pattern.
Do you think that V2-specific VPN tests should go to `components/BUILD.gn`, while the rest (shared and V1 unit tests) should live in `test/BUILD.gn`? Won't that become confusing?

components/brave_vpn/browser/v2/brave_vpn_service_impl.cc
alekspop26: Makes sense, did the split.

components/brave_vpn/browser/v2/brave_vpn_service_impl.h
alekspop26: Fixed.

components/brave_vpn/browser/v2/BUILD.gn
alekspop26: Fixed.

components/brave_vpn/browser/v2/brave_vpn_service_impl_unittest.cc
alekspop26: Fixed.

components/brave_vpn/browser/v2/BUILD.gn
alekspop26: Fixed.

components/brave_vpn/browser/v2/BUILD.gn
alekspop26: Fixed (now there is a base dep: `base/notimplemented.h`).

components/brave_vpn/browser/v2/brave_vpn_service_impl.cc
alekspop26: Fixed.

components/brave_vpn/browser/v2/brave_vpn_service_impl.h
alekspop26: Fixed.

components/brave_vpn/browser/v2/brave_vpn_service_impl.cc
alekspop26: Fixed.

test/BUILD.gn
bridiver: There are lots of existing component unit tests that are in the wrong place. All new tests for all components should go in components/BUILD.gn. Existing tests should be migrated over time. The issue is that tests in test/BUILD.gn often use Profile and other chrome classes which should not be used for component tests

alekspop26: Thanks for clarifying that, will put the V2 tests under `components` - once there are some.

bridiver: btw this does not apply to browser tests which cannot go in components because they necessarily use chrome  classes

bridiver (approved):

PR #36936, opened 2026-06-02
[VPN 2.0] Create stub Agent application

### LINKED ISSUE #54619: Implement stub Agent application
Implement stub Agent application:
- conditional build (same as BraveVpnService v2)
- setup cross-platform GN build system
- setup source tree inside brave-core
- stub main.cc
- stub unit tests running both locally and in CI
- distribution and bundling

alekspop26:
Create a stub VPN agent application that can run independently of the browser, in non-privileged mode. It is supposed to display some UI (tray icon) in the future, so on Mac, it must be placed into an app bundle. Code that will be shared with a helper app, is placed into a shared directory.

Implements part of https://github.com/brave/brave-browser/issues/54619

### DIFF
--- a/components/brave_vpn/app/v2/agent/BUILD.gn
+++ b/components/brave_vpn/app/v2/agent/BUILD.gn
@@ -1,0 +1,83 @@
+# Copyright (c) 2026 The Brave Authors. All rights reserved.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import("//brave/components/brave_vpn/common/buildflags/buildflags.gni")
+if (is_mac) {
+  import("//build/config/mac/rules.gni")
+  import("//build/util/branding.gni")
+}
+if (is_win) {
+  import("//chrome/process_version_rc_template.gni")
+}
+
+assert(enable_brave_vpn_v2)
+
+source_set("core") {
+  sources = [
+    "agent_app.cc",
+    "agent_app.h",
+  ]
+
+  public_deps = [ "//base" ]
+}
+
+if (is_win) {
+  process_version_rc_template("version_resources") {
+    sources = [ "win/agent.ver" ]
+    output = "$target_gen_dir/brave_vpn_agent_app.rc"
+  }
+}
+
+executable("agent") {
+  output_name = "brave_vpn_agent_app"
+
+  sources = [ "main.cc" ]
+  deps = [
+    ":core",
+    "//base",
+    "//base:base_static",
+  ]
+
+  if (is_mac) {
+    frameworks = [ "AppKit.framework" ]
+  }
+
+  if (is_win) {
+    deps += [
+      ":version_resources",
+      "//build/win:default_exe_manifest",
+    ]
+
+    public_configs = [ "//build/config/win:windowed" ]
+  }
+}
+
+if (is_mac) {
+  vpn_agent_product_full_name = "$chrome_product_full_name VPN Agent"
+  vpn_agent_bundle_id = "$chrome_mac_bundle_id.vpn.agent"
+
+  # Assembles the agent executable and Info.plist into an app bundle.
+  mac_app_bundle("brave_vpn_agent_bundle") {
+    info_plist = "mac/Info.plist"
+    output_name = vpn_agent_product_full_name
+    extra_substitutions = [ "MAC_BUNDLE_IDENTIFIER=$vpn_agent_bundle_id" ]
+
+    sources = [ "main.cc" ]
+    frameworks = [ "AppKit.framework" ]
+
+    deps = [
+      ":core",
+      "//base",
+      "//base:base_static",
+    ]
+  }
+
+  # Drops the assembled app into the browser framework's Helpers directory.
+  bundle_data("brave_vpn_agent_bundle_data") {
+    sources = [ "$root_out_dir/$vpn_agent_product_full_name.app" ]
+    outputs = [ "{{bundle_contents_dir}}/Helpers/{{source_file_part}}" ]
+    public_deps = [ ":brave_vpn_agent_bundle" ]
+  }
+}
--- a/components/brave_vpn/app/v2/agent/agent_app.cc
+++ b/components/brave_vpn/app/v2/agent/agent_app.cc
@@ -1,0 +1,23 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "brave/components/brave_vpn/app/v2/agent/agent_app.h"
+
+#include "base/logging.h"
+
+namespace brave_vpn {
+namespace v2 {
+
+AgentApp::AgentApp() = default;
+
+AgentApp::~AgentApp() = default;
+
+int AgentApp::Run() {
+  VLOG(1) << "Hello from the Brave VPN Agent!";
+  return 0;
+}
+
+}  // namespace v2
+}  // namespace brave_vpn
--- a/components/brave_vpn/app/v2/agent/main.cc
+++ b/components/brave_vpn/app/v2/agent/main.cc
@@ -1,0 +1,55 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "base/at_exit.h"
+#include "base/command_line.h"
+#include "base/logging.h"
+#include "base/logging/logging_settings.h"
+#include "base/process/memory.h"
+#include "brave/components/brave_vpn/app/v2/agent/agent_app.h"
+#include "build/build_config.h"
+
+namespace {
+inline constexpr char kVpnAppLogFile[] = "log-file";
+
+void InitLogging(const base::CommandLine& command_line) {
+  logging::LoggingSettings settings;
+  settings.logging_dest =
+      logging::LOG_TO_SYSTEM_DEBUG_LOG | logging::LOG_TO_STDERR;
+  base::FilePath log_file_path;
+  if (command_line.HasSwitch(kVpnAppLogFile)) {
+    settings.logging_dest |= logging::LOG_TO_FILE;
+    log_file_path = command_line.GetSwitchValuePath(kVpnAppLogFile);
+    settings.log_file_path = log_file_path.value().c_str();
+  }
+  logging::InitLogging(settings);
+}
+}  // namespace
+
+#if BUILDFLAG(IS_WIN)
+int APIENTRY wWinMain(HINSTANCE, HINSTANCE, wchar_t*, int) {
+#else
+int main(int argc, char* argv[]) {
+#endif
+  base::AtExitManager exit_manager;
+#if BUILDFLAG(IS_WIN)
+  base::CommandLine::Init(0, nullptr);
+#else
+  base::CommandLine::Init(argc, argv);
+#endif
+  const auto& command_line = *base::CommandLine::ForCurrentProcess();
+
+  base::EnableTerminationOnHeapCorruption();
+  base::EnableTerminationOnOutOfMemory();
+#if BUILDFLAG(IS_WIN)
+  base::win::RegisterInvalidParamHandler();
+  base::win::SetupCRT(command_line);
+#endif
+
+  InitLogging(command_line);
+
+  brave_vpn::v2::AgentApp agent_app;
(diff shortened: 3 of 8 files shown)


netzenbot (approved):

thypon:
Revolved the command line finding. Let's check if there is any message/cmdline passing issue, in the general review

components/brave_vpn/app/v2/agent/linux/main.cc
```diff
@@ -0,0 +1,25 @@
...
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "base/at_exit.h"
+#include "base/command_line.h"
+#include "base/process/memory.h"
+#include "brave/components/brave_vpn/v2/app/agent/agent_app.h"
+
+int main(int argc, char* argv[]) {
+  base::AtExitManager exit_manager;
+  base::CommandLine::Init(argc, argv);
```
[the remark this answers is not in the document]
bridiver: I don't know that this check is appropriate, as discussed in DM all chromium processes take command line args

components/brave_vpn/app/v2/agent/win/agent.ver
```diff
@@ -0,0 +1,3 @@
+INTERNAL_NAME=brave_vpn_agent_app
+ORIGINAL_FILENAME=brave_vpn_agent_app.exe
+PRODUCT_FULLNAME=Brave Browser VPN Agent
```
bsclifton, line 3: Something to think about - I'm not sure if each version of Brave would potentially have its own agent?
For example, if you had installed Nightly browser and then purchase VPN, does it install Nightly version of the agent?

Raising because it might need to have the channel name in the product (not only here but other places too, like macOS plist).

components/brave_vpn/app/v2/agent/BUILD.gn
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  if (is_linux) {
+    sources += [ "linux/main.cc" ]
+  }
+}
+
+if (is_mac) {
+  vpn_agent_product_full_name = "$chrome_product_full_name VPN Agent"
+  vpn_agent_bundle_id = "$chrome_mac_bundle_id.vpn.agent"
```
bsclifton, line 59: Similar to above comment (RE: channel; ex release/beta/nightly), Bundle is different per channel. I'm not sure if that is resolved with this branding variable
https://source.chromium.org/chromium/chromium/src/+/main:build/util/branding.gni

components/brave_vpn/app/v2/agent/BUILD.gn
```diff
@@ -0,0 +1,104 @@
...
+    "//base:base_static",
+  ]
+
+  if (is_mac) {
+    sources += [ "mac/main.mm" ]
+    frameworks = [ "AppKit.framework" ]
+  }
+
+  if (is_win) {
+    sources += [ "win/main.cc" ]
```
cdesouza-chromium: It is not the practice in Chromium in general to have `main.cc` split in different platform files, as `main.cc` is already supposed to be short. Is there a particular reason you considered for this split? I suspect it would be simpler to have them together and just use build flags, in this particular case, and it also goes a bit against the grain. It also doesn't look that they differ that much to gain much in being separated, which demands some shared library placement for things that could just live there with `main.cc`.

For an example, but there are plenty of cases like this: https://source.chromium.org/chromium/chromium/src/+/main:content/shell/app/shell_main.cc;l=28

https://source.chromium.org/chromium/chromium/src/+/main:chrome/enterprise_companion/main.cc;l=43

components/brave_vpn/app/v2/shared/app_utils.cc
```diff
@@ -0,0 +1,29 @@
...
+#include "brave/components/brave_vpn/app/v2/shared/app_utils.h"
+
+#include "base/logging.h"
+#include "base/logging/logging_settings.h"
+#include "brave/components/brave_vpn/app/v2/shared/switches.h"
+
+namespace brave_vpn {
+namespace v2 {
+
+void AppInitLogging(const base::CommandLine& command_line) {
```
cdesouza-chromium: Tying back to my other comment about `main.cc` being unified, this also would be a better fit as part of main, as it is common in most of the other executables in the repository, unless I'm missing something.

components/brave_vpn/app/v2/shared/switches.h
```diff
@@ -0,0 +1,19 @@
...
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#ifndef BRAVE_COMPONENTS_BRAVE_VPN_APP_V2_SHARED_SWITCHES_H_
+#define BRAVE_COMPONENTS_BRAVE_VPN_APP_V2_SHARED_SWITCHES_H_
+
+namespace brave_vpn {
+namespace v2 {
+namespace switches {
+
+inline constexpr char kVpnAppLogFile[] = "log-file";
```
cdesouza-chromium: Same on the vein of the previous comment. It looks like this could all live in a single `main.cc`, as the only existing consumer. Of course this could change, but in general it is important to let things live near where they are being used.

See https://google.github.io/styleguide/cppguide.html#Local_Variables

components/brave_vpn/app/v2/shared/app_utils.cc
alekspop26: This is a shared code and will be used in the helper. Just trying to avoid copy-paste here.

components/brave_vpn/app/v2/shared/switches.h
alekspop26: Same as above: will be used in the helper. Also, a bunch of other switches will be added in the next CR.

components/brave_vpn/app/v2/agent/win/agent.ver
alekspop26: We want to have VPN working for different channels, right? E.g. Public channel and nightly channel? In that case, they have to register their own agents and helpers, right?

components/brave_vpn/app/v2/agent/BUILD.gn
alekspop26: Yes, bundling is different per channel, that's why I decided to use the branding variable here which has the proper suffix appended.

components/brave_vpn/app/v2/agent/BUILD.gn
alekspop26: I'm trying to avoid all those #ifdefs to improve readability.
Since both agent and helper are going to have a lot of platform-specific code, I decided to introduce the split right away, even for the stub. But I'm also ok to move everything back to a single source file (I'll just have to move some macOS specific code to a dedicated MM anyway).

components/brave_vpn/app/v2/agent/win/agent.ver
alekspop26: >if you had installed Nightly browser and then purchase VPN, does it install Nightly version of the agent?

Yes, I think that's the only way it's going to work - specifically, because of Mojo scrambling. In other words, both agent, helper, and browser must be built together to work together. Hence that channel dependency.
In that approach, if the user purchased VPN in the Nightly browser and installed it, and then opened the Release browser, it is supposed to install another VPN agent, for the Release browser, automatically (while the local state is shared).

components/brave_vpn/app/v2/shared/app_utils.cc
cdesouza-chromium: At this junction, `AppInitLogging` has only one consumer, and that consumer is `components/brave_vpn/app/v2/agent/agent_app.cc`, so we follow the style guide guidance, and keep the function with internal linkage, `agent_app.cc`, in an anonymous namespace (or even better in `main`).

See: https://google.github.io/styleguide/cppguide.html#Internal_Linkage

Going back to the matter of small PRs, this also involve keeping PRs withing their own scope. We approach things this way, with incremental steps, so each change makes sense on their own. In some cases we may have no other way but to anticipate concerns, but that tends to lend its own evidence too. For this PR, it is more beneficial to keep things simple, with internal linkage, than run ahead of ourselves with architectural designs for which the demand is not present yet. If we change course tomorrow, we have less of a risk of accumulating the cost of wrong assumptions and having to or forgetting to go back and correcting things.

If the next PR requires you to use `AppInitLogging` in another place, then it will be trivial to move it to a shared location, and the need for it will be self-evident, but at this moment, I'm not sure we need that yet.

components/brave_vpn/app/v2/shared/switches.h
cdesouza-chromium: Check the other replies.

components/brave_vpn/app/v2/agent/BUILD.gn
cdesouza-chromium: We always prefer to avoid compile time `#if`s when that can be done without code duplication, and it doesn't significantly increase the total amount of code. To illustrate how this is done some times, `HeadlessContentMainDelegate` has to do what you are doing here, setting `NSApp.activationPolicy` to `NSApplicationActivationPolicyAccessory`, so in the class header they did this:

```cxx
#if BUILDFLAG(IS_MAC)
  void PlatformPreBrowserMain();
#endif
```

But the definition of that function they place in `headless/lib/headless_content_main_delegate_mac.mm`, leaning on the fact the `gn` treats suffixes like `_mac` in a special way. So they kept the macro pollution to a minimum, and yet didn't introduce duplication.

In this PR though, there is quite a lot of duplication between the three files, so it is better to have a single `main.cc` source.

components/brave_vpn/app/v2/agent/agent_app.cc
```diff
@@ -0,0 +1,35 @@
...
+namespace brave_vpn {
+namespace v2 {
+
+AgentApp::AgentApp() : initialized_(false) {}
+
+AgentApp::~AgentApp() = default;
+
+std::optional<int> AgentApp::Initialize(const base::CommandLine& command_line) {
+  CHECK(!initialized_) << "Initialize should only be called once.";
+  AppInitLogging(command_line);
```
cdesouza-chromium: Looking a bit closer in this code, I suspect we would be better off just calling `AppInitLogging` on main, and removing `Initialize` for now as we don't have a proper use for it yet (nor for `initialized_`, which I think can be removed too)

Also, bear in mind that we prefer factory functions over the use of `Init()` methods, as `Init()` methods can have undocumented half constructed states, and factory functions sidestep the need for internal states like `initialized_`. See https://abseil.io/tips/42

cdesouza-chromium (commented):
Adding a few replies... we are gonna get there. Thanks for your patience.

components/brave_vpn/app/v2/agent/win/agent.ver
bsclifton: OK this is perfect! It wasn't obvious to me that the channel was being appended. We do want per-channel VPN 🙂

components/brave_vpn/app/v2/agent/BUILD.gn
alekspop26: Thanks for the explanation. Moved everything into a single `main.cc` with compile-time #ifs.
The NSApp code is not super useful right now (until we actually have the message loop being pumped), so I removed that.

components/brave_vpn/app/v2/shared/app_utils.cc
alekspop26: Thanks, makes sense, fixed.

components/brave_vpn/app/v2/shared/switches.h
alekspop26: Done.

components/brave_vpn/app/v2/agent/agent_app.cc
alekspop26: Thanks, removed the Init() call for now, will replace with a factory method if needed in the future.

components/brave_vpn/app/v2/agent/win/agent.ver
alekspop26: Oh yes, that's quite obscure. I checked the generated plist, and indeed it is appended:
```
	<key>CFBundleDisplayName</key>
	<string>Brave Browser Development VPN Agent</string>
	<key>CFBundleExecutable</key>
	<string>Brave Browser Development VPN Agent</string>
	<key>CFBundleIdentifier</key>
	<string>com.brave.Browser.development.vpn.agent</string>
```

bsclifton: woohoo! thanks! 😄

cdesouza-chromium (approved):
LGTM

components/brave_vpn/app/v2/agent/main.cc
```diff
@@ -0,0 +1,55 @@
...
+#if BUILDFLAG(IS_WIN)
+int APIENTRY wWinMain(HINSTANCE, HINSTANCE, wchar_t*, int) {
+#else
+int main(int argc, char* argv[]) {
+#endif
+  base::AtExitManager exit_manager;
+#if BUILDFLAG(IS_WIN)
+  base::CommandLine::Init(0, nullptr);
+#else
+  base::CommandLine::Init(argc, argv);
```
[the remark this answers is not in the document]
alekspop26: The same concern has already been raised above: https://github.com/brave/brave-core/pull/36936#discussion_r3349919705
Resolving as duplicate.

bsclifton (approved):

PR #36937, opened 2026-06-02
[iOS] Stop feeding HistorySearchOptions with begin and end date

### LINKED ISSUE #55837: [iOS] Bump search history  time range
Currently it's 24h since 1.89.x release. it possibly causes users  tosee no search history if there is no search was performed in the past 24h. 
We could bump the time range to 72h or 7 days. Maybe even an option in settings for user to pick a preferred time range.

nuo-xu:
Resolves https://github.com/brave/brave-browser/issues/55837

Test:
1. open brave and search with some query via omnibox
2. forward your device date/time more than 24h
3. open brave and start typing some query in omnibox
4. verify you will still see some search histories in the list

### DIFF
--- a/ios/brave-ios/Sources/Brave/Frontend/Sync/BraveCore/History/HistoryAPIExtensions.swift
+++ b/ios/brave-ios/Sources/Brave/Frontend/Sync/BraveCore/History/HistoryAPIExtensions.swift
@@ -28,13 +28,9 @@ extension BraveHistoryAPI {
       return nil
     }
 
-    let current = Date()
     let options = HistorySearchOptions(
       maxCount: 100,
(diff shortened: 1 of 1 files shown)


kylehickinson (approved):

StephenHeaps (approved):

Uni-verse:
Verified on `iPhone 12` running `iOS 26.3.1` using version `1.93 (62)`

- Ensured that search histories are shown in the list after 24h using the test plan in https://github.com/brave/brave-core/pull/36937#issue-4573520244
- Ensured that consecutive searches are shown in the search history. 

example | example | example
--|--|--
IMG_0448 | IMG_0449 | IMG_0452

PR #36941, opened 2026-06-02
[iOS] Use OriginFilter for origin-scoped JavaScriptFeature's

kylehickinson:
This introduces a chromium src override to add additional OriginFilter's used in Brave's JavaScriptFeature's and replaces the manual host/origin checks that were being used already


netzenbot (approved):

StephenHeaps (approved):

nuo-xu (approved):

chromium_src/ios/web/js_messaging/origin_filter.mm
```diff
@@ -0,0 +1,58 @@
...
+        @"https://talk.brave.software",
+      ];
+    case web::OriginFilter::kBraveAccount:
+      return @[
+        @"https://account.brave.com",
+        @"https://account.bravesoftware.com",
+        @"https://account.brave.software",
+      ];
+    default:
+      return web::GetOriginList_Chromium(filter);
```
cdesouza-chromium: We are inserting a default case at this switch, and at the next, which means we lose the compiler check for all values in the enum.

I think it will be better to have these cases add to the top of the switch with a plaster. There are other cases like this. See https://github.com/brave/brave-core/pull/36525 for documentation about this.

chromium_src/ios/web/public/js_messaging/origin_filter.h
```diff
@@ -0,0 +1,17 @@
...
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#ifndef BRAVE_CHROMIUM_SRC_IOS_WEB_PUBLIC_JS_MESSAGING_ORIGIN_FILTER_H_
+#define BRAVE_CHROMIUM_SRC_IOS_WEB_PUBLIC_JS_MESSAGING_ORIGIN_FILTER_H_
+
+// Adds Brave specific origin filters to the OriginFilter enum
+#define kGoogleSearch \
+  kGoogleSearch, kYouTube, kBraveSearch, kBraveTalk, kBraveAccount
```
cdesouza-chromium: We should be inserting with plaster here too.

chromium_src/ios/web/js_messaging/origin_filter.mm
kylehickinson: @cdesouza-chromium Replaced the overrides with plaster here: https://github.com/brave/brave-core/pull/36941/commits/372b33cd53fe82d798281a6d777921ece09148be

goodov (approved):

stoletheminerals (approved):

PR #36952, opened 2026-06-03
Account: remove alert from `<brave-account-forgot-password-dialog>`.

szilardszaloki:
Temporarily addresses https://github.com/brave/brave-browser/issues/55638.


fmarier (approved):
Let's not close the underlying issue because we will need to make this warning conditional once we have Sync integrated.

bsclifton (approved):
string reviewers ++

PR #36954, opened 2026-06-03
[Wallet] Don't allow SignMessage for v1 tx payload

### LINKED ISSUE #55334: Wallet should disregard future signature versions in SOL cointype
Currently V1 signature type is not accepted by the SOL network (but is coming in soon via [SIMD-0385](https://github.com/solana-foundation/solana-improvement-documents/blob/cddb637c56916b7432f75d20ebf8bf74f69e794b/proposals/0385-transaction-v1.md?plain=1#L4)), the wallet will still allows signing for this V1 transaction in the signMessage format which we normally would limit. With a durable nonce transaction there's a small possibility that a user signs a durable nonce transaction and then once V1 is enabled and no new durable nonces txn gets produced then the transaction could be stored and forwarded at that time.

We should prevent `signMessage` from signing transactions in the same way we currently do for legacy/V0

Currently it has a bunch of encoding issues:

Image

supermassive:
Resolves https://github.com/brave/brave-browser/issues/55334


netzenbot (approved):

kdenhartog (approved):

components/brave_wallet/browser/solana_message.cc
```diff
@@ -489,6 +491,99 @@ std::optional<SolanaMessage> SolanaMessage::Deserialize(
                        std::move(instructions), std::move(*addr_table_lookups));
 }
 
+// static
+bool SolanaMessage::DeserializeAsV1(base::span<const uint8_t> bytes) {
```
cmazakas, line 495: This function looks good but the name is a tad confusing. `DeserializeAsV1` gave me the impression we'd actually deserialize the byte buffer into a concrete V1 object, when in reality it's just checking that the bytes conform to a specific protocol.

cmazakas (approved):

PR #36955, opened 2026-06-03
Rewrite .install.sh.toml to .install.sh.yaml

mherrmann:
Pure dev concern. The TOML format is deprecated.


cdesouza-chromium (approved):
LGTM

PR #36967, opened 2026-06-03
[Wallet]: Display Warning for Approve All Transactions

### LINKED ISSUE #24268: Display warning setApprovalForAll is called on an NFT collection
|Twitter | https://twitter.com/mulligan/status/1548383796126289921 |
-|-

`setApprovalForAll` is becoming a common exploit to drain all NFTs of an owner in a particular collection. We should display an appropriate warning when such an approval is requested.

### Tech spec

Use the `ABIDecoder` to parse the above transaction into a dedicated `TransactionType`. On the frontend, use the same warning style that we use for unlimited ERC20 approvals.

Douglashdaniel:
## Description 

Will now display a warning for `setApprovalForAll` transactions.

Resolves <https://github.com/brave/brave-browser/issues/24268>
Security <https://github.com/brave/internal/issues/1757>


components/brave_wallet_ui/components/extension/confirm_send_transaction/confirm_send_transaction.tsx
```diff
@@ -397,6 +450,15 @@ export function ConfirmSendTransaction() {
         isOpen={showEditNetworkFee}
         onCancel={() => setShowEditNetworkFee(false)}
       />
+
+      {/* Approval for all warning */}
+      {isApprovalForAllTx && showApprovalForAllWarning && (
+        <ApprovalForAllWarning
+          onConfirm={onConfirmAndCloseWarning}
+          onReject={onReject}
+          address={transactionDetails.recipient}
```
onyb: I think this is the NFT contract address. What we need to warn about here is the `spender`, which can only be figured out by parsing the calldata in core.

onyb (changes requested):

netzenbot (approved):

components/brave_wallet_ui/components/extension/confirm_send_transaction/confirm_send_transaction.tsx
onyb: Fixed here: [03a1362](https://github.com/brave/brave-core/pull/36967/commits/03a136295110ac1de53316f50a4126800dbfc1f3)

onyb (approved):
frontend++

bsclifton (approved):
string reviewers ++

supermassive (approved):

kdenhartog:
spotted a few bypasses for this, so working on some tests and confirming there's some fixes build and work first then will commit them to this branch.

kdenhartog:
I've added some additional fixes for the bypasses now that look to robustly catch the possibility of bypasses via multicalls. It would be useful for reviewers who have approved to re-look at this to make sure it's still good.

cc @onyb @supermassive @brave/string-reviewers-team

kdenhartog (approved):
This LGTM now for security after the changes I added

components/brave_wallet/browser/eth_data_parser.cc
```diff
@@ -536,6 +563,73 @@ std::optional<LiFiBridgeData> LiFiBridgeDataDecode(
...
+
+  std::vector<std::string> result;
+  for (const auto& inner_call : decoded.value()[array_index].GetList()) {
+    auto inner_bytes = PrefixedHexStringToBytes(inner_call.GetString());
+    if (!inner_bytes) {
+      continue;
+    }
+    auto found =
+        FindAllNestedSetApprovalForAll(base::span(*inner_bytes), depth + 1);
+    result.insert(result.end(), std::make_move_iterator(found.begin()),
```
supermassive: That's a potential recursion with 1K depth and it doesn't look safe then.
Should be a queue instead.

components/brave_wallet/browser/eth_data_parser.cc
```diff
@@ -536,6 +563,73 @@ std::optional<LiFiBridgeData> LiFiBridgeDataDecode(
   };
 }
 
+// Returns all operator addresses from granting setApprovalForAll calls in
+// `data`, unwrapping multicalls.
+std::vector<std::string> FindAllNestedSetApprovalForAll(
+    base::span<const uint8_t> data,
+    size_t depth) {
+  if (data.size() < 4) {
+    return {};
```
supermassive: Seems like this(and most of `return{}` below) should fail whole parsing

supermassive (changes requested):

components/brave_wallet/browser/eth_data_parser.cc
kdenhartog: Thanks, wasn't sure what the proper way to handle this was but originally it suggested depth of 5 which seemed like it would effect usability. Will get this updated.

components/brave_wallet/browser/eth_data_parser.cc
kdenhartog: Would you prefer return wallet internal error for this or like a nullopt?

components/brave_wallet/browser/eth_data_parser_unittest.cc
```diff
@@ -311,6 +311,209 @@ TEST(EthDataParser, GetTransactionInfoFromDataERC721TransferFrom) {
   EXPECT_EQ(tx_args[2], "0xf");
 }
```
onyb, line 313: @kdenhartog I couldn't find unit-tests for the `kMulticallWithDeadlineSelector` and `kMulticallWithPreviousBlockHashSelector` variants. Can you please look into it?

PR #36971, opened 2026-06-03
[CodeHealth] [Wallet] dramatically speedup RetireTxByStatus

cmazakas:
Update RetireTxByStatus to avoid several copies.

### DIFF
--- a/components/brave_wallet/browser/tx_state_manager.cc
+++ b/components/brave_wallet/browser/tx_state_manager.cc
@@ -264,13 +264,16 @@ void TxStateManager::RetireTxByStatus(base::DictValue& txs,
       (status == mojom::TransactionStatus::Confirmed ? "confirmed_time"
                                                      : "created_time");
 
-  const std::string* oldest_tx_meta_id = nullptr;
+  // Points into `txs`; left at end() until we find a candidate to retire.
+  auto oldest_it = txs.end();
   uint64_t num_txs = 0;
 
   // Everything is smaller than this.
   base::Time oldest_time = base::Time::Max();
 
-  for (const auto [tx_meta_id, value] : txs) {
+  for (auto it = txs.begin(); it != txs.end(); ++it) {
+    const auto& value = it->second;
+
     const auto* dict = value.GetIfDict();
     if (!dict) {
       continue;
@@ -304,12 +307,12 @@ void TxStateManager::RetireTxByStatus(base::DictValue& txs,
     }
 
     oldest_time = *t;
-    oldest_tx_meta_id = &tx_meta_id;
+    oldest_it = it;
   }
 
   if (num_txs > max_num) {
-    CHECK(oldest_tx_meta_id);
-    txs.Remove(*oldest_tx_meta_id);
+    CHECK(oldest_it != txs.end());
+    txs.erase(oldest_it);
   }
 }
 
--- a/components/brave_wallet/browser/tx_state_manager.cc
+++ b/components/brave_wallet/browser/tx_state_manager.cc
@@ -264,7 +264,7 @@ void TxStateManager::RetireTxByStatus(base::DictValue& txs,
       (status == mojom::TransactionStatus::Confirmed ? "confirmed_time"
                                                      : "created_time");
 
-  const std::string* oldest_tx_meta_id;
+  const std::string* oldest_tx_meta_id = nullptr;
   uint64_t num_txs = 0;
 
   // Everything is smaller than this.
--- a/components/brave_wallet/browser/tx_state_manager.cc
+++ b/components/brave_wallet/browser/tx_state_manager.cc
@@ -152,9 +152,9 @@ bool TxStateManager::AddOrUpdateTx(const TxMeta& meta) {
     if (is_add) {
       // We only keep most recent 1k confirmed plus rejected tx metas per
       // network
-      RetireTxByStatus(*update, meta.chain_id(),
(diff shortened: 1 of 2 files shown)


nuo-xu (approved):
iOS++

components/brave_wallet/browser/tx_state_manager.cc
```diff
@@ -250,24 +256,60 @@ void TxStateManager::RetireTxByStatus(const std::string& chain_id,
...
+      continue;
+    }
+
+    auto t = base::ValueToTime(time_val);
+    if (!t || *t >= oldest_time) {
+      continue;
+    }
+
+    oldest_time = *t;
+    oldest_tx_meta_id = &tx_meta_id;
```
netzenbot: If every matching tx entry is missing the `time_key` field, `oldest_tx_meta_id` is never initialized but `num_txs` can still exceed `max_num`, making `CHECK(oldest_tx_meta_id)` undefined behavior. Initialize to `nullptr` (e.g., `const std::string* oldest_tx_meta_id = nullptr;`) so the CHECK is well-defined and will fire correctly in that degenerate case.

components/brave_wallet/browser/tx_state_manager.cc
```diff
@@ -250,24 +256,60 @@ void TxStateManager::RetireTxByStatus(const std::string& chain_id,
...
+    const auto* time_val = dict->Find(time_key);
+    if (!time_val) {
+      continue;
+    }
+
+    auto t = base::ValueToTime(time_val);
+    if (!t || *t >= oldest_time) {
+      continue;
+    }
+
```
netzenbot, line 310: The loop `for (const auto [tx_meta_id, value] : txs)` copies the key into `tx_meta_id`, so `oldest_tx_meta_id = &tx_meta_id` stores a pointer to a stack variable that goes out of scope at the end of each iteration — a dangling pointer. `txs.Remove(*oldest_tx_meta_id)` then dereferences that dead pointer. Change the loop to `for (const auto& [tx_meta_id, value] : txs)` so the structured binding holds references to the actual map entries, making the pointer valid. Additionally, `oldest_tx_meta_id` is declared uninitialized; if all matching entries are missing the time key, `CHECK(oldest_tx_meta_id)` reads an indeterminate pointer (UB); initializing it to `nullptr` would at least make the CHECK predictable.

components/brave_wallet/browser/tx_state_manager.cc
```diff
@@ -250,24 +256,60 @@ void TxStateManager::RetireTxByStatus(const std::string& chain_id,
...
+  const char* time_key =
+      (status == mojom::TransactionStatus::Confirmed ? "confirmed_time"
+                                                     : "created_time");
+
+  const std::string* oldest_tx_meta_id;
+  uint64_t num_txs = 0;
+
+  // Everything is smaller than this.
+  base::Time oldest_time = base::Time::Max();
```
netzenbot, line 274: `oldest_tx_meta_id` is uninitialized. If every matching transaction entry is missing its time field (`time_val == nullptr`), the loop increments `num_txs` but never assigns `oldest_tx_meta_id`. When `num_txs > max_num`, `CHECK(oldest_tx_meta_id)` evaluates an uninitialized pointer — undefined behavior. Initialize it to `nullptr` and update the check to treat a null pointer as a logic error with a clear message, or use `std::optional<std::string_view>` to make the "not yet found" state explicit.

components/brave_wallet/browser/tx_state_manager.cc
```diff
@@ -250,24 +256,60 @@ void TxStateManager::RetireTxByStatus(const std::string& chain_id,
...
+    if (!t || *t >= oldest_time) {
+      continue;
+    }
+
+    oldest_time = *t;
+    oldest_tx_meta_id = &tx_meta_id;
+  }
+
+  if (num_txs > max_num) {
+    CHECK(oldest_tx_meta_id);
```
netzenbot: The pointer `oldest_tx_meta_id` is declared uninitialized (`const std::string* oldest_tx_meta_id;`) and is only assigned inside the loop when a valid time value is found. If all matching transactions in storage lack the `time_key` field, `num_txs` can exceed `max_num` but `oldest_tx_meta_id` is never set — making `CHECK(oldest_tx_meta_id)` undefined behavior and `txs.Remove(*oldest_tx_meta_id)` a data corruption risk. Initialize it to `nullptr` and guard the remove with `if (oldest_tx_meta_id)`. https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-027

netzenbot (approved):

components/brave_wallet/browser/tx_state_manager.cc
```diff
@@ -239,7 +244,8 @@ std::vector<std::unique_ptr<TxMeta>> TxStateManager::GetTransactionsByStatus(
   return result;
 }
 
-void TxStateManager::RetireTxByStatus(const std::string& chain_id,
+void TxStateManager::RetireTxByStatus(base::DictValue& txs,
```
supermassive: nit: I'd suggest to pass a ref to ScopedTxsUpdate highlight that mutation will be persisted

supermassive (approved):

components/brave_wallet/browser/tx_state_manager.cc
cmazakas: This is done!

PR #36981, opened 2026-06-04
Various backup results service improvements

DJAndries:
Resolves


bridiver:
Can you please update the PR to explain what exactly is changing? If there are several distinct improvements it should probably be multiple PRs

PR #36983, opened 2026-06-04
[Sidebar V2] Apply panel shadow

simonhong:

Sidebar V2 uses upstream's SidePanel, which doesn't carry Brave's
rounded-corner drop shadow. Add a shadow around the open side panel that
matches the V1 / contents-area shadow (BraveContentsViewUtil::CreateShadow)
without ever drawing outside the browser view.

Instead of attaching the shadow to the panel's own layer (which would slide
with the panel and leak past the window edge during the open/close
animation), add BraveSidePanelShadowOverlayView - a decorative sibling of the
side panel that:

* tracks the panel's bounds and visibility by observing it,
* paints the shadow into its own MasksToBounds layer clamped to the browser view,
   so the blur is never drawn outside the window, and
* is stacked just below the panel, so the panel covers the inner half of the
   blur and only the outer glow shows.

The shadow shape is inset by the panel's border insets so it hugs the visible
rounded panel (smaller than the raw panel bounds) rather than the padded
edges.

BraveSidePanelHeader now paints to its own non-opaque, top-rounded layer so it
composites above the overlay. Without a layer it rendered into the browser-view
root layer beneath the overlay, and the blur was drawn over the header instead
of behind it.

TEST=SidebarBrowserTest.SidebarV2PanelShadow

Shadow around panel when rounded corners is on
image

No shadow when rounded corners is off
image

### DIFF
--- a/browser/ui/views/frame/brave_side_panel_shadow_overlay_view.cc
+++ b/browser/ui/views/frame/brave_side_panel_shadow_overlay_view.cc
@@ -44,6 +44,10 @@ BraveSidePanelShadowOverlayView::BraveSidePanelShadowOverlayView(
   layer()->SetMasksToBounds(true);
   SetVisible(false);
 
+  auto* panel = browser_view_->side_panel();
+  CHECK(panel);
+  panel_observation_.Observe(panel);
+
   panel_shape_ = AddChildView(std::make_unique<views::View>());
 
   const int radius = views::LayoutProvider::Get()->GetCornerRadiusMetric(
@@ -53,24 +57,9 @@ BraveSidePanelShadowOverlayView::BraveSidePanelShadowOverlayView(
   shadow_->SetVisible(false);
 }
 
-BraveSidePanelShadowOverlayView::~BraveSidePanelShadowOverlayView() {
-  if (observed_panel_) {
-    observed_panel_->RemoveObserver(this);
-    observed_panel_ = nullptr;
-  }
-}
+BraveSidePanelShadowOverlayView::~BraveSidePanelShadowOverlayView() = default;
 
 void BraveSidePanelShadowOverlayView::UpdateShadowVisibility() {
-  SyncToSidePanel();
-}
-
-void BraveSidePanelShadowOverlayView::AddedToWidget() {
-  if (!observed_panel_) {
-    if (auto* panel = browser_view_->side_panel()) {
-      observed_panel_ = panel;
-      observed_panel_->AddObserver(this);
-    }
-  }
   SyncToSidePanel();
 }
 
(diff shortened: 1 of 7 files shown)


browser/ui/BUILD.gn
```diff
@@ -701,6 +701,11 @@ source_set("ui") {
         "views/side_panel/side_panel_resize_widget.cc",
         "views/side_panel/side_panel_resize_widget.h",
       ]
+
+      sources += [
+        "views/frame/brave_side_panel_shadow_overlay_view.cc",
```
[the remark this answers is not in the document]
simonhong: Deleted.

browser/ui/BUILD.gn
```diff
@@ -701,6 +701,11 @@ source_set("ui") {
...
+        "views/frame/brave_side_panel_shadow_overlay_view.h",
```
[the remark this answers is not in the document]
simonhong: deleted.

browser/ui/views/frame/brave_side_panel_shadow_overlay_view.h
```diff
@@ -0,0 +1,72 @@
...
+                               views::View* starting_view,
+                               bool is_visible) override;
+  void OnViewIsDeleting(views::View* observed_view) override;
+
+  // Syncs this view's bounds, child shape and visibility to the side panel's
+  // current bounds, visibility and rounded-corners state.
+  void SyncToSidePanel();
+
+  const raw_ref<BrowserView> browser_view_;
+  raw_ptr<views::View> observed_panel_ = nullptr;
```
sangwoo108: How about using ScopedObservation and its GetSource() and IsObservingSource() in the implementations?

browser/ui/views/frame/brave_side_panel_shadow_overlay_view.cc
```diff
@@ -0,0 +1,142 @@
...
+    observed_panel_ = nullptr;
+  }
+}
+
+void BraveSidePanelShadowOverlayView::UpdateShadowVisibility() {
+  SyncToSidePanel();
+}
+
+void BraveSidePanelShadowOverlayView::AddedToWidget() {
+  if (!observed_panel_) {
```
sangwoo108: Q. As this view never be detached after it's added to widget, this should be always true?

browser/ui/views/frame/brave_side_panel_shadow_overlay_view.cc
```diff
@@ -0,0 +1,142 @@
...
+void BraveSidePanelShadowOverlayView::OnViewIsDeleting(
+    views::View* observed_view) {
+  observed_view->RemoveObserver(this);
+  if (observed_view == observed_panel_) {
+    observed_panel_ = nullptr;
+  }
+}
+
+void BraveSidePanelShadowOverlayView::SyncToSidePanel() {
+  auto* panel = browser_view_->side_panel();
```
sangwoo108: Should this be the same with |observed_panel_|? If it should, how about using |observed_panel_| instead of getting |side_panel()| for consistency?

sangwoo108 (approved):
LGTM

browser/ui/views/frame/brave_side_panel_shadow_overlay_view.cc
simonhong: Right, and this method is deleted as we can start observing in ctor.

browser/ui/views/frame/brave_side_panel_shadow_overlay_view.cc
simonhong: Done.

browser/ui/views/frame/brave_side_panel_shadow_overlay_view.h
simonhong: Updated. much clear.

PR #36995, opened 2026-06-04
[ads][CodeHealth] Use raw_ref for ResourceComponent member

tmancey:
`resource_component_` was held as a nullable `raw_ptr` but the factory always provides a valid object. Changing it to `raw_ref` removes the null guards throughout and adds a `CHECK_DEREF` in the factory so a missing component crashes immediately.

Resolves


cdesouza-chromium (approved):
LGTM

bridiver (approved):

aseren (approved):

PR #37006, opened 2026-06-04
Upgrade from Chromium 150 to Chromium 151

### LINKED ISSUE #56094: Upgrade from Chromium 150 to Chromium 151
**Description**
Upgrade from Chromium 150 to Chromium 151

https://chromium.googlesource.com/chromium/src/+log/150.0.7871.114..151.0.7922.19?pretty=fuller&n=10000

**QA tests:**

**Desktop Affected areas:**
- Verify autoplay indicator appears correctly in the location bar (CHIP).
- Verify Brave-hosted Manifest V2 extensions can still be installed and loaded.
- Verify installing an unpacked MV2 extension shows a deprecation warning.
- Verify no MV2 "disabled extensions" dialog appears at startup.
- Verify Private and Guest windows render with the correct theme colors (title bar, toolbar, backgrounds).
- Verify custom/renamed window titles display correctly.
- Verify Tab Search opens and returns results from its entry point.
- Verify that Settings About page correctly shows Brave/Chromium versions (2 lines) and that Brave version is a clickable link.

**Desktop Known Upstream Changes:**
- Sharing hub button in the omnibox has been removed.

**iOS Affected areas:**
- Verify internal WebUI pages (brave:// / chrome://) load correctly.
- Verify "Send Tab to Self" (sending tabs between devices) works.

**Android Affected areas:**
- Verify Omnibox focus and actions hasn't regressed.
- Verify "Show Reading mode" is reachable from the app menu.
- Verify NTP infobar (https://github.com/brave/brave-core/pull/28203) still appear and function after the Android infobar refactor. It is now shown in the form of snackbar, which is expected. There is known issue with close button https://github.com/brave/brave-browser/issues/57260
- Verify the adaptive toolbar button and toolbar position (top/bottom) behave correctly.
- Verify media pause/suspend behavior works from the media notification.
- Verify translate and grid-view icons render correctly.
- New icon for tab groups tab in overview mode is expected.

cdesouza-chromium:
Resolves https://github.com/brave/brave-browser/issues/56094


zenparsing (approved):
Rewards 👍

DJAndries (approved):
p3a++

cmazakas (approved):
Wallet changes LGTM!

fmarier (approved):

patches/chrome-browser-ui-webui-help-BUILD.gn.patch
```diff
@@ -1,8 +1,16 @@
...
-index 17d8d3e2832c4d572054a7c866690703e9fe87a2..4716b9609e0d8279ca872459cec0115ee742f914 100644
+index 17d8d3e2832c4d572054a7c866690703e9fe87a2..aca3b84d13641685f2c7521293591947dea29f6c 100644
 --- a/chrome/browser/ui/webui/help/BUILD.gn
 +++ b/chrome/browser/ui/webui/help/BUILD.gn
-@@ -49,7 +49,7 @@ source_set("help") {
+@@ -8,6 +8,7 @@ source_set("help") {
+   public = [ "version_updater.h" ]
+   public_deps = [ "//base" ]
+   deps = [ "//chrome/browser/upgrade_detector" ]
++  deps += [ "//brave/browser:sparkle_buildflags" ]
```
bridiver: this should be `//brave/browser/ui/webui/help` group with sparkle_buildflags as a public dep

fallaciousreasoning (approved):
lgtm

browser/brave_wallet/brave_wallet_provider_delegate_impl_helper.cc
```diff
@@ -54,8 +53,8 @@ void UnlockWallet() {
 }
 
 void ShowWalletOnboarding(content::WebContents* web_contents) {
-  Browser* browser =
-      web_contents ? chrome::FindBrowserWithTab(web_contents) : nullptr;
+  BrowserWindowInterface* browser =
+      GlobalBrowserCollection::GetInstance()->FindBrowserWithTab(web_contents);
```
cypt4: https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/ui/browser_window/public/browser_collection.h;drc=372e20708d25bac45de7d495bdafa0a11d5aa91f;l=78

Let's check wc != null and in ShowAccountCreation

patches/chrome-browser-ui-webui-help-BUILD.gn.patch
mkarolin: In `//brave/browser/ui/webui/help` or in `//brave/chromium_src/browser/ui/webui/help`?

browser/brave_wallet/brave_wallet_provider_delegate_impl_helper.cc
mkarolin: Would https://github.com/brave/brave-core/pull/37006/commits/df9ba2d816b50e1d3ea5aff45f70f84be73d7ee9 work?

kylehickinson (approved):
ios++

rewrite/crypto/unexportable_key_software_unsecure.cc.yaml
```diff
@@ -5,13 +5,17 @@
...
+      The software provider does not support `ECDSA_SHA384` key generation, so
+      it must fall through to the `continue;  // Not supported` arm in each
+      `switch (algo)`. This adds the missing `case` to all three switches
+      (`SelectAlgorithm`, `GenerateSigningKeySlowly` and
+      `GenerateAttestationKeySlowly`), grouping it with `RSA_PSS_SHA256` which
+      already takes that path.
     re_pattern: '(switch\s*\(algo\).*?)(\n\s*continue;\s*// Not supported)'
     re_flags: [DOTALL]
-    count: 2
+    count: 3
```
bridiver, line 18: using count here seems to make this more fragile, is there a reason not to just replace all?

rewrite/chrome/android/java/src/org/chromium/chrome/browser/sync/settings/ManageSyncSettings.java.yaml
```diff
@@ -26,10 +26,13 @@ substitutions:
...
-      Force the signed-in account name to empty in updateSyncPreferences so the
-      following Google-account check is always skipped for Brave.
+      Use an empty AccountInfo for primaryAccount in updateSyncPreferences
+
+      This forces the signed-in account name passed to
+      onGoogleActivityControlsClicked to be empty, keeping Brave's Google
+      account details out of the activity controls flow.
     re_pattern:
-      '(String signedInAccountName =\n)( *)(CoreAccountInfo\.getEmailFrom\()'
+      'AccountInfo primaryAccount =
```
bridiver: why is this pattern spread across two lines?

rewrite/third_party/rust/hashbrown/v0_16/BUILD.gn.yaml
```diff
@@ -0,0 +1,26 @@
+# Copyright (c) 2026 The Brave Authors. All rights reserved.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+substitutions:
+  - description: |
+      Append Brave-specific deps and features to the hashbrown crate.
+    re_pattern: '(  proc_macro_configs \+= _configs_to_add\n)\}'
```
bridiver: for this kind of stuff I think we should be matching the target and adding things the end rather than matching some specific line in the target that could change over time

rewrite/chrome/android/java/src/org/chromium/chrome/browser/sync/settings/ManageSyncSettings.java.yaml
samartnik: this is a presubmit fix, likely due to 80-col limit for yaml files

samartnik: I shortened regex to fit in 1 line f8fbc4335eb0d17893f31e08a3e846400a17f29d

chromium_src/chrome/browser/passage_embeddings/chrome_passage_embeddings_service_controller.cc
```diff
@@ -1,61 +1,28 @@
 /* Copyright (c) 2026 The Brave Authors. All rights reserved.
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this file,
- * You can obtain one at https://mozilla.org/MPL/2.0/. */
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
```
bridiver: this change is backwards, no?

build/config/win/config.gni
```diff
@@ -5,4 +5,4 @@
 
 # True for CPU targets whose PDB can exceed the 8 GiB limit imposed by the
 # upstream default /pdbpagesize:8192.
-brave_win_use_large_pdb = target_cpu == "x86"
+brave_win_use_large_pdb = true
```
bridiver: I think we need to look into what's going on here and why we have to enable this for all builds and not just debug non-component builds

chromium_src/chrome/browser/ui/views/permissions/chip/permission_dashboard_controller.cc
```diff
@@ -17,6 +17,9 @@ bool IsAutoplay(content_settings::PageSpecificContentSettings* content_settings,
                 ContentSettingImageModel* model) {
   // If upstream adds other types to the dashboard we may need to update this
   // code.
+  if (model->image_type() == ContentSettingImageModel::ImageType::kSensors) {
```
bridiver: this seems fragile, shouldn't we just ignore everything except that ones we explicitly care about?

chromium_src/chrome/browser/extensions/manifest_v2_handler_unittest.cc
```diff
@@ -0,0 +1,15 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#include "extensions/browser/manifest_v2_handler.h"
+
+// Pass extension id as an additional parameter to
+// mv2_handler->ShouldBlockExtensionInstallation so that the extension can
+// be checked against the Brave-hosted list.
+#define ShouldBlockExtensionInstallation(...) \
```
bridiver: why are we using chromium_src for some of these and plaster for others?

browser/ui/tabs/brave_tab_features.cc
```diff
@@ -68,7 +70,14 @@ void BraveTabFeatures::Init(TabInterface& tab, Profile* profile) {
 #endif
 
 #if BUILDFLAG(ENABLE_PSST)
-  if (base::FeatureList::IsEnabled(psst::features::kEnablePsst)) {
+  // The page action controller can be null when features::kPageActionsMigration
```
bridiver, line 74: this seems overly permission then? If page_action_controller() should only be null when kPageActionsMigration then we should capture that in the conditional, not just in the comments

rewrite/.rustfmt.toml.yaml
```diff
@@ -0,0 +1,24 @@
...
+    re_pattern: '(ignore = \[.+?",\n)\]'
+    re_flags: [DOTALL]
+    replace: |-
+      \1  "brave/third_party/",
+        "brave/tools/crates/vendor/",
+        "brave/third_party/wasm/vendor/",
+      ]
+  - description: |
+      Do not format submodules.
+    re_pattern: '\]\n\Z'
```
bridiver: this is a bit of a confusing pattern match, shouldn't this just match the end of the file?

patches/components-content_settings-core-browser-content_settings_policy_provider.cc.patch
```diff
@@ -1,36 +1,53 @@
...
+index 637e76ba55111213f1175ece3a1a17e7007e48dd..bae778110f8c9f00710a45cd4f239dfddcd36d70 100644
 --- a/components/content_settings/core/browser/content_settings_policy_provider.cc
 +++ b/components/content_settings/core/browser/content_settings_policy_provider.cc
-@@ -51,6 +51,7 @@ struct PrefsForManagedContentSettingsMapEntry {
+@@ -51,6 +51,10 @@ struct PrefsForManagedContentSettingsMapEntry {
  // verifies this invariant or documents any necessary deviation.
  constexpr PrefsForManagedContentSettingsMapEntry
      kPrefsForManagedContentSettingsMap[] = {
-+        BRAVE_MANAGED_CONTENT_SETTINGS
++        {kManagedBraveShieldsDisabledForUrls,
```
bridiver, line 9: rather than convert this to plaster, it seems like we should be subclassing `content_settings::PolicyProvider` and overriding in HostContentSettingsMap?

rewrite/third_party/blink/renderer/core/html/canvas/html_canvas_element.h.yaml
```diff
@@ -0,0 +1,25 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+substitutions:
+  - description: |
+      Add Brave's synchronous toDataURL overloads.
+
+      Brave exposes toDataURL variants that take an explicit ScriptState so it
+      can be called off the main flow; declared right after the destructor.
+    re_pattern: '(  ~HTMLCanvasElement\(\) override;\n)'
```
bridiver, line 12: This is another case where I feel like we should be targeting a section of the class declaration instead of inserting after a specific method

browser/ui/messages/android/java/src/org/chromium/chrome/browser/ui/messages/snackbar/BraveSnackbarView.java
```diff
@@ -88,6 +100,191 @@ public void makeClickable(@Nullable Runnable clickCallback) {
...
+    }
+
+    /**
+     * The {@link SnackbarView} is reused across snackbars (the manager updates it in place rather
+     * than recreating it). After the base class repopulates the view, re-apply the
+     * action-below-message layout if this is the snackbar that requested it, otherwise revert it so
+     * the stacked layout and close button don't leak into an unrelated snackbar.
+     */
+    @Override
+    boolean update(Snackbar snackbar) {
```
bridiver, line 143: I'm confused about why there are so many changes for a seemingly simple upstream change that didn't touch anything outside of the code it removed?

chromium_src/chrome/browser/passage_embeddings/chrome_passage_embeddings_service_controller.cc
mkarolin: Restored in https://github.com/brave/brave-core/pull/37006/commits/b00b95b4c99ff2b387c078e1232deb7e097e172f

chromium_src/chrome/browser/ui/views/permissions/chip/permission_dashboard_controller.cc
mkarolin: Updated in https://github.com/brave/brave-core/pull/37006/commits/52b5edd377efe0d0a61b01027ddbf3755ee36f11

chromium_src/chrome/browser/extensions/manifest_v2_handler_unittest.cc
mkarolin: I think we typically do not patch/plaster tests, but we have done overrides for tests before. For non-test files, I think I only resorted to plaster files when overrides weren't feasible.

patches/chrome-browser-ui-webui-help-BUILD.gn.patch
bridiver: I feel like it should be in `//brave`, but what do we typically do in these cases?

patches/chrome-browser-ui-webui-BUILD.gn.patch
```diff
@@ -11,3 +11,12 @@ index b1c6b05cb71b6c97215c3659775d8996ad2293c0..771943c8043b91c327b6f8b96c7e20ae
  }
  
  source_set("webui") {
+@@ -333,7 +335,7 @@ source_set("webui_util") {
+ # //ui/resources only defines under "not is_android or is_desktop_android or
+ # enable_webui_ntp"; gate the target to where it was compiled inside :ui.
+ if (enable_extensions_core || enable_webui_ntp ||
+-    enable_webui_contextual_tasks_composebox) {
++    enable_webui_contextual_tasks_composebox || is_android) {
```
bridiver, line 19: for follow-up shouldn't we just be enabling `enable_webui_ntp`? That's what we're using this for anyway, right? Maybe there is other code we can cleanup?

rewrite/chrome/browser/ui/webui/BUILD.gn.yaml
```diff
@@ -12,3 +12,16 @@ substitutions:
...
       }
+  - description: |
+      Define `favicon_source` on regular Android too.
+
+      Upstream only builds FaviconSource where it was previously compiled into
+      //chrome/browser/ui:ui, which excludes regular Android. Brave's WebUIs
+      (wallet, AI chat, rewards, etc.) need FaviconSource on Android; the
+      chromium_src override of favicon_source.cc maps the default favicon
+      resources so the source compiles there.
+    re_pattern: 'if \(enable_extensions_core \|\| enable_webui_ntp \|\|\n    enable_webui_contextual_tasks_composebox\) \{\n  source_set\("favicon_source"\)'
```
bridiver: avoid patterns with fixed whitespace, looks like you already had to modify this because of it? https://github.com/brave/brave-core/pull/37006/changes/56587c7fc9f25b293b5e3b876cd3471868e77dec

browser/ui/BUILD.gn
```diff
@@ -186,11 +186,14 @@ source_set("ui") {
       "//brave/browser/brave_rewards",
       "//chrome/browser:flags",
       "//chrome/browser/lifetime",
-      "//chrome/browser/ui/webui:theme_source",
     ]
 
     public_deps += [ "//chrome/browser/ui/webui:favicon_source" ]
 
+    if (!is_android) {
+      deps += [ "//chrome/browser/ui/webui:theme_source" ]
```
bridiver, line 193: `!is_android || enable_webui_ntp` to match the guard on `theme_source`

browser/BUILD.gn
```diff
@@ -63,6 +64,29 @@ source_set("brave_global_features_impl") {
...
 }
 
+# Additions to `//chrome/browser:browser`. These sources used to be injected
+# directly into `:browser`, but `:browser` has been modularised down to zero
+# sources (and locked down against regaining any), so we mirror upstream's model
+# and keep our sources in our own leaf target that `:browser` depends on.
+# Targets with circular includes back into `:browser` aggregate must depend on
+# `:browser_public_dependencies`, which is what breaks the build-graph cycle in
+# place of using `allow_circular_includes_from`.
+source_set("browser_impl") {
```
bridiver: I'm confused about why we're using `browser_impl` here instead of following upstream `core`?

test/BUILD.gn
```diff
@@ -1031,6 +1031,7 @@ test("brave_browser_tests") {
     "//brave/components/brave_rewards/core",
     "//brave/components/email_aliases/buildflags",
     "//chrome/browser/ui:incognito_allowed_url",
+    "//chrome/browser/ui/views/translate",
```
bridiver, line 1037: ?

browser/ui/messages/android/java/src/org/chromium/chrome/browser/ui/messages/snackbar/BraveSnackbarView.java
samartnik: They removed the views we relied on for our informers, now we use snakbar UI as a base for our informers instead. The amount of changes added due to requirement to have multi-line message with a long action on its own line and a close button. This was discussed in slack with @aguscruiz

patches/ui-webui-resources-tools-bundle_js.gni.patch
```diff
@@ -1,8 +1,16 @@
...
-index dc7540cfe05c6e8779814cc432c1dbb8e0460a47..d35b4bca42d4ea323a8b302d8087f98bb3987c19 100644
+index dc7540cfe05c6e8779814cc432c1dbb8e0460a47..0c2f384c2fe5c98be830733332342d754d669186 100644
 --- a/ui/webui/resources/tools/bundle_js.gni
 +++ b/ui/webui/resources/tools/bundle_js.gni
-@@ -123,6 +123,7 @@ template("bundle_js") {
+@@ -70,6 +70,7 @@ template("bundle_js") {
+       outputs += [ "$out_folder/rollup.config.mjs" ]
+     }
+ 
++    if (true) { brave_excludes = [ "chrome://resources/brave/leo.bundle.js", "//resources/brave/leo.bundle.js", "chrome://resources/brave/opaque_ke.bundle.js", "//resources/brave/opaque_ke.bundle.js", ] if (defined(invoker.excludes)) { brave_excludes += invoker.excludes } args += [ "--exclude" ] + brave_excludes } else
```
bridiver: `brave_excludes` should be defined in a gni file instead of hard-coding here (and also fixup other uses of this same list)

rewrite/chrome/browser/ui/views/bubble/webui_bubble_manager.h.yaml
```diff
@@ -0,0 +1,20 @@
+# Copyright (c) 2026 The Brave Authors. All rights reserved.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+substitutions:
+  - description: |
```
bridiver, line 7: wouldn't this be simpler and easier to understand as a single subtitution?

rewrite/chrome/browser/ui/webui/extensions/extensions_ui.cc.yaml
```diff
@@ -0,0 +1,23 @@
...
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+substitutions:
+  - description: |
+      Add Brave's extensions resources to the WebUI data source.
+    re_pattern:
+      '(content::WebUIDataSource\*
+      CreateAndAddExtensionsSource\(Profile.+?\n)(  return source;)'
```
bridiver: Since we're matching inside a method I think this would be more clear if `{}` were included in the pattern instead of using `Profile`. We should try to have some consistent pattern for matching methods to insert at the beginning or end

bridiver: for instance here we do something similar to what I would expect https://github.com/brave/brave-core/pull/37006/changes/6f89e901f7ad6afa83a62e8a136384f087ec22d8#diff-2d04af4e01a5f954893db1e9a57b6cabfa380ebcd82b013e3a8cf5c251a59404R10

browser/ui/tabs/brave_tab_features.cc
mkarolin: Changed in 77455349de0065abb3c7dd2237aea645aa2ea8c4

patches/ui-webui-resources-tools-bundle_js.gni.patch
mkarolin: Updated in https://github.com/brave/brave-core/pull/37006/commits/aaf8ddfd7f985bd9692f7a22caed53d80dd1c66d

rewrite/crypto/unexportable_key_software_unsecure.cc.yaml
cdesouza-chromium: I left the count there, because I do want this to flag when it changes during rebase, so we can inspect it.

rewrite/third_party/rust/hashbrown/v0_16/BUILD.gn.yaml
cdesouza-chromium: Fixed in 4654e368cc8.

rewrite/.rustfmt.toml.yaml
cdesouza-chromium: Fixed in 7869af152fb

patches/components-content_settings-core-browser-content_settings_policy_provider.cc.patch
cdesouza-chromium: There are two things here about this:

1. These changes at the tail end are supposed to be strict plaster conversions, so I tried to avoid refactoring whenever possible.
2. The issue with overrides is where the source of truth is. If these arrays start being accessed through some other means, and the values should be in them, then we are having a bug in wait by moving this data out of these arrays.

rewrite/third_party/blink/renderer/core/html/canvas/html_canvas_element.h.yaml
cdesouza-chromium: I have been toying with inserting on the the destructor when we want something in the public section, as that makes things more reliable. I can't think of any other single point that clearly says "Public section". Do you have any suggestions?

patches/chrome-browser-ui-webui-BUILD.gn.patch
cdesouza-chromium: cc @samartnik

rewrite/chrome/browser/ui/webui/BUILD.gn.yaml
cdesouza-chromium: @samartnik I've pushed b28a1bc68d8, which simplifies this to: 

```
    re_pattern: 'if \(([^)]+)\) \{\n  source_set\("favicon_source"\)'
    replace: |-
      if (\1 || is_android) {
        source_set("favicon_source")
```

browser/BUILD.gn
cdesouza-chromium: Fixed in 1a42aefae00

test/BUILD.gn
cdesouza-chromium: Fixed in 1f9fc8669df and ea22b08c52b but I need to keep an eye if this is gonna get undone during rebase.

rewrite/chrome/browser/ui/views/bubble/webui_bubble_manager.h.yaml
cdesouza-chromium: Fixed in c381b64d748. You be the judge.

rewrite/chrome/browser/ui/webui/extensions/extensions_ui.cc.yaml
cdesouza-chromium: Fixed in d5b74112471

browser/ui/BUILD.gn
samartnik: fixed here cc002bf2bfaa503676a7a017ce9020fa31c077fb

patches/chrome-browser-ui-webui-BUILD.gn.patch
samartnik: created follow up issue https://github.com/brave/brave-browser/issues/57017

browser/ui/messages/android/java/src/org/chromium/chrome/browser/ui/messages/snackbar/BraveSnackbarView.java
samartnik: I forgot to add second commit that resulted to this change with actual removal of the views, I'll amend the commit message https://chromium.googlesource.com/chromium/src/+/d14807e91ac3c810fb4ed6040a7f6973002c7102

rewrite/components/sync/protocol/device_info_specifics.proto.yaml
```diff
@@ -0,0 +1,21 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+substitutions:
+  - description: |
+      Import Brave's device info specifics.
+    re_pattern: '(import "components/sync/protocol/sync_enums.proto";\n)'
+    replace: |
+      \1import "brave/components/sync/protocol/brave_device_info_specifics.proto";
+  - description: |
```
AlexeyBarabash, line 12: maybe there should be a new line before the 2nd `description:`?

AlexeyBarabash (approved):
Android LGTM

No big changes for sync owners, LGTM, verified Sync works, simple scenario Desktop<=>Android<=>Desktop bookmarks/history/passwords.

rewrite/chrome/android/java/src/org/chromium/chrome/browser/sync/settings/ManageSyncSettings.java.yaml
bridiver: oic, @cdesouza-chromium maybe we want to disable presubmit for this since the formatting could potentially mess with the regex?

rewrite/chrome/browser/ui/views/tabs/tab_style_views.cc.yaml
```diff
@@ -0,0 +1,73 @@
+# Copyright (c) 2026 The Brave Authors. All rights reserved.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+substitutions:
+  - description: |-
+      Expose TabStyleViewsImpl's private members to BraveVerticalTabStyle.
+
```
bridiver, line 9: These empty lines don't seem helpful, especially below where there are 3 of them in a row. I think it's easier to see what is substitution without them

patches/chrome-browser-BUILD.gn.patch
```diff
@@ -1,20 +1,21 @@
...
-     deps += [
-@@ -5355,6 +5356,7 @@ group("browser_public_dependencies") {
+   # This target intentionally has no sources of its own and no
+   # allow_circular_includes_from; it only aggregates dependencies. Do NOT add
+@@ -4764,6 +4765,8 @@ group("browser_public_dependencies") {
        "//components/offline_pages/core/background:background_offliner",
      ]
    }
-+  import("//brave/browser/sources.gni") public_deps += brave_chrome_browser_public_deps
++  import("//brave/browser/sources.gni")
```
bridiver: why was this patch changed to two lines?

patches/chrome-browser-password_manager-android-BUILD.gn.patch
```diff
@@ -1,16 +1,22 @@
...
  
    resources_package = "org.chromium.chrome.browser.password_manager"
 +  import("//brave/browser/password_manager/android/java_sources.gni") sources += brave_browser_password_manager_java_sources deps += brave_browser_password_manager_deps
  }
  
  generate_jni("jni_headers") {
-@@ -210,6 +211,7 @@ generate_jni("jni_headers") {
+   visibility = [
+     ":*",
++    "//brave/browser:browser_impl",
```
bridiver: what specifically triggered the need for this?

rewrite/chrome/browser/ui/views/tabs/tab_style_views.cc.yaml
cdesouza-chromium: I have been keeping the strings like `docstrings` in python, where the short subject line says **what**, while the longer description elaborates on **what** and **why**. This is a pattern I have uses everywhere, and I had even added to that document I shared with you earlier.

patches/chrome-browser-password_manager-android-BUILD.gn.patch
samartnik: Here is the error without it:
```
ERROR at //build/config/BUILDCONFIG.gn:573:5: Dependency not allowed.
    target(_target_type, target_name) {
    ^----------------------------------
The item //brave/browser:core
can not depend on //chrome/browser/password_manager/android:jni_headers
because it is not in //chrome/browser/password_manager/android:jni_headers's visibility list: [
  //chrome/browser/password_manager/android:*
  //chrome/browser:browser
  //chrome/browser/password_manager/android:jni_headers
```

patches/chrome-browser-BUILD.gn.patch
cdesouza-chromium: Fixed in cc88e264b9c, although this actually should be converted to a plaster.

rewrite/chrome/android/java/src/org/chromium/chrome/browser/sync/settings/ManageSyncSettings.java.yaml
cdesouza-chromium: I went out of my way to migrate these files to `yaml` so we would have auto-format for them 😄 between other things.

patches/chrome-browser-password_manager-android-BUILD.gn.patch
bridiver: what happens if we remove that dependency from `//brave/browser:core`? Does it trigger another error? I'm trying to find the root cause

rewrite/chrome/android/java/src/org/chromium/chrome/browser/sync/settings/ManageSyncSettings.java.yaml
cdesouza-chromium: Fixed in 1e4bc600762. Just use `re_pattern: |-`, and it prevents string formatting for regexes.

patches/ui-webui-resources-tools-bundle_js.gni.patch
bridiver: is there a follow-up to fix all the other places where we have this list hardcoded?

rewrite/third_party/blink/renderer/core/html/canvas/html_canvas_element.h.yaml
bridiver: I guess this gets a little complex with a one-off regex to match the end of the public section, but in principle something like this `(public:(?:(?!\n[ \t]*(?:private|protected)[ \t]*:|\n[ \t]*\};)[\s\S])*)`

patches/chrome-browser-password_manager-android-BUILD.gn.patch
samartnik: Then we can't include this file
```
ERROR at //brave/browser/password_manager/android/password_ui_view_android.cc:53:11: Can't include this header from here.
#include "chrome/browser/password_manager/android/jni_headers/PasswordUiView_jni.h"
```
Generally we need this dep for backporting built-in Password Manager that they removed in the upstream a while back https://github.com/brave/brave-core/commit/d7256b1461eafc3accae64690815ea60d6f15b1a

patches/components-content_settings-core-browser-content_settings_policy_provider.cc.patch
bridiver: that's reasonable

rewrite/chrome/browser/ui/views/location_bar/star_view.cc.yaml
```diff
@@ -0,0 +1,21 @@
+# Copyright (c) 2026 The Brave Authors. All rights reserved.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at https://mozilla.org/MPL/2.0/.
+
+substitutions:
+  - description: |
+      Cancel StarView::OnBubbleWidgetChanged with an early return.
```
bridiver, line 8: please remember to include the "why" as well like above https://github.com/brave/brave-core/pull/37006/changes/6a03f62e25de2cc41e7a0a0b70bc8fc98796d42f#diff-cbe42e2a66d599d8212fa065e76bd3b1772156f6637483e6d782662550f12896R10

chromium_src/chrome/browser/extensions/manifest_v2_handler_unittest.cc
bridiver: well plaster is intended to eventually replace chromium_src defines so it seems a bit odd to create new overrides and plaster patches for the same change. Also some of the overrides below are not tests.

bridiver: It seems like we're approaching the end of mv2 support though? Should we maybe be doing this through managed policy instead of this?

patches/chrome-browser-password_manager-android-BUILD.gn.patch
bridiver: It looks like we should have a dep on `//chrome/browser/password_manager/android:password_manager_android` instead which exposes this as a public dep

rewrite/chrome/browser/ui/views/tabs/tab_style_views.cc.yaml
bridiver: there are some that appear to deviate though like https://github.com/brave/brave-core/pull/37006/changes/1c2be69ddd505b82a40fba84c712868c38955262#diff-740a47a621b0a73da2f69dd4caa6a1b2ffd9aa4323d085aacf58b26fbce7ec66R35-R36 and https://github.com/brave/brave-core/pull/37006/changes/1c2be69ddd505b82a40fba84c712868c38955262#diff-740a47a621b0a73da2f69dd4caa6a1b2ffd9aa4323d085aacf58b26fbce7ec66R58 ?

build/config/win/config.gni
bridiver: cc @bsclifton

bridiver: it seems odd that we're exceeding the larger limit even for non-debug builds

rewrite/third_party/blink/renderer/core/html/canvas/html_canvas_element.h.yaml
bridiver: the above regex would probably be better as a regex macro since it's rather complex

patches/chrome-browser-password_manager-android-BUILD.gn.patch
samartnik: yep, it works, pushed fix here 597912a4faec94dc403ec2de3ab75e283af5b9fa

patches/ui-webui-resources-tools-bundle_js.gni.patch
mkarolin: I think patches/ui-webui-resources-tools-bundle_js_excludes.gni.patch is the only other one where I see hardcoded files. I can replace that one as well.

patches/chrome-browser-ui-webui-help-BUILD.gn.patch
mkarolin: I see we have `chromium_src/chrome/app/BUILD.gn` for example, but I don't know that we have a rule for it.

rewrite/chrome/browser/ui/views/tabs/tab_style_views.cc.yaml
cdesouza-chromium: The first link is wrong, but the second one is just a multi-paragraph long description although that one is getting to close on the **how**, which is what I'm trying to give guidance for the description to not be about.

patches/ui-webui-resources-tools-bundle_js.gni.patch
mkarolin: Updated in https://github.com/brave/brave-core/pull/37006/commits/92a9f0b087ea56f9c84be1d8f2eb8fcabc1a567c

rewrite/chrome/browser/ui/views/tabs/tab_style_views.cc.yaml
cdesouza-chromium: Fixed in 2714a7c7141a7fc286dd27aaf90ddfcf3a1fc7b2

rewrite/chrome/browser/ui/views/location_bar/star_view.cc.yaml
cdesouza-chromium: Fixed in e6c6deb863dce0dfcb3566f61ccef3042367210c

patches/chrome-browser-ui-webui-help-BUILD.gn.patch
mkarolin: Added in https://github.com/brave/brave-core/pull/37006/commits/be0234a4e0c8fa3a00379c3f44461ec7fd298714

chromium_src/chrome/browser/extensions/manifest_v2_handler_unittest.cc
mkarolin: There's still a discussion going on in [slack](https://bravesoftware.slack.com/archives/CU5UXM3EC/p1780612889277919) as to what exactly we are doing with this. If we go with enabling all MV2 extensions all these changes to check for specific extension ids will be unnecessary.

mkarolin: Reverted this commit in a73e546ada11bd4ecb9461be70039f791d6bfbfd and added different fix in b29c8981039cbca983df89e1e705c9122a146344

patches/chrome-browser-ui-webui-help-BUILD.gn.patch
bridiver: This accidentally went to the wrong link https://github.com/brave/brave-core/commit/be0234a4e0c8fa3a00379c3f44461ec7fd298714#r191878922

build/config/win/config.gni
mkarolin: The failure only happens for `Static` build. In `Release` build (at least locally) the pdb size for unit_tests.exe is `8,182,644,736`, while for `Static` build is claims to need `8,653,455,360`. I suppose if we want to be more precise with the patch, we could add an `else if (!is_component_build) { <our change> }` instead (or even add `&& !is_official_build`).

rewrite/third_party/blink/renderer/core/html/canvas/html_canvas_element.h.yaml
cdesouza-chromium: yeah, I'm landing slowly the `ast-grep` stuff, and then we are gonna have the full machinery in place for both ast-grep and generalised regex macros.

build/config/win/config.gni
bridiver: `!is_component_build && !is_official_build`

mkarolin: Updated in https://github.com/brave/brave-core/pull/37006/commits/0f6aab3130dfcb179fa542781b771d1a1e7f754c

bridiver (approved):

build/config/win/config.gni
mkarolin: Added a follow up issue https://github.com/brave/brave-browser/issues/57086

bsclifton (approved):
Strings LGTM++  Also briefly went over conflict patches and they look good too!

mkarolin:
**When uplifting, also uplift:**
* ✅ https://github.com/brave/brave-browser/issues/57210 (https://github.com/brave/brave-core/pull/38098)
* ✅ https://github.com/brave/brave-browser/issues/57243 (https://github.com/brave/brave-core/pull/38103)
* ✅ https://github.com/brave/brave-browser/issues/57222 (https://github.com/brave/brave-core/pull/38103)
* ✅ https://github.com/brave/brave-browser/issues/57237 (https://github.com/brave/brave-core/pull/38103)
* ✅ https://github.com/brave/brave-browser/issues/57240 (https://github.com/brave/brave-core/pull/38103)
* ✅ https://github.com/brave/brave-browser/issues/57260 (https://github.com/brave/brave-core/pull/38106)
* ✅ https://github.com/brave/brave-browser/issues/57275 (https://github.com/brave/brave-core/pull/38144)
* ✅ https://github.com/brave/brave-browser/issues/57102 (https://github.com/brave/brave-core/pull/38147)
* ✅ https://github.com/brave/brave-browser/issues/57273 (https://github.com/brave/brave-core/pull/38151)
* ✅ https://github.com/brave/brave-browser/issues/57279 (https://github.com/brave/brave-core/pull/38142)
* ✅ https://github.com/brave/brave-browser/issues/57294 (https://github.com/brave/brave-core/pull/38153)

PR #37008, opened 2026-06-05
[VPN 2.0] Add GetLastConnectionError API to BraveVpnService

### LINKED ISSUE #55762: Add GetLastConnectionError API to the service
In skus_internals_ui.cc, we make a direct call to network connection manager to get last connection error for the UI.
To support v2 architecture, we should break the unneded dependency and move the API to BraveVpnService.

- Add service AP GetLastConnectionError
- Remove v1 includes from skus_internals_ui
- Remove "//brave/components/brave_vpn/browser/connection:api" dep in the corresponding sources.gni

alekspop26:
Currently, SkusInternalsUI uses a call to the VPN connection manager to get a last VPN error. This is relevant only to Architecture 1.0, since in Architecture 2.0 there is no connection manager in the browser. Hence we need to implement an implementation-agnotic API in the BraveVpnService, to access the last VPN connection error in SkusInternalsUI. This change introduces this new API.

Resolves https://github.com/brave/brave-browser/issues/55762

### DIFF
--- a/browser/ui/BUILD.gn
+++ b/browser/ui/BUILD.gn
@@ -993,7 +993,6 @@ source_set("ui") {
     ]
   }
   if (enable_brave_vpn_v1) {
-    deps += [ "//brave/components/brave_vpn/browser/connection:api" ]
     if (is_win) {
       deps += [
         "//brave/components/brave_vpn/common/win",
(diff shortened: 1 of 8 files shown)


simonhong (approved):
++

PR #37009, opened 2026-06-05
[AI Chat]: Enable skills syncing

### LINKED ISSUE #56107: [AI Chat]: Sync AI Chat Skills
At the moment, AI Chat skills are not synced. These are stored locally as a pref, so its fairly straightforward to add them to sync.

fallaciousreasoning:
Resolves https://github.com/brave/brave-browser/issues/56107

### DIFF
--- a/chromium_src/components/sync_preferences/common_syncable_prefs_database.cc
+++ b/chromium_src/components/sync_preferences/common_syncable_prefs_database.cc
@@ -10,9 +10,14 @@
 
 #include "base/containers/fixed_flat_map.h"
 #include "base/containers/map_util.h"
+#include "brave/components/ai_chat/core/common/buildflags/buildflags.h"
 #include "brave/components/containers/buildflags/buildflags.h"
 #include "brave/components/email_aliases/buildflags/buildflags.h"
 #include "components/search_engines/search_engines_pref_names.h"
+
+#if BUILDFLAG(ENABLE_AI_CHAT)
+#include "brave/components/ai_chat/core/common/pref_names.h"
+#endif
 
 #if BUILDFLAG(ENABLE_CONTAINERS)
 #include "brave/components/containers/core/browser/pref_names.h"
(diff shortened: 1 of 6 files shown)


chromium_src/components/sync_preferences/common_syncable_prefs_database.cc
```diff
@@ -80,6 +88,20 @@ constexpr auto kBraveCommonSyncablePrefsAllowlist = base::MakeFixedFlatMap<
...
         },
     },
+#endif
+#if BUILDFLAG(ENABLE_AI_CHAT)
+    {
+        ai_chat::prefs::kBraveAIChatSkills,
+        {
+            brave_syncable_prefs_ids::kBraveAIChatSkills,
+            syncer::PREFERENCES,
+            sync_preferences::PrefSensitivity::kNone,
```
fallaciousreasoning, line 98: @brave/sec-team does this make sense for the sensitivity?

petemill:
I haven't made this kind of change myself, so would be good to get review by someone in @brave/sync-reviewers, and couldn't hurt to get an explicit security review.

chromium_src/components/sync_preferences/common_syncable_prefs_database.cc
ShivanKaul: Skills are just saved prompts, right? Seems correct.

ShivanKaul:
I'm not opposed in general, but why are we bundling this in with `Settings`? I don't think this is very intuitive. Left a comment here: https://github.com/brave/reviews/issues/2203#issuecomment-4635894683

components/ai_chat/core/common/DEPS
```diff
@@ -1,4 +1,5 @@
 include_rules = [
+  "+components/pref_registry",
```
netzenbot, line 2: Adding `+components/pref_registry` to `common/DEPS` violates the [common/ dep rule](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-071): `PrefRegistrySyncable` is a browser-process concept and `common/` should only depend on deps safe for every process (`//base`, mojom, shared structs). The syncable flag constant is only meaningful in the browser process; consider moving the syncable pref registration into `browser/` code where `PrefRegistrySyncable` is appropriate.

PR #37020, opened 2026-06-05
[Sidebar V2] Use window-corner radius when panel is at window edge

simonhong:

The side panel's content used a single radius on all corners. When the
panel sits flush against the window edge (the sidebar UI is hidden), its
inner bottom corner is a window corner and should use the larger
window-corner radius instead.

GetPanelContentsRoundedCorners() now takes a BrowserWindowInterface so it
can reach both the prefs and the BraveBrowserView. It applies the
window-corner radius to the panel's inner bottom corner (lower-right when
right-aligned, lower-left when left-aligned) only while the sidebar is
hidden, and the regular radius otherwise.
  
Sidebar UI visibility can change at runtime (show-on-mouse-over, pinning),
so SidebarContainerView runs a callback whenever the control view's
visibility changes; BraveBrowserView wires it to SidePanel::UpdateBorder()
to recompute the corners.

TEST=SidebarBrowserTest.SidebarV2GetPanelContentsRoundedCorners,
           SidebarBrowserTest.SidebarV2ContentCornersFollowSidebarVisibility,
           SidePanelUtilsTest.*

https://github.com/user-attachments/assets/dfb9cbe8-1d86-4704-9b02-c01cac0d4fd4

Screenshot 2026-06-08 at 14 33 10 Screenshot 2026-06-08 at 14 34 46

### DIFF
--- a/browser/ui/views/side_panel/side_panel_utils.cc
+++ b/browser/ui/views/side_panel/side_panel_utils.cc
@@ -5,10 +5,17 @@
 
 #include "brave/browser/ui/views/side_panel/side_panel_utils.h"
 
+#include "brave/browser/ui/views/frame/brave_browser_view.h"
 #include "brave/common/pref_names.h"
+#include "chrome/browser/ui/browser_window/public/browser_window_interface.h"
 #include "chrome/browser/ui/views/chrome_layout_provider.h"
+#include "chrome/common/pref_names.h"
 #include "components/prefs/pref_service.h"
 #include "ui/views/layout/layout_provider.h"
+
+using views::ShapeContextTokensOverride::kRoundedCornersBorderRadius;
+using views::ShapeContextTokensOverride::
+    kRoundedCornersBorderRadiusAtWindowCorner;
 
 namespace brave {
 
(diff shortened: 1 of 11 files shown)


browser/ui/views/side_panel/side_panel_utils.cc
```diff
@@ -5,37 +5,73 @@
...
-                                                    bool has_header) {
-  // When Brave's rounded corners are off, the panel has no rounded border, so
-  // the content shouldn't be rounded either.
-  if (!prefs->GetBoolean(kWebViewRoundedCorners)) {
+gfx::RoundedCornersF GetPanelContentsRoundedCorners(
+    BrowserWindowInterface* browser_window_interface,
+    bool has_header) {
+  auto* prefs = browser_window_interface->GetProfile()->GetPrefs();
+  auto* browser_view = BraveBrowserView::From(
+      BrowserView::GetBrowserViewForBrowser(browser_window_interface));
```
netzenbot: The comment says 'Can null during the startup' but the unit test explicitly documents that this null only occurs in unit-test environments ('BrowserView::GetBrowserViewForBrowser() returns null in unit-test environments (no real native window)'). Per [ARCH-024](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-024), add `CHECK_IS_TEST()` before the early return to make the test-only nature explicit and prevent confusion about whether null is a valid production state.

browser/ui/views/side_panel/side_panel_utils.cc
```diff
@@ -5,37 +5,73 @@
...
 
+using views::ShapeContextTokensOverride::kRoundedCornersBorderRadius;
+using views::ShapeContextTokensOverride::
+    kRoundedCornersBorderRadiusAtWindowCorner;
+
 namespace brave {
 
 bool ShouldShowSidePanelHeader(SidePanelEntryId id) {
   return id == SidePanelEntryId::kReadingList ||
          id == SidePanelEntryId::kBookmarks;
```
netzenbot, line 24: nit: `GetPanelContentsRoundedCorners` extracts both `PrefService*` and `BraveBrowserView*` from `BrowserWindowInterface*`, using it as a bag-of-stuff object rather than the most specific dependencies needed. Per [ARCH-003](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-003), consider passing `PrefService*` and `BraveBrowserView*` directly.

browser/ui/views/side_panel/side_panel_utils.cc
```diff
@@ -5,37 +5,73 @@
...
 
-gfx::RoundedCornersF GetPanelContentsRoundedCorners(PrefService* prefs,
-                                                    bool has_header) {
-  // When Brave's rounded corners are off, the panel has no rounded border, so
-  // the content shouldn't be rounded either.
-  if (!prefs->GetBoolean(kWebViewRoundedCorners)) {
+gfx::RoundedCornersF GetPanelContentsRoundedCorners(
+    BrowserWindowInterface* browser_window_interface,
+    bool has_header) {
+  auto* prefs = browser_window_interface->GetProfile()->GetPrefs();
```
netzenbot: nit: The merged condition only explains the `!browser_view` arm but drops the explanation for `!prefs->GetBoolean(kWebViewRoundedCorners)`. The old code had a clear comment for that path. Consider either separating them into two `if` statements with individual comments, or expanding the comment to cover both conditions (e.g. "Null during startup, or rounded corners pref is disabled."). ([Document Non-Obvious Failure Branches](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-021))

browser/ui/views/side_panel/side_panel_utils.cc
```diff
@@ -5,37 +5,73 @@
...
+  }
+
+  // When the sidebar is visible it sits between the panel and the window edge,
+  // so the panel's bottom corner is not a window corner — use regular radius.
+  if (browser_view->IsSidebarVisible()) {
+    return rounded_corners;
+  }
+
+  // When the sidebar is hidden the panel touches the window edge directly. The
+  // inner bottom corner (lower-right for a right-aligned panel, lower-left for
```
netzenbot, line 58: nit: `const auto rounded_corners_border_radius_at_window_corner` stores an `int` (the return type of `GetCornerRadiusMetric`). Prefer `const int` here — it's short and descriptive, and `auto` adds nothing. See [CS-053](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-053).

browser/ui/views/side_panel/side_panel_utils.cc
simonhong: Updated comment.

browser/ui/views/side_panel/side_panel_utils.cc
simonhong: deleted test only condition as it doesn't happen in production.

browser/ui/views/side_panel/side_panel_utils.cc
simonhong: Updated.

browser/ui/views/side_panel/side_panel_utils.cc
simonhong: Updated to receive BrowserView* only.

browser/ui/views/sidebar/sidebar_container_view.h
```diff
@@ -86,6 +87,11 @@ class SidebarContainerView : public sidebar::Sidebar,
   bool IsFullscreenForCurrentEntry() const;
   void UpdateBorder();
 
+  // Runs `callback` whenever the sidebar control view's visibility changes, so
+  // the side panel can recompute its content corner radii (which depend on
+  // sidebar visibility — see GetPanelContentsRoundedCorners()).
+  void SetSidebarVisibilityChangedCallback(base::RepeatingClosure callback);
```
sangwoo108: nit: I think the name of this can be misleading. Maybe SetSidebarControlViewVisibilityChangedCallback() could be better.

sangwoo108 (approved):
LGTM

browser/ui/views/sidebar/sidebar_container_view.h
simonhong: Renamed. need to f/u other places also.

goodov (approved):

PR #37031, opened 2026-06-05
[VPN 2.0] Revise BraveVpnServiceObserver

alekspop26:
BraveVPNServiceObserver is an existing observer that works with VPN Architecture 1.0. It can be reused for Architecture 2.0 mostly as-is, with minor changes.

This PR renames BraveVPNServiceObserver to BraveVpnServiceObserver to follow Chromium's naming convention for acronyms. In class names one must follow the rule that acronyms are treated like words.

Implements part of https://github.com/brave/brave-browser/issues/54599


netzenbot (approved):

simonhong (approved):
++

PR #37033, opened 2026-06-05
[Polkadot] [Wallet] update suggested Polkadot account names to be accurate

cmazakas:
Right now when a user attempts to create a Polkadot account the suggested name will present an account index that totals _all_ Polkadot accounts. 

In reality, users either create a testnet or mainnet Polkadot account and the UI needs to reflect this accordingly as internally, we only have one coin type that represents Polkadot: DOT. 

It would be unsustainable to introduce a new coin type for each parachain so we simply update the account counting logic at each call-site.


netzenbot (approved):

Douglashdaniel (approved):
++ Wallet front-end

PR #37035, opened 2026-06-05
[Polkadot] [Wallet] update polkadot account descriptions

cmazakas:
Update the displayed wallet strings to show "Polkadot Testnet" for testnet accounts.

This is useful because we're now supporting parachains, each of which can have their own unique names which makes the usage of "Polkadot Westend" confusing for users.


mkarolin (approved):
`strings`++

PR #37036, opened 2026-06-05
[Android] Prompt restart when an Origin Play Store purchase is auto-restored

### LINKED ISSUE #56078: Crash on opening a page after buying Origin
### Description

The browser crashes when user opens a page after buying Origin without restarting the browser. We don't force users to restart the browser, we only ask.

The browser always fetches the origin key on first launch in case of Play Store purchase, even for first installation. 
If the key already exists, we apply it and go into the state that crashes. 

IMPORTANT: Browser restart fixes the issue.

#### Testcase 1 - reinstall

1. Install the browser
1. Buy origin via play store
1. Restart
1. Uninstall Brave
1. Install Brave
1. Run Brave
1. Search `test` through address bar OR open nytimes.com

#### Testcase 2 - purchased origin without restart

1. Install the browser
1. Buy origin via play store (do not restart)
1. Search `test` through address bar OR open nytimes.com

### Actual result

Page crashes

[Crash Reports.zip](https://github.com/user-attachments/files/28574483/Crash.Reports.zip)

Image

Image

I tried to upload the crash but it was stuck

Image

### Expected result

Page does not crash

### Reproduces how often

Easily reproduced

### Brave version

Reproduced on:

BraveMonoarm.apk v1.91.169 C149  on Lenovo TB-8506FS (Android 11.0)
BraveMonoarm.apk v1.91.164 C149  on Lenovo TB-8506FS (Android 11.0)

Does not reproduce on

BraveMonoarm.apk v1.91.163 C148 on Lenovo TB-8506FS (Android 11.0)

cc @aseren @tmancey

SergeyZhukovsky:
On a fresh profile with a prior Play Store Origin purchase, `BraveActivity` auto-restores it at startup via `verifyPurchase()`. This enforces the Origin policies (e.g. disabling Brave Wallet) live in the running session, but unlike the in-settings purchase and the native web-purchase flows it never prompted a restart. The browser then honours the freshly enforced policy per navigation while already-configured renderers do not, which can leave the two out of sync (e.g. the renderer keeps injecting `window.ethereum` and requests a binder the browser no longer registers).

Open the Origin settings screen when a Play Store purchase is auto-restored, reusing the existing fetching-spinner/restart snackbar, so the user restarts and the policies take effect cleanly at boot. The prompt fires only on the auto-restore transition (the subscription pref is inactive when `verifyPurchase` runs); the in-settings first purchase keeps its own snackbar.

Resolves: https://github.com/brave/brave-browser/issues/56078

### DIFF
--- a/android/java/org/chromium/chrome/browser/brave_origin/BraveOriginSettingsLauncherHelper.java
+++ b/android/java/org/chromium/chrome/browser/brave_origin/BraveOriginSettingsLauncherHelper.java
@@ -16,13 +16,14 @@ import org.chromium.chrome.browser.settings.BraveOriginPreferences;
 import org.chromium.chrome.browser.settings.SettingsNavigationFactory;
 
 /**
- * Opens the Brave Origin settings screen from native when a purchase is first detected, so the user
- * can restart to apply the newly enforced policies.
+ * Opens the Brave Origin settings screen when a purchase is first detected (from native for web
+ * purchases, or from Java when a Play Store purchase is auto-restored), so the user can restart to
+ * apply the newly enforced policies.
  */
 @NullMarked
 public class BraveOriginSettingsLauncherHelper {
     @CalledByNative
-    private static void showOriginSettingsForRestart() {
+    public static void showOriginSettingsForRestart() {
         Activity activity = ApplicationStatus.getLastTrackedFocusedActivity();
         if (activity == null) {
             return;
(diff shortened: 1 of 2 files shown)


samartnik (approved):
++

btlechowski:
Verified with v1.93.38 C149 on Lenovo TB-8506FS (Android 11.0)


Verified test plan from https://github.com/brave/brave-browser/issues/56078

Important: the crash in the original issue is still reproducible. Only a prompt to restart the browser is shown

Verified prompt to restart is shown after Origin Play store purchase is auto restored

image

Verified tapping Restart actually restart the browser
image

Verified no crash when opening pages
image

PR #37062, opened 2026-06-07
[ai-chat][msan] `ConversationHandler::Observer` string use

cdesouza-chromium:
This PR changes some of the events in `ConversationHandler::Observer` to
just use `const std::string&` for cases where `std::string` uses cannot
be completely removed form the call flow, so using `std::string_view`
can becomes a source of extra copies, and passing values a `std::string`
reference simplifies things.

Resolves https://github.com/brave/internal/issues/1763
Resolves https://github.com/brave/internal/issues/1765
Resolves https://github.com/brave/internal/issues/1764
Resolves https://github.com/brave/internal/issues/1766
Resolves https://github.com/brave/internal/issues/1767
Resolves https://github.com/brave/internal/issues/1768

### DIFF
--- a/components/ai_chat/core/browser/conversation_handler.h
+++ b/components/ai_chat/core/browser/conversation_handler.h
@@ -79,33 +79,31 @@ class ConversationHandler : public mojom::ConversationHandler,
 
   class Observer : public base::CheckedObserver {
    public:
-    ~Observer() override {}
-
     // Called when the conversation history changess
     virtual void OnRequestInProgressChanged(ConversationHandler* handler,
-                                            bool in_progress) {}
+                                            bool in_progress) = 0;
     virtual void OnConversationEntryAdded(
         ConversationHandler* handler,
         mojom::ConversationTurnPtr& entry,
-        std::optional<PageContents> associated_content_value) {}
+        std::optional<PageContents> associated_content_value) = 0;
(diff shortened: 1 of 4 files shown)


netzenbot (approved):

atuchin-m (approved):

components/ai_chat/core/browser/conversation_handler.h
```diff
@@ -79,33 +79,31 @@ class ConversationHandler : public mojom::ConversationHandler,
...
         const std::string& conversation_uuid,
-        const std::string& title) {}
+        const std::string& title) = 0;
     virtual void OnConversationTokenInfoChanged(
         const std::string& conversation_uuid,
         uint64_t total_tokens,
-        uint64_t trimmed_tokens) {}
-    virtual void OnAssociatedContentUpdated(ConversationHandler* handler) {}
+        uint64_t trimmed_tokens) = 0;
+    virtual void OnAssociatedContentUpdated(ConversationHandler* handler) = 0;
```
petemill, line 106: This is an observer class, these functions should not have been changed to be made required

cdesouza-chromium: https://github.com/brave/brave-core/pull/37097

Thanks for pointing that out.

PR #37063, opened 2026-06-07
Android: Keep HTML fullscreen on Home press

### LINKED ISSUE #36825: Youtube videos don't switch to landscape when activating full screen
## Description 

Previously, when watching YouTube on android and switching to full screen, the view changed automatically to landscape, even when the phone doesn't switch to landscape when horizontal. Now, it stays in portrait mode. So, in order to have the video show on the full screen instead of a small portion of it, I have to enable automatic rotation, which I really don't like, and rotate the phone after switching to full screen.

So, in short, it forces me to enable automatic rotation, which I really don't like.

Note: when I open a video in Rumble, I get the expected behavior of showing the video in landscape mode.

See this thread: https://community.brave.com/t/youtube-full-screen-stays-in-portrait-mode/536178

`Mattches` replied in that thread that this is an upstream issue from chromium. I suggest to add a specific fix for youtube, since watching youtube videos without ads is a killer feature of brave, and since I wouldn't be surprised if a fix won't come from upstream, since this change in behavior might not be a real bug, and it's in the interest of Google to make the web experience of youtube slightly more annoying.

## Steps to reproduce 

   1. Have the phone in portrait mode.
   2. Go to youtube, open a video.
   3. Click the `[ ]` button to start fullscreen mode

## Actual result 

The video is in full screen, but the phone remains in portrait mode, so the video is very small.

## Expected result

The phone should switch to landscape mode, so the video will fill the entire screen.

## Issue reproduces how often 

Always.

## Version/Channel Information:

- Can you reproduce this issue with the current Play Store version? Yes.
- Can you reproduce this issue with the current Play Store Beta version? 
- Can you reproduce this issue with the current Play Store Nightly version?

## Device details
   - Install type (ARM, x86): ARM
   - Device type (Phone, Tablet, Phablet): Phone
   - Android version: 13; Build/TP1A.220624.014 (Galaxy A51, One UI 5.1)
   
## Brave version

1.63.169, Chromium 122.0.6261.111

### Website problems only
- Does the issue resolve itself when disabling Brave Shields? Yes
- Does the issue resolve itself when disabling Brave Rewards? Yes
- Is the issue reproducible on the latest version of Chrome? Probably (according to comment saying it's upstream)

### LINKED ISSUE #43067: Tapping full screen button in YouTube not working properly sometimes.
### Description

While watching videos in YouTube. Sometimes while tapping the full screen button, the video is not stretching to full screen. Rather it's getting stuck to a corner of the screen _(shown in image)_. Though it's not happening every time, it's frequent.
Temporary solution: rotating the phone 180° making the video full screen.

### Steps to reproduce

1. Open https://m.youtube.com/
2. Select any video and play it.
3. Then tap the ful screen button on the down-right corner.
4. You'll see the video is shrinked to the a corner of the screen. While the rest of the screen is black.
5. Rotating the phone 180° may solve it for that specific video.

### Actual result

Image

Video is not stretched to full screen.

### Expected result

The video should ne stretched completely over the screen.Image

### Reproduces how often

Intermittent issue

### Brave version

Brave - Beta 1.74.36, Chromium 132.0.6834.57

### Device

- Brand/model: Samsung M32
- Android version: 13
- One UI version: 5.1

### Channel information

- [ ] release (stable)
- [x] beta
- [ ] nightly

### Reproducibility

### Miscellaneous information

_No response_

rushabhT3:
Fixes YouTube fullscreen exiting when pressing Home button. Overrides onActivityStateChange to preserve persistent fullscreen mode during video playback.

Resolves brave/brave-browser#43067
Resolves brave/brave-browser#36825

Resolves

### DIFF
--- a/chromium_src/chrome/android/java/src/org/chromium/chrome/browser/fullscreen/FullscreenHtmlApiHandlerBase.java
+++ b/chromium_src/chrome/android/java/src/org/chromium/chrome/browser/fullscreen/FullscreenHtmlApiHandlerBase.java
@@ -1,0 +1,36 @@
+// Copyright 2026 The Brave Authors
+// Copyright 2024 The Chromium Authors
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+package org.chromium.chrome.browser.fullscreen;
+
+import android.app.Activity;
+
+import org.chromium.base.ActivityState;
+import org.chromium.base.supplier.NonNullObservableSupplier;
+import org.chromium.chrome.browser.multiwindow.MultiWindowModeStateDispatcher;
+
+public class FullscreenHtmlApiHandlerBase
+        extends org.chromium.chrome.browser.fullscreen.FullscreenHtmlApiHandlerBase {
+
+    public FullscreenHtmlApiHandlerBase(
+            Activity activity,
+            NonNullObservableSupplier<Boolean> areControlsHidden,
+            boolean exitFullscreenOnStop,
+            MultiWindowModeStateDispatcher multiWindowDispatcher) {
+        super(activity, areControlsHidden, exitFullscreenOnStop, multiWindowDispatcher);
+    }
+
+    @Override
+    public void onActivityStateChange(Activity activity, int newState) {
+        // BRAVE FIX: keep fullscreen when Home is pressed during video
+        if (newState == ActivityState.STOPPED && mExitFullscreenOnStop) {
+            if (mTabInFullscreen != null && mTabInFullscreen.isPlayingMedia()) {
+                // Don't exit — preserve fullscreen for PIP/background play
+                return;
+            }
+        }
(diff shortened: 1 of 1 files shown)


rushabhT3:
@brave/android-reviewers @brave/chromium-src-reviewers

This PR fixes YouTube (and any HTML5 video) exiting fullscreen when pressing the Home button on Android.

Root cause: upstream `FullscreenHtmlApiHandlerBase.onActivityStateChange()` calls `exitFullscreen()` on Activity STOPPED. This override keeps persistent fullscreen alive if media is playing, allowing proper PIP/background transition instead of a forced exit.

CI needs approval to run — first-time contributor. Happy to adjust the check (currently using `isPlayingMedia()`) if there's a better signal.

rushabhT3:
@brave-builds can you provide a test APK for android?

PR #37065, opened 2026-06-08
[git-cr] Adding `git cr commit --fixup=drop`

cdesouza-chromium:
This is just a shortcut to `brockit drop`, which provides a commit to
serve as a rebase event to drop a target commit.


emerick (approved):

AlexeyBarabash (approved):
++

PR #37068, opened 2026-06-08
Add best practice: don't bind unowned view types into callbacks

bridiverbot:
## Summary

Adds a new memory/lifetime best practice (CSM-037): don't bind _unowned_
"view" types into `base::BindOnce` / `base::BindRepeating`.

`base::Bind*` already rejects the unsafe raw-pointer/reference cases at compile
time (forcing `base::Unretained`, refusing raw pointers to ref-counted types,
etc.), but it does **not** catch non-owning view types like `std::string_view`
and `base::span<T>`. `Bind*` copies a view by value — pointer + length, not the
data — so an unowned view dangles if nothing keeps its backing storage alive
until the callback runs. The guidance is to bind an owning copy (`std::string`,
`std::vector<T>`) instead.

Exception documented: passing a view is fine when the caller owns the backing
store and guarantees it outlives the callback.

## Notes

- Docs-only change under `docs/best-practices/`.

### DIFF
--- a/docs/best-practices/coding-standards-memory.md
+++ b/docs/best-practices/coding-standards-memory.md
@@ -762,3 +762,77 @@ pointing to string literals (if they may point to heap-allocated objects, use
 Blink renderer code using Oilpan, and any other code whose objects are allocated
 outside PartitionAlloc (V8 heap, Java heap, etc.), cannot use `raw_ptr<T>` or
 `raw_ref<T>`.
+
+---
+
+<a id="CSM-037"></a>
+
+## ❌ Don't Bind Unowned View Types into Callbacks
+
+**`base::Bind*` already rejects the unsafe raw-pointer and reference cases at
+compile time (it forces `base::Unretained`, refuses raw pointers to
+ref-counted types, etc.), but it does _not_ catch non-owning "view" types like
+`std::string_view` and `base::span<T>`.** `base::BindOnce` /
+`base::BindRepeating` copy bound arguments by value, and copying a view copies
+only its pointer and length — not the underlying data. If nothing keeps the
+backing storage alive until the callback runs, the view dangles and you get a
+use-after-free. Bind an owning copy (`std::string`, `std::vector<T>`) instead.
+
+```cpp
+// ❌ WRONG - string_view bound into a posted task; backing buffer may be gone
+void Schedule(std::string_view name) {
+  task_runner_->PostTask(
+      FROM_HERE,
+      base::BindOnce(&MyClass::OnReady, weak_factory_.GetWeakPtr(), name));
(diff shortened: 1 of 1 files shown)


docs/best-practices/coding-standards-memory.md
```diff
@@ -762,3 +762,66 @@ pointing to string literals (if they may point to heap-allocated objects, use
...
+// ✅ ALSO CORRECT - the view points into a member this object owns, and the
+// bound WeakPtr ensures the callback only runs while `this` (and therefore
+// `buffer_`) is still alive. No copy needed.
+class Decoder {
+ public:
+  void Start() {
+    task_runner_->PostTask(
+        FROM_HERE,
+        base::BindOnce(&Decoder::Decode, weak_factory_.GetWeakPtr(),
+                       base::span(buffer_)));  // view into owned member
```
cdesouza-chromium: For a case like this, we should be using `base::raw_span` rather than a naked span.

https://source.chromium.org/chromium/chromium/src/+/main:base/memory/raw_span.h

docs/best-practices/coding-standards-memory.md
```diff
@@ -762,3 +762,66 @@ pointing to string literals (if they may point to heap-allocated objects, use
...
+```
+
+The same applies to `base::span<T>` — bind an owning container
+(`std::vector<T>`) rather than the span. Prefer making the callback target take
+an owning type (`const std::string&`, `std::vector<T>`) so the bound copy is
+forced to own its data.
+
+**Exception:** Passing a view is fine when the caller owns the backing store and
+guarantees it outlives the callback. The hazard is specifically an _unowned_
+view: one whose backing storage no one is keeping alive until the callback runs.
```
cdesouza-chromium: The exception should be either the use of `raw_span`, or the storage being static, which applies for `span` to arrays, or strings.

cdesouza-chromium (commented):
LGTM % few things to change

docs/best-practices/coding-standards-memory.md
bridiverbot: Done — the example now binds `base::raw_span<const uint8_t>(buffer_)` instead of a naked span, so the stored view gets dangling-pointer protection.

docs/best-practices/coding-standards-memory.md
bridiverbot: Reworked the exception into two explicit cases: (1) use `base::raw_span` rather than a naked `base::span`, and (2) the backing storage being static (a `span` over a static/constexpr array, or a `string_view` of a string literal). Added a static-storage example alongside the raw_span one.

cdesouza-chromium (approved):

PR #37069, opened 2026-06-08
[AI Chat]: Fix completions being split by inline search results

### LINKED ISSUE #56169: [AI Chat]: When receiving results for an inline search event, content up to the event is split and hidden inside the `Thinking` expand/collapse
1. Enter a query like "CAN you search for the top bars in "

If the response has `inline-search` results, they should not be rendered inside the `Task Complete` expand/collapse

Bad (expanded): 
Image

Bad (collapsed):
Image

This fixes an issue that was largely made more visible with https://github.com/brave/brave-core/pull/35757

fallaciousreasoning:
Resolves https://github.com/brave/brave-browser/issues/56169

Solution is to not split on `inline-search` events.


netzenbot (approved):

petemill (commented):
I have a couple things we might want to consider:
1. That `InlineSearchEvent` might not be the only ignorable event. Anything that isn't LLM-generated (which is "single-threaded") and is supplied by a separate stream on the aichat server could be ignored, e.g. `SearchQueriesEvent`, `SearchStatusEvent`?

2. Perhaps completion isn't the only stream that can be interrupted. What if the LLM is in the middle of a tool call generation and the inline search results come back at that point? Looks like completion and tool use are the only 2 events that we combine, so maybe we should apply this logic to both?

components/ai_chat/core/browser/conversation_handler.cc
```diff
@@ -1324,36 +1325,58 @@ void ConversationHandler::UpdateOrCreateLastAssistantEntry(
...
+      // from this delta is inline search results. Inline search events are
+      // rendered inline within the completion text (matched by query), so they
+      // must not split one response into multiple completion events, which
+      // would fragment the answer into separate task steps and render only the
+      // trailing fragment as the final response. Any other event type (e.g.
+      // tool use) is a genuine boundary that still starts a new completion
+      // event, preserving per-step "thinking" alongside tool use.
+      mojom::ConversationEntryEvent* existing_completion = nullptr;
+      for (auto& existing : base::Reversed(*entry->events)) {
+        if (existing->is_inline_search_event()) {
```
DJAndries: just mirroring Pete's concerns: should this be `is_tool_use_event` instead?

fallaciousreasoning: Nah, this one should definitely be `is_inline_search_event()` - ToolUseEvents should still cause a split, but I think what Pete is saying is that ToolUseEvents shouldn't be split by inline search events either.

I'm going to update this so that we get a list of events that don't cause a split.

DJAndries (approved):

components/ai_chat/core/browser/conversation_handler.cc
```diff
@@ -72,6 +73,69 @@ using ai_chat::mojom::ConversationTurn;
...
+      return false;
+  }
+}
+
+// If the previous non-splitting event is the same type as the current one, we
+// should join them instead of creating a new one. Otherwise, completions and
+// tool uses could be split (i.e. by an inline search event).
+mojom::ConversationEntryEvent* GetExistingEvent(
+    mojom::ConversationEntryEvent* current,
+    const std::vector<mojom::ConversationEntryEventPtr>& events) {
```
petemill: How about `MaybeGetEventToAppend`?

components/ai_chat/core/browser/conversation_handler.cc
```diff
@@ -72,6 +73,69 @@ using ai_chat::mojom::ConversationTurn;
 
 constexpr size_t kDefaultSuggestionsCount = 4;
 
+// Determines whether recieving an event of this type should trigger a new
+// streamed event. For example, an inline-search event shouldn't trigger a
+// new completion event (the completion before the inline search should be
+// modified).
```
petemill: nit - "trigger" might not be the best choice here since the event in question isn't the trigger, the "streamable event" is. Maybe something like:

```suggestion
// Determines whether receiving an event of type |event_tag| after a streaming event (e.g. completion), should potentially 
// cause a new streamable event to be created if the same streamable event is received afterwards. For example, an inline-search event shouldn't cause a
// new completion event (the completion before the inline search should be
// appended to).
```

Alternatively:

```suggestion
// Determines whether a streamable event (e.g. completion) should be interrupted by an event of type |event_tag|
// after which, if a new streamable chunk is received, a new streamable event is created, or whether the previous streamable event should be appended to.
```

Or some combination!

petemill: And maybe add to the comment something like
> Rough guide is that anything that's not generated in the single thread of LLM thought should be non-splitting. Anything the server would generate in a separate task should be non-splitting. Anything that's generated by the main LLM thread should be appended, if it's a streamable type.

petemill (approved):
Just have some naming nits. Nice work!

kjozwiak:

```
Brave | 1.93.67 Chromium: 149.0.7827.115 (Official Build) nightly (64-bit)
-- | --
Revision | 78ed9aadfce163ccde56bd17004ef904d7b5ef65
OS | Windows 11 Version 25H2 (Build 26200.8655)
```

See https://bravesoftware.slack.com/archives/CHGKGMHDJ/p1781559384898679?thread_ts=1781220911.075429&cid=CHGKGMHDJ for more context.

Using the STR/Cases outlined via https://github.com/brave/brave-browser/issues/56169#issue-4609799219, ensured that the entire answer is being displayed when `Task complete` as been collapsed as per the following:

**`Expanded`** | **`Collapsed`**
-----------------|-------------------
Screenshot 2026-06-15 173620 | Screenshot 2026-06-15 173342

PR #37070, opened 2026-06-08
Set disabled icon for scroll buttons

### LINKED ISSUE #56013: Consider a different look (grayed out) for the scrollable horizontal tab strip buttons when they aren't clickable
### Description

Would be nice if the scrollable horizontal tab strip buttons had a visual change for when there aren't any tabs to the left or the right to navigate to. 

Similar to the browser navigation buttons, how the "forward" nav button has a slightly different display to indicate that clicking on it won't go anywhere if there isn't at least one page to navigate to.

Note, not the same as https://github.com/brave/brave-browser/issues/55322 as this issue appears to removes the hover state.

### Steps to reproduce

1. Under brave://settings/appearance have "Use scrollable tab strip" enabled
2. Select a minimum tab width (I'm using "large")
3. Enable "Show scroll buttons on horizontal tab strip"
4. Open a bunch of tabs so you get the buttons
5. Navigate all the way to the left or all the way to the right of your tab strip
6. Notice that the tab strip button is still lit up (not grayed out) and clickable even though there's no where to navigate to

Note for QA: please check both light and dark themes

### Actual result

Example | Example
--- | ---
Image | Image

### Expected result

something similar to the forward browser navigation button in the above screenshot

### Reproduces how often

Easily reproduced

```
Brave | 1.91.165 Chromium: 149.0.7827.23 (Official Build) (arm64)
-- | --
Revision | fbc9ab393bb06368f8da826728d82a0724fcb118
OS | macOS Version 26.5 (Build 25F71)
```

### Channel information

- [ ] release (stable)
- [ ] beta
- [ ] nightly

### Reproducibility

### Miscellaneous information

cc @rebron @sangwoo108

sangwoo108:
TabStripControlButton::UpdateIcon() is missing disabled icon. But the scroll buttons are disabled when the tab strip is not scrollable , so the icon should be updated to reflect that.

Resolves https://github.com/brave/brave-browser/issues/56013

### DIFF
--- a/browser/ui/views/frame/brave_tab_strip_region_view.cc
+++ b/browser/ui/views/frame/brave_tab_strip_region_view.cc
@@ -29,6 +29,7 @@
 #include "ui/base/l10n/l10n_util.h"
 #include "ui/base/metadata/metadata_header_macros.h"
 #include "ui/base/metadata/metadata_impl_macros.h"
+#include "ui/base/models/image_model.h"
 #include "ui/events/event.h"
 #include "ui/views/animation/ink_drop.h"
 #include "ui/views/controls/button/button.h"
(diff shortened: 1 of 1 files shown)


simonhong (approved):
++

PR #37074, opened 2026-06-08
[iOS] Gate Brave News widget on Brave News admin policy

### LINKED ISSUE #56182: Disable Brave News Widget when Brave News is unavailable via policy
This widget doesn't technically use the users Brave News preference however it is associated with Brave News, so we should disable the widget from loading News data when the Brave News preference is disabled via admin policy

kylehickinson:
This disables the widget from downloading news clustering data when Brave News is disabled by policy

Resolves https://github.com/brave/brave-browser/issues/56182

### DIFF
--- a/ios/brave-ios/App/BraveWidgets/NewsTopics/NewsTopicsModel.swift
+++ b/ios/brave-ios/App/BraveWidgets/NewsTopics/NewsTopicsModel.swift
@@ -3,6 +3,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+import BraveWidgetsModels
 import CodableHelpers
 import Foundation
 import OrderedCollections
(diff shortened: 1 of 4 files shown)


StephenHeaps (approved):

Uni-verse:
Verified on `iPhone 12` running `iOS 26.3.1` using version `1.93 (71)`

- ensured that Brave News widgets are gated/disabled in the widget menu and also display as `Brave News is Disabled` on the home screen after admin policy is disabled.
- ensured that tapping on the disabled widgets on the home screen does not load any news articles.
- ensured that re-enabling Brave News in Origin settings will show news articles on NTP as well as enable the widgets to show articles (once they update). 
- ensured that Brave News has not regressed. 

Update: Additional verification: https://github.com/brave/brave-core/pull/37301#issuecomment-4735438984

example | example | example
--|--|--
IMG_0458 | IMG_0463 | IMG_0462

PR #37078, opened 2026-06-08
Add best practice: choose parameter types by ownership, use T&& to force a move

bridiverbot:
## Summary

Adds memory/lifetime best practice (CSM-038): choose a function parameter type
from what the function does with the argument, and never take a plain by-value
copyable owning type like `std::string` — by-value hides intent and permits a
silent copy.

For a `std::string` argument the choice is always one of:

| Intent                            | Parameter type        |
| --------------------------------- | --------------------- |
| Read-only, no ownership           | `std::string_view`    |
| Borrow the object (no copy/store) | `const std::string&`  |
| Take ownership (transfer)         | `std::string&&`       |

When the function takes ownership, an rvalue-reference parameter (`std::string&&`)
**forces the caller to `std::move`** (or pass a temporary), so the transfer is
explicit at the call site and a silent copy is impossible. The same
`view` / `const&` / `&&` split applies to other copyable movable owning types
(`std::vector<T>`, `base::Value`, …).

Documents the move-only exception: pass `std::unique_ptr<T>`, `mojo::Remote<T>`,
`base::OnceCallback`, etc. **by value** — the type already prevents a silent
copy, so by-value is the idiomatic transfer signature and should not be flagged
as needing `&&`. Cross-references CSM-031 and CSM-037.

## Notes

- Docs-only change under `docs/best-practices/`.

### DIFF
--- a/docs/best-practices/coding-standards-memory.md
+++ b/docs/best-practices/coding-standards-memory.md
@@ -836,3 +836,79 @@ class Decoder {
 constexpr auto kMagic = std::to_array<uint8_t>({0x7f, 'E', 'L', 'F'});
 base::BindOnce(&Validate, base::span(kMagic));
 ```
+
+---
+
+<a id="CSM-038"></a>
+
+## ✅ Choose Parameter Types by Ownership; Use `T&&` to Force a Move
+
+**Pick a function parameter type from what the function does with the argument,
+and never take a plain by-value copyable owning type like `std::string` —
+by-value hides intent and permits a silent copy.** For a `std::string`
+argument, the choice is one of three:
+
+| Intent                                          | Parameter type       |
+| ----------------------------------------------- | -------------------- |
+| Read-only (default)                             | `std::string_view`   |
+| Read-only, needs a null-terminated `const char*`| `base::cstring_view` |
+| Read-only, but downstream needs a `std::string` | `const std::string&` |
+| Take ownership (transfer)                       | `std::string&&`      |
+
+**`std::string_view` is the default for read-only parameters.** When a
+`string_view` won't do, two narrower non-owning options exist before you fall
+back to borrowing an owning `std::string`:
+
+- **Passing to a C API that wants a null-terminated `const char*`.** A plain
+  `std::string_view` is not guaranteed null-terminated, so its `data()` cannot be
+  handed to a C API. Use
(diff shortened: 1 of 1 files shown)


docs/best-practices/coding-standards-memory.md
```diff
@@ -836,3 +836,58 @@ class Decoder {
...
+
+**Pick a function parameter type from what the function does with the argument,
+and never take a plain by-value copyable owning type like `std::string` —
+by-value hides intent and permits a silent copy.** For a `std::string`
+argument, the choice is always one of three:
+
+| Intent                            | Parameter type        |
+| --------------------------------- | --------------------- |
+| Read-only, no ownership           | `std::string_view`    |
+| Borrow the object (no copy/store) | `const std::string&`  |
```
bridiver: This is wrong, const std::string& really only makes sense when using string_view would force a copy at some point. For instance dealing with mojo generated APIs or existing code that already uses std::string converting from string_view would force a copy

bridiver (changes requested):

docs/best-practices/coding-standards-memory.md
bridiverbot: Good call — fixed. std::string_view is now the documented default for read-only params, and const std::string& is narrowed to the case where taking a string_view would force a std::string to be materialized (and copied) downstream — e.g. mojo-generated APIs or existing code that already takes const std::string& / std::string. Updated the table, the explanation, and the example comment (StoreViaMojo).

bridiverbot: Also added the second const std::string& case you noted: a C API needing a null-terminated const char*. Since string_view isn't guaranteed null-terminated, you can't pass its data() to a C API; const std::string& lets you call .c_str() without materializing a temporary just to null-terminate. Both cases (downstream copy + C API c_str()) are now listed.

bridiverbot: Switched the C-API case to base::cstring_view (base/strings/cstring_view.h) — a non-owning, bounds-safe view of a NUL-terminated string that exposes .c_str() and is the intended replacement for const char*. Preferred over const std::string& there since it needs no ownership. const std::string& now only covers the case where a downstream mojo/legacy API actually needs a std::string (would copy from a view).

PR #37087, opened 2026-06-08
Fixes early deletion of old NTT component data after country change

### LINKED ISSUE #56095: [ads] NTT component is not downloaded after changing variation seed country
### Description

Follow up to: https://github.com/brave/brave-browser/issues/47909

The old component is removed, but we still try to serve NTTs.

IMPORTANT: It fixes itself after browser restart

### Steps to reproduce

1. Clean profile
2. Run Brave with `--disable-variations-seed-fetch-throttling`
3. Check `brave://components` for `NTP Sponsored Images`
4. Enable VPN from another country
5. Wait till `NTP Sponsored Images` is updated to new country

### Actual result

Old `NTP Sponsored Images` component is removed
New `NTP Sponsored Images` component is not downloaded

We still try to serve the NTT despite no component:
Image

### Expected result

Old `NTP Sponsored Images` component is removed
New `NTP Sponsored Images` component is downloaded
NTTs are served from new component

### Reproduces how often

Easily reproduced

### Brave version

Brave v1.91.169 on Lenovo TB-8506FS (Android 11.0)

cc @aseren @tmancey @iambrianfung

aseren:
Previously the old NTT component was unregistered synchronously before the new component's data was available. This could caused an issue when Brave Ads served NTT which is not presented on disk showing error page on NTP. The PR fixes the issue by deferring unregistration to after the new component's JSON is loaded so the old component remains active until the replacement is ready.

The issue which was fixed in the PR caused dump without crashes at: https://github.com/brave/brave-core/blob/master/components/ntp_background_images/browser/ntp_background_images_service.cc#L70.
Backtrace shows that the issue started happening from January 22nd: [link](https://brave.sp.backtrace.io/p/brave/triage?time=quarter&filters=JTVCJTVCJTIyY2FsbHN0YWNrJTIyJTJDJTIyY29udGFpbnMlMjIlMkMlMjJIYW5kbGVDb21wb25lbnREYXRhJTIyJTVEJTVE&aggregations=((version%2Cdistribution)%2C(guid%2Cunique)%2C(classifiers%2Chead))&similarity=false).

Resolves https://github.com/brave/brave-browser/issues/56095


components/ntp_background_images/browser/ntp_background_images_service.cc
```diff
@@ -388,6 +387,13 @@ void NTPBackgroundImagesService::OnGetSponsoredComponentJsonData(
   for (auto& observer : observers_) {
     observer.OnSponsoredImagesDataDidUpdate(sponsored_images_data_.get());
   }
+
+  if (pending_unregister_old_component_id_ && component_update_service_) {
+    // Unregister previous component.
```
tmancey, line 392: nit: comment repeats next line.

components/ntp_background_images/browser/ntp_background_images_service.cc
```diff
@@ -238,12 +238,11 @@ void NTPBackgroundImagesService::RegisterSponsoredImagesComponent() {
     return;
   }
 
-  if (sponsored_images_component_id_) {
-    // Unregister previous component.
-    component_update_service_->UnregisterComponent(
-        *sponsored_images_component_id_);
-  }
+  // Unregistration of the old component is pending until the new one is ready.
+  pending_unregister_old_component_id_ = sponsored_images_component_id_;
```
tmancey, line 242: Because this is async, `RegisterSponsoredImagesComponent` can be called again before the new component loads. If the country changes twice before the first new component loads, the pending old component ID gets overwritten and is never unregistered. Similarly, if the new components JSON is invalid or another failure the pending ID is never cleared, so if the country changes again before a successful load the same overwrite happens and the original component is orphaned.

components/ntp_background_images/browser/ntp_background_images_service.cc
aseren: The comment was copied from the old code.
But yeah, agreed that it is not needed

components/ntp_background_images/browser/ntp_background_images_service.cc
aseren: Great catch! I thought about them before and found them as edge cases because they require variation service to fire twice in a short period of time. Also this PR fixes invalid JSON dumps and invalid JSON shouldn't be a case again.

tmancey: not tested but something like
```
const std::optional<std::string> last_id = sponsored_images_component_id_;
sponsored_images_component_id_ = sponsored_images_component->id;
sponsored_images_installed_dir_ = base::FilePath();

RegisterNTPSponsoredImagesComponent(
    ...,
    base::BindRepeating(
        &NTPBackgroundImagesService::OnSponsoredComponentReady,
        weak_factory_.GetWeakPtr(), last_id));
```
then `OnSponsoredComponentReady` carries the last ID through to `OnGetSponsoredComponentJsonData` and unregisters it there. This avoids the new member entirely and fixes some of these edge cases.

tmancey: I believe `sponsored_images_component_id_` was likely the cause of the issue because it is set regardless of whether the component was successful. It, and possibly other variables, should only be set after the component has completed successfully.

tmancey: This should fix all race conditions above.

components/ntp_background_images/browser/ntp_background_images_service_unittest.cc
```diff
@@ -727,4 +759,77 @@ TEST_F(NTPBackgroundImagesServiceTest,
...
+  EXPECT_TRUE(observer_.on_sponsored_images_updated);
+  observer_.on_sponsored_images_updated = false;
+
+  EXPECT_CALL(component_update_service(),
+              UnregisterComponent(std::string(us_component->id)))
+      .Times(0);
+
+  // Simulate GB sponsored images component installation failure when it is
+  // registered but its JSON never delivered.
+  service_->set_country_code_for_testing("GB");
```
netzenbot, line 820: The `EXPECT_THAT(service_->pending_unregister_old_component_id_, ...)` check inspects a private implementation field rather than observable behavior. The `EXPECT_CALL(...UnregisterComponent...).Times(0)` mock expectation already verifies what matters externally — that unregistration does not occur. Consider removing the internal-state check. See [TI-002](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-002).

components/ntp_background_images/browser/ntp_background_images_service.cc
```diff
@@ -388,6 +387,13 @@ void NTPBackgroundImagesService::OnGetSponsoredComponentJsonData(
   for (auto& observer : observers_) {
     observer.OnSponsoredImagesDataDidUpdate(sponsored_images_data_.get());
   }
+
+  if (pending_unregister_old_component_id_ && component_update_service_) {
+    // Unregister previous component.
+    component_update_service_->UnregisterComponent(
```
netzenbot, line 393: nit: `component_update_service_` is only null in tests (the existing `RegisterSponsoredImagesComponent` path dereferences it directly without a null check). Per [ARCH-024](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-024), null checks that only apply in test configurations should be preceded by `CHECK_IS_TEST()` to document the intent and prevent confusion about whether null is a valid production state.

components/ntp_background_images/browser/ntp_background_images_service.h
```diff
@@ -106,6 +106,12 @@ class NTPBackgroundImagesService {
       DoNotGetSponsoredRichMediaContentIfWallpaperRelativeUrlReferencesParent);
   FRIEND_TEST_ALL_PREFIXES(NTPBackgroundImagesServiceTest,
                            SponsoredImageWithMissingImageUrlTest);
+  FRIEND_TEST_ALL_PREFIXES(
```
netzenbot, line 109: nit: `friend class NTPBackgroundImagesServiceTest` (line 90) already grants full access to all private members, so these `FRIEND_TEST_ALL_PREFIXES` macros are redundant. See [TI-011](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-011).

aseren:
I was able to reproduce the issue by adding `--variations-seed-fetch-interval=1` which updates variation seed every one minute and switched VPN to different countries back and forth. It is caused by a bug in chromium updater code which showed up only when we started updating sponsored component on a fly when country changes, instead of updating it on browser startup. I've patched chromium code and found that the issue didn't appear in my testing.

The current PR does a plaster fix of a race condition/bug in current NTP sponsored images code, but it does not fix the issue I reproduced. So I'm closing the PR as discussed with Terry it would be better to properly re-write NTP sponsored images code instead of making tons of plasters.

tmancey:
@aseren this needs fixing as it fixes a different race than the Chromium bug you found.

PR #37097, opened 2026-06-08
[ai-chat] `ConversationHandler::Observer` stubs brough back

cdesouza-chromium:
This is a follow up to a previous PR which had erroneously replaced the
stubs with pure virtuals, however that's not the practice with observes.

See https://github.com/brave/brave-core/pull/37062


petemill (approved):
thank you!

netzenbot (approved):

PR #37113, opened 2026-06-09
[toolchain] Build Rust/WASM with `LZMA_API_STATIC`

cdesouza-chromium:
We are running into issues in the intel nodes, but this can also happen
in arm systems, when the local homebrew install has deployed a different
version from the one expected during link.

We are setting this change to get this dependency built and statically
linked. It is of note that upstream already does this on Linux, but they
most likely don't have homebrew installs on their bots, and that's why
they are not hitting this on Mac.


AlexeyBarabash (approved):
LGTM!

PR #37116, opened 2026-06-09
Bump orchard crate to the 0.14 version

cypt4:
Dependency review https://github.com/brave/reviews/issues/2272
Follow up code changes : https://github.com/brave/brave-core/pull/37122
Resolves https://github.com/brave/internal/issues/1758

Testplan: Verify that basic shielded functionality is working and user is able to spend old shielded balance.

PR #37122, opened 2026-06-09
Update zcash code to work with orchard 0.14

cypt4:
This PR updates brave rust code to work with latest orchard crate and also changes librustzcash to the latest trimmed compatible version.
Resolves : https://github.com/brave/internal/issues/1758
Dependency review : https://github.com/brave/reviews/issues/2272
Review for target branch : https://github.com/brave/brave-core/pull/37116


kdenhartog (approved):
no concerns with doing this. It doesn't have an affect on the new change to orchard 0.14 and the change to the proof system

components/brave_wallet/browser/zcash/zcash_wallet_service_unittest.cc
```diff
@@ -1619,7 +1619,7 @@ TEST_F(ZCashWalletServiceUnitTest, ShieldFunds_FailsOnNetworkError) {
 #define MAYBE_UnshieldFunds UnshieldFunds
 #endif
 
-// https://3xpl.com/zcash/transaction/5e07d5b298f2862f2332effc833f5fe9157eb9ca00e03f394663e81b397515a9
+// https://3xpl.com/zcash/transaction/821edadc1bc51e7dc7a57c01eb766292a88d3836ad7f83f4e3e505100cef2300
```
cmazakas, line 1622: Maybe a dumb question but does upgrading the crate truly necessitate updating these tests? The updates themselves look okay but I'm just wondering, are we adding incompatible backwards-breaking changes here?

cmazakas (approved):
The changes looks fine on paper, even though I will confess I'm not a huge fan of how much the tests had to change and we introduced a couple of new Rust functions as well. If we mutate one of the new Rust functions, a unittest should fail.

components/brave_wallet/browser/zcash/zcash_wallet_service_unittest.cc
cypt4: Yes, since tx body changes in the new version due crypto math update.

components/brave_wallet/browser/zcash/rust/librustzcash/BUILD.gn
```diff
@@ -51,24 +59,30 @@ rust_static_library("zcash_primitives") {
...
 
-  # TODO(cypt4): Fix unused imports and remove this flag.
-
-  rustflags = [ "-Aunused-imports" ]
+  # Brave trims this crate down to the Orchard serialization surface, which
+  # leaves some upstream helpers unused. Allow unused imports / dead code rather
+  # than further diverging from upstream source.
+  rustflags = [
+    "-Aunused-imports",
+    "-Adead_code",
```
bridiver, line 75: how much dead code are we talking about? What's the binary size before/after the bump?

netzenbot (approved):

components/brave_wallet/browser/zcash/rust/librustzcash/BUILD.gn
cypt4: a quick comparison between this PR mac nightly artifact and artifact from another PR (37002) says sizes are almost the same. This PR is even 20kb smaller.

cypt4: dead_code affects ~40 lines in zcash_primitives/src/encoding.rs

bridiver (approved):
Code owner ok

components/brave_wallet/browser/zcash/zcash_wallet_service_unittest.cc
kdenhartog: Yeah, for context the reason for this update was that the network was updated to prevent a cryptography bug which allowed for the possibility of minting additional ZEC in the orchard shielded pool. When that network update went online, it produced a breaking change on how proofs are generated hence the need for the additional updates.

PR #37126, opened 2026-06-09
[Wallet] update hex parsing for uint256_t to use internal function template

cmazakas:
Directly invoke the primary function template responsible for parsing hex strings into integral types in wallet.

`IsValidHexString` ensures the input string is prefixed by `0x`, which makes the `.substr(2)` call sound when obtaining the portion of the string we will directly parse.

CodeHealth follow-up for https://github.com/brave/brave-core/pull/37071

### DIFF
--- a/components/brave_wallet/common/hex_utils.cc
+++ b/components/brave_wallet/common/hex_utils.cc
@@ -16,6 +16,7 @@
 #include "base/numerics/safe_conversions.h"
 #include "base/strings/strcat.h"
 #include "base/strings/string_number_conversions.h"
+#include "base/strings/string_number_conversions_internal.h"
 #include "base/strings/string_util.h"
 #include "base/strings/string_view_util.h"
 
@@ -126,19 +127,18 @@ bool HexValueToUint256(std::string_view hex_input, uint256_t* out) {
   if (!out) {
     return false;
   }
+
   if (!IsValidHexString(hex_input)) {
     return false;
   }
-  auto hex_string = hex_input.substr(2);
-  if (hex_string.length() > 64) {
-    return false;
+
+  hex_input = hex_input.substr(2);
+  if (hex_input.empty()) {
+    *out = 0;
+    return true;
   }
-  *out = 0;
-  for (char c : hex_string) {
-    (*out) <<= 4;
-    (*out) += static_cast<uint256_t>(base::HexDigitToInt(c));
-  }
-  return true;
+
+  return base::internal::HexStringToIntImpl(hex_input, *out);
 }
 
 std::optional<uint256_t> HexValueToUint256(std::string_view hex_input) {


netzenbot (approved):

components/brave_wallet/common/hex_utils.cc
```diff
@@ -126,19 +127,18 @@ bool HexValueToUint256(std::string_view hex_input, uint256_t* out) {
...
-    (*out) += static_cast<uint256_t>(base::HexDigitToInt(c));
+
+  hex_input = hex_input.substr(2);
+  if (hex_input.empty()) {
+    *out = 0;
+    return true;
   }
-  return true;
+
+  return base::internal::HexStringToIntImpl(hex_input, *out);
```
supermassive, line 141: Afaik `base::internal` is not allowed outside of base and I have once had a PR reverted with such reasoning.

cdesouza-chromium: I think this is okay because it is code they have and we don't. Having said that I think in a follow up PR we can do a bit better here.

We have these calls in Wallet to `base::internal` code that we need

```cxx
std::optional<uint256_t> Base10ValueToUint256(std::string_view input) {
  uint256_t output;
  return base::internal::StringToIntImpl(input, output) ? std::optional(output)
                                                        : std::nullopt;
}

std::optional<int256_t> Base10ValueToInt256(std::string_view input) {
  int256_t output;
  return base::internal::StringToIntImpl(input, output) ? std::optional(output)
                                                        : std::nullopt;
}

std::string Uint256ValueToBase10(uint256_t value) {
  return base::internal::IntToStringT<std::string>(value);
}
```

And now we are adding this:

```
return base::internal::HexStringToIntImpl(hex_input, *out);
```

So, it doesn't need to be done on this PR, it can be a follow up, but it seems we keep calling these internals, when actually we should just add/expose the variants we need `base/strings/string_number_conversions.cc` and `base/strings/string_number_conversions.h`


Something like this:
```cxx
bool StringToUint256(std::u16string_view input, uint256_t* output) {
  return internal::StringToIntImpl(input, *output);
}
```

I think all these cases fit extension of `//base` code rather than we directly using the internals, and also extend `chromium_src/base/strings/string_number_conversions_unittest.cc`. Does that make sense?

cdesouza-chromium (approved):
LGTM % left a suggestion for a follow up to reduce the direct use of `base::internal::` outside `//base`.

components/brave_wallet/common/hex_utils.cc
supermassive: I'm OK with how it is now actually, just wanted confirmations from someone from core owners.
And approach to extend base is also looks natural way of fixing this for us.

supermassive (approved):

components/brave_wallet/common/hex_utils.cc
cmazakas: Added a corresponding CodeHealth issue here https://github.com/brave/brave-browser/issues/56289

PR #37137, opened 2026-06-09
[Android] Fix home-screen PWA/shortcut icons disappearing after reboot

### LINKED ISSUE #52423: "Add to homescreen" shortcuts lost after reboot
### Description

"Add to homescreen" shortcuts lost after reboot

### Steps to reproduce

1. go to web page in latest Brave on Android
2. "add to home screen"
3. launcher icon is added
4. reboot
5. 

### Actual result

* after adding launcher icon
* reboot
* shortcut is gone

### Expected result

* after adding launcher icon
* reboot
* shortcut is present

### Reproduces how often

Easily reproduced

### Brave version

Brave 1.86.142 Chromium 144.0.7559.97

### Device

- Brand/model: Pixel 10 Pro XL
- Android version: Android 16/Build/BP4A.260105.004.E1

### Channel information

- [x] release (stable)
- [ ] beta
- [ ] nightly

### Reproducibility

### Miscellaneous information

Chrome for android doesn't actually even add it in the first place. 

Yesterday though, Brave didn't add it either, and I had to go through a bunch of different options to get it to add at all. 

This behavior was present on the brand new pixel from the moment I transferred from old device to new box after unboxing the new phone. I got it working, and got all my home screen shortcuts added, then rebooted and they are all gone. Multiple tests afterwards show Brave adding the shortcut but it is gone after reboot.

Victor-root:
Resolves https://github.com/brave/brave-browser/issues/52423

## 🐛 Problem
On Android, icons added via **"Add to Home screen"** — both installed **PWAs** and plain **shortcuts** — **vanish after a reboot**. They appear fine, then disappear on restart. Reproduced across several brands/launchers (Oppo, Xiaomi, Pixel…), on stock **and** third-party launchers (Nova). **Chrome on the same device is unaffected.**

## 🔍 Root cause
Brave doesn't mint WebAPKs, so "Add to Home screen" always pins a shortcut via `ShortcutManager`. Android persists the shortcut's launch `Intent` extras as **Android Binary XML**, where each serialized string is capped at **65535 bytes** (modified UTF-8). The icon is stored as a base64 string in `WebappConstants.EXTRA_ICON`; a high-res icon overflows that cap, so `ShortcutService` fails to write the package file and **drops every Brave shortcut** on the next reboot:

```
ShortcutService: Failed to write to file .../com.brave.browser.xml
java.io.IOException: Modified UTF-8 length too large: -83039
  PersistableBundle.saveToXml → ShortcutService.writeTagExtra → ShortcutPackage.saveShortcut
```

The icon shows briefly because `requestPinShortcut()` adds it to the launcher's in-memory list first; the on-disk write fails, so it's gone after reboot. As the file is **per-package**, one oversized PWA icon also takes down co-existing plain shortcuts (→ both disappear). Chrome avoids this by installing real WebAPKs, which never go through `ShortcutService`.

## 🛠️ Fix
Route the pinned-shortcut icon through a new `BravePwaShortcutIconHelper` that **calls the upstream `BitmapHelper.encodeBitmapAsString()`** and **downscales the bitmap only when its base64 would otherwise exceed the limit** (floored at 48px). Only the in-Intent copy shrinks — the home-screen glyph is a separate full-res `Bitmap` and is unaffected, and the MAC stays consistent (computed over the same encoded string). Applied via a direct patch to `ShortcutHelper.java`, with unit tests covering the helper.

## ✅ Testing
Built locally (`arm64`, Debug) and **tested on a physical Oppo Reno 13 Pro (Android 16)**:
- Added a **PWA** ("Install") **and** a **plain shortcut** ("Add to Home screen") → **rebooted** → **both survive** ✅
- Before this change, both **vanished** after reboot.
- Same result on the **Oppo stock launcher** and **Nova Launcher**.

(@SergeyZhukovsky also reproduced the failure on a Pixel with a large icon 🙏)

## 📝 Notes
- Minimal: no behavior change unless the encoded icon would exceed the limit.
- A floor on the downscale keeps the splash/stored icon reasonable.
- Happy to adjust the implementation as needed before merge.

### DIFF
--- a/android/brave_java_sources.gni
+++ b/android/brave_java_sources.gni
@@ -36,6 +36,7 @@ brave_java_sources = [
   "../../brave/android/java/org/chromium/chrome/browser/BraveLandscapeHelper.java",
   "../../brave/android/java/org/chromium/chrome/browser/BraveLaunchIntentDispatcher.java",
   "../../brave/android/java/org/chromium/chrome/browser/BraveLocalState.java",
+  "../../brave/android/java/org/chromium/chrome/browser/BravePwaShortcutIconHelper.java",
   "../../brave/android/java/org/chromium/chrome/browser/BraveRelaunchUtils.java",
   "../../brave/android/java/org/chromium/chrome/browser/BraveRewardsBalance.java",
   "../../brave/android/java/org/chromium/chrome/browser/BraveRewardsExternalWallet.java",
--- a/android/java/org/chromium/chrome/browser/BravePwaShortcutIconHelper.java
+++ b/android/java/org/chromium/chrome/browser/BravePwaShortcutIconHelper.java
@@ -1,0 +1,74 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+package org.chromium.chrome.browser;
+
+import android.graphics.Bitmap;
+
+import androidx.annotation.VisibleForTesting;
+
+import org.chromium.build.annotations.NullMarked;
+import org.chromium.build.annotations.Nullable;
+import org.chromium.chrome.browser.browserservices.intents.BitmapHelper;
+
+/**
+ * Helper that bounds the size of a PWA icon before it is base64-encoded into a pinned-shortcut
+ * Intent.
+ *
+ * <p>On Android 12+ the system ShortcutService persists pinned shortcuts as Binary XML, where every
+ * string is capped at 65535 bytes. A high-resolution icon's base64 string overflows that cap, the
+ * whole package XML write fails atomically, and ALL of Brave's pinned shortcuts get dropped on the
+ * next reboot. Chrome avoids this by minting WebAPKs (icons live in APK drawables), a Google-only
+ * backend Brave cannot use, so Brave falls back to the legacy webapp-shortcut path that carries the
+ * icon in the Intent.
+ *
+ * <p>Downscaling here keeps the encoded icon under the limit. Only the in-Intent copy shrinks; the
+ * home-screen glyph is a separate full-resolution Bitmap and is unaffected. The MAC stays
+ * consistent because it is computed over this same encoded string.
+ */
+@NullMarked
+public class BravePwaShortcutIconHelper {
+    // Keep the encoded icon comfortably below the 65535-byte Binary XML string
+    // limit to leave headroom for the rest of the shortcut's persisted strings.
+    @VisibleForTesting static final int MAX_ENCODED_ICON_CHARS = 60000;
+
+    // Floor so the in-app icon stays recognizable even for pathological inputs.
+    @VisibleForTesting static final int MIN_ICON_SIZE_PX = 48;
+
+    // Bound on shrink iterations; convergence is fast but guard against edge
+    // cases (e.g. incompressible noise) where a single pass is not enough.
+    private static final int MAX_ATTEMPTS = 10;
+
+    /**
+     * Encodes {@code bitmap} as a base64 PNG string like {@link
+     * BitmapHelper#encodeBitmapAsString(Bitmap)}, downscaling first if needed so the result fits
+     * within the ShortcutService Binary XML string limit.
+     *
+     * @param bitmap The icon to encode, may be null.
+     * @return the base64-encoded (and possibly downscaled) icon string.
+     */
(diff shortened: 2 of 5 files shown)


SergeyZhukovsky:
@Victor-root thank you so much for your effort and I appreciate that you helped to solve that problem. We can't merge your current PR as is because there are things that we do different in Brave. I'm going to leave some comments on why we do it different and also I created a branch with the changes that acceptable https://github.com/brave/brave-core/commit/27a2f4f6aadfe64d4d675989e4f51c24758ffb37. I can create a PR from my branch but I would prefer you do it based on my commit so it's merged under your name. Go ahead, check it and create please and we are going to approve and merge the fix.

android/java/org/chromium/chrome/browser/webapps/BravePwaShortcutIconHelper.java
```diff
@@ -0,0 +1,105 @@
...
+            }
+
+            current = Bitmap.createScaledBitmap(current, newWidth, newHeight, /* filter= */ true);
+            encoded = encode(current);
+        }
+
+        return encoded;
+    }
+
+    // Mirrors ShortcutHelper.encodeBitmapAsString so the result round-trips through
```
SergeyZhukovsky: We don't usually do it as upstream Chromium could potentially add some valuable checks or functionality inside the legit BitmapHelper.encodeBitmapAsString and we are going to miss it. The best way is to call the upstream function and add something that we want to do right before or after it's call.

android/java/org/chromium/chrome/browser/webapps/BravePwaShortcutIconHelper.java
```diff
@@ -0,0 +1,105 @@
...
+
+    private BravePwaShortcutIconHelper() {}
+
+    /**
+     * Returns a base64 PNG encoding of {@code bitmap} in the same format read back by
+     * {@code ShortcutHelper.decodeBitmapFromString}, guaranteed to fit Android's pinned-shortcut
+     * persistence limit. The bitmap is downscaled only when its full-size encoding would otherwise
+     * exceed the limit.
+     */
+    public static String encodeForPinnedShortcut(@Nullable Bitmap bitmap) {
```
SergeyZhukovsky: we want to add some tests to that functionality

rewrite/chrome/android/java/src/org/chromium/chrome/browser/ShortcutHelper.java.toml
```diff
@@ -0,0 +1,17 @@
...
+Route the webapp home-screen icon through BravePwaShortcutIconHelper so the
+base64 icon stored in the pinned shortcut's Intent (WebappConstants.EXTRA_ICON)
+stays under Android's ShortcutManager persistence limit (65535 bytes in modified
+UTF-8). A high-resolution icon otherwise overflows it, Android's ShortcutService
+fails to persist com.brave.browser.xml, and every Brave shortcut in that file is
+dropped on the next reboot. See
+https://github.com/brave/brave-browser/issues/52423.
+"""
+pattern = 'String encodedIcon = BitmapHelper.encodeBitmapAsString(icon);'
+replace = 'String encodedIcon = org.chromium.chrome.browser.webapps.BravePwaShortcutIconHelper.encodeForPinnedShortcut(icon);'
```
SergeyZhukovsky: we better do a direct patch in that particular case inside ShortcutHelper.java. We still don't use plaster approach widely, it's experimental. We try to use bytecode replacing technique on Android, but in that particular case the call that has to be replaced right in the middle of the function, bytecode works great when the whole function has to be replaced or extended.

Victor-root:
Thank you so much, @SergeyZhukovsky 🙏 I added you as co-author since the final, clean implementation is yours, it felt only fair.

This genuinely means a lot to me: I've been dealing with this bug for about a year, and as a heavy PWA user it was honestly close to pushing me away from Brave which I really didn't want, because I love the browser and wanted to stay on it. So I'm really grateful you took the time to dig into it and get it fixed properly. Thanks again for the help, and for letting it land under my name! 🙌

SergeyZhukovsky (approved):
++

samartnik (approved):
java patching ++

SergeyZhukovsky:
@stoletheminerals could you please take a look, does the change look fine in terms of security?

stoletheminerals (approved):

SergeyZhukovsky:
CI is green https://github.com/brave/brave-core/pull/37167, we have all approves, merging it. Thank you very much again @Victor-root! We will have it bake on nightly and beta channels first to see if there any issues with it. It should appear on stable channel on July 28th https://github.com/brave/brave-browser/wiki/Brave-Release-Schedule

PR #37139, opened 2026-06-09
[VPN 2.0] Fix incorrect bundling of VPN agent

### LINKED ISSUE #56246: Fix incorrect bundling of VPN agent
Looks like `framework_bundle_data` is used in both bundle deps and transitive public deps.
Which results in the VPN agent bundle being added to all of the framework's bundles.
Need to fix this, so it's only added to the primary framework bundle.

alekspop26:
The `framework_bundle_data` group is used in both bundle deps and transitive public deps. That results in the VPN agent bundle being added to all of the framework's bundles. To fix this, a new group, `vpn_apps_bundle_data`, is added that is appended only to browser's framework bundle deps (`brave_chrome_framework_bundle_deps`), but not public deps (`brave_chrome_framework_public_deps`) applied transitively. This ensures that the VPN agent's bundle is added only to the top-level framework's bundle.

Resolves https://github.com/brave/brave-browser/issues/56246

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -326,9 +326,11 @@ if (is_android) {
 if (is_mac) {
   group("framework_bundle_data") {
     deps = [ "components/brave_extension/extension/brave_extension:brave_extension_framework_bundle_data" ]
+  }
 
-    if (enable_brave_vpn_v2_bundling) {
-      deps += [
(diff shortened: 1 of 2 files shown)


netzenbot (approved):

cdesouza-chromium (approved):
LGTM

PR #37141, opened 2026-06-10
[Sidebar V2] Consider panel visibility also for contents's bottom radius

simonhong:

Don't use window-corner radius to web contents when panel or sidebar UI(control view) are visible.

TEST=RoundedCornersWithSidePanelVisibleButSidebarControlViewHiddenTest

Use non window-corner radius: fixed by this PR
image

Use window-corner radius: existing behavior
image

Use non window-corner radius: existing behavior
image

Use non window-corner radius: existing behavior
image


sangwoo108 (approved):
LGTM

PR #37143, opened 2026-06-10
Fix rsync timeout in updates with Omaha 4 on macOS

### LINKED ISSUE #56252: Updates with Omaha 4 sometimes brick the browser
Both @LaurenWags and @Brave-Matt reported cases where Brave (or Origin) were unlaunchable on macOS with the error message "The application Brave [Browser/Origin] Nightly can't be opened".

Image

Investigating the associated update logs showed that the update process ran into a timeout and got interrupted mid-way.

This can only happen on Nightly and Beta for now, because Omaha 4 is not enabled on Release.

mherrmann:
This is a rewrite of #36914 that hopefully resolves https://github.com/brave/brave-browser/issues/56252.

# Test plan


cdesouza-chromium (approved):
LGTM

PR #37013, opened 2026-06-05
[Brave News]: Enable sync

fallaciousreasoning:
Resolves https://github.com/brave/brave-browser/issues/23054

PR #37144, opened 2026-06-10
[Brave News]: Add sync support

### LINKED ISSUE #23054: Include custom Brave News sources and/or setup in Sync
## Description 

Many users spend a lot of time configuring Brave News to their liking, especially now with custom RSS feeds. Users should be able to Sync this data across devices just like History, Bookmarks or other browser data. 

cc @petemill 

## Miscellaneous Information:

Found this users thread as I was writing this issue:
https://community.brave.com/t/add-brave-news-to-synchronization/397708

fallaciousreasoning:
Resolves https://github.com/brave/brave-browser/issues/23054
Closes https://github.com/brave/brave-core/pull/37013
Closes https://github.com/brave/go-sync/pull/429

I had a play around with more complicated sync strategies here but merging gets a lot more complicated unless we add timestamps to individual entires. I think for the most part this should be a pretty massive improvement.

One case where this won't work properly is if you make edits on two different devices, the first edit will win.

### DIFF
--- a/chromium_src/chrome/browser/sync/prefs/chrome_syncable_prefs_database.cc
+++ b/chromium_src/chrome/browser/sync/prefs/chrome_syncable_prefs_database.cc
@@ -47,6 +47,16 @@ enum {
   kProfileContentSettingsPartitionedExceptionsFingerprintingV2 = 300027,
   kProfileContentSettingsPartitionedExceptionsBraveShields = 300028,
   kProfileContentSettingsPartitionedExceptionsBraveSpeedreader = 300029,
+  // Brave News prefs (mirrors
+  // brave/components/brave_news/common/pref_names.h). Synced as part of the
+  // "Settings" (PREFERENCES) sync type.
+  kBraveNewsShouldShowToolbarButton = 300030,
+  kBraveNewsNewTabPageShowToday = 300031,
+  kBraveNewsOptedIn = 300032,
+  kBraveNewsSources = 300033,
+  kBraveNewsChannels = 300034,
+  kBraveNewsDirectFeeds = 300035,
+  kBraveNewsOpenArticlesInNewTab = 300036,
 };
 }  // namespace brave_syncable_prefs_ids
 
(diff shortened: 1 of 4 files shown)


fallaciousreasoning:
**Note for reviewers:** This is my first syncable pref, so if something looks wrong it probably is.

components/brave_news/common/DEPS
```diff
@@ -1,5 +1,6 @@
 include_rules = [
   "+components/prefs",
+  "+components/pref_registry",
```
netzenbot, line 3: nit: `pref_registry_syncable.h` is only included in `pref_names.cc`, so this allowlist entry belongs in `specific_include_rules` keyed to `pref_names\.cc` rather than the broad `include_rules`. See [BS-030](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-030).

aguscruiz:
HUGEEEEEE

chromium_src/chrome/browser/sync/prefs/chrome_syncable_prefs_database.cc
```diff
@@ -197,6 +207,42 @@ const auto& BraveSyncablePreferences() {
...
+        sync_preferences::PrefSensitivity::kNone,
+        sync_preferences::MergeBehavior::kNone}},
+      {"brave.news.channels",
+       {brave_syncable_prefs_ids::kBraveNewsChannels, syncer::PREFERENCES,
+        sync_preferences::PrefSensitivity::kNone,
+        sync_preferences::MergeBehavior::kNone}},
+      {"brave.today.userfeeds",
+       {brave_syncable_prefs_ids::kBraveNewsDirectFeeds, syncer::PREFERENCES,
+        sync_preferences::PrefSensitivity::kNone,
+        sync_preferences::MergeBehavior::kNone}},
```
fallaciousreasoning, line 241: I'm a bit worried about this on initial sync

PR #37149, opened 2026-06-10
Fix upstream windows build flake by ignoring .rsp in reorder-imports.

goodov:
Resolves
```
21:31:46    File "D:\ws\x64-nightly\src\brave\vendor\depot_tools\bootstrap-2@3_11_8_chromium_35_bin\python3\bin\Lib\shutil.py", line 256, in copyfile
21:31:46      with open(src, 'rb') as fsrc:
21:31:46           ^^^^^^^^^^^^^^^
21:31:46  PermissionError: [Errno 13] Permission denied: 'initialexe\\chrome.exe.rsp'
```

The correct GN dep is there, so it's not a dependency issue, but a file removal issue.

### DIFF
--- a/chromium_src/build/win/reorder-imports.py
+++ b/chromium_src/build/win/reorder-imports.py
@@ -8,7 +8,7 @@ import glob
 import override_utils
 
 
-@override_utils.override_function(glob, name='iglob')
+@override_utils.override_function(glob)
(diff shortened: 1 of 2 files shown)


cdesouza-chromium (approved):
LGTM

PR #25345, opened 2024-08-27
Get scriptlets working in about:blank frames

antonok-edm:
Resolves https://github.com/brave/brave-browser/issues/40703

## Submitter Checklist:

- [x] I confirm that no [security/privacy review is needed](https://github.com/brave/brave-browser/wiki/Security-reviews) and no other type of reviews are needed, or that I have [requested](https://github.com/brave/reviews/issues/new/choose) them
- [x] There is a [ticket](https://github.com/brave/brave-browser/issues) for my issue
- [x] Used Github [auto-closing keywords](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue) in the PR description above
- [x] Wrote a good [PR/commit description](https://google.github.io/eng-practices/review/developer/cl-descriptions.html)
- [x] Squashed any review feedback or "fixup" commits before merge, so that history is a record of what happened in the repo, not your PR
- [x] Added appropriate labels (`QA/Yes` or `QA/No`; `release-notes/include` or `release-notes/exclude`; `OS/...`) to the associated issue
- [x] Checked the PR locally:
  * `npm run test -- brave_browser_tests`, `npm run test -- brave_unit_tests` [wiki](https://github.com/brave/brave-browser/wiki/Tests)
  * `npm run presubmit` [wiki](https://github.com/brave/brave-browser/wiki/Presubmit-checks), `npm run gn_check`, `npm run tslint`
- [ ] Ran `git rebase master` (if needed)

## Reviewer Checklist:

- [ ] A security review [is not needed](https://github.com/brave/brave-browser/wiki/Security-reviews), or a link to one is included in the PR description
- [ ] New files have MPL-2.0 license header
- [ ] Adequate test coverage exists to prevent regressions
- [ ] Major classes, functions and non-trivial code blocks are well-commented
- [ ] Changes in component dependencies are properly reflected in `gn`
- [ ] Code follows the [style guide](https://chromium.googlesource.com/chromium/src/+/HEAD/styleguide/c++/c++.md)
- [ ] Test plan is specified in PR before merging

## After-merge Checklist:

- [ ] The associated issue milestone is set to the smallest version that the
  changes has landed on
- [ ] All relevant documentation has been updated, for instance:
  - [ ] https://github.com/brave/brave-browser/wiki/Deviations-from-Chromium-(features-we-disable-or-remove)
  - [ ] https://github.com/brave/brave-browser/wiki/Proxy-redirected-URLs
[...]

PR #37153, opened 2026-06-10
[farbling] Support schemes for farbling with inherited http/https origin

### LINKED ISSUE #56048: Shields is not applied on non HTTP / HTTPS sites
We have some valid use-cases where shield is [not applied](https://github.com/brave/brave-core/blob/master/components/brave_shields/core/browser/brave_shields_utils.cc#L863-L865) on non-HTTP/HTTPS URL schemes like: `brave://`, `file://` etc as they don't have an associated content settings to them. For HTTP/HTTPS sites, we have the various [BRAVE_WEBCOMPAT_*](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/chromium_src/components/content_settings/core/common/content_settings_types.mojom?L32-50) content settings types that allows us to turn "off" / "on" shields for each type independently. 

As a consequence of not applying shields, an http/https site can embed or open a blob:// or data:// which can read the true values of various farbled surfaces (for example: [screen dimensions](https://hackerone.com/reports/3777660) and [exfiltrating un-farbled values via local storage](https://test-website-a.pages.dev/blob-farbling-bypass/))

We should investigate the various URL schemes that maybe meaningful to be included under shields, and a way to integrate them properly. 

Past references: 
- https://github.com/brave/brave-browser/issues/36972 
- https://github.com/brave/brave-browser/issues/24984
- https://github.com/brave/brave-browser/issues/23170#issue-1254541710

r0hit0303:
This change:

a) Replace the use of `GetLastCommittedURL` with `GetLastCommittedOrigin` or `url::Origin(GetLastCommittedURL)` to correctly fetch the right Content Settings for the current URL. Relying on `GetLastCommittedURL` can create a new Content settings in the context of inherited http/https schemes like `blob://` instead of reusing the existing one from the inherited origin.  Please refer to the draft [document](https://docs.google.com/document/d/1X_hRvZhaK2gTfE0sw-0iVxKhKl4UIeTEfg4ctQxbQns/edit?usp=sharing) for more details.

b) Removes the TODO in `brave_blob_screen_farbling_browsertest` to match the expected behaviour when farbling is enabled. 

c) Adds new unit tests specific to getting farbling token when considering the various URLs etc. 

Resolves: https://github.com/brave/brave-browser/issues/56048

### Manual testing (results): 
1) Go to https://test-website-a.pages.dev/blob-farbling-bypass/
2) Run the tests. 

#### Brave (after this change)

image

#### Brave (before this change) 

image

#### Chrome - No protections
image

### DIFF
--- a/browser/brave_shields/brave_shields_tab_helper.cc
+++ b/browser/brave_shields/brave_shields_tab_helper.cc
@@ -33,6 +33,7 @@
 #include "content/public/browser/navigation_controller.h"
 #include "content/public/browser/navigation_handle.h"
 #include "content/public/browser/reload_type.h"
+#include "content/public/browser/render_frame_host.h"
 #include "content/public/browser/security_principal.h"
 #include "content/public/browser/web_contents.h"
 #include "net/base/features.h"
@@ -323,7 +324,10 @@ void BraveShieldsTabHelper::SetBraveShieldsAdBlockOnlyModePromptDismissed() {
 }
 
 GURL BraveShieldsTabHelper::GetCurrentSiteURL() const {
-  return web_contents()->GetLastCommittedURL();
+  return web_contents()
+      ->GetPrimaryMainFrame()
+      ->GetLastCommittedOrigin()
+      .GetURL();
 }
 
 GURL BraveShieldsTabHelper::GetFaviconURL(bool refresh) {
--- a/browser/brave_shields/brave_shields_web_contents_observer.cc
+++ b/browser/brave_shields/brave_shields_web_contents_observer.cc
@@ -38,6 +38,7 @@
 #include "extensions/buildflags/buildflags.h"
 #include "mojo/public/cpp/bindings/associated_remote.h"
 #include "third_party/blink/public/common/associated_interfaces/associated_interface_provider.h"
+#include "url/origin.h"
 
 #if BUILDFLAG(ENABLE_CONTAINERS)
 #include "brave/components/containers/content/browser/storage_partition_utils.h"
@@ -318,12 +319,17 @@ void BraveShieldsWebContentsObserver::SendShieldsSettings(
   DCHECK(navigation_handle);
   RenderFrameHost* rfh = navigation_handle->GetRenderFrameHost();
 
-  const GURL& primary_url =
+  // Relying on url::Origin::Create correctly extracts the embedded origin for
+  // blob: URLs (e.g. blob:https://example.com/uuid to https://example.com) and
+  // works for all other schemes as well. Origin is more secure than working
+  // with last committed URLs.
+  const GURL primary_url =
       navigation_handle->GetParentFrameOrOuterDocument()
           ? navigation_handle->GetParentFrameOrOuterDocument()
                 ->GetOutermostMainFrame()
-                ->GetLastCommittedURL()
-          : navigation_handle->GetURL();
+                ->GetLastCommittedOrigin()
+                .GetURL()
+          : url::Origin::Create(navigation_handle->GetURL()).GetURL();
 
   HostContentSettingsMap* host_content_settings_map =
       HostContentSettingsMapFactory::GetForProfile(rfh->GetBrowserContext());
--- a/browser/farbling/brave_blob_screen_farbling_browsertest.cc
+++ b/browser/farbling/brave_blob_screen_farbling_browsertest.cc
@@ -146,22 +146,9 @@ class BraveBlobScreenFarblingBrowserTest
     }));
 
     // Validate.
-    bool should_match = true;
-    // When the screen fingerprinting feature is enabled AND fingerprinting is
-    // blocked (DEFAULT) AND the blob is in a popup, the popup bypasses farbling
-    // and sees real values while the parent sees farbled values — so they won't
-    // match.
-    // TODO(https://github.com/brave/brave-browser/issues/56048): Remove this
-    // override once shields are supported on blob:// URLs.
-    if (GetParam() && !fingerprinting_allowed_ &&
-        blob_container_type_ == BlobContainerType::kPopUpWindow) {
-      should_match = false;
-    }
-
-    EXPECT_EQ(should_match,
-              content::EvalJs(MainFrame(),
-                              "storedScreenValuesMatch('parent', 'blob')")
-                  .ExtractBool())
+    EXPECT_EQ(true, content::EvalJs(MainFrame(),
+                                    "storedScreenValuesMatch('parent', 'blob')")
+                        .ExtractBool())
         << "param=" << GetParam()
         << ", container=" << static_cast<std::size_t>(blob_container_type_)
         << ", allowed=" << fingerprinting_allowed_;
@@ -170,18 +157,6 @@ class BraveBlobScreenFarblingBrowserTest
     if (blob_container_type_ == BlobContainerType::kPopUpWindow) {
       ClosePopup();
     }
-  }
-
-  void RunTests() {
-    AllowFingerprinting(/*allow=*/false);
-    NavigateToBlob();
-
-    AllowFingerprinting(/*allow=*/true);
-    NavigateToBlob();
-  }
-
-  void set_blob_container_type(const BlobContainerType blob_container_type) {
-    blob_container_type_ = blob_container_type;
   }
 
   void RunTests() {
--- a/browser/farbling/brave_blob_screen_farbling_browsertest.cc
+++ b/browser/farbling/brave_blob_screen_farbling_browsertest.cc
@@ -198,7 +198,6 @@ class BraveBlobScreenFarblingBrowserTest
 
  private:
   base::test::ScopedFeatureList feature_list_;
-  raw_ptr<Browser> pop_up_browser_;
   GURL blob_test_url_;
   BlobContainerType blob_container_type_ = BlobContainerType::kUnset;
   bool fingerprinting_allowed_ = false;
--- a/browser/farbling/brave_blob_screen_farbling_browsertest.cc
+++ b/browser/farbling/brave_blob_screen_farbling_browsertest.cc
@@ -185,8 +185,6 @@ class BraveBlobScreenFarblingBrowserTest
   }
 
   void RunTests() {
-    ui_test_utils::SetAndWaitForBounds(*browser(), kTestWindowBounds);
-
     AllowFingerprinting(/*allow=*/false);
     NavigateToBlob();
 
--- a/browser/farbling/brave_blob_screen_farbling_browsertest.cc
+++ b/browser/farbling/brave_blob_screen_farbling_browsertest.cc
@@ -184,8 +184,23 @@ class BraveBlobScreenFarblingBrowserTest
     blob_container_type_ = blob_container_type;
   }
 
+  void RunTests() {
+    ui_test_utils::SetAndWaitForBounds(*browser(), kTestWindowBounds);
+
+    AllowFingerprinting(/*allow=*/false);
+    NavigateToBlob();
+
+    AllowFingerprinting(/*allow=*/true);
+    NavigateToBlob();
+  }
+
+  void set_blob_container_type(const BlobContainerType blob_container_type) {
+    blob_container_type_ = blob_container_type;
+  }
+
  private:
   base::test::ScopedFeatureList feature_list_;
+  raw_ptr<Browser> pop_up_browser_;
   GURL blob_test_url_;
   BlobContainerType blob_container_type_ = BlobContainerType::kUnset;
   bool fingerprinting_allowed_ = false;
--- a/browser/farbling/brave_blob_screen_farbling_browsertest.cc
+++ b/browser/farbling/brave_blob_screen_farbling_browsertest.cc
@@ -84,11 +84,12 @@ class BraveBlobScreenFarblingBrowserTest
     }
   }
 
-  void AllowFingerprinting(bool allow) {
+  void AllowFingerprinting(bool allow, std::optional<GURL> url = std::nullopt) {
     fingerprinting_allowed_ = allow;
     brave_shields::SetFingerprintingControlType(
         HostContentSettingsMapFactory::GetForProfile(browser()->profile()),
-        allow ? ControlType::ALLOW : ControlType::DEFAULT, blob_test_url_);
+        allow ? ControlType::ALLOW : ControlType::DEFAULT,
+        url.has_value() ? url.value() : blob_test_url_);
   }
 
   content::RenderFrameHost* MainFrame() const {
@@ -171,9 +172,11 @@ class BraveBlobScreenFarblingBrowserTest
     blob_container_type_ = blob_container_type;
   }
 
+ protected:
+  GURL blob_test_url_;
+
  private:
   base::test::ScopedFeatureList feature_list_;
-  GURL blob_test_url_;
   BlobContainerType blob_container_type_ = BlobContainerType::kUnset;
   bool fingerprinting_allowed_ = false;
   raw_ptr<Browser> pop_up_browser_ = nullptr;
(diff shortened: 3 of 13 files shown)


components/brave_shields/core/browser/brave_shields_utils.cc
r0hit0303, line 286: We seem to be doing this even for invalid URLs (i.e `url.is_valid()` returns false). I wonder in this case we shouldn't just simply return `true` i.e ensuring we continue to do farbling. In my previous patchset, I was returning false here which broke couple of tests. [1](https://ci.brave.com/job/brave-core-build-pr-origin-macos-arm64/job/PR-37153/1/). For example [here](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/components/brave_shields/core/test/brave_shields_utils_unittest.cc?L282) we expected `true` i.e apply shields on an empty GURL. I'm not sure what was the historical reason behind this.

ShivanKaul:
Can we also check against https://github.com/brave/brave-core/pull/25345?

components/content_settings/renderer/brave_content_settings_agent_impl.cc
```diff
@@ -359,7 +359,7 @@ BraveContentSettingsAgentImpl::GetBraveShieldsSettings(
     ContentSettingsType webcompat_settings_type) {
   blink::WebLocalFrame* frame = render_frame()->GetWebFrame();
   const GURL primary_url(GetTopFrameOriginAsURL(frame));
-  if (!primary_url.SchemeIsHTTPOrHTTPS()) {
+  if (!primary_url.SchemeIsHTTPOrHTTPS() && !primary_url.SchemeIsBlob()) {
```
ShivanKaul: do we even need this check?

r0hit0303: You are right. We probably don't need it. My rationale to add that was to consider cases where the top frame origin was actually a blob:// scheme and if we wanted to save any web-compat settings for that. However, given that blob URLs are uniquely generated for each session by the browser storage, it probably doesn't make sense indeed to persist any web-compat settings for them like you said, and the defaults are good here.

r0hit0303:
Thanks Shivan for your comments. The change is still in draft I wanted to look into few things before. 

> Can we also check against https://github.com/brave/brave-core/pull/25345?

I'm unable to open [scriptlet.localframes.xyz](https://scriptlet.localframes.xyz/) which was required for manual testing. Anton's change did also include [browser tests](https://github.com/brave/brave-core/pull/25345/changes#diff-a22832051483ad7ab6e9d1689654bbef9abf591a7b4a5c423812f098626b5d35) ([code](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/browser/brave_shields/ad_block_service_browsertest.cc?L2625)) which is passing for this change. If I can get the source-code for `scriptlet.localframes.xyz` I'm happy to host it locally and re-verify the manual testing process.

components/brave_shields/core/browser/brave_shields_utils.cc
r0hit0303: I've made the intent more explicit in the latest change that we want the brave shields to be enabled when `url.is_valid()` returns false. Otherwise, it went through the whole ```ContentSetting setting =
      map->GetContentSetting(url, GURL(), ContentSettingsType::BRAVE_SHIELDS);``` flow which didn't make much sense to me for invalid URLs. Please let me know if I missed something.

r0hit0303:
@antonok-edm Could you ptal at this change first with the added context of (https://github.com/brave/brave-core/pull/37153#issuecomment-4673155810)? Thanks

r0hit0303:
I'm moving the refactoring changes related to tests as a separate PR to help with this review.  Changing the status to draft in the meantime.

r0hit0303:
@goodov Could you ptal? Thanks

goodov (approved):
lgtm

chromium_src/third_party/blink/renderer/core/frame/local_dom_window.cc
```diff
@@ -44,6 +45,26 @@ using brave::BlockScreenFingerprinting;
...
+// underlying brave session cache is the same, which helps to keep the same
+// farbling seed for both the opener and the openee ensuring they see the same
+// farbled values.
+ExecutionContext* GetContextFromOpenerIfPossible(
+    const DOMWindow* opener,
+    ExecutionContext* current_context) {
+  if (auto* local_opener = DynamicTo<LocalDOMWindow>(opener)) {
+    if (ExecutionContext* opener_context =
+            local_opener->GetExecutionContext()) {
+      return opener_context;
```
goodov: I think the idea is right, but I wonder if the `opener()` should be used somehow automatically as part of `BraveSessionCache::From`.

`LocalDOMWindow` implements `ExecutionContext`, so technically you can upcast it via `DynamicTo` and access `opener()` inside.

r0hit0303: Thanks for the suggestion. This also helps to keep the logic centralized across all the different fingerprinting surfaces. 

I had one small unrelated doubt on the change. I was thinking about cross origin case where `a.com` opens `b.com` as a popup. In this case should `b.com` and `a.com` see the same farbled values or different? I think with change they would get the same value (please see the browser test in the latest ps). I'm leaning here to say same is fine because `a.com` and `b.com` could probably communicate over `postMessages` their respective values anyways? although, I'm not sure if `Cross-Origin-Opener-Policy` would affect this communication.

browser/farbling/brave_blob_screen_farbling_browsertest.cc
```diff
@@ -212,3 +202,72 @@ IN_PROC_BROWSER_TEST_P(BraveBlobScreenFarblingBrowserTest,
...
+                                     .ExtractString();
+
+    // 4. Different origins get different per-origin farble seeds only when the
+    // feature is enabled AND fingerprinting is not allowed. In every other
+    // case (feature disabled, or fingerprinting explicitly allowed) both
+    // windows see the same real screen values.
+    const bool farbling_active = GetParam() && !allow;
+    if (farbling_active) {
+      EXPECT_NE(a_screen, b_screen)
+          << "a.com and b.com should see different farbled screen values";
```
r0hit0303: @goodov This fails locally due to this change I think

r0hit0303: I just checked this test against a clean branch and it's failing there as well. So, a.com and b.com (opened by a.com) are already getting the same value. I'm not sure why though. 

**Update:** 
So, both `a.com` and `b.com` are getting farbled but they end up getting the same farbled values because they both had the window width setup to 400 x 400 in the test, and [ChromeClientImpl::BraveGetScreenInfos ](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/chromium_src/third_party/blink/renderer/core/page/chrome_client_impl.cc?L35-62) seems to rely only on the window attributes to deterministically choose the width that's just greater than 400 amongst the hardcoded accepted screen values which I missed. 

The attributes which would be affected are the screenX/screenY in the `local_dom_window.cc` via [FarbleInteger](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/chromium_src/third_party/blink/renderer/core/frame/local_dom_window.cc?L77). I'm rewriting the test now to check the behaviour.

chromium_src/third_party/blink/renderer/core/frame/local_dom_window.cc
r0hit0303: Also adding @pes10k to the discussion here, as we were discussing offline about sharing the opener's context for more insights on this.

pes10k: My suggestion was to not rely on walking frame trees or explicitly using the opener to try and walk to the parent. Thats fraught and easy to do wrong (you need to account for nested instances, the opener might not be the top level document, etc. (this is all the same with the local frames paper we wrote, that showed that most tools do this wrong for about:blank frames)

My strong suggestion is to avoid all that, and to *always* use two things to determine the right farble seed:
the immediate frame's `window.origin`, and the top level document's `window.origin`. Those properties should be easy to get to and not require and walking and climbing around the frame tree.

r0hit0303: > My strong suggestion is to avoid all that, and to always use two things to determine the right farble seed:
the immediate frame's window.origin, and the top level document's window.origin

I'm not sure if we can do that without changing the core design of the BraveSessionCache. Maybe @goodov can chime in here if I'm off but
[BraveSessionCache](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/third_party/blink/renderer/core/farbling/brave_session_cache.cc?L260) is strongly tied to `ExecutionContext` i.e each new context gives a new `BraveSessionCache`. This change is simply trying to feed in the opener's `ExecutionContext` to get the associated `BraveSessionCache` of the opener so as to get the same farbled seed. 

@ShivanKaul Is it a strong requirement to have the pop-up window share the same execution context as the opener given that both are getting farbled anyways?   If not, then we don't need to make extra considerations around trying to match the farbled values.

ShivanKaul: Hmm, I can't think of a reason for `blob` URLs to be different than HTTP/HTTPS URLs when it comes to popup windows. The privacy problem was that the `blob` page being opened doesn't get Shields protections _and_ is the same origin so gets same storage etc.

r0hit0303: That's my understanding as well that for blob URLs opened in a popup window should get the same farbling seed out-of-the-box. However, that's not happening in practice (I removed the changes made to brave_session_cache.cc and ran the manual test again. Please see the result below.) until we relied on the opener's execution context. 

I think the main question here I need to answer is why the execution context is coming different for: blob popup window and the main window opener as they both have the same eTLD+1(!). I'll investigate this further but really helpful to get all these different points on my initial concern. 

**W/O changes to brave_session_cache**
Result of https://test-website-a.pages.dev/blob-farbling-bypass/

image

pes10k: Like i mentioned, i dont know the particulars of the implementation in chromium, @goodov or @bridiver or others are way better to ask about that than me.

The two points im trying to make are: 

1. its very easy to make mistakes if you're determining the security origin (and so, storage area, farble seed, etc) based on a frame's URL, or by walking the frame tree (`window.opener()`, etc). The right security origin (and so, the correct storage area etc) will always be `window.origin` or [`GetSecurityOrigin`](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/execution_context/execution_context.cc;drc=362dbaf3c1048a93929fff003e784d537f3ebfbc;l=391). And so, if you're asking about openers, then I think it might be digging in the wrong direction.

2.  None of this should require or involve needing to change the design of `BraveSessionCache`.
Conceptually at least, storage areas should be stored in something like a doubly nested hashtable, with the correct storage area and related values (inc the farbling seed) for a third party frame found by first indexing into the session cache with the eTLD+1 of the security origin of the top level document, and then the eTLD+1 of the security origin of the immediate frame.

r0hit0303: Appreciate all the help here. Farbling is still quite new to me and I was not sure why the underlying token wasn't the same for blob:// which inherits the same origin as you said. Going via the opener was a bad idea and I should have investigated a bit more on why the farbling token didn't match out of the box. I've found the issue  and have updated the change and also added new tests for checking the farbling token behaviour. PTAL. Resolving this thread.

netzenbot (approved):

pes10k:
@r0hit0303 

with respect to the [localframes test](https://github.com/brave/brave-core/pull/37153#issuecomment-4679252714), if you didn't find the code, just let me know what was being tested and I'm happy to recreate the test case(s)

components/brave_shields/core/browser/brave_shields_utils.cc
```diff
@@ -224,16 +225,21 @@ void SetBraveShieldsEnabled(HostContentSettingsMap* map,
                             bool enable,
                             const GURL& url,
                             PrefService* local_state) {
-  if (url.is_valid() && !url.SchemeIsHTTPOrHTTPS()) {
+  // Resolve inherited origins (e.g. blob:https://host/id to https://host/) so
+  // that scheme guards and pattern generation always use the true origin.
+  const url::Origin origin = url::Origin::Create(url);
+  const GURL origin_url = origin.GetURL();
```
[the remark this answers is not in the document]
r0hit0303: I'm actually adding considerations on origin which were missing before. We were working on plain URLs to make decisions about brave shields settings. This change transforms those URLs to Origin to URL derived from the origin, to keep the scope of change minimal. Ideally, these methods should have their signature changed to Origin. We could slowly deprecate the public methods which relies on GURL and add a new corresponding method that takes in Origin. I'll follow up on that after discussing with the various stakeholders after this change.

r0hit0303:
If it's ok with everyone I think we could first merge this fix, and then I can continue with refactoring the brave_shields_utils.cc (https://github.com/brave/brave-browser/issues/56805). 

**Manual testing**

https://github.com/user-attachments/assets/e55bd99f-80a2-4c0d-8293-496ec5b3267f

antonok-edm (approved):
lgtm, thanks!

browser/brave_shields/brave_shields_tab_helper.cc
```diff
@@ -323,7 +324,10 @@ void BraveShieldsTabHelper::SetBraveShieldsAdBlockOnlyModePromptDismissed() {
 }
 
 GURL BraveShieldsTabHelper::GetCurrentSiteURL() const {
-  return web_contents()->GetLastCommittedURL();
+  return web_contents()
+      ->GetPrimaryMainFrame()
+      ->GetLastCommittedOrigin()
+      .GetURL();
```
pes10k, line 330: This seems odd to me.  The URL will the the same as the origin here, and Chromium discourages folks from passing around and making S+P decisions based on GURL (as a broader way of encoruaging folks to stick to the security origin).  Do you really want / need the GURL here?

browser/brave_shields/brave_shields_web_contents_observer.cc
```diff
@@ -318,12 +319,17 @@ void BraveShieldsWebContentsObserver::SendShieldsSettings(
...
+  // with last committed URLs.
+  const GURL primary_url =
       navigation_handle->GetParentFrameOrOuterDocument()
           ? navigation_handle->GetParentFrameOrOuterDocument()
                 ->GetOutermostMainFrame()
-                ->GetLastCommittedURL()
-          : navigation_handle->GetURL();
+                ->GetLastCommittedOrigin()
+                .GetURL()
+          : url::Origin::Create(navigation_handle->GetURL()).GetURL();
```
pes10k, line 332: Same here.  Can you just omit the GURL here, since the GURL is kinda a footgun (a unknowing developer might thing you actually want the frame URL), and the chromium docs urge not doing this (using GURL to store the origin) for exactly this foot-gun reason

https://chromium.googlesource.com/chromium/src/+/HEAD/docs/security/origin-vs-url.md#don_t-use-the-type-to-store-origins

pes10k:
making my comments as non-blocking, since from slack @r0hit0303 says that getting rid of GURL here (and replacing with Origin) is planned in future changes, and beyond the scope here

pes10k (approved):

browser/brave_shields/brave_shields_tab_helper.cc
r0hit0303: We are following on the big refactor which will pass the url::Origin at https://github.com/brave/brave-browser/issues/56805. For now, I want to keep this fix lean to resolve the privacy issue at hand. Thanks

r0hit0303:
Thanks @antonok-edm and @pes10k for the approval. 

@fmarier and @thypon for visibility on this change from security side in-case there are any concerns.

+@goodov / @ShivanKaul for optional re-review. 

+all: The refactoring work is out-of-scope for this fix and is tracked separately at https://github.com/brave/brave-browser/issues/56805.

components/brave_shields/core/test/brave_shields_utils_unittest.cc
```diff
@@ -1076,3 +1084,183 @@ TEST_F(BraveShieldsUtilTest, GetDomainBlockingType) {
...
+      map(), GURL("https://example.com/page1"), {});
+  const auto t2 = brave_shields::GetFarblingToken(
+      map(), GURL("https://example.com/page2?q=1#anchor"), {});
+  EXPECT_EQ(t1, t2);
+}
+
+// A blob URL whose inner origin is https://example.com must yield the same
+// token as a plain https://example.com URL, because both resolve to the same
+// effective origin (https://example.com/).
+TEST_F(BraveShieldsUtilTest, FarblingToken_BlobUrl_SameTokenAsOriginUrl) {
```
bridiver, line 1131: This isn't actually testing `GetFarblingToken`, it's testing `Origin::Create` so it isn't needed here

components/brave_shields/core/test/brave_shields_utils_unittest.cc
```diff
@@ -1076,3 +1084,183 @@ TEST_F(BraveShieldsUtilTest, GetDomainBlockingType) {
...
+      url::Origin::Create(
+          GURL("blob:https://example.com/550e8400-e29b-41d4-a716-446655440000"))
+          .GetURL(),
+      {});
+  const auto https_token = brave_shields::GetFarblingToken(
+      map(), GURL("https://example.com/some/path"), {});
+  EXPECT_EQ(blob_token, https_token);
+}
+
+// A blob URL whose inner origin is a subdomain must resolve to the same token
```
bridiver, line 1146: same as https://github.com/brave/brave-core/pull/37153/changes#r3531647334

components/brave_shields/core/test/brave_shields_utils_unittest.cc
```diff
@@ -1076,3 +1084,183 @@ TEST_F(BraveShieldsUtilTest, GetDomainBlockingType) {
...
+      map(), GURL("https://sub.example.com"), {});
+  const auto blob_sub_token = brave_shields::GetFarblingToken(
+      map(),
+      url::Origin::Create(GURL("blob:https://sub.example.com/some-uuid"))
+          .GetURL(),
+      {});
+  EXPECT_EQ(sub_token, blob_sub_token);
+}
+
+// A blob URL for a subdomain must share the token of the root domain.
```
bridiver, line 1163: same as https://github.com/brave/brave-core/pull/37153/changes#r3531647334

components/brave_shields/core/test/brave_shields_utils_unittest.cc
```diff
@@ -1076,3 +1084,183 @@ TEST_F(BraveShieldsUtilTest, GetDomainBlockingType) {
...
+  const auto blob_sub_token = brave_shields::GetFarblingToken(
+      map(),
+      url::Origin::Create(GURL("blob:https://sub.example.com/some-uuid"))
+          .GetURL(),
+      {});
+  EXPECT_EQ(root_token, blob_sub_token);
+}
+
+// Two completely different origins must get different tokens.
+TEST_F(BraveShieldsUtilTest, FarblingToken_DifferentOrigins_DifferentTokens) {
```
bridiver, line 1183: We didn't already have tests for this stuff? That's a bit concerning.

components/brave_shields/core/test/brave_shields_utils_unittest.cc
r0hit0303: The refactoring change is planned to be followed up separately from this privacy fix like mentioned [earlier](https://github.com/brave/brave-core/pull/37153#issuecomment-4896790217). When that will happen these tests will not compile, and will fail as they are passing GURL now. If we remove these tests now, it means the author would need to keep in mind to add them later. If you feel too strongly about your comment, then do re-open this comment and I'll file a separate issue to track that and remove these tests. Please note, I'd like to fix the underlying privacy issue asap so in-case there were other points that we could follow-up separately from this change, then that will help us to expedite this fix faster. Thanks.

components/brave_shields/core/test/brave_shields_utils_unittest.cc
r0hit0303: Ack. I found these browser tests [here](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/browser/farbling/brave_farbling_browsertest.cc?L159-163) but maybe I missed something. I think it's better to keep these **unit** tests as it's testing the core method. Please re-open this if needed.

goodov (approved):

r0hit0303:
Thanks @goodov for the review

PR #37155, opened 2026-06-10
[iOS] Gate usage of safe builtin prompt communication to web schemes

kylehickinson:
This change ensures that `sendWebKitMessageSynchronously` which uses `window.prompt` under the hood can only be used on web pages, since application URLs are expected to use mojo and `MojoFacade` forces this requirement. It also exposes a new `windowHasWebScheme` that can be used in JavaScript features to avoid running scripts on WebUI

### DIFF
--- a/ios/web/js_messaging/resources/safe_builtins.ts
+++ b/ios/web/js_messaging/resources/safe_builtins.ts
@@ -37,12 +37,24 @@ class SafeBuiltins {
     message: object | string,
   ) => any | null
 
+  // Whether or not the window.location has a web scheme (http/https/data)
+  readonly windowHasWebScheme: () => boolean
+
   constructor() {
     // Setup private refs to capture in safe builtin functions
     const webkitMessageHandlers = window.webkit.messageHandlers
     const windowPrompt = window.prompt.bind(window)
     const jsonStringify = JSON.stringify.bind(JSON)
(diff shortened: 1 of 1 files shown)


ios/web/js_messaging/resources/safe_builtins.ts
```diff
@@ -37,12 +37,24 @@ class SafeBuiltins {
...
     const windowPrompt = window.prompt.bind(window)
     const jsonStringify = JSON.stringify.bind(JSON)
     const jsonParse = JSON.parse.bind(JSON)
+    const windowProtocol = window.location.protocol
+
+    this.windowHasWebScheme = () => {
+      return (
+        windowProtocol === 'https:'
+        || windowProtocol === 'http:'
+        || windowProtocol === 'data:'
```
stoletheminerals, line 55: do we need it to run on `data:` too?

kylehickinson: Mostly have it to match the native function `web::UrlHasWebScheme` in Chromium: https://source.chromium.org/chromium/chromium/src/+/main:ios/web/common/url_scheme_util.mm;l=14

stoletheminerals (approved):

netzenbot (approved):

PR #37171, opened 2026-06-10
[VPN 2.0] Add LaunchDaemon plist for Brave VPN Helper on macOS

### LINKED ISSUE #54626: Implement stub Helper application
Implement stub Helper application:
-  conditional build (same as BraveVpnService v2)
-  setup cross-platform GN build system
-  setup source tree inside brave-core
-  stub main.cc
-  ~stub unit tests running both locally and in CI~ (no stub tests needed)
-  distribution and bundling

alekspop26:
Add launchd.plist template and wire it into the browser bundle at Contents/Library/LaunchDaemons/ so SMAppService.daemon(plistName:) can register the privileged helper. Use process_version with @var@ substitution to support variable replacement in plist keys, which is required for the MachServices endpoint name. The generated plist is placed into the browser's `brave_app` target deps rather than the framework, since SMAppService requires it in the top-level app bundle.

Resolves https://github.com/brave/brave-browser/issues/54626

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -726,7 +726,10 @@ if (is_mac) {
     deps = [ ":brave_app_plist" ]
 
     if (enable_brave_vpn_v2_bundling) {
-      deps += [ "//brave/components/brave_vpn/app/v2/agent:brave_vpn_agent_launchd_plist_data" ]
+      deps += [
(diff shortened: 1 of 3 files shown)


bsclifton (approved):
LGTM! compiled locally to make sure `.plist` file is populated correctly 👍

PR #37177, opened 2026-06-11
[Brave News] Give browser/brave_news its own source_set targets

bridiverbot:
## Summary

Follow-up to the review comment on https://github.com/brave/brave-core/pull/36245 (the Chromium 149→150 bump): the production `browser/brave_news` sources were being dumped into the aggregated `//chrome/browser` target via `browser/brave_news/sources.gni` instead of living in their own GN target.

This converts them to proper targets in `browser/brave_news/BUILD.gn`, matching the `browser/brave_adaptive_captcha` layout:

- `source_set("brave_news")` — public, holds the headers with their `public_deps`.
- `source_set("impl")` — `visibility = [ "//chrome/browser" ]`, holds the `.cc` files (incl. the Android factory under `is_android`).

`browser/sources.gni` now depends on `//brave/browser/brave_news` and `:impl` (guarded by `enable_brave_news`) instead of appending the slurped sources/deps, and `browser/brave_news/sources.gni` is deleted.

### Deps that the monolith used to provide transitively

`gn check` surfaced deps that were free while these sources were compiled inside `//chrome/browser`:

- `:impl` now declares the chrome service factories it uses (`content_settings`, `favicon`, `history`, `policy`, `profiles`), plus `//brave/browser:browser_process` and `//brave/components/brave_shields/core/browser`.
- `brave_news_tab_helper.cc` includes `chrome/browser/feed/rss_links_fetcher.h`, which on desktop lives in the `//chrome/browser` monolith. `:impl` can't depend on `//chrome/browser` (cycle), so it's added to `brave_chrome_browser_allow_circular_includes_from`, matching `serp_metrics:impl`.
- Consumers of the now-extracted headers — `//brave/browser/ui` and `//brave/browser/ui/webui/brave_new_tab_page_refresh` — now depend on `//brave/browser/brave_news`.

## Testing

- `gn check` (via `npm run gn_check`): **clean** — 0 errors across the tree.
- `gn gen`: valid.
- A targeted compile could not be completed in my environment due to a pre-existing missing asset (`node_modules/@brave/leo/icons/social-github-color.svg`) unrelated to this change — worth a normal CI/build pass.


fallaciousreasoning (approved):

cdesouza-chromium (approved):
LGTM

browser/sources.gni
```diff
@@ -741,7 +739,12 @@ brave_chrome_browser_sources += brave_browser_web_discovery_sources
 brave_chrome_browser_sources += brave_browser_webcompat_reporter_sources
 
 brave_chrome_browser_deps += brave_browser_autocomplete_deps
-brave_chrome_browser_deps += brave_browser_brave_news_deps
+if (enable_brave_news) {
```
bridiver: please move this below with the rest of the conditional deps

browser/brave_news/BUILD.gn
```diff
@@ -18,6 +84,7 @@ source_set("browser_tests") {
     ]
 
     deps = [
+      ":brave_news",
```
bridiver, line 87: doesn't this also need the impl target?

browser/sources.gni
bridiverbot: Done — moved the `enable_brave_news` block down to join the other conditional `if (enable_X) { brave_chrome_browser_deps += [...] }` blocks (just after the `enable_widevine` one) in 8b6f51a3ba2.

browser/brave_news/BUILD.gn
bridiverbot: It can't — `:impl` has `visibility = [ "//chrome/browser" ]`, so a direct dep here would fail `gn check`. It doesn't need to either: `browser_tests` already depends on `//chrome/browser`, which links `:impl` via `brave_chrome_browser_deps`, so the implementation is pulled in transitively at link time. The `:brave_news` dep is only here to provide the public headers the browsertests include.

bridiver: but does it need to depend on `//chrome/browser` for other reasons? We could change `impl` visibility if it allows us to remove `//chrome/browser` dep here

bridiverbot: Yeah, it needs `//chrome/browser` independently of `brave_news:impl`, so widening `:impl` visibility wouldn't let us drop it:

- `direct_feed_fetcher_browsertest.cc` calls `g_browser_process->system_network_context_manager()->GetSharedURLLoaderFactory()` — `SystemNetworkContextManager`'s impl is in the `//chrome/browser` monolith (`chrome/browser/net`). And that test doesn't touch `brave_news:impl` at all; it exercises `//brave/components/brave_news/browser:direct_feed_fetcher`.
- `brave_news_tab_helper_browsertest.cc` uses `Browser` / `BrowserView` / `browser()->tab_strip_model()` / `Profile`.

So I kept `//chrome/browser` + the `:brave_news` header dep and left `:impl` visibility as-is. Happy to split these out if you'd prefer, but it'd take more than just the visibility change.

boocmp (approved):

PR #37182, opened 2026-06-11
[screenshot] Add capture engine for visible area and full page

sangwoo108:
Add brave/browser/ui/screenshot/, a View-agnostic capture orchestrator that the toolbar UI (next commit) will drive. Two modes:

Visible area:
  RenderWidgetHostView::CopyFromSurface -> SkBitmap -> PNG (encoded on
  ThreadPool) -> Save As.

Full page: Uses screenshot::PrintPreviewExtractor

Save As default filename is screenshot-YYYY-MM-DD-HHMMSS.png

A `kBraveScreenshot` base::Feature gates the feature; the toolbar wiring comes in the next commit.

part of https://github.com/brave/brave-browser/issues/56140

### DIFF
--- a/browser/about_flags.cc
+++ b/browser/about_flags.cc
@@ -129,6 +129,7 @@
 #include "brave/browser/ui/darker_theme/features.h"
 #include "brave/browser/ui/focus_mode/focus_mode_features.h"
 #include "brave/browser/ui/page_info/features.h"
+#include "brave/browser/ui/screenshot/features.h"
 #endif
 
 #define EXPAND_FEATURE_ENTRIES(...) __VA_ARGS__,
@@ -828,6 +829,20 @@ constexpr flags_ui::FeatureEntry::Choice kVerticalTabCollapseDelayChoices[] = {
 #define BRAVE_WORKSPACE_FEATURE_ENTRY
 #endif
 
+#if defined(TOOLKIT_VIEWS)
+#define BRAVE_SCREENSHOT_FEATURE_ENTRY                                        \
+  EXPAND_FEATURE_ENTRIES({                                                    \
+      "brave-screenshot",                                                     \
+      "Enable Brave Screenshot",                                              \
+      "Adds a toolbar button for capturing the visible area or full page as " \
+      "a PNG.",                                                               \
+      kOsDesktop,                                                             \
+      FEATURE_VALUE_TYPE(screenshot::features::kBraveScreenshot),             \
+  })
+#else
+#define BRAVE_SCREENSHOT_FEATURE_ENTRY
+#endif
+
 // Keep the last item empty.
 #define LAST_BRAVE_FEATURE_ENTRIES_ITEM
 
@@ -1425,6 +1440,7 @@ constexpr flags_ui::FeatureEntry::Choice kVerticalTabCollapseDelayChoices[] = {
       kOsDesktop | kOsAndroid,                                                 \
       FEATURE_VALUE_TYPE(brave_origin::features::kBraveOrigin),                \
   })                                                                           \
+  BRAVE_SCREENSHOT_FEATURE_ENTRY                                               \
   LAST_BRAVE_FEATURE_ENTRIES_ITEM  // Keep it as the last item.
 namespace flags_ui {
 namespace {
--- a/browser/ai_chat/print_preview_extraction_delegate_impl.h
+++ b/browser/ai_chat/print_preview_extraction_delegate_impl.h
@@ -15,6 +15,10 @@
 
 static_assert(BUILDFLAG(ENABLE_PRINT_PREVIEW));
 
+namespace content {
+class WebContents;
+}  // namespace content
+
 namespace ai_chat {
 
 // Adapts the browser-layer PrintPreviewExtractor to the AI Chat
@@ -22,7 +26,8 @@ namespace ai_chat {
 class PrintPreviewExtractionDelegateImpl
     : public AssociatedWebContentsContent::PrintPreviewExtractionDelegate {
  public:
-  explicit PrintPreviewExtractionDelegateImpl(
+  PrintPreviewExtractionDelegateImpl(
+      content::WebContents* web_contents,
       std::unique_ptr<screenshot::PrintPreviewExtractor> extractor);
   ~PrintPreviewExtractionDelegateImpl() override;
 
@@ -36,6 +41,7 @@ class PrintPreviewExtractionDelegateImpl
 
  private:
   std::unique_ptr<screenshot::PrintPreviewExtractor> extractor_;
+  raw_ptr<content::WebContents> web_contents_;
 };
 
 }  // namespace ai_chat
--- a/browser/ui/browser_window/internal/BUILD.gn
+++ b/browser/ui/browser_window/internal/BUILD.gn
@@ -22,8 +22,10 @@ source_set("internal") {
     "//brave/components/brave_vpn/common/buildflags",
     "//brave/components/email_aliases/buildflags",
     "//brave/components/playlist/core/common/buildflags",
+    "//chrome/browser/profiles:profile",
     "//chrome/browser/tab_group_sync:factories",
     "//chrome/browser/ui/browser_window",
+    "//printing/buildflags",
   ]
 
   if (enable_playlist) {
@@ -48,10 +50,6 @@ source_set("internal") {
     ]
   }
 
-  deps += [
-    "//chrome/browser/profiles:profile",
-    "//printing/buildflags",
-  ]
   if (enable_print_preview) {
     deps += [ "//brave/browser/screenshot" ]
   }
--- a/browser/ui/browser_window/internal/BUILD.gn
+++ b/browser/ui/browser_window/internal/BUILD.gn
@@ -8,6 +8,7 @@ import("//brave/components/brave_vpn/common/buildflags/buildflags.gni")
 import("//brave/components/email_aliases/buildflags/buildflags.gni")
 import("//brave/components/playlist/core/common/buildflags/buildflags.gni")
 import("//build/config/ui.gni")
+import("//printing/buildflags/buildflags.gni")
 
 source_set("internal") {
   sources = []
@@ -39,11 +40,20 @@ source_set("internal") {
   if (!is_android) {
     deps += [
       "//brave/browser/ui/focus_mode",
+      "//brave/browser/ui/screenshot",
       "//brave/browser/ui/tabs:tab_menu",
       "//brave/browser/ui/tabs:tree_tab_session_manager",
       "//brave/browser/ui/views/page_info",
       "//chrome/browser/ui/tabs:features",
     ]
+  }
+
+  deps += [
+    "//chrome/browser/profiles:profile",
+    "//printing/buildflags",
+  ]
+  if (enable_print_preview) {
+    deps += [ "//brave/browser/screenshot" ]
   }
 
   if (enable_brave_rewards) {
--- a/browser/ui/browser_window/internal/browser_window_features.cc
+++ b/browser/ui/browser_window/internal/browser_window_features.cc
@@ -5,12 +5,15 @@
 
 #include "chrome/browser/ui/browser_window/public/browser_window_features.h"
 
+#include "base/containers/flat_map.h"
 #include "base/feature_list.h"
+#include "base/functional/bind.h"
 #include "base/memory/ptr_util.h"
 #include "base/notreached.h"
 #include "brave/browser/ui/brave_browser_window.h"
 #include "brave/browser/ui/focus_mode/focus_mode_controller.h"
 #include "brave/browser/ui/focus_mode/focus_mode_utils.h"
+#include "brave/browser/ui/screenshot/screenshot_controller.h"
 #include "brave/browser/ui/sidebar/sidebar_controller.h"
 #include "brave/browser/ui/sidebar/sidebar_utils.h"
 #include "brave/browser/ui/tabs/brave_browser_tab_menu_model_delegate.h"
@@ -21,11 +24,14 @@
 #include "brave/components/brave_vpn/common/buildflags/buildflags.h"
 #include "brave/components/email_aliases/buildflags/buildflags.h"
 #include "brave/components/playlist/core/common/buildflags/buildflags.h"
+#include "chrome/browser/profiles/profile.h"
 #include "chrome/browser/tab_group_sync/tab_group_sync_service_factory.h"
 #include "chrome/browser/ui/browser.h"
 #include "chrome/browser/ui/browser_window/public/browser_window_interface.h"
 #include "chrome/browser/ui/tabs/features.h"
 #include "chrome/browser/ui/views/frame/browser_view.h"
+#include "printing/buildflags/buildflags.h"
+#include "ui/gfx/native_ui_types.h"
 
 #if BUILDFLAG(ENABLE_BRAVE_REWARDS)
 #include "brave/browser/brave_rewards/rewards_service_factory.h"
@@ -34,6 +40,11 @@
 
 #if BUILDFLAG(ENABLE_BRAVE_VPN)
 #include "brave/browser/ui/brave_vpn/brave_vpn_controller.h"
+#endif
+
+#if BUILDFLAG(ENABLE_PRINT_PREVIEW)
+#include "brave/browser/screenshot/print_preview_extractor_factory.h"
+#include "chrome/browser/ui/webui/print_preview/print_preview_ui.h"
 #endif
 
 #if !BUILDFLAG(ENABLE_BRAVE_REWARDS)
@@ -145,12 +156,42 @@ void BrowserWindowFeatures::InitPostBrowserViewConstruction(
   brave_vpn_controller_ = std::make_unique<BraveVPNController>(browser_view);
 #endif
 
+#if BUILDFLAG(ENABLE_PRINT_PREVIEW)
+  {
+    auto extractor = screenshot::CreatePrintPreviewExtractor(
+        base::BindRepeating(
+            []() -> base::IDMap<printing::mojom::PrintPreviewUI*>& {
+              return printing::PrintPreviewUI::GetPrintPreviewUIIdMap();
+            }),
+        base::BindRepeating([]() -> base::flat_map<int, int>& {
+          return printing::PrintPreviewUI::GetPrintPreviewUIRequestIdMap();
+        }));
+    screenshot_controller_ = std::make_unique<screenshot::ScreenshotController>(
+        browser_view->GetProfile(),
+        base::BindRepeating(
+            [](BrowserView* bv) -> gfx::NativeWindow {
+              return bv->GetNativeWindow();
+            },
+            browser_view),
+        base::BindRepeating([]() { return base::FilePath(); }),
+        std::move(extractor));
+  }
+#else
+  screenshot_controller_ = std::make_unique<screenshot::ScreenshotController>(
+      browser_view->GetProfile(), base::BindRepeating(
+                                      [](BrowserView* bv) -> gfx::NativeWindow {
+                                        return bv->GetNativeWindow();
+                                      },
+                                      browser_view));
+#endif
+
   BrowserWindowFeatures_ChromiumImpl::InitPostBrowserViewConstruction(
       browser_view);
 }
 
 void BrowserWindowFeatures::TearDownPreBrowserWindowDestruction() {
   BrowserWindowFeatures_ChromiumImpl::TearDownPreBrowserWindowDestruction();
+  screenshot_controller_.reset();
 #if BUILDFLAG(ENABLE_BRAVE_VPN)
   brave_vpn_controller_.reset();
 #endif
--- a/browser/ui/browser_window/internal/browser_window_features.cc
+++ b/browser/ui/browser_window/internal/browser_window_features.cc
@@ -173,7 +173,6 @@ void BrowserWindowFeatures::InitPostBrowserViewConstruction(
               return bv->GetNativeWindow();
             },
             browser_view),
-        base::BindRepeating([]() { return base::FilePath(); }),
         std::move(extractor));
   }
 #else
--- a/browser/ui/screenshot/BUILD.gn
+++ b/browser/ui/screenshot/BUILD.gn
@@ -1,0 +1,55 @@
+# Copyright (c) 2026 The Brave Authors. All rights reserved.
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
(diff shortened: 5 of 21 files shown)


bridiver:
Please make sure to fix presubmit warnings

browser/about_flags.cc
```diff
@@ -1418,13 +1433,7 @@ constexpr flags_ui::FeatureEntry::Choice kVerticalTabCollapseDelayChoices[] = {
   BRAVE_FORCE_POPUP_TO_BE_OPENED_IN_NEW_TAB_FEATURE_ENTRY                      \
```
netzenbot, line 1433: This change deletes the existing `brave-origin` feature entry (FEATURE_VALUE_TYPE(brave_origin::features::kBraveOrigin)) and replaces it with `BRAVE_SCREENSHOT_FEATURE_ENTRY`. The brave-origin entry appears unrelated to this PR and should be kept; add the screenshot entry alongside it rather than replacing it.

browser/about_flags.cc
```diff
@@ -1418,13 +1433,7 @@ constexpr flags_ui::FeatureEntry::Choice kVerticalTabCollapseDelayChoices[] = {
...
-  EXPAND_FEATURE_ENTRIES({                                                     \
-      "brave-origin",                                                          \
-      "Enable Brave Origin",                                                   \
-      "Enables Brave Origin features and settings.",                           \
-      kOsDesktop | kOsAndroid,                                                 \
-      FEATURE_VALUE_TYPE(brave_origin::features::kBraveOrigin),                \
-  })                                                                           \
+  BRAVE_SCREENSHOT_FEATURE_ENTRY                                               \
   LAST_BRAVE_FEATURE_ENTRIES_ITEM  // Keep it as the last item.
 namespace flags_ui {
```
netzenbot, line 1445: This replaces the `brave-origin` flag entry with the screenshot macro, removing `brave-origin` from about:flags entirely. The screenshot entry should be appended without deleting `brave-origin` (looks like an accidental drop, likely a rebase artifact). ([Keep PRs Focused on a Single Purpose](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-070))

browser/about_flags.cc
```diff
@@ -1418,13 +1433,7 @@ constexpr flags_ui::FeatureEntry::Choice kVerticalTabCollapseDelayChoices[] = {
   BRAVE_FORCE_POPUP_TO_BE_OPENED_IN_NEW_TAB_FEATURE_ENTRY                      \
   EMAIL_ALIASES_FEATURE_ENTRIES                                                \
   BRAVE_WORKSPACE_FEATURE_ENTRY                                                \
```
netzenbot, line 1435: This replaces the `brave-origin` feature entry rather than adding the screenshot one alongside it, which removes Brave Origin from chrome://flags. `brave_origin::features::kBraveOrigin` is still defined and used, so the screenshot entry should be appended, not substituted for the brave-origin entry.

browser/ui/screenshot/screenshot_controller_unittest.cc
```diff
@@ -0,0 +1,205 @@
...
+
+  // Drives the pipeline from OnVisibleAreaCopied() onward, bypassing
+  // CopyFromSurface (which returns kNotImplemented in the test environment).
+  // Sets the controller into the same busy state that CaptureVisibleArea()
+  // would set before dispatching CopyFromSurface.
+  void InjectBitmap(SkBitmap bitmap, ScreenshotController::ResultCallback cb) {
+    controller_->busy_ = true;
+    controller_->pending_callback_ = std::move(cb);
+    controller_->OnVisibleAreaCopied(std::move(bitmap));
+  }
```
netzenbot, line 74: The full-page capture path (`CaptureFullPage` / `OnFullPageChunks` / `StitchAndEncode`) has no test coverage here. `StitchAndEncode` has meaningful logic worth testing — multi-chunk stitching, dimension computation, and the `kMaxStitchedBytes` cap. Consider adding a helper that injects chunk results (mirroring `InjectBitmap`) to exercise it. ([PRs Should Include Reasonable Test Coverage](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-039))

browser/ui/screenshot/screenshot_controller.h
```diff
@@ -0,0 +1,133 @@
+// Copyright (c) 2026 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#ifndef BRAVE_BROWSER_UI_SCREENSHOT_SCREENSHOT_CONTROLLER_H_
+#define BRAVE_BROWSER_UI_SCREENSHOT_SCREENSHOT_CONTROLLER_H_
+
+#include <memory>
```
netzenbot, line 9: nit: `ScreenshotController` only uses `profile_` for `DownloadPrefs::FromBrowserContext(profile_)`, which accepts a `content::BrowserContext*`. Per [Pass the Most Specific Dependency](https://github.com/brave/brave-core/tree/master/docs/best-practices/architecture.md#ARCH-003), consider storing a `content::BrowserContext*` instead of `Profile*`.

browser/ui/screenshot/BUILD.gn
```diff
@@ -0,0 +1,53 @@
...
+    "//chrome/test:test_support",
+    "//content/test:test_support",
+    "//printing/buildflags",
+    "//skia",
+    "//testing/gtest",
+    "//ui/gfx",
+    "//ui/gfx/codec",
+    "//ui/shell_dialogs:test_support",
+  ]
+}
```
netzenbot, line 55: nit: `screenshot_controller.h` is a public header of this target and includes `ui/shell_dialogs/select_file_dialog.h` (base class), `SkBitmap.h`, `ui/gfx/native_ui_types.h`, and `printing/buildflags/buildflags.h`. Header-visible deps should be in `public_deps` so consumers get the transitive include paths (see the sibling `focus_mode` target). ([Use `public_deps` for Header-File Includes in BUILD.gn](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-024))

browser/ui/screenshot/screenshot_controller.cc
```diff
@@ -0,0 +1,352 @@
...
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+  busy_ = false;
+  pending_png_.clear();
+  if (select_dialog_) {
+    select_dialog_->ListenerDestroyed();
+    select_dialog_.reset();
+  }
+}
+
+}  // namespace screenshot
```
netzenbot, line 361: nit: `WritePngFile` only reads `bytes` (and `base::WriteFile` accepts `base::span<const uint8_t>`), so `base::span<const uint8_t>` would be more flexible here than `const std::vector<uint8_t>&`. ([Use `base::span` at API Boundaries](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-017))

browser/ai_chat/print_preview_extraction_delegate_impl.h
```diff
@@ -35,6 +40,7 @@ class PrintPreviewExtractionDelegateImpl
   void CaptureImages(CaptureImagesCallback callback) override;
 
  private:
+  raw_ptr<content::WebContents> web_contents_;
```
netzenbot: nit: per Chromium convention, place unowned `raw_ptr<>` members after owning members. The new `web_contents_` raw_ptr should go after `extractor_` (the `unique_ptr`) rather than before it. ([Place raw_ptr<> Members Last in Class Declarations](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-memory.md#CSM-013))

browser/ui/screenshot/screenshot_controller.h
```diff
@@ -0,0 +1,133 @@
...
+
+  void OnEncoded(std::optional<std::vector<uint8_t>> png);
+  void ShowSaveDialog(std::vector<uint8_t> png);
+  void ShowSaveDialogWithPath(base::FilePath default_path);
+  // Reply callback for WritePngFile posted from FileSelected().
+  void OnFileWritten(base::FilePath path, bool ok);
+  void FinishWithError(Error error);
+  void Reset();
+
+  raw_ptr<Profile> profile_;
```
netzenbot: nit: per Chromium convention, place unowned `raw_ptr<>` members after owning members. `profile_` (a `raw_ptr`) is declared ahead of the owned `print_preview_extractor_` (`unique_ptr`); consider ordering owning members first. ([Place raw_ptr<> Members Last in Class Declarations](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-memory.md#CSM-013))

browser/ui/screenshot/screenshot_controller.cc
```diff
@@ -0,0 +1,351 @@
...
+                     weak_factory_.GetWeakPtr()));
+}
+
+#if BUILDFLAG(ENABLE_PRINT_PREVIEW)
+void ScreenshotController::OnFullPageChunks(ChunkResult result) {
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+
+  if (!result.has_value()) {
+    VLOG(1) << "Full page capture failed: " << result.error();
+    FinishWithError(Error::kCaptureFailed);
```
netzenbot, line 243: nit: prefer `DVLOG(1)` over `VLOG(1)` for this retained diagnostic log. DVLOG(1) avoids release binary bloat while still being enableable via `--v=1`/`--vmodule`. ([Use `DVLOG(1)` for Retained Logging](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-067))

browser/ui/browser_window/internal/BUILD.gn
```diff
@@ -39,13 +40,22 @@ source_set("internal") {
...
       "//brave/browser/ui/tabs:tab_menu",
       "//brave/browser/ui/tabs:tree_tab_session_manager",
       "//brave/browser/ui/views/page_info",
       "//chrome/browser/ui/tabs:features",
     ]
   }
 
+  deps += [
+    "//chrome/browser/profiles:profile",
+    "//printing/buildflags",
```
simonhong: we could insert these two above deps = [] list.

browser/ui/browser_window/internal/browser_window_features.cc
```diff
@@ -145,12 +156,42 @@ void BrowserWindowFeatures::InitPostBrowserViewConstruction(
   brave_vpn_controller_ = std::make_unique<BraveVPNController>(browser_view);
 #endif
 
+#if BUILDFLAG(ENABLE_PRINT_PREVIEW)
+  {
+    auto extractor = screenshot::CreatePrintPreviewExtractor(
```
simonhong, line 161: Should we create |extractor| here?
As ScreenshotController depends on print preview extractor,
ScreenshotController's ctor can create extractor.
I think it just add unnecessary print preview dependency to BrowserWindowFeatures.

Then, maybe we don't need two types of ctor?

browser/ui/screenshot/screenshot_controller.h
```diff
@@ -0,0 +1,133 @@
...
+    kUserCancelled,
+    kWriteFailed,
+  };
+  using ResultCallback =
+      base::OnceCallback<void(base::expected<base::FilePath, Error>)>;
+
+  // `parent_window_getter` is invoked just-in-time at Save-As dialog
+  // creation, after the browser window is fully initialized. Constructing the
+  // controller during BraveToolbarView::Init() (when browser->window() is
+  // still null) is intentional.
```
simonhong: This comment is invalid. It's not created from toolbar.

browser/ui/browser_window/internal/browser_window_features.cc
```diff
@@ -145,12 +156,42 @@ void BrowserWindowFeatures::InitPostBrowserViewConstruction(
...
+          return printing::PrintPreviewUI::GetPrintPreviewUIRequestIdMap();
+        }));
+    screenshot_controller_ = std::make_unique<screenshot::ScreenshotController>(
+        browser_view->GetProfile(),
+        base::BindRepeating(
+            [](BrowserView* bv) -> gfx::NativeWindow {
+              return bv->GetNativeWindow();
+            },
+            browser_view),
+        base::BindRepeating([]() { return base::FilePath(); }),
```
simonhong: It looks like both ctor doesn't need dir_getter?
FilePath() is passed and and below uses null.

browser/ui/browser_window/internal/browser_window_features.cc
sangwoo108: Yes, in order to avoid circular deps to the c/b/u target, it's intended design - due to deps to PrintPreviewUI

browser/ui/browser_window/internal/BUILD.gn
sangwoo108: done.

browser/ui/screenshot/screenshot_controller.h
```diff
@@ -0,0 +1,133 @@
...
+  using NativeWindowGetter = base::RepeatingCallback<gfx::NativeWindow()>;
+
+  // Returns the directory used as the default Save As path. If empty, falls
+  // back to DownloadPrefs::FromBrowserContext(profile_)->DownloadPath().
+  // Tests can inject a simpler implementation to avoid the full download
+  // service stack.
+  using DownloadDirGetter = base::RepeatingCallback<base::FilePath()>;
```
simonhong: If this dir getter is for test only, I think we can set via test only setter instead of ctor's args.

browser/ui/screenshot/screenshot_controller.h
sangwoo108: Good catch. removed.

browser/ui/screenshot/screenshot_controller.h
sangwoo108: Yup, SetDownloadDirGetterForTesting() would be better.

browser/ui/screenshot/screenshot_controller.h
```diff
@@ -0,0 +1,133 @@
...
+#endif
+  ScreenshotController(const ScreenshotController&) = delete;
+  ScreenshotController& operator=(const ScreenshotController&) = delete;
+  ~ScreenshotController() override;
+
+  bool busy() const { return busy_; }
+
+  // Returns false if `web_contents` has no live RenderWidgetHostView or a
+  // capture is already in flight.
+  bool CanCapture(content::WebContents* web_contents) const;
```
simonhong: don't need to worry about using raw pointer instead of WebContents::GetWeakPtr() in this class?

browser/ui/screenshot/screenshot_controller.h
```diff
@@ -0,0 +1,128 @@
...
+class BrowserContext;
+class WebContents;
+}  // namespace content
+
+namespace screenshot {
+class PrintPreviewExtractor;
+
+// Owned by BrowserWindowFeatures (one per Browser). Orchestrates capture of a
+// screenshot (visible area or full page), encodes the result as PNG, and
+// drives a Save As dialog. Does not expose ai_chat:: types to callers.
```
simonhong: what `Does not expose ai_chat:: types to callers` means?

* I'll leave a comment if it's last review comment. Until then, don't need to push several times :)

browser/ui/screenshot/screenshot_controller.h
sangwoo108: I think it's okay because web contents are referenced from UI thread synchronously.

browser/ui/screenshot/screenshot_controller.h
```diff
@@ -0,0 +1,128 @@
...
+
+  // Captures the visible viewport area or the full page of `web_contents`,
+  // encodes the result as PNG, and drives a Save As dialog.
+  void CaptureVisibleArea(content::WebContents* web_contents,
+                          ResultCallback done);
+#if BUILDFLAG(ENABLE_PRINT_PREVIEW)
+  void CaptureFullPage(content::WebContents* web_contents, ResultCallback done);
+#endif  // BUILDFLAG(ENABLE_PRINT_PREVIEW)
+
+  void set_download_dir_for_testing(base::FilePath path) {
```
simonhong: here and many other places
```suggestion
  void set_download_dir_for_testing(const base::FilePath& path) {
```

browser/ui/screenshot/screenshot_controller.cc
```diff
@@ -0,0 +1,349 @@
...
+  }
+  return web_contents->GetRenderWidgetHostView() != nullptr;
+}
+
+void ScreenshotController::CaptureVisibleArea(
+    content::WebContents* web_contents,
+    ResultCallback done) {
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+  if (!CanCapture(web_contents)) {
+    std::move(done).Run(base::unexpected(Error::kNoTab));
```
simonhong: passing `kNoTab` is ambiguous here. It could be busy

browser/ui/screenshot/screenshot_controller.h
```diff
@@ -0,0 +1,128 @@
...
+  void OnEncoded(std::optional<std::vector<uint8_t>> png);
+  void ShowSaveDialog(std::vector<uint8_t> png);
+  void ShowSaveDialogWithPath(base::FilePath default_path);
+  // Reply callback for WritePngFile posted from FileSelected().
+  void OnFileWritten(base::FilePath path, bool ok);
+  void FinishWithError(Error error);
+  void Reset();
+
+  NativeWindowGetter parent_window_getter_;
+  bool busy_ = false;
```
simonhong: nit: Maybe we can use `pending_callback_` as busy bit?

browser/ui/screenshot/screenshot_controller.h
```diff
@@ -0,0 +1,128 @@
...
+  ScreenshotController(content::BrowserContext* profile,
+                       NativeWindowGetter parent_window_getter,
+                       std::unique_ptr<screenshot::PrintPreviewExtractor>
+                           print_preview_extractor);
+#endif
+  ScreenshotController(const ScreenshotController&) = delete;
+  ScreenshotController& operator=(const ScreenshotController&) = delete;
+  ~ScreenshotController() override;
+
+  bool busy() const { return busy_; }
```
simonhong: no one calls this

browser/ui/screenshot/screenshot_controller.cc
```diff
@@ -0,0 +1,348 @@
...
+void ScreenshotController::OnFullPageChunks(ChunkResult result) {
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+
+  if (!result.has_value()) {
+    DVLOG(1) << "Full page capture failed: " << result.error();
+    FinishWithError(Error::kCaptureFailed);
+    return;
+  }
+  base::ThreadPool::PostTaskAndReplyWithResult(
+      FROM_HERE, {base::TaskPriority::USER_VISIBLE},
```
simonhong: maybe we want `base::TaskShutdownBehavior::SKIP_ON_SHUTDOWN` here and above?

simonhong (commented):
I don't have more.

browser/ui/screenshot/screenshot_controller_unittest.cc
```diff
@@ -0,0 +1,306 @@
...
+
+  Result result = future.Get();
+  ASSERT_TRUE(result.has_value());
+
+  std::optional<std::vector<uint8_t>> png_bytes =
+      base::ReadFileToBytes(save_path);
+  ASSERT_TRUE(png_bytes);
+  SkBitmap stitched = gfx::PNGCodec::Decode(*png_bytes);
+  EXPECT_EQ(stitched.width(), 64);
+  EXPECT_EQ(stitched.height(), 80);  // 48 + 32
```
netzenbot, line 311: nit: this verifies the stitched dimensions but not the vertical ordering of the chunks. Since StitchAndEncode composes chunks top-to-bottom in order, consider also sampling a pixel from each band (e.g. red in the top 48px, green below) so a future ordering regression is caught. ([Verify Ordering in Tests That Return Ordered Data](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-018))

browser/ui/screenshot/screenshot_controller.h
sangwoo108: TIL: I found interesting aspect of this. I saw many places where path is passed by value(https://source.chromium.org/search?q=%5C(base::FilePath%5C%20&ss=chromium&start=11) . When a class consumes the path and stores it to a member variable, they usually `std::move(path)` and pass the path by value in order to make it sure that the argument is rvalue. And the other cases are where the path is bound to a task that will be posted to another thread. This also guarantees that the path is rvalue that can be safely passed to another thread.

---

but I think `const &` is good enough for this case.

sangwoo108: One open question is that why they don't use FilePath&& in those cases.

browser/ui/screenshot/screenshot_controller.cc
sangwoo108: Good point. Made CanCapture returns Error::kBusy.

browser/ui/screenshot/screenshot_controller.h
simonhong: If call pass the ownership to method, we can use `FilePath&&`.

I think using `const FilePath&` as param type is standard way in chromium.
If method needs copy, just copy is fine. Only one copy.
If method just uses it, no copy happened.

If use `FilePath`, copy happend always even that method just refers it.

browser/ui/screenshot/screenshot_controller.h
sangwoo108: Good call.

browser/ui/screenshot/screenshot_controller.h
sangwoo108: Removed

browser/ui/screenshot/screenshot_controller.cc
sangwoo108: Good call

browser/ui/screenshot/screenshot_controller.h
sangwoo108: Removed

browser/ui/screenshot/screenshot_controller_unittest.cc
sangwoo108: Done.

simonhong (approved):
++

petemill (approved):
strings++ but please add a screenshot (😆) to the PR description. Did not check the c++

sangwoo108:
> strings++ but please add a screenshot (😆) to the PR description. Did not check the c++

The visible screenshot would arrive at next commit :) Thanks

PR #37184, opened 2026-06-11
Updated best practices for changing Java members visibility

### LINKED ISSUE #56620: Update plaster best practices to be used for Java members visibility
### Description

Update plaster best practices to be used for Java members visibility

AlexeyBarabash:
Resolves https://github.com/brave/brave-browser/issues/56620

### DIFF
--- a/docs/best-practices/plaster.md
+++ b/docs/best-practices/plaster.md
@@ -68,3 +68,63 @@ Reserve `re_pattern` for when you need regex features like whitespace matching,
 character classes, or structural patterns.
 
 ---
+
+<a id="PLSTR-003"></a>
+
+## ✅ Use Plaster to change private fields/methods visibility in Java for new code
+
+```yaml
+# ✅ CORRECT - simple visibility replacement for a field
+pattern: 'private final @Nullable View mNtpHeader;'
+replace: 'protected final @Nullable View mNtpHeader;'
+```
+
+```java
+// ❌ WRONG - Bytecode manipulation for field visibility change
+        deleteField(sBraveFeedSurfaceCoordinatorClassName, "mNtpHeader");
+        makeProtectedField(sFeedSurfaceCoordinatorClassName, "mNtpHeader");
+```
+
+```yaml
+# ✅ CORRECT - simple visibility replacement for a method
+pattern: 'private void runMenuItemEnterAnimations()'
+replace: 'protected void runMenuItemEnterAnimations()'
+```
+
+```java
+// ❌ WRONG - Bytecode manipulation for method visibility change
+
+// BraveAppMenuDummySuper.java
+class BraveAppMenuDummySuper extends AppMenu
+
+public void runMenuItemEnterAnimations() {
+    assert false : "This class usage should be removed in the bytecode!";
+}
+
+// BraveAppMenu.java
+public class BraveAppMenu extends BraveAppMenuDummySuper
+
+@Override
+public void runMenuItemEnterAnimations() {
+    // We do nothing here as we don't want any fancy animation for the menu.
+}
+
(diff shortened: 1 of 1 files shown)


docs/best-practices/plaster.md
```diff
@@ -68,3 +68,63 @@ Reserve `re_pattern` for when you need regex features like whitespace matching,
...
+    // We do nothing here as we don't want any fancy animation for the menu.
+}
+
+// BraveAppMenuClassAdapter.java
+    changeSuperName(sBraveAppMenuClassName, sAppMenuClassName);
+    makePublicMethod(sAppMenuClassName, "runMenuItemEnterAnimations");
+```
+
+```java
+// 💡 Note: you can still use Bytecode to replace the static method
```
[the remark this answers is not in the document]
AlexeyBarabash: yes, you are right, that was confusing.
updated at https://github.com/brave/brave-core/pull/37184/commits/57f5c6d683cc8390334f0e0e1462b4150529e26f

docs/best-practices/plaster.md
```diff
@@ -68,3 +68,62 @@ Reserve `re_pattern` for when you need regex features like whitespace matching,
 character classes, or structural patterns.
 
 ---
+
+<a id="PLSTR-003"></a>
```
bridiver, line 72: @bbondy do you think we should split these up by source file type or just keep them all together?

bridiver: oh wait nvm, this is plaster.md

docs/best-practices/plaster.md
```diff
@@ -68,3 +68,62 @@ Reserve `re_pattern` for when you need regex features like whitespace matching,
...
+
+```java
+// ❌ WRONG - Bytecode manipulation for field visibility change
+        deleteField(sBraveFeedSurfaceCoordinatorClassName, "mNtpHeader");
+        makeProtectedField(sFeedSurfaceCoordinatorClassName, "mNtpHeader");
+```
+
+```yaml
+# ✅ CORRECT - simple visibility replacement for a method
+pattern: 'private void runMenuItemEnterAnimations()'
```
bridiver: We should avoid matching on the method params, that can be fragile. I think the examples would be better as regex with match groups, wdyt @cdesouza-chromium ? I'm not opposed to just doing `private void runMenuItemEnterAnimations(` -> `protected void runMenuItemEnterAnimations(` since it's pretty simple, but usually we try to avoid duplication. I guess there's some tradeoffs here so I'll defer to @cdesouza-chromium

AlexeyBarabash: Updated example to avoid include arguments into `pattern`
https://github.com/brave/brave-core/pull/37184/commits/e9d93e708b601a142c13ce5ffe18eb1053014654

I don't see advantages to use groups capture like
```yaml
re_pattern: 'private (void runMenuItemEnterAnimations)'
replace: 'protected \1'
``` 
straight replace requires less efforts to understand it.  

but I can update the doc, if this options seems better. @cdesouza-chromium WDYT?

netzenbot (approved):

docs/best-practices/plaster.md
bridiver: yes, straight replace is simpler to read, but more error prone. I also don't want people to get in the habit of copying everything because that's not great in general. Match group replacements are so common that people will get used to it quickly

AlexeyBarabash: This is strong reason, I agree completely,

updated at https://github.com/brave/brave-core/pull/37184/changes/4a097b86920cc76209f1331e3dd7afdb14170196 to use group replacement in the example

AlexeyBarabash:
closing this as Java rewrites are planned

PR #37185, opened 2026-06-11
Move npm install to sync/init pre-commands.

### LINKED ISSUE #56287: Run npm install before sync in a visible way
Right now `npm run sync` runs a full sync and runs DEPS hooks one of which calls `npm install` internally to sync `node_modules` directory to the required state.

We should call install explicitly on sync/init instead, and after migration to `pnpm` let it do the same in a clear way.

Note: `pnpm` has a feature to perform auto-install on any script call, but we will disable it as it takes ~500ms on each `build`, `test`, ... commands, which is not great. Also our build calls some scripts directly via `npm`, which also add this crazy overhead during the build.

goodov:
Right now `npm run sync` runs DEPS hooks, one of which calls `npm install` internally to sync `node_modules` directory to the required state.

This PR makes this `install` call be explicit before `sync/init`. The reason for this change is to make `install` be visible and in the right order.

CI always call `npm install` (without `--yes`) explicitly before running any package script, so this change is a no-op for CI.

Developers usually only call `npm run sync` after rebase and rely on the hidden `npm install` during `bootstrap.py`. This call should be done _before_ any package script is run to sync `node_modules` for any command that might need it.

Resolves https://github.com/brave/brave-browser/issues/56287


package.json
```diff
@@ -10,7 +10,7 @@
     "init": "cd ../../ && npm run --prefix src/brave sync -- --init",
     "delete_string_translations": "node ./build/commands/scripts/commands.js delete_string_translations",
     "create_dist": "node ./build/commands/scripts/commands.js build --target=create_dist",
-    "sync": "node ./build/commands/scripts/sync.js",
+    "sync": "npm install --no-save --no-fund --no-audit && node ./build/commands/scripts/sync.js",
```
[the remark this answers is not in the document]
goodov: I didn't add it intentionally here. The install is now explicit and any required interaction will be surfaced without blind approvals.

diracdeltas, line 13: you can do `npm ci` here and `--no-save` is not needed anymore. also can we add the security defaults from https://github.blog/changelog/2026-06-09-upcoming-breaking-changes-for-npm-v12/ ? so it becomes

`"npm ci --no-fund --no-audit --ignore-scripts --allow-git=none --allow-remote=none`

goodov: I wish it was that easy.. We can't use `npm ci` for brave-core build process, because this command recreates `node_modules` from scratch, which is very slow and triggers a lot of targets to be rebuilt for developers. That's one of the reasons why we're switching to pnpm (this PR is a first step).

`--ignore-scripts --allow-git=none --allow-remote=none` this needs testing and also this needs to be added on CI to make sense. I'm not going to cover it here, because the goal of this PR is to simplify pnpm switch down the line, which will have all of this by default.

diracdeltas (approved):

mkarolin (approved):
++

fallaciousreasoning (approved):

petemill (approved):
Please can you add a little to the PR description or Issue about why this is better than running in sync, and why removing -y is ok including for CI (i.e. why it was added in the first place and why it isn't relevant anymore).

package.json
bridiver: We're switching to `pnpm`? When was that decided/discussed? I'm not necessarily opposed, it's just news to me.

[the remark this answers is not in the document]
bridiver: What was the problem before? It can waste a lot of time to come back to your computer after running a command and find out it did nothing because it was waiting for a prompt. What blind approvals are now explicit?

netzenbot (approved):

package.json
goodov: It was once again triggered by an attempt to roll `npm ci` everywhere, and we had a discussion about other package managers in this context.

https://bravesoftware.slack.com/archives/C02QUN5HRRT/p1754450052662819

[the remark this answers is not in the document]
goodov: There was a problem in WDP install https://github.com/brave/brave-browser/issues/21750

I don't think we ever hit that in brave-core, because on CI we always call `npm install` as a build step, and we never pass `--yes` on CI. The flag was added just as a safe measure at the same time I fixed WDP.

goodov:
> Please can you add a little to the PR description or Issue about why this is better than running in sync, and why removing -y is ok including for CI (i.e. why it was added in the first place and why it isn't relevant anymore).

Adjusted description and added the history bit about initial `--yes` introduction here https://github.com/brave/brave-core/pull/37185#discussion_r3400443359

PR #37195, opened 2026-06-11
[iOS] Add Farbling JavaScriptFeature

### LINKED ISSUE #53983: Support fingerprint protections when using single profile web view configuration
Support enabling the fingerprint protections which is powered by JavaScript when the `UseProfileWebViewConfiguration` feature flag is enabled.

kylehickinson:
This adds support for fingerprint protections while the `UseProfileWebViewConfiguration` feature flag is enabled

Resolves https://github.com/brave/brave-browser/issues/53983

### Testplan

1. Enable `UseProfileWebViewConfiguration` / single configuration feature flag and re-launch.
2. Visit https://dev-pages.brave.software/fingerprinting/farbling.html and complete the steps

### DIFF
--- a/ios/browser/brave_shields/BUILD.gn
+++ b/ios/browser/brave_shields/BUILD.gn
@@ -13,16 +13,24 @@ source_set("brave_shields") {
     "brave_shields_settings_service_factory_bridge.mm",
     "cookie_control_javascript_feature.h",
     "cookie_control_javascript_feature.mm",
+    "farbling_args.h",
+    "farbling_args.mm",
+    "farbling_javascript_feature.h",
+    "farbling_javascript_feature.mm",
   ]
 
   deps = [
     ":cookie_control_js",
+    ":farbling_js",
     "//base",
     "//brave/components/brave_shields/core/browser",
+    "//brave/components/brave_shields/core/common",
+    "//brave/components/brave_shields/core/common:mojom",
     "//brave/components/brave_shields/ios/browser",
     "//brave/ios/browser/api/profile",
     "//brave/ios/browser/keyed_service",
     "//brave/ios/browser/shared/prefs",
+    "//components/content_settings/core/browser",
     "//components/keyed_service/core",
     "//components/prefs",
     "//components/user_prefs",
@@ -32,10 +40,16 @@ source_set("brave_shields") {
     "//ios/chrome/browser/shared/model/profile:profile_keyed_service_factory",
     "//ios/web/public",
     "//ios/web/web_state/ui:wk_web_view_configuration_provider_header",
+    "//third_party/abseil-cpp:absl",
     "//url",
   ]
 
-  public_deps = [ "//ios/web/public/js_messaging" ]
+  public_deps = [
+    "//base",
+    "//brave/ios/web/js_messaging:message_handler_token",
+    "//brave/ios/web/js_messaging:randomized_message_handler_name",
+    "//ios/web/public/js_messaging",
+  ]
 
   frameworks = [ "Foundation.framework" ]
 }
@@ -44,4 +58,24 @@ optimize_ts("cookie_control_js") {
   visibility = [ ":brave_shields" ]
   sources = [ "resources/cookie_control.ts" ]
   deps = [ "//ios/web/public/js_messaging:gcrweb" ]
+}
+
+source_set("unit_tests") {
+  testonly = true
+
+  sources = [ "farbling_args_unittest.mm" ]
+
+  deps = [
+    ":brave_shields",
+    "//base",
+    "//base/test:test_support",
+    "//brave/components/brave_shields/core/common",
+    "//testing/gtest",
+  ]
+}
+
+optimize_ts("farbling_js") {
+  visibility = [ ":brave_shields" ]
+  sources = [ "resources/farbling.ts" ]
+  deps = [ "//brave/ios/web/js_messaging:util_scripts" ]
 }
--- a/ios/browser/brave_shields/farbling_args.mm
+++ b/ios/browser/brave_shields/farbling_args.mm
@@ -1,0 +1,150 @@
+// Copyright (c) 2026 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#include "brave/ios/browser/brave_shields/farbling_args.h"
+
+#include <cstdint>
+#include <string>
+#include <string_view>
+#include <vector>
+
+#include "base/containers/fixed_flat_set.h"
+#include "base/feature_list.h"
+#include "base/strings/string_util.h"
+#include "base/token.h"
+#include "brave/components/brave_shields/core/common/features.h"
+#include "third_party/abseil-cpp/absl/random/random.h"
+
+namespace brave_shields {
+
+namespace {
+
+// Seeded pseudo-random generator used to deterministically derive farbling
+// values for a given origin. This is the same generator that the rest of
+// Brave's farbling code uses (see `BraveSessionCache` and
+// `BraveFarblingService`), seeded from the origin's persistent farbling token.
+using FarblingPRNG = absl::random_internal::randen_engine<uint64_t>;
+
+constexpr double kMaxUInt64AsDouble = static_cast<double>(UINT64_MAX);
+
+// The parts that are combined to form a randomly generated fake plugin name.
+// An empty entry represents an omitted part.
+constexpr auto kPluginNameFirstParts = base::MakeFixedFlatSet<std::string_view>(
+    {"Chrome", "Chromium", "Brave", "Web", "Browser", "OpenSource", "Online",
+     "JavaScript", "WebKit", "Web-Kit", "WK", ""});
+
+constexpr auto kPluginNameSecondParts =
+    base::MakeFixedFlatSet<std::string_view>(
+        {"PDF", "Portable Document Format", "portable-document-format",
+         "document", "doc", "PDF and PS", "com.adobe.pdf", ""});
+
+constexpr auto kPluginNameThirdParts = base::MakeFixedFlatSet<std::string_view>(
+    {"Viewer", "Renderer", "Display", "Plugin", "plug-in", "plug in",
+     "extension", ""});
+
+// A list of fake voice names used to generate a fake `SpeechSynthesizer` voice.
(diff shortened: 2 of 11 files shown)


ios/browser/brave_shields/BUILD.gn
```diff
@@ -32,10 +42,14 @@ source_set("brave_shields") {
     "//ios/chrome/browser/shared/model/profile:profile_keyed_service_factory",
     "//ios/web/public",
     "//ios/web/web_state/ui:wk_web_view_configuration_provider_header",
+    "//third_party/abseil-cpp:absl",
     "//url",
   ]
```
netzenbot, line 45: nit: `farbling_javascript_feature.h` includes `message_handler_token.h` / `randomized_message_handler_name.h` and holds `MessageHandlerToken`/`RandomizedMessageHandlerName` as value members, so per [BS-024](https://github.com/brave/brave-core/tree/master/docs/best-practices/build-system.md#BS-024) these two deps belong in `public_deps` rather than `deps` so consumers of this header get the transitive include paths.

ios/browser/brave_shields/farbling_javascript_feature.mm
```diff
@@ -0,0 +1,107 @@
...
+              web::JavaScriptFeature::FeatureScript::TargetFrames::kAllFrames,
+              web::JavaScriptFeature::FeatureScript::ReinjectionBehavior::
+                  kInjectOncePerWindow,
+              base::BindRepeating(&FarblingJavaScriptFeature::GetReplacements,
+                                  base::Unretained(this)))}) {}
+
+FarblingJavaScriptFeature::~FarblingJavaScriptFeature() = default;
+
+// static
+FarblingJavaScriptFeature* FarblingJavaScriptFeature::GetInstance() {
```
netzenbot, line 48: nit: This shields/content-settings path checks `request_url.is_valid()` but not the scheme. Consider an early return for non-HTTP(S) URLs (`request_url.SchemeIsHTTPOrHTTPS()`) before the `GetFingerprintMode`/`GetFarblingToken` lookups to skip irrelevant schemes (e.g. about:, data:). ([Short-Circuit on Non-HTTP(S) URLs](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-032))

ios/browser/brave_shields/farbling_args.mm
```diff
@@ -0,0 +1,159 @@
...
+
+// Returns a random element from `values` drawn from `prng`, matching the Swift
+// `Array.seededRandom()` helper. `values` must not be empty.
+template <typename Container>
+std::string_view NextElement(FarblingPRNG& prng, const Container& values) {
+  const int index =
+      NextIntInRange(prng, 0, static_cast<int>(values.size()) - 1);
+  return *(values.begin() + index);
+}
+
```
netzenbot, line 78: nit: These dict keys are written here and read back with the same literals in farbling_args_unittest.mm. Consider defining named constants for the JSON property keys to prevent typos and ease refactoring. ([Use Named Constants for JSON Property Keys](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards-apis.md#CSA-054))

ios/browser/brave_shields/farbling_args.mm
```diff
@@ -0,0 +1,159 @@
...
+    plugins.Append(std::move(plugin));
+  }
+  return plugins;
+}
+
+}  // namespace
+
+base::DictValue MakeFarblingArgs(const base::Token& farbling_token) {
+  // Seed the generator from the origin's persistent farbling token, matching
+  // how the rest of Brave's farbling code derives its seed. This replaces the
```
netzenbot: nit: This comment describes the change and the prior Swift implementation rather than the code as it stands. A future reader with no knowledge of the migration would just need to know the seed source. Consider trimming "This replaces the previous Swift implementation which seeded srand48 with a per-session HMAC-derived domain key." See https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-019

ios/browser/brave_shields/farbling_javascript_feature.mm
```diff
@@ -0,0 +1,107 @@
...
+      addEntriesFromDictionary:handler_name_.GetPlaceholderReplacements()];
+  return [replacements copy];
+}
+
+std::optional<std::string>
+FarblingJavaScriptFeature::GetScriptMessageHandlerName() const {
+  return handler_name_.GetScriptMessageHandlerName();
+}
+
+bool FarblingJavaScriptFeature::GetFeatureRepliesToPrompts() const {
```
netzenbot, line 67: nit: Definition order here doesn't match the header. The header declares ScriptMessageReceivedWithReply before GetFeatureRepliesToPrompts, so swapping these two definitions would keep them in sync. See https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-017

StephenHeaps (commented):
I have request blocking in `ios/browser/brave_shields/request_blocking/`, and cosmetic filtering in `ios/browser/brave_shields/cosmetic_filtering/` with their own `BUILD.gn` / target.

I don't have a strong preference, but seems like it could be better organization to split these JavaScript Features up?

kylehickinson:
@StephenHeaps No strong preference in that regard, I can shift this into its own subfolder to match up

kylehickinson:
As discussed in DMs, Farbling uses the `BraveShieldsSettingsFactory`, so best to avoid possible circular deps and keep this in the same folder, may revisit overall folder structure in the future

ios/browser/brave_shields/farbling_args.mm
kylehickinson: since these are only duplicated in tests I'm going to leave it for now, the keys themselves would end up being in an anonymous namespace and not accessible anyways

netzenbot (approved):

StephenHeaps (commented):
With the single config flag enabled and shields enabeld, I'm seeing the same values for plugins, hardwareConcurrency, etc across the 2 different test site domains 🤔
Image
Image

StephenHeaps (approved):
Checked a couple different devices against previous issue I reported and farbling values differ per-test site now. LGTM, thanks.

ios/browser/brave_shields/farbling_args.mm
```diff
@@ -0,0 +1,150 @@
...
+// values for a given origin. This is the same generator that the rest of
+// Brave's farbling code uses (see `BraveSessionCache` and
+// `BraveFarblingService`), seeded from the origin's persistent farbling token.
+using FarblingPRNG = absl::random_internal::randen_engine<uint64_t>;
+
+constexpr double kMaxUInt64AsDouble = static_cast<double>(UINT64_MAX);
+
+// The parts that are combined to form a randomly generated fake plugin name.
+// An empty entry represents an omitted part.
+constexpr auto kPluginNameFirstParts = base::MakeFixedFlatSet<std::string_view>(
```
bridiver, line 34: why are we managing a static list here? Don't we have the same values in desktop?

bridiver: really the same thing applies for most of this file I think? This doesn't look like code that should be ios specific?

ios/browser/brave_shields/resources/farbling.ts
```diff
@@ -0,0 +1,327 @@
+// Copyright (c) 2026 The Brave Authors. All rights reserved.
```
bridiver, line 1: it's not clear to me what this script is or how it is used, it could use some comments to explain

ios/browser/brave_shields/resources/farbling.ts
```diff
@@ -0,0 +1,327 @@
...
+
+  // 2. Farble plugin data
+  // Injects fake plugins with fake mime-types
+  // Random plugins are determined by the plugin data
+  const farblePlugins = (fakePluginData: PluginData[]) => {
+    if (fakePluginData.length === 0) {
+      return
+    }
+    // Function that create a fake mime-type based on the given fake data
+    const makeFakeMimeType = (fakeData: MimeTypeData): MimeType => {
```
bridiver, line 93: it would be better I think to keep this logic in native and just pass the relevant `window` information to it rather than apply the farbling here where it's more difficult to test? It would also make it easier to share the code with desktop/android

bridiver: We generally try to keep logic like this outside of JS anyway so we could have something like `maybeFarblePlugins` that passes in the existing plugin data and returns the (maybe) farbled version depending on whether it is enabled or not

ios/browser/brave_shields/resources/farbling.ts
kylehickinson: Its a port of https://github.com/brave/brave-core/blob/90d9436232ae13ab70c64a379d06b84ff9a50835/ios/brave-ios/Sources/Brave/Frontend/UserContent/UserScripts/Scripts_Dynamic/Scripts/Paged/FarblingProtectionScript.js

Not sure what I would say other than "overriding common JS APIs used to fingeprint users"?

ios/browser/brave_shields/farbling_args.mm
kylehickinson: ported from https://github.com/brave/brave-core/blob/90d9436232ae13ab70c64a379d06b84ff9a50835/ios/brave-ios/Sources/Brave/Frontend/Browser/UserScripts/FarblingProtectionHelper.swift#L50

ios/browser/brave_shields/resources/farbling.ts
bridiver: Well outside of this PR it wouldn't even be clear that it's used for a javascript_feature so yes, just some basic explanation would be useful

ios/browser/brave_shields/resources/farbling.ts
bridiver: I guess since we're moving an existing ios script we shouldn't refactor it here, but for follow-up

ios/browser/brave_shields/farbling_args.mm
bridiver: for follow-up

bridiver (approved):
content setting changes ok

ios/browser/brave_shields/resources/farbling.ts
kylehickinson: Added a `@fileoverview` below the license like other Chromium scripts with a note about what this script does

ios/browser/brave_shields/farbling_args.mm
kylehickinson: https://github.com/brave/brave-browser/issues/56664

stoletheminerals (approved):

PR #37202, opened 2026-06-11
Upgrade from Chromium 149.0.7827.103 to Chromium 149.0.7827.115 (1.92.x)

emerick:
Resolves https://github.com/brave/brave-browser/issues/56299


kjozwiak (approved):
Uplift into `1.92.x` approved 👍

PR #37204, opened 2026-06-11
[iOS] Remove opaque toolbar/navigation bar backgrounds

### LINKED ISSUE #56315: Remove opaque toolbar/navigation bar backgrounds
In preparation for Liquid Glass, and to match new design styles, drop explicit backgrounds on navigation bars & toolbars

### Test
In UI that had navigation bars at the top (bookmarks, settings, history, etc), ensure they dont have a solid opaque background initially. On iOS 18 when you scroll down it will become a standard toolbar background and on 26 it will be a soft progressive blur when scrolling

kylehickinson:
Removes any opaque backgrounds in preparation for adopting Liquid Glass with the exception of some legacy UI in Brave Wallet (which is not presented anymore and would be removed in the future) and `BraveVPNInstallViewController` which needs to be replaced as its using `#colorLiteral`s still.

Resolves https://github.com/brave/brave-browser/issues/56315


netzenbot (approved):

StephenHeaps (commented):
Should these changes be behind the `LiquidGlassMode.isEnabled` check?

kylehickinson:
@StephenHeaps No, design wants them removed them even on iOS 18

StephenHeaps (approved):

ios/brave-ios/Sources/Brave/Frontend/Shields/ShieldsPanelView.swift
```diff
@@ -87,7 +87,7 @@ struct ShieldsPanelView: View {
     }
     .background(Color(braveSystemName: .containerBackground))
     .frame(idealWidth: 360, alignment: .center)
-    .toolbar(.hidden)
+    .toolbarBackground(.hidden, for: .navigationBar)
```
StephenHeaps: Something off here
Image

StephenHeaps (changes requested):

ios/brave-ios/Sources/Brave/Frontend/Shields/ShieldsPanelView.swift
StephenHeaps: Probably still need `.toolbar(.hidden, for: .navigationBar)` or `.toolbar(.hidden)` in addition to hiding background?

kylehickinson: Yeah I will replace it with `toolbarVisibility` API (18+) after https://github.com/brave/brave-core/pull/37270 lands

kylehickinson: Fixed in latest push

StephenHeaps (approved):

PR #37206, opened 2026-06-11
[skus] Migrate HttpResponse cxx bridge off CxxVector/CxxString

bridiverbot:
## Summary

Part of brave/brave-browser#55722 (auditing/cleaning up `rust::Str`/`rust::String`/`CxxString`/`CxxVector` misuse across the cxx FFI bridges). This is the brave-core SKUs piece.

The `HttpResponse` struct passed across the Rust↔C++ cxx bridge borrowed `&CxxVector<CxxString>` (headers) and `&CxxVector<u8>` (body). Sharing `std::vector`/`std::string` allocations with Rust this way relies on matching allocators across the FFI boundary, which is the pattern flagged in the audit.

This change switches `HttpResponse` to **owned** `Vec<String>`/`Vec<u8>`, matching the sibling `HttpRequest` struct (which already uses owned types), and deep-copies the data on the C++ side into a `rust::Vec` before handing it over.

## Notes

- Header values are constructed with `rust::String::lossy()` rather than the throwing `rust::String(const std::string&)` constructor. Server response header values are not guaranteed to be UTF-8, and Chromium is built with exceptions disabled — the throwing ctor would abort on a non-UTF-8 header. The Rust side already validates header names/values with `HeaderName::try_from`/`HeaderValue::try_from` and returns a clean `InternalError`, so `lossy()` preserves the prior graceful-failure behavior.
- The body is copied byte-for-byte into a `rust::Vec<uint8_t>`, equivalent to the previous `std::vector<uint8_t>` assignment.

## Test plan

- [x] `npm run build -- --target=brave/components/skus/browser:browser`
- [x] `npm run build -- --target=brave/components/skus/browser:unit_tests`

### DIFF
--- a/components/skus/browser/rs/cxx/src/httpclient.rs
+++ b/components/skus/browser/rs/cxx/src/httpclient.rs
@@ -46,14 +46,13 @@ impl TryFrom<http::Request<Vec<u8>>> for ffi::HttpRequest {
     }
 }
 
-impl From<ffi::HttpResponse<'_>> for Result<http::Response<Vec<u8>>, InternalError> {
-    fn from(resp: ffi::HttpResponse<'_>) -> Self {
+impl From<ffi::HttpResponse> for Result<http::Response<Vec<u8>>, InternalError> {
+    fn from(resp: ffi::HttpResponse) -> Self {
         match resp.result.code {
             ffi::SkusResultCode::Ok => {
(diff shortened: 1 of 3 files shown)


cmazakas (approved):
Rust changes LGTM!

PR #37211, opened 2026-06-12
[toolchain] Use --stage 1 for prebuilt-rustc wasm std build

cdesouza-chromium:
We are trying to work around reproducibility failures involving building
rust. This failures have been occasioned by upstream, but they affect us
nonetheless.

This PR changes the build for the wasm32 std build to `--stage=1`, as
the previous PR had erroneously used `--stage=0`. This is necessary to
collect the rustlib systroot.


AlexeyBarabash (approved):
++

PR #37212, opened 2026-06-12
[Sidebar V2] Update panel's default minimum width

simonhong:

Apply our minimum width. Patched because it's defined and used in the same file.


cdesouza-chromium (approved):
LGTM

PR #37216, opened 2026-06-12
[farbling] Add browser tests for blob:// rendered inside iframe

r0hit0303:
This change adds more tests to check the current behaviour of farbling when the blob:// URL is loaded in an iframe. The [expectation](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/third_party/blink/renderer/core/farbling/brave_session_cache.cc?L58-60) is that iframe receives the same farbling seed as the top frame so they both should see the same values. The test checks that. 

The main change to support farbling on blob:// URLs when opened as a new window is tracked at https://github.com/brave/brave-core/pull/37153

Related https://github.com/brave/brave-browser/issues/56048

### DIFF
--- a/browser/farbling/brave_blob_screen_farbling_browsertest.cc
+++ b/browser/farbling/brave_blob_screen_farbling_browsertest.cc
@@ -124,8 +124,6 @@ class BraveBlobScreenFarblingBrowserTest
     NavigateToBlobIframe();
   }
 
-  // Like NavigateToBlob but the blob URL is loaded inside an iframe rather
-  // than a new popup window, so the pass/fail title is set on the main frame.
   void NavigateToBlobIframe() {
     ASSERT_TRUE(ui_test_utils::NavigateToURL(browser(), blob_test_url_));
     content::ExecuteScriptAsync(Parent(),
--- a/browser/farbling/brave_blob_screen_farbling_browsertest.cc
+++ b/browser/farbling/brave_blob_screen_farbling_browsertest.cc
@@ -6,6 +6,7 @@
 #include <string>
 
 #include "base/path_service.h"
+#include "base/test/run_until.h"
 #include "base/test/scoped_feature_list.h"
 #include "brave/components/brave_shields/core/browser/brave_shields_utils.h"
 #include "brave/components/constants/brave_paths.h"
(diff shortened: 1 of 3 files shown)


r0hit0303:
@cdesouza-chromium Could you ptal? Thanks

browser/farbling/brave_blob_screen_farbling_browsertest.cc
```diff
@@ -117,6 +113,25 @@ class BraveBlobScreenFarblingBrowserTest
...
+    NavigateToBlobIframe();
+
+    SetFingerprintingSetting(/*allow=*/true);
+    NavigateToBlobIframe();
+  }
+
+  void NavigateToBlobIframe() {
+    ASSERT_TRUE(ui_test_utils::NavigateToURL(browser(), blob_test_url_));
+    content::ExecuteScriptAsync(Parent(),
+                                "startBlobScreenFingerprintingIframeTest()");
```
netzenbot, line 131: `TitleWatcher` here only registers `u"pass"`, but `startBlobScreenFingerprintingIframeTest` can set the title to `"fail"` or `"timeout"`. Those outcomes will never wake the watcher, so a real failure hangs until the overall browser-test timeout rather than failing fast. Register the failure titles too (e.g. add `u"fail"`/`u"timeout"` to the watcher) so `WaitAndGetTitle()` returns and the assertion reports the actual title. ([Search for Existing Chromium Patterns](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-isolation.md#TI-006))

test/data/blob-fingerprinting.html
```diff
@@ -73,6 +76,54 @@
...
+          return;
+        }
+
+        if (localStorage.getItem('blob_done') !== 'true') return;
+
+        clearInterval(poller);
+        URL.revokeObjectURL(blobUrl);
+        document.title = storedScreenValuesMatch('parent', 'blob')
+            ? 'pass'
+            : 'fail';
```
netzenbot: This JS-side `setInterval` poller with a hardcoded 50ms interval and 200-attempt timeout is the unreliable-timing pattern [TN-002](https://github.com/brave/brave-core/tree/master/docs/best-practices/testing-navigation.md#TN-002) warns against. Prefer driving the wait from C++ with `base::test::RunUntil()` polling the `blob_done` localStorage flag (e.g. via `EvalJs`), so the test waits on the real condition with no arbitrary timeout in the page.

ShivanKaul (approved):

r0hit0303:
Thanks @ShivanKaul for the review

PR #37217, opened 2026-06-12
[query-filter] Update test only code in query filter component installer

### LINKED ISSUE #56328: Refactor query_filter_component_installer to make testing logic more clear
### Description

cc @cdesouza-chromium who had the idea for improvement around this [block] (https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/components/query_filter/browser/query_filter_component_installer.cc?L52-56) to be moved as a lambda directly in the PostTaskAndReplyWithResult callback in lieu of [OnQueryFilterFileRead](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/components/query_filter/browser/query_filter_component_installer.cc?L121).

r0hit0303:
This change slightly updates the [callback](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/components/query_filter/browser/query_filter_component_installer.cc?L114) logic after ReadQueryFilterFile is fired. Previously, the result of [ReadQueryFilterFile](https://sourcegraph.com/r/github.com/brave/brave-core@023dc2f6a5f03d105c703f2f2109f27c1719b8d7/-/blob/components/query_filter/browser/query_filter_component_installer.cc?L40) was directly plugged into [OnQueryFilterFileRead](https://sourcegraph.com/r/github.com/brave/brave-core/-/blob/components/query_filter/browser/query_filter_component_installer.cc?L49) which took care of not only the core responsibility (i.e populating the query filter data), but also had some test-only logic hidden away. 

This change modifies the callback to instead calling `OnQueryFilterFileRead directly, it calls a lambda that breaks the previous work of `OnQueryFilterFileRead` into distinct individual parts and simplifying the `OnQueryFilterFileRead` code. 

Resolves https://github.com/brave/brave-browser/issues/56328

### DIFF
--- a/components/query_filter/browser/query_filter_component_installer.cc
+++ b/components/query_filter/browser/query_filter_component_installer.cc
@@ -23,7 +23,6 @@
 #include "brave/components/query_filter/common/features.h"
 #include "components/component_updater/component_installer.h"
 #include "components/component_updater/component_updater_service.h"
-#include "third_party/abseil-cpp/absl/cleanup/cleanup.h"
 
 namespace {
 inline constexpr char kQueryFilterComponentName[] = "Query Filter";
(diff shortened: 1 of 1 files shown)


r0hit0303:
@cdesouza-chromium could you ptal? Thanks

components/query_filter/browser/query_filter_component_installer.cc
```diff
@@ -118,7 +110,21 @@ void QueryFilterComponentInstallerPolicy::ComponentReady(
       base::BindOnce(
           &ReadQueryFilterFile,
           install_dir.AppendASCII(query_filter::kQueryFilterJsonFile)),
-      base::BindOnce(&OnQueryFilterFileRead, version));
+      base::BindOnce(
+          [](base::Version version, const std::string& json_data) {
+            if (OnQueryFilterFileRead(json_data)) {
+              query_filter::QueryFilterData::GetInstance()->UpdateVersion(
+                  version);
```
cdesouza-chromium: I don't think we need to move `UpdateVersion` to this block. It can be left inside `OnQueryFilterFileRead`

r0hit0303: Done

components/query_filter/browser/query_filter_component_installer.cc
```diff
@@ -46,16 +45,10 @@ std::string ReadQueryFilterFile(const base::FilePath& path) {
   return contents;
 }
 
+// Reads the |json_data| and updates the component version iff the current rules
+// were updated.
```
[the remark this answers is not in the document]
r0hit0303: I don't think the suggestion is correct.

netzenbot (approved):

cdesouza-chromium (approved):
LGTM

PR #37226, opened 2026-06-12
[VPN 2.0] Implement crash reporting to VPN apps

alekspop26:
The change implements Chromium crash reporting for Agent and Helper apps, similar to what we currently have for Wireguard service on Windows, but cross-platform. For now, no crash uploading will happen - local reporting only. Code is shared between the apps, not duplicated.

Implements part of https://github.com/brave/brave-browser/issues/55909

### DIFF
--- a/components/brave_vpn/app/v2/shared/app_utils.cc
+++ b/components/brave_vpn/app/v2/shared/app_utils.cc
@@ -5,29 +5,64 @@
 
 #include "brave/components/brave_vpn/app/v2/shared/app_utils.h"
 
+#include "base/base_paths.h"
+#include "base/check.h"
+#include "base/environment.h"
 #include "base/logging.h"
 #include "base/logging/logging_settings.h"
+#include "base/path_service.h"
+#include "brave/components/brave_vpn/app/v2/shared/switches.h"
+#include "build/build_config.h"
 
-namespace brave_vpn {
-namespace v2 {
-namespace app_utils {
+#if BUILDFLAG(IS_LINUX)
+#include "base/nix/xdg_util.h"
+#endif
+
+namespace brave_vpn::v2::app_utils {
 namespace {
-inline constexpr char kVpnAppLogFile[] = "log-file";
-}
+inline constexpr base::FilePath::CharType kBraveCompanyName[] =
+    FILE_PATH_LITERAL("BraveSoftware");
+}  // namespace
 
 void InitLogging(const base::CommandLine& command_line) {
   logging::LoggingSettings settings;
   settings.logging_dest =
       logging::LOG_TO_SYSTEM_DEBUG_LOG | logging::LOG_TO_STDERR;
   base::FilePath log_file_path;
-  if (command_line.HasSwitch(kVpnAppLogFile)) {
+  if (command_line.HasSwitch(switches::kVpnAppLogFile)) {
     settings.logging_dest |= logging::LOG_TO_FILE;
-    log_file_path = command_line.GetSwitchValuePath(kVpnAppLogFile);
+    log_file_path = command_line.GetSwitchValuePath(switches::kVpnAppLogFile);
     settings.log_file_path = log_file_path.value().c_str();
   }
   logging::InitLogging(settings);
 }
 
-}  // namespace app_utils
-}  // namespace v2
-}  // namespace brave_vpn
+base::FilePath GetUserDataDir(const std::string& product_name,
+                              bool is_privileged_process) {
+  CHECK(!product_name.empty());
+  base::FilePath result;
+#if BUILDFLAG(IS_WIN)
+  result = base::PathService::CheckedGet(is_privileged_process
+                                             ? base::DIR_COMMON_APP_DATA
+                                             : base::DIR_LOCAL_APP_DATA);
+#elif BUILDFLAG(IS_MAC)
+  result =
+      is_privileged_process
+          ? base::FilePath(FILE_PATH_LITERAL("/Library/Application Support"))
+          : base::PathService::CheckedGet(base::DIR_APP_DATA);
+#elif BUILDFLAG(IS_LINUX)
+  if (is_privileged_process) {
+    result = base::FilePath(FILE_PATH_LITERAL("/var/lib"));
+  } else {
+    auto env = base::Environment::Create();
+    result = base::nix::GetXDGDirectory(
+        env.get(), base::nix::kXdgConfigHomeEnvVar, base::nix::kDotConfigDir);
+  }
+#else
+#error unsupported platform
+#endif
+  return result.Append(kBraveCompanyName)
+      .Append(base::FilePath::FromUTF8Unsafe(product_name));
+}
+
+}  // namespace brave_vpn::v2::app_utils
--- a/components/brave_vpn/app/v2/shared/app_utils.h
+++ b/components/brave_vpn/app/v2/shared/app_utils.h
@@ -6,16 +6,44 @@
 #ifndef BRAVE_COMPONENTS_BRAVE_VPN_APP_V2_SHARED_APP_UTILS_H_
 #define BRAVE_COMPONENTS_BRAVE_VPN_APP_V2_SHARED_APP_UTILS_H_
 
-#include "base/command_line.h"
+#include <string>
 
-namespace brave_vpn {
-namespace v2 {
-namespace app_utils {
+#include "base/command_line.h"
+#include "base/files/file_path.h"
+#include "build/build_config.h"
+
+namespace brave_vpn::v2::app_utils {
 
 void InitLogging(const base::CommandLine& command_line);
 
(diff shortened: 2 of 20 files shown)


components/brave_vpn/app/v2/shared/crash_reporter_client.cc
```diff
@@ -0,0 +1,74 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
```
netzenbot, line 1: nit: this file uses std::move (in the CrashReporterClient constructor) but doesn't include <utility>. Per IWYU, add `#include <utility>` rather than relying on a transitive include. See [CS-001](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-001).

components/brave_vpn/app/v2/shared/crash_reporter_client_win.cc
```diff
@@ -0,0 +1,96 @@
+/* Copyright (c) 2026 The Brave Authors. All rights reserved.
```
netzenbot, line 1: nit: GetProductNameAndVersion uses std::unique_ptr<FileVersionInfo> but the file doesn't include <memory>. Per IWYU, add `#include <memory>`. See [CS-001](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-001).

components/brave_vpn/app/v2/shared/crash_reporter_client.cc
```diff
@@ -0,0 +1,74 @@
...
+
+CrashReporterClient::~CrashReporterClient() = default;
+
+bool CrashReporterClient::IsRunningUnattended() {
+  // Never upload crash reports.
+  return true;
+}
+
+}  // namespace v2
+}  // namespace brave_vpn
```
netzenbot: nit: Prefer `DVLOG(1)` over `VLOG(1)` for retained debug logging. `VLOG` compiles into release builds and bloats the binary, whereas `DVLOG` is stripped from release builds and can still be enabled via `--v=1`. See [CS-067](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-067).

components/brave_vpn/app/v2/shared/app_utils_mac.cc
```diff
@@ -0,0 +1,49 @@
...
+    for (base::FilePath checked_path = e.Next(); !checked_path.empty();
+         checked_path = e.Next()) {
+      if (checked_path.BaseName().Extension() == ".framework") {
+        framework_path = checked_path;
+        break;
+      }
+    }
+  }
+
+  if (!framework_path.empty()) {
```
netzenbot, line 53: nit: Prefer `DVLOG(1)` over `VLOG(1)` here. `DVLOG` is stripped from release builds (avoiding binary bloat) but can still be enabled with `--v=1` for debugging. See [CS-067](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-067).

components/brave_vpn/app/v2/shared/app_utils.h
```diff
@@ -6,13 +6,23 @@
...
 
+#include <string>
+
 #include "base/command_line.h"
+#include "base/files/file_path.h"
+#include "build/build_config.h"
 
 namespace brave_vpn {
 namespace v2 {
 namespace app_utils {
```
netzenbot: nit: GetUserDataDir() has non-obvious contract worth documenting in the header: it returns an empty FilePath when product_name is empty, and the path differs by platform and is_privileged_process. A short comment describing these would help callers. ([✅ Method Documentation Should Describe the Contract](https://github.com/brave/brave-core/tree/master/docs/best-practices/documentation.md#DOC-004))

components/brave_vpn/app/v2/shared/crash_reporter_client.cc
alekspop26: Fixed.

components/brave_vpn/app/v2/shared/crash_reporter_client.cc
alekspop26: Fixed.

components/brave_vpn/app/v2/shared/crash_reporter_client_win.cc
alekspop26: Fixed.

components/brave_vpn/app/v2/shared/app_utils_mac.cc
alekspop26: Fixed.

components/brave_vpn/app/v2/shared/app_utils.h
alekspop26: Fixed.

netzenbot (approved):

components/brave_vpn/app/v2/shared/switches.h
```diff
@@ -0,0 +1,22 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+#ifndef BRAVE_COMPONENTS_BRAVE_VPN_APP_V2_SHARED_SWITCHES_H_
+#define BRAVE_COMPONENTS_BRAVE_VPN_APP_V2_SHARED_SWITCHES_H_
+
+namespace brave_vpn {
+namespace v2 {
+namespace switches {
```
goodov: nit: prefer nested namespaces like `namespace brave_vpn::v2::switches {}` where applicable. it's allowed and makes code bit nicer.

components/brave_vpn/app/v2/shared/app_utils.cc
```diff
@@ -5,29 +5,68 @@
...
+#endif
 
 namespace brave_vpn {
 namespace v2 {
 namespace app_utils {
 namespace {
-inline constexpr char kVpnAppLogFile[] = "log-file";
-}
+inline constexpr base::FilePath::CharType kBraveCompanyName[] =
+    FILE_PATH_LITERAL("BraveSoftware");
```
goodov, line 24: a bit concerned this is hardcoded instead of some known build-time constant. there's no one for this purpose?

components/brave_vpn/app/v2/shared/app_utils.cc
```diff
@@ -5,29 +5,68 @@
...
+  } else {
+    auto env = base::Environment::Create();
+    result = base::nix::GetXDGDirectory(
+        env.get(), base::nix::kXdgConfigHomeEnvVar, base::nix::kDotConfigDir);
+  }
+#else
+#error unsupported platform
+#endif
+  return result.Append(kBraveCompanyName)
+      .Append(base::FilePath::FromUTF8Unsafe(product_name));
```
goodov, line 65: why this is done by hand here? isn't there some shared helper to do that? if not, can you at least write a unit test that will compare both chrome-internal path generation with this version?

components/brave_vpn/app/v2/shared/app_utils.cc
alekspop26: Discussed on Slack. Quick recap: the constant is mainly hardcoded, and exists in different places:
1. On Windows: [redefined](https://github.com/brave/brave-core/blob/master/chromium_src/chrome/install_static/chromium_install_modes.h#L26) in Brave code.
2. On Mac: is read by Chromium from app bundle's CrProductDirName.
3. On Linux: [hardcoded](https://github.com/brave/brave-core/blob/master/chromium_src/chrome/common/chrome_paths_linux.cc#L28) in the data dir path.
Hence, the best approach seems to keep the constant re-defined in the VPN apps code. It also reduces their dependency on the browser branding. While the company name is obviously the same, VPN apps' branding is conceptually different from the Browser's.

components/brave_vpn/app/v2/shared/app_utils.cc
alekspop26: It's already hard-coded in some Brave calls, e.g. [here](https://github.com/brave/brave-core/blob/master/chromium_src/chrome/common/chrome_paths_linux.cc#L28) for Linux. No shared helper exists. I'll add a unit test though -- good idea.

components/brave_vpn/app/v2/shared/switches.h
alekspop26: Fixed.

components/brave_vpn/app/v2/shared/app_utils.cc
alekspop26: Added the unit tests. Since on mac, branding tests would require bundling and a proper value of `CrProductDirName`, I excluded Mac, but on Windows and Linux, the test will work.

goodov (approved):

PR #37230, opened 2026-06-13
Transition tests to using mv3 extensions.

mkarolin:
Resolves

This is an AI-assisted change.

### DIFF
--- a/browser/ephemeral_storage/blob_url_browsertest.cc
+++ b/browser/ephemeral_storage/blob_url_browsertest.cc
@@ -258,13 +258,16 @@ IN_PROC_BROWSER_TEST_F(BlobUrlPartitionEnabledBrowserTest,
       RegisterBlobs(a_site_ephemeral_storage_url_);
 
   extensions::TestExtensionDir test_extension_dir;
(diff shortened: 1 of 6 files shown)


browser/ephemeral_storage/blob_url_browsertest.cc
```diff
@@ -258,13 +258,16 @@ IN_PROC_BROWSER_TEST_F(BlobUrlPartitionEnabledBrowserTest,
       RegisterBlobs(a_site_ephemeral_storage_url_);
```
netzenbot, line 258: nit: avoid the positional reference "below" ([CS-024](https://github.com/brave/brave-core/tree/master/docs/best-practices/coding-standards.md#CS-024)). Reference the operation explicitly, e.g. "the cross-origin blob fetch in the loop", so the comment stays correct if code is inserted.

mkarolin: Addressed in [b93afd8](https://github.com/brave/brave-core/pull/37230/commits/b93afd8ca8b3bb06313473de81ec90610b47fd6a)

boocmp (approved):

